PR #227, opened 2026-06-04
fix(reader-activation): clear local data on account switch (NPPM-2899)

jason10lee:
### All Submissions:

Follow-up to NPPM-2721 (clear localStorage namespace on logout). That fix cleared the visitor's local reader data on **logout**, but not when a session moves directly from one authenticated reader to another with no intervening anonymous pageload (e.g. a shared device where reader B logs in over reader A's still-warm session). Reader B would then inherit reader A's stored activity, pageviews, and donor status from `localStorage`.

This adds an **account-switch** trigger: when the server confirms a *different authenticated reader* than the one stored locally, the reader namespace is cleared and reseeded for the new reader. It's gated so a previously-**anonymous** visitor — including one who only submitted an email to a newsletter prompt (stored but never authenticated) — keeps their accumulated engagement when they log in; **only a true authenticated→authenticated switch wipes**. NPPM-2721's logout, cached-page, and orphaned-email behavior is unchanged. The clear decision is factored into a pure, unit-tested `shouldClearReaderData()` predicate covering all three triggers.

Closes [NPPM-2899: RAS: clear local reader data on account switch (A→B) without an anonymous pageload](https://linear.app/a8c/issue/NPPM-2899/).

1. Enable Reader Activation. Log in as reader **A**, browse a few posts (accumulates `activity`/`pageviews`; set `is_donor` if convenient).
2. In the same browser, end up authenticated as a **different** reader **B** without an anonymous pageload in between (e.g. simulate by logging in as B over A, or by serving a config with `authenticated_email` = B's email).
3. On B's authenticated pageload, confirm B does **not** inherit A's data: `np_reader_<blog_id>_activity` / `_pageviews` / `_is_donor` are gone and `np_reader_<blog_id>_reader` is reseeded to B.
4. **A→A** (same reader refresh/re-login): activity is **preserved** (no wipe).
5. **Carryover preserved:** anonymous browsing then login → activity carries over; a newsletter-email lead (unauthenticated stored email) logging in under a *different* email → activity preserved.
6. **NPPM-2721 regressions:** logout → namespace cleared; a cached *anonymous* page served to a still-logged-in browser → **not** cleared.
7. `cd plugins/newspack-plugin && npm test` → reader-activation suite green, incl. the new `shouldClearReaderData` truth-table (11 cases) and the A→B / A→A / lead integration tests.

### Other information:

* [x] Have you successfully ran tests with your changes locally? (106 green, lint clean)


plugins/newspack-plugin/src/reader-activation/index.js
```diff
@@ -710,27 +764,18 @@ function init() {
...
+	// NPPM-2899).
+	if (
+		shouldClearReaderData( {
+			authenticatedEmail: data?.authenticated_email,
+			initialEmail,
+			storedReader: currentReader,
+			hasAuthReaderCookie: !! getCookie( 'np_auth_reader' ),
+		} )
+	) {
 		clearReaderStore();
```
dkoo, line 798: **Suggestion:** Account-switch clear also drops reader B's own server data for one pageload — On an A→B switch, `clearReaderStore()` runs `store.clear()`, which (in `store.js:402`) sets `window.newspack_reader_data.items = {}`. On a normal authenticated pageload the server localizes B's reader items — including read-only keys like `is_donor` / `is_newsletter_subscriber` (`class-reader-data.php:152`) — and `init()`'s trailing `store.rehydrate()` (`index.js:809`) consumes them to populate the store. Because `clear()` empties that in-memory snapshot first, the trailing rehydrate is a no-op for B. The only other rehydrate path (`EVENTS.session` in `store.js:314`) fires via `hydrateSession()` after an *in-page* auth transition, not on a fresh authenticated pageload — so B's read-only flags are missing from the store for the switch pageload and only reappear on B's next navigation (when items are re-localized and, stored==B now, no clear fires). The cross-reader leakage this PR targets is correctly fixed; this is the inverse gap — B briefly loses its *own* state. A fix on this path (rehydrate B's server snapshot after the clear) would close it.

plugins/newspack-plugin/src/reader-activation/index.test.js
```diff
@@ -281,4 +281,177 @@ describe( 'init() post-logout clearing (NPPM-2721)', () => {
 		expect( reader.authenticated ).toBe( false );
 		expect( reader.email ).toBeUndefined();
 	} );
+
+	it( 'authenticated switch A→B wipes the prior reader data', () => {
```
dkoo, line 300: **Suggestion:** No bootInit integration test for account-switch with np_auth_reader cookie present — The predicate-level suite covers account-switch both with `hasAuthReaderCookie: true` (line 337) and `false` (line 425), but the `bootInit()` integration case for A→B (line 285) doesn't seed an `np_auth_reader` cookie. Since "the account-switch trigger deliberately ignores `np_auth_reader`" is the PR's headline safety claim, an end-to-end `bootInit` case with `cookies: { np_auth_reader: 'b@example.com' }` + config `authenticated_email: 'b@example.com'` would prove `init()` wires the cookie through without it suppressing the clear — closing the gap between the pure predicate and the real call site. Not blocking; the predicate is the load-bearing part and it is tested.

plugins/newspack-plugin/src/reader-activation/index.js
```diff
@@ -701,6 +701,60 @@ function handlePush( ...args ) {
...
+ *   - account switch (NPPM-2899): a previously-AUTHENTICATED reader is replaced by a
+ *     server-confirmed DIFFERENT reader. Gated on storedReader.authenticated === true
+ *     so an unauthenticated stored email (e.g. a newsletter lead) logging in under a
+ *     different email keeps its anonymous carryover. Keys on authenticated_email
+ *     (server-confirmed), never the intention cookie.
+ *   - logout (NPPM-2721): server anonymous AND a prior session was authenticated.
+ *   - orphaned (NPPM-2721): server anonymous AND a stale stored email != intention.
+ *
+ * @param {Object}  args
+ * @param {string}  args.authenticatedEmail  newspack_ras_config.authenticated_email ('' when anonymous).
```
dkoo, line 725: **Nit:** Two cookie names are easy to conflate on a future read — The predicate correctly distinguishes `np_auth_intention` (identity, via `initialEmail`) from `np_auth_reader` (the cached-page guard, via `hasAuthReaderCookie`). The JSDoc and comments are self-consistent, but the two `np_auth_*` cookies play very different roles and are easy to mix up later — a one-word reminder at each use site ("intention cookie = identity", "reader cookie = cache guard") would help the next reader. No change required.

plugins/newspack-plugin/src/reader-activation/index.js
```diff
@@ -701,6 +701,60 @@ function handlePush( ...args ) {
...
+ * @param {string}  args.initialEmail        authenticated_email || np_auth_intention cookie.
+ * @param {Object}  args.storedReader        Persisted reader ({ email, authenticated }).
+ * @param {boolean} args.hasAuthReaderCookie Whether the np_auth_reader cookie is present.
+ * @return {boolean} Whether to clear the namespace.
+ */
+export function shouldClearReaderData( { authenticatedEmail, initialEmail, storedReader, hasAuthReaderCookie } ) {
```
dkoo, line 731: **Nit:** New named export widens the module's importable contract — Exporting `shouldClearReaderData` is fine — it's a pure predicate, tree-shakeable, and not attached to `window.newspackReaderActivation`, so no new third-party-reachable destructive surface. The only consideration: a named export from this entry becomes part of the module's importable contract for other bundles. If the intent is test-only, an `@access private` / "exported for testing" line in the JSDoc would discourage other code from depending on it as a supported API. Purely optional.

plugins/newspack-plugin/src/reader-activation/index.js
```diff
@@ -701,6 +701,60 @@ function handlePush( ...args ) {
...
+	// switch from such a (non-occurring) cached-authenticated response, so gating on
+	// it would only risk suppressing legitimate A→B clears.
+	if ( storedReader?.authenticated === true && authed && stored && authed !== stored ) {
+		return true;
+	}
+
+	// Remaining triggers require the server to report anonymous. A present
+	// np_auth_reader cookie means "not anonymous" (NPPM-2721 cached-page guard).
+	// Use the normalized `authed` so a whitespace-only email is handled consistently.
+	const serverSaysAnonymous = ! authed && ! hasAuthReaderCookie;
```
dkoo, line 760: **Nit:** Whitespace-only authenticated_email now normalizes to anonymous (behavioral delta) — `serverSaysAnonymous` now derives from the normalized `authed` (`norm(authenticatedEmail)`) rather than the raw `data?.authenticated_email`. A degenerate whitespace-only `authenticated_email` (`'   '`) flips from "not anonymous" (pre-fix) to "anonymous", which could newly trigger the logout/orphaned clear. The server only ever sends `''` or a real email, so this is a non-occurring input and the normalization is intentional/consistent — just a genuine delta from NPPM-2721 rather than a pure refactor, worth a one-line note if not already covered.

plugins/newspack-plugin/src/reader-activation/index.js
dkoo, line 798: **Nit:** EVENTS.reader doesn't surface the intermediate anonymous reseed on a switch — On A→B, `clear()` emits `EVENTS.data` per wiped key (and for the reseeded anonymous reader), then `init()` emits a single `EVENTS.reader` with the *new* B identity. Consumers listening on `EVENTS.reader` therefore see the transition go straight to B without observing the reset step (the anonymous reseed is only on `EVENTS.data`). This matches the existing logout path so it's consistent — but account-switch is the first case where a *different authenticated* identity replaces another within one `init()`. Worth confirming no analytics/segmentation handler keyed on `EVENTS.reader` needs to observe the reset to drop per-reader derived caches; if any do, they'd need to also watch `EVENTS.data` for the `reader` key.

dkoo (approved):
Passes manual testing. The core fix is correct, well-reasoned, and unusually well-tested — the A→B isolation goal is genuinely met (B no longer inherits A's activity/donor state) with no false-positive data-loss path, the predicate is cleanly pure, and `init()` re-reads after the clear so the stale email can't leak back into the reseed. One functional gap is worth a look before merge (non-blocking, see inline): on the account-switch pageload the clear also discards reader B's *own* server-localized read-only data, which the trailing `store.rehydrate()` would normally hydrate, so B can briefly read as a non-donor to prompt segmentation until the next navigation. The remaining notes are nits.

jason10lee:
Thanks, @dkoo!  Will review the suggestion and nits--the suggestion looks doable in this PR on a quick look!

plugins/newspack-plugin/src/reader-activation/index.js
jason10lee: Addressed in 2d98600. Captured the current reader's server snapshot (`newspack_reader_data.items`) before `clearReaderStore()` and restore it on the authenticated switch path (`if ( authenticated && serverReaderItems )`), so the trailing `store.rehydrate()` repopulates reader B's own read-only flags instead of no-op'ing on the emptied cache. Scoped to the switch case — logout/orphaned clears stay anonymous and intentionally empty. Added an integration test that seeds B's server items and asserts `is_donor`/`is_newsletter_subscriber` are restored while reader A's writable carryover stays wiped.

plugins/newspack-plugin/src/reader-activation/index.test.js
jason10lee: Addressed in 2d98600. Added a `bootInit` A→B integration case with `cookies: { np_auth_reader: 'b@example.com' }` + `authenticated_email: 'b@example.com'`, asserting the clear still fires (A's data wiped, reader becomes B). Confirmed the test is meaningful by temporarily making the switch branch consult the cookie — the new test went red — then reverted.

plugins/newspack-plugin/src/reader-activation/index.js
jason10lee: Addressed in 2d98600. Added a one-line reminder at each use site: `np_auth_intention = identity` on the `initialEmail` line and `np_auth_reader = cache guard` on the `hasAuthReaderCookie` line, plus the same distinction inline in the predicate's `@param` docs.

plugins/newspack-plugin/src/reader-activation/index.js
jason10lee: Addressed in 2d98600. Added `@access private` to `shouldClearReaderData`'s JSDoc with a note that it's exported only for unit testing and shouldn't be depended on from other bundles.

plugins/newspack-plugin/src/reader-activation/index.js
jason10lee: Addressed in 2d98600. Added a one-line note documenting the intentional delta from NPPM-2721: deriving `serverSaysAnonymous` from the normalized `authed` makes a whitespace-only `authenticated_email` normalize to anonymous, but the server only ever sends `''` or a real email so the input is non-occurring.

plugins/newspack-plugin/src/reader-activation/index.js
jason10lee: Investigated — confirmed safe, no behavior change needed (doc note added in 2d98600). All four `EVENTS.reader` consumers (account-link labels in reader-activation-auth, the gifted-post reload in content-banner, and reader-registration's `endLoginFlow`) react to the *current* reader's `authenticated` state, not to a reset transition; none keeps a per-reader derived cache that must observe the intermediate anonymous reseed within one `init()`. `clear()` already emits a per-key `EVENTS.data` (`value: undefined`) for every wiped public key, which is the channel any cache-invalidation consumer keyed on `detail.key` would use. Documented this at the `EVENTS.reader` emit.

jason10lee:
Thanks for the thorough review, @dkoo! All addressed in 2d98600:

**Suggestion — the account-switch clear also dropped reader B's *own* server data for one pageload:** Fixed. `clearReaderStore()` empties `newspack_reader_data.items`, so the trailing `store.rehydrate()` was a no-op for B. Now I capture B's server snapshot before the clear and restore it on the authenticated switch path, so rehydrate repopulates B's own read-only flags — no more brief non-donor read for prompt segmentation. Scoped to the switch case (logout/orphaned clears stay anonymous and intentionally empty), and covered by a new integration test that asserts B's `is_donor`/`is_newsletter_subscriber` are restored while reader A's writable carryover stays wiped.

**Nits:**
- **bootInit A→B + `np_auth_reader`:** Added the end-to-end integration case proving `init()` wires the cookie through without it suppressing the clear (verified meaningful — it goes red if the switch branch consults the cookie).
- **Named export:** Marked `shouldClearReaderData` `@access private` (exported for unit testing only).
- **Cookie-name conflation:** Added reminders at each use site — `np_auth_intention` = identity vs `np_auth_reader` = cache guard.
- **Whitespace-only `authenticated_email`:** Documented the intentional normalization delta from NPPM-2721 (non-occurring input; server sends `''` or a real email).
- **`EVENTS.reader` on a switch:** Investigated — confirmed safe. All four consumers react to the current reader's `authenticated` state, not a reset transition; per-key wipes are already surfaced on `EVENTS.data`. Documented at the emit.

Replied inline on each thread with specifics. 7 reader-activation suites green (108 tests); lint clean.

PR #253, opened 2026-06-09
ci(workflows): catch & auto-fix Dependabot pnpm.overrides drift

miguelpeixe:
### All Submissions:

Follow-up to #252. Adds two independent layers of defense against the lockfile/`package.json` `pnpm.overrides` drift that broke `main` in [run 27210804365](https://github.com/Automattic/newspack-workspace/actions/runs/27210804365):

**1. CI safety net — `.github/workflows/ci.yml`**

The `changed-packages` detector currently skips `Install deps` when no source package is affected. That misses a class of PRs that only modify root dependency files (the Dependabot lockfile-only case, but also any manual root edit). The detector now additionally checks `git diff` for changes to `pnpm-lock.yaml`, root `package.json`, or `pnpm-workspace.yaml`; if any are touched, `any=true` so `Install deps` runs. The JS / PHPCS / PHPUnit matrix jobs continue to skip cleanly because their matrices stay empty.

**2. Dependabot auto-fix — `.github/workflows/dependabot-sync-overrides.yml` (new)**

Runs on every Dependabot PR via `pull_request_target`. When the PR diff touches `pnpm-lock.yaml`, it compares the lockfile's `overrides:` block against `package.json`'s `pnpm.overrides` and, for any key already present in `package.json` whose value differs, updates `package.json` to match. Commits and pushes via the same `GH_TOKEN` PAT pattern used by `dependabot-branch-auto-update.yml`, so CI re-fires against the synced state and the existing auto-merge can complete.

Safety:

* Gated on `github.actor == 'dependabot[bot]'`.
* `pull_request_target` runs the workflow definition from `main`, not from the PR, so a hostile PR cannot modify the workflow that holds the PAT.
* The Python sync only updates *existing* keys in `package.json.pnpm.overrides`; it never adds or removes entries. So a hostile lockfile diff cannot inject arbitrary overrides into `package.json`.
* Commit author is `dependabot[bot]` so the existing `auto-merge.yml` (which checks PR author) continues to treat the PR as auto-mergeable. The repo does not dismiss stale reviews, so the prior approval survives the sync commit.

1. Confirm both workflow files parse (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml')); yaml.safe_load(open('.github/workflows/dependabot-sync-overrides.yml'))"`).
2. After merge, the next Dependabot PR that bumps a pinned override exercises the auto-fix path end-to-end. Expected behaviour: Dependabot opens the PR, sync workflow runs, detects the `package.json` mismatch, pushes a follow-up commit, CI re-runs and passes, auto-merge completes.
3. (Out-of-band) Manually open a no-op PR that only edits `pnpm-lock.yaml` (e.g. add then immediately revert a whitespace change) and confirm `Install deps` runs in CI on that PR rather than being skipped.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — N/A, CI workflow change. Verified locally that the sync logic correctly classifies the three current overrides as matching against the post-#252 state.
* [x] Have you successfully run tests with your changes locally? — YAML validated; sync script dry-run executed against current repo state.

What this does **not** cover:

* Composer overrides — out of scope; Dependabot's composer support is separate.
* Overrides Dependabot *adds* that don't exist in `package.json` — by design we don't touch new keys. If we ever start tracking a new transitive override, a human still adds it to `package.json` first.


.github/workflows/dependabot-sync-overrides.yml
```diff
@@ -0,0 +1,170 @@
...
+    # a synchronize event triggered by a human collaborator amending the PR,
+    # github.actor would be the human and the sync would skip exactly when a
+    # manual edit most needs reconciling. PR author stays dependabot[bot].
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      # PAT, not GITHUB_TOKEN: a push from GITHUB_TOKEN does not re-fire CI on
+      # the PR, which would leave required checks pending forever. Same reason
+      # as dependabot-branch-auto-update.yml.
+      - name: Checkout PR branch
```
adekbadek, line 58: **Suggestion:** Pin the “never execute tree content” invariant with a comment — The security of this `pull_request_target` job rests on a single invariant: it reads the checked-out PR tree only as **data** (git plumbing + `yaml.safe_load`/`json.load`), and never *executes* anything from it. That's what neutralizes the standard `pull_request_target` RCE vector — there's no `pnpm install`, build, or lifecycle-script run anywhere in the privileged job. Because anyone with write access can push to a Dependabot branch, that invariant is load-bearing. A future edit that adds an innocent-looking `pnpm install` or feeds lockfile data into a shell command would silently reopen a remote-code-execution path against the PAT. Worth a one-line comment next to the checkout step stating the constraint, so it survives future maintenance.

.github/workflows/dependabot-sync-overrides.yml
```diff
@@ -0,0 +1,170 @@
...
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as f:
+              f.write(flag + '\n')
+          PY
+
+      - name: Commit and push
+        if: steps.sync.outputs.changed == 'true'
+        env:
+          # Keep this aligned with SYNC_COMMIT_SUBJECT in the loop-break gate
+          # above; the gate matches HEAD's subject against this exact string.
+          SYNC_COMMIT_SUBJECT: 'build(deps): sync package.json pnpm.overrides with lockfile'
```
adekbadek: **Nit:** `SYNC_COMMIT_SUBJECT` duplicated across two steps — The exact commit-subject string lives in two places (the loop-break gate at line 56 and the commit step at line 161), with a comment asking the maintainer to keep them aligned by hand. A job- or workflow-level `env:` key would make it single-source and remove the drift risk that the comment is there to guard against. The inline comments mitigate it, so this is a readability nit, not a defect.

adekbadek (approved):
All good, one suggestion

.github/workflows/dependabot-sync-overrides.yml
miguelpeixe: Addressed in e342930 — added a `SECURITY INVARIANT` comment block above the checkout step spelling out that this `pull_request_target` job reads the PR tree as **data only** (git plumbing + `yaml.safe_load`/`json.load`) and must never execute anything from it (no `pnpm install`, build, or lifecycle scripts). That pins the load-bearing constraint so a future edit can't silently reopen the RCE path against the PAT.

.github/workflows/dependabot-sync-overrides.yml
miguelpeixe: Addressed in e342930 — hoisted `SYNC_COMMIT_SUBJECT` to a single job-level `env:` key (inherited by every step's run shell) and removed both step-level copies plus the hand-alignment comment. The loop-break gate and the commit step now reference the same single source, so they can't drift.

PR #258, opened 2026-06-09
fix(audience): unblock the ESP connect flow in Integrations

claudiulodro:
### All Submissions:

On a fresh Newspack install it was impossible to connect the Newsletter ESP integration from Audience → Integrations. The card said "Connect" until a master list was selected, but the screen for selecting a master list was unreachable while the card was in the "Connect" state. A dead end.

<details>
<summary>Claudiu's original analysis of the dead end</summary>

> The label logic (settings-section.js:81): the button says "Connect" when is_set_up is false, "Enable" when it's true (and once enabled, the card flips to "Configure" via CardFeature).
>
> The catch is what is_set_up means for the ESP integration (class-esp.php:53). It requires two conditions:
>
> A newsletter provider is connected in Newspack Newsletters (is_service_provider_configured() — API key set, etc.), and
>
> A master list ID is stored in the integration settings (mailchimp_audience_id, active_campaign_master_list, or constant_contact_list_id).
>
> So merely connecting a provider does not change the button — it keeps saying "Connect" until a master list/audience is also chosen. And this creates a dead-end loop in the UI:
>
> When is_set_up is false, needsSetup is true, so clicking "Connect" navigates to the Newspack Newsletters settings page (setup_url) — even if the provider is already connected there and the only missing piece is the master list.
>
> The master list picker lives in the integration's own configure view (#/settings/esp), but that view is unreachable from the card while it's in the "Connect" state: the configure button and the "more" menu only render when the integration is enabled, and the enable action is also hijacked to goToSetup while needsSetup is true.
>
> So a user who connects Mailchimp and comes back to the Integrations screen sees "Connect", clicks it, lands back on the Newsletters settings page, and has no path from this screen to select the audience that would complete setup.

</details>

#### The reworked flow

The card's action is now derived from which prerequisite is actually missing, rather than from one conflated `is_set_up` check:

| Card state | Action | What it does |
| --- | --- | --- |
| Newsletters plugin inactive | **Activate** | Activates the plugin. Holds a minimum 2s busy state so the action registers. |
| No provider connected | **Connect** | Real handoff to the Newsletters settings screen, with a return banner. |
| Provider connected, required settings missing | **Enable** | Opens a modal collecting the missing required fields (the audience), then saves and enables in one step. |
| Provider is Manual / Other | **Change provider** | Red "Requires an API-based ESP" badge. Handoff to switch provider. |
| Enabled | **Configure** | The settings view, plus Logs and Disable. |

"Finish set up" is gone. The modal replaces it, so selecting an audience and enabling is one action rather than two screens.

https://github.com/user-attachments/assets/1aa48614-034e-4954-becf-727ea359f159

#### Answering the architecture question from the original description

> this change introduces a new `is_connected` state, so the class has `is_set_up` and `is_connected`. We can probably avoid this if you hate it, or we can keep it if you think it's an important distinction.

Keeping it. The distinction is exactly what fixes the bug. `is_connected` answers "is there a usable provider", `is_set_up` answers "is this integration ready to sync" (connected **and** a master list stored). Collapsing them back into one flag is what made "Connect" unable to tell "no provider" apart from "provider but no audience", which is the dead end. The two states now route to two different places.

#### Also in this PR

- **`Change provider` instead of `Connect` when the provider is Manual.** The site *is* connected, just to a provider with no contact-sync API, so "Connect" was false. The label is supplied by the integration itself (`get_unsupported_action_label()`), so no ESP-specific copy lives in the shared card loop.
- **The card grid no longer unmounts on the post-activation refetch.** `fetchSettings` swapped the whole grid for a "Loading…" line on every call, which is right on first mount and wrong on a refetch, so activating made the page jump. It now opts out of the loading gate and returns its promise, so the card holds "Activating…" until fresh data lands. Pre-existing on `main`; the 2s busy window made it obvious.
- **Unsaved changes are guarded in the Configure view.** Editing a field and navigating away used to lose it silently. Uses the existing `useUnsavedChangesDialog` hook, and discarding now genuinely discards.
- **A failed settings save tells you.** `handleSave` had no `.catch`, so a failed save produced no feedback and an unhandled rejection. It now surfaces an error notice and preserves your unsaved edits so you can retry.

#### Provider brand marks

Each ESP now shows its own brand mark instead of a generic envelope — on the Audience → Integrations card once a provider is selected, and on the Newsletters settings provider cards. Both render through a single new shared component, `IntegrationIcon` (`packages/components`), which owns the mark and its badge background in one place: adding a provider is one icon file, one map entry, and one background rule. Manual shows an `@` on a dark badge (it has no brand logo); the generic envelope shows only when no provider is selected. `CardFeature` and the core card gained a raw-icon slot so the self-contained badge isn't double-wrapped.

You need a site with Newspack Newsletters installed and an API-based ESP account (Mailchimp is easiest).

> [!IMPORTANT]
> **Fresh sites auto-enable the ESP integration**, so the card will show "Enabled" / "Configure" rather than "Connect". To see the connect flow, hit **Disable** on the card first. This is pre-existing framework behaviour, not something this PR changes. Note also that an integration that is enabled but not connected is served by **Configure**, which routes through the same handoff.

1. Deactivate Newspack Newsletters. The card shows a disabled-looking state with a "Requires Newspack Newsletters" badge and an **Activate** button.
2. Click **Activate**. The button shows "Activating…" for ~2s. The card should stay put throughout, with no flash or page jump, then land on **Connect**.
3. Click **Connect**. You land on Newsletters settings with a "Return to Integrations after completing configuration" banner.
4. Select Mailchimp, paste an API key, save. Click **Back to Integrations**.
5. The card now shows **Enable**. Click it. A modal appears asking for the audience (the required field that is still missing).
6. Pick an audience and click **Enable**. The modal closes, a "Newsletter ESP enabled." snackbar appears, and the card flips to Enabled / Configure.

Then the unsupported path:

7. Go to Newsletters settings and switch the provider to **Manual / Other**. Back on Integrations, the card is muted with a red **Requires an API-based ESP** badge and a **Change provider** button that routes back to the provider picker. The card's icon is the Manual `@` mark on a dark badge; connected API providers (Mailchimp/ActiveCampaign/Constant Contact) show their brand logo.

And the unsaved-changes guard:

8. With the integration enabled, open **Configure**, edit a field, and try to navigate away. You get a "You have unsaved changes that will be lost" dialog. Discarding actually discards: return to the view and the edit is gone.

### Known follow-ups (not in this PR)

- An integration that is **enabled and then becomes unsupported** (ESP on Mailchimp, provider switched to Manual) loses its More menu, so no Disable and no Logs from that card. `CardFeature` gates the dropdown on `isConfigureState = enabled && ! requirements`. Nobody is stranded since "Change provider" is a way forward, but fixing it touches every card.
- `pendingChanges` does double duty as both the view-scoped live draft and the request-scoped in-flight/retry buffer. Splitting those is the principled cleanup behind the guard work here.
- The wizard notices store appends without deduping by id, so call sites that use a stable id expecting "replace" get "stack".

### Other information:

Tests: JS 41 suites / 372 tests. PHP `Test_ESP` (21) and `Test_Integrations` (62) green.


claudiulodro:
@thomasguillot "volunteered" to address this issue, so I've switched this to Needs Design Review. Feel free to close this PR if you'd rather start from scratch. Thanks!

thomasguillot:
Two follow-ups came out of the review on this PR. Both are pre-existing and neither blocks merging here, but they are worth doing next. Picking them up in a PR stacked on this branch.

### 1. `pendingChanges` does double duty

`pendingChanges` in `views/integrations/index.js` is simultaneously two different things:

- the **view-scoped live draft** (what you are typing in the Configure view), and
- the **request-scoped in-flight / retry buffer** (the payload a save is currently attempting, and the only surviving copy if that save fails).

Those two have genuinely different lifecycles, and every defect found while adding the unsaved-changes guard landed exactly where they disagree:

- the draft was never cleared on discard, so "Discard changes" discarded nothing and the value came back on revisit with Save enabled
- clearing it on unmount then destroyed the retry buffer when a save failed
- guarding that with a pre-indexed ref read the wrong integration's flag on a route change

Each of those is a symptom of one cause. The shape that removes the class: keep the draft local to `ConfigureView`, so unmount-discards is structural and the cleanup effect disappears entirely, and let the request layer own the in-flight payload plus its failure/retry state.

Worth flagging for whoever picks this up: the obvious version of this refactor, "just move `pendingChanges` into `ConfigureView`", is wrong. It fixes the first and third symptoms and makes the second one structurally unfixable, because a draft that dies at unmount cannot survive a failed save at all. The split has to be a split, not a move.

### 2. An enabled integration that becomes unsupported loses its More menu

Enable the ESP on Mailchimp, then switch the provider to Manual. The card correctly shows the red "Requires an API-based ESP" badge and a "Change provider" action, but the More menu disappears, so there is no Disable and no Logs from that card.

`CardFeature` renders the dropdown only when `isConfigureState && !! moreControls?.length`, and `isConfigureState = enabled && ! requirements` (`packages/components/src/card-feature/index.tsx:172`, `:118`). An unsupported reason always populates `requirements`, so the dropdown is suppressed for exactly the integrations that are enabled but non-functional.

Nobody is stranded, since "Change provider" is a real way forward, and the gating is untouched by this PR. But an enabled integration you cannot disable from its own card is the wrong end state, and fixing the gating affects every card in the product rather than just this one, so it wants its own blast radius.

thomasguillot:
Two follow-up PRs stack on top of this branch, both from issues surfaced while working in this area:

- **#632** `refactor(audience): split the configure-view draft from its retry buffer` (base `fix/newsletter-setup`). Splits the overloaded `pendingChanges` into a ConfigureView-local draft and a parent-owned retry buffer, making the discard and route-change fixes from this PR structural rather than guarded. Also fixes two in-flight-save edges found in review (edits lost if you keep typing during a save; a phantom "unsaved changes" state after a save completes across a remount).
- **#641** `fix(components): keep the more menu on enabled unsupported cards` (base `pr-a/split-pending-changes`, stacked on #632). An enabled-but-unsupported card (ESP after switching the provider to Manual) was losing its More menu, so no Disable or Logs. Gates the CardFeature dropdown on operability instead of `isConfigureState`, so it affects every card and got wider QA.

Stack order: #258 → #632 → #641. Merge bottom-up; each base retargets as the one below it lands.

packages/components/src/integration-icons/index.js
```diff
@@ -1,6 +1,29 @@
...
 export { default as wisepops } from './wisepops';
+
+// ESP provider brand marks keyed by the slug the backend reports
+// (`Newspack_Newsletters::service_provider()`). Rendered via IntegrationIcon.
+export const espProviderIcons = {
+	active_campaign: activeCampaign,
+	mailchimp,
+	constant_contact: constantContact,
+	manual: atSymbol,
+};
```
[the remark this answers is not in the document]
thomasguillot: `atSymbol` from `@wordpress/icons` is itself a renderable `<SVG fill="currentColor">` element, same shape as the local brand marks, so `IntegrationIcon` renders it directly and the Manual `@` mark shows correctly. No `<Icon>` wrapper needed.

packages/components/src/integration-icon/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * background. A new provider needs an icon file, an `espProviderIcons` entry,
+ * and a background rule in this component's stylesheet.
+ */
+const IntegrationIcon = ( { provider, className }: { provider: string; className?: string } ) => {
+	const node = Object.prototype.hasOwnProperty.call( espProviderIcons, provider ) ? espProviderIcons[ provider ] : null;
+	if ( ! node ) {
+		return null;
+	}
+	return (
+		<span className={ classnames( 'newspack-integration-icon', `newspack-integration-icon--${ provider.replace( /_/g, '-' ) }`, className ) }>
```
miguelpeixe: **[nit] Decorative brand icon lacks `aria-hidden`**

The integration/provider brand icon renders without `aria-hidden="true"`. In every usage it sits beside the visible integration name (the card title), so it duplicates that text for assistive tech. Adding `aria-hidden="true"` (and `focusable="false"`) on the icon wrapper is the standard decorative-icon treatment.

plugins/newspack-plugin/src/wizards/audience/views/integrations/settings-section.js
```diff
@@ -54,67 +58,139 @@ export const SettingsSection = ( { integrations, loading, activating = {}, onTog
...
+									<CardFeature
+										key={ id }
+										title={ name }
+										description={ description }
+										icon={ cardIcon }
+										enabled={ isEnabled }
+										requirements={ requirements }
+										requirementsActionable={ canActivate || canFixUnsupported }
+										enableLabel={ enableLabel }
+										busy={ isActivating }
```
miguelpeixe: **[nit] No busy feedback on Enable / Disable**

`busy` on the card is wired only to `isActivating`. The `toggling` state exists in `index.js` and is passed through `sharedProps`, but `SettingsSection` never destructures it — so a direct **Enable** / **Disable** click gives no in-flight/busy feedback on the card (unlike Activate). Harmless (enable/disable are idempotent server-side and the list refetches on success) and pre-existing, but the new `busy` infrastructure this PR adds could naturally cover it.

plugins/newspack-plugin/src/wizards/audience/views/integrations/enable-modal.js
```diff
@@ -0,0 +1,113 @@
...
+				<VStack spacing={ 6 } className="newspack-integration-enable-modal__content">
+					{ error && <Notice isError noticeText={ error } /> }
+					{ missingFields.map( field => (
+						<SettingsField
+							key={ field.key }
+							field={ field }
+							value={ getValue( field ) }
+							onChange={ value => setValues( prev => ( { ...prev, [ field.key ]: value } ) ) }
+						/>
+					) ) }
```
miguelpeixe, line 103: **[nit] Benign modal state-update-after-unmount ordering**

On a successful modal enable, the parent's `onEnable` resolves and runs `setEnablingId( null )` (unmounting the modal) in one microtask, then the modal's own `.finally( () => setEnabling( false ) )` runs in the next — i.e. after the component is told to unmount. Benign on the WordPress-bundled React 18 (it no-ops state updates on unmounted components), but it's an ordering smell worth tidying.

miguelpeixe (approved):
Approving — no blockers. A high-quality foundational fix that holds up on every axis that matters: the `is_connected` / `is_set_up` split genuinely breaks the fresh-install dead-end without introducing a new one (traced across every card state), the ESP `is_set_up()` semantics are unchanged for the real sync/pull consumers (no data-loss regression), the EnableModal handles partial save/enable failure cleanly and can't submit an incomplete payload, and the security surface is tight — a server-derived, same-site-validated handoff and an allowlisted, `activate_plugins`-gated activation path. Test coverage (PHP + JS) is thorough.

A few notes, none blocking:

- **`is_connected()` has no direct unit test.** It's new and load-bearing (it drives the Connect-vs-Enable branch) but is only exercised indirectly. A direct stored-state test (toggling the provider option / credentials) would close the loop symmetrically with the existing `test_esp_is_set_up_*` cases.
- **(Pre-existing — noticed in the enable handler this PR touches.)** The `/settings/{id}/enabled` route declares no `args`, so `enabled` isn't type-coerced and the handler relies on PHP truthiness. The first-party client always sends a JSON boolean, so it's fine today, but a non-JSON caller sending `"false"` would be truthy and *enable* instead of disable. `'enabled' => [ 'type' => 'boolean' ]` (or `rest_sanitize_boolean`) would make the contract explicit. Low-risk (the route is `manage_options`-gated) and out of this PR's scope — just flagging while nearby.
- **The new clickable card header** (`headerIsButton`) has no explicit `:focus-visible` style; it stays keyboard-accessible via the UA outline, but an explicit ring consistent with the card's hover/active state would be more on-brand.

Three small inline nits below.

PR #264, opened 2026-06-09
fix(checkout): sync repeat trial pricing in modal checkout

rbcorrales:
### All submissions:

* [x] Have you followed the Newspack contributing guidelines?
* [x] Have you checked to ensure there are not other open Pull Requests for the same update or change?

This fixes checkout pricing for anonymous shoppers whose email addresses belong to users who have already used a subscription trial.

Previously, the modal checkout flow could validate billing details without resolving the email to the matching existing user for trial eligibility. The payment step could still show trial pricing and a zero-due-today amount, while the completed order charged the subscription amount.

This change keeps the modal product summary and complete transaction button aligned with WooCommerce checkout recalculation. It also applies the repeat-trial eligibility check during standard WooCommerce checkout refreshes when the guest email is present only in serialized `post_data`, which can happen when the checkout URL is accessed directly. The account-association behavior remains scoped to modal checkout.

Closes NPPM-2838.

1. Configure WooCommerce, WooCommerce Subscriptions, and a supported Stripe test gateway for checkout.
2. Create a subscription product with a free trial and add it to a page using the Newspack modal checkout flow.
3. Create or identify a user who already has a subscription to that product, so the user is no longer eligible for the free trial.
4. Open the subscription page in an anonymous browser session.
5. Start modal checkout with the email address for the user who has already used the trial.
6. Continue to the payment step.
7. Confirm that the "Complete transaction" button shows the amount due today, not zero.
8. Confirm the transaction details show the paid amount due today and the recurring total.
9. Return to billing details and repeat with a new email address that has not used the trial.
10. Confirm that the "Complete transaction" button still shows zero due today for the eligible new trial user.
11. Add the same trial subscription product to the cart and open the standard WooCommerce checkout URL directly.
12. Enter the email address for the user who has already used the trial.
13. Confirm that the standard checkout order review also recalculates to the paid amount due today.

### Other information:

* [x] Have you added an explanation of what your changes do and why you would like us to include them?


plugins/newspack-blocks/tests/test-modal-checkout.php
```diff
@@ -0,0 +1,362 @@
...
+			}
+
+			/**
+			 * Get the product subtotal.
+			 *
+			 * @return string
+			 */
+			public function get_product_subtotal() {
+				return $this->subtotal;
+			}
```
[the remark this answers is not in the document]
rbcorrales: Addressed in 4515684. The test cart double now accepts `$product` and `$quantity`, matching the production `WC_Cart::get_product_subtotal()` call shape, then ignores those values before returning the configured subtotal.

plugins/newspack-plugin/tests/mocks/wc-mocks.php
```diff
@@ -451,6 +451,10 @@ class WC_Subscriptions {
 	 * caller passed the expected sign-up-fee mode and orders_to_include list.
 	 */
 	class WC_Subscriptions_Switcher {
+		public static function cart_contains_switches() {
+			return false;
+		}
```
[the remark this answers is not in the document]
rbcorrales: Addressed in 45156845. The `WC_Subscriptions_Switcher` mock now accepts the optional `$item_action` argument, so calls like `cart_contains_switches( 'any' )` no longer trigger a PHP 8 argument count error in tests.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1767,6 +1769,85 @@ public static function order_button_text_with_price() {
 		);
 	}
 
+	/**
+	 * Get the cart product summary shown at the top of modal checkout.
+	 *
```
adekbadek, line 1800: **Suggestion:** Helper duplicates the cart-validation preamble still inline in the template — `get_cart_product_summary()` reproduces the single-item cart gauntlet (`get_cart_contents_count() === 1`, `array_key_first`, the `woocommerce_cart_item_product` / `woocommerce_checkout_cart_item_visible` filters) that still lives inline in `render_before_checkout_form()`. The template calls the new helper for the inner summary but re-derives `$_product` and the visibility check itself just above, so the same conditions are evaluated twice per render and the two copies can diverge over time. Consider having the template lean entirely on the helper (or extracting the shared guard).

plugins/newspack-blocks/src/modal-checkout/index.js
```diff
@@ -208,6 +208,72 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
...
+				let productSummaryRequest = false;
+				function requestUpdatedProductSummary() {
+					if ( productSummaryRequest ) {
+						productSummaryRequest.abort();
+					}
+					const request = $.ajax( {
+						url: newspackBlocksModalCheckout.ajax_url,
+						method: 'POST',
+						data: {
+							action: 'get_cart_product_summary',
```
adekbadek, line 240: **Suggestion:** get_cart_product_summary AJAX fallback omits the modal_checkout flag — The `get_cart_product_summary` request sends only `action`, with no `modal_checkout=1` and no serialized `post_data`, so `is_modal_checkout()` is false while it renders and the trial-limiting filters do not fire during it. The primary path (`updateProductSummaryFromOrderReview`, reading the order-review DOM) is correct because the order-review recalc carries the flag; this AJAX path only runs as a fallback when there is no `tr.cart_item` in the order-review table. The charge itself is unaffected (the order POST carries `modal_checkout=1`), so this is a display-consistency concern rather than a mischarge — but it means the fallback could momentarily render an un-limited price. Either add `modal_checkout: 1` to the request data so the endpoint is recognized as modal checkout, or confirm the fallback is never reached for trial products. `get_cart_total` has the same blind spot but only emits a number.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -636,8 +636,13 @@ public static function limit_free_trials_to_one_per_user( $trial_length, $produc
 
 		$user_id = get_current_user_id();
 
-		// If not logged in, try to get the user ID from the billing email.
-		if ( ! $user_id && method_exists( 'Newspack_Blocks\Modal_Checkout', 'get_user_id_from_email' ) ) {
+		// If not logged in during modal checkout, try to get the user ID from the billing email.
+		if (
+			! $user_id &&
+			method_exists( 'Newspack_Blocks\Modal_Checkout', 'is_modal_checkout' ) &&
+			\Newspack_Blocks\Modal_Checkout::is_modal_checkout() &&
```
adekbadek: **Suggestion:** Scoping the trial email-lookup to modal checkout drops it from standard checkout — This gates the `billing_email`-to-user resolution behind `is_modal_checkout()`. Before this change `limit_free_trials_to_one_per_user()` resolved the email for *any* logged-out request; now it only does so during modal checkout. Confirmed against the real WooCommerce Subscriptions stack: a repeat-trial email submitted through standard (non-modal) checkout now returns the full trial length (the $0 trial is offered again), whereas during modal checkout it correctly returns 0. Logged-in repeat subscribers are still caught everywhere via `get_current_user_id()`, so the behavioural change is bounded to logged-out guests on a non-modal path. The PR description frames the scoping as intentional ("so standard checkout behavior is not expanded"). Worth confirming explicitly: does any in-scope site expose a non-modal subscription checkout to guests? If so, this re-opens the repeat-trial loophole there. If standard guest checkout is genuinely out of scope (subscriptions always funnel through modal checkout), this is fine as-is.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/class-newspack-test-modal-checkout.php
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Test double for the modal checkout class.
+ *
+ * @package Newspack\Tests
+ */
+
+/**
+ * Test double for the modal checkout class when newspack-blocks is not loaded.
+ */
+class Newspack_Test_Modal_Checkout {
```
adekbadek, line 11: **Suggestion:** Test double reimplements the real Modal_Checkout logic (drift risk) — `Newspack_Test_Modal_Checkout::get_user_id_from_email()` and `is_modal_checkout()` are near-verbatim copies of the real `Modal_Checkout` methods. Because this is a hand-maintained reimplementation rather than the code under test, it can silently drift: if the real parser's sanitization or `post_data` handling changes, these newspack-plugin tests keep passing against stale logic and stop proving the cross-plugin integration. Consider aliasing to the real class when newspack-blocks is loaded and treating this double purely as a "blocks not loaded" stub, with a comment that it must mirror the real contract. (The newspack-blocks `test-modal-checkout.php` already exercises the real implementation directly.)

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1767,6 +1769,85 @@ public static function order_button_text_with_price() {
...
+			$cart_item_key
+		); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+
+		ob_start();
+		echo wp_kses( $product_name, $allowed_html ) . ': ';
+		if ( function_exists( 'wc_get_formatted_cart_item_data' ) ) {
+			echo wp_kses( wc_get_formatted_cart_item_data( $cart_item ), $allowed_html );
+		}
+		echo wp_kses( $product_subtotal, $allowed_html );
+		return wp_kses( ob_get_clean(), $allowed_html );
```
adekbadek: **Nit:** Redundant double wp_kses pass — `get_product_subtotal()` already returns sanitized price HTML and `wc_get_formatted_cart_item_data()` returns trusted markup; wrapping each fragment in `wp_kses()` and then `wp_kses()`-ing the assembled buffer again is belt-and-suspenders. Harmless and defensible for echoed output, just redundant. The `bdi` / `span` / `small` allowlist additions themselves are correct and necessary for `wc_price()` output.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1966,7 +2043,25 @@ public static function ensure_base_default_customer_address( $option_value ) {
 	 * @return false|int User ID if found by email address, false otherwise.
 	 */
 	public static function get_user_id_from_email() {
```
adekbadek, line 2066: **Nit:** Email→subscription enumeration surface (pre-existing, not amplified) — Parsing `billing_email` out of serialized `post_data` is new code, but the capability it exposes — an anonymous shopper supplying an arbitrary email and inferring whether it belongs to an existing subscriber by observing whether trial pricing is denied — already existed via the top-level `billing_email` field this function read before. The change widens the read to the serialized field WooCommerce actually sends (required for the fix) without amplifying the abuse: the resolved user only ever removes trial eligibility, never grants access or discloses PII. Sanitization is correct (`is_string` guard → `wp_unslash` → `wp_parse_str` → `sanitize_email`). Flagging for awareness only; no action needed.

plugins/newspack-blocks/src/modal-checkout/index.js
```diff
@@ -208,6 +208,72 @@ import { domReady, onCheckoutPlaceOrderProcessing } from './utils';
...
+						return false;
+					}
+
+					const $productName = $cartItem.find( '.product-name' ).clone();
+					$productName.find( '.product-quantity' ).remove();
+					const productName = $productName.text().replace( /\s+/g, ' ' ).trim();
+					const productTotal = $cartItem.find( '.product-total' ).text().replace( /\s+/g, ' ' ).trim();
+					if ( productName && productTotal ) {
+						return setProductSummary( `${ productName }: ${ productTotal }`, false );
+					}
```
adekbadek, line 270: **Nit:** DOM and AJAX summary paths render different markup — `updateProductSummaryFromOrderReview` writes the summary with `setProductSummary( …, false )` → `.text()` (plain text), while the AJAX fallback writes `setProductSummary( response )` → `.html()` (server-rendered markup including the `<bdi>` / price spans). A repeat-trial user can therefore see a plain-text price in one state and styled markup in another, depending on which branch ran. If visual consistency matters, prefer one source of truth; at minimum a comment explaining why both formats are acceptable would help the next maintainer.

adekbadek (changes requested):
Targets the root cause cleanly and is well-tested. 

One behavioural point worth an explicit confirmation before merge (left inline): gating the email lookup behind `is_modal_checkout()` also removes it from standard (non-modal) guest checkout, where the prior code ran for any logged-out request — this might cause issues on certain configurations.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
rbcorrales: Addressed in a439465. I widened the repeat-trial eligibility check to standard checkout refreshes too, because direct checkout access can hit this path with the billing email only in serialized `post_data`. I kept account association scoped to modal checkout, so this only affects price and trial eligibility.

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Acknowledged. I am leaving the behavior as-is for this PR because the lookup only removes trial eligibility and is needed for the display fix. I noted the privacy/abuse tradeoff in the internal Linear thread here: https://linear.app/a8c/issue/NPPM-2838/#comment-8db38fa0

plugins/newspack-blocks/src/modal-checkout/index.js
rbcorrales: Addressed in 4a3bccd. I kept the DOM path text-only and added a comment explaining the split: the AJAX path can render server markup because PHP runs it through `wp_kses`, while the DOM path avoids introducing an unsanitized HTML sink.

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Addressed in 589a689. The summary is now assembled once and sanitized once with the allowed-html list, rather than sanitizing each fragment and then the full buffer.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/class-newspack-test-modal-checkout.php
rbcorrales: Addressed in 349f032. I added a comment tying the fallback test double to `Newspack_Blocks\Modal_Checkout::get_user_id_from_email()`, so future changes know the contract must stay aligned. The real implementation is still covered directly in the newspack-blocks tests.

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Addressed in fbfb3f4. I extracted `get_cart_product_summary_context()` and the template now reuses that context when rendering the summary, so the single-item and visibility checks are not duplicated.

plugins/newspack-blocks/src/modal-checkout/index.js
rbcorrales: Addressed in e081497. Both `get_cart_product_summary` and `get_cart_total` now send `modal_checkout: 1` and serialized checkout `post_data`, so repeat-trial checks can resolve the entered email during the refresh and fallback paths.

rbcorrales:
Thanks for the review @adekbadek. I addressed the feedback. The main behavioral change after the review is that repeat-trial eligibility now also applies to standard checkout refreshes when the guest email is only available in serialized `post_data`, while account association remains modal-only. I also updated the PR description and testing steps to cover that path.

adekbadek (approved):
Re-reviewed the changes since my last pass. The repeat-trial loophole is now closed on standard checkout too — trial-limiting is un-gated from modal checkout in both filters (`limit_free_trials_to_one_per_user` and `subscriptions_product_limited_for_user`), and the email lookup no-ops safely outside checkout, so it only bites on a real repeat-subscriber email. Both paths are test-covered. The cart-guard duplication, double `wp_kses`, AJAX modal flag, and DOM/AJAX markup split are all addressed.

One eyes-open note, not blocking: un-gating widens the email→subscriber enumeration surface to standard checkout as well (removes trial eligibility only — no PII or access), an accepted tradeoff per NPPM-2838.

Approving.

PR #269, opened 2026-06-10
fix(newsletters): harden ActiveCampaign sends against API timeouts

adekbadek:
### All Submissions:

Hardens ActiveCampaign newsletter sends against ActiveCampaign-side API hangs, and makes a retried send safe.

Background: a publisher hit `cURL error 28: Operation timed out after 45002ms with 0 bytes received` when sending to their main list. Root cause is on ActiveCampaign's side – specific campaigns got into a state where AC's V1 `/admin/api.php` calls referencing them never respond, tripping our 45s request timeout. It self-resolved with no Newspack change. We can't fix the AC hang, but we can stop it from corrupting our send flow, stop surfacing a raw cURL string to publishers, and – most importantly – make sure a retry never sends the newsletter twice.

The send is a sequence of side-effectful steps (delete old campaign → upload message → create campaign → trigger send), and any step can time out. The core insight: a timed-out `campaign_status` (the trigger) is ambiguous – AC may have started sending before the HTTP response hung. So the fix makes the sequence verify state before acting.

1. **Verify the stored campaign's state before resending (double-send safety).** On entry, `send()` reads the campaign's actual status from AC rather than blindly deleting and recreating it. If it's already scheduled/sending/sent, the send is treated as done (no resend). If it's a confirmed draft, it's safe to recreate and send. If its state can't be verified (the status check itself times out) or it's in an indeterminate paused/stopped state, `send()` fails safe with a publisher-facing notice and does **not** resend – a stranded "needs review" is far better than two newsletters to the list.

2. **Bounded cleanup timeout.** Deleting a previous campaign is best-effort cleanup, so it uses a short, bounded timeout (15s) instead of the 45s default – a stuck prior campaign can't consume the request budget. Required plumbing a caller-supplied `timeout` through the v1/v3 request methods, which previously silently dropped it.

3. **Publisher-friendly timeout error.** A cURL-28 / "timed out" transport failure is rephrased into actionable guidance instead of the raw cURL message; the original error is preserved as `WP_Error` data for logging/support tooling. Other transport errors pass through unchanged.

**Scope note:** this is the synchronous, idempotent foundation. It does not auto-retry through a hang – the publisher still sees a failure on the spot – but the next attempt (manual re-publish, or the existing scheduled-send retry) now resumes **safely** instead of risking a duplicate send. A fully async/queued send with automatic retry is a larger follow-up that can sit on top of this.

The state gate closes the **timeout-retry** race (a second send attempt after a prior attempt hung), which is the failure mode behind this ticket. It does **not** serialize two genuinely concurrent in-flight `send()` calls (e.g. a manual publish racing the scheduled-retry cron, or a double-click): there is no lock around the `ac_campaign_id` read-modify-write, so both could read the same draft and pass the gate. That concurrent-invocation race is pre-existing and out of scope here; closing it would need a real lock (e.g. a transient/meta guard) and is a separate change.

Closes NPPM-2878.

1. Run the suite: from `plugins/newspack-newsletters`, `n test-php --filter ActiveCampaignResilienceTest` – all tests pass.
2. The suite is a self-proving spec covering: caller-timeout plumbing (v1 + v3), bounded cleanup timeout, cURL-28 humanization (v1 + v3) with the original error preserved, non-timeout errors passing through unchanged, and the send-state gate: an already-dispatched campaign is not resent, an unverifiable campaign fails safe, a paused/stopped campaign is flagged for review, and a confirmed draft is recreated and sent (asserting it force-deletes the old draft without a redundant status re-check).
3. Regression: `n test-php --filter ActiveCampaignApiRequestsTest` still passes.
4. The refactored request layer (incl. the new caller-timeout path) was also exercised against the live ActiveCampaign API, read-only.

### Other information:


plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -472,6 +493,31 @@ public function api_v3_request( $resource, $method = 'GET', $options = [] ) {
...
+	 * @return WP_Error The original error, or a friendlier one for timeouts.
+	 */
+	private function humanize_transport_error( $error ) {
+		$message = $error->get_error_message();
+		if ( false !== stripos( $message, 'timed out' ) || false !== stripos( $message, 'cURL error 28' ) ) {
+			return new \WP_Error(
+				'newspack_newsletters_active_campaign_timeout',
+				__( 'ActiveCampaign did not respond in time. This is usually a temporary issue on ActiveCampaign\'s end. Please wait a few minutes and try again; if the problem continues, try sending a fresh copy of the newsletter.', 'newspack-newsletters' )
+			);
+		}
```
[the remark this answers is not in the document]
adekbadek: Addressed in ae0498d71 — the humanized timeout now carries the original transport failure (`original_error_code` / `original_error_message` / `original_error_data`) as WP_Error data, so the publisher sees the friendly message while logging/support tooling keep the underlying cURL detail. Added a test asserting the original error is preserved.

adekbadek:
Retargeted to `release` — this is a hotfix for NPPM-2878. The two commits were rebased cleanly onto `origin/release` (no conflicts; the edited methods are identical between `main` and `release`). As a `fix:` it will ship as a patch release and promote back to `main`/`alpha` via the normal release flow.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1309,6 +1416,63 @@ public function save( $meta_id, $post_id, $meta_key ) {
...
+			// during an AC incident (exactly when this matters) carries the
+			// generic api_error code, not the timeout code, and must still block.
+			return new \WP_Error(
+				'newspack_newsletters_active_campaign_unverified_campaign',
+				__( 'Newspack could not confirm with ActiveCampaign whether this newsletter had already started sending, so it was not resent (to avoid a duplicate). Please wait a few minutes and try again, or check the campaign status in ActiveCampaign.', 'newspack-newsletters' )
+			);
+		}
+		// A successful response with no campaign means it no longer exists in
+		// ActiveCampaign: nothing was dispatched, so it is safe to recreate.
+		$status = isset( $campaigns[0]['status'] ) ? (string) $campaigns[0]['status'] : self::STATUS_DRAFT;
```
dkoo, line 1482: **Suggestion:** Empty campaign_list defaults to "draft" → recreate-and-send (strong suggestion: fix or document) — When `campaign_list` returns a successful response with no campaign row, `$status` falls back to `STATUS_DRAFT` (`'0'`), which `get_campaign_dispatch_state()` maps to `CAMPAIGN_DRAFT` — the **only** dispatch state that lets `send()` proceed to recreate and trigger a fresh campaign to the entire list. Every other uncertain case in this function deliberately fails safe; this one defaults toward the maximally destructive action.

Per the maintainer, ActiveCampaign's API is documented to **never** return an empty campaign list for a campaign that was previously dispatched and then deleted — so in practice this path should not occur. A runtime probe confirms the supporting half: AC's missing-campaign response (`result_code=0`, "Nothing is returned") is converted to a `WP_Error` and fails safe as `unverified_campaign`; only a hypothetical *successful-but-empty* list would reach the `draft` default. So this is **not a blocker** — but because the consequence if the assumption ever broke is a duplicate send to the full list (amplified up to 10× by the scheduled-retry loop), it's worth either a one-line defensive fix (route the no-row case to `CAMPAIGN_NEEDS_REVIEW`) or an explicit in-code comment recording that the safety here rests on AC never returning an empty success for a dispatched-then-deleted campaign.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1309,6 +1416,63 @@ public function save( $meta_id, $post_id, $meta_key ) {
...
+			case self::STATUS_SCHEDULED:
+			case self::STATUS_SENDING:
+			case self::STATUS_COMPLETED:
+				return self::CAMPAIGN_ALREADY_DISPATCHED;
+			default:
+				// Paused (3) and stopped (4) are campaigns that were sending and
+				// got halted, so they have likely already delivered to part of
+				// the list; an unknown status is equally indeterminate. None are
+				// safe to recreate-and-resend, so route them to manual review.
+				return self::CAMPAIGN_NEEDS_REVIEW;
```
dkoo, line 1502: **Suggestion:** STATUS_DISABLED is "deletable" but routed to needs-review — silent asymmetry — `STATUS_DISABLED` (`'6'`) is in `DELETABLE_STATUSES` (line 62), yet in the dispatch-state `switch` it falls through `default` → `CAMPAIGN_NEEDS_REVIEW`. That's a defensible choice (a disabled campaign is genuinely indeterminate), but the asymmetry is silent: a future reader comparing the two constants will wonder why `6` is simultaneously "deletable" and "needs review," and may "fix" it by adding it to the safe-to-recreate set. A one-line comment in the `default` case noting that disabled is intentionally *not* treated as a fresh draft would prevent that.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1319,11 +1483,30 @@ public function save( $meta_id, $post_id, $meta_key ) {
 	public function send( $post ) {
 		$post_id = $post->ID;
 
-		$error = null;
-
-		/** Clean up existing campaign. */
+		// A campaign created by a previous attempt may already have been
```
dkoo, line 1516: **Suggestion:** Concurrent send() calls can still both pass the draft gate (pre-existing) — There's no lock around the read-modify-write of the `ac_campaign_id` post meta, so two near-simultaneous `send()` calls (e.g. a manual publish racing the scheduled-retry cron, or a double-click publish) can both read the stored campaign as a draft, both pass the gate, and both `create_campaign()` + trigger `campaign_status`. The base class's `is_newsletter_sent()` short-circuit only helps once one call has *completed*; it doesn't serialize concurrent in-flight calls. This is **pre-existing** and out of scope for a timeout fix — but since the PR's headline is double-send safety, it's worth a sentence in the description clarifying the gate closes the timeout-retry race, not a true concurrent-invocation race.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -155,6 +203,43 @@ public function api_v3_request( $resource, $method = 'GET', $options = [] ) {
...
+	 * ANY error, so a timeout that this match ever missed would still block a
+	 * resend rather than slip through.
+	 *
+	 * @param WP_Error $error A WP_Error returned by wp_safe_remote_request.
+	 *
+	 * @return WP_Error The original error, or a friendlier one for timeouts.
+	 */
+	private function humanize_transport_error( $error ) {
+		$message = $error->get_error_message();
+		if ( false !== stripos( $message, 'timed out' ) || false !== stripos( $message, 'cURL error 28' ) ) {
```
dkoo, line 242: **Suggestion:** Timeout detection keys on an English substring — `humanize_transport_error()` detects timeouts via `stripos( $message, 'timed out' )` and `stripos( $message, 'cURL error 28' )`. The `'timed out'` branch is locale/transport-fragile (cron/CLI sends or non-English builds may phrase it differently), so the publisher could still see a raw cURL string. The `'cURL error 28'` token is locale-stable (cURL error *numbers* aren't translated), so the important case is covered — and the docblock correctly notes send-safety doesn't depend on this match (the dispatch-state gate fails safe on any error). So this is purely a residual UX gap, not a safety issue; flagging only so the `'timed out'` fragility is a known limitation.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -15,6 +15,52 @@
...
+	 * Timeout, in seconds, for best-effort cleanup requests (deleting a
+	 * previously-created campaign before a new send). ActiveCampaign
+	 * occasionally stops responding to calls that reference a particular
+	 * campaign; when that happens the campaign_list/campaign_delete cleanup must
+	 * fail fast instead of consuming the whole request budget and stranding the
+	 * send behind a stuck campaign it was only trying to tidy up.
+	 *
+	 * @var int
+	 */
+	const CLEANUP_REQUEST_TIMEOUT = 15;
```
dkoo, line 41: **Nit:** CLEANUP_REQUEST_TIMEOUT docblock now understates its role — `CLEANUP_REQUEST_TIMEOUT` is documented purely as a "best-effort cleanup" timeout, but `get_campaign_dispatch_state()` now also reuses it for the *safety-critical* status check. The 15s value is fine there (a fast failure correctly routes to the fail-safe `unverified_campaign` branch — exactly when AC is slow, you *want* to bail quickly). But the docblock should mention this second, non-cleanup use so the constant isn't later "fixed" by someone who thinks it only affects disposable cleanup calls.

plugins/newspack-newsletters/tests/test-active-campaign-resilience.php
```diff
@@ -0,0 +1,524 @@
...
+					],
+				],
+			] // 0 = draft.
+		);
+
+		$result = $active_campaign->send( get_post( $post_id ) );
+
+		$this->assertTrue( $result, 'A draft prior campaign is safe to recreate and send.' );
+		$this->assertSame( '777', (string) get_post_meta( $post_id, 'ac_campaign_id', true ), 'The fresh campaign id replaces the stored draft.' );
+		$this->assertContains( 'campaign_create', $this->called_actions, 'A fresh campaign must be created.' );
```
dkoo, line 521: **Nit:** Draft path doesn't assert the old campaign is cleaned up — `test_send_recreates_and_sends_when_prior_campaign_is_draft` asserts `campaign_create` and `campaign_status` fire, but not that the confirmed-draft path issues a `campaign_delete` (with `force=true`, skipping the status re-check). Since skipping that re-check on `force=true` is a behavior this PR deliberately introduced (`delete_campaign` line 1356), an assertion that the draft path calls `campaign_delete` exactly once and does *not* make a second `campaign_list` would lock it in.

dkoo (commented):
Strong, defensively-reasoned fix, and the test suite is excellent — it pins every send-state branch, the bounded-timeout plumbing, and timeout humanization. The send-state gate fails safe in the right direction (refuse-to-resend on any uncertainty), and the caller-supplied `timeout` now correctly survives the `$args + $options` merge. No blockers.

One item worth a deliberate decision: `get_campaign_dispatch_state()` defaults the single unverifiable case — a successful-but-empty `campaign_list` — to `draft`, i.e. recreate-and-send (the most destructive outcome, re-run up to 10× by the scheduled-send loop). ActiveCampaign is documented never to return an empty list for a dispatched-then-deleted campaign, and a runtime check confirmed its missing-campaign response (`result_code=0`) fails safe — so this shouldn't occur in practice. But since the cost if it ever did is a duplicate send to the full list, consider either a one-line guard (route the no-row case to needs-review) or an in-code comment recording that the safety rests on that AC behavior. The rest are smaller suggestions and pre-existing nits.

Note: the live ActiveCampaign API wasn't exercised across real accounts.

**Notes on code outside this PR's changed lines** (can't be posted inline, so collected here):
- `plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php:191` — **Suggestion:** api_v3_request treats HTTP 400 as success (pre-existing) — The error gate `if ( 400 < $response_code || ! empty( $response_body['errors'] ) )` uses strict `>`, so an HTTP `400 Bad Request` with no `errors` array in its body slips through as a valid response. Pre-existing (not introduced here), but it undercuts the resilience goal — a `400` could feed bad data into the dispatch-state reasoning. Since this PR is specifically hardening the AC path, changing to `400 <=` while you're here would be a cheap improvement; otherwise worth a note that it's knowingly left alone.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Kept the behavior and documented it (in 3b25b94ed). The empty-list branch still treats an absent campaign row as a deleted/missing campaign that is safe to recreate, which preserves resend recovery when a campaign was removed in AC out-of-band. Added an in-code comment recording that this safety rests on AC never returning an empty *success* for a dispatched-then-deleted campaign (it returns `result_code=0`, which lands in the `is_wp_error()` fail-safe branch), and that the cost if that ever broke is a duplicate send — so the default must not be widened without re-verifying that AC behavior.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Added a comment in the `default` case (3b25b94ed) noting that disabled (6) is intentionally *not* treated as a fresh draft despite appearing in `DELETABLE_STATUSES` — "deletable" only means the cleanup path may remove it, not that its dispatch state is known safe to resend.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Agreed — pre-existing and out of scope. No code change; clarified the PR description to state that the gate closes the timeout-retry race (a second attempt after a hang), not a genuinely concurrent-invocation race, since there is no lock around the `ac_campaign_id` read-modify-write. That would need a real lock and is a separate change.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Documented the limitation in the `humanize_transport_error()` docblock (3b25b94ed): the `cURL error 28` token is locale-stable so the important case is always caught, while the `timed out` fallback is locale/transport-fragile and may miss — a residual UX gap only, never a safety issue (the dispatch-state gate fails safe on any error).

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
adekbadek: Extended the `CLEANUP_REQUEST_TIMEOUT` docblock (3b25b94ed) to note its second, safety-critical use in `get_campaign_dispatch_state()` — and that a fast failure there is desirable, so it should not be lowered on the assumption it only affects disposable cleanup calls.

plugins/newspack-newsletters/tests/test-active-campaign-resilience.php
adekbadek: Added assertions in 8dce40f26: the draft path now asserts `campaign_delete` fires exactly once and that `campaign_list` runs only once (the dispatch-state gate), locking in that the `force=true` delete skips the redundant status re-check.

adekbadek:
Thanks for the careful read. Addressing the two points from your review summary:

**Empty `campaign_list` defaults to `draft`.** Kept the behavior and documented it in 3b25b94ed (also replied on the inline thread). The empty-list case is the genuine "campaign deleted/missing in AC" signal, so treating it as safe-to-recreate preserves resend recovery for an out-of-band deletion; routing it to needs-review would block that legit case to guard against an AC behavior you confirmed is unreachable in practice (`result_code=0` already fails safe). Added an in-code comment recording that the safety rests on that AC behavior and that the cost if it ever broke is a duplicate send.

**`api_v3_request` treats HTTP 400 as success (pre-existing).** Fixed while here in 3b25b94ed — changed the gate from `400 < $response_code` to `400 <=`, so a bare `400 Bad Request` with no `errors` array is now treated as an error instead of slipping through. Added a short comment explaining why 400 is included.

All `ActiveCampaignResilienceTest` tests pass (18 tests, 55 assertions) and both changed files are PHPCS-clean.

dkoo (approved):
Thanks, @adekbadek!

PR #274, opened 2026-06-10
fix(newsletters): render layout previews correctly in Safari/WebKit

thomasguillot:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?

Newsletter layout previews (the Layouts list DataView, the editor's layout sidebar, and the init-modal layout picker) rendered correctly in Chromium but broke in Safari/WebKit: the site logo blew out to its intrinsic size, social links lost their flex layout, the site-logo "Drop files to upload" overlay showed through, and custom layout fonts/colors fell back to defaults.

**Root cause.** `NewsletterPreview` got the core editor stylesheets into the `BlockPreview` iframe by hand-injecting `<link>`/`<style>` nodes into the iframe `<head>` on load. Gutenberg owns that iframe document and re-renders it, discarding anything injected outside its own channels. Chromium's timing left the injected nodes in place; WebKit wiped them, so the iframe was missing core block/editor CSS.

**Fix.** Route styles through Gutenberg's own channels — the same way the Post/Site editors do — so they persist across browsers:

- **Core stylesheets** (`wp-block-library`, `-theme`, `wp-edit-blocks`, `wp-components`) via `settings.__unstableResolvedAssets`, referenced by URL and baked into the iframe `srcDoc`. Seeded only where that setting is empty (admin-shell surfaces) and never on the live editor, so the editor canvas is never reloaded.
- **Global styles + default fonts** via a seeded `settings.styles` entry (safe `{ css }`, no raw HTML concatenation).
- **Per-layout overrides** (body/header fonts, background/text colors, custom CSS) via `BlockPreview`'s `additionalStyles` prop, with `!important` on the generated rules so they win over both the seeded defaults and the editor's higher-specificity, variable-based font rules on the in-editor preview surfaces.

This removes the WebKit-fragile hand-injection: the redundant stylesheet clone, the duplicated per-instance/global `<style>` injection, and the dead `#newspack-newsletters__layout-css` element (whose only consumer in `posts-inserter/posts-preview.js` is also removed — that preview still gets fonts via `useCustomFontsInIframe`). Net result is a smaller component with no per-block CSS patches.

1. Build newspack-newsletters. Open Newsletters → Layouts list **in Safari**.
2. Prebuilt layout thumbnails render correctly: logo sized to its container, social icons in a styled row, no "Drop files to upload" text.
3. Create a saved layout with a distinct custom body/header font and custom background/text colors (or use a few seeded ones). Its thumbnail renders in that font and those colors in Safari — and so does its preview in the **editor's layout sidebar / the init-modal layout picker** (the surfaces where `editor.css` competes; `!important` ensures the layout's styles win).
4. Confirm all of the above still render correctly in **Chrome** (no regression).
5. Open the newsletter editor and confirm the main editor canvas behaves normally (the seed is guarded to never touch the live editor's resolved assets).

### Other information:


plugins/newspack-newsletters/src/components/newsletter-preview/index.js
```diff
@@ -43,24 +53,46 @@ const withSamplePostsInserter = blocks => {
...
-	const [ css, setCss ] = useState( '' );
 	const [ isReady, setIsReady ] = useState( false );
 
+	// Admin-shell previews lack the editor-provided assets, so seed them; skip
+	// the live editor (populated string) so its canvas isn't reloaded.
+	const { updateSettings } = useDispatch( blockEditorStore );
+	const resolvedAssets = useSelect( select => select( blockEditorStore ).getSettings().__unstableResolvedAssets, [] );
 	useEffect( () => {
-		const cssRules = [];
+		if ( typeof resolvedAssets?.styles === 'string' ) {
```
dkoo, line 79: **Suggestion.** This guard is the only thing keeping `updateSettings` off the shared `@wordpress/block-editor` store that the live editor canvas (where `NewsletterPreview` is also mounted, via `newsletter-editor/layout/index.js`) and the init-modal picker use — and it infers "am I in the admin shell" purely from the absence of a private `__unstableResolvedAssets.styles` string. I confirmed it works in the current WP (the live editor populates that string before any preview mounts, so the guard returns early), so this is a forward-compatibility concern rather than a current bug. Still, an explicit admin-shell prop/context flag from the mount sites would be more robust than inferring the environment from an `__unstable` field — at minimum, worth an inline comment documenting the assumption.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
```diff
@@ -1,21 +1,31 @@
...
 import { getTemplateBlocks } from '../../editor/blocks/posts-inserter/utils';
 
 const POSTS_INSERTER = 'newspack-newsletters/posts-inserter';
 
+const CORE_STYLESHEET_IDS = [ 'wp-block-library-css', 'wp-block-library-theme-css', 'wp-edit-blocks-css', 'wp-components-css' ];
+
+const DEFAULT_FONTS_CSS =
+	'body *:not(code) { font-family: georgia, serif; } body h1, body h2, body h3, body h4, body h5, body h6 { font-family: arial, sans-serif; }';
+
+const buildResolvedStyles = () =>
```
dkoo: **Suggestion.** `buildResolvedStyles()` reads core stylesheets by hardcoded id and joins their `outerHTML`. `.filter( Boolean )` tolerates a missing id, but silently — if an expected sheet isn't enqueued on some surface (the Layouts list page only force-enqueues `wp-edit-blocks`; the other three rely on incidental enqueue) or is renamed/split in a future WP, that style just disappears from the preview with no diagnostic. Consider logging when an expected id is absent, or sourcing the list from the same enqueue contract the PHP side guarantees rather than scraping the DOM.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
```diff
@@ -43,24 +53,46 @@ const withSamplePostsInserter = blocks => {
...
-			cssRules.push( `h1, h2, h3, h4, h5, h6 { font-family: ${ meta.font_header }; }` );
+			rules.push( `body h1, body h2, body h3, body h4, body h5, body h6 { font-family: ${ meta.font_header } !important; }` );
+		}
+		if ( meta.background_color ) {
+			rules.push( `body { background-color: ${ meta.background_color } !important; }` );
+		}
+		if ( meta.text_color ) {
+			rules.push( `body { color: ${ meta.text_color } !important; }` );
 		}
 		if ( meta.custom_css ) {
```
dkoo, line 107: **Nit.** `meta.font_*`, `background_color`, `text_color`, and `custom_css` are concatenated raw into the `additionalStyles` string. Not a new XSS surface (preview-only iframe, publisher-controlled, same trust model as before) — but the old path ran everything through `getScopedCss()`'s CSSOM parse, which normalized/dropped malformed input. Now a malformed value or an unclosed brace in `custom_css` can leak into or break the surrounding seeded rules in the preview (not the sent email). Minor; just flagging that the parser safety-net is gone.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
```diff
@@ -1,21 +1,31 @@
...
 import './style.scss';
-import { getScopedCss } from '../../newsletter-editor/styling';
 import { getSamplePosts } from '../../editor/blocks/posts-inserter/sample-posts';
 import { getTemplateBlocks } from '../../editor/blocks/posts-inserter/utils';
 
 const POSTS_INSERTER = 'newspack-newsletters/posts-inserter';
 
+const CORE_STYLESHEET_IDS = [ 'wp-block-library-css', 'wp-block-library-theme-css', 'wp-edit-blocks-css', 'wp-components-css' ];
+
+const DEFAULT_FONTS_CSS =
```
dkoo, line 20: **Nit.** `DEFAULT_FONTS_CSS` hardcodes the default fonts that the editor's styling layer also expresses (via CSS variables / the custom-styles defaults). Two sources of truth for the same defaults will drift over time — consider deriving from a shared constant.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
```diff
@@ -43,24 +53,46 @@ const withSamplePostsInserter = blocks => {
...
-	const instanceId = useInstanceId( NewsletterPreview );
-	const elementId = `preview-${ instanceId }`;
-	const [ css, setCss ] = useState( '' );
 	const [ isReady, setIsReady ] = useState( false );
 
+	// Admin-shell previews lack the editor-provided assets, so seed them; skip
+	// the live editor (populated string) so its canvas isn't reloaded.
+	const { updateSettings } = useDispatch( blockEditorStore );
+	const resolvedAssets = useSelect( select => select( blockEditorStore ).getSettings().__unstableResolvedAssets, [] );
 	useEffect( () => {
```
dkoo, line 78: **Nit.** Seeding the global block-editor store from inside each mounted preview's effect means every card on the Layouts grid races to write the same settings. It's idempotent and guarded, so functionally safe — but a single setup at the admin-shell screen mount would be a cleaner ownership model and would sidestep the environment-inference concern above. Optional/architectural.

dkoo (approved):
## Summary

Approving. A well-reasoned fix that moves newsletter layout-preview styling onto Gutenberg's supported channels (`settings.__unstableResolvedAssets`, `settings.styles`, and `BlockPreview`'s `additionalStyles`) instead of hand-injecting `<link>`/`<style>` nodes that WebKit discards. It nets out smaller and cleanly removes the dead `#newspack-newsletters__layout-css` element together with its only consumer. CI is green.

## Verification

Tested in Chrome against an isolated environment built from this branch:

- **Layouts list** renders correctly — logos sized to their container, social icons in a styled row, images intact, no "Drop files to upload" overlay. DOM inspection confirms the preview iframes' `<head>` now carries the seeded core stylesheets (`wp-block-library`, `wp-edit-blocks`, `wp-components`) — the browser-agnostic mechanism that makes the fix hold under WebKit.
- **Init-modal layout picker** and a custom font/color layout render with the per-layout overrides winning (cream background, monospace/cursive fonts) on the surfaces where `editor.css` competes — the `!important` does its job.
- **Live editor canvas** is untouched: the guard is dormant there because `__unstableResolvedAssets.styles` is already a populated string, so the seeding effect returns early — confirming the "never touches the live editor" intent.
- Env debug log is clean.

I could not pixel-verify Safari/WebKit itself in this environment (no WebKit automation available), but the seeding mechanism the fix depends on was confirmed present in the iframe, and it's browser-agnostic by construction.

A few inline suggestions below — none blocking. One small thing worth a test: there's no unit coverage pinning the seeding guard (undefined resolved-assets → seed once; string present → skip), which would protect the private-API assumption noted inline.

## Out of scope (for a separate follow-up)

Separately noticed: when editing a *saved layout* directly, the layout editor canvas can briefly fail to show the layout's custom background color (it applies correctly in the previews and in the newsletter editor). That path lives in `src/newsletter-editor/styling/index.js` (`ApplyStyling`), which this PR doesn't touch and where the preview component isn't even mounted — so it's independent of this change. Flagging it for a separate PR rather than expanding scope here.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
thomasguillot: Added a dev-gated diagnostic (`process.env.NODE_ENV !== 'production'`, matching `empty-state/index.js`): `buildResolvedStyles()` now `console.warn`s the specific missing ids. Only `wp-edit-blocks` is force-enqueued on the Layouts list, so the other three can legitimately be absent on some surface — sourcing from a guaranteed contract isn't possible, but the warning surfaces it instead of failing silently. Production stays quiet. ([4d8108284](https://github.com/Automattic/newspack-workspace/commit/4d8108284))

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
thomasguillot: Documented the assumption inline rather than threading a prop, since the full admin-shell flag would touch every mount site for a forward-compat concern that isn't a current bug. The comment now spells out that the environment is inferred from the private `__unstableResolvedAssets.styles` field — live editor populates it before any preview mounts, admin shell leaves it undefined — and notes that if a future WP populates it everywhere, an explicit mount-site flag would be needed. ([4d8108284](https://github.com/Automattic/newspack-workspace/commit/4d8108284))

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
thomasguillot: Leaving as-is. The old `getScopedCss()` CSSOM normalization was part of the hand-injection path this PR deliberately removed, and re-adding a parse step would bring that complexity back. As you note, it's preview-only, publisher-controlled, and the same trust model as before — a malformed value breaks the preview, not the sent email. Not worth re-introducing the parser for a no-functional-change safety net. Flagging acknowledged.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
thomasguillot: Real concern — they've actually already drifted (the styling layer defaults header font to `Arial, Helvetica, sans-serif` while this preview uses `arial, sans-serif`). But the only shared source is `fontOptgroups` in `newsletter-editor/styling/index.js`, a labeled option structure; extracting a shared defaults constant is a cross-module change that risks the editor styling layer and warrants visual verification. That's beyond a polish pass on this PR, so I'm leaving it for a dedicated follow-up rather than doing it half-way here.

plugins/newspack-newsletters/src/components/newsletter-preview/index.js
thomasguillot: Agreed it's the cleaner ownership model, but leaving it for now: the effect is idempotent and guarded, so the racing writes are functionally safe, and hoisting setup to a single admin-shell mount point would touch every mount site (`layouts-list`, `init-modal` layout picker, `send-button`). That's an architectural change rather than a polish fix, and it's coupled to the environment-inference question above — both are better handled together in a follow-up if we move to an explicit admin-shell flag.

PR #283, opened 2026-06-11
fix(content-gate): cascade taxonomy content rules to child terms (NPPD-1670)

adekbadek:
### All Submissions:

Fixes [NPPD-1670](https://linear.app/a8c/issue/NPPD-1670). Content gates with taxonomy-term content rules now cascade to a targeted term's child terms for hierarchical taxonomies (categories), matching WooCommerce Memberships.

Previously `Content_Restriction_Control::get_post_gates()` matched a post's term IDs against a rule's term IDs with a flat `array_intersect`, so a rule targeting a *parent* category did **not** gate posts assigned only to a *child* category. Restricted content leaked: a publisher reported no paywall appearing on child-category posts.

The fix adds a private `expand_hierarchical_terms()` helper that expands a hierarchical-taxonomy rule's target terms to include all descendants (via `get_term_children()`) before the intersect. Key properties:

- **Evaluation-time expansion**, so child terms added *after* a rule is saved are covered automatically (matches Woo's dynamic behavior).
- **Symmetric** for inclusion and exclusion rules.
- **No-op** for non-hierarchical taxonomies (tags), and no public-contract change (private method only) — the cross-repo `newspack-manager` consumer is unaffected.

1. Create a category tree (parent → child → grandchild) and a post assigned **only** to the child or grandchild category (not the parent).
2. Create a content gate with a content rule targeting the **parent** category, with registration or custom access active.
3. Visit the child/grandchild post as a logged-out reader → the gate (paywall/regwall) now appears. Before this change it did not.
4. Flip the rule to an **exclusion** rule on the parent → the child/grandchild post is no longer gated, while posts outside the subtree still are.
5. Or run the unit tests: `n test-php --filter test_content_rules` (covers inclusion cascade, exclusion cascade, the one-directional guarantee, and non-hierarchical passthrough).

### Rollout note

Because expansion is evaluation-time, on deploy every existing gate targeting a **parent** category recomputes its gated-post set with no migration step — this is a behavior-changing deploy, not a transparent fix. A per-site audit of the currently-live Access-Control sites (which gates expand and by how much) is tracked on [NPPD-1670](https://linear.app/a8c/issue/NPPD-1670).

### Other information:

* [x] Have you successfully run tests with your changes locally? (`Test_Content_Gates`: 56 tests / 167 assertions green; PHPCS clean.)

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
@@ -189,7 +189,12 @@ class Content_Restriction_Control {
 					if ( ( ! $is_exclusion && ! $terms ) || is_wp_error( $terms ) ) {
 						continue 2;
 					}
-					if ( $is_exclusion ? ! empty( array_intersect( $terms, $content_rule['value'] ) ) : empty( array_intersect( $terms, $content_rule['value'] ) ) ) {
+					// For hierarchical taxonomies, a rule targeting a term also covers
+					// that term's descendants, mirroring WooCommerce Memberships' cascade.
+					// The helper also casts the rule's term IDs to integers, matching the
+					// integer IDs from wp_get_post_terms() on the other side of the intersect.
+					$target_terms = self::expand_hierarchical_terms( $content_rule['value'], $taxonomy );
+					if ( $is_exclusion ? ! empty( array_intersect( $terms, $target_terms ) ) : empty( array_intersect( $terms, $target_terms ) ) ) {
 						continue 2;
 					}
 				}
@@ -204,6 +209,36 @@ class Content_Restriction_Control {
 			$post_gates[] = $gate;
 		}
 		return $post_gates;
+	}
+
+	/**
+	 * Expand a set of taxonomy term IDs to include descendant terms when the
+	 * taxonomy is hierarchical.
+	 *
+	 * A content rule targeting a parent term should also match content assigned
+	 * only to that term's descendants, mirroring WooCommerce Memberships' cascade
+	 * behavior. Expansion happens at evaluation time so newly-added child terms
+	 * are covered without re-saving the rule. Non-hierarchical taxonomies (e.g.
+	 * tags) have no descendants and are returned as integer-cast IDs unchanged.
+	 *
+	 * @param array        $term_ids Term IDs from a content rule's value.
+	 * @param \WP_Taxonomy $taxonomy Taxonomy object the term IDs belong to.
+	 *
+	 * @return int[] De-duplicated term IDs including descendants.
+	 */
+	private static function expand_hierarchical_terms( $term_ids, $taxonomy ) {
+		$term_ids = array_map( 'intval', (array) $term_ids );
+		if ( ! $taxonomy->hierarchical ) {
+			return $term_ids;
+		}
+		$expanded = $term_ids;
+		foreach ( $term_ids as $term_id ) {
+			$children = get_term_children( $term_id, $taxonomy->name );
+			if ( ! is_wp_error( $children ) ) {
+				$expanded = array_merge( $expanded, array_map( 'intval', $children ) );
+			}
+		}
+		return array_values( array_unique( $expanded ) );
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
@@ -28,6 +28,26 @@ class Content_Restriction_Control {
 	private static $post_gate_layout_id_map = [];
 
 	/**
+	 * Request-scoped cache of the gates that apply to a post, keyed by post ID.
+	 * The gate set a post matches depends only on the post and the (request-stable)
+	 * gate configuration, so it is safe to memoise for the request lifetime. Caches
+	 * the empty (no-gates) result too, which matters for the Premium Newsletters
+	 * cron loop that evaluates the same post for many readers.
+	 *
+	 * @var array<int,array>
+	 */
+	private static $post_gates_map = [];
+
+	/**
+	 * Request-scoped cache of a hierarchical term's descendant IDs, keyed by
+	 * "{taxonomy}:{term_id}". The term hierarchy is request-stable, so this avoids
+	 * re-walking the tree across content rules and across get_post_gates() calls.
+	 *
+	 * @var array<string,int[]>
+	 */
+	private static $term_descendants_map = [];
+
+	/**
 	 * Post meta key for exempting a post from access control restrictions.
 	 *
 	 * @var string
@@ -121,6 +141,9 @@ class Content_Restriction_Control {
 		if ( ! $post_id ) {
 			return [];
 		}
+		if ( array_key_exists( $post_id, self::$post_gates_map ) ) {
+			return self::$post_gates_map[ $post_id ];
+		}
 		$is_newsletter = false;
 		if ( class_exists( 'Newspack\Newsletters\Subscription_Lists' ) && Subscription_Lists::CPT && get_post_type( $post_id ) === Subscription_Lists::CPT ) {
 			$is_newsletter = true;
@@ -128,7 +151,8 @@ class Content_Restriction_Control {
 
 		$gates = Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish', $is_newsletter );
 		if ( empty( $gates ) ) {
-			return [];
+			self::$post_gates_map[ $post_id ] = [];
+			return self::$post_gates_map[ $post_id ];
 		}
 
 		$post_gates = [];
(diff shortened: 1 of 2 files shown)


plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -206,6 +211,36 @@ public static function get_post_gates( $post_id = null ) {
...
+	 * behavior. Expansion happens at evaluation time so newly-added child terms
+	 * are covered without re-saving the rule. Non-hierarchical taxonomies (e.g.
+	 * tags) have no descendants and are returned as integer-cast IDs unchanged.
+	 *
+	 * @param array        $term_ids Term IDs from a content rule's value.
+	 * @param \WP_Taxonomy $taxonomy Taxonomy object the term IDs belong to.
+	 *
+	 * @return int[] De-duplicated term IDs including descendants.
+	 */
+	private static function expand_hierarchical_terms( $term_ids, $taxonomy ) {
```
dkoo: **Suggestion:** Memoise hierarchical expansion within a request — `is_post_restricted()` is invoked many times per singular page render — directly from `Content_Gate::is_post_restricted()` (metering, gifting, comment filters, restricted-excerpt rendering, the gate output path) and once per restricted list per queued user inside the Premium Newsletters cron (`class-premium-newsletters.php:321`). Each call now fans out into `wp_get_post_terms()` + `get_term_children()` per taxonomy rule term.

`get_term_children()` is option-cached at the hierarchy layer, but it still calls `get_term()` per descendant to recurse — deep category trees aren't free. The existing `$post_gate_id_map` only memoises a *positive* restriction result inside `is_post_restricted()`; it does not short-circuit `get_post_gates()` and never memoises the negative case.

Two cheap wins:

1. A `static $memo = []` keyed by `$taxonomy->name . ':' . $term_id` inside `expand_hierarchical_terms()` — eliminates redundant recursion across rules and across calls.
2. A request-scoped `[$post_id => $gates]` static cache around `get_post_gates()` — mirrors the `$post_gate_id_map` pattern further down and is the bigger lever for the cron loop.

Neither needs invalidation logic — both are request-scoped.

**Consider:** Memoise `expand_hierarchical_terms()` per (taxonomy, term_id) for the request lifetime; ideally also memoise `get_post_gates()` per `$post_id` (mirroring the `$post_gate_id_map` pattern in `is_post_restricted()`), so the cron loop in `Premium_Newsletters::check_access()` does not repeat the term-tree walk per user per list.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -189,7 +189,12 @@ public static function get_post_gates( $post_id = null ) {
 					if ( ( ! $is_exclusion && ! $terms ) || is_wp_error( $terms ) ) {
 						continue 2;
 					}
-					if ( $is_exclusion ? ! empty( array_intersect( $terms, $content_rule['value'] ) ) : empty( array_intersect( $terms, $content_rule['value'] ) ) ) {
+					// For hierarchical taxonomies, a rule targeting a term also covers
+					// that term's descendants, mirroring WooCommerce Memberships' cascade.
+					// The helper also casts the rule's term IDs to integers, matching the
+					// integer IDs from wp_get_post_terms() on the other side of the intersect.
+					$target_terms = self::expand_hierarchical_terms( $content_rule['value'], $taxonomy );
```
dkoo: **Nit:** Acknowledge the int-cast as a quiet behaviour fix — The inline comment frames `array_map( 'intval', (array) $term_ids )` as a defensive mirror of `wp_get_post_terms( ..., 'fields' => 'ids' )`. In practice the test suite stores some rule IDs as strings (e.g. `(string) $page_id` in the specific_posts tests), and `array_intersect()` falls back to string-equivalence — so the new `intval` is actually doing real normalisation work, not just defence. It's almost certainly correct and harmless, but worth either reflecting in the comment or adding a one-line test that stores a stringified term ID in a category rule's `value` and confirms it still gates.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
dkoo: **Nit:** Optional: type the new helper signature — Non-blocking — typed signatures aren't enforced repo-wide yet. The docblock already says `int[]` and `\WP_Taxonomy`, but the signature is untyped:

```php
private static function expand_hierarchical_terms( $term_ids, $taxonomy ) {
```

If you want to pin it:

```php
private static function expand_hierarchical_terms( array $term_ids, \WP_Taxonomy $taxonomy ): array {
```

The function already dereferences `$taxonomy->hierarchical` / `$taxonomy->name`, so requiring the object (not the slug) is the right contract — typing the signature makes that explicit and lets static analysis catch a future caller that passes a slug. Take it or leave it.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -313,6 +313,137 @@ public function test_content_rules() {
...
+				'name'     => 'Other Category',
+			]
+		);
+
+		// Posts assigned only to a descendant term, never directly to the parent.
+		$parent_post     = $this->factory->post->create( [ 'post_category' => [ $parent_cat ] ] );
+		$child_post      = $this->factory->post->create( [ 'post_category' => [ $child_cat ] ] );
+		$grandchild_post = $this->factory->post->create( [ 'post_category' => [ $grandchild_cat ] ] );
+		$other_post      = $this->factory->post->create( [ 'post_category' => [ $other_cat ] ] );
+		$this->post_ids  = array_merge( $this->post_ids, [ $parent_post, $child_post, $grandchild_post, $other_post ] );
```
dkoo, line 358: **Nit:** Call `reset_restriction_cache()` between scenarios in `test_content_rules_hierarchical_child_terms` — The test runs three sequential rule configurations (inclusion targeting parent, exclusion targeting parent, inclusion targeting child) and calls `get_post_gates()` between them. Today `get_post_gates()` does not consult `$post_gate_id_map` so there's no false cache hit. Other multi-scenario tests in this file (institutional-access for example) do call `reset_restriction_cache()` between scenarios — adding the call here matches the convention and future-proofs the test if the caching strategy in `get_post_gates()` ever changes (see the memoise suggestion).

dkoo (approved):
Approve with comments — no blockers. The fix is small, well-targeted, and the test coverage hits the cases that matter (parent rule + child-only post, exclusion symmetry, tag passthrough, the one-directional guarantee). Two things worth a close look before merge: (1) on **deploy**, every existing gate whose rule targets a parent category will silently widen — confirm the per-site audit promised in the rollout note runs and surfaces newly-captured posts; (2) `expand_hierarchical_terms()` is invoked inside a hot loop that runs on every restricted page render, so memoising the helper (and `get_post_gates()` per-post) costs nothing and avoids repeating `get_term_children()` per call. The other comments are cleanups: an empty-value early-continue mirroring the `post_types` guard, an optional typed signature on the new helper, and a small test-hygiene nit. There is a latent Premium Newsletters cron path (`class-premium-newsletters.php:321`) that converts an `is_post_restricted()` true into an ESP unsubscribe — no Newspack site uses it in production yet, but it's worth keeping in mind for whichever publisher adopts newsletter gating first.

Before deploy, run the per-site audit promised in the rollout note: enumerate every active gate's content-rule taxonomy terms, expand them through `get_term_children()`, and list any posts newly captured by the cascade. (Newsletter-list CPTs feed an ESP unsubscribe path that isn't in production use today, but worth including the list in the audit so it's ready the day a publisher turns newsletter gating on.)

- `plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php:184` — **Suggestion:** Early-continue when `$content_rule['value']` is empty — The `post_types` branch a few lines above already short-circuits an empty rule value. The taxonomy branch does not: an empty inclusion rule still runs `get_taxonomy()` + `wp_get_post_terms()` + the new `expand_hierarchical_terms()`, then `array_intersect()` against `[]`, then falls through `continue 2` — same outcome, more work. ```php if ( empty( $content_rule['value'] ) ) { continue 2; } ``` placed just before the `get_taxonomy()` call makes the intent explicit and avoids the unnecessary lookups — meaningful on the Premium Newsletters per-user-per-list cron path. **Consider:** Early-continue in `get_post_gates()` when `$content_rule['value']` is empty (mirroring the `post_types` empty-value guard at line 175), so the taxonomy branch skips `get_taxonomy()` + `wp_get_post_terms()` for a rule that can never match.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
adekbadek: Done in a9de7c548. `get_post_gates()` is now memoized per post (caching the empty result too) and `expand_hierarchical_terms()` per `{taxonomy}:{term_id}` for the request, so the Premium Newsletters cron loop walks each term tree once across readers. Both caches are request-scoped (no invalidation needed) and are cleared by `reset_restriction_cache()`.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
adekbadek: Done in a9de7c548. Added a test that stores a stringified parent term ID in the category rule's value and asserts it still cascades to gate the child post, and reworded the call-site comment to note the `intval` is load-bearing normalization (string rule values vs integer post-term IDs in `array_intersect()`), not merely defensive.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
adekbadek: Done in a9de7c548. Signature is now `( array $term_ids, \WP_Taxonomy $taxonomy ): array`; the call site passes `(array) $content_rule['value']` so the prior defensive cast is preserved.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
adekbadek: Done in a9de7c548. Added `reset_restriction_cache()` between scenarios here and in `test_content_rules` (now required since `get_post_gates()` is memoized), and added the two new caches (`post_gates_map`, `term_descendants_map`) to the reset helper.

PR #286, opened 2026-06-11
fix(newsletters): upgrade DataViews to 16 and fix list styling

thomasguillot:
Linear: [DSGNEWS-183](https://linear.app/a8c/issue/DSGNEWS-183/newsletters-admin-shell-dataviews-lists-render-outdated-styling)

## Summary

Upgrades the Newsletters admin-shell DataViews lists to `@wordpress/dataviews@16.0.0` and fixes the resulting list styling.

The admin-shell bundle was compiling `@wordpress/dataviews@10.3.0` (DEWP bundles dataviews rather than externalizing it). That older version drew a stray `<hr>` separator in the row **Actions** kebab and used an older table column-sizing model. Bumping to 16.0.0 removes the separator and matches core's current DataViews behavior. The bump is scoped to this plugin — the workspace root stays on 10.x, and webpack resolves the plugin's own 16.0.0 copy.

## Changes

- **`package.json`** — add `@wordpress/dataviews: ^16.0.0` devDependency (plus the matching `pnpm-lock.yaml` entry). Removes the stray Actions-menu separator.
- **`style.scss`**
  - Add `.dataviews-view-table col.dataviews-view-table__col-first-data { width: auto !important; }` to re-assert the primary column as flexible. newspack-plugin's **global** `commons.css` still ships dataviews 10.x CSS whose `col:not(.dataviews-view-table__col-primary){width:1%}` rule wrongly matches 16.x's renamed `__col-first-data` column, which otherwise left no column to absorb table slack.
  - Remove two overrides now redundant on 16.x (a `.components-base-control` margin reset and a pagination select `line-height`), each verified dead by toggling live in the browser.
  - Reorganize the file into logical sections, add two SASS mixins (`inline-icon`, `flex-column-fill`), and nest the previously-flat `.newspack-newsletters-quick-edit-modal__*` selectors under one block.

## Testing

- Verified the SCSS refactor compiles to **equivalent** CSS — identical declaration count (2914) and rule count (1707) before/after.
- `lint:scss` passes.
- Browser-checked all five DataViews screens (Newsletters, Ads, Advertisers, Layouts lists + the Layout picker modal): separator gone, column widths correct, Quick Edit slide-in panel and reduced-motion behavior intact.
- Reviewed by Codex and Claude — 0 findings.


plugins/newspack-newsletters/src/admin-shell/style.scss
```diff
@@ -427,5 +267,141 @@ body.newspack-newsletters-admin-screen--bundled .newspack-newsletters-admin-moun
...
+@media (prefers-reduced-motion: reduce) {
+	.newspack-newsletters-quick-edit-modal__overlay,
+	.components-modal__frame.newspack-newsletters-quick-edit-modal,
+	.components-modal__screen-overlay.newspack-newsletters-quick-edit-modal__overlay.is-animating-out .components-modal__frame {
+		animation: none;
+	}
+}
+
+// newspack-plugin's global commons bundle ships dataviews 10.x CSS that wrongly pins __col-first-data to 1%.
+.dataviews-view-table col.dataviews-view-table__col-first-data {
```
adekbadek, line 387: **Suggestion:** The `width: auto !important` on `__col-first-data` counters a `col[...]:not(...){width:1%}` rule loaded globally by newspack-plugin's `commons.css` (still on dataviews 10) — and, as it turns out, also shipped by dataviews 16's own bundled stylesheet via its legacy `__col-primary` selector, so this override is needed even standalone. `!important` is the only reliable win here, since enqueue order across two independently-versioned plugins isn't controllable — so the approach is sound. It also fails safe: once newspack-plugin ships dataviews 16, its rule stops matching `__col-first-data` and this override degrades to a harmless no-op rather than breaking. Worth opening a follow-up to realign newspack-plugin's dataviews version and drop this override then, referenced from the inline comment, so it doesn't silently outlive its purpose.

plugins/newspack-newsletters/package.json
```diff
@@ -49,6 +49,7 @@
 		"qs": "^6.15.2"
 	},
 	"devDependencies": {
+		"@wordpress/dataviews": "^16.0.0",
```
adekbadek, line 52: **Nit:** `@wordpress/dataviews` is compiled into the shipped `dist/admin-shell.js` (imported via the `/wp` subpath across the five screens), so it's semantically a runtime dependency, yet it's declared under `devDependencies`. Defensible — it matches the repo's existing pattern (bundled `lodash` also sits in `devDependencies`) and this plugin ships `dist/` and is never installed as a consumer package — so purely a consistency nit, not a correctness issue.

adekbadek (approved):
Approve. But is there a plan to bump to 16 for the whole repo? I'm not a fan of per-plugin divergence here.

thomasguillot:
Good call on the divergence. Tracked in [DSGNEWS-195](https://linear.app/a8c/issue/DSGNEWS-195/realign-newspack-plugin-dataviews-to-16-and-drop-newsletters-admin) to bump newspack-plugin (and the rest of the repo) to dataviews 16 and drop this override once that lands — at which point the `col-first-data` rule stops matching and it's a no-op anyway.

Added a plain `// TODO` next to the override so it doesn't silently outlive its purpose.

PR #287, opened 2026-06-11
feat(colors)!: redefine neutral 700/800, add overlay token system

thomasguillot:
### All Submissions:

Propagates the updated Newspack Design System neutral + overlay colors across the workspace.

**Neutral ramp:**
- Neutral 700 `#000000b3` (was black at 70% opacity) → `#5d5d5d` (solid) — now a dark-mode border token, not an opacity value. This also restores monotonic ordering of the neutral scale on white.
- Neutral 800 `#3e3e3e` → `#393939`

**New `overlay-*` token system** — a digital-only "black at stepped opacity" scrim ramp. Added to `newspack-colors` (`$overlay-000…$overlay-1000` plus rounded `--rgba` companions via a new `to-rgba()` helper), to the Newspack UI custom properties (`--newspack-ui-color-overlay-0…-100`, ÷10 naming mirroring neutrals), and to the block theme `theme.json` (`--wp--custom--color--overlay-*`).

**Scrim migration:** every place that was using `neutral-70` as a translucent scrim — Newspack UI modals, the overlay menu/search blocks, the block theme overlay mask and Jetpack instant-search overlay, post-carousel caption gradients, the republication-tracker modal shadow, and the newspack-blocks carousel arrows — now points at the `overlay-70` / `overlay-700` scrim token. This is value-preserving (`#000000b3`), and it means flipping Neutral 700 to a solid value no longer silently turns those surfaces into opaque gray.

**Theme palettes:** `contrast-2` (block theme + style variations) and the classic theme `secondary-variation` realigned to the new neutral values.

#### ⚠️ BREAKING CHANGE

`newspack-colors` is bumped to **2.0.0**. The exported `neutral-700` token changes both value (`#000000b3` → `#5d5d5d`) and meaning (was a 70% black overlay, now a solid border). Downstream consumers that relied on `neutral-700` for translucency must migrate to the new `overlay-700` token. All in-repo consumers are migrated in this PR.

> Reviewed via a Codex + Claude multi-reviewer loop (3 rounds, converged to 0 findings).

1. Build the affected units: `n build newspack-plugin`, `n build newspack-blocks`, `n build newspack-block-theme`. All should compile cleanly.
2. **Modals:** open a Newspack UI modal (e.g. the reader-activation auth modal). Confirm the backdrop scrim is the usual ~70% black translucency (content dimmed but visible behind), not a solid gray.
3. **Overlay menu / search:** trigger the fullscreen overlay menu and the overlay search block. Confirm the background is the translucent 70% scrim, not opaque gray.
4. **Block theme overlays:** open the mobile overlay menu (overlay mask) and, if Jetpack Search is active, the instant-search overlay. Confirm both scrims remain translucent.
5. **Carousel arrows (newspack-blocks):** on a homepage carousel, hover the prev/next arrows (translucent darken), then keyboard-tab to an arrow. Confirm a clearly visible focus ring (white inner stroke) over both light and dark slide images — this is the accessibility fix.
6. **Post-carousel (block theme):** confirm the caption gradient overlay still reads over images.
7. **Editor palette:** in the block theme, open a color picker. Confirm "Contrast / Two" is `#5D5D5D` and the new `overlay-*` custom colors are available.
8. **Dark style variations:** switch to a dark style variation and confirm `neutral-70`/`-80` borders and fills use the new solid values.
9. **Republication tracker (classic theme):** open the republication modal and confirm its drop shadow renders.

### Other information:


packages/colors/colors.module.scss
```diff
@@ -130,8 +136,8 @@ $neutral-300: #ddd;
 $neutral-400: #ccc;
 $neutral-500: #949494;
 $neutral-600: #6c6c6c;
-$neutral-700: #000000b3;
-$neutral-800: #3e3e3e;
+$neutral-700: #5d5d5d;
```
dkoo, line 139: **Suggestion:** Standalone-repo siblings still consume `--wp--custom--color--neutral-70` as a scrim — The four legacy standalone repos (`Automattic/newspack-plugin`, `newspack-blocks`, `newspack-block-theme`, `newspack-theme`) still reference `--wp--custom--color--neutral-70` / `--newspack-ui-color-neutral-70` / `$neutral-700` as a translucent scrim — verified by grep against the `repos/` checkouts in this workspace:\n\n- `repos/newspack-plugin/src/blocks/overlay-menu/style.scss:157`\n- `repos/newspack-plugin/src/blocks/overlay-search/style.scss:25`\n- `repos/newspack-block-theme/src/scss/blocks/newspack/_post-carousel.scss:18`\n- `repos/newspack-block-theme/src/scss/overlays/_base.scss:15`\n- `repos/newspack-blocks/src/blocks/carousel/view.scss:185`\n- `repos/newspack-theme/newspack-theme/sass/plugins/republication-tracker-tool.scss`\n\nLow risk in practice: `newspack-block-theme` is in a controlled release state, so the update order between the block theme (which redefines this CSS-custom-property as solid `#5d5d5d`) and the plugins that consume it as a scrim is managed at release time. Worth a follow-up to migrate the standalone-repo references to `overlay-70` so the new ramp is fully adopted and the legacy contract can eventually be retired.

packages/colors/colors.module.scss
```diff
@@ -149,6 +155,34 @@ $neutral-800--rgb: to-rgb($neutral-800);
...
+$overlay-100: #00000026;
+$overlay-200: #00000033;
+$overlay-300: #0000004d;
+$overlay-400: #00000066;
+$overlay-500: #00000080;
+$overlay-600: #00000099;
+$overlay-700: #000000b3;
+$overlay-800: #000000cc;
+$overlay-900: #000000e6;
+$overlay-1000: #000000f2;
```
dkoo, line 172: **Suggestion:** `overlay-1000` is ~95% alpha, not fully opaque — `$overlay-1000: #000000f2` resolves to alpha **0.949** (`242/255`), which `to-rgba()` rounds to `rgba(0, 0, 0, 0.95)`. The ramp tops out below opaque on purpose (mirrored by an `overlay-000` floor of ~5%), but a consumer reaching for `overlay-1000` expecting a fully-opaque scrim — by analogy with `neutral-1000: #000` — will get something subtly translucent. Either keep the ramp as-is and document the floor/ceiling intent in a comment near the `// Overlay: Hex` block, **or** make `overlay-1000` actually opaque (`#000000ff`) so the high end of the ramp matches `neutral-1000`. Not a bug, but the asymmetry between `neutral-1000 = #000` and `overlay-1000 ≈ 95% black` is a footgun worth either a code comment or a value tweak.

packages/colors/colors.module.scss
```diff
@@ -149,6 +155,34 @@ $neutral-800--rgb: to-rgb($neutral-800);
...
+$overlay-100--rgba: to-rgba($overlay-100);
+$overlay-200--rgba: to-rgba($overlay-200);
+$overlay-300--rgba: to-rgba($overlay-300);
+$overlay-400--rgba: to-rgba($overlay-400);
+$overlay-500--rgba: to-rgba($overlay-500);
+$overlay-600--rgba: to-rgba($overlay-600);
+$overlay-700--rgba: to-rgba($overlay-700);
+$overlay-800--rgba: to-rgba($overlay-800);
+$overlay-900--rgba: to-rgba($overlay-900);
+$overlay-1000--rgba: to-rgba($overlay-1000);
```
dkoo, line 186: **Nit:** Consider exporting the `overlay-*--rgba` companions on `:export` — The new `$overlay-*--rgba` Sass variables are defined (lines 173–184) but not added to the `:export {}` block (lines 223–325), so JS/CSS-Modules consumers only get the hex form (e.g. `overlay-700: #000000b3`). The hex form is fine for `background-color` / `border-color` use, but if a JS consumer ever wanted to interpolate the rounded `rgba(0, 0, 0, 0.7)` form (e.g. for `box-shadow` strings in JSX), they'd need to re-derive it. Cheap to add now; awkward to add later because consumers will have written their own conversion.

plugins/newspack-plugin/docs/insights/component-design-spec.md
```diff
@@ -56,16 +56,16 @@ Two-namespace convention with documented rules:
 - `$alert-yellow` — StalePill base color (via color-mix)
 
 **SERIES_PALETTE** for multi-series charts (LineChart, PieChart):
-1. `primary-500` — anchor blue
+1. `primary-600` — anchor blue
```
dkoo: **Nit:** SERIES_PALETTE doc neither matches the prior code nor the current code — The doc lists token names (`primary-600`, `quaternary-700`, `secondary-700`, `tertiary-700`, `neutral-600`, `primary-300`), but `src/wizards/insights/tabs/components/_insights-charts.scss` hardcodes the series colors literally (`#1c5ea8`, `#3eb24f`, `#d98b27`, `#8a4fbd`, `#c0392b`, `#1f8a8a`, `#6b7785`) — and those hex values map to **none** of the listed tokens in either the old or the new doc text. This drift predates the PR (the old `primary-500` claim was wrong too), so the edit doesn't make anything worse — but updating the token names while the code is hex-literal entrenches the mismatch. Either align the SCSS to actually use the colors module, or trim the doc to describe what's really there.

dkoo (approved):
Manual testing passes! Approve with comments — no blockers. In-repo migration is clean and value-preserving (`#000000b3` everywhere); the breaking change is correctly bumped to `newspack-colors@2.0.0` and the new `theme.json` overlay-* custom properties land at the document root exactly as expected (browser-verified). A handful of sibling 50%-black scrims (modal-checkout, homepage-articles, carousel captions) are exact `overlay-500` candidates left un-migrated — worth folding in if scope allows, otherwise a clean follow-up. `overlay-1000` caps at ~95% alpha rather than fully opaque, which is asymmetric with `neutral-1000: #000` and worth either a code comment or a value tweak. Cross-repo coordination with the standalone sibling repos that still consume `--wp--custom--color--neutral-70` as a scrim is noted as a low-risk suggestion — the block-theme release state controls the update order, so the harm scenario is theoretical in practice. A brief release note about the `contrast-2` preset repainting from `#3E3E3E` to `#5D5D5D` on existing block content would help publishers. Two minor doc/export polish nits round it out.

File a follow-up to migrate the standalone-repo scrim references (overlay-menu / overlay-search / post-carousel / carousel / republication-tracker) to the new `overlay-70` token, so the new ramp is fully adopted and the legacy `--wp--custom--color--neutral-70` contract can eventually be retired.

Add a one-line release note for publishers: any block content using the `contrast-2` preset slug (`has-contrast-2-color` / `-background-color` CSS classes) will repaint from `#3E3E3E` to `#5D5D5D` after the block-theme update — visible on already-published pages.

- `plugins/newspack-blocks/src/blocks/carousel/view.scss:57` — **Suggestion:** Carousel caption scrim still hand-built at 50% black — Lines 57 and 167 use `rgba(np-colors.$neutral-1000, 0.5)` — same value as `$overlay-500`. The same file already adopts `$overlay-700` for the arrow hover/focus background on line 185, so the partial migration is inconsistent (both sites are scrims on top of imagery).
- `plugins/newspack-blocks/src/blocks/homepage-articles/view.scss:303` — **Suggestion:** Homepage articles overlay scrim left un-migrated — `background: rgba(np-colors.$neutral-1000, 0.5)` — another exact `$overlay-500` candidate, semantically a card-image scrim like the carousel one above.
- `plugins/newspack-blocks/src/modal-checkout/modal.scss:33` — **Suggestion:** Modal checkout scrim is a natural `overlay-500` candidate — `background: rgba(np-colors.$neutral-1000, 0.5)` is exactly `$overlay-500` (`#00000080`). Migrating it would let the modal-checkout backdrop join the same token-driven system as the newspack-ui modal (which the PR migrates to `--newspack-ui-color-overlay-70`). If the PR's intent is to centralize all stepped scrims under the new ramp, this is the most prominent miss; if scope was strictly 70% scrims, ignore.
- `plugins/newspack-plugin/src/plugins-screen/plugins-screen.scss:24` — **Nit:** Admin plugins-screen loading scrim uses raw `rgba` — `background: rgba(0, 0, 0, 0.5)` on the loading overlay. This file is admin-context, where the design system would normally route through `--newspack-ui-color-overlay-50` (mapped from `$overlay-500`). Low priority — it's a transient admin overlay — but it's a token gap the new ramp now closes cleanly.

packages/colors/colors.module.scss
thomasguillot: The monorepo is the source of truth here — the standalone `repos/` checkouts are superseded by it, not an independent shipping target, so there's no separate migration to track. Every in-monorepo scrim is already on `overlay-70`.

packages/colors/colors.module.scss
thomasguillot: Kept the value and documented the intent: added a comment on the overlay block noting the ramp deliberately spans 5%–95% alpha (never fully transparent or opaque), and that solid black is `$neutral-1000`. The asymmetry with `neutral-1000` is by design — an `overlay-*` token is always a translucent scrim; opaque black is `neutral-1000`.

packages/colors/colors.module.scss
thomasguillot: Keeping `:export` hex-only. No `$*-rgb` companion is exported today — they're Sass-only by convention — so exporting just the overlay rgba forms would break that uniformity. No JS consumer reads an overlay or rgb key, and 8-digit hex already drops straight into a `box-shadow` string, so there's nothing to interpolate yet. I'll export the full companion family if a real consumer shows up.

plugins/newspack-plugin/docs/insights/component-design-spec.md
thomasguillot: Reverted the palette-slug edit (back to pre-PR), so this PR no longer touches that section. The series colors are intentionally hardcoded hex in `_insights-charts.scss` — the brand palette has no distinct token for the categorical teal/red hues — so the doc-vs-code drift is pre-existing and out of scope here. The footgun line keeps the required `#5d5d5d` update.

thomasguillot:
Addressed the out-of-line scrim suggestions in 2d92313:

- **modal-checkout backdrop** + **carousel caption** → `$overlay-700` (caption darkens 50%→70%, matching the newspack-ui modal scrim).
- **carousel arrow base** + **homepage-articles card scrim** → `$overlay-500` (the arrow keeps its hover/focus delta to `$overlay-700`, so the 50%→70% hover affordance is preserved).
- **admin plugins-screen modal scrim** → `rgba($black, 0.35)` from `@wordpress/base-styles`, matching WordPress's own `.components-modal__screen-overlay` for that admin context rather than introducing a newspack token there.

PR #291, opened 2026-06-12
feat(insights): empty-state pattern for all-zero sections (NPPD-1694)

kmwilkerson:
## Summary

The empty-state pattern for Gates tab sections that would otherwise render as a row of zeros, landing on `feat/insights-rsm`. Closes [[NPPD-1694](https://linear.app/a8c/issue/NPPD-1694)](https://linear.app/a8c/issue/NPPD-1694). It adds two reusable pieces, an `<EmptyMetricSection>` shared component and a `zeroFallback` count fallback on `MetricCard`, and wires the first and only consumer in this PR: the Gates tab's Paid reader conversion section. The trigger was test pubs 90-day window rendering Paid conversion as `0%`, `0%`, `$0.00`, `$0.00`, correct numbers (17 paywall attempts, zero conversions in the window) that read as a broken dashboard. That section now renders an honest, lightly diagnostic empty state, and individual zero cards read as `0 of 17` or `0 conversions` instead of `0%` or `$0.00`.

The Free reader conversion section is deferred to [[NPPD-1702](https://linear.app/a8c/issue/NPPD-1702)](https://linear.app/a8c/issue/NPPD-1702). It needs a hub-side BigQuery change (the regwall rates are a precomputed column with no underlying counts). The Paid side is fully self-contained and needs no hub changes.

### Pre-flight decisions

Surfaced before writing code, where the codebase disagreed with the original plan.

- Built on the current design-system primitives. This branch was reconciled onto the latest `feat/insights-rsm` after the design-system refactors landed. `<EmptyMetricSection>` renders its identity via the shared `SectionHeading` (the `SectionHeader` wrapper) and its callout via a non-dismissible `@wordpress/components` `Notice`, the same primitive `InfoCallout` now uses, rather than hand-rolled markup. The `zeroFallback` rendering re-fits onto the `Card`-based `MetricCard` with no change to its value or secondary regions.
- Free section deferred ([[NPPD-1702](https://linear.app/a8c/issue/NPPD-1702)](https://linear.app/a8c/issue/NPPD-1702)). The paywall numerator is computed locally via the Woo-order join, so the Paid section is buildable end to end here. The regwall (Free) rates come from a precomputed hub column with no underlying counts, and the BigQuery proxy only names an allowlisted query, so the SQL lives on the Newspack Manager hub. Rather than regress the working regwall percentages, Free lands with that hub change.
- `conversions_count` reuses the existing `denominator` field. The orchestrator already passes the conversions count into the envelope's `denominator` for both currency scorecards, so there is no redundant field. The only genuinely missing rate field was `numerator`, which was computed then discarded.
- The currency-total rule follows the ticket table. A currency-total card shows `0 conversions` when conversions equals 0 and attempts are greater than 0.
- `MetricCard` has no `'rate'` format. Rate scorecards render as `format='percent'`, and total-vs-average is not encoded on the card, so `zeroFallback` keys behavior off `format` plus a `currencyRole` discriminator plus the numerator and denominator counts.
- `zeroFallback` is built in the `scalarToCard` adapter path, not inline on `<MetricCard>`, since every Gates scorecard already flows through `scalarToMetricCardProps`.
- Error-state guard. The Paid section treats a zero total as a genuine empty state only when the Direct scalar actually computed (`state !== 'error'`). An errored Direct query leaves the totals at 0, which would otherwise mis-render as "no paywall attempts" instead of surfacing the card's error treatment.
- The null glyph is matched exactly to the Performance-by-gate table's null cell (same glyph and `aria-label`); counts use the app's shared `formatNumber`.

### `<EmptyMetricSection>`, shared component

`tabs/components/EmptyMetricSection.tsx` (new). `SectionHeading` (title plus caption) over a non-dismissible info `Notice`. The `{N}` body interpolation substitutes a `formatNumber`-formatted `signalCount` only when supplied and the placeholder is present; a stray `{N}` is left visible so a missing prop is caught in QA. Distinct from `SectionEmpty`, the plain "no rows" paragraph for collection metrics.

### `MetricCard` count fallback

`tabs/components/MetricCard.tsx` gains a `zeroFallback` prop, keyed on two counts (`denominator` = attempts, `numerator` = conversions) rather than on `value`. Rendering matches the ticket table: rate shows `0 of 17`; currency-total shows `0 conversions`; currency-average shows the null glyph plus "No conversions in this window"; and any zero opportunity count shows the null glyph plus "No {attempts} in this window". The comparison delta is suppressed under any fallback hero. Labels are pre-pluralized translatable nouns with no runtime string-munging.

### Gates envelope additions (PHP and TS)

`numerator` now rides alongside `denominator` on rate scalars, threaded from the already-computed paywall Woo join (null where the count is not computed locally). A pure, unit-tested `Gates_Metric::paywall_section_totals()` derives `paywall_attempts_total` (Direct denominator) and `paywall_conversions_total` (inclusive max across attributions) with no extra query. `is_window_all_error()` skips the new scalar totals. Mirrored in `api/gates.ts` and `gates-fixture.php`.

### Paid section, first consumer

`tabs/gates/PaidReaderConversionSection.tsx` detects state from the section totals in the component, not the orchestrator: `attempts === 0` gives `no_opportunity`; `conversions === 0` gives `no_conversions` (with `signalCount`); otherwise the four scorecards, each carrying its `zeroFallback`.

## How to test

1. Check out `feat/insights-empty-state` and run `n build newspack-plugin`.
2. Fixture mode (no BigQuery): with `NEWSPACK_INSIGHTS_FIXTURE_MODE` on, visit Insights → Gates. `?_fixture_state=empty` renders the Paid section's `no_opportunity` empty state; the default `populated` variant renders the four scorecards with the count fallback wired.
3. Test pub, 90-day (real data): the Paid reader conversion section renders the `no_conversions` empty state with `{N}` = 17.
4. A publisher with normal paywall data renders the four scorecards; an individual zero card shows `0 of N` or `0 conversions` rather than `0%` or `$0.00`.
5. `n test-php --filter Test_Gates_Metric`: 39 tests, 145 assertions green (existing plus 6 new). PHPCS clean.
6. `tsc --noEmit`, ESLint, the production build, and the 29 React tests across `EmptyMetricSection`, `MetricCard`, and `PaidReaderConversionSection` all pass. The `.tsx` tests are now collected by Jest ([[NPPD-1683](https://linear.app/a8c/issue/NPPD-1683)](https://linear.app/a8c/issue/NPPD-1683)'s `testMatch` fix has landed on this base).

## Heads-up

- Free reader conversion is a tracked follow-up ([[NPPD-1702](https://linear.app/a8c/issue/NPPD-1702)](https://linear.app/a8c/issue/NPPD-1702)). It needs the hub to return registration-impression and registration counts on the regwall queries. The in-repo wiring mirrors the Paid side and lands with the hub change.
- The `configuration_missing` `EmptyMetricSectionState` is defined but currently unused on Gates; no orchestrator path emits a "gates not configured" signal yet.

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/fixtures/gates-fixture.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/fixtures/gates-fixture.php
@@ -236,6 +236,34 @@ return function ( string $variant = 'populated', bool $compare = false ): array 
 		];
 	};
 
+	// --- Paid-section empty-state smoke variants (NPPD-1694). ---
+	if ( 'paid_no_conversions' === $variant ) {
+		// 17 paywall attempts, zero conversions → the section's `no_conversions`
+		// empty state with {N} = 17 (Richland Source's live scenario).
+		$current                              = $build( 1.0 );
+		$current['paywall_attempts_total']    = 17;
+		$current['paywall_conversions_total'] = 0;
+		return [
+			'tab_error' => false,
+			'current'   => $current,
+			'previous'  => null,
+		];
+	}
+	if ( 'paid_zero_cards' === $variant ) {
+		// Section has data (Influenced converted) so the grid renders, but the
+		// Direct cards are zero → exercises the per-card count fallback:
+		// "0 of 320" (rate), "0 conversions" (currency total), "—" (currency avg).
+		$current                                       = $build( 1.0 );
+		$current['paywall_conversion_direct']          = $scalar( 0.0, 320, 'rate', 0 );
+		$current['total_paywall_revenue_direct']       = $scalar( 0.0, 0, 'currency' );
+		$current['avg_revenue_per_paywall_conversion'] = $scalar( 0.0, 0, 'currency' );
+		return [
+			'tab_error' => false,
+			'current'   => $current,
+			'previous'  => null,
+		];
+	}
+
 	return [
 		'tab_error' => false,
 		'current'   => $build( 1.0 ),
--- a/plugins/newspack-plugin/includes/wizards/insights/fixtures/gates-fixture.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/fixtures/gates-fixture.php
@@ -53,11 +53,15 @@ return function ( string $variant = 'populated', bool $compare = false ): array 
 				'value'            => $zero_value( $type ),
 				'computable'       => false,
 				'denominator'      => null,
+				'numerator'        => null,
 				'placeholder_type' => $type,
 				'error_code'       => 'bigquery_proxy_http_error',
 				'error_message'    => 'HTTP 500',
 			];
 		}
+		// Scalar section-total metadata (NPPD-1694) — zero in the all-error window.
+		$current['paywall_attempts_total']    = 0;
+		$current['paywall_conversions_total'] = 0;
 		$collections = [
 			'conversion_funnel'      => 'stages',
 			'exposures_distribution' => 'buckets',
@@ -87,9 +91,13 @@ return function ( string $variant = 'populated', bool $compare = false ): array 
 				'value'            => $zero_value( $type ),
 				'computable'       => false,
 				'denominator'      => 'rate' === $type ? 0 : null,
+				'numerator'        => 'rate' === $type ? 0 : null,
 				'placeholder_type' => $type,
 			];
 		}
+		// Zero attempts → the Paid section renders its no_opportunity empty state.
+		$current['paywall_attempts_total']    = 0;
+		$current['paywall_conversions_total'] = 0;
 		$current['conversion_funnel']      = [
 			'state'  => 'empty',
 			'stages' => [],
@@ -110,12 +118,13 @@ return function ( string $variant = 'populated', bool $compare = false ): array 
 	}
 
 	// --- Populated variant (default). ---
-	$scalar = static function ( $value, $denominator, string $type ) {
+	$scalar = static function ( $value, $denominator, string $type, $numerator = null ) {
 		return [
 			'state'            => 'populated',
 			'value'            => $value,
 			'computable'       => true,
 			'denominator'      => $denominator,
+			'numerator'        => $numerator,
 			'placeholder_type' => $type,
 		];
 	};
@@ -131,11 +140,16 @@ return function ( string $variant = 'populated', bool $compare = false ): array 
 			// Section 2 — Free reader conversion.
 			'regwall_conversion_direct'          => $scalar( round( 0.071 * $f, 3 ), null, 'rate' ),
 			'regwall_conversion_influenced_7d'   => $scalar( round( 0.123 * $f, 3 ), null, 'rate' ),
-			// Section 3 — Paid reader conversion.
-			'paywall_conversion_direct'          => $scalar( round( 0.021 * $f, 3 ), null, 'rate' ),
-			'paywall_conversion_influenced_14d'  => $scalar( round( 0.038 * $f, 3 ), null, 'rate' ),
+			// Section 3 — Paid reader conversion. Paywall rate cards carry both
+			// numerator (matched Woo orders) and denominator (attempts) so fixture
+			// mode exercises the NPPD-1694 count fallback.
+			'paywall_conversion_direct'          => $scalar( round( 0.021 * $f, 3 ), (int) round( 320 * $f ), 'rate', (int) round( 7 * $f ) ),
+			'paywall_conversion_influenced_14d'  => $scalar( round( 0.038 * $f, 3 ), (int) round( 290 * $f ), 'rate', (int) round( 11 * $f ) ),
 			'total_paywall_revenue_direct'       => $scalar( round( 4180.5 * $f, 2 ), (int) round( 47 * $f ), 'currency' ),
 			'avg_revenue_per_paywall_conversion' => $scalar( round( 88.95 * $f, 2 ), (int) round( 47 * $f ), 'currency' ),
+			// Section 3 empty-state totals (NPPD-1694).
+			'paywall_attempts_total'             => (int) round( 320 * $f ),
+			'paywall_conversions_total'          => (int) round( 11 * $f ),
 			// Section 4 — How readers convert. Funnel uses Richland Source's shape
 			// so both drop-off deltas exceed 20% and render in the error color.
 			'conversion_funnel'                  => [
--- a/plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
@@ -34,6 +34,7 @@ use Newspack\Insights\Woo_Order_Resolver;
  *   value: int|float,
  *   computable: bool,
  *   denominator: int|null,
+ *   numerator: int|null,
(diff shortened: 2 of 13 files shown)


kmwilkerson (commented):
Empty-state pattern for all-zero Gates scorecards — no blockers. The Paid reader conversion section now renders honest zeros (`0 of 17` / `0 conversions` / em-dash) and a clear `no_opportunity` / `no_conversions` callout instead of a row of `0%`/`$0.00`, built on the current design-system primitives (Card, SectionHeading, Notice). A couple of things worth a look before merge: a partial-query-failure edge where an errored Influenced query can be hidden by the `no_conversions` empty state, and the `@phpstan-type` typedef not listing the new `numerator` key; the remaining items are nits. Verified by browser-testing all five render states and running the Gates PHPUnit suite.

- `plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php:36` — **Suggestion:** @phpstan-type ScalarMetric typedef is missing the new numerator key — `error_scalar()` and `populated_scalar()` now always emit a `numerator` (int|null), but the `@phpstan-type ScalarMetric` definition at the top of the class still lists only `denominator`. Since this typedef is the documented contract for the scalar shape and adding `numerator` is the point of the change, add `numerator: int|null,` so static analysis covers the new field.

plugins/newspack-plugin/src/wizards/insights/tabs/gates/PaidReaderConversionSection.tsx
```diff
@@ -17,67 +24,141 @@ import { __ } from '@wordpress/i18n';
...
+
+	const attempts = current.paywall_attempts_total;
+	const conversions = current.paywall_conversions_total;
+
+	// The section totals are derived from the Direct scalar's denominator/numerator,
+	// which are null when that query errors — coercing both totals to 0. A zero total
+	// is only a *genuine* empty state when the Direct metric actually computed;
+	// otherwise we fall through to the scorecards so each card surfaces its own error
+	// treatment rather than a misleading "no paywall attempts" empty state.
+	const dataKnown = current.paywall_conversion_direct.state !== 'error';
```
kmwilkerson: **Suggestion:** **Suggestion:** Partial query failure can hide behind the no_conversions empty state — The empty-state guard only checks `paywall_conversion_direct.state`, but `paywall_conversions_total` is `max(direct.numerator, influenced.numerator)` and the Influenced query can fail independently of Direct. If Direct returns a real 0 conversions while the Influenced query errors (its null numerator coerces to 0 in `paywall_section_totals`), the section renders the `no_conversions` empty state and the errored Influenced card is never shown — the failure surfaces as a benign "no conversions" message instead of the card's error treatment. Consider also falling through to the scorecards when `paywall_conversion_influenced_14d.state === 'error'`, and add a test for the Direct-ok / Influenced-errored combination (the current suite covers Direct-errored only).

plugins/newspack-plugin/includes/wizards/insights/fixtures/gates-fixture.php
```diff
@@ -87,9 +91,13 @@
 				'value'            => $zero_value( $type ),
 				'computable'       => false,
 				'denominator'      => 'rate' === $type ? 0 : null,
+				'numerator'        => 'rate' === $type ? 0 : null,
```
kmwilkerson, line 94: **Nit:** **Nit:** Fixture empty-variant gives regwall rate cards numerator 0 vs production null — The empty variant sets `numerator => 'rate' === $type ? 0 : null`, so regwall rate cards get `numerator: 0`, whereas production regwall metrics (precomputed rates via `compute_metric_from_proxy`) always emit `numerator: null`. Harmless today (the regwall section passes no `zeroFallback`), but it's a fixture-vs-production divergence — if the UI ever distinguishes `numerator === null` from `0` for a regwall card, fixture mode wouldn't reproduce production. Consider aligning the empty-variant regwall numerators to `null`, or a one-line note.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
```diff
@@ -134,14 +135,20 @@ private function error_scalar( string $placeholder_type, \WP_Error $error ): arr
...
 	 * @param string    $placeholder_type One of 'count', 'rate', 'currency', 'decimal'.
+	 * @param int|null  $numerator        Optional numerator (NPPD-1694). Surfaced only
+	 *                                    for rate scorecards whose underlying count is
+	 *                                    computed locally (the paywall Woo join); null
+	 *                                    everywhere the count isn't available, e.g. the
+	 *                                    precomputed-rate regwall cards.
 	 * @return array
 	 */
-	private function populated_scalar( $value, bool $computable, ?int $denominator, string $placeholder_type ): array {
+	private function populated_scalar( $value, bool $computable, ?int $denominator, string $placeholder_type, ?int $numerator = null ): array {
```
kmwilkerson, line 146: **Nit:** **Nit:** populated_scalar() $value lacks a native type hint — While editing this signature to add `$numerator`, consider promoting `$value` to a native union type (`int|float $value`) to match the file's otherwise fully-typed signatures and the PHP 8.3 target — the `@param int|float $value` PHPDoc already states the intent. Pre-existing on the method, but this is a natural moment.

plugins/newspack-plugin/src/wizards/insights/tabs/components/MetricCard.tsx
```diff
@@ -109,7 +145,54 @@ const MetricCard = ( props: MetricCardProps ) => {
...
+					__( '%1$s of %2$s', 'newspack-plugin' ),
+					formatNumber( numerator ),
+					formatNumber( denominator )
+				);
+			} else if ( currencyRole === 'total' ) {
+				// Currency total → "0 conversions", symmetric with the rate card.
+				fallbackHero = sprintf(
+					/* translators: 1: conversions count (0 here), 2: plural "conversions". */
+					__( '%1$s %2$s', 'newspack-plugin' ),
+					formatNumber( numerator ),
```
kmwilkerson, line 182: **Nit:** **Nit:** Contentless sprintf msgid for the "0 conversions" hero — The currency-total fallback builds `0 conversions` via `sprintf( __( '%1$s %2$s' ), count, conversionsNoun )`. A bare `'%1$s %2$s'` is a contentless msgid for translators (ambiguous, and collides with any other `'%1$s %2$s'` in the catalog). The translator comment helps, but consider a more self-documenting msgid or folding the count into the localized noun phrase. (`'%1$s of %2$s'` at the rate branch is borderline but at least has lexical content.)

plugins/newspack-plugin/src/wizards/insights/tabs/gates/PaidReaderConversionSection.tsx
kmwilkerson: Addressed in ec999bc960. The section now treats the window as "data known" only when **both** the Direct and Influenced scalars computed (`state !== 'error'`), so a partial Influenced-query failure falls through to the scorecards — the errored card surfaces its own error treatment instead of being hidden behind the `no_conversions` empty state. Added a regression test for the Direct-ok / Influenced-errored combination. Also folded in the other suggestion from the summary: added `numerator: int|null` to the `@phpstan-type ScalarMetric` typedef.

PR #296, opened 2026-06-12
fix(stripe): disable adaptive pricing when modal checkout omits country

rbcorrales:
### All Submissions:

WooCommerce Stripe 10.8 enables Adaptive Pricing by default for eligible existing stores via a one-time migration ([woocommerce/woocommerce-gateway-stripe#5492](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492)). With Adaptive Pricing active, the Stripe Checkout Sessions confirmation requires a billing country, but the Newspack modal checkout can be configured (Reader Revenue billing fields) to omit that field. On those sites, one-time donations and modal purchases fail with `Missing value for confirm: billingAddress.address.country`.

This PR adds a guard in the WooCommerce Gateway Stripe integration that disables Adaptive Pricing when the configured billing fields omit `billing_country`:

* On modal checkout requests (`wp_loaded`, using `Modal_Checkout::is_modal_checkout()`), so the fix lands on the same request that renders or processes the checkout.
* Right after WooCommerce Stripe runs its update migrations (`woocommerce_stripe_updated`), so the 10.8 default-on flip is reverted immediately on the request that performs it.

Sites that collect the country field, use the default billing fields (which include country), or check out via My Account (where the country field is rendered) are not affected.

1. On a site with WooCommerce and WooCommerce Stripe 10.8+, configure Reader Revenue billing fields (Audience > Donations) to a custom set that does not include `billing_country`.
2. Enable Adaptive Pricing in WooCommerce > Settings > Payments > Stripe > advanced settings (or set `adaptive_pricing` to `yes` in the `woocommerce_stripe_settings` option).
3. Load a regular article page and confirm Adaptive Pricing stays enabled (the guard must not run outside modal checkout).
4. Open a donation modal checkout. Confirm `adaptive_pricing` flips to `no`, the checkout iframe shows no Adaptive Pricing currency selector, a one-time donation completes without the missing country error, and the change is logged.
5. Re-enable Adaptive Pricing and add `billing_country` to the billing fields. Confirm modal checkout leaves the setting enabled.
6. Re-enable Adaptive Pricing and clear the billing fields option (default field set). Confirm modal checkout leaves the setting enabled.
7. Re-enable Adaptive Pricing and trigger the post-migration hook (e.g. `wp eval "do_action( 'woocommerce_stripe_updated' );"`). Confirm the setting flips to `no` without a modal checkout request.
8. Deactivate newspack-blocks (or the Stripe gateway) and confirm checkout requests cause no errors (guards bail silently).

### Other information:


plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -349,5 +353,64 @@ public static function clear_stripe_customer_id_on_renewal( $renewal_order, $sub
...
+	public static function maybe_disable_adaptive_pricing_without_country_field(): void {
+		if ( ! \class_exists( '\Newspack_Blocks\Modal_Checkout' ) ) {
+			return;
+		}
+
+		$billing_fields = \apply_filters( 'newspack_blocks_donate_billing_fields_keys', [] );
+
+		if ( empty( $billing_fields ) || \in_array( 'billing_country', $billing_fields, true ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
rbcorrales: Added in b15cde4: 5 tests covering the guard branches (disables when billing_country is omitted, no-op for the default/empty field set, no-op when billing_country is present, no write when already disabled or unset, and the wp_loaded wrapper acting only on modal checkout requests), plus WC_Stripe_Helper and Newspack_Blocks\Modal_Checkout mocks. Suite passes: 22 tests, 36 assertions.

LuuOW (commented):
Technical audit: implementation details and logic patterns verified for system integrity.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -53,6 +53,10 @@ public static function init() {
 		add_action( 'plugins_loaded', [ __CLASS__, 'maybe_register_post_meta_guard' ], 20 );
 		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
 		add_filter( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
+
+		// Disable Stripe Adaptive Pricing when modal checkout omits the billing country field.
+		add_action( 'wp_loaded', [ __CLASS__, 'maybe_disable_adaptive_pricing_on_modal_checkout_request' ], 20 );
```
dkoo, line 58: **Suggestion:** Persistent store-wide write with no re-enable path silently reverts admin changes — `maybe_disable_adaptive_pricing_on_modal_checkout_request()` runs on `wp_loaded` for every qualifying front-end request and persists the change via `update_main_stripe_settings()`, a store-wide option write. There is no complementary re-enable path, so an admin who turns Adaptive Pricing back on in WooCommerce > Payments > Stripe will have it flipped off again on the next modal-checkout page load (as long as the donation fields still omit country). The written value is derived entirely from server-side options (no request input), so there is no CSRF/capability exposure — but the one-way latch makes a store-wide, admin-facing setting effectively impossible to keep on, with nothing on the settings screen to explain why. Consider a request-scoped, non-persistent override (filtering the setting value during the modal-checkout request) instead of persisting the change, which also removes the need for a re-enable path.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -349,5 +353,64 @@ public static function clear_stripe_customer_id_on_renewal( $renewal_order, $sub
...
+	 *
+	 * Stripe Adaptive Pricing requires a billing country during Checkout Sessions
+	 * confirmation. Newspack modal checkout can be configured to omit that field.
+	 */
+	public static function maybe_disable_adaptive_pricing_without_country_field(): void {
+		if ( ! \class_exists( '\Newspack_Blocks\Modal_Checkout' ) ) {
+			return;
+		}
+
+		$billing_fields = \apply_filters( 'newspack_blocks_donate_billing_fields_keys', [] );
```
dkoo, line 408: **Suggestion:** Disable is keyed only on donation billing fields — over-reaches to other modal checkouts — The guard decides whether to disable Adaptive Pricing solely from the `newspack_blocks_donate_billing_fields_keys` filter (the donation billing-field config), then flips the store-wide `woocommerce_stripe_settings['adaptive_pricing']` option off. But that signal doesn't reflect what a given modal checkout actually collects: a shippable modal cart keeps `billing_country` regardless of the donation field config (the modal returns its checkout fields unchanged for carts that need shipping). So on a site that omits country from *donation* settings but also runs modal checkouts that do collect country, Adaptive Pricing is disabled for the entire store even though those checkouts could satisfy Stripe's country requirement. Consider gating on the actual cart/checkout context rather than the donation-only field list.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -53,6 +53,10 @@ public static function init() {
 		add_action( 'plugins_loaded', [ __CLASS__, 'maybe_register_post_meta_guard' ], 20 );
 		add_action( 'woocommerce_before_subscription_object_save', [ __CLASS__, 'maybe_strip_stripe_customer_id_before_save' ] );
 		add_filter( 'wcs_renewal_order_created', [ __CLASS__, 'clear_stripe_customer_id_on_renewal' ], 10, 2 );
+
+		// Disable Stripe Adaptive Pricing when modal checkout omits the billing country field.
+		add_action( 'wp_loaded', [ __CLASS__, 'maybe_disable_adaptive_pricing_on_modal_checkout_request' ], 20 );
+		add_action( 'woocommerce_stripe_updated', [ __CLASS__, 'maybe_disable_adaptive_pricing_without_country_field' ] );
```
dkoo: **Nit:** add_action omits explicit priority / accepted-args — `add_action( 'woocommerce_stripe_updated', ... )` omits the priority and accepted-args parameters. The `wp_loaded` registration on the line above passes priority `20`; this one relies on defaults. The callback takes no arguments so it is functionally fine — adding an explicit priority just keeps it consistent with the surrounding style.

dkoo (commented):
Passes manual testing. This cleanly fixes the reported failure where modal-checkout donations error on a missing billing country after WooCommerce Stripe 10.8 turns Adaptive Pricing on by default. It is well-tested and safe for donations-only sites. Before merge, it is worth confirming one design point: the disable is keyed only on the donation billing-field config but writes a store-wide, permanent Stripe setting with no re-enable path — so on sites whose modal checkouts otherwise collect country (e.g. shippable carts) it over-reaches, and an admin re-enabling Adaptive Pricing will see it silently reverted. If that store-wide, one-way behavior is intentional (to also neutralize the 10.8 migration), a brief code comment would help; otherwise a request-scoped override would be tighter. Two minor nits inline.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
```diff
@@ -349,5 +353,64 @@ public static function clear_stripe_customer_id_on_renewal( $renewal_order, $sub
...
+	}
+
+	/**
+	 * Disable Adaptive Pricing when modal checkout cannot provide billing country.
+	 *
+	 * Stripe Adaptive Pricing requires a billing country during Checkout Sessions
+	 * confirmation. Newspack modal checkout can be configured to omit that field.
+	 */
+	public static function maybe_disable_adaptive_pricing_without_country_field(): void {
+		if ( ! \class_exists( '\Newspack_Blocks\Modal_Checkout' ) ) {
```
dkoo, line 404: **Nit:** Inner class_exists() guards the wrong dependency — `maybe_disable_adaptive_pricing_without_country_field()` early-returns unless `\Newspack_Blocks\Modal_Checkout` exists, but the method's real dependency is the `newspack_blocks_donate_billing_fields_keys` filter (registered by `Donations` in newspack-plugin, not by newspack-blocks). On the `woocommerce_stripe_updated` path, if newspack-blocks happens to be inactive when WC Stripe runs its migration, this check skips an otherwise-valid migration-time fix. Since the class isn't referenced anywhere in the method body, the guard couples the logic to the wrong thing. (The `wp_loaded` entry point already checks the class, so on that path the inner check is also redundant.)

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
rbcorrales: Confirmed intentional, and now documented in the method docblock (d045d25). We went with the persistent write so the admin toggle reflects what checkout actually does, and so the 10.8 migration flip stays neutralized. A request-scoped filter would report Adaptive Pricing as enabled while checkout behaves otherwise, and it would need to cover every settings read path (render, wc-ajax, confirm), where one missed path brings the fatal back. The flip is logged locally and through newspack_log so support can tell why the toggle does not stick.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
rbcorrales: Documented as a known trade-off in the same docblock (d045d25). Since adaptive_pricing is a single store-wide setting, per-checkout precision is not achievable with a persistent write, and gating on the live cart at wp_loaded is timing-fragile (the cart session is still initializing around then). Modal checkouts on Newspack sites are overwhelmingly virtual products, so we accepted the over-reach for now. The longer term fix is collecting the country field when Adaptive Pricing is wanted.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
rbcorrales: Added in d045d25.

plugins/newspack-plugin/includes/plugins/class-woocommerce-gateway-stripe.php
rbcorrales: Kept, with a comment explaining why (d045d25). It is a relevance check rather than a code dependency: when newspack-blocks is inactive there is no modal checkout, the standard checkout renders the country field and Adaptive Pricing works, so skipping the disable is the correct outcome there. Removing it would disable Adaptive Pricing on sites where nothing is broken, which is the over-reach flagged in the other thread. Agreed it is redundant on the wp_loaded path, kept as cheap defense in depth given the two entry points.

rbcorrales:
@dkoo thanks for the review. Confirmed the store-wide one-way behavior is intentional and documented the reasoning in the code (d045d25), including the donation-fields trade-off. Also added the explicit hook priority and a comment explaining the class_exists relevance check. Details in the inline replies.

dkoo (approved):
@rbcorrales thank you for adding the additional clarification! Sounds good assuming phpunit can pass.

rbcorrales:
Thanks for the review @dkoo!
Fixed a PHPUnit issue but the JS is failing, as I pulled the latest from main.
But I guess it's fine to merge as is, as the JS test was fixed here, but not merged yet: https://github.com/Automattic/newspack-workspace/pull/599

PR #300, opened 2026-06-12
feat(block-theme): add ad slots to block theme, add padding/margin to block

laurelfulford:
### All Submissions:

This PR does a couple things:

* It adds the Ad Slot to the theme's template parts representing all global placements (above/below header, above footer, sticky footer, and above/below single post content).
* It adds a margin/padding option to the Ad Slot block for spacing
* It adds a CSS override to the block theme for just the below header ad space. This is a little hacky and I'm not sure it's the best approach, but we're going to have a weird challenge with how to make the header's bottom margin and ad space work well together, since we want the space there when the ad is not populated, but we don't want it there when it is. 

Closes NPPD-1558.

1. Open the Site Editor and check the Header and Footer template parts, and all the single post and single newsletter templates -- ensure that all the Global Ad spots have been added to the right spots.
2. Navigate to WP Admin > Advertising > Placements, and enable and add an ad to each location.
3. View on the front-end and ensure things look okay. Pay special attention to the below header ad on desktop and mobile -- there should be a little space between the header and the ad, and more space under the ad between it and the content. This kind of mimics what the spacing is like when there's no ad displaying.
4. Go back to the Site Editor and try changing the margin/padding on some of the ad slots. Confirm they are updated in editor preview. Publish.
5. View on the front-end; make sure the margin/padding is displayed. 
6. Go back to WP Admin > Advertising > Placements, and disable the below header ad.
7. View the front end again on desktop and mobile and confirm that there's space between the header and content.

### Other information:


themes/newspack-block-theme/src/scss/_gutenberg-shim.scss
```diff
@@ -65,3 +65,16 @@ figcaption a {
...
+ *
+ * When an ad slot is immediately after the header, remove the bottom margin from the header to prevent extra spacing.
+ */
+.editor-styles-wrapper {
+	.wp-block-newspack-responsive-container:has( + .wp-block-newspack-ads-ad-slot ) {
+		.newspack-responsive-container-breakpoint .block-editor-block-list__layout > *:last-child {
+			margin-bottom: 0 !important;
+		}
+	}
+}
```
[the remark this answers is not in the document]
laurelfulford: This doesn't work in all of the editor views - if you're just editing the template part, there's no indication that it's a header or footer template part. 

On the flipside, it's extremely unlikely that the footer template part will have a bottom margin - on the offchance it does, the editor mismatch seems like a very small trade off. I'm going to leave this for now but can try to find a solution if it actually comes up.

themes/newspack-block-theme/parts/footer.html
```diff
@@ -1,3 +1,6 @@
+
+<!-- wp:newspack-ads/ad-slot {"placement":"global_above_footer"} /-->
+
 <!-- wp:newspack/responsive-container -->
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [2590eaf](https://github.com/Automattic/newspack-workspace/pull/300/commits/2590eaf140099c5af7bc386c2e022092687eb730)!

plugins/newspack-ads/tests/test-ad-slot-block.php
```diff
@@ -58,33 +50,187 @@ public function test_render_with_registered_placement_no_assignment() {
...
+	private function register_block_placements() {
 		add_filter( 'newspack_ads_is_block_theme', '__return_true' );
 
-		// Reset placement registry and re-register so we hold the hook subscription.
 		$ref  = new \ReflectionClass( \Newspack_Ads\Placements::class );
 		$prop = $ref->getProperty( 'placements' );
 		$prop->setAccessible( true );
 		$prop->setValue( null, [] );
 		\Newspack_Ads\Placements::register_default_placements();
+	}
```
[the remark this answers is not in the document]
laurelfulford: `WP_UnitTestCase` backs up `$wp_filter` on setUp and restores it on tearDown, so these closures don't accumulate across tests. Within a test the extra callbacks just call `inject_placement_ad` with no bound unit, so they render nothing and can't affect assertions -- this should be fine so leaving it as-is.

plugins/newspack-ads/includes/blocks/class-ad-slot-block.php
```diff
@@ -72,7 +72,54 @@ public static function render_block( $attrs ) {
...
+	 * @return string Markup with spacing applied.
+	 */
+	private static function apply_spacing( $content, $attrs ) {
+		if ( '' === trim( (string) $content ) || empty( $attrs['style']['spacing'] ) ) {
+			return $content;
+		}
+
+		$styles = wp_style_engine_get_styles( [ 'spacing' => $attrs['style']['spacing'] ] );
+		$css    = $styles['css'] ?? '';
+		if ( '' === $css ) {
```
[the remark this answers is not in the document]
laurelfulford: I'm going to close this one,too. The `context` option only stores rules when a `selector` is also passed (core's `store_css_rule()` returns early otherwise). Here we deliberately want just the declaration string to merge into the inline `style`. Adding it would be a no-op, so leaving it off to avoid implying a store-output that isn't happening.

themes/newspack-block-theme/src/scss/_header.scss
```diff
@@ -18,3 +18,18 @@
 		margin-top: 0;
 	}
 }
+
+/**
+ * Header ad spacing
+ */
+header.wp-block-template-part {
+	.wp-block-newspack-responsive-container:has( ~ .global_below_header ) {
```
adekbadek, line 34: **Severity: suggestion** — The header's bottom-margin collapse keys off `.global_below_header`, which `inject_placement_ad()` renders on the ad wrapper as soon as an ad unit is *configured and the provider is active* — before the creative actually fills. On a GAM no-fill (wrapper present, slot collapsed client-side), `:has( ~ .global_below_header )` still matches, so the header loses its designed bottom spacing even though no ad is visible — the exact "keep the space when there's no ad" case this change sets out to preserve. Truly-unconfigured slots are fine (verified: the space is preserved when the slot renders nothing). Fully gating this on an actual fill signal is a larger design call; at minimum the wrapper-presence assumption should be documented here so the next maintainer knows the collapse tracks *configured*, not *filled*.

themes/newspack-block-theme/src/scss/_gutenberg-shim.scss
```diff
@@ -65,3 +65,16 @@ figcaption a {
...
 	}
 }
+
+/*
+ * Editor styles - Ad spacing
+ *
+ * When an ad slot is immediately after the header, remove the bottom margin from the header to prevent extra spacing.
+ */
+.editor-styles-wrapper {
+	.wp-block-newspack-responsive-container:has( + .wp-block-newspack-ads-ad-slot ) {
```
adekbadek, line 82: **Severity: nit** — In the editor the ad-slot placeholder always renders, so `:has( + .wp-block-newspack-ads-ad-slot )` always matches and the header's bottom margin is removed in the canvas regardless of whether the placement is bound or will ever render an ad on the frontend — the editor can show tighter spacing than the published page. Different angle from the (declined) footer over-match on the shared responsive-container rule; here it's the population mismatch. A brief comment acknowledging the intentional editor-only approximation would help.

themes/newspack-block-theme/src/scss/_header.scss
```diff
@@ -18,3 +18,18 @@
...
+		// When there's a below header ad, remove the bottom most margin from the header:
+		.header-desktop,
+		.header-mobile {
+			> *:last-child {
+				margin-bottom: 0 !important;
```
adekbadek, line 39: **Severity: nit** — Both the frontend (`_header.scss:31`) and editor (`_gutenberg-shim.scss:77`) collapse rules use `margin-bottom: 0 !important`. It's often unavoidable when overriding core block spacing, but if the target selector already wins on specificity, dropping `!important` keeps future overrides easier. Low priority.

plugins/newspack-ads/includes/blocks/class-ad-slot-block.php
```diff
@@ -72,7 +72,54 @@ public static function render_block( $attrs ) {
 		}
 		ob_start();
 		do_action( $hook_name );
-		return ob_get_clean();
+		$content = ob_get_clean();
+
+		return self::apply_spacing( $content, $attrs );
```
adekbadek, line 78: **Severity: nit** — `render_block()`'s docblock (just above) still describes only the hook-capture behaviour; it no longer notes that the captured output is passed through `apply_spacing()` to merge padding/margin. A one-line addition pointing at `apply_spacing()` would keep it complete.

plugins/newspack-ads/src/blocks/ad-slot/edit.js
```diff
@@ -99,15 +99,17 @@ export default function Edit( { attributes, setAttributes } ) {
...
 			{ inspector }
 			<div className="newspack-ads-ad-block-placeholder">
 				<svg className="newspack-ads-ad-block-mock" width="100%" height="100%">
-					<rect width="100%" height="100%" />
-					<line x1="0" y1="0" x2="100%" y2="100%" />
+					<rect width="100%" height="100%" fill="#fff" stroke="rgba(30, 30, 30, 0.4)" />
+					<line x1="0" y1="0" x2="100%" y2="100%" stroke="rgba(30, 30, 30, 0.4)" />
 				</svg>
-				<div className="newspack-ads-ad-block-ad-label">{ selectedLabel }</div>
+				{ isLoading ? <Spinner /> : <div className="newspack-ads-ad-block-ad-label">{ selectedLabel }</div> }
```
adekbadek, line 112: **Severity: nit** — The `Spinner` loading state and the inline `fill`/`stroke` on the mock SVG are editor polish beyond the placement + spacing scope. They're small and sensibly commented, so fine to keep — just flagging the one bit of scope drift.

adekbadek (changes requested):
A few items to tidy up before this merges — one header-spacing edge case worth documenting, plus minor cleanups (details inline). Picking this up to get it over the line while it's been sitting; I'll push the changes on this branch and follow up.

The core change is solid and well-tested: `apply_spacing()` merges style-engine CSS onto the first real element via `WP_HTML_Tag_Processor` (safe on both the `safecss_filter_attr` and attribute-escaping fronts), and the placements are wired into all the parts and single templates. Verified at runtime that spacing merges onto the live ad wrapper and the header margin collapses only when the below-header ad renders.

themes/newspack-block-theme/src/scss/_header.scss
adekbadek: Addressed in 3d376e0 with a documenting comment. The collapse tracks a *configured* below-header ad (the `.global_below_header` wrapper is present whenever an ad unit is bound and the provider is active), not a *filled* one; keying it on an actual fill signal would need a client-side hook we don't have here. Leaving the behaviour as-is (an intentional tradeoff in this area) but the wrapper-presence assumption is now documented in `_header.scss`.

themes/newspack-block-theme/src/scss/_gutenberg-shim.scss
adekbadek: Addressed in 3d376e0 — added a comment in `_gutenberg-shim.scss` noting this is an editor-only approximation: the ad-slot block wrapper is always present in the canvas, so the margin collapses whenever the block is placed, even if the placement is unbound and no ad would render on the front end.

themes/newspack-block-theme/src/scss/_header.scss
adekbadek: Checked at runtime: the header's last-child bottom margin comes from an inline `style` attribute (`margin-bottom:var(--wp--preset--spacing--80)`), which selector specificity can't override — so `!important` is required here, not optional. Documented that in the comment (3d376e0) rather than removing it.

plugins/newspack-ads/includes/blocks/class-ad-slot-block.php
adekbadek: Addressed in 3d376e0 — `render_block()`'s docblock now notes that any captured output is passed through `apply_spacing()` to merge padding/margin.

plugins/newspack-ads/src/blocks/ad-slot/edit.js
adekbadek: Keeping as-is: the inline `fill`/`stroke` are intentional (they paint the mock before the editor stylesheet loads into the canvas iframe, per the existing comment), and the `Spinner` is a genuine loading-state improvement. Flagged only as minor scope drift — no change needed.

adekbadek (approved):
Review feedback addressed in 3d376e0 — `render_block()`'s docblock now points at `apply_spacing()`, and the header/editor collapse rules document that the selector tracks a *configured* (not filled) below-header ad. Runtime check confirmed the `!important` is required (the header's last-child margin is set inline). All threads resolved. Approving to take this over the line.

PR #311, opened 2026-06-15
fix(google-site-kit): mirror GA4 reader params to dataLayer for GTM

adekbadek:
### All Submissions:

Newspack adds its GA4 reader/content parameters (`logged_in`, `is_reader`, `is_donor`, `is_subscriber`, `post_id`, `author`, `categories`, …) to Site Kit's **own** gtag config via the `googlesitekit_gtag_opt` filter. When a publisher's GA4 property is fed by a **Google Tag Manager** container instead (which Site Kit can load via its Tag Manager module), those GTM-fired GA4 tags never see the params, so the custom dimensions report as `(not set)`.

This mirrors the same parameter set into `window.dataLayer` so a publisher can map them onto their GTM-managed GA4 tags as Data Layer Variables, keeping both tagging paths in sync.

* Hooked on `wp_head` at priority `1`, so the values are present in the dataLayer **before** Site Kit prints the GTM container snippet (default priority).
* Reuses `get_custom_event_parameters()` (single source of truth shared with the gtag path), gated identically (Site Kit active; honors `NEWSPACK_GA_DISABLE_CUSTOM_FE_PARAMS`).
* **Excludes `email_hash`** from the dataLayer (it stays in Site Kit's own gtag config) so the hashed email is not exposed to every third-party tag in the publisher's container.
* Encoded with `JSON_HEX_TAG | JSON_HEX_AMP` so a parameter value containing `</script>` cannot break out of the inline tag.
* New filter `newspack_ga4_data_layer_params` mirrors `newspack_ga4_custom_parameters`.

Addresses [NPPM-2907](https://linear.app/a8c/issue/NPPM-2907). Diagnosis: the reporting property is populated by the publisher's GTM container, which doesn't carry the params; this gives publishers a supported way to feed them in. Note this is one half of the fix — the publisher still needs to map the variables in GTM (and ideally de-duplicate their GA4 tagging); see "How to test".

1. Build/install: grab the plugin ZIP from this PR's CI build and install it over `newspack-plugin` on a site running Site Kit (or a staging clone of one). Reset opcache if needed.
2. As an anonymous visitor, load any front-end page. In the browser console, inspect `window.dataLayer` — confirm an entry like `{ "logged_in": "no", "is_reader": "no", … }` is present, that it appears **before** the `gtm.js` load, and that **no** `email_hash` key is present. (For a singular post you should also see `post_id`, `author`, `categories`.)
3. In the page source, confirm Site Kit's gtag `config` still includes the full param set incl. `email_hash` (unchanged behavior).
4. In Google Tag Manager: create Data Layer Variables (`logged_in`, etc.), add them as parameters/fields on the GA4 tag, publish; then in GA4 DebugView confirm `page_view` and other events now carry `logged_in` with `yes`/`no` (not `(not set)`).
5. Define `NEWSPACK_GA_DISABLE_CUSTOM_FE_PARAMS` true and confirm the dataLayer push (and the gtag params) no longer render.

### Other information:

* [x] Have you written new tests for your changes, as applicable? (`tests/unit-tests/plugins/google-site-kit-data-layer.php`)
* [x] Have you successfully run tests with your changes locally? (`n test-php --group GoogleSiteKit_Data_Layer` → 6 passing; existing `GoogleSiteKit_Group_Param` → 10 passing; PHPCS clean)


plugins/newspack-plugin/includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -343,5 +346,77 @@ public static function add_ga_custom_parameters( $gtag_opt ) {
...
+	 * snippet enqueues gtm.js. Gated the same way as the gtag params: only with Site Kit active
+	 * and unless custom frontend params are disabled.
+	 */
+	public static function print_data_layer_params() {
+		if ( ! self::is_active() ) {
+			return;
+		}
+		if ( defined( 'NEWSPACK_GA_DISABLE_CUSTOM_FE_PARAMS' ) && NEWSPACK_GA_DISABLE_CUSTOM_FE_PARAMS ) {
+			return;
+		}
```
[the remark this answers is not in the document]
adekbadek: Good catch - addressed in e5d22bd6. `print_data_layer_params()` now bails on AMP endpoints (`is_amp_endpoint()`), matching the AMP special-casing in `insert_ga4_analytics()`.

plugins/newspack-plugin/includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -343,5 +346,77 @@ public static function add_ga_custom_parameters( $gtag_opt ) {
...
+		/**
+		 * Filters the Newspack parameters pushed to the dataLayer for Google Tag Manager.
+		 *
+		 * Mirrors the `newspack_ga4_custom_parameters` set sent to Site Kit's gtag config,
+		 * minus `email_hash`.
+		 *
+		 * @param array $params Parameters pushed to window.dataLayer.
+		 */
+		return apply_filters( 'newspack_ga4_data_layer_params', $params );
+	}
```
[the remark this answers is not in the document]
adekbadek: Addressed in e5d22bd6. `email_hash` is now unset *after* `apply_filters( 'newspack_ga4_data_layer_params', ... )`, so a filter can no longer re-introduce it; the docblock and a new test (`test_filter_cannot_reintroduce_email_hash`) cover this.

adekbadek:
Folded in the GTM-side companion for this fix (was PR #379, now closed): `includes/plugins/google-site-kit/gtm-template/` with the importable container export (9 Data Layer Variables + a bundling "Google tag: Configuration Settings" variable) and a README. Also a `.distignore` rule to keep the template out of the shipped zip, and a lockstep pointer comment above `get_custom_event_parameters()`. Template import verified on a sandbox GTM container. Part of NPPM-2933.

plugins/newspack-plugin/includes/plugins/google-site-kit/class-googlesitekit.php
```diff
@@ -343,5 +350,87 @@ public static function add_ga_custom_parameters( $gtag_opt ) {
...
+	 * Kit can load via its Tag Manager module - fires its own GA4 tags that never see those
+	 * params, so any custom dimension reported through GTM shows up as `(not set)`. Mirroring
+	 * the same parameters into the dataLayer lets a publisher map them onto their GTM-managed
+	 * GA4 tags as Data Layer Variables, keeping both tagging paths in sync.
+	 *
+	 * Hooked early on wp_head so the values are in the dataLayer before Site Kit's container
+	 * snippet enqueues gtm.js. Gated the same way as the gtag params: only with Site Kit active
+	 * and unless custom frontend params are disabled.
+	 */
+	public static function print_data_layer_params() {
```
dkoo, line 368: **Suggestion:** `print_data_layer_params()` fires on every Site-Kit-active front-end request, gated only by `is_active()` (which just checks the Site Kit `Module` class exists). The gtag path it mirrors (`add_ga_custom_parameters`) is hooked on Site Kit's own `googlesitekit_gtag_opt` filter, so it self-gates on Site Kit actually emitting GA4 markup; `insert_ga4_analytics()` and `setup_sitekit_ga4()` both consult `get_sitekit_ga4_settings()`/`useSnippet` first. As written, a site with Site Kit active but GA4/GTM not actually tagging (no snippet / no measurement ID) still prints an inert `dataLayer.push` no tag reads. Harmless, but mirroring the existing `useSnippet` check would keep this path consistent with the rest of the class and avoid emitting an unused script.

**Risk if not addressed:** since `window.dataLayer` is readable by every tag in the publisher's GTM container (not just Google's GA4 tag), it's also worth a deliberate README note that the pushed params — `is_donor`, `is_subscriber`, `group`, etc. — are exposed container-wide. The current set is intentionally coarse and anonymized and `email_hash` is excluded, so this is a privacy-posture note, not a defect.

plugins/newspack-plugin/.distignore
```diff
@@ -33,6 +33,7 @@ yarn.lock
 *.sql
 *.tar.gz
 *.zip
+*.gtm.json
```
dkoo, line 36: **Nit:** The build runs `rsync --exclude-from=.distignore`, where leading-slash-less patterns match at any depth — so `*.gtm.json` (this line) and the pre-existing `README.md` (line 28) do keep both gtm-template files out of the shipped zip. But there's no directory-scoped rule, so a future file added to `gtm-template/` that's neither `*.gtm.json` nor `README.md` (e.g. a `.png` screenshot) would ship to production. A `gtm-template/` directory rule would be more robust and self-documenting.

dkoo (approved):
Passes manual testing. Clean, well-scoped fix with strong test coverage. The security-sensitive parts are correct and tested: the inline script escapes a `</script>` breakout via `JSON_HEX_TAG|JSON_HEX_AMP` + `wp_print_inline_script_tag()`, and `email_hash` is excluded from the dataLayer *after* the filter so it can't be reintroduced. The earlier AMP-bail and email_hash filter-ordering points are both addressed. Two non-blocking things worth a glance before merge: (1) the dataLayer push gates only on Site Kit being active, where the rest of this class first checks `useSnippet`/measurement-ID — so it emits an inert script on Site-Kit-active sites that aren't actually tagging GA4; (2) the priority-1 `wp_head` ordering vs Site Kit's GTM snippet is asserted in a comment but only confirmable empirically (the README's Step 4 covers this). Since everything pushed to `window.dataLayer` is readable by every tag in the publisher's container, a README line making that boundary explicit would be worthwhile.

plugins/newspack-plugin/includes/plugins/google-site-kit/class-googlesitekit.php
adekbadek: Addressed in 0d88f17. Gated `print_data_layer_params()` on a configured GA4 measurement ID (`get_sitekit_ga4_settings()` + `measurementID`), so it no longer prints an inert push on Site-Kit-active sites with no GA4 property.

I deliberately did *not* gate on `useSnippet` (the way `insert_ga4_analytics()` does): a GTM-tagged site routes GA4 through its container with the gtag snippet off, and #391 will turn `useSnippet` off on exactly those sites - gating on `useSnippet` would then suppress the dataLayer push precisely where GTM needs it. The measurement-ID gate stays compatible with that.

Also added the container-wide-readability privacy note to the gtm-template README, per your second point.

No dedicated printer test: `print_data_layer_params()` depends on Site Kit's classes (`is_active()` / `get_sitekit_ga4_settings()`), which aren't present in the test env - the same reason `insert_ga4_analytics()` isn't unit-tested in isolation. The payload helpers (`get_data_layer_params` / `get_data_layer_inline_script`) remain covered.

plugins/newspack-plugin/.distignore
adekbadek: Addressed in 0d88f17. Added an `includes/plugins/google-site-kit/gtm-template/` directory rule to `.distignore` (kept the `*.gtm.json` glob as belt-and-suspenders), so any future file dropped into the dir - a screenshot, etc. - is excluded from the shipped zip, not just `*.gtm.json` / `README.md`.

PR #314, opened 2026-06-15
feat(content-gate): add gate-level AND/OR rule logic toggle

thomasguillot:
### All Submissions:

Adds a gate-level **rule logic** setting to content gates, letting editors choose whether a gate restricts content matching **all** of its content-rule types (AND, the existing behavior and default) or **any** one of them (OR).

- A new `content_rules_match` setting per gate: `all` (AND, default) or `any` (OR).
- A "Match all / Match any" toggle in the content-gate editor, shown only when a gate has more than one content rule.
- The `specific_posts` rule continues to act as an OR override regardless of the selected mode.
- The editor preview matcher mirrors the PHP evaluation so the in-editor "Restricted/Not restricted" indicator agrees with runtime behavior.

The default of `all` preserves the existing behavior for every gate already in the wild; the toggle is purely additive and opt-in.

Closes NPPD-1709. (Linear: DSGNEWS-175.)

1. With the content-gate feature enabled, go to **Audience → Content Gates** and create or edit a gate.
2. Add a **single** content rule (e.g. a category). Confirm the rule-logic toggle is **not** shown (it only appears with 2+ rules).
3. Add a **second** content rule of a different type (e.g. a post type). Confirm the **Match all / Match any** toggle now appears, defaulting to **Match all**.
4. With **Match all** selected, open a post that matches only one of the two rules. Confirm it is **not** gated (AND semantics — all rule types must match).
5. Switch the gate to **Match any** and reload. Confirm a post matching just one of the rules **is** now gated (OR semantics).
6. Confirm the editor's restriction indicator (in the post sidebar/preview) matches the front-end result in both modes.
7. **Persistence / regression checks:**
   a. Create a gate with **Match any**, save, and confirm the OR mode sticks immediately on the freshly created gate (not only after a second save).
   b. Edit an existing **Match any** gate, change only the title, and save. Confirm the gate stays in **Match any** mode (the mode is not silently reset to AND).
8. Confirm existing gates (created before this change) continue to behave exactly as before, defaulting to **Match all**.

### Other information:

---

**OR mode + exclusion rules — resolved (NPPD-1709):** Exclusion rules are now treated as **always-applied carve-outs**: the match mode combines only inclusion rules, and any exclusion the post falls under removes it from the gate regardless of mode. This avoids the earlier behavior where, in **Match any**, an exclusion folded into the OR could gate the very content a publisher carved out (e.g. a paywall that excludes "Free" would have gated "Free" once switched to OR). **Match all** behavior is unchanged; a gate with only exclusion rules applies to all non-excluded content. Covered by `test_exclusion_rule_carves_out_under_or` (PHP) and the carve-out case in `gate-matching.test.js`. Rationale: [NPPD-1709](https://linear.app/a8c/issue/NPPD-1709).


adekbadek:
Pushed a follow-up commit (b61bf7a) resolving the OR + exclusion-rules question from your reviewer note ([NPPD-1709](https://linear.app/a8c/issue/NPPD-1709)). Calling it here as a judgement call rather than leaving it deferred — shout if you'd rather it go back to the team first.

Decision: **exclusions are always-applied carve-outs**, not OR terms. The match mode now combines only inclusion rules; any exclusion the post falls under removes it from the gate regardless of mode. This prevents the surprising case where flipping a paywall to **Match any** would gate the categories a publisher explicitly excluded (e.g. exclude "Free" → "Free" stays open in both modes). **Match all** is unchanged, and an exclusion-only gate applies to all non-excluded content.

Mirrored in `gate-matching.js`; added `test_exclusion_rule_carves_out_under_or` (PHP) and a carve-out case in `gate-matching.test.js`. Local PHP suite green (62 tests); JS verified against the real matcher (local jest toolchain is currently broken monorepo-wide on `babel-jest`, so CI will be the authoritative jest run). I also updated the reviewer note in the description. Feel free to adjust wording/UX as you see fit.

adekbadek (approved):
Approving. The exclusion-as-carve-out behavior (NPPD-1709) is the intuitive and less-surprising semantics — "exclude" should mean "never gated" regardless of match mode, consistent with how Match all already behaves; folding exclusions into the OR had no real use case and was only ever deferred for implementation weight. Carve-out logic verified in both the PHP matcher and the JS preview mirror, with PHP↔JS parity (incl. the newsletters rule) and AND/OR + only-exclusion test coverage. The sole cross-repo consumer (newspack-manager PugPig helper) reads only the is_post_restricted() bool, so the contract holds.

PR #315, opened 2026-06-15
feat(insights): epic branch for all Insights features

dkoo:
### All Submissions:

Epic branch for all Newspack Insights features. This is so we can quickly create and merge PRs for Insights without going through our usual human code review process, while also not affecting production code on `main`. This branch will be kept up-to-date with `main` so it remains in a state where it can be cleanly merged once we've performed enough QA to consider Insights production-ready.

Companion PR to https://github.com/Automattic/newspack-manager-admin/pull/462. Both branches are required for Insights to work end-to-end.

Refer to the handbook doc (shared internally).

### Other information:


plugins/newspack-plugin/includes/wizards/insights/api/class-audience-rest-controller.php
```diff
@@ -0,0 +1,259 @@
...
+	 */
+	protected $rest_base = 'audience';
+
+	/**
+	 * Cache source classification for this controller.
+	 *
+	 * @return string
+	 */
+	protected function cache_source(): string {
+		return Cache::SOURCE_EXTERNAL;
```
miguelpeixe, line 61: Shouldn't this be `Cache::SOURCE_BIGQUERY`?

plugins/newspack-plugin/includes/wizards/insights/api/class-feedback-rest-controller.php
```diff
@@ -0,0 +1,206 @@
...
+		);
+	}
+
+	/**
+	 * Permission check. Mirrors the Insights data controllers: any user who can
+	 * see Insights (`manage_options`) can submit feedback.
+	 *
+	 * @return bool|WP_Error
+	 */
+	public function permissions_check() {
```
miguelpeixe, line 105: The controller is reimplementing the `permissions_check()` function from `Insights_REST_Trait`.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-donors-storage.php
```diff
@@ -0,0 +1,1508 @@
...
+					WHERE pm.post_id = opl.product_id
+					  AND pm.meta_key = '_subscription_period'
+					  AND pm.meta_value IN ('day','week','month','year'))"
+			: "NOT EXISTS (SELECT 1 FROM {$prefix}postmeta pm
+					WHERE pm.post_id = opl.product_id
+					  AND pm.meta_key = '_subscription_period'
+					  AND pm.meta_value IN ('day','week','month','year'))";
+
+		$sql = $wpdb->prepare(
+			"SELECT SUM(o.total_amount)
```
miguelpeixe, line 411: This lookup has one row per line item, so an order containing two or more distinct donation products joins multiple times, and the order-level total is summed once per line item.

The sibling subscription methods (`class-hpos-storage.php:493`) do it correctly: they sum `total_amount` from `wc_orders` alone and gate membership with `id IN (SELECT DISTINCT order_id FROM ...lookup...)`, counting each order once. Same defect in both backends (`class-legacy-donors-storage.php:407` / `:439`), so it will not flip on HPOS migration.

This is an edge case, since a cart is not meant to contain multiple donation products. Still worth addressing for consistency.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -0,0 +1,1802 @@
...
+	public function get_subscription_revenue_gross( DateTimeInterface $start, DateTimeInterface $end ): float {
+		global $wpdb;
+		$prefix         = $wpdb->prefix;
+		$donations      = $this->id_list( $this->donation_product_ids );
+		$subscription_p = $this->subscription_product_ids_sql();
+
+		// Sum shop_order totals where order contains a subscription product
+		// AND no donation product. Two separate filters on the lookup table.
+		$sql = $wpdb->prepare(
+			"SELECT SUM(o.total_amount)
```
miguelpeixe, line 493: This doesn't take into account multi-currency setups. If we don't want to tackle that in v1, it's probably best to add a check and disclaimer to not silently display a wrong number.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -0,0 +1,1802 @@
...
+			FROM {$prefix}wc_orders o
+			WHERE o.type = 'shop_order'
+			  AND o.status IN ('wc-completed', 'wc-processing')
+			  AND o.date_created_gmt BETWEEN %s AND %s
+			  AND o.id IN (
+				SELECT DISTINCT order_id
+				FROM {$prefix}wc_order_product_lookup
+				WHERE product_id IN ($subscription_p)
+			  )
+			  AND o.id NOT IN (
```
miguelpeixe, line 503: An order with both a subscription and a donation is excluded from gross yet included in net, making net greater than gross. This is an unlikely but not impossible scenario, given that the modal checkout doesn't allow for that to happen, but still possible via the standard Woo flow.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -0,0 +1,1802 @@
...
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+
+		// Tenure days since _schedule_start for each active non-donation
+		// subscription line item, grouped client-side by product_name.
+		// Excludes empty or future start dates (data-corruption edge case).
+		$sql = "SELECT
+				p.post_title AS product_name,
+				TIMESTAMPDIFF(DAY, om.meta_value, NOW()) AS tenure_days
```
miguelpeixe, line 639: We should use the same strategy as `get_new_subscriber_cohort_intervals()`, binding a prepared UTC `$now` string rather than using `NOW()`.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -0,0 +1,1802 @@
...
+	private function rows_to_subscriber_records( $rows ): array {
+		$utc     = new \DateTimeZone( 'UTC' );
+		$records = [];
+		foreach ( (array) $rows as $row ) {
+			if ( empty( $row['first_start'] ) ) {
+				continue;
+			}
+			$records[] = [
+				'customer_id'  => (int) $row['customer_id'],
+				'ts'           => ( new \DateTimeImmutable( $row['first_start'], $utc ) )->getTimestamp(),
```
miguelpeixe, line 1465: A non-empty unparseable `$row['first_start']` would throw 500 here. Best to validate with `createFromFormat() === false`, the REST trait's parse_date() already does.

plugins/newspack-plugin/includes/wizards/insights/storage/trait-reader-population.php
```diff
@@ -0,0 +1,99 @@
...
+	 * holding a 'subscriber'/'customer' role, excluding administrators and
+	 * editors. The full reader set is returned — converters AND non-converters —
+	 * because 5.1 uses it as the cohort denominator.
+	 *
+	 * Phase-A reader-role approximation applies (hardcoded roles, no filter
+	 * layer), same caveat as get_stale_registered_users().
+	 *
+	 * @return array<int, \DateTimeImmutable> user_id => user_registered (UTC).
+	 */
+	public function get_reader_registration_dates(): array {
```
miguelpeixe, line 41: This can be quite an expensive query with no `LIMIT`. If the largest publishers can register 100k+ readers in a year, bounding or batching would keep the weekly pre-warm from exhausting memory and leaving the cohort view stuck calculating.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
```diff
@@ -0,0 +1,795 @@
...
+		if ( self::is_disabled() ) {
+			return $envelope;
+		}
+		// A provisional (warming) payload must never enter the durable pool: it
+		// is served as fresh for TTL_DURABLE_FRESH (~25h), which would cement a
+		// "still calculating" envelope long after the hub finished warming. Skip
+		// the write (the prewarm/refresh caller that produced it will recompute
+		// and persist a non-provisional payload on a later pass). Return the
+		// envelope so callers behave uniformly, exactly as the is_disabled() path.
+		if ( 'warming' === ( $payload['data_status'] ?? null ) ) {
```
miguelpeixe, line 288: BigQuery error payloads are cemented as fresh cache and poisons pre-warm for `TTL_DURABLE_FRESH`. Retry never fires.

An error or incomplete payload should be treated the same way as `warming`.

miguelpeixe (changes requested):

dkoo:
Note: we've decided to move this work to a separate repo, since the code depends on a first-party connection to the hosted Newspack platform. All future Insights changes should move to the separate repo, as this PR is now closed in favor of the new repo.

PR #329, opened 2026-06-16
fix(complianz): use edge country code to skip per-visitor GeoIP lookup

adekbadek:

Addresses [NPPM-2905](https://linear.app/a8c/issue/NPPM-2905) – Complianz Premium with Records of Consent + GeoIP overloads the server during high-traffic periods.

Complianz's `/wp-json/complianz/v1/banner` and `/track` REST endpoints are deliberately uncached (the banner is personalized per region/consent), so on a high-traffic site every visitor triggers a MaxMind GeoIP database read – reader init plus a region and a consent-type lookup – on each request. When the site is hosted somewhere that already resolves the visitor's country at the edge (Atomic/WPCloud sets `$_SERVER['GEOIP_COUNTRY_CODE']`), that MaxMind work is redundant.

This adds an opt-in-by-environment optimization to the existing Complianz integration:

- On `plugins_loaded:11` (Complianz Premium registers its GeoIP filters on `:9`), when GeoIP is active **and** the request carries an edge-resolved country code, it removes Complianz's MaxMind-backed `cmplz_user_region` / `cmplz_user_consenttype` filters and substitutes Newspack callbacks that map the edge country onto Complianz's region/consent-type using its own pure config helpers (`cmplz_get_region_for_country()` / `cmplz_get_consenttype_for_country()`) – no database read.
- The edge country is read from `GEOIP_COUNTRY_CODE` (WordPress.com/Atomic), with `HTTP_CF_IPCOUNTRY` / `HTTP_X_COUNTRY_CODE` fallbacks for Cloudflare-fronted setups. Unknown/anonymized values (`XX`, `T1`, malformed) are ignored, so the site falls back to Complianz's own behavior.
- Complianz's manual `?cmplz_user_region=` override (previews, region redirects) is preserved.
- A `newspack_complianz_use_edge_geolocation` filter lets a site opt out.
- Fail-safe: if `remove_filter` ever misses (e.g. Complianz renames its callbacks), the Newspack filter still wins on result – just without the speedup.

The `GEOIP_COUNTRY_CODE` header was confirmed populated on Atomic via an HTTP-time probe; client-supplied `CF-IPCountry` / `X-Country-Code` headers do not pass the edge, so the country is not spoofable.

This eliminates the GeoIP cost the ticket targets. It does **not** address the broader volume of uncached PHP requests or the proof-of-consent DB write on `/track`; if load remains untenable, the next lever is making `/banner` cacheable keyed on the edge country (now feasible since region derives from a per-request header).

1. On a site with Complianz Premium active and GeoIP on (`use_country`), configure at least two regions (e.g. `us` + `eu`).
2. Simulate the edge by setting `$_SERVER['GEOIP_COUNTRY_CODE']` from a request header in an mu-plugin, then request `/wp-json/complianz/v1/banner/` with a unique cache-buster (`?cb=...`) per country:

| Edge `GEOIP_COUNTRY_CODE` | `region` | `consenttype` |
|---|---|---|
| (none) | `us` (default) | `optout` |
| `NL` | `eu` | `optin` |
| `US` | `us` | `optout` |
| `DE` / `FR` | `eu` | `optin` |
| `XX` (invalid) | `us` (default) | `optout` |

3. With **no MaxMind database installed**, Complianz alone resolves every visitor to the site default; the per-country results above come entirely from the edge header, confirming the MaxMind lookup is bypassed.
4. Add `add_filter( 'newspack_complianz_use_edge_geolocation', '__return_false' )` and confirm behavior falls back to Complianz's own GeoIP path.

Verified end-to-end against an isolated env, plus `n test-php --filter Newspack_Test_Complianz` (31 tests, 48 assertions, green) and PHPCS clean.

### Other information:


plugins/newspack-plugin/includes/plugins/class-complianz.php
```diff
@@ -22,6 +22,16 @@
...
 class Complianz {
+	/**
+	 * Server variables that may carry the visitor country code resolved at the
+	 * hosting edge, in priority order. `GEOIP_COUNTRY_CODE` is the server-level
+	 * GeoIP value set on WordPress.com / Atomic (WPCloud); the remaining two are
+	 * common Cloudflare / proxy fallbacks.
+	 *
+	 * @var string[]
+	 */
+	const EDGE_COUNTRY_HEADERS = [ 'GEOIP_COUNTRY_CODE', 'HTTP_CF_IPCOUNTRY', 'HTTP_X_COUNTRY_CODE' ];
```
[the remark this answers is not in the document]
adekbadek: Addressed in c00f87e. Dropped the generic `HTTP_X_COUNTRY_CODE` fallback; the default now trusts only `GEOIP_COUNTRY_CODE` (a non-HTTP server variable a client cannot set via a request header), with other edge headers opt-in through a new `newspack_complianz_edge_country_headers` filter. Verified end-to-end that a client-supplied `CF-IPCountry` header is now ignored (falls back to the site default region).

plugins/newspack-plugin/includes/plugins/class-complianz.php
```diff
@@ -30,6 +40,126 @@ public static function init() {
...
+		// Preserve Complianz's manual region override.
+		if ( isset( $_GET['cmplz_user_region'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			$region = sanitize_title( wp_unslash( $_GET['cmplz_user_region'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			if ( ! cmplz_has_region( $region ) && function_exists( 'cmplz_select_region_outside_supported_regions' ) ) {
+				$region = cmplz_select_region_outside_supported_regions( $region );
+			}
+		}
+
+		return $region;
+	}
```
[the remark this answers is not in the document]
adekbadek: Addressed in c00f87e. `edge_user_region()` now returns the incoming `$region` unchanged when `cmplz_get_region_for_country()` / `cmplz_get_regions()` are unavailable.

plugins/newspack-plugin/includes/plugins/class-complianz.php
```diff
@@ -30,6 +40,126 @@ public static function init() {
...
+			// cmplz_get_consenttype_for_country() is a pure config lookup (no database read).
+			$user_consenttype = cmplz_get_consenttype_for_country( $country_code );
+			if ( in_array( $user_consenttype, cmplz_get_used_consenttypes(), true ) ) {
+				$consenttype = $user_consenttype;
+			} else {
+				$consenttype = 'other';
+			}
+		}
+		return $consenttype;
 	}
```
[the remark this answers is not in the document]
adekbadek: Addressed in c00f87e. `edge_user_consenttype()` now returns the incoming `$consenttype` unchanged when `cmplz_get_consenttype_for_country()` / `cmplz_get_used_consenttypes()` are unavailable.

plugins/newspack-plugin/includes/plugins/class-complianz.php
```diff
@@ -30,6 +55,160 @@ public static function init() {
...
+		 * @param bool $use_edge_geolocation Whether to use edge geolocation. Default true.
+		 */
+		if ( ! apply_filters( 'newspack_complianz_use_edge_geolocation', true ) ) {
+			return;
+		}
+
+		// Only take over a filter we actually removed. If Complianz ever renames or
+		// re-prioritizes its GeoIP callbacks, remove_filter() no-ops and we leave its
+		// native behaviour in place rather than running both filters in parallel.
+		if ( remove_filter( 'cmplz_user_region', 'cmplz_user_region', 20 ) ) {
```
dkoo: **Suggestion:** The `remove_filter( 'cmplz_user_region', 'cmplz_user_region', 20 )` (and the consent-type equivalent at priority 10) hardcodes Complianz Premium's current internal hook priorities. The return-value gate is the right call — if a future Complianz release re-prioritizes those callbacks, `remove_filter` returns `false`, the swap is declined, and native MaxMind behavior is preserved. But that failure is invisible: the optimization just stops working, and the first symptom would be the MaxMind load regression this PR fixes, rediscovered from scratch on a high-traffic site. Consider a one-time `Logger::log()` / `newspack_log` when `cmplz_geoip_enabled()` is true and a valid edge country is present but `remove_filter` returned `false`, so a Complianz update that breaks the coupling is observable rather than a silent regression.

plugins/newspack-plugin/includes/plugins/class-complianz.php
```diff
@@ -30,6 +55,160 @@ public static function init() {
...
+	public static function edge_user_consenttype( $consenttype ) {
+		// Fail safe: if Complianz's helpers are unavailable, leave the consent type untouched.
+		if ( ! function_exists( 'cmplz_get_consenttype_for_country' ) || ! function_exists( 'cmplz_get_used_consenttypes' ) ) {
+			return $consenttype;
+		}
+		$country_code = self::get_edge_country_code();
+		if ( '' !== $country_code ) {
+			// cmplz_get_consenttype_for_country() is a pure config lookup (no database read).
+			$user_consenttype = cmplz_get_consenttype_for_country( $country_code );
+			if ( in_array( $user_consenttype, cmplz_get_used_consenttypes(), true ) ) {
```
dkoo, line 243: **Suggestion:** This `'other'` fallback is spec-valid — `'other'` is a canonical Complianz consent type and Complianz's own `cmplz_sanitize_consenttype()` coerces unknowns to it. The subtlety: it gates the country's resolved type through `cmplz_get_used_consenttypes()` (the site's *configured-region* types), whereas Complianz's sanitizer checks the full type list (`optin`/`optout`/`other`/`optinstats`). A country whose region legitimately resolves to `optinstats` on a site where that type isn't in the used set would be coerced to `'other'` here, while native code would propagate the real type. It's an extreme edge case and `'other'` is still valid, but gating against the full consent-type list (or a brief inline note) would make the parity intent explicit.

plugins/newspack-plugin/tests/unit-tests/plugins/complianz.php
```diff
@@ -215,4 +318,238 @@ public function test_pixel_blocked_when_settings_and_complianz_active() {
...
+	public function test_edge_consenttype_uses_country_when_in_use() {
+		$_SERVER['GEOIP_COUNTRY_CODE'] = 'NL';
+		$this->assertSame( 'optin', Complianz::edge_user_consenttype( 'optout' ) );
+	}
+
+	/**
+	 * A country whose consent type is not configured falls back to 'other'.
+	 */
+	public function test_edge_consenttype_falls_back_to_other() {
+		// 'FR' is not mapped, so cmplz_get_consenttype_for_country returns 'other'.
```
dkoo: **Nit:** The comment reads "'FR' is not mapped, so `cmplz_get_consenttype_for_country` returns 'other'" — but the mock it relies on returns `false` for an unmapped country (with its own comment noting real Complianz returns `false`, not `'other'`). The `'other'` actually comes from the production code's `else` branch, not the helper. The assertion is correct; only the comment misstates where `'other'` originates. A one-line fix avoids future confusion about Complianz's contract.

plugins/newspack-plugin/tests/unit-tests/plugins/complianz.php
```diff
@@ -215,4 +318,238 @@ public function test_pixel_blocked_when_settings_and_complianz_active() {
...
+	public function test_applying_filters_after_swap_yields_edge_values() {
+		self::$geoip_enabled           = true;
+		$_SERVER['GEOIP_COUNTRY_CODE'] = 'NL';
+		$this->register_premium_geoip_filters();
+
+		Complianz::maybe_use_edge_geolocation();
+
+		$this->assertSame( 'eu', apply_filters( 'cmplz_user_region', 'us' ) );
+		$this->assertSame( 'optin', apply_filters( 'cmplz_user_consenttype', 'optout' ) );
+	}
```
dkoo, line 618: **Nit:** Tests cover the `?cmplz_user_region=` override in `edge_user_region` directly and the region/consent-type swap end-to-end, but not the interaction where an override region drives the consent type through `cmplz_get_consenttype_for_country()`'s internal re-application of `cmplz_user_region` (the re-entrant path the docblock calls out). That path is the trickiest behavior in the change and the most likely to break on a Complianz refactor; an assertion for `?cmplz_user_region=us` + edge country `NL` yielding the US consent type would lock it in (the mock helper would need to honor the region filter to exercise it faithfully).

dkoo (approved):
Wasn't able to test manually without Complianz Pro, but the code checks out! No blockers.

Solid, defensively-written optimization — it targets exactly the redundant per-visitor MaxMind read that NPPM-2905 is about, and the fail-safes (helper `function_exists` guards, the `remove_filter`-return-value gate, sentinel rejection of `XX`/`T1`) are the right way to couple to a third-party plugin's internals. I confirmed against Complianz's source that the `'other'` consent-type fallback and the `?cmplz_user_region=` override both match its own behavior, and that nothing in Newspack persists or fans out the resolved region/consent value — a wrong-but-valid value is bounded to a single visitor's single request. No blockers; a few refinements below worth a look before merge. Approving with comments.

plugins/newspack-plugin/includes/plugins/class-complianz.php
adekbadek: Done. `maybe_use_edge_geolocation()` now captures each `remove_filter()` return and, when GeoIP is active with a valid edge country but either callback wasn't found at its expected priority, fires `Logger::newspack_log( 'newspack_complianz_geoip_swap_failed', ... )` so a Complianz release that breaks the coupling surfaces to Manager instead of silently regressing to per-visitor MaxMind lookups. It's rate-limited to once/day via a transient so uncached banner/track requests don't flood the log. Covered by `test_logs_once_when_swap_fails` (also asserts the once-only behaviour) and `test_no_log_when_swap_succeeds`.

plugins/newspack-plugin/includes/plugins/class-complianz.php
adekbadek: I checked the real Complianz Premium source (pro/filters-actions.php) in a running env: native `cmplz_user_consenttype()` does exactly `in_array( $type, cmplz_get_used_consenttypes() )` else `'other'` -- so this code is a faithful mirror, not an accidental narrowing. Widening it to the full `cmplz_sanitize_consenttype()` type list would actually *diverge* from what a non-optimized site shows (your optinstats example would resolve to the real type here but to 'other' natively), reintroducing exactly the drift the swap avoids. So I took the note option deliberately rather than changing the gate, and added an inline comment stating the parity intent and warning against widening it. Flagging that I chose the lesser-looking option on purpose here -- happy to revisit if you read native behaviour differently.

plugins/newspack-plugin/tests/unit-tests/plugins/complianz.php
adekbadek: Fixed. The comment now reads that 'FR' maps to no region so `cmplz_get_consenttype_for_country` returns false, and the 'other' comes from `edge_user_consenttype()`'s own else branch -- not the helper.

plugins/newspack-plugin/tests/unit-tests/plugins/complianz.php
adekbadek: Added `test_edge_consenttype_follows_override_region_through_reentrant_path`: with edge country NL (eu→optin) and `?cmplz_user_region=us`, after the swap it asserts the consent type resolves to US's optout, proving the override region flows through `cmplz_get_consenttype_for_country()`'s internal re-application of `cmplz_user_region`. To exercise it faithfully I made the mock honour the region filter: `cmplz_get_consenttype_for_country` now does `apply_filters( 'cmplz_user_region', cmplz_get_region_for_country( $country ) )` then maps region→type (matching real Complianz's `$regions[$region]['type']`), replacing the old country→type shortcut. Existing consenttype tests still pass unchanged.

PR #369, opened 2026-06-19
feat(insights): first-conversion-date storage helpers for Tab 3

dkoo:
### All Submissions:

Adds two per-customer-set storage helpers to the Insights data layer that return each customer's **first conversion timestamp**:

- `get_first_subscription_order_dates( int[] $customer_ids )` — earliest non-donation subscription start per customer.
- `get_first_donation_order_dates( int[] $customer_ids )` — earliest completed/processing donation order date per customer.

Both return `array<int, \DateTimeImmutable>` keyed by `customer_id` (UTC), are added to the relevant storage interface with HPOS and legacy implementations, and are exposed through uncached pass-throughs on the `Subscribers_Metric` / `Donors_Metric` wrappers.

This is the storage foundation for the Conversion Journey (Tab 3) "Phase B" remediation. A follow-up PR consumes these helpers to compute registration→conversion lag and to anchor BigQuery source-matching to each local record — keeping WP/Woo as the identity spine. The first-conversion definitions deliberately mirror the existing `get_new_subscribers_in_window()` / `get_new_donors_in_window()` aggregates (`MIN(date) … GROUP BY customer_id`), so "first subscription/donation" stays consistent across metrics. **Backend-only; no user-facing behavior changes in this PR.**

Closes NPPD-1616.

1. Check out this branch in the consumer plugin and confirm PHPUnit passes for the new contract tests: from `plugins/newspack-plugin`, run `n test-php --filter 'first_subscription_order_dates'` and `n test-php --filter 'first_donation_order_dates'` — expect 5 passing tests each (interface presence, empty-list guard for both backends, and metric-wrapper delegation).
2. Confirm no regressions in the wider group: `n test-php --group insights` (note: any pre-existing `Source_Matcher not found` errors are an unrelated stale-classmap artifact in local checkouts and clear with `composer dump-autoload`; CI installs a fresh classmap).
3. Confirm the empty-input guard does no DB work: both `get_first_subscription_order_dates([])` and `get_first_donation_order_dates([])` return `[]` immediately (covered by the new tests).
4. Lint on the host (the workspace ruleset resolves from the repo root): from `plugins/newspack-plugin`, run `./vendor/bin/phpcs includes/wizards/insights/storage includes/wizards/insights/metrics` — expect no errors on the changed files.
5. Optional live check (requires WooCommerce data): in an environment with subscription and donation orders, call the two `Subscribers_Metric` / `Donors_Metric` helpers with a known customer-ID set and confirm the returned dates match each customer's earliest subscription start / earliest completed donation order.

### Other information:

<sub>Note for reviewers: the ten new tests are grouped together as one block in the test file rather than distributed across the existing section-header comments — they cover one feature (two symmetric helpers) and read more clearly together.</sub>

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
@@ -1145,6 +1145,11 @@ class HPOS_Storage implements Storage_Interface {
 		// (om.meta_value != '') so a blank value can't yield a bogus epoch
 		// date. The window-count method drops blanks implicitly via its
 		// BETWEEN bounds, so results reconcile on healthy data.
+		// MIN(om.meta_value) is a lexical comparison: wc_orders_meta.meta_value
+		// is a string column, and _schedule_start is stored zero-padded
+		// `Y-m-d H:i:s`, so lexical order equals chronological order. The
+		// donation helper aggregates a real datetime column, so it carries no
+		// such assumption.
 		$sql = "SELECT o.customer_id, MIN(om.meta_value) AS first_start
 			FROM {$prefix}wc_orders o
 			JOIN {$prefix}wc_orders_meta om
--- a/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
@@ -1140,7 +1140,11 @@ class HPOS_Storage implements Storage_Interface {
 
 		// Earliest non-donation subscription _schedule_start per customer,
 		// scoped to the given customer set. Mirrors the inner aggregate of
-		// get_new_subscribers_in_window() (same first-start definition).
+		// get_new_subscribers_in_window() (same first-start definition), with
+		// one added guard: this query also excludes empty _schedule_start
+		// (om.meta_value != '') so a blank value can't yield a bogus epoch
+		// date. The window-count method drops blanks implicitly via its
+		// BETWEEN bounds, so results reconcile on healthy data.
 		$sql = "SELECT o.customer_id, MIN(om.meta_value) AS first_start
 			FROM {$prefix}wc_orders o
 			JOIN {$prefix}wc_orders_meta om
--- a/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
@@ -1121,4 +1121,48 @@ class HPOS_Storage implements Storage_Interface {
 			$rows
 		);
 	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @param int[] $customer_ids Customer IDs to look up.
+	 * @return array<int, \DateTimeImmutable>
+	 */
+	public function get_first_subscription_order_dates( array $customer_ids ): array {
+		if ( empty( $customer_ids ) ) {
+			return [];
+		}
+
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+		$ids       = $this->id_list( $customer_ids );
+
+		// Earliest non-donation subscription _schedule_start per customer,
+		// scoped to the given customer set. Mirrors the inner aggregate of
+		// get_new_subscribers_in_window() (same first-start definition).
+		$sql = "SELECT o.customer_id, MIN(om.meta_value) AS first_start
+			FROM {$prefix}wc_orders o
+			JOIN {$prefix}wc_orders_meta om
+				ON om.order_id = o.id AND om.meta_key = '_schedule_start'
+			JOIN {$prefix}woocommerce_order_items oi
+				ON oi.order_id = o.id AND oi.order_item_type = 'line_item'
+			JOIN {$prefix}woocommerce_order_itemmeta oim
+				ON oim.order_item_id = oi.order_item_id AND oim.meta_key = '_product_id'
+			WHERE o.type = 'shop_subscription'
+			  AND oim.meta_value NOT IN ($donations)
+			  AND om.meta_value != ''
+			  AND o.customer_id IN ($ids)
+			GROUP BY o.customer_id";
+
+		$rows = $wpdb->get_results( $sql, ARRAY_A );
+		$map  = [];
+		foreach ( (array) $rows as $row ) {
+			if ( empty( $row['first_start'] ) ) {
+				continue;
+			}
+			$map[ (int) $row['customer_id'] ] = new \DateTimeImmutable( $row['first_start'], new \DateTimeZone( 'UTC' ) );
+		}
+		return $map;
+	}
 }
--- a/plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-donors-storage.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-donors-storage.php
@@ -923,7 +923,9 @@ class Legacy_Donors_Storage implements Donors_Storage_Interface {
 		// Legacy CPT equivalent of HPOS_Donors_Storage::get_first_donation_order_dates():
 		// earliest completed/processing donation order post_date_gmt per
 		// _customer_user, scoped to the given set. Mirrors get_new_donors_in_window().
-		$sql = "SELECT cust.meta_value AS customer_id, MIN(p.post_date_gmt) AS first_donation_date
+		// _customer_user is cast to UNSIGNED (it is stored as a string) to match the
+		// other list-scoped legacy donor queries and align grouping with the int keys returned.
+		$sql = "SELECT CAST(cust.meta_value AS UNSIGNED) AS customer_id, MIN(p.post_date_gmt) AS first_donation_date
 			FROM {$prefix}posts p
 			JOIN {$prefix}postmeta cust
 				ON cust.post_id = p.ID AND cust.meta_key = '_customer_user'
@@ -931,8 +933,8 @@ class Legacy_Donors_Storage implements Donors_Storage_Interface {
 			WHERE p.post_type = 'shop_order'
 			  AND p.post_status IN ('wc-completed', 'wc-processing')
 			  AND opl.product_id IN ($donations)
-			  AND cust.meta_value IN ($ids)
-			GROUP BY cust.meta_value";
+			  AND CAST(cust.meta_value AS UNSIGNED) IN ($ids)
+			GROUP BY CAST(cust.meta_value AS UNSIGNED)";
 
 		$rows = $wpdb->get_results( $sql, ARRAY_A );
 		$map  = [];
--- a/plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-donors-storage.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-donors-storage.php
@@ -905,6 +905,47 @@ class Legacy_Donors_Storage implements Donors_Storage_Interface {
 	}
 
 	/**
+	 * {@inheritDoc}
+	 *
+	 * @param int[] $customer_ids Customer IDs to look up.
+	 * @return array<int, \DateTimeImmutable>
+	 */
+	public function get_first_donation_order_dates( array $customer_ids ): array {
+		if ( empty( $customer_ids ) ) {
+			return [];
+		}
+
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+		$ids       = $this->id_list( $customer_ids );
+
+		// Legacy CPT equivalent of HPOS_Donors_Storage::get_first_donation_order_dates():
+		// earliest completed/processing donation order post_date_gmt per
+		// _customer_user, scoped to the given set. Mirrors get_new_donors_in_window().
+		$sql = "SELECT cust.meta_value AS customer_id, MIN(p.post_date_gmt) AS first_donation_date
+			FROM {$prefix}posts p
+			JOIN {$prefix}postmeta cust
+				ON cust.post_id = p.ID AND cust.meta_key = '_customer_user'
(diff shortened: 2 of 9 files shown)


plugins/newspack-plugin/includes/wizards/insights/storage/class-storage-interface.php
```diff
@@ -303,4 +303,20 @@ public function count_active_non_donation_subscribers_by_customer_ids( array $cu
...
 	 */
 	public function get_stale_registered_users(): int;
+
+	/**
+	 * Earliest non-donation subscription start (`_schedule_start`) per customer,
+	 * restricted to the given customer set. Same first-start-per-customer
+	 * definition as {@see get_new_subscribers_in_window()}, but returns the
+	 * dates rather than counting a window. Used by Tab 3 (Conversion Journey)
+	 * to compute registration→subscription lag and to anchor the BQ
+	 * source-match window.
```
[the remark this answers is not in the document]
dkoo: Good call — clarified in 5b6aeb9. The docblock now notes that the implementations additionally exclude empty `_schedule_start` (so a blank value can't yield a bogus epoch date), and that the window-count metric drops blanks implicitly via its `BETWEEN` bounds, so the two reconcile on healthy data.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -1121,4 +1121,48 @@ function ( $row ) {
...
+		}
+
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+		$ids       = $this->id_list( $customer_ids );
+
+		// Earliest non-donation subscription _schedule_start per customer,
+		// scoped to the given customer set. Mirrors the inner aggregate of
+		// get_new_subscribers_in_window() (same first-start definition).
```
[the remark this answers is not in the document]
dkoo: Done in 5b6aeb9 — the inline comment now spells out the added `om.meta_value != ''` guard and how it relates to the window-count method's implicit `BETWEEN` exclusion.

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
```diff
@@ -1053,4 +1053,50 @@ function ( $row ) {
...
+		}
+
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+		$ids       = $this->id_list( $customer_ids );
+
+		// Legacy CPT equivalent of HPOS_Storage::get_first_subscription_order_dates():
+		// earliest non-donation subscription _schedule_start per _customer_user,
+		// scoped to the given customer set. Mirrors get_new_subscribers_in_window().
```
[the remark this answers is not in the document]
dkoo: Done in 5b6aeb9 — the comment now calls out the `start.meta_value != ''` guard and notes Tab 6's window logic drops blanks implicitly via `BETWEEN` instead.

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
```diff
@@ -1053,4 +1053,50 @@ function ( $row ) {
...
+				ON start.post_id = p.ID AND start.meta_key = '_schedule_start'
+			JOIN {$prefix}woocommerce_order_items oi
+				ON oi.order_id = p.ID AND oi.order_item_type = 'line_item'
+			JOIN {$prefix}woocommerce_order_itemmeta oim
+				ON oim.order_item_id = oi.order_item_id AND oim.meta_key = '_product_id'
+			WHERE p.post_type = 'shop_subscription'
+			  AND oim.meta_value NOT IN ($donations)
+			  AND start.meta_value != ''
+			  AND cust.meta_value IN ($ids)
+			GROUP BY cust.meta_value";
```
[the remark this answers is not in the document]
dkoo: Agreed — fixed in 5b6aeb9. `_customer_user` is now `CAST(... AS UNSIGNED)` in SELECT/WHERE/GROUP BY, matching `count_active_non_donation_subscribers_by_customer_ids` and aligning grouping with the int keys returned to callers.

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-donors-storage.php
```diff
@@ -904,6 +904,47 @@ public function get_donations_by_tier( DateTimeInterface $start, DateTimeInterfa
...
+		$sql = "SELECT cust.meta_value AS customer_id, MIN(p.post_date_gmt) AS first_donation_date
+			FROM {$prefix}posts p
+			JOIN {$prefix}postmeta cust
+				ON cust.post_id = p.ID AND cust.meta_key = '_customer_user'
+			JOIN {$prefix}wc_order_product_lookup opl ON opl.order_id = p.ID
+			WHERE p.post_type = 'shop_order'
+			  AND p.post_status IN ('wc-completed', 'wc-processing')
+			  AND opl.product_id IN ($donations)
+			  AND cust.meta_value IN ($ids)
+			GROUP BY cust.meta_value";
```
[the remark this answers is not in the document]
dkoo: Agreed — fixed in 5b6aeb9, same `CAST(... AS UNSIGNED)` treatment in SELECT/WHERE/GROUP BY to match the other list-scoped legacy donor queries.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-journey-storage.php
```diff
@@ -348,6 +348,110 @@ public function test_donors_metric_count_completed_donation_delegates(): void {
 		$this->assertSame( 0, $metric->count_completed_donation_order_customers_by_customer_ids( [] ) );
 	}
 
+	/**
+	 * HPOS_Storage exposes get_first_subscription_order_dates().
+	 */
+	public function test_hpos_storage_implements_first_subscription_order_dates(): void {
```
dkoo, line 354: **Severity:** suggestion

The four new methods have no DB-backed test coverage — these tests assert method existence, the empty-list guard, and mock-based wrapper delegation, but none exercise the actual aggregate against seeded orders. A regression in the `GROUP BY` / `MIN` / status filter / donation product-set filter, or an HPOS-vs-legacy divergence on identical data, would pass CI undetected. Consider a DB-backed integration test that seeds one customer with both a donation and a non-donation subscription and asserts both backends return the same `DateTimeImmutable`. (I verified both paths manually against real seeded data for this review, which covers the immediate case but not future drift.)

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
```diff
@@ -1053,4 +1053,50 @@ function ( $row ) {
...
+
+		global $wpdb;
+		$prefix    = $wpdb->prefix;
+		$donations = $this->id_list( $this->donation_product_ids );
+		$ids       = $this->id_list( $customer_ids );
+
+		// Legacy CPT equivalent of HPOS_Storage::get_first_subscription_order_dates():
+		// earliest non-donation subscription _schedule_start per _customer_user,
+		// scoped to the given customer set. Mirrors get_new_subscribers_in_window().
+		$sql = "SELECT cust.meta_value AS customer_id, MIN(start.meta_value) AS first_start
```
dkoo: **Severity:** nit

`MIN(start.meta_value)` here (and `MIN(p.post_date_gmt)` in `Legacy_Donors_Storage::get_first_donation_order_dates()`) is a lexicographic comparison over a string meta column. It's correct only because `_schedule_start` / `post_date_gmt` are stored as zero-padded `Y-m-d H:i:s`, where lexical order equals chronological order. A one-line comment noting that assumption would protect against a future format change. The HPOS path is unaffected (real `datetime` column).

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
dkoo: **Severity:** nit

Legacy guest orders persist `_customer_user = '0'`; if `0` is ever passed in `$customer_ids`, those rows would group under a bogus `0` key. Callers pass real registered-reader IDs so this isn't reachable today, and it matches the sibling list-scoped queries — but a defensive positive-int filter at the boundary would make the contract explicit. (Distinct from the existing suggestion on this line to cast `_customer_user` to UNSIGNED.)

dkoo (commented):
Clean, well-scoped, low-risk addition that faithfully extends the existing Insights storage pattern. The SQL is interpolated only with `intval`-cast values (no injection surface), HPOS/legacy parity holds, and a live legacy-backend check against seeded orders — including a cancelled donation order and a donation-product subscription dated earlier than the valid records — returned each customer's correct earliest timestamp in UTC, with the status and donation-exclusion filters behaving correctly and empty input returning `[]`. No blockers from review or testing — recommending merge once the note below is considered.

One thing worth a glance before merge: the four new SQL paths have no DB-backed regression test, so a future change to the aggregate could regress silently (details inline). Two minor nits also noted. Heads-up that the closing reference `NPPD-1616` points at a Tab 6 UI ticket that's already Done — looks like a mis-link for this Tab 3 storage work.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-journey-storage.php
dkoo, line 354: **Severity:** suggestion

The four new methods have no DB-backed test coverage — these tests assert method existence, the empty-list guard, and mock-based wrapper delegation, but none exercise the actual aggregate against seeded orders. A regression in the `GROUP BY` / `MIN` / status filter / donation product-set filter, or an HPOS-vs-legacy divergence on identical data, would pass CI undetected. Consider a DB-backed integration test that seeds one customer with both a donation and a non-donation subscription and asserts both backends return the same `DateTimeImmutable`. (I verified both paths manually against real seeded data for this review, which covers the immediate case but not future drift.)

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
```diff
@@ -1053,4 +1053,55 @@ function ( $row ) {
...
+
+		// Legacy CPT equivalent of HPOS_Storage::get_first_subscription_order_dates():
+		// earliest non-donation subscription _schedule_start per _customer_user,
+		// scoped to the given customer set. Mirrors get_new_subscribers_in_window(),
+		// with one added guard: this query also excludes empty _schedule_start
+		// (start.meta_value != '') so a blank value can't yield a bogus epoch date
+		// — the window-count method drops blanks implicitly via its BETWEEN bounds.
+		// _customer_user is cast to UNSIGNED (it is stored as a string) to match the
+		// other list-scoped legacy queries and align grouping with the int keys returned.
+		$sql = "SELECT CAST(cust.meta_value AS UNSIGNED) AS customer_id, MIN(start.meta_value) AS first_start
```
dkoo, line 1085: **Severity:** nit

`MIN(start.meta_value)` here (and `MIN(p.post_date_gmt)` in `Legacy_Donors_Storage::get_first_donation_order_dates()`, line 928) is a lexicographic comparison over a string meta column. It's correct only because `_schedule_start` / `post_date_gmt` are stored as zero-padded `Y-m-d H:i:s`, where lexical order equals chronological order — so results agree with the HPOS path (real `datetime` column). A one-line comment noting that assumption would guard against a future format change.

dkoo (commented):
Re-reviewing against the current head (`5b6aeb9c8`); this supersedes my earlier review, which was posted against the prior commit `fcc51a19d` and now shows as outdated. The latest commit resolves the earlier round of feedback — `_customer_user` is now cast to `UNSIGNED` in both legacy queries (aligning grouping with the integer keys returned), and the empty-`_schedule_start` guard plus its parity caveat vs `get_new_subscribers_in_window()` are documented in the interface and the implementations.

The change remains clean, well-scoped, and low-risk: the SQL is interpolated only with `intval`-cast values (no injection surface) and HPOS/legacy parity holds. Verified live on the legacy backend against real WooCommerce data — donation orders and non-donation subscriptions created through the real code paths (WooCommerce Subscriptions + Newspack donation products): both helpers return each customer's correct earliest timestamp in UTC, `MIN` correctly selects the earliest of multiple records, donation subscriptions are correctly excluded from the subscription helper, output matches an independent `MIN(...) GROUP BY` cross-check, and empty input returns `[]`. No blockers — recommending merge.

Two minor, non-blocking items remain (inline).

plugins/newspack-plugin/includes/wizards/insights/storage/class-legacy-storage.php
dkoo: Added in 6068930 — with a small correction after verifying the column types. `_schedule_start` is a *string* column in **both** backends (`wp_postmeta.meta_value` is `longtext`, `wp_wc_orders_meta.meta_value` is `text`), so the lexical-vs-chronological assumption applies to both subscription helpers — I added the note to `Legacy_Storage` **and** `HPOS_Storage::get_first_subscription_order_dates()`. The donation helpers aggregate real `datetime` columns (`wp_posts.post_date_gmt` and `wp_wc_orders.date_created_gmt`), so `MIN()` there is natively chronological and needs no caveat — left those untouched.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-journey-storage.php
dkoo: Agreed this is a real gap, and a good future addition. Deferring it for now rather than adding it in this PR: the insights PHPUnit bootstrap deliberately doesn't install the WooCommerce DB tables, so the existing Tab 6/7 storage methods have no SQL-body tests either — a DB-backed integration test would need WC-table seeding wired into the harness, which is a larger change than this storage slice warrants. Both backends were verified live against seeded orders for this review (including the MIN/donation-exclusion/status cases), so the immediate behavior is covered; the drift risk you describe is the part that stays open. Leaving this thread unresolved as a tracked follow-up for a DB-backed insights test harness — your call on whether to file it separately.

PR #372, opened 2026-06-19
ci(post-release): name conflicting files in merge-failure Slack alert

jason10lee:
### All Submissions:

When the post-release `release → alpha`/`release → main` back-merge hits a conflict, the Slack alert previously said only that a merge failed and linked the build — leaving the reader to open the Actions log to find out what conflicted or who to involve. This makes the alert self-contained and routable.

The alert now names each conflicting file and, for each, the **incoming release-side PR (as a clickable link) and its author**, so a reader can tell what needs reconciling and who to ping straight from the message:

> ⚠️ Post-release merge to `main` failed for: `Automattic/newspack-workspace`.
> Conflicting files:
> • `path/to/file` — incoming: [#000](https://github.com/Automattic/newspack-workspace/pull/000) (PR Author)
> Check <…|the build> for details.

Attribution identifies the *incoming* change being merged forward (not sole blame — the merge is mutual), resolved from the squash commit that last touched each file.

Built to be robust on this already-failed path:

- **Size-safe.** The named list is capped (first 10 files + "…and N more"); a hard line-boundary trim backstops Slack's 3000-char section limit so a many-conflict merge can't produce an oversized payload that Slack rejects (which would drop the *entire* alert). The build link is always preserved.
- **JSON-safe.** The payload is built with node's `JSON.stringify` rather than hand-rolled string interpolation, correctly escaping the variable-length file list.
- **Always notifies.** If the node payload build fails, it falls back to a minimal hand-rolled alert instead of going silent; it never aborts the script (`set -e`) or leaves subsequent merges / status reporting unreported.
- **Edge cases.** Non-ASCII paths stay literal (`core.quotePath=false`) so attribution matches; PR/author parsing is robust to merge commits (`--no-merges`), tabs in commit subjects, and issue-vs-PR refs; gracefully degrades when a commit has no `(#NNN)`.

When there are no conflicting paths to report, the message is unchanged from before.

1. In a throwaway git repo, create a `release` branch and a default branch that both modify the same line of a file, so `git merge release` conflicts; give the release-side commit a `(#NNN)` subject.
2. Run `post-release.sh` (or exercise `notify_slack` / `attribute_conflicts`) with Slack env vars pointed at a test channel, and trigger the failing merge path.
3. Confirm the conflicting path is captured *before* `git merge --abort`, and that the posted message lists it under `Conflicting files:` with the incoming PR rendered as a clickable Slack link (`<url|#NNN>`) and the author name.
4. Confirm size-safety: with many (or pathologically long-pathed) conflicts, the rendered text stays ≤ 2900 chars, the file list is capped with "…and N more", and the build link survives (backticks stay balanced).
5. Confirm graceful degradation: a no-conflict failure path renders the previous (unchanged) message; a commit with no `(#NNN)` shows just the file (or author); a forced node failure still posts the minimal fallback alert.
6. Confirm the generated payload is valid JSON in the multi-file, no-conflict, and fallback cases.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? (No test harness exists for `post-release.sh`; verified manually — `bash -n`, `shellcheck`, node payload generation across size/edge cases, and simulated conflicting merges.)


.github/scripts/post-release.sh
```diff
@@ -89,16 +89,130 @@ restore_workspace_deps_and_commit() {
...
+# "<file><TAB><PR><TAB><author>" row per input path ($1 = newline-separated
+# paths). This identifies the *incoming* change being merged forward, not sole
+# blame — the merge is mutual (the target branch changed the file too). PR and
+# author are left empty when unresolvable (commit has no "(#NNN)", etc.).
+attribute_conflicts() {
+  local files="$1"
+  [ -z "$files" ] && return 0
+  local mb f meta subj author pr
+  mb=$(git merge-base HEAD release 2>/dev/null || true)
+  while IFS= read -r f; do
```
dkoo, line 103: **Nit:** Paths with embedded newlines would mis-parse (pathological) — Both the upstream `git diff --name-only` capture (lines 245/262) and `attribute_conflicts`’ `while IFS= read -r f` loop over `<<< "$files"` split on newlines, so a conflicting path containing a literal `\n` would be split into two phantom entries. `core.quotePath=false` deliberately does *not* escape newlines, so this is a real (if pathological) gap. In practice nobody commits a newline in a path, and the blast radius is only a cosmetically-wrong Slack alert on an already-failed job — not worth fixing, noted for completeness.

.github/scripts/post-release.sh
```diff
@@ -89,16 +89,130 @@ restore_workspace_deps_and_commit() {
...
+      meta=$(git log "$mb"..release -1 --no-merges --format='%s%x09%an' -- "$f" 2>/dev/null || true)
+      if [ -n "$meta" ]; then
+        # Split from the right: the format is "<subject><TAB><author>" and an
+        # author name has no tab, so a literal tab inside the subject can't
+        # corrupt the author field.
+        subj=${meta%$'\t'*}
+        author=${meta##*$'\t'}
+        # Prefer the trailing "(#NNN)" squash-merge PR ref; fall back to the last
+        # bare "#NNN" so an issue ref earlier in the subject isn't mistaken for it.
+        pr=$(printf '%s' "$subj" | grep -oE '\(#[0-9]+\)' | tail -1 | tr -cd '0-9' || true)
```
dkoo, line 118: **Nit:** `tr -cd ‘0-9’` is safe only because `tail -1` runs first — `grep -oE ‘\(#[0-9]+\)’ | tail -1 | tr -cd ‘0-9’` is correct because `tail -1` reduces the match to a single `(#NNN)` before `tr` strips to digits. If `tail` ever emitted more than one line (it can’t here), `tr -cd` would concatenate all digits into one bogus PR number. The safety lives in a different pipe stage than the extraction, which makes the idiom slightly fragile to future edits. Cosmetic only.

.github/scripts/post-release.sh
```diff
@@ -89,16 +89,130 @@ restore_workspace_deps_and_commit() {
...
+      channel: process.env.SLACK_CHANNEL_ID,
+      blocks: [{ type: "section", text: { type: "mrkdwn", text } }],
+    }));
+  ') || {
+    # node missing/erroring must not abort the script (set -e) on this
+    # already-failed path. Don't go silent either: fall back to a minimal
+    # hand-rolled alert (fixed text, no variable conflict list → no JSON-escaping
+    # hazard) so the team is still notified the merge failed.
+    echo "[post-release] Slack payload build failed; sending minimal fallback alert."
+    payload="{\"channel\":\"$SLACK_CHANNEL_ID\",\"blocks\":[{\"type\":\"section\",\"text\":{\"type\":\"mrkdwn\",\"text\":\"⚠️ Post-release merge to \`$target\` failed for: \`$GITHUB_REPOSITORY\`. Check <$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|the build> for details.\"}}]}"
```
dkoo, line 208: **Nit:** Fallback alert hand-rolls JSON — safety rests on an external constraint — The always-notify fallback interpolates `$target`, `$SLACK_CHANNEL_ID`, `$GITHUB_REPOSITORY`, etc. directly into a JSON string literal. This is safe today: those values are constrained by GitHub/git to `[A-Za-z0-9._/-]`, so no JSON metacharacter (`"`, `\`, newline) can appear. Worth a one-line note that the safety depends on that external constraint rather than on escaping — if this fallback shape were ever copied to a context with looser inputs it would break. No change needed.

dkoo (approved):
Didn't attempt to manually test—code review only. Approve — nits only, no blockers.

This is a careful, heavily-commented change that gets the tricky Bash right. `local payload` is declared separately from its assignment, so the `|| { fallback }` actually fires when node errors (a `local payload=$(…)` would have masked node’s exit status). Untrusted content (filenames, author names, commit subjects) reaches node exclusively via `process.env` — never a shell `eval` — so there’s no command-injection surface. The size-trim is line-boundary-aware and keeps the build link and backticks balanced.

Importantly, the failure signal never depends on the new naming logic being correct: on any failure of the enriched alert, the job still exits non-zero via `sync_failed`, the run still goes red, and a minimal fallback alert is still posted. The three inline notes are cosmetic and need no action before merge.

.github/scripts/post-release.sh
jason10lee: No argument here, including on "not worth fixing." A newline in a tracked path would have to get past everything upstream of this script first, and the worst case is a cosmetically wrong Slack line on a job that already failed.

.github/scripts/post-release.sh
jason10lee: Leaving it. `tail -1` can't emit more than one line, so the extraction is correct as written, and moving the safety into the same pipe stage would be churn on a line that works.

.github/scripts/post-release.sh
jason10lee: Leaving the fallback as-is, but the invariant is worth stating somewhere, so: `$target`, `$SLACK_CHANNEL_ID` and `$GITHUB_REPOSITORY` are constrained to `[A-Za-z0-9._/-]` by GitHub and git, which is why no JSON metacharacter can reach the string literal. That's an external guarantee the line doesn't enforce itself — worth knowing before this shape gets copied somewhere with looser inputs. Recording it here rather than spending a CI cycle on a comment-only commit against an approved PR.

PR #373, opened 2026-06-19
feat(insights): conversion source-mix + time-to-convert distributions

dkoo:
### All Submissions:

Brings five Insights → Conversion Journey (Tab 3) metrics from "coming soon" to live data, built on a WP/Woo identity spine with BigQuery supplying only the conversion **source** (gate / prompt / direct), attached to each local record by a timestamp-window match — no GA cookie is ever cast to a Woo customer ID.

- **3.2 / 3.3 Source mix (new subscribers / new donors)** — remediates the production bug where these returned ~0. Counts now come from Woo conversions in the window; each conversion's source is matched to the BigQuery checkout-exposure event just before it.
- **4.2 / 4.3 Time to subscribe / donate** — cumulative "what share converted by day N" curves, split by source. Registration→conversion lag comes from Woo; the per-source split comes from a BigQuery registration-source lookup behind a cheap presence probe. When that source layer is absent or unavailable, the curve still renders (unsegmented) instead of disappearing.
- **4.4 Subscriber → donor lag** — a pure-Woo cumulative curve of how long subscribers take to also donate, hidden until at least 50 readers have done both.

4.2 / 4.3 / 4.4 always use **all available history** regardless of the selected date range (a caption now says so on each); 3.2 / 3.3 follow the dashboard date range. This is primarily a backend change plus that one caption — the chart components and response types were already in place.

Part of NPPD-1692.

1. Check out this branch and run the PHP unit tests from `plugins/newspack-plugin`: `n test-php --filter Test_Conversion_Metric` (expect 87 passing) and `n test-php --group insights` (no regressions).
2. Run the JS tests: `npm test -- HowLongConversionsTakeSection` — the new test asserts the all-history caption renders on exactly the three snapshot charts (4.2 / 4.3 / 4.4), not on 4.1.
3. Lint on the host: from `plugins/newspack-plugin`, `./vendor/bin/phpcs includes/wizards/insights/storage includes/wizards/insights/metrics` and `npm run lint:js -- src/wizards/insights/tabs/conversion/HowLongConversionsTakeSection.tsx` — expect no errors.
4. Confirm graceful degradation logic: the time-to-convert curves fall back to a single unsegmented (all-"direct") series when the BigQuery presence probe returns zero or the source query errors (covered by the `*_probe_zero_all_direct` and `*_registrations_error_all_direct` unit tests).
5. **Live verification (required before this ships, after the hub deploys):** on a real Audience site (e.g. blockclub), open the Conversion Journey tab and confirm 3.2 / 3.3 show non-zero, sensible source splits (vs the current ~0), the 4.2 / 4.3 / 4.4 curves populate, and 4.4 stays hidden below 50 cross-converters. The local dev site cannot exercise the hub BigQuery queries (it loads `newspack-manager-client`, not the hub).

### Other information:

<sub>Notes for reviewers: BigQuery `event_timestamp` is microseconds — converted to seconds before timestamp matching. Storage epochs are built in PHP from UTC datetime strings (not SQL `UNIX_TIMESTAMP`, which would use the session timezone). The new `rows_to_records` / `rows_to_lags` helpers are duplicated per storage class to match the existing self-contained convention (no shared base; `id_list`/`fmt` are likewise duplicated). `Woo_Order_Resolver` is intentionally left in place — it still backs the deferred influenced metrics and is removed in a later PR. Woo SQL bodies are live-verified rather than unit-tested (no WooCommerce tables in the PHPUnit bootstrap; Phase A precedent).</sub>

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
@@ -746,6 +746,129 @@ final class Conversion_Metric {
 	}
 
 	/**
+	 * Empirical CDF over a list of day-lags: one point per distinct day that has
+	 * conversions, with the running fraction of the cohort converted by that day.
+	 *
+	 * @param int[] $days_values Day lags (already truncated to the lookback).
+	 * @return array<int, array{day:int, cumulative_pct:float}>
+	 */
+	private function cumulative_distribution( array $days_values ): array {
+		$total = count( $days_values );
+		if ( 0 === $total ) {
+			return [];
+		}
+		$per_day = [];
+		foreach ( $days_values as $day ) {
+			$day             = (int) $day;
+			$per_day[ $day ] = ( $per_day[ $day ] ?? 0 ) + 1;
+		}
+		ksort( $per_day );
+		$running = 0;
+		$points  = [];
+		foreach ( $per_day as $day => $count ) {
+			$running += $count;
+			$points[] = [
+				'day'            => (int) $day,
+				'cumulative_pct' => round( $running / $total, 4 ),
+			];
+		}
+		return $points;
+	}
+
+	/**
+	 * Registration source events for the time-to-convert distributions, behind
+	 * the BQ probe gate. Window is a trailing 365 days ending now (snapshot).
+	 * Returns [ ['ts'=>int seconds, 'source'=>string], … ], or [] when the probe
+	 * reports no registrations or any BQ call fails (degrade to all-direct).
+	 *
+	 * @return array<int, array{ts:int, source:string}>
+	 */
+	private function registration_source_events(): array {
+		$utc   = new \DateTimeZone( 'UTC' );
+		$end   = new \DateTimeImmutable( 'now', $utc );
+		$start = $end->modify( '-365 days' );
+
+		$probe = $this->proxy->query( 'conversion_journey_has_registrations_in_window', $start, $end );
+		if ( is_wp_error( $probe ) || ! is_array( $probe ) || empty( $probe[0] ) || ! is_array( $probe[0] ) ) {
+			return [];
+		}
+		$first_row = $probe[0];
+		$count     = (int) reset( $first_row ); // Single COUNT column; read defensively regardless of alias.
+		if ( $count <= 0 ) {
+			return [];
+		}
+
+		$reg = $this->proxy->query( 'conversion_journey_registrations_with_source', $start, $end );
+		if ( is_wp_error( $reg ) || ! is_array( $reg ) ) {
+			return [];
+		}
+		$events = [];
+		foreach ( $reg as $row ) {
+			if ( ! is_array( $row ) ) {
+				continue;
+			}
+			$events[] = [
+				'ts'     => intdiv( (int) ( $row['reg_ts'] ?? 0 ), 1000000 ),
+				'source' => (string) ( $row['source'] ?? Source_Matcher::SOURCE_DIRECT ),
+			];
+		}
+		return $events;
+	}
+
+	/**
+	 * Build a multi-series time-to-convert distribution. Lag = first_ts −
+	 * registered_ts in whole days, kept to [0,365]. Each reader's source comes
+	 * from matching their user_registered to a BQ registration event within
+	 * ±WINDOW_REGISTRATION_SECONDS; unmatched → direct. Always emits the three
+	 * source groups (zero-filled when empty).
+	 *
+	 * @param array<int, array<string,int>> $rows         Conversion-lag rows.
+	 * @param string                        $first_ts_key Row key for the first-conversion epoch.
+	 * @return array{state:string, groups:array}
+	 */
+	private function build_time_to_convert_distribution( array $rows, string $first_ts_key ): array {
+		$events          = $this->registration_source_events();
+		$records         = [];
+		$lag_by_customer = [];
+		foreach ( $rows as $row ) {
+			$lag = intdiv( (int) $row[ $first_ts_key ] - (int) $row['registered_ts'], 86400 );
+			if ( $lag < 0 || $lag > 365 ) {
+				continue;
+			}
+			$cid                     = (int) $row['customer_id'];
+			$records[]               = [
+				'key' => $cid,
+				'ts'  => (int) $row['registered_ts'],
+			];
+			$lag_by_customer[ $cid ] = $lag;
+		}
+
+		$map         = Source_Matcher::attach_sources( $records, $events, Source_Matcher::WINDOW_REGISTRATION_SECONDS, Source_Matcher::WINDOW_REGISTRATION_SECONDS );
+		$days_by_src = [
+			'gate'   => [],
+			'prompt' => [],
+			'direct' => [],
+		];
+		foreach ( $lag_by_customer as $cid => $lag ) {
+			$source                   = $map[ $cid ] ?? Source_Matcher::SOURCE_DIRECT;
+			$days_by_src[ $source ][] = $lag;
+		}
+
+		$groups = [];
+		foreach ( self::SOURCES as $source ) {
+			$groups[] = [
+				'label'  => $source,
+				'points' => $this->cumulative_distribution( $days_by_src[ $source ] ),
+			];
+		}
+
+		return [
+			'state'  => empty( $lag_by_customer ) ? 'empty' : 'populated',
+			'groups' => $groups,
+		];
+	}
+
+	/**
 	 * Source-mix via the Woo identity spine. The BQ query supplies exposure
 	 * events (attempt_ts in microseconds + gate/popup params); each Woo
 	 * conversion record is matched to the nearest preceding event within
@@ -892,32 +1015,36 @@ final class Conversion_Metric {
 	}
 
 	/**
-	 * Time-to-subscribe cumulative distribution (4.2) — three series by
-	 * source (gate / prompt / direct).
+	 * Time-to-subscribe cumulative distribution (4.2) — three series by source.
+	 * Snapshot: ignores the window (all-history). May become windowed later
+	 * (e.g. holiday-season period comparison).
 	 *
-	 * Phase B `coming_soon` placeholder. Not yet wired to BigQuery.
-	 *
-	 * @param DateTimeInterface $start Window start.
-	 * @param DateTimeInterface $end   Window end.
+	 * @param DateTimeInterface $start Window start (ignored — snapshot).
+	 * @param DateTimeInterface $end   Window end (ignored — snapshot).
 	 * @return array{state: string, groups: array}
 	 */
 	public function get_time_to_subscribe_distribution( DateTimeInterface $start, DateTimeInterface $end ): array {
 		unset( $start, $end );
-		return $this->coming_soon_collection( 'groups' );
+		return $this->build_time_to_convert_distribution(
+			$this->subscribers_metric->get_subscription_conversion_lags(),
+			'first_sub_ts'
+		);
 	}
 
 	/**
 	 * Time-to-donate cumulative distribution (4.3) — three series by source.
+	 * Snapshot: ignores the window (all-history). May become windowed later.
 	 *
-	 * Phase B `coming_soon` placeholder. Not yet wired to BigQuery.
-	 *
-	 * @param DateTimeInterface $start Window start.
-	 * @param DateTimeInterface $end   Window end.
+	 * @param DateTimeInterface $start Window start (ignored — snapshot).
+	 * @param DateTimeInterface $end   Window end (ignored — snapshot).
 	 * @return array{state: string, groups: array}
 	 */
 	public function get_time_to_donate_distribution( DateTimeInterface $start, DateTimeInterface $end ): array {
 		unset( $start, $end );
-		return $this->coming_soon_collection( 'groups' );
+		return $this->build_time_to_convert_distribution(
+			$this->donors_metric->get_donation_conversion_lags(),
+			'first_donation_ts'
+		);
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
@@ -54,6 +54,7 @@ defined( 'ABSPATH' ) || exit;
 use DateTimeInterface;
 use Newspack\Insights\BigQuery_Proxy_Client;
 use Newspack\Insights\Donors_Metric;
+use Newspack\Insights\Source_Matcher;
 use Newspack\Insights\Subscribers_Metric;
 use Newspack\Insights\Woo_Order_Resolver;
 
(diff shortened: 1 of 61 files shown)


plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
```diff
@@ -699,103 +700,228 @@ public function get_source_mix_registrations( DateTimeInterface $start, DateTime
...
+	 * @param array<int, array{customer_id:int, ts:int}> $records    Woo conversion records.
+	 * @param DateTimeInterface                          $start      Window start.
+	 * @param DateTimeInterface                          $end        Window end.
+	 * @return array
+	 */
+	private function compute_source_mix( string $query_name, array $records, DateTimeInterface $start, DateTimeInterface $end ): array {
+		$bq = $this->proxy->query( $query_name, $start, $end );
+		// Precedence intentional: a genuine proxy error must surface as the 'error' envelope, not be masked as 'empty'.
+		if ( is_wp_error( $bq ) ) {
+			return $this->error_collection( 'slices', $bq );
```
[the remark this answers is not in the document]
dkoo: Addressed in e819e3a5, both points. (1) Reordered so `compute_source_mix` checks `empty($records)` **before** the BQ call — a window with zero conversions is now `empty` (not `error`) and skips the proxy round-trip entirely, since BigQuery only attributes source to conversions that already exist in Woo. (2) A non-array success body or any non-array row now returns `malformed_collection('slices')` (state `error`, consistent with the other BQ-backed metrics) instead of silently degrading to all-direct; added `test_source_mix_subscribers_malformed_bq_body` to cover it. Good catch — chasing this also surfaced a stale REST-controller test (it asserted `time_to_subscribe` was still `coming_soon`), now fixed.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-metric.php
```diff
@@ -820,22 +820,54 @@ private function source_mix_rows(): array {
...
 		$proxy = $this->createMock( BigQuery_Proxy_Client::class );
 		$proxy->method( 'query' )
 			->with( 'conversion_journey_source_mix_subscribers' )
-			->willReturn( $rows );
-
-		// Each bucket gets count_completed_orders called once; return 1 for gate
-		// (2 rows), 1 for prompt (1 row), 1 for direct (1 row).
-		$resolver = $this->createMock( Woo_Order_Resolver::class );
-		$resolver->method( 'count_completed_orders' )->willReturn( 1 );
+			->willReturn( $bq_rows );
```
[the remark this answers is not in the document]
dkoo: Verified this does **not** fail — the full `Test_Conversion_Metric` suite is green (the actual CI failure was an unrelated stale assertion in `Test_Conversion_REST_Controller`, now fixed in e819e3a5). PHPUnit's `with()` only checks the constraints you pass and ignores actual arguments beyond them, so `->with('conversion_journey_source_mix_subscribers')` against `query(name, start, end)` passes — and it intentionally asserts the correct query name is dispatched. Leaving as-is.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-metric.php
```diff
@@ -889,51 +921,105 @@ public function test_source_mix_subscribers_returns_error_state_on_proxy_error()
...
+		];
+
+		$donors = $this->createMock( Donors_Metric::class );
+		$donors->method( 'get_new_donor_records_in_window' )->willReturn( $records );
+
 		$proxy = $this->createMock( BigQuery_Proxy_Client::class );
 		$proxy->method( 'query' )
 			->with( 'conversion_journey_source_mix_donors' )
-			->willReturn( $rows );
+			->willReturn( $bq_rows );
```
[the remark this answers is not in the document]
dkoo: Same as the subscribers test above — verified green, not a failure. The single-arg `with()` is an intentional assertion that the right query name is dispatched; PHPUnit ignores the extra `start`/`end` args beyond the provided constraint. Leaving as-is.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-metric.php
```diff
@@ -1863,6 +1951,113 @@ public function test_fixture_error_variant() {
...
+	 * @param array<string,mixed> $map query_name => rows|WP_Error.
+	 * @return BigQuery_Proxy_Client
+	 */
+	private function proxy_by_query( array $map ): BigQuery_Proxy_Client {
+		$proxy = $this->createMock( BigQuery_Proxy_Client::class );
+		$proxy->method( 'query' )->willReturnCallback(
+			static function ( $query_name ) use ( $map ) {
+				return $map[ $query_name ] ?? [];
+			}
+		);
```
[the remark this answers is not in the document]
dkoo: Verified non-issue: PHP does not warn when a closure declaring fewer parameters is called with extra positional args, so `willReturnCallback(fn($query_name) => …)` invoked as `query(name, start, end)` is fine and the suite is green. Left as-is to keep the helper minimal.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-conversion-metric.php
```diff
@@ -1863,6 +1951,113 @@ public function test_fixture_error_variant() {
...
+	public function test_source_mix_subscribers_proxy_error(): void {
+		$subs = $this->createMock( Subscribers_Metric::class );
+		$subs->method( 'get_new_subscriber_records_in_window' )->willReturn(
+			[
+				[
+					'customer_id' => 1,
+					'ts'          => 1700000000,
+				],
+			] 
+		);
```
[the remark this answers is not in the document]
dkoo: Fixed in e819e3a5 — stripped the trailing whitespace (this one was real, unlike the WP `( x )` paren-spacing the earlier reviewers flagged).

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
```diff
@@ -699,103 +700,237 @@ public function get_source_mix_registrations( DateTimeInterface $start, DateTime
...
-		$rows = $joined['rows'];
+	/**
+	 * Registration source events for the time-to-convert distributions, behind
+	 * the BQ probe gate. Window is a trailing 365 days ending now (snapshot).
+	 * Returns [ ['ts'=>int seconds, 'source'=>string], … ], or [] when the probe
+	 * reports no registrations or any BQ call fails (degrade to all-direct).
+	 *
+	 * @return array<int, array{ts:int, source:string}>
+	 */
+	private function registration_source_events(): array {
```
dkoo, line 991: **Suggestion:** registration_source_events() is uncached — `registration_source_events()` fires two BigQuery queries (the presence probe + the source lookup) and is called once each by the subscribe and donate distributions, so it runs twice per request — and up to four times in comparison mode — always returning the same trailing-365-day data. Memoizing the result on the instance (as `fetch_paid_attempts_woo_join` already does) would collapse these to a single fetch.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
```diff
@@ -699,103 +700,237 @@ public function get_source_mix_registrations( DateTimeInterface $start, DateTime
...
+		$utc   = new \DateTimeZone( 'UTC' );
+		$end   = new \DateTimeImmutable( 'now', $utc );
+		$start = $end->modify( '-365 days' );
```
dkoo, line 997: **Suggestion:** 4.2/4.3 source split bucketed to 'direct' for >365-day-old registrations — `registration_source_events()` loads registration-source events only for the trailing 365 days (line 789), but the lag rows feeding 4.2/4.3 are all-history (`get_subscription_conversion_lags` / `get_donation_conversion_lags`). Any converter who registered more than a year ago is still in the CDF but has no matching source event, so they're silently attributed `direct` — systematically inflating the `direct` series for older cohorts while the curve still includes them. Either bound the lag rows to the same 365-day window, or derive the source-event window from the registration timestamps actually being graphed. At minimum, a code comment near the 365-day window would help a future reader reconcile the all-history lag span with the 365-day source span.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
```diff
@@ -699,103 +700,237 @@ public function get_source_mix_registrations( DateTimeInterface $start, DateTime
...
+		// nothing to attribute, and a proxy error here would be a false 'error'.
+		if ( empty( $records ) ) {
+			return [
+				'state'  => 'empty',
+				'total'  => 0,
+				'slices' => [],
+			];
+		}
+
+		$bq = $this->proxy->query( $query_name, $start, $end );
```
dkoo: **Suggestion:** Source-mix misses exposure events just before the window start — `compute_source_mix()` fetches BigQuery exposure events for `[$start, $end]` (line 896), but `attach_sources()` matches each Woo conversion to an event up to `WINDOW_ORDER_SECONDS` (1800s) *before* it (line 926, before-only window). A conversion within 30 minutes of the window start therefore can't see its gate/prompt exposure event if that event fell just before `$start`, and is misclassified `direct`. Fetching events from `$start - WINDOW_ORDER_SECONDS` before matching would close the boundary gap. Magnitude is small (30 minutes of a multi-day window), but it systematically biases start-of-window conversions toward `direct`.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -1174,4 +1174,127 @@ public function get_first_subscription_order_dates( array $customer_ids ): array
...
+				FROM {$prefix}wc_orders o
+				JOIN {$prefix}wc_orders_meta om
+					ON om.order_id = o.id AND om.meta_key = '_schedule_start'
+				JOIN {$prefix}woocommerce_order_items oi
+					ON oi.order_id = o.id AND oi.order_item_type = 'line_item'
+				JOIN {$prefix}woocommerce_order_itemmeta oim
+					ON oim.order_item_id = oi.order_item_id AND oim.meta_key = '_product_id'
+				WHERE o.type = 'shop_subscription'
+				  AND oim.meta_value NOT IN ($donations)
+				  AND om.meta_value != ''
```
dkoo, line 1287: **Suggestion:** 3.2/3.3 source-mix total can diverge from the 3.1 headline count — The new records query `get_new_subscriber_records_in_window` adds `AND om.meta_value != ''` to the `_schedule_start` join (line 1206, and the legacy counterpart in `class-legacy-storage.php`), but the pre-existing count query `get_new_subscribers_in_window` (line 144) has no such guard. For a customer whose earliest `_schedule_start` is blank, the count query's `MIN()` picks the blank (excluded from the window) while the records query skips it and `MIN()`s the next non-empty start (which can fall inside the window). So the source-mix total (3.2/3.3) can disagree with the headline new-subscriber count (3.1) for the same window. The records behavior is arguably the more correct one — worth either adding the same guard to the count query or documenting the intended difference. (The donor side uses the never-blank `date_created_gmt` and has no such asymmetry.)

dkoo (commented):
Self-review (due diligence). No blockers, no security issues — the core mechanics (microseconds→seconds, UTC epochs, cumulative-curve boundaries, the ≥50 visibility gate, division-by-zero guards, all-`direct` graceful degradation, HPOS/legacy parity) are correct and covered by the PHP and JS unit tests. The comments below are accuracy and comparison-mode-cost refinements on what is an admin-only, read-only dashboard; none block merge.

The two most worth acting on before ship: the three all-history distribution methods should move from `build_window()` to `build_snapshot()` (they re-run for the `previous` window in comparison mode today), and the several independent paths that bias source attribution toward `direct` are worth at least a comment, since they compound on the same charts.

Note the live BigQuery source attribution still needs verification on a real Audience site — it can't run in the local env.

- `plugins/newspack-plugin/includes/wizards/insights/api/class-conversion-rest-controller.php:362` — **Suggestion:** Window-independent distributions re-run in comparison mode — `get_time_to_subscribe_distribution`, `get_time_to_donate_distribution`, and `get_subscriber_to_donor_lag_distribution` are wired into `build_window()` (lines 362–364), but their own docblocks declare them snapshot/all-history ("ignores the window"). `build_window()` runs once per window, so in comparison mode these three re-run for the `previous` window and return byte-identical all-history data — exactly the redundancy `build_snapshot()` (computed once at line 295, merged into both windows) exists to avoid. These are among the tab's most expensive Woo/BQ queries; moving the keys into `build_snapshot()` halves their cost when comparison is active.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
```diff
@@ -699,103 +700,237 @@ public function get_source_mix_registrations( DateTimeInterface $start, DateTime
...
 			];
+			$lag_by_customer[ $cid ] = $lag;
 		}
 
-		// Bucket rows by source. Each row's source is classified:
-		// gate   — gate_post_id is non-empty.
-		// prompt — gate_post_id is empty but popup_id is non-empty.
-		// direct — both are empty.
-		$buckets = [
+		$map         = Source_Matcher::attach_sources( $records, $events, Source_Matcher::WINDOW_REGISTRATION_SECONDS, Source_Matcher::WINDOW_REGISTRATION_SECONDS );
```
dkoo, line 1071: **Nit:** Near-simultaneous registrants can swap source attribution — For the time-to-convert split, `attach_sources()` uses a symmetric `WINDOW_REGISTRATION_SECONDS` window with greedy nearest-match, one event consumed per record (line 846). On a busy site where several readers register within that window of each other, the nearest-match tiebreak can attribute one reader's source to another. It's an accuracy nuance, not a defect — a one-line comment noting that cross-registrant misattribution within the window is accepted would keep a future maintainer from mistaking it for a bug.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
```diff
@@ -1174,4 +1174,127 @@ public function get_first_subscription_order_dates( array $customer_ids ): array
...
 	}
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @param DateTimeInterface $start Window start.
+	 * @param DateTimeInterface $end   Window end.
+	 * @return array<int, array{customer_id:int, ts:int}>
+	 */
+	public function get_new_subscriber_records_in_window( DateTimeInterface $start, DateTimeInterface $end ): array {
```
dkoo, line 1188: **Nit:** Guest conversions collapse to a single matcher key (customer_id 0) — Guest orders carry `customer_id = 0` (HPOS) / a blank `_customer_user` cast to `0` (legacy), so all guest conversions in a window share one key when fed to `Source_Matcher`. With the single-consume guard, at most one guest can be attributed a non-`direct` source per window; the rest fall to `direct`. Equivalent across both storage backends, so not a divergence — but if guest subscribers/donors are common, the source mix under-attributes them. An explicit `customer_id > 0` filter in the records query (or a comment) would make the intent clear.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
dkoo: Done in 8f367076 — memoized on a `$registration_source_events_cache` instance property; the two BQ queries now run once per request (the subscribe and donate distributions share the cached result, including in comparison mode).

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
dkoo: Done in 8f367076 — bounded the 4.2/4.3 cohort to the same trailing-365-day registration window `registration_source_events()` covers (filter `registered_ts >= now − 365d`), so converters who registered before the source window no longer get silently bucketed `direct`. Comment added. (Updated the 4.2/4.3 unit tests to `time()`-relative registration timestamps since the cohort is now wall-clock-relative.)

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
dkoo: Done in 8f367076 — `compute_source_mix()` now fetches the BQ exposure events from `$start − 1 day` (the events-only query; the Woo `$records` stay `[$start,$end]`). The proxy formats dates as Ymd/day-granular, so a full day is the right widening to let a conversion just after `$start` match an exposure that fell just before it. Comment added.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
dkoo: Done in 8f367076 — added the same `om.meta_value != ''` / `start.meta_value != ''` guard to `get_new_subscribers_in_window` (HPOS + legacy), so 3.1 reconciles with the 3.2/3.3 records. Heads-up: this method is shared with Tab 6, but the change only affects the degenerate blank-earliest-`_schedule_start` edge case.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-conversion-metric.php
dkoo: Done in 8f367076 — added a comment at the registration-anchor `attach_sources()` call noting that greedy single-consume nearest-match can swap sources between readers registering within ±120s on a busy site, an accepted accuracy nuance.

plugins/newspack-plugin/includes/wizards/insights/storage/class-hpos-storage.php
dkoo: Done in 8f367076 — added `AND o.customer_id > 0` (HPOS) / `AND CAST(cust.meta_value AS UNSIGNED) > 0` (legacy) to all four records queries (sub + donor), so guests (who collapse to a single matcher key) are excluded from source attribution. Comment added. Net effect on the total is ≤1 (guests already collapse to one `customer_id = 0` group in the count).

dkoo:
Re: the review-body suggestion — **window-independent distributions re-running in comparison mode** (`class-conversion-rest-controller.php`): fixed in 8f367076. Moved `time_to_subscribe_distribution`, `time_to_donate_distribution`, and `subscriber_to_donor_lag_distribution` out of `build_window()` and into `build_snapshot()` (computed once, merged into both `current` and `previous`), so in comparison mode they no longer re-run for the `previous` window — matching their own snapshot docblocks and halving their Woo/BQ cost when comparison is active. Full insights group green (357), host phpcs clean. The live BigQuery source verification on a real Audience site remains the one outstanding gate (can't run in the local env).

PR #376, opened 2026-06-20
fix(woocommerce): suppress payment notice on equivalent membership

jason10lee:
### All Submissions:

Follow-up to [NPPM-2926: Update Payment Notice still fires for subscribers holding both active and expired subscriptions](https://linear.app/a8c/issue/NPPM-2926/update-payment-notice-still-fires-for-subscribers-holding-both-active) hotfix [#301](https://github.com/Automattic/newspack-workspace/pull/301), which scoped the "update your payment method" notice to recoverable (`on-hold`/`pending`) statuses.

That hotfix addressed the issue, but we spotted an unreported gap that was worth following up on.  The existing suppression only treats two subscriptions as covering "the same product" when they are tiers of one grouped/variable product — it compares at the **product** level, not the **membership** level. A reader who holds equivalent active access bought through a *different* product (e.g. Newsroom Pro Monthly vs. a Special Offer product that grant the same membership plan) still saw the notice.

This PR suppresses the notice when the reader holds equivalent active access:

- **`Memberships::user_has_equivalent_active_access()`** (new, public) — true when the purchased product's membership plan is also granted by another **active** subscription (via the existing `get_user_subscription_for_membership_plan()`), or the reader is an **active** member of that plan (`wc_memberships_is_user_active_member()`).
- **`Memberships::get_plan_ids_for_product()`** (new, private) — resolves the membership plans a product grants.
- **`WooCommerce_Update_Payment_Notice`** — after the existing product-level (Layer 1) check, captures the variation-accurate purchased product and calls the new helper; suppresses on a match. ~9 lines, no API surface change.

Behavior is additive and fail-closed: if no equivalent access is found, the notice fires exactly as before.

#### Tests

- New WC Memberships mock layer (`tests/mocks/wc-memberships-mocks.php`) — WC Memberships isn't in the PHPUnit bootstrap.
- Unit coverage for membership-equivalence suppression and the variation/grouped/pending-cancel guards.

1. `n test-php --filter Newspack_Test_WooCommerce_Update_Payment_Notice` from `plugins/newspack-plugin` → 24 tests green.
2. Manual (WooCommerce + Subscriptions + Memberships): create a membership plan granted by two different products; give a reader an `on-hold`/`pending` subscription on product A and an **active** subscription on product B (same plan). Confirm the "update your payment method" notice does **not** appear. Cancel/expire the active subscription and confirm the notice returns.

### Other information:

Cross-repo impact review: no breaking changes; new `Memberships` method is additive, no consumer updates required.


dkoo (commented):
Solid, well-scoped follow-up to #301 that closes the membership-level equivalence gap NPPM-2926 describes. The logic fails closed, only runs for readers who already hold an `on-hold`/`pending` subscription that needs payment, and the self-match guards are correct: Layer 2 can't self-match because the active-status set (`active`/`pending-cancel`) is disjoint from the recoverable set (`on-hold`/`pending`), and Layer 3 skips the membership tied to the evaluated subscription via `exclude_subscription_id`. I traced the core scenario (active sub B + on-hold sub A, same plan) for both membership-linkage cases and both correctly suppress, while the genuinely-failing case still shows the notice. The 25 unit tests cover the real two-different-products-one-plan case plus the variation/grouped/pending-cancel guards, and the separate-process isolation for the WC Memberships mock is justified. No blockers.

Two non-blocking edges worth a look before merge:

- **Multi-plan products** — the helper suppresses as soon as one granted plan is covered, so a reader who still holds only part of a multi-plan product's access could stop being nagged for a failed payment that's needed to restore the rest.
- **Grouped-parent resolution** — `get_plan_ids_for_product()` runs the grouped-parent lookup on the variation-accurate product, whereas the Layer-1 code it mirrors first resolves to the parent; the two can diverge for a variation whose parent belongs to a grouped product.

Minor, optional: a defensive `(array)` cast on `wc_memberships_get_user_active_memberships()` and resolving the user's active subscriptions once (instead of once per plan) would tighten the helper.

Verification: unit suite green and the helper confirmed fail-closed when Memberships is inactive; a full live WooCommerce + Subscriptions + Memberships multi-product repro was not run, so runtime coverage is partial.

Release risk: LOW.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -863,6 +863,112 @@ public static function get_user_subscription_for_membership_plan( $user_id, $mem
...
+		// Build the set of product IDs that could be configured on a plan: the
+		// purchased product/variation, its parent variable product, and any
+		// grouped parents (mirrors the resolution in WooCommerce_Update_Payment_Notice).
+		$candidate_ids = [ (int) $product->get_id() ];
+		$parent_id     = (int) $product->get_parent_id();
+		if ( $parent_id > 0 ) {
+			$candidate_ids[] = $parent_id;
+		}
+		if ( class_exists( 'WC_Subscriptions_Product' ) && method_exists( 'WC_Subscriptions_Product', 'get_visible_grouped_parent_product_ids' ) ) {
+			foreach ( (array) \WC_Subscriptions_Product::get_visible_grouped_parent_product_ids( $product ) as $grouped_id ) {
```
dkoo: **Severity:** suggestion

`get_plan_ids_for_product()` runs `WC_Subscriptions_Product::get_visible_grouped_parent_product_ids()` on the variation-accurate `$product`, but the Layer-1 resolution this mirrors (`class-woocommerce-update-payment-notice.php:175-184`) reassigns `$product` to the **parent** variable product *before* the grouped lookup. In real WC Subscriptions, grouped children are top-level products, so passing a variation can return `[]` where passing the parent would return the grouped parent — a plan that grants the grouped parent, purchased via a variation, is then missed. `$candidate_ids` already contains the direct `parent_id`, so this only bites when the plan is configured against the grouped parent specifically. Narrow, and it fails in the safe direction (notice still shows), but to match Layer 1 exactly, run the grouped lookup against the resolved parent id too. No test covers the variation→grouped combination.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -863,6 +863,112 @@ public static function get_user_subscription_for_membership_plan( $user_id, $mem
...
+					continue;
+				}
+				// Only subscription-tied memberships expose get_subscription_id();
+				// a plain (e.g. manually granted) membership is always "other" access.
+				if ( $exclude_subscription_id
+					&& method_exists( $membership, 'get_subscription_id' )
+					&& (int) $membership->get_subscription_id() === (int) $exclude_subscription_id ) {
+					continue;
+				}
+				return true;
```
dkoo: **Severity:** suggestion

When a subscription product grants more than one membership plan, this early `return true` treats a single matching active plan as equivalent access to the whole product (the same any-plan logic repeats in the Layer 2 loop below). A reader who still holds only Plan A active would stop seeing the payment notice for a failed product that also grants Plan B, even though paying is still required to restore Plan B. This is the one edge that fails toward *under-notifying*. Consider tracking covered plan ids across both layers and suppressing only once all of `$plan_ids` are covered.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -863,6 +863,112 @@ public static function get_user_subscription_for_membership_plan( $user_id, $mem
...
+		}
+
+		// Layer 3 — currently-active membership access. Enumerate the user's
+		// active memberships so we can skip one tied to the subscription under
+		// evaluation: a membership kept active by the same on-hold/pending
+		// subscription is not "other" access, and counting it would falsely
+		// suppress the very notice that subscription needs. This mirrors
+		// Layer 2's active-status self-match proofing. NPPM-2926.
+		if ( function_exists( 'wc_memberships_get_user_active_memberships' ) ) {
+			foreach ( wc_memberships_get_user_active_memberships( $user_id ) as $membership ) {
```
dkoo, line 911: **Severity:** nit

`wc_memberships_get_user_active_memberships( $user_id )` is iterated directly. The real API returns an array (and the sibling call at line 826 assigns to a variable first), so this is safe in practice; a defensive `(array)` cast would match the cast you already apply to the grouped-parent result on line 954. Very low priority.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
```diff
@@ -863,6 +863,112 @@ public static function get_user_subscription_for_membership_plan( $user_id, $mem
...
+		}
+
+		// Layer 2 — another active subscription that grants the same plan.
+		// get_user_subscription_for_membership_plan() resolves through
+		// WooCommerce_Connection::get_active_subscriptions_for_user(), which
+		// filters to ACTIVE_SUBSCRIPTION_STATUSES — so the on-hold/pending
+		// subscription under evaluation can never self-match here.
+		foreach ( $plan_ids as $plan_id ) {
+			if ( self::get_user_subscription_for_membership_plan( $user_id, $plan_id ) ) {
+				return true;
```
dkoo: **Severity:** nit

The Layer 2 loop calls `get_user_subscription_for_membership_plan()` per plan, and each call re-runs `wcs_get_users_subscriptions()` via `WooCommerce_Connection::get_active_subscriptions_for_user()` — an N+1 over `$plan_ids`. It only executes for a logged-in reader who already holds an on-hold/pending subscription needing payment, so real-world impact is negligible; resolving the user's active subscriptions once would avoid it if plan counts ever grow.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
jason10lee: This reproduces. `get_visible_grouped_parent_product_ids()` goes through `get_parent_ids()`, which matches a grouped product's `_children` postmeta — and that holds top-level product IDs, never variation IDs, so a variation finds no grouped parent at all. I wrote the test before the fix: a variation whose parent is a grouped child returned `[]` instead of the plan. The grouped lookup now runs against the resolved parent as well, matching Layer 1. Fixed in fc56ae7c1.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
jason10lee: This one was worth changing — under-notifying about money is the wrong direction to fail. Suppression now needs every plan the product grants to be covered, with both layers contributing coverage instead of either short-circuiting to `true`. A reader holding Plan A but not Plan B keeps the notice. Fixed in 0ff075a61, with tests for partial coverage and full coverage.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
jason10lee: Went to the source on this one rather than reason about it. `wc_memberships_get_user_active_memberships()` delegates to `get_user_memberships()`, which has a single return path, initializes `$user_memberships = array()` up front, and only ever reassigns it to an array. It can't return a non-array, so a cast would be guarding a state the API cannot reach. The `(array)` on the grouped-parent line stays because that one wraps a `$wpdb->get_col()` result.

plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php
jason10lee: Keeping the shape. Hoisting the lookup means reworking `get_user_subscription_for_membership_plan()` and `WooCommerce_Connection::get_active_subscriptions_for_user()`, both shared, for a path that only runs when a logged-in reader already has an on-hold subscription, across plan counts in the low single digits. The fix for the multi-plan thread above does tighten it: Layer 2 now skips plans Layer 3 already covered and stops at the first uncovered one, so the common single-plan case is still a single lookup.

dkoo (approved):
Re-review of the delta d9803e61..0ff075a6 (two follow-up commits addressing my prior two suggestions). Both are fully and correctly addressed; delta is clean.

1. Grouped-parent resolution for variations (fc56ae7c) — get_plan_ids_for_product() now runs get_visible_grouped_parent_product_ids() against the resolved parent product in addition to the purchased product. Since WC Subscriptions' grouped _children meta only holds top-level product IDs (never variation IDs), a variation purchase previously found no grouped parent; resolving the parent first now mirrors the Layer-1 lookup. wc_get_product() is null-guarded and results are deduped via array_unique( array_filter() ). Fully addressed.

2. Every granted plan must be covered (0ff075a6) — user_has_equivalent_active_access() no longer returns true on the first covered plan. Layer 3 now accumulates covered plan IDs; the final loop over all plan IDs returns false on the first plan neither layer covers, and returns true only if every plan is covered. Verified: the Layer-2 condition is not inverted (no active subscription for a plan means uncovered means return false), the empty plan set is short-circuited to false before the loop (fail-closed), the exclude_subscription_id self-match guard is preserved, and the single-plan common case still costs at most one Layer-2 lookup. Fully addressed.

Tests (+77 lines) exercise exactly the delta: test_plan_lookup_matches_grouped_parent_of_variation registers the grouped parent against the top-level product only (8101 to 8100) and asserts the variation (8102) still resolves to the plan, which fails without the parent lookup; test_no_equivalent_access_when_only_some_plans_are_covered asserts a partially-covered two-plan product still shows the notice; test_equivalent_access_when_all_plans_are_covered asserts suppression when the two plans are covered by a membership and a separate active subscription respectively. The mocks (get_visible_grouped_parent_product_ids keyed by product id, wcs_create_subscription, active-membership globals) support the assertions. Both files pass php -l.

Approving.

PR #377, opened 2026-06-21
feat(insights): direct donation conversion metrics from order meta (NPPD-1745)

kmwilkerson:
## What

Migrates the **direct donation** conversion metrics (Tab 5 / Prompts) off the GA4-attempt → `Woo_Order_Resolver` join and onto **WooCommerce order meta** (`_newspack_popup_id`), which is anonymous-inclusive and carries a real `customer_id`. This is the NPPD-1685 spike's recommended direct-metric fix.

The GA4 attempt event has no usable `customer_id` (GA4 `user_id` is unpopulated; the only id is the cookie `user_pseudo_id`), so the cookie→`customer_id` cast can't match real orders — dropping anonymous-at-attempt donors. The originating prompt id is already written onto the cleared order, so the direct metrics read it directly with no hub join.

Investigation + measurements (internal): `Automattic/newspack-insights` → `investigations/nppd-1685.md`. Ticket: **NPPD-1745**.

## Changes (commits, dependency order)

- **Storage reader** `get_prompt_attributed_donation_conversions` (legacy + HPOS): initial completed donation orders carrying the popup id, renewals excluded, grouped by popup id. Dedupe is **EXISTS + correlated `MIN`**, not a JOIN — the popup-id meta is written more than once per order, so a JOIN double-counts both count and revenue.
- **`Donors_Metric`** cached delegator for the reader.
- **`get_donation_revenue_direct`** sums the order-meta map (pure-local; survives a hub outage). **`get_donation_conversion_direct`** + **per-prompt rate** use an order-meta numerator ÷ hub donation-impressions denominator, via a shared `donation_rate_value()` with a coherence guard (no fabricated >100% across the two surfaces) and unified —/0% em-dash semantics.
- **Tab-error banner** scoped to hub-backed metrics (`METRIC_SOURCES` classification + `is_window_all_error`) so a surviving local card doesn't suppress the "whole tab failed" banner.
- **Tests:** a DB-backed integration harness (reusable fixtures trait stands up the WC order tables; legacy + HPOS), the duplicate-meta 2× regression anchor, coherence-guard + em-dash coverage, and the scoped-banner regression.

## Tests

`n test-php --group insights` → **365/365** green.

## Scope / not in this PR

- **Influenced-14d** metrics still use `Woo_Order_Resolver` (genuine cross-session linkage) — separate effort; the resolver is now influenced-only.
- **Source-mix (Tab 3)** and the **Tab-3 banner scoping** are on the #373 track (coordination), not here.
- A lower-urgency write-side cleanup (make the popup-id order meta `unique`) to be filed separately; the reader is robust to historical duplicates regardless.

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/api/class-prompts-rest-controller.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/api/class-prompts-rest-controller.php
@@ -256,11 +256,12 @@ class Prompts_REST_Controller extends WP_REST_Controller {
 	/**
 	 * Assemble the top-level response.
 	 *
-	 * `tab_error` is true only when every metric in the current window reports
-	 * `state: 'error'` — i.e. the whole tab failed to load (e.g. the BigQuery
-	 * proxy is down/misconfigured). React renders a tab-level error banner in
-	 * that case; otherwise each section renders its own error/empty/populated
-	 * treatment.
+	 * `tab_error` is true only when every **hub-backed** metric in the current
+	 * window reports `state: 'error'` — i.e. the hub (BigQuery proxy) is
+	 * down/misconfigured. Local (Woo order-meta) cards survive a hub outage and do
+	 * not suppress the banner (NPPD-1745; see {@see self::is_window_all_error()}).
+	 * React renders a tab-level error banner in that case; otherwise each section
+	 * renders its own error/empty/populated treatment.
 	 *
 	 * @param Prompts_Metric         $metric        Orchestrator.
 	 * @param DateTimeImmutable      $start         Current window start.
@@ -289,26 +290,39 @@ class Prompts_REST_Controller extends WP_REST_Controller {
 	}
 
 	/**
-	 * Whether every metric in a window payload reports `state: 'error'`.
+	 * Whether every **hub-backed** metric in a window reports `state: 'error'`.
 	 *
-	 * Returns `false` as soon as any metric is not in the error state (the `window`
-	 * key is date metadata, not a metric, so it's skipped). A metric missing a
-	 * `state` key is treated as non-error, so the banner only shows on an
-	 * unambiguous all-failed window.
+	 * Scoped to hub-backed metrics (NPPD-1745): as cards move to local Woo order-meta
+	 * sourcing, a local card always succeeds even when the hub is fully down. Under
+	 * the old "every metric errored" rule the "whole tab failed" banner could then
+	 * never fire — a half-broken tab that looks trustworthy. So the banner now fires
+	 * when all hub-backed (and hybrid) metrics error, even though a surviving local
+	 * card still renders. The hub-backed-vs-local classification is declared
+	 * explicitly on {@see \Newspack\Insights\Prompts_Metric::METRIC_SOURCES}, not
+	 * inferred and not hardcoded here.
+	 *
+	 * Returns `false` as soon as any hub-backed metric is not in the error state.
+	 * Returns `false` for a window with no recognizable hub-backed metric at all
+	 * (nothing to declare failed).
 	 *
 	 * @param array $window The shape returned by `build_window()`.
 	 * @return bool
 	 */
 	private static function is_window_all_error( array $window ): bool {
-		foreach ( $window as $key => $value ) {
-			if ( 'window' === $key ) {
+		$saw_hub_backed = false;
+		foreach ( Prompts_Metric::METRIC_SOURCES as $key => $source ) {
+			if ( 'local' === $source ) {
 				continue;
 			}
-			if ( ! is_array( $value ) || ! isset( $value['state'] ) || 'error' !== $value['state'] ) {
+			if ( ! isset( $window[ $key ] ) || ! is_array( $window[ $key ] ) || ! isset( $window[ $key ]['state'] ) ) {
+				continue;
+			}
+			$saw_hub_backed = true;
+			if ( 'error' !== $window[ $key ]['state'] ) {
 				return false;
 			}
 		}
-		return true;
+		return $saw_hub_backed;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


kmwilkerson:
## Code-review (8-angle, high effort)

Ran a multi-angle review (line-by-line, removed-behavior, cross-file, SQL deep-dive, rate/banner semantics, tests, cleanup, conventions). Summary: solid, well-tested; no crashers on the common path. The real issues are the new tab-error banner having a hole for one publisher class, and the order-meta numerator vs GA4-impressions denominator being two surfaces that can disagree.

### Fixed in 19841cc840
- **Subscription augmentation now WC-gated.** `get_performance_by_prompt` gated the new donation read on `woocommerce_active()` but still ran the subscription resolver path (`wc_get_orders()`) unconditionally → potential fatal on a non-WC publisher. Now gated symmetrically (+ a regression assertion that subscription columns also degrade on non-WC).
- **Memoized `prompts_performance_by_prompt`.** The rate denominator and the per-prompt table each queried it independently for the same window (2 identical hub round-trips, 4 with comparison). Now one round-trip per window.
- **`tear_down()`** (was camelCase `tearDown()`, inconsistent with the suite's snake_case convention; the WC-active seam filter is removed there).
- **Sharpened the revenue-card comment** so the intentional (local/hub-resilient) conversions-based em-dash isn't "fixed" to match the rate card.

### Open — design calls for reviewer/Copilot
1. **Non-WC + hub-down suppresses the "whole tab failed" banner.** `donation_conversion_direct` is the only `hybrid` card but short-circuits to `populated` on non-WC *before* touching the hub, so on a non-WC publisher during a full hub outage `is_window_all_error` sees a non-error hub-backed card and returns false — the exact gap `METRIC_SOURCES` was added to close, for non-WC sites. The controller test currently asserts this as expected.
2. **Coherence guard can null the entire headline donation rate.** Numerator (order-meta, server-side, anonymous-inclusive) vs denominator (GA4 donation-intent impressions, consent/adblock-undercounted) are different populations; when aggregate conversions exceed aggregate impressions, the whole tab rate renders an em-dash despite real donations. Edge, but high-consequence.
3. **`fetch_donation_prompt_impressions` swallows a malformed/relabeled hub response** as `0` impressions (→ em-dash), whereas the sibling table returns `malformed_collection`. A hub fault or intent rename is hidden instead of surfaced (and the hybrid card then doesn't count toward the banner).
4. **HPOS renewal exclusion is unverified against live HPOS data** (`NOT EXISTS` on `wc_orders_meta`); on a mid-migration site the renewal marker may still be in `postmeta`. Self-flagged in code; flagging for the first HPOS-publisher check.
5. **`METRIC_SOURCES` is a hand-maintained parallel list to `build_window`** (matches today; no drift guard — a future `build_window` key missing from the map is silently excluded from the banner and the test can't catch it). Candidate to move onto the metric envelope.
6. **Tab-level vs per-prompt rate can visibly diverge** (shared formula, different inputs: aggregate vs per-row guard) — the docblock's "cannot diverge" is overstated.
7. **Per-prompt `subscription_conversion_rate` has no coherence guard** (can render >100%), unlike the donation column — subscriptions weren't migrated, so they didn't get the guard.

Items 1, 2 are the ones I'd want a decision on before this leaves draft.

kmwilkerson:
## Review responses

Resolutions below; all committed with the `insights` group green (368 tests). 1/3/5 + the 6 docblock land in the fix commit on this branch.

### Fixed (before this leaves draft)

**1 — non-WC + hub-down banner hole.** Agreed; this is exactly the gap `METRIC_SOURCES` exists to close, reopened for non-WC sites. `is_window_all_error` now skips `hybrid` cards when WooCommerce is inactive — they short-circuit to a not-applicable empty state *before* touching the hub, so a hub outage can't make them error and their surviving empty state must not count as a hub-backed survivor. The genuinely hub-backed (`hub`) cards still drive the decision. Threaded `Prompts_Metric::woocommerce_active()` (now public) into the check. Flipped the controller test that asserted the bug, and added `test_tab_error_fires_on_non_wc_hub_outage` pinning both directions (non-WC fires, WC-active still suppresses on a real survivor).

**3 — malformed hub response swallowed as 0.** Agreed. `fetch_donation_prompt_impressions` now returns `bigquery_proxy_malformed_rows` on a non-array response (mirroring the sibling table's `malformed_collection`), so the hybrid card errors and counts toward the banner instead of rendering an em-dash off a fabricated 0 denominator. Added a test.

**5 — `METRIC_SOURCES` drift.** Best structural catch. Added a drift-guard test: every state-bearing `build_window` key must have a `METRIC_SOURCES` entry, so a future card added to the window but not the map can't silently drop out of the banner. Moving the source onto the metric envelope is the cleaner long-term fix — filing that as a follow-up; the test closes the silent-failure path now.

### Documented as known limitations (no behavior change)

**2 — coherence guard can null the headline rate.** Keeping the guard. Order-meta numerator (complete, server-side) over GA4 impressions (consent/adblock-undercounted) are different populations, so the denominator can legitimately be smaller — an em-dash is the honest render, and a >100% rate would be worse. It's the same denominator-fidelity issue as the gate `checkout_impressions` case. Documented as a known limitation on `rate_value()`; follow-up filed for a more complete impressions source.

**4 — HPOS renewal exclusion unverified on live HPOS.** Self-flagged in code already; same HPOS caveat as the rest of this work — keeping as a documented known caveat for the first HPOS-publisher check.

### Smaller

**6 — "cannot diverge" overstated.** Fixed the docblock: the tab-level card and per-prompt rows share one formula + coherence guard, but their *values* legitimately differ (aggregate vs per-row inputs); what they can't do is drift to a different *rule* (e.g. one rendering a >100% rate the other suppresses).

**7 — per-prompt `subscription_conversion_rate` has no coherence guard.** Correct, and by design here — subscriptions aren't migrated in this PR. They get the guard in the subscription-direct migration (NPPD-1746), where the order-meta rate and coherence guard land together. That work also proactively mirrors 1 (non-WC banner hole, on the Gates tab) and 3 (malformed → error) to the gate/popup paths so the same gaps don't reopen there.

kmwilkerson:
**Correction to my earlier comment:** these review fixes (#1/#3/#5/#6/#2) were committed locally but missed this PR's merge window, so they did **not** land here. They're split out into a standalone follow-up PR against `feat/insights-rsm`: #383. (#7 lands with the subscription migration, NPPD-1746.)

PR #382, opened 2026-06-22
fix(popups): preview ignores autosave older than saved prompt (NPPM-2940)

jason10lee:
### All Submissions:

Prompt previews could show blocks that had already been removed and saved away. The preview renderer always preferred the editor's autosave over the saved prompt, but Gutenberg's autosave is a no-op once the post is clean — so after you delete a block and do a full save, the older autosave (still containing the block) lingered and the preview kept rendering it.

This makes the preview use the autosave only when it is actually newer than the saved prompt: a freshly-saved prompt previews its current content, while in-progress (unsaved) edits still preview correctly.

**Scope:** this addresses the *preview-shows-removed-content* symptom of NPPM-2940 (Problem 2). The "overlay layout/alignment lost on the front end" symptoms (Problems 1 & 3) have a different root cause and are handled by #216.

Addresses [NPPM-2940: CLTampa prompt Edit view doesn't match what displays on the frontend](https://linear.app/a8c/issue/NPPM-2940/cltampa-prompt-edit-view-doesnt-match-what-displays-on-the-frontend).

1. Create a prompt (any placement). Add a Heading block, then a Paragraph. Full **Save (Update)**.
2. Click **Preview** — confirm both blocks show.
3. Delete the Heading block. Full **Save (Update)** again.
4. Click **Preview** — before this PR the removed Heading still appears; with this PR only the Paragraph shows.
5. **Regression check:** make an edit but do *not* save; click **Preview** — the unsaved edit still appears.
6. `cd plugins/newspack-popups && n test-php --filter test_preview_` → both new `ModelTest` cases pass; full popups suite green (213 tests).

### Other information:

* [x] Have you successfully run tests with your changes locally? (213/213 popups suite green)


plugins/newspack-popups/includes/class-newspack-popups-model.php
```diff
@@ -290,9 +290,17 @@ public static function get_preview_query_options() {
...
+		// The editor's autosave only holds the freshest content while there are
+		// unsaved changes. After a full save it can be OLDER than the saved post
+		// (Gutenberg's autosave() no-ops on a clean post), so prefer it only when
+		// it is actually newer than the saved post — otherwise a block removed
+		// before the last save would linger in the preview (NPPM-2940).
+		$saved       = get_post( $post_id );
 		$autosave    = wp_get_post_autosave( $post_id );
-		$post_object = $autosave ? $autosave : get_post( $post_id );
+		$post_object = $saved;
+		if ( $autosave && ( ! $saved || strtotime( $autosave->post_modified_gmt ) > strtotime( $saved->post_modified_gmt ) ) ) {
```
dkoo, line 335: **Suggestion:** The tie-break deliberately prefers the saved post when `post_modified_gmt` is equal, which is exactly what guarantees a removed-then-saved block can never resurface — the point of the fix. The trade-off: an unsaved edit made within the same one-second timestamp bucket as the preceding full save briefly loses to the saved post and is absent from the preview until the next autosave bumps the timestamp. The window is sub-second and self-heals, so this is fine as-is; flagging it so the behavior reads as a conscious choice rather than a surprise.

dkoo, line 335: **Suggestion:** Optional: comparing `post_modified_gmt` via `strtotime()` works because both operands get the same (server-timezone) conversion, so relative ordering is preserved. A direct string comparison (`$autosave->post_modified_gmt > $saved->post_modified_gmt`) would be equivalent — MySQL `Y-m-d H:i:s` sorts lexicographically the same as chronologically — and would sidestep timezone/DST and zero-date (`strtotime( '0000-00-00 00:00:00' ) === false`) edge cases. Not required.

plugins/newspack-popups/tests/test-model.php
```diff
@@ -255,6 +255,158 @@ public function test_retrieve_preview_popup_with_invalid_id() {
...
+		);
+	}
+
+	/**
+	 * NPPM-2940: after a full save, the editor's stale autosave can be OLDER
+	 * than the saved post (Gutenberg's autosave() no-ops on a clean post). The
+	 * preview must render the freshly-saved post, not the older autosave whose
+	 * blocks were since removed.
+	 */
+	public function test_preview_prefers_saved_post_when_newer_than_autosave() {
```
dkoo, line 292: **Nit:** Optional: the four new tests cover autosave-older, autosave-newer, tie, and missing-saved-post, but not the most common runtime path — a saved post present with no autosave at all (`wp_get_post_autosave()` returns `false`). It's the trivial default branch, so low value, but a one-line case would close the matrix.

dkoo (approved):
Passes manual testing. Release risk: LOW. Correct, well-targeted fix for NPPM-2940 Problem 2: the preview now prefers the saved prompt and only uses the editor autosave when it is genuinely newer (by `post_modified_gmt`), so a block that was removed and saved no longer lingers in the preview, while in-progress unsaved edits still preview.

Verified end-to-end against a running site: a real stale autosave (resolved by `wp_get_post_autosave()` and still containing the removed block) is correctly ignored in favor of the saved post, and a newer unsaved autosave still shows. The full popups PHPUnit suite is green (215/215), the four new preview tests genuinely traverse the real `wp_get_post_autosave` path, PHPCS is clean, and there are no debug-log warnings.

No blockers — only a few optional notes below.

plugins/newspack-popups/includes/class-newspack-popups-model.php
jason10lee: That's the intended trade-off, and the comment above the check is already describing it — prefer the autosave only when it's actually newer, so a block removed before the last save can't reappear. The sub-second window clears on the next autosave.

jason10lee: I worked through the edge cases and the two forms agree on all of them. On a zero date `strtotime()` returns `false`, and comparing `false` against an int casts to bool, so the zero-dated side loses — the same result a lexicographic compare gives. Both operands are GMT in one format, so timezone and DST can't separate them either. Keeping `strtotime()`; it says "compare times" more plainly than a string compare.

plugins/newspack-popups/tests/test-model.php
jason10lee: Agreed the matrix should cover it — no-autosave is the most common path. Added in 64ceb2904.

Worth flagging what turned up while doing it: this branch needed a merge from `main` first, and `main` has since added a permission gate to `retrieve_preview_popup()`. The gate makes the missing-saved-post case unreachable — `get_post_type()` returns `false` for a post that doesn't resolve, so it denies before the autosave fallback runs. The test that asserted the old fallback went red on the merge and now asserts the gate's behavior instead. The gate is the stronger claim and should win, but it does narrow what this PR does, so it's worth a look.

PR #387, opened 2026-06-23
hotfix(newsletters): persist send-config before send (NPPM-2935, NPPM-2929)

jason10lee:
### All Submissions:

* [x] Does your code follow the WordPress and VIP Go coding standards? (PHPCS clean on changed files)
* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

**Root cause.** The ESP send is triggered from `pre_post_update`, which fires *inside* `wp_update_post()` — **before** WordPress's REST posts controller writes the request's post meta (`meta->update_value()` runs after `wp_update_post()`). So when a publisher changes send-config and publishes in one request, the send reads the **previously-stored** meta:

- **NPPM-2935** — the newsletter goes to the *old* list/segment. Worse, for ActiveCampaign `create_campaign()` computes `segmentid = $has_configured_segment ? $send_sublist_id : 0`, and `segmentid = 0` means **the entire audience** — so a freshly-selected segment that hasn't been persisted yet sends to everyone (irreversible).
- **NPPM-2929** — a sender set in the same request isn't persisted yet, so the send aborts with "Please input sender name and email address."

**Fix.** Persist the request's send-config on `rest_pre_insert_{cpt}`, which fires in `prepare_item_for_database` *before* `wp_update_post()`, so the send reads current values. A single-source `SEND_CONFIG_META_KEYS` allowlist (`send_list_id`, `send_sublist_id`, `senderName`, `senderEmail`) drives the early write. It is partial-update safe (only writes keys present in the request), write-only-if-changed, and uses `wp_slash()` to mirror the normal meta write so the value the send reads is byte-identical to what is finally stored. It runs only after the route's `edit_post` permission check, and these keys carry no custom sanitize/auth callbacks, so no REST guarantee is bypassed. The provider send/sync code and the segment guard are unchanged. Also removes the now-orphaned `api_set_post_meta()` dead code, whose purpose this handler serves correctly.

**Relation to NPPM-2879.** That fix addressed a *Mailchimp-specific* wrong-audience symptom (resetting the campaign's segment snapshot in the Mailchimp PATCH path) — a different root cause (an ESP API snapshot quirk). This change is provider-agnostic: it fixes the underlying REST meta-write ordering that affects every ESP. The two are complementary and do not overlap.

Closes: **NPPM-2935** and **NPPM-2929** (Linear; auto-linked via branch name).

1. On an ActiveCampaign (or Mailchimp) newsletter, change the Send-To list/segment on an existing draft and Send/Publish in the same action without a separate save — it now goes to the **newly-selected** list/segment, and a freshly-selected segment is never sent as "entire audience."
2. With sender name/email populated via a saved layout, send a test/publish — no spurious "Please input sender name and email address."
3. Run the new regression suite: `Newsletter_Send_Config_Persistence_Test` (8 tests — ordering, whole-audience segment gate, sender freshness, partial-update safety, create no-op, unauthorized-caller rejection, allowlist lock, dead-code removal). Verified against the `release` plugin: `OK (8 tests)`, and the full plugin suite is green (`OK (221 tests)`).

### Other information:

### DIFF
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
@@ -19,6 +19,14 @@ final class Newspack_Newsletters {
 	const API_NAMESPACE                     = 'newspack-newsletters/v1';
 
 	/**
+	 * Send-config meta keys the ESP send path reads. Single source of truth for
+	 * the fields that must be persisted before the send fires (NPPM-2935/2929).
+	 * Keep in lockstep with the send-config get_post_meta reads in the provider
+	 * send/sync path (e.g. Active_Campaign::create_campaign / sync).
+	 */
+	const SEND_CONFIG_META_KEYS = [ 'send_list_id', 'send_sublist_id', 'senderName', 'senderEmail' ];
+
+	/**
 	 * Supported fonts.
 	 *
 	 * @var array
@@ -82,6 +90,7 @@ final class Newspack_Newsletters {
 		add_filter( 'post_row_actions', [ __CLASS__, 'display_view_or_preview_link_in_admin' ] );
 		add_filter( 'jetpack_relatedposts_filter_options', [ __CLASS__, 'disable_jetpack_related_posts' ] );
 		add_action( 'save_post_' . self::NEWSPACK_NEWSLETTERS_CPT, [ __CLASS__, 'save' ], 10, 3 );
+		add_filter( 'rest_pre_insert_' . self::NEWSPACK_NEWSLETTERS_CPT, [ __CLASS__, 'persist_send_config_before_send' ], 10, 2 );
 		add_action( 'admin_enqueue_scripts', [ __CLASS__, 'branding_scripts' ] );
 		add_filter( 'newspack_theme_featured_image_post_types', [ __CLASS__, 'support_featured_image_options' ] );
 		add_filter( 'gform_force_hooks_js_output', [ __CLASS__, 'suppress_gravityforms_js_on_newsletters' ] );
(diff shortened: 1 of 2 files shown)


plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -550,6 +559,60 @@ public static function save( $post_id, $post, $update ) {
...
+			// reads before REST schema validation rejects the request. null is allowed
+			// (it clears the field, matching the normal meta path's effect on read).
+			if ( null !== $value && ! is_scalar( $value ) ) {
+				continue;
+			}
+			// Skip a redundant write when unchanged (defense-in-depth; keeps unchanged saves byte-identical).
+			if ( (string) $value === (string) get_post_meta( $prepared_post->ID, $key, true ) ) {
+				continue;
+			}
+			update_post_meta( $prepared_post->ID, $key, wp_slash( $value ) );
```
dkoo, line 620: **Suggestion:** The early write persists the raw `$request['meta']` value via `update_post_meta( …, wp_slash( $value ) )`, while the normal REST meta write applies `sanitize_meta()` before the same slashed write. Today the persisted value is byte-identical because none of the four `SEND_CONFIG_META_KEYS` registers a `sanitize_callback` — but that equality is now load-bearing and invisible at the registration site. If a `sanitize_callback` is ever added to one of these keys (e.g. `sanitize_email` on `senderEmail`), the send would read the *unsanitized* early-write value while the finally-stored value is sanitized, re-diverging the two. Consider a one-line comment at the meta registration pointing back to `SEND_CONFIG_META_KEYS`, or running the value through `sanitize_meta()` here so the guarantee is self-enforcing.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -550,6 +559,60 @@ public static function save( $post_id, $post, $update ) {
...
+	 * skipped (the REST schema rejects them moments later anyway). The early write
+	 * is committed during prepare and is intentionally not rolled back if the
+	 * enclosing post update later fails — acceptable because the send only fires on
+	 * a successful status transition within that same update.
+	 *
+	 * @param stdClass        $prepared_post Post object about to be inserted/updated.
+	 * @param WP_REST_Request $request       The REST request.
+	 * @return stdClass The unchanged prepared post.
+	 */
+	public static function persist_send_config_before_send( $prepared_post, $request ) {
```
dkoo, line 595: **Nit:** `rest_pre_insert_{cpt}` is meant to mutate the prepared-post object, not to persist data; writing meta to the DB inside it is a side-effect in a filter. It's a defensible trade-off for this hotfix — there's no other hook that runs after the route's permission check but before the `pre_post_update` send — and the docblock is honest about the intentional non-rollback. Flagging it only as a candidate for a cleaner follow-up: deferring the send to after `rest_insert_{cpt}` / `save_post` rather than racing meta persistence against it.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
dkoo, line 620: **Nit:** The normal REST meta write calls `current_user_can( 'edit_post_meta', $post_id, $key )` per key before writing; this early write doesn't. Harmless today because all four keys register `auth_callback => '__return_true'` (a no-op) and the route's `edit_post` gate already applies — but, like the sanitize coupling above, it silently depends on those `auth_callback`s staying permissive. No action needed for the hotfix.

dkoo (approved):
Passes manual testing. This is a correct, tightly-scoped fix for a genuinely irreversible bug. The `rest_pre_insert_{cpt}` placement commits the request's send-config before every `pre_post_update`-driven send, the `SEND_CONFIG_META_KEYS` allowlist exactly matches the keys the send path reads across the active ESPs, and the scheduled (cron) send path is unaffected because its meta is already committed by the time it fires. The dead-code removal is clean — `api_set_post_meta()` had no remaining route or client reference. No blockers; two forward-looking notes below worth a glance before merge.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
jason10lee: Thanks — agreed the parity is load-bearing and shouldn't be invisible. One clarification from testing: `update_post_meta()` runs `sanitize_meta()` internally, so a registered `sanitize_callback` *is* applied by the early write too — I confirmed a key registered with `sanitize_email` stores identically whether written via the early path or the normal route, so `sanitize_callback` specifically wouldn't re-diverge. The real gaps are the ones you flag below: the early write skips the REST schema-validation layer and the per-key `edit_post_meta` check. I've added a comment at the `send_list_id` registration (`19ae6e91a`) noting these keys must stay permissive strings and that any restrictive schema/auth must be mirrored in `persist_send_config_before_send()`. Went with the doc-comment over routing through `sanitize_meta()` here, since `update_post_meta` already applies it and that wouldn't cover the schema/cap gap anyway.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
jason10lee: Agreed this is the cleaner long-term shape. Persisting in a filter is a deliberate trade-off here because it's the only hook that runs after the route's permission check but before the `pre_post_update` send. I'll track the deferred-send refactor (move persistence to `rest_insert`/`save_post` so it isn't racing the send) as a separate follow-up rather than widen the hotfix's scope. Thanks for the thorough review!

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
jason10lee: Folded into the same registration-site comment (`19ae6e91a`): it notes the early write relies on `auth_callback => __return_true` plus the route's `edit_post` gate, and to mirror any future restrictive `auth_callback` in `persist_send_config_before_send()`. Agreed — no behavioral change needed for the hotfix.

PR #390, opened 2026-06-23
hotfix(wc-memberships): skip expiry filter for non-subscription memberships

adekbadek:

The `wc_memberships_expire_user_membership` filter fires for **every** user membership, not only subscription-tied ones. `Membership_Expiry::prevent_membership_expiration()` assumed a `\WC_Memberships_Integration_Subscriptions_User_Membership` and called `get_subscription_id()` on the passed object. For a plain `\WC_Memberships_User_Membership` (a membership not linked to a subscription) that method does not exist, so the call threw:

```
PHP Fatal error: Uncaught Error: Call to undefined method WC_Memberships_User_Membership::get_subscription_id() in .../wc-memberships/class-membership-expiry.php:39
```

The handler now bails out early via `method_exists()` when the membership is not subscription-tied, and the docblock type is widened to the base class.

This is a hotfix onto `release`: the fix already landed on `main`/`alpha` (`9b37834a`, originally Automattic/newspack-plugin#4743) but is missing from the released 6.43.x line, which is what production sites hitting this fatal are running – this should have been a hotfix from the get-go. 

1. On a site with WooCommerce Memberships, create a membership that is **not** tied to a subscription (e.g. a manually-assigned or order-based membership).
2. Trigger the membership expiry path for that membership (or expire a subscription-linked membership for a user who also holds a non-subscription one).
3. Before this change the request fatals with the undefined-method error above; after it, expiry proceeds without error.

A unit test (`tests/unit-tests/plugins/wc-memberships/membership-expiry.php`) covering the non-subscription bail-out is included.

### Other information:

* [x] The change is a cherry-pick of an already-reviewed-and-merged commit (`9b37834a`, Automattic/newspack-plugin#4743) on `main`/`alpha`.

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/wc-memberships/class-membership-expiry.php
+++ b/plugins/newspack-plugin/includes/plugins/wc-memberships/class-membership-expiry.php
@@ -28,11 +28,16 @@ class Membership_Expiry {
 	/**
 	 * Prevent membership expiration if there are other active subscriptions.
 	 *
-	 * @param bool                                                      $cancel_membership Whether to cancel the membership when the subscription is cancelled (default true).
-	 * @param \WC_Memberships_Integration_Subscriptions_User_Membership $user_membership The subscription-tied membership.
+	 * @param bool                            $cancel_membership Whether to cancel the membership when the subscription is cancelled (default true).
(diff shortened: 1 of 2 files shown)


adekbadek:
Merging since this was already approved in https://github.com/Automattic/newspack-plugin/pull/4743 and the issue is affecting live sites.

PR #396, opened 2026-06-23
fix(components): avoid interactive nesting in CoreCard header

thomasguillot:
### All Submissions:

`CardSettingsGroup` defaulted `onHeaderClick` to a noop and always forwarded it, so `CoreCard` rendered every settings-group header as a `<button>`. On the access-control cards that also carry a `ToggleControl` and an "Edit layout" link, those interactive elements ended up nested inside the header `<button>` — invalid HTML that is already shipped.

The header now renders as a `<button>` only when a real `onHeaderClick` handler is supplied **and** the header has no interactive children (no `toggle` action type, no header action, no actions menu); otherwise it renders as a `<div>`. `CardSettingsGroup` no longer injects a noop `onHeaderClick`, so the header type reflects whether the card is genuinely clickable.

Net effect: clickable selector cards stay keyboard-focusable buttons, while cards that own their own interactive controls become valid `<div>` headers.

Closes DSGNEWS-176.

1. Enable content gates (`define( 'NEWSPACK_CONTENT_GATES', true )`) and open **Audience → Access control**, then edit a gate.
2. Inspect the **Registered access** and **Paid access** cards: their header is now a `<div>`, and the toggle + "Edit layout" link render and operate normally (toggling enables/disables, the link opens the layout editor). Confirm the browser console shows no invalid-DOM-nesting warnings.
3. Inspect the **Restrict all** / **Choose specific** selector cards: their header is still a `<button>` and remains clickable and keyboard-focusable to switch the selection.
4. Check other `CardSettingsGroup` consumers for no regression: the **Newsletters → Settings** provider cards stay clickable, and the premium-newsletter / institutions access cards behave like the gate access cards above.
5. Verify no visual regression in card spacing or header borders across the above.

### Other information:


packages/components/src/card/core-card.js
```diff
@@ -80,11 +80,16 @@ const CoreCard = ( {
 	if ( noBorder ) {
 		otherProps.isBorderless = true;
 	}
+	// A button header would nest its interactive children (toggle, header action, or actions menu)
+	// inside a <button>, which is invalid. Only render the header as a button when a real click
+	// handler is supplied and the header has no interactive children of its own.
+	const hasInteractiveHeaderChildren = actionType === 'toggle' || !! headerAction || actions?.length > 0;
+	const headerIsButton = !! onHeaderClick && ! hasInteractiveHeaderChildren;
```
adekbadek, line 87: **Suggestion:** The new `headerIsButton` branching has no direct test. Existing coverage only exercises the button path (clicking a selector/provider card); nothing asserts that a header with a `toggle`, `headerAction`, or `actions` renders as a `<div>` rather than a `<button>` — which is exactly the invalid-nesting bug this PR fixes. A small test rendering `CoreCard` with `actionType="toggle"` + `headerAction` and asserting the header element is not a `button` would lock the fix against regression. (The `card/` directory has no existing test harness, so this is net-new scaffolding — hence a suggestion, not a blocker.)

packages/components/src/card/core-card.js
```diff
@@ -80,11 +80,16 @@ const CoreCard = ( {
 	if ( noBorder ) {
 		otherProps.isBorderless = true;
 	}
+	// A button header would nest its interactive children (toggle, header action, or actions menu)
+	// inside a <button>, which is invalid. Only render the header as a button when a real click
+	// handler is supplied and the header has no interactive children of its own.
+	const hasInteractiveHeaderChildren = actionType === 'toggle' || !! headerAction || actions?.length > 0;
```
adekbadek: **Suggestion:** `hasInteractiveHeaderChildren` covers `toggle` / `headerAction` / `actions`, but not `isDraggable`, which renders interactive `<Button>` move controls (chevron up/down) directly inside the header (lines 111-124). No current consumer combines `isDraggable` with `onHeaderClick`, so this is latent — but if one ever does, the header renders as a `<button>` wrapping those `<Button>`s, reintroducing exactly the invalid nesting this PR removes. Consider adding `|| isDraggable` to the guard, or documenting the invariant.

**Risk if not addressed:** a future draggable-and-clickable card silently ships invalid nested-button markup again, with no test to catch it.

adekbadek (approved):
Approving with suggestions

packages/components/src/card/core-card.js
thomasguillot: Added `src/card/core-card.test.js` (bfce779) with the net-new harness. It asserts the header renders as `<button>` for a plain `onHeaderClick`, and as a non-button when combined with `toggle`, `headerAction`, `actions`, or `isDraggable` — locking the invalid-nesting fix against regression.

packages/components/src/card/core-card.js
thomasguillot: Good catch — added `|| isDraggable` to the guard (bfce779), so a draggable-and-clickable header now renders as a `<div>` instead of wrapping the move `<Button>`s in a `<button>`. Covered by a dedicated test case.

PR #135, opened 2026-05-27
feat(reader-auth): unify auth + post-reg verification flows

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guidelines](https://github.com/Automattic/newspack-workspace/blob/main/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Three related reader-auth improvements, landed together because they share the same code paths:

**1. Post-registration email verification flow** (newspack-plugin + newspack-newsletters)

Every brand-new reader account is now offered an immediate OTP email-verification step before the page reloads, so they can verify without having to visit My Account later. The flow is wired into three registration entry points:

- The Reader Registration block — already had its own verification UI, but only when present on a gated post with a verification requirement; this PR refactors it onto a shared helper and offers the verification flow unconditionally.
  - Also presents the post-registration newsletter signup form (if configured) after completing or dismissing verification.
- The auth modal's unified signin/register form — previously had no verification step.
  - Also presents the post-registration newsletter signup form (if configured) after completing or dismissing verification.
- The Newsletter Subscription block — previously had no verification step.

The flow is: registration succeeds → verification modal appears → reader either clicks "Send code" (auth modal opens in OTP state to enter the code) or dismisses → newsletters signup modal appears (if configured) → page reloads. Dismissing any modal in the chain still leads to the newsletters prompt and reload, so the reader is never stranded.

**2. Unified signin + register auth modal form** (newspack-plugin)

Eliminates the dual-state auth modal. The two h2 titles, the "Create account" / "Sign into existing account" toggle buttons, and the `data-action="register"`-only elements are gone. There's now a single email field; the server decides whether to sign in or register based on whether the email is already associated with an account (mirroring how the Reader Registration block already works). The `#register_modal` and `#signin_modal` URL hashes both open the unified form.

**3. Cross-plugin verification API** (newspack-plugin)

Three additive contract changes so other plugins can opt into the verification flow:

- New `Reader_Activation::get_verification_payload( $user )` PHP helper — single source of truth for the `verified` / `verification_nonce` fields the frontend needs.
[...]

PR #400, opened 2026-06-23
fix(reader-verification): change "go back" button to "skip for now"

dkoo:
### All Submissions:

In the post-registration verification modal, changes the "Go back" button (which just closes the modal) to "Skip for now" instead. This makes it clearer what the button actually does and doesn't suggest to readers that verification is required at this point.

This PR doesn't change anything about when the verification modal appears—it will still always appear after registering via the auth modal, Registration block, or Newsletter Subscription block unless the `newspack_show_post_registration_verification` overrides this—but even if verification **is** actually required in the particular flow (such as when entering this flow from a content gate that requires verification to bypass) skipping will cause the page to refresh and continue to show the content gate but with a verification prompt until the reader completes verification.

Related to NPPD-1760.

1. Follow any of the flows from testing steps in #135 (A, B, or C)
2. When you get to the modal that says "We'll send a verification code to [email address]", confirm that the cancel button now says "Skip for now" instead of "Go back".

### Other information:


dkoo:
@thomasguillot tagging you for a design review since this touches on the UX for the post-reg verification flow.

thomasguillot (approved):
🚢

PR #402, opened 2026-06-23
Nextdoor integration: Use country-specific servers for different countries

claudiulodro:
### All Submissions:

This PR adjusts the Nextdoor integration to fix an issue people in non-US countries encountered when connecting their site.

It will be difficult to test this PR unless you are in Canada or Australia. In the US, go through the Nextdoor connection flow and verify it still works nicely. In Canada or Australia, the connection flow should throw an `Invalid input: country` error before this fix, and should connect nicely afterwards.

1. https://help.newspack.com/share-posts-to-nextdoor/

### Other information:


plugins/newspack-plugin/includes/optional-modules/nextdoor/class-controller.php
```diff
@@ -193,6 +193,11 @@ public static function api_start_oauth( $request ) {
 			return $account_response;
 		}
 
+		// Persist the country so subsequent calls (claim_page, articles) target the same region.
+		$settings            = Nextdoor::get_settings();
+		$settings['country'] = $country;
+		Nextdoor::update_settings( $settings );
+
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-workspace/pull/402/commits/d3ecbfc0a76079343171ae83fe57f8c50a5eb752

plugins/newspack-plugin/includes/optional-modules/class-nextdoor.php
```diff
@@ -322,6 +323,44 @@ public static function get_available_roles() {
...
+		/**
+		 * Filter the map of ISO 3166-1 alpha-2 country codes to Nextdoor API base URLs.
+		 *
+		 * @see https://developer.nextdoor.com/reference/displaying-availability
+		 *
+		 * @param array $hosts Map of country code to API base URL.
+		 */
+		$hosts = apply_filters( 'newspack_nextdoor_api_hosts', $hosts );
+
+		return isset( $hosts[ $country ] ) ? $hosts[ $country ] : $hosts['US'];
```
adekbadek, line 361: **Suggestion:** The fallback indexes `$hosts['US']`, but `$hosts` has just been passed through the public `newspack_nextdoor_api_hosts` filter — a caller that returns a map without a `US` key turns this into an undefined-key warning and a `null` URL, which then flows into `trailingslashit( null ) . ltrim( ... )` and produces a broken request for every unknown country. Fall back to a hardcoded literal (`'https://nextdoor.com'`) so the default can't be filtered away.

plugins/newspack-plugin/includes/optional-modules/nextdoor/class-auth.php
```diff
@@ -123,11 +123,16 @@ public static function refresh_access_token( $client_id, $client_secret, $access
...
+		// injected from env/constants are not written back to the database.
+		$saved = get_option( Nextdoor::SETTINGS_SLUG, [] );
+		if ( ! is_array( $saved ) ) {
+			$saved = [];
+		}
+		$saved['access_token']     = $token_data['access_token'];
+		$saved['token_expires_at'] = time() + $token_data['expires_in'];
 
-		Nextdoor::update_settings( $settings );
+		Nextdoor::update_settings( $saved );
```
adekbadek, line 135: **Suggestion:** The read-raw-option → `is_array` guard → merge → `update_settings()` block (plus its 2-line "why" comment) is copy-pasted verbatim across four sites (here and line 225, and `class-controller.php` lines 204/236). Each copy re-implements a security invariant — never persist centralized secrets — that lives only in a repeated comment. Extracting a single helper, e.g. `Nextdoor::update_saved_settings( array $values )` that reads the raw option, merges, and writes, would pin the invariant in one place.

**Consider:** routing the settings-save endpoint through the same helper — `api_update_nextdoor_settings()` in `includes/wizards/newspack/class-nextdoor-section.php:158-172` still does `get_settings()` → `update_settings()`, so on a platform with centralized credentials, saving the Nextdoor settings screen persists the shared client secret into `wp_options`. That's the same leak this PR closes elsewhere, left open at the most-exercised save path (out of this diff, but the helper would cover it for free).

plugins/newspack-plugin/includes/optional-modules/class-nextdoor.php
```diff
@@ -322,6 +323,44 @@ public static function get_available_roles() {
...
+	 * Nextdoor operates region-specific API servers; calling the wrong region
+	 * returns "Invalid input: country" during account creation.
+	 *
+	 * @see https://developer.nextdoor.com/reference/displaying-availability
+	 *
+	 * @param string|null $country ISO 3166-1 alpha-2 country code. If null, the
+	 *                             country saved in settings is used.
+	 * @return string API base URL (no trailing slash).
+	 */
+	public static function get_api_base_url( $country = null ) {
```
adekbadek, line 338: **Suggestion:** `get_api_base_url()` is pure, self-contained branching logic — case/whitespace normalization (`ca` → `CA`), unknown-country → US fallback, `null`/`''` → saved-country-then-US default, and the filter hook — and it's the crux of the fix, yet there's no test (no Nextdoor test file exists yet). A small unit test here would lock the region-mapping contract against regression cheaply and document the intended behavior.

plugins/newspack-plugin/includes/optional-modules/nextdoor/class-controller.php
```diff
@@ -193,6 +193,16 @@ public static function api_start_oauth( $request ) {
...
 		}
 
+		// Persist the country so subsequent calls (claim_page, articles) target the same region.
+		// Read the raw option (not Nextdoor::get_settings()) so centralized credentials
+		// injected from env/constants are not written back to the database.
+		$saved = get_option( Nextdoor::SETTINGS_SLUG, [] );
+		if ( ! is_array( $saved ) ) {
+			$saved = [];
+		}
+		$saved['country'] = strtoupper( $country );
```
adekbadek, line 203: **Nit:** `strtoupper( $country )` on persist omits the `trim()` that `get_api_base_url()` applies (`strtoupper( trim( ... ) )`). Harmless in practice — the REST `sanitize_text_field` callback already trims — but mirroring the normalization keeps the persisted value and the lookup consistent (and falls out naturally if the write goes through a shared helper).

adekbadek (approved):
Pre-approving, because the region fix itself is correct and well-scoped, but some things are worth addressing. The host map is a pure lookup with a safe US fallback, the `$country` override threads through only the onboarding call, and existing US publishers keep working with an empty stored country. The settings write-back hardening is a real improvement. No blockers; the notes below are worth a look before merge, and the security-fix completeness one (the settings-save endpoint still writes back `get_settings()`) is the most impactful.

- `plugins/newspack-plugin/includes/optional-modules/nextdoor/class-controller.php:44` — **Suggestion:** The `country` arg is only `sanitize_text_field`'d, with no validation against the supported set. An unsupported code (e.g. `GB`) makes `get_api_base_url()` fall back to the US host while `create_account()` still sends `country => GB` in the request body — reproducing the exact `Invalid input: country` error this PR exists to eliminate, just for a different country. Adding a `validate_callback` (or `enum`) derived from `Nextdoor::get_available_countries()` rejects unsupported regions up front with a clear local error. The UI dropdown constrains this today, but the endpoint is the trust boundary.
- `plugins/newspack-plugin/includes/optional-modules/nextdoor/class-auth.php:172` — **Nit:** Out of scope / pre-existing (this PR only touched the token write-back below): `handle_oauth_callback()` processes the OAuth `code` from `$_GET` on `init` with no `state` parameter validation — the standard OAuth CSRF defense. An attacker holding an authorization code for their own Nextdoor account could trick a logged-in admin into hitting the callback and bind the site to the attacker's account. Worth a follow-up ticket rather than anything for this PR.

PR #404, opened 2026-06-24
fix(group-subscription): gate management UX behind content gates flag (NPPD-1752)

adekbadek:
### All Submissions:

The group subscription management UX shipped in #148 (live since `newspack@6.44.0-alpha.1`) was not gated behind the Access Control feature flag (`NEWSPACK_CONTENT_GATES`). On un-migrated sites still running Woo Memberships, the reader-facing group endpoint and management flows could surface even though Access Control is not the source of truth.

This gates the entire reader-facing group management surface behind the flag, composing cleanly with the existing `Memberships::is_active()` suppression:

- **`Group_Subscription_MyAccount::init()`** – early-returns when the flag is off, so the `/group/` My Account endpoint, its rewrite rule, query var, and management UX are never registered. The rewrite-rules option version was bumped (`_v2` → `_v3`) so the first request after the flag turns on re-flushes and regenerates the endpoint's rewrite rule.
- **`Group_Subscription_Invite::init()`** – early-returns before registering its email-config hooks. Static helpers remain callable.
- **`Group_Subscription_API::register_routes()`** – early-returns before registering the REST routes that back the reader UX and the admin meta box.
- **`My_Account_UI_v1`** – the group nav entry now requires `Content_Gate::is_newspack_feature_enabled()` in addition to the existing `! Memberships::is_active()` check.

The publisher-facing group-label controls (`Group_Subscription_Settings`, `Audience_Wizard`) were already gated per-callback; no change needed there.

This is an alpha-hotfix: the management UX is already on `alpha`, so this PRs to `alpha` first. A companion PR targets `main` (merges post-release).

#### Note on gating style

This gates at the single registration chokepoint (`init()` / `register_routes()`) rather than per-callback like `Group_Subscription_Settings`. These classes are pure-UX with no must-always-run hooks, so the chokepoint is sufficient and clearer. `Group_Subscription_Settings::admin_enqueue_scripts()` remains ungated (pre-existing, out of scope).

Closes NPPD-1752.

1. With `NEWSPACK_CONTENT_GATES` **undefined/off** and Woo Memberships active, confirm the `/my-account/group/` endpoint 404s, the `newspack-group-subscription/v1` REST routes are absent, and no group nav entry shows in My Account.
2. Define `NEWSPACK_CONTENT_GATES` (true) and load any page once (to flush rewrite rules). Confirm the `/my-account/group/` endpoint resolves, the REST routes register, and the group nav entry appears for an eligible group owner/manager.
3. Run the unit tests: `n test-php --filter Group_Subscription` and `n test-php --filter Group_Management_Nav` (both green).

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
@@ -25,6 +25,12 @@ class Group_Subscription_API {
 	 * Register REST API routes.
 	 */
 	public static function register_routes() {
+		// The group management REST routes back the reader-facing My Account UX
+		// and the admin meta box, both gated behind the Access Control feature
+		// flag. Don't register the routes on un-migrated sites.
+		if ( ! Content_Gate::is_newspack_feature_enabled() ) {
+			return;
+		}
 		\register_rest_route(
 			self::NAMESPACE,
 			'/search-users',
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
@@ -60,6 +60,13 @@ class Group_Subscription_Invite {
 	 * Initialize hooks.
 	 */
 	public static function init() {
+		// Invite acceptance and the invite email config are part of the group
+		// management UX, gated behind the Access Control feature flag. The
+		// static invite helpers remain available regardless; only the hooks are
+		// gated.
+		if ( ! Content_Gate::is_newspack_feature_enabled() ) {
+			return;
+		}
 		add_filter( 'newspack_email_configs', [ __CLASS__, 'add_email_config' ] );
 		add_action( 'template_redirect', [ __CLASS__, 'process_invite_request' ] );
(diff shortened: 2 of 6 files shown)


adekbadek:
Companion `main` PR (merges post-release): #405

jason10lee (approved):
Approving. Straightforward, well-scoped fix — the gating sits at the right altitude (the `init()` / `register_routes()` chokepoints), composes cleanly with the existing `Memberships::is_active()` suppression, and leaves all the security-relevant surfaces (REST permission callbacks, nonces, capability checks) untouched, so it can only *remove* surface when the flag is off. The `_v2`→`_v3` rewrite-rules bump is the right re-flush mechanism.

I verified the gate in an isolated env: with `NEWSPACK_CONTENT_GATES` off, zero `newspack-group-subscription/v1` routes register and neither the `group` nor `manage-members` My Account endpoint is wired; flipping the flag on brings all five routes and both endpoints back. Any failure mode here would be pretty attenuated — the sort of thing our automated checks are well-positioned to catch — and there's no publisher-facing risk, since nothing outside newspack-plugin consumes this surface.

One optional follow-up, not a blocker: the unit tests only exercise the flag-*on* path, so the flag-*off* suppression isn't asserted in CI (the process-global `define()` in `set_up()` makes the off-case awkward — a `@runInSeparateProcess` test would close it). Minor: test-step 1 says `/my-account/group/` "404s" with the flag off; on an already-upgraded alpha site it renders the dashboard instead (suppression still holds).

LGTM.

PR #405, opened 2026-06-24
fix(group-subscription): gate management UX behind content gates flag (NPPD-1752)

adekbadek:
### All Submissions:

**Companion to #404 (alpha-hotfix).** Identical change targeting `main`; merges post-release. The fix is already on `alpha` via #404, so this PR brings `main` in sync without waiting for the alpha→main release merge.

The group subscription management UX shipped in #148 (live since `newspack@6.44.0-alpha.1`) was not gated behind the Access Control feature flag (`NEWSPACK_CONTENT_GATES`). On un-migrated sites still running Woo Memberships, the reader-facing group endpoint and management flows could surface even though Access Control is not the source of truth.

This gates the entire reader-facing group management surface behind the flag, composing cleanly with the existing `Memberships::is_active()` suppression:

- **`Group_Subscription_MyAccount::init()`** – early-returns when the flag is off, so the `/group/` My Account endpoint, its rewrite rule, query var, and management UX are never registered. The rewrite-rules option version was bumped (`_v2` → `_v3`) so the first request after the flag turns on re-flushes and regenerates the endpoint's rewrite rule.
- **`Group_Subscription_Invite::init()`** – early-returns before registering its email-config hooks. Static helpers remain callable.
- **`Group_Subscription_API::register_routes()`** – early-returns before registering the REST routes that back the reader UX and the admin meta box.
- **`My_Account_UI_v1`** – the group nav entry now requires `Content_Gate::is_newspack_feature_enabled()` in addition to the existing `! Memberships::is_active()` check.

The publisher-facing group-label controls (`Group_Subscription_Settings`, `Audience_Wizard`) were already gated per-callback; no change needed there.

#### Note on gating style

This gates at the single registration chokepoint (`init()` / `register_routes()`) rather than per-callback like `Group_Subscription_Settings`. These classes are pure-UX with no must-always-run hooks, so the chokepoint is sufficient and clearer. `Group_Subscription_Settings::admin_enqueue_scripts()` remains ungated (pre-existing, out of scope).

Closes NPPD-1752.

1. With `NEWSPACK_CONTENT_GATES` **undefined/off** and Woo Memberships active, confirm the `/my-account/group/` endpoint 404s, the `newspack-group-subscription/v1` REST routes are absent, and no group nav entry shows in My Account.
2. Define `NEWSPACK_CONTENT_GATES` (true) and load any page once (to flush rewrite rules). Confirm the `/my-account/group/` endpoint resolves, the REST routes register, and the group nav entry appears for an eligible group owner/manager.
3. Run the unit tests: `n test-php --filter Group_Subscription` and `n test-php --filter Group_Management_Nav` (both green).

### Other information:


adekbadek:
Alpha-hotfix PR (merge first): #404

jason10lee (approved):
Approving — and noting up front that this is still a draft. I'm reading that as simply parked behind the release of its companion alpha PR (#404) rather than unfinished work, so I'm comfortable approving now; re-ping me if the scope changes before it's marked ready.

This is the companion to the alpha gating PR, and the same clean, well-scoped change. The gating composes correctly with the extra REST state-guards and the recategorized invite email already on `main`: those live inside the gated callbacks/filters, so when the flag is off they're simply unreachable rather than half-wired.

I verified the gate in an isolated env against `main`: with `NEWSPACK_CONTENT_GATES` off, zero `group-subscription/v1` routes register, neither My Account endpoint is wired, and the group-invite email config is absent from the Emails set; flipping the flag on brings all of it back — including the invite email config the new `emails-section.php` test covers. Nice touch that this test attaches/detaches the filter with a proper `set_up`/`tear_down` + cache reset.

Same optional, non-blocking follow-up as the alpha PR: the content-gate tests only exercise the flag-*on* path (the process-global `define()` in `set_up()` makes the off-case awkward — a `@runInSeparateProcess` test would close it). Failure modes here would be subtle and well within what our automated checks catch.

LGTM.

PR #406, opened 2026-06-24
fix(build): non-interactive pnpm + robust standalone repo builds

jason10lee:
### All Submissions:

Makes `n build` / `n ci-build` for a standalone `repos/` checkout install and build more reliably:

* **No silent failures.** A failed Composer/JS install or build for a standalone repo now fails the command instead of being swallowed into a green build. (Improves both `n build` and `n ci-build`.)
* **Reproducible CI installs.** Under `n ci-build`, a standalone repo's JS install is frozen/clean *where the package manager + lockfile support it* — `pnpm install --frozen-lockfile` with a `pnpm-lock.yaml`, `npm ci` with a `package-lock.json`. A lockfile-less repo, a Yarn repo, or plain `n build` falls back to a normal install so it still builds.
* **Non-interactive where it matters.** The pnpm frozen install uses a *scoped* `CI=true` so it doesn't block on a TTY-less prompt — without exporting `CI` globally (which would flip pnpm's frozen-lockfile default onto dev-mode/workspace installs).

Composer keeps dev deps (consistent with the monorepo composer paths) and runs `--no-interaction`. Monorepo (`plugins/`/`themes/`) builds are unchanged.

1. Place a standalone checkout at `repos/plugins/<name>/` with a `composer.json` and a `package.json` declaring a `build` script.
2. `n build <name>` installs the repo's Composer + JS deps and runs its own `build`; a monorepo `n build <plugin>` still works.
3. With a committed lockfile, `n ci-build <name>` uses a frozen/clean install and doesn't hang on a pnpm prompt in a TTY-less context.
4. Remove the lockfile, re-run `n ci-build <name>` → falls back to a plain install and still builds (no frozen-lockfile failure).
5. Break the install (unresolvable Composer requirement) → `n build <name>` exits non-zero instead of reporting success.
6. After editing a monorepo plugin's `package.json`, `n build` / `n ci-build` is unaffected (no spurious frozen-lockfile failure).

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — no `bin/` test harness in this repo.
* [x] Have you successfully run tests with your changes locally? — verified in an isolated env; reviewed via multi-agent + cross-provenance (magi) code review, which caught and fixed a `CI=true` scoping regression.


bin/build-repos.sh
```diff
@@ -33,18 +33,54 @@ package_filter_for_dir() {
...
+        return 0
+    fi
+
+    # Detect package manager from lockfile; default npm for a lockfile-less repo
+    # (npm's lenient hoisting is the safest default, and matches the prior code).
     local pm="npm"
-    [ -f "$dir/pnpm-lock.yaml" ] && pm="pnpm"
-    [ -f "$dir/yarn.lock" ] && pm="yarn"
-    ( cd "$dir" && "$pm" install )
+    if [ -f "$dir/pnpm-lock.yaml" ]; then
```
dkoo, line 58: **Severity:** suggestion

This `if`/`elif` changes dual-lockfile precedence versus the previous code. The old logic used sequential assignments (`pm="pnpm"` then `pm="yarn"`), so a checkout containing **both** `pnpm-lock.yaml` and `yarn.lock` resolved to **yarn** (last assignment wins). With `if`/`elif`, pnpm now wins (first match). For a repo mid-migration that still carries both lockfiles, this silently switches the tool — and in CI, `pnpm install --frozen-lockfile` against a stale pnpm lock would fail where the yarn path would have succeeded. If yarn precedence was intentional, check `yarn.lock` before `pnpm-lock.yaml`; otherwise this is fine as-is, just worth confirming the flip is deliberate.

dkoo (commented):
Solid, well-commented hardening of the standalone `repos/` build path. The core claims hold up: failed composer/JS install/build now propagate via `|| return 1`, and that return value reaches the script's exit code correctly — no `set -e` needed, since the call is the last command in the `*)` case arm (reproduced: a failing build exits non-zero). The frozen/clean CI installs are correctly guarded by lockfile presence, the scoped (non-exported) `CI=true` on the pnpm frozen install is right, and the monorepo `all`/workspace-package paths are genuinely untouched. Release risk: LOW.

Two non-blocking items are inline. The one worth a real second look: the package-manager detection changed from sequential assignments (last-wins) to `if`/`elif` (first-wins), which flips precedence for a checkout that has both a `pnpm-lock.yaml` and a `yarn.lock` — previously yarn won, now pnpm does. In CI that repo would then run `pnpm install --frozen-lockfile` against a possibly-stale pnpm lock and fail where the yarn path would have succeeded. Dual committed lockfiles are a rare/transient state, so it's a suggestion rather than a blocker — but if yarn precedence was intentional, it's worth restoring.

One pre-existing note, out of this PR's scope: the workspace `pnpm install` (line 105) still isn't guarded, so a workspace-install failure during `n ci-build <standalone>` would be masked by the standalone build's exit status. Worth a follow-up if "green means green" is meant to cover the whole invocation.

bin/build-repos.sh
```diff
@@ -33,18 +33,54 @@ package_filter_for_dir() {
...
+    # Install JS deps. In CI mode use a reproducible install where the package
+    # manager + lockfile support it (pnpm is only selected when its lockfile
+    # exists, so --frozen-lockfile is always satisfiable; npm ci needs a
+    # package-lock.json). Everything else -- Yarn (avoids the Classic-vs-Berry
+    # frozen-flag split), a lockfile-less repo, or dev mode -- gets a plain
+    # install so it still builds. CI=true is scoped to the pnpm frozen install
+    # (non-interactive in a TTY-less CI shell) and is deliberately NOT exported:
+    # a global CI=true also flips pnpm's frozen-lockfile default onto the
+    # dev-mode and workspace installs, which would break them on a stale lockfile.
+    if [ "$MODE" = "ci" ] && [ "$pm" = "pnpm" ]; then
```
dkoo, line 73: **Severity:** nit

`build_standalone_repo` reads `MODE` as a script-level global rather than taking it as a parameter. It's correct today because `MODE="$2"` is assigned before the only call site, but the function is *defined* above that assignment — if a call site is ever added earlier in the file, `MODE` would be unset and every standalone build would silently take the dev-mode plain-install branch. Passing mode as `$2` (mirroring `dir="$1"`) would make the dependency explicit and order-independent. Purely maintainability; no current defect.

bin/build-repos.sh
jason10lee: The flip is real and deliberate. `main` does resolve to yarn for a both-lockfiles checkout (sequential `&&`, last assignment wins), and the `if`/`elif` here makes it pnpm. Going with pnpm-first: it's the workspace's package manager, and the branch it guards (`pnpm install --frozen-lockfile`) is only reachable when `pnpm-lock.yaml` exists, so it's always satisfiable. #410 puts the same `if`/`elif` in `test-js.sh`; merging this one first makes the two agree.

bin/build-repos.sh
jason10lee: Leaving this one. Shell resolves variables when a function runs, not where it's defined, so declaring `build_standalone_repo` above `MODE="$2"` isn't what would create the exposure — only call order could, and the one call site sits well after the assignment. `MONOREPO_ROOT` and `REPOS_PATH` are read the same way, as script-level config, so passing only `MODE` explicitly would make it the exception rather than the pattern.

bin/build-repos.sh
dkoo: Confirmed against the code — `origin/main`'s `build-repos.sh` resolves a both-lockfiles checkout to **yarn** (sequential assignments, last wins), and this branch flips it to **pnpm** via `if`/`elif` (first match). Agreed the pnpm-first choice is deliberate and defensible: it's the workspace's package manager, and the `pnpm install --frozen-lockfile` branch is only reachable when `pnpm-lock.yaml` is present. #410 uses the identical structure, so the two agree once this lands first. Resolving — thanks for confirming the flip is intentional.

bin/build-repos.sh
dkoo: Fair — you're right that the shell resolves the variable at call time, not where the function is defined, so the only real exposure is call order, and the sole call site sits well after `MODE="$2"`. And since `MONOREPO_ROOT`/`REPOS_PATH` are read the same script-global way, passing only `MODE` explicitly would make it the exception rather than the pattern. Non-issue; leaving as-is.

dkoo (approved):
Approving after manual QA. Ran all the named commands from the PR description against a real standalone `repos/` checkout (`newspack-manager-admin`), and they behave exactly as described:

- `n build <name>` / `n ci-build <name>` install the repo's Composer + JS deps and run its own build; a monorepo `n build <plugin>` is unaffected.
- The CI path uses a frozen/clean install where the lockfile supports it and doesn't hang on pnpm's TTY-less purge prompt — the scoped `CI=true` works as intended.
- Removing the lockfile falls back to a normal install and still builds; breaking the install exits non-zero instead of reporting a green build.

(Builds take a while because each run re-reconciles the root workspace deps *and* the named repo's — see the note below — but the behavior is correct throughout.)

Both earlier review threads are resolved: the both-lockfiles yarn→pnpm precedence flip is deliberate (pnpm-first, mirrored by #410 so the two agree once this lands first), and the `MODE` script-global read is consistent with the file's existing `MONOREPO_ROOT`/`REPOS_PATH` pattern.

One pre-existing, non-blocking observation (not introduced by this PR): `build-repos.sh` runs the full workspace `pnpm install` at the monorepo root before every single-repo build, so `n build <standalone>` re-reconciles all ~19 workspace projects each time — that's the bulk of the runtime, and that root install's failure isn't guarded. Worth a follow-up someday; out of scope here.

LGTM. 🚀

PR #407, opened 2026-06-24
fix(collections): preserve inner blocks in card indicator

adekbadek:
### All Submissions:

Monorepo port of [Automattic/newspack-plugin#4726](https://github.com/Automattic/newspack-plugin/pull/4726) (by @laurelfulford). The original PR fixed this in the legacy `newspack-plugin` repo, but the same bug still exists unfixed in the monorepo, which is now the primary release source. Code is from the original PR; this PR carries it to where it ships.

In Collections, the 'Card' Collection Indicator Style reassembled post content from `$block['innerHTML']`, which drops inner blocks - e.g. `core/list-item` children of a `core/list`, or columns in a Columns block. Switching to `render_block( $block )` preserves inner blocks (and renders dynamic blocks correctly).

Also includes:
- A regression test (`test_insert_after_nth_block_preserves_inner_blocks`) asserting `<li>` items survive.
- An update to the existing test for the WP 7.0+ `wp-block-paragraph` class now emitted by `render_block()`.

Closes NPPM-2831.

1. On a site running Collections, go to Newspack > Settings > Collections and set 'Collection Indicator Style' to 'Card'.
2. Create a post with a List block containing several items, assigned to a published collection. Publish.
3. View the post on the front-end; before this fix the list renders as an empty `<ul></ul>` (items dropped). After it, the items appear.
4. Smoke test that the post otherwise renders correctly.

Unit-level: `n test-php --filter Content_Inserter` (14 tests, 56 assertions). The regression test fails on the old `innerHTML` code (empty `<ul>`) and passes with `render_block()`.

### Other information:


plugins/newspack-plugin/includes/collections/class-content-inserter.php
```diff
@@ -264,7 +264,8 @@ function ( $block ) use ( $blocks_to_skip_empty ) {
 		// Insert after the nth block.
 		$rendered_content = '';
 		foreach ( $parsed_blocks as $index => $block ) {
-			$rendered_content .= $block['innerHTML'];
+			// Use render_block() so inner blocks (e.g. core/list-item children of core/list) and dynamic blocks are emitted.
+			$rendered_content .= render_block( $block );
```
dkoo, line 273: **Blocker:** render_block() at the_content priority 1 re-enters the indicator filter before the run-once guard is set — `maybe_insert_collection_indicators()` runs on `the_content` at priority 1 and only sets its re-entrancy guard `self::$the_content_has_rendered = true` at the *end* (line 118), after `insert_after_nth_block()` (line 106) has already called `render_block()` on every parsed block. Core dynamic blocks re-apply `the_content` while rendering — confirmed: `core/post-content` (wp-includes/blocks/post-content.php:50), plus `core/template-part` and `core/post-time-to-read`. So when a card-style post that is in a published collection contains e.g. a Query Loop (`core/query` › `core/post-template` › `core/post-content`), rendering it here re-enters this filter with the guard still `false` and `in_the_loop()` still `true` (it reads the global main query, which stays in-loop during a synchronous render). The nested call then injects the *outer* post's collection card into the *queried* post's content. I reproduced this in a live env (WP 7.0): the `collection-card` markup appears both before and inside the nested region. This is a regression vs. pre-PR behavior, where the dynamic block had empty `innerHTML` and was dropped here, then rendered by core's `do_blocks` at priority 9 when the guard was already set — so no leak. Fix: set `self::$the_content_has_rendered = true` *before* calling `insert_after_nth_block()` (move it above the `switch`), so any nested re-entry short-circuits at the guard check. Add a regression test with a `core/post-content`/Query Loop block in a card-style collection post.

plugins/newspack-plugin/tests/unit-tests/collections/class-test-content-inserter.php
```diff
@@ -264,20 +264,44 @@ public function test_insert_after_nth_block() {
 		$block_content = "<!-- wp:paragraph -->\n<p>First paragraph.</p>\n<!-- /wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p>Second paragraph.</p>\n<!-- /wp:paragraph -->\n\n<!-- wp:paragraph -->\n<p>Third paragraph.</p>\n<!-- /wp:paragraph -->";
 		$block_result  = Content_Inserter::insert_after_nth_block( $block_content, $insert_html, 2 );
 
-		$this->assertStringContainsString( '<p>First paragraph.</p>', $block_result, 'First paragraph should be present.' );
-		$this->assertStringContainsString( '<p>Second paragraph.</p>', $block_result, 'Second paragraph should be present.' );
+		// WP 7.0+ adds a `wp-block-paragraph` class to rendered paragraph blocks.
```
dkoo: **Suggestion:** Test couples to the WP 7.0+ wp-block-paragraph class — The updated assertions hardcode `<p class="wp-block-paragraph">`, a class core only started emitting in WP 7.0+. CI runs against the latest stable WP so it passes today, but the suite becomes brittle across WP versions (it would fail on pre-7.0, or if core changes the class again). Consider asserting on the inner text (`First paragraph.`) plus ordering (via `strpos`) instead of the core-generated wrapper class — which is what the new inner-block test already does (it asserts on `<li>` content). This keeps the test focused on the behavior under review rather than a core rendering detail.

plugins/newspack-plugin/tests/unit-tests/collections/class-test-content-inserter.php
```diff
@@ -264,20 +264,44 @@ public function test_insert_after_nth_block() {
...
 		$this->assertGreaterThan( $second_pos, $insert_pos, 'Inserted content should come after second paragraph.' );
 		$this->assertLessThan( $third_pos, $insert_pos, 'Inserted content should come before third paragraph.' );
 	}
 
+	/**
+	 * Test insert_after_nth_block preserves inner blocks (e.g., list items).
+	 *
+	 * Regression: a `core/list` block stores its `<li>` items as `core/list-item`
+	 * inner blocks, not in `innerHTML`. The same applies to columns, groups,
+	 * buttons, etc. Reassembling parsed blocks must use `render_block()` so
```
dkoo, line 350: **Nit:** New regression test could also lock in inner-block placement and ordering — `test_insert_after_nth_block_preserves_inner_blocks` correctly fails on the old `innerHTML` code and passes now, so it genuinely guards the regression. Optional strengthening: also assert the surrounding `<ul class="wp-block-list">` wrapper and the inserted-content ordering, so it locks in that list items render *within* their parent rather than merely appearing somewhere in the output.

dkoo (changes requested):
Confirmed the fix and the blocker (details inline) in manual testing.

Thanks for porting this — the core fix is right and the regression test is a good guard. `render_block()` is the correct way to reassemble parsed blocks so inner blocks (list items, columns) and dynamic blocks survive, and I confirmed the existing suite (14 tests / 56 assertions) passes on this branch.

One blocker before merge: rendering blocks inside the `the_content` priority-1 filter introduces a re-entrancy regression. `self::$the_content_has_rendered` is set only *after* `insert_after_nth_block()` runs, so a dynamic block that re-applies `the_content` during that render (Query Loop / `core/post-content`, `core/template-part`, `core/post-time-to-read`) re-enters the indicator filter with the guard still false. The result: a card-style post that is in a published collection and contains such a block injects the *outer* post's collection card into the *nested/queried* post's content — and that output is cached for all readers. I reproduced this in a live env (WP 7.0). It's a one-line fix plus a regression test. There's also a smaller test-brittleness suggestion below.

plugins/newspack-plugin/includes/collections/class-content-inserter.php
adekbadek: Fixed in da9f45b. Reproduced the leak first with a failing regression test (`test_card_indicator_not_reinjected_into_nested_the_content`): a dynamic block that re-applies `the_content` during render produced two cards (outer + nested), exactly as you described. Moved `self::$the_content_has_rendered = true` above the `switch` so any re-entry short-circuits at the guard check; the test now asserts the card renders exactly once. Confirmed the normal single-post render is unchanged (card once, list items intact). Great catch - this was latent in the original PR too.

plugins/newspack-plugin/tests/unit-tests/collections/class-test-content-inserter.php
adekbadek: Done in da9f45b. The paragraph assertions now check the inner text (`First paragraph.`) plus ordering via `strpos`, dropping the WP-version-coupled `wp-block-paragraph` wrapper class - matching the approach the inner-block test already uses.

plugins/newspack-plugin/tests/unit-tests/collections/class-test-content-inserter.php
adekbadek: Done in da9f45b. `test_insert_after_nth_block_preserves_inner_blocks` now also asserts the `<ul class="wp-block-list">` wrapper, that both items render within it (between `<ul>` and `</ul>`), that the inserted content precedes the list, and that item order is preserved.

adekbadek:
Thanks for the thorough manual review @dkoo - all three points addressed in da9f45b:

- **Blocker (re-entrancy / cached card leak into nested `the_content`):** Reproduced first with a failing regression test (a dynamic block that re-applies `the_content` during render yielded two cards), then fixed by moving the `self::$the_content_has_rendered` guard above the `switch` so nested re-entry short-circuits. Test now asserts exactly one card; normal single-post render confirmed unchanged. This was latent in the original PR too, so good thing we re-reviewed on the way over.
- **Suggestion (WP 7.0 `wp-block-paragraph` coupling):** Assertions now check inner text + ordering instead of the core wrapper class.
- **Nit (lock in inner-block placement):** The inner-blocks test now also asserts the `<ul class="wp-block-list">` wrapper, that items render within it, and the insert/order positions.

Full suite green (15 tests, 63 assertions).

dkoo (approved):
Thanks for the updates, @adekbadek! All three points from the previous review are addressed and verified. The guard is now set before the blocks are rendered, so a dynamic block that re-applies `the_content` during the card render (Query Loop / `core/post-content`, etc.) short-circuits on re-entry instead of leaking the outer post's collection card into nested/queried content; there's a dedicated regression test asserting the card renders exactly once. The test assertions no longer couple to the WP 7.0+ `wp-block-paragraph` class, and the inner-blocks test now locks in list-wrapper placement and ordering.

I re-verified on the live front-end: the Query Loop collection post that previously rendered three cards (two leaked into the queried siblings) now renders exactly one, with the list-block post still rendering its items correctly. Full suite green (15 tests / 63 assertions). LGTM — thanks for the quick, thorough turnaround.

PR #408, opened 2026-06-24
feat(content-gate): add pre-save checklist and gate preferences

thomasguillot:
## Summary

Adds a deliberate "pre-save" experience plus a publisher Preferences modal to the Newspack gate editor. That editor is shared by **Audience › Access control** (content gates) and **Premium Newsletters**, so this flow applies to both surfaces. New gates now default to **Inactive** so nothing goes live by accident, and saving surfaces a review panel that mirrors the access summary before you commit.

Closes [NPPD-1708](https://linear.app/a8c/issue/NPPD-1708/add-an-option-to-save-a-gate-as-a-draft)

## What's included

- **Pre-save panel.** A right-side slide-in shown on Save: "Are you ready to save?", a Status (Active/Inactive) choice, a read-only summary mirroring the access-control card, and an "Always show pre-save checks" toggle.
- **Gate Preferences modal** (kebab › Preferences). Per-user "Enable pre-save checks" and per-site "Default status for new gates", each with explanatory help text.
- **Status controls.** "Set to active/inactive" in both the edit kebab and the gate card. The edit kebab uses a `MenuGroup` divider: Rename · Set to active/inactive · Delete · ─── · Preferences.
- **Save disabled until dirty.** The gate/newsletter Save and the Preferences Save stay disabled unless there is actually something to save.
- **Shared summary.** `getGateSummarySections()` powers both the list card and the panel so they can't drift.
- **Works on Premium Newsletters too.** The gate Edit component is shared with the premium-newsletters wizard, so the pre-save panel, status controls, Preferences, and Save-disabled behavior all apply there as well. The `/preferences` route and localized preference keys were added to that wizard, with `Content_Gate` as the single source of truth.
- **Backend.** New gates honor the sent status (default `draft`), a `newspack_content_gate_default_status` option, and a `/preferences` REST route.

## Notes

- **Infra:** adds a workspace-wide `.npmrc` that hoists `babel-jest`, `@babel/preset-env`, and `@babel/preset-typescript` so `newspack-scripts` jest resolves them under pnpm's strict layout. Flagging since it is workspace-level.
- Copilot's automated review could not run (monthly quota exceeded); the change went through a multi-reviewer (Codex + Claude) pass instead.

## Test plan

- `npm test`: 278/278 JS tests pass
- `npm run lint:js`, `npm run tsc`, `npm run lint:php`: clean
- `npm run build`: compiles
- Browser-QA'd the full flow: create › pre-save panel › uncheck › save › edit › Preferences › re-enable.

## Screenshots

<table>
  <tr>
    <td width="50%"><b>Add new gate</b><br>gate-01-add-new-form</td>
    <td width="50%"><b>Pre-save panel</b><br>gate-02-presave-panel</td>
  </tr>
  <tr>
    <td><b>Edit kebab</b><br>gate-03-edit-kebab</td>
    <td><b>Preferences</b><br>gate-04-preferences-modal</td>
  </tr>
  <tr>
    <td><b>Gate card kebab</b><br>gate-05-card-kebab</td>
    <td></td>
  </tr>
</table>

## Walkthrough

https://github.com/user-attachments/assets/88e3897d-bb33-40bc-a415-b6ea80475004


plugins/newspack-plugin/includes/content-gate/class-content-gate-api.php
```diff
@@ -112,7 +112,7 @@ public static function sanitize_gate( $gate ) {
 		$sanitized = [
 			'title'         => isset( $gate['title'] ) ? sanitize_text_field( $gate['title'] ) : __( 'Untitled Content Gate', 'newspack-plugin' ),
 			'priority'      => isset( $gate['priority'] ) ? intval( $gate['priority'] ) : 0,
-			'status'        => isset( $gate['status'] ) && ! empty( $gate['id'] ) ? self::sanitize_status( $gate['status'], $gate['id'] ) : 'draft',
+			'status'        => isset( $gate['status'] ) ? self::sanitize_status( $gate['status'], $gate['id'] ?? 0 ) : 'draft',
```
adekbadek: **Suggestion:** Omitted status on update resets a gate to draft — `sanitize_gate()` is the `sanitize_callback` for the update route too, and it now unconditionally injects a `status` key (`'draft'` when absent). That makes the preserve-fallback in `update_gate_settings()` (`class-content-gate.php:1285`, `... : $post->post_status`) dead code: an update request that omits `status` resets an already-published gate to draft, and a draft gate stops enforcing (`class-content-restriction-control.php:160`). This behaviour is unchanged from before the PR and the shipped UI always sends `status`, so it only bites programmatic callers — but the polarity is revenue-adverse. Confirmed reproducible against a running env: a REST update of a published gate that omits `status` silently reset it to `draft` (a draft gate stops enforcing). The `content_rules_match` field a few lines up already got a guard against exactly this ("only include when the request provided it"); `status` deserves the same, or fall back to the current post status on update.

**Consider:** If any automation or external integration updates gate settings via the REST route, it must include the gate's current `status` in the payload. Because the update route's sanitizer now always fills in `status` (defaulting to draft when omitted), an update that leaves status out will reset a published, enforcing gate to draft — and a draft gate stops gating, so previously-protected content silently becomes freely readable.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -1326,6 +1326,66 @@ public static function get_post_statuses() {
...
+	/**
+	 * Get the default status ('publish' or 'draft') for newly created gates.
+	 *
+	 * Defaults to 'draft' (inactive) so new gates are set up before going live.
+	 * Only affects gates created going forward; existing gates keep their own
+	 * status. Publishers can change this default in the Access control preferences.
+	 *
+	 * @return string
+	 */
+	public static function get_default_new_gate_status() {
```
adekbadek, line 1362: **Suggestion:** Default-status option is enforced client-side only — `get_default_new_gate_status()` is consumed only to populate the localized `default_gate_status` value for the editor JS (`:101` / `:230`); it is never read by `create_gate()`. So the "Default status for new gates" site option is authoritative only in the browser — a gate created via REST/PHP without a `status` gets `draft` from `sanitize_gate()` regardless of the option. As shipped this is the *safe* direction (draft), so no action is strictly required; if you want the option to be authoritative server-side, see the caveat below about the Memberships path. Also worth confirming: this option is a single site-wide value shared by both the Access control and Premium Newsletters editors, so changing it in one changes it in the other — the modal copy doesn't hint at that cross-feature scope.

**Consider:** Do not change `create_gate()`'s `post_status => 'publish'` fallback to draft or to the new default-status option. That fallback is reached only by direct PHP callers, and the only ones are the WooCommerce Memberships auto-gate creators (`class-memberships.php:150` and `:286`, which pass no status). Memberships recognises a gate only when it is published (`class-memberships.php:217`), so a Memberships gate created as draft becomes invisible to enforcement and the membership-restricted content is served free until the gate is manually published.

**Consider:** Note the 'default status for new gates' preference is a single site-wide option shared by both the Audience content-gates and Premium Newsletters editors; changing it in one surface changes it in the other. It is also honoured only in the wizard UI (it pre-fills the status control in JS) and is not read when a gate is created server-side, so setting the default to 'publish' will not cause gates created outside the standard UI flow to go live.

plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -526,6 +549,31 @@ public function update_gate_priorities( $request ) {
 		return rest_ensure_response( $updated_gates );
 	}
 
+	/**
```
adekbadek: **Suggestion:** Duplicated /preferences route + handler across both wizards — The `/preferences` route registration and the `update_preferences()` method are byte-for-byte identical here and in `class-premium-newsletters-wizard.php` (route ~139, handler ~405). Both write the same per-user meta and the same site-wide `newspack_content_gate_default_status` option. Fold the route args + handler into a shared trait or a `Content_Gate` helper (e.g. `Content_Gate::register_preferences_route( $this )` / `handle_preferences_request( $request )`) so the enum/validation and the read-back response can't drift between the two surfaces. This is the highest-value cleanup in the PR.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/preferences-modal/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+	const isDirty = checksEnabled !== presaveChecksEnabled || status !== initialStatus;
+
+	const handleSave = () => {
+		setIsSaving( true );
+		apiFetch( {
+			path: `/newspack/v1/wizard/${ slug }/preferences`,
+			method: 'POST',
+			data: { presave_checks_enabled: checksEnabled, default_gate_status: status },
+		} )
```
adekbadek, line 45: **Suggestion:** Preferences save has no error handling — `handleSave()` chains `.then().finally()` with no `.catch`. On a failed `apiFetch`, the promise rejects unhandled (a console "unhandled promise rejection"), the modal silently stays open, and the publisher gets no feedback — they clicked Save and nothing visibly happened. Add a `.catch` that surfaces an error notice (the parent owns `addNotice`; pass an `onError` or render a local error).

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/save-panel/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+			return;
+		}
+		if ( reducedMotion ) {
+			onCancel();
+			return;
+		}
+		setIsClosing( true );
+		setTimeout( onCancel, SLIDE_OUT_MS );
+	};
+
```
adekbadek, line 62: **Suggestion:** Pre-save Modal has no accessible name — The `Modal` uses `__experimentalHideHeader` with no `contentLabel`/`aria.label`, so it has no accessible name for screen readers (and WP logs a dev warning). The visible "Are you ready to save?" prompt is a plain `<strong>` span, not associated via `aria-labelledby`. Add a `contentLabel` (e.g. `__( 'Review before saving', 'newspack-plugin' )`) or wire `aria-labelledby` to the intro.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -1326,6 +1326,66 @@ public static function get_post_statuses() {
...
+	/**
+	 * Set the default status for newly created gates.
+	 *
+	 * @param string $status Either 'publish' or 'draft'.
+	 *
+	 * @return string The stored status.
+	 */
+	public static function set_default_new_gate_status( $status ) {
+		$status = in_array( $status, [ 'publish', 'draft' ], true ) ? $status : 'draft';
+		update_option( self::DEFAULT_STATUS_OPTION, $status );
```
adekbadek: **Nit:** Autoload the default-status option off — `set_default_new_gate_status()` calls `update_option()` without an explicit autoload flag, so the option is autoloaded on every request. It's read only in admin (wizard localization + the preferences endpoint), never on the frontend, so pass `false` for autoload to keep it out of the frontend `alloptions` payload.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -68,11 +72,14 @@ const getContentTypeFromRules = ( rules: GateContentRule[] ): 'all' | 'custom' |
 };
 
 const Edit = ( { match, updateGatesData, slug = AUDIENCE_CONTENT_GATES_WIZARD_SLUG, isNewsletter = false }: ContentGateEditProps ) => {
+	const [ defaultGateStatus, setDefaultGateStatus ] = useState< GateStatus >(
+		( ( window as any ).newspackAudienceContentGates?.default_gate_status as GateStatus ) || 'draft' // eslint-disable-line @typescript-eslint/no-explicit-any
```
adekbadek: **Nit:** Type the localized window keys instead of casting — The two new localized keys (`presave_checks_enabled`, `default_gate_status`) are read/written via `( window as any ).newspackAudienceContentGates` in five spots (76, 109, 267, 534-535), each with an eslint-disable. Adding these keys to the existing ambient `Window` typing would remove the casts and give real type-checking on the localized payload.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -178,49 +192,91 @@ const Edit = ( { match, updateGatesData, slug = AUDIENCE_CONTENT_GATES_WIZARD_SL
 		[ gate, contentRules, contentRulesMatch, registration, customAccess, status, title ]
```
adekbadek, line 191: **Nit:** handleCreate/handleSave omit isFetching from deps — `handleCreate` and `handleSave` reference `isFetching` in their inner guard but omit it (and `resetError`/`resetNotices`/`slug`/`history`/`addNotice`/`wizardApiFetch`/`updateGatesData`) from their `useCallback` dep arrays, so they close over a stale `isFetching`. In practice `handleSaveButtonClick` (which does depend on `isFetching`) guards first, so this is redundant-guard staleness rather than a live bug, and it matches the file's pre-existing style — either complete the deps or drop the redundant inner guard.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
```diff
@@ -178,49 +192,91 @@ const Edit = ( { match, updateGatesData, slug = AUDIENCE_CONTENT_GATES_WIZARD_SL
...
+		presaveChecksEnabled: keepChecks,
+	}: {
+		status: GateStatus;
+		presaveChecksEnabled: boolean;
+	} ) => {
+		setShowSavePanel( false );
+		if ( keepChecks !== presaveChecksEnabled ) {
+			setPresaveChecksEnabled( keepChecks );
+			( window as any ).newspackAudienceContentGates.presave_checks_enabled = keepChecks; // eslint-disable-line @typescript-eslint/no-explicit-any
+			apiFetch( {
```
adekbadek, line 268: **Nit:** Fire-and-forget swallows preference-save failure — `apiFetch(...).catch( () => {} )` fully swallows a failed persistence of the "always show pre-save checks" preference. The local state and the `window` global are updated optimistically first, so on failure the UI claims a preference that was never persisted and silently reverts on the next page load. At minimum log the failure; ideally reconcile or notify.

adekbadek (changes requested):
Some issues found

plugins/newspack-plugin/includes/content-gate/class-content-gate-api.php
thomasguillot: Fixed in beb62906b: sanitize_gate() now only includes `status` when the request provides it, so the preserve-fallback in update_gate_settings() is live again and an update that omits status keeps the stored post status. Follow-up commits 1874df766, 8bddbca59, 3d99df7ce and 15d22f3e0 extend the same sparse contract to the rest of the payload (content rules, title/priority, nested registration/custom_access, metering), each with a regression test.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
thomasguillot: Addressed in beb62906b by making the option authoritative server-side for REST creates: both wizards' create callbacks fill an omitted status via the new Content_Gate::with_default_new_gate_status(). create_gate()'s publish fallback is untouched, so the Memberships auto-gate path behaves exactly as before. This also became necessary once sanitize_gate() stopped injecting draft, since an omitted status on create would otherwise have fallen through to the publish fallback. The preference remains a single site-wide option shared by both editors, now honored consistently in the UI and over REST; the modal copy is unchanged.

plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
thomasguillot: Fixed in beb62906b: the route registration and handler now live in the shared Newspack\Wizards\Traits\Content_Gate_Preferences trait, used by both wizards, so the validation and read-back response can't drift.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/preferences-modal/index.tsx
thomasguillot: Fixed in beb62906b: a failed save now renders an inline error Notice inside the modal (server message when available) and the modal stays open for retry.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/save-panel/index.tsx
thomasguillot: Fixed in beb62906b: the Modal is labelled via aria={ { labelledby } } wired to the visible "Are you ready to save?" prompt, so the accessible name matches the visible text and the dev warning is gone.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
thomasguillot: Fixed in beb62906b: the option is written with autoload false.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
thomasguillot: Fixed in beb62906b: both keys are typed on the existing ambient Window interface (presave_checks_enabled as boolean | string to reflect wp_localize_script stringification) and all five casts with their eslint-disables are gone.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
thomasguillot: Fixed in beb62906b with a third option: the guards now read the isSaving ref, which is live across renders. Completing the dep arrays would have made both callbacks change identity every render (resetError and wizardApiFetch aren't referentially stable), churning the header-actions effect; dropping the guard entirely would have left the isNew "Edit layout" call sites with no double-submit protection.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/index.tsx
thomasguillot: Fixed in beb62906b: on failure the optimistic state and the window global roll back to the persisted value and an error notice is shown.

thomasguillot:
All nine review threads are addressed, each with a reply and commit reference. Beyond the direct fixes, an internal multi-reviewer pass on the fix commit surfaced that the status finding generalized: any omitted field in a gate update payload was destructive (rules, title/priority, nested registration/custom_access, metering). Commits 1874df766, 8bddbca59, 3d99df7ce and 15d22f3e0 make the whole sanitizer sparse so partial REST updates preserve stored values at every level, with regression tests for each (gate test class is now 69 tests). Full suite: 1604 PHP tests and 278 JS tests passing.

adekbadek (approved):
All nine threads are resolved and verified against a running env: `Content_Gates` suite is green (69 tests) with meaningful sparse-preservation regressions, and the generalization of the `status` finding to a fully sparse sanitizer is the right call. Note the frontend counterpart to this same field-preservation class of bug landed separately in #527 (dropping `gate_layout_id` from the editor payload), so #408 will want main merged in; the two are complementary layers, not overlapping.

PR #410, opened 2026-06-24
fix(n): test-js runs a standalone repo's own test script

jason10lee:
### All Submissions:

Makes `n test-js <name>` work for standalone `repos/` checkouts.

A standalone `repos/` checkout isn't part of the pnpm workspace, so a `pnpm --filter` selector matches nothing and silently runs zero tests. `n test-js <standalone>` now installs and runs the repo's own `test` script with its own package manager (detected from its lockfile). A missing `test` script or `package.json` is a clean no-op.

This branch originally also derived the workspace filter from each package's real `package.json` name. `main` shipped that independently in `f4449a806`, so the merge here takes main's version and drops the equivalent helper from this branch.

Two entrypoint problems surfaced in review. `./n` derived the target project from the working directory alone, so `n test-js <name>` discarded the name and failed with "You must be inside one of the repos" from anywhere but the project's own directory. The `test-js` case now reads the name argument the way `build`, `ci-build` and `watch` do. It invokes the container script directly with the folder quoted, following `watch` rather than `build`, because the value can now come from the command line: interpolated unquoted into `sh -c`, a name carrying a space or a `;` would word-split or be read as syntax by the container's shell.

Routing the argument also exposed a latent bug in `bin/test-js.sh`. `find_project` runs inside a command substitution, so its `exit 1` ended only that subshell. An unrecognised name left the project path empty, fell through to the workspace branch, and ran `pnpm install` plus `--filter ""` across every package, turning a typo into a full-workspace test run instead of an error. The status is propagated at the call site now.

`bin/tests/` gains a spec for each. Two small changes let those scripts run on the host: `bin/test-js.sh` sources its helpers relative to itself rather than by absolute container path, and `bin/resolve-project-path.sh` takes environment overrides for its mount paths. In the container `bin/` is mounted at `/var/scripts` and none of those variables is set, so both resolve exactly as before.

`n test-php <name>` has the same missing-argument gap and is deliberately left alone: it forwards `${@:2}` to phpunit, so it needs option-aware parsing to tell a project name from a `--filter` flag rather than the four-line fix that suits `test-js`.

1. `n test-js <name>` from outside the project directory runs that project's JS tests. This is the form the docs describe, and the one that previously failed.
2. Place a standalone checkout at `repos/plugins/<name>/` with a `package.json` declaring a `test` script; `n test-js <name>` installs its deps with the package manager its lockfile implies and runs that `test` script.
3. A standalone repo with no `test` script (or no `package.json`) is a clean no-op, not an error.
4. `n test-js <typo>` reports the unknown name and exits non-zero, installing nothing.
5. `bash bin/tests/test-test-js-project-resolution.sh` and `bash bin/tests/test-n-test-js-name-argument.sh` both pass. CI runs them in the `Shell specs` job.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — Two specs in `bin/tests/`, run by CI's `Shell specs` job. Both were checked against the unfixed code: reverting the abort makes the first fail with `pnpm --filter ""` in its output, and the second fails against `main`'s dispatcher and again against a routed-but-unquoted one.
* [x] Have you successfully run tests with your changes locally? — All three `bin/tests/` specs pass. End-to-end in the container, a standalone checkout ran its own suite (17 tests, exit 0) and an unknown name exited 1. `shellcheck -S error` is clean on every changed file; the two pre-existing `SC2145` findings in `n` are untouched.


bin/test-js.sh
```diff
@@ -18,15 +18,59 @@ find_project() {
...
+    if ! grep -q '"test"[[:space:]]*:' "$dir/package.json" 2>/dev/null; then
+        echo "No \"test\" script in $(basename "$dir")/package.json; nothing to test" >&2
+        return 0
+    fi
+    # Detect package manager from lockfile; default npm for a lockfile-less repo
+    # (matches `n build` so the two never pick a different PM for one checkout).
+    local pm="npm"
+    if [ -f "$dir/pnpm-lock.yaml" ]; then
+        pm="pnpm"
+    elif [ -f "$dir/yarn.lock" ]; then
```
dkoo, line 52: **Severity:** suggestion

Package-manager precedence here diverges from `bin/build-repos.sh`. `build_standalone_repo()` uses two sequential `&&` assignments, so when a repo contains *both* `pnpm-lock.yaml` and `yarn.lock` the yarn check wins and `n build` picks yarn. This `if/elif` picks pnpm in that same case. The inline comment above claims this "matches `n build` so the two never pick a different PM for one checkout" — but that invariant does not hold for the both-lockfiles case: `n build` would install/run with yarn while `n test-js` uses pnpm on the same checkout, so tests could resolve against a different dependency graph than the build. Real-world impact is small (two lockfiles in one repo is unusual), but since parity with the sibling is the stated goal, either mirror `build-repos.sh` exactly (sequential `&&`, yarn-wins) or soften the comment to match the actual behavior.

dkoo (commented):
This is a clean, low-risk change. It correctly fixes the real bug — filtering by directory name (`newspack-plugin`) matches no workspace package, so `pnpm --filter` silently runs no tests — by resolving the real package name from `package.json`, and it mirrors the already-merged `bin/build-repos.sh` pattern for standalone `repos/` checkouts. Exit codes propagate in both branches and interpolations are quoted. One non-blocking suggestion inline; nothing here blocks merge.

Release risk: LOW.

bin/test-js.sh
jason10lee: The invariant doesn't hold against `main` as it stands — you're right about that. It's the other half of #406, which moves `build-repos.sh` to the same pnpm-first `if`/`elif`. Once that lands the two files agree and the comment is accurate, so #406 merges first.

Leaving this thread open until it does: it's the reminder that these two land in order.

dkoo: Agreed — verified that the parity comment becomes accurate once #406 lands (both files then use the identical pnpm-first `if`/`elif`), so the #406 → #410 merge order is the right call, and keeping this thread open as that reminder makes sense.

Separately: running the fix end-to-end surfaced an unrelated entrypoint issue (`n test-js <name>` never reaches the new standalone path) — I'll file that as its own review rather than muddy this thread.

dkoo (changes requested):
Following up after running this end-to-end: **`n test-js <name>` never reaches the new standalone-repo logic — the name argument is silently ignored.**

### Repro (workspace root, this branch)
```
$ n test-js newspack-manager-admin
Error: You must be inside one of the repos to run this command.
```
`newspack-manager-admin` is checked out at `repos/plugins/newspack-manager-admin`, has a `test` script (`newspack-scripts test`) and a `package-lock.json` — so the standalone path *should* run, but the command errors before `bin/test-js.sh` is ever invoked.

### Root cause
This PR updates `bin/test-js.sh` (the container-side runner) but not `./n` (the host dispatcher that decides *which* project to target). `./n` derives `target_folder` **only from `PWD`**, and its `test-js` case just calls `require_target_folder` — unlike `build`/`ci-build`/`watch`, it never reads `$2`:

```bash
# ./n — build honors the name argument:
build)
    if [[ -n "$2" ]]; then target_folder=$2; fi
    ...
# ./n — test-js does not:
test-js)
    require_target_folder
    docker exec ... /var/scripts/test-js.sh $target_folder   # $2 ignored
```

So `n test-js <name>` from anywhere but *inside* the target directory hits `require_target_folder` and errors. The new standalone path is only reachable via `cd repos/plugins/<name> && n test-js` (no argument).

### Why this blocks
- The PR's own **"How to test" can't be run as written** — every step passes a name (`n test-js newspack-manager-admin`, `n test-js newspack-plugin`).
- It breaks the **"matches `n build`" parity** this PR and #406 repeatedly cite: `n build <name>` works by name from anywhere; `n test-js <name>` doesn't.
- Not fixed by #406 — that PR only touches `bin/build-repos.sh`; `./n`'s `test-js` case is byte-identical across `main`, #406, and this branch.

### Suggested fix — mirror `build` in `./n`
```bash
test-js)
    if [[ -n "$2" ]]; then target_folder=$2; fi
    require_target_folder
    docker exec ... /var/scripts/test-js.sh "$target_folder"
    ;;
```
(`test-php` has the same gap — worth fixing in the same pass for consistency, though it's beyond this PR's stated scope.)

The lockfile-detection change inside `test-js.sh` itself is correct; this is purely about the entrypoint not routing a `<name>` argument to it.

bin/test-js.sh
jason10lee: #406 landed on `main` on 2026-08-10 and moved `build_standalone_repo()` to the same pnpm-first `if`/`elif` — `bin/build-repos.sh:58-62` on `origin/main`. The two files agree now, including the both-lockfiles case, so the invariant the comment claims holds.

Rebased onto main in `6c3039a64`.

jason10lee:
Fixed in `d746acc88` — the `test-js` case reads `$2` as you suggested. It invokes the container script directly with the folder quoted, following `watch` rather than `build`, since the value can now come from argv.

Routing it surfaced a second bug (`3d0a9c90a`): `find_project`'s `exit 1` ended only its command-substitution subshell, so an unknown name left `PROJECT_DIR` empty and ran `pnpm install` plus `--filter ""` across all 19 packages. Now `|| exit 1` at the call site.

Both are specced in `bin/tests/` (`d0d100c9a`, `3daa73b86`) and green in `Shell specs`; each fails against the unfixed code. Two small changes made that possible — `bin/test-js.sh` sources its helpers relative to itself instead of by absolute `/var/scripts/` path, and `resolve-project-path.sh` takes environment overrides for its mount paths. That leaves `test-js.sh` as the only container-side script sourcing relatively; it resolves to the same files there.

`test-php` left alone. It forwards `${@:2}` to phpunit, so `n test-php --filter Foo` would read `--filter` as the project name — option-aware parsing rather than the four-line change here, so it seemed better as its own PR. Happy to take it if you'd rather.

Description updated. Thanks for the end-to-end run; the repro was exact.

bin/test-js.sh
```diff
@@ -18,19 +22,58 @@ find_project() {
...
+# `test` script in-place with their own package manager. A missing package.json
+# or test script is a no-op (matching pnpm --filter's silent skip for a package
+# with no test script).
+test_standalone_repo() {
+    local dir="$1"
+    if [ ! -f "$dir/package.json" ]; then
+        echo "No package.json in $dir; nothing to test" >&2
+        return 0
+    fi
+    if ! grep -q '"test"[[:space:]]*:' "$dir/package.json" 2>/dev/null; then
```
dkoo: 🟢 Non-blocking nit: this regex also matches a *dependency* named `test`. A package with `"devDependencies": { "test": "^1.0.0" }` and no `test` script would pass this check and then fail on `npm run test` with "Missing script", turning the intended no-op into a non-zero exit. The workspace branch below already reads `package.json` through node for the package name; the same form works here and removes the ambiguity:

```bash
if [ -z "$(node -p "require('$dir/package.json').scripts?.test ?? ''" 2>/dev/null)" ]; then
```

Rare in practice, so fine to leave for a follow-up.

dkoo (approved):
Re-ran the repro from my July review on `861ac29b4`, and the name argument reaches the standalone path now. From the workspace root, `n test-js <standalone>` installs and runs that checkout's own `test` script, a failing script propagates its exit status (a fixture returning `exit 7` came back as 7), and a checkout with no `test` script exits 0 with the "nothing to test" message. `n test-js definitely-not-a-project` prints the not-found message and exits 1 before any install runs, which confirms the command-substitution fix on `find_project`. For the workspace path, `n test-js newspack-plugin` from the root ran `pnpm install` and then only the newspack-plugin suite, and `n test-js newspack-theme` resolved to the theme package the same way.

On the host, all three `bin/tests/` specs pass, and `shellcheck -S error` reports only the two pre-existing `SC2145` findings in `n`. CI is green, including the `Shell specs` job. I also checked the package-manager parity thread: `bin/build-repos.sh` on `main` has the same pnpm → yarn → npm order, so the comment in `test_standalone_repo` holds now.

One non-blocking suggestion inline on the `test` script detection. Release risk: LOW. This touches dev tooling only, and the container defaults are unchanged.

bin/test-js.sh
dkoo: Confirmed on `origin/main`: `build_standalone_repo()` in `bin/build-repos.sh` uses the same pnpm → yarn → npm `if`/`elif`, so the parity comment holds for the both-lockfiles case too. Resolving.

bin/test-js.sh
jason10lee: Took the node read; it's in `83d42960a`. One change from your snippet: `node -p` exits non-zero on an unparseable `package.json` or a missing node, and stdout is empty in both cases, so a bare `[ -z … ]` can't separate "no test script" from "couldn't read it" and would print `nothing to test` and return 0. That skips a repo's whole suite while exiting clean, which is the same shape as the two bugs this PR already fixes, so an unreadable manifest returns 1 and names the file instead.

`bin/tests/test-test-js-script-detection.sh` pins both paths plus a control on a real test script. It fails 5 assertions against the pre-fix code: the dependency-named-`test` fixture reaches `npm install` / `npm run test`, and the truncated manifest exits 0.

PR #411, opened 2026-06-24
fix(env): provision plugin vendor/ on env up to avoid activation fatal

jason10lee:
### All Submissions:

A fresh isolated environment (`n env up`) links the monorepo plugins but never installs their Composer `vendor/` — so activating a plugin that requires its autoloader (e.g. `newspack-plugin`, `newspack-popups`) fatals on a missing `vendor/autoload.php`. This adds a `bin/ensure-vendor.sh` step at env-up that runs a **runtime-only** (`--no-dev`) per-plugin Composer install — all plugin activation needs, without dev dependencies or the slow JS build — so a fresh env is never vendorless. (Dev dependencies for running a plugin's tests still come from `n build` / `n ci-build`.)

* **Idempotent.** Skips any plugin whose `vendor/autoload.php` is already present, so warm starts stay instant.
* **Non-fatal.** If a Composer install fails, it warns actionably and leaves the env usable rather than tearing down an otherwise-working environment.

Only monorepo plugins are provisioned; standalone `repos/` checkouts manage their own dependencies.

1. `n env up <name>` on a fresh environment; confirm the `[ensure-vendor]` step installs Composer deps for the monorepo plugins.
2. Activate a plugin that requires its autoloader (e.g. `newspack-plugin`) — it activates without a fatal about a missing `vendor/autoload.php`.
3. Re-run `n env up <name>` — confirm `ensure-vendor` skips already-provisioned plugins (no redundant installs).
4. Force a Composer failure for one plugin — confirm env-up reports the failure (including how many other plugins were provisioned) but still completes and the env stays usable.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — no `bin/` test harness in this repo.
* [x] Have you successfully run tests with your changes locally? — verified in an isolated environment: `ensure-vendor` ran at env-up across the monorepo plugins, with error propagation + graceful degradation confirmed. No CI coverage for `bin/` here, so verification is manual.


bin/env.sh
```diff
@@ -470,6 +470,13 @@ MIGRATE
             docker exec "$container_name" wp --allow-root rewrite flush --hard >/dev/null 2>&1
             docker exec "$container_name" chown "$run_user":"$run_user" /var/www/html/.htaccess 2>/dev/null || true
         fi
+        # Provision Composer vendor/ for the migrated monorepo plugins, so a
+        # later `n setup` / plugin activation doesn't fatal on a missing
+        # vendor/autoload.php (the foundation-smoke failure mode). Idempotent;
+        # skips plugins whose vendor/ is already present. On failure it warns
+        # (actionably) rather than tearing down an otherwise-usable env.
+        docker exec "$container_name" bash /var/scripts/ensure-vendor.sh || \
```
dkoo, line 478: **Severity:** suggestion

This provisioning runs *after* several WP-CLI commands above (`wp option get`, `wp core install`, `wp rewrite flush`) that bootstrap WordPress and load active plugins. On a fresh env this is fine — only `newspack-theme` is active at this point and the Newspack plugins aren't activated until `n setup`. But for an already-provisioned env whose DB has a Newspack plugin active while its mounted `vendor/` is absent (e.g. a worktree whose vendor was never built or was removed), those earlier commands can fatal on the missing `vendor/autoload.php` before provisioning runs, leaving the env only partially repaired. It self-heals — the earlier failures are non-fatal here so this block still runs, and a second `n env up` comes up clean — but moving this provisioning before the first plugin-loading WP-CLI command would repair such envs in a single pass. Non-blocking.

bin/ensure-vendor.sh
```diff
@@ -0,0 +1,47 @@
...
+installed=0
+
+while IFS= read -r dir; do
+	[ -d "$dir" ] || continue
+	[ -f "$dir/composer.json" ] || continue
+	name=$(basename "$dir")
+	# Already provisioned — nothing to do (keeps warm starts instant).
+	[ -f "$dir/vendor/autoload.php" ] && continue
+	echo "[ensure-vendor] installing composer deps for $name"
+	if composer install --working-dir "$dir" --no-interaction --no-dev; then
```
dkoo, line 29: **Severity:** suggestion

Non-worktree plugins are bind-mounted from the single shared host tree (`./plugins`), so every env writes into the same `plugins/<name>/vendor/`. `n env up --all` is sequential so it's safe, but two concurrent `n env up <name>` invocations (a documented parallel-session habit in this workspace) could both observe a missing `vendor/autoload.php` on a cold start and both launch `composer install` into the same directory, which can corrupt `vendor/`. The idempotency guard closes the window after the first success, so this only bites on genuinely-concurrent cold starts. A per-dir `flock` would eliminate it; at minimum a comment acknowledging the shared-tree write would help.

bin/ensure-vendor.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+#
+# Ensure migrated monorepo plugins have their Composer vendor/ installed.
```
dkoo, line 3: **Severity:** nit

`get_all_project_dirs` iterates both `/newspack-plugins` and `/newspack-themes`, and `newspack-theme` / `newspack-block-theme` both ship a `composer.json`, so this also provisions themes. The header and the user-facing messages ("migrated monorepo plugins", "those plugins will fatal on activation") read as plugin-only, so a theme failure would be misreported as a plugin. Behaviorally correct — just consider broadening the wording to "plugins/themes".

dkoo (commented):
**Release risk: LOW** — no blockers; a small, well-guarded env-provisioning change.

This is a clean fix for the vendorless-fresh-env activation fatal. The shell is careful: every expansion is quoted, the loop is idempotent (skips dirs that already have `vendor/autoload.php`, so warm `n env up` stays effectively instant), per-plugin failures are accumulated rather than aborting mid-loop, and non-fatality is enforced at the call site so a composer failure warns instead of tearing down an otherwise-usable env. `--no-dev` is the right scope — plugin activation needs only the runtime autoloader (`require`, not `require-dev`), and test deps are correctly deferred to `n build`/`n ci-build`. Scope is correctly monorepo-only; standalone `repos/` checkouts are untouched.

A few non-blocking suggestions inline: the provisioning runs after WP-CLI commands that can load active plugins (an ordering edge case for already-active-but-vendorless envs), a possible race between concurrent cold `n env up`s writing the shared `vendor/`, and the plugin-vs-theme wording. None block merge.

bin/env.sh
jason10lee: Moved ahead of the WP-CLI calls in `a045c0d33`. It needs no database, so nothing is lost by running it first, and the partially-repaired case you describe now heals in one pass instead of two.

Confirmed the ordering you identified: the earliest plugin-loading call is the `wp db query` probe inside the wait loop, so provisioning now sits above the loop entirely rather than just above `wp core is-installed`.

bin/ensure-vendor.sh
jason10lee: Took the `flock` rather than the comment — `flock` is in the image (util-linux 2.37.2), so the window can be closed rather than documented. The second waiter re-checks `vendor/autoload.php` inside the lock and skips.

Verified with two concurrent workers against one directory: one installs, the other blocks and skips inside the lock.

bin/ensure-vendor.sh
jason10lee: Broadened, and you're right that it was misleading — `get_all_project_dirs()` iterates `$PLUGINS_PATH` *and* `$THEMES_PATH` (`resolve-project-path.sh:40-49`), and both themes ship a `composer.json`. Header and all four user-facing messages now say plugins/themes.

Also made the script executable while here; it was committed `100644`.

PR #155, opened 2026-05-29
feat(emails-rr): card expiry warning email (NPPD-1524)

kmwilkerson:
## Summary

Adds a Newspack-managed **Card expiry warning** email — sent ~14 days before a saved payment method expires on an active WooCommerce subscription, prompting the reader to update their card before renewal fails. Lives in the Reader revenue chip; requires WooCommerce Subscriptions.

Resolves [NPPD-1524](https://linear.app/a8c/issue/NPPD-1524).

image

## Stacking

This was originally stacked on slice 2b (#146); now that the slice stack has merged, it is **rebased onto `main`** and contains only its own card-expiry delta.

## What's in the slice

**Backend:**
- `Card_Expiry_Warning` class registers a new Newspack-managed Pattern A email via `newspack_email_configs` with `chip='reader-revenue'`, `recipient='reader'`, `recommended=true`, and an explicit `trigger_description`.
- Daily `wp_schedule_event` cron scans active subscriptions for saved payment methods within N days of expiry (default 14, filterable via `newspack_card_expiry_warning_days`).
- **Two-prefix idempotency schema**, both keyed on `token_id:expiry_month/year`:
  - `SEEDED_META` — written by the first-deploy seed pass (see below) *without* sending. Blocks a send on the normal path, but is **bypassable** by the explicit CLI backfill.
  - `SENT_META` — written after a successful `Emails::send_email()`. **Always** blocks (even under the CLI bypass), so a warning is never sent twice for the same card cycle.
  - A reader who updates their card clears the marker via the `woocommerce_subscription_payment_method_updated` hook, so the new card's eventual expiry re-triggers normally.
- Block-template email at `includes/templates/reader-revenue-emails/card-expiry-warning.php`, using the standard Newspack substitution tokens (`*BILLING_FIRST_NAME*`, `*CARD_LAST_4*`, `*EXPIRY_DATE*`, `*RENEWAL_DATE*`, `*UPDATE_PAYMENT_URL*`, etc.).
- Group subscriptions: warning recipient resolves to the customer on the subscription (the group sub owner for institutional access).

**Discoverability tweak:** small one-line change to slice 2b's `Email_Preview::get_sample_substitutions()` — `*RENEWAL_DATE*` sample shifts from `+1 year` to `+30 days` for a more realistic card-expiry preview context. No 2b behavior change.

## Publisher-respect machinery

[...]

PR #412, opened 2026-06-24
feat(emails): two-phase idempotent transactional send (NPPD-1768)

kmwilkerson:
## Summary

Implements the durable **two-phase idempotent send** tracked as a follow-up from the card-expiry warning ([NPPD-1524](https://linear.app/a8c/issue/NPPD-1524) / #155). Resolves [NPPD-1768](https://linear.app/a8c/issue/NPPD-1768).

The card-expiry warning shipped with a single-phase "send → mark SENT (with bounded retry)". That leaves one window open: if the process dies after the mail is accepted but before the SENT marker persists, a later cron pass re-sends. The bounded retry shrinks that window but can't close it, and it offers no protection against two passes (cron + a manual CLI backfill) picking the same token at once.

This PR closes both with a reusable two-phase claim — **claim → send → confirm** — and wires card-expiry to it.

## Two product/design decisions 

1. **Stale-claim policy → re-send (over-send posture).** When a pass finds a claim left behind by a process that died mid-send, it cannot know whether the mail went out. It **re-sends** and logs it: for these warnings a rare duplicate is far less harmful than a silent miss (no warning → failed renewal → involuntary churn). Net effect vs. today: strictly better — it adds a concurrency lock while keeping the same (now rarer) hard-death duplicate risk.
2. **Build it reusable, not card-expiry-only.** The same send-then-mark shape exists in the commerce emails (`WooCommerce_Connection`: receipt / welcome / cancellation), where user-triggered concurrency and retries are more common than a daily cron. So the logic lives in a standalone helper those flows can adopt next.

## The helper — `includes/emails/class-idempotent-send.php`

`Idempotent_Send::send( $entity, [...] )` runs a claim → send → confirm state machine on the entity's own metadata:

1. **SENT** marker matches the idempotency value → already delivered, skip.
2. **PENDING** claim matches:
   - **recent** (within the grace window) → another pass is mid-send; skip — a **best-effort guard** (the claim is a non-atomic read-then-write, so it narrows but does not fully eliminate a simultaneous double-send; see below).
   - **stale** (older than grace) → re-send per the over-send policy, logged.
3. otherwise → write the PENDING claim **durably** (never send without a durable claim) → send → promote the claim to SENT durably.

Notes:
- **`WC_Data`-agnostic.** The entity is duck-typed on `get_meta` / `update_meta_data` / `delete_meta_data` / `save`, so the helper works for a `WC_Subscription` *or* a `WC_Order` and carries no hard WooCommerce dependency.
- **Filterable grace window** (`newspack_idempotent_send_grace_seconds`, default 1h).
- **Value-match semantics preserved** — a marker carrying a *different* value (e.g. a card's prior expiry tuple) doesn't block a new value, so the next cycle correctly sends.
- **`clear_on_send`** — companion keys (card-expiry's SEEDED mark) cleared in the same confirm save.
- Bounded `save_with_retry` is now the single source of truth for the retry behavior.

## Card-expiry wiring

`Card_Expiry_Warning::maybe_send_warning()` now delegates the send/mark to the helper (net **−104 / +82** lines; the bespoke `save_subscription_with_retry` is removed). The SEEDED gate + CLI `$bypass_idempotency` stay card-expiry-specific — bypass only relaxes the SEEDED pre-check; the SENT gate stays authoritative inside the helper. `clear_sent_flag` (card-method-update hook) now also wipes an interrupted PENDING claim.

## Testing

- **New helper suite** (`tests/unit-tests/idempotent-send.php`, 17 tests): claim-written-before-send, SENT blocks without sending, recent-claim skips, stale-claim re-sends, send-failure releases the claim, **claim survives a failed confirm save** (the durability invariant — verified via a stage-then-commit fake that models WC's save() as the commit point), failed-release leaves the claim (logged), claim-can't-persist → no send, filterable grace, value-change not blocked, non-array claim handled, multi-key `clear_on_send`, the `is_claimed()` predicate, and retry-budget succeed/give-up.
- **Card-expiry suite** updated (28 green): `clear_sent_flag` now asserts the PENDING claim is cleared, the gate asserts a recent claim blocks / a stale claim re-sends, and the retry tests point at the shared helper.
- Regression: full `WooCommerce_Subscriptions_Integration` group (44) green.
- **Real-WC smoke scenario** (`tests/integration/card-expiry-warning-smoke.php`, scenario 12): drives the two-phase path against a real `WC_Subscription` — the array PENDING marker round-trips through WC `save()`/`get_meta()`, a recent claim blocks a resend, a stale claim re-sends + promotes to SENT. This script isn't in CI (see NPPD-1795), but was run manually and **passes**.
- Lint clean against the plugin and the stricter workspace-root standard.

## Follow-on

The commerce emails can adopt `Idempotent_Send` next — the reusable shape is the bigger payoff. The two-phase path is now exercised against a real subscription by the smoke scenario above (manual); folding that realistic exercise into **CI** is tracked separately in [NPPD-1795](https://linear.app/a8c/issue/NPPD-1795).

## Concurrency (honest scope)

The PENDING claim is a **best-effort guard, not a mutual-exclusion lock** — a non-atomic read-then-write. It reliably catches *overlapping* passes (a daily cron still running when an operator starts a CLI backfill) but two *perfectly simultaneous* passes can still both send. This is **strictly better than before** (the old path had zero concurrency protection); a hard guarantee would need an atomic primitive (DB unique index / `GET_LOCK` / `wp_cache_add`). For a daily cron + occasional backfill, the residual is a rare duplicate — the accepted failure direction.

## Review hardening 

Addressed in the second commit: unified the skip decision behind a shared `Idempotent_Send::is_claimed()` predicate (so the cron scan, the CLI backfill count, and `send()` agree, and the gate skips a recent claim before building a recipient); the failed-send claim-release now checks + logs its save result; the grace filter is computed lazily (after the SENT short-circuit); marked the helper `final`; and the test fake now models WC's stage-then-commit so the **durability** invariant (a claim survives a failed confirm save) is actually verified — plus tests for the failed-release, non-array claim, multi-key clear, the `is_claimed` predicate, and a `tear_down` for the grace filter.


plugins/newspack-plugin/includes/emails/class-idempotent-send.php
```diff
@@ -0,0 +1,294 @@
...
+	 * @param int    $attempts Max attempts (>= 1).
+	 * @return array{saved: bool, last_error: string} Whether the save landed,
+	 *               and the last error message if it did not.
+	 */
+	public static function save_with_retry( $entity, int $attempts = self::DEFAULT_SAVE_ATTEMPTS ): array {
+		$attempts   = max( 1, $attempts );
+		$last_error = '';
+		for ( $attempt = 1; $attempt <= $attempts; $attempt++ ) {
+			try {
+				$entity->save();
```
rbcorrales, line 400: This treats any non-throwing `save()` as durable, but the current caller is a `WC_Subscription`. `WC_Order::save()` catches `Exception` in `WC_Abstract_Order::save()` and still returns the object ID, and `WC_Data::save_meta_data()` writes meta rows without propagating add/update failures back to this helper. That means we can send without a persisted `PENDING` claim, or return after confirm while `SENT` never landed.

Can this verify the expected marker after save with a fresh persisted read, or use a checked persistence primitive for the claim and confirm markers? The tests should also model the real `WC_Order`/`WC_Subscription` save semantics, not only a fake that throws.

plugins/newspack-plugin/includes/emails/class-idempotent-send.php
```diff
@@ -0,0 +1,294 @@
...
+			Logger::log(
+				sprintf( 'Could not persist the pending claim for "%s"; skipping the send this pass.', $idem ),
+				$header,
+				'error'
+			);
+			return false;
+		}
+
+		// 2. Send.
+		$sent = (bool) call_user_func( $args['send'] );
```
rbcorrales: If the callback throws, this leaves the durable `PENDING` claim in place because only the false-return branch releases it. `wp_mail` filters/actions can throw through `Emails::send_email()`, and cron/CLI already catch `Throwable` per pair, so this can suppress retries until grace and later log the next attempt as stale even when the first attempt never sent.

Could this catch `Throwable` around the callback, best-effort release the claim with the same retry/logging path, then rethrow so existing cron/CLI failure handling still sees the original error?

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-card-expiry-warning.php
```diff
@@ -694,92 +719,38 @@ public static function maybe_send_warning( $subscription, $token, bool $bypass_i
...
+		// Two-phase idempotent send (NPPD-1768): claim a PENDING marker
+		// durably, send, then promote the claim to SENT. A process death
+		// between send and promote leaves a claim that a later pass
+		// reconciles — recent claims are skipped as a concurrency lock,
+		// stale claims re-send (over-send beats a missed expiry warning).
+		// On a confirmed send the helper also clears the SEEDED mark so the
+		// "at most one of {SEEDED, SENT}" invariant holds. The SENT gate is
+		// re-checked inside the helper, so it stays authoritative even under
+		// `$bypass_idempotency` (which only relaxes the SEEDED pre-check).
+		return Idempotent_Send::send(
```
rbcorrales, line 741: This uses the subscription object that was loaded during discovery. In the CLI path, pairs are loaded and filtered before the confirmation prompt, then sent later. In cron, all pairs are loaded before the loop. If another process writes `PENDING` or `SENT` after discovery, this object can miss the marker and the helper's first `is_claimed()` check uses stale in-memory meta.

Can we reload or force-read the subscription immediately before entering `Idempotent_Send::send()`?

plugins/newspack-plugin/tests/integration/card-expiry-warning-smoke.php
```diff
@@ -722,10 +722,90 @@ function () {
...
+
+$reloaded = wcs_get_subscription( $sub_id );
+$stored   = $reloaded->get_meta( $pending_key, true );
+if ( ! is_array( $stored ) || ( $stored['value'] ?? null ) !== $expiry_key ) {
+	smoke_fail( 'PENDING claim did not round-trip as an array through WC meta.', 'got: ' . var_export( $stored, true ) );
+} else {
+	$mails = [];
+	$sent  = Card_Expiry_Warning::maybe_send_warning( $reloaded, $token2 );
+	if ( false === $sent && 0 === count( $mails ) ) {
+		smoke_pass( 'Recent PENDING claim round-tripped as an array and blocked a resend (0 emails).' );
```
rbcorrales, line 765: This new pass increments `$passed`, but `$total` at the top of the script is still `12`. With the two new scenario 12 pass paths plus cleanup, a fully passing smoke run reaches 14 and exits nonzero at the final equality check.

Please update the expected total/header comments, then rerun the manual smoke script so the real-WC proof can pass cleanly.

plugins/newspack-plugin/includes/emails/class-idempotent-send.php
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		// 3. Confirm: promote PENDING → SENT (and clear companion keys).
+		$entity->delete_meta_data( $pending_key );
+		$entity->update_meta_data( $args['sent_key'], $idem );
+		foreach ( $clear_on_send as $key ) {
+			$entity->delete_meta_data( $key );
+		}
+		if ( ! self::save_with_retry( $entity, $save_attempts )['saved'] ) {
+			Logger::log(
```
rbcorrales: These degraded-state logs are the only signal for rare failure modes, but `Logger::log` is gated by `NEWSPACK_LOG_LEVEL` and is often off in production. The `save_with_retry()` `last_error` is also discarded at the call sites, so support/Manager would not get the subscription/order context or the failure reason when the helper takes a degraded path.

Can these operator-actionable paths also emit Manager-visible Newspack logging and include caller-provided context plus `last_error`?

rbcorrales (changes requested):
I think this needs a few changes before merge.

The main blocker is that the helper still does not prove the `PENDING` and `SENT` markers persisted under real `WC_Order` / `WC_Subscription` save semantics. Since `WC_Order::save()` can catch exceptions and still return the object ID, the current `save_with_retry()` success path can treat a marker as durable when it may not have landed.

I also left inline notes for the throwing-send path leaving a fresh `PENDING` claim behind, the stale subscription object window before the claim/send attempt, the smoke script counter, and production-visible logging for degraded states.

Once the marker persistence check and throwing-callback cleanup are addressed, I think the rest is mostly hardening and follow-up work.

plugins/newspack-plugin/includes/emails/class-idempotent-send.php
kmwilkerson: Added a `read_fresh` verifier: after each claim and confirm save, the helper re-reads the marker from storage (card-expiry passes a closure using `WC_Data::read_meta_data( true )`, a forced data-store read) and treats an unpersisted marker as a failure — so a swallowed `WC_Abstract_Order::save()` no longer counts as durable (claim → no send; confirm → degraded + claim survives). Tests now model the swallow case (`save()` returns success but persists nothing), not only a throwing save.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/includes/emails/class-idempotent-send.php
kmwilkerson: The send callback is now wrapped in try/catch: on `Throwable` the helper releases the pending claim (best-effort save) and **rethrows**, so cron/CLI's existing per-pair `Throwable` handling still sees the original error and no orphan claim is left to suppress retries. Added `test_throwing_send_releases_claim_and_rethrows`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-card-expiry-warning.php
kmwilkerson: `maybe_send_warning()` now reloads the subscription with a forced-fresh meta read (`reload_fresh()` → `read_meta_data( true )`) before the gate and the send, so a `PENDING`/`SENT` marker written by a concurrent pass after discovery is visible to both `is_claimed()` and the helper.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/tests/integration/card-expiry-warning-smoke.php
kmwilkerson: Fixed — `$total` is now 14 (11 single-check scenarios + 2 checks in scenario 12 + cleanup), so a fully passing run exits 0. Reran the script: scenario 12 (the two-phase real-WC proof) passes. The 4 remaining failures are a pre-existing container quirk (payment-token expiry reads back as `00/`); I reproduced the identical failures on `main`'s smoke script, so they're unrelated to this PR.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/includes/emails/class-idempotent-send.php
kmwilkerson: The degraded paths (claim not durable, release didn't land, confirm not durable) now emit `Logger::newspack_log` — which fires the `newspack_log` action consumed by Manager regardless of `NEWSPACK_LOG_LEVEL` — including the caller-provided `context` (the subscription id) and the `save_with_retry` `last_error`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

rbcorrales (changes requested):
The five points all look addressed. One small follow-up:

In `Idempotent_Send::send()`, the thrown-send catch releases the claim with a bare `save_with_retry()`, while the false-return branch verifies the release with `persisted( ..., '' )` and emits `log_degraded( 'newspack_idempotent_send_release_failed', ... )` when it does not land.

So if the send throws and WC swallows the release write, a recent `PENDING` claim can still survive undetected and unlogged to Manager.

Could the catch use the same verified-release plus degraded-log path, then rethrow? The only thing to watch is that verifier/logging failures should not mask the original send error.

kmwilkerson:
Good catch — fixed in 3da90d77d9. The thrown-send `catch` now uses the **same verified-release path** as the false-return branch:

- `delete_meta_data( $pending_key )` → `persisted( …, $pending_key, '' )` (the fresh-read verify) → `log_degraded( 'newspack_idempotent_send_release_failed', … )` when the release doesn't land. So a swallowed release after a throw is now detected and surfaced to Manager, not silent.
- The release+verify+log is wrapped in an inner `try { … } catch ( \Throwable $cleanup_error )` that's intentionally swallowed, and the original `$send_error` is **always rethrown** afterward — so a verifier or logging failure can never mask the original send error.

New test `test_throwing_send_with_swallowed_release_logs_and_rethrows` pins exactly that: the send throws **and** WC swallows the release write → the claim survives, the Manager `newspack_log` fires with `newspack_idempotent_send_release_failed`, and the original `RuntimeException` still propagates. Helper suite 21 green, PHPCS clean against the root standard.

rbcorrales (approved):
Verified the new commit. The thrown-send catch now uses the same verified release as the false-return branch: it deletes `PENDING`, re-reads with `persisted( ..., '' )`, and emits the Manager-visible `newspack_idempotent_send_release_failed` log when the release does not land, all wrapped so a verifier or logging failure cannot mask the original send error, which is still rethrown.

`test_throwing_send_with_swallowed_release_logs_and_rethrows` exercises the swallowed-release case and asserts the original error propagates, the claim is detected as surviving, and the degraded log fires. This closes the last caveat from my review. Approving.

Thanks!

PR #415, opened 2026-06-24
fix(group-subscription): include owner in member counts

dkoo:
## Description

Fixes [NPPM-2949](https://linear.app/a8c/issue/NPPM-2949/june-29-alpha-testing-bug-group-counts-are-off-on-cards) — group member counts excluded the owner/manager, so cards read "0 of X members" when the owner was the only member, and disagreed with the group's Members tab.

**Root cause:** `Group_Subscription::get_members()` returns only users carrying the `_newspack_group_subscription` meta — it excludes the owner (the subscription customer, returned by `get_managers()`). Surfaces that did `count( get_members() )` were off by one. The Members tab was already correct because it merges managers + members.

This is an **alpha-hotfix** (based off `alpha`).

## Changes

- **New additive helpers** on `Group_Subscription` (backward compatible — this contract is consumed by `newspack-manager`/`newspack-blocks`):
  - `get_all_members()` — deduped managers ∪ members
  - `get_member_count()` — owner-inclusive member count
  - `get_member_capacity()` — `limit + managers`, or `null` when unlimited
- **Group picker cards** — owner-inclusive count + capacity.
- **Admin subscription list-table column** — owner-inclusive count + capacity.
- **Admin meta box** — header count includes the owner; the owner is now rendered as a non-removable "(owner)" row so the header matches the list (the admin JS tallies `<li>` count, so it stays in sync after add/remove/invite).
- **Members tab badge** — unchanged (already owner-inclusive).
- **Setting label** — "Group subscription member limit" → **"Group subscription member limit (in addition to owner)"**, with a matching description. Limit enforcement is unchanged: the limit genuinely is the number of members allowed *in addition to* the owner.

### Display semantics

The configured limit is "members in addition to the owner", so the total capacity shown in "X of Y members" is `limit + owner`:

| State (limit = 10) | Card / column |
|---|---|
| Owner only | `1 of 11 members` |
| Owner + 1 member | `2 of 11 members` |
| Full | `11 of 11 members` |

## Testing

Automated:
- New `tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php` (6 tests covering owner inclusion, dedup, and capacity).
- Full `WooCommerce_Subscriptions_Integration` group: **44/44 pass**. PHPCS clean.

Manual:
1. As a user owning multiple group subscriptions, visit `/my-account/group/`. With only the owner, a card reads **"1 of N members"** (not "0 of N").
2. Open a group → the Members tab badge and the card agree.
3. Add a member → card increments and still matches the Members tab.
4. In wp-admin → Subscriptions, the list-table column and the group meta box both count the owner; the meta box shows the owner as a non-removable "(owner)" row.
5. Edit a subscription product → the member-limit field reads "… (in addition to owner)".

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
@@ -360,10 +360,26 @@ class Group_Subscription_Settings {
 		$product  = \wc_get_product( WooCommerce_Subscriptions::get_subscription_product_id( $subscription ) );
 		$members  = Group_Subscription::get_members( $subscription );
 		$managers = Group_Subscription::get_managers( $subscription );
-		// The owner/manager(s) are rendered (non-removable) at the top of the list below,
-		// so exclude any manager that also carries member meta to avoid a duplicate row.
-		$member_only = array_values( array_diff( array_map( 'intval', $members ), array_map( 'intval', $managers ) ) );
 		$invites  = Group_Subscription_Invite::get_invites( $subscription );
+		// Resolve the rows once, applying the same guards used when rendering below, so the
+		// header count always matches the rendered list (and the admin JS, which re-tallies the
+		// list items on add/remove/invite). The owner/manager(s) render as non-removable rows;
+		// members exclude any manager that also carries member meta (avoiding a duplicate row)
+		// and must be readers.
+		$manager_users = [];
+		foreach ( array_map( 'intval', $managers ) as $manager_id ) {
+			$manager_user = get_user_by( 'id', $manager_id );
+			if ( $manager_user ) {
+				$manager_users[] = $manager_user;
+			}
+		}
+		$member_users = [];
+		foreach ( array_diff( array_map( 'intval', $members ), array_map( 'intval', $managers ) ) as $member_id ) {
+			$member_user = get_user_by( 'id', $member_id );
+			if ( $member_user && Reader_Activation::is_user_reader( $member_user ) ) {
+				$member_users[] = $member_user;
+			}
+		}
 		?>
 		<div class="newspack-group-subscription__container" data-subscription-id="<?php echo \esc_attr( $subscription->get_id() ); ?>">
 			<div class="newspack-group-subscription__settings">
@@ -427,8 +443,9 @@ class Group_Subscription_Settings {
 						sprintf(
 							// translators: %d: The number of group members.
 							__( 'Group members (<span class="newspack-group-subscription__members-count">%d</span>)', 'newspack-plugin' ),
-							// The owner counts as a member, so include managers alongside members and pending invites.
-							Group_Subscription::get_member_count( $subscription ) + count( array_values( $invites ) )
+							// Count exactly the rows rendered below (owner(s) + reader-members + invites)
+							// so the header never drifts from the list.
+							count( $manager_users ) + count( $member_users ) + count( $invites )
 						)
 					);
 					?>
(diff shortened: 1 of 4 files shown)


thomasguillot:
Splitting the other June 29 group-subscription alpha bug out separately: NPPM-2947 (manual subscription not inheriting product group settings) is now handled in #419. This PR stays focused on NPPM-2949, with no overlap between the two.

thomasguillot:
## Code review

**Verdict: solid, ship-ready.** Bug is correctly fixed, the contract change is backward-compatible, and the edge cases are genuinely covered by tests.

### What's right
- **Root cause is correct.** `get_members()` only returns users carrying `_newspack_group_subscription` meta, which excludes the owner (the subscription customer), so `count( get_members() )` was off by one everywhere it fed a display. The three new additive helpers (`get_all_members`, `get_member_count`, `get_member_capacity`) are the right shape and don't disturb the existing contract consumed by `newspack-manager`/`newspack-blocks`.
- **Numerator/denominator stay consistent.** `get_member_count` = `unique( filter( managers ∪ members ) )` and `get_member_capacity` = `limit + count( filter( managers ) )`. Both apply the same `array_filter` to managers, so the ownerless `customer_id = 0` edge (`get_managers()` → `[0]`) collapses to "0 of 10" on both sides rather than a phantom "0 of 11". Dedup also handles an owner who additionally carries member meta. Both are explicitly tested.
- **Display vs. enforcement semantics line up.** The limit stays owner-exclusive for enforcement (`class-group-subscription-invite.php` still checks `count( get_members() ) >= limit`), while display becomes owner-inclusive over `limit + owner`. "Full" renders as `11 of 11` exactly when enforcement blocks the next add. The relabel to "(in addition to owner)" makes that read as intentional rather than a fudge.
- **Coverage is complete.** Every count-display surface is accounted for: group picker card, list-table column, and meta box header are updated; the Members tab (`group-subscription-members.php`) was already owner-inclusive; `group.php`'s `$members` only feeds an `is_completely_empty` check, not a count.

### Minor notes (non-blocking)
1. **Meta box header vs. rendered `<li>` count can drift for non-reader members.** The header uses `get_member_count()` (no reader filter), but rendered member rows still apply `! is_user_reader( $user )` (and managers render with only a `get_user_by` existence check). If a member isn't a reader or the user record is gone, the header counts them but no `<li>` renders, and since `admin.js` re-tallies `<li>` on add/remove/invite, the number would jump on the first mutation. This is **pre-existing** (old `count( $members )` had the same gap vs. the filtered render), so not a regression, but this is a natural moment to align the two if you want the header rock-solid on load.
2. **`get_managers()` runs twice per card** (via `get_member_count` → `get_all_members`, and again in `get_member_capacity`), and `get_member_capacity` re-fetches settings the picker already loaded into `$settings`. Cheap (no extra DB query in the manager path), so fine to leave, just flagging the redundancy.
3. Trivial: in `group-picker.php`, `$settings` is still fetched but `limit` is no longer read from it there. Confirm it's still used for the name (it is) so the assignment isn't dead.

thomasguillot (approved):
Approving with some minor notes — full review in https://github.com/Automattic/newspack-workspace/pull/415#issuecomment-4802341856

dkoo:
Thanks for the thorough review! Dispositions on the three notes:

**1. Meta-box header vs. rendered `<li>` drift — fixed** in 70cd18811.

The meta box now resolves the manager/member rows into arrays **once**, applying the same guards used for rendering (manager: `get_user_by` existence; member: existence + `is_user_reader`), and the header counts those exact arrays:

```php
count( $manager_users ) + count( $member_users ) + count( $invites )
```

The loops then render the same arrays, so the header and the list are derived from a single source and can't diverge — and `admin.js`'s `<li>` re-tally stays aligned on add/remove/invite. Verified against a real non-reader (editor-role) member: the old `get_member_count()` header would read **3** while only **2** rows render; the new header reads **2 = 2 rendered rows**.

**2. `get_managers()` twice + re-fetch settings per card — leaving as-is.** Agreed it's cheap (the manager path is query-free). The only way to dedupe it is to pass `$settings`/manager IDs into `get_member_capacity()`, which would couple a public helper consumed by `newspack-manager`/`newspack-blocks` to caller-side state — a worse API trade for a sub-millisecond, zero-DB win.

**3. `$settings` in `group-picker.php` — confirmed still used** for the card name (`$settings['name']`), so not dead. 👍

PR #418, opened 2026-06-25
fix(newsletters): ad publish meta error, invalid dates, advertiser counts

thomasguillot:
## Summary

Fixes issues found in June 29 alpha testing of the Newsletters admin UX modernization (#2163, #4735). All changes are in `newspack-newsletters` (Newsletter Ads / Advertisers admin).

### NPPM-2950 — "not allowed to edit tracking_impressions" + "Invalid Date"

- A new ad had no `tracking_impressions` / `tracking_clicks` meta row, so the block editor's round-trip of the registered default `0` was rejected by `auth_callback => __return_false` on publish. We seed those rows in `ad_default_fields`, so the round-trip is an unchanged no-op and the read-only contract is preserved.
- The legacy ad editor stored `start_date` / `expiry_date` as ISO datetime, which rendered as "Invalid Date". A `sanitize_callback` normalizes the meta to `Y-m-d` on save (strictly ISO + `checkdate`), and the status helper / list / Quick Edit JS tolerate legacy datetime values.

### NPPM-2945 — advertiser ad counts wrong

WP's term count was publish-only while the Ads list shows draft / scheduled / pending ads, so an advertiser whose only ad was non-published showed `0`. The advertiser taxonomy also spans the newsletters CPT, so newsletters tagged with an advertiser inflated the count. Both are fixed with a custom `update_count_callback` that counts only ads (`Ads::CPT`) across the relevant statuses, plus a one-time recount (on `admin_init`) for existing terms.

### NPPM-2946 — layouts "won't open"

Working as designed: prebuilt layouts are read-only (Duplicate only); duplicate or create a layout first, then it opens in the editor. No code change.

## Hardening (from multi-reviewer pass)

- Ad lifecycle "today" anchored to the site timezone (`current_time`) across the status helper, the SQL status filter, and `is_ad_active`.
- SQL status buckets normalize date meta (`STR_TO_DATE(LEFT(meta_value,10), '%Y-%m-%d') + INTERVAL 0 DAY`) so the list filter agrees with the status badge for legacy datetime and rejects malformed dates exactly as PHP does.
- A REST guard rejects invalid `start_date` / `expiry_date` writes with `400` and leaves existing meta unchanged, instead of silently clearing the boundary.
- The legacy ad editor emits and compares site-local `Y-m-d` (normalized on read), fixing UTC day-shift and same-day expiry selection.

## Testing

- `n test-php` ads / advertisers / tracking suites pass; new tests cover the meta seeding, date normalization, legacy-datetime filter agreement, invalid-date rejection, the REST guard, and ads-only counts.
- Verified end-to-end against a real REST publish: `200` with the seed in place, the exact `403` without it; `start_date` persists as `Y-m-d`.

Fixes NPPM-2950, NPPM-2945.

### DIFF
--- a/plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
+++ b/plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
@@ -10,7 +10,9 @@ namespace Newspack\Newsletters\Admin;
 defined( 'ABSPATH' ) || exit;
 
 use Newspack_Newsletters\Ads;
+use WP_Error;
 use WP_Post;
+use WP_REST_Request;
 
 /**
  * Adds a read-only `newspack_newsletters_ad_status` REST field on the
@@ -47,6 +49,62 @@ class Ads_List_REST {
 			2
 		);
 		add_filter( 'posts_clauses', [ __CLASS__, 'apply_meta_sort_clauses' ], 10, 2 );
+		add_filter( 'rest_request_before_callbacks', [ __CLASS__, 'guard_invalid_dates' ], 10, 3 );
+	}
+
+	/**
+	 * Reject REST writes that supply a non-empty but non-ISO start_date or
+	 * expiry_date — e.g. '06/29/2026'. An empty string is an intentional
+	 * clear and is allowed through; sanitize_ad_date normalises valid ISO
+	 * datetime strings to a bare Y-m-d before they reach the meta table.
+	 *
+	 * @param mixed           $response Earlier filter result; passed through if non-null.
+	 * @param array           $handler  Route handler details.
+	 * @param WP_REST_Request $request  Incoming REST request.
+	 * @return mixed WP_Error with status 400 when an invalid date is found, otherwise $response.
+	 */
+	public static function guard_invalid_dates( $response, $handler, $request ) {
+		unset( $handler );
+
+		if ( null !== $response ) {
+			return $response;
+		}
+
+		if ( ! $request instanceof WP_REST_Request ) {
+			return $response;
+		}
+
+		$route   = $request->get_route();
+		$pattern = '#^/wp/v2/' . preg_quote( Ads::CPT, '#' ) . '(?:/\d+)?$#';
+		if ( ! preg_match( $pattern, $route ) ) {
+			return $response;
+		}
+
+		$method = $request->get_method();
+		if ( ! in_array( $method, [ 'POST', 'PUT', 'PATCH' ], true ) ) {
+			return $response;
+		}
+
+		$meta = $request->get_param( 'meta' );
+		if ( ! is_array( $meta ) ) {
+			return $response;
+		}
+
+		foreach ( [ 'start_date', 'expiry_date' ] as $key ) {
+			if ( ! isset( $meta[ $key ] ) ) {
+				continue;
+			}
+			$value = $meta[ $key ];
+			if ( is_string( $value ) && '' !== $value && '' === Ads::sanitize_ad_date( $value ) ) {
+				return new WP_Error(
+					'newspack_newsletters_ad_invalid_date',
+					__( 'Start and expiration dates must be valid YYYY-MM-DD dates.', 'newspack-newsletters' ),
+					[ 'status' => 400 ]
+				);
+			}
+		}
+
+		return $response;
 	}
 
 	/**
@@ -128,7 +186,7 @@ class Ads_List_REST {
 		}
 
 		global $wpdb;
-		$today = gmdate( 'Y-m-d' );
+		$today = current_time( 'Y-m-d' );
 
 		$post_status_set = [];
 		$bucket_clauses  = [];
@@ -147,14 +205,14 @@ class Ads_List_REST {
 				case 'expired':
 					$post_status_set    = array_merge( $post_status_set, [ 'publish', 'private' ] );
 					$bucket_clauses[]   = $wpdb->prepare(
-						"( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'expiry_date' AND meta_value <> '' AND meta_value < %s ) )",
+						"( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'expiry_date' AND ( STR_TO_DATE( LEFT( meta_value, 10 ), '%%Y-%%m-%%d' ) + INTERVAL 0 DAY ) < %s ) )",
 						$today
 					);
 					break;
 				case 'scheduled':
 					$post_status_set    = array_merge( $post_status_set, [ 'publish', 'private', 'future' ] );
 					$bucket_clauses[]   = $wpdb->prepare(
-						"( ( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'start_date' AND meta_value <> '' AND meta_value > %s ) ) OR {$wpdb->posts}.post_status = 'future' )",
+						"( ( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'start_date' AND ( STR_TO_DATE( LEFT( meta_value, 10 ), '%%Y-%%m-%%d' ) + INTERVAL 0 DAY ) > %s ) ) OR {$wpdb->posts}.post_status = 'future' )",
 						$today
 					);
 					break;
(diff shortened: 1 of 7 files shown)


plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -340,6 +344,100 @@ public static function ad_default_fields( $post_id, $post, $update ) {
 			return;
 		}
 		update_post_meta( $post_id, 'position_in_content', 100 );
+
+		// Seed the read-only counters so the editor's round-trip of the
+		// default value is an unchanged no-op, not a denied meta insert.
+		add_post_meta( $post_id, 'tracking_impressions', 0, true );
```
dkoo, line 349: **Suggestion:** Stale tracking counter may re-trigger the publish error after impressions accrue — Seeding `tracking_impressions`/`tracking_clicks` rows correctly fixes the *new-ad* publish path: with no meta row, core's `update_meta_value()` reads `null` (not the registered default), skips the value-equality short-circuit, and hits `auth_callback => __return_false`. Seeding a real `0` makes that round-trip a genuine no-op. But once an ad serves impressions, the counter is incremented server-side (`update_post_meta` near lines 688/709) and the editor's `useSelect` will round-trip the now-*stale* `0`; the value no longer equals the stored count, so the next save of that ad re-raises "not allowed to edit tracking_impressions". Worth confirming the admin-shell/editor strips these readonly counters from the save payload — if it doesn't, the bug resurfaces for any ad that has accrued impressions.

plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -340,6 +344,100 @@ public static function ad_default_fields( $post_id, $post, $update ) {
...
+	 * do not inflate the count.
+	 *
+	 * @param int[]               $terms    Array of term-taxonomy IDs to update.
+	 * @param \WP_Taxonomy|string $taxonomy The taxonomy object or name.
+	 */
+	public static function update_advertiser_term_count( $terms, $taxonomy ) {
+		global $wpdb;
+
+		$taxonomy_name = is_object( $taxonomy ) ? $taxonomy->name : $taxonomy;
+		$statuses      = [ 'publish', 'private', 'future', 'draft', 'pending' ];
```
dkoo: **Suggestion:** Advertiser count omits auto-draft, diverging from the list's draft bucket — `update_advertiser_term_count` counts `['publish','private','future','draft','pending']`, but the Ads list's `draft` bucket (`filter_rest_query`) also includes `auto-draft`. An `auto-draft` ad with an assigned advertiser would appear in the list but not be reflected in the term count, so the count and the list can disagree by the auto-draft rows. Auto-drafts rarely carry taxonomy terms so impact is small, but since the stated goal is for the count to match the Ads list, consider including `auto-draft` (and confirm `trash` is intentionally excluded).

plugins/newspack-newsletters/src/ads/editor/index.js
```diff
@@ -43,7 +45,7 @@ function AdEdit() {
 	const { saveEntityRecord } = useDispatch( 'core' );
 	const { removeEditorPanel } = useDispatch( editPostStore );
 	const messages = [];
-	if ( expiryDate && ! isInTheFuture( expiryDate ) ) {
+	if ( expiryDate && expiryDate < format( 'Y-m-d', new Date() ) ) {
```
dkoo: **Suggestion:** Past-date warning uses browser-local date while the server uses site time — The "expiration date is set in the past" warning compares `expiryDate` against `format( 'Y-m-d', new Date() )`, which is the *browser-local* date, whereas the PHP lifecycle now anchors "today" to the site timezone via `current_time( 'Y-m-d' )`. When the editor's browser timezone differs from the site timezone, near midnight this client-side warning can show or suppress incorrectly for the same date the server will classify differently. Impact is limited to the advisory message (not the actual active/expired decision), but a site-timezone-aware date for the comparison would make the warning agree with the server.

plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
```diff
@@ -147,25 +205,25 @@ public static function filter_rest_query( $args, $request ) {
 				case 'expired':
 					$post_status_set    = array_merge( $post_status_set, [ 'publish', 'private' ] );
 					$bucket_clauses[]   = $wpdb->prepare(
-						"( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'expiry_date' AND meta_value <> '' AND meta_value < %s ) )",
+						"( {$wpdb->posts}.post_status IN ( 'publish', 'private' ) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = 'expiry_date' AND ( STR_TO_DATE( LEFT( meta_value, 10 ), '%%Y-%%m-%%d' ) + INTERVAL 0 DAY ) < %s ) )",
```
dkoo, line 208: **Nit:** SQL status-bucket subqueries are non-sargable — The per-row correlated `EXISTS` subqueries filter on `meta_key` (indexed) plus `STR_TO_DATE( LEFT( meta_value, 10 ), ... )` (not sargable), so each bucket effectively scans the matching postmeta rows per ad. `$wpdb->prepare` usage is correct and there's no injection vector (only `$today` is bound, via `%s`; the `%%Y-%%m-%%d` literals are correctly escaped). Fine for the typically low volume of newsletter ads — flag if this list is ever expected to scale to large ad tables.

plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -340,6 +344,100 @@ public static function ad_default_fields( $post_id, $post, $update ) {
...
+	 *
+	 * Accepts only a leading ISO `Y-m-d` (optionally followed by a time
+	 * component, e.g. `2026-06-29T23:59:59`), matching the SQL filter's
+	 * `STR_TO_DATE(LEFT(meta_value,10),'%Y-%m-%d')` exactly. Any value
+	 * that does not begin with a valid calendar date returns ''.
+	 *
+	 * @param mixed $value Raw meta value.
+	 * @return string `Y-m-d`, or '' when empty or not a valid leading ISO date.
+	 */
+	public static function sanitize_ad_date( $value ) {
```
dkoo, line 365: **Nit:** PHP regex and SQL parser disagree on hand-malformed single-digit dates — `sanitize_ad_date`'s regex requires two-digit month/day (`^\d{4}-\d{2}-\d{2}`), but the SQL `STR_TO_DATE( LEFT( meta_value, 10 ), '%Y-%m-%d' )` accepts single-digit components (e.g. `2026-6-9` parses in SQL but is rejected by PHP). Such a row would be bucketed by the list filter but treated as "no constraint" by `get_status_for_post`, so the filter and the status badge could disagree. This is unreachable via the `sanitize_callback` (which normalizes on write) and via realistic legacy data (the old editor wrote two-digit ISO), so impact is essentially nil — noting it only because the docblock claims the regex matches the SQL exactly. The genuinely-impossible-date cases (`2026-02-30`, zero-day) DO align thanks to `+ INTERVAL 0 DAY` coercing them to NULL.

plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -340,6 +344,100 @@ public static function ad_default_fields( $post_id, $post, $update ) {
...
+				'taxonomy'   => self::ADVERTISER_TAX,
+				'hide_empty' => false,
+				'fields'     => 'tt_ids',
+			]
+		);
+		if ( is_wp_error( $term_ids ) ) {
+			return;
+		}
+		if ( ! empty( $term_ids ) ) {
+			wp_update_term_count_now( $term_ids, self::ADVERTISER_TAX );
```
dkoo, line 442: **Nit:** One-time advertiser recount runs synchronously on admin_init — `maybe_recount_advertiser_terms` is correctly option-gated (idempotent, writes the `_v2` flag only after success, and bails on `WP_Error` without setting it), so this is safe. The recount itself runs synchronously on the admin page load that first trips it. Advertiser terms are low-cardinality so this is fine in practice; if a site ever had thousands of advertiser terms it would be worth deferring to a scheduled event. Noting only — no change needed for the expected scale.

dkoo (approved):
Passes manual testing. All three reported behaviors are fixed and were confirmed end-to-end in an isolated environment: a new Newsletter Ad now publishes without the "not allowed to edit tracking_impressions" error, start/expiry dates store and render as clean `Y-m-d` (no "Invalid Date"), and advertiser counts are correct — draft/scheduled ads are now counted and newsletters tagged with an advertiser no longer inflate the count. The SQL preparation, site-timezone anchoring, and the `+ INTERVAL 0 DAY` impossible-date handling are all correct and well covered by the new tests. No blockers — the inline comments are minor suggestions and edge-case notes. The one worth a quick look before merge: whether the editor re-submits a now-stale `tracking_impressions` value for ads that have already accrued impressions, which could resurface the same publish error for those ads.

- `plugins/newspack-newsletters/src/admin-shell/screens/ads-list/fields.js:177` — **Nit:** Sort/export getValue returns un-sliced legacy datetime — `formatDate` was correctly updated to slice to `Y-m-d` for *display*, but the `start_date`/`expiry_date` field `getValue` callbacks still return raw `item?.meta?.start_date || ''`. For a legacy row whose meta is still an ISO datetime, sorting/export uses the full datetime string. ISO is lexically sortable so sorting still works, but the exported value would carry the time portion; slicing here too would keep it consistent with the display.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Fixed. Confirmed live first: when the counter accrues after the editor loads, the next save sends the stale value and 403s exactly as you describe — the editor round-trips the full meta object, so the seed alone wasn't enough. Added an apiFetch middleware that strips `tracking_impressions`/`tracking_clicks` from ad save payloads, and re-tested the same scenario: it now saves 200 with the server counter untouched.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Fixed. Added `auto-draft` to the counted statuses so it matches the list's draft bucket, and bumped the one-time recount sentinel to `_v3` so already-recounted sites pick up the new semantics. `trash` stays excluded by design (it's a separate list kind). Added tests for the auto-draft count and the recount refresh.

plugins/newspack-newsletters/src/ads/editor/index.js
thomasguillot: Fixed. The warning now compares against `wpDate( 'Y-m-d' )` (the non-localized site-timezone `date` from `@wordpress/date`), so the advisory agrees with the server's `current_time()` classification.

plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
thomasguillot: Acknowledged — leaving as-is. Newsletter ad volume is low, so the per-bucket `EXISTS` scans are fine; flagged for a revisit if the ad table ever grows large.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Good catch on the docblock overclaiming. Reworded it: the regex is intentionally stricter than `STR_TO_DATE` on single-digit components, but every value is normalized on write, so stored data always agrees. No behavior change.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Keeping it synchronous, consistent with your read — advertiser terms are low-cardinality, so the recount is fast. Note the auto-draft fix bumps the sentinel to `_v3`, which re-runs this once on upgraded sites; same low-cardinality cost, so no async migration.

thomasguillot:
Addressed the review feedback in c48cfdf37.

- **Stale tracking counter 403:** reproduced it live first (the editor round-trips the full meta object, so a counter that accrued after load sends a stale value and 403s), then fixed with an `apiFetch` middleware that strips the read-only `tracking_impressions`/`tracking_clicks` from ad save payloads. Re-verified: the same stale scenario now saves `200` with the server counter untouched.
- **Advertiser count:** now includes `auto-draft` to match the list's draft bucket, and the one-time recount sentinel is bumped to `_v3` so already-recounted sites pick up the new semantics. Added tests for the auto-draft count and the recount refresh.
- **Past-date warning:** compares against site-timezone `wpDate( 'Y-m-d' )` so the advisory agrees with the server's `current_time()` classification.
- **Docblock:** corrected the "matches SQL exactly" claim (the regex is intentionally stricter on single-digit components; values are normalized on write so stored data always agrees).
- **Sort/export:** `getValue` slices legacy ISO datetime to `Y-m-d` for consistency with the display.
- **SQL non-sargable** and **synchronous recount** nits: kept as-is given low ad volume and low-cardinality advertiser terms.

All inline threads replied to and resolved. Tests, lint, PHPCS, and build are green; the counter fix was verified end-to-end in a live editor.

PR #419, opened 2026-06-25
fix(group-subscription): inherit product settings on manual save

thomasguillot:
## Summary

Fixes **NPPM-2947**, found in June 29 alpha testing of the Group Subscriptions My Account feature (#148).

When an admin manually creates a WooCommerce subscription (wp-admin → Add subscription) and assigns a product that has group subscriptions enabled, the new subscription did not inherit the product's group settings. The owner had no group controls in My Account until they manually toggled the per-subscription override, even though the helper text says the subscription inherits from the product.

### Root cause

`Group_Subscription_Settings::save_group_subscription_meta()` runs on every `woocommerce_process_shop_order_meta` save and unconditionally wrote the meta-box checkbox state, which `update_subscription_settings()` diffs against the *effective* (product-inherited) value. On the Add-subscription screen the group meta box renders **before** the product line item is linked, so the "Group subscription enabled" checkbox renders unchecked. Saving then wrote an explicit `enabled = 'no'` override that defeated product inheritance.

### Fix

Dirty-tracking via a rendered baseline. The meta box emits hidden `{key}_baseline` fields capturing the values it rendered, and the save handler only persists a setting the admin actually changed away from its baseline:

- Untouched control (e.g. the unchecked-at-creation checkbox) → no override written → the subscription keeps inheriting the product's group settings.
- Intentional opt-out (unchecking a rendered-checked box) → explicit `enabled = 'no'` override is still written.

Because an inheriting subscription can change its effective group status without writing `enabled` meta, the handler also keeps the cached group-subscription ID set (consumed by the admin group/non-group list filters) in sync: it refreshes the cache whenever the rendered baseline differs from the post-save effective status, in either direction.

No change to the read/inheritance path (already correct) and no UI redesign.

## Testing

- `n test-php --group WooCommerce_Subscriptions_Integration` passes; +8 new tests covering the save dirty-tracking and group-ID cache invalidation (inheritance on/off, checked-matches-inherited, and non-group no-op).
- Verified against real WooCommerce in an isolated env: a manually-created subscription whose product enables groups reports `is_group_subscription = true` and the owner manages it after the Add-subscription save (previously `false`).
- PHPCS clean on all changed files. Reviewed via a multi-pass adversarial review loop (Codex + a second reviewer); cache-invalidation gaps it surfaced are fixed and covered by tests.

Fixes NPPM-2947.

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
@@ -360,6 +360,9 @@ class Group_Subscription_Settings {
 		$invites = Group_Subscription_Invite::get_invites( $subscription );
 		?>
 		<div class="newspack-group-subscription__container" data-subscription-id="<?php echo \esc_attr( $subscription->get_id() ); ?>">
+			<input type="hidden" name="<?php echo \esc_attr( self::GROUP_SUBSCRIPTION_META_PREFIX . 'enabled_baseline' ); ?>" value="<?php echo \esc_attr( \wc_bool_to_string( $settings['enabled'] ) ); ?>" />
+			<input type="hidden" name="<?php echo \esc_attr( self::GROUP_SUBSCRIPTION_META_PREFIX . 'limit_baseline' ); ?>" value="<?php echo \esc_attr( (int) $settings['limit'] ); ?>" />
+			<input type="hidden" name="<?php echo \esc_attr( self::GROUP_SUBSCRIPTION_META_PREFIX . 'name_baseline' ); ?>" value="<?php echo \esc_attr( $settings['name'] ); ?>" />
 			<div class="newspack-group-subscription__settings">
 				<h3><?php \esc_html_e( 'Settings', 'newspack-plugin' ); ?></h3>
 				<p>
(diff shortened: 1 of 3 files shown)


plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -493,21 +496,52 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+		}
+
+		if ( ! empty( $changed ) ) {
+			self::update_subscription_settings( $subscription, $changed );
+		}
+
+		// Effective group status can flip via inherited product settings without a meta write; refresh the cached ID set when it changed.
+		if ( isset( $_POST[ $prefix . 'enabled_baseline' ] ) ) {
+			$baseline_enabled = \wc_string_to_bool( sanitize_text_field( wp_unslash( $_POST[ $prefix . 'enabled_baseline' ] ) ) );
+			if ( $baseline_enabled !== self::get_subscription_settings( $subscription )['enabled'] ) {
```
dkoo, line 544: **Suggestion:** The post-save cache-sync reads back the effective status via `get_subscription_settings( $subscription )['enabled']`, which resolves the linked product. On the Add-subscription screen — the exact scenario this PR fixes — the product line item may not yet be linked at `woocommerce_process_shop_order_meta` time, so this read can return the un-inherited default rather than the true post-link value, leaving the `newspack_group_subscription_ids` transient briefly stale. This is harmless: the transient only drives the admin list-table group/non-group filter, and it self-heals via its 5-minute TTL plus the product-save / trash / delete clear hooks. Worth a one-line comment documenting the self-healing fallback so the next reader doesn't mistake it for an access-control path. **Consider:** the block also re-runs `get_subscription_settings()` (a product load + the `newspack_group_subscription_settings` filter) that `update_subscription_settings()` already computed for its diff — you could compute the effective settings once after the conditional save and reuse it here to avoid the redundant load.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -493,21 +496,52 @@ public static function save_group_subscription_meta( $subscription_id, $subscrip
...
+				: 0,
+			'name'    => isset( $_POST[ $prefix . 'name' ] )
+				? sanitize_text_field( wp_unslash( $_POST[ $prefix . 'name' ] ) )
+				: '',
+		];
+
+		$changed = [];
+		foreach ( [ 'enabled', 'limit', 'name' ] as $key ) {
+			$baseline_field = $prefix . $key . '_baseline';
+			if ( ! isset( $_POST[ $baseline_field ] ) ) {
```
dkoo, line 514: **Nit:** If a `{key}_baseline` hidden field is ever absent from the POST (a stale cached admin page, JS stripping the form), that key is silently skipped and the setting is never persisted even when the admin changed it. That's a deliberate fail-safe given the bug being fixed, but it does mean a legitimate edit can be dropped if the baseline field goes missing — acceptable, just flagging the trade-off.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -242,4 +242,190 @@ public function test_name_falls_back_to_singular_group_label() {
...
+		$subscription = $this->make_subscription_with_product( [ 'enabled' => 'yes' ] );
+		$prefix       = Group_Subscription_Settings::GROUP_SUBSCRIPTION_META_PREFIX;
+		set_transient( Group_Subscription_Settings::GROUP_SUBSCRIPTION_IDS_TRANSIENT, [ 999 ], MINUTE_IN_SECONDS );
+
+		// The product is no longer group-enabled at save time; the box rendered checked (baseline 'yes') and was left untouched.
+		wc_create_mock_product(
+			[
+				'id'   => 123,
+				'meta' => [ $prefix . 'enabled' => 'no' ],
+			]
```
dkoo: **Nit:** Trailing whitespace after the closing `]` on this line (`git diff --check` flags it; PHPCBF would strip it). Cosmetic only.

dkoo (approved):
Passes manual testing. Well-targeted fix for the manual-subscription inheritance bug, with strong test coverage and a clean security posture — the nonce check is intact, all new `{key}_baseline` POST reads are unslashed + sanitized, the hidden fields are `esc_attr`-escaped, and meta writes go through the CRUD API so HPOS is handled. The dirty-tracking approach (persist only on divergence from the rendered baseline) correctly preserves both product inheritance for untouched controls and the intentional opt-out. No blockers — approving. A couple of small things worth a glance before merge, none blocking.

- `plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php:432` — **Suggestion:** The new tests cover the `enabled` and `limit` dirty-tracking branches (write + no-write) and both cache-clear directions, but the `name` branch has no dedicated coverage. Since `name` falls back to the product name / group label in `get_subscription_settings()`, a regression in its baseline equality check wouldn't be caught. Worth adding a `test_save_*_name_*` pair mirroring the limit tests.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
thomasguillot: Added a comment documenting the self-healing fallback (stale read on the Add screen, recovered via the transient TTL + product save/trash/delete hooks, never an access-control path).

Skipped the reuse optimization on purpose: `update_subscription_settings()` computes the *pre-write* `$previous_settings` and only runs on a diff, while this block needs the *post-write* effective status and runs even when nothing was saved. Reusing that value would compare the wrong state. The product load is already cached, so the redundant call is cheap.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
thomasguillot: Confirmed — that dropped-edit-on-missing-baseline behavior is the intended fail-safe for the bug this fixes. Leaving as-is.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
thomasguillot: Stripped.

thomasguillot:
Re: the name dirty-tracking coverage suggestion — added `test_save_does_not_override_name_when_unchanged_from_baseline` (no override written, still inherits the product name) and `test_save_writes_name_override_when_changed_from_baseline`, mirroring the existing limit pair. Both green.

PR #424, opened 2026-06-25
fix(memberships): sync team member end date to team expiration

chickenn00dle:
### All Submissions:

**Ensures a team member's access expires no later than their team membership's expiration date.**

When a member is added to a WooCommerce Memberships **team**, the member's user membership was not always given an end date. Teams only propagates the team's expiration onto a member's membership when the member already had an existing membership, or when the team is tied to a WooCommerce subscription. For **manually-managed teams** — a team with a fixed end date but no subscription behind it — a freshly created member membership was left with the plan-relative end date, which resolves to *empty* for unlimited/subscription-typed plans. The result: the member never expired, keeping access indefinitely even after the team membership lapsed.

This adds a hook on `wc_memberships_for_teams_add_team_member` that stamps the team's expiration date onto the new member's user membership:

- **Manually-managed (non-subscription) teams** — members now inherit the team's end date instead of getting none.
- **Only ever shortens down to the team end date** — a member who already holds a *longer* end date (e.g. a separately purchased individual membership) keeps it.
- **Schedules the per-membership expiry event** — calling `set_end_date()` (re)schedules `wc_memberships_user_membership_expiry`, so members expire on time on their own.
- **Subscription-tied teams are unaffected** — the hook runs at priority 20, after Teams' own subscription integration (priority 10); for those teams the member end date already matches the team end date, so nothing changes.

Closes NPPM-2932.

Regression (manually-managed team — the bug):
1. With WooCommerce Memberships + Memberships for Teams active, create a membership plan with **no** access length (unlimited) and **no** subscription product.
2. Create a team on that plan (e.g. via Memberships → Teams) and set a fixed **Team membership end date** in the future.
3. Add a member to the team.
4. Open the member's user membership and confirm its **end date now matches the team's end date** (before this PR it was empty).
5. Confirm a scheduled `wc_memberships_user_membership_expiry` action exists for that membership at the team end date (Tools → Scheduled Actions), so the member will expire when the team does.

Preserve longer access:
6. Give a user an individual membership on the same plan with an end date **further out** than the team's end date, then add them to the team. Confirm their later end date is **preserved**, not shortened.

No regression (subscription-tied team):
7. Create a subscription-product-backed team and add a member. Confirm the member's end date still tracks the subscription/team end date as before.

### Other information:

---

## For AI

**Root cause.** Two code paths set a new team member's user-membership end date:

1. `Team::add_member()` (in `woocommerce-memberships-for-teams`) creates the membership via `wc_memberships_create_user_membership()` with **no** end date. Core then derives the end date from `MembershipPlan::get_expiration_date( $now, $args )` — plan/product length relative to *now*, which is **empty** for unlimited (and auto-renew-less subscription) plans. The team's own `_membership_end_date` is only synced in the *existing-membership* branch (`959-964`), never the *new-membership* branch (`981-994`).
2. The Teams Subscriptions integration `adjust_team_member_user_membership_data()` would set it from the subscription, but it is guarded by `if ( $subscription )` and bails entirely for non-subscription teams.

Manually-managed teams (no `_subscription_id`, fixed `_membership_end_date`) hit neither, so members get no end date.

**Fix.** `Newspack\Teams_For_Memberships::sync_member_end_date_to_team()` on `wc_memberships_for_teams_add_team_member` (priority 20, after the subscription integration's 10). It reads `$team->get_membership_end_date( 'timestamp' )`, returns early for unlimited teams, and only sets the member end date when it is shorter than the team end (`$member_end < $team_end`), preserving longer individual memberships. It mirrors `Team::set_membership_end_date()`'s status handling (reactivate if future, expire if past).

**Scope note.** This is the forward fix only. Existing members on affected sites (members already created with empty/incorrect end dates) need a separate one-off backfill that sets each member's end date to their team's end date via `set_end_date()` (so the expiry event reschedules). The fragile difference-based adjustment in `Team::set_membership_end_date()` (which produces drifting dates when an admin edits the team end date) is a vendor-side design issue left for upstream.

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
+++ b/plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
@@ -401,10 +401,10 @@ class Teams_For_Memberships {
 	 * expires even after the team membership lapses.
 	 *
 	 * This enforces the invariant that a team member's access ends no later than the team's
-	 * expiration date, regardless of subscription status, while preserving a longer end date a
-	 * member may already hold from a separately purchased individual membership. Calling
-	 * set_end_date() also (re)schedules the per-membership expiry event, so the member expires on
-	 * time on their own.
+	 * expiration date, regardless of subscription status, unless the member already holds a longer
+	 * finite end date from a separately purchased individual membership. Calling set_end_date()
+	 * also (re)schedules the per-membership expiry event, so the member expires on time on their
+	 * own.
 	 *
 	 * @see https://linear.app/a8c/issue/NPPM-2932
 	 *
@@ -412,10 +412,11 @@ class Teams_For_Memberships {
 	 * @param \SkyVerge\WooCommerce\Memberships\Teams\Team        $team            The team instance.
 	 * @param \WC_Memberships_User_Membership                     $user_membership The related user membership instance.
 	 */
-	public static function sync_member_end_date_to_team( $member, $team, $user_membership ) {
+	public static function sync_member_end_date_to_team( $member, $team, $user_membership ): void {
 		if (
 			! is_a( $team, '\SkyVerge\WooCommerce\Memberships\Teams\Team' ) ||
-			! is_a( $user_membership, '\WC_Memberships_User_Membership' )
+			! is_a( $user_membership, '\WC_Memberships_User_Membership' ) ||
+			! method_exists( $team, 'get_membership_end_date' )
 		) {
 			return;
 		}
@@ -430,22 +431,27 @@ class Teams_For_Memberships {
 		$team_end   = (int) $team_end;
 		$member_end = (int) $user_membership->get_end_date( 'timestamp' );
 
-		// Only ever shorten access down to the team end date; never override a longer end date the
-		// member already holds (e.g. a separately purchased individual membership). A missing end
-		// date is read as 0 and corrected here.
+		// Set the member end to the team end unless the member already holds a longer finite end
+		// date (e.g. a separately purchased individual membership). A missing end date reads as 0
+		// and is corrected here.
 		if ( $member_end >= $team_end ) {
 			return;
 		}
 
-		$user_membership->set_end_date( $team_end );
+		// Pass a MySQL/UTC date string for consistency with this file's sibling set_end_date() call.
+		$user_membership->set_end_date( gmdate( 'Y-m-d H:i:s', $team_end ) );
 
-		// Mirror Team::set_membership_end_date(): reactivate if the new end date is in the future,
-		// or expire if the team has already lapsed.
-		$now = time();
-		if ( $team_end > $now && $user_membership->is_expired() ) {
+		// If the team is still current, reactivate a membership that was previously expired -- this
+		// also covers an existing, expired individual membership being absorbed into a team.
+		// is_active() does not auto-reactivate, so this has to be explicit.
+		//
+		// We intentionally do NOT force a membership added to an already-lapsed team to `expired`
+		// here. Setting the past end date is enough: is_active() lazily expires it on the next
+		// access check. Forcing the status now would synchronously fire
+		// wc_memberships_user_membership_status_changed, which removes the reader from the plan's
+		// newsletter lists at the ESP -- undesirable to trigger en masse during a (bulk) add.
+		if ( $team_end > time() && $user_membership->is_expired() ) {
 			$user_membership->update_status( 'active' );
-		} elseif ( $team_end <= $now ) {
-			$user_membership->update_status( 'expired' );
 		}
 
 		$user_membership->add_note( __( 'Membership end date synced to the team membership expiration date.', 'newspack-plugin' ) );
--- a/plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
+++ b/plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
@@ -390,21 +390,27 @@ class Teams_For_Memberships {
 	}
 
 	/**
-	 * Stamp the team's expiration date onto a member's user membership when they are added.
+	 * Fill in a member's user-membership end date from the team when it has none.
 	 *
-	 * Teams only propagates the team end date to a member's user membership when the member
-	 * already had an existing membership (see Team::add_member()), or when the team is tied to a
-	 * WooCommerce subscription (see the Teams Subscriptions integration, which bails for
-	 * non-subscription teams). For manually-managed teams that have a fixed end date but no
-	 * subscription, a newly created member membership is left with the plan-relative end date,
-	 * which resolves to empty for unlimited/subscription-typed plans. The member then never
-	 * expires even after the team membership lapses.
+	 * Teams propagates the team end date to a member's user membership when the member already had
+	 * an existing membership (see Team::add_member()), or when the team is tied to a WooCommerce
+	 * subscription (see the Teams Subscriptions integration, which bails for non-subscription
+	 * teams). For manually-managed teams that have a fixed end date but no subscription, a newly
+	 * created member membership is left with the plan-relative end date, which resolves to empty
+	 * for unlimited/subscription-typed plans. The member then never expires even after the team
+	 * membership lapses.
 	 *
-	 * This enforces the invariant that a team member's access ends no later than the team's
-	 * expiration date, regardless of subscription status, unless the member already holds a longer
-	 * finite end date from a separately purchased individual membership. Calling set_end_date()
-	 * also (re)schedules the per-membership expiry event, so the member expires on time on their
-	 * own.
+	 * This fills that specific gap: when the member's user membership has no end date at all, it
+	 * inherits the team's. We deliberately only fill a *missing* end date and never override one
+	 * that is already set -- whether by the plan's access length, a separately purchased individual
+	 * membership, or upstream Team::add_member() (which reconciles existing memberships and their
+	 * status on its own). That keeps us from fighting fixed-length plans' midnight-snapped dates,
+	 * shortening a longer independently-held membership, or duplicating upstream's membership note.
+	 *
+	 * Calling set_end_date() also (re)schedules the per-membership expiry event so the member
+	 * expires on time on their own; is_active() lazily expires the membership if the team end is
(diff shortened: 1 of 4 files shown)


plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
```diff
@@ -386,6 +388,68 @@ public static function restore_team_meta_on_renewal( $action, $item ) {
...
+	 * set_end_date() also (re)schedules the per-membership expiry event, so the member expires on
+	 * time on their own.
+	 *
+	 * @see https://linear.app/a8c/issue/NPPM-2932
+	 *
+	 * @param \SkyVerge\WooCommerce\Memberships\Teams\Team_Member $member          The team member instance.
+	 * @param \SkyVerge\WooCommerce\Memberships\Teams\Team        $team            The team instance.
+	 * @param \WC_Memberships_User_Membership                     $user_membership The related user membership instance.
+	 */
+	public static function sync_member_end_date_to_team( $member, $team, $user_membership ) {
```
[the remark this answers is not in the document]
chickenn00dle: Addressed in 3ccfb5ac1. Added PHPUnit coverage for `sync_member_end_date_to_team()` in `tests/unit-tests/plugins/teams-for-memberships.php` — cases for no end date (stamps the team end), extend-shorter, preserve-longer (no shortening), unlimited team no-op, reactivate-expired on a future team, no force-expire on an already-lapsed team, and the non-Team guard. Added two namespaced mock classes so the `is_a()` guards are exercised. Full group passes (18 tests, 51 assertions).

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
```diff
@@ -386,6 +388,74 @@ public static function restore_team_meta_on_renewal( $action, $item ) {
...
+			return;
+		}
+
+		$team_end   = (int) $team_end;
+		$member_end = (int) $user_membership->get_end_date( 'timestamp' );
+
+		// Set the member end to the team end unless the member already holds a longer finite end
+		// date (e.g. a separately purchased individual membership). A missing end date reads as 0
+		// and is corrected here.
+		if ( $member_end >= $team_end ) {
```
dkoo: **Suggestion:** Finite plan-relative end dates later than the team end aren't capped — The `$member_end >= $team_end` guard preserves any member end date already at or beyond the team end, assuming it came from a separately purchased individual membership. But a *newly created* team membership on a plan with a finite access length gets a plan-relative end date (e.g. now + 1 year); if that lands later than a team capped to, say, 30 days, the guard treats it as a pre-existing purchase and skips the sync — so the member keeps access past the team's expiration and the intended invariant ("access ends no later than the team end") isn't enforced for finite-length plans. The date comparison alone can't distinguish a pre-existing individual membership from the plan default on the just-created team membership. (The inverse case — an *unlimited* individual membership on the same plan reading as end date `0` and being shortened to the team end — matches what upstream `Team::add_member()` already does, so it's not a regression.) Whether this matters depends on whether manually-managed teams here ever use finite-length plans; the Compliance Week case uses unlimited/subscription-typed plans, which this fix handles correctly.

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
```diff
@@ -386,6 +388,74 @@ public static function restore_team_meta_on_renewal( $action, $item ) {
...
+		// also covers an existing, expired individual membership being absorbed into a team.
+		// is_active() does not auto-reactivate, so this has to be explicit.
+		//
+		// We intentionally do NOT force a membership added to an already-lapsed team to `expired`
+		// here. Setting the past end date is enough: is_active() lazily expires it on the next
+		// access check. Forcing the status now would synchronously fire
+		// wc_memberships_user_membership_status_changed, which removes the reader from the plan's
+		// newsletter lists at the ESP -- undesirable to trigger en masse during a (bulk) add.
+		if ( $team_end > time() && $user_membership->is_expired() ) {
+			$user_membership->update_status( 'active' );
```
dkoo: **Suggestion:** Reactivation synchronously enrolls members into ESP newsletter lists — The reactivation branch calls `update_status('active')` to revive a previously-expired membership absorbed into a still-current team. In WC Memberships an active transition fires `wc_memberships_user_membership_saved`, which newspack-newsletters hooks to enroll the reader into the plan's newsletter lists at the ESP (`add_user_to_lists` → a synchronous, unbatched ESP API call). That enrollment is **not** short-circuited for an expired→active transition — the skip only applies when the *previous* status was already active (`class-woocommerce-memberships.php:296`). The method's own comment deliberately avoids the symmetric *removal* side effect when force-expiring, but the reactivation it does perform has the same class of side effect in the additive direction. For a single add this is one extra ESP call; the concern is the bulk path, where the add-member action fires once per imported row.

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
```diff
@@ -386,6 +388,74 @@ public static function restore_team_meta_on_renewal( $action, $item ) {
...
+		}
+
+		$user_membership->add_note( __( 'Membership end date synced to the team membership expiration date.', 'newspack-plugin' ) );
```
dkoo, line 451: **Nit:** Possible duplicate note / redundant re-write on fixed-date plans — In the existing-membership branch, upstream `Team::add_member()` has already set the end date and added its own note before this hook runs. Because fixed-access-length plans snap the end date to local midnight and `Team::set_membership_end_date()` applies its own timezone adjustment, the stored member end can land a few hours short of `$team->get_membership_end_date('timestamp')`, making `$member_end < $team_end` true. The hook then re-sets the date (idempotent) and adds a second, generic note ("Membership end date synced to…") on top of upstream's specific one. Cosmetic, but consider re-setting / noting only when the value actually changes materially.

dkoo (commented):
Passes manual testing. Thanks for the thorough write-up and the added unit coverage — the helper's branches are well tested and the upstream status semantics are faithfully mirrored. This is a solid forward fix for a real paid-access bug. Two things worth deciding consciously before merge, neither blocking the Compliance Week fix:

1. The reactivation branch (`update_status('active')`) synchronously enrolls a previously-expired member into the plan's ESP newsletter lists via newspack-newsletters' `add_user_to_lists` (`wc_memberships_user_membership_saved`), and that path is not short-circuited for an expired→active transition. This is the same class of ESP churn the method's own comment avoids for the removal direction — it matters most during a bulk member add (the CSV import calls `add_member` per row).
2. A finite plan-relative end date that lands later than the team end is treated as a pre-existing individual purchase and left uncapped, so the new "access ends no later than the team" invariant isn't enforced for finite-length plans. It doesn't affect the unlimited/subscription-typed plans this fix targets, so it may be fine — just confirm it's intended.

Also a heads-up that the separate backfill for members already created with empty/incorrect end dates still needs to happen for affected sites.

dkoo (approved):

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
chickenn00dle: Great catch — removed the reactivation branch entirely in 7948dee37. It turned out to be redundant: upstream \`Team::add_member()\` already reactivates an absorbed previously-expired membership (it calls \`update_status('active')\` once the end date it just set is in the future), and freshly created memberships are already \`active\`. So the hook no longer touches status at all — no synchronous ESP enrollment (or removal) fires from it in either direction, which kills the bulk-add churn concern. \`is_active()\` still handles lazy expiry for the past-end case.

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
chickenn00dle: Fixed in 7948dee37. The hook now only acts when the membership has **no** end date at all, so on fixed-date (and any finite-length) plans where upstream has already set a midnight-snapped date, it early-returns — no idempotent re-write, no second note. We only fill the specific gap that unlimited/subscription-typed plans leave empty.

plugins/newspack-plugin/includes/plugins/class-teams-for-memberships.php
chickenn00dle: You're right, and after reworking this (7948dee37) I leaned into scoping rather than capping. The hook now only fills a **missing** end date and never overrides one the plan/upstream already set — so a finite-length plan keeps its plan-relative date and is *not* capped to the team end by this hook. I went this way because capping would mean overwriting a plan-computed date, which is exactly what reintroduces the midnight-snap re-write and the shorten-a-longer-independent-membership risks from the other two threads. The fix is deliberately scoped to the unlimited/subscription-typed plans where the date comes out empty (Compliance Week's case). I updated the docblock to claim only that narrower invariant. If finite-length team plans turn out to need true capping, I think that belongs upstream in `Team::add_member()` rather than here — happy to file a follow-up if you think it's worth it. WDYT?

chickenn00dle:
Thanks for the thorough review, @dkoo — really helpful. Pushed 7948dee37, which reworks the helper around both points:

**1. Reactivation → ESP enrollment.** Removed the status block entirely. It was redundant: upstream `Team::add_member()` already reactivates an absorbed expired membership, and new memberships are created active. The hook now only ever *fills a missing end date* and never touches status, so no synchronous ESP enrollment/removal fires from it during a (bulk) add. `is_active()` covers lazy expiry for the past-end case.

**2. Finite plan-relative date not capped.** Rather than cap (which would mean overriding a plan-computed date and reintroduces the midnight-snap re-write + shorten-a-longer-membership risks from the inline threads), I narrowed the hook to only fill an end date when there is none. Finite-length plans keep their plan-relative date; the fix is scoped to the unlimited/subscription-typed plans where the date comes out empty (the actual Compliance Week case), and the docblock now claims only that narrower invariant. If finite-length team plans need true capping, I think that belongs upstream in `Team::add_member()` — happy to file a follow-up. Replied on the thread with the details.

**Backfill.** Confirmed — the one-off reconciliation for members already created with empty/incorrect end dates still needs to run on affected sites; tracking it separately from this forward fix.

Also added/updated unit coverage for the new behavior (fill-missing, skip-existing, no status change, lapsed-team, unlimited-team, non-Team guard). Resolved the three inline threads.

PR #429, opened 2026-06-25
feat: add tag labels to the Carousel block

jason10lee:
### All Submissions:

A polish pass that fills a component gap (tag labels in the **Carousel block**) and makes the label treatment consistent across every context that renders post labels.

**Carousel block (`newspack-blocks`)**
- New **Show tag labels** toggle (`showTagLabels`, default `true`) in the Carousel's Post Meta settings, mirroring the Content Loop (Homepage Posts) block.
- Editor preview and front-end (`view.php`) render tag labels above the headline, after the category/sponsor labels. The taxonomy fetch is gated on the attribute, and the labels-only case (categories off, no sponsors) renders correctly.
- Carousel tag-label color honors the theme Customizer (`.wpnbpc` added to the inline color selector).

**Label order + layout (`newspack-theme`) — also corrects the already-merged Content Loop block + single-post / archive / search rendering**
- Tag labels now render **after** the category (and sponsor) label, consistently across **single posts, archives, search, blog index, Content Loop, and Carousel** (`entry-header.php`, `content-archive.php`, `content-excerpt.php`), matching the "after any category or sponsorship labels" intent.
- Category and tag labels now sit **inline on one line** in all of those contexts. In the Content Loop / Carousel blocks they previously stacked because the block CSS `.wpnbha/.wpnbpc .cat-links{display:flex}` beat the generic inline rule on specificity; this adds block-scoped rules at higher specificity (excluding sponsor rows so their own flex/gap layout wins deterministically). The editor-preview SCSS was updated to match.

> ⚠️ This is a **theme** change, so the layout/order fix also applies to the already-merged Content Loop block and single posts / archives / search — not just the Carousel.

**Rollout note:** `showTagLabels` defaults to `true` (consistent with Content Loop). Any existing Carousel block on a site that already uses flagged tag labels will begin rendering them on next load. Worth a release-notes mention.

Related: [NPPD-383: "Breaking News" or "Developing" Label](https://linear.app/a8c/issue/NPPD-383). Builds on parent PRs Automattic/newspack-plugin#4381, Automattic/newspack-blocks#2282, Automattic/newspack-theme#2613.

1. With the Newspack plugin and a Newspack theme active: go to **Posts → Tags**, edit a tag, check **Display as label**, optionally set **Label text**, and save.
2. Assign the tag to a couple of published posts (give them a category too).
3. Add a **Carousel** block showing those posts. In the block's **Post Meta** settings, confirm the **Show tag labels** toggle (on by default). On the front end, confirm the label appears above the headline, **after the category, on one line**.
4. Confirm the same category-then-tag, single-line treatment in: a **Homepage Posts** block, a **single post**, a **category/tag archive**, and a **search results** page.
5. Confirm the labels-only case (category off, no sponsor) still shows the tag label.
6. **Appearance → Customize → Tag Labels**: pick a color and confirm it applies to Carousel labels too.
7. Toggle **Show tag labels** off and confirm the labels disappear.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — N/A (presentational: block render + theme template/CSS).
* [x] Have you successfully run tests with your changes locally? — lint (JS/SCSS/PHPCS) clean; CI green; verified live in an isolated env across single post, Content Loop, Carousel, archive, and search.


themes/newspack-theme/newspack-theme/sass/plugins/newspack-tag-labels.scss
```diff
@@ -36,18 +36,42 @@ amp-script .tag-labels {
...
+// Tag labels render after the category (and sponsor) label and sit on the same line,
+// mirroring the sponsor → category convention. The category `.cat-links` is immediately
+// followed by the tag-labels `.cat-links.tag-labels` in every context (single posts,
+// home, archive, search, and the Content Loop / Carousel blocks), so keep it inline-block
+// and space the tag labels that follow it. The sponsor `.cat-links.sponsor-label` keeps
+// its own layout (from newspack-sponsors), so it is excluded here.
+.cat-links:not( .sponsor-label ):has( + .tag-labels ) {
+	display: inline-block;
+}
+.cat-links + .tag-labels {
```
dkoo, line 48: **Suggestion:** The global `.cat-links + .tag-labels` inline rule doesn't fire when the preceding label is a sponsor `.cat-links.sponsor-label` (deliberately excluded from the `:has()` inline rule so sponsors keep their own flex/gap layout). In the native-sponsors, sponsor-only-display branch that means the tag labels can wrap onto a new line below the sponsor rather than sitting inline. The inline comment suggests this is intentional, but it's the one branch where the same-line goal doesn't hold — worth a quick visual check on a sponsored post that also has flagged tags before merge.

plugins/newspack-blocks/src/blocks/carousel/block.json
```diff
@@ -72,6 +72,10 @@
 			"type": "boolean",
 			"default": false
 		},
+		"showTagLabels": {
```
dkoo, line 75: **Nit:** `showTagLabels` defaults to `true`, so any existing Carousel block with no saved value will begin rendering flagged tag labels on the next load, with no editor action. This matches the already-merged Content Loop default, so it's consistent — just worth a release-notes line so the rollout isn't a surprise for publishers.

themes/newspack-theme/newspack-theme/sass/plugins/newspack-tag-labels.scss
```diff
@@ -36,18 +36,42 @@ amp-script .tag-labels {
 }
 /* stylelint-enable */
 
+// Tag labels render after the category (and sponsor) label and sit on the same line,
+// mirroring the sponsor → category convention. The category `.cat-links` is immediately
+// followed by the tag-labels `.cat-links.tag-labels` in every context (single posts,
+// home, archive, search, and the Content Loop / Carousel blocks), so keep it inline-block
+// and space the tag labels that follow it. The sponsor `.cat-links.sponsor-label` keeps
+// its own layout (from newspack-sponsors), so it is excluded here.
+.cat-links:not( .sponsor-label ):has( + .tag-labels ) {
```
dkoo, line 45: **Nit:** The category's inline adjustment depends on the CSS `:has()` selector. On a browser without `:has()` support the category stays block-level and the tag labels drop to the next line — graceful degradation, and `:has()` is broadly supported now, so this is acceptable; noting only because it's the one hard dependency the layout introduces.

dkoo (approved):
Manually tested. Approve — clean presentational change with no blockers. The Carousel tag-label support mirrors the already-merged Content Loop (Homepage Posts) block: same `showTagLabels` default, same `$attributes['showTagLabels'] ?? true` render gate, and the same category → tag DOM order. The theme-wide reorder (tag labels always after the category/sponsor label, inline on one line) is internally consistent across all three templates and both the sponsor and no-sponsor branches, and the SCSS matches the new DOM order.

Verified in an isolated env: correct category → tag ordering in single posts, archives, search, the Content Loop block, and the Carousel; the labels-only case (category off) renders; the toggle-off case hides labels and skips the taxonomy fetch; the Customizer tag-label color reaches the Carousel `.wpnbpc` labels; and no PHP notices/warnings are logged. A few low-priority notes are inline.

themes/newspack-theme/newspack-theme/sass/plugins/newspack-tag-labels.scss
jason10lee: Confirmed and fixed in `cf8a94de9`. Checked in an env with native sponsors, category display hidden, on a sponsored post carrying a flagged tag.

The wrap is real, but it lands in the Content Loop rather than the Carousel. `homepage-articles/view.scss` nests `.cat-links { display: flex }` inside `.wpnbha` (0,2,0), which overrides the theme's `.sponsor-label { display: inline-flex }` (0,1,0) — so excluding `.sponsor-label` from the inline re-assertion leaves it block-level. `.wpnbpc` never sets `display` on `.cat-links`, so the Carousel keeps `inline-flex` and was already correct.

Before:

Content Loop before the fix: SPONSORED on one line, EXCLUSIVE wrapped below

After:

Content Loop after the fix: SPONSORED and EXCLUSIVE inline on one line

The fix restores the sponsor's own display inside the block wrappers, without touching the flex/gap layout the exclusion exists to protect:

```scss
.cat-links.sponsor-label:has( + .tag-labels ) {
	display: inline-flex;
}
```

Carousel with the fix, as a control — unchanged:

Carousel with the fix: SPONSORED and EXCLUSIVE inline, unchanged from before

Both Content Loop shots are the same fixture with only the compiled CSS toggled — I reverted the SCSS, rebuilt, and re-shot rather than reusing an earlier capture, since the fixture had changed in between. `lint:scss` is clean.

plugins/newspack-blocks/src/blocks/carousel/block.json
jason10lee: Agreed on both counts — `showTagLabels` defaults to `true`, so an existing Carousel block starts rendering flagged labels on the next load with no editor action, matching the already-merged Content Loop default.

Leaving the default as-is for that consistency.

themes/newspack-theme/newspack-theme/sass/plugins/newspack-tag-labels.scss
jason10lee: Agreed, and left as-is.

Worth noting the fix in `cf8a94de9` adds one more `:has()` selector rather than introducing a second dependency — on a browser without support, the sponsor stays block-level and the labels drop below it, which is the same graceful degradation you describe here.

PR #432, opened 2026-06-25
feat(newsletters): add redirect-to-URL option to subscription form

kmwilkerson:
### All Submissions:

Adds a "redirect to URL" option to the **Newsletter Subscription Form** block (`newspack-newsletters/subscribe`), matching the redirect-after-action capability the **Checkout Button** block already has. After a successful subscribe, an optional **click-through Continue button** sends the reader to a custom URL or back to the previous page.

This deliberately **matches** the Checkout Button's pattern (click-through, not auto-redirect). A separate follow-up will replace the click-through with an auto-redirect in **both** blocks — out of scope here.

| Layer | File | Change |
|---|---|---|
| Attributes | `src/blocks/subscribe/block.json` | `afterSuccessBehavior` (`""`), `afterSuccessButtonLabel` (`"Continue"`), `afterSuccessURL` — same names/defaults as the Checkout Button. |
| Editor | `src/blocks/subscribe/edit.js` | New "After subscribe" inspector panel (Select + Button label + Custom URL), plus the Continue button in the Success-state preview for WYSIWYG parity. |
| SSR | `src/blocks/subscribe/index.php` | Exposes config as `data-after-success-*` on the block container (only when a behavior is set). |
| Frontend | `src/blocks/subscribe/view.js` | Builds + wires the Continue button in the `status === 200` success branch; click → `window.location.href` (`custom`) / `window.history.back()` (`referrer`). |

**Notes on matching the Checkout Button:**

- The Checkout Button's Continue+redirect lives entirely inside the modal-checkout iframe (`thankyou.php` button + `modal.js` close handler), which the modal-less subscribe form can't reuse — so it's re-implemented natively in the subscribe success path. Reader-facing behavior is identical.
- The button is built in `view.js`, not SSR: `view.js` bails early when there's no `<form>` (the no-JS/GET already-subscribed render), so an SSR'd button would be inert. SSR only ships the `data-*` config.
- Empty behavior = no button (inline success message, as today); `custom` with an empty URL = no-op — both match the Checkout Button.
- No URL validation / open-redirect guard is added — deliberately matching the Checkout Button, which trusts the publisher-entered URL via `window.location.href` (`esc_url()` already strips `javascript:`/`data:` at render).

Addresses NPPD-1553.

1. Add a **Newsletter Subscription Form** block to a page; in the inspector open **After subscribe** and set **Post-signup button → Go to a custom URL**, enter a URL, and optionally change the button label.
2. On the front end, subscribe with a valid email. On success the message appears with a **Continue** button; clicking it navigates to the custom URL.
3. Repeat with **Go to the previous page** → Continue does `history.back()`.
4. With **Show the success message** (default) → no button, success message unchanged.
5. In the editor, toggle **Edited State → Success** and confirm the Continue button appears when a behavior is configured.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — N/A.
* [x] Have you successfully run tests with your changes locally? — `phpcs` (root ruleset) clean; build + prettier clean. Reviewed with `/code-review` (one low finding — Continue button now inherits the submit button's full class list for block-theme / named-palette parity; 2nd commit). Browser-tested on a local site against the built bundle and server-rendered pages:
  * Custom URL → submit → "Take the survey" button → click navigates to the custom URL.
  * Previous page → submit → "Go back" button → click does `history.back()`.
  * Default (Show success message) → no Continue button, success message unchanged.
  * SSR `data-after-success-*` correct for custom/referrer/default; absent when no behavior set. No PR-introduced debug-log warnings.


plugins/newspack-newsletters/src/blocks/subscribe/edit.js
```diff
@@ -412,6 +443,14 @@ export default function SubscribeEdit( {
 										allowedFormats={ [ 'core/bold', 'core/italic' ] }
 									/>
 								</div>
+								{ afterSuccessBehavior && (
```
adekbadek: **Suggestion:** Editor preview shows a button that won't render on the front end — The editor preview renders the Continue button whenever `afterSuccessBehavior` is set, but the front end suppresses it when the behavior is `custom` and the URL is empty (`view.js` bails on `'custom' === behavior && ! url`). An author who picks *Go to a custom URL* and leaves the URL blank sees a Continue button in the editor that never appears on the published page. Mirroring the front-end guard in the preview (hide it when `'custom' === afterSuccessBehavior && ! afterSuccessURL`) keeps the editor WYSIWYG.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
```diff
@@ -170,12 +170,24 @@ function render_block( $attrs ) {
...
+	$after_success_label    = $attrs['afterSuccessButtonLabel'] ?? '';
 	// phpcs:enable
 	ob_start();
 	?>
 	<div
 		class="wp-block-newspack-newsletters-subscribe newspack-newsletters-subscribe <?php echo esc_attr( get_block_classes( $attrs ) ); ?>"
 		data-success-message="<?php echo \esc_attr( $attrs['successMessage'] ); ?>"
+		<?php if ( $after_success_behavior ) : ?>
+			data-after-success-behavior="<?php echo \esc_attr( $after_success_behavior ); ?>"
+			data-after-success-url="<?php echo \esc_url( $after_success_url ); ?>"
```
adekbadek, line 189: **Suggestion:** esc_url() differs from the Checkout Button's esc_attr() and can silently drop unusual URLs — `data-after-success-url` is escaped with `esc_url()`, whereas the Checkout Button escapes the same publisher value with `esc_attr()`. `esc_url()` is the safer choice and normal `https://` URLs and same-site absolute paths (`/thank-you`) round-trip fine, so keeping it is reasonable — but it isn't literal parity, and it silently rewrites unusual-but-intended values. Verified against the block's own renderer: `javascript:alert(1)` is dropped to empty (good), but a bare `foo/bar` (no leading slash) is rewritten to `http://foo/bar` — i.e. the button would navigate to a wrong host rather than to a same-site `foo/bar` path, with no editor feedback. Worth a one-line note that the escaping intentionally differs, and a thought on whether a relative path without a leading slash is a case publishers might hit.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
```diff
@@ -28,6 +28,55 @@ function domReady( callback ) {
...
+	button.className = 'newspack-newsletters-subscribe__continue ' + ( submitButton?.className || 'submit-button' );
+	const submitStyle = submitButton?.getAttribute( 'style' );
+	if ( submitStyle ) {
+		button.setAttribute( 'style', submitStyle );
+	}
+	button.addEventListener( 'click', () => {
+		if ( 'custom' === behavior ) {
+			window.location.href = url;
+		} else if ( 'referrer' === behavior ) {
+			window.history.back();
```
adekbadek: **Suggestion:** 'Go to the previous page' dead-ends when there's no session history — `window.history.back()` is a no-op when there's no in-session history entry (reader landed directly, opened in a new tab, or the previous entry is cross-origin) — the reader clicks Continue and nothing happens, with no feedback. `document.referrer` is a more reliable "previous page" signal: prefer it when it's same-origin, and fall back to `history.back()` (see suggested code). Since this PR's aim is Checkout-Button parity, the identical dead-end lives in the Checkout Button's `modal.js` close handler (`newspack-blocks/src/modal-checkout/modal.js`, ~line 604, same `window.history.back()`) — fix both in this PR so they don't diverge.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
```diff
@@ -28,6 +28,55 @@ function domReady( callback ) {
...
+		}
+	} );
+	responseContainer.appendChild( button );
```
adekbadek, line 84: **Suggestion:** Continue button gets no focus and the response region isn't announced — After success the form is replaced and the Continue button is appended, but focus stays where the (now-detached) submit button was and nothing announces the new control to assistive tech (the response container has no `aria-live`). Since the feature's whole purpose is to move the reader onward, moving focus to the new button — or making the response region a live region — would materially help keyboard and screen-reader users. The missing `aria-live` is pre-existing on the success message, so this is an enhancement rather than a regression.

plugins/newspack-newsletters/src/blocks/subscribe/block.json
```diff
@@ -70,6 +70,17 @@
 			"type": "string",
 			"default": "Thank you for signing up!"
 		},
+		"afterSuccessBehavior": {
```
adekbadek, line 73: **Nit:** afterSuccessBehavior default "" vs the Checkout Button's undefined — `afterSuccessBehavior` is given a `""` default here while the Checkout Button leaves it undefined. Behaviorally identical — `index.php` reads it with `?? ''` and the editor coerces with `'' !== afterSuccessBehavior` — so no functional impact; noting only as a small divergence from the block being mirrored.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
```diff
@@ -28,6 +28,55 @@ function domReady( callback ) {
...
+	const behavior = container.getAttribute( 'data-after-success-behavior' );
+	if ( ! behavior ) {
+		return;
+	}
+	const url = container.getAttribute( 'data-after-success-url' );
+	// Match the Checkout Button: a 'custom' redirect with no URL is a no-op.
+	if ( 'custom' === behavior && ! url ) {
+		return;
+	}
+	const label = container.getAttribute( 'data-after-success-label' ) || 'Continue';
```
adekbadek, line 64: **Nit:** Front-end "Continue" fallback label isn't translatable — The `|| 'Continue'` fallback in `view.js` is a hardcoded English string, and the `afterSuccessButtonLabel` default in `block.json` is likewise the literal `"Continue"` (not run through `__()`), so an untouched button renders English regardless of locale. In practice SSR emits `data-after-success-label` from the attribute, so the JS fallback only fires if an author sets a behavior and then clears the label — low impact, and it matches the Checkout Button's own untranslated default. Noted for parity; if localization matters, sourcing the default from a translated localized string would fix both blocks.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
adekbadek, line 84: **Nit:** No idempotency guard on button insertion — `maybeRenderAfterSuccessButton` unconditionally `appendChild`s. It's safe today because `endFlow` runs once per submit and the success branch detaches the form (preventing re-submission), but a cheap guard — bail if `responseContainer.querySelector( '.newspack-newsletters-subscribe__continue' )` already exists — would prevent a second button if that control flow ever changes.

adekbadek (changes requested):
Works fine, but some changes are needed

- `plugins/newspack-newsletters/src/blocks/subscribe/view.js:84` — **Suggestion:** Already-subscribed / reloaded readers never get the redirect — The button and redirect only fire for a live AJAX subscribe in the current page view. When a reader is already subscribed, `render_block` emits the block with `data-status="200"` and no `<form>`, and `view.js` returns early at `if ( ! form )` before it can build the button. So a returning or already-subscribed reader — including one who simply reloads after subscribing — gets no Continue affordance. Likely acceptable for the stated "redirect the reader who just engaged" use case, but it's a behavioral gap worth an explicit decision/note since it isn't called out.

plugins/newspack-newsletters/src/blocks/subscribe/edit.js
kmwilkerson: Good catch — fixed in 930772a. The preview now mirrors the front-end guard and hides the Continue button when the behavior is `custom` with an empty URL, so the editor matches what publishes.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
kmwilkerson: Fixed in 930772a. `referrer` now prefers a same-origin `document.referrer` and only falls back to `window.history.back()` when there is no usable referrer, so a reader who landed directly or opened in a new tab no longer dead-ends. Applied the same change to the Checkout Button (`newspack-blocks/src/modal-checkout/modal.js`) so the two stay in sync.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
kmwilkerson: Added the guard in 930772a — `maybeRenderAfterSuccessButton` now bails if a `.newspack-newsletters-subscribe__continue` already exists in the response container, so a future control-flow change can't append a duplicate.

kmwilkerson: Partly addressed in 930772a: focus now moves to the Continue button after it's appended, which is the strongest win here since the whole point of the button is to move the reader onward. Left the `aria-live` on the response region alone — as you noted it's pre-existing on the success message, so wiring it up is a separate enhancement rather than something this PR regresses.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
kmwilkerson: Kept `esc_url()` (it's the safer choice and normal URLs / same-site absolute paths round-trip fine), and added a code comment above it in 930772a noting the escaping intentionally diverges from the Checkout Button's `esc_attr()`, plus the leading-slash caveat: a schemeless relative path without a leading slash (`foo/bar`) gets rewritten to a different host, so publishers should use `/foo/bar` or an absolute URL for a same-site destination.

plugins/newspack-newsletters/src/blocks/subscribe/block.json
kmwilkerson: Leaving the `""` default as-is. The editor and `index.php` both rely on it being a string (`'' !== afterSuccessBehavior`, `?? ''`), so an explicit default is safer here than mirroring the Checkout Button's `undefined` — and as you noted the behavior is identical either way. Noting the small divergence is fair; I'd rather keep the defined default than risk an `undefined` slipping through the editor guard.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
kmwilkerson: Leaving this for parity with the Checkout Button, which has the same untranslated default. In practice SSR always emits `data-after-success-label` (defaulting to `Continue` from `block.json`), so the JS `|| 'Continue'` fallback only fires if an author sets a behavior and then clears the label — low-impact enough that sourcing a translated default is better done as a follow-up that fixes both blocks together rather than diverging one of them here.

kmwilkerson:
Thanks for the review — changes are in 930772a.

**Applied**
- **Editor preview parity** — the preview now hides the Continue button when the behavior is `custom` with an empty URL, matching the front end.
- **`referrer` no longer dead-ends** — prefers a same-origin `document.referrer`, falls back to `history.back()`. Same fix applied to the Checkout Button (`modal-checkout/modal.js`) so the two don't diverge.
- **Focus** moves to the Continue button after success, so keyboard/SR users land on the actionable control.
- **Idempotency guard** on button insertion.
- **`esc_url()`** kept, with a code comment noting the intentional divergence from the Checkout Button's `esc_attr()` and the schemeless-relative-path caveat (use a leading slash / absolute URL for same-site).

**Kept as-is, with rationale in the threads**
- `block.json` `""` default (vs the Checkout Button's `undefined`) — the editor/PHP guards rely on it being a string.
- Untranslated `'Continue'` JS fallback — matches the Checkout Button; better fixed as a follow-up across both blocks. Left the response region's missing `aria-live` alone since it's pre-existing.

**On the already-subscribed / reloaded reader gap** — this is the intended behavior: the Continue button only appears for the reader who just subscribed in the current page view. A returning or already-subscribed reader (including one who reloads after subscribing) is server-rendered with no form, so `view.js` returns early and shows the plain success message. That matches the "redirect the reader who just engaged" intent; surfacing it for the no-form render would mean SSR'ing the button. I've added a comment at that early return documenting the decision so it's explicit rather than incidental.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
```diff
@@ -28,11 +28,106 @@ function domReady( callback ) {
...
+	button.textContent = label;
+	// Inherit the form button's classes and inline color styling for full visual
+	// parity (block-theme `wp-element-button`, named-palette `has-*` color classes,
+	// and custom hex styles). The spinner is a child node, and `in-progress` lives
+	// on the form, so the submit button carries no transient classes to strip.
+	button.className = 'newspack-newsletters-subscribe__continue ' + ( submitButton?.className || 'submit-button' );
+	const submitStyle = submitButton?.getAttribute( 'style' );
+	if ( submitStyle ) {
+		button.setAttribute( 'style', submitStyle );
+	}
```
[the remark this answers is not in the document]
kmwilkerson: Good catch — fixed in 973f98d. Confirmed the gap: in the default style variation the submit button's red background / white text come from the `button[type="submit"]` selector, which the Continue button doesn't match (it's `type="button"` — kept deliberately so it can't submit a surrounding form — and it lives in `__response`, a sibling of the now-replaced form). The copied `className`/inline style only cover the *customized*-color cases, so a default-color button rendered unstyled.

Added a base rule in `style.scss` giving `.newspack-newsletters-subscribe__continue` the same default `background-color`/`color`/`cursor` as the submit button. The named-palette classes and inline colors copied from the submit button in `view.js` are more specific, so a customized button still overrides the default. The modern style variation already targets `.submit-button` (which the Continue button carries) and sets no background/color itself, so it's unaffected.

plugins/newspack-newsletters/src/blocks/subscribe/edit.js
adekbadek: Confirmed in 930772a — the preview conditional now matches the front-end guard, so the editor no longer previews a button that will not publish.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
adekbadek: Accepted — keeping esc_url() is the right call, and the comment now documents the intentional divergence from the Checkout Button plus the schemeless-relative-path caveat.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
adekbadek: Confirmed — goToPreviousPage() prefers a same-origin document.referrer and falls back to history.back(), and the identical fix is mirrored in the Checkout Button modal.js so the two do not drift. Verified by reading the diff against both blocks; build and CI are green.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
adekbadek: Confirmed — focus now moves to the Continue button, which is the substantive win here. Agreed the response-region aria-live is a separate, pre-existing enhancement and fine to leave for a follow-up.

plugins/newspack-newsletters/src/blocks/subscribe/block.json
adekbadek: Agreed — keeping the defined empty-string default is the safer choice since the editor guard and index.php both treat it as a string.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
adekbadek: Agreed — parity with the Checkout Button untranslated default is fine here; a translated default is better landed as a follow-up that fixes both blocks together.

plugins/newspack-newsletters/src/blocks/subscribe/view.js
adekbadek: Confirmed in 930772a — the guard bails when a Continue button already exists in the response container.

adekbadek (approved):
All eight threads addressed: the two behavioral suggestions (the `referrer` dead-end and Continue-button focus) are fixed, the referrer fix is correctly mirrored into the Checkout Button's `modal.js` so the blocks don't diverge, and the `esc_url()` divergence plus the already-subscribed no-form behavior are now documented in-code — the remaining items are reasoned won't-fixes on nits. Re-verified each fix by reading the diff against its thread, with a clean `newspack-newsletters` build, PHPCS, and 51/51 CI green (didn't re-drive the browser — these are refinements to the render/navigation path verified in the first pass). Approving.

PR #435, opened 2026-06-25
feat(tags): make private tags opt-out and add reader-data toggle

wil-gerken:
### All Submissions:

Productizes the Private Tags feature for full release. Two changes:

1. **Inverts the feature flag to a per-site opt-out.** The feature is now **on by default**. A site can opt out via the `NEWSPACK_PRIVATE_TAGS_DISABLED` constant or the `newspack_private_tags_enabled` filter. `is_enabled()` is kept (logic changed, not deleted) so a site running a conflicting custom implementation on the same `np_private_tag` meta key can stay on its own code. Init is deferred to `after_setup_theme` so the filter can be registered from any plugin or theme.

2. **Adds a `reader_data` behavior toggle.** `filter_reader_activity()` — which strips private tag IDs from the client-side reader-activation data — was the one always-on behavior; it is now a publisher-toggleable setting like the other nine. Default is on (strip), matching today's behavior. It appears as the last toggle in the renamed "Data and integrations" group of the private-tags customizer.

**⚠️ This un-gates the feature for all publishers.** The admin **Tags** UI (Private column, Quick Edit checkbox, "(private)" label, add/edit checkbox) now appears for everyone — intended. The reader-data strip becomes a toggle (default on = unchanged behavior).

**Rollout coordination (the shared `np_private_tag` key).** The feature now recognizes any tag carrying the `np_private_tag` meta key, so sites running a *custom* implementation on that key must opt out before this reaches release, to avoid double-handling. Status:
- **TT** — the `NEWSPACK_PRIVATE_TAGS_DISABLED` constant is set on their production site. ✅
- **`newspack-custom-code` sites** — opt-out added to the `newspack-private-tags` plugin in Automattic/newspack-custom-code#275 (covers any publisher running that plugin).

Deactivating a custom plugin later is the migration path: core takes over automatically (same key, zero data migration).

Reference: NPPD-1466
Related: Automattic/newspack-custom-code#275

1. With no constant defined, confirm the feature is **on by default** — Newspack → Settings → Advanced Settings shows the "Customize where private tags are hidden" card.
2. Mark a tag private and add it to a post. Confirm the private tag is hidden from: post tag links, its archive page (404), CSS classes, GAM ad targeting, and the Yoast sitemap, and that its ID is stripped from the `newspack_reader_data` reader-activation data.
3. Turn on "Customize where private tags are hidden", then uncheck **"Remove private tags from audience management data"** and save. Reload the post → the private tag ID returns to the reader-activation data; the other behaviors are unaffected.
4. Define `NEWSPACK_PRIVATE_TAGS_DISABLED` as `true` (or add a `newspack_private_tags_enabled` → `false` filter from a plugin/theme) → the feature is fully off and the settings card is hidden.
5. Confirm the admin Tags screen shows the Private column with a checkmark on private tags, the "(private)" label, and the "Private tag" checkbox on the add/edit forms.

### Other information:

<sub>phpcs clean; <code>n test-php --group private-tags</code> → 83 tests / 137 assertions OK; eslint clean. Verified end-to-end on a live site (all 10 behaviors + admin UI + both opt-out paths).</sub>


jason10lee (approved):
Reviewed independently in an isolated environment (full plugin + JS test suites, eslint, PHPCS against the monorepo root ruleset — all green), plus a cross-repo impact pass and a multi-model code review. Clean, well-documented change, and the test delta genuinely strengthens coverage. **Approved for merge once `NEWSPACK_PRIVATE_TAGS_DISABLED` is set where needed** (the rollout section already tracks this). A couple of non-blocking notes:

**Suggestions (non-blocking)**
1. Consider help text on the new "Remove private tags from audience management data" toggle — unchecking it re-exposes private tag IDs in client-side reader data (round-trippable to names via REST), so a one-liner would make opting out an informed choice.
2. The `NEWSPACK_PRIVATE_TAGS_DISABLED` *constant* opt-out path isn't covered by a test (the filter path is). It's honestly hard to test in-process, so a `@runInSeparateProcess` test or an explicit note in the trait would be options to close the gap.

**Nits**
- The new `@constant` docblock omits the `@status` line the plugin's other constant docs carry.  A couple agentic reviewers seemed to think this was important.  🤷‍♂️ 

---

**Verification detail**
- PHPUnit (newspack-plugin): 1833 / 5349 OK (9 pre-existing env skips). `--group private-tags`: 83 / 137 OK.
- JS: 28 suites / 313 tests pass. eslint clean. PHPCS clean (root ruleset).
- `after_setup_theme` deferral verified safe: the action is registered at plugin load, well before the hook fires, and every behavior hook targets `init`/admin/REST/render. The only `is_enabled()` callers are `init()` and a render-time helper — nothing evaluates it earlier, so the filter opt-out path stays consistent.
- Default behavior preserved: `reader_data => true` plus the `all` master default mean the reader-activation strip still runs unchanged for existing sites.

**Cross-repo impact detail**
- Additive surface (new constant + `newspack_private_tags_enabled` filter + new toggle). No other repo calls the core `Private_Tags` class or filters `newspack_reader_activity_article_view`, so reader-activation consumers see unchanged data by default.
- The shared `np_private_tag` key has a single in-workspace custom implementation (Texas Tribune's), covered by the opt-out constant per the rollout section.

The earlier automated review comments all look addressed in the current revision.

wil-gerken:
Thanks Jason! Thanks for the detailed review and verification notes. Excellent.

A few quick replies:

* **Help text on the toggle:** I skipped this one because unchecking any of the toggles re-exposes private tags, and this is actually the least direct exposure since it’s only the term IDs. The others — Yoast, feeds, CSS classes, and GAM — can surface the readable name or slug directly. Singling this one out for a warning felt a bit off, so I kept it consistent.

* **Constant opt-out test:** There was already a note in there for that — over by the filter test in `tests/unit-tests/tags/test-private-tags.php` — so I think we’re good. Hopefully it won’t ever be used anyway.

* **`@status` docblock:** Nice catch. I added that to match the other constants.

Thanks again!

wil-gerken:
@jason10lee,

One more heads-up before I merge.

After your approval and my doc tweak, Copilot’s re-review made a fair point about `is_enabled()`: the `newspack_private_tags_enabled` filter ran *after* the constant check, so a filter returning `true` could re-enable the feature even when `NEWSPACK_PRIVATE_TAGS_DISABLED` was set.

Since that constant is our opt-out kill switch — and what’s shielding Texas Tribune — it felt wrong for a stray filter to be able to override it. It also didn’t match the docblock, which frames both the constant and filter as opt-outs, not re-enables.

I changed it so the constant has final precedence: it now hard-returns `false` before the filter runs, and the filter only applies to the default-on path. Conservative direction — a stronger kill switch, nothing loosened.

I also added a process-isolated test for this (`test_disable_constant_takes_precedence_over_filter`), since a constant can’t be undefined mid-process. PHPCS is clean, `--group private-tags` is green (84/138), CI is passing, and Copilot’s re-review came back with no comments.

Since this is a behavior tweak on logic you reviewed, I wanted to loop you in before merging. I don’t think it needs a deep re-review — just checking that it sits right with you.

Thanks again!

jason10lee:
> After your approval and my doc tweak, Copilot’s re-review made a fair point about `is_enabled()`: the `newspack_private_tags_enabled` filter ran _after_ the constant check, so a filter returning `true` could re-enable the feature even when `NEWSPACK_PRIVATE_TAGS_DISABLED` was set.
> 
> Since that constant is our opt-out kill switch — and what’s shielding Texas Tribune — it felt wrong for a stray filter to be able to override it. It also didn’t match the docblock, which frames both the constant and filter as opt-outs, not re-enables.

Yeah, I could see the value of allowing a filter to do that if we wanted another plugin to control this programmatically.  I don't see one at the moment, and if you don't, as Claude likes to say, Ya Ain't Gonna Need It.  So this is fine by me!

PR #443, opened 2026-06-26
fix(insights): bq-internal paywall influenced rate (tab 4 gates)

dkoo:
## What & why

The Insights **Tab 4 Gates** `paywall_conversion_influenced_14d` metric was the last one still computed via the buggy `Woo_Order_Resolver` (a GA4-cookie → `(int)customer_id` cast that drops anonymous attempts to customer 0 → ~0 Woo matches). This PR rewires it to read a **BQ-internal `{rate, denominator}`** result the hub computes directly from confirmed conversions (`form_submission_success`) — no downstream Woo join — and removes `Woo_Order_Resolver` from `Gates_Metric` entirely.

The already-remediated metrics (paywall direct/revenue/avg, both regwall) are **untouched** (verify-only).

## Changes

- **`compute_influenced_rate_from_proxy()`** added to `Gates_Metric` (ported from `Conversion_Metric`): reads one row `{ paywall_conversion_influenced_rate: float|null, conversion_denominator: int }`. Handles empty window, null rate (no converters → non-computable with a real denominator), schema drift (missing column → `data_missing`), and proxy errors.
- **`get_paywall_conversion_influenced_14d`** now calls it; **`compute_paywall_rate_from_proxy` deleted**.
- **`Woo_Order_Resolver` removed from `Gates_Metric`** (import, property, constructor param). The *class* is NOT deleted — Donors and Prompts still use it; the class deletion is a later PR.
- **`data_missing`** added to `Gates_Metric::populated_scalar` (it previously carried `numerator` but no `data_missing`).
- **`CACHE_PREFIX` v1 → v2** and **`Gates_REST_Controller::cache_schema_version()` override added** — the prefix was previously dead (the controller `use`d the cache trait but never overrode the version), so a response-shape change couldn't bust stale caches. Now wired.
- **`METRIC_SOURCES['paywall_conversion_influenced_14d']` reclassified `hybrid` → `hub`** — the metric is now fully hub-computed, so a hub failure must always count toward the tab-error banner (previously skipped on non-WC sites).

## Coordinated hub PR

This depends on the hub builder `build_paywall_conversion_influenced_14d` (in `newspack-manager-admin`) emitting the new `{rate, denominator}` shape. That hub PR is the coordinated counterpart (link to follow). The consumer is independently testable here because the tests mock the proxy. Note: the hub builder is also touched by the intraday PR (newspack-manager-admin#469); that conflict is a trivial predicate-swap reconciliation handled at merge.

## Testing

- Full insights group: **478 tests green**. Root-ruleset PHPCS clean.
- New/updated tests cover: rate+denominator read, null-rate non-computable, `data_missing` on schema drift, proxy error, the `data_missing` scalar flag, the cache-version wiring, and the `hub` reclassification in the non-WC hub-outage banner test.

Base: `feat/insights-rsm` (Insights epic). Part of NPPD-1692.

## Bundled consistency fix (post-Copilot)

Also hardens `Conversion_Metric::compute_influenced_rate_from_proxy` to reject a non-integer numeric denominator (erroring instead of silently truncating), matching the integer-only count validation in `Gates_Metric` and `Prompts_Metric`. Small cross-tab consistency follow-up to the Copilot review on this PR.

## Review-driven additions

- **Influenced-only conversions no longer masked:** the BQ-internal influenced metric reports a denominator (count of paywall converters) and no numerator, so `paywall_section_totals()` now gates the Paid section's empty state on the influenced *denominator* — an influenced-only window (direct 0, influenced positive) renders its scorecards instead of "No paywall conversions."
- **`data_missing` surfaced on Tab 4:** the backend's schema-drift flag is now wired through the Gates frontend (`GatesScalarMetric` type + a `scalarToCard` branch mirroring the Conversion tab), so drift shows the shared "data could not be loaded" note instead of a believable 0%. Full JS suite green (670).

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
@@ -425,73 +425,53 @@ final class Gates_Metric {
 	}
 
 	/**
-	 * Compute the influenced-14d paywall conversion rate, converter-denominated (NPPD-1764).
+	 * Read a BQ-internal influenced rate metric: one row carrying a precomputed
+	 * SAFE_DIVIDE rate (null when there are no converters) and an integer denominator.
 	 *
-	 * = distinct subscribers in the window whose conversion had a paywall-capable gate
-	 * exposure in a PRIOR session within 14d ÷ ALL new subscribers in the window. I.e.
-	 * "of our subscribers, what share were paywall-influenced" — user-level, matching the
-	 * doc's conceptual framing and Tab 3's 7.1–7.4 converter framing (NPPD-1766). (The
-	 * Direct rate is exposure-denominated by design — a different lens — so the two are
-	 * intentionally not the same base.)
-	 *
-	 * Replaces the prior attempt-denominated rate (÷ `count($rows)`), which keyed the
-	 * denominator off influenced checkout *attempts* — deflated on anonymous-conversion-
-	 * heavy publishers, so the rate read inflated.
-	 *
-	 * - **Numerator** stays hub/GA4-cross-session-sourced and Woo-matched (`count_unique_
-	 *   completed_users`): order meta has no session timing to place a prior-session
-	 *   exposure, so influenced can't move to the order-meta model the way Direct did.
-	 *   It remains anonymous-undercounted (NPPD-1685); NPPD-1747 improves its completeness
-	 *   for the whole influenced family — do NOT build a gates-specific cross-session join.
-	 * - **Denominator** is the anonymous-inclusive Woo identity spine
-	 *   ({@see Subscribers_Metric::get_new_subscribers_in_window()}). Because the GA4-matched
-	 *   numerator is not a strict subset of it, `rate_value()` suppresses any >100% to a
-	 *   non-computable em-dash (the coherence guard, parity with the Direct/Prompts rates).
-	 *
-	 * Local denominator + hub numerator → 'hybrid' (see METRIC_SOURCES).
-	 *
-	 * @param string            $query_name Catalog name (the influenced rows to match).
-	 * @param DateTimeInterface $start      Window start.
-	 * @param DateTimeInterface $end        Window end.
+	 * @param string            $query_name      Catalog query name.
+	 * @param string            $rate_key        Rate column key.
+	 * @param string            $denominator_key Denominator column key.
+	 * @param DateTimeInterface $start           Window start.
+	 * @param DateTimeInterface $end             Window end.
 	 * @return array
 	 */
-	private function compute_paywall_rate_from_proxy(
+	private function compute_influenced_rate_from_proxy(
 		string $query_name,
+		string $rate_key,
+		string $denominator_key,
 		DateTimeInterface $start,
 		DateTimeInterface $end
 	): array {
-		if ( ! $this->woocommerce_active() ) {
-			// Non-WC publisher: no local subscribers to denominate against. Empty
-			// state, not a fake 0% (NPPD-1737 Option A scoping).
-			return $this->populated_scalar( 0.0, false, 0, 'rate', 0 );
-		}
-
 		$rows = $this->proxy->query( $query_name, $start, $end );
 		if ( is_wp_error( $rows ) ) {
 			return $this->error_scalar( 'rate', $rows );
 		}
-		if ( ! is_array( $rows ) ) {
-			// Malformed hub response (not an array): surface as an error, not a confident
-			// 0% — parity with the donation/subscription malformed-rows handling (NPPD-1745
-			// #3). An empty array is valid (0 influenced) and handled below.
-			return $this->error_scalar( 'rate', new \WP_Error( 'bigquery_proxy_malformed_rows', __( 'The query returned an unexpected shape.', 'newspack-plugin' ) ) );
+		if ( empty( $rows ) ) {
+			return $this->populated_scalar( 0.0, false, null, 'rate' );
 		}
-
-		// Numerator: distinct subscribers whose conversion was paywall-influenced (the
-		// hub influenced rows, Woo-matched). Empty rows → 0 influenced, which is a real
-		// 0% against the subscriber denominator below — not "no data".
-		$numerator = empty( $rows ) ? 0 : $this->woo_resolver->count_unique_completed_users( $rows );
-
-		// Denominator: all new subscribers in the window (converters), not influenced
-		// attempts — anonymous-inclusive, the same Woo spine the source-mix totals use.
-		$denominator = $this->subscribers_metric()->get_new_subscribers_in_window( $start, $end );
-
-		$rate = $this->rate_value( $numerator, $denominator );
-		return null === $rate
-			? $this->populated_scalar( 0.0, false, $denominator, 'rate', $numerator )
-			: $this->populated_scalar( $rate, true, $denominator, 'rate', $numerator );
+		if ( ! is_array( $rows[0] ) || ! array_key_exists( $rate_key, $rows[0] ) || ! array_key_exists( $denominator_key, $rows[0] ) ) {
+			return $this->populated_scalar( 0.0, false, null, 'rate', null, true );
+		}
+		$denominator = $rows[0][ $denominator_key ];
+		if ( ! is_numeric( $denominator ) ) {
+			return $this->error_scalar(
+				'rate',
+				new \WP_Error( 'bigquery_proxy_malformed_value', __( 'The query returned a non-numeric denominator.', 'newspack-plugin' ) )
+			);
+		}
+		$denominator = (int) $denominator;
+		$rate        = $rows[0][ $rate_key ];
+		if ( null === $rate ) {
+			return $this->populated_scalar( 0.0, false, $denominator, 'rate' );
+		}
+		if ( ! is_numeric( $rate ) ) {
+			return $this->error_scalar(
+				'rate',
+				new \WP_Error( 'bigquery_proxy_malformed_value', __( 'The query returned a non-numeric value.', 'newspack-plugin' ) )
+			);
+		}
+		return $this->populated_scalar( (float) $rate, $denominator > 0, $denominator, 'rate' );
 	}
-
 
 	/**
 	 * Fetch (memoized per window) the `gates_performance_by_gate` hub rows, so the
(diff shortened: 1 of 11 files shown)


dkoo:
Coordinated hub PR (the BQ-internal builder this reads): Automattic/newspack-manager-admin#470.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
```diff
@@ -422,74 +413,54 @@ private function read_optional_count( array $row, string $key ): ?int {
...
+			return $this->populated_scalar( 0.0, false, null, 'rate', null, true );
+		}
+		$denominator = $rows[0][ $denominator_key ];
+		if ( ! is_numeric( $denominator ) ) {
+			return $this->error_scalar(
+				'rate',
+				new \WP_Error( 'bigquery_proxy_malformed_value', __( 'The query returned a non-numeric denominator.', 'newspack-plugin' ) )
+			);
+		}
+		$denominator = (int) $denominator;
```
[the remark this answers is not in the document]
dkoo: Done in 488983e9. The denominator is a BigQuery `COUNT(DISTINCT)` (always a non-negative integer), so a non-integer is malformed rather than truncatable — it now errors via the same `(float) $x !== (float) (int) $x` check this class already uses for count-like fields, instead of silently casting. The ported `Conversion_Metric` reader has the same looseness; I've filed a follow-up to harden it identically for parity.

plugins/newspack-plugin/tests/unit-tests/insights/class-test-gates-metric.php
```diff
@@ -1301,155 +1300,111 @@ static function ( string $query_name ) {
...
-	private function resolver_with_unique_completed( int $count ): Woo_Order_Resolver {
-		$resolver = $this->createMock( Woo_Order_Resolver::class );
-		$resolver->method( 'count_unique_completed_users' )->willReturn( $count );
-		return $resolver;
+	public function test_constructor_needs_no_woo_order_resolver() {
+		$ref = new \ReflectionClass( \Newspack\Insights\Gates_Metric::class );
+		$src = file_get_contents( $ref->getFileName() ); // phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown -- local file path from ReflectionClass, not a remote URL.
+		$this->assertStringNotContainsString( 'Woo_Order_Resolver', $src );
+		$this->assertInstanceOf( \Newspack\Insights\Gates_Metric::class, new \Newspack\Insights\Gates_Metric() );
 	}
```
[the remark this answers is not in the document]
dkoo: Done in 488983e9. Replaced the `file_get_contents` source-string check with reflection over the constructor parameters and typed properties (asserting none resolves to `Woo_Order_Resolver`), keeping the zero-arg instantiation assertion. No filesystem IO, and it won't false-trip on an incidental mention in a comment.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
```diff
@@ -422,74 +413,58 @@ private function read_optional_count( array $row, string $key ): ?int {
...
-		// Denominator: all new subscribers in the window (converters), not influenced
-		// attempts — anonymous-inclusive, the same Woo spine the source-mix totals use.
-		$denominator = $this->subscribers_metric()->get_new_subscribers_in_window( $start, $end );
-
-		$rate = $this->rate_value( $numerator, $denominator );
-		return null === $rate
-			? $this->populated_scalar( 0.0, false, $denominator, 'rate', $numerator )
-			: $this->populated_scalar( $rate, true, $denominator, 'rate', $numerator );
+		if ( ! is_array( $rows[0] ) || ! array_key_exists( $rate_key, $rows[0] ) || ! array_key_exists( $denominator_key, $rows[0] ) ) {
+			return $this->populated_scalar( 0.0, false, null, 'rate', null, true );
```
dkoo, line 441: **Suggestion:** data_missing is emitted here but never surfaced on Tab 4 — On schema drift this branch returns a populated zero with `data_missing => true`, mirroring how `Conversion_Metric` flags an untrustworthy zero. But the Gates frontend never consumes the flag: `src/wizards/insights/tabs/gates/scalarToCard.ts` has no `data_missing` branch (the Conversion tab's `scalarToCard.ts` does), and the `GatesScalarMetric` type in `src/wizards/insights/api/gates.ts` doesn't declare the field. So in a deploy-skew / schema-drift scenario the card renders an ordinary `0%` instead of the "Some data could not be loaded" note `MetricNote` already supports — the drift signal this PR adds has no observable effect on Tab 4 yet. The frontend wiring is out of scope for a backend-only PR, but worth a paired follow-up so the new flag isn't dead.

**Risk if not addressed:** A hub schema change that escapes notice during the coordinated rollout makes every site's paywall-influenced card show a believable 0% for as long as the drift persists, with no visible 'data missing' signal to prompt investigation — a quietly wrong number on an internal analytics card, no data mutation.

**Consider:** If the hub influenced shape ever drifts, the Gates paywall-influenced card should show the same "data could not be loaded" note the Conversion tab already shows. Add a `data_missing` field to the Gates scalar type and a `data_missing` branch in the Gates scalar-to-card mapping so the flag the backend now emits is actually surfaced.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
```diff
@@ -422,74 +413,58 @@ private function read_optional_count( array $row, string $key ): ?int {
...
+		if ( null === $rate ) {
+			return $this->populated_scalar( 0.0, false, $denominator, 'rate' );
+		}
+		if ( ! is_numeric( $rate ) ) {
+			return $this->error_scalar(
+				'rate',
+				new \WP_Error( 'bigquery_proxy_malformed_value', __( 'The query returned a non-numeric value.', 'newspack-plugin' ) )
+			);
+		}
+		return $this->populated_scalar( (float) $rate, $denominator > 0, $denominator, 'rate' );
```
dkoo, line 465: **Suggestion:** Influenced-only conversions get hidden by the Paid-section empty state — The new hub-shaped influenced metric returns `{rate, denominator}` only — it never sets a numerator. But `paywall_section_totals()` (line 832) still gates the Paid section's empty state on `max( (int) $direct['numerator'], (int) $influenced['numerator'] )`. With the influenced numerator now always `null`→`0`, an influenced-only window — direct conversions `0` but a positive influenced rate — coerces `paywall_conversions_total` to `0`, and `PaidReaderConversionSection.tsx` then renders the "No paywall conversions in this timeframe" empty state, replacing all four scorecards including the influenced card that holds a real, non-zero rate. The section-total computation lives in this file, so the backend half is in scope: consider deriving the total from the rate/denominator (or teaching the empty-state gate to account for a positive influenced rate) so a genuine influenced result isn't masked.

**Risk if not addressed:** For the duration a publisher's direct count sits at 0 while influenced is positive, the whole Paid reader conversion section reads as 'no conversions' — understating paywall performance to that publisher on their dashboard. No money or data is affected; the harm is a misleading admin chart that undersells the feature.

**Consider:** Because the new influenced method no longer reports a conversion count, the Paid section's empty-state gate should not treat an influenced-only window as "no conversions". Either derive the section total from the rate/denominator rather than a numerator that no longer exists, or have the empty-state check account for a positive influenced rate, so a real influenced result isn't hidden behind the "No paywall conversions" message.

dkoo (commented):
Self review. Clean, well-tested backend port — no blockers. The new proxy-read parsing faithfully mirrors the `Conversion_Metric` original (every branch: empty rows, schema drift, null rate, non-numeric error, computable path), the cache-version wiring is now correct and idiomatic, and the `hybrid`→`hub` reclassification correctly makes a hub outage count toward the tab-error banner on non-WC sites. Two follow-ups worth a look before this surfaces to publishers: (1) the new hub shape carries no conversion *count*, but the Paid section's empty-state still gates on a numerator — so an influenced-only window (direct conversions 0, influenced rate positive) renders as "No paywall conversions" and hides the real influenced scorecard; (2) the backend now emits `data_missing` on schema drift, but the Gates frontend never reads it, so drift renders as a confident 0% instead of a "data could not be loaded" note. Both are display-only on an internal dashboard (no data/money/send impact); neither blocks merge, but (1) is partly backend (the section-total computation lives in this file).

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
dkoo: Fixed in 88ec3ba. `paywall_section_totals()` now reads the influenced **denominator** instead of its (now-absent) numerator: the BQ-internal influenced metric reports `conversion_denominator` = COUNT(DISTINCT) of paywall converters, which is a real conversion count, so `max( direct numerator, influenced denominator )` again keeps an influenced-only window (direct 0, influenced positive) rendering its scorecards. Added `test_paywall_section_totals_influenced_only_renders` and updated the two existing totals tests to the new influenced shape (denominator = converter count, no numerator).

plugins/newspack-plugin/includes/wizards/insights/metrics/class-gates-metric.php
dkoo: Agreed — the backend now emits `data_missing` but Tab 4's frontend doesn't read it yet, so it's dead until wired. Keeping it out of this backend PR as you noted; filed a paired follow-up to add `data_missing` to `GatesScalarMetric` (`src/wizards/insights/api/gates.ts`) and a `data_missing` branch in `src/wizards/insights/tabs/gates/scalarToCard.ts`, mirroring the Conversion tab, so schema drift surfaces the "Some data could not be loaded" `MetricNote` instead of a confident 0%. Happy to fold it into this PR instead if you'd prefer it land together.

dkoo: Folded into this PR after all (86e0095d). `GatesScalarMetric` now declares `data_missing`, and `tabs/gates/scalarToCard.ts` has the same `data_missing` → `{ dataMissing: true }` branch as the Conversion tab, so schema drift surfaces the shared "data could not be loaded" `MetricNote` instead of a 0%. Added `tabs/gates/scalarToCard.test.ts` (error > data_missing > normal precedence) and updated the gates scalar fixtures for the new required field. Full JS suite green (670).

PR #445, opened 2026-06-28
feat(newspack-ui): standardize snackbars + accessible notices

thomasguillot:
## What

First of two PRs for the snackbar audit ([DSGNEWS-185](https://linear.app/a8c/issue/DSGNEWS-185/audit-snackbar-usage-across-newspack-workspace)). Scoped to the `newspack-ui` / my-account (front-end) snackbar and notices. The wp-admin snackbar standardization (`WizardSnackbar`, advertising wizard, story-budget) is a separate follow-up PR.

## Layout

- **Single position.** Collapsed the four `newspack-ui__snackbar` position variants into one: top-right, 24px inset (`spacer-5`), slide in from the right. The other three variants were dead code.
- **Admin-bar aware via CSS variable.** `top` now uses `var(--wp-admin--admin-bar--height, 0px)`, which WordPress sets responsively, so the breakpoint overrides are gone and it falls back to 0 on the public front end.
- **Grid item layout.** Each `__snackbar__item` is a CSS grid (`icon | content | close`, the icon and close columns added only when present via `:has()`). The icon pins to the top, the content wraps, and the close button is vertically centered.
- **Content-sized, no width jump.** Each item sizes to its own content and is right-aligned, capped at `--newspack-ui-width-s` (410px). A short snackbar no longer resizes when a longer one appears.
- **Graceful reflow.** On dismiss an item slides out at full height, then its slot collapses, so the items below animate up instead of jumping. Inactive and reused notices (e.g. `openNotice(el, false)`) collapse to zero height too, so they reserve no space in the stack.

## Motion

- Removed the animated countdown bar (`__content::after`).
- Slide only (no opacity fade), since the item travels fully off-screen.
- `prefers-reduced-motion: reduce` disables the animation entirely.

## Buttons

- New `newspack-ui__button--ghost-dark` variant for ghost buttons on dark surfaces (text `neutral-0`, hover `neutral-80`, light focus ring).
- The snackbar close button now uses the button system (`--icon --ghost-dark`), so hover, focus-visible, and transitions come for free. It is 36px, vertically centered, with a 4px right inset.

## Error icon

- Error-type snackbars automatically render the `newspack-ui__svg-icon--error` icon in the icon column, on both the PHP (`print_notices`) and JS (`createNotice`) paths.

## Tokens

- Renamed the modal-coupled width tokens to generic ones: `--newspack-ui-width-{xs,s,m,l}` (was `--newspack-ui-modal-width-*`, plus a new `xs` of 300px). All consumers (modals, woo overrides, my-account, newspack-theme) were updated; no aliases remain.

## Accessibility

Notices previously had no ARIA, so screen readers never announced them. Now:

- `type` drives the announcement politeness and the error icon (no color returns): `error` is assertive, everything else is polite.
- Snackbars announce via two persistent, visually-hidden live regions (`role="status"`/`aria-live="polite"` and `role="alert"`/`aria-live="assertive"`) created inside `.newspack-ui`. `openNotice()` writes the message into the matching region, so both on-load notices (e.g. my-account messages after a redirect) and dynamic ones (group actions, copy-link) are announced reliably. On-load activation is deferred a tick so the regions are registered before content is injected.
- Inline `newspack-ui__notice` markup carries `role="alert"` (`--error`) / `role="status"` (else).

## API

- `Newspack_UI::add_notice()` / `createNotice()` carry `type` (default `success`); the string shorthand `add_notice( $msg, 'error' )` is still supported. `corner` was removed (single position). Notices are stored flat and rendered in one container.
- Error call sites pass `error` so they announce assertively and show the error icon; `login_needed` stays polite (it is a call to action, not an error).

## Demo

- The `?ui-demo` Notices section has the snackbar triggers in a horizontal stack, a "Show persistent snackbar" example (icon + dismissible close), and "Show snackbar" now uses `createNotice` so it demonstrates the full auto-hide and reflow.

## Verification

- `lint:scss` / `lint:js` / `phpcs` clean for changed files; built and confirmed in the enqueued `dist/`.
- Verified on the live `?ui-demo` and my-account: single and stacked snackbars, content sizing, graceful reflow, 36px centered ghost-dark close, error icon, and the admin-bar offset.


plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -59,34 +59,41 @@ public static function enqueue_assets() {
...
-				'corner'         => 'top-right',
-				'type'           => 'success',
+				'type'           => 'success', // Severity; drives the ARIA announcement only ('error' announces assertively, anything else politely).
 				'id'             => uniqid(),
 				'autohide'       => true, // If false, the notice will have a close button.
 				'active_on_load' => true, // Whether the notice should be visible on page load.
 			]
 		);
-		self::$notices[ $notice['corner'] ][ $notice['id'] ] = $notice;
+		self::$notices[ $notice['id'] ] = $notice;
```
adekbadek, line 111: **Suggestion:** Type the rewritten `add_notice()` for PHP 8.3 — `add_notice()` is being rewritten here, so it's the moment to type it to the 8.3 target — `add_notice( string $message, string|array $args = [] ): string` — and declare the property `private static array $notices = []` (:23). The sibling entry point `My_Account::add_notice()` (reader-activation/class-my-account.php:62) already does exactly this (`string $message, string $type = 'success' ): void`), so aligning keeps the two notice APIs visibly consistent. Not a correctness issue.

plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -98,36 +105,34 @@ public static function print_notices() {
...
-							<?php endif; ?>
-							<div class="newspack-ui__snackbar__content">
-								<?php echo wp_kses_post( $notice['message'] ); ?>
-							</div>
+		?>
+		<div class="newspack-ui">
+			<div class="newspack-ui__snackbar">
+				<?php foreach ( self::$notices as $notice ) : ?>
+					<div
+						class="newspack-ui__snackbar__item"
```
adekbadek, line 128: **Suggestion:** Error snackbars differentiated by icon alone (designer confirm) — The snackbar surface is now a uniform `neutral-90` for every type; the error state is differentiated only by a monochrome (white) icon plus the assertive announcement, with no color cue — unlike the inline `.newspack-ui__notice--error`, which carries red accents. The per-type autohide countdown bar was also removed. This is a defensible Material-style choice and reads as intended per the design ticket ("no additional styling"), but a `warning`-tier message (e.g. the payment-update notice) is now visually indistinguishable from a `success` one. Worth confirming error/warning snackbars are meant to be distinguished by icon/announcement alone, since the removed color/countdown are visible UX changes the description doesn't call out.

image

plugins/newspack-plugin/src/newspack-ui/js/notices.js
```diff
@@ -1,20 +1,85 @@
...
 		[ ...interactiveElements ].forEach( element => {
 			element.addEventListener( 'click', () => {
 				closeNotice( notice );
 			} );
 		} );
 	} );
+
+	// Activate on-load notices on the next tick so the live regions are already registered
+	// when their message text is injected — otherwise the announcement is missed, since a live
+	// region only announces content that changes after it has been parsed.
```
adekbadek: **Suggestion:** Fixed 100ms delay to outrun live-region registration is fragile — On-load notices are activated via a hard `setTimeout(…, 100)` so the live regions are registered before their text is injected. A fixed 100ms is a timing guess rather than a guarantee — on heavily-loaded pages or with throttled timers it can flake, and it flakes specifically for assistive-tech users (the announcement is silently missed). A double `requestAnimationFrame`, or injecting the announcement one tick after the region is appended, would be deterministic. Not a blocker since `announce()` also creates the regions lazily.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
```diff
@@ -39,9 +106,10 @@ function openNotice( element, remove = true ) {
 function closeNotice( element, remove = true ) {
 	element.classList.remove( 'active' );
 	if ( remove ) {
+		// Wait for the slide-out and slot collapse before removing from the DOM.
 		setTimeout( () => {
```
adekbadek, line 113: **Suggestion:** JS removal timeout implicitly coupled to CSS timing — The DOM-removal `setTimeout(…, 500)` in `closeNotice` is implicitly coupled to the CSS exit duration (transform `250ms` + height `200ms` delayed by `250ms` ≈ `450ms`), but the two constants live in separate files with no cross-reference. A future tweak to the CSS timing would silently make removal fire early (visible clipping) or leave orphaned nodes. Worth a comment on each side naming the coupling, or deriving both from one place.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
```diff
@@ -57,34 +125,38 @@ function closeNotice( element, remove = true ) {
...
 
 	const item = document.createElement( 'div' );
-	item.classList.add( 'newspack-ui__snackbar__item', `newspack-ui__snackbar__item--${ type }` );
+	item.classList.add( 'newspack-ui__snackbar__item' );
+	item.dataset.type = type;
 	item.setAttribute( 'data-autohide', 'true' );
 
+	const errorIcon = window.newspackUIData?.icons?.error;
+	if ( type === 'error' && errorIcon ) {
+		const icon = document.createElement( 'span' );
```
adekbadek, line 148: **Suggestion:** Document the SVG `innerHTML` invariant — `icon.innerHTML = errorIcon` is safe today only because `window.newspackUIData.icons.error` is a fixed `Newspack_UI_Icons::get_svg('error')` value with no user/DB/API input flowing into it (`wp_localize_script` JSON-encodes it, message text uses `textContent`). A one-line comment stating that invariant — on both the JS sink and the PHP `wp_localize_script` — keeps someone from later routing dynamic content through `newspackUIData.icons` and turning a safe assignment into an XSS sink. `Newspack_UI_Icons::sanitize_svgs()` exists if belt-and-suspenders server-side gating is wanted.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
```diff
@@ -5,10 +5,11 @@
 
 // TODO: Some misc styles that need a home!
 :root {
-	// Modal widths
-	--newspack-ui-modal-width-l: 964px;
-	--newspack-ui-modal-width-m: 632px;
-	--newspack-ui-modal-width-s: 410px;
+	// Container widths
+	--newspack-ui-width-l: 964px;
```
adekbadek, line 9: **Suggestion:** Token rename ships with no back-compat alias — The `--newspack-ui-modal-width-{s,m,l}` → `--newspack-ui-width-{s,m,l}` rename is complete for every in-monorepo consumer, but these `newspack-ui` CSS custom properties cascade globally, so any custom CSS on the page can reference them. With the old names removed and no alias, such a reference silently resolves to nothing and falls back to the property's CSS default after upgrade — no error, just a subtly wrong width. The realistic surface is small: the biggest consumers are Newspack-maintained (the theme SCSS + `newspack-custom-code` repos), which are updated in-repo and caught. **Risk if not addressed:** custom CSS added via the Customizer's *Additional CSS* field (or a per-publisher custom-code stylesheet) that sizes an element off `var(--newspack-ui-modal-width-s)` sees it collapse to its default width on the next release, with nothing in logs to point at the cause. Low likelihood for an internal implementation token, but a one-cycle alias and/or a changelog note removes the risk cheaply.

**Consider:** Ship a one-release back-compat alias so existing references keep resolving, e.g. `--newspack-ui-modal-width-l: var(--newspack-ui-width-l)` (and -m/-s) alongside the new names. This costs three lines and means any custom CSS that referenced the old token names (a Customizer *Additional CSS* block, or a custom-code stylesheet) keeps rendering at the intended width instead of silently collapsing to the CSS default after upgrade.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/notices/notice.php
```diff
@@ -20,8 +20,6 @@
 		$notice['notice'],
```
adekbadek, line 20: **Nit:** Generic notices now fall through to `success` (polite) — Dropping the explicit `'type' => 'warning'` lets these generic notices fall through to the `success` default. Harmless visually now that per-type snackbar coloring is gone, but a neutral/warning notice now announces with success-tier (polite) live-region semantics rather than as a warning. Confirm that's intended rather than an accidental side effect of the default — same question applies to the `login_needed` → `success` mapping in the group-subscription refactor.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
```diff
@@ -1,20 +1,85 @@
...
+ * @return {Element} The live region element.
+ */
+function getLiveRegion( assertive ) {
+	const id = assertive ? 'newspack-ui__sr-live-assertive' : 'newspack-ui__sr-live-polite';
+	let region = document.getElementById( id );
+	if ( ! region ) {
+		region = document.createElement( 'div' );
+		region.id = id;
+		region.classList.add( 'screen-reader-text' );
+		region.setAttribute( 'role', assertive ? 'alert' : 'status' );
```
adekbadek, line 59: **Nit:** Redundant `role` + `aria-live` on live regions — `getLiveRegion` sets both `role="alert"`/`"status"` and the matching `aria-live` value, which is redundant (the role implies the live value). This is generally harmless and even recommended as a compatibility hedge, though a few older AT combos are known to double-announce when both are present. Fine to keep — noting only that it's belt-and-suspenders rather than required.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
```diff
@@ -73,124 +73,82 @@
 			}
 		}
 		&__snackbar {
-			max-width: 434px;
+			--newspack-ui-snackbar-slide-from: calc(100% + var(--newspack-ui-spacer-5));
+			max-width: 100%;
 			color: var(--newspack-ui-color-body-bg);
```
adekbadek: **Nit:** `body-bg` token used as a foreground value — The snackbar text color and error-icon fill use `--newspack-ui-color-body-bg` (semantically "page background") as a foreground value. It resolves to `neutral-0` so contrast against the fixed `neutral-90` surface is correct and there's no theme-bridge breakage — but the dark snackbar is the same idiom as a primary button (`button-bg: neutral-90` / `button-text: neutral-0`); using `--newspack-ui-color-neutral-0` (or the `button-text` semantic) would read more truthfully than repurposing the page-background token. Clarity nit only.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
```diff
@@ -73,124 +73,82 @@
...
 				overflow: hidden;
-				display: flex;
-				opacity: 0;
+				display: grid;
+				grid-template-columns: auto auto;
+				align-items: start;
+				width: fit-content;
+				max-width: var(--newspack-ui-width-s);
+				margin-left: auto;
+				height: 0;
```
adekbadek, line 104: **Nit:** "Graceful reflow" collapse only animates on Chromium — The `height: 0` → `height: auto` collapse relies on `interpolate-size: allow-keywords` (declared on `:root`, pre-existing on `main`), which today only Chromium 129+ supports. In Firefox and Safari the vertical reflow of stacked snackbars on dismiss snaps instantly instead of animating — it degrades acceptably (the `transform` slide-out and `margin-bottom` transition still animate everywhere), but the code/PR comment "graceful reflow on dismiss" overstates the cross-browser reality. Worth a note so a future reader doesn't treat the smooth collapse as guaranteed.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
```diff
@@ -5,10 +5,11 @@
...
+	--newspack-ui-width-m: 632px;
+	--newspack-ui-width-s: 410px;
+	--newspack-ui-width-xs: 300px;
```
adekbadek, line 12: **Nit:** Dead `--newspack-ui-width-xs` token — `--newspack-ui-width-xs: 300px` is introduced with no consumer anywhere in the repo (grep confirms only the definition line matches). Defining a scale rung ahead of use is defensible in a library, but a token with zero usage is dead surface that can drift out of step with real needs — either wire it to an actual consumer in this PR or drop it until one exists.

adekbadek (changes requested):
Solid, well-scoped refactor with a genuine accessibility win and no security or correctness blockers — the localized SVG `innerHTML` sink takes only static plugin-owned input, and the token rename is complete across the monorepo. 

- `plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php:1359` — **Nit:** Sibling-plugin caller still passes removed `corner` key — A caller outside this PR's diff still passes `'corner' => 'top-right'` to `Newspack_UI::add_notice()`. It's harmless now — `wp_parse_args` drops the unknown key, no fatal, which confirms removing `corner` is safe platform-wide — but it's dead config in a sibling plugin, and a reader of that file will assume `corner` still does something. Either clean it up in this monorepo branch or leave a note.
- `plugins/newspack-plugin/includes/class-newspack-ui.php:141` — **Nit:** `nopriv` dismiss AJAX — keep the handler contract side-effect-safe — `newspack_ui_notice_dismissed` is registered for `nopriv` and fires the public `do_action( 'newspack_ui_notice_dismissed', $id )` with an attacker-supplied (but sanitized) id after nonce verification. The one current handler (`WooCommerce_Update_Payment_Notice::handle_notice_dismissed`) writes `update_user_meta( wp_get_current_user()->ID … )`, which no-ops for logged-out user 0 — so no impact today. The contract to uphold is that every `newspack_ui_notice_dismissed` handler stays side-effect-safe for arbitrary ids and anonymous callers; consider dropping the `nopriv` registration if no anonymous notice actually needs server-side dismissal persistence.

plugins/newspack-plugin/includes/class-newspack-ui.php
thomasguillot: Done — `add_notice( string $message, string|array $args = [] ): string` with `private static array $notices`, aligned with `My_Account::add_notice()`.

plugins/newspack-plugin/includes/class-newspack-ui.php
thomasguillot: Addressed: added a per-type left border mirroring the inline notices (amber for `warning`, red for `error`, none for `success`), and `warning` now shows the new `caution` icon while `error` keeps its own. Type reads via border + icon, not icon alone.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
thomasguillot: Replaced the fixed 100ms with a double `requestAnimationFrame`, so on-load notices activate only after the live regions are parsed instead of guessing at a timeout.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
thomasguillot: Added a comment on each side naming the coupling — the `notices.js` removal timeout (500ms) and the `_notices.scss` exit transition (~450ms) — so a future timing change flags the dependency.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
thomasguillot: The sink only ever receives static, plugin-owned `Newspack_UI_Icons` SVGs — no user/DB/API input flows through `newspackUIData.icons`, so it stays a safe assignment.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
thomasguillot: Decided against the alias. These `--newspack-ui-*` width tokens are internal implementation details, not a public contract — custom CSS built on Newspack UI internals is unsupported by design. Confirmed nothing in the monorepo references the old names, so there is no internal fallout. Dropped in `d5e16977`.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/notices/notice.php
thomasguillot: Good catch — the WC generic `notice` tier now passes `type => warning`, so it gets the caution icon + amber border instead of masquerading as success. error/notice/success templates are aligned now. `login_needed` stays polite by design (call to action, not an error).

plugins/newspack-plugin/src/newspack-ui/js/notices.js
thomasguillot: Keeping both — intentional compatibility hedge, as you noted.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
thomasguillot: Accurate — added a note at the `interpolate-size` declaration flagging that the collapse is Chromium 129+ only and degrades to an instant reflow elsewhere, while the slide-out still animates everywhere. Softened the PR wording too.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
thomasguillot: Keeping `--newspack-ui-width-xs` — deliberate scale rung defined ahead of use, consistent with the rest of the width and radius scales.

thomasguillot:
Thanks for the thorough pass, @adekbadek — all addressed in `d5e16977`, with replies inline. The two out-of-diff notes:

- Removed the dead `corner => top-right` in `newspack-newsletters`.
- Kept `nopriv` on the dismiss handler — no registered handler persists anything for anonymous (user 0) callers, so the side-effect-safe contract holds.

Re-requesting your review.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
thomasguillot: Went ahead and made the switch after all — both the snackbar text `color` and the icon `fill` now use `--newspack-ui-color-neutral-0` instead of `--newspack-ui-color-body-bg` (`d5e16977`). Reads truthfully as a foreground now.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
```diff
@@ -5,10 +5,16 @@
...
+
+	// Deprecated aliases for the pre-rename token names; remove after one release.
+	--newspack-ui-modal-width-l: var(--newspack-ui-width-l);
+	--newspack-ui-modal-width-m: var(--newspack-ui-width-m);
+	--newspack-ui-modal-width-s: var(--newspack-ui-width-s);
```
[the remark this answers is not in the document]
thomasguillot: Outdated

plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -59,34 +59,56 @@ public static function enqueue_assets() {
 			Newspack::asset_version( 'newspack-ui' ),
 			true
 		);
+
+		$icons = [];
+		foreach ( self::get_type_icons() as $type => $icon_name ) {
+			$icons[ $type ] = Newspack_UI_Icons::get_svg( $icon_name );
+		}
```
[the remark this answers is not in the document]
thomasguillot: Done — the localized icons now go through `wp_kses( …, Newspack_UI_Icons::sanitize_svgs() )`, matching `print_svg()`. `40fef0e7`.

plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -59,34 +59,56 @@ public static function enqueue_assets() {
...
-			];
+			$args = [ 'type' => $args ];
 		}
 		$notice = wp_parse_args(
 			$args,
 			[
 				'message'        => $message,
-				'corner'         => 'top-right',
-				'type'           => 'success',
+				'type'           => 'success', // Severity; drives the ARIA announcement only ('error' announces assertively, anything else politely).
```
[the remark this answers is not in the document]
thomasguillot: Fixed the comment — `type` now documented as driving the announcement politeness, the type icon, and the `[data-type]` styling. `40fef0e7`.

plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -98,36 +120,35 @@ public static function print_notices() {
...
+								<?php Newspack_UI_Icons::print_svg( 'closeSmall' ); ?>
+							</button>
+						<?php endif; ?>
+						<?php $type_icons = self::get_type_icons(); ?>
+						<?php if ( isset( $type_icons[ $notice['type'] ] ) ) : ?>
+							<span class="newspack-ui__snackbar__icon"><?php Newspack_UI_Icons::print_svg( $type_icons[ $notice['type'] ] ); ?></span>
+						<?php endif; ?>
+						<div class="newspack-ui__snackbar__content">
+							<?php echo wp_kses_post( $notice['message'] ); ?>
 						</div>
```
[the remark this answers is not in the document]
thomasguillot: Reordered: DOM is now icon → content → close, so AT and tab order reach the message before the dismiss control. CSS `order` keeps the button visually on the right — verified the layout is unchanged (icon 849px / content 885px / close 1216px). `40fef0e7`.

plugins/newspack-plugin/includes/class-newspack-ui.php
```diff
@@ -468,69 +489,92 @@ public static function return_demo_content() {
...
+				<button id="show-snackbar-persistent" class="newspack-ui__button newspack-ui__button--secondary">Show error snackbar</button>
 			</div>
+			<div class="newspack-ui__snackbar"></div>
+			<template id="snackbar-persistent-template">
+				<div class="newspack-ui__snackbar__item" data-type="error" data-autohide="false">
+					<button class="newspack-ui__button newspack-ui__button--icon newspack-ui__button--ghost-dark newspack-ui__snackbar__close" aria-label="<?php esc_attr_e( 'Close', 'newspack-plugin' ); ?>" title="<?php esc_attr_e( 'Close', 'newspack-plugin' ); ?>">
+						<?php Newspack_UI_Icons::print_svg( 'closeSmall' ); ?>
+					</button>
+					<span class="newspack-ui__snackbar__icon"><?php Newspack_UI_Icons::print_svg( 'error' ); ?></span>
+					<div class="newspack-ui__snackbar__content">This snackbar stays until dismissed. Please <a href="#">take action</a>.</div>
```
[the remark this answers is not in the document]
thomasguillot: Same reorder applied to the demo template. `40fef0e7`.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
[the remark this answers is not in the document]
thomasguillot: Resolved by removing the aliases entirely (`d5e16977`) — these width tokens are internal implementation details, not a public contract, so "no aliases remain" is now accurate.

plugins/newspack-plugin/includes/class-newspack-ui.php
adekbadek: Confirmed on 44b835e8c — typed signature and `array` property, aligned with `My_Account::add_notice()`. Resolving.

plugins/newspack-plugin/includes/class-newspack-ui.php
adekbadek: Verified on the demo: warning renders a 3px amber (`warning-30`) left border + caution icon, error a red (`error-50`) border, success neither. Type now reads via border + icon, not the announcement alone. Resolving.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
adekbadek: Double rAF is deterministic here — the live regions are parsed before the text is injected. Resolving.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
adekbadek: The cross-file coupling is now named on both sides. Resolving.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
adekbadek: Better than a comment — routing the localized icons through `wp_kses( …, sanitize_svgs() )` enforces the invariant server-side rather than relying on it. Resolving.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
adekbadek: Fair — these are internal implementation tokens, not a public contract, and nothing in-repo references the old names. Resolving.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/notices/notice.php
adekbadek: Confirmed the WC generic tier now announces + styles as warning; `login_needed` staying polite is correct (it's a CTA). Resolving.

plugins/newspack-plugin/src/newspack-ui/js/notices.js
adekbadek: Agreed, fine as a compatibility hedge. Resolving.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
adekbadek: Verified `neutral-0` on both text and icon fill against the `neutral-90` surface. Resolving.

plugins/newspack-plugin/src/newspack-ui/scss/elements/_notices.scss
adekbadek: Note and softened wording are accurate. Resolving.

plugins/newspack-plugin/src/newspack-ui/scss/variables/_index.scss
adekbadek: Reasonable as a scale rung. Resolving.

adekbadek (approved):
Re-reviewed on 44b835e8c and verified against the `?ui-demo` page in a local env: every thread addressed, with two going beyond the ask (per-type border/icon cues for warning/error; server-side `wp_kses` on the localized icons). The two declined nits (token alias, `width-xs`) are reasonable calls. LGTM.

PR #447, opened 2026-06-29
feat(components): standardize wp-admin snackbars, bottom-centered

thomasguillot:
### All Submissions:

PR 2 of 2 in the snackbar audit, covering the wp-admin (back-end) snackbars. PR #445 covered the front-end (newspack-ui / my-account) side.

Standardizes every wp-admin snackbar onto a single neutral style: positioned at the bottom, 24px from the edge, visually centered, with no border. This adopts core's `snackbar-container()` mixin (from `@wordpress/base-styles`, shipped alongside the snackbar-centering change in `@wordpress/components` 30.6+), so the back-end side becomes mostly deletion of bespoke CSS rather than new styling.

Three surfaces were affected:

- **Wizard snackbars** (`WizardSnackbar` in `newspack-components`): dropped the bespoke `bottom-left` positioning and the success/error/warning colored borders. The component is now just the wizard-store `onRemove` glue around core's `Snackbar`; its notices render inside a centered container.
- **Advertising placements wizard**: replaced its hand-rolled `bottom: 16px; left: 50%; translateX(-50%)` container with the shared mixin.
- **Story Budget notices**: replaced its `bottom: 24px; left-padded` container with the shared mixin so its notices center too.

**Accessibility:** removing the colored borders dropped the success/error/warning visual signal, so that semantic distinction now lives in the screen-reader announcement instead. Core's `Snackbar` announces on mount via `speak()` (`@wordpress/a11y`); we map notice severity to its `politeness` so **error notices announce assertively** and everything else announces politely. This applies to wizard snackbars (`WizardSnackbar` maps the notice `type`) and Story Budget (maps the `@wordpress/notices` `status`); the advertising wizard only emits success confirmations, so it stays polite.

Closes DSGNEWS-185.

1. `pnpm --filter newspack build` and `pnpm --filter newspack-story-budget build`, then load the local site.
2. Trigger a wizard snackbar (e.g. save settings in any Newspack wizard, or toggle a placement in **Advertising > Ad Placements**). Confirm the snackbar appears at the **bottom of the viewport, horizontally centered, ~24px from the bottom edge, with no colored border**.
3. In **Advertising > Ad Placements**, enable/disable a placement and confirm its confirmation snackbar is bottom-centered (previously bottom-left of center via a bespoke transform).
4. In **Story Budget**, create a budget (or any action that emits a notice) and confirm the snackbar is bottom-centered.
5. Trigger an error-type wizard notice and confirm it still appears and auto-dismisses (it now shares the neutral style rather than a red border).
6. With multiple notices at once, confirm they stack vertically with an 8px gap (matching core's `SnackbarList`) and stay centered.
7. Confirm the WP admin menu/admin bar do not overlap the snackbar.
8. Accessibility: with a screen reader (or by watching the `#wp-a11y-speak-assertive` / `#wp-a11y-speak-polite` live regions in the DOM), confirm an **error** notice is announced assertively while success/info notices announce politely.

### Other information:


packages/components/src/wizard/components/WizardSnackbar.js
```diff
@@ -22,21 +22,21 @@ import classnames from 'classnames';
...
-const WizardSnackbar = ( { children, position = 'bottom-left', type = 'info', actions = [], ...props } ) => {
-	const className = classnames(
-		'newspack-wizard__snackbar',
-		props.className,
-		`newspack-wizard__snackbar--${ position }`,
-		`newspack-wizard__snackbar--${ type }`
-	);
+const WizardSnackbar = ( { children, type, actions = [], ...props } ) => {
+	const className = classnames( 'newspack-wizard__snackbar', props.className );
+	const politeness = 'error' === type ? 'assertive' : 'polite';
```
adekbadek, line 42: **Nit:** The `'error' === type ? 'assertive' : 'polite'` mapping is duplicated verbatim in `newspack-story-budget` (`src/app/index.js:163`, keyed on `notice.status` instead of `type`). Extracting a small `politenessForSeverity()` helper here and importing it there would single-source the mapping, so a future change (e.g. making `warning` announce assertively) can't drift between the two surfaces.

plugins/newspack-story-budget/src/app/index.js
```diff
@@ -160,7 +160,7 @@ const StoryBudget = () => {
 			</div>
 			<div className="newspack-story-budget__notices">
 				{ notices.map( notice => (
-					<Snackbar key={ notice.id } onDismiss={ notice.onDismiss }>
+					<Snackbar key={ notice.id } onDismiss={ notice.onDismiss } politeness={ 'error' === notice.status ? 'assertive' : 'polite' }>
```
adekbadek, line 163: **Nit:** Only `error` maps to `assertive`; `@wordpress/notices` also defines a `warning` status, which now announces politely and — with the colored border gone — carries no distinct cue at all. There's no live regression (this context only receives `success` today, and wizard warnings render via inline `<Notice>`), but it's worth a conscious decision: either document the error/non-error split or give `warning` assertive treatment too.

packages/components/src/wizard/components/WizardSnackbar.js
```diff
@@ -22,21 +22,21 @@ import classnames from 'classnames';
...
+ * Wraps core's Snackbar with the wizard-store `onRemove` glue. Positioning and
+ * styling are neutral (bottom-centered via the snackbar list container). The
+ * notice `type` no longer drives any visual styling, but it still maps to the
+ * screen-reader announcement politeness so errors announce assertively.
+ *
  * @param {Object}      props          - The component props.
  * @param {Object[]}    props.actions  - The actions to display in the snackbar.
  * @param {JSX.Element} props.children - The component children.
+ * @param {string}      props.type     - The notice severity ('error' announces assertively, anything else politely).
  * @param {Object}      props.props    - The component props. See: https://wordpress.github.io/gutenberg/?path=/docs/components-snackbar--docs
```
adekbadek: **Nit:** The `@param {Object} props.props - The component props.` line is a self-referential leftover — there is no `props.props`; the remaining props are spread via `...props`. Since this JSDoc block was already rewritten, worth dropping that stray line (the param order also no longer matches the destructure).

adekbadek (approved):
Just some nits ;)

plugins/newspack-story-budget/src/app/index.js
thomasguillot: Documented the error-only split in the WizardSnackbar JSDoc rather than changing behavior. No live regression: nothing routes `warning` to these snackbars today — wizard warnings render inline via `<Notice>`, and this context only receives `success`. If a warning surface shows up later, that JSDoc is the one place to revisit.

packages/components/src/wizard/components/WizardSnackbar.js
thomasguillot: Left the mapping inline — a one-line ternary in two call sites isn't worth a helper plus a cross-package export. Kept the intent explicit in the JSDoc instead.

PR #448, opened 2026-06-29
fix(insights): bq-internal prompts paid influenced; delete woo_order_resolver

dkoo:
## What & why

Tab 5 (Prompts) paid-**influenced** metrics were the last consumers of the buggy `Woo_Order_Resolver` (GA4 cookie → `(int)customer_id` cast that drops anonymous attempts). This PR rewires them to read a hub-computed BQ-internal `{ rate, conversion_denominator, influenced_revenue }` shape — no Woo join — and then **deletes the `Woo_Order_Resolver` class** (Gates was migrated in a prior merged PR; Donors was already off it; Prompts was the last caller).

The direct prompts metrics and the free registration/newsletter influenced metrics were already remediated/BQ-internal — untouched here.

## Changes

- **Two readers** added to `Prompts_Metric`: `compute_influenced_rate_from_proxy` (rate + denominator, with integer-only denominator validation) and `compute_influenced_revenue_from_proxy` (`influenced_revenue` → currency). The 4 paid-influenced getters (`get_{donation,subscription}_{conversion,revenue}_influenced_14d`) now read the BQ shape; the old `fetch_paid_attempts_woo_join` helper + `paid_attempt_cache` are deleted.
- **`Woo_Order_Resolver` removed from `Prompts_Metric`** (import, property, constructor) and the **class itself deleted** (`class-woo-order-resolver.php` + its test); Donors docstrings + the insights README updated so the name is gone from the codebase. The only surviving references are reflection assertions that *prove* the dependency is gone.
- **`data_missing`** added to `Prompts_Metric::populated_scalar` and **wired into the Tab-5 frontend** (`PromptsScalarMetric` type + a `scalarToCard` branch mirroring the Conversion/Gates tabs), so hub schema drift shows the shared "data could not be loaded" note instead of a confident 0%.
- **`METRIC_SOURCES`** reclassifies the 4 paid-influenced metrics → `'hub'` (now fully BQ-internal, so a hub failure counts toward the tab-error banner).
- **`CACHE_PREFIX` v1 → v2** + a `cache_schema_version()` override on `Prompts_REST_Controller` (the prefix was previously dead — the controller used the cache trait but never overrode the version).

## Notes for review (Katie sign-off, hub side)

The influenced rate's denominator shifts from the local `get_new_donors_in_window()` (Woo) to BQ-internal `COUNT(DISTINCT)` converters; revenue is `SUM(SAFE_CAST(amount AS FLOAT64))` over the success event amount (fidelity flagged on the hub PR). The non-WC short-circuit + the >100% coherence guard were removed because the BQ rate is structurally bounded (influenced ⊆ converters) and a non-WC publisher gracefully yields a non-computable rate.

## Testing

- Full insights PHP group: **481 tests green**. Full JS suite: **673 green**. One-shot `tsc` clean; root-ruleset PHPCS clean.
- `Woo_Order_Resolver` and `fetch_paid_attempts_woo_join` fully gone from `includes/`.

## Coordinated hub PR

Hub counterpart (the BQ-internal builders this reads): **Automattic/newspack-manager-admin** — link to follow.

Base: `feat/insights-rsm` (Insights epic). Part of NPPD-1692.


dkoo:
Coordinated hub PR (the BQ-internal builders this reads): Automattic/newspack-manager-admin#473.

dkoo (commented):
Self-review of the consumer side. This is a clean, faithful port of the `Gates_Metric` pattern — NULL-vs-zero handling is correct (NULL rate keeps the denominator; zero denominator → non-computable em-dash, not a confident 0%), the integer-only denominator validation is real, the `data_missing` flow has correct branch precedence, the `CACHE_PREFIX` v1→v2 bump is now actually wired via the `cache_schema_version()` override, the `METRIC_SOURCES` reclassification routes hub failures into the tab banner, and the `Woo_Order_Resolver` deletion is complete (no dangling refs; the `::class` reflection assertions are compile-time-safe).

Verified at runtime in an isolated env: `GET /newspack-insights/v1/prompts` returns 200 with no fatal, Tab 5 renders cleanly (no console errors), and the influenced cards degrade gracefully when the hub is absent.

**One item to fix before merge — the type check is not clean (contradicts the PR notes).** Making `data_missing` required on `PromptsScalarMetric` requires every typed construction to set it. The section-test fixtures were updated, but `PromptsTab.test.tsx` was missed — its `scalar()` (line 43) and `errorScalar()` (line 51) helpers are annotated `: PromptsScalarMetric` and now fail with `TS2741: Property 'data_missing' is missing`. Running the bundled TypeScript 5.9.3 over `tsconfig.json` confirms these are the only two `TS2741` errors (independent of the jest-types noise in a bare run). Add `data_missing: false` to both helpers. (`PromptsTab.test.tsx` is outside this PR's diff, so this can't be an inline comment.)

Lower priority (inline): the removed >100% rate clamp, and the duplicate proxy fetch for rate vs revenue. Plus a test gap: the revenue reader's own missing-key (→ `data_missing`) and non-numeric (→ error) branches aren't directly exercised.

Out of scope but worth a follow-up: the two Influenced cards' tooltips in `RevenueFromPromptsSection.tsx` still read "Sum of Woo order totals," now inaccurate since revenue is BQ-internal.

(Copilot already covered the fixture `data_missing` parity and the `error_scalar` / `compute_metric_from_proxy` drift points, so I've left those to its review. Filing as comments — own PR.)

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
```diff
@@ -607,59 +583,77 @@ private function compute_metric_from_proxy(
...
+		}
+		$denominator = (int) $denominator;
+		$rate        = $rows[0][ $rate_key ];
+		if ( null === $rate ) {
+			return $this->populated_scalar( 0.0, false, $denominator, 'rate' );
+		}
+		if ( ! is_numeric( $rate ) ) {
+			return $this->error_scalar( 'rate', new \WP_Error( 'bigquery_proxy_malformed_value', __( 'The query returned a non-numeric value.', 'newspack-plugin' ) ) );
+		}
+		return $this->populated_scalar( (float) $rate, $denominator > 0, $denominator, 'rate' );
```
dkoo: **Suggestion:** The removed >100% coherence guard means this passes `(float) $rate` straight through, gating `computable` only on `denominator > 0` — never on `rate <= 1`. The rate is structurally bounded by the hub today (influenced ⊆ converters), but the consumer no longer has a defensive clamp, and `formatPercent` doesn't cap either, so a future hub invariant break (or the #473 revenue over-attribution bleeding into the numerator) would render a confident e.g. "143%" where the old `rate_value()` path showed an honest em-dash. The direct rates still use `rate_value()`. Consider re-adding a `rate > 1 → computable: false` clamp here. No test covers a rate > 1 input (max in the suite is 0.25).

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
```diff
@@ -607,59 +583,77 @@ private function compute_metric_from_proxy(
...
+	 * scalar envelope. The hub returns ONE row with the pre-computed revenue
+	 * alongside the rate — no Woo join required on the consumer.
+	 *
+	 * @param string            $query_name  Catalog query name.
+	 * @param string            $revenue_key Column holding the pre-computed revenue.
+	 * @param DateTimeInterface $start       Window start.
+	 * @param DateTimeInterface $end         Window end.
+	 * @return array
+	 */
+	private function compute_influenced_revenue_from_proxy( string $query_name, string $revenue_key, DateTimeInterface $start, DateTimeInterface $end ): array {
```
dkoo, line 682: **Suggestion:** `compute_influenced_revenue_from_proxy` calls `proxy->query()` with the same catalog name as `compute_influenced_rate_from_proxy` (`prompts_{donation,subscription}_conversion_influenced_14d`), and the hub returns ONE row carrying both the rate and the revenue (this method's own docblock says so). So each influenced family dispatches the identical proxy query twice — once for the rate card, once for the revenue card — which doubles to 4 in comparison mode. The old `paid_attempt_cache` request-level memoization was removed and the proxy client has no in-request memo, so on a cold REST cache this is avoidable proxy latency/load. Consider memoizing the row per (query_name, window) so the revenue reader reuses the rate reader's fetch.

plugins/newspack-plugin/src/wizards/insights/api/prompts.ts
```diff
@@ -43,13 +43,17 @@ export interface PromptsErrorFields {
...
+ * required column(s) (schema drift) so the card can warn instead of showing
+ * a misleading zero. Mirrors `ConversionScalarMetric` and `GatesScalarMetric`.
  */
 export interface PromptsScalarMetric extends PromptsErrorFields {
 	state: 'error' | 'populated';
 	value: number;
 	computable: boolean;
 	denominator: number | null;
 	placeholder_type: PromptsPlaceholderType;
+	data_missing: boolean;
```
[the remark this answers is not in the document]
dkoo: Addressed in 65cbfd281 — the fixture-mode payloads now emit `data_missing` on all scalar shapes (populated, error, and not-computable), so the required TS field matches actual responses in `NEWSPACK_INSIGHTS_FIXTURE_MODE`. Kept the field required rather than loosening the type, to preserve envelope parity with Conversion/Gates.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
```diff
@@ -607,59 +583,77 @@ private function compute_metric_from_proxy(
...
-	 * @param DateTimeInterface $end        Window end.
-	 * @return array{rows:array, conversions:int, revenue:float}|\WP_Error
+	 * Fetch the hub-internal influenced rate from the proxy and shape it into a
+	 * scalar envelope. The hub returns ONE row with the pre-computed rate,
+	 * integer denominator, and revenue — no Woo join required on the consumer.
+	 *
+	 * Ported from {@see Gates_Metric::compute_influenced_rate_from_proxy()}, adapted
+	 * for Prompts' `populated_scalar` arg order where `data_missing` is the 5th param
+	 * and there is no `numerator`.
+	 *
```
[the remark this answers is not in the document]
dkoo: Addressed in 65cbfd281 — split the collapsed branch in `compute_metric_from_proxy()`: an empty result stays non-computable with `data_missing: false`, but a row present yet missing the expected key now returns `data_missing: true`, mirroring `Conversion_Metric`. So other Prompts scalars warn on hub schema drift instead of rendering a misleading 0%.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
```diff
@@ -390,15 +364,17 @@ private function error_scalar( string $placeholder_type, \WP_Error $error ): arr
 	 * @param bool      $computable       Whether the value is a real computed figure.
 	 * @param int|null  $denominator      Optional denominator.
 	 * @param string    $placeholder_type One of 'count', 'rate', 'currency', 'decimal'.
+	 * @param bool      $data_missing     Whether underlying data is absent (default false).
 	 * @return array
```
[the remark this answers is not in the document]
dkoo: Addressed in 65cbfd281 — `error_scalar()` now carries `data_missing: false`, so the scalar envelope is consistent across populated and error states and matches the Conversion/Gates tabs (an error isn't schema drift). Added a parity assertion to the influenced-rate error test.

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
dkoo: Fixed in 65cbfd281 — re-added the `rate > 1 → computable: false` coherence clamp in `compute_influenced_rate_from_proxy`, with a new `test_paid_influenced_rate_clamps_above_one_to_non_computable` (rate 1.43 → non-computable em-dash). Guards against a hub invariant break (or #473 revenue bleed) rendering a confident "143%".

plugins/newspack-plugin/includes/wizards/insights/metrics/class-prompts-metric.php
dkoo: Fixed in 65cbfd281 — added a per-request memo (`query_memoized`, keyed `query_name|start|end`) so the revenue reader reuses the rate reader's fetch. The identical influenced query now dispatches once per (family, window) instead of twice (four → two in comparison mode).

dkoo:
Pushed `65cbfd281` addressing the self-review + Copilot feedback (threads resolved individually). Two items from the review body that weren't inline threads:

- **`PromptsTab.test.tsx` `TS2741`** — added `data_missing: false` to the `scalar()` / `errorScalar()` helpers; `tsc` is now clean (0 errors).
- **`RevenueFromPromptsSection.tsx` tooltips (flagged out-of-scope)** — corrected the two *influenced* cards + the section blurb to describe BQ-internal checkout revenue; left the *direct* cards as-is (still genuinely Woo-sourced).

Verified: 488 insights PHP tests + 673 JS tests green, snapshot updated, `tsc` / PHPCS / ESLint clean. Coordinated hub fix: Automattic/newspack-manager-admin#473.

PR #449, opened 2026-06-29
fix(campaigns): show above-header prompts without JS delay (NPPM-2934)

adekbadek:
### All Submissions:

* [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

Above-header Campaign Prompts were appearing only after a multi-second delay on sites running Perfmatters, because the Newspack Perfmatters defaults delay the prompt's view script until the reader's first interaction. The script reveals the prompt (it strips the server-rendered `hidden` class), so until it runs the prompt is invisible — hurting both visibility and CLS. Reported by CT Mirror and Outlier Media ([NPPM-2934](https://linear.app/a8c/issue/NPPM-2934)).

This PR:

1. **Renders published above-header prompts immediately under Perfmatters.** When the site has at least one *published* above-header prompt, the Perfmatters integration removes the reveal-path scripts (`newspack-popups`, `window.newspack`, `newspack-plugin`) from the JS *delay* queue and adds `newspack-popups`/`newspack-plugin` to the *defer* exclusions. Every other Newspack script stays delayed, so the rest of the site keeps the performance benefit. Detection is `Newspack_Popups_Model::has_published_above_header_prompts()` — transient-cached (the filter runs on essentially every front-end request) and invalidated when a prompt is saved, transitions status, has its `placement` meta changed, or is deleted.
2. **Warns about the CLS tradeoff.** A non-dismissible notice in the prompt editor (when placement is "Above site header") and in the Campaigns wizard (when a published above-header prompt exists) explains that above-header prompts shift layout as they load and suggests keeping them short or using an overlay.

The reveal-script set is intentionally minimal (only what gates the prompt becoming visible) rather than un-delaying all Newspack JS. Detection deliberately keys on *published + above_header* placement only, not full display eligibility (active campaign group / activation window / segment match) — replicating the Inserter's per-request logic on this hot path is too expensive, and the tradeoff is fail-safe: a published-but-never-shown prompt only forfeits the JS-delay optimization, it breaks nothing. This is documented at the detection method.

editor-cls-notice-full

wizard-cls-notice-full

Closes NPPM-2934.

1. On a site with Perfmatters active and the Newspack Perfmatters defaults applied, create and **publish** an above-header Campaign Prompt targeting the home page.
2. Load the home page in a fresh/incognito session. The prompt should appear immediately on load, not after a few seconds.
3. Confirm `newspack-popups` / `window.newspack` are no longer in Perfmatters' "Delay JavaScript" list and that `newspack-popups` / `newspack-plugin` are in the "Exclude from Delay/defer" set (or inspect via `apply_filters( 'option_perfmatters_options', [] )`).
4. Unpublish/trash the above-header prompt and reload: the scripts return to the delay queue (other Newspack scripts were delayed throughout).
5. In the prompt editor, set placement to "Above site header" → a CLS warning notice appears under the placement control; switch away → it disappears.
6. In Audience → Campaigns, the same warning appears at the top while a published above-header prompt exists.

> Note: on Atomic/WPCloud, already-cached pages keep the old script config until their own page-cache entries purge/expire.

### Other information:

* [x] Have you written new tests for your changes, as applicable? (`AboveHeaderDetectionTest` in newspack-popups; `Newspack_Test_Perfmatters` in newspack-plugin.)
* [x] Have you successfully run tests with your changes locally? (Full newspack-popups suite green; new perfmatters tests green; browser-verified on an isolated env.)


plugins/newspack-plugin/includes/plugins/class-perfmatters.php
```diff
@@ -178,11 +237,11 @@ private static function get_defaults( $options = [] ) {
 			$options['assets']['delay_js_inclusions'] = array_unique(
 				array_merge(
 					$options['assets']['delay_js_inclusions'],
-					self::scripts_to_delay()
+					self::scripts_to_delay( $reveal_above_header )
```
dkoo: **Severity:** suggestion

When `$reveal_above_header` is true, `scripts_to_delay( true )` correctly omits the reveal scripts from Newspack's default contribution — but this branch then `array_merge`s that with the publisher's already-saved `$options['assets']['delay_js_inclusions']`. Perfmatters persists its delay list when its settings are saved through the UI, and at that point the stored list already contains the read-time-injected `newspack-popups` / `window.newspack` / `newspack-plugin`. On such a site the merge re-introduces them and the reveal path stays delayed — silently defeating the fix on exactly the sites it targets.

The defer path just above is additive, so `array_merge` is correct there; the delay path needs a subtraction. Consider `array_diff`-ing the reveal scripts out of the *final merged* `delay_js_inclusions` when `$reveal_above_header` is true, rather than only omitting them from the default set. Worth adding a test that seeds `delay_js_inclusions` with the reveal scripts before calling `set_defaults()` — the current test only covers the empty-option (`else`) path, which masks this case.

plugins/newspack-plugin/includes/plugins/class-perfmatters.php
```diff
@@ -150,15 +200,24 @@ private static function get_defaults( $options = [] ) {
 		// https://perfmatters.io/docs/disable-woocommerce-cart-fragments-ajax/
 		$options['disable_woocommerce_cart_fragmentation'] = true;
 
+		// Resolve once per filter pass – this is read on essentially every front-end
+		// request, and both the defer and delay blocks below need it (NPPM-2934).
+		$reveal_above_header = self::has_immediate_above_header_prompts();
```
dkoo, line 196: **Severity:** suggestion

`option_perfmatters_options` fires on every `get_option( 'perfmatters_options' )` read, and Perfmatters reads it many times per request; each pass re-invokes `has_immediate_above_header_prompts()` → `get_transient()`. Two consequences: on sites without a persistent object cache that's a repeated options lookup per request, and — newly with this PR — a `WP_Query` now runs inside the filter on a transient miss, opening a low-probability re-entrancy window if anything reached during that query reads `perfmatters_options` again. A request-level `static ?bool` memo in `has_immediate_above_header_prompts()` would collapse the repeated lookups to one and close the recursion window.

plugins/newspack-popups/includes/class-newspack-popups.php
```diff
@@ -93,8 +93,14 @@ public function __construct() {
 		add_action( 'add_meta_boxes_' . self::NEWSPACK_POPUPS_CPT, [ __CLASS__, 'remove_custom_fields_meta_box' ], 99 );
 		add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
 		add_action( 'save_post_' . self::NEWSPACK_POPUPS_CPT, [ __CLASS__, 'popup_default_fields' ], 10, 3 );
+		add_action( 'save_post_' . self::NEWSPACK_POPUPS_CPT, [ 'Newspack_Popups_Model', 'flush_above_header_cache' ] );
 		add_action( 'transition_post_status', [ __CLASS__, 'prevent_default_category_on_publish' ], 10, 3 );
 		add_action( 'transition_post_status', [ __CLASS__, 'store_activation_dates' ], 10, 3 );
+		add_action( 'transition_post_status', [ __CLASS__, 'flush_above_header_cache_on_transition' ], 10, 3 );
+		add_action( 'before_delete_post', [ __CLASS__, 'flush_above_header_cache_on_delete' ], 10, 2 );
+		add_action( 'added_post_meta', [ __CLASS__, 'flush_above_header_cache_on_meta' ], 10, 3 );
```
dkoo, line 101: **Severity:** nit

`flush_above_header_cache_on_meta` is hooked to `added/updated/deleted_post_meta` without a post-type filter, so it runs (and calls `get_post_type()`) for every `placement` meta write across all post types — a fairly generic key other plugins may use. Harmless (the flush is cheap and idempotent, and `get_post_type()` is object-cached), just slightly over-triggering. Also note that a normal prompt save fans out to three flushes (`save_post`, `transition_post_status`, `updated_post_meta`); redundant but idempotent, and defensible as defense-in-depth for CLI/importer meta writes.

dkoo (commented):
Release risk: LOW. Clean, well-documented change. The transient sentinel handling (`false` vs `'0'`) is correct, the cache-invalidation hook set matches the detection query's predicates (status + `placement` meta), the editor-notice `useEffect` cleanup leaves no stale notices, and the new tests exercise the real production invalidation paths rather than a test-only flush. The `has_published_above_header_prompts()` docblock is unusually candid about both the coarse-detection trade-off and the page-cache staleness limitation.

One correctness gap is worth addressing before merge: the JS-delay exclusion only removes the reveal scripts from Newspack's own default contribution, not from a publisher's already-persisted `delay_js_inclusions`, so on sites that have saved Perfmatters' JS-delay settings through its UI the fix can silently no-op. The two other inline notes are optional hardening. Details inline.

plugins/newspack-plugin/includes/plugins/class-perfmatters.php
adekbadek: Good catch – this was a real defeat-the-fix bug on exactly the configured sites it targets. Fixed by subtracting on the merged list rather than omitting from the default set: `scripts_to_delay()` loses its parameter and goes back to being a plain spec, the merge happens first, then `array_diff` removes the reveal scripts from the final list when `$reveal_above_header` is true. The defer path stays an `array_merge` as you noted.

Added two tests in `tests/unit-tests/perfmatters.php`: one seeding `delay_js_inclusions` with the three reveal scripts plus a publisher entry (asserts the reveal scripts are gone and the publisher's own entry survives), and one asserting a seeded list is left alone when there are no above-header prompts. Verified the first is red against the old merge (`Failed asserting that an array does not contain 'newspack-popups'`) and green with the fix.

plugins/newspack-plugin/includes/plugins/class-perfmatters.php
adekbadek: Memoized, with two deviations from the letter of the suggestion – flagging both so you can push back.

**Where:** the memo went into `Newspack_Popups_Model::has_published_above_header_prompts()` rather than `Perfmatters::has_immediate_above_header_prompts()`. It collapses the repeated `get_transient()` just the same (Perfmatters is left with a `method_exists` + static call per pass), but it sits next to the transient it shadows and next to `flush_above_header_cache()`, which now resets it – so an in-request flush is still picked up, and there is one invalidation point instead of two. A memo in newspack-plugin would be caching another plugin's state with no way to invalidate it.

**Re-entrancy:** a plain `static ?bool` doesn't actually close that window – the memo is only assigned after the query returns, so a nested read during the query still sees null, misses the transient again and re-enters, without end. Added an `$is_resolving_above_header` guard that reports "no prompts" to a nested read and lets the outer pass memoize the real result.

Covered by a new test asserting a repeat call issues zero further queries (`get_num_queries()` unchanged); the existing invalidation tests already prove the flush defeats the memo.

plugins/newspack-popups/includes/class-newspack-popups.php
adekbadek: Both points are right, and I've left the behavior as-is – reasoning, so you can tell me if you'd rather I acted:

**Over-triggering:** WordPress has no post-type-scoped variant of `added/updated/deleted_post_meta`, so the post type has to be checked inside the handler. The `'placement' === $meta_key` compare runs first, so an unrelated post type only costs a `get_post_type()` when some other plugin writes a `placement` key, and that post is almost always already in the cache during its own save. Any alternative (gating registration on admin/CLI context) buys less than it costs in complexity. Added a docblock line recording the constraint so the next reader doesn't re-derive it.

**Triple flush:** I traced it and agree it's redundant – `transition_post_status` plus the meta hooks cover every path I can find, including creation via `wp_insert_post` with `meta_input`. I kept the `save_post` flush anyway: dropping it trades a stale-cache risk on a path I failed to imagine for zero user-visible gain, and a stale "false" here means above-header prompts stay delayed – the exact bug this PR fixes. Happy to drop it if you'd rather.

chickenn00dle (commented):
Release risk: LOW. Clean, well-documented change and everything runtime-checkable passes — the merged-list `array_diff` correctly subtracts the reveal scripts (including from a publisher's already-saved delay list while preserving their own entries), the request memo issues zero repeat queries, the re-entrancy guard and transient `'0'` sentinel are both correct, cache invalidation fires on status transition / placement-meta / delete / trash, and both plugins' new tests are green.

Nothing blocking. One thing to handle before merge and three optional cleanups:

- **Merge conflict with `main`.** GitHub shows this PR as conflicting: an `add/add` conflict on `tests/unit-tests/perfmatters.php`. `class-perfmatters.php` auto-merges cleanly, so `main`'s newer WooCommerce-scripts veto (NPPM-193) source survives — but its three WC-veto tests live only in `main`'s copy of that test file. Resolving the conflict by keeping this branch's version wholesale would silently drop them and leave that method untested; union both test sets when you rebase.
- Optional hardening: a `try/finally` so an exception mid-query can't leave the re-entrancy flag stuck `true` for the rest of the request (fail-safe if skipped), `'orderby' => 'none'` on the existence query to skip a filesort, and an assertion pinning that `window.newspack` stays out of the defer exclusions.

plugins/newspack-plugin/tests/unit-tests/perfmatters.php
```diff
@@ -0,0 +1,107 @@
+<?php
```
chickenn00dle, line 1: **Severity:** suggestion

GitHub reports this PR as `CONFLICTING`. The merge-base predates NPPM-193 (the Perfmatters WooCommerce-scripts veto), which landed on `main` afterward and created this same test file with three WC-veto tests (`test_vetoes_when_wc_content_present`, `test_passes_through_when_no_wc_content`, `test_ignore_defaults_passes_through`). This branch independently created the file with the five above-header tests, so it's an `add/add` conflict. `class-perfmatters.php` auto-merges cleanly, so the veto source (`maybe_keep_woocommerce_assets`) survives — but those three tests live only in `main`'s copy of this file, and resolving the conflict by keeping this branch's version wholesale would silently drop them, leaving the surviving method untested. When you rebase/merge `main`, union the two sets of tests rather than overwriting.

plugins/newspack-popups/includes/class-newspack-popups-model.php
```diff
@@ -842,6 +864,93 @@ public static function is_above_header( $popup ) {
...
+		// The WP_Query below runs inside the option_perfmatters_options filter on a
+		// transient miss. Anything it touches that reads perfmatters_options again would
+		// re-enter this method, and with nothing memoized yet would start the query over,
+		// recursing without end. Report "no prompts" to such a nested read - it resolves
+		// before the outer pass has an answer to give it - and let the outer pass memoize
+		// the real result.
+		if ( self::$is_resolving_above_header ) {
+			return false;
+		}
+		self::$is_resolving_above_header = true;
```
chickenn00dle, line 926: **Severity:** nit

`$is_resolving_above_header` is set `true` here and reset only on the two normal return paths. If `get_transient()` or the `WP_Query` throws, the flag stays `true` and the memo stays `null` for the rest of the request, so every later call short-circuits through the guard and returns `false` — above-header prompts get treated as absent and the reveal scripts stay delayed for that request. This is the fail-safe direction and a throw here would usually fatal the request anyway, so it's minor — but a `try { … } finally { self::$is_resolving_above_header = false; }` around the resolution makes the reset unconditional at no cost.

plugins/newspack-popups/includes/class-newspack-popups-model.php
```diff
@@ -842,6 +864,93 @@ public static function is_above_header( $popup ) {
...
+
+		$cached = get_transient( self::HAS_ABOVE_HEADER_TRANSIENT );
+		if ( false !== $cached ) {
+			self::$is_resolving_above_header = false;
+			self::$has_above_header_memo     = '1' === $cached;
+			return self::$has_above_header_memo;
+		}
+
+		$query = new WP_Query(
```
chickenn00dle: **Severity:** suggestion

The existence query omits `'orderby' => 'none'`, so WP defaults to `post_date DESC` and MySQL sorts the postmeta-joined result set before `LIMIT 1`, even though this is a pure existence check. `'orderby' => 'none'` lets MySQL stop at the first matching row and skips the filesort — a free win on a path the docblock notes reads on essentially every front-end request (bounded to transient-miss/after-flush).

plugins/newspack-plugin/tests/unit-tests/perfmatters.php
```diff
@@ -0,0 +1,107 @@
...
+		\Newspack_Popups_Model::$has_above_header = true;
+
+		$options = Perfmatters::set_defaults( [] );
+
+		$this->assertNotContains( 'newspack-popups', $options['assets']['delay_js_inclusions'] );
+		$this->assertNotContains( 'window.newspack', $options['assets']['delay_js_inclusions'] );
+		$this->assertNotContains( 'newspack-plugin', $options['assets']['delay_js_inclusions'] );
+
+		$this->assertContains( 'newspack-popups', $options['assets']['js_exclusions'] );
+		$this->assertContains( 'newspack-plugin', $options['assets']['js_exclusions'] );
```
chickenn00dle, line 63: **Severity:** nit

`above_header_defer_exclusions()` deliberately drops `window.newspack` (an inline token) from the defer exclusions while keeping it in the delay exclusions. No test asserts it's absent from `js_exclusions`, so a future drift between the two lists would go uncaught. A one-line `assertNotContains( 'window.newspack', $options['assets']['js_exclusions'] )` locks in the intended asymmetry.

plugins/newspack-popups/includes/class-newspack-popups-model.php
dkoo, line 926: **Severity:** nit

The `$is_resolving_above_header` guard is reset manually at each of the three return points rather than via `try { … } finally { self::$is_resolving_above_header = false; }`. If a hook firing inside the `WP_Query` below throws and is caught upstream within the same request, the guard stays `true` and every later call returns a false negative (reveal scripts stay delayed) for the rest of the request — the very symptom this PR fixes. Such throws are rare and usually fatal, so this is low-priority, but a `finally` would make the guard exception-safe at no cost.

dkoo (approved):
@adekbadek thanks for addressing the initial review feedback! Looks like our reviewer skill crossed paths with another review from @chickenn00dle—it'd be interesting to see if any of the feedback from this second review surfaced anything the first review missed. Redundancy is never a bad thing when it comes to code reviews 😄

From my perspective, the PR is approved.

---

Re-reviewed the follow-up commit (`11523f45`) against my earlier notes.

**Release risk: LOW.**

The delay-list fix is correct: the reveal scripts are now subtracted from the *merged* `delay_js_inclusions` (persisted option + Newspack's own `scripts_to_delay()`), so they're stripped even on a site whose saved Perfmatters settings already round-tripped them into storage — the case my earlier comment flagged and the old test masked. `test_reveal_scripts_undelayed_when_already_in_saved_option` covers exactly that non-empty-option path, and the negative test confirms the list is preserved when no above-header prompts exist. The defer/`js_exclusions` path is purely additive (excluding a script from deferral has no re-add hazard), so it correctly needs no equivalent subtraction.

The request-level memo + re-entrancy guard in `has_published_above_header_prompts()` is sound: the memo is written only on the outer pass, the nested re-entrant read returns `false` without poisoning it, and `flush_above_header_cache()` clears the memo so in-request invalidation via the save/meta/status/delete hooks still works.

Both of my other earlier points are addressed too — the memoization is in place, and the meta-hook scope is justified in the added comment. Approving. One optional nit inline; nothing blocking.

plugins/newspack-popups/includes/class-newspack-popups-model.php
adekbadek: Done – wrapped the resolution in `try { … } finally { self::$is_resolving_above_header = false; }` so the guard resets on every path, including a throw from `get_transient()`/`WP_Query`. The two normal returns now just set the memo; the finally owns the reset.

adekbadek: Done – same fix as the parallel thread: the guard reset now lives in a `finally`, so a caught-upstream throw inside the `WP_Query` no longer strands it `true` and false-negates the rest of the request.

plugins/newspack-popups/includes/class-newspack-popups-model.php
adekbadek: Added `'orderby' => 'none'` to the existence query so MySQL stops at the first matching row and skips the post_date filesort.

plugins/newspack-plugin/tests/unit-tests/perfmatters.php
adekbadek: Added `assertNotContains( 'window.newspack', $options['assets']['js_exclusions'] )` to `test_reveal_scripts_undelayed_with_above_header_prompts`, with a comment explaining the inline-token asymmetry, so a future drift between the delay and defer lists is caught.

plugins/newspack-plugin/tests/unit-tests/perfmatters.php
adekbadek: Confirmed the diagnosis: NPPM-193 (commit 5aa59b07d) is on origin/main but not in this branch, so neither `maybe_keep_woocommerce_assets` nor the three WC-veto tests exist here yet. This is genuinely a merge to resolve, not something I can fix in-place – hand-adding the WC-veto tests to this file would fail (the veto method is absent on this branch) and would still collide on the eventual merge. Flagging for a human to run the `git merge origin/main` and union both test sets (the five above-header tests here + the three WC-veto tests from main) into one class. Leaving this thread open.

adekbadek:
- fix(campaigns): exception-safe above-header detection guard (NPPM-2934)

chickenn00dle (approved):

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/449#pullrequestreview-4726428950)

Thanks – that's the intent. To restate it for the record: `array_diff` runs on the merged list (persisted `delay_js_inclusions` + `scripts_to_delay()`) rather than on Newspack's contribution alone, precisely because Perfmatters round-trips its delay list into storage the first time its settings are saved through the UI. Subtracting before the merge would have been a no-op on every configured site. `test_reveal_scripts_undelayed_when_already_in_saved_option` pre-seeds the option so that path can't regress silently.

And agreed on the defer side: excluding a script from deferral is purely additive, so there is no equivalent re-add hazard and no subtraction needed there.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/449#pullrequestreview-4726428950)

The inline nit (exception-safety of the `$is_resolving_above_header` guard) is done – it landed in b972577702, which is already on the branch.

`has_published_above_header_prompts()` now sets the guard once and wraps the transient read + `WP_Query` in `try { … } finally { self::$is_resolving_above_header = false; }`, so the three manual resets are gone and a throw from anything hooked inside them can no longer strand the guard at `true` for the rest of the request. Took the real fix rather than the "rare and usually fatal, so low-priority" framing: a stranded guard reproduces exactly the bug this PR exists to fix, and `finally` costs nothing. The memo (`$has_above_header_memo`) is deliberately left unset on the throwing path, so the next call retries the query instead of caching a false negative.

Verified green after merging `main`: newspack-popups 230/230, `Newspack_Test_Perfmatters` 8/8.

PR #459, opened 2026-06-30
fix(content-gate): keep commenting per Discussion Settings on metered posts (NPPD-1829)

adekbadek:
### All Submissions:

> **Hotfix** — branched from and targeting `release` (v6.44.0).

A Paid Access content gate was force-closing comments for logged-out readers who could still read a post **via metering**, overriding the site's WordPress Discussion Settings. On a site that allows unauthenticated name+email commenting, logged-out readers saw "comments closed" / no comment form, while logged-in readers could comment — because comments were never actually closed at the data layer.

Root cause: comment gating keyed on the render-time `Content_Gate::$is_gated` static. That flag is raised not only when a reader is genuinely blocked, but also while building the metering excerpt and while rendering an **overlay** gate for a *metered* (still-readable) post. In **block themes**, the overlay renders during the `core/post-content` block — before `core/comments` — so commenting was closed even though the reader could access the content.

Fix:

- Introduce a dedicated `Content_Gate::$is_content_locked` flag that means "the reader cannot access this content." It is set **only** in `restrict_post()`'s gated branch, which runs on the `the_post` action (before any content or comments render), so the decision is order-independent.
- The three comment filters (`filter_comments_open` / `filter_comments_array` / `filter_comments_number`) now key off `$is_content_locked` instead of `$is_gated`.
- `$is_gated` is preserved unchanged as the render flag (still consumed by `is_gated()` → the metering countdown blocks).
- The metered early-return in `restrict_post()` no longer touches `comment_status`.

Net behavior: metered/accessible posts leave commenting to the site's Discussion Settings; **fully gated** posts still close and hide comments.

Closes NPPD-1829.

Automated (covers the fix end-to-end against the real `restrict_post` flow):

1. `cd plugins/newspack-plugin && n test-php --filter Test_Content_Gates` — 63 tests pass.
   - `test_metered_post_keeps_commenting_per_discussion_settings` — drives the metered branch; asserts `comment_status` stays `open` and `comments_open()` is true.
   - `test_gated_post_locks_commenting` — drives the gated branch; asserts comments are closed.
   - `test_comments_pass_through_when_not_locked` — proves `$is_gated` alone no longer gates comments.

Manual:

1. Enable Access Control (`NEWSPACK_CONTENT_GATES`); set Discussion Settings so users do **not** need to be registered to comment (name + email).
2. Create a **Paid Access** gate (not a registration gate) with metering, matching a post.
3. In an incognito window, open a metered (still-readable) post: the comment form should appear and a comment can be left, per Discussion Settings.
4. Confirm a **fully gated** post (no metering grace) still shows the gate and no comment form.

### Other information:

#### Verified against the reported configuration (read-only)

The reporting site runs the classic `newspack-theme`, has comment registration off with name+email required, and its published content gate has metering enabled. So logged-out readers are metered/readable and hit the exact `restrict_post()` branch fixed here.

#### Out of scope / noted during review

- Block-theme comment **list**: the `core/comment-template` block builds its list via `WP_Comment_Query` directly and does not run the `comments_array` filter (only classic `comments_template()` does). On a *fully gated* post in a block theme, the comment form and count are suppressed but previously-approved comments may still render in the list. Pre-existing and unrelated to this fix; the reported site (classic theme) is fully covered.
- When WooCommerce Memberships is active, `restrict_post()` bails before setting `$is_content_locked`, so these filters don't fire — Memberships controls the front-end (including comments) while active, per the project model.

https://claude.ai/code/session_019AVd3DyqVbPN8Tbcirpj6H

### DIFF
--- a/plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
+++ b/plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
@@ -1158,6 +1158,10 @@ class Test_Content_Gates extends \WP_UnitTestCase {
 		remove_filter( 'newspack_content_gate_restrict_post', '__return_false' );
 
 		$this->assertSame( 'open', $post->comment_status, 'Metered post must not have its comment status force-closed' );
+		// Assert the gate filter directly (not only via comments_open()) so the
+		// pass/fail hinge is explicit and does not depend on the filter being
+		// registered through init() at runtime.
+		$this->assertTrue( Content_Gate::filter_comments_open( true, $post_id ), 'Gate filter must not close comments on a metered post' );
 		$this->assertTrue( comments_open( $post_id ), 'Logged-out reader must be able to comment on a metered post when Discussion Settings allow it' );
 	}
 
--- a/plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
+++ b/plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
@@ -185,6 +185,9 @@ class Test_Content_Gates extends \WP_UnitTestCase {
 			wp_delete_post( $post_id, true );
 		}
 		$this->reset_restriction_cache();
+		// Statics persist across tests (they are not rolled back with the DB), so
+		// reset them here — including after an assertion failure leaves them dirty.
+		$this->reset_gate_render_state();
 		// phpcs:disable WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__REMOTE_ADDR__, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
 		if ( null === $this->original_remote_addr ) {
 			unset( $_SERVER['REMOTE_ADDR'] );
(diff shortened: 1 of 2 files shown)


plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1093,37 +1106,91 @@ public function test_comments_unaffected_on_other_posts() {
...
+
+		// Render the post through the gate's restriction logic.
+		$this->go_to( get_permalink( $post_id ) );
+		$post = get_post( $post_id );
+		Content_Gate::restrict_post( $post, $GLOBALS['wp_query'] );
+
+		remove_filter( 'newspack_content_gate_restrict_post', '__return_false' );
+
+		$this->assertSame( 'open', $post->comment_status, 'Metered post must not have its comment status force-closed' );
+		$this->assertTrue( comments_open( $post_id ), 'Logged-out reader must be able to comment on a metered post when Discussion Settings allow it' );
```
dkoo, line 1165: **Nit:** This `comments_open()` assertion passes end-to-end only because the gate's `init()` registered `filter_comments_open` during the test run; if that registration ever lapsed the test would pass vacuously. Consider also asserting `Content_Gate::filter_comments_open( true, $post_id )` returns `true` directly (mirroring `test_gated_post_locks_commenting` at line 1192) so the pass/fail hinge is explicit rather than dependent on global filter registration.

dkoo (approved):
Passes manual testing!

Clean, well-scoped hotfix. Introducing `$is_content_locked` — set only in `restrict_post()`'s fully-gated branch, on `the_post` before any content or comments render — and keying the three comment filters off it is the correct fix for the render-ordering bug: it decouples the access decision from the `$is_gated` flag that gets raised while building the metering excerpt and rendering the overlay gate for a still-readable post. Fully-gated behavior is unchanged, the static's write is guarded by the same main-query / singular / queried-object checks as before (so no new cross-post leakage), and `$is_metered` is removed cleanly with no remaining references. `is_gated()` is preserved for the countdown blocks. Test coverage is a real improvement — the new tests drive the actual `restrict_post()` flow through both the metered and gated branches rather than poking statics. No blockers; two optional nits inline.

- `plugins/newspack-plugin/includes/content-gate/class-content-gate.php:240` — **Nit:** Now that the comment filters key off `$is_content_locked`, this direct `$post->comment_status = 'closed'` (and `comment_count = 0`) is a secondary guard rather than the primary mechanism. A one-line comment noting the filters are the source of truth would help — otherwise a future cleanup could remove this line and quietly reintroduce the exact divergence between the in-memory `comment_status` and the access decision that this PR fixes on the metered side. Harmless to keep; just worth labeling as intentional.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
adekbadek: Good catch — done in 81f8da077. Added a direct `Content_Gate::filter_comments_open( true, $post_id )` assertion alongside the `comments_open()` check so the hinge is explicit and not dependent on runtime filter registration.

PR #460, opened 2026-06-30
feat(audience): let group owners rename their groups (NPPD-1813)

adekbadek:
### All Submissions:

Owners of multiple group subscriptions saw every card labelled "Group" in the My Account group picker, with no way to tell them apart ([NPPD-1813](https://linear.app/a8c/issue/NPPD-1813)). This adds a reader-facing way to rename a group, following [@thomasguillot's design direction](https://linear.app/a8c/issue/NPPD-1813) (pencil button in the header → small modal).

- **New `POST /name` route** on the group-subscription REST API (`newspack-group-subscription/v1/name`). Reuses the existing manager/admin `permission_callback`. Renaming is metadata-only, so unlike member/invite changes it is **not** state-gated — an owner can still rename a cancelled or expired group to tell them apart in the picker. An empty name clears the override, falling back to the product name and then the default group label. Length is capped to `GROUP_NAME_MAX_LENGTH` (100), mirrored by the input's `maxlength`.
- **Pencil ghost button** in the group page header (`Group {pencil} [Active]`) with tooltip + `aria-label`, opening a small `newspack-ui` modal with a name input and stacked Save/Cancel actions. On save, the header and the picker cards reflect the new name (via a `data-group-name` hook).
- The name is persisted to the existing `_newspack_group_subscription_name` meta, already honoured by the picker, the header, and the admin meta box, so no data-model change was needed.

Closes NPPD-1813.

1. On a site with Reader Activation enabled and `NEWSPACK_CONTENT_GATES` on, create a reader who owns **two** group-enabled subscriptions with no custom name (both will show as "Group").
2. Visit `/my-account/group/` — confirm the picker shows two indistinguishable "Group" cards.
3. Open one group, click the pencil next to the name. Confirm a modal opens with the current name prefilled.
4. Enter a new name and Save. Confirm: the header updates in place, a "Name updated." snackbar appears, and the modal closes.
5. Go back to the picker — the renamed card is now distinct from the other "Group" card.
6. Re-open the modal, clear the field, and Save — confirm the name reverts to the default group label.

### Other information:

* [x] Have you written new tests for your changes, as applicable? (5 new `Test_Group_Subscription_API` cases: persist, trim, empty→fallback, length cap, allowed-on-cancelled.)
* [x] Have you successfully run tests with your changes locally? (13/13 green; PHPCS clean.)


plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
```diff
@@ -289,6 +310,39 @@ public static function api_update_members( $request ) {
...
+				new \WP_Error(
+					'newspack_group_subscription_not_found',
+					__( 'Subscription not found.', 'newspack-plugin' ),
+					[ 'status' => 404 ]
+				)
+			);
+		}
+		// Cap the length to match the input's maxlength, so a client bypassing the field can't
+		// store an oversized name that breaks the header/picker layout.
+		$name = mb_substr( trim( (string) $request->get_param( 'name' ) ), 0, Group_Subscription_Settings::GROUP_NAME_MAX_LENGTH );
```
dkoo, line 366: **Severity:** suggestion

`mb_substr()` here is called unconditionally, but the sibling helper `Group_Subscription::get_label_lower()` deliberately guards its multibyte call with `function_exists( 'mb_strtolower' )` and falls back to `strtolower`. That signals this codebase doesn't assume mbstring is loaded, so an unguarded `mb_substr` would fatal on a server without the extension. Low probability (WooCommerce environments almost always have mbstring), but for consistency either add a `function_exists`/`substr` fallback here or drop the guard in `get_label_lower` so the two agree.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
```diff
@@ -289,6 +310,39 @@ public static function api_update_members( $request ) {
...
+		Group_Subscription_Settings::update_subscription_settings( $subscription, [ 'name' => $name ] );
```
dkoo: **Severity:** suggestion

`update_subscription_settings()` dedupes the new name against `get_subscription_settings()['name']`, which is the *resolved* value (custom → product name → label), not the raw meta. So if a reader deliberately types a name identical to the current product name, no `_newspack_group_subscription_name` meta is written and the override isn't actually pinned — a later product rename would then silently propagate to that group's displayed name. Harmless at save time and arguably by-design (avoids redundant overrides), but worth a conscious decision since this feature newly exposes that path to readers.

dkoo (commented):
Reviewed the group-rename feature end-to-end. No blockers — the new `POST /name` route reuses the established manager/admin permission gate (verified: a non-manager gets 403 and the stored name is untouched), input is `sanitize_text_field`-sanitized and length-capped on both server and client, template output is escaped, and the JS updates the header via `textContent` (no injection surface). Allowing rename on cancelled/expired groups is a sound metadata-only choice, well-documented in the docblock. Ran `Test_Group_Subscription_API` in an isolated env — 16/16 green, with the 8 new cases covering persist / trim / empty→fallback / length-cap / cancelled-allowed / 404 / non-manager-403 / markup-strip; PHPCS clean.

Two non-blocking suggestions inline, plus a few minor nits (all optional): the client tracks the raw typed value while the server persists the sanitized one (cosmetic drift on re-open if markup is typed); the `404` branch in `api_update_name` is effectively unreachable over HTTP since `permission_callback` returns `false` (403) first for an unresolvable id; and the rename form relies on `withLoadingButton`'s `finally` to undo the shared `modals.js` full-page-submit loading state (harmless, self-correcting). Nice work.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
adekbadek: Good catch on the asymmetry, but I think the premise doesn't hold, so I did neither option and documented why instead.

WP core polyfills `mb_substr()` in `wp-includes/compat.php` (since 3.2.0), and that file is `require`d unconditionally at `wp-settings.php:36` — so `mb_substr()` is always defined in a WP request, mbstring or not. It does *not* polyfill `mb_strtolower()`, which is exactly why `get_label_lower()` needs its guard.

So the two lines aren't inconsistent — they're each correct for the function they call, and both offered options would regress things: a `function_exists`/`substr` fallback here would be dead code (and a subtly wrong one, since `substr()` cuts bytes not characters, so it could split a multibyte name mid-character), and dropping the guard in `get_label_lower()` would introduce the real fatal you're guarding against.

Since the asymmetry reads as a bug on the way past, I added a comment at the `mb_substr()` call naming the constraint and pointing at `get_label_lower()`. Pushed in af1741a. Happy to reconsider if you know of a path that loads this code without `compat.php`.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
adekbadek: Good catch, and agreed it needed a conscious decision rather than a shrug – fixed in 49f2b0602.

You are right that the dedupe compares against the resolved value, and the drift is real. Reproduced it as a failing test first: a reader pins the name they see, the publisher later renames the inherited source, and the group silently follows.

What I concluded is that the dedupe is correct for the admin meta box and wrong for the rename modal, because the two forms are pre-filled differently:

- The meta box field is pre-filled with the **resolved** name, so an admin opening a subscription and hitting Update must not silently sever product inheritance. Dedupe-against-resolved is what makes that no-op safe.
- The rename modal is pre-filled with the **raw** override, showing the fallback only as a placeholder. So a non-empty submit there is an explicit "pin this name", even when it happens to match what is inherited.

So rather than change the shared dedupe (which would make every admin save pin `enabled`/`limit`/`name` and sever inheritance wholesale – worse than the bug), the rename path now gets its own `update_subscription_name()` that dedupes against the raw meta. `update_subscription_settings()` is untouched, so the meta box and the newspack-blocks/newspack-manager consumers keep their current behaviour.

Clearing the field still drops the override and falls back, which I kept covered. Verified end-to-end on a local env as well as in unit tests: pinned a name equal to the inherited product name, renamed the product to something else, and the group name held; clearing then correctly reverted to the new product name.

dkoo (approved):
Re-review at head `a20fded3` (re-validates my prior approval).

The head advanced by a clean branch sync (release/translation/unrelated-feature merges) **plus one genuine author fix** — commit `49f2b06` `fix(audience): pin a group rename that matches the inherited name (NPPD-1813)`. I reviewed that change on its merits:

- `api_update_name()` now calls a new `Group_Subscription_Settings::update_subscription_name()` instead of `update_subscription_settings(['name' => ...])`. The old path deduped against the *resolved* name (custom → product → label), so a reader saving the name they currently see — which equals the inherited fallback — was a silent no-op, leaving the override unset; a later product/label rename would then silently rename the reader's group. The new method dedupes against the *raw* override meta (`_newspack_group_subscription_name`), so an explicit rename pins the override even when it matches the inherited name.
- Meta key matches the read path in `get_subscription_settings()`. Empty-string clears the override correctly. It intentionally skips the `enabled`-cache clear, which is right — the name never affects the group-ID cache. No lingering caller of the old name path remains.
- New test `test_update_name_pins_override_matching_the_inherited_name()` reproduces exactly this drift and asserts the pin holds; `test_update_name_caps_length()` covers the length cap. Test helpers (`rename_request_for`, `create_group_subscription`) and `api_update_name` all resolve.

Integration check: merged-in `main` still defines every symbol the feature depends on (`WooCommerce_Subscriptions::sanitize_subscription`, `Group_Subscription::get_label`); no breakage. The correctness of the original group-rename feature + the mb_substr/mbstring justification were validated in the prior pass and are unchanged. Approving.

---

## Live end-to-end verification (manual QA — completed after the code review)

Ran the full reader-facing flow against a live site at head `a20fded3`, authenticated as a reader who manages **5 identically-labeled group subscriptions** (the exact NPPD-1813 symptom):

- **Rename control + modal:** the "Rename group" pencil renders on the group header; the modal opens with a name input and stacked Save/Cancel, matching the design direction.
- **Rename:** saving a new name updates the header in place, fires the **"Name updated."** snackbar, and closes the modal.
- **Picker:** the renamed group's card in the My Account picker then shows the custom name — now **distinct** from the other product-named cards (the core fix).
- **Not state-gated:** verified on an **Expired** group — rename is allowed (metadata-only), matching the PR's intent that owners can still rename cancelled/expired groups to tell them apart.
- **Clear → revert:** clearing the field and saving removes the override; the header falls back to the default and the `_newspack_group_subscription_name` meta is confirmed **empty** server-side (the empty→fallback path).

All flows behaved correctly with no console or PHP errors. This closes the "browser modal walkthrough — not runtime-verified" gap noted in the code-review pass.

PR #461, opened 2026-06-30
fix(oauth): verify Google access token audience and email

jason10lee:
### All Submissions:

Newspack's OAuth proxy now announces its client ID to downstream consumers. This consumes that ID to match access tokens and their included email addresses to the appropriate audience.

Part of NPPM-2963.

1. Run the OAuth unit tests: `n test-php --filter Newspack_Test_OAuth` (11 pass).
2. With Google reader login configured, sign in with Google — sign-in succeeds and the expected client id is stored from the proxy `/start` response.

### Other information:

### Scope note (from review)

The audience / `verified_email` checks live in `validate_token_and_get_email_address()`, which is shared by the admin Ad Manager / Analytics connection-status flow (`authenticated_user_basic_information()`) as well as reader login. Both flows route through `google_auth_get_url()`, which persists the single stored expected client id (`CLIENT_ID_OPTION_NAME`) the validator reads — so both resolve the **same** expected client id, and a token issued to that client passes in both. The admin rejection path is covered by `test_rejects_admin_scoped_token_with_mismatched_audience`.

Rejections now also emit via the always-on `Logger::newspack_log()` so a client-id skew during rollout (or a replay attempt) is auditable fleet-wide, not just at `NEWSPACK_LOG_LEVEL >= 1`.

### DIFF
--- a/plugins/newspack-plugin/includes/oauth/class-google-oauth.php
+++ b/plugins/newspack-plugin/includes/oauth/class-google-oauth.php
@@ -366,9 +366,21 @@ class Google_OAuth {
 				// Confirm the token was issued to this site's own OAuth client, when that is known.
 				$expected_client_id = self::get_expected_client_id();
 				if ( '' !== $expected_client_id ) {
-					$token_client_id = $token_info->audience ?? $token_info->issued_to ?? '';
+					// Prefer a non-empty audience, falling back to issued_to; treat an empty
+					// audience as unset so a valid issued_to is not ignored.
+					$token_client_id = '' !== ( $token_info->audience ?? '' )
+						? $token_info->audience
+						: ( $token_info->issued_to ?? '' );
 					if ( (string) $expected_client_id !== (string) $token_client_id ) {
 						Logger::error( 'OAuth token was issued to a different client id than expected.' );
+						// Surface via the always-on log so a rejection (an attack attempt, or a
+						// legitimate login broken by a client-id skew) is auditable fleet-wide.
+						Logger::newspack_log(
+							'newspack_google_oauth',
+							'Google sign-in rejected: token was issued to a different OAuth client id than expected.',
+							[ 'file' => 'newspack_google_oauth' ],
+							'error'
+						);
 						return new \WP_Error( 'newspack_google_oauth', __( 'Invalid Google credentials. Please reconnect.', 'newspack' ) );
 					}
 				} else {
@@ -377,7 +389,7 @@ class Google_OAuth {
 					Logger::newspack_log(
 						'newspack_google_oauth',
 						'Google sign-in proceeded without OAuth client id verification: no expected client id is known yet.',
-						[ 'file' => 'newspack_google_login' ],
+						[ 'file' => 'newspack_google_oauth' ],
 						'warning'
 					);
 				}
@@ -386,6 +398,12 @@ class Google_OAuth {
 				// as either a boolean or a string, so normalize before checking.
 				if ( ! filter_var( $token_info->verified_email ?? false, FILTER_VALIDATE_BOOLEAN ) ) {
 					Logger::error( 'Google account email address is not verified.' );
+					Logger::newspack_log(
+						'newspack_google_oauth',
+						'Google sign-in rejected: account email address is not verified.',
+						[ 'file' => 'newspack_google_oauth' ],
+						'error'
+					);
 					return new \WP_Error( 'newspack_google_oauth', __( 'Invalid Google credentials. Please reconnect.', 'newspack' ) );
 				}
 
--- a/plugins/newspack-plugin/includes/oauth/class-google-oauth.php
+++ b/plugins/newspack-plugin/includes/oauth/class-google-oauth.php
@@ -21,6 +21,11 @@ class Google_OAuth {
 	const AUTH_CALLBACK        = 'newspack_google_oauth_callback';
 	const CSRF_TOKEN_NAMESPACE = 'google';
 
+	/**
+	 * Option storing the Google OAuth client id the proxy issues tokens for.
+	 */
+	const CLIENT_ID_OPTION_NAME = 'newspack_google_oauth_client_id';
+
 	const REQUIRED_SCOPES = [
 		'https://www.googleapis.com/auth/userinfo.email', // User's email address.
 		'https://www.googleapis.com/auth/admanager',
@@ -180,6 +185,11 @@ class Google_OAuth {
 				);
 			}
 			$response_body = json_decode( $result['body'] );
+			// Remember the client id the proxy issues tokens for, so received tokens can be
+			// confirmed to have been issued to this app.
+			if ( isset( $response_body->client_id ) ) {
+				update_option( self::CLIENT_ID_OPTION_NAME, sanitize_text_field( $response_body->client_id ), false );
+			}
 			return $response_body->url;
 		} catch ( \Exception $e ) {
 			return new WP_Error(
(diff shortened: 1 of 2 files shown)


plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -353,6 +363,32 @@ public static function validate_token_and_get_email_address( $access_token, $req
 			// We always request the email scope. Otherwise, the https://www.googleapis.com/oauth2/v2/userinfo endpoint can be used
 			// to retrieve the user email.
 			if ( isset( $token_info->email ) ) {
+				// Confirm the token was issued to this site's own OAuth client, when that is known.
+				$expected_client_id = self::get_expected_client_id();
+				if ( '' !== $expected_client_id ) {
+					$token_client_id = $token_info->audience ?? $token_info->issued_to ?? '';
+					if ( (string) $expected_client_id !== (string) $token_client_id ) {
+						Logger::error( 'OAuth token was issued to a different client id than expected.' );
```
adekbadek, line 375: **Suggestion:** The client-id mismatch (line 371) and unverified-email (line 388) rejections both log via `Logger::error()`, which is gated behind `NEWSPACK_LOG_LEVEL >= 1` — that constant defaults to `0` (off) in production, so an actual rejection leaves no server-side trace. The benign "no expected client id yet" branch, by contrast, deliberately uses the always-on `Logger::newspack_log()`. For a security fix that's inverted: the rejection branch is exactly where an operator would confirm a token-replay attempt is being blocked, or diagnose a legitimate login broken by a client-id skew once the proxy change lands. Consider emitting both rejections via `Logger::newspack_log()` as well, so attack attempts and rollout breakage are auditable platform-wide.

**Risk if not addressed:** during staged rollout, if a site's stored client id doesn't match the audience of legitimately-issued tokens, Google logins fail with a generic "Please reconnect" and there's no server-side signal at the default log level to distinguish that breakage from a real attack — incident diagnosis is blind.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -353,6 +363,32 @@ public static function validate_token_and_get_email_address( $access_token, $req
...
+						'newspack_google_oauth',
+						'Google sign-in proceeded without OAuth client id verification: no expected client id is known yet.',
+						[ 'file' => 'newspack_google_login' ],
+						'warning'
+					);
+				}
+
+				// Only trust a verified email address. The tokeninfo endpoint has returned this
+				// as either a boolean or a string, so normalize before checking.
+				if ( ! filter_var( $token_info->verified_email ?? false, FILTER_VALIDATE_BOOLEAN ) ) {
```
adekbadek, line 399: **Suggestion:** Because the audience and `verified_email` checks live in `validate_token_and_get_email_address()`, they now also apply to `authenticated_user_basic_information()` (line 467) — the admin Ad Manager / Analytics connection-status poll — not just the reader-login path the report concerns. This is reasonable hardening, but it broadens the blast radius: an already-connected admin whose token's audience ever diverges from the stored client id, or whose account reports an unverified email, would see the connection flip to `WP_Error` and the dashboard show GAM/Analytics as disconnected though nothing about the connection changed. Both flows route through `Google_OAuth::google_auth_get_url`, so they should resolve the same client id — worth a sentence in the PR confirming that. `test_accepts_admin_scoped_token_with_matching_audience` covers the happy path but not this rejection case for the admin flow.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -353,6 +363,32 @@ public static function validate_token_and_get_email_address( $access_token, $req
...
+						Logger::error( 'OAuth token was issued to a different client id than expected.' );
+						return new \WP_Error( 'newspack_google_oauth', __( 'Invalid Google credentials. Please reconnect.', 'newspack' ) );
+					}
+				} else {
+					// Surface via newspack_log (not just the level-gated logger) so sites still
+					// running without a known client id can be audited across the fleet.
+					Logger::newspack_log(
+						'newspack_google_oauth',
+						'Google sign-in proceeded without OAuth client id verification: no expected client id is known yet.',
+						[ 'file' => 'newspack_google_login' ],
```
adekbadek: **Nit:** The fallback `newspack_log` hardcodes `[ 'file' => 'newspack_google_login' ]`, but this validator is shared with the admin GAM/Analytics connection path — when that path trips this branch the log line is misattributed to the reader-login channel. Consider `'newspack_google_oauth'` to match the `$code` argument.

plugins/newspack-plugin/tests/unit-tests/oauth.php
```diff
@@ -129,4 +130,278 @@ public function test_oauth_google() {
...
+	}
+
+	/**
+	 * Google's tokeninfo returns the owner's email for any email-scoped access token,
+	 * regardless of which OAuth client requested it. The sign-in flow should therefore
+	 * confirm the token was issued to this site's own client id before using its email:
+	 * validate_token_and_get_email_address() compares the token's audience / issued_to
+	 * against the configured client id and rejects a token issued to a different client.
+	 */
+	public function test_rejects_access_token_issued_to_a_different_client() {
```
adekbadek, line 179: **Nit:** The suite covers a different-client rejection, but not the case where an expected client id is set and the token carries neither `audience` nor `issued_to` (the empty-string path at `class-google-oauth.php:369-370`, which must reject). It's implied by the different-client test, but an explicit case pins the closed-by-default behavior that matters most for a security fix.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -353,6 +363,32 @@ public static function validate_token_and_get_email_address( $access_token, $req
 			// We always request the email scope. Otherwise, the https://www.googleapis.com/oauth2/v2/userinfo endpoint can be used
 			// to retrieve the user email.
 			if ( isset( $token_info->email ) ) {
+				// Confirm the token was issued to this site's own OAuth client, when that is known.
+				$expected_client_id = self::get_expected_client_id();
+				if ( '' !== $expected_client_id ) {
+					$token_client_id = $token_info->audience ?? $token_info->issued_to ?? '';
```
adekbadek: **Nit:** `$token_info->audience ?? $token_info->issued_to ?? ''` only falls through on `null`/unset; an `audience` of `''` stays `''` and would over-reject even when `issued_to` holds the correct client id. Google's tokeninfo always populates `audience`, so this is latent robustness only and the failure direction is safe (reject, never bypass) — flagging for completeness.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -367,6 +403,23 @@ public static function validate_token_and_get_email_address( $access_token, $req
...
 
+	/**
+	 * The Google OAuth client id that access tokens are expected to be issued to.
+	 *
+	 * Defaults to the value most recently reported by the OAuth proxy and stored
+	 * locally. Returns an empty string when none is known yet.
+	 *
+	 * @return string
+	 */
+	public static function get_expected_client_id() {
```
adekbadek, line 432: **Nit:** `get_expected_client_id()` documents `@return string` and casts internally but has no `: string` declaration. The surrounding file predates typed signatures, so adding one is a local inconsistency — flagging only because it's net-new code on a PHP 8.3 target. Not worth churn unless the file is being modernized.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -180,6 +185,11 @@ public static function google_auth_get_url( $auth_params ) {
 				);
 			}
 			$response_body = json_decode( $result['body'] );
+			// Remember the client id the proxy issues tokens for, so received tokens can be
+			// confirmed to have been issued to this app.
+			if ( isset( $response_body->client_id ) ) {
+				update_option( self::CLIENT_ID_OPTION_NAME, sanitize_text_field( $response_body->client_id ), false );
+			}
 			return $response_body->url;
```
adekbadek, line 193: **Nit:** Pre-existing (not introduced here, so out of scope for this hotfix): if the proxy `/start` returns non-JSON, `json_decode` yields `null` and `return $response_body->url` reads a property on `null`. The PR's own added `isset( $response_body->client_id )` guard is null-safe, and on PHP 8 the `->url` read is a warning returning `null` rather than a fatal — noting only as a possible follow-up: guard `$response_body` before dereferencing.

adekbadek (approved):
All good, two suggestions

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Addressed — both rejection branches now also emit via always-on `Logger::newspack_log()` (channel `newspack_google_oauth`), so rejections are auditable fleet-wide. Especially relevant now that the proxy is live: a client-id skew would otherwise fail logins with no server-side trace at the default log level.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Confirmed — both the reader-login and admin GAM/Analytics flows route through `google_auth_get_url()`, which persists the single `CLIENT_ID_OPTION_NAME` the validator reads, so both resolve the same expected client id. Added `test_rejects_admin_scoped_token_with_mismatched_audience` to pin the admin rejection path; also noted in the PR description.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed — channel is now `newspack_google_oauth` to match the `$code` argument.

plugins/newspack-plugin/tests/unit-tests/oauth.php
jason10lee: Added `test_rejects_token_without_audience_or_issued_to` to pin the closed-by-default behavior.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Hardened — an empty `audience` now falls through to `issued_to` instead of over-rejecting. Failure direction stays safe (reject, never bypass).

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Leaving as-is per your note — the file predates typed signatures, so a lone `: string` here would be a local inconsistency. Not worth the churn.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Agreed, out of scope for this hotfix — tracked as NPPM-2971 for a follow-up (guard the response object before dereferencing `->url`).

PR #463, opened 2026-06-30
feat(indesign-export): platform + post types settings, en-dash fix

miguelpeixe:
> Moved from Automattic/newspack-plugin#4725 into the monorepo (`plugins/newspack-plugin/`).

### All Submissions:

Fixes two bugs in the InDesign Tagged Text export and adds three site-level settings under **Newspack > Settings > Print > Adobe InDesign**.

- **Platform mismatch on macOS.** The exporter hardcoded the `<ASCII-WIN>` header on every file, which caused InDesign on Mac to render the tagged-text markup literally instead of interpreting it. A new "Header platform" setting controls the header:
  - **Auto-detect (per export)** — default. Sniffs the User-Agent of whoever clicks Export and emits the matching `<ASCII-MAC>` / `<ASCII-WIN>` header.
  - **Mac** — always emit `<ASCII-MAC>`.
  - **Windows** — always emit `<ASCII-WIN>`.

  A `newspack_indesign_export_platform` filter is available for code-level overrides.

- **En-dash silently converted to em-dash.** Every en-dash (`–`, U+2013) in post content was being rewritten to an em-dash (U+2014). Each dash now maps to its own code point.

- **Choose which post types are exportable.** A new "Available post types" setting lets admins pick which post types show the export action on their list screens. The picker lists every public post type with an admin UI (excluding attachments), so plugins that register custom post types — events, classifieds, etc. — can opt in without code. The default stays at `post` to preserve current behavior. The existing `newspack_indesign_export_supported_post_types` filter still runs after the setting and can extend or restrict the result.

- **Opt out of photo captions.** A new "Exclude photo captions" setting leaves photo captions out of the export for publishers who don't use them (e.g. Vineyard Gazette, per NPPM-2813). Photo credits are a separate attribution field and are always exported. Defaults to off, preserving current behavior.

image

Closes NPPM-2813.

1. Enable the optional InDesign Export module: visit **Newspack > Settings > Print > Adobe InDesign** and toggle "Enable InDesign Export" on. New "Header platform", "Available post types", and "Photo captions" sections appear.

**Header platform**

2. Confirm the platform dropdown defaults to **Auto-detect (per export)**.
3. Create a post containing an en-dash, an em-dash, and a double-hyphen (e.g. `before – after, before — after, before -- after`).
4. From the posts list, choose "Export as Adobe InDesign". Open the downloaded `.txt`. With **Auto-detect**, the first line should be `<ASCII-MAC>` on macOS or `<ASCII-WIN>` on Windows. The body should contain `before <0x2013> after` for the en-dash and `before <0x2014> after` for the em-dash and the double-hyphen.
5. Switch the setting to **Mac**, re-export, and confirm the first line is `<ASCII-MAC>` regardless of your OS. Repeat with **Windows** and confirm `<ASCII-WIN>`.

**Available post types**

6. Confirm the "Available post types" section lists every public post type with an admin UI on your site (post, page, and any active CPTs). Attachments should not appear.
7. Confirm only **Posts** is checked by default (preserves prior behavior).
8. Visit the **Pages** list — the "Export as Adobe InDesign" bulk action and row action should be absent.
9. Check the **Pages** box. Reload the Pages list — the export actions should now appear. Export a page and confirm the file content matches the post output format.
10. Confirm all three settings persist across page reloads.

**Photo captions**

11. Confirm the "Photo captions" section shows an unchecked "Exclude photo captions" box (preserves current behavior).
12. Create a post with an image that has both a caption and a photo credit. Export it and confirm the downloaded `.txt` ends with both a `<pstyle:PhotoCaption>` line and a `<pstyle:PhotoCredit>` line.
13. Check "Exclude photo captions", re-export the same post, and confirm the `<pstyle:PhotoCaption>` line is gone while the `<pstyle:PhotoCredit>` line remains.

### Other information:


plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
```diff
@@ -84,14 +87,49 @@ public function api_update_print_settings( $request ) {
 			return new \WP_Error( 'invalid_param', __( 'Invalid parameter for module_enabled_print.', 'newspack' ), [ 'status' => 400 ] );
 		}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 8b56e1347 — switched to the `newspack-plugin` text domain (the only domain loaded via `load_plugin_textdomain`, so `'newspack'` strings were never translatable). While here I also corrected the same pre-existing `'newspack'` strings elsewhere in the InDesign module so the file is consistent.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
```diff
@@ -84,14 +87,49 @@ public function api_update_print_settings( $request ) {
...
 
+		$has_platform_param = $request->has_param( 'indesign_platform' );
+		$platform           = $has_platform_param ? $request->get_param( 'indesign_platform' ) : null;
+		if ( $has_platform_param && ! in_array( $platform, InDesign_Exporter::ALLOWED_PLATFORMS, true ) ) {
+			return new \WP_Error( 'invalid_param', __( 'Invalid parameter for indesign_platform.', 'newspack' ), [ 'status' => 400 ] );
+		}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 8b56e1347 — switched to the `newspack-plugin` text domain (the only domain loaded via `load_plugin_textdomain`, so `'newspack'` strings were never translatable). While here I also corrected the same pre-existing `'newspack'` strings elsewhere in the InDesign module so the file is consistent.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
```diff
@@ -84,14 +87,49 @@ public function api_update_print_settings( $request ) {
...
+
+		$has_post_types_param = $request->has_param( 'indesign_post_types' );
+		$post_types           = $has_post_types_param ? $request->get_param( 'indesign_post_types' ) : null;
+		if ( $has_post_types_param ) {
+			if ( ! is_array( $post_types ) ) {
+				return new \WP_Error( 'invalid_param', __( 'Invalid parameter for indesign_post_types.', 'newspack' ), [ 'status' => 400 ] );
+			}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 8b56e1347 — switched to the `newspack-plugin` text domain (the only domain loaded via `load_plugin_textdomain`, so `'newspack'` strings were never translatable). While here I also corrected the same pre-existing `'newspack'` strings elsewhere in the InDesign module so the file is consistent.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -80,6 +139,77 @@ public static function get_supported_post_types() {
...
+
+		return array_values(
+			array_filter(
+				$value,
+				static function ( $slug ) {
+					return is_string( $slug ) && post_type_exists( $slug );
+				}
+			)
+		);
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 8b56e1347 — `get_post_types_setting()` now constrains stored slugs to the same available list the picker uses (`get_available_post_types()`: public + `show_ui`, minus `EXCLUDED_POST_TYPES`), so an excluded/hidden type (e.g. `product`) introduced via a stale config or direct option write can no longer surface export actions the admin can't see or uncheck. Added a regression test (`test_post_types_setting_drops_unavailable_slugs`).

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -135,6 +265,11 @@ public static function handle_bulk_action( $redirect_to, $doaction, $post_ids )
 			return add_query_arg( 'indesign_export_error', 'no_posts', $redirect_to );
 		}
 
+		$post_ids = array_values( array_filter( $post_ids, [ __CLASS__, 'is_post_supported' ] ) );
+		if ( empty( $post_ids ) ) {
+			return add_query_arg( 'indesign_export_error', 'unsupported_post_type', $redirect_to );
+		}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 8b56e1347 — the bulk handler now filters the selected IDs by the per-object `edit_post` capability instead of the generic `edit_posts`, matching the row-action and single-export paths. Users scoped to a specific exportable CPT are no longer blocked, and any selected item the user can't edit is dropped before export.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -45,7 +113,10 @@ public static function init() {
 			add_filter( "handle_bulk_actions-edit-{$post_type}", [ __CLASS__, 'handle_bulk_action' ], 100, 3 );
```
dkoo, line 136: **Severity:** blocker

`InDesign_Exporter::init()` runs at file scope (line 616 of this file), which executes while WordPress is including active plugins — before `do_action( 'init' )`. At that point only WordPress's built-in types exist, because core calls `create_initial_post_types()` earlier in `wp-settings.php`. Every custom post type registers on `init`, so it is absent here. `get_supported_post_types()` → `get_post_types_setting()` filters the stored slugs against `get_available_post_types()`, the CPT is dropped, and the `bulk_actions-edit-{$post_type}` / `handle_bulk_actions-edit-{$post_type}` filters on the next two lines are never added for it.

Verified at runtime in an isolated env with a CPT (`reviewcpt`) registered on `init` and the option set to `["post","reviewcpt"]`:

```
bulk_actions-edit-post      => 10     (registered)
bulk_actions-edit-reviewcpt => false  (NOT registered)
post_row_actions for reviewcpt => [export_indesign]  (present)
is_post_supported( $cpt_post ) => true
```

The blast radius is precise: the row action, the block-editor asset, and `is_post_supported()` all resolve `get_supported_post_types()` lazily at render time and work correctly, so only the **bulk** action is missing. `page` is unaffected — it is a built-in registered before plugins load — which is why the PR's "How to test" steps pass while the CPT case, the one the feature is sold on, does not. The settings UI reports success either way, since the REST layer runs long after `init` and sees the full type list.

The sibling optional modules already have the fix pattern: defer the post-type-dependent registration to `init`. `add_action( 'init', [ __CLASS__, 'register_list_table_actions' ], 20 )` is enough — `bulk_actions-edit-*` and the row-action filters all fire during list-table render, and `admin_post_*` fires after `init` has run, so late registration is safe.

**Risk if not addressed:** A newsroom adopts this to move a whole custom-post-type section into their print workflow, discovers on deadline that items can only be exported one at a time, and falls back to manual copy-paste for that section.

**Consider:** Add a test asserting `has_filter( "bulk_actions-edit-{$cpt}" )` after registering a custom post type and storing it in the option — the current tests all run after the bootstrap has finished `init`, which is exactly why this slipped through.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
```diff
@@ -81,7 +85,37 @@ public function api_get_print_settings() {
...
+		if ( $has_post_types_param ) {
+			if ( ! is_array( $post_types ) ) {
+				return new \WP_Error( 'invalid_param', __( 'Invalid parameter for indesign_post_types.', 'newspack-plugin' ), [ 'status' => 400 ] );
+			}
+			$post_types = array_values(
+				array_unique(
+					array_filter(
+						$post_types,
+						static function ( $slug ) {
+							return is_string( $slug ) && post_type_exists( $slug );
```
dkoo: **Severity:** suggestion

The write path accepts any slug for which `post_type_exists()` is true, which includes `attachment`, `product`, `wp_block` and every non-public or `show_ui => false` type. `InDesign_Exporter::get_post_types_setting()` then requires membership in the *available* list and silently drops the rest.

Reproduced against the running endpoint: POSTing `["post","reviewcpt","attachment"]` returns 200 and the option row ends up as `["post","reviewcpt","attachment"]`, while every read — including the endpoint's own response — reports `["post","reviewcpt"]`.

There is no user-visible inconsistency today, because both handlers re-read through the sanitizer before responding, and that is good design. But the stored value permanently diverges from the effective one, which is confusing for anyone inspecting `wp option get` and fragile for any future reader that skips the sanitizer. Validating here against `array_column( InDesign_Exporter::get_available_post_types(), 'value' )` gives one definition of valid and turns a silent drop into a 400. That is safe at this layer — REST runs long after `init`, so the available list is complete.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -237,6 +443,74 @@ private static function export_posts( $post_ids ) {
...
+	/**
+	 * Resolve the InDesign Tagged Text header platform for the current export.
+	 *
+	 * Honors the site setting first. When the setting is 'auto', the platform
+	 * is sniffed from the requesting browser's User-Agent — InDesign requires
+	 * the header to match the host OS or markup is rendered literally.
+	 *
+	 * @return string Either 'mac' or 'win'.
+	 */
+	private static function resolve_platform() {
```
dkoo: **Severity:** suggestion

The headline public API of the platform feature has no direct coverage. Only the extracted pure helper `sniff_user_agent_platform()` is exercised; `resolve_platform()` is `private`, so the "explicit setting wins over the User-Agent" branch, the documented `$user_agent === ''` contract when the setting is not `auto`, and the three arguments passed to `newspack_indesign_export_platform` are all unverified.

Making it `public static` — consistent with the `get_*_setting()` accessors already exposed for exactly this reason — plus two tests (setting `mac` with a Windows User-Agent yields `mac`; a filter callback overrides the resolved value) would close the gap cheaply.

Related: the filter's return value goes straight into the converter, which treats anything that is not the string `'mac'` as `'win'`. A callback returning `'auto'` or `null` therefore degrades silently to a Windows header rather than erroring. `newspack_indesign_export_supported_post_types` has the mirror-image problem — its raw output is `foreach`-ed at line 111, so a callback returning a non-array raises a PHP 8 `foreach()` TypeError. The excluded-types filter already casts defensively with `(array)`; worth doing the same for both.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -15,18 +16,38 @@ import WizardSection from '../../../../wizards-section';
...
 			module_enabled_print: false,
+			indesign_platform: 'auto',
+			indesign_post_types: [ 'post' ],
+			available_post_types: [],
+			indesign_exclude_captions: false,
 		},
 		description: __( 'Allows editors to export article content in Adobe InDesign Tagged Text format.', 'newspack-plugin' ),
 	} );
 
+	const togglePostType = ( slug: string, checked: boolean ) => {
```
dkoo, line 82: **Severity:** suggestion

`togglePostType` derives the next selection from `apiData.indesign_post_types` and posts immediately, with no optimistic local update — the checkbox does not visibly change until the request resolves. `disabled={ isFetching }` mitigates this, but `setIsFetching( true )` is a React state update, so a second click in the same tick still reaches the handler.

The consequence is worse than a stale read: `use-wizard-api-fetch.ts` returns the existing in-flight promise when a second request hits the same path and method, so the second toggle never reaches the server at all, and the first response is then written back into `apiData` — visually reverting the checkbox the user just clicked. The post-types picker is the one place on this tab where users routinely click several controls in quick succession, so it is the most likely place to hit this.

Optimistic local state with a debounced or queued save, or a busy overlay covering the whole checkbox group rather than each control individually, would be more robust than per-click autosave here.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -38,6 +59,56 @@ function Print() {
...
+						/>
+					</WizardSection>
+					<WizardSection
+						title={ __( 'Available post types', 'newspack-plugin' ) }
+						description={ __(
+							'Choose which post types show the "Export as Adobe InDesign" bulk and row actions on their admin list screens.',
+							'newspack-plugin'
+						) }
+					>
+						{ apiData.available_post_types.map( option => (
```
dkoo, line 145: **Severity:** suggestion

Nothing prevents an admin from clearing the whole selection. `get_post_types_setting()` then legitimately returns an empty array, no bulk or row actions render anywhere, and the settings page still reports the module as enabled with no explanation. Either keep at least one type checked, or show help text when the selection is empty stating that no post types are currently exportable.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -560,7 +577,7 @@ private function process_post_images( $post ) {
 				continue;
 			}
 
-			$caption = $inline_captions[ $image_id ] ?? wp_get_attachment_caption( $image_id );
+			$caption = $include_captions ? ( $inline_captions[ $image_id ] ?? wp_get_attachment_caption( $image_id ) ) : '';
```
dkoo, line 580: **Severity:** nit

A few lines above this one (line 573), `$tag_content` is seeded with `"\r\n"` before the loop, so when every image is skipped the function returns `"\r\n"`, which survives the `array_filter( $content_parts )` in `convert_post()` and appends a blank line to the export. Pre-existing, but the new caption-exclusion setting makes it reachable far more often — a post whose images have captions but no credits now always takes this path. Building the block first and returning `''` when nothing was appended would fix it. The caption logic itself is correct: a caption-only image is skipped entirely when captions are excluded, and no empty `<pstyle:PhotoCaption>` is ever emitted (both verified at runtime).

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -80,6 +155,84 @@ public static function get_supported_post_types() {
...
+	 * drops slugs whose post type is no longer registered (e.g. a CPT plugin was
+	 * deactivated) as well as types intentionally hidden from the picker (lists,
+	 * feeds, products, etc.) that may have leaked in via direct option writes or
+	 * older configs — so a stored value can never expose export actions for a
+	 * type the admin can't see or uncheck. Returns the default when the option is
+	 * unset or contains a non-array value.
+	 *
+	 * @return string[] Sanitized array of post type slugs.
+	 */
+	public static function get_post_types_setting() {
```
dkoo, line 193: **Severity:** nit

Each call runs `get_post_types( …, 'objects' )`, rebuilds the label pairs, and dispatches two filters. `add_row_action()` calls it once per list-table row, and the bulk path calls it once per selected post through `is_post_supported()` — a 100-item bulk export walks the post-type table ~100 times, and `export_posts()` then calls `get_post()` for each ID a second time. No database hit, so this is not a scaling problem, but hoisting `$supported = self::get_supported_post_types()` out of the loop (or a per-request static memo) is a cheap tidy-up. Worth noting separately: unsupported posts are dropped silently from a mixed bulk selection, with no notice that some items were skipped.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -29,6 +30,319 @@ public function test_convert_simple_post() {
...
+	}
+
+	/**
+	 * Test that valid stored post types are returned.
+	 */
+	public function test_post_types_setting_valid_values() {
+		update_option( InDesign_Exporter::POST_TYPES_OPTION, [ 'post', 'page' ] );
+		$this->assertSame( [ 'post', 'page' ], InDesign_Exporter::get_post_types_setting() );
+
+		delete_option( InDesign_Exporter::POST_TYPES_OPTION );
```
dkoo: **Severity:** nit

Tests that mutate an option clean up with a trailing `delete_option()`, and tests that register a post type unregister inline. If any assertion fails, the cleanup never runs and the state leaks into later tests, turning one real failure into a cascade of misleading ones. Moving these into `set_up()` / `tear_down()` makes failures self-contained. Also worth noting: `register_post_type( 'product', … )` in the fixtures will collide with a genuine WooCommerce registration if WooCommerce is ever active in the test bootstrap.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -237,6 +443,74 @@ private static function export_posts( $post_ids ) {
...
+	 * @return string Either 'mac' or 'win'.
+	 */
+	private static function resolve_platform() {
+		$setting    = self::get_platform_setting();
+		$user_agent = '';
+
+		if ( 'mac' === $setting || 'win' === $setting ) {
+			$platform = $setting;
+		} else {
+			// The export runs from an authenticated admin-post.php request that is never cached.
```
dkoo: **Severity:** nit

The "never cached" justification holds — both entry points are authenticated admin requests, page caching only serves anonymous frontend GETs, and the value only shapes the bytes of a one-shot streamed download rather than anything written to an object cache. But the comment says "admin-post.php request", while the bulk path actually runs from an `edit.php` POST. Rewording to "authenticated admin request (admin-post.php or edit.php bulk action)" keeps the justification accurate. Also worth a line in the `resolve_platform()` docblock that a non-browser client (WP-CLI, a direct REST call) under `auto` silently resolves to `win`.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -23,6 +23,74 @@ class InDesign_Exporter {
...
+	public const POST_TYPES_DEFAULT = [ 'post' ];
+
+	/**
+	 * Option name storing whether photo captions are excluded from exports.
+	 *
+	 * Photo credits are a separate attribution field and are always exported.
+	 *
+	 * @var string
+	 */
+	public const CAPTIONS_OPTION = 'newspack_indesign_export_exclude_captions';
```
dkoo: **Severity:** nit

`PLATFORM_OPTION`/`PLATFORM_DEFAULT` and `POST_TYPES_OPTION`/`POST_TYPES_DEFAULT` match, but `CAPTIONS_OPTION` pairs with `EXCLUDE_CAPTIONS_DEFAULT`. Since the option genuinely stores "exclude captions" — the option string itself is `newspack_indesign_export_exclude_captions` — renaming to `EXCLUDE_CAPTIONS_OPTION` would align all three pairs and remove the polarity ambiguity at call sites.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -15,18 +16,38 @@ import WizardSection from '../../../../wizards-section';
...
 	} );
 
+	const togglePostType = ( slug: string, checked: boolean ) => {
+		const next = new Set( apiData.indesign_post_types );
+		if ( checked ) {
+			next.add( slug );
+		} else {
+			next.delete( slug );
+		}
+		apiFetchToggle( { ...apiData, indesign_post_types: Array.from( next ) }, true );
```
dkoo: **Severity:** nit

Each save spreads the whole `apiData`, which includes the server-computed `available_post_types` array of `{ value, label }` objects. The handler never reads that parameter, so this is harmless, but it inflates every request and blurs which fields of `PrintData` are actually writable. Splitting the writable fields from the derived ones in the type would make the contract clearer. Minor and related: `errorMessage` is only surfaced on the top action card, so a failed checkbox save reports its error far from the control that caused it.

dkoo (changes requested):
Release risk: HIGH — the post-types setting, which is the headline feature, does not register the bulk export action for any custom post type, and the settings UI gives no sign that anything is wrong. Three of the four changes (platform header, en-dash fix, caption opt-out) are correct and verified end to end in an isolated environment, the security posture is good (nonces plus per-object `edit_post` on both export paths, and an `is_post_supported()` guard so widening the post-type surface doesn't open an unguarded endpoint), and PHPCS and the 61 new tests are clean. The one blocker is a hook-ordering problem with a well-established fix used by the sibling optional modules. Worth a second look while you're in there: the REST write path and the option reader disagree on what a valid post type is, and `resolve_platform()` plus its new filter have no test coverage.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Fixed in af3b93ef4 — moved the post-type-dependent registration into a new `register_list_table_actions()` hooked on `init` priority 20, so custom post types (which register on `init`) are present when the `bulk_actions-edit-{$post_type}` filter names are built. Added `test_register_list_table_actions_registers_bulk_action_for_custom_post_type()`, which registers a CPT, stores it in the option, calls the deferred method, and asserts `has_filter( 'bulk_actions-edit-reviewcpt', … )` is registered — it fails against the old file-scope registration and passes now.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
miguelpeixe: Fixed in af3b93ef4 — the write path now validates each slug against `array_column( InDesign_Exporter::get_available_post_types(), 'value' )` and returns a 400 on any unavailable or non-string slug, instead of silently dropping it. The stored option can no longer diverge from the effective value the reader returns. Updated the REST tests to match: `rejects_unavailable_post_types`, `rejects_non_string_post_types`, plus a `dedupes_post_types` case that keeps the valid-dedup path covered.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Fixed in af3b93ef4 — `resolve_platform()` is now `public static`, and its return value is normalized to exactly `mac`/`win` after the filter runs, so a callback returning `auto`, `null`, or anything else degrades to `win` (the converter's default header) rather than leaking an invalid platform. Added direct coverage: `test_resolve_platform_setting_overrides_user_agent` (explicit setting beats the UA), `test_resolve_platform_auto_sniffs_user_agent`, `test_resolve_platform_filter_overrides_and_normalizes`, and `test_resolve_platform_filter_receives_context` (asserts the three args passed to `newspack_indesign_export_platform`). Also mirrored the excluded-types filter's defensiveness by casting `newspack_indesign_export_supported_post_types` output with `(array)`, with a regression test for a non-array callback return.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in af3b93ef4 — the picker now keeps an optimistic local mirror of the selection (`selectedPostTypes`) so a checkbox flips on click, and the save is debounced (500ms) so a rapid series of clicks coalesces into a single POST carrying the final selection. That avoids the in-flight dedupe swallowing the second toggle and reverting the box the user just clicked. The local state is re-synced from the server value on load and after each successful save.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in af3b93ef4 — added a non-dismissible warning `Notice` under the checkbox group that appears when the selection is empty, stating that the "Export as Adobe InDesign" actions won't appear anywhere until at least one post type is selected. I went with the help-text option rather than forcing a minimum so the admin can still deliberately clear the list.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Fixed in af3b93ef4 — `process_post_images()` now builds the block into an initially-empty string and returns `''` when nothing was appended, so a fully-skipped image set (e.g. caption-only images with captions excluded) no longer leaves a `\r\n` for `array_filter()` in `convert_post()` to keep. The `"\r\n"` seed is prepended only when there's real content. Added `test_caption_only_image_adds_no_content_when_captions_excluded`, which asserts the export with the image is byte-identical to the same post without it.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Partially addressed in af3b93ef4 — hoisted `get_supported_post_types()` out of the per-ID loop in the bulk handler, which previously re-resolved it once per selected post through `is_post_supported()`. I deliberately skipped a global static memo: the option can change within a single request (the unit tests exercise exactly that), so a static cache would need `update_option`/`delete_option` invalidation, which felt disproportionate for a no-DB registry read. The per-row `add_row_action()` call is left as-is for the same reason. Happy to add the memo + invalidation hooks if you'd rather. The "unsupported items silently dropped from a mixed bulk selection" point is noted as a separate follow-up.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
miguelpeixe: Fixed in af3b93ef4 — added `set_up()`/`tear_down()` that reset the three options, unregister any test-registered post types, and restore `$_SERVER['HTTP_USER_AGENT']`, and removed the inline trailing cleanups so a failed assertion mid-test can no longer leak state into later tests. On the WooCommerce collision: the `product` registration is now torn down unconditionally (guarded by `post_type_exists()`), and the suite stays green; I kept `product` since it's the concrete `EXCLUDED_POST_TYPES` case being asserted.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Fixed in af3b93ef4 — reworded the inline comment to "authenticated admin request (admin-post.php for a single export, or an edit.php bulk action) that is never cached", and added a line to the `resolve_platform()` docblock noting that a non-browser client (WP-CLI, a direct REST call) under `auto` has no User-Agent and therefore resolves to `win`.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Fixed in af3b93ef4 — renamed `CAPTIONS_OPTION` → `EXCLUDE_CAPTIONS_OPTION` and updated every reference (the reader accessor, the REST writer, and the tests), so all three option/default pairs now share the same polarity.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in af3b93ef4 — saves now go through a `save()` helper that strips the server-derived `available_post_types` from the payload, and `useWizardApiFetchToggle`'s `dataToSend` is widened to `Partial<T>` so only writable fields are sent. The `errorMessage`-placement note is a fair follow-up; I left it as-is for this pass since it's still surfaced on the top action card and reshaping error routing is a bit out of scope here.

miguelpeixe:
Thanks for the thorough review, @dkoo — all points addressed in af3b93ef4.

**Blocker — CPT bulk export action never registered (hook ordering).** Fixed. Moved the post-type-dependent registration into a new `register_list_table_actions()` hooked on `init` priority 20, matching the sibling optional modules, so custom post types registered on `init` are present when the `bulk_actions-edit-{$post_type}` filter names are built. Added `test_register_list_table_actions_registers_bulk_action_for_custom_post_type()`, which registers a CPT, stores it in the option, calls the deferred method, and asserts `has_filter( 'bulk_actions-edit-reviewcpt', … )` — it fails against the old file-scope registration and passes now.

**REST write path vs. option reader disagreement.** Fixed. The write path now validates each slug against `array_column( InDesign_Exporter::get_available_post_types(), 'value' )` and returns a 400 on any unavailable or non-string slug instead of silently dropping it, so the stored option can no longer diverge from the effective value. REST tests updated accordingly (`rejects_unavailable_post_types`, `rejects_non_string_post_types`, `dedupes_post_types`).

**`resolve_platform()` and its filter had no coverage.** Fixed. Made it `public static` and normalized the return to exactly `mac`/`win` after the filter (so a callback returning `auto`/`null` degrades to `win` rather than an invalid platform), then added four tests covering the explicit-setting-wins branch, the auto/UA branch, the filter override + normalization, and the three args passed to `newspack_indesign_export_platform`. Also cast the `newspack_indesign_export_supported_post_types` filter output with `(array)` (with a non-array-callback regression test) to match the excluded-types filter's defensiveness.

I also picked up the nits: the converter blank-line leak (byte-identical baseline test), hoisting `get_supported_post_types()` out of the bulk loop, centralized `set_up`/`tear_down` cleanup, the comment/docblock wording, and the `EXCLUDE_CAPTIONS_OPTION` rename. Details are in the per-thread replies. PHPCS is clean and the InDesign suite passes (70 tests / 185 assertions).

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -15,18 +17,78 @@ import WizardSection from '../../../../wizards-section';
...
+			indesign_exclude_captions: false,
 		},
 		description: __( 'Allows editors to export article content in Adobe InDesign Tagged Text format.', 'newspack-plugin' ),
 	} );
 
+	// Optimistic mirror of the post-type selection so a checkbox flips on click
+	// instead of waiting for the round trip. Kept in sync with the server value
+	// on load and after every successful save.
+	const [ selectedPostTypes, setSelectedPostTypes ] = useState< string[] >( apiData.indesign_post_types );
+	useEffect( () => {
```
dkoo, line 54: **Severity:** suggestion

The optimistic mirror has no failure path. `setSelectedPostTypes( nextPostTypes )` runs unconditionally on click, and the only thing that ever re-syncs it is this effect, which keys on `apiData.indesign_post_types`.

`apiData` is written in exactly one place — the hook's `onSuccess: setApiData`. On a rejected POST, `useWizardApiFetch`'s `catchCallback` calls `setError`, fires `onError`, and re-throws; `thenCallback`/`onSuccess` never run. So `apiData` keeps its old reference, this effect never re-fires, and the checkbox stays on the optimistic value indefinitely. Nothing reloads the view either — this section does not pass `refreshOn`.

Before this commit the checkbox rendered straight off `apiData` and self-corrected the moment a save failed, so this is a new regression rather than a pre-existing gap. It compounds with error placement: `errorMessage` is only rendered on the top-level action card (line 98), a section away from the control that actually failed, so the user's signal is an unrelated-looking error at the top of the page while the checkbox below reads as saved.

Reverting `selectedPostTypes` from an `onError`, or re-deriving it from `apiData` when `errorMessage` becomes non-null, would close this.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -15,18 +17,78 @@ import WizardSection from '../../../../wizards-section';
...
+		[]
+	);
+
+	/**
+	 * Persist a change, sending only the writable fields. `available_post_types`
+	 * is derived server-side (read-only), so it is stripped from the payload.
+	 */
+	const save = ( overrides: Partial< PrintData > ) => {
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const { available_post_types, ...writable } = { ...apiData, ...overrides };
```
dkoo: **Severity:** suggestion

`save()` still builds a full payload out of `{ ...apiData, ...overrides }`, and it now does so from inside a 500 ms timer, which turns a latent issue into a reachable one in two directions.

**Stale overwrite, both ways.** The debounced callback closes over `apiData` as of the click that scheduled it. If a platform or captions save resolves inside that window, the later post-types POST re-sends the old platform/captions values and reverts them; conversely, toggling captions while a checkbox save is still pending posts the pre-click `indesign_post_types`. The second case self-heals when the timer fires, but not if the component unmounted first (see the neighbouring comment).

**Permanent loss of a stored post type.** `get_post_types_setting()` filters the stored array down to the currently *available* types before returning it, so a slug whose CPT is temporarily gone — its plugin deactivated during an update, or the type added to `newspack_indesign_export_excluded_post_types` — is silently absent from the GET payload. Any subsequent save then writes that reduced array back over the real option. Reproduced in a live environment: with the option at `["post","reviewcpt","gone_cpt"]`, a save that changed only the platform dropdown left it at `["post","reviewcpt"]`. Re-activating the plugin no longer restores that post type's export actions, and nothing tells the admin it happened.

Both collapse into one fix, and the server already supports it: `api_update_print_settings()` requires only `module_enabled_print` to be present and `has_param()`-gates every other field, so `save()` can send `{ module_enabled_print: apiData.module_enabled_print, ...overrides }` and nothing else. That also removes the need to strip `available_post_types`. Worth noting the widened `Partial<T>` on `apiFetchToggle` is currently *wider* than this endpoint accepts — omitting `module_enabled_print` yields a 400 — so it is a latent trap for the next caller unless the required field is threaded through explicitly.

**Risk if not addressed:** a publisher deactivates a CPT plugin to troubleshoot something unrelated, an editor adjusts the InDesign header platform in the same session, and that post type quietly disappears from the export configuration — surfacing later as "the export button vanished for our events" with nothing in the UI or logs connecting the two.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
```diff
@@ -15,18 +17,78 @@ import WizardSection from '../../../../wizards-section';
...
+	// instead of waiting for the round trip. Kept in sync with the server value
+	// on load and after every successful save.
+	const [ selectedPostTypes, setSelectedPostTypes ] = useState< string[] >( apiData.indesign_post_types );
+	useEffect( () => {
+		setSelectedPostTypes( apiData.indesign_post_types );
+	}, [ apiData.indesign_post_types ] );
+
+	const saveTimer = useRef< ReturnType< typeof setTimeout > | undefined >();
+	useEffect(
+		() => () => {
```
dkoo, line 61: **Severity:** suggestion

This cleanup clears the pending timer without flushing it, so a checkbox change made within 500 ms of leaving the view is dropped after the box already flipped optimistically.

That is a short window, but it is genuinely reachable here: the settings wizard renders each section through a react-router `<Switch>`/`<Route>` (`packages/components/src/wizard/index.js`), so switching tabs unmounts this component outright. Turning the module off in the card above unmounts the whole section too, and a plain navigate-away does the same. In every case the only scheduled POST is cancelled and the user has seen the checkbox change.

Flushing the pending save in the cleanup would fix it, though sending only the changed fields immediately — per the neighbouring comment — removes the need for the debounce at all, since the dedupe behaviour it was working around goes away once each save carries the final value.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
```diff
@@ -81,7 +85,40 @@ public function api_get_print_settings() {
...
+				return new \WP_Error( 'invalid_param', __( 'Invalid parameter for indesign_post_types.', 'newspack-plugin' ), [ 'status' => 400 ] );
+			}
+			// Validate against the same "available" list the settings UI offers
+			// (public + show_ui, minus excluded types), which is the single source
+			// of truth for a valid slug. Rejecting here — rather than silently
+			// dropping unknown slugs — keeps the stored option from diverging from
+			// the effective value the reader returns. Safe at this layer: REST runs
+			// long after `init`, so the available list is complete.
+			$available_post_types = array_column( InDesign_Exporter::get_available_post_types(), 'value' );
+			$post_types           = array_values( array_unique( $post_types ) );
```
dkoo: **Severity:** nit

`array_unique()` defaults to `SORT_STRING`, so it stringifies every element — but it runs here, one line before the `is_string()` check that would have rejected a non-string. A nested-array element (trivially POSTable as JSON) therefore raises "Array to string conversion" during dedupe, before validation gets to reject it.

Confirmed against the running endpoint: POSTing `indesign_post_types: ["post", ["post"]]` emits the warning and *then* returns the correct 400. So the outcome is right and only an authenticated admin can reach it — this is log noise, not a security or correctness issue. It also means `1` and `'1'` collapse into one entry before either is validated.

Validating element types and membership first, then `array_values( array_unique( ... ) )` on the survivors, makes the rejection path warning-free.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -595,4 +943,266 @@ public function test_image_caption_and_credit_special_characters() {
...
+
+		$this->assertNotFalse(
+			has_filter( 'bulk_actions-edit-reviewcpt', [ InDesign_Exporter::class, 'add_bulk_action' ] ),
+			'The bulk export action must be registered for a configured custom post type.'
+		);
+		$this->assertNotFalse(
+			has_filter( 'handle_bulk_actions-edit-reviewcpt', [ InDesign_Exporter::class, 'handle_bulk_action' ] )
+		);
+
+		// Remove the filters this test added (options/CPTs are handled in tear_down).
```
dkoo: **Severity:** nit

The new `tear_down()` docblock justifies centralizing cleanup so "a failed assertion mid-test can't leak a registration into later tests" — but this test's own filter cleanup runs inline *after* four assertions. If any of them fails, PHPUnit aborts the method and `bulk_actions-edit-reviewcpt`, `handle_bulk_actions-edit-reviewcpt`, `post_row_actions` and `page_row_actions` stay hooked for the rest of the process, since `tear_down()` only resets options, post types and the User-Agent.

On the `TEST_POST_TYPES` sweep: I checked the claim that it could unregister genuinely production-registered types, and it does not today — the optional module is inactive in the test bootstrap, so `InDesign_Exporter::init()` bails, and none of `product`, `newspack_collection`, `newspack_nl_list`, `partner_rss_feed`, `event` or `flyer` exists when this class runs. The sweep is safe as written; it is just coupled to that bootstrap state, because it unregisters by name with no record of who registered each type. Tracking the slugs a test actually registered would decouple it.

Separately, `test_caption_only_image_adds_no_content_when_captions_excluded()` asserts only that the two exports match, so it would still pass if the image block stopped being parsed at all. Asserting that the same fixture with `include_captions => true` *does* contain `PhotoCaption` would pin it.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -39,15 +107,39 @@ public static function init() {
...
+	 * Register the list-table bulk and row export actions.
+	 *
+	 * Hooked to `init` (priority 20) so every configured post type — including
+	 * custom post types registered on `init` — is present when the supported
+	 * list is resolved. The `bulk_actions-edit-{$post_type}` filter names are
+	 * built from that list here; the shared row-action filters are added
+	 * alongside them. Both sets of filters fire during list-table render, well
+	 * after `init`, so late registration is safe.
+	 */
+	public static function register_list_table_actions() {
```
dkoo, line 132: **Severity:** nit

Deferring to `init` priority 20 fixes the custom-post-type case, and it is comfortably safe against the Newspack ecosystem — every CPT registration across `plugins/` hooks `init` at priority 1–10.

The residual edge is that this now freezes the *bulk* action set at that moment, while `add_row_action()` still resolves `get_supported_post_types()` live at render time. So a site that hooks `newspack_indesign_export_supported_post_types` or `newspack_indesign_export_excluded_post_types` later than `init:20` — or on `wp_loaded`/`admin_init` — gets the row action but not the bulk action, which is the same half-working shape as the bug just fixed, just reached from a different direction. The same applies to a third-party CPT registered after priority 20.

Since both filters are advertised extension points, a line in their docblocks stating that they must be hooked before `init` priority 20 would make the contract explicit.

dkoo (approved):
Release risk: LOW — no confirmed blockers, no contract-breaking changes, and the fixes were verified at runtime rather than read.

The blocker is genuinely resolved: deferring registration to `init` priority 20 puts custom post types back in the bulk-action set, and there is now a regression test that would catch it coming back. The REST validation, the capability gate through the refactor, the platform normalization, and the blank-line fix all check out in a live environment.

What is worth a second pass before merge is the new optimistic-checkbox code, where three reviewers converged on the same area. A failed save now leaves the checkbox showing a selection that was never persisted, and the debounced save still posts a whole snapshot of the settings object — which both clobbers concurrent edits and, when a stored post type's plugin is temporarily inactive, permanently drops it from the option. Sending only the changed fields fixes all three at once and the endpoint already supports it. None of this blocks merge.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in 2eb2a2ae9 — `apiFetchToggle` now returns the request promise, and the debounced save attaches a `.catch()` that snaps `selectedPostTypes` back to the last server-acknowledged value. That value is read from an `apiDataRef` kept current by effect, so the re-sync survives stale closures and fires on consecutive identical failures too (unlike keying an effect on `errorMessage`). The action-card error message still surfaces the failure.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in 2eb2a2ae9 — `save()` no longer spreads an `apiData` snapshot. The payload carries only `module_enabled_print` (which the endpoint requires) plus the explicitly changed fields, and the module flag is read from `apiDataRef` at send time — so neither a stale render nor the debounce timer can write back platform/captions/post-types values, in either direction.

Relatedly, the toggle hook now passes `updateCacheMethods: [ 'GET' ]` on POST so a successful save refreshes the store's GET cache. Without that, revisiting the tab remounts the view from the first-load cache snapshot (POSTs never updated the GET cache), and any later save would have re-persisted those stale values — the reachable-any-time flavor of this same overwrite. The hook's other consumers (rss/nextdoor/collections) get the cache refresh for free.

plugins/newspack-plugin/src/wizards/newspack/views/settings/print/index.tsx
miguelpeixe: Fixed in 2eb2a2ae9 — the unmount cleanup now flushes the pending save (held in a `pendingSave` ref) instead of discarding it, so a change made within the debounce window of switching tabs still persists. The flushed request reads the module flag from the ref, and its state updates land on the unmounted component as no-ops.

plugins/newspack-plugin/includes/wizards/newspack/class-print-section.php
miguelpeixe: Fixed in 2eb2a2ae9 — validation now runs before dedupe: every element passes the `is_string()` + availability check first, and `array_unique()` only ever sees a known-string array. Added `test_api_update_print_settings_rejects_nested_array_post_types` to pin the warning-free 400.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
miguelpeixe: Fixed in 2eb2a2ae9 — the four `remove_filter()` calls moved into `tear_down()` (removing a never-added filter is a no-op, so they run unconditionally for every test) and the inline cleanup after the assertions is gone, so a mid-test failure can no longer leak the hooks.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Agreed on leaving this as documented behavior — every CPT registration across the Newspack ecosystem hooks `init` at priority 1–10, and a site-level callback on the supported-types filters added after `init` 20 degrades to row-actions-only rather than nothing, which is the right failure direction. Happy to revisit if a concrete site setup surfaces it.

miguelpeixe:
Second pass from the approval review is in — 2eb2a2ae9 addresses all six notes:

- **Optimistic checkbox cluster**: saves now send only the module flag + explicitly changed fields (read from a ref at send time), successful POSTs refresh the store's GET cache so a remounted tab can neither display nor re-save a stale first-load snapshot, failed saves snap the optimistic selection back, and unmount flushes the pending debounced save instead of dropping it.
- **REST nit**: `indesign_post_types` validation runs before `array_unique()`, with a new nested-array 400 regression test.
- **Test nit**: the CPT test's filter cleanup moved into `tear_down()`.
- **`init` 20 freeze**: kept as documented behavior (reply on the thread).

PHPCS and ESLint clean; both InDesign PHPUnit suites pass (87 tests, 235 assertions).

PR #464, opened 2026-06-30
feat(audience): subscription products (Plans) wizard

miguelpeixe:
### All Submissions:

Adds a new **"Plans"** management page under **Audience** — a DataViews UI for managing WooCommerce Subscriptions products: name, price + billing period, active-subscriber counts, category, status, variations, group/multi-seat subscriptions, and content unlocks. List, create, and edit subscription products without leaving the wizard.

The page is **gated behind the `NEWSPACK_PLANS_UI` feature flag** and only registers when WooCommerce Subscriptions is active. It shows the **plain product price** — the dynamic-pricing "effective price" overlay is intentionally a separate PR (the integration), so this PR has **zero dependency on the dynamic-pricing engine**.

First of a three-way split of `feat/rsm-pricing-rules-dataviews`: **(1) Plans (this PR)**, (2) Pricing Rules wizard, (3) the engine-pricing integration onto Plans. PRs 1 and 2 are independent; PR 3 stacks on this one.

1. Add `define( 'NEWSPACK_PLANS_UI', true );` to `wp-config.php` on a site with **WooCommerce Subscriptions** active.
2. Go to **Audience** — a **Plans** submenu appears. Without the flag (or without WC Subscriptions) it does not register.
3. The list shows subscription products with price/period, active-subscriber counts, category, and status. Create a new product and edit an existing one; confirm variations, group-subscription (multi-seat) settings, and content unlocks save correctly.
4. Confirm prices shown are the **plain product price** — there is no dynamic-pricing/effective-price column in this PR.

### Other information:


miguelpeixe:
Addressed Copilot's review in `2ec87255c`:

- **`product-edit.tsx` redirect during render** — moved into a `useEffect` so render stays pure (StrictMode-safe).
- **`subscription-products.php` translated term names** — switched to locale-independent default term names; UI labels stay translated separately.
- **`get_active_subscription_count()` N+1 `wcs_get_subscriptions()`** — added a per-request memo of subscription-id lists, keyed by product id, so repeated/aggregated ids resolve once.
- **`README.md` stale "RSM prototype / Not for PR to main" header** — rewritten to the shipped **Plans** reality.
- **No Plans-specific PHPUnit tests for the create/update REST endpoints** — acknowledged. Worthwhile as a follow-up: this PR reconstructs an existing branch with no upstream Plans tests, so REST-endpoint coverage is better scoped on its own rather than bundled here.

dkoo (changes requested):
@miguelpeixe This review combines a `/code-review` pass with hands-on manual QA in an isolated env (WCS 9.0 + Stripe test). Overall it's a clean, well-built DataViews CRUD wizard: the Woo write paths all match the wiring map, input validation is thorough, and type-locking on edit is correct and important. Design-system review came back clean (correct color tokens, correct components, only the dead CSS), and QA passed every §18 non-engine path — 49/49 REST+DB assertions + group/multi-seat + browser smoke, with the engine regression harness unaffected.

Requesting changes on a few substantive items — mainly the deletion-guard coverage and a silent data-loss path on variable subs, plus a capability decision and an easy cache-invalidation fix. **The detailed findings are left as inline comments on the relevant lines.**

_Scope note: reviewed as Layer-1 (plain product price); the engine effective-price surfaces are #466._

**Two nits without a clean single anchor:**
- Create-vs-edit donation-flag write asymmetry (create writes `_newspack_is_donation` only when ticked; edit always writes `yes`/`no`) — a documented standards question, not a bug.
- Unchecked variation `save()` return in `create_variable_subscription()` — a silent failure could leave a parent missing a plan.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+	 * @return string The wizard name.
+	 */
+	public function get_name() {
+		return esc_html__( 'Audience Management / Plans', 'newspack-plugin' );
+	}
+
+	/**
+	 * Register the endpoints needed for the wizard screens.
+	 */
+	public function register_api_endpoints() {
```
dkoo, line 114: 🔴 **Active-subscriber deletion guard is effectively absent for whole-plan deletion.** This registers only READABLE + CREATABLE (collection) and EDITABLE (single) — there's no `DELETABLE` route, and the Plans UI wires no delete/trash action. Nothing hooks core WooCommerce deletion (`pre_trash_post`/`before_delete_post`), and `product_has_active_subscriptions()` is only called inside `sync_variable_variations()` to guard removing a *variation row* mid-edit. Net (confirmed in manual QA): a subscription product with active subscribers can be trashed/deleted from the **core Products page with no warning**. Suggest hooking `pre_trash_post`/`before_delete_post` on `product`/`product_variation` to block or warn when active subs exist, and widening the status set beyond `active`/`pending-cancel` (misses `on-hold`/`pending`/suspended-but-resumable).

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/product-form.tsx
```diff
@@ -0,0 +1,521 @@
...
+				setError( __( 'Select at least one subscription to bundle.', 'newspack-plugin' ) );
+				return;
+			}
+			payload.bundled_product_ids = bundled;
+		} else if ( type === 'variable-subscription' ) {
+			if ( plans.some( plan => ( ! plan.existing && ! plan.label.trim() ) || ! isValidPrice( plan.price ) ) ) {
+				setError( __( 'Each plan needs a label and a valid price.', 'newspack-plugin' ) );
+				return;
+			}
+			payload.variations = plans.map( plan => ( {
```
dkoo, line 202: 🔴 **Variable-sub group settings clobber divergent per-variation values on save (data-loss risk).** This spreads a single `groupFields` value into *every* variation → `sync_variable_variations()` writes it to all of them, overwriting divergent per-variation group settings set via the core Products UI. The single value is seeded from a collapsed parent summary (`get_group_subscription_summary()`); on divergence (limit `-1`/"Varies") the form silently defaults to `0` (L108). Non-blocking (the Plans UI can't create divergent settings), but silent data loss when they exist. Suggest surfacing group settings per variation, or detecting divergence server-side and refusing to overwrite a collapsed/`-1` value.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Audience Subscription Products Wizard.
+ */
+class Audience_Subscription_Products extends Wizard {
```
dkoo, line 19: 🔴 (decision) **Capability is `manage_options`, not `manage_woocommerce`.** All three CRUD routes inherit the base wizard's `manage_options`; the sibling group-subscription API uses `manage_woocommerce` for equivalent writes. Not a security issue — `manage_options` is the stricter gate — but it locks WooCommerce shop-managers out of a product-management page and is inconsistent with product-CRUD semantics. Either override `$capability = 'manage_woocommerce'` or document the choice.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+				}
+			}
+		}
+
+		// Reindex inner maps to plain lists.
+		foreach ( $map as $product_id => $product_gates ) {
+			$map[ $product_id ] = array_values( $product_gates );
+		}
+
+		set_transient( self::GATE_MAP_TRANSIENT, $map, 5 * MINUTE_IN_SECONDS );
```
dkoo, line 1526: 🟡 **"Unlocks" gate-map transient is never invalidated.** 5-min TTL here, but nothing calls `delete_transient( self::GATE_MAP_TRANSIENT )` on gate CRUD — so editing/creating/deleting a content gate isn't reflected in the Unlocks column for up to 5 minutes (surfaced in manual QA). Clean hooks exist: `Content_Gate::update_gate_settings()` / `update_gate_setting()` and the gate `before_delete_post` (`class-institution.php` already models the save→invalidate pattern). Read-only column, so minor — but easy.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+		}
+
+		// Set the product_type term explicitly (WC's save doesn't reliably do it here);
+		// without this the parent reads back as 'simple' and has no variations.
+		wp_set_object_terms( $parent_id, 'variable-subscription', 'product_type' );
+
+		foreach ( $clean as $plan ) {
+			$variation = new \WC_Product_Variation();
+			$variation->set_parent_id( $parent_id );
+			$variation->set_attributes( [ 'billing-period' => $plan['label'] ] );
```
dkoo: 🟡 **Hardcoded variation attribute key `'billing-period'`.** Works only because `sanitize_title('Billing period') === 'billing-period'`; would silently break variations to "Any" if the attribute name is ever localized/changed. Derive the key from `sanitize_title()` of the name. (Same at L732.)

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+		// call is a query — so resolve each id at most once.
+		static $ids_by_product = [];
+		$subscription_ids = [];
+		foreach ( $product_ids as $product_id ) {
+			if ( ! isset( $ids_by_product[ $product_id ] ) ) {
+				$subscriptions = \wcs_get_subscriptions(
+					[
+						'product_id'             => $product_id,
+						'subscription_status'    => self::ACTIVE_SUBSCRIPTION_STATUSES,
+						'subscriptions_per_page' => -1,
```
dkoo: 🟡 **Active-sub count pulls all subscriptions into memory.** `subscriptions_per_page => -1` hydrates every matching `WC_Subscription` just to `count()` them, once per product/variation per list load. Fine at typical Newspack sizes, but a latency hazard on a plan with tens of thousands of subscribers. Consider caching the counts (alongside the gate-map transient) or a lighter count path.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/style.scss
```diff
@@ -0,0 +1,96 @@
...
+
+	// "Unlocks" — content gates this product unlocks (Access control).
+	&__unlocks {
+		display: flex;
+		flex-wrap: wrap;
+		gap: 4px;
+	}
+
+	// Edit modal.
+	&__modal {
```
dkoo: ⚪ Dead CSS: `__modal`, `__modal-value`, `__modal-divider`, `__modal-note` (and `__add-modal*` above) are unused now that the form is a full-page route. (`__modal-label` *is* still used — keep it, though the name's misleading.)

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -0,0 +1,1476 @@
...
+	 * Read a product's base subscription price.
+	 *
+	 * Distinguishes "not set" (null) from an explicit 0 so the UI can render the
+	 * difference faithfully.
+	 *
+	 * @param \WC_Product $product The product or variation.
+	 *
+	 * @return float|null The price, or null when not set.
+	 */
+	private static function read_subscription_price( $product ) {
```
dkoo, line 1220: ⚪ `! isset( $raw )` here is dead — `get_meta()` always returns a set value; `'' === $raw` does the real work.

dkoo:
@miguelpeixe another UX-based thing that came to mind while manually testing this: should we consider adding a styled modal confirmation when leaving a plan's edit view with unsaved changes, as in the Access Control UI? That's becoming a more widespread pattern across our wizard UIs so we might want to follow the trend here.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in a4d585a7d. Added `pre_trash_post`/`pre_delete_post` guards (`block_subscription_product_deletion`) that refuse to trash/delete a subscription product — or, for a variable subscription, any of its variations — while it still has a renewable subscription, from anywhere in wp-admin (Products list, bulk actions, programmatic). Widened the set to a new `BLOCKING_SUBSCRIPTION_STATUSES` (`active`, `pending`, `on-hold`, `pending-cancel`); the informational subscriber count keeps the narrower `ACTIVE_SUBSCRIPTION_STATUSES`. WordPress surfaces its standard failure notice when a delete is blocked.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/product-form.tsx
miguelpeixe: Fixed server-side in a4d585a7d (more robust than gating the UI). `sync_variable_variations()` now detects pre-existing divergence (`existing_group_settings_diverge()`) and, when the existing variations already differ, leaves each one's group meta untouched instead of overwriting it with the collapsed value. New variations still take the submitted value, so the Plans UI can no longer clobber per-variation settings configured via the core product editor.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Done in a4d585a7d — went with `manage_woocommerce`. `protected $capability = 'manage_woocommerce'` overrides the base wizard's `manage_options` and covers the menu page + all three REST routes (shared `Wizard::api_permissions_check()`). Now consistent with the sibling group-subscription API and product-CRUD semantics, and shop managers aren't locked out.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in a4d585a7d. `invalidate_gate_map_cache()` (clears the transient + per-request memo) now runs on `save_post`/`before_delete_post` for gate post types and on `custom_access` meta writes (`updated_post_meta`/`added_post_meta`) — the latter because `update_custom_access_settings()` writes the subscription rule via post meta and bypasses `save_post`.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in a4d585a7d. Added a `BILLING_PERIOD_ATTRIBUTE_NAME` const + a `billing_period_attribute_key()` helper (`sanitize_title()` of the name); all five sites now derive the key from it (both create + sync `set_attributes()`, plus the two `get_attribute()` reads), so a renamed/localized attribute name stays in sync.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in a4d585a7d. Replaced `wcs_get_subscriptions( … per_page => -1 )` with `wcs_get_subscriptions_for_product( $id, 'ids', … )` — a single lightweight query that returns IDs with no `WC_Subscription` hydration. The per-request memo is kept; the deletion guard's existence check uses the same call with `limit => 1`.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/style.scss
miguelpeixe: Fixed in a4d585a7d. Removed `__modal`, `__modal-value`, `__modal-divider`, `__modal-note`, and the `__add-modal*` blocks. Kept `__modal-label` (still used by the read-only Unlocks caption) and left its name as-is.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in a4d585a7d. Dropped the `! isset( $raw )` half — `get_meta()` always returns a set value, so `'' === $raw` does the real work.

miguelpeixe:
Thanks for the thorough review + hands-on QA, @dkoo. Pushed **a4d585a7d** (fixes) and **5513da9a9** (the modal); the inline threads are answered individually. The two review-body nits and the unsaved-changes suggestion:

**Create-vs-edit donation-flag write asymmetry** — Fixed in a4d585a7d. All three create paths now call `set_donation_flag( $product, $is_donation )` (always writing `yes`/`no`) — the same helper `api_update_product()` uses — so create and edit are symmetric and the create-only conditional write is gone.

**Unchecked variation `save()` in `create_variable_subscription()`** — Fixed in a4d585a7d. Each variation `save()` return is now checked; on failure it rolls back (force-deletes the already-saved variations + the parent) and returns a `create_failed` error, so a partial save can no longer leave a parent missing a plan.

**Unsaved-changes confirmation modal** (from the conversation thread) — Added in 5513da9a9. The plan editor now uses the same `useUnsavedChangesDialog` guard as the Access Control edit view: dirty state is tracked via a first-render snapshot of the form, the guard covers hash-router nav + refresh/close, and it stands down while a save is in flight. Passed lint but I have not browser-tested it — worth a look on your next pass.

miguelpeixe:
Follow-up fix in **4ccdae4ff**: removing a plan from a variable subscription when that plan still had active subscribers was a silent no-op. The save returned success, but the variation was kept (correctly, to avoid orphaning the subscription), while the parent’s attribute options had already been rebuilt without it, leaving the product inconsistent.

Now:
- **Server (authoritative):** `sync_variable_variations()` checks up-front, before any mutation, and rejects the save with a clear 409 (`Can’t remove a plan that still has active subscribers: <label>. Cancel or move those subscriptions first.`). Uses the widened blocking-status set, so nothing is half-applied.
- **UI:** each variation now carries an `active_subscriptions` count (free — shares the per-request subscription memo), and the plan’s **Remove** button is disabled with a `“N active subscriber(s) — can’t be removed”` helptext.

image

dkoo (approved):
@miguelpeixe Thanks — all the review feedback is addressed, and I re-verified the substantive fixes live on an isolated env (WCS 9.0 + Stripe test). Approving. 🎉

Findings, all resolved:

- **P1 (capability)** → `manage_woocommerce` on the wizard + all three REST routes. ✅ _(live: a `manage_woocommerce`-only user is allowed; a no-cap user gets 403)_
- **P2 (count perf)** → `wcs_get_subscriptions_for_product( …, 'ids' )` — no more `WC_Subscription` hydration just to count. ✅
- **P3 (hardcoded `billing-period` key)** → derived via `sanitize_title()`. ✅
- **P4 (deletion guard)** → now hooks `pre_trash_post` + `pre_delete_post` on core product/variation deletion, with the status set widened to `[active, pending, on-hold, pending-cancel]`, plus the wizard-side 409 + disabled Remove. ✅ _(live: trashing/deleting a product with an active — or on-hold — subscriber is blocked, including via a child variation on a variable-sub parent; a no-subscriber product still trashes normally)_
- **P9 (Unlocks gate-map staleness)** → `invalidate_gate_map_cache()` on gate save/delete. ✅ _(live: editing a gate is reflected in the Unlocks map immediately, not after the 5-min TTL)_
- **P10 (variable-sub group-setting clobber)** → server-side divergence guard. ✅ _(live: divergent per-variation group settings are preserved when a collapsed value is saved; uniform variations still receive it)_
- **P5 / P6 + the two unanchored nits** → all cleaned up.

Nice work — thorough pass. 🙏

PR #466, opened 2026-06-30
feat(audience): engine-priced effective price on Plans (integration)

miguelpeixe:
### All Submissions:

Wires the **dynamic-pricing engine** into the **Plans** wizard (PR 1) so each subscription product shows its **engine-priced effective price** and the **applied pricing-rule stack** — not just the plain product price.

- Adds `Subscription_Policy_Resolver` — the **single integration seam** between the Plans UI and the standalone `woocommerce-dynamic-pricing` engine. It composes all active rules over a product's purchase cycle for the effective price, lists the matching rules with the winner flagged, and projects the per-cycle price trajectory. Gracefully degrades to the base price when the engine is inactive, the product is invalid, or it's engine-excluded (e.g. donations) — no fabricated data.
- Surfaces it in the DataViews list (and **per-variation** for variable subscriptions): base → effective price, applied-policy chips with the winner distinguished, and a per-cycle schedule tooltip.
- Reads route exclusively through `Subscription_Policy_Resolver::resolve()` (and the `newspack_subscription_policy_resolution` filter); the return shape is the UI contract.

**Third of a three-way split** of `feat/rsm-pricing-rules-dataviews`: (1) Plans, (2) Pricing Rules, **(3) the engine-pricing integration (this PR)**. **Stacks on PR 1 (Plans)** — based on `feat/audience-plans-ui`, so review/merge that one first; this PR's diff is just the Layer-2 integration on top.

1. On a site with **WooCommerce Subscriptions** + the **woocommerce-dynamic-pricing** engine active, and `define( 'NEWSPACK_PLANS_UI', true );` in `wp-config.php`.
2. Create a pricing rule (**Audience → Pricing Rules**) that targets subscriptions.
3. Go to **Audience → Plans** — subscription products show the **effective price** (≠ base when a rule applies), the applied-rule chips with the **winning** rule flagged, and a per-cycle schedule tooltip. Variable subscriptions resolve **per variation**.
4. Deactivate the engine plugin → the list falls back to the plain base price with no rule chips (no errors).

### Other information:


miguelpeixe:
Addressed Copilot's review in `e1c536ecb`:

- **`null` base price → fabricated `$0.00`** (per-variation, non-variable subs, and the no-priced-variation fallback) — added a `resolve_policy()` guard that returns an empty resolution (em dash, no rules) when the base price is unset, instead of letting the resolver coerce `null` → `0.0` and prime the engine against a fabricated base.
- **`formatAmount()` used `toFixed()`** (ignored locale separators) — now matches the PHP base `price_label`: the currency payload carries the `number_format_i18n` decimal/thousand separators, and `formatAmount` groups the integer part and uses the locale decimal separator.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
```diff
@@ -0,0 +1,263 @@
...
+		$repository = $engine->repository();
+		$matching   = $repository ? $repository->for_context( self::context_for( $product, $base, 1 ) ) : [];
+
+		$rules = [];
+		foreach ( $matching as $rule ) {
+			$rule_id = (string) $rule->id;
+			$rules[] = self::policy(
+				$rule_id,
+				(string) $rule->strategy_id,
+				get_the_title( (int) $rule_id ),
```
dkoo: 🔴 **Blocker — HTML entities render literally in the chips.** `get_the_title()` runs `wptexturize`, so a rule named `A & B – C` emits `A &#038; B &#8211; C`, and the React `<Badge text>` (`policy-cells.tsx:48`) renders the string node **without decoding** → the literal entity shows in the Plans "Applied rules" chip (measured). Fix: `html_entity_decode( get_the_title( (int) $rule_id ), ENT_QUOTES )` here **and** at the `project_schedule` copy (~L190) — exactly what this PR's modal bridge already does at `class-dynamic-pricing-bridges.php:183`.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
```diff
@@ -0,0 +1,263 @@
...
+
+		// Project the composed price across the subscription's cycles, attributing the
+		// winning rule to each segment — the same walk the impact preview uses. The
+		// purchase cycle (segment 0) sets the headline effective price + winning rule.
+		$schedule   = self::project_schedule( $engine, $product, $base );
+		$effective  = $schedule ? (float) $schedule[0]['amount'] : $base;
+		$winning_id = $schedule ? (string) $schedule[0]['rule_id'] : '';
+
+		$repository = $engine->repository();
+		$matching   = $repository ? $repository->for_context( self::context_for( $product, $base, 1 ) ) : [];
```
dkoo: **F2 (display-only over-inclusion):** the chip list comes from `for_context()` (scope + active-window only), so condition-gated rules **over-list** — a segment-gated rule renders as an "Applied rules" chip on products/readers it doesn't actually apply to. Manual QA confirmed the effective **price** is correct (the price path goes through `resolve()`→`matching_rules()`, which excludes it), so this is chip-accuracy, not a money bug. Suggest building `$rules` from `$engine->matching_rules( self::context_for( $product, $base, 1 ) )` so the chips match what actually applies.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
```diff
@@ -0,0 +1,263 @@
...
+	 * winning at the segment's first cycle). Mirrors the impact preview's
+	 * projection so the Plans tooltip matches what buyers are charged over time.
+	 *
+	 * @param \Automattic\WooCommerce\DynamicPricing\Pricing_Engine $engine  The engine.
+	 * @param \WC_Product                                           $product The product.
+	 * @param float                                                 $base    The base recurring price.
+	 *
+	 * @return array<int, array{from_cycle:int, amount:float, rule_id:string, rule_label:string}>
+	 */
+	private static function project_schedule( $engine, $product, $base ) {
```
dkoo: **F3 + F4 (correctness + duplication):** `project_schedule()` re-implements the engine's cycle walk **and** feeds it the wizard's injected `_subscription_price` base — which diverges from every other engine surface (they derive base via `Amount_Calculator::base_price_for()`). Measured for the APFS product: injected base **$6** vs engine base **$20**, so a `percent_of_base` rule would show a Plans price that disagrees with the storefront/My-Account schedule (the "table matches what buyers see" guarantee). The public `Schedule_Projector::project_for_product( WC_Product, ?WC_Customer )` already does this walk correctly (the admin inspector uses it) — calling it fixes the base divergence (F3) **and** removes the duplication (F4), and inherits its per-request memo (helps F5). If per-segment winner attribution is needed it doesn't return `rule_id` — add a rule-id-carrying variant to the engine rather than forking the walk here.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
```diff
@@ -0,0 +1,263 @@
...
+		$horizon  = \Automattic\WooCommerce\DynamicPricing\Schedule_Projector::HORIZON;
+		$decimals = function_exists( 'wc_get_price_decimals' ) ? wc_get_price_decimals() : 2;
+		$segments = [];
+		$previous = null;
+
+		for ( $cycle = 1; $cycle <= $horizon; $cycle++ ) {
+			$decision = $engine->resolve( self::context_for( $product, $base, $cycle ) );
```
dkoo: **F5 (perf):** this runs `resolve()` HORIZON(=24)× per product **and per variation** on every list load — ~2,880 decide/compose/condition passes for 40 products × 3 variations. The repository set is request-cached (not a DB N+1), but the compose work isn't memoized. Adopting `project_for_product()` (F4) inherits the projector's `$product_memo`; otherwise memoize per `product_id` within the request. Worth measuring on a large catalog before merge.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/product-form.tsx
```diff
@@ -0,0 +1,537 @@
...
+									__next40pxDefaultSize
+								/>
+							) }
+						</VStack>
+					) }
+					<CheckboxControl label={ __( 'This is a donation', 'newspack-plugin' ) } checked={ isDonation } onChange={ setIsDonation } />
+				</VStack>
+			</Grid>
+
+			{ isEdit && initial && ( initial.unlocks.length > 0 || initial.policy.policies.length > 0 ) && (
```
dkoo: **F6:** the edit form's "Policies & access" section renders only the row-level `initial.policy` (the representative / lowest-priced variation). For a variable sub where a **non-representative** variation carries a rule, this hides or misrepresents it — even though the API now resolves `variation.policy` per plan. Consider rendering per-variation policies here.

dkoo (changes requested):
@miguelpeixe Human review of the effective-price integration — code review + hands-on manual QA on an integrated env (#464 now merged + #465 + engine **PR #2**, WCS 9.0 + Stripe test). **Requesting changes** — one display **blocker** (encoded entities in the chips) plus a correctness cluster (F3/F4) and a couple of medium items. The happy path is solid: base→effective, per-variation, and §21 cross-surface all match the engine's resolve, and donations/bundles correctly degrade to base in the integrated build.

## 🔴 Blocker
- **HTML entities render literally in the "Applied rules" chips** (inline at the resolver). A rule named `A & B – C` shows `&#038;`/`&#8211;`. One-line fix, and this PR's own modal bridge already does it.

## Correctness / quality (inline)
- **F3 + F4** — the resolver hand-rolls the cycle walk with an injected `_subscription_price` base instead of calling the public `Schedule_Projector::project_for_product()`; diverges from the storefront base for APFS-scheme products (measured $6 vs $20). The projector fixes both the divergence and the duplication.
- **F2 (display-only)** — the chip list uses `for_context()` (scope+window) not `matching_rules()`, so condition-gated rules over-list. Manual QA confirmed the effective *price* is correct — chip-accuracy, not a money bug — but still worth `matching_rules()`.
- **F5 (perf)** — up to 24×N un-memoized `resolve()` per page. **F6** — the edit form shows only the representative variation's policy.

## Design question (donations & bundles)
Manual QA confirmed the Donations and Plan-bundles tabs are **correctly empty** — those products are engine-excluded, so no rule applies (this also resolves the earlier donation-pricing concern). That raises a design question: **if those products can never be affected, why show the "Applied rules" and "Effective price" columns on those tabs at all — and why let a rule target donation/grouped products in the first place?** Worth aligning the UI (and maybe the scope picker) with the exclusion.

## Lower / notes
- **APFS-in-Plans (low):** the simple+APFS product is omitted from the Plans list (query lists only subscription types), yet it shows as *affected* in the Pricing Rules impact preview — the two surfaces disagree. (List-scope is #464's, surfaced here.)
- **Winner attribution (nit):** with a compose-`min` price tie, the "winning" chip is order-dependent (e.g. a flat $5 rule ties a 50%-off rule on a $10 product).
- **F7:** `DEFAULT_CURRENCY` missing separator fields.
- Confirmed **intentional** (not issues): pricing model is fixed on edit by design, and Priority/compose/lock are preset-and-hidden for named goals (Custom exposes them).

## Suggestions (non-blocking)
Highlight the winning rule in the chip/popover (there's currently no visual winner indication); link the Plans applied-rule chips to the rule edit view.

The core resolve is correct and the cross-surface agreement is solid — the blocker + F3/F4 are the things to land before merge. Happy to re-verify.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
miguelpeixe: Fixed in 17f0377d5 — `html_entity_decode( get_the_title( (int) $rule_id ), ENT_QUOTES )` on the chip label.

The second spot you flagged (the `project_schedule` copy at ~L190) is gone entirely — the schedule now comes from the engine's `Schedule_Projector`, so there's no second `get_the_title()` to decode. See the F3/F4 thread.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
miguelpeixe: Fixed in 17f0377d5 — the chip list now comes from `$engine->matching_rules( self::context_for( $product, $base, 1 ) )` instead of the repository's `for_context()`. `matching_rules()` runs the strategy `applies_to` + `passes_conditions` gate, so condition-gated (segment/reader) rules no longer over-list as chips.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
miguelpeixe: Fixed in 17f0377d5 — replaced `project_schedule()` with the engine's public `Schedule_Projector::project_for_product()`, and the base now comes from `Amount_Calculator::base_price_for()`.

No engine change was needed after all. The winner attribution that would have required the rule-id-carrying variant is **unused by the UI**: `is_winning` and the segment `rule_id`/`rule_label` are never read in `policy-cells.tsx` — the chips deliberately don't single out a winner, and the schedule popover renders only cycle + amount. So I dropped those fields and adopted `project_for_product()` as-is. That fixes the base divergence (F3 — now $20 via `base_price_for()`, not the injected `_subscription_price`), removes the forked walk (F4), and inherits the projector's per-request memo (F5).

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscription-policy-resolver.php
miguelpeixe: Addressed via F4 in 17f0377d5 — adopting `project_for_product()` inherits the projector's per-request `$product_memo`, so the 24-cycle walk is memoized per product within the request rather than re-run on each read. (Per-variation resolution still runs once per variation, but each is now memoized.)

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/product-form.tsx
miguelpeixe: Fixed in 17f0377d5 — the edit form's "Policies & access" section now lists per-plan policies for variable subscriptions: it maps over each priced variation and renders that variation's `variation.policy` (labeled by plan), instead of only the row-level representative `initial.policy`. A simple/single subscription still shows its one resolution, unlabeled.

miguelpeixe:
@dkoo thanks for the thorough review + manual QA. Rebased on `main` (resolved the add/add conflicts left by the squash-merged #464 Layer 1) and pushed the fixes in 17f0377d5. Point by point:

**🔴 Blocker (entities in chips)** — fixed with `html_entity_decode( …, ENT_QUOTES )` on the chip label. The `project_schedule` copy is gone (see F3/F4), so there's no second decode site.

**F2 (chip over-listing)** — chips now come from `matching_rules()` (runs the condition gate) instead of `for_context()`.

**F3 + F4 (base divergence + duplication)** — `project_schedule()` is replaced by the engine's public `Schedule_Projector::project_for_product()`, with the base from `Amount_Calculator::base_price_for()`. This needed **no engine change**: the per-segment winner attribution that would've required the rule-id-carrying variant is unused by the UI (`is_winning` / `rule_id` / `rule_label` are never read in `policy-cells.tsx`), so I dropped those fields and used the projector as-is — fixing the $6-vs-$20 base and inheriting its memo.

**F5 (perf)** — inherited from the projector's per-request `$product_memo`.

**F6 (edit form)** — the "Policies & access" section now renders per-plan policies for variable subs.

**F7 (`DEFAULT_CURRENCY` separators)** — added `decimal_separator` / `thousand_separator` to the fallback (the type required them; only the live payload had them).

**Design question (donations & bundles)** — hid the "Applied rules" and "Effective price" columns on the Donations and Bundles tabs (engine-excluded → always empty). The scope-picker half (why a rule can *target* donation/grouped products) lives in the Pricing Rules UI (#465), so I left that there.

**Winner-attribution nit** — moot now: since the winner flag was unused, I removed it rather than fix the tie-order.

**Suggestions** — "link chips to the rule edit view" and "highlight the winning rule" left as follow-ups; the latter is what would need the rule-id-carrying projector variant you described.

**APFS-in-Plans (low)** — that's the #464 list-scope surfaced here; leaving it for that PR.

Happy to re-verify.

dkoo (approved):
**Approving.** All eight items from the previous review are addressed, and the two remaining follow-ups are cheap and non-blocking. Re-reviewed at `17f0377d5` and re-confirmed at the current head `a6d8e309` (pure `origin/main` merge — the four #466-owned files are byte-identical between the two, and the auto-resolved merge leaves every main-side file identical to `main`).

### First, a correction we owe you

Our previous review asserted that "the winner is surfaced inside the schedule popover via `is_winning`," and a QA-table row repeated it. **That was wrong, and your pushback was right.** We verified it two ways:

1. At the old head, `git grep 'is_winning\|rule_label' 15b5e7c12 -- plugins/newspack-plugin/src` matches **only** `types.d.ts:16,28` — no `.tsx` referenced any of the three fields.
2. `policy-cells.tsx`: `PolicyChips` renders `p.id`/`p.label`/`p.adjustment_label` only; `ScheduleList` renders `seg.from_cycle`/`seg.amount` only.

The winner was only ever surfaced *implicitly* (segment 0's amount is the headline effective price), never explicitly. Dropping the three fields loses nothing user-visible, and `types.d.ts` was updated in the same commit, so there's no type drift. Our claim was inferred from the payload flow rather than a rendered-UI check — our mistake.

### Verification

| Item | Verdict |
|---|---|
| **B2** entity-encoded chips | **Fixed** — `html_entity_decode(…, ENT_QUOTES)` at resolver `:136`; the second decode site is genuinely gone with `project_schedule()`, and projector segments carry no label at all, so the chip is the only title-bearing string left. Confirmed in-browser. |
| **F2** chip over-listing | **Fixed** — `:129` now uses `matching_rules()`, which runs `is_excluded` + `applies_to()` + `passes_conditions()` (engine `class-pricing-engine.php:95-118`) — exactly the gate `for_context()` skipped. Confirmed in-browser: the segment-gated rule no longer chips. |
| **F3** base divergence | **Fixed**, with a narrow residual (below). |
| **F4** duplicated cycle-walk | **Fixed** — `project_for_product()` replaces the fork; the engine's `walk()` is line-for-line what was re-implemented, and derives base internally. No engine change needed. |
| **F5** perf | **Partial** — see below. |
| **F6** per-plan policies | **Fixed** — `policyRows` per priced variation, gate switched to `policyRows.length`. |
| **F7** currency separators | **Fixed** — kills the literal `"undefined"` thousands separator in `formatAmount`. |
| **Hiding the columns on Donations / Bundles** | **Right call.** Two-layer hiding is correct (dropping from `fields` removes the picker entry, from `view.fields` the rendered column), and tab switching remounts the list, so the mount-only initializer is safe. These products are engine-excluded, so the columns are structurally always empty — that's noise, not information. Blocking rules from *targeting* them is a complementary guarantee that belongs to #465, not an alternative. |

### Follow-ups (non-blocking)

1. **`base_price` is no longer rounded.** `build()` rounds `effective_price` but passes `base_price` through raw, while `EffectivePrice` decides whether to draw the `base → effective` arrow with a strict `!==` and `schedule[0].amount` *is* rounded. A price with >2 decimals could render a pointless `$20.00 → $20.00`. One-line hardening: `round( (float) $base_price, $decimals )` in `build()`.
2. **F3 residual.** `class-audience-subscription-products.php:1031-1034` still short-circuits to `empty_policy()` when `_subscription_price` is empty, so a subscription with no meta price never reaches the resolver and renders "—" even though the engine would price it. The meta is still the gatekeeper, one layer up from the fix.
3. **F5 is structurally right but the cost is unchanged.** `project_for_product()` *is* memoized, so the claim is literally true — but the memo only dedupes repeat lookups of the same product, and the list resolves each product/variation exactly once. Per-page cost stays at 24 × (products + variations) engine passes. Bounded by a catalog that's realistically a few dozen rows in one admin request, so we're not asking for a fix — worth a note only so the cost is documented if this ever grows a per-reader or storefront caller.
4. **Stale docblocks.** `policy-cells.tsx:6-7` and `:93` still promise winner attribution the payload no longer carries; the PHP side was updated but these weren't. Also still open from the original review: the resolver's `@type string $type Policy type: promo|season|winback|loyalty` (`:57`) is stale vs. the `strategy_id` actually passed.
5. **Latent, no action needed.** The Price column reads `_subscription_price` meta while the effective-price base now comes from `base_price_for()`. We flagged this as a probable user-visible inconsistency — **on measurement it isn't one**, and we're recording the correction rather than the original claim. We swept every product the list query returns plus every variation: all nine subscription/variation rows agree exactly, because `WC_Subscriptions_Product::get_price()` resolves to the same meta. We also engineered a drift (`_regular_price` 20 vs `_subscription_price` 12) and `base_price_for()` still returned 12. The only diverging type is `grouped` (base reads 0.00, since a grouped product is neither a subscription nor priced) — and grouped products render *exclusively* in the Plan bundles tab, which this PR removes both columns from. So it's masked by this very fix. Worth knowing only if the column ever returns for grouped rows.

Also verified benign: the schedule now resolves under trigger `'product_inspection'` while chips use `'subscription_products'` — no matcher, strategy, or repository reads `$ctx->trigger` anywhere in the engine, so there's no behavioral risk today, only a latent coupling if a trigger-sensitive rule ever lands.

### QA

Manual browser QA passed on an isolated env running the integrated #464+#465+#466 build against engine PR #2: decoded chips, correct chip gating, columns absent from both the table and the picker on Donations/Bundles and restored on tab-switch, and base→effective values correct across all rows. One item — F6's per-plan policy rows — is **not** verified in-browser, blocked by an unrelated regression on `main` that renders the single-plan edit view blank. We're root-causing that separately; it reproduces on plain `main` with none of this PR's code, so it isn't yours and shouldn't hold this up. F6 is verified by code reading.

dkoo:
Follow-up on this from the prior comment:

> One item — F6's per-plan policy rows — is not verified in-browser, blocked by an unrelated regression on `main` that renders the single-plan edit view blank. We're root-causing that separately; it reproduces on plain `main` with none of this PR's code, so it isn't yours and shouldn't hold this up. F6 is verified by code reading.

#672 addresses the regression at the component level. It was unrelated to any work for Dynamic Pricing.

miguelpeixe:
Follow-ups from the approving review addressed in b596cacfa, point by point:

**1. `base_price` no longer rounded** — Fixed as suggested: `build()` now rounds both prices to store decimals (null-safe), so the strict `!==` in `EffectivePrice` can't draw a `$20.00 → $20.00` arrow from an unrounded base.

**2. F3 residual (`_subscription_price` as gatekeeper)** — Fixed. `resolve_policy()` no longer short-circuits on a null meta price; the null base is forwarded and the resolver decides. With the engine active it derives its own base (`base_price_for()`), so a meta-less product now resolves and displays. The original Copilot guarantee is preserved on both remaining edges: the no-engine/invalid/excluded path passes the null through (em dash, never a coerced `0.0`), and a *genuinely* unpriced product (no catalog price, derived base `<= 0`, no rules) returns null pricing rather than a fabricated `$0.00` row — while an unpriced product that a rule *does* price resolves normally, matching what the engine-filtered storefront would charge.

**3. F5 cost documentation** — Added the note you asked for at the projector call site: the memo dedupes repeat lookups of the *same* product within the request, not distinct products, so a list request walks the horizon once per row — fine for an admin catalog read; a per-reader/storefront caller would need a broader cache first.

**4. Stale docblocks** — Swept: `policy-cells.tsx` header + `EffectivePrice` docblock no longer promise winner attribution, and the resolver's `@type string $type` now documents the strategy id actually passed. Also caught three spots beyond the list: README (§Layer 2 "winning policy distinguished" and the seam section's "winner flagged") and the decorative `promo|season|winback|loyalty` union in `types.d.ts`. While in there: added the missing `schedule` key to `resolve()`'s return docblock and made `base_price`/`effective_price` honestly `number | null` in the TS types (with the `EffectivePrice` guard narrowing both).

**5. Price-column/`base_price_for()` latency** — No action, per your note; thanks for recording the measurement.

From the earlier review's non-blocking suggestions: the **applied-rule chips now link to the rule's edit view** (`newspack-audience-pricing-rules#/edit/{id}`), with an underline-on-hover affordance. **Winner highlighting** is deliberately not added — it would reverse the winner-attribution drop this review endorsed, and doing it honestly needs the projector to carry per-segment rule attribution (plus a tie-break story for the compose-`min` case you flagged). Happy to revisit as an engine-side follow-up if product wants it.

PR #470, opened 2026-06-30
fix(popups): harmonize overlay prompt preview with inline

jason10lee:
### All Submissions:

Overlay prompt previews now apply the same capability and post-type checks the
inline placement already uses, so both placements behave consistently: preview
content is loaded only for users who can manage prompts, and only for the
prompts CPT.

Part of NPPM-2962.

1. Run the model tests: `n test-php --filter ModelTest` (11 pass).
2. As an admin, preview a prompt from the editor (overlay and inline) — both render as before.
3. As a logged-out visitor, the overlay preview now shows nothing, matching the inline placement.

### Other information:

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-model.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-model.php
@@ -290,6 +290,14 @@ final class Newspack_Popups_Model {
 	 * @return array|null Popup object array, or null if the post id does not resolve to a post.
 	 */
 	public static function retrieve_preview_popup( $post_id ) {
+		// Previews render unsaved prompt content, so restrict them to users who can manage
+		// prompts and to the prompts CPT, mirroring the plugin's other preview entry points.
+		if ( ! Newspack_Popups::is_user_admin() ) {
+			return null;
+		}
+		if ( Newspack_Popups::NEWSPACK_POPUPS_CPT !== get_post_type( $post_id ) ) {
+			return null;
+		}
 		// Up-to-date post data is stored in an autosave.
 		$autosave    = wp_get_post_autosave( $post_id );
 		$post_object = $autosave ? $autosave : get_post( $post_id );
--- a/plugins/newspack-popups/includes/class-newspack-popups-model.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-model.php
@@ -284,10 +284,14 @@ final class Newspack_Popups_Model {
 	}
 
 	/**
-	 * Retrieve popup preview CPT post.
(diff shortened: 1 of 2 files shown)


plugins/newspack-popups/includes/class-newspack-popups-model.php
```diff
@@ -290,6 +290,14 @@ public static function get_preview_query_options() {
 	 * @return array|null Popup object array, or null if the post id does not resolve to a post.
```
adekbadek: **Suggestion:** Docblock no longer describes the return contract — The function now also returns `null` when the capability check fails or the post isn't the prompts CPT, but the docblock still documents only the not-found case (`@return array|null … or null if the post id does not resolve to a post`). Update the `@return` line (and ideally the one-line summary at :287) to state the access gating, so the contract is discoverable without reading the body.

plugins/newspack-popups/tests/test-model.php
```diff
@@ -255,6 +255,102 @@ public function test_retrieve_preview_popup_with_invalid_id() {
 		);
 	}
 
+	/**
+	 * A preview should only load content for users who can manage prompts.
+	 */
+	public function test_retrieve_preview_popup_denies_logged_out_user() {
```
adekbadek: **Nit:** Test doesn't isolate the capability gate — `test_retrieve_preview_popup_denies_logged_out_user` creates a plain (non-prompt) post, so *both* gates would deny it — it doesn't isolate the capability gate as the name implies. The intended isolation is already covered by `test_retrieve_preview_popup_denies_logged_out_user_for_prompt` (prompt CPT + logged out). Harmless overlap; worth a tighter comment or dropping the case.

adekbadek (approved):
Works as described, left two non-blocking comments.

plugins/newspack-popups/includes/class-newspack-popups-model.php
jason10lee: Addressed — the docblock now documents the access contract: the summary notes it is for users who can manage prompts, and the `@return` spells out the null cases (cannot manage prompts, not a prompt, or the id does not resolve).

plugins/newspack-popups/tests/test-model.php
jason10lee: Dropped the redundant case — the capability gate is isolated by `test_retrieve_preview_popup_denies_logged_out_user_for_prompt` and the CPT gate by `test_retrieve_preview_popup_denies_non_prompt_post_type`, so the plain-post-both-gates test added no unique coverage.

PR #471, opened 2026-06-30
fix(blocks): limit articles endpoint to viewable post types

jason10lee:
### All Submissions:

The public `/articles` REST endpoint now restricts the requested `postType` to
publicly viewable post types, matching what WordPress exposes on the front end.
Non-viewable post types are dropped; a request for only non-viewable types
returns no results. The shared query builder is unchanged, so the editor
endpoints and block rendering are unaffected.

Part of NPPM-2964.

1. Run the block tests: `n test-php --filter HomepagePostsBlockTest` (6 pass).
2. As a logged-out visitor, call `/wp-json/newspack-blocks/v1/articles` with a non-public post type — no results are returned.
3. Posts and publicly viewable custom post types are returned as before.

### Other information:

### Accepted trade-off (from review)

Because `/articles` now filters requested post types through `is_post_type_viewable()`, the Sponsors CPT (`public=false`, deliberately listable in Homepage Posts / Carousel per [newspack-sponsors#52](https://github.com/Automattic/newspack-sponsors/pull/52)) is dropped from the endpoint. A Sponsors-only block still renders **page 1 via SSR**, but **load-more / infinite-scroll returns empty** — graceful, and only for a Sponsors-only block with more sponsors than the page size.

This is intentional: the predicate must **not** be loosened to `show_in_rest`, since `public=false` + `show_in_rest=true` CPTs (`np_content_gate`, `newspack_popups_cpt`, …) would then be re-exposed. Restoring Sponsors load-more behind a dedicated allow-list filter is tracked as a follow-up in **NPPM-2972**.

### DIFF
--- a/plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -129,6 +129,21 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 			$exclude_ids = [];
 		}
 
+		// This endpoint is public, so restrict it to publicly viewable post types — matching
+		// what WordPress exposes on the front end. The editor endpoints are capability-gated.
+		$attributes['postType'] = self::filter_viewable_post_types( isset( $attributes['postType'] ) ? $attributes['postType'] : [ 'post' ] );
+		if ( empty( $attributes['postType'] ) ) {
+			// Every requested post type was non-viewable; return nothing rather than
+			// substituting a different post type.
+			return rest_ensure_response(
+				[
+					'items' => [],
+					'ids'   => [],
+					'next'  => '',
+				]
+			);
+		}
+
 		$article_query_args = Newspack_Blocks::build_articles_query( $attributes, apply_filters( 'newspack_blocks_block_name', 'newspack-blocks/homepage-articles' ) );
 
 		// If using exclude_ids, don't worry about pagination. Just get the next postsToShow number of results without the excluded posts. Otherwise, use standard WP pagination.
@@ -199,6 +214,20 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 				'next'  => $next_url,
 			]
 		);
+	}
+
+	/**
+	 * Restrict a list of post types to those that are publicly viewable.
+	 *
+	 * Used to keep the public articles endpoint from exposing post types that
+	 * WordPress would not surface on the front end. Non-viewable types are dropped;
+	 * the returned list may be empty when none qualify (the caller then returns no results).
+	 *
+	 * @param array|string $post_types Requested post type(s).
+	 * @return array Viewable post types (may be empty).
+	 */
+	private static function filter_viewable_post_types( $post_types ) {
+		return array_values( array_filter( (array) $post_types, 'is_post_type_viewable' ) );
 	}
 
 	/**
--- a/plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
+++ b/plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
@@ -131,17 +131,11 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 
 		// This endpoint is public, so restrict it to publicly viewable post types — matching
 		// what WordPress exposes on the front end. The editor endpoints are capability-gated.
-		$attributes['postType'] = self::filter_viewable_post_types( isset( $attributes['postType'] ) ? $attributes['postType'] : [ 'post' ] );
+		$attributes['postType'] = self::filter_viewable_post_types( $attributes['postType'] );
 		if ( empty( $attributes['postType'] ) ) {
 			// Every requested post type was non-viewable; return nothing rather than
 			// substituting a different post type.
-			return rest_ensure_response(
-				[
-					'items' => [],
-					'ids'   => [],
-					'next'  => '',
-				]
-			);
+			return self::articles_response();
 		}
 
 		$article_query_args = Newspack_Blocks::build_articles_query( $attributes, apply_filters( 'newspack_blocks_block_name', 'newspack-blocks/homepage-articles' ) );
@@ -207,11 +201,23 @@ class WP_REST_Newspack_Articles_Controller extends WP_REST_Controller {
 			);
 		}
 
+		return self::articles_response( $items, $ids, $next_url );
+	}
+
+	/**
+	 * Build the articles endpoint response.
+	 *
+	 * @param array  $items Rendered article items.
+	 * @param array  $ids   Post ids in the response.
+	 * @param string $next  URL for the next page, or empty string when there is none.
+	 * @return WP_REST_Response
+	 */
+	private static function articles_response( $items = [], $ids = [], $next = '' ) {
 		return rest_ensure_response(
 			[
 				'items' => $items,
 				'ids'   => $ids,
-				'next'  => $next_url,
+				'next'  => $next,
 			]
 		);
 	}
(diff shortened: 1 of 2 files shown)


plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -79,4 +79,153 @@ public function test_hpb_wp_query() {
...
+			[
+				'post_type'   => 'newspack_secret_cpt',
+				'post_status' => 'publish',
+			]
+		);
+		wp_set_current_user( 0 );
+
+		$controller = new WP_REST_Newspack_Articles_Controller();
+		$request    = new WP_REST_Request( 'GET', '/newspack-blocks/v1/articles' );
+		$request->set_param( 'postType', [ 'newspack_secret_cpt' ] );
```
adekbadek: **Suggestion:** The realistic specific-posts attack is `postType=['post']` (which survives the viewability filter) combined with `specificPosts=[<non-viewable id>]`. As written, this test sets `postType=['newspack_secret_cpt']`, which `filter_viewable_post_types()` empties, so the request short-circuits on the empty-guard and never reaches `build_articles_query()` — the `WP_Query` `post_type AND post__in` semantics that actually block the attack stay unexercised. Consider adding a case with a viewable `postType` and a non-viewable `specificPosts` ID so the real guarantee is locked in.

plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -129,6 +129,21 @@ public function get_items( $request ) {
 			$exclude_ids = [];
 		}
 
+		// This endpoint is public, so restrict it to publicly viewable post types — matching
+		// what WordPress exposes on the front end. The editor endpoints are capability-gated.
+		$attributes['postType'] = self::filter_viewable_post_types( isset( $attributes['postType'] ) ? $attributes['postType'] : [ 'post' ] );
```
adekbadek: **Suggestion:** Heads-up on a bounded, accepted side effect (not a blocker): listing sponsors in Homepage Posts / Carousel is a deliberate feature (newspack-sponsors#52 + newspack-blocks#746) — the Sponsors CPT is `public=false` + `show_in_rest=true` + `supports 'newspack_blocks'`, so it's selectable in the block. `is_post_type_viewable()` returns false for it, so this filter drops it from `/articles`: the first page still renders via SSR but load-more / infinite-scroll returns empty (graceful, no error; only affects a Sponsors-only block with more sponsors than the page size). This is the right call — the predicate must **not** be loosened to `show_in_rest`, since `public=false` + `show_in_rest=true` config CPTs (`np_institution`, `np_content_gate`, `newspack_popups_cpt`, …) would be re-exposed by `show_in_rest` but are correctly blocked by `is_post_type_viewable()`. **Please note this trade-off in the PR description** (link #52). **Suggested follow-up** (separate PR, so this hotfix stays minimal): restore Sponsors load-more by widening the predicate to `is_post_type_viewable( $type ) || post_type_supports( $type, 'newspack_blocks' )`, ideally behind a `newspack_blocks_articles_allowed_post_types` filter that newspack-sponsors hooks — noting it re-permits anonymous enumeration of sponsor bodies platform-wide, which is a conscious (low-sensitivity) widening.

adekbadek: **Nit:** The `isset( $attributes['postType'] ) ? ... : [ 'post' ]` fallback is effectively unreachable: `$attributes` is built via `wp_parse_args( ..., <schema defaults> )` just above and the `postType` schema default is `['post']`, so the key is always set. This can be simplified to `self::filter_viewable_post_types( $attributes['postType'] )`. (It does lock in one benign behavior change: an explicit `postType => []` now yields an empty result instead of falling back to `post` — safer, and no legitimate caller does that.)

plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -79,4 +79,153 @@ public function test_hpb_wp_query() {
 		self::assertEquals( 1, count( $query->posts ), 'There is one post returned.' );
 		self::assertEquals( $post_id, $query->posts[0]->ID, 'The post returned is the one with the CAP author assigned.' );
 	}
+
+	/**
+	 * The public /articles endpoint must not return posts from non-viewable post types.
+	 */
+	public function test_articles_endpoint_excludes_non_viewable_post_types() {
+		register_post_type(
```
adekbadek: **Nit:** `unregister_post_type()` runs only on the happy path at the end of each test; a mid-test assertion failure throws before it, leaking the registered CPT into later tests. A shared helper that cleans up in `tearDown()` (or try/finally) would make the suite order-independent. The identical `register_post_type( 'newspack_secret_cpt', ... )` block is also copy-pasted across three tests — a small `register_non_viewable_cpt()` helper would name the intent and drop the boilerplate.

plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
```diff
@@ -129,6 +129,21 @@ public function get_items( $request ) {
...
+		if ( empty( $attributes['postType'] ) ) {
+			// Every requested post type was non-viewable; return nothing rather than
+			// substituting a different post type.
+			return rest_ensure_response(
```
adekbadek: **Nit:** The empty-response literal `[ 'items' => [], 'ids' => [], 'next' => '' ]` is duplicated here and at the final `return`. Minor, but if a field is ever added to the response shape it's easy to update one path and miss the other.

adekbadek (approved):
This closes the unauthenticated post-type disclosure cleanly. `is_post_type_viewable()` is the right predicate — stricter than the suggested `show_in_rest` and matching what WordPress surfaces on the front end.

Two things worth a look, neither blocking: a Homepage Posts / Carousel block set to list the Sponsors post type (selectable, but `public=false`) still renders its first page via SSR while load-more now returns empty — a graceful but silent divergence worth noting in the PR description as an accepted trade-off; and the specific-posts test exercises the short-circuit path rather than the realistic attack. Plus a few nits.

adekbadek (approved):
_(Duplicate submission — double-posted by tooling. The active review with inline comments is [the one above](https://github.com/Automattic/newspack-workspace/pull/471#pullrequestreview-4607238924); its 5 inline comments here have been removed.)_

plugins/newspack-blocks/tests/test-homepage-posts-block.php
jason10lee: Added `test_articles_endpoint_excludes_non_viewable_specific_post_under_viewable_type` — viewable `postType=[post]` plus a non-viewable `specificPosts` id — which reaches the WP_Query `post_type` + `post__in` intersection and asserts the id is not surfaced. Kept the existing short-circuit test too.

plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
jason10lee: Confirmed as an accepted trade-off and noted in the PR description (link to newspack-sponsors#52). Restoring Sponsors load-more behind a `newspack_blocks_articles_allowed_post_types` allow-list filter is tracked in NPPM-2972 — kept out of this hotfix so it stays minimal, and per your note the predicate is deliberately not loosened to `show_in_rest`.

jason10lee: Simplified to `self::filter_viewable_post_types( $attributes[postType] )` — `wp_parse_args` with the schema default always sets the key, so the fallback was dead code.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
jason10lee: Extracted `register_non_viewable_cpt()` / `register_viewable_cpt()` helpers and moved cleanup to `tear_down()` (tracking registered types), so the suite is order-independent and the boilerplate is gone.

plugins/newspack-blocks/src/blocks/homepage-articles/class-wp-rest-newspack-articles-controller.php
jason10lee: Deduped via a private `articles_response( $items, $ids, $next )` helper used by both the empty-guard and the final return.

PR #472, opened 2026-06-30
feat: explicit breadcrumb API for Newspack admin headers

thomasguillot:
### All Submissions:

Standardizes the Newspack admin header across every wizard and admin screen on a single shared `Page` header (built on `@wordpress/admin-ui`) with an explicit, url-driven breadcrumb API.

Previously each surface derived its breadcrumbs through a different mechanism (route matching, DOM patching, ad-hoc trails). This replaces all of them with one `{ label, url? }[]` trail authored per surface: the last item is the page's single `<h1>`, earlier items link when they carry a `url`. The result is a consistent header (breadcrumbs, branded logo, badges, and right-aligned actions) on Dashboard, Settings, Audience, Advertising, Newsletters, Story Budget, Popups, Listings, Network, Sponsors, Collections, and Multibranded surfaces.

Alongside the header work, this PR includes the related UI polish:

- **Header actions** — primary actions live in the page header instead of the content body: Ad Unit save/add, Campaigns/Segments "Add", Donations "Save", Premium Newsletters "Add".
- **Emails** — the Settings button moved into the DataViews toolbar (compact).
- **Copy** — Title Case applied to breadcrumb leaves and screen labels (Advertising, Access Control, Premium Newsletters, Audience), and "Add New X" shortened to "Add X" to match WordPress 6.8.
- **Layout** — wizard content fills to the section max-width (1006px) inside the flex-column header; page-breadcrumb min-height scoped to a dedicated `HStack` wrapper so it no longer leaks onto card headers.
- **Fixes** — `withWizardScreen` no longer requires a Router (standalone Multibranded no longer crashes); discarding unsaved changes from a breadcrumb/back link now navigates correctly.

07-dashboard
27-newsletters-layouts
02-advertising-placements
levels-4

Part of DSGNEWS-143. (Tabbed navigation is still to be sorted out in follow-up work.)

1. Rebuild: `pnpm --filter newspack-components build`, then `pnpm --filter newspack build` (and `newspack-newsletters`, `newspack-story-budget`, `newspack-popups`, `newspack-multibranded-site`). Hard-reload admin pages (cache-bust) between builds.
2. Visit each admin surface and confirm the header shows the correct breadcrumb trail with exactly one `<h1>` (the last crumb), the Newspack logo, and any actions right-aligned: Dashboard, Settings (all tabs incl. Additional Brands), Audience (Configuration/Payment/Emails/Advanced Settings, Campaigns, Segments, Access Control, Donations), Advertising (Providers/Placements/Settings, Google Ad Manager, Add/Edit Ad Unit), Newsletters (list/Layouts/Advertising/Advertisers/Settings/Premium), Story Budget, Sponsors, Collections, Listings, Network.
3. Confirm breadcrumb links: on Advertising GAM sub-screens, "Providers" and "Google Ad Manager" are links; on Access Control sub-screens (Metered Countdown, Content Gifting, Institutions) "Access Control" links back to the gates list.
4. Advertising: confirm "Display Ads", "Add New Ad Unit"→"Add Ad Unit", "Edit Ad Unit" copy; Save/Cancel and Add Ad Unit render in the page header.
5. Audience: confirm "Add Campaign" and "Add Segment" render in the page header (and only on the list, not the edit screen); the create-campaign modal still opens. Donations shows "Save" in the header. On the Emails screen the "Settings" button sits in the DataViews toolbar (compact) and opens the settings modal.
6. Access Control / Premium Newsletters: verify Title Case labels (Access Control, Content Rules, Registered/Paid Access, Metered Countdown, Content Gifting, Advanced Settings, Auto Signup, Premium Newsletters, Add Premium Newsletter).
7. Unsaved-changes guard: on Content Gifting / Metered Countdown, change a value, click the "Access Control" breadcrumb (or the back arrow) → confirm the discard modal appears; "Discard changes" navigates to the gates list; "Cancel" keeps you on the page with changes intact and the URL restored.
8. Layout: on card-heavy wizards (e.g. Campaigns, Segments) confirm the content fills to ~1006px and card headers are not forced to the header min-height.
9. Standalone Multibranded (without the Newspack plugin active): confirm the Brands admin screen renders without a Router crash.

### Other information:


packages/components/src/page/index.js
```diff
@@ -0,0 +1,56 @@
...
+				</HStack>
+			}
+			badges={ badges }
+			actions={ actions }
+			hasPadding={ false }
+		>
+			{ children }
+		</AdminUIPage>
+	);
+};
```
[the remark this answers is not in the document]
thomasguillot: Good catch. Rather than drop it, I wired `subTitle` through to admin-ui's `Page` (we may reintroduce subtitles later) in 66e5d0d3c — subtitles now render and the test asserting it passes.

plugins/newspack-plugin/includes/wizards/class-listings-wizard.php
```diff
@@ -124,7 +124,8 @@ public function add_page() {
 			'edit_posts', // Copied from Listings plugin...see docblock note above.
 			$this->slug,
 			'',
-			$icon
+			$icon,
+			'6.7'
 		);
```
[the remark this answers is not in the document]
thomasguillot: Changed to numeric 6 to match the Collections CPT (menu_position => 6) and Story Budget menus. Worth noting the position is effectively cosmetic for Listings: it is a wizard, so its final placement under the Newspack menu comes from the menu_order filter in class-wizards.php (via parent_menu_order = 3), which overrides this argument on every admin load.

packages/components/src/breadcrumbs/index.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * WordPress dependencies.
+ */
+import { __ } from '@wordpress/i18n';
+import { Link, Stack, Text } from '@wordpress/ui';
```
adekbadek, line 5: **Suggestion:** Shared header imports undeclared packages — `Breadcrumbs` imports `Link, Stack, Text` directly from `@wordpress/ui`, but `packages/components/package.json` declares only `@wordpress/admin-ui` (2.4.0). The webpack `FORCE_BUNDLE` set in `packages/scripts` additionally bundles `@wordpress/theme` and `@wordpress/style-runtime`, neither declared in any Newspack manifest. All three resolve today only via a transitive hoist of admin-ui's dependency tree (`@wordpress/ui` is an unstable `0.x`). A directly-imported package should be a direct dependency: declare all three in `packages/components` (pinned to the versions admin-ui 2.4.0 currently resolves) so a lockfile regen or an admin-ui bump can't silently break the build.

packages/components/src/breadcrumbs/index.js
```diff
@@ -0,0 +1,63 @@
...
+						) }
+						<Text variant="body-lg" aria-hidden="true" className="newspack-breadcrumbs__separator">
+							/
+						</Text>
+					</li>
+				) ) }
+				{ last && last.label && (
+					<li>
+						{ /* eslint-disable-next-line jsx-a11y/heading-has-content -- content is supplied via the Text children through @wordpress/ui's render prop. */ }
+						<Text variant="heading-lg" render={ <h1 /> } className="newspack-breadcrumbs__current">
```
adekbadek, line 53: **Suggestion:** Page h1 nested inside the breadcrumb nav landmark; current crumb lacks aria-current — The page's single `<h1>` (the last crumb, line 53) is rendered inside `<nav aria-label="Breadcrumbs">` (line 33). Placing the primary page heading inside a breadcrumb navigation landmark is semantically unusual for assistive tech, and the current crumb carries no `aria-current="page"`. Consider marking the current crumb with `aria-current="page"` and, if the h1 is meant to be the document's main heading, keeping it outside the `nav` — otherwise landmark navigation surfaces the page title as breadcrumb content.

packages/components/src/wizard/index.js
```diff
@@ -53,6 +54,43 @@ const ResetHeaderData = () => {
...
+	if ( ! breadcrumbItems.length && headerText ) {
+		breadcrumbItems = [ { label: headerText } ];
+	}
+	// A section can supply render-time current-page breadcrumb(s) via
+	// headerData.sectionName — either a single label (e.g. the integration name,
+	// or Add/Edit) or an ordered array of `{ label, url? }` crumbs when the leaf
+	// needs its own linked ancestors (e.g. an integration's Logs subpage). Each
+	// crumb is appended in turn, skipping one that just repeats the current
+	// trailing label so the same label never renders twice.
+	if ( sectionName ) {
```
adekbadek: **Suggestion:** Leaf breadcrumb authored two different ways — The leaf ("current page") crumb is authored two ways across surfaces, reconciled only by a trailing-label string-equality dedup. Some routes put the leaf in the static `breadcrumbs` array *and* set an identical `sectionName` in an effect (e.g. content-gates `countdown-banner`/`content-gifting`/`institutions`), so the same label is declared twice and de-duplicated at render; other routes (the gate `Edit` route) leave the static trail at the parent and rely solely on the dynamic `sectionName`. A maintainer editing one label but not its twin gets a doubled crumb ("Metered Countdown / Metered Countdown"), and it's non-obvious which mechanism owns the leaf on any given route. Pick one convention — dynamic leaves via `sectionName`, static leaves in `breadcrumbs`, never both — and document it where the trail is assembled.

packages/components/src/wizard/index.js
```diff
@@ -53,6 +54,43 @@ const ResetHeaderData = () => {
...
+		const extraCrumbs = ( Array.isArray( sectionName ) ? sectionName : [ { label: sectionName } ] ).filter( crumb => crumb?.label );
```
adekbadek: **Suggestion:** sectionName trail assembly / dedup is untested — The `sectionName` append/dedup logic (string-vs-array normalization, the `filter(crumb => crumb?.label)`, and the trailing-label dedup) is the trickiest new code in the PR and has no test. `breadcrumbs-select.test.js` thoroughly covers `activeBreadcrumbs`, but nothing exercises the array-`sectionName` path, the string path, or the dedup that the dual-authoring above depends on. A small unit test around the trail assembly (or a pure helper extracted from it) would be the highest-value test to add, since a regression here silently doubles or drops a crumb.

plugins/newspack-newsletters/src/admin-shell/page-header.js
```diff
@@ -79,9 +80,8 @@ export default function PageHeader() {
 	}
 
 	if ( newspackHeader ) {
-		// Reuse the wizard's `__header__actions` class so its flex/spacing aligns us next to the breadcrumb.
 		return createPortal(
```
adekbadek, line 83: **Suggestion:** Header-actions portal relies on an undocumented render-once contract — The newsletters admin-shell portals its header actions into `#newspack-wizards-admin-header-actions` — a DOM node owned by a *separate* React root (newspack-plugin's admin-header bundle, `admin-header/index.tsx`). This works only because that node is rendered once as a static empty `<div>` from `window.newspackWizardsAdminHeader` and newspack-plugin never re-renders or unmounts it. If the admin-header app ever gains state and re-renders (or is remounted), React reconciliation would wipe the portalled action buttons. The `MutationObserver` handles the mount-ordering race well, but the cross-root "this node must never re-render" contract is implicit — document it at the `<div>` definition on the producing side so a future change there doesn't silently break newsletters' actions.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/index.js
```diff
@@ -82,7 +82,7 @@ export default function NewslettersListScreen() {
 				icon={ envelope }
 				title={ __( 'Get started with newsletters', 'newspack-newsletters' ) }
```
adekbadek, line 83: **Suggestion:** Empty Newsletters list renders two h1s — Verified in the browser: with zero newsletters, the list page renders **two** `<h1>`s — the breadcrumb leaf ("All Newsletters") plus the `EmptyState` `title` ("Get started with newsletters"). The `EmptyState` heading is pre-existing (this PR only touched copy here), but the PR's new header h1 now competes with it, so the "exactly one `<h1>`" invariant this PR establishes is violated in the empty state. Consider rendering the `EmptyState` title at a lower heading level — matching the h1→h2 demotion already applied to the emails `PageHeading` elsewhere in this PR. Empty-state only; populated lists show a single h1.

plugins/newspack-plugin/includes/wizards/traits/trait-wizards-admin-header.php
```diff
@@ -61,14 +61,20 @@ public function admin_header_enqueue() {
...
 		wp_enqueue_script( 'newspack-wizards-admin-header' );
 		wp_localize_script(
 			'newspack-wizards-admin-header',
 			'newspackWizardsAdminHeader',
 			[
-				'tabs'  => $this->tabs,
-				'title' => $this->title,
-			] 
+				'tabs'        => $this->tabs,
+				'breadcrumbs' => array_values( $breadcrumbs ),
```
adekbadek, line 81: **Suggestion:** Public breadcrumb filter result passed to array_values() unguarded — `array_values( $breadcrumbs )` runs on the return of the public `newspack_wizards_admin_header_breadcrumbs` filter (line 65). The trait's own default returns an array, but this is a publicly hookable filter — a third-party/customer callback returning `null` or a non-array yields a `TypeError` under PHP 8 and fatals the admin header for every screen. A cheap `is_array()` guard (falling back to `$this->breadcrumbs` or `[]`) makes the header robust to a misbehaving filter.

packages/components/src/wizard/index.js
```diff
@@ -123,7 +162,105 @@ const Wizard = (
...
+		</>
+	);
+
+	const headerActions =
+		actions?.length > 0 ? (
+			<>
+				{ mainActions.map( ( action, index ) => (
+					<Button
+						key={ index }
+						className="newspack-wizard__header__actions__main"
```
adekbadek: **Nit:** Header-action class names retain __header__ segment — The moved header-action buttons keep `newspack-wizard__header__actions__main` / `__more` class names even though there is no longer a `newspack-wizard__header` element (actions now live in `Page`'s actions slot). The SCSS still keys off these names so nothing breaks, but the `__header__` segment is now misleading — a rename would keep the CSS namespace honest.

plugins/newspack-newsletters/includes/admin/pages/class-settings-page.php
```diff
@@ -48,4 +48,16 @@ public function get_label(): string {
...
 	}
+
+	/**
+	 * Explicit admin-header breadcrumb trail.
+	 *
+	 * @return array<array{label: string}>
+	 */
+	public function get_wizard_breadcrumbs(): ?array {
+		return [
+			[ 'label' => __( 'Newsletters', 'newspack-newsletters' ) ],
```
adekbadek, line 59: **Nit:** Newsletters Settings parent crumb is not a back-link — Verified in the browser: on the Newsletters → Settings screen, the "Newsletters" parent crumb is authored without a `url` (`[ 'label' => __( 'Newsletters' ) ]`), so it renders as plain text rather than a link back to the newsletters list. Given the ticket's goal of linkable breadcrumbs for easier navigation, consider giving the parent the list URL. Low-stakes and internally consistent (the list page authors the same non-linked parent), so purely optional.

plugins/newspack-plugin/includes/wizards/newsletters/class-newsletters-wizard.php
```diff
@@ -119,8 +119,8 @@ function() {
 				// Add the admin header.
 				$this->admin_header_init(
 					[
-						'title' => $this->get_name(),
-						'tabs'  => $this->get_tabs(),
+						'breadcrumbs' => $this->admin_screens[ $this->get_screen_slug() ],
```
adekbadek: **Nit:** admin_screens indexed without the ?? [] guard used elsewhere — `admin_header_init([ 'breadcrumbs' => $this->admin_screens[ $this->get_screen_slug() ] ])` indexes `admin_screens` without the `?? []` fallback that the same lookup uses at line 367 (and that `get_header_text()` uses). A screen slug missing from the map would emit a PHP 8 "Undefined array key" warning and pass `null` through to the localized data. Mirror the `?? []` fallback here (and in the listings/network wizards) for consistency and warning-safety.

adekbadek (commented):
Not blockers, but a few things are worth taking a look.

packages/components/src/breadcrumbs/index.js
thomasguillot: Fixed in 3357c92. Declared `@wordpress/ui`, `@wordpress/theme` and `@wordpress/style-runtime` as direct dependencies of `newspack-components`, pinned to the versions admin-ui 2.4.0 currently resolves (0.15.0 / 0.15.0 / 0.4.0), so a lockfile regen or admin-ui bump can't silently drop them.

packages/components/src/wizard/index.js
thomasguillot: Fixed in 3357c92. Extracted the sectionName append/dedup into a pure `appendSectionName()` helper in `breadcrumbs-select.js` and added unit tests covering the string path, the array path, the trailing-label dedup, no-label skipping, non-adjacent duplicates, and immutability.

plugins/newspack-newsletters/src/admin-shell/page-header.js
thomasguillot: Fixed in 3357c92. Documented the render-once contract at the producing `<div id="newspack-wizards-admin-header-actions">` in `admin-header/index.tsx`, noting it's a cross-root portal target that must never re-render or gain state.

plugins/newspack-plugin/includes/wizards/traits/trait-wizards-admin-header.php
thomasguillot: Fixed in 3357c92. Added an `is_array()` guard that falls back to the default trail before `array_values()`, so a filter callback returning a non-array can't fatal the header.

plugins/newspack-plugin/includes/wizards/newsletters/class-newsletters-wizard.php
thomasguillot: Fixed in 3357c92. Mirrored the `?? []` fallback on the `admin_screens[...]` lookup in the newsletters, listings, and network wizards.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/index.js
thomasguillot: Fixed in 7e58172. The empty-state title came from `SectionHeader`'s `pageHeader`, which forced an `<h1>` — hence two on the empty list. `pageHeader` no longer sets the tag: the level follows the `heading` prop (default `h2`) and `pageHeader` now only drives the enlarged/centered styling. Title sizing moved onto a dedicated `__title` class (base-styles `heading-x-large()` mixin) so it's tag-independent — demoting the heading can't change the look, and genuine `h2`s elsewhere in a header aren't affected. The same class of bug on the Access Control and Institutions onboarding screens (both `pageHeader` consumers) and the two admin modals is demoted in the same pass. Verified before/after in the browser: exactly one `h1` per screen now; the only visual delta is the title line-height tightening 32→24px (the canonical mixin).

packages/components/src/breadcrumbs/index.js
thomasguillot: By design the `Breadcrumbs` component is the single owner of each screen's canonical `<h1>` — the trail leaf and the page title are deliberately one element — so we're keeping it as-is and not restructuring the nav landmark or the leaf's markup here. The concrete problem (more than one `h1` per page) is fixed at the source in 7e58172 by demoting every other in-content heading to `h2`, so the breadcrumb `h1` is now genuinely the only one on every screen. aria-current on the leaf is a reasonable follow-up but sits inside this component we're intentionally leaving untouched in this PR.

plugins/newspack-newsletters/includes/admin/pages/class-settings-page.php
thomasguillot: This is intentional — `Newsletters` is the top-level parent of the section, so it stays plain text rather than linking back to itself. It's the same convention across the newsletters admin pages (list, ads, advertisers all author a non-linked parent), so leaving as-is.

packages/components/src/wizard/index.js
thomasguillot: Renamed in 58703d7. The dead `__header__` segment is gone — `newspack-wizard__header__actions__*` → `newspack-wizard__actions__*` across `wizard/index.js` (the 5 class refs) and the two overflow-menu selectors in `with-wizard-screen/style.scss`, kept in sync. No other consumers in the repo.

packages/components/src/wizard/index.js
thomasguillot: Fixed in c6fe441. Each Access Control route now authors its current-page leaf in exactly one place:

- **Constant labels** (Metered Countdown, Content Gifting, Institutions list) — leaf stays in the static `breadcrumbs` array; dropped the redundant `sectionName`.
- **Runtime labels** (Add/Edit Institution) — `sectionName` owns the leaf; the static trail is ancestors-only, now with a linked `Institutions` crumb, matching the gate editor pattern. This also restores the previously-skipped Institutions step in those two trails.

`appendSectionName` now documents the rule, and the trailing-label dedup is kept only as a defensive guard rather than something any route relies on.

packages/components/src/confirm-dialog/index.tsx
```diff
@@ -76,22 +79,37 @@ function ConfirmDialog(
...
 			return;
 		}
 		const unblock = history.block( ( location: string, action: string ) => {
+			// Let our own confirmed navigation through instead of re-blocking it.
+			if ( bypassBlock.current ) {
+				return undefined;
+			}
 			pendingNavigation.current = () => {
-				unblock();
+				bypassBlock.current = true;
```
adekbadek, line 104: The bypassBlock toggles aren't wrapped in try/finally, so a throw in history.replace/history.push (here and in handleOnCancel at line 84) leaves bypassBlock stuck true — silently disabling unsaved-changes protection for the rest of the session. A try/finally around each true/false pair closes that off.

packages/components/src/confirm-dialog/index.tsx
```diff
@@ -76,22 +79,37 @@ function ConfirmDialog(
...
 		onCancel();
-	}, [ onCancel, pendingNavigation ] );
+	}, [ onCancel, history ] );
 
 	// Block navigation when there are unsaved changes.
 	useEffect( () => {
 		if ( ! when ) {
 			return;
 		}
 		const unblock = history.block( ( location: string, action: string ) => {
```
adekbadek, line 98: This block/bypass path is the one correctness-sensitive piece with no test coverage — the POP re-sync (history.replace(history.location) before pushing the target) and the REPLACE-vs-push branch are exactly the kind of subtle behavior a regression test should pin, especially since breadcrumb links now navigate via anchor POP. Worth a follow-up test.

adekbadek (approved):
Re-review with the branch checked out: all four robustness fixes from the first pass are verified in code (the public-filter is_array() guard, the ?? [] fallbacks in the listings/network/newsletters wizards, and the cross-root render-once portal doc), and the c6fe441 leaf-authoring refactor is sound — the restored Institutions middle step is an improvement and no current-page label is lost. Approving; the confirm-dialog notes below are non-blocking follow-ups, and the branch is currently in conflict with main so it needs a rebase before merge.

packages/components/src/confirm-dialog/index.tsx
thomasguillot: Fixed in 1ba90c7. Both bypassBlock pairs (the cancel re-sync and the confirmed-navigation replay) are now wrapped in try/finally, so a throw in history.replace/push can't leave the blocker disabled for the rest of the session.

packages/components/src/confirm-dialog/index.tsx
thomasguillot: Added in 1ba90c7. New confirm-dialog unit tests pin the block/bypass path: cancel stays on the current page and re-syncs the URL, a confirmed push replays with the re-sync replace first, a blocked REPLACE replays via replace rather than push, and a blocked POP (back/breadcrumb) is re-pushed after the re-sync. Two regression tests also cover the try/finally fix above (the blocker stays armed when the replayed navigation or cancel re-sync throws); both fail against the previous implementation.

thomasguillot:
Rebased on main (the wizard actions MenuGroup/separator change and the gate-summary refactor are reconciled, with the title-cased labels carried into gate-summary.tsx), and both confirm-dialog follow-ups are addressed in 1ba90c7.

plugins/newspack-plugin/src/wizards/audience/views/campaigns/index.js
```diff
@@ -21,30 +21,36 @@ import { stringify } from 'qs';
...
-const { HashRouter, Redirect, Route, Switch } = Router;
+const { HashRouter, NavLink, Redirect, Route, Switch } = Router;
 
 const headerText = __( 'Audience Management / Campaigns', 'newspack-plugin' );
 
+const ROOT = [ { label: __( 'Audience Management', 'newspack-plugin' ) } ];
+
 const tabbedNavigation = [
 	{
 		label: __( 'Campaigns', 'newpack-plugin' ),
```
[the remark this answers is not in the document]
thomasguillot: Fixed in d5b8f50 (all three, plus a pre-existing instance of the same typo in with-wizard).

plugins/newspack-plugin/src/wizards/audience/views/campaigns/index.js
```diff
@@ -21,30 +21,36 @@ import { stringify } from 'qs';
...
 		path: '/campaigns',
 		exact: true,
+		breadcrumbs: [ ...ROOT, { label: __( 'Campaigns', 'newspack-plugin' ) } ],
 	},
 	{
 		label: __( 'Segments', 'newpack-plugin' ),
```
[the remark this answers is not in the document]
thomasguillot: Fixed in d5b8f50 (all three, plus a pre-existing instance of the same typo in with-wizard).

plugins/newspack-plugin/src/wizards/audience/views/campaigns/index.js
```diff
@@ -21,30 +21,36 @@ import { stringify } from 'qs';
...
 		path: '/segments',
 		exact: false,
+		breadcrumbs: [ ...ROOT, { label: __( 'Segments', 'newspack-plugin' ) } ],
 	},
 	{
 		label: __( 'Settings', 'newpack-plugin' ),
```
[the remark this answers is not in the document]
thomasguillot: Fixed in d5b8f50 (all three, plus a pre-existing instance of the same typo in with-wizard).

PR #473, opened 2026-06-30
feat(insights): pre-warm Insights tabs with durable cache

dkoo:
### All Submissions:

**Phase 1 of NEWS-2581** — improve the perceived loading speed of the Insights dashboard. This is consumer-side only; the hub-side incremental BigQuery cache is a follow-up (Phase 2).

Today the windowed Insights tabs compute on demand and cache in short-lived transients, which this stack stores in memcached and can evict under memory pressure — so the first dashboard view (and any view after an eviction) pays the full BigQuery/WooCommerce round-trip. This PR makes the common case instant:

- **Daily pre-warm.** On `admin_init` (guarded, at most once per day), a background Action Scheduler job warms the five fixed dropdown timeframes — Last 7 / 30 / 90 days, This month, Last month — for every windowed tab. The work runs off the request path; nothing is computed inline on `admin_init`. So by the time an editor opens Insights, the data is usually already cached.
- **Durable, eviction-proof storage.** Pre-warmed windows are stored in non-autoloaded options (DB-backed) instead of transients, so they survive memcached eviction. Reads prefer the durable entry, then fall back to the existing transient/compute path — durable storage is written only by the warm job, so option growth is bounded to presets × tabs (orphaned windows are pruned as the rolling presets shift each day).
- **Stale-while-revalidate.** A cached view always renders instantly; if its data is older than ~25 hours, a background refresh is scheduled so the next view is current. The visible "last updated" timestamp and the manual refresh control are unchanged.

No UI changes. The Advertising tab (Google Ad Manager) keeps its own existing scheduled refresh and is not pre-warmed. Behavior is gated by the existing `NEWSPACK_INSIGHTS_ENABLED` flag, skips when `NEWSPACK_INSIGHTS_FIXTURE_MODE` is on, and fully no-ops (including BigQuery spend) when `NEWSPACK_INSIGHTS_CACHE_DISABLED` is set.

This PR also extracts the date/permission/window-validation helpers that were duplicated across all eight Insights REST controllers into a shared trait — a behavior-preserving refactor that nets out to fewer lines and gives Phase 2 a single seam for window handling.

Closes NEWS-2581.

Prerequisites: a site connected to the Insights BigQuery proxy (Newspack Manager Admin hub), with `define( 'NEWSPACK_INSIGHTS_ENABLED', true );` and `NEWSPACK_INSIGHTS_FIXTURE_MODE` off.

1. **Unit tests + lint.** From `plugins/newspack-plugin`: `n test-php --group insights` (597 passing) and `npm run lint:php` (0 errors).
2. **Warm is scheduled, not inline.** As an admin, load any wp-admin page (not necessarily Insights). Confirm a `newspack_insights_prewarm` action is enqueued in the `newspack-insights` group (Tools → Scheduled Actions, or `wp action-scheduler list --group=newspack-insights`), and that the admin page itself did not block on data computation.
3. **Durable options are written and not autoloaded.** After the warm action runs, confirm `wp option list --search='newspack_insights_warm_*'` lists entries for the windowed tabs, and that they are stored with `autoload = no` (so they don't bloat the autoload cache).
4. **Fast render across presets.** Open Insights and switch the timeframe across Last 7 / Last 30 / Last 90 / This month / Last month. Each should render from cache quickly; the "last updated" timestamp reflects the warm time, not the page-load time.
5. **Once per day.** Reload an admin page again the same day and confirm a second warm is NOT enqueued (the daily marker guards it). Clear the marker (`wp option delete newspack_insights_last_prewarm_date`) and confirm the next admin page load enqueues a fresh warm.
6. **Prune on day-roll.** After the rolling presets shift (next day, or by hand-warming with an earlier "today"), confirm yesterday's now-orphaned `newspack_insights_warm_*` windows are pruned and only the current five presets per tab remain.
7. **Stale-while-revalidate.** With a durable entry older than ~25h (e.g. edit its `computed_at`), open that tab: it renders the cached data immediately AND a `newspack_insights_warm_refresh` action is enqueued; after it runs, the entry's `computed_at` is current.
8. **Manual refresh persists on a warmed preset.** On a pre-warmed preset window (e.g. Last 30 days), note the "last updated" time, then use the tab's manual **Refresh** control. Confirm the data updates AND that reloading Insights **keeps** the refreshed data with the new "last updated" time — it must not revert to the older pre-warmed numbers. (Refresh overwrites the durable entry in place and resets the freshness clock.) On a custom or comparison-enabled window (never pre-warmed), confirm Refresh still works via the transient path and creates no durable entry (`wp option list --search='newspack_insights_warm_*'` shows nothing new for that window).
9. **Capability gate.** As a non-admin (no `manage_options`), load wp-admin and confirm no warm is enqueued.
10. **Escape hatches.** Set `define( 'NEWSPACK_INSIGHTS_CACHE_DISABLED', true );` and confirm no warm is scheduled or run (no BigQuery queries fire) and tabs still load via the compute path. Set `NEWSPACK_INSIGHTS_FIXTURE_MODE` and confirm warming is skipped.
11. **Refactor parity.** Confirm the Insights REST endpoints still validate dates, reject bad/inverted ranges and unpaired comparison params, and enforce `manage_options` exactly as before (the shared-trait extraction is behavior-preserving).

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
@@ -29,13 +29,14 @@ trait Cached_Controller_Trait {
 	abstract protected function tab_slug(): string;
 
 	/**
-	 * Response-shape version mixed into the cache key. Override and bump it when
-	 * a controller's payload shape changes so cached payloads from a prior shape
-	 * don't survive a deploy. Default empty: no version component, so controllers
-	 * that don't opt in keep their existing key shape (no surprise cache-bust).
+	 * Global envelope cache-schema version mixed into the cache key for every
+	 * Insights controller. Returns {@see Cache::ENVELOPE_SCHEMA_VERSION} by
+	 * default so ALL tabs are versioned uniformly — no per-tab opt-in required.
+	 * Override only when a single tab needs an independent bump that must NOT
+	 * bust the rest of the pre-warm cycle.
 	 */
 	protected function cache_schema_version(): string {
-		return '';
+		return Cache::ENVELOPE_SCHEMA_VERSION;
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
@@ -10,6 +10,7 @@
 
 namespace Newspack\Insights;
 
+use DateTimeImmutable;
 use WP_REST_Request;
 use WP_REST_Response;
 
@@ -35,6 +36,43 @@ trait Cached_Controller_Trait {
 	 */
 	protected function cache_schema_version(): string {
 		return '';
+	}
+
+	/**
+	 * Base-window payload builder — same shape as a no-comparison GET. Each
+	 * controller implements it by delegating to its existing build_response()
+	 * with null comparison.
+	 *
+	 * @param DateTimeImmutable $start Window start (00:00:00, site tz).
+	 * @param DateTimeImmutable $end   Window end (23:59:59, site tz).
+	 * @return array
+	 */
+	abstract public function build_window_payload( DateTimeImmutable $start, DateTimeImmutable $end ): array;
+
+	/**
+	 * Build and durably store one pre-warmed base window. Uses the controller's
+	 * own tab/source/versioned key so the entry key-matches the GET read path.
+	 *
+	 * @param DateTimeImmutable $start Window start.
+	 * @param DateTimeImmutable $end   Window end.
+	 */
+	public function warm_window( DateTimeImmutable $start, DateTimeImmutable $end ): void {
+		$key_parts = $this->versioned_key_parts_from(
+			$start->format( 'Y-m-d' ),
+			$end->format( 'Y-m-d' ),
+			null,
+			null
+		);
+		Cache::store_durable(
+			$this->tab_slug(),
+			$this->cache_source(),
+			$key_parts,
+			$this->build_window_payload( $start, $end ),
+			[
+				'start' => $start->format( 'Y-m-d' ),
+				'end'   => $end->format( 'Y-m-d' ),
+			]
+		);
 	}
 
 	/**
@@ -77,18 +115,22 @@ trait Cached_Controller_Trait {
 	}
 
 	/**
-	 * Canonical window components used as the cache key.
+	 * Pure cache-key derivation from raw window strings, with the response-shape
+	 * version prepended when the controller sets one.
 	 *
-	 * @param WP_REST_Request $request Incoming request.
-	 * @return array{0:string,1:string,2:?string,3:?string}
+	 * @param string      $start Window start (Y-m-d).
+	 * @param string      $end   Window end (Y-m-d).
+	 * @param string|null $cs    Comparison start or null.
+	 * @param string|null $ce    Comparison end or null.
+	 * @return array
 	 */
-	private static function cache_key_parts( WP_REST_Request $request ): array {
-		return [
-			(string) $request->get_param( 'start' ),
-			(string) $request->get_param( 'end' ),
-			$request->get_param( 'compare_start' ) ? (string) $request->get_param( 'compare_start' ) : null,
-			$request->get_param( 'compare_end' ) ? (string) $request->get_param( 'compare_end' ) : null,
-		];
+	private function versioned_key_parts_from( string $start, string $end, ?string $cs, ?string $ce ): array {
+		$parts   = [ $start, $end, $cs, $ce ];
+		$version = $this->cache_schema_version();
+		if ( '' !== $version ) {
+			array_unshift( $parts, $version );
+		}
+		return $parts;
 	}
 
 	/**
@@ -100,12 +142,12 @@ trait Cached_Controller_Trait {
 	 * @return array
 	 */
 	private function versioned_cache_key_parts( WP_REST_Request $request ): array {
-		$parts   = self::cache_key_parts( $request );
-		$version = $this->cache_schema_version();
-		if ( '' !== $version ) {
-			array_unshift( $parts, $version );
-		}
-		return $parts;
+		return $this->versioned_key_parts_from(
+			(string) $request->get_param( 'start' ),
+			(string) $request->get_param( 'end' ),
+			$request->get_param( 'compare_start' ) ? (string) $request->get_param( 'compare_start' ) : null,
+			$request->get_param( 'compare_end' ) ? (string) $request->get_param( 'compare_end' ) : null
+		);
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
@@ -53,10 +53,16 @@ trait Cached_Controller_Trait {
 	 * Build and durably store one pre-warmed base window. Uses the controller's
 	 * own tab/source/versioned key so the entry key-matches the GET read path.
 	 *
+	 * Returns the versioned key parts the entry was stored under, so the
+	 * pre-warm caller can build the prune keep-list from the exact key written
+	 * (single source of truth — avoids re-computing key parts in a separate
+	 * code path that may diverge from cache_schema_version()).
+	 *
 	 * @param DateTimeImmutable $start Window start.
 	 * @param DateTimeImmutable $end   Window end.
+	 * @return array Versioned key parts passed to Cache::store_durable().
 	 */
-	public function warm_window( DateTimeImmutable $start, DateTimeImmutable $end ): void {
+	public function warm_window( DateTimeImmutable $start, DateTimeImmutable $end ): array {
 		$key_parts = $this->versioned_key_parts_from(
 			$start->format( 'Y-m-d' ),
 			$end->format( 'Y-m-d' ),
@@ -73,6 +79,7 @@ trait Cached_Controller_Trait {
 				'end'   => $end->format( 'Y-m-d' ),
 			]
 		);
+		return $key_parts;
 	}
 
 	/**
@@ -137,6 +144,13 @@ trait Cached_Controller_Trait {
 	 * Canonical window key parts with the response-shape version prepended when
 	 * the controller sets one. An empty version leaves the window parts
 	 * untouched, so a non-overriding controller's cache key is unchanged.
+	 *
+	 * NOTE — truthiness coupling: empty-string compare_start / compare_end are
+	 * treated as absent here (falsy → null) to match the behaviour of
+	 * Insights_REST_Trait::parse_window_args(), which treats a falsy param as
+	 * "no comparison window". Both sides must change together if either switches
+	 * to isset() / !== '' checks; otherwise cache keys and parsed comparison
+	 * windows will disagree on what "absent" means.
 	 *
 	 * @param WP_REST_Request $request Incoming request.
 	 * @return array
--- a/plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
@@ -1,0 +1,180 @@
+<?php
+/**
+ * Newspack Insights — shared REST helpers.
+ *
+ * Date parsing/validation, the window-args parser, the collection params spec,
+ * and the permission check, shared by every Insights REST controller. Extracted
+ * from the per-controller copies (NEWS-2581) so the GET path, the refresh path,
+ * and the pre-warm path agree on window handling and so Phase 2 (hub incremental
+ * cache) has a single date/window seam.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Insights;
+
+use DateTimeImmutable;
+use DateTimeZone;
+use Exception;
+use WP_Error;
+use WP_REST_Request;
+
+defined( 'ABSPATH' ) || exit;
+
+trait Insights_REST_Trait {
+
+	/**
+	 * Permission check — requires the Insights wizard capability.
+	 *
+	 * @return bool|WP_Error
+	 */
+	public function permissions_check() {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return new WP_Error(
+				'newspack_insights_rest_forbidden',
+				__( 'You do not have permission to view Insights data.', 'newspack-plugin' ),
+				[ 'status' => rest_authorization_required_code() ]
+			);
+		}
+		return true;
+	}
+
+	/**
+	 * REST validate_callback for a Y-m-d date string.
+	 *
+	 * @param mixed $value Value.
+	 * @return bool|WP_Error
+	 */
+	public function validate_date_string( $value ) {
+		if ( ! is_string( $value ) || '' === $value ) {
+			return new WP_Error(
+				'newspack_insights_invalid_date',
+				__( 'Date must be a non-empty YYYY-MM-DD string.', 'newspack-plugin' ),
+				[ 'status' => 400 ]
+			);
+		}
+		$parsed = DateTimeImmutable::createFromFormat( 'Y-m-d', $value, $this->site_timezone() );
+		if ( ! $parsed || $parsed->format( 'Y-m-d' ) !== $value ) {
+			return new WP_Error(
+				'newspack_insights_invalid_date',
+				/* translators: %s: the invalid date string */
+				sprintf( __( 'Invalid date "%s". Expected YYYY-MM-DD.', 'newspack-plugin' ), $value ),
+				[ 'status' => 400 ]
+			);
+		}
+		return true;
+	}
+
+	/**
+	 * Collection params: start, end (required) + compare_start, compare_end.
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$base = [
+			'type'              => 'string',
+			'sanitize_callback' => 'sanitize_text_field',
+			'validate_callback' => [ $this, 'validate_date_string' ],
+		];
+		return [
+			'start'         => array_merge(
+				$base,
+				[
+					'description' => __( 'Inclusive window start date (YYYY-MM-DD, site timezone).', 'newspack-plugin' ),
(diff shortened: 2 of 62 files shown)


plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -0,0 +1,253 @@
...
+		foreach ( self::$tabs as $tab => $warmer ) {
+			$keep = [];
+			foreach ( $windows as $w ) {
+				try {
+					$warmer( $w['start'], $w['end'] );
+				} catch ( \Throwable $e ) {
+					self::log_failure( $tab, $w['preset'], $e );
+					continue;
+				}
+				$keep[] = [ $w['start']->format( 'Y-m-d' ), $w['end']->format( 'Y-m-d' ), null, null ];
```
dkoo: **Blocker:** The prune keep-list is built from **unversioned** window parts — `[ $w['start'], $w['end'], null, null ]` — but `warm_window()` (trait-cached-controller.php:60) stores each durable entry under `versioned_key_parts_from(...)`, which prepends `cache_schema_version()` when non-empty. Three controllers return a non-empty version: Gates (`Gates_Metric::CACHE_PREFIX`), Conversion (`Conversion_Metric::CACHE_PREFIX`) and Prompts (`Prompts_Metric::CACHE_PREFIX`). So `prune_durable()` hashes the unversioned key, which never matches the stored versioned hash, deletes every entry this run just warmed for those three tabs, and rewrites the index to phantom hashes the GET path (versioned key) will never read. Net: zero durable-cache benefit for Gates/Conversion/Prompts, and the SWR refresh — which only fires on a durable hit — becomes unreachable.

**Risk if not addressed:** because those three tabs' `build_window_payload()` fans out ~13/23/24 BigQuery-proxy calls per window with no per-request memo, the discarded daily warm plus the now-always-uncached page views make net BigQuery spend and hub load *higher* than with no caching at all, and the index keeps accumulating phantom hashes that no future prune can reclaim.

**Consider:** derive the prune keep-list from the same versioned key parts `warm_window()` stores under (e.g. expose the controller's `versioned_key_parts_from(start, end, null, null)`), rather than reconstructing raw window parts in the Prewarm loop — ideally from one shared helper so the warm-write key and the prune-keep key can't diverge again. And add a unit test that drives the real controller warmer (not a stub storing unversioned parts) for a versioned tab, asserting the GET read path returns the warmed entry after a run — the current test passes only because it bypasses `cache_schema_version()`.

plugins/newspack-plugin/includes/wizards/insights/class-insights-preset-windows.php
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * Compute every fixed preset window, anchored to $today, in $today's
+	 * timezone. Each "Last N days" preset is an inclusive N-day window ending
+	 * today (subtract N-1). Start is 00:00:00, end is 23:59:59, matching the
+	 * REST controllers' parse_date() boundaries.
+	 *
+	 * @param DateTimeImmutable $today Site-timezone "now" (e.g. current_datetime()).
+	 * @return array<int, array{preset:string, start:DateTimeImmutable, end:DateTimeImmutable}>
+	 */
+	public static function all( DateTimeImmutable $today ): array {
```
dkoo, line 34: **Suggestion:** The pre-warm key-match hinges on the browser-computed "today" (`useDateRange.ts`, browser-local time) equaling the server-computed "today" (`current_datetime()`, site timezone). When an admin's browser timezone differs from the site timezone and they straddle a date boundary, the dropdown requests a window whose Y-m-d strings differ by a day from the warmed key — every preset misses and falls through to compute, silently defeating Phase 1 for that user. It degrades gracefully (no error) and the browser-local computation predates this PR, but since the whole feature hinges on this parity it's worth anchoring the JS preset computation to the site timezone or documenting the limitation.

dkoo (commented):
Self-review of my own Phase 1 caching PR, posted as a comment. Overall this is solid: durable storage hygiene (non-autoloaded writes, prefixed naming, bounded prune), capability gating, off-request scheduling and Action Scheduler de-dup are all sound, and the trait extraction is behavior-preserving across the eight controllers.

One blocker should be fixed before this merges: the daily prune builds its keep-list from unversioned window key parts, but the three BigQuery-backed tabs (Gates, Conversion, Prompts) store durable entries under a versioned key, so the prune deletes every entry it just warmed for those tabs in the same run. The result is no caching benefit for the most expensive tabs, wasted daily BigQuery spend on warm work that's immediately discarded, an unreachable stale-while-revalidate path, and a durable index that accumulates phantom hashes. The existing prewarm test hides this because its stub warmer stores unversioned parts — so the fix (derive the prune keep-list from the same versioned key parts warm_window() stores under) should land together with a test that drives the real controller warmer.

Two suggestions worth a look: the pre-warm key-match depends on the browser's local "today" equaling the site-timezone "today" (a mismatch across a date boundary silently defeats the cache for that user); and a durable entry stored without a window is served stale forever because the refresh path early-returns on the empty window. The remaining items are nits.

Also still open from the automated pass: the trait's permission check is still hardcoded to manage_options (reuse get_required_capability()), some trailing whitespace, and the refresh handler's date parse could round-trip-validate like the trait already does.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
```diff
@@ -140,6 +173,150 @@ public static function peek( string $tab, string $source, array $key_parts ): ?a
...
+			return null;
+		}
+		$stored = get_option( self::durable_option( $tab, $key_parts ), null );
+		if ( ! is_array( $stored ) || ! isset( $stored['payload'], $stored['computed_at'], $stored['source'] ) ) {
+			return null;
+		}
+		if ( $stored['source'] !== $source ) {
+			return null;
+		}
+		if ( ! isset( $stored['window'] ) || ! is_array( $stored['window'] ) ) {
```
dkoo: **Suggestion:** `peek_durable()` synthesizes an empty window (`['start' => '', 'end' => '']`) for any durable entry lacking one. In `store()` that empty window is passed to the `newspack_insights_durable_stale` action, and `on_durable_stale()` early-returns on empty start/end — so once such an entry goes stale it's served forever and never SWR-refreshed. Every entry `store_durable()` writes carries a window, so this only bites a malformed/legacy row, but the failure mode is a permanently-stale-but-served entry with no self-healing. Consider treating a windowless durable entry as unservable (return null) so it falls through to recompute.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
```diff
@@ -140,6 +173,150 @@ public static function peek( string $tab, string $source, array $key_parts ): ?a
...
+		$hashes = get_option( $option, [] );
+		if ( ! is_array( $hashes ) ) {
+			$hashes = [];
+		}
+		foreach ( $hashes as $hash ) {
+			if ( ! in_array( $hash, $keep_hashes, true ) ) {
+				delete_option( 'newspack_insights_warm_' . $tab . '_' . $hash );
+			}
+		}
+		update_option( $option, array_values( $keep_hashes ), false );
```
dkoo: **Nit:** `prune_durable()` rewrites the index with `array_values( $keep_hashes )` — the hashes the caller *asked* to keep — rather than the hashes that actually survived from the prior index. If `$keep_key_parts` ever lists a window that was never stored, the index references a non-existent option. Latent today (run_prewarm only adds to `$keep` after a successful warm), but intersecting `$keep_hashes` with the existing index would make it self-consistent regardless of caller behavior — and would have contained the blocker above to a no-op instead of a phantom-index leak.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -0,0 +1,253 @@
...
+	 * A partial success (some tabs/windows succeeded) does stamp the marker —
+	 * real work was done.
+	 */
+	public static function run_prewarm(): void {
+		if ( ! self::is_warmable() ) {
+			return;
+		}
+		$windows     = Preset_Windows::all( current_datetime() );
+		$any_success = false;
+		foreach ( self::$tabs as $tab => $warmer ) {
```
dkoo, line 153: **Nit:** `run_prewarm()` relies on every section's `register_hooks()` having populated `self::$tabs` in the async-runner request. It holds today, but if the registry is ever empty when the job fires, the run silently no-ops without warming or stamping the marker, and the next `admin_init` re-enqueues — a quiet re-schedule loop rather than a visible failure. A debug log when `self::$tabs` is empty would make that observable.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
```diff
@@ -60,7 +62,38 @@ public static function store( string $tab, string $source, array $key_parts, cal
 		}
 
 		$cooldown_until = self::SOURCE_BIGQUERY === $source ? self::bq_cooldown_until( $tab ) : null;
-		$key            = self::transient_key( $tab, $key_parts );
+
+		// Durable warm store takes precedence over the transient. Pre-warmed
+		// preset windows live here (memcached-eviction-proof). A stale entry is
+		// still served instantly; we fire an action so the pre-warm layer can
+		// schedule an async background refresh (stale-while-revalidate).
+		$durable = self::peek_durable( $tab, $source, $key_parts );
```
dkoo, line 79: **Nit:** On the comparison-window GET path (`compare_start`/`compare_end` present), `store()` still issues a `peek_durable()` `get_option()` that's guaranteed to miss — durable storage only ever holds base (no-comparison) windows. One wasted uncached option read per comparison request (not autoloaded, so negligible); could be skipped when the key parts indicate a comparison window.

plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
```diff
@@ -100,12 +142,12 @@ private static function cache_key_parts( WP_REST_Request $request ): array {
...
-		$parts   = self::cache_key_parts( $request );
-		$version = $this->cache_schema_version();
-		if ( '' !== $version ) {
-			array_unshift( $parts, $version );
-		}
-		return $parts;
+		return $this->versioned_key_parts_from(
+			(string) $request->get_param( 'start' ),
+			(string) $request->get_param( 'end' ),
+			$request->get_param( 'compare_start' ) ? (string) $request->get_param( 'compare_start' ) : null,
```
dkoo, line 163: **Nit:** `versioned_cache_key_parts()` treats an empty-string `compare_start`/`compare_end` as absent via truthiness, and `parse_window_args()` (trait-insights-rest.php:133) uses the same truthiness to decide comparison mode — so they agree today. Just flagging the implicit coupling: if either side later switches to `isset()` / `!== ''`, cache keys and parsed windows could diverge. Not a bug now.

plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
```diff
@@ -0,0 +1,180 @@
...
+	public function permissions_check() {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return new WP_Error(
+				'newspack_insights_rest_forbidden',
+				__( 'You do not have permission to view Insights data.', 'newspack-plugin' ),
+				[ 'status' => rest_authorization_required_code() ]
+			);
+		}
+		return true;
+	}
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9. permissions_check() now calls Insights_Wizard::get_required_capability() instead of hardcoding manage_options, so the REST routes and the pre-warm scheduler share one capability source.

plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
```diff
@@ -0,0 +1,180 @@
...
+			'validate_callback' => [ $this, 'validate_date_string' ],
+		];
+		return [
+			'start'         => array_merge(
+				$base,
+				[
+					'description' => __( 'Inclusive window start date (YYYY-MM-DD, site timezone).', 'newspack-plugin' ),
+					'required'    => true,
+				] 
+			),
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9 (trailing whitespace removed); thread auto-resolved on push.

plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
```diff
@@ -0,0 +1,180 @@
...
+			'end'           => array_merge(
+				$base,
+				[
+					'description' => __( 'Inclusive window end date (YYYY-MM-DD, site timezone).', 'newspack-plugin' ),
+					'required'    => true,
+				] 
+			),
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9 (trailing whitespace removed); thread auto-resolved on push.

plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
```diff
@@ -0,0 +1,180 @@
...
+			'compare_start' => array_merge(
+				$base,
+				[
+					'description' => __( 'Optional comparison window start. Must pair with compare_end.', 'newspack-plugin' ),
+					'required'    => false,
+				] 
+			),
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9 (trailing whitespace removed); thread auto-resolved on push.

plugins/newspack-plugin/includes/wizards/insights/api/trait-insights-rest.php
```diff
@@ -0,0 +1,180 @@
...
+			'compare_end'   => array_merge(
+				$base,
+				[
+					'description' => __( 'Optional comparison window end. Must pair with compare_start.', 'newspack-plugin' ),
+					'required'    => false,
+				] 
+			),
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9 (trailing whitespace removed); thread auto-resolved on push.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -0,0 +1,253 @@
...
+		$tab = $args['tab'] ?? '';
+		if ( ! isset( self::$tabs[ $tab ] ) ) {
+			return;
+		}
+		$tz    = wp_timezone();
+		$start = DateTimeImmutable::createFromFormat( 'Y-m-d', (string) ( $args['start'] ?? '' ), $tz );
+		$end   = DateTimeImmutable::createFromFormat( 'Y-m-d', (string) ( $args['end'] ?? '' ), $tz );
+		if ( ! $start || ! $end ) {
+			return;
+		}
```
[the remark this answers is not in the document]
dkoo: Fixed in 5d19092a9. run_warm_refresh() now rejects Action Scheduler date args that do not round-trip: it verifies start->format('Y-m-d') equals the original arg for both start and end before warming, so a rolled-over date like 2026-02-30 is discarded.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Fixed in 5d19092a9. warm_window() now returns the versioned key parts it stored under, and run_prewarm() builds the prune keep-list from that returned value (single source of truth, no unversioned reconstruction), so the keep hash always matches the stored hash. prune_durable() also intersects the existing index with the keep set. Added a regression test that drives the real Gates_REST_Controller (non-empty cache_schema_version) and asserts the durable entry survives the prune under the versioned key -- it fails against the pre-fix code. Separately, c65c2ad88 consolidates envelope versioning to one global Cache::ENVELOPE_SCHEMA_VERSION so every tab is versioned uniformly.

plugins/newspack-plugin/includes/wizards/insights/class-insights-preset-windows.php
dkoo: Fixed in e09aeb3c3. computeRangeForPreset() now anchors today to the site timezone via a getSiteToday() helper (window.newspackInsights.timezone through Intl.DateTimeFormat en-CA), so the dropdown windows match the server-side Preset_Windows the warm keys are built from. Falls back to browser-local when no timezone is configured.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
dkoo: Fixed in 5d19092a9. peek_durable() now returns null when the stored window is missing or empty, so an unserviceable entry falls through to a live recompute instead of being served forever-stale with no SWR path.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
dkoo: Fixed in 5d19092a9. prune_durable() now rewrites the index as array_intersect(existing_hashes, keep_hashes), so a key that was never stored cannot survive a prune cycle.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Fixed in 5d19092a9. run_prewarm() now logs a warning via Logger::newspack_log when the tab registry is empty before returning, so a missing-registry no-op is observable instead of silent.

plugins/newspack-plugin/includes/wizards/insights/class-cache.php
dkoo: Deferring. This is a single non-autoloaded option read on comparison-mode requests only (durable storage only ever holds base windows), which the note itself calls negligible. Short-circuiting on a comparison key adds branching to the hot read path for no measurable gain; happy to revisit if profiling ever surfaces it.

plugins/newspack-plugin/includes/wizards/insights/api/trait-cached-controller.php
dkoo: Addressed in 5d19092a9 with a documenting comment. versioned_cache_key_parts() now notes that empty-string compare params are treated as absent via truthiness to match parse_window_args(), and that both sides must change together if either moves to isset() / !== '' checks.

PR #475, opened 2026-07-01
fix(salesforce): verify WooCommerce webhook signature

jason10lee:
### All Submissions:

The Salesforce sync webhook endpoint now verifies the WooCommerce HMAC
signature (`X-WC-Webhook-Signature`) over the raw request body, delegating to
WooCommerce's own `WC_Webhook::generate_signature()`.

For that verification to be meaningful the webhook needs a signing secret, which
WooCommerce does not set for programmatically created webhooks. So the sync
webhook is now given a strong secret on creation, and existing webhooks that
lack one are backfilled (on `init`, via `platform_check`). Legitimate
WooCommerce deliveries (which always include the signature) are unaffected.

Part of NPPM-2967.

1. Run the tests: `n test-php --filter Newspack_Test_Salesforce` (6 pass).
2. A request with a valid webhook id but a missing or incorrect signature is rejected; a correctly signed delivery is accepted.
3. The sync webhook is created with a signing secret, and an existing webhook without one is backfilled on the next request.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/class-salesforce.php
+++ b/plugins/newspack-plugin/includes/class-salesforce.php
@@ -202,17 +202,31 @@ class Salesforce {
 	 * @return bool|WP_Error
 	 */
 	public static function api_validate_webhook( $request ) {
-		$is_valid   = false;
-		$webhook_id = $request->get_header( 'X-WC-Webhook-ID' );
-
-		if ( $webhook_id ) {
-			$is_valid = self::get_webhook() === (int) $webhook_id;
+		$webhook_id = (int) $request->get_header( 'X-WC-Webhook-ID' );
+		if ( ! $webhook_id || self::get_webhook() !== $webhook_id ) {
+			return self::webhook_forbidden_error();
 		}
 
-		if ( $is_valid ) {
-			return true;
+		// Verify the WooCommerce webhook signature over the raw request body, so only
+		// WooCommerce (which holds the shared secret) can invoke the sync handler.
+		$webhook   = function_exists( 'wc_get_webhook' ) ? \wc_get_webhook( $webhook_id ) : null;
+		$signature = $request->get_header( 'X-WC-Webhook-Signature' );
+		if ( ! $webhook || empty( $signature ) ) {
+			return self::webhook_forbidden_error();
+		}
+		if ( ! hash_equals( $webhook->generate_signature( $request->get_body() ), $signature ) ) {
+			return self::webhook_forbidden_error();
 		}
 
+		return true;
+	}
+
+	/**
+	 * The standard error returned for an invalid webhook request.
+	 *
+	 * @return \WP_Error
+	 */
+	private static function webhook_forbidden_error() {
 		return new \WP_Error(
 			'newspack_rest_forbidden',
 			esc_html__( 'Invalid webhook request.', 'newspack' ),
--- a/plugins/newspack-plugin/includes/class-salesforce.php
+++ b/plugins/newspack-plugin/includes/class-salesforce.php
@@ -30,6 +30,12 @@ class Salesforce {
 	];
 
 	/**
+	 * Autoloaded one-shot flag, set once the sync webhook is confirmed to have a
+	 * signing secret, so platform_check() stops re-loading the webhook on every init.
+	 */
+	const WEBHOOK_SECRET_BACKFILLED = 'newspack_salesforce_webhook_secret_backfilled';
+
+	/**
 	 * If this site is a duplicate (staging or dev clone) site,
 	 * don't handle webhooks which communicate with external services.
 	 *
@@ -67,7 +73,7 @@ class Salesforce {
 
 		if ( $is_newspack && ! $webhook_id ) {
 			self::create_webhook();
-		} elseif ( $is_newspack && $webhook_id ) {
+		} elseif ( $is_newspack && $webhook_id && ! get_option( self::WEBHOOK_SECRET_BACKFILLED ) ) {
 			self::ensure_webhook_secret( $webhook_id );
 		}
 	}
@@ -82,9 +88,21 @@ class Salesforce {
 	 */
 	private static function ensure_webhook_secret( $webhook_id ) {
 		$webhook = \wc_get_webhook( $webhook_id );
-		if ( $webhook && '' === $webhook->get_secret() ) {
+		if ( ! $webhook ) {
+			return;
+		}
+		if ( '' === $webhook->get_secret() ) {
+			// Deploy-window note: backfilling changes the secret, so any delivery WooCommerce
+			// already queued and signed with the old (empty) secret will 403 on arrival, and
+			// WooCommerce does not auto-retry. The window is the first init after deploy and
+			// self-heals immediately.
 			$webhook->set_secret( wp_generate_password( 50, true, true ) );
 			$webhook->save();
+		}
+		if ( '' !== $webhook->get_secret() ) {
+			// One-shot: once a secret is present, record it (autoloaded) so platform_check()
+			// stops re-loading the webhook on every init.
+			update_option( self::WEBHOOK_SECRET_BACKFILLED, true, true );
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


plugins/newspack-plugin/includes/class-salesforce.php
```diff
@@ -202,17 +202,31 @@ public static function api_permissions_check() {
...
-			return true;
+		// Verify the WooCommerce webhook signature over the raw request body, so only
+		// WooCommerce (which holds the shared secret) can invoke the sync handler.
+		$webhook   = function_exists( 'wc_get_webhook' ) ? \wc_get_webhook( $webhook_id ) : null;
+		$signature = $request->get_header( 'X-WC-Webhook-Signature' );
+		if ( ! $webhook || empty( $signature ) ) {
+			return self::webhook_forbidden_error();
+		}
+		if ( ! hash_equals( $webhook->generate_signature( $request->get_body() ), $signature ) ) {
+			return self::webhook_forbidden_error();
```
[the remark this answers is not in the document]
jason10lee: Addressed — the sync webhook now gets a strong secret on creation, and existing webhooks that lack one are backfilled on `init` via `platform_check()`, so deliveries can be verified. Thanks for catching this.

plugins/newspack-plugin/tests/mocks/wc-mocks.php
```diff
@@ -776,3 +776,41 @@ function wc_get_template( $template_name, $args = [] ) {
...
+	public function save() {
+		if ( ! $this->id ) {
+			$this->id = count( self::$registry ) + 1;
+		}
+		self::$registry[ $this->id ] = $this;
+		return $this->id;
+	}
+	public function generate_signature( $payload ) {
+		return base64_encode( hash_hmac( 'sha256', $payload, wp_specialchars_decode( $this->secret, ENT_QUOTES ), true ) );
+	}
```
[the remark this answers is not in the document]
jason10lee: Addressed — added `set_user_id()` and `delete()` to the WC_Webhook mock.

plugins/newspack-plugin/includes/class-salesforce.php
```diff
@@ -67,6 +67,24 @@ public static function platform_check() {
...
+	 * Ensure an existing webhook has a signing secret.
+	 *
+	 * Webhooks created before signature verification was added were saved without a
+	 * secret; without one, delivery signatures cannot be verified meaningfully.
+	 *
+	 * @param int $webhook_id Webhook id.
+	 */
+	private static function ensure_webhook_secret( $webhook_id ) {
+		$webhook = \wc_get_webhook( $webhook_id );
+		if ( $webhook && '' === $webhook->get_secret() ) {
```
adekbadek: **Suggestion:** Backfill re-loads the webhook on every init for the life of the site — Once a secret exists, `platform_check()` still calls `wc_get_webhook( $webhook_id )` on every `init` (every front-end and admin request) just to confirm the secret is already set — a permanent per-request object load for what is a one-time migration. It's object-cached so usually a cache hit rather than a DB round-trip, but it never does anything after the first backfill. Consider gating it behind a one-shot autoloaded flag (e.g. an option set once the backfill succeeds) so the steady state is a cheap option read rather than a webhook load.

plugins/newspack-plugin/includes/class-salesforce.php
```diff
@@ -202,17 +220,31 @@ public static function api_permissions_check() {
...
+			return self::webhook_forbidden_error();
 		}
 
-		if ( $is_valid ) {
-			return true;
+		// Verify the WooCommerce webhook signature over the raw request body, so only
+		// WooCommerce (which holds the shared secret) can invoke the sync handler.
+		$webhook   = function_exists( 'wc_get_webhook' ) ? \wc_get_webhook( $webhook_id ) : null;
+		$signature = $request->get_header( 'X-WC-Webhook-Signature' );
+		if ( ! $webhook || empty( $signature ) ) {
```
adekbadek: **Suggestion:** Validator doesn't reject an empty-secret webhook (defense-in-depth) — `api_validate_webhook()` verifies `hash_equals( $webhook->generate_signature( $body ), $signature )` but never rejects a webhook whose stored secret is empty. With an empty key the signature over the request body is reproducible by anyone who can POST that body, so the check provides no protection in that state — confirmed by direct test: an empty-secret webhook plus an attacker-computed signature is *accepted*. In practice this is unreachable via HTTP because `ensure_webhook_secret()` runs on `init` (before the REST permission callback) and backfills a real secret, and the non-WooCommerce path deletes the webhook option entirely — so there is no live bypass today. But the validator's security then rests on hook ordering rather than on the check itself. A one-line guard — reject when `'' === $webhook->get_secret()` — makes the gate self-contained and is cheap insurance for a fix going straight to `release`.

**Risk if not addressed:** if the `init` backfill ever stops running before dispatch (e.g. a future refactor moves or removes the `platform_check` `init` hook) while the webhook-id check still passes, the endpoint becomes forgeable again — an unauthenticated caller could drive the sync handler for any enumerable order id, causing Salesforce write-amplification, API-quota exhaustion, and order-note spam.

plugins/newspack-plugin/tests/unit-tests/salesforce.php
```diff
@@ -0,0 +1,148 @@
...
+		self::assertSame( '', $webhook->get_secret(), 'Precondition: the webhook has no secret.' );
+
+		Salesforce::platform_check();
+
+		self::assertNotEmpty(
+			wc_get_webhook( $webhook->get_id() )->get_secret(),
+			'An existing webhook without a secret is backfilled with one.'
+		);
+	}
+}
```
adekbadek, line 205: **Suggestion:** Test coverage gaps: empty-secret path, id guard, body-tamper — The suite proves rejection (no signature, wrong signature, wrong secret) and acceptance (correct signature), but misses the most security-relevant edge of the original vulnerability: a webhook whose stored secret is empty. A test asserting that `api_validate_webhook()` rejects an empty-secret webhook would pin down the guard suggested above. Two smaller gaps: no test for an absent or mismatched `X-WC-Webhook-ID` (the pre-existing first guard this builds on, so a regression there would be silent), and no body-tamper test (sign one body, send a different one) to prove directly that the signature is verified over the *raw body* rather than only inferring it from the wrong-secret case.

plugins/newspack-plugin/includes/class-salesforce.php
```diff
@@ -67,6 +67,24 @@ public static function platform_check() {
...
+	 *
+	 * Webhooks created before signature verification was added were saved without a
+	 * secret; without one, delivery signatures cannot be verified meaningfully.
+	 *
+	 * @param int $webhook_id Webhook id.
+	 */
+	private static function ensure_webhook_secret( $webhook_id ) {
+		$webhook = \wc_get_webhook( $webhook_id );
+		if ( $webhook && '' === $webhook->get_secret() ) {
+			$webhook->set_secret( wp_generate_password( 50, true, true ) );
```
adekbadek, line 99: **Nit:** Deploy-window: in-flight deliveries signed with the old empty secret get 403'd — A webhook that existed before this fix was signed by WooCommerce with an empty secret; the first post-deploy request backfills a real secret, so any delivery WooCommerce already queued and signed with the old empty secret will 403 on arrival, and WooCommerce does not auto-retry failed deliveries. The window is a handful of orders at the exact moment of the first `init` after deploy, and it self-heals immediately. Worth a one-line code comment (and/or a changelog note) so a publisher who sees a couple of unsynced orders at deploy time has an explanation.

plugins/newspack-plugin/includes/class-salesforce.php
```diff
@@ -202,17 +220,31 @@ public static function api_permissions_check() {
...
-		if ( $is_valid ) {
-			return true;
+		// Verify the WooCommerce webhook signature over the raw request body, so only
+		// WooCommerce (which holds the shared secret) can invoke the sync handler.
+		$webhook   = function_exists( 'wc_get_webhook' ) ? \wc_get_webhook( $webhook_id ) : null;
+		$signature = $request->get_header( 'X-WC-Webhook-Signature' );
+		if ( ! $webhook || empty( $signature ) ) {
+			return self::webhook_forbidden_error();
+		}
+		if ( ! hash_equals( $webhook->generate_signature( $request->get_body() ), $signature ) ) {
```
adekbadek, line 263: **Nit:** No logging on the rejection path — The receiver returns a generic 403 on every rejection with no log trail. For a publisher-facing "my Salesforce sync stopped working" report, a `Logger::log` line distinguishing the failure modes (id matched but signature mismatched vs. secret missing vs. no signature header) would materially cut debugging time. Optional — this class does no logging on the reject path today.

adekbadek (approved):
Approve with comments — no blockers.

plugins/newspack-plugin/includes/class-salesforce.php
jason10lee: Applied — `platform_check()` now gates the backfill behind an autoloaded one-shot flag (`WEBHOOK_SECRET_BACKFILLED`), set once a secret is confirmed present, so steady state is a cheap option read rather than a per-init webhook load.

plugins/newspack-plugin/includes/class-salesforce.php
jason10lee: Added — `api_validate_webhook()` now fails closed when the webhook secret is empty, so the gate is self-contained rather than resting on the init-before-dispatch ordering. We had this queued as NPPM-2968; since the backfill runs on init before the permission callback, adding it now is 403-safe, so NPPM-2968 is closed as done-here.

plugins/newspack-plugin/tests/unit-tests/salesforce.php
jason10lee: Added three tests: empty-secret webhook rejected (pins the guard above, even with a matching empty-key signature), mismatched `X-WC-Webhook-ID` rejected, and a body-tamper test (sign one body, send another) proving verification is over the raw body.

plugins/newspack-plugin/includes/class-salesforce.php
jason10lee: Added a code comment in `ensure_webhook_secret()` explaining the deploy-window 403s (deliveries signed with the old empty secret, no WooCommerce auto-retry, self-heals on the first init).

plugins/newspack-plugin/includes/class-salesforce.php
jason10lee: Added — `webhook_forbidden_error()` now takes a reason and writes a distinguishing `Logger::log` line (id mismatch / no signature / no secret / signature mismatch), while the public 403 stays generic.

PR #476, opened 2026-07-01
fix(newsletters): add ad status control (active/inactive) to editor

adekbadek:

The Newsletter Ad editor exposed no usable post-status control: the editor stripped the native "Summary" panel, and the ads list Quick Edit deferred status "to the full editor" — so publishers had no way to pause a running ad or bring a draft live from the editor. (See the ads-list Quick Edit companion in #478.)

This adds a single **Status: Active / Inactive** control to the editor's Ad settings panel:

- **Active** → `post_status = publish`, **Inactive** → `post_status = draft`. `publish` is what gates ad insertion (`Ads::get_newsletter_ads()` serves publish-only), so Active/Inactive maps directly to "runs" / "never runs".
- The status is only written back on an actual change, so pre-existing edge statuses (`future`/`private`/`pending`) are preserved when a publisher edits other fields without touching the control.
- The native "Summary" panel stays removed on purpose. Ads have no use for its Scheduled / Private / Pending / Password options — scheduling is driven by the start/expiration date meta, and non-`publish` statuses never serve. Exposing only Active/Inactive avoids publishers setting a status that silently prevents an ad from running.

Labels match the ads list Quick Edit control (#478) so the two surfaces read consistently.

### Regression origin

The editor stopped showing any status UI when `removeEditorPanel( 'post-status' )` was introduced in the legacy `newspack-newsletters` repo by commit [`265642f9`](https://github.com/Automattic/newspack-newsletters/commit/265642f944d0d308580347d062eba7b00935f561) ("feat: ads taxonomies and fields", #1262, part of the #1272 revenue-optimizations epic; Sep 2023). That commit created `src/ads/editor/index.js`, superseding `src/ads-admin/editor/index.js`, which did not remove the panel — and [#1262](https://github.com/Automattic/newspack-newsletters/pull/1262)'s description never mentioned the removal. It rode along silently in the editor rewrite that also introduced date-driven activation, so the panel was most likely stripped on the assumption dates would govern display, without accounting for `post_status` still gating insertion. This PR keeps the native panel off (correct, given the unwanted statuses) but adds back a purpose-built control.

1. Open a Newsletter Ad in the block editor.
2. Confirm the Document sidebar shows **no** native Status/Visibility/Publish-date panel (no Scheduled/Private/Password options anywhere).
3. In the **Ad settings** panel, confirm a **Status** control (Active / Inactive) at the top, reflecting the ad's current state (published ad → Active).
4. Switch to **Inactive**, save, and confirm the ad's `post_status` becomes `draft` (unpublished; the ads list shows Inactive) and it is no longer served.
5. Switch back to **Active**, save, and confirm it publishes (Active in the list, eligible for insertion).

### Other information:

* [x] Have you successfully run tests with your changes locally? (ESLint clean; JS suite green; verified end-to-end in an isolated env — Active↔Inactive round-trips to publish/draft and persists server-side.)

### DIFF
--- a/plugins/newspack-newsletters/src/ads/editor/index.js
+++ b/plugins/newspack-newsletters/src/ads/editor/index.js
@@ -7,7 +7,7 @@ import { useSelect, useDispatch } from '@wordpress/data';
 import { Fragment, useState } from '@wordpress/element';
 import { PluginDocumentSettingPanel, PluginPrePublishPanel, store as editPostStore } from '@wordpress/edit-post';
 import { registerPlugin } from '@wordpress/plugins';
-import { ToggleControl, TextControl, DatePicker, Notice, RangeControl, Button, Modal } from '@wordpress/components';
+import { ToggleControl, TextControl, DatePicker, Notice, RadioControl, RangeControl, Button, Modal } from '@wordpress/components';
 import { date as wpDate, format, isInTheFuture } from '@wordpress/date';
 import { SelectControl } from 'newspack-components';
 import AdPlacements from '../../components/ad-placements';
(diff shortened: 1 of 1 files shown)


jason10lee (approved):
Approved, and thanks for the fix!  Straightforward addition; new custom control renders cleanly and works as described.

Agentic review turned up no side effects or serious concerns.  It did have two suggestions that I'd consider *non-blocking*:

1.  **Editor loses "Move to trash" / publish-date again.** Because it re-hides the native panel, those affordances that the pure revert would have restored are gone. Not a gap — the DataView list still owns Trash/Restore/Delete — but it's a deliberate trade worth naming.
2. **Two-step activation + a leaky modal.** Picking "Active" only sets intent; the publisher still clicks Publish and sees the native pre-publish confirmation with "Visibility: Public / Publish: Immediately" — i.e. the exact WP vocabulary the control is trying to abstract away peeks through. Minor seam.

PR #478, opened 2026-07-01
fix(newsletters): add status control to ads quick edit

thomasguillot:
### All Submissions:

Adds a Status control (Draft / Active) to the top of the Quick Edit panel on the Newsletter Ads list. Until now an ad's status could not be changed from the list, and the full ad editor hides the status panel, so publishers had no way to pause a running ad or bring a draft live without workarounds.

The control maps to the ad's post status: Active publishes the ad, Draft unpublishes it. Because a published ad's list label is derived from its start and expiration dates, an Active ad still reads as Scheduled or Expired according to those dates, and help text in the panel explains this. Setting an ad to Draft is a manual override that immediately pulls a currently running ad out of rotation without touching its dates. The status is only sent when it actually changes, so edge statuses like private or future are preserved when the publisher edits other fields.

Screenshot 2026-07-01 at 09 46 32@2x

1. Go to Newsletters > Advertising (`edit.php?post_type=newspack_nl_ads_cpt&page=newspack-newsletters-ads-list`).
2. Create or pick a published ad that is within its date window and confirm its Status column reads Active.
3. Open Quick Edit and confirm a Status control (Draft / Active) appears at the top with Active selected.
4. Switch it to Draft and Save, then confirm the row's Status becomes Draft (the ad is now unpublished and will not be served).
5. Open Quick Edit on that draft, switch to Active and Save, then confirm the row returns to Active.
6. On an ad with a future start date, set it to Active and Save, then confirm it reads Scheduled (the dates govern the published label).
7. On an ad with a past expiration date, set it to Active, then confirm it reads Expired.
8. Confirm that changing only the Status enables the Save button, and that saving without changing Status does not alter the post status.

### Other information:

### Note on the list Status column

The DataViews Status column (`Ads_List_REST`, unchanged here) labels the `draft` kind "Inactive" to match the Quick Edit control, but still classifies `private` ads as publish-equivalent (Active/Scheduled/Expired) where the control shows them as Inactive. This is intentional and matches the editor counterpart #476, which also leaves the column as-is: `private` ads are not reachable through the ads UI and are served publish-only, so the divergence is pre-existing and low-impact. Reconciling the column classification is out of scope for this hotfix.

### DIFF
--- a/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
+++ b/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
@@ -48,7 +48,8 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 		const value = item?.meta?.price;
 		return value ? String( value ) : '';
 	} )();
-	const initialStatus = [ 'publish', 'private', 'future' ].includes( item?.status ) ? 'active' : 'draft';
+	// `private` ads are never served (the serving queries are publish-only), so treat them as Draft rather than Active.
+	const initialStatus = [ 'publish', 'future' ].includes( item?.status ) ? 'active' : 'draft';
 
 	const [ status, setStatus ] = useState( initialStatus );
 	const [ advertiserSelections, setAdvertiserSelections ] = useState( initialAdvertiserSelections );
--- a/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
+++ b/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
@@ -1,12 +1,13 @@
 /**
- * Quick Edit panel for the ads list — advertiser, placement,
+ * Quick Edit panel for the ads list — status, advertiser, placement,
  * category, start/expiry dates, price.
  *
- * Status / insertion strategy / position stay in the full editor.
+ * Insertion strategy / position stay in the full editor; status is
+ * editable here.
  */
 
 import apiFetch from '@wordpress/api-fetch';
-import { FormTokenField, TextControl } from '@wordpress/components';
+import { FormTokenField, RadioControl, TextControl } from '@wordpress/components';
 import { useEffect, useMemo, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { emailAd } from 'newspack-icons';
@@ -47,7 +48,9 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 		const value = item?.meta?.price;
 		return value ? String( value ) : '';
 	} )();
+	const initialStatus = [ 'publish', 'private', 'future' ].includes( item?.status ) ? 'active' : 'draft';
 
+	const [ status, setStatus ] = useState( initialStatus );
 	const [ advertiserSelections, setAdvertiserSelections ] = useState( initialAdvertiserSelections );
 	const [ placementSelections, setPlacementSelections ] = useState( initialPlacementSelections );
 	const [ categorySelections, setCategorySelections ] = useState( initialCategorySelections );
@@ -57,6 +60,7 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 	const [ isBusy, setIsBusy ] = useState( false );
 
 	const isDirty =
+		status !== initialStatus ||
 		startDate !== initialStartDate ||
 		expiryDate !== initialExpiryDate ||
 		price !== initialPrice ||
@@ -97,6 +101,9 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 			categories: categorySelections.map( s => s.id ),
 			meta,
 		};
+		if ( status !== initialStatus ) {
+			data.status = status === 'active' ? 'publish' : 'draft';
+		}
 		try {
 			await apiFetch( { path: `${ POSTS_PATH }/${ item.id }`, method: 'POST', data } );
 			notifySuccess( __( 'Ad updated.', 'newspack-newsletters' ) );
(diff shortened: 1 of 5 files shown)


thomasguillot:
Copilot's review on the superseded main PR #477 flagged that the panel's `status` (and every sibling field) initializes from `item` once and would not resync if `item` changed while the panel stayed mounted. Fixed in 3e4478f01 by adding `key={ item.id }` to <AdsQuickEditPanel> so it remounts and reinitializes all fields on item change.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -47,7 +48,9 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 		const value = item?.meta?.price;
 		return value ? String( value ) : '';
 	} )();
+	const initialStatus = [ 'publish', 'private', 'future' ].includes( item?.status ) ? 'active' : 'draft';
```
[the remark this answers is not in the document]
thomasguillot: Good catch. Confirmed the serving queries (`Ads::get_newsletter_ads()` and `Ads_Placements::get_ad_by_placement()`) use `get_posts()` with no `post_status`, so they default to `publish` only and never serve `private` ads. Went with option (b): `initialStatus` now treats only `publish`/`future` as Active and maps `private` to Draft, so the control reflects serving reality. `private` is still preserved on save when untouched (status is only sent when changed). Added a test for the private case. Fixed in 47c2274e0. Note the list column (`get_status_for_post`) still classifies `private` as publish-equivalent; reconciling that is a separate, broader change since `private` is not reachable through the ads UI anyway.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -47,7 +48,10 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 		const value = item?.meta?.price;
 		return value ? String( value ) : '';
 	} )();
+	// `private` ads are never served (the serving queries are publish-only), so treat them as Draft rather than Active.
+	const initialStatus = [ 'publish', 'future' ].includes( item?.status ) ? 'active' : 'draft';
```
adekbadek: `private` maps to Draft here, but the list Status column (`class-ads-list-rest.php`, unchanged in this PR) buckets `private` alongside `publish` and shows Active/Scheduled/Expired from the dates. So a private ad can read `Active` in the column yet `Draft` in Quick Edit. Pre-existing and low-impact — private ads are served publish-only, so Draft is arguably the truer label — but worth a one-line note in the description that the column label is knowingly left as-is.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -97,6 +102,9 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 			categories: categorySelections.map( s => s.id ),
 			meta,
 		};
+		if ( status !== initialStatus ) {
+			data.status = status === 'active' ? 'publish' : 'draft';
```
adekbadek, line 106: Re-activating a scheduled (`future`) ad sends `status: 'publish'`, discarding the scheduled `post_date`. This matches the documented "Draft is a manual override" intent, so no change needed — but it's uncovered: only the preserve-when-untouched case is tested. A `future → draft` assertion would pin the intended flatten behaviour.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -121,6 +129,16 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 			canSave={ canSave }
 			saveLabel={ __( 'Save', 'newspack-newsletters' ) }
 		>
+			<RadioControl
```
adekbadek, line 132: Minor: the ads CPT uses default post capabilities, so POSTing `status: 'publish'` needs `publish_posts`. An edit-but-not-publish role would see Active and only fail at save (graceful `notifyError`). Theoretical given the page's audience, but noting it.

adekbadek (approved):
Correct list-side counterpart to #476. The publish path is ESP-send-safe (the ads CPT registers no `transition_post_status` hook, so setting status can't trigger a campaign send) and unchanged statuses are preserved because `data.status` is only sent when the radio moves. The `key={ quickEditItem.id }` addition also fixes a latent stale-state bug across all Quick Edit fields, not just status. Nits inline, none blocking.

adekbadek (approved):
Re-approving on `4db37fa3`. The Draft → Inactive rename resolves the labeling nit: "Inactive" no longer claims a specific `post_status`, so it covers both `draft` and `private` and the Quick Edit control stays consistent with the list Status column. Behaviour is unchanged (save mapping still `active`→`publish`, else→`draft`) and the tests were updated to match — suite green locally. The remaining `future→draft` test and publish-cap notes stand as optional.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
thomasguillot: Added a note to the PR description as suggested. Keeping the list column classification as-is intentionally: it matches the editor counterpart #476, which also leaves `Ads_List_REST` untouched, and private ads aren't reachable through the ads UI. Both status controls (this one and #476) map private → Inactive; the column divergence is pre-existing and out of scope for the hotfix.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
thomasguillot: Added coverage in e02a6633c: a `future` ad toggled to Inactive POSTs `status: 'draft'`, pinning the intended flatten-the-schedule behavior.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
thomasguillot: Acknowledged — leaving as-is. It's theoretical for this page's audience (ad managers have publish rights), and an edit-but-not-publish role already gets a graceful `notifyError` on save rather than a silent failure. Same behavior as the editor counterpart #476.

thomasguillot:
@adekbadek heads up on one post-approval change (bfb85c59e): I relabeled the DataViews Status column's `draft` kind from "Draft" to "Inactive" so the list column speaks the same Active/Inactive language as the Quick Edit control — a draft/pending ad now reads "Inactive" in both places. Label-only: `Ads_List_REST` classification is untouched, so Active/Scheduled/Expired/Trash are unchanged, and the private-ad edge stays as documented (and consistent with #476). Quick re-glance welcome.

PR #479, opened 2026-07-01
fix(newsletters): activate ads in one step from status toggle

adekbadek:

Follow-up to #476 (ad editor Active/Inactive control) and #478 (ads list Quick Edit status), both merged.

**1. One-step activation.** Review feedback on the editor control: picking **Active** only set intent — the publisher then had to click WP's native **Publish** button, which opens the pre-publish confirmation exposing "Visibility: Public / Publish: Immediately". That's the exact WP vocabulary the Active/Inactive control is meant to abstract away, leaking through a two-step flow. This makes the toggle a true on/off switch: selecting Active/Inactive applies **and persists** the status in one step (`editPost` + `savePost`), the same immediate-commit model as the list Quick Edit. Because activation no longer routes through the native Publish button, the pre-publish panel (and its raw Visibility/Publish wording) never appears.

Disabling the pre-publish panel via `disablePublishSidebar()` was rejected: that setting lives in the global `core` preferences scope, so it would suppress pre-publish checks for posts and newsletters too.

**2. Date advisories moved into the sidebar.** Because one-step activation bypasses the native Publish flow, the date advisories that were only rendered in `PluginPrePublishPanel` (expiration in the past, expiry before start, or the display-window summary) would otherwise never be seen. They now also render as a Notice at the top of the Ad settings panel, visible during normal editing and activation.

1. Open a **draft** Newsletter Ad in the editor.
2. In **Ad settings**, click **Status → Active**. Confirm it activates immediately — no pre-publish panel, no "Visibility / Publish" wording — and persists (reload; `post_status` is `publish`, the list shows Active).
3. Click **Status → Inactive** and confirm it saves straight to `draft` the same way.
4. Set an **expiration date in the past** (or an expiry before the start date). Confirm a warning notice appears at the top of the Ad settings panel ("This ad will not be displayed…") without needing to open the native Publish flow.

### Other information:

* [x] Have you successfully run tests with your changes locally? (ESLint clean via wp-scripts lint-js; verified end-to-end in an isolated env — one-click Active/Inactive commits to publish/draft with no pre-publish panel, and the date advisory renders in the sidebar.)


plugins/newspack-newsletters/src/ads/editor/index.js
```diff
@@ -117,6 +117,15 @@ function AdEdit() {
 	// to Inactive, and only write `publish`/`draft` back on an actual change.
 	const statusControl = [ 'publish', 'future' ].includes( status ) ? 'active' : 'inactive';
 
+	// Toggle acts as an on/off switch: apply and persist the new status in one
+	// step. Saving directly avoids sending the publisher through the native
+	// Publish button, whose pre-publish panel exposes WP's "Visibility /
+	// Publish" vocabulary that this control is meant to hide.
+	const setStatus = value => {
```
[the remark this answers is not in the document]
adekbadek: Good catch — addressed in eab5720. The date advisories (expiry-in-past / expiry-before-start / display-window) now also render as a Notice at the top of the Ad settings panel, so they're visible during normal editing and one-step activation, not only in the pre-publish panel the toggle bypasses.

plugins/newspack-newsletters/src/ads/editor/index.js
```diff
@@ -117,17 +117,27 @@ function AdEdit() {
...
+		editPost( { status: value === 'active' ? 'publish' : 'draft' } );
+		savePost();
```
dkoo, line 127: **Suggestion:** The status `RadioControl` isn't disabled while this save is in flight. The sibling `disable-auto-ads.js` disables its control during save (`disabled={ isSaving || ... }`); here a publisher can click Active→Inactive→Active rapidly and dispatch overlapping saves. Core's internal `isSavingPost` guard makes data corruption unlikely, but there's no visual feedback that a save is underway. Relatedly, the returned `savePost()` promise is neither awaited nor `.catch`-handled — on a failed save the control keeps showing the intended state with no local error feedback (the editor still surfaces its own notice). Consider selecting `isSavingPost` from `core/editor` and passing `disabled` to the control to match the sibling component.

dkoo, line 127: **Suggestion:** `savePost()` persists the full edited record, so toggling status now also commits any other pending sidebar edits (price, start/expiry dates) the publisher had staged. This is generally the intended "one post, one save" behavior and matches the native button, but it's a subtle change from the prior flow where those were only committed on an explicit Publish — worth confirming it's intended.

dkoo (approved):
Clean, well-scoped change — makes the Newsletter Ad status toggle a true one-step switch. I verified it end-to-end in an isolated env: **Active** commits to `publish` and **Inactive** to `draft` in a single step with no pre-publish panel (both DB-confirmed), and a past expiration date surfaces the warning Notice at the top of the Ad settings sidebar. No blockers.

The `editPost({ status })` + `savePost()` idiom is correct: `@wordpress/data` dispatches synchronously, so `savePost()` persists the just-set status, and this matches the existing `disable-auto-ads.js` pattern. Two non-blocking notes inline.

Release risk: LOW — single editor-JS file, no PHP/contract/schema surface, verified against a running env.

plugins/newspack-newsletters/src/ads/editor/index.js
adekbadek: Good catch – done in 1d704fe. The panel now selects `isSavingPost` from `core/editor` and passes `disabled={ isSaving }` to the `RadioControl`, matching `disable-auto-ads.js`. Verified in a local env by polling every 25ms across a toggle: the control goes enabled → disabled for the whole save window → enabled, and never samples as enabled while a save is in flight, so the rapid Active→Inactive→Active click can no longer dispatch overlapping saves.

On the unhandled `savePost()` promise: I looked into adding a `.catch`, and left it off deliberately. `core/editor`'s `savePost` catches the failure internally and dispatches its own error notice rather than rejecting, so the returned promise resolves either way – a `.catch` here would be dead code, and the sibling doesn't have one for the same reason. On a failed save the behaviour is now the same as every other control in the editor: the edit stays dirty with the intended status, the editor surfaces the error notice, and the re-enabled Update button is the retry path. Happy to add local error feedback if you'd rather it be explicit here.

adekbadek: Confirmed intended. `savePost()` here is deliberately the same "one post, one save" the native Publish button performs – the whole point of the change is that the publisher never reaches for Publish, so the status toggle has to be the thing that persists the record. There is no separate "save just these fields" affordance on this CPT, and the sidebar edits (price, start/expiry) are already staged in editor state and would land on the next Publish or autosave regardless, so folding them into the toggle save is not skipping a partial-commit path – it just moves the single commit point onto the control the publisher actually uses. Good to flag it explicitly, though.

PR #486, opened 2026-07-01
fix(insights): fan out pre-warm into per-window actions

dkoo:
### All Submissions:

Fixes a production bug in the Insights daily pre-warm (Phase 1 of NEWS-2581), found while testing on a live site.

**Root cause.** `Prewarm::run_prewarm()` warmed *every* registered tab × all 5 preset windows × their many BigQuery-proxy calls in a **single** Action Scheduler action. On a real site that runs 12–16 minutes, far exceeding Action Scheduler's 300-second failure watchdog — so every run is marked "failed after 300 seconds," and when PHP's execution limit kills the process first the run dies partway, leaving the cache only partially warmed (and Action Scheduler then retries, producing overlapping long-running jobs).

Evidence from the live site's Action Scheduler logs (one representative action):

```
02:37:50  action started via Async Request
02:43:48  action marked as failed after 300 seconds. Unknown error occurred...
02:53:58  action complete via Async Request      <-- actually finished ~16 min after starting
```

**Fix — fan out into one small action per (tab, window):**

- `maybe_schedule()` (on `admin_init`, once/day) enqueues a lightweight **dispatcher** action and stamps the daily marker immediately.
- The dispatcher (`run_prewarm`) does **no** warming/BQ — it just enqueues one `WARM_WINDOW_ACTION` per tab × preset window, so it cannot time out.
- `run_warm_window()` warms a **single** window (one `build_window_payload`, worst case ~tens of seconds) and then prunes that tab to its current preset keys.
- The stale-while-revalidate path is unified onto the same per-window action.

Marker semantics: it means "fan-out scheduled today." A window that fails is re-enqueued with a bounded backoff by the per-window handler (Action Scheduler does NOT auto-retry a single async action), and once retries are exhausted the action is marked failed and visible in the AS admin -- rather than re-running the whole warm.

No behavior change to the cached data itself; this only changes *how* the warm work is scheduled. Once deployed, the per-window prune also self-heals any partially-warmed/orphaned durable state left by the previous monolithic runs.

Part of NEWS-2581 (Phase 1 hardening).

Prereqs: a site with `NEWSPACK_INSIGHTS_ENABLED` true, `NEWSPACK_INSIGHTS_FIXTURE_MODE` off, connected to the Insights BigQuery proxy.

1. **Unit tests + lint:** from `plugins/newspack-plugin`, `n test-php --group insights` (619 passing) and `npm run lint:php` (0 errors).
2. **Dispatcher fans out, warms nothing inline:** as an admin, load any wp-admin page. Confirm a single `newspack_insights_prewarm` (dispatcher) action is enqueued, and that it in turn enqueues one `newspack_insights_warm_window` action per tab × preset (Tools → Scheduled Actions, or `wp action-scheduler list --group=newspack-insights`). The dispatcher itself completes near-instantly.
3. **Per-window actions succeed under the watchdog:** confirm each `newspack_insights_warm_window` action completes in well under 300 seconds (no "failed after 300 seconds"), and that `wp option list --search='newspack_insights_warm_*'` fills in for all warmed tabs (including gates/conversion/prompts, which previously timed out).
4. **Once-per-day:** reload an admin page the same day and confirm no second dispatcher is enqueued (daily marker). Clear the marker (`wp option delete newspack_insights_last_prewarm_date`) and confirm the next admin load re-enqueues.
5. **Prune / self-heal:** after a full warm, confirm only the current five preset windows per tab remain in `newspack_insights_warm_*` (orphaned prior-day windows pruned).
6. **Fast render:** open Insights and switch across all five presets — each renders from cache; "last updated" reflects the warm time.
7. **Escape hatches:** with `NEWSPACK_INSIGHTS_CACHE_DISABLED` set, confirm no dispatcher or window actions run (no BigQuery calls); with `NEWSPACK_INSIGHTS_FIXTURE_MODE`, warming is skipped.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
+++ b/plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
@@ -4,9 +4,10 @@
  *
  * Warms the five fixed timeframe presets for every registered windowed tab into
  * the durable cache, so opening Insights renders from cache. Scheduling is
- * triggered on admin_init (once/day, guarded) and the actual work runs in an
- * async Action Scheduler job — never inline on a request. Stale durable entries
- * trigger a background refresh (stale-while-revalidate). NEWS-2581 Phase 1.
+ * triggered on admin_init (once/day, guarded) and fans out into one small async
+ * Action Scheduler job per (tab, window) — so no single job times out.
+ * Stale durable entries trigger a background refresh (stale-while-revalidate).
+ * NEWS-2581 Phase 1.
  *
  * @package Newspack
  */
@@ -23,15 +24,15 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Prewarm {
 
-	const WARM_ACTION         = 'newspack_insights_prewarm';
-	const WARM_REFRESH_ACTION = 'newspack_insights_warm_refresh';
-	const WARM_GROUP          = 'newspack-insights';
-	const MARKER_OPTION       = 'newspack_insights_last_prewarm_date';
+	const WARM_ACTION        = 'newspack_insights_prewarm';
+	const WARM_WINDOW_ACTION = 'newspack_insights_warm_window';
+	const WARM_GROUP         = 'newspack-insights';
+	const MARKER_OPTION      = 'newspack_insights_last_prewarm_date';
 
 	/**
-	 * Registry: tab slug => warmer callable( DateTimeImmutable, DateTimeImmutable ): void.
+	 * Registry: tab slug => [ 'warmer' => callable, 'key_for' => callable ].
 	 *
-	 * @var array<string, callable>
+	 * @var array<string, array{warmer: callable, key_for: callable}>
 	 */
 	private static $tabs = [];
 
@@ -55,26 +56,26 @@ final class Prewarm {
 		self::$hooks_registered = true;
 		add_action( 'admin_init', [ __CLASS__, 'maybe_schedule' ] );
 		add_action( self::WARM_ACTION, [ __CLASS__, 'run_prewarm' ] );
-		add_action( self::WARM_REFRESH_ACTION, [ __CLASS__, 'run_warm_refresh' ] );
+		add_action( self::WARM_WINDOW_ACTION, [ __CLASS__, 'run_warm_window' ] );
 		add_action( 'newspack_insights_durable_stale', [ __CLASS__, 'on_durable_stale' ], 10, 3 );
 	}
 
 	/**
-	 * Register a tab's warmer. The warmer builds + durably stores one base
-	 * window and RETURNS the versioned key parts it stored under. Typically
-	 * [ $controller, 'warm_window' ] — the Cached_Controller_Trait implementation
-	 * returns the key array from warm_window() so the prune keep-list is derived
-	 * from the actual stored key rather than a re-derived approximation.
+	 * Register a tab's warmer and key resolver.
 	 *
-	 * @param string   $tab    Tab slug.
-	 * @param callable $warmer fn( DateTimeImmutable $start, DateTimeImmutable $end ): array.
+	 * @param string   $tab     Tab slug.
+	 * @param callable $warmer  fn( DateTimeImmutable $start, DateTimeImmutable $end ): array — warms + stores a window and returns the versioned key parts.
+	 * @param callable $key_for fn( DateTimeImmutable $start, DateTimeImmutable $end ): array — returns the versioned key parts WITHOUT warming.
 	 */
-	public static function register_tab( string $tab, callable $warmer ): void {
-		self::$tabs[ $tab ] = $warmer;
+	public static function register_tab( string $tab, callable $warmer, callable $key_for ): void {
+		self::$tabs[ $tab ] = [
+			'warmer'  => $warmer,
+			'key_for' => $key_for,
+		];
 	}
 
 	/**
-	 * Test-only: clear the in-memory registry.
+	 * Test-only: clear the in-memory registry and hooks flag.
 	 */
 	public static function reset_registry_for_tests(): void {
 		self::$tabs             = [];
@@ -102,6 +103,11 @@ final class Prewarm {
 	/**
 	 * Schedule an async warm at most once per day on admin_init. Never computes
 	 * inline. Guards on environment, capability, and the daily marker.
+	 *
+	 * Marker semantics: the marker now means "fan-out was scheduled today", not
+	 * "warming completed successfully." Individual window failures are retried
+	 * automatically by Action Scheduler on each per-window job; a re-fan-out on
+	 * the same day is never needed.
 	 */
 	public static function maybe_schedule(): void {
 		if ( ! self::is_warmable() ) {
@@ -120,18 +126,18 @@ final class Prewarm {
 			return;
 		}
 		as_enqueue_async_action( self::WARM_ACTION, [], self::WARM_GROUP );
+		// Stamp the marker immediately: "fan-out scheduled today." Retry of
+		// individual failed windows is handled by Action Scheduler per-job
+		// retries, not by re-fanning-out the same day.
+		update_option( self::MARKER_OPTION, self::today(), false );
 	}
 
 	/**
-	 * WARM_ACTION handler. Warms every registered tab × preset, prunes orphaned
-	 * windows, and records the daily marker.
+	 * WARM_ACTION handler. Fans out one WARM_WINDOW_ACTION per (tab, window).
 	 *
-	 * The daily marker is only stamped when at least one window warmed
-	 * successfully across all tabs. A total-failure run (e.g. a BQ-proxy outage
-	 * that throws on every window) leaves the marker unchanged so the next
-	 * admin_init can re-enqueue a retry, still de-duped by as_has_scheduled_action.
-	 * A partial success (some tabs/windows succeeded) does stamp the marker —
-	 * real work was done.
+	 * This dispatcher does NO warming, pruning, or BQ calls — it cannot time
+	 * out. Each per-window job is retried independently by Action Scheduler if
+	 * it fails.
 	 */
 	public static function run_prewarm(): void {
 		if ( ! self::is_warmable() ) {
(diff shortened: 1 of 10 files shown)


plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -148,33 +154,68 @@ public static function run_prewarm(): void {
...
+		$end   = DateTimeImmutable::createFromFormat( 'Y-m-d', (string) ( $args['end'] ?? '' ), $tz );
+		if ( ! $start || $start->format( 'Y-m-d' ) !== (string) ( $args['start'] ?? '' ) ) {
+			return;
+		}
+		if ( ! $end || $end->format( 'Y-m-d' ) !== (string) ( $args['end'] ?? '' ) ) {
+			return;
+		}
+		try {
+			( self::$tabs[ $tab ]['warmer'] )( $start->setTime( 0, 0, 0 ), $end->setTime( 23, 59, 59 ) );
+		} catch ( \Throwable $e ) {
```
dkoo, line 228: The marker-semantics rationale (lines 108-110, 129-132) rests on "individual window failures are retried automatically by Action Scheduler," but the code doesn't deliver that. `run_warm_window` wraps the warmer in `try { … } catch ( \Throwable $e ) { self::log_failure(…); }` and then returns normally, so Action Scheduler calls `mark_complete()` — the action is recorded **complete, not failed**, no `action_scheduler_failed_execution` fires, and it's invisible in AS's failed-actions admin (the Newspack log still records it via `log_failure`). Moreover, a single `as_enqueue_async_action` is **not** auto-retried even when it fails: `ActionScheduler_Abstract_QueueRunner::process_action()` only reschedules when `$action->get_schedule()->is_recurring()`. So with the daily marker stamped eagerly in `maybe_schedule` (line 132, before any warm runs), a BigQuery-proxy outage that coincides with the nightly warm leaves those windows cold for the whole day with no same-day recovery — and stale-while-revalidate can't backfill them because `on_durable_stale` only fires on an *existing* durable entry (a never-warmed window has none). This is the partial-cache-with-no-recovery mode the hotfix set out to remove, narrowed to the exception path.

**Risk if not addressed:** if the hub BigQuery proxy has a multi-hour outage overlapping the nightly warm, every windowed tab silently degrades to per-view live BigQuery computes cached only in evictable memcached for the rest of the day, while Action Scheduler's failed-actions list shows all-green — so the outage is easy to miss in the place ops would look for it.

**Consider:** either drop the catch (or re-throw after logging) so the action is marked `failed` and is visible/monitorable, or add an explicit bounded re-enqueue on failure (`as_schedule_single_action( time() + backoff, … )` with an attempt cap in the args). Whichever you pick, update the marker-semantics comment and the PR description so the stated recovery path matches the code — single async actions are not auto-retried.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -148,33 +154,68 @@ public static function run_prewarm(): void {
...
+			self::log_failure( $tab, $args['start'] . '..' . $args['end'], $e );
+		}
+		// Prune orphaned durable entries for this tab regardless of warm result.
+		// The five current-preset keys are always included so no valid entry is removed.
+		$keep = array_map(
```
dkoo, line 255: The warmer is called with the `start`/`end` from the job args (computed at dispatcher time), but the prune keep-list is recomputed from `Preset_Windows::all( current_datetime() )` at *job-execution* time. If the dispatcher enqueues just before local midnight and the window job runs just after, the freshly-warmed entry is keyed to yesterday's window while the keep-list is today's five — so `prune_durable` deletes the entry this job just wrote. The same shape happens when `on_durable_stale` re-warms yesterday's shifted preset (e.g. yesterday's `last-30`): full window recompute, then immediately pruned away. Self-heals on the next day's fan-out, but it's wasted BigQuery work and a one-day-stale gap. Consider anchoring the keep-list to the same date the job warmed (derive it from the job's `start`/`end`), or skipping the prune-delete of the key just written.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -148,33 +154,68 @@ public static function run_prewarm(): void {
...
+	 * the keep-list, so a not-yet-warmed window is never wrongly deleted.
+	 *
+	 * @param array $args [ 'tab' => string, 'start' => 'Y-m-d', 'end' => 'Y-m-d' ].
+	 */
+	public static function run_warm_window( array $args ): void {
+		if ( ! self::is_warmable() ) {
+			return;
 		}
+		$tab = $args['tab'] ?? '';
+		if ( ! isset( self::$tabs[ $tab ] ) ) {
```
dkoo, line 213: `run_warm_window` returns early on an unregistered `$tab` (line 193) or non-round-tripping dates (lines 199-204) with no log. A malformed/unknown-tab arg in production would indicate a registration-timing or serialization bug, but it vanishes silently — only the warm-failure path logs. A `Logger::newspack_log` at warning level on these branches would make a fan-out/registry desync diagnosable from logs rather than needing a repro.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -120,18 +126,18 @@ public static function maybe_schedule(): void {
 			return;
 		}
 		as_enqueue_async_action( self::WARM_ACTION, [], self::WARM_GROUP );
+		// Stamp the marker immediately: "fan-out scheduled today." Retry of
+		// individual failed windows is handled by Action Scheduler per-job
+		// retries, not by re-fanning-out the same day.
+		update_option( self::MARKER_OPTION, self::today(), false );
```
dkoo, line 147: The marker is stamped eagerly here (right after enqueuing the dispatcher), which is deliberate and the right call to prevent re-fan-out storms. One residual edge distinct from the per-window failure case above: if `as_enqueue_async_action` on line 128 succeeds but the *dispatcher* action is later canceled/purged before it runs (AS admin action, failed deploy, queue flush), the marker still blocks any re-schedule until tomorrow, leaving that day unwarmed. Acceptable given a first read will still compute live; noting so the trade-off is explicit.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
```diff
@@ -55,26 +56,26 @@ public static function init(): void {
...
 	 *
-	 * @param string   $tab    Tab slug.
-	 * @param callable $warmer fn( DateTimeImmutable $start, DateTimeImmutable $end ): array.
+	 * @param string   $tab     Tab slug.
+	 * @param callable $warmer  fn( DateTimeImmutable $start, DateTimeImmutable $end ): array — warms + stores a window and returns the versioned key parts.
+	 * @param callable $key_for fn( DateTimeImmutable $start, DateTimeImmutable $end ): array — returns the versioned key parts WITHOUT warming.
 	 */
-	public static function register_tab( string $tab, callable $warmer ): void {
-		self::$tabs[ $tab ] = $warmer;
+	public static function register_tab( string $tab, callable $warmer, callable $key_for ): void {
```
dkoo, line 83: `register_tab` now requires the third `$key_for` param with no default. Making it mandatory is the right call — a missing resolver would silently break pruning, and all in-tree callers pass it. Just flagging that any out-of-tree or future 2-arg caller now fatals rather than degrading; fine for an internal API.

dkoo (commented):
Self-review of my own Phase-1 hardening PR, posted as a comment. The fan-out is the right fix and is well-executed: the dispatcher does zero BigQuery work so it can't time out, each per-window job completes in seconds, the prune keep-list is built from all five current presets (no self-wipe of the other four), per-window dedup via `as_has_scheduled_action` is correct on both the fan-out and SWR paths, dates are round-trip-validated before use, and the insights suite passes at 619.

One thing to reconcile before merge — the failure-recovery contract. The PR states individual window failures are "retried automatically by Action Scheduler," but that isn't what happens: `run_warm_window` catches and swallows the warmer's exception, so Action Scheduler marks the job **complete**, not failed (no `failed` status, invisible in AS's failed-actions view — though `log_failure` does still record it to the Newspack log). And even if the exception were rethrown, Action Scheduler does not auto-retry a single `as_enqueue_async_action` — only recurring actions get rescheduled (verified against the vendored AS `process_action`). Since the daily marker is stamped before any warm runs, a BigQuery-proxy outage during the nightly warm leaves those windows cold for the day with no same-day recovery, and stale-while-revalidate can't heal a never-warmed window (it only fires on an existing durable entry). This isn't a user-facing correctness bug — a cold window still returns live data — but the code and its stated contract disagree. Options: add an explicit bounded re-enqueue on failure (`as_schedule_single_action` with a short backoff + attempt cap), or drop the try/catch so failures surface as `failed`, and correct the marker rationale to describe the real recovery path (next-day fan-out / manual refresh).

The rest are low-severity: a warm/prune anchor mismatch that can self-prune a just-warmed non-current window near midnight, a silent no-op on unknown-tab/bad-date args worth a diagnostic log, and a couple of nits.

---

**Nit (not inline-anchorable — `on_durable_stale` is unchanged here):** `TTL_DURABLE_FRESH` is 25h against a ~24h schedule, so a warmed entry goes stale ~1h before the next day's fan-out. Any admin page-load in that hour triggers an SWR `newspack_insights_warm_window` per stale window per tab — extra BigQuery load that the scheduled fan-out is about to redo anyway. Not a bug (SWR is doing its job), just noting the overlap; a slightly longer freshness bound or coordinating with the schedule would avoid the redundant pre-fan-out refreshes.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Fixed in e4a0c9fcd. Correct on both points: a caught-and-returned failure was recorded complete, and a single async action is not auto-retried by Action Scheduler. run_warm_window now re-enqueues a failed window with a bounded linear backoff (WARM_MAX_ATTEMPTS = 3) via as_schedule_single_action, and re-throws once exhausted so the action is marked failed and visible in the AS admin. Corrected the marker-semantics comments and the PR description to match. Added tests for the re-enqueue and the exhaustion-rethrow paths.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Fixed in e4a0c9fcd. run_warm_window now captures the key warm_window() returns and adds it to the prune keep-list, so a job never deletes the entry it just wrote -- covering both the midnight-roll race and the SWR re-warm of a shifted preset. Added test_run_warm_window_keeps_just_warmed_key_off_preset.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Fixed in e4a0c9fcd. Added log_skip() (Logger::newspack_log, warning) on both early returns -- unregistered tab and non-round-tripping dates -- so a fan-out/registry desync is diagnosable from logs. Added a test asserting the log fires on an unregistered tab.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Acknowledged, leaving as an accepted trade-off. A dispatcher canceled/purged before it runs will skip that day's warm (a first view still computes live). Keeping the eager marker; noted explicitly.

plugins/newspack-plugin/includes/wizards/insights/class-insights-prewarm.php
dkoo: Acknowledged. Making key_for mandatory is intentional: all in-tree callers pass it, and a missing resolver should fatal loudly rather than silently break pruning. No change.

PR #488, opened 2026-07-01
fix(group-subscription): validate invite before creating an account

jason10lee:
### All Submissions:

The group subscription invite handler now validates the invitation before
creating a reader account for a new invitee, so account creation (and the email
verification and sign-in that follow) only proceeds for a valid invitation. The
validation mirrors the checks already used when accepting an invite.

Part of NPPM-2966.

1. Run the tests: `n test-php --filter Test_Group_Subscription_Invite` (2 pass).
2. An invite request with an invalid key does not create an account; a valid invitation still creates one.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
+++ b/plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
@@ -13,6 +13,8 @@ use Newspack\Group_Subscription_Invite;
  */
 class Test_Group_Subscription_Invite extends WP_UnitTestCase {
 
+	const REDIRECTED = 'group-invite-test-redirected'; // phpcs:ignore Squiz.Commenting.VariableComment.Missing
+
 	public static function set_up_before_class() { // phpcs:ignore Squiz.Commenting.FunctionComment.Missing
 		parent::set_up_before_class();
 		require_once dirname( __DIR__, 4 ) . '/mocks/wc-mocks.php';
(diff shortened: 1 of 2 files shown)


plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -0,0 +1,106 @@
...
+		add_filter( 'wp_redirect', $redirect, 1 );
+		try {
+			Group_Subscription_Invite::process_invite_request();
+		} catch ( \Exception $e ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedCatch
+			// The handler ends in a redirect + exit; the filter above turns that into
+			// a catchable exception so the test can continue.
+			unset( $e );
+		}
+		remove_filter( 'wp_redirect', $redirect, 1 );
+	}
```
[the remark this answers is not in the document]
jason10lee: Addressed — the helper now throws a typed marker only for the redirect, re-throws anything else, fails if no redirect occurs, and removes the filter in a `finally`.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -0,0 +1,116 @@
...
+	 * An invalid invite key must not create a reader account for a new email.
+	 */
+	public function test_invalid_key_does_not_create_account() {
+		$email = 'nppm2966-invitee@example.test';
+		self::assertFalse( get_user_by( 'email', $email ), 'Precondition: no account for the email.' );
+
+		$_GET['action']       = Group_Subscription_Invite::QUERY_ARG;
+		$_GET['key']          = 'this-key-is-not-valid';
+		$_GET['email']        = $email;
+		$_GET['subscription'] = '999';
```
dkoo, line 71: **Nit:** Optional (non-blocking): the invalid-key test uses a non-existent subscription id (`999`), so it exits `is_valid_invite()` at the very first check (`sanitize_subscription` / `has_status`) rather than exercising the key-mismatch, email-mismatch, or expired-but-active branches. Since the fix's whole point is per-condition invite validation, a case with a real active subscription but a wrong/expired key (or a mismatched email) would more precisely pin the gate's behavior. Not required for this hotfix — the valid-key test confirms the conjunction passes when all four sub-checks hold.

dkoo (approved):
Approve. This closes the unauthenticated account-creation / forced-login hole (CWE-287) by validating the invite before any account is created, verified, or logged in. The new `is_valid_invite()` check is a strict subset of the validation `accept_invite()` already performs (active-subscription status, invite-key lookup, exact email match, not-expired), so it can't let an invalid/expired/mismatched invite through and can't reject a legitimate invitee that would previously have succeeded; `accept_invite()` still re-validates after registration as defense-in-depth. I reproduced the fix end-to-end in an isolated env: an unauthenticated request with an invalid key now redirects with `error_invite_invalid` and creates no account or session, while a valid seeded invite still creates the account and logs the invitee in. One optional, non-blocking test-coverage note below.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
jason10lee: Added `test_wrong_key_with_active_subscription_does_not_create_account` and `test_mismatched_email_with_valid_key_does_not_create_account` — both seed a real active subscription with a valid invite, then submit a wrong key / mismatched email respectively, exercising the invite-key lookup and email-match branches (past the subscription check) and asserting no account is created. Thanks for the end-to-end repro!

PR #497, opened 2026-07-02
feat(components): rework tabbed navigation on @wordpress/ui tabs

thomasguillot:
### All Submissions:

Stacked on #472. Replaces the bespoke `TabbedNavigation` component with an implementation built on `@wordpress/ui` Tabs, so wizard/admin tabs match the WordPress design system used by the shared page header.

- The shared `TabbedNavigation` keeps its public API: `Wizard`, `withWizardScreen`, the Setup wizard, and Story Budget consume it unchanged. Route-driven selection, `activeTabPaths` wildcards, `exact`, and `disableUpcoming` all behave as before, and tabs now share one route matcher with breadcrumbs.
- The wizards admin header (PHP-localized tabs on CPT/taxonomy screens, e.g. Newsletters ads) now renders the same shared component instead of raw markup; `forceSelected` maps to a `selected` item flag. Because these tabs are cross-page links with no in-page panel to control, they render as a `nav` of anchors with `aria-current`, styled to match the tabs widget.
- Tabs render as real anchors, so middle/cmd-click opens in a new tab. Plain clicks route through `history.push` (like the previous `NavLink`) so `history.block` guards such as unsaved-changes dialogs keep working; a raw hashchange would reach the router as an uninterceptable POP.
- wp-admin doesn't define the `--wpds-*` design tokens, so the components package ships them with the shared `Page` header styles: `page/style.scss` inlines `@wordpress/theme`'s prebuilt `design-tokens.css` (via the physical path; resolving the public specifier through sass/webpack leaks a bogus script dependency into consumers' asset manifests) and remaps the wpds brand tokens onto the Newspack primary scale on `:root`. The active-tab indicator override to `primary-600` is scoped to the tab bar, and every token usage carries a wp-vars fallback for contexts where the sheet isn't loaded.
- Routed page content renders inside the active `Tabs.Panel`, so each tab's `aria-controls` points at real panel content. Panels stay 1:1 with the tabs, which `@wordpress/ui` dev builds validate.

Part of DSGNEWS-143.

1. Go to Audience → Campaigns. The tabs (Campaigns / Segments / Settings) should render in the WP design-system style with a `primary-600` underline under the active tab. Click through them: URL hash, breadcrumb, and content should stay in sync.
2. Keyboard: focus the tab list, use Left/Right arrows to move between tabs (focus moves without navigating), then Enter to activate. Cmd/middle-click a tab and confirm it opens in a new browser tab.
3. Go to Newspack → Settings. Navigate to Additional Brands → Add New Brand (`#/additional-brands/new`): the "Additional Brands" tab should stay selected on this hidden subpage (`activeTabPaths` wildcard). Navigating away from a dirty brand form should still trigger the unsaved-changes dialog.
4. Go to Newsletters → Advertising (PHP-driven admin-header tabs). "Ads" should be selected; clicking "Advertisers" performs a full page load and selects it there.
5. Run the Setup wizard (`admin.php?page=newspack-setup-wizard`): tabs are hidden on the welcome screen; mid-flow, upcoming tabs are disabled (not clickable, no href) while previous tabs remain enabled.
6. Go to Story Budget and switch between Stories / Budgets.
7. Confirm the tab bar remains sticky below the admin bar on scroll (≥600px viewports).

### Other information:

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 		"prepare": "husky || true; pnpm run build:packages",
 		"lint": "pnpm --filter './plugins/*' --filter './themes/*' run lint:js && composer phpcs",
 		"lint:changed": "pnpm --filter '...[origin/main]' run lint:js",
-		"test:js": "pnpm --filter './plugins/*' --filter './themes/*' run test",
+		"test:js": "pnpm --filter './packages/*' --filter './plugins/*' --filter './themes/*' run --if-present test",
 		"test:php": "composer test",
 		"format": "pnpm --filter './plugins/*' --filter './themes/*' run format:js",
 		"cm": "commit"
--- a/package.json
+++ b/package.json
@@ -77,8 +77,7 @@
 			"@typescript-eslint/parser": "8.61.0",
 			"prettier": "npm:wp-prettier@^3.0.3",
 			"react": "18.3.1",
-			"react-dom": "18.3.1",
-			"@wordpress/private-apis": "1.49.0"
+			"react-dom": "18.3.1"
 		},
 		"patchedDependencies": {
 			"multi-semantic-release": "patches/multi-semantic-release.patch"
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
 			"@typescript-eslint/parser": "8.61.0",
 			"prettier": "npm:wp-prettier@^3.0.3",
 			"react": "18.3.1",
-			"react-dom": "18.3.1"
+			"react-dom": "18.3.1",
+			"@wordpress/private-apis": "1.49.0"
 		},
 		"patchedDependencies": {
 			"multi-semantic-release": "patches/multi-semantic-release.patch"
--- a/packages/components/src/divider/style.scss
+++ b/packages/components/src/divider/style.scss
@@ -46,6 +46,6 @@
 	}
 
 	&--variant-tertiary {
-		background-color: wp-colors.$gray-100;
+		background-color: var( --wpds-color-stroke-surface-neutral-weak );
 	}
 }
--- a/packages/components/src/divider/style.scss
+++ b/packages/components/src/divider/style.scss
@@ -46,6 +46,8 @@
 	}
 
 	&--variant-tertiary {
-		background-color: var( --wpds-color-stroke-surface-neutral-weak );
+		// Fallback for consumers whose bundle doesn't carry the wpds token sheet
+		// (it ships with the Page header's stylesheet).
+		background-color: var( --wpds-color-stroke-surface-neutral-weak, #{wp-colors.$gray-100} );
 	}
 }
--- a/packages/components/src/page/index.js
+++ b/packages/components/src/page/index.js
@@ -3,6 +3,7 @@
  */
 import { NavigableRegion } from '@wordpress/admin-ui';
 import { __experimentalHStack as HStack } from '@wordpress/components'; // eslint-disable-line @wordpress/no-unsafe-wp-apis
+import { cloneElement, isValidElement } from '@wordpress/element';
 import { Icon } from '@wordpress/icons';
 import { Stack, Text } from '@wordpress/ui';
 
@@ -22,20 +23,29 @@ import './style.scss';
  * Newspack page region: a sticky header block (breadcrumbs, actions, optional
  * tabbed navigation) followed by the page content.
  *
+ * This intentionally does not use `@wordpress/admin-ui`'s `Page`: that
+ * component has no slot for content rendered between the header and the body
+ * inside the sticky region, which is exactly where the tab bar must live. Until
+ * admin-ui grows such a slot, the header markup is assembled here from the same
+ * design-system primitives and tokens, so it can be swapped back later.
+ *
  * @param {Object} props
  * @param {Array}  props.breadcrumbItems    Trail items: `{ label, url? }`.
  * @param {*}      [props.badges]
  * @param {*}      [props.subTitle]
  * @param {*}      [props.actions]
- * @param {*}      [props.tabbedNavigation] Tabbed navigation, rendered inside the sticky header block.
+ * @param {*}      [props.tabbedNavigation] A `TabbedNavigation` element; its bar renders inside
+ *                                          the sticky header block and the page children render
+ *                                          inside its active tab panel.
  * @param {string} [props.className]
  * @param {*}      props.children
  * @return {JSX.Element} Page component.
  */
 const Page = ( { breadcrumbItems = [], badges, subTitle, actions, tabbedNavigation, className, children } ) => {
 	const currentLabel = breadcrumbItems[ breadcrumbItems.length - 1 ]?.label;
-	return (
-		<NavigableRegion className={ classnames( 'newspack-page', className ) } ariaLabel={ currentLabel }>
+
+	const renderShell = ( tabBar, content ) => (
+		<>
 			<Stack direction="column" className="newspack-page__header-region">
 				<Stack direction="column" className="newspack-page__header">
 					<Stack direction="row" gap="sm" justify="space-between">
@@ -60,9 +70,17 @@ const Page = ( { breadcrumbItems = [], badges, subTitle, actions, tabbedNavigati
 						</Text>
 					) }
 				</Stack>
-				{ tabbedNavigation }
+				{ tabBar }
 			</Stack>
-			{ children }
+			{ content }
+		</>
+	);
+
+	return (
+		<NavigableRegion className={ classnames( 'newspack-page', className ) } ariaLabel={ currentLabel }>
+			{ isValidElement( tabbedNavigation )
+				? cloneElement( tabbedNavigation, { renderShell, content: children } )
+				: renderShell( null, children ) }
 		</NavigableRegion>
 	);
 };
--- a/packages/components/src/page/index.js
+++ b/packages/components/src/page/index.js
@@ -1,9 +1,10 @@
 /**
  * WordPress dependencies.
  */
-import { Page as AdminUIPage } from '@wordpress/admin-ui';
+import { NavigableRegion } from '@wordpress/admin-ui';
 import { __experimentalHStack as HStack } from '@wordpress/components'; // eslint-disable-line @wordpress/no-unsafe-wp-apis
 import { Icon } from '@wordpress/icons';
+import { Stack, Text } from '@wordpress/ui';
 
 /**
  * External dependencies.
@@ -18,40 +19,51 @@ import Breadcrumbs from '../breadcrumbs';
 import './style.scss';
 
 /**
- * Newspack page header + content region, built on @wordpress/admin-ui's Page.
- *
- * The branded Newspack logo is supplied as admin-ui's decorative (aria-hidden)
- * `visual` slot. The breadcrumb trail (with the current page as the single h1)
- * is supplied as admin-ui's `breadcrumbs` ReactNode.
+ * Newspack page region: a sticky header block (breadcrumbs, actions, optional
+ * tabbed navigation) followed by the page content.
  *
  * @param {Object} props
- * @param {Array}  props.breadcrumbItems Trail items: `{ label, url? }`.
+ * @param {Array}  props.breadcrumbItems    Trail items: `{ label, url? }`.
  * @param {*}      [props.badges]
  * @param {*}      [props.subTitle]
  * @param {*}      [props.actions]
+ * @param {*}      [props.tabbedNavigation] Tabbed navigation, rendered inside the sticky header block.
  * @param {string} [props.className]
  * @param {*}      props.children
  * @return {JSX.Element} Page component.
  */
-const Page = ( { breadcrumbItems = [], badges, subTitle, actions, className, children } ) => {
+const Page = ( { breadcrumbItems = [], badges, subTitle, actions, tabbedNavigation, className, children } ) => {
 	const currentLabel = breadcrumbItems[ breadcrumbItems.length - 1 ]?.label;
 	return (
-		<AdminUIPage
-			className={ classnames( 'newspack-page', className ) }
-			ariaLabel={ currentLabel }
-			visual={ <Icon icon={ newspack } /> }
-			breadcrumbs={
-				<HStack className="newspack-page__breadcrumbs" justify="flex-start">
-					<Breadcrumbs items={ breadcrumbItems } />
-				</HStack>
-			}
-			badges={ badges }
-			subTitle={ subTitle }
-			actions={ actions }
-			hasPadding={ false }
-		>
+		<NavigableRegion className={ classnames( 'newspack-page', className ) } ariaLabel={ currentLabel }>
+			<Stack direction="column" className="newspack-page__header-region">
+				<Stack direction="column" className="newspack-page__header">
+					<Stack direction="row" gap="sm" justify="space-between">
+						<Stack direction="row" gap="sm" align="center" justify="start">
+							<div className="newspack-page__header-visual" aria-hidden="true">
+								<Icon icon={ newspack } />
+							</div>
+							<HStack className="newspack-page__breadcrumbs" justify="flex-start">
+								<Breadcrumbs items={ breadcrumbItems } />
+							</HStack>
+							{ badges }
+						</Stack>
+						{ actions && (
+							<Stack direction="row" gap="sm" align="center" className="newspack-page__header-actions">
+								{ actions }
+							</Stack>
+						) }
+					</Stack>
+					{ subTitle && (
+						<Text render={ <p /> } variant="body-md" className="newspack-page__header-subtitle">
+							{ subTitle }
+						</Text>
+					) }
+				</Stack>
+				{ tabbedNavigation }
+			</Stack>
 			{ children }
-		</AdminUIPage>
+		</NavigableRegion>
 	);
 };
 
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -1,3 +1,4 @@
+@use "../../../colors/colors.module" as colors;
 @use "~@wordpress/base-styles/colors" as wp-colors;
 @use "~@wordpress/base-styles/mixins" as wp-mixins;
 @use "~@wordpress/base-styles/variables" as wp-vars;
@@ -12,6 +13,30 @@
 // build time.
 /* stylelint-disable-next-line no-invalid-position-at-import-rule -- sass @use must precede the CSS @import; postcss-import inlines it at build time. */
 @import "@wordpress/theme/src/prebuilt/css/design-tokens.css";
+
+// Newspack brand mapping: the wpds brand color tokens ship in WordPress
+// blue; remap them onto the Newspack primary scale so every design-system
+// component (focus rings, brand buttons, toggles) renders in brand colors
+// without per-component overrides.
+/* stylelint-disable plugin-wpds/no-setting-wpds-custom-properties -- deliberate Newspack brand override of the wpds brand tokens. */
+:root {
+	--wpds-color-background-interactive-brand-strong: #{colors.$primary-600};
+	--wpds-color-background-interactive-brand-strong-active: #{colors.$primary-800};
+	--wpds-color-background-interactive-brand-weak-active: #{colors.$primary-000};
+	--wpds-color-background-surface-brand: #{colors.$primary-000};
+	--wpds-color-background-thumb-brand: #{colors.$primary-600};
+	--wpds-color-background-thumb-brand-active: #{colors.$primary-600};
+	--wpds-color-foreground-interactive-brand: #{colors.$primary-600};
+	--wpds-color-foreground-interactive-brand-active: #{colors.$primary-800};
+	--wpds-color-foreground-interactive-brand-disabled: #{colors.$primary-400};
+	--wpds-color-stroke-focus: #{colors.$primary-600};
+	--wpds-color-stroke-interactive-brand: #{colors.$primary-600};
+	--wpds-color-stroke-interactive-brand-active: #{colors.$primary-800};
+	--wpds-color-stroke-interactive-brand-disabled: #{colors.$primary-100};
+	--wpds-color-stroke-surface-brand: #{colors.$primary-000};
+	--wpds-color-stroke-surface-brand-strong: #{colors.$primary-600};
+}
+/* stylelint-enable plugin-wpds/no-setting-wpds-custom-properties */
 
 [role="region"].newspack-page {
 	background-color: wp-colors.$white;
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -1,4 +1,5 @@
 @use "~@wordpress/base-styles/colors" as wp-colors;
+@use "~@wordpress/base-styles/mixins" as wp-mixins;
 @use "~@wordpress/base-styles/variables" as wp-vars;
 
 [role="region"].newspack-page {
@@ -6,9 +7,11 @@
 }
 
 .newspack-page__header-region {
-	position: sticky;
-	top: var(--wp-admin--admin-bar--height);
-	z-index: 3;
+	@include wp-mixins.break-medium {
+		position: sticky;
+		top: var(--wp-admin--admin-bar--height);
+		z-index: 3;
+	}
 }
 
 .newspack-page__header {
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -5,18 +5,41 @@
 	background-color: wp-colors.$white;
 }
 
-.newspack-page > [class*="__header"] {
+.newspack-page__header-region {
+	position: sticky;
 	top: var(--wp-admin--admin-bar--height);
+	z-index: 3;
 }
 
-.newspack-page > .newspack-tabbed-navigation {
-	position: static;
+.newspack-page__header {
+	/* stylelint-disable-next-line plugin-wpds/no-unknown-ds-tokens -- the plugin's token list lags the prebuilt design-tokens.css names. */
+	background: var(--wpds-color-background-surface-neutral-strong);
+	border-block-end: var(--wpds-border-width-xs) solid var(--wpds-color-stroke-surface-neutral-weak);
+	padding: var(--wpds-dimension-padding-lg) var(--wpds-dimension-padding-2xl);
 }
 
-.newspack-page [class*="header-visual"] svg path {
-	transform: scale(1.1111);
-	transform-box: view-box;
-	transform-origin: center;
+.newspack-page__header-actions {
+	flex-shrink: 0;
+}
+
+.newspack-page__header-subtitle {
+	/* stylelint-disable-next-line plugin-wpds/no-unknown-ds-tokens -- the plugin's token list lags the prebuilt design-tokens.css names. */
+	color: var(--wpds-color-foreground-content-neutral-weak);
+	padding-block-end: var(--wpds-dimension-padding-xs);
+}
+
+.newspack-page__header-visual {
+	display: grid;
+	flex-shrink: 0;
+	height: 24px;
+	place-items: center;
+	width: 24px;
+
+	svg path {
+		transform: scale(1.1111);
+		transform-box: view-box;
+		transform-origin: center;
+	}
 }
 
 .newspack-page__breadcrumbs {
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -3,6 +3,13 @@
 @use "~@wordpress/base-styles/variables" as wp-vars;
 
 // Design-system tokens ride with the Page header into any consuming bundle.
+// Known tech debt: this reaches past the package's public exports entry
+// (`@wordpress/theme/design-tokens.css`) into its internal layout, because
+// postcss-import cannot resolve exports maps, and resolving the public
+// specifier through sass/webpack leaks a bogus `wp-theme/design-tokens.css`
+// script dependency into consumers' asset manifests, which silently blocks
+// their enqueues. If upstream moves `src/prebuilt/`, this fails loudly at
+// build time.
 /* stylelint-disable-next-line no-invalid-position-at-import-rule -- sass @use must precede the CSS @import; postcss-import inlines it at build time. */
 @import "@wordpress/theme/src/prebuilt/css/design-tokens.css";
 
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -2,6 +2,10 @@
 @use "~@wordpress/base-styles/mixins" as wp-mixins;
 @use "~@wordpress/base-styles/variables" as wp-vars;
 
+// Design-system tokens ride with the Page header into any consuming bundle.
+/* stylelint-disable-next-line no-invalid-position-at-import-rule -- sass @use must precede the CSS @import; postcss-import inlines it at build time. */
+@import "@wordpress/theme/src/prebuilt/css/design-tokens.css";
+
 [role="region"].newspack-page {
 	background-color: wp-colors.$white;
 }
--- a/packages/components/src/page/style.scss
+++ b/packages/components/src/page/style.scss
@@ -5,6 +5,14 @@
 	background-color: wp-colors.$white;
 }
 
+.newspack-page > [class*="__header"] {
+	top: var(--wp-admin--admin-bar--height);
+}
+
+.newspack-page > .newspack-tabbed-navigation {
+	position: static;
+}
+
 .newspack-page [class*="header-visual"] svg path {
 	transform: scale(1.1111);
 	transform-box: view-box;
--- a/packages/components/src/tabbed-navigation/index.js
+++ b/packages/components/src/tabbed-navigation/index.js
@@ -1,7 +1,6 @@
 /**
  * WordPress dependencies.
  */
-import { ThemeProvider } from '@wordpress/theme';
 import { Tabs } from '@wordpress/ui';
 
 /**
@@ -19,19 +18,6 @@ const { useHistory, useLocation } = Router;
 
 const getItemValue = item => item.path || item.href;
 
-/**
- * Whether a navigation item is the active one.
- *
- * Router-driven items match on `path` (exact by default, prefix when
- * `exact: false`) and optionally on `activeTabPaths` (a trailing `*` makes an
- * entry a prefix match) so hidden subpages can keep a parent tab selected.
- * Outside a router, an item is active when explicitly marked `selected` or
- * when its `href` is the current URL.
- *
- * @param {Object}      item     Navigation item.
- * @param {string|null} pathname Current router pathname, or null when rendered outside a router.
- * @return {boolean} Whether the item is active.
- */
 export const isItemActive = ( item, pathname ) => {
 	if ( item.selected ) {
 		return true;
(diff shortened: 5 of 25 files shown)


packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,149 @@
...
+							return (
+								<Tabs.Tab
+									key={ getItemValue( item ) }
+									value={ getItemValue( item ) }
+									disabled={ isDisabled }
+									nativeButton={ false }
+									render={
+										// eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/anchor-is-valid -- content is supplied via the Tab children through @wordpress/ui's render prop, and disabled tabs intentionally drop the href.
+										<a href={ isDisabled ? undefined : href } onClick={ event => onTabClick( event, item ) } />
+									}
```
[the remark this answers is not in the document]
thomasguillot: @wordpress/ui Tabs is built on @base-ui/react (not Ariakit). Its useButton merges our onClick so that when \`disabled\` it calls preventDefault and returns without invoking the external handler, and useFocusableWhenDisabled already sets \`aria-disabled="true"\` on the non-native anchor. So navigation is already blocked (mouse + keyboard) and the semantics are guaranteed by the library, on top of the dropped href and CSS \`pointer-events: none\`. Applied your readability point anyway: the onClick is now omitted entirely when the tab is disabled, so we no longer lean on the library to swallow it.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,149 @@
...
+		return true;
+	}
+	if ( Array.isArray( item.activeTabPaths ) ) {
+		return item.activeTabPaths.some( path => ( path.endsWith( '*' ) ? pathname.startsWith( path.slice( 0, -1 ) ) : path === pathname ) );
+	}
+	if ( false === item.exact && item.path ) {
+		return pathname.startsWith( item.path );
+	}
+	return false;
+};
```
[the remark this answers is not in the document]
thomasguillot: Added a focused suite at packages/components/src/tabbed-navigation/index.test.js covering isItemActive across all branches: the \`selected\` flag, href matching outside a router, exact path, the default no-prefix-match, \`exact: false\` prefix matching, and \`activeTabPaths\` exact + wildcard (including the hidden-subpage parent-tab case). Mirrors the existing wizard/breadcrumbs-select.test.js pattern.

packages/components/src/tabbed-navigation/index.js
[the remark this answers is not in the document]
thomasguillot: Landed in 2cf572515.

packages/components/src/tabbed-navigation/index.js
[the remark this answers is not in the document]
thomasguillot: Landed in 2cf572515.

packages/components/src/tabbed-navigation/style.scss
```diff
@@ -2,81 +2,35 @@
...
+			box-shadow: none;
+		}
 
-				&:hover {
-					color: inherit;
-				}
-			}
+		&[data-disabled] {
+			pointer-events: none;
 		}
```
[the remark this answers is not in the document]
thomasguillot: Good catch, confirmed the same aria-disabled vs data-disabled mismatch here. Fixed in ecad070.

packages/components/src/tabbed-navigation/style.scss
```diff
@@ -2,81 +2,35 @@
...
-		display: flex;
-		flex-wrap: wrap;
-		margin: 0;
-	}
-
-	li {
-		margin-bottom: 0;
-		width: 100%;
+	a[role="tab"] {
+		color: wp-colors.$gray-900;
```
adekbadek, line 70: **Severity:** blocker

This unconditional `color: wp-colors.$gray-900` on `a[role="tab"]` overrides `@wordpress/ui`'s own disabled rule (`&[data-disabled] { color: var(--wpds-color-foreground-interactive-neutral-disabled) }`), so disabled tabs render identical to enabled ones.

Verified in a browser on the Setup wizard at `#/settings`: the disabled "Services" and "Design" tabs compute to `rgb(30, 30, 30)` at `opacity: 1` — the exact same colour as the active tab. The design-system token is present and correct (`--wpds-color-foreground-interactive-neutral-disabled` resolves to `#8d8d8d`) and the elements do carry `data-disabled`, so the DS would style them properly if this rule let it. The library's `cursor: default` survives only because nothing here overrides it.

This is a regression against the previous stylesheet, which gave disabled tabs `opacity: 0.3` and `cursor: not-allowed`. A publisher part-way through the Setup wizard now sees tabs that look completely normal and clickable, and clicking them does nothing at all (`pointer-events: none`) — no dimming, no `not-allowed` cursor, no feedback of any kind. It also works against the goal of the PR, since the design system's disabled treatment is being adopted everywhere except the one state where it carries the most meaning.

Scoping the rule to `a[role="tab"]:not([data-disabled])`, or dropping the blanket `color` override and letting the tokens drive it, restores the disabled affordance.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,104 @@
...
+										onClick={ isDisabled ? undefined : event => onTabClick( event, item ) }
+									/>
+								}
+							>
+								{ item.label }
+							</Tabs.Tab>
+						);
+					} ) }
+				</Tabs.List>
+				{ displayedItems.map( item => (
```
adekbadek: **Severity:** blocker

One empty `Tabs.Panel` is rendered per tab so the tab/panel count check passes, while the routed page content lives outside `Tabs.Root` entirely. In the shipped DOM each tab therefore carries `aria-controls` pointing at an empty `role="tabpanel"`. A screen-reader user activates "Segments", is told they have opened a tab panel, follows `aria-controls`, and lands in an empty region. The WAI-ARIA tabs pattern requires the panel to *contain* the tab's content, and the validation that forces these panels to exist is precisely the check that exists to prevent this shape — its message says mismatched associations "can break screen reader navigation and violate WAI-ARIA Tabs pattern requirements".

The cost/benefit is inverted: `TabsValidationProvider` is gated on `process.env.NODE_ENV !== 'production'` and tree-shakes out of production builds, so production pays broken tabpanel semantics purely to silence a development-mode assertion. The admin-header case is the sharpest — those "tabs" are plain links that trigger a full page load, yet they now claim `role="tab"` and own an empty panel.

For the `Wizard` / `withWizardScreen` / Story Budget consumers the routed content is right there and could be rendered *inside* the matching `Tabs.Panel`, which would make the ARIA true and delete the workaround. The admin header, where the content is PHP-rendered below the mount, cannot do that — which is itself the signal that it should stay navigation (links with `aria-current`) styled to match, rather than borrowing the ARIA of a widget it is not implementing.

If the empty panels do stay, this line needs a comment recording the constraint: the two `displayedItems.map()` calls must now remain in lockstep forever or development builds hard-crash the admin, and nothing in the file says so.

packages/components/src/tabbed-navigation/index.test.js
```diff
@@ -0,0 +1,79 @@
+/**
```
adekbadek, line 7: **Severity:** suggestion

This suite does not execute in any CI job, so it currently provides no regression protection.

`packages/components` ships no `test` script; the workspace root's `test:js` filters only `./plugins/*` and `./themes/*`; and `newspack-plugin`'s jest run (`rootDir` = the plugin, `testMatch` `<rootDir>/**/*test.js?(x)`) does not crawl the symlinked `packages/` directory — its run is exactly the 28 suites under `src/`, with zero mentions of `tabbed-navigation`. CI is green without this file ever being loaded.

When forced to run with an explicit `rootDir`, `packages/components/src` yields 9 suites / 54 tests and they all pass, including this one — so the tests are correct, just dead. That also means the rest of the package's existing suite (breadcrumbs, page, wizard, with-wizard-screen, …) is unprotected today, which is a pre-existing gap rather than something this PR caused. But this PR is what adds a new test into that dead pool, in direct response to a request for test coverage, so it seems worth fixing here: a `"test": "newspack-scripts test"` script in `packages/components/package.json` plus adding `./packages/*` to the root `test:js` filter would make all 54 tests real.

packages/components/src/tabbed-navigation/index.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+	describe( 'outside a router (pathname is null)', () => {
+		afterEach( () => {
+			delete window.location;
+			window.location = new URL( 'http://localhost/' );
+		} );
+
+		it( 'is active when the href matches the current URL', () => {
+			delete window.location;
+			window.location = new URL( 'http://example.com/wp-admin/admin.php?page=ads' );
```
adekbadek, line 50: **Severity:** suggestion

Every assertion here exercises `isItemActive`, which is the least fragile thing in the rewrite — a pure function over strings. The genuinely new behaviour, the code that actually replaced `NavLink`, is untested: `onTabClick`'s modifier-key and `button` bail-outs, the `preventDefault` + `history.push` interception that keeps `history.block` (unsaved-changes) guards working, the `disableUpcoming` index arithmetic including the "no tab matches the route, so every tab is disabled" case, and the `hasRoutedItems` component swap that decides whether the router hooks are called at all.

A few render-level tests with a `MemoryRouter` — click a tab and assert `history.push` fired, cmd-click and assert it did not, assert a `history.block` guard is consulted — would cover the parts a regression would actually break. All of these behaviours were verified by hand in a browser for this review and do work today; the point is that nothing will catch it when they stop.

packages/components/src/divider/style.scss
```diff
@@ -46,6 +46,6 @@
 	}
 
 	&--variant-tertiary {
-		background-color: wp-colors.$gray-100;
+		background-color: var( --wpds-color-stroke-surface-neutral-weak );
```
adekbadek: **Severity:** suggestion

The design-token stylesheet is `@import`ed from `page/style.scss` only, but `divider`, `tabbed-navigation` and `breadcrumbs` all read `--wpds-*` custom properties. Inside `newspack-plugin` this happens to work, because `Page` lands in the same bundle.

`newspack-components` is a published package, though: any consumer that imports `Divider` or `TabbedNavigation` without also importing `Page` gets undefined custom properties, and `background-color: var( --wpds-color-stroke-surface-neutral-weak )` then resolves to nothing — an invisible divider, and a tab bar with no bottom border, with no error to explain it.

`breadcrumbs/style.scss` already guards against exactly this with a fallback (`var( --wpds-color-stroke-surface-neutral, #dbdbdb )`). Either give every `--wpds-*` reference a fallback the same way, or import the token sheet from each stylesheet that consumes tokens (postcss-import dedupes within a bundle), so token availability is not implicitly coupled to `Page` being in the tree.

plugins/newspack-plugin/src/wizards/admin-header/index.tsx
```diff
@@ -16,24 +16,21 @@ export function WizardsAdminHeader( {
...
-								</li>
-							);
-						} ) }
-					</ul>
-				</div>
-			) }
-		</Page>
+		<Page
+			breadcrumbItems={ breadcrumbs }
+			actions={ <div id="newspack-wizards-admin-header-actions" /> }
```
adekbadek, line 27: **Severity:** suggestion

The previous version carried a six-line comment on this `actions` prop recording that `#newspack-wizards-admin-header-actions` is a cross-root portal target — the newsletters admin shell portals its header actions into it from a separate React root — and that this only works because the node is rendered once and never re-rendered ("do not give this component state that could re-render"). That comment is gone, while the same change adds a stateful subtree (`Tabs.Root` / `Tabs.List`, which keeps its own resize/overflow state) directly beneath it.

The state stays local to `Tabs.List`, so the portal target is not reconciled today and nothing is broken. But the invariant is still live and the only record of it has been deleted, which sets up the next person to add state here and break the newsletters header actions silently. Worth restoring on the `actions` prop.

package.json
```diff
@@ -79,7 +79,8 @@
 			"@typescript-eslint/parser": "8.61.0",
 			"prettier": "npm:wp-prettier@^3.0.3",
 			"react": "18.3.1",
-			"react-dom": "18.3.1"
+			"react-dom": "18.3.1",
+			"@wordpress/private-apis": "1.49.0"
```
adekbadek: **Severity:** suggestion

`@wordpress/private-apis` is not in `dependency-extraction-webpack-plugin`'s bundled-packages list, so every Newspack entry externalises it to the `wp.privateApis` global — the built asset manifests list `wp-private-apis` as a script dependency, meaning the browser always uses WP core's copy no matter what pnpm resolves. This override therefore only reshapes the `node_modules` graph (jest, type-checking, and downstream consumers of the published `newspack-components`), while exact-pinning `1.49.0` across all 12 plugins and every shared package — a wide blast radius for a change with no runtime effect.

It is also silent about intent: the next `@wordpress/*` bump that wants a newer `private-apis` will be quietly held back by this line with nothing to explain why. Could the PR state what symptom this actually fixed, and whether a scoped override (or simply declaring the dependency in `packages/components`) would do the same job?

Worth noting the duplicates that *do* get bundled are untouched: `@wordpress/admin-ui` resolves to both 2.3.0 and 2.4.0, and `@wordpress/ui` to both 0.15.0 and 0.16.0.

packages/components/src/page/index.js
```diff
@@ -18,40 +19,51 @@ import Breadcrumbs from '../breadcrumbs';
...
-				<HStack className="newspack-page__breadcrumbs" justify="flex-start">
-					<Breadcrumbs items={ breadcrumbItems } />
-				</HStack>
-			}
-			badges={ badges }
-			subTitle={ subTitle }
-			actions={ actions }
-			hasPadding={ false }
-		>
+		<NavigableRegion className={ classnames( 'newspack-page', className ) } ariaLabel={ currentLabel }>
```
adekbadek: **Severity:** suggestion

The base branch deliberately built this on `@wordpress/admin-ui`'s `Page` (visual slot, breadcrumbs slot, badges, subtitle, actions); this change swaps it for `NavigableRegion` plus a hand-assembled `Stack` header, and drops the `showSidebarToggle` / `SidebarToggleFill` slot along the way.

The motivation is legible from the diff — admin-ui's `Page` has no slot for content that sits between the header and the body *inside* the sticky region, which is exactly where the tabs need to go — but nothing in the code or the commits says so. The practical effect is that Newspack now owns a private copy of the design system's page-header markup, padding and tokens, and will drift from admin-ui as that package evolves.

Given that DSGNEWS-143's goal is alignment with the design system, it would help to record the reason (a comment on `Page`, or the PR body) and ideally open an upstream issue for the missing slot, so this can be unwound later rather than quietly becoming a permanent fork.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,104 @@
...
  */
 import './style.scss';
 
-const { NavLink, useHistory } = Router;
+const { useHistory, useLocation } = Router;
 
-const TabbedNavigation = ( { items, className, disableUpcoming, children = null } ) => {
+const getItemValue = item => item.path || item.href;
+
+export const isItemActive = ( item, pathname ) => {
```
adekbadek, line 23: **Severity:** suggestion

`packages/components/src/wizard/breadcrumbs-select.js` already contains a route matcher over the same `{ path, exact, activeTabPaths }` item shape, with its own test file — and the wildcard expression on line 32 here is a byte-for-byte copy of it. The two now disagree in two ways: the breadcrumb matcher treats path matching as *prefix by default* (exact only when `exact === true`), while `isItemActive` treats it as *exact by default* (prefix only when `exact === false`); and the breadcrumb matcher falls through to path matching when `activeTabPaths` misses, whereas `isItemActive` returns early.

The practical consequence shows up in Story Budget, whose items are bare `{ label, path }` with no `exact` key: on `#/stories/new` — a real, reachable route, since the "Add Story" button links to it — no tab is highlighted at all, while the breadcrumb still resolves to "Stories". This is not a regression (the old `NavLink` also defaulted `exact` to `true` and behaved the same way), so it is pre-existing. But this PR is the moment the two matchers become near-identical twins thirty lines apart in the same package.

Extracting one matcher parameterised by the exact-default, and having both call sites use it, would define the wildcard grammar once and let Story Budget's nested routes keep their highlight.

packages/components/src/page/style.scss
```diff
@@ -1,14 +1,52 @@
 @use "~@wordpress/base-styles/colors" as wp-colors;
+@use "~@wordpress/base-styles/mixins" as wp-mixins;
 @use "~@wordpress/base-styles/variables" as wp-vars;
 
+// Design-system tokens ride with the Page header into any consuming bundle.
+/* stylelint-disable-next-line no-invalid-position-at-import-rule -- sass @use must precede the CSS @import; postcss-import inlines it at build time. */
+@import "@wordpress/theme/src/prebuilt/css/design-tokens.css";
```
adekbadek, line 15: **Severity:** suggestion

`@wordpress/theme@0.16.0` publishes `"./design-tokens.css": "./src/prebuilt/css/design-tokens.css"` in its `exports` map, so the supported specifier is `@wordpress/theme/design-tokens.css`. Importing `@wordpress/theme/src/prebuilt/css/design-tokens.css` reaches past that into the package's internal layout, and will break the day upstream moves `src/prebuilt/` — on a caret-ranged dependency on a `0.x` package, that is not hypothetical.

plugins/newspack-story-budget/src/app/index.js
```diff
@@ -100,60 +100,60 @@ const StoryBudget = () => {
...
-			<AppHeader breadcrumbItems={ breadcrumbItems } />
-			<TabbedNavigation items={ navigationItems } />
-			<div className="newspack-story-budget__content">
-				<Switch>
-					<Route path="/stories">
-						<AppHeaderActions>
-							{ canManage && (
-								<Button variant="primary" href="#/stories/new">
-									{ __( 'Add Story', 'newspack-story-budget' ) }
+			<AppHeader breadcrumbItems={ breadcrumbItems } tabbedNavigation={ <TabbedNavigation items={ navigationItems } /> }>
```
adekbadek, line 103: **Severity:** suggestion

`plugins/newspack-story-budget/src/style.scss` still carries a "Fix Tabbed Navigation" block (`position: static`, `z-index: 10`, `li a { box-sizing: border-box; }`) written to compensate for the old implementation. None of it applies any more: the component is no longer the sticky element (the `Page` header region is), `z-index` is inert without positioning, and there are no `<li>`s left to select. The PR that removes the markup those rules were written against is the right place to remove the rules.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,104 @@
...
 		}
-		if ( Array.isArray( item?.activeTabPaths ) ) {
-			return item.activeTabPaths.some( path => {
-				if ( path.endsWith( '*' ) ) {
-					const basePath = path.slice( 0, -1 );
-					return pathname.startsWith( basePath );
-				}
-				return item.activeTabPaths.includes( pathname );
-			} );
+		if ( event.defaultPrevented || event.metaKey || event.ctrlKey || event.shiftKey || event.altKey || event.button !== 0 ) {
```
adekbadek: **Severity:** nit

Non-primary pointer buttons dispatch `auxclick`, not `click`, so React's `onClick` never sees `event.button !== 0`. Middle-click already opens the tab in a new browser tab purely because the element is a real `<a href>` — confirmed in a browser — not because of this guard. Harmless, but it sits in the very handler the PR description credits for making middle-click work, so a reader will reasonably assume it is load-bearing. Either drop it, or note that it is belt-and-braces.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,104 @@
...
+		return true;
+	}
+	if ( null === pathname ) {
+		return Boolean( item.href ) && window.location.href === item.href;
+	}
+	if ( item.path === pathname ) {
+		return true;
+	}
+	if ( Array.isArray( item.activeTabPaths ) ) {
+		return item.activeTabPaths.some( path => ( path.endsWith( '*' ) ? pathname.startsWith( path.slice( 0, -1 ) ) : path === pathname ) );
```
adekbadek: **Severity:** nit

`'/segments*'` strips to `/segments` and would also match `/segments-old`, while the `exact === false` branch three lines down deliberately anchors with `+ '/'` and has a test (`does not match a prefix-colliding sibling`) proving it does. Every current caller writes `'/x/*'`, so the trailing slash saves it in practice, but the two branches encode different rules for the same idea. Anchoring the wildcard the same way — or documenting that the `*` must follow a `/` — removes the trap.

packages/components/src/tabbed-navigation/index.js
```diff
@@ -1,60 +1,104 @@
...
 /**
  * Internal dependencies.
  */
 import './style.scss';
 
-const { NavLink, useHistory } = Router;
+const { useHistory, useLocation } = Router;
 
-const TabbedNavigation = ( { items, className, disableUpcoming, children = null } ) => {
+const getItemValue = item => item.path || item.href;
```
adekbadek: **Severity:** nit

`getItemValue` returns `undefined` for an item with neither `path` nor `href`. Such an item yields `key={ undefined }` (a React key collision if there is more than one) and `value={ undefined }` on both its `Tabs.Tab` and its `Tabs.Panel`, which base-ui treats as an index-based value rather than the intended no-op. No current caller does this, but the item contract is otherwise permissive — every field is optional — so either filter these out alongside `isHiddenInTabbedNavigation`, or fall back to the index.

adekbadek (changes requested):
The refactor is sound and the risky parts hold up under a live browser, but a few things needs to be addressed.

packages/components/src/tabbed-navigation/style.scss
thomasguillot: Fixed in 1dac76d, with one correction to the diagnosis: the wpds tab styles are injected inside @layer wp-ui, and wp-admin's unlayered `a { color }` beats any layered rule regardless of specificity, so dropping our override entirely would leave disabled tabs admin-link blue, not token gray. The enabled color is now scoped to `:not([data-disabled]):not([aria-disabled="true"])`, and the disabled state restates the DS token from our unlayered rule: `color: var(--wpds-color-foreground-interactive-neutral-disabled, ...)`. Verified in the browser on Setup at #/settings: disabled tabs compute to rgb(141,141,141).

packages/components/src/tabbed-navigation/index.js
thomasguillot: Fixed in 1dac76d. Router consumers now render the routed content inside the active tab's `Tabs.Panel`: `Page` hands its children to `TabbedNavigation` through a layout callback, so the bar stays in the sticky header region while the active panel (the one `aria-controls` points at) contains the actual page content; inactive panels unmount. The admin header takes the other path you suggested: href-only items now render as a plain `nav` of links with `aria-current`, styled to match, with no tab ARIA at all. The 1:1 tab/panel pairing is documented at the panel map. Verified in the browser across Setup, Audience, Newsletters, and Story Budget.

packages/components/src/tabbed-navigation/index.test.js
thomasguillot: Fixed in 3d889fc. `packages/components` now has `"test": "newspack-scripts test"`, the root `test:js` includes `./packages/*` with `--if-present`, and the jest config ignores the compiled `dist/`/`shared/` copies (anchored at the package root so plugin `src/shared/` tests keep running). The package suite is 85 tests and CI's per-package job picks it up whenever components change.

packages/components/src/tabbed-navigation/index.test.js
thomasguillot: Added in 1dac76d. The suite now renders through `MemoryRouter` and covers the parts you listed: click fires `history.push` and moves selection, cmd-click is left to the browser, a `history.block` guard returning false keeps the route and selection unchanged, `disableUpcoming` disables later tabs (and all tabs when no route matches), the href-only branch renders nav links instead of a tabs widget, and the active panel contains the content its tab's `aria-controls` points at. jestSetup gained a ResizeObserver polyfill (3d889fc) so the real @wordpress/ui Tabs renders under jsdom.

packages/components/src/divider/style.scss
thomasguillot: Fixed in 0e2bbfa and 1dac76d. Every `--wpds-*` reference in `divider` and `tabbed-navigation` now carries a fallback, using the wp-vars/base-styles equivalents rather than hardcoded values (`$gray-100`, `$border-width`, `$default-font-size`, `$grid-unit-20`, `$border-width-focus-fallback`), so a consumer without the Page bundle degrades to the WordPress admin values instead of nothing.

plugins/newspack-plugin/src/wizards/admin-header/index.tsx
thomasguillot: Restored. The cross-root portal comment is back on the `actions` div in `admin-header/index.tsx` (it came back as part of rebasing this branch onto the updated base, which is where it was introduced), so the render-once contract is recorded next to the new `Tabs`-adjacent code. Verified in the browser that the newsletters admin-shell still portals its actions in: the list page shows "Add Newsletter" in the header.

package.json
thomasguillot: Removed in 96056ee. You were right that it had no runtime effect; it was added to collapse a transient 1.48/1.49 duplication during development, and after this branch's dependency alignment the graph resolves a single `private-apis@1.49.0` without it (verified by removing it, regenerating the lockfile, and re-running tests and builds). The same commit aligns `newspack-components`' `@wordpress/style-runtime` to `^0.5.0`, which is what `ui@0.16` resolves. The remaining bundled 0.15/0.4 duplicates come from `admin-ui@2.4.0`'s own internals and go away only when admin-ui moves to ui 0.16 upstream; `admin-ui@2.3.0` is pulled by `@wordpress/edit-post` and is externalized at runtime, so it never lands in our bundles.

packages/components/src/page/index.js
thomasguillot: Documented in 1dac76d. The `Page` docblock now records why admin-ui's `Page` isn't used (no slot for content between the header and the body inside the sticky region, which is where the tab bar must live) and that the markup is assembled from the same primitives and tokens so it can be swapped back once admin-ui grows such a slot. I'll file the upstream slot request separately.

packages/components/src/tabbed-navigation/index.js
thomasguillot: Fixed in b8c7234 and 1dac76d. Both call sites now use a single `matchesRoute` in `packages/components/src/route-match.js`: prefix matching by default at segment boundaries, exact for the root path and `exact: true` opt-ins, anchored wildcards, and fall-through from `activeTabPaths` to path matching. Tab selection also prefers the most specific match so nested sibling paths can't outrank each other. Story Budget's #/stories/new now keeps the Stories tab highlighted, verified in the browser.

packages/components/src/page/style.scss
thomasguillot: Tried the public specifier both ways and it can't work in this toolchain, so the deep path stays with the constraint recorded (1dac76d). Details: postcss-import (via @wordpress/postcss-plugins-preset) has no exports-map support, so `@import "@wordpress/theme/design-tokens.css"` fails outright. Routing it through sass/webpack instead resolves at build time for src consumers, but for package-dist consumers it leaks a bogus `wp-theme/design-tokens.css` script dependency into the built asset manifests, and an unregistered handle makes WordPress silently skip the whole enqueue; I caught the newsletters admin-shell script disappearing this way in the browser. The deep path fails loudly at build time if upstream moves `src/prebuilt/`, which is the least-bad failure mode available.

plugins/newspack-story-budget/src/app/index.js
thomasguillot: Removed in 4ece476.

packages/components/src/tabbed-navigation/index.js
thomasguillot: Dropped in 1dac76d. The handler comment now notes that middle clicks dispatch auxclick and are handled by the native anchor href, so nothing implies the guard was load-bearing.

packages/components/src/tabbed-navigation/index.js
thomasguillot: Fixed in b8c7234. Wildcards are anchored at a segment boundary in the shared matcher: `'/segments*'` matches `/segments` and `/segments/123` but never `/segments-old`, same as the `exact: false` branch, with tests pinning it.

packages/components/src/tabbed-navigation/index.js
thomasguillot: Fixed in 1dac76d. `getItemValue` falls back to an index-based value for items with neither `path` nor `href`, so keys stay unique and base-ui gets a real value.

thomasguillot:
All review feedback is addressed. Summary of the round:

- **Rebased onto the updated base branch** (472 was rebased onto main earlier today), so the merge conflict is gone.
- **Both blockers fixed in 1dac76d**: routed content now renders inside the active `Tabs.Panel` (real ARIA), the PHP admin header renders as plain navigation links with `aria-current`, and disabled tabs show the design system's disabled color. One finding from the disabled-tab fix: the wpds styles sit in `@layer wp-ui`, so wp-admin's unlayered link color always wins over them; our unlayered rule restates the DS token instead.
- **Suggestions**: package tests wired into CI (3d889fc, 85 tests), render-level tab tests added, one shared route matcher for tabs and breadcrumbs (b8c7234, also fixes Story Budget's nested-route highlight), wpds token fallbacks via wp-vars (0e2bbfa), Page fork documented, private-apis override dropped and style-runtime aligned (96056ee), stale Story Budget CSS removed (4ece476).
- **Design-tokens specifier**: kept the deep path after testing both public-specifier routes; details and the failure mode they cause are on the thread.
- **Nits**: dead button guard dropped, wildcards anchored, `getItemValue` index fallback.

Verified in the browser on Setup (disabled tabs), Audience Campaigns (panel content, tab navigation), Newsletters list (nav-link header, portalled actions), and Story Budget (#/stories/new highlight). Suites: components 85, newspack-plugin 330, newsletters 204, all passing.

packages/components/src/page/style.scss
```diff
@@ -1,14 +1,84 @@
...
+// Design-system tokens ride with the Page header into any consuming bundle.
+// Known tech debt: this reaches past the package's public exports entry
+// (`@wordpress/theme/design-tokens.css`) into its internal layout, because
+// postcss-import cannot resolve exports maps, and resolving the public
+// specifier through sass/webpack leaks a bogus `wp-theme/design-tokens.css`
+// script dependency into consumers' asset manifests, which silently blocks
+// their enqueues. If upstream moves `src/prebuilt/`, this fails loudly at
+// build time.
+/* stylelint-disable-next-line no-invalid-position-at-import-rule -- sass @use must precede the CSS @import; postcss-import inlines it at build time. */
+@import "@wordpress/theme/src/prebuilt/css/design-tokens.css";
```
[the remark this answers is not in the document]
thomasguillot: Good catch, the description was describing an earlier iteration. Updated it to match the shipped approach: tokens ride with the Page header styles (prebuilt design-tokens.css inlined in page/style.scss) with the brand remap on :root, no ThemeProvider. The physical-path import is deliberate; reasoning is on the design-tokens thread from the previous round.

packages/components/src/tabbed-navigation/style.scss
adekbadek: Verified fixed on d0fa260, and your correction to the diagnosis is right — I checked the layering and unlayered `a { color }` does beat the layered DS rule, so restating the token from an unlayered rule is the correct fix rather than dropping the override.

Browser, Setup wizard at `#/settings`: disabled "Services"/"Design" now compute to `rgb(141, 141, 141)` (the `#8d8d8d` token) against `rgb(30, 30, 30)` for the active tab, `pointer-events: none`, no `href`. The enabled/disabled split is visible rather than pixel-identical.

packages/components/src/tabbed-navigation/index.js
adekbadek: Verified fixed on d0fa260, both paths.

Routed: on Setup `#/settings` the active tab 's `aria-controls` resolves to a panel containing the real content (5,258 chars, "Site Profile" heading, form controls); inactive panels unmount, so no empty `role="tabpanel"` is left in the DOM. Story Budget `#/stories/new` keeps Stories selected with the content in its panel.

Link-driven: the newsletter Ads screen renders `<nav aria-label="Secondary navigation">` with zero tab/tablist/tabpanel ARIA anywhere on the page and `aria-current="page"` on the active link. Splitting the two cases is a better outcome than what I proposed.

packages/components/src/tabbed-navigation/index.test.js
adekbadek: Verified fixed on 3d889fc. The `JS / newspack-components` CI job on this run now actually executes the suite — the job log shows `> newspack-scripts test`, `PASS src/tabbed-navigation/index.test.js`, `PASS src/route-match.test.js`, `Test Suites: 12 passed`, `Tests: 85 passed`. The check was green before this without the file ever loading, so this is the real difference.

packages/components/src/tabbed-navigation/index.test.js
adekbadek: Verified on d0fa260. The suite now covers the behaviour rather than just the pure function, and `consults history.block guards before navigating` asserts both that the pathname is unchanged and that selection does not drift — which is exactly the failure mode that matters.

Confirmed the same thing in a browser against a dirty Segment form: clicking another tab fires "There are unsaved changes to this segment. Discard changes?", dismissing keeps the route on `#/segments/new`, the Segments tab stays selected, and the typed value survives.

packages/components/src/divider/style.scss
adekbadek: Verified fixed on 0e2bbfa/1dac76d. Every `--wpds-*` reference in `divider` and `tabbed-navigation` now carries a fallback, so a Page-less consumer degrades to the wp-admin values instead of nothing.

plugins/newspack-plugin/src/wizards/admin-header/index.tsx
adekbadek: Verified restored on d0fa260, and it reads better where it now sits — directly above the `Page` call, so the render-once contract is stated next to the `Tabs`-adjacent code rather than on the prop alone.

package.json
adekbadek: Verified removed on 96056ee — the override is gone from the root `package.json` and CI is green, so the graph resolves a single copy without it. The remaining admin-ui-internal duplicates are upstream's to shed; no objection to leaving those.

packages/components/src/page/index.js
adekbadek: Resolved by the docblock on d0fa260 — recording the missing-slot constraint and that the markup is assembled from the same primitives so it can be swapped back is what this needed. Filing the upstream slot request separately is the right follow-up.

packages/components/src/tabbed-navigation/index.js
adekbadek: Verified fixed on b8c7234. Both call sites share `matchesRoute`, and the two matchers no longer disagree: prefix-by-default at segment boundaries, exact for root and `exact: true`, anchored wildcards, and fall-through from `activeTabPaths`. `route-match.test.js` pins each of those.

The most-specific-match preference is a good catch beyond what I raised — confirmed in a browser that Story Budget keeps Stories selected on `#/stories/new` and `#/stories/sites`.

packages/components/src/page/style.scss
adekbadek: Fair — the constraint is real and the comment records it well, including that it fails loudly at build time if upstream moves `src/prebuilt/`. Silently skipped enqueues would be a worse trade. Leaving this as-is.

plugins/newspack-story-budget/src/app/index.js
adekbadek: Verified removed on 4ece476.

packages/components/src/tabbed-navigation/index.js
adekbadek: Verified on d0fa260 — the comment now attributes middle-click to the native anchor href, so nothing reads as load-bearing.

packages/components/src/tabbed-navigation/index.js
adekbadek: Verified fixed on b8c7234. `matchesWildcard` anchors at the `/` boundary and `route-match.test.js` pins that `/segments*` matches `/segments` and `/segments/123` but not `/segments-old`, so both branches now encode the same rule.

packages/components/src/tabbed-navigation/index.js
adekbadek: Verified fixed on d0fa260 — `getItemValue` falls back to `item-${ index }`, so keys stay unique and base-ui gets a real value.

adekbadek (approved):
All findings addressed and re-verified in a browser on d0fa260: both blockers are genuinely fixed, and the components suite now actually runs in CI (85 tests).

PR #499, opened 2026-07-02
feat(export): post-Memberships subscription & user CSV exports (NPPD-1730)

adekbadek:
### All Submissions:

Deactivating WooCommerce Memberships (a standard Access Control migration step) removes the membership CSV export some publishers rely on for recurring operations – mailing lists, revenue-recognition user pulls, product segmentation. This adds a first-party replacement to newspack-plugin:

- **Batched CSV exporters** (`includes/export/`): `Subscriptions_CSV_Exporter` and `Users_CSV_Exporter` extend WooCommerce's `WC_CSV_Batch_Exporter` (temp file in uploads, page stepping, CSV-injection escaping inherited). Subscriptions rows include schedule dates, products (multi-item concatenated), totals, customer identity, and full billing/shipping – the address data the old Memberships export never had. Users rows include core fields, roles, and WC billing/shipping meta.
- **Export the current filtered view**: an "Export CSV" button on the subscriptions list (HPOS + legacy CPT) and the users list serializes the active list filters (status, product, customer, payment method, group, search, month / role, search) and translates them into paged queries. Paging is `limit`/`offset`-based via `wcs_get_orders_with_meta_query()` with a SQL count – deliberately not `wcs_get_subscriptions( [ 'paged' => n ] )`, whose `paged` arg is silently dropped, and not the unpaged `product_id` path. List sorting is intentionally not carried over (export order is always ID ascending — insert-stable, so offset pagination stays consistent when rows are created mid-export).
- **AJAX-stepped flow** (WC product-exporter style): the browser drives one page per request, so no long request ever runs; the download is nonce-protected, capability re-checked, and deletes the file on send. Temp files live in a hardened `uploads/newspack-csv-exports/` dir with server-generated random filenames; a daily cron sweeps abandoned files.
- **WP-CLI**: `wp newspack export-subscriptions` / `wp newspack export-users` share the same exporters and query translation, with per-batch object-cache clearing so large exports keep flat memory.
- **Extensibility contract**: `newspack_{subscriptions,users}_export_{headers,row,query_args}` filters let other features add columns – NPPD-1719 (group-subscription export) will use this exactly the way Memberships-for-Teams added `team_*` columns to the old Memberships export.
- Promotes `Group_Subscription_Settings::apply_group_filter()` from private to public for reuse by the subscriptions exporter (backward-compatible visibility widening).

Capabilities: subscriptions export requires `manage_woocommerce`; users export requires `list_users` **and** `manage_woocommerce` (the CSV carries Woo billing PII that `list_users` alone does not imply).

Closes NPPD-1730.

1. On a site with WooCommerce + WC Subscriptions and some subscriptions, go to WooCommerce > Subscriptions, apply filters (e.g. status Active + a product), and click **Export CSV**. The progress indicator steps to 100% and a CSV downloads whose rows exactly match the filtered list (count and content).
2. Toggle HPOS (`wp option update woocommerce_custom_orders_table_enabled no` + sync) and repeat on the legacy list – same result.
3. Go to Users, filter by a role, click **Export CSV** – the CSV rows match the filtered user count, with billing/shipping meta columns.
4. Try the download URL with a wrong nonce or as a user without the capability – it is rejected; confirm `uploads/newspack-csv-exports/` holds no files after a completed download.
5. Set a billing first name to `=HYPERLINK("https://evil.example","x")` and export – the cell arrives prefixed with `'` (formula injection neutralized).
6. CLI: `wp newspack export-subscriptions --status=active --product=<id> --output=/tmp/subs.csv` and `wp newspack export-users --role=subscriber --output=/tmp/users.csv` – row counts match the equivalent admin exports.
7. Scale: seeded 40,000 subscriptions – browser export completes in ~2m45s (800 steps, constant pace), CLI in ~1m24s with flat memory.

### Other information:

https://claude.ai/code/session_01WquT3V87kgu5Bv2ihuXZUJ


plugins/newspack-plugin/src/csv-export/index.js
```diff
@@ -0,0 +1,53 @@
+/* globals newspackCsvExport */
+/**
+ * Drives the batched CSV export from the admin list tables: one AJAX request
+ * per page (WooCommerce product-exporter style), then a nonce-protected
+ * download of the assembled file.
+ */
```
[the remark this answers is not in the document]
adekbadek: Applied in `1adc45f6a`, then reverted in `a376578cf` with runtime evidence: the `public-path` import makes webpack split the shared runtime into the `commons` chunk, and the entry's bootstrap then waits for that chunk (`a.O(void 0,[4223],...)`) — which standalone entries never enqueue, so the module never executes and the button silently dies (verified broken in a test env, verified working after the revert). The convention only holds for the wizard entries, which enqueue `commons` alongside. This entry is a self-contained vanilla bundle with no split chunks, matching the other standalone entries (`group-subscription-admin`, `indesign-export`).

plugins/newspack-plugin/includes/export/class-users-csv-exporter.php
```diff
@@ -0,0 +1,200 @@
...
+			'user_email'      => __( 'Email', 'newspack-plugin' ),
+			'display_name'    => __( 'Display Name', 'newspack-plugin' ),
+			'first_name'      => __( 'First Name', 'newspack-plugin' ),
+			'last_name'       => __( 'Last Name', 'newspack-plugin' ),
+			'roles'           => __( 'Roles', 'newspack-plugin' ),
+			'user_registered' => __( 'Registered Date', 'newspack-plugin' ),
+		];
+		foreach ( self::get_address_meta_keys() as $meta_key ) {
+			$columns[ $meta_key ] = ucwords( str_replace( '_', ' ', $meta_key ) );
+		}
```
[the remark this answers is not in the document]
adekbadek: Fixed in `1adc45f6a`: the address column labels now come from a shared `CSV_Batch_Exporter::get_address_column_labels()` map with explicit `__()` strings, used by both the users and subscriptions exporters (one source of truth for ids + labels).

plugins/newspack-plugin/includes/export/class-csv-batch-exporter.php
```diff
@@ -0,0 +1,95 @@
...
+		$upload_dir = \wp_upload_dir();
+		$dir        = \trailingslashit( $upload_dir['basedir'] ) . CSV_Exports::EXPORTS_DIR;
+		if ( ! is_dir( $dir ) ) {
+			\wp_mkdir_p( $dir );
+			// Direct file ops are fine here: the dir is under uploads.
+			// phpcs:disable WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents, WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_file_put_contents
+			file_put_contents( \trailingslashit( $dir ) . 'index.html', '' );
+			file_put_contents( \trailingslashit( $dir ) . '.htaccess', 'deny from all' );
+			// phpcs:enable
+		}
```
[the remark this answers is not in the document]
adekbadek: Fixed in `1adc45f6a`: `get_exports_dir()` now bails when `wp_mkdir_p()` fails, before attempting the hardening writes; downstream file writes surface their own errors (and the AJAX handler reports an unwritable-uploads error instead of serving an empty CSV).

plugins/newspack-plugin/includes/export/class-csv-batch-exporter.php
```diff
@@ -0,0 +1,164 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Shared surface for Newspack CSV exporters: captured list params and temp
+ * files staged in a hardened uploads subdirectory.
+ *
+ * This class must only be loaded after WooCommerce's WC_CSV_Batch_Exporter
+ * abstract (see CSV_Exports::load_exporter_dependencies()).
+ */
+abstract class CSV_Batch_Exporter extends \WC_CSV_Batch_Exporter {
```
chickenn00dle, line 19: **Severity:** suggestion

`composer.json` classmaps the whole `includes` directory (no `exclude-from-classmap`), so after `composer dump-autoload` `CSV_Batch_Exporter`, `Subscriptions_CSV_Exporter`, and `Users_CSV_Exporter` all land in the classmap. But this class `extends \WC_CSV_Batch_Exporter`, which WooCommerce loads on demand (`require_once`, never via an autoloader). Any autoload trigger for these classes while WC's export framework isn't loaded — a bare `class_exists( 'Newspack\Users_CSV_Exporter' )` (autoload defaults to `true`), an `instanceof`, static analysis — would require this file, hit the `extends`, fail to load the parent, and fatal at class-declaration time. The intended runtime path is safe (`get_exporter()` only instantiates after `load_exporter_dependencies()`), so this is latent — but these three are the only classmapped classes that extend a lazily-loaded WC abstract. Cheap to defuse by adding the three export files to `exclude-from-classmap` (they're always loaded explicitly via `require_once`).

plugins/newspack-plugin/includes/export/class-csv-exports.php
```diff
@@ -0,0 +1,451 @@
...
+		if ( ! self::validate_export_filename( $filename, $type ) ) {
+			\wp_die( \esc_html__( 'Invalid download link.', 'newspack-plugin' ), '', 403 );
+		}
+		$exporter->set_filename( $filename );
+		// A served export is deleted on send; a replayed link would otherwise
+		// quietly download a headers-only CSV.
+		if ( ! file_exists( $exporter->get_export_file_path() ) ) {
+			\wp_die( \esc_html__( 'This download link has expired. Please run the export again.', 'newspack-plugin' ), '', 410 );
+		}
+		$exporter->export();
```
chickenn00dle: **Severity:** suggestion

The admin download delegates to WooCommerce's `WC_CSV_Batch_Exporter::export()`, which does `send_content( get_headers_row_file() . get_file() )` — and `get_file()` is a `file_get_contents()` of the whole data file concatenated into one PHP string. Peak memory on the download request is roughly the full file size, so a large filtered export (100k+ rows × ~40 PII columns) can exceed `memory_limit` and fail the download *after* the multi-step build already succeeded. `save_to()` (the CLI path) was deliberately written to stream via `stream_copy_to_stream` to avoid this ~2x peak; the admin path can't reuse it because it hands off to WC. Consider a streamed admin download (`readfile`/`fpassthru` of the temp file) or documenting a practical row ceiling for the button.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
```diff
@@ -0,0 +1,323 @@
...
+			$args['status'] = array_keys( \wcs_get_subscription_statuses() );
+		} else {
+			$args['status'] = [ \wcs_sanitize_subscription_status_key( $status ) ];
+		}
+
+		// Product filter: resolve to a subscription ID set, intersected with
+		// any other ID restriction. Never pass product_id to the query
+		// functions directly — that path runs unpaged.
+		if ( ! empty( $params['_wcs_product'] ) ) {
+			$product_subscription_ids = array_keys( \wcs_get_subscriptions_for_product( absint( $params['_wcs_product'] ) ) );
```
chickenn00dle: **Severity:** suggestion

`build_query_args()` runs inside `prepare_data_to_export()` on every step, so for a product-filtered export `wcs_get_subscriptions_for_product()` — a `LIMIT -1` query returning *all* matching subscription IDs — executes once per page, and the full ID array is passed as `post__in` on each page's query. For a product with tens of thousands of subscriptions that's O(pages × full-set) redundant work plus a large `post__in` per query. (The customer path at :138 is cheaper — WCS caches it.) Consider resolving the product ID set once per run and caching it (e.g. a transient keyed by the export filename). The empty-result case is already safe: `set_post__in_query_var` maps `[]` to a no-match sentinel.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
```diff
@@ -0,0 +1,323 @@
...
+		$date_created = $subscription->get_date_created();
+
+		$row = [
+			'subscription_id'     => $subscription->get_id(),
+			'status'              => $subscription->get_status(),
+			'date_created'        => $date_created ? $date_created->date( 'Y-m-d H:i:s' ) : '',
+			'start_date'          => self::get_date_field( $subscription, 'start' ),
+			'trial_end_date'      => self::get_date_field( $subscription, 'trial_end' ),
+			'next_payment_date'   => self::get_date_field( $subscription, 'next_payment' ),
+			'last_payment_date'   => self::get_date_field( $subscription, 'last_payment' ),
```
chickenn00dle: **Severity:** suggestion

The `last_payment_date` column calls `get_date_field( $subscription, 'last_payment' )`, and WCS deprecated the `last_payment` date type ("...deprecated due to ambiguity — it actually returns the date created for the last order... Use `last_order_date_created` or `last_order_date_paid`"). Confirmed at runtime: every exported subscription emits a `PHP Deprecated: WC_Subscription::get_date ... last_payment` line to the debug log — one per row, so a 40k-row export writes 40k deprecation lines. Switch to `last_order_date_created` (or `last_order_date_paid`) to silence it and pin the semantics.

plugins/newspack-plugin/includes/export/class-csv-exports.php
```diff
@@ -0,0 +1,451 @@
...
+		);
+	}
+
+	/**
+	 * AJAX handler: process one page of the export.
+	 */
+	public static function ajax_export() {
+		\check_ajax_referer( self::AJAX_NONCE_ACTION, 'security' );
+
+		$type = isset( $_POST['export'] ) ? \sanitize_key( $_POST['export'] ) : '';
```
chickenn00dle: **Severity:** nit

`sanitize_key( $_POST['export'] )` here and `sanitize_key( $_GET['export'] )` at :380 skip `wp_unslash()`. Functionally safe (`sanitize_key()` strips backslashes too), but the house rule is `wp_unslash()`-before-sanitize, and the sibling reads at :376/:394 already do. Worth aligning.

plugins/newspack-plugin/includes/cli/class-export.php
```diff
@@ -0,0 +1,178 @@
...
+			// memory (the admin AJAX flow is immune — one page per request).
+			if ( function_exists( '\WP_CLI\Utils\wp_clear_object_cache' ) ) {
+				\WP_CLI\Utils\wp_clear_object_cache();
+			}
+		} while ( $percent < 100 );
+
+		if ( ! $exporter->save_to( $output ) ) {
+			WP_CLI::error( sprintf( 'Could not write to %s.', $output ) );
+		}
+		WP_CLI::success( sprintf( 'Exported %d rows to %s.', $exported, $output ) );
```
chickenn00dle: **Severity:** nit

When the no-progress guard breaks at :162, the loop still calls `save_to()` and prints `WP_CLI::success( 'Exported N rows...' )`, so a partial CSV ships under a success message. The preceding `WP_CLI::warning` covers it — minor — but consider making the final message reflect the incomplete state.

chickenn00dle (commented):
Reviewed in an isolated env with WooCommerce + Subscriptions. No blockers — this is a careful, defense-in-depth implementation, and I confirmed the security-critical behaviors at runtime: CSV row counts match the DB across all-statuses/active/role/product filters, formula-injection is neutralized (`=`, `+`, `-`, `@` prefixed with `'`; commas/quotes/newlines quoted), the empty dataset yields a clean headers-only file, and the capability gate holds (only an admin passes; editor/author/subscriber/customer are denied both exports). Cross-type filename guard and path-traversal rejection also hold; 29 unit tests pass and CI is green.

The comments below are all non-blocking. The one I'd most want a look at before merge is the classmap/autoload footgun — it's latent today but cheap to defuse. The rest are scale notes for very large sites plus a per-row deprecation notice I hit at runtime.

plugins/newspack-plugin/includes/export/class-csv-batch-exporter.php
adekbadek: Done. Added the three export files to `exclude-from-classmap` in composer.json so a stray autoload trigger (a `class_exists()` with autoload on, an `instanceof`, static analysis) can no longer require them before WooCommerce's `WC_CSV_Batch_Exporter` abstract is loaded and fatal at the `extends`. They remain reachable via the explicit `require_once` chain in `load_exporter_dependencies()`. Regenerated the classmap and confirmed the three exporter classes are gone from it while `CSV_Exports` (and the CLI class) stay.

plugins/newspack-plugin/includes/export/class-csv-exports.php
adekbadek: Done. Added `CSV_Batch_Exporter::stream_export()` and switched the admin download to it instead of WC's `export()`. It sends the same headers, clears any active output buffer, echoes the small headers row, then `fpassthru()`s the data temp file (streamed, not `file_get_contents()`'d whole), then unlinks the temp files and dies — so peak memory stays flat regardless of export size, matching the reason `save_to()` streams for the CLI path. Browser-verified end-to-end: a 459-row/117 KB users export downloads as `text/csv` (200), and a replay of the link returns 410 since the temp file is deleted on send.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
adekbadek: Done. `build_query_args()` now takes an optional per-run `$cache_key` (defaulting to '' so it stays pure for the direct callers and unit tests), and `prepare_data_to_export()` passes the run's filename. The product-filter set is resolved through a new `get_product_subscription_ids()` helper that memoizes `wcs_get_subscriptions_for_product()` in a short-lived transient keyed by that filename — so it runs once per run instead of once per page. Keyed by the (unique-per-run) filename rather than the product id, a later run always re-resolves against current data, so there's no cross-run staleness. The empty-set case is still cached safely and maps to the `[0]` no-match sentinel.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
adekbadek: Done. Switched `last_payment` to `last_order_date_created`, which is the exact behavioral equivalent per WCS's own deprecation note (`last_payment` returned the last order's created date), so exported values are unchanged. Updated the test fixture's date key to match. Verified against a live 258-row subscriptions export: `debug.log` now shows 0 `last_payment` (and 0 `last_order_date`) deprecation lines, down from one per row.

plugins/newspack-plugin/includes/export/class-csv-exports.php
adekbadek: Done. Both reads now `wp_unslash()` before `sanitize_key()` (`$_POST['export']` and `$_GET['export']`), matching the house rule and the sibling reads.

plugins/newspack-plugin/includes/cli/class-export.php
adekbadek: Done. The final message now branches on completion: only a full run (`$percent >= 100`) prints `WP_CLI::success( 'Exported N rows...' )`; a run that hit the no-progress guard prints `WP_CLI::warning( 'Export incomplete: wrote N rows... before stopping early.' )`, so a partial CSV no longer ships under a success message.

adekbadek:
@chickenn00dle Re-requesting review – the review comments have been addressed.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
```diff
@@ -0,0 +1,363 @@
...
+		$args['offset']   = ( $this->get_page() - 1 ) * $this->get_limit();
+		$args['paginate'] = true;
+		$args['return']   = 'ids';
+
+		// Do not "simplify" this to wcs_get_subscriptions( [ 'paged' => n ] ):
+		// that function accepts paged but never applies it (page 1 forever),
+		// and its product_id path runs an unpaged limit=-1 query.
+		$results = \wcs_get_orders_with_meta_query( $args );
+
+		$this->total_rows = (int) $results->total;
```
chickenn00dle: **Severity:** suggestion

**Suggestion:** Offset paging here (and in the users exporter) is insert-stable (ID-ascending, new rows land past the moving window) but **not delete-stable**, and `total_rows` is recomputed from the live query each page. If rows leave the filtered set mid-export — a hard delete, or more commonly a status change that drops a row out of a status-filtered view (a renewal/expiration scheduled action firing during a multi-minute run) — later rows slide back into an already-consumed offset and are silently skipped; a large enough shrink can also push the cumulative exported count past the recomputed total, ending the run early. The CSV downloads and presents as complete with no snapshot caveat, so for the revenue-recognition/accounting use case a publisher could reconcile on under-counted data.

**Consider:** the fix pattern already exists in this PR — the product-filter path memoizes its subscription ID set for the whole run in a transient. Snapshotting the target ID set once on step 1 for the unfiltered/status paths, or at minimum captioning the export as a point-in-time snapshot in the button help text, would close it.

plugins/newspack-plugin/includes/export/class-csv-exports.php
```diff
@@ -0,0 +1,453 @@
...
+	 */
+	public static function download_export_file() {
+		if ( ! isset( $_GET['action'] ) || self::DOWNLOAD_ACTION !== $_GET['action'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return;
+		}
+		if ( ! \wp_verify_nonce( \sanitize_key( \wp_unslash( $_GET['nonce'] ?? '' ) ), self::DOWNLOAD_NONCE_ACTION ) ) {
+			\wp_die( \esc_html__( 'Invalid download link.', 'newspack-plugin' ), '', 403 );
+		}
+
+		$type = isset( $_GET['export'] ) ? \sanitize_key( \wp_unslash( $_GET['export'] ) ) : '';
```
chickenn00dle, line 396: **Severity:** nit

**Nit:** The unit tests cover the security-relevant helpers (filename↔type binding via `validate_export_filename`, `save_to` streaming/failure, the cleanup sweep), but the actual `download_export_file()` and `ajax_export()` handlers — nonce failure, capability re-check, expired-file 410, cross-type filename rejection at the handler level — are only exercised transitively. Given this is a PII export with the auth logic living in those two handlers, a couple of handler-level tests (or a note that they were manually verified in the env) would close the gap.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
```diff
@@ -0,0 +1,363 @@
...
+		$transient = '' !== $cache_key
+			? 'newspack_export_product_subs_' . md5( $cache_key . ':' . $product_id )
+			: '';
+		if ( '' !== $transient ) {
+			$cached = \get_transient( $transient );
+			if ( is_array( $cached ) ) {
+				return $cached;
+			}
+		}
+		$ids = array_map( 'intval', array_keys( \wcs_get_subscriptions_for_product( $product_id ) ) );
```
chickenn00dle, line 230: **Severity:** nit

**Nit:** `get_product_subscription_ids()` caches the full subscription-ID set for a product in a transient. On a very large site a single popular product could push that array past the ~1MB object-cache entry ceiling. Edge case, and the same unbounded set is passed as `post__in` regardless (an inherent WCS-filter scaling trait, not net-new) — informational rather than actionable.

chickenn00dle (approved):
Re-review — no blockers. All prior-round comments (an earlier review pass plus Copilot) are addressed and I re-verified the load-bearing fixes are sound: `exclude-from-classmap` correctly keeps the exporter subclasses out of the autoloader until WC's abstract exists (they're `require_once`d on demand), `stream_export()` keeps download memory flat and fails rather than shipping a headers-only CSV, and the per-run product-ID transient resolves the expensive query once per run. CI is green and the PHP export tests pass (29).

The security model holds: per-step nonce + per-type capability re-check, an unguessable random filename bound to its capability-checked type, a hardened temp dir (deny-all `.htaccess` + empty index), delete-on-send with a 410 on replay, a daily cron sweep, and the users export gated behind `manage_woocommerce` so a `list_users`-only editor can't pull billing PII.

I checked the one runtime-verifiable concern (a possible `post__not_in` PII over-export in the non-group subscriptions filter under HPOS) against the running env and the WC/WCS source: it does **not** materialize — HPOS's orders-table query remaps `post__not_in` to `exclude` before running, and the legacy path reaches `WP_Query` which honors it natively, so a non-group export correctly excludes group subscriptions.

Aside from this, left a few non-blocking comments. In either case, this is an approval from me.

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
adekbadek: Fixed the half that silently truncates, and surfaced the rest.

`total_rows` is now pinned to what page 1 counted (`CSV_Batch_Exporter::pin_total_rows()`, one int in a per-run transient keyed by the export filename — the same per-run mechanism as the product-ID memoization). A set that shrinks mid-run can no longer push the cumulative exported count past a recomputed total and end the run early; the run keeps paging until a page comes back empty. Both surfaces then flag that: the AJAX handler returns an `Rows were removed or changed while the export was running, so this file may be incomplete…` notice that replaces the "Export complete" status, and the CLI's existing no-progress guard (now sharing the same `page_was_empty()` definition) warns as before. The button help text also names the export a point-in-time snapshot.

I did not snapshot the whole target ID set for the unfiltered/status paths: that generalizes the unbounded-array trait you flagged on `get_product_subscription_ids()` (thread on line 225) from "a popular product's subscriptions" to "every subscription/user on the site", and it would replay that set as `post__in` on every page — worst exactly on the large sites where a multi-minute run makes the race likely. So rows that slide back into an already-consumed offset can still be missed; that residual is documented on `pin_total_rows()`, and it is now visible to the publisher rather than presenting as a complete CSV. Happy to go further if you'd rather take the snapshot cost.

Covered by `test_total_rows_pinned_across_a_shrinking_run` (asserts the pre-fix behavior would have reported 100% over half the rows).

plugins/newspack-plugin/includes/export/class-csv-exports.php
adekbadek: Added `tests/unit-tests/export/class-csv-exports-handlers.php` — 8 tests driving `download_export_file()` and `ajax_export()` directly (wp_die intercepted via the `wp_die_handler`/`wp_die_ajax_handler` filters, so the bail-out status and message are assertable):

- download: invalid nonce → 403; capability revoked after the link was minted → 403 (the re-check, not just the nonce); cross-type filename (`users` filename through the `subscriptions` path) → 403; replayed link with the file already sent → 410; `users` export refused for a fully-capable admin while WooCommerce is inactive.
- ajax: invalid nonce → `check_ajax_referer` refuses with 403; valid nonce + insufficient capability → JSON error; a step-2 request carrying another type's filename is not adopted (the run restarts under a filename of its own type).

plugins/newspack-plugin/includes/export/class-subscriptions-csv-exporter.php
adekbadek: Agreed it's inherent rather than net-new, so no behavior change — but I documented the degradation on `get_product_subscription_ids()` so a future reader doesn't have to re-derive it: if the array outgrows the object cache's per-entry ceiling, `set_transient()` simply doesn't stick, `get_transient()` returns `false`, and every page re-runs the query — i.e. exactly the pre-cache behavior, no correctness cliff. The same unbounded set goes to `post__in` either way.

This is also why I did not generalize the ID-set snapshot to the unfiltered/status paths for the offset-paging thread — see the reply there.

adekbadek:
@chickenn00dle Re-requesting review – commits landed after your approval that need a look before this merges:
- fix(export): pin CSV export totals so a shrinking set can't truncate
- test(export): align product-subscription mock with the shared fixture

plugins/newspack-plugin/includes/export/class-csv-batch-exporter.php
```diff
@@ -0,0 +1,258 @@
...
+	 *
+	 * The pin is a single int in a transient keyed by the export filename,
+	 * which is unique per run, so a later run always recounts.
+	 *
+	 * @param int $live_total Total rows the current page's query reported.
+	 */
+	protected function pin_total_rows( int $live_total ): void {
+		$transient = 'newspack_export_total_' . md5( $this->get_filename() );
+		if ( 1 === $this->get_page() ) {
+			\set_transient( $transient, $live_total, HOUR_IN_SECONDS );
```
chickenn00dle: **Suggestion:** One-hour pin TTL is short relative to the runs it protects — The pin is stored with `HOUR_IN_SECONDS`, but the rest of the subsystem assumes a run can live much longer — `CSV_Exports::cleanup_stale_files()` only sweeps export files older than `DAY_IN_SECONDS`. The AJAX path never calls `set_limit()`, so it runs at WooCommerce's default of 50 rows per request: a 100k-row subscriptions export is 2,000 round trips, each paging `wcs_get_orders_with_meta_query()` and hydrating 50 subscriptions, which will exceed an hour on a large site. A long CLI run does the same.

Only page 1 ever writes the pin, so once it expires the `false === $pinned` branch falls through to `$live_total` for every remaining page and the run silently reverts to exactly the pre-fix behavior — with no signal on either surface. The same applies if a persistent object cache evicts the entry. Either refresh the TTL on read (re-`set_transient()` in the `false !== $pinned` branch, making it a sliding window tied to activity) or just use `DAY_IN_SECONDS` to match the window the export files themselves get.

plugins/newspack-plugin/includes/export/class-csv-exports.php
```diff
@@ -0,0 +1,464 @@
...
+			if ( $exporter->get_total_exported() > 0 && ( ! file_exists( $file_path ) || 0 === filesize( $file_path ) ) ) {
+				\wp_send_json_error(
+					[ 'message' => __( 'The export file could not be written. Please check uploads directory permissions.', 'newspack-plugin' ) ]
+				);
+			}
+			\wp_send_json_success(
+				[
+					'step'       => 'done',
+					'percentage' => 100,
+					'notice'     => $percent < 100
```
chickenn00dle: **Suggestion:** The incomplete-export notice can't fire when fewer than `limit` rows leave the set — The notice is gated on `$percent < 100`, but the run usually reaches the terminal empty page with `$percent` already at exactly 100, so it stays silent on much of the shrinkage it was written to catch.

WooCommerce derives `get_total_exported()` as `(page - 1) * limit + rows_written_this_page` — an arithmetic that assumes every prior page was full. Take a pinned total of 100 at the default limit of 50, and suppose ten already-exported rows leave the set after page 1: page 2 (offset 50) returns 40 rows, so `total_exported` is 90 and `$percent` is 90; page 3 (offset 100) returns nothing, and `total_exported` becomes `(3-1)*50 + 0 = 100`, so `$percent` is 100. `page_was_empty()` is true and the handler takes the done branch, but `$percent < 100` is false — `notice` is `''`. The publisher gets 90 rows, ten rows that still exist were never exported, and the UI says the export completed. The notice only fires when the shrink is at least a full page (`k >= limit`, or `k >= N mod limit`), so with the default limit of 50 you need roughly 25–50 rows to vanish mid-run before anyone is warned.

Worth being precise about what's new here: the *miss itself* is pre-existing — I simulated the pre-delta live-recount behavior and it loses exactly the same rows — and you already document rows sliding back into a consumed offset as an accepted limitation in `pin_total_rows()`'s docblock. So this is not a regression, and the delta is strictly better. But the docblock says the run "keeps paging until a page comes back empty ... which is where both surfaces flag the result as a possibly-incomplete snapshot", and that second half isn't true in the partial-page case, which is the likelier one.

Reaching an empty page on page 2+ *always* implies incompleteness — a clean run terminates at `total_exported == $total_rows` on a non-empty page, short final page included. So gating both surfaces on `$exporter->page_was_empty() && $exporter->get_page() > 1` covers the whole class; the page guard keeps a legitimately empty result set (where `get_percent_complete()` returns 100 for `total_rows === 0`) from being flagged. The CLI guard at `includes/cli/class-export.php:162` has the same `$percent < 100 &&` gate and one extra symptom — with `$percent` at 100 the `do/while` at `:173` exits normally and `:181` prints `WP_CLI::success( 'Exported 100 rows...' )` for a file holding 90, since `$exported` is read from the terminal empty page.

plugins/newspack-plugin/tests/unit-tests/export/class-subscriptions-csv-exporter.php
```diff
@@ -0,0 +1,551 @@
...
+
+	/**
+	 * A result set that shrinks mid-run (rows deleted, or a status change
+	 * dropping them out of a status-filtered view) must not cut the export
+	 * short: the total stays pinned to what page 1 counted, so the run ends on
+	 * an empty page — which both the AJAX handler and the CLI report as a
+	 * possibly-incomplete snapshot — instead of reporting a truthful-looking
+	 * 100% over a half-written CSV.
+	 */
+	public function test_total_rows_pinned_across_a_shrinking_run() {
```
chickenn00dle, line 523: **Suggestion:** The pinned-total test picks the one shrink magnitude that works — `test_total_rows_pinned_across_a_shrinking_run` removes subscriptions 3 and 4 — a shrink of 2 against `limit = 2`, i.e. exactly one full page. For a 4-row fixture at limit 2 that is the *only* magnitude that produces the notice; a shrink of 1 lands in the gap described above and would pass a 100%-complete run over a short file. So the single test guarding this mechanism exercises the convenient case and never the partial-page one.

It also asserts only on the exporter (`page_was_empty()`, `get_percent_complete()`) and never on what either consumer does with them, and the new handler test file — thorough as it is on authorization — has no test that drives `ajax_export()` through a shrinking multi-page run to assert the `notice` key. A case that removes one *already-exported* row and then steps pages 2 and 3, asserting a non-empty `notice` in the handler payload, would fail today and pass under the gate change above.

plugins/newspack-plugin/tests/unit-tests/export/class-csv-exports-handlers.php
```diff
@@ -0,0 +1,286 @@
...
+	 * HTTP status passed to the last intercepted wp_die().
+	 *
+	 * @var int
+	 */
+	private $die_status = 0;
+
+	/**
+	 * Intercept wp_die() so the handlers' bail-out paths are assertable.
+	 */
+	public function set_up() {
```
chickenn00dle, line 40: **Nit:** Handler tests don't reset the WCS mock globals they depend on — `test_ajax_export_does_not_adopt_a_cross_type_filename()` relies on "No subscriptions exist", but the class never resets the `$subscriptions_database` global that the `wcs_get_orders_with_meta_query()` mock pages over, and other suites that sort earlier (`tests/unit-tests/content-gate/`) set it in `set_up()` without clearing it in `tear_down()`. It passes today only because the last such class happens to leave it empty. If leftovers were ever present the failure would be doubly confusing: `get_total_exported()` would be greater than 0 while the mock exporter writes no file, so `ajax_export()` would take the "file could not be written" branch rather than `done`. Adding `global $subscriptions_database; $subscriptions_database = [];` to `set_up()` — matching what the sibling exporter test does at line 28 — makes it order-independent; `$wcs_mock_orders_with_meta_query_result` and `$wcs_mock_hpos_enabled` deserve the same.

plugins/newspack-plugin/tests/unit-tests/export/class-csv-exports-handlers.php
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	/**
+	 * The users export requires WooCommerce to be active (its CSV carries WC
+	 * billing PII), so an administrator with every capability is still refused
+	 * while WooCommerce is absent.
+	 */
+	public function test_download_users_export_requires_woocommerce() {
+		$this->login_as_export_capable_admin();
+		$this->assertFalse( class_exists( 'WooCommerce' ), 'This test asserts the WooCommerce-inactive behavior.' );
```
chickenn00dle: **Nit:** WooCommerce-absent precondition depends on undeclared file ordering — `test_download_users_export_requires_woocommerce` asserts `class_exists( 'WooCommerce' )` is false as its precondition, but `tests/unit-tests/my-account.php:496` does `eval( 'class WooCommerce {}' )`, which is irreversible for the rest of the PHPUnit process. This passes only because `unit-tests/export/` sorts ahead of `unit-tests/my-account.php` — load-bearing ordering that nothing declares. The assertion fails loudly rather than passing vacuously, which is the right call, but skipping the test when the shim is already defined would remove the dependency.

Relatedly, the `define( 'WC_ABSPATH', '/woocommerce/' )` at line 45 is process-wide and irreversible, so every later test class sees WooCommerce's path constant defined. Nothing else reads it today, but `defined( 'WC_ABSPATH' )` is a common "is WooCommerce active" proxy — and this class is otherwise careful about exactly that kind of assumption. Defining it in `tests/mocks/wc-mocks.php` alongside the other WC stand-ins would make it explicit and shared.

chickenn00dle (approved):
Re-reviewed the two commits pushed since my approval. Scoped the diff to exactly your delta (head with `799d4e85` + `0ee804d3` reverted as the base) so the `main` merge in between didn't muddy it.

The pin is a real improvement and I verified it never behaves worse than the code it replaces: I reimplemented WooCommerce's `get_total_exported()` / `get_percent_complete()` semantics and swept shrink scenarios tracking exported row *identities*, and in every case the pinned version either matches or beats the live-recount version. Not blocking on any of this.

One thing worth a look before merge, in the finding below: the incomplete-export notice is gated on `$percent < 100`, which means it fires for a shrink of a full page or more and stays silent for anything smaller — including the single-subscription-expires case your own docblock names as the motivating one. The underlying miss is pre-existing and you already document it as accepted; what's new is that the thing added to *warn* about it doesn't fire in the likelier half of the cases. Cheap to close.

The rest are follow-ups you can take or leave.

- `plugins/newspack-plugin/includes/export/class-csv-exports.php:158` — **Nit:** `get_exporter()` return type is narrower than what callers use — `get_exporter()` is declared `?\WC_CSV_Batch_Exporter`, but callers now invoke four methods that only exist on the Newspack subclass — `page_was_empty()` (line 339 here and `includes/cli/class-export.php:162`), plus the pre-existing `set_list_params()`, `get_export_file_path()` and `save_to()`. Narrowing the declared type to `?CSV_Batch_Exporter` is safe — the parent is only resolved on use, so it doesn't force an early load ahead of WooCommerce — and makes the contract match reality.
- `plugins/newspack-plugin/includes/export/class-csv-exports.php:349` — **Nit:** The pin transient is never deleted, including where the run is known to be over — Nothing ever deletes the pin. Core's `delete_expired_transients()` (on the daily `wp_scheduled_delete` event) eventually reaps the option pair on sites without a persistent object cache, so this is bounded rather than a leak — but the class already owns a daily sweep for the analogous file artifacts, and a `delete_transient()` here in the `done` branch plus after `save_to()` in the CLI costs nothing and drops the reliance on core cron actually firing. The TTL still covers the abandoned-run case.

plugins/newspack-plugin/includes/export/class-csv-batch-exporter.php
adekbadek: Went with both halves of the suggestion in e24dd08: the pin is now written with `DAY_IN_SECONDS`, matching the window `cleanup_stale_files()` gives the export files, and every page re-sets it, so the window slides with activity rather than expiring under a run that outlives it. A page that finds the pin gone still falls back to the live count — nothing can be done about an object-cache eviction — but that now takes a full day of no progress instead of an hour of steady progress.

The other half of the lifetime: the pin was never deleted, so it sat for the full TTL after the run finished. Both surfaces now call `clear_pinned_total()` once the run is over.

`test_pinned_total_expiry_slides_with_each_page` winds the expiry back to a minute out and asserts the next page pushes it back.

plugins/newspack-plugin/includes/export/class-csv-exports.php
adekbadek: Fixed in e24dd08 — the analysis is right, and the arithmetic walkthrough made the fix obvious. Both surfaces now gate on `$exporter->ended_short()`, a new method on the base exporter that is exactly the `get_page() > 1 && page_was_empty()` condition you describe, so the whole class is covered rather than only shrinkage of a full page or more.

The CLI had the extra symptom too: the loop now breaks on `ended_short()` and carries that as a flag into the final message, instead of re-deriving completeness from `$percent`. `$exported` is also read off the last page that actually wrote rows, since on the terminal empty page the counter reports the pinned total rather than what the file holds — so "wrote 90 rows" no longer says 100.

The docblock that claimed both surfaces flag the empty-page case is now true, and `ended_short()` carries the reason the percentage can't be the gate.

plugins/newspack-plugin/tests/unit-tests/export/class-subscriptions-csv-exporter.php
adekbadek: Both tests added in e24dd08.

`test_a_sub_page_shrink_still_ends_short` removes one already-exported row from the 4-row fixture at limit 2: page 2 writes one row (75%), page 3 comes back empty and the counter climbs back to exactly 100 over a three-row file. It asserts that 100 explicitly, so the gap is documented rather than implied, then asserts `ended_short()`.

`test_ajax_export_flags_a_run_that_ended_short` drives `ajax_export()` through the same shrink over three steps and asserts a non-empty `notice` in the payload (plus that the run's pinned total is gone afterwards). It fails on the old `$percent < 100` gate with "Failed asserting that a string is not empty" and passes on the new one — I checked both ways round.

That test needed the batch-exporter mock to actually write the data file, since the file is the handler's only evidence a page was written; it now appends per page and unlinks on page 1, like the real class, and both export suites clear the staged files in `tear_down()`.

plugins/newspack-plugin/tests/unit-tests/export/class-csv-exports-handlers.php
adekbadek: Done in e24dd08 — `set_up()` now resets `$subscriptions_database` and `$wcs_mock_hpos_enabled` and unsets `$wcs_mock_orders_with_meta_query_result`, matching the sibling exporter suite, so the result set an export sees here no longer depends on which class ran first. The new shrinking-run test seeds that global deliberately, which would have made the leak much easier to hit.

plugins/newspack-plugin/tests/unit-tests/export/class-csv-exports-handlers.php
adekbadek: Both addressed in e24dd08. `WC_ABSPATH` is now defined in `tests/mocks/wc-mocks.php` alongside the other WC stand-ins, with a note that code treats it as an "is WooCommerce loaded" proxy — so it is shared and explicit rather than a side effect of this class's `set_up()`. And `test_download_users_export_requires_woocommerce` now skips when a `WooCommerce` shim is already defined, naming `tests/unit-tests/my-account.php` as the source, so it no longer depends on undeclared file ordering. It still runs (unskipped) in a full-suite run.

adekbadek:
On the `get_exporter()` return type: narrowed to `?CSV_Batch_Exporter` in e24dd08. The callers do rely on the subclass surface — `set_list_params()`, `get_export_file_path()`, `stream_export()`, `save_to()`, and now `ended_short()` and `clear_pinned_total()` — so the declared type was understating the contract.

---

On the pin transient never being deleted: both surfaces now drop it where the run is known to be over — `clear_pinned_total()` in `ajax_export()`'s `done` branch, and after `save_to()` in the CLI (before the write-failure bail, so a failed `--output` doesn't leave it behind either). The pin's own expiry is now a sliding day rather than a fixed hour, which made deleting it at the end matter more, not less.

adekbadek:
@chickenn00dle Re-requesting review – commits landed after your approval that need a look before this merges:
- fix(export): flag every short export run and keep the pinned total alive

chickenn00dle (approved):

PR #501, opened 2026-07-02
test(blocks): load Tag_Labels stub in homepage-posts test

jason10lee:
### All Submissions:

Fast follow to #500 (resolves NPPM-2973). When the automated `release → main`
back-merge was reconciled manually in #500, the combined
`test-homepage-posts-block.php` was based on the release-side (#471) version,
which predated the tag-labels work — so `main`'s top-of-file
`require_once __DIR__ . '/class-newspack-tag-labels-stub.php';` was dropped.

Without it the `Newspack\Tag_Labels` stub never loads, so the two tag-labels
contract tests (`test_tag_labels_rest_field_shape`,
`test_tag_labels_rest_field_empty_returns_false`) **silently skip**. This
restores the one-line `require_once` so they run again.

1. `n test-php --filter HomepagePostsBlockTest` — 9 tests pass with **0 skipped**
   (previously 2 skipped). Verified locally: 9 tests / 19 assertions (was 11 with
   the tag-labels tests skipped).

### Other information:


chickenn00dle (approved):
Release risk: LOW — test-only diff, no blockers, no contract surface, and no production code path touched (`.distignore` excludes `/tests`).

The change is a faithful, minimal restore. I traced the history and it matches the description exactly: the stub and its `require_once` landed together in `b5ef988bc` (#2282), and merge commit `164fce8c3` — the manual `release → main` reconciliation in #500 — rebuilt this file from the release-side version that predated the tag-labels work, dropping the require while leaving the stub file on disk. The header region is now byte-identical to `b5ef988bc`, so this is a clean revert rather than a new design decision.

The mechanism is right, too. A per-file `require_once` for a stub used by exactly one test file matches the prevailing pattern here (`test-checkout-button-block.php:8`, `newspack-popups/tests/test-segmentation-newsletter-link.php:11`), and `bootstrap.php` is reserved for things every test in the suite needs — hoisting it there would define `\Newspack\Tag_Labels` for the whole suite for no benefit.

I checked two hazards and both are clear. PHPUnit runs `bootstrap.php` to completion before it parses test files, so the stub's `if ( ! class_exists( ... ) )` guard cannot lose a race or shadow a real class that loads later. And because `$stub_labels` defaults to `null`, `Newspack_Blocks::get_tag_labels()` returns exactly what it would return with the class absent, and the stub defines no `display()` method — so no other test in the shared PHPUnit process changes behavior.

One thing worth confirming before merge: that the two restored tests actually report as running rather than skipped. If they still skip, the fix would look successful without being so. I left one non-blocking note about the skip guard itself, which is the reason this regression reached a green CI in the first place — worth a follow-up, but out of scope here.

PR #507, opened 2026-07-02
fix(listings): authorize self-serve premium listing create/delete

dkoo:
### All Submissions:

Security hotfix for a broken-access-control issue in the self-serve listings feature (`Products_Ui`). Its front-end create and delete handlers were gated only by a static-action nonce. A nonce confirms a request came from the user's own session, but it is **not** an authorization check — so any logged-in low-privilege user (e.g. a listings subscriber) could:

- **Delete**: trash arbitrary posts, pages, CPTs, products, or orders by id — the delete handler trashed whatever post id it was handed.
- **Create**: mint premium listings past their paid quota, attributed to another account — the create handler trusted attacker-supplied `customer_id`/`subscription_id`.

This PR adds per-object authorization:

- **Delete** — only the owner of a self-serve premium listing may trash it: the target must be an Event/Marketplace listing created via a subscription and owned by the current user.
- **Create** — the subscription must belong to the current user, be active and premium, and have included listings remaining; the new listing is attributed to the current user rather than an attacker-supplied id.

Notes:
- A raw `current_user_can( 'delete_post', … )` gate is intentionally **not** used: these listing CPTs register with the default `post` capability type, and subscribers/customers lack the `delete_*` capabilities, so that check would break legitimate self-serve deletion of a user's own listings. Ownership is the correct gate.
- Hardening only — no change to the legitimate self-serve create/delete UX.

Closes NPPM-2965.

1. On a site with WooCommerce, WooCommerce Subscriptions, and newspack-listings active, define `NEWSPACK_LISTINGS_SELF_SERVE_ENABLED` and create the self-serve products from Listings settings.
2. As a premium-listing subscriber, open the WooCommerce **View subscription** page and confirm the legitimate flows are unchanged: you can still create (up to quota) and delete your **own** Event/Marketplace listings.
3. Copy the delete nonce from one of your own Delete links, then send `GET /?customer_id=<you>&subscription_id=<yours>&delete=<a post id you do NOT own>&redirect_uri=%2Fwp-admin%2F&_wpnonce=<nonce>`. Confirm the target post is **not** trashed (previously it was).
4. With the same nonce, try to delete **another customer's** listing id — confirm it is **not** trashed.
5. Try to create a listing against a `subscription_id` you don't own, or past your included quota — confirm no listing is created and no listing is attributed to another account.
6. Run the automated tests: from `plugins/newspack-listings`, `n test-php` — includes the new `ProductsUiTest` (8 cases covering delete authorization). Full listings suite: 12 tests / 48 assertions passing.

### Other information:

### DIFF
--- a/plugins/newspack-listings/includes/products/class-products-ui.php
+++ b/plugins/newspack-listings/includes/products/class-products-ui.php
@@ -18,6 +18,27 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Products_Ui extends Products {
 	/**
+	 * Number of included listings granted to premium subscribers.
+	 *
+	 * @var int
+	 */
+	private $total_included_listings;
+
+	/**
+	 * Nonce action string for creating a listing.
+	 *
+	 * @var string
+	 */
+	private $create_nonce;
+
+	/**
+	 * Nonce action string for deleting a listing.
+	 *
+	 * @var string
+	 */
+	private $delete_nonce;
+
+	/**
 	 * Constructor.
 	 */
 	public function __construct() {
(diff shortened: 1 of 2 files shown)


plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -561,10 +690,15 @@ public function create_or_delete_premium_listing() {
...
 
-		// Get order info and remaining premium listings.
-		$subscription = new \WC_Subscription( $subscription_id );
-		$args         = [
-			'post_author' => $customer_id,
+		// Verify the current user owns the subscription and has quota remaining
+		// before creating a listing; the nonce alone is not authorization (NPPM-2965).
+		if ( ! $this->user_can_create_premium_listing( $subscription_id ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
dkoo: Fixed in 06ea791c. `$subscription_id` is now cast to `int` once, immediately after create-nonce verification, so the ownership/quota checks and the stored `listing_subscription` meta all use the same normalized value. A value like `123abc` now stores `123` and is correctly counted toward the quota.

plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -524,6 +545,108 @@ public function listing_subscription_manage_premium_listings( $subscription ) {
...
+	 *
+	 * @param int      $subscription_id Subscription ID to create the listing against.
+	 * @param int|null $user_id         User ID to check. Defaults to the current user.
+	 * @return bool True if the user may create a listing.
+	 */
+	public function user_can_create_premium_listing( $subscription_id, $user_id = null ) {
+		if ( null === $user_id ) {
+			$user_id = get_current_user_id();
+		}
+		$user_id         = (int) $user_id;
```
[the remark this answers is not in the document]
dkoo: Added in 06ea791c. The quota logic is now extracted into a WC-free `get_remaining_included_listings()` helper with unit coverage for the **quota-exhausted** path (full allowance when unused, decrement per listing, floor at zero, ignores other subscriptions), plus a fail-closed test asserting create authorization is denied when WooCommerce Subscriptions is unavailable. The remaining create branches (subscription ownership + active/premium status) call `wcs_get_subscription()`/`WC_Subscription`, which the newspack-listings PHPUnit harness can't load (no WooCommerce in the test bootstrap). Those were instead verified end-to-end in an isolated env with WooCommerce + Subscriptions, driving the real handler over authenticated HTTP requests.

plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -561,10 +708,19 @@ public function create_or_delete_premium_listing() {
...
-		$subscription = new \WC_Subscription( $subscription_id );
-		$args         = [
-			'post_author' => $customer_id,
+		// Normalize the subscription id once so the ownership/quota checks and the
+		// stored listing meta all use the same value (NPPM-2965).
+		$subscription_id = (int) $subscription_id;
+
+		// Verify the current user owns the subscription and has quota remaining
+		// before creating a listing; the nonce alone is not authorization (NPPM-2965).
+		if ( ! $this->user_can_create_premium_listing( $subscription_id ) ) {
```
jason10lee, line 717: **Suggestion:** The quota check is check-then-insert with no lock: `user_can_create_premium_listing()` reads the current count here and the insert happens further down, with nothing serializing them. Two near-simultaneous create requests that both observe `remaining == 1` will both pass and produce 11 listings against a 10-slot subscription. Impact is bounded (a small one-time overage per burst) and self-inflicted — the user is racing their own quota — so it's acceptable for a hotfix; a hard cap would need a lock or insert-then-recount, which is out of scope here. Noting for awareness.

plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -524,6 +545,126 @@ public function listing_subscription_manage_premium_listings( $subscription ) {
...
+			return false;
+		}
+
+		// The subscription must belong to the requesting user.
+		if ( (int) $subscription->get_user_id() !== $user_id ) {
+			return false;
+		}
+
+		// Must be an active premium subscription.
+		$is_premium = 'active' === $subscription->get_status()
```
jason10lee, line 626: **Suggestion:** `get_status() === 'active'` excludes `pending-cancel` (cancelled but still within the paid period), which WooCommerce Subscriptions treats as still-entitled. This is **not a regression** — the existing create-button UI already gates on `'active' === get_status()` (line 428), so a `pending-cancel` subscriber never sees the button, and this server check matches that definition. Flagging only so it's a conscious product decision: if `pending-cancel` subscribers should still be able to create, this check and the UI gate should change together.

plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -524,6 +545,126 @@ public function listing_subscription_manage_premium_listings( $subscription ) {
...
+		$subscription_id = (int) $subscription_id;
+		if ( ! $subscription_id ) {
+			return 0;
+		}
+
+		$existing_listings = get_posts(
+			[
+				'meta_key'       => self::POST_META_KEYS['listing_subscription'],
+				'meta_value'     => $subscription_id, // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+				'post_status'    => [ 'draft', 'future', 'pending', 'private', 'publish' ],
```
jason10lee, line 655: **Nit:** `get_remaining_included_listings()` counts only `draft/future/pending/private/publish`, so a trashed listing no longer consumes quota — "10 included listings" is effectively "10 concurrent non-trashed listings," not a lifetime cap. This is consistent with the pre-existing display logic (same status set), so it's almost certainly the intended "delete to free a slot" semantic; concurrent listings stay capped and there's no cross-user impact. Worth a one-line confirmation that unlimited lifetime churn is acceptable.

plugins/newspack-listings/includes/products/class-products-ui.php
```diff
@@ -561,10 +708,19 @@ public function create_or_delete_premium_listing() {
...
+
+		// Verify the current user owns the subscription and has quota remaining
+		// before creating a listing; the nonce alone is not authorization (NPPM-2965).
+		if ( ! $this->user_can_create_premium_listing( $subscription_id ) ) {
+			return;
+		}
+
+		// Attribute the new listing to the current user, not an attacker-supplied id.
+		$args = [
+			'post_author' => get_current_user_id(),
```
jason10lee, line 723: **Nit:** This line replaced `'post_author' => $customer_id`, so `customer_id` is no longer used for anything after the fix (delete ignores it too). Yet the args guard at the top of the handler still bails when `customer_id` is empty. It's harmless because the legitimate UI always includes it, but a future reader could think it still carries authorization weight — consider dropping it from the required-args check, or adding a short comment that it's retained only for URL/backward compatibility.

plugins/newspack-listings/tests/test-products-ui.php
```diff
@@ -0,0 +1,233 @@
...
+	 */
+	public function test_remaining_included_listings_zero_for_invalid_subscription() {
+		$this->assertSame( 0, $this->products_ui->get_remaining_included_listings( 0 ) );
+	}
+
+	/**
+	 * Create authorization fails closed when WooCommerce Subscriptions is
+	 * unavailable (as in this test harness) — no subscription can be verified.
+	 */
+	public function test_cannot_create_premium_listing_without_woocommerce() {
```
jason10lee, line 229: **Nit:** Because WooCommerce Subscriptions is absent in the listings test harness, the only create-path test asserts the fail-closed case (returns false when `wcs_get_subscription` is unavailable). The security-relevant rejection branches — subscription-owned-by-another-user and non-premium — rest on code inspection only (the quota-exhausted path *is* covered via `get_remaining_included_listings`). This is a harness limitation, not an omission; a fast follow could stub `wcs_get_subscription` to lock those two branches down.

jason10lee (approved):
Thanks, and approved!  PHPUnit passes (18 tests / 55 assertions) and CI is green. No blockers:  just a couple suggestions from agentic review we'd consider low severity, and a few nits.

dkoo:
Thanks, @jason10lee! Since this is a security-oriented hotfix, I'm going to merge and defer the suggestions for now.

PR #512, opened 2026-07-02
refactor(plugin): replace tachyons utility classes with stacks

thomasguillot:
### All Submissions:

`newspack-plugin` bundled the entire Tachyons utility library into `dist/admin.css` (74 KB raw, ~95% of the whole file), shipped on every wp-admin page, to power roughly 30 class usages. This PR replaces all of that usage with layout components and drops the dependency, taking **`admin.css` from 77.6 KB to 6.4 KB**.

- All 27 classic `Card buttonsCard` call sites (newspack-plugin, newspack-ads, and packages/components itself) are replaced with `HStack`, and the `buttonsCard` class mapping and SCSS are removed from the classic Card. Modal action footers share a `.newspack-modal__footer` class that restores the old vertical spacing. The `__experimentalCoreCard` path (used by the audience segment-group cards) is untouched.
- Raw layout divs (`flex justify-between items-end` and friends) become `HStack`/`VStack` with `spacing` props; margin-only utilities (`ma0`, `mt2`, `mr2`...) become two semantic notice modifiers and a few scoped SCSS rules.
- The Tachyons import and the dead `table.fixed` override are removed from `src/admin/style.scss`, the dependency is dropped from `package.json` and the lockfile, and the inert `ma0`/`mt0` classes are removed from the WooCommerce my-account templates (confirmed dead: `admin.css` never loads on the frontend).
- Tachyons also shipped a global `* { box-sizing: border-box }` reset. That reset was only re-applied to modern wizards (scoped to `.newspack-wizard__content:has(> .newspack-wizard__sections)`), so legacy `withWizard` screens (components demo, Setup, Newsletters, Advertising) fell back to `content-box` and rendered ~48px too wide. The reset is now applied to all `.newspack-wizard`/`.newspack-wizard__content` UI, restoring the correct widths.
- The webhook endpoint modal is now `size="full"`, giving the two-column Actions grid room to lay out instead of overflowing the medium modal.
- A pre-refactor inventory of every touched call site, grouped by wizard, is documented in [a comment on the Linear issue](https://linear.app/a8c/issue/DSGNEWS-190/replace-tachyons-utility-classes-with-stacks-and-drop-the-dependency) for future rework reference.

**Follow-up cleanup (from code review):** the first pass missed a few Tachyons class usages outside the main wizard surface; these are now converted too, so the dependency is fully gone (a workspace-wide grep for Tachyons class names returns zero):

- `mv0` on the Header Bidding GAM orders list → a `VStack spacing={ 0 }` connected list in `newspack-ads`, with the connected-card border/overlap rules re-homed into the ads SCSS (previously they relied on Tachyons' `.mv0` margin reset leaking in from `newspack-plugin`'s `admin.css`).
- `relative` on `GroupedSelectControl` → a scoped `newspack-grouped-select-control__field` positioning wrapper (without it the dropdown chevron re-anchored to the whole labeled control and drifted up).
- `o-50` / `gray` in the Setup and plugin-card wizards → namespaced `.newspack-is-dimmed` / `.newspack-text-muted` (the latter backed by `$gray-700`).
- Removed the orphaned `campaigns/analytics` view (dead code — a `withWizardScreen` "Coming soon" placeholder imported nowhere since the Audience IA rework).
- The border-box reset now also covers the wizards admin header (`#newspack-wizards-admin-header`), which is a sibling of the `.newspack-wizard` subtree and so fell outside the earlier reset.

Notes for reviewers:

- The `pnpm-lock.yaml` diff includes ~80 lines of `tarball:` metadata on unrelated entries; that is what the pinned pnpm 10.33 now writes when touching the lockfile, not version changes.

Closes DSGNEWS-190.

1. Run `pnpm install` and `pnpm build`, then confirm `plugins/newspack-plugin/dist/admin.css` is ~6 KB and contains no Tachyons rules (`grep -c "\.ma0" dist/admin.css` returns 0).
2. Settings > Connections: open the Mailchimp API key modal (footer buttons right-aligned with spacing above), check the GA4 custom events Save/Reset button pair, and the Webhook Endpoints section header (title left, button right, spacing below).
3. Webhooks: add/edit an endpoint and confirm the modal opens full-width, the "Send a test request" row and the Save footer are right-aligned, and the Actions checkboxes lay out in two columns.
4. Advertising > Ad Units (GAM connected or legacy mode): the "Add New Ad Unit" and "Save" rows are right-aligned; in legacy mode the Network Code field and Save button sit side by side, bottom-aligned.
5. Setup wizard: the "setup already completed" notice sits flush at the top; on the completed screen the four cards keep equal height with buttons pinned bottom-right.
6. Audience > Campaigns: open the Add New Campaign and rename modals, duplicate a prompt, and open the prompt settings modal; all footers are right-aligned. On the Audience setup complete screen the "Publish campaign" button is centered.
7. Newsletters settings: while data loads, the spinner appears centered.
8. Components Demo (`admin.php?page=newspack-components-demo`): confirm the cards match `main` width (not wider), and button rows, Waiting rows (spinner labels visible), Handoff buttons, the demo modal footer, and the icon gallery tiles all render as before.
9. Settings > Advanced Settings: select non-default featured image or template values for all posts and confirm the warning notices keep tight spacing.
10. WooCommerce My Account (frontend): edit account screens are visually unchanged (removed classes were dead code).
11. Settings > Connections: open a **grouped** SelectControl (e.g. a typography selector) and confirm the dropdown chevron stays vertically centered on the select, including when a label/help text is shown.
12. Confirm a workspace-wide grep for bare Tachyons classNames (`mv0`, `o-50`, `relative`, `gray`, …) in `src` returns nothing.

### Other information:


plugins/newspack-plugin/src/wizards/newspack/views/settings/connections/webhooks/modals/upsert.tsx
```diff
@@ -121,7 +121,7 @@ const Upsert = ( {
 					<Notice isError noticeText={ __( 'Request Error: ', 'newspack-plugin' ) + editing.disabled_error } />
 				) }
 				{ testResponse.success && <Notice isSuccess noticeText={ `${ testResponse.message }: ${ testResponse.code }` } /> }
-				<Grid columns={ 1 } gutter={ 16 } className="mt0">
+				<Grid columns={ 1 } gutter={ 16 }>
```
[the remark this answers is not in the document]
thomasguillot: Good catch. Set `noMargin` on the Grid so it no longer picks up the base 32px top margin; the bottom spacing is preserved by the following Divider's own top margin.

plugins/newspack-plugin/src/admin/style.scss
```diff
@@ -262,10 +261,13 @@ h1 {
...
-	position: static;
-	table-layout: fixed;
+// Notice spacing inside wizard layouts.
+.newspack-notice--flush {
+	margin: 0;
+}
+
+.newspack-notice--spaced {
+	margin: 8px 0;
 }
```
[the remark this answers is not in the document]
thomasguillot: Right, split into `--spaced-top` and `--spaced-bottom` (using the `$grid-unit-10` token) so each notice keeps its original single-sided spacing.

plugins/newspack-plugin/src/admin/style.scss
```diff
@@ -146,15 +146,17 @@ h1 {
 	background: #{wp-colors.$white};
 }
 
-// Only apply styles if there are sections and is immediate descendent.
+.newspack-wizard,
+.newspack-wizard *,
+.newspack-wizard__content,
+.newspack-wizard__content * {
+	box-sizing: border-box;
```
chickenn00dle, line 162: **Severity:** suggestion

Tachyons previously shipped a global `* { box-sizing: border-box }` that applied to every element on any admin page loading `admin.css`. This reset scopes it to the `.newspack-wizard` / `.newspack-wizard__content` subtrees, which covers the wizard body (`class-wizard.php` renders it as `<div class="newspack-wizard …">`). One surface falls outside these selectors: the admin header, rendered by `trait-wizards-admin-header.php` as a separate `#newspack-wizards-admin-header` div (containing `.newspack-wizard__header`) that is a sibling of — not nested in — `.newspack-wizard`, and its own `admin-header.css` doesn't set box-sizing. That header CSS is minimal (mostly flex, no explicit widths), so the risk is low, but since the previously-global reset is gone it's worth a quick visual check on the admin header + tabbed navigation that nothing reverted to `content-box`.

plugins/newspack-plugin/src/wizards/audience/views/campaigns/analytics/index.js
```diff
@@ -28,11 +29,11 @@ const PopupAnalytics = () => (
 					{ __( ' to see how Campaign data is being recorded in GA.', 'newspack-plugin' ) },
 				</>
 			</p>
-			<Card buttonsCard noBorder>
+			<HStack justify="flex-start" spacing={ 4 } wrap>
```
chickenn00dle: **Severity:** suggestion

The old `.newspack-card__buttons-card` carried `margin: 32px 0`. Modal footers restore the top half via `.newspack-modal__footer { margin-top: 32px }`, but the `buttonsCard` → `HStack` replacements that didn't pick up that class lose the 32px vertical margin entirely — here (the help-page button) and the `componentsDemo` rows (Web Previews / Handoff / Modal-open / button demos). The dev demo is low-stakes; here it means up to 32px less space above the button row. Worth confirming the spacing still reads right, or adding a small scoped margin if it collapsed.

chickenn00dle (commented):
Static mapping audit of the tachyons → stacks refactor. Up front: I couldn't run this in a browser (environment limitation), so this is a static review of the class/component mapping, not a visual verification — the spacing/layout parity items below are worth a quick visual pass before merge.

Overall this is a clean, well-scoped swap. **Release risk: LOW.** The mechanics check out against the `@wordpress/components` source:

- `HStack spacing={ 4 }` → 16px gap matches the old `.newspack-card__buttons-card` gap; `spacing={ 2 }` → 8px matches the old `mr2` / `code marginTop: 8`.
- An explicit `justify` prop correctly overrides the default `alignment="edge"` (in `useHStack`, `...otherProps` is spread after `...align`), so `justify="flex-end"` and friends behave as intended.
- `expanded` defaults to `true` → `width: 100%`, so dropping `w-100` is a no-op.
- `alignment="bottom"` → `align-items: flex-end` matches the old `items-end` (ad-units legacy row, webhooks header).
- The SCSS translations all match: `.newspack-setup-completed__card` (= `flex flex-column justify-between`), `.newspack-webhooks__header` margin-bottom 32px (= `mb4`), and `--spaced-top` / `--spaced-bottom` using `$grid-unit-10` (8px, = the old single-sided `mt2` / `mb2`).
- No leftover tachyons utility classes in the touched packages, no JS querying the removed class names, and the `segment-group` `buttonsCard` usages correctly stay on the untouched `CoreCard` path (which keeps its own class + SCSS).

Two things worth a look, both inline below. (`newspack-multibranded-site` still loads its own tachyons from a CDN — that's independent of this PR, no action.)

plugins/newspack-plugin/src/wizards/audience/views/campaigns/analytics/index.js
thomasguillot: Good catch on the two non-modal cases — handled separately:

**The help-page button here:** this whole `analytics` view turned out to be orphaned. `PopupAnalytics` is a `withWizardScreen` "Coming soon" placeholder left over from the old Campaigns wizard — imported nowhere and rendered on no route since the Audience IA rework (the `.pot` strings are the only other references to it). Rather than add a margin to a screen nobody sees, I removed the dead view (`index.js` + `style.scss`).

**componentsDemo rows:** the 32px doesn't fully disappear — the row's `margin-top` was collapsing against each section heading's ~17px bottom margin, so dropping it nets ~15px on a dev-only page. Verified in-browser (before/after side-by-side): visually indistinguishable, so left as-is.

plugins/newspack-plugin/src/admin/style.scss
thomasguillot: Addressed — extended the border-box reset to cover the admin header subtree too, so it no longer relies on the old global Tachyons reset:

```scss
.newspack-wizard,
.newspack-wizard *,
.newspack-wizard__content,
.newspack-wizard__content *,
#newspack-wizards-admin-header,
#newspack-wizards-admin-header * {
	box-sizing: border-box;
}
```

Now the header + tabbed nav stay border-box regardless of what `admin-header.css` sets.

plugins/newspack-plugin/src/admin/style.scss
```diff
@@ -153,15 +153,19 @@ h1 {
 	background: #{wp-colors.$white};
 }
 
-// Only apply styles if there are sections and is immediate descendent.
+.newspack-wizard,
+.newspack-wizard *,
```
chickenn00dle, line 157: **Severity:** suggestion

**Suggestion:** The new `box-sizing: border-box` reset is scoped to `.newspack-wizard`, `.newspack-wizard__content`, and `#newspack-wizards-admin-header`. But the newspack `Modal` wraps `@wordpress/components` Modal, which renders via a React portal to `document.body` — outside all three scoped roots — and WP's own modal CSS sets no `box-sizing`, so modal (and popover) content previously got border-box *only* from Tachyons' former global `*{box-sizing:border-box}`. It now reverts to content-box. This PR does most of its work inside modals (`newspack-modal__footer` HStacks, the header-bidding-gam "Manage Orders" VStack of bordered ActionCards). Practically likely benign — box-sizing only changes layout for elements with an explicit width/height *plus* padding/border, and most modal content is `width:auto` block flow — but it's a genuine scope gap the earlier box-sizing rounds didn't cover.

**Consider:** extend the reset to `.newspack-modal, .newspack-modal *` (and popover content), or eyeball the converted modals — the header-bidding-gam Manage Orders modal (bordered cards with `margin-top:-1px` overlaps) is the highest-box-model surface.

plugins/newspack-ads/src/wizard-settings/header-bidding-gam/style.scss
```diff
@@ -1,6 +1,30 @@
+@use "~@wordpress/base-styles/colors" as wp-colors;
+
 .newspack-card.newspack-action-card.newspack-card--has-grey-header.newspack-card--has-children
 .newspack-header-bidding-gam-action-card
 .newspack-action-card__region-top {
 	background-color: #fff;
```
chickenn00dle: **Severity:** suggestion

**Suggestion:** `background-color: #fff;` — this file imports `wp-colors` and uses `wp-colors.$gray-300` two lines down, so this should be `wp-colors.$white` for token consistency. It's a re-homed legacy rule, but it's newly written into this file, so worth tokenizing now.

chickenn00dle (approved):
Re-review — no blockers. This is a clean, well-scoped presentational refactor, and I verified the parts that matter for a refactor: `admin.css` rebuilds to 6.4KB (from 77.6KB) with zero Tachyons selectors, the dependency is fully gone from every `package.json` and the lockfile, and — the load-bearing check — a thorough static scan for references to the now-undefined Tachyons classes across `plugins/` and `packages/` came back **empty** in all admin-loaded code. Spacing equivalences hold (the `mt4`/`mb4` → 32px restorations, `buttonsCard` gap → `HStack spacing={4}`, the `.newspack-modal__footer` and `--spaced-top/-bottom` rules), the JS/TSX wrappers are purely presentational (handlers/props/keys preserved), and the prior-round fixes are sound.

The visual layout of the converted surfaces still wants a manual pass — I verified the build and the static mapping, not the rendered pixels. The one thing worth eyes there is inline below: the scoped border-box reset doesn't reach portaled modal/popover content, which the old global Tachyons reset did.

A couple of low-severity non-blocking token/cleanup notes are inline too. Nice reduction.

- `packages/components/src/card/index.js:27` — **Nit:** The classic `Card` still destructures/accepts the `buttonsCard` prop, but the line that applied `newspack-card__buttons-card` (and its SCSS) was removed, so the prop is now inert on the classic card. No regression — the only remaining `buttonsCard` consumers pass `__experimentalCoreCard` and route to `core-card.js` where it still applies — just a leftover prop that could confuse future readers.

plugins/newspack-plugin/src/admin/style.scss
thomasguillot: Checked this and the reset isn't actually missing. `@wordpress/components` ships its own box-sizing reset scoped to the portalled frame:

```css
.components-modal__frame { box-sizing: border-box }
.components-modal__frame *,
.components-modal__frame ::before,
.components-modal__frame ::after { box-sizing: inherit }
```

`.components-popover` has the identical pattern. Confirmed both in the runtime CSS (`wp-includes/css/dist/components/style.min.css`, WP 7.0.2) and in `node_modules/@wordpress/components/build-style/style.css`.

Since the frame is border-box and every descendant inherits it, modal and popover content never depended on Tachyons' former global reset. That covers the header-bidding-gam "Manage Orders" cards as well. No change needed.

plugins/newspack-ads/src/wizard-settings/header-bidding-gam/style.scss
thomasguillot: Tokenized: `background-color: wp-colors.$white`. Compiled output is unchanged (`#fff`).

PR #522, opened 2026-07-03
fix(blocks): allow opted-in post types on the public /articles endpoint

jason10lee:
### All Submissions:

Restores **Sponsors load-more** in Homepage Posts / Carousel (resolves NPPM-2972).

NPPM-2964 (#471) narrowed the public `/articles` endpoint to `is_post_type_viewable()` post types. The Sponsors CPT is `public=false`, so it's dropped there — a Sponsors-only block renders page 1 via SSR (which queries directly, not through `/articles`) but load-more / infinite-scroll returns empty.

**Design — explicit opt-in, not a broad marker:**
- **newspack-blocks** adds a `newspack_blocks_articles_allowed_post_types` filter: `is_post_type_viewable( $type ) || in_array( $type, $allowed, true )`. Default `[]` → **no behavior change** for any site that doesn't opt in.
- **newspack-sponsors** hooks it to allow **only** its own CPT — one explicit, documented, per-CPT exposure decision, living next to the CPT.

I deliberately **did not** widen via `post_type_supports( 'newspack_blocks' )` (the other option floated on the ticket). That marker is declared broadly (`post`, `page`, Events Calendar, Listings, Collections, Sponsors) and isn't a "safe to expose anonymously" signal — using it would auto-expose any *future* `public=false` CPT that opts into block support. The explicit filter keeps the anonymous-exposure decision auditable and owned by the plugin registering the CPT.

**Why this is a low-sensitivity opt-in:** Sponsor data (name, logo, disclosure/byline) is already surfaced publicly as sponsor labels; the endpoint still returns only `publish` status to anonymous callers; and #471's `specificPosts` protection still holds for every type *not* on the allow-list (WP_Query `post_type` ∧ `post__in` intersection).

1. `n test-php --filter HomepagePostsBlockTest` — the new `test_articles_endpoint_allows_filter_allowlisted_post_type` passes: an allow-listed non-viewable CPT is served **and** a non-allow-listed one is still dropped.
2. With newspack-sponsors active, a Sponsors-only Homepage Posts block with more sponsors than the page size: load-more now returns sponsors (previously empty). Page-1 SSR already worked via the same `build_articles_query`, so this only realigns the REST path with it.

### Other information:

_Note: the Sponsors-side opt-in isn't unit-tested (newspack-sponsors has no PHPUnit suite); the filter mechanism is covered in newspack-blocks, and the Sponsors query itself is exercised by the existing page-1 SSR path._


plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -330,4 +330,34 @@ public function test_tag_labels_rest_field_empty_returns_false() {
...
+		add_filter( 'newspack_blocks_articles_allowed_post_types', $filter );
+		wp_set_current_user( 0 );
+
+		$controller = new WP_REST_Newspack_Articles_Controller();
+		$request    = new WP_REST_Request( 'GET', '/newspack-blocks/v1/articles' );
+		$request->set_param( 'postType', [ $allowed, $excluded ] );
+		$request->set_param( 'postsToShow', 10 );
+		$ids = $controller->get_items( $request )->get_data()['ids'];
+
+		remove_filter( 'newspack_blocks_articles_allowed_post_types', $filter );
```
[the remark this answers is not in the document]
jason10lee: Good catch — wrapped the `get_items()` call in `try`/`finally` so `remove_filter()` always runs even if the call throws.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -330,4 +330,37 @@ public function test_tag_labels_rest_field_empty_returns_false() {
...
+		$request->set_param( 'postsToShow', 10 );
+		try {
+			$ids = $controller->get_items( $request )->get_data()['ids'];
+		} finally {
+			// Always drop the global filter so a failure here can't leak into later tests.
+			remove_filter( 'newspack_blocks_articles_allowed_post_types', $filter );
+		}
+
+		self::assertContains( $allowed_id, $ids, 'An allow-listed non-viewable post type is served.' );
+		self::assertNotContains( $excluded_id, $ids, 'A non-viewable post type not on the allow-list is still dropped.' );
```
chickenn00dle, line 366: **Severity:** suggestion

The new test proves an allow-listed CPT's published posts are served and non-listed types are dropped, but it doesn't assert the security-critical invariant the fix hinges on: that an unpublished (draft/private) post of the allow-listed CPT stays hidden from an anonymous caller. Creating a `draft` post of the allowed type and asserting `assertNotContains( $draft_id, $ids )` under `wp_set_current_user( 0 )` would lock the status gate in place against future regressions in `build_articles_query()`. I confirmed at runtime that drafts are correctly excluded today — this just guards it going forward.

plugins/newspack-sponsors/includes/class-core.php
```diff
@@ -51,6 +51,24 @@ public function __construct() {
...
+	 * The CPT is public=false (no standalone URLs), so newspack-blocks otherwise
+	 * drops it from that unauthenticated endpoint. Its published data is already
+	 * surfaced publicly as sponsor labels/bylines, so this is a deliberate,
+	 * low-sensitivity opt-in of published sponsors only.
+	 *
+	 * @param string[] $post_types Post type slugs the endpoint may serve.
+	 * @return string[]
+	 */
+	public static function allow_in_articles_endpoint( $post_types ) {
+		$post_types[] = self::NEWSPACK_SPONSORS_CPT;
```
chickenn00dle, line 71: **Severity:** nit

`allow_in_articles_endpoint()` appends via `$post_types[] = ...` without guarding that `$post_types` is an array. It's safe today (the controller seeds the filter with `[]` and casts the final result), but a defensive `(array) $post_types` cast would keep the callback robust if another filter callback in the chain ever returns a non-array.

plugins/newspack-sponsors/includes/class-core.php
```diff
@@ -51,6 +51,24 @@ public function __construct() {
 		add_filter( 'newspack_ads_should_show_ads', [ __CLASS__, 'suppress_ads' ], 10, 2 );
 		add_filter( 'newspack_ads_ad_targeting', [ __CLASS__, 'ad_targeting' ], 10, 2 );
 		add_action( 'pre_get_posts', [ __CLASS__, 'ensure_only_sponsors' ] );
+		add_filter( 'newspack_blocks_articles_allowed_post_types', [ __CLASS__, 'allow_in_articles_endpoint' ] );
```
chickenn00dle: **Severity:** nit

This `add_filter()` omits the explicit priority and accepted-args count that the sibling `add_filter()` calls just above it use (`10, 2`). Adding `, 10, 1` would keep it consistent with the surrounding convention.

chickenn00dle (approved):
Release risk: LOW. Clean, well-scoped fix. The new `newspack_blocks_articles_allowed_post_types` filter defaults to `[]`, so sites that don't opt in see no change, and the allow-list is layered additively on top of `is_post_type_viewable()` rather than replacing it. The security boundary holds: the status gate lives downstream in `build_articles_query()`, which hard-defaults `post_status` to `['publish']` for anonymous callers (widening only with `edit_others_posts`) and sets `has_password => false` — so opting a `public=false` CPT into the endpoint only ever exposes its published, non-password-protected posts, never drafts or private content. I verified this at runtime against an isolated env with newspack-sponsors active: the sponsors CPT is now served to anonymous `/articles` requests (published sponsors returned, a draft sponsor excluded), a non-allow-listed non-viewable CPT is still dropped, and a normal `post` request is unaffected. The full newspack-blocks PHPUnit suite passes. No blockers — just a few minor items worth a glance before merge, most valuably a regression test that pins the draft/private exclusion the fix relies on.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
jason10lee: Added in `048767fb0` as `test_articles_endpoint_allowlist_still_hides_unpublished` — published, draft and private posts of the allow-listed type, asserting under `wp_set_current_user( 0 )` that only the published one comes back.

Checked that it can actually fail: loosening `$included_post_statuses` at `class-newspack-blocks.php:617` to include `draft` makes it fail on exactly that assertion, so the guard is tied to the status gate rather than passing incidentally.

plugins/newspack-sponsors/includes/class-core.php
jason10lee: Done in `048767fb0` — `$post_types = (array) $post_types;` ahead of the append.

plugins/newspack-sponsors/includes/class-core.php
jason10lee: Done in `048767fb0` — now `add_filter( 'newspack_blocks_articles_allowed_post_types', [ __CLASS__, 'allow_in_articles_endpoint' ], 10, 1 );`, matching the siblings above it.

PR #523, opened 2026-07-03
fix(oauth): guard unparseable proxy /start response

jason10lee:
### All Submissions:

Resolves NPPM-2971 (follow-up from the #461 / NPPM-2963 review, @adekbadek).

In `Google_OAuth::google_auth_get_url()`, if the OAuth proxy `/start` returns a
body that isn't valid JSON, `json_decode()` yields `null`. The success path then
did `return $response_body->url` — on PHP 8 a warning that returns `null`, which
silently propagates downstream as a non-`WP_Error` failure.

- **Success path:** guard `$response_body` with `is_object()` **and** `isset( ->url )`
  before dereferencing, returning a clean `WP_Error` if the response can't be used.
- **Error (non-200) path:** the sibling `property_exists( $parsed_data, 'message' )`
  three lines up had the same hazard (a `TypeError` on a `null`, caught but ugly);
  guarded it with `is_object()` too so the whole method is robust to a non-JSON proxy
  response.

Pre-existing robustness only — not a live bug (the existing `isset( ->client_id )`
guard was already null-safe, and PHP 8 warns rather than fatals).

1. `n test-php --filter Newspack_Test_OAuth` — two new cases pass: an unparseable
   `/start` body and a parseable-but-url-less body each return a `WP_Error` (not `null`).

### Other information:


plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -185,6 +185,14 @@ public static function google_auth_get_url( $auth_params ) {
...
 			}
 			$response_body = json_decode( $result['body'] );
+			// Guard against an unparseable proxy response so we return a clean error rather
+			// than dereferencing null (a PHP warning that then propagates null downstream).
+			if ( ! is_object( $response_body ) || ! isset( $response_body->url ) ) {
+				return new WP_Error(
+					'newspack_google_oauth',
+					__( 'Could not parse the authentication response.', 'newspack' )
+				);
+			}
```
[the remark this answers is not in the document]
jason10lee: Good call — tightened the guard to `empty( $response_body->url ) || ! is_string( $response_body->url )`, so an empty/`0`/non-string url now returns a WP_Error too. Added a test for the empty-url case.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -185,6 +185,14 @@ public static function google_auth_get_url( $auth_params ) {
...
+			// error rather than dereferencing null or returning a non-string url downstream.
+			if ( ! is_object( $response_body ) || empty( $response_body->url ) || ! is_string( $response_body->url ) ) {
+				return new WP_Error(
+					'newspack_google_oauth',
+					__( 'Could not parse the authentication response.', 'newspack' )
+				);
+			}
 			// Remember the client id the proxy issues tokens for, so received tokens can be
 			// confirmed to have been issued to this app.
 			if ( isset( $response_body->client_id ) ) {
```
adekbadek: **Blocker:** The guard stops one line short: an unguarded `client_id` can silently disable the token-audience check — The new guard type-checks `url`, but the very next statement writes `client_id` with no type check at all — and `client_id` is the one field on this response that carries security meaning.

`isset()` passes for any non-null value, including an array. `sanitize_text_field()` returns `''` for an array or object (WP core's `_sanitize_text_fields()` short-circuits on `is_object() || is_array()`). So `update_option( CLIENT_ID_OPTION_NAME, '' )` overwrites a previously-good client id with an empty string. `get_expected_client_id()` then returns `''`, and that is exactly the value that makes `validate_token_and_get_email_address()` take the `else` branch and **proceed without verifying the token's audience** (`if ( '' !== $expected_client_id )`, ~line 376) — the check #461 added to stop a Google token issued to some other application from signing a reader in.

Reproduced end to end on this branch, in the real WP runtime: with a stored client id of `good-client-id.apps.googleusercontent.com`, a `/start` response of `{"url":"https://accounts.google.com/...","client_id":[]}` returned the url happily, left `client_id` as `''`, and `get_expected_client_id()` returned `''` — audience check off.

This is persisted state, so it does not clear when the proxy recovers: every Google sign-in from that point on is accepted without the check, silently, until a later well-formed response happens to restore the value. A change whose stated purpose is surviving a malformed response from this endpoint should not leave open the one field that can turn off a security control. Require the same shape you now require of `url`, and don't let a bad value replace a good one:

```php
if ( ! empty( $response_body->client_id ) && is_string( $response_body->client_id ) ) {
 update_option( self::CLIENT_ID_OPTION_NAME, sanitize_text_field( $response_body->client_id ), false );
}
```

**Risk if not addressed:** If that empty value is stored, the site permanently stops checking that a Google access token was issued to its own application, because the reader sign-in callback proceeds without verification whenever the stored client id is empty. Every subsequent reader sign-in accepts any Google token carrying a verified email, including one obtained through an unrelated Google application. That is the exact check added in #461 to stop reader accounts being taken over via a substituted token, and it stays off until a later well-formed response happens to restore the value — the proxy recovering does not undo it.

**Consider:** Only store the client id when it is a non-empty string, and leave the previously stored value untouched otherwise. An unusable value from the proxy should not be allowed to replace a known-good one, because the empty value is what turns the token check off.

plugins/newspack-plugin/tests/unit-tests/oauth.php
```diff
@@ -450,4 +450,83 @@ public function test_rejects_admin_scoped_token_with_mismatched_audience() {
...
+
+		$result = Google_OAuth::google_auth_get_url(
+			[
+				'csrf_token'     => 'csrf-token-123',
+				'scope'          => 'https://www.googleapis.com/auth/userinfo.email',
+				'redirect_after' => 'https://example.org/',
+			]
+		);
+
+		self::assertTrue( is_wp_error( $result ), 'An unparseable /start response must return a WP_Error, not null.' );
```
adekbadek, line 536: **Blocker:** Two of the three new tests pass against the unpatched code, and the path that actually fatalled has none — The tests do not encode the fix. Reverting only the production guard and re-running them, **`test_start_response_unparseable_yields_error` and `test_start_response_missing_url_yields_error` both still pass**. Only the empty-url case goes red.

The reason: PHPUnit converts the PHP 8 warning *"Attempt to read property 'url' on null"* into a `PHPUnit\Framework\Error\Warning`, which **is** an `\Exception` — so the method's own `catch ( \Exception $e )` swallows it and returns a `WP_Error` whose message is literally `Attempt to read property "url" on null`. `assertTrue( is_wp_error( $result ) )` cannot tell that apart from the guard firing. The tests would not catch the guard's removal, and the PR description's "two new cases pass" reads as proof of the fix when both would have passed without it.

Asserting the message pins the branch — verified green on this code and red on the unpatched code:

```php
self::assertSame(
 'Could not parse the authentication response.',
 $result->get_error_message()
);
```

Two cases are also missing, and they're the ones that matter most:

1. **The non-200 path — the only one that actually fatalled.** All three new tests stub `[ 'code' => 200 ]`. A 502 with an HTML body (an ordinary gateway outage) was an uncaught `TypeError` pre-fix; that's the regression most worth a test, and it has none.
2. **The happy path is unpinned.** Nothing asserts a well-formed `/start` response still returns its url — `test_start_response_client_id_is_persisted()` discards the return value. As written, all three new tests would still pass if the guard were `if ( true ) { return new WP_Error( … ); }`, so a future over-tightening of this condition could break Google sign-in for every publisher with the suite still green.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -176,7 +176,7 @@ public static function google_auth_get_url( $auth_params ) {
 			if ( 200 !== $result['response']['code'] ) {
 				$error_text  = __( 'Request failed.', 'newspack' );
```
adekbadek: **Suggestion:** The new failure paths are silent, and the non-200 branch throws away the HTTP status — Both new error paths log nothing, and this branch collapses every non-200 into a bare `'Request failed.'`, discarding `$result['response']['code']` entirely.

That trades a noisy fatal for a silent generic notice on the exact failure this PR exists to surface. Before the change, a malformed `/start` response at least left a fatal in the PHP error log — which is how NPPM-2971 came to be filed at all. After it, the proxy can start returning HTML error pages and the only trace anywhere is a transient toast in one admin's browser. `Google_OAuth::api_google_auth_get_url()` (the admin "connect Google account" path) doesn't log its `WP_Error` either.

Every other failure in this class announces itself via `Logger::error()`, and the class already reaches for `Logger::newspack_log()` when it wants an anomaly auditable platform-wide (see the client-id skew handling ~line 386). Since NPPM-2971 is fundamentally an observability bug, a `Logger::error()` — with the status code and a truncated body — plus the status folded into the `WP_Error` (message or data) is what makes the next proxy outage diagnosable from logs rather than from a publisher report.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -176,7 +176,7 @@ public static function google_auth_get_url( $auth_params ) {
...
 				$parsed_data = json_decode( $result['body'] );
-				if ( property_exists( $parsed_data, 'message' ) ) {
+				if ( is_object( $parsed_data ) && property_exists( $parsed_data, 'message' ) ) {
```
adekbadek: **Suggestion:** `$parsed_data->message` still has the exact type-confusion just fixed for `url` — This line was hardened against a `null` `$parsed_data`, but `->message` is still trusted to be a string. A proxy response of `{"message":{"code":500}}` yields a `WP_Error` whose message is an array, which then flows into `Google_Login::api_google_auth_get_url()` → `sprintf( __( 'Failed to get Google OAuth URL: %s' ), $url->get_error_message() )` and into `wp_die( esc_html( … ) )` elsewhere — array-to-string conversion notices, which is the same class of downstream damage this PR closes for `url`.

Applying the same discipline makes the two branches symmetric (`isset()` is also the more idiomatic guard than `property_exists()` for a decoded `stdClass`):

```php
if ( is_object( $parsed_data ) && ! empty( $parsed_data->message ) && is_string( $parsed_data->message ) ) {
 $error_text = $parsed_data->message;
}
```

**Risk if not addressed:** If the proxy's response shape changes in a way this code does not anticipate, the one remaining unguarded field that has security meaning — the client id — absorbs it, while the field that was guarded keeps working, so the failure looks like a non-event.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -185,6 +185,14 @@ public static function google_auth_get_url( $auth_params ) {
 				);
 			}
 			$response_body = json_decode( $result['body'] );
+			// Guard against an unparseable or malformed proxy response so we return a clean
+			// error rather than dereferencing null or returning a non-string url downstream.
+			if ( ! is_object( $response_body ) || empty( $response_body->url ) || ! is_string( $response_body->url ) ) {
```
adekbadek: **Suggestion:** The proxy-supplied url reaches a same-origin navigation sink without a scheme check — The guard now enforces "non-empty string", but nothing constrains the value to an `http(s)` URL — and both consumers hand it straight to a navigation sink. The reader-facing flow opens `window.open( 'about:blank', … )` and then assigns `authWindow.location = data` (`src/reader-activation-auth/google-oauth.js:53` and `:93`). An `about:blank` popup opened by the page **inherits the opener's origin**, so a `javascript:` URL assigned to its `location` would execute in the site's own origin — on a path any unauthenticated visitor can reach (`/login/google` gates only on `is_oauth_configured()`).

The trust boundary is the proxy over TLS, so this is defense-in-depth rather than a live exploit. But this guard is the natural and cheap place to close the shape, and it removes the dependency on the upstream service staying healthy:

```php
if ( ! is_object( $response_body ) || empty( $response_body->url ) || ! is_string( $response_body->url ) || ! wp_http_validate_url( $response_body->url ) ) {
```

**Consider:** Confirm the returned address uses an http or https scheme before handing it back, since the browser assigns it directly to a popup that shares the site's origin.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -185,6 +185,14 @@ public static function google_auth_get_url( $auth_params ) {
...
+				return new WP_Error(
+					'newspack_google_oauth',
+					__( 'Could not parse the authentication response.', 'newspack' )
```
adekbadek: **Suggestion:** New reader-facing string uses a text domain the plugin never extracts — The plugin header declares `Text Domain: newspack-plugin`, and `class-newspack.php` loads that domain — so strings tagged `'newspack'` are never extracted and never translated. Confirmed against the shipped catalogue: `languages/newspack-plugin.pot` contains `"Google OAuth is not configured."` (tagged `newspack-plugin`, from `class-google-login.php`) but has **zero** entries for the `'newspack'`-domain strings in this file, e.g. `"Session token mismatch."`.

This string reaches readers, not just admins — `Google_Login::api_google_auth_get_url()` returns this `WP_Error` straight out of the public `/login/google` route — so it lands untranslated for every non-English publisher. Local file consistency argues for `'newspack'`; correctness argues for `'newspack-plugin'`, which is what the rest of `includes/oauth/` already uses. The ~677 legacy `'newspack'` usages are a separate cleanup, but new strings shouldn't be added to the pile.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -185,6 +185,14 @@ public static function google_auth_get_url( $auth_params ) {
...
+				);
+			}
```
adekbadek, line 223: **Suggestion:** The `catch` still cannot catch what actually broke here — widen it to `\Throwable` — This fixes the two known dereferences, but the structure that let the original failure escape as a 500 is untouched. The `catch ( \Exception $e )` on line 202 cannot catch `\TypeError`, `\Error`, or `\ArgumentCountError` — they extend `\Error`, not `\Exception`. That is precisely why the pre-fix `property_exists( null, … )` surfaced as an uncaught fatal rather than the `WP_Error` this method contracts to return (`@return string|WP_Error`).

As it stands the catch-all only ever catches the `\Exception` thrown by `OAuth::authenticate_proxy_url()`; everything it was presumably meant to backstop, it silently doesn't. `catch ( \Throwable $e )` is a one-word change that makes the documented error contract actually hold, and it's the difference between patching the two dereferences we happened to find and this endpoint being unable to take the site down.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
```diff
@@ -176,7 +176,7 @@ public static function google_auth_get_url( $auth_params ) {
 			if ( 200 !== $result['response']['code'] ) {
```
adekbadek: **Nit:** Prefer the WP HTTP retrieval helpers over raw array indexing — `$result['response']['code']` and `$result['body']` are raw array reads on a value whose shape is only guaranteed by convention — any `pre_http_request` filter can return an array without a `response` key, which is exactly the "the transport lied to us" scenario this PR exists to harden against. The sibling client in this same directory already models the pattern end to end (`class-google-oauth-ga4-client.php`: `is_wp_error()` → `wp_remote_retrieve_response_code()` → `wp_remote_retrieve_body()` → `is_array( $decoded )`), and `validate_token_and_get_email_address()` / `token_has_scope()` in this very file use `wp_remote_retrieve_*` too — so the raw access here is the outlier.

plugins/newspack-plugin/tests/unit-tests/oauth.php
```diff
@@ -450,4 +450,83 @@ public function test_rejects_admin_scoped_token_with_mismatched_audience() {
 			'An admin-scoped token with a mismatched audience must be rejected.'
 		);
 	}
+
+	/**
+	 * Stub the proxy /start endpoint with a canned 200 response body.
+	 *
+	 * @param string $body Raw response body.
+	 */
+	private function stub_start_response( $body ) {
```
adekbadek: **Nit:** The new helper duplicates the stub already inlined in `test_start_response_client_id_is_persisted()` — `stub_start_response()` is a verbatim generalisation of the stub that `test_start_response_client_id_is_persisted()` already open-codes (~lines 365-391): same `set_api_key()`, same conditional `define()` of `NEWSPACK_GOOGLE_OAUTH_PROXY`, same `pre_http_request` filter keyed on the same URL fragment. Extracting the helper and leaving the duplicate behind is the worst of both — two copies of the proxy response shape that can now drift. Fold the existing test onto the helper, which also puts the happy-path url assertion right next to the failure cases.

(For the record, the two mechanics that look risky are fine: `WP_UnitTestCase_Base` snapshots and restores `$wp_filter` in `tear_down()`, so the filter does not leak between tests, and the guarded `define()` matches what the surrounding tests already do.)

adekbadek (changes requested):
The guard is correct, and it fixes more than it claims: pre-fix, a 502 with an HTML body from the proxy was an **uncaught `TypeError`** — a 500 on the public `/login/google` route — not the "caught but ugly" error described here. Verified in the real WP runtime, outside PHPUnit.

Two things should be fixed before this merges.

Two follow-ups in this same file, outside this diff but the identical bug class, worth folding in or filing: `api_google_auth_revoke()` (~line 300) indexes `wp_safe_remote_post()`'s result as an array with no `is_wp_error()` check, so any timeout or DNS failure is an uncaught fatal in a REST callback; and `validate_token_and_get_email_address()` (~line 356) does `json_decode()` then reads `->scope` with no object guard, on the reader sign-in path.

jason10lee:
Both blockers reproduced as described. All nine items are addressed across four commits: b4fc919 (client id), 50d0e61 (hardening + logging), bb042b6 (tests), 6f3c5c7 (status code).

**On "it fixes more than it claims":** confirmed. `property_exists( null, … )` raises a `TypeError`, which `catch ( \Exception )` does not catch — so the pre-fix 502-with-an-HTML-body was an uncaught fatal on the public route, not a caught error. The catch is now `\Throwable`, so the documented `string|WP_Error` contract holds.

**On the client id blocker:** reproduced end to end. Starting from a stored `good-client-id...`, a `/start` response carrying an array `client_id` returned the url and left `get_expected_client_id()` returning `''`. Fixed as suggested — a value is only stored when it is a non-empty string, so an unusable response cannot replace a known-good one, and the unusable case is now logged rather than silently swallowed.

**On the tests:** confirmed. With only the production guard reverted, `test_start_response_unparseable_yields_error` and `test_start_response_missing_url_yields_error` both still passed, returning a `WP_Error` whose message was literally `Attempt to read property "url" on null` — PHPUnit's warning-to-exception conversion swallowed by the method's own catch, as you described. Every failure path now asserts its message, and we mutation-tested the result: removing the url guard fails 4 tests, reverting the client id guard fails 1, reverting the `->message` guard fails 1. The suite no longer passes without the fix.

We also added the two missing cases: a 502 with an HTML body (the path that actually fatalled), and a happy-path assertion pinning that a well-formed response still returns its url.

**On the status code:** adopted as suggested (6f3c5c7) — the proxy's status now becomes the REST response status, so a proxy 502 surfaces as a 502 rather than a blanket 500. We checked the two things that could have made this risky and neither holds: nothing in the plugin's JS branches on the response status (`WizardApiError.statusCode` is stored but never read), and WP core's `rest_convert_error_to_response()` only honours `data['status']` when `is_numeric()`, so a malformed transport response falls back to 500 rather than emitting a bogus status. An impact review across the monorepo and the Manager repos found no consumer of these endpoints' status outside `newspack-plugin`.

**The two out-of-diff follow-ups** are filed as NPPM-3003 — both are the same bug class, and neither is reachable from the `/start` path this PR is scoped to.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Confirmed and fixed in b4fc919. We reproduced the full chain first: a stored `good-client-id...` plus a `/start` response with an array `client_id` left `get_expected_client_id()` returning `""`. Adopted your suggested shape — stored only when a non-empty string, so a bad response cannot replace a good value — and added a `Logger::error()` on the rejected-value branch so it isn't silent. Regression test in bb042b6 (`test_start_response_unusable_client_id_does_not_replace_stored_value`); it fails if the guard is reverted to `isset()`.

plugins/newspack-plugin/tests/unit-tests/oauth.php
jason10lee: Confirmed — both tests did still pass with the guard reverted, and the returned message was `Attempt to read property "url" on null`, the swallowed-warning path you identified. Fixed in bb042b6: every failure case now asserts its message, so the guard is what has to produce the error. Added the two missing cases — a 502 with an HTML body, and a happy-path assertion that a well-formed response still returns its url — and folded the client-id test onto the shared helper so it pins the happy path. Mutation-checked: removing the url guard now fails 4 tests, not 1.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61 and 6f3c5c7. Both failure paths now `Logger::error()` with the status code and a truncated body (capped at 500 chars, since the proxy can fail with a whole HTML error page), and the status is folded into the `WP_Error` data as `status` — so a proxy 502 becomes a REST 502 instead of a blanket 500. We verified no consumer branches on the status of these endpoints, in this plugin or in the Manager repos, and WP core falls back to 500 when the value isn't numeric. Covered by `test_start_non_200_html_body_yields_error`.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61, using your suggested guard verbatim (`is_object` + `! empty` + `is_string`). Covered by `test_start_non_200_non_string_message_yields_generic_error`, which fails if the guard reverts to `property_exists`.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61 — added `wp_http_validate_url()` to the guard. Agreed it's defense-in-depth rather than a live exploit, but it's cheap and it removes the dependency on the proxy staying healthy. `test_start_response_non_http_url_yields_error` covers a `javascript:` url.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61. We confirmed the diagnosis before changing it: the plugin header and `load_plugin_textdomain()` both declare `newspack-plugin`, and the shipped `.pot` has zero entries for the `'newspack'`-domain strings in this file. New strings in this method now use `newspack-plugin`. The ~677 legacy usages are left alone, as you suggested.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61. We verified the premise directly — `property_exists( null, … )` throws a `TypeError`, and `$e instanceof \Exception` is `false` for it — so the old catch never backstopped the failure this PR exists to fix. This is the change that makes the `@return string|WP_Error` contract true.

plugins/newspack-plugin/includes/oauth/class-google-oauth.php
jason10lee: Fixed in 50d0e61 — now `wp_remote_retrieve_response_code()` / `wp_remote_retrieve_body()`, matching `class-google-oauth-ga4-client.php` and the other methods in this file.

plugins/newspack-plugin/tests/unit-tests/oauth.php
jason10lee: Fixed in bb042b6. `test_start_response_client_id_is_persisted()` now uses `stub_start_response()` (extended to take a status code, for the 502 case), and it asserts the returned url as well — so the happy path is pinned next to the failure cases, as you suggested.

adekbadek (approved):
Re-reviewed after the follow-up commits. Both blockers are resolved and verified against the real WP runtime: the `client_id` write now stores only a non-empty string (a bad value logs and keeps the stored id, so the token-audience check can't be silently disabled), and the tests now pin the error message — reverting the guard turns 4 of them red, so they encode the fix rather than a swallowed PHP warning. The added url-scheme validation, status-in-`WP_Error`, and logging are all sound, and surfacing the proxy status as the REST status checks out given no consumer branches on it.

One non-blocking follow-up: the same unguarded-response bug class this PR fixes still lives in two sibling methods in this file — `api_google_auth_revoke()` (indexes `$result['response']['code']` with no `is_wp_error()` check → fatal on a timeout/DNS failure) and `validate_token_and_get_email_address()` (reads `->scope` on an unguarded `json_decode()`, on the public reader sign-in path). Worth folding into this PR or filing as a follow-up — your call which.

PR #544, opened 2026-07-07
feat(subscribers): add discounts demo prototype

thomasguillot:
> [!WARNING]
> **Do not merge.** This is a design-exploration sandbox layered on top of the prototype PR #295. It targets `feat/subscribers-demo-prototype`, not `main`.

## What this is

A second, fully independent copy of the subscribers demo that adds mocked **member product discount** management, exploring the admin UX for [NPPD-1794: Implement WCM product discount functionality](https://linear.app/a8c/issue/NPPD-1794/implement-wcm-product-discount-functionality) and the [AC Product Discount user stories](https://linear.app/a8c/document/ac-product-discount-user-stories-4389131fa2e3). The existing subscribers demo stays untouched and simple; this copy is where the discount UX evolves.

Reachable at `admin.php?page=newspack-subscribers-discounts-demo` when the site is in Newspack debug mode. The original demo remains at `admin.php?page=newspack-subscribers-demo`, and the two never share state (separate localStorage namespaces).

## Discounts tab

A third tab beside Subscribers and Groups: a DataViews list of discount rules with **Audience** (plan), **Discount** ("£51.00 off" / "10% off", following the site's WooCommerce currency), **Applies to** ("5 products", "Books category", "All products", with exclusions noted), **Status** (Active/Paused) and **Created** columns. Row actions: Edit, Pause/Resume and Delete, each confirming via the demo's small-modal grammar with snackbar results. Five seeded rules cover every state, including a Yucatan-style fixed amount off a hand-picked list of five books.

## Rule editor (two-step modal, products first)

- **Step 1, who and what**: audience plan picker (Digital / Print / Team) and targeting via a segmented control: specific products (token search against a mock catalog), a category, or the whole store, with an exclusions field for the latter two.
- **Step 2, discount**: fixed amount or percentage, with a **live preview table** (product, price, member price) updating as you type. Fixed mode adds an "or set the member price" input: type the target price and the discount is computed, the arithmetic publishers currently do by hand. This is the affordance the user stories call out as the single biggest improvement over WooCommerce Memberships.

## Global settings

A small Settings modal on the Discounts tab with the two cross-rule knobs from the user stories: apply member discounts to on-sale products (toggle), and best-discount-only vs. combine when multiple rules match.

## Profile Benefits section

A Benefits section on the subscriber profile (after Subscriptions) listing the discounts that reader's plans grant, each opening the rule editor. Group members inherit the owner's plan benefits: a reader whose only access comes through a Team Monthly group shows the seeded "15% off Courses" rule.

## Mock data

- `data/mock-catalog.js`: ~15 store products shaped like a real news storefront (books, a variable-format photo album, guided talks, events, multi-thousand courses) in Books / Events / Courses categories.
- `data/mock-discounts.js`: the five seeded rules plus localStorage-backed CRUD, settings, and pricing/label helpers, unit-tested (`mock-discounts.test.js`).

## How the copy works

Full copy of `src/wizards/subscribersDemo/` to `src/wizards/subscribersDiscountsDemo/` (deliberate: total independence over diff size; drift is accepted). Only functional renames differ: wizard slug, dist bundle (webpack auto-discovers the folder), avatars REST route, mount id, and the localStorage prefix (`newspack-subscribers-discounts-demo:*` with its own `DATA_VERSION`). The inline config global and CSS class namespace intentionally keep their original names since the two demos never load on the same page. The first commit is the verbatim copy plus registration; each discount feature lands as its own commit for reviewability.

## Testing

- `pnpm build` emits `dist/subscribersDiscountsDemo.{js,css,asset.php}`; full jest suite green (30 suites / 283 tests).
- Smoke-tested in the browser: list, both editor steps with live preview and linked member-price inputs, pause/delete/settings flows, localStorage isolation between the demos, and the Benefits section on a group-member-only profile.
- The original demo's diff against the base branch is empty apart from the two registration lines (`class-newspack.php`, `class-wizards.php`).


thomasguillot:
Superseded by #547: after design discussion we landed on a "2 and 2" IA — Subscribers/Groups stay people management in the base demo, and discounts now sit beside a plan-centric Subscriptions tab in a dedicated subscriptions demo (all the discount work from this PR was transplanted there, including review fixes).

PR #555, opened 2026-07-08
feat(content-gate): add in-editor gate preview (NPPD-1871)

adekbadek:

Adds a **Preview** button to the gate layout editor (`np_gate_layout`) that autosaves the layout and opens a `WebPreview` modal iframe of a recent published post with the gate force-rendered — mirroring the newspack-popups prompt preview.

Closes NPPD-1871

**How it works** (parallels popups):
- **Editor** (`src/content-gate/editor/preview.js`): Preview button in `PluginPostStatusInfo` → `autosave()` → opens `WebPreview` at the preview post's permalink with `ngp_id=<layout id>` plus the six layout metas as abbreviated query params (autosaves don't persist meta). In-iframe article links are rewritten to keep the preview active on navigation.
- **Server** (`includes/content-gate/class-gate-preview.php`): a new `Gate_Preview` controller detects the preview request (capability-gated on `edit_others_pages`, filterable, no nonce — popups parity), then force-drives the display pipeline via existing filter seams at `PHP_INT_MAX`: forces the queried post restricted, points the layout/gate IDs at the previewed layout, short-circuits metering, substitutes the layout's **autosaved** content via a new `newspack_gate_layout_content` filter, overrides the six layout metas from the query params, and hides the admin bar.

**Parity decision (no Woo Memberships bypass):** the preview keeps parity with the real front-end. On WCM-active sites the front-end is still controlled by Memberships, so rather than show a misleadingly ungated post, the button is simply **not offered there** (no preview data localized when `Memberships::is_active()`). It's also not offered on the legacy `np_memberships_gate` layout CPT.

https://github.com/user-attachments/assets/253f5d84-b5ba-4e71-8d71-3e50b3c2db14

1. On a site with `NEWSPACK_CONTENT_GATES` defined and Woo Memberships **inactive**, create a content gate (Audience → Access Control), then edit one of its layouts.
2. Make an **unsaved** edit to the layout content, then click **Preview** in the status panel.
3. The modal iframe shows a recent published post gated with your unsaved content; the admin bar is hidden and article links inside the iframe carry the preview params.
4. Switch Style to **Overlay** (still unsaved) and Preview again → overlay renders at the chosen position/size.
5. Negative: open the same preview URL logged out → you see the genuine gate behavior only (no unsaved-content leak, style override ignored).
6. Parity: activate Woo Memberships → the Preview button is absent from the layout editor.

### Other information:

- 12 new unit tests in `tests/unit-tests/content-gate/gate-preview.php`; full content-gate group green (372 tests). PHPCS + ESLint clean.
- The WCM-parity absence of the button is verified by logic/unit gating; the full Woo + Memberships browser check is left to env QA.
- Cross-repo note: `newspack-manager` consumes the content-gate contract. The new `newspack_gate_layout_content` filter and `Gate_Preview` class are additive/backward-compatible.

https://claude.ai/code/session_01GQzJpwGHATH8KNrqCGZbwq


plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,442 @@
...
+	 *
+	 * @return mixed|null Sanitized value, or null if invalid (stored meta wins).
+	 */
+	private static function sanitize_meta_override( $meta_key, $raw ) {
+		switch ( $meta_key ) {
+			case 'style':
+				return in_array( $raw, [ 'inline', 'overlay' ], true ) ? $raw : null;
+			case 'inline_fade':
+			case 'use_more_tag':
+				return rest_sanitize_boolean( $raw );
```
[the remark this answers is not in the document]
adekbadek: Good catch — fixed in 5a2f81dbcde977d0f4ce7b96651737597246be3d. Boolean params now only accept recognized true/false strings; anything else (e.g. `ngp_if=abc`) is dropped so the stored meta wins, matching the docblock. Added a regression assertion.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,442 @@
...
+	 */
+	public static function filter_is_post_restricted( $restricted, $post_id ) {
+		if ( ! self::is_preview_request() ) {
+			return $restricted;
+		}
+		if ( empty( $post_id ) || (int) $post_id === (int) get_queried_object_id() ) {
+			return true;
+		}
+		return $restricted;
+	}
```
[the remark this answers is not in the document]
adekbadek: Fixed in 5a2f81dbcde977d0f4ce7b96651737597246be3d. The filter now resolves an empty `$post_id` to the queried object and forces restriction only when the target equals the queried post; an unrelated post id passes through, and nothing is forced when there is no singular queried object. Added a test covering all three cases.

plugins/newspack-plugin/src/content-gate/editor/preview.js
```diff
@@ -0,0 +1,74 @@
...
+	};
+
+	const onWebPreviewLoad = iframeEl => {
+		if ( ! iframeEl ) {
+			return;
+		}
+		[ ...iframeEl.contentWindow.document.querySelectorAll( 'a[href^="' + frontendUrl + '"]' ) ].forEach( anchor => {
+			anchor.setAttribute( 'href', addQueryArgs( anchor.getAttribute( 'href' ), query ) );
+		} );
+	};
```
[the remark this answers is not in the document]
adekbadek: Fixed in 5a2f81dbcde977d0f4ce7b96651737597246be3d. The in-iframe link rewrite is now wrapped in try/catch so a cross-origin DOMException (mapped domains, scheme mismatch) is logged and swallowed — the preview still opens; only the link-param carry-over is skipped.

adekbadek:
Addressed all three review comments in 5a2f81d (stricter boolean drop, queried-post-only force-restrict, cross-origin iframe guard) — pushed, threads resolved. @copilot please re-review.

plugins/newspack-plugin/tests/unit-tests/content-gate/gate-preview.php
```diff
@@ -0,0 +1,431 @@
...
+	 * A subscriber hitting the preview URL is not a preview request and gets no
+	 * forced gating.
+	 */
+	public function test_subscriber_with_param_is_not_gated() {
+		wp_set_current_user( $this->subscriber_id );
+		$this->set_query_param( Gate_Preview::PREVIEW_QUERY_PARAM, $this->layout_id );
+
+		$this->assertFalse( Gate_Preview::is_preview_request(), 'Subscriber with the param is not a preview request.' );
+
+		$this->go_to( get_permalink( $this->canvas_post_id ) );
```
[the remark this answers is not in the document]
adekbadek: Good catch — fixed in a2c19eec8eda707755ebb7f7ddbb55b4689838f0. The test now carries `ngp_id` in the `go_to()` URL (which rebuilds $_GET), so the render assertion genuinely exercises a subscriber on the preview URL rather than the post-wipe no-param case. Same fix already applied to the admin-preview test.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,496 @@
...
+				$normalized = strtolower( trim( (string) $raw ) );
+				if ( in_array( $normalized, [ 'true', '1', 'yes', 'on' ], true ) ) {
+					return true;
+				}
+				if ( in_array( $normalized, [ 'false', '0', 'no', 'off' ], true ) ) {
+					return false;
+				}
+				return null;
+			case 'visible_paragraphs':
+				return is_numeric( $raw ) ? absint( $raw ) : null;
```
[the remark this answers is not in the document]
adekbadek: Good catch — fixed in b78510cfc6f2ad469a769c8a1b3eb7fd41b5e2e5. `visible_paragraphs` now clamps with `max( 0, (int) $raw )`, matching get_visible_paragraphs(), so a negative param resolves to 0 instead of its absolute value. Added a regression assertion.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,499 @@
...
+	 * Find the gate that references the given layout.
+	 *
+	 * Gates link to layouts (not the reverse) via their registration /
+	 * custom_access `gate_layout_id`. Scans both the regular and newsletter gate
+	 * sets. Returns the first matching gate ID regardless of its post status;
+	 * the caller decides how to treat a draft parent.
+	 *
+	 * The layout→parent-gate mapping is memoized per request: this is consulted
+	 * from filter_gate_post_id, which fires many times per preview render
+	 * (has_gate, metering, analytics, overlay), and each scan runs two uncached
```
chickenn00dle, line 267: **Suggestion** — During a preview, `filter_is_post_restricted()` forces restriction on whatever post the request queries, and `ngp_vp` (visible_paragraphs) is clamped to `max( 0, (int) )` with no upper bound. A user who satisfies the preview capability could hand-edit the URL to point the preview at a genuinely restricted post and set `ngp_vp` very high, so the excerpt builder emits the full body as the "excerpt". This isn't reachable for standard roles — `edit_others_pages` is held by Editors/Admins, who already pass the `edit_post` exemption in `is_post_restricted()` and see the ungated post anyway — so the only exposure is an exotic custom role granting `edit_others_pages` without post-edit rights. Since the capability is the sole documented guard, a defense-in-depth upper bound on `ngp_vp` (or scoping the force-restrict to the canvas post id) would close the edge at no cost.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,499 @@
...
+			'preview_post'       => self::preview_post_permalink(),
+			'frontend_url'       => get_site_url(),
+			'query_param'        => self::PREVIEW_QUERY_PARAM,
+			'preview_query_keys' => self::PREVIEW_QUERY_KEYS,
+		];
+	}
+
+	/**
+	 * Force the queried post to be restricted during a preview.
+	 *
```
chickenn00dle, line 360: **Suggestion** — `new WP_Query( $args )` already runs the query during construction and stores the results in `$query->posts`; calling `$query->get_posts()` again re-executes the SQL. Use `$posts = $query->posts;` to avoid the second query (the WordPress/VIP duplicate-query pattern). Low blast radius here — editor-triggered, `no_found_rows` + `posts_per_page => 1` — but it's a free fix, and the mirrored newspack-popups version carries the same wart.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,499 @@
...
+	 * The request-immutable signals are memoized for the request lifetime since
+	 * this is consulted from hot filters. The capability check is evaluated
+	 * fresh on every call, never cached: it depends on the current user, which
+	 * may not be established when the first hot filter fires — caching a `false`
+	 * from an early call would silently disable the preview for the whole
+	 * request. The cache is skipped under IS_TEST_ENV, where a single suite
+	 * drives multiple users and query states through this method (precedent:
+	 * Content_Gate::is_newspack_feature_enabled()).
+	 *
+	 * @return bool
```
chickenn00dle, line 147: **Nit** — The preview capability defaults to `edit_others_pages`, but the gate-layout CPT registers with no `capability_type`, so it maps to *post* capabilities — editing others' layouts actually needs `edit_others_posts`. A role with `edit_others_posts` but not `edit_others_pages` could author a layout yet get no Preview button (and vice-versa). It's intentional newspack-popups parity and filterable via `newspack_gate_preview_capability`, so functionally fine — just worth a one-line comment noting the mismatch is deliberate.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
```diff
@@ -0,0 +1,499 @@
...
+	 * @param bool $show Whether to show the admin bar.
+	 *
+	 * @return bool
+	 */
+	public static function filter_show_admin_bar( $show ) {
+		if ( $show && self::is_preview_request() ) {
+			return false;
+		}
+		return $show;
+	}
```
chickenn00dle, line 540: **Nit** — `filter_layout_meta()` calls `get_preview_meta_overrides()` — which re-reads `$_GET` and re-runs sanitization for all six keys — on every matching meta read for the previewed layout during a render. It's preview-request-only and cheap, but memoizing the overrides array per request would mirror the caching discipline already applied to `request_targets_preview()` and `get_layout_parent_gate_id()`.

chickenn00dle (commented):
Release risk: LOW — no blockers. Careful, well-documented work: the preview drives the real content-gate render pipeline through the existing filter seams rather than reimplementing gate logic, so there's no divergent second implementation to drift from the front-end. The security model holds — every force-render filter no-ops unless the capability + valid-layout + feature-on + Memberships-off check passes; metering is short-circuited so the previewing admin's allowance is never consumed; and the preview shows the gate prompt over a public canvas post, never unlocked article content. A few low-priority items worth a glance before merge (all inline).

Two process notes: editor/visual verification of the preview UX was **not** performed here (no browser environment available), so please leave that for a manual/env QA pass — it's the feature's core value. And the branch currently reports a merge conflict with `main` that'll need a rebase.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
adekbadek: Added a defense-in-depth ceiling: `ngp_vp` now clamps to a new `PREVIEW_MAX_VISIBLE_PARAGRAPHS` (50) constant, so even the exotic-role case can't crank the count to dump a gated post's full body via the excerpt builder's `array_slice( $paragraphs, 0, $count )`. Went with the upper bound rather than scoping the force-restrict to the canvas post: the leak flows through the layout's `visible_paragraphs` meta override (which applies regardless of which post is queried), and a genuinely-restricted target is restricted independently of our force-restrict, so scoping that filter alone wouldn't close it. Fully scoping the whole preview to the canvas post would mean re-querying the canvas id inside the memoized `request_targets_preview()`, which risks caching a `false` before the main query resolves. Covered by a new assertion in test_meta_overrides_are_sanitized.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
adekbadek: Fixed – now reads `$query->posts` (already populated by the constructor) instead of calling `get_posts()` a second time, with a comment noting the WordPress/VIP duplicate-query pattern.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
adekbadek: Added a note in the `newspack_gate_preview_capability` docblock spelling out that the gate-layout CPT has no `capability_type` and so maps to post caps (`edit_others_posts`), that the `edit_others_pages` default is deliberate newspack-popups parity, and that the filter is the escape hatch for roles holding one capability but not the other.

plugins/newspack-plugin/includes/content-gate/class-gate-preview.php
adekbadek: Memoized `get_preview_meta_overrides()` per request, mirroring the caching discipline in `get_layout_parent_gate_id()` (request-static cache, skipped under `IS_TEST_ENV` since a single suite drives multiple `$_GET` states through it). filter_layout_meta() no longer re-reads $_GET and re-sanitizes all six keys on every matching meta read during a render.

adekbadek:
- fix(content-gate): harden gate preview per review round 2 (NPPD-1871)
- style(content-gate): align equals sign in overlay gate render (NPPD-1871)

chickenn00dle (approved):
Second-round review — no blockers. All nine round-1 comments are addressed, CI is green, and the change is additive/backward-compatible (new `newspack_gate_layout_content` filter + new `Gate_Preview` class; the one listed content-gate consumer, newspack-manager, hooks none of the changed seams). I traced the security model end-to-end and it holds: every force-render filter no-ops for a logged-out or non-capable visitor even with a fully hand-crafted `ngp_id`+`ngp_*` URL, because the capability check is re-evaluated fresh on every call; metering is short-circuited so the admin's allowance isn't consumed; and the preview only ever shows the gate prompt over a public `post`, never unlocked article content.

Approving with just one non-blocking note:

The most substantive note: the client-side gate script (`gate.js`) has no preview awareness, so an admin preview fires a real `seen` impression — it bumps the admin's own `paywall_hits` (which persists and syncs to the ESP `Paywall Hits` field via a sticky `max()` merge) and dispatches a GA4 `seen` event. The server-side metering short-circuit deliberately protects the metering allowance but not this path. It's admin-volume and scoped to the admin's own contact record (no reader-facing amplifier), so it's a suggestion, not a blocker — but worth either suppressing reporting when `ngp_id` is present, or noting it in the PR description.

PR #556, opened 2026-07-08
feat(reader-activation): add --skip-lists and --fields to esp sync CLI (NPPD-1883)

adekbadek:
### All Submissions:

Adds two orthogonal flags to the `wp newspack esp sync` WP-CLI command so a site's existing ESP (e.g. ActiveCampaign) contacts can be backfilled with the new Content Access metadata fields without side effects. Needed because those fields were added after the sync was built, so dormant contacts never received them, and a plain re-sync would resubscribe unsubscribed readers and rewrite every field.

- **`--skip-lists`** — upserts each contact *without* a master list, so an unsubscribed contact is not resubscribed. Missing contacts are still created (list-less). Honored by integrations that read the sync options (the built-in ESP integration does).
- **`--fields=<name1,name2>`** — restricts both the *computed* and the *pushed* metadata to just the named fields (raw keys or display labels, case-insensitive) and leaves the reader's name untouched, so a scoped backfill can't rewrite names or other fields.

Under the hood:

- Threads an `$options` array (`skip_lists`, `fields`) through `Contact_Sync::sync_contact() → sync() → push_to_integrations() → ESP::push_contact_data()`. The abstract `push_contact_data()` stays 3-argument (a 4th optional param fatals existing 3-arg implementations); only the ESP override adds `$options`. The optional-param convention is documented in the integrations README.
- Adds `Metadata::resolve_field_labels()` (canonicalizes tokens to labels; dedupes synonymous keys) and a `$fields` param on `Metadata::get_contact_with_metadata()` for compute-side class filtering, with a special case so requesting a payment field still runs the legacy class that computes it.
- Push-side field/name filtering lives in a shared `Contact_Sync::prepare_contact_for_integration()` helper, reused by the dry-run preview.
- Failed pushes under these flags are **not** auto-retried (the retry handler rebuilds the full contact and would push it with the master list, undoing the intent); operators re-run the affected `--offset` window.
- CLI pre-flight validates `--fields` (unknown / unavailable / not-enabled-as-outgoing) before any batch runs.
- Fixes a pre-existing counter bug so the command reports `Synced X (Y errors, Z skipped)` across all four sync branches, not just the order-ids branch.

Enables the eventual backfill invocation:

```
wp newspack esp sync --skip-lists --fields="Content Access,Content Access Source,Content Access Group" --sync-context="Content Access backfill (NPPD-1883)"
```

1. On a site with the `NEWSPACK_CONTENT_GATES` flag on, Reader Activation enabled, a configured ESP, and the Content Access fields enabled as outgoing fields:
2. Pre-flight errors surface before any work: `wp newspack esp sync --dry-run --user-ids=<id> --fields="Not A Real Field"` (unknown), a field not enabled as outgoing (targeted error), and with the flag off, `--fields="Content Access"` reports the field as unavailable.
3. `wp newspack esp sync --dry-run --skip-lists --fields="Content Access,Content Access Source,Content Access Group" --user-ids=<id>` reports the correct totals and (with `NEWSPACK_LOG_LEVEL` on) logs a per-integration payload containing exactly the three prefixed keys with lists skipped.
4. A real run (`NEWSPACK_FORCE_ALLOW_ESP_SYNC` + a hook on `newspack_newsletters_pre_add_contact`) confirms `$lists === false`, no `name`, and exactly the three prefixed Content Access keys reach the upsert.
5. Regression: a plain `wp newspack esp sync --dry-run --user-ids=<id>` is unchanged (full payload, master list) and now reports a non-zero count.
6. `n test-php` — the new `Test_Contact_Sync_Options` and `Test_RAS_Contact_Sync_Options` suites pass and the full suite stays green.

### Other information:


plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -206,15 +283,15 @@ private static function push_to_integrations( $contact, $context, $existing_cont
...
+			$integration_contact = self::prepare_contact_for_integration( $integration, $contact, $options );
 
 			// Added logging here to more easily monitor integration sync data. Can be removed once integrations are released.
 			if ( 'legacy' !== Metadata::get_version() ) {
 				Logger::log( sprintf( 'Syncing contact %s for integration %s with context "%s".', $integration_contact['email'] ?? 'unknown', $integration_id, $context ) );
 				Logger::log( $integration_contact );
 			}
 
-			$result = $integration->push_contact_data( $integration_contact, $context, $existing_contact );
+			$result = $integration->push_contact_data( $integration_contact, $context, $existing_contact, $options );
```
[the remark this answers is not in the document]
adekbadek: PHP does not warn or error when a method is called with more positional arguments than it declares — the surplus args are silently discarded (still reachable via `func_get_args()`). Verified two ways: a direct `$obj->method($a,$b,$c,$d)` on a 3-param method prints with no notice, and the full test suite passes with `Failing_Sample_Integration` (which declares the 3-param signature) receiving the 4-arg call here. The alternative — adding `$options` to the abstract `push_contact_data()` — is what actually breaks BC: it is a fatal "declaration must be compatible" error for every existing 3-param override (also verified). So the 4-arg call is safe and intentional; no change needed.

plugins/newspack-plugin/includes/reader-activation/integrations/README.md
```diff
@@ -238,6 +238,15 @@ When a contact needs to be synced, the framework calls `push_contact_data()` on
...
 `prepare_contact()` is a no-op when the site is still on the legacy metadata schema (where the metadata classes pre-filter), which keeps newly-built integrations compatible with un-migrated sites.
 
+### Optional `$options` parameter
+
+`Contact_Sync` may pass a fourth `$options` array to `push_contact_data()` carrying operator-driven sync scoping (currently used by the `wp newspack esp sync` CLI):
+
+- `skip_lists` (bool) — upsert the contact without adding it to any list, so an unsubscribed contact isn't resubscribed.
+- `fields` (string[]|null) — the canonical field labels the sync is scoped to (already applied to the metadata before your method is called).
+
+The abstract signature intentionally stays three-parameter (`push_contact_data( $contact, $context, $existing_contact )`) so existing three-param implementations keep working — PHP silently ignores the extra argument. Add `$options = []` to your override only if the integration needs to react to these flags (the built-in `esp` integration reads `skip_lists`). Integrations that ignore `$options` behave exactly as before.
```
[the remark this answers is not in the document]
adekbadek: The README statement is accurate — PHP silently discards surplus positional arguments (no warning). I have expanded this paragraph in the latest commit to say so explicitly and to record why the parameter is not added to the abstract (doing so fatals existing 3-param overrides).

plugins/newspack-plugin/tests/unit-tests/cli/class-test-ras-contact-sync-options.php
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	public function test_defaults_when_no_options_passed() {
+		$options = $this->parse( [] );
+		$this->assertSame(
+			[
+				'skip_lists' => false,
+				'fields'     => null,
+			],
+			$options
```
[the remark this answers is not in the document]
adekbadek: Fixed in the latest commit.

plugins/newspack-plugin/tests/unit-tests/reader-activation-sync/class-test-contact-sync-options.php
```diff
@@ -0,0 +1,443 @@
...
+			'push_to_integrations',
+			[
+				$contact,
+				'ctx',
+				null,
+				[
+					'skip_lists' => true,
+					'fields'     => $this->content_access_labels,
+				],
+			]
```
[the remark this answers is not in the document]
adekbadek: Fixed in the latest commit.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -168,22 +172,95 @@ public static function sync( $contact, $context = '', $existing_contact = null )
...
+	private static function prepare_contact_for_integration( $integration, $contact, $options = [] ) {
+		if ( ! empty( $options['fields'] ) ) {
+			unset( $contact['name'] );
+		}
+
+		$integration_contact = $integration->prepare_contact( $contact );
+
+		if ( empty( $options['fields'] ) ) {
+			return $integration_contact;
+		}
```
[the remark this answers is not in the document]
adekbadek: Good catch — moved the `unset()` to operate on `$integration_contact` after `prepare_contact()` in ced01abe9, so the name guarantee holds regardless of what an integration override does. Now matches the docblock.

plugins/newspack-plugin/includes/reader-activation/integrations/class-esp.php
```diff
@@ -366,19 +366,24 @@ public function can_sync( $return_errors = false ) {
...
 	 */
-	public function push_contact_data( $contact, $context = '', $existing_contact = null ) {
+	public function push_contact_data( $contact, $context = '', $existing_contact = null, $options = [] ) {
 		$can_sync = $this->can_sync( true );
 		if ( $can_sync->has_errors() ) {
 			return $can_sync;
 		}
 
-		$master_list_id = $this->get_master_list_id();
+		$master_list_id = ! empty( $options['skip_lists'] ) ? false : $this->get_master_list_id();
```
chickenn00dle, line 445: **Severity:** suggestion

When `skip_lists` is set, `master_list_id` becomes `false`, which flows through `Newspack_Newsletters_Contacts::upsert()` to the provider's `upsert_contact()` with an empty list set. Base providers (ActiveCampaign, Constant Contact) handle that as a list-less create/update and write the metadata — but the Mailchimp provider overrides `upsert_contact()` and returns `WP_Error( 'No lists found.' )` on an empty list set *before* writing any merge fields. So on a Mailchimp site the list-less backfill writes no metadata for any contact (it fails loud — each contact is counted as an error in the tally, not silent). Since a `--skip-lists` backfill is the headline use case, consider detecting a Mailchimp provider in the CLI pre-flight (alongside the not-enabled-fields check) and erroring early with an actionable message, or at minimum documenting the Mailchimp limitation next to the `--skip-lists` flag.

**Risk if not addressed:** if the list-less backfill is run on Mailchimp without first confirming the metadata lands, the Content Access fields are never written and any downstream gating/segmentation depending on them operates on unpopulated data until someone notices.

plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
```diff
@@ -68,9 +85,10 @@ protected static function log( $message, $data = [] ) {
 	 *   @type int         $config['max_batches'] Maximum number of batches to process.
 	 *   @type bool        $config['is_dry_run'] True if a dry run.
 	 *   @type string      $config['context'] Context of the sync.
+	 *   @type array       $config['options'] Sync options ( `skip_lists` bool, `fields` string[]|null ).
 	 * }
 	 *
-	 * @return int|\WP_Error Number of synced contacts or WP_Error.
+	 * @return array|\WP_Error Results tally ( `processed`, `errors`, `skipped` ) or WP_Error.
 	 */
 	private static function sync_contacts( $config ) {
```
chickenn00dle, line 93: **Severity:** suggestion

The reworked results tally (`record_result()`, the per-branch `skipped++`, the reset-on-entry, and the new `Synced/Would sync X (Y errors, Z skipped)` line) is the pre-existing counter bug this PR sets out to fix, but nothing drives `sync_contacts()` and asserts the returned `['processed','errors','skipped']` tally. A single test running a small batch with a success/error/skip mix (e.g. an inactive user under `--active-only`) and asserting the tally would lock in the fix that motivated the change.

plugins/newspack-plugin/includes/reader-activation/sync/class-metadata.php
```diff
@@ -365,23 +365,115 @@ public static function get_all_fields( $only_available = false ) {
...
+	 *
+	 * @param string[] $inputs Field tokens to resolve.
+	 *
+	 * @return string[]|\WP_Error Ordered, de-duplicated labels, or a WP_Error for
+	 *                            the first token that is unknown
+	 *                            (`newspack_esp_sync_unknown_field`) or known but
+	 *                            currently unavailable
+	 *                            (`newspack_esp_sync_unavailable_field`).
+	 */
+	public static function resolve_field_labels( array $inputs ) {
```
chickenn00dle: **Severity:** nit

The new helpers (`resolve_field_labels`, `build_field_lookup`, `class_handles_any_field`, and the CLI's `parse_sync_options` / `options_are_default` / `prepare_contact_for_integration`) omit return type declarations, where the PHP 8.3 target favors them and a sibling method in this same file (`get_grouped_default_fields(): array`) already uses one. They'd annotate cleanly (`: array|\WP_Error`, `: array`, `: bool`). Purely a modernization nudge on the new surface — PHPCS won't flag it.

chickenn00dle (commented):
Nice, careful piece of work — the flags are cleanly plumbed, backward compatibility with the default (no-flags) path is preserved end-to-end, and the two new test suites assert the behaviors that matter. No blockers. A few things worth a look before merge.

**Release risk: ELEVATED**, for one provider-specific reason: on Mailchimp the list-less path (`--skip-lists`, the primary backfill mode) is rejected by the provider and writes no metadata, so the backfill is effectively a no-op there. It fails loud — every contact returns an error in the results tally, so it's recoverable once the output is read — but the feature's stated purpose isn't achieved on Mailchimp as written. Worth either detecting this in the pre-flight (like the not-enabled-fields check) or documenting the Mailchimp limitation next to the flag. See the inline note.

One operational caution to call out too: `--fields` and `--skip-lists` are orthogonal, and a field-only run *without* `--skip-lists` still upserts each contact to the configured master list — which can re-subscribe readers who had opted out. Since the two flags read as interchangeable, it's worth a line in the command docs that a name/subscription-preserving backfill needs *both* flags together.

Everything else is minor (see inline comments). This is code-only review — not runtime-exercised against a live ESP.

plugins/newspack-plugin/includes/reader-activation/integrations/class-esp.php
adekbadek: Done the real fix rather than just documenting: `parse_sync_options()` now hard-fails the pre-flight when `--skip-lists` is combined with a Mailchimp provider (`newspack_esp_sync_skip_lists_mailchimp`), since Mailchimp's `upsert_contact()` override returns `No lists found.` before writing any merge fields, so a list-less backfill would push metadata for no one. Added a note to the `--skip-lists` help text as well, plus tests covering the Mailchimp error and the non-Mailchimp pass-through.

plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
adekbadek: Added `tests/unit-tests/cli/class-test-ras-contact-sync-tally.php`, which drives `sync_contacts()` via reflection and asserts the returned `['processed','errors','skipped']` tally: a processed+skipped mix (active reader vs. an inactive reader under `--active-only`), an errors+skipped mix (failing integration), and a reset-between-runs case (the static tally must not accumulate across calls). Runs are field-scoped to the non-WC `Content Access` field so contact computation skips the WooCommerce classes — the shared `WC_Subscription` mock doesn't extend `WC_Order`, so building a subscribed reader's full metadata trips an unrelated mock-fidelity error.

plugins/newspack-plugin/includes/reader-activation/sync/class-metadata.php
adekbadek: Added return type declarations to all six new methods: `resolve_field_labels(): array|\WP_Error`, `build_field_lookup(): array`, `class_handles_any_field(): bool`, `options_are_default(): bool`, `prepare_contact_for_integration(): array`, and `parse_sync_options(): array|\WP_Error`.

adekbadek:
- fix(reader-activation): block --skip-lists on Mailchimp esp sync (NPPD-1883)

chickenn00dle (approved):

PR #564, opened 2026-07-08
feat(newsletters): block-based WooCommerce email rendering engine (behind flag)

chickenn00dle:
## Summary

**Adds a block-based WooCommerce email-editor rendering engine for newsletters, behind a feature flag that is off by default.** This is the `epic/editor-refactor` line of work landing on `main`.

Legacy newsletters render through MJML. This epic introduces a parallel path built on the vendored `woocommerce/email-editor` package (the same engine MailPoet uses), giving the editor and the sent email 1:1 parity and unlocking WC-native blocks that MJML couldn't support.

- **Feature-flagged, off by default** — gated on `Feature_Flag::is_enabled()` (option `newspack_newsletters_use_woo_renderer`, constant `NEWSPACK_NEWSLETTERS_WOO_RENDERER`, or the `newspack_newsletters_use_woo_renderer` filter). With the flag off, nothing changes: the legacy MJML path is untouched.
- **When enabled** — newsletters render via the WC engine end to end (editor preview, sent email, engine stamped on send), the editor becomes theme-native, and the allow-list opens up table / gallery / media-text / cover (plus experimental audio/video behind a second opt-in filter).
- **Block parity + full-bleed** — per-block renderer overrides bring buttons, quotes, separators, social links, posts-inserter, share and the ad block to email parity; `alignfull` background sections and cover blocks bleed edge-to-edge while content stays centered.
- **Send-path parity** — the final post-processing MJML bakes into the saved HTML now runs on the WC output too: **UTM + click-tracking link rewriting**, the **open tracking pixel**, **`custom_css`**, and **conditional-content merge-tag guards** so segment-targeted blocks are shown/hidden per recipient (rather than shipping unguarded to the whole list).

**Regression note:** every existing site keeps the MJML renderer until the flag is explicitly enabled, so this merge is a no-op by default.

### How to test

**Scenario A — regression (flag OFF, the default):**
1. On a Newspack site running this branch, don't touch the flag.
2. Create/edit a newsletter — confirm the editor, preview, and a test send all behave exactly as they do on `main` (MJML output, unchanged).

**Scenario B — new engine (flag ON):**
1. Enable it: `wp option update newspack_newsletters_use_woo_renderer 1` (or define the constant / add the filter).
2. Open a newsletter — the editor is now theme-native and the preview matches the email 1:1.
3. Insert the newly-supported blocks (table, gallery, media-text, cover) — they appear in the inserter and render correctly.
4. Add an `alignfull` cover or background group — it bleeds edge-to-edge while the inner content stays at content width; check in a desktop client and Outlook if possible.
5. Send a test newsletter and inspect it in MailHog (http://localhost:8025) — HTML is produced by the WC engine and matches the editor.
6. Confirm the producing engine is stamped on the sent newsletter.

**Scenario C — send-path parity (flag ON):**
1. Add a paragraph link, a button, and social links — confirm the sent hrefs carry `utm_medium=email` (link tracking parity).
2. With click tracking on, add an ad block — confirm the ad's links are proxied through the click endpoint (`np_newsletters_click`).
3. Confirm the **open tracking pixel** (``) is present, and that `custom_css` post meta is applied in the sent HTML.
4. Add a Group with conditional content (segment guard) — confirm the ESP merge tags (`conditionalBefore` / `conditionalAfter`) wrap the block in the sent HTML, including when that group is an `alignfull` full-bleed band.

### Submission checklist
- [ ] Follows coding standards (PHP/JS lint clean)
- [ ] Tests pass (`n test-php` in `plugins/newspack-newsletters`)
- [ ] Changelog handled by semantic-release

---

<details>
<summary><strong>For AI / reviewers — architecture, rebase details, conflict resolutions</strong></summary>

### Scope
- Rebased epic (23 commits, 54 files, +8750/−319) plus a review round (3 commits, +397/−40), entirely within `plugins/newspack-newsletters` (+ `composer.json`/`composer.lock` pulling in `woocommerce/email-editor@2.14.0` via jetpack-autoloader).
- Squash-merged PRs #316 → #511 into `epic/editor-refactor`.

### Architecture (`includes/email-renderers/`)
- **`Feature_Flag`** — single gate: option `newspack_newsletters_use_woo_renderer` (default `false`) → constant override → filter override.
- **`Renderer_Controller`** — `render_wc( $post )` drives the WC engine; `active_engine()` resolves MJML vs WC; stamps the producing engine on send. A `render_block` filter (priority 20, after the package's own priority-10 callback) wraps conditional-content blocks in their ESP merge-tag guards, and a `finalize_html()` pass after the full-bleed transform applies `process_links()` (UTM + click), injects `custom_css` into `<head>`, and injects the open pixel before `</body>` — the send-path parity the MJML template bakes in.
- **`Editor_Bootstrap`** — boots the WC email-editor in the block editor when the flag is on (the boot is always-on but behaviorally inert flag-off; the editor takeover is flag-gated in `Newspack_Newsletters_Editor`).
- **`Theme_Json_Builder`** / **`Email_Defaults`** / **`Fonts`** — build a full newsletter `theme.json` from post meta so editor and email share one style source (reference model is vanilla WP, not legacy MJML). `build()` is a pure meta→theme.json translator (no flag branch).
- **`Block_Renderer_Registry`** — self-registering `blocks/class-*.php` overrides (ad, button, column, posts-inserter, quote, separator, share, social-links) for email parity. The ad renderer runs its HTML through `process_links()` with the *ad's* post context so ad links are click-tracked.
- **`Full_Bleed_Sections`** — hoists top-level `alignfull` + `has-background` groups (and `email-block-cover` cover blocks) to body-level `width:100%` bands, re-centering inner content to `CONTENT_WIDTH` (660px) with an Outlook MSO ghost-table pin.
- **`Newspack_Newsletters\Tracking\Pixel::get_pixel_markup()`** — shared open-pixel markup, reused by both the MJML body hook (wrapped in `<mj-raw>`) and the WC `finalize_html()` pass.
- REST: `api_get_post_html` (`post-html` endpoint) renders the saved newsletter server-side for the editor preview (read-only; the WC engine re-fetches by ID, so no live `content` override).
- Editor JS (`src/editor/index.js`) sets per-block `supports.align`; posts-inserter preview injects email-matching typography into the iframed `BlockPreview`.

### Review round (addressing @adekbadek's feedback)
Three commits on top of the rebase — full PHP suite green (657 tests), PHPCS clean:
- **`fix(newsletters): close WC email render parity gaps`** — the blocker (conditional-content guards) plus send-path parity: `process_links`/UTM/click, open pixel, `custom_css`, ad-link click tracking, and the preview-text fallback.
- **`fix(newsletters): address WC-renderer review nits`** — flag-consistent `Email_Defaults` (filters registered unconditionally; callbacks re-check the flag), `Theme_Json_Builder::build()` as a pure translator, missing type declarations, `@package` consistency, and an under-privileged (`edit_others_posts`) permission test.
- **`docs(newsletters): note WC package version and always-on bootstrap`** — the `2.14.0` version assumption + Woo-core-override risk, and why the editor bootstrap is always-on but inert flag-off.

**Kept as conscious choices (reasoning in-thread):** font/spacing scale duplication (deduping means refactoring the load-bearing legacy renderer), the render-failed-vs-empty error channel, and the send-time renderer stamp on legacy sends.

### Rebase
Rebased 24 commits from the old `epic/editor-refactor` onto latest `origin/main` (main had advanced ~301 commits since the branch point). Result: 23 commits, no markers, full PHP suite green (637 tests, 2780 assertions, 3 env-gated skips). Backup ref: `backup/epic-editor-refactor-prerebase`.

- **Dropped** the auto-generated `chore(release): 4.4.0-epic-editor-refactor.1 [skip ci]` prerelease commit (semantic-release artifact, would only conflict with main's own release history).
- **`class-newspack-newsletters.php`** — kept the new `api_get_post_html` method; did **not** resurrect `api_set_post_meta`, which `main` removed as orphaned dead code in #387 (there's a `test_orphaned_api_set_post_meta_is_removed` regression guard — it passes).
- **`posts-preview.js`** — kept epic's new `usePostItemStyles` hook; dropped the old `useLayoutStyle`, which `main` removed in #274 when the layout-css injection moved into `newsletter-preview/index.js`. Ref list updated to `[ ref, useIframeBorderFix, usePostItemStyles, useCustomFontsInIframe() ]`.

</details>


plugins/newspack-newsletters/tests/email-renderers/test-feature-flag.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Test that the NEWSPACK_NEWSLETTERS_WOO_RENDERER constant overrides the option.
+	 */
+	public function test_constant_overrides_option() {
+		if ( ! defined( 'NEWSPACK_NEWSLETTERS_WOO_RENDERER' ) ) {
+			define( 'NEWSPACK_NEWSLETTERS_WOO_RENDERER', true );
+		}
+		delete_option( 'newspack_newsletters_use_woo_renderer' );
+		$this->assertTrue( Feature_Flag::is_enabled() );
+	}
```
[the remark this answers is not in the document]
chickenn00dle: Good catch — a constant can't be undefined, so this would leave the flag forced on for the rest of the process. Added `@runInSeparateProcess` + `@preserveGlobalState disabled` so it can't leak. Fixed in 4eebb44de.

plugins/newspack-newsletters/newspack-newsletters.php
```diff
@@ -34,9 +34,32 @@
...
+	require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/vendor/autoload.php';
+} else {
+	add_action(
+		'admin_notices',
+		function() {
+			echo '<div class="notice notice-error"><p>';
+			echo esc_html__( 'Newspack Newsletters is missing its Composer dependencies. Please run "composer install" in the plugin directory.', 'newspack-newsletters' );
+			echo '</p></div>';
+		}
+	);
```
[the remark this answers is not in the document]
chickenn00dle: Fixed in 4eebb44de — `function ()` with the space to satisfy WP PHPCS.

chickenn00dle:
Fixed the PHPUnit failure in `85d4ec943`.

Root cause was a latent test-ordering bug that the test-isolation fix above surfaced: `Editor_Bootstrap::merge_theme_json()` memoized the built `WP_Theme_JSON` in a process-global `static` cache keyed only by post ID. Post IDs recycle across tests (per-test DB rollback), so `test_render_wc_applies_per_newsletter_background` got a stale theme from an earlier render and the per-newsletter background never applied. Adding `@runInSeparateProcess` to the constant test just shifted the post-ID sequence enough to expose it.

Fix mirrors the existing `Ads::reset_inserted_ads()` reset in `render_wc()`: the cache is now a resettable class property, cleared at the start of each render. This also closes a real (if narrow) production edge — a second render of a post whose colors changed mid-request would otherwise reuse the stale theme. Full suite green locally (648 tests, 0 failures).

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
```diff
@@ -0,0 +1,148 @@
...
+			$preheader = (string) get_post_meta( $post->ID, 'preview_text', true );
+			$result    = $renderer->render(
+				$render_post,
+				(string) $post->post_title,
+				$preheader,
+				(string) get_bloginfo( 'language' ),
+				'',
+				Editor_Bootstrap::TEMPLATE_SLUG
+			);
+			return isset( $result['html'] ) ? Full_Bleed_Sections::transform( (string) $result['html'] ) : '';
```
adekbadek: **Blocker:** Conditional-content guards dropped — segment-only block sends to the whole list — Blocks can carry `conditionalBefore`/`conditionalAfter` attributes — ESP merge-tag guards that show/hide a block per recipient segment. The MJML renderer wraps guarded blocks in `<mj-raw>` conditional tags (`class-newspack-newsletters-renderer.php:1819-1820, 1950-1951`) so the ESP applies the guard. `render_wc()` returns package HTML *without* these wrappers, and `retrieve_email_html()` sends the saved output verbatim — so a flag-on newsletter using conditional content delivers the guarded block **unguarded to every recipient**. Sends are irreversible, so segment-only or sensitive copy would reach readers explicitly meant not to see it. Either wrap the guarded blocks in the WC path or disable the conditional-content controls until the WC engine supports them; this must be closed before the flag is enabled for any publisher using conditional content.

plugins/newspack-newsletters/composer.json
```diff
@@ -3,7 +3,9 @@
 	"description": "Newsletter authoring plugin.",
 	"type": "wordpress-plugin",
 	"require": {
-		"campaignmonitor/createsend-php": "^7.0"
+		"campaignmonitor/createsend-php": "^7.0",
+		"automattic/jetpack-autoloader": "^5.0",
+		"woocommerce/email-editor": "2.14.0"
```
adekbadek, line 8: **Suggestion:** jetpack-autoloader may load a different email-editor version than the pinned 2.14.0 — `woocommerce/email-editor` is pinned to `2.14.0`, but WooCommerce core ships the same package via its own jetpack-autoloader, which loads the single highest version site-wide. On a Woo site the actually-loaded classes may differ from the pinned `2.14.0` that `Full_Bleed_Sections` and the block overrides target (they depend on the exact rendered markup and the renderer APIs). The render code degrades gracefully (structure mismatch → no-op, bad override → fail closed), so it's not a correctness blocker, but the compatibility assumption is implicit — document the supported package range or assert the loaded version so a future Woo bump that changes canvas markup fails loudly rather than silently producing unbanded emails. The autoloader switch also changes class loading for the whole plugin, so it's worth platform-wide smoke testing.

plugins/newspack-newsletters/includes/email-renderers/blocks/class-ad.php
```diff
@@ -0,0 +1,136 @@
...
+		// re-enters this renderer via do_blocks() and would recurse indefinitely.
+		if ( in_array( $ad_post->ID, self::$render_stack, true ) ) {
+			return '';
+		}
+
+		// do_blocks() routes inner blocks through the WC email pipeline because
+		// Content_Renderer::initialize() keeps its render_block hook active during render_wc().
+		self::$render_stack[] = $ad_post->ID;
+		try {
+			$html = (string) do_blocks( $ad_post->post_content );
```
adekbadek, line 69: **Suggestion:** WC-rendered ad links aren't routed through click tracking — The WC ad path renders the ad post via `do_blocks( $ad_post->post_content )` and returns it directly. The MJML ad path renders through `post_to_mjml_components( $ad_post )`, which calls `process_links( ..., $ad_post )` so ad URLs are rewritten through the click endpoint. WC-rendered ad links therefore won't be proxied or attributed when Newspack click tracking is on — the same class of gap as the newsletter-body link processing.

plugins/newspack-newsletters/includes/email-renderers/class-email-defaults.php
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	const DEFAULT_BUTTON_BORDER_RADIUS = '4px';
+
+	/**
+	 * Wire up wp_theme_json_data_default only when the WC renderer flag is on.
+	 *
+	 * @return void
+	 */
+	public static function init() {
+		if ( ! Feature_Flag::is_enabled() ) {
```
adekbadek: **Suggestion:** Email_Defaults reads the flag at load time — filter-based enablement misses the canvas defaults — `Email_Defaults::init()` reads `Feature_Flag::is_enabled()` at plugin-load time and returns early if off, so its `wp_theme_json_data_default` filters are wired based on the load-time value — unlike the render path, which re-reads the flag per request. A site enabling the renderer via the `newspack_newsletters_use_woo_renderer` *filter* (added on a later hook) would get the render path but not these editor-canvas defaults, so the canvas and the email would silently disagree. Resolve the flag consistently, or document that the filter mechanism only reliably drives the render side.

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
```diff
@@ -0,0 +1,148 @@
...
+		};
+		add_filter( 'the_content', $inject_content, 0 );
+
+		// Save/restore so nested render_wc() calls leave the outer post intact.
+		$previous             = self::$rendering_post;
+		self::$rendering_post = $post;
+		try {
+			$container = \Automattic\WooCommerce\EmailEditor\Email_Editor_Container::container();
+			$renderer  = $container->get( \Automattic\WooCommerce\EmailEditor\Engine\Renderer\Renderer::class );
+			$preheader = (string) get_post_meta( $post->ID, 'preview_text', true );
```
adekbadek: **Suggestion:** Preview text bypasses the get_preview_text() fallback — The preheader is read as raw `preview_text` meta. The MJML path uses `Newspack_Newsletters_Renderer::get_preview_text( $post )`, which falls back to a trimmed text version of the newsletter when the meta is empty. Flag-on newsletters without an explicit preview text will therefore show a blank inbox preheader. Route this through `get_preview_text()` for parity.

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
adekbadek: **Suggestion:** WC render skips process_links (UTM/click tracking), the open pixel, and custom_css — `render_wc()` returns the package HTML directly; unlike the MJML path it never runs `Newspack_Newsletters_Renderer::process_links()` (UTM + click-tracking link rewriting, applied at `class-newspack-newsletters-renderer.php:1964` and baked into the saved HTML), the tracking-pixel body hook, or `custom_css` injection. Because `retrieve_email_html()` hands the saved `newspack_email_html` to the ESP verbatim, a flag-on newsletter's links ship without UTM/click attribution and with no open pixel. This is a silent regression — no error, reporting just goes dark. Apply the equivalent final post-processing to the WC output before it's saved.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
```diff
@@ -0,0 +1,375 @@
...
+ * the WC renderer consumes. Non-destructive: reads meta, never writes it.
+ */
+class Theme_Json_Builder {
+	/**
+	 * Font-size scale (slug => CSS size). Mirrors Newspack_Newsletters_Renderer::get_font_size()
+	 * so presets resolve to the same pixel values as the legacy editor.
+	 *
+	 * @var array
+	 */
+	const FONT_SIZES = [
```
adekbadek: **Suggestion:** Font/spacing scales are hand-copied from the legacy renderer and will drift — `FONT_SIZES` / `SPACING_SIZES` are copied from `Newspack_Newsletters_Renderer::get_font_size()` / `get_spacing_value()` (per the docblocks). If the legacy MJML scale ever changes, the two silently diverge and the WC engine renders different sizes than MJML. Source both from one shared constant/method instead of mirroring.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
```diff
@@ -0,0 +1,375 @@
...
+					'typography' => [
+						'fontFamily' => $header_font,
+					],
+				],
+			],
+		];
+
+		// Emit email-safe button styles when the WC renderer is active. The WC email
+		// package drops CSS-var and rem values, so resolve radius and padding to px here.
+		if ( Feature_Flag::is_enabled() ) {
```
adekbadek: **Suggestion:** Theme_Json_Builder::build() branches on the global flag despite a pure-translator docblock — `build()` gates the button radius/padding emission on `Feature_Flag::is_enabled()`, but the class docblock describes it as a pure meta→theme.json translator whose output should depend only on `$post`. A concrete effect: the `test_render_wc_*` tests call `render_wc()` without enabling the flag, so `build()` silently takes the no-button-styles branch — a different path than production (flag on). Move the flag decision to the caller, or drop the inner check since `build()` is only reached during a WC render.

plugins/newspack-newsletters/newspack-newsletters.php
```diff
@@ -98,6 +122,10 @@
 require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/includes/admin/class-admin-shell.php';
 require_once NEWSPACK_NEWSLETTERS_PLUGIN_FILE . '/includes/class-wizard-bridge.php';
 
+// Boot the WooCommerce Email Editor package. Must run before the `init` hook
+// so the editor's own `init` callbacks (CPT, templates) are registered in time.
+\Newspack\Newsletters\Email_Renderers\Editor_Bootstrap::init();
```
adekbadek, line 141: **Suggestion:** Editor_Bootstrap boots the WC package unconditionally, not behind the flag — `Editor_Bootstrap::init()` runs unconditionally at plugin load — the one piece of this feature not behind `Feature_Flag`. Even with the flag off it boots the WC email-editor container on every request, opts the newsletters CPT into the WC editor (forcing the load-bearing `init:11` CPT re-registration in `class-editor-bootstrap.php`), and registers global filters (`block_type_metadata_settings`), a `sent` post status, and a block template. It's behaviorally neutral for legacy sites in practice (no editor assets load unless `woocommerce_is_email_editor_page` is true), but it means the PR's "flag off = true no-op" claim is a *behaviorally-neutralized boot*, not an absent one. Consider gating the bootstrap on `Feature_Flag::is_enabled()` (stamp-on-send and ad registration don't need the package booted), or document why the always-on boot is required and confirm the CPT opt-in is inert on flag-off sites.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -983,6 +1011,41 @@ public static function api_get_mjml( $request ) {
...
+		$post = get_post( $request['post_id'] );
+		if ( ! $post instanceof \WP_Post || ! self::validate_newsletter_id( $post->ID ) ) {
+			return new \WP_Error(
+				'newspack_newsletters_no_post',
+				__( 'Newsletter not found.', 'newspack-newsletters' ),
+				[ 'status' => 404 ]
+			);
+		}
+		$html = \Newspack\Newsletters\Email_Renderers\Renderer_Controller::render_wc( $post );
+		if ( '' === $html ) {
```
adekbadek, line 1039: **Nit:** api_get_post_html conflates render-failed with rendered-empty — `api_get_post_html` maps an empty render string to a 500 "Failed to render." `render_wc()` returns `''` both on a caught exception and on genuinely empty output, so a legitimately content-less draft previewed in the editor would surface a hard error rather than an empty preview. The WC engine always emits the template shell so `''` should only mean failure today, but the overloaded error channel is fragile if package output ever changes.

plugins/newspack-newsletters/includes/email-renderers/class-feature-flag.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class Feature_Flag {
+	const OPTION = 'newspack_newsletters_use_woo_renderer';
+
+	/**
+	 * Whether the WC renderer is enabled. Constant overrides option; filter wins last.
+	 *
+	 * @return boolean
+	 */
+	public static function is_enabled() {
```
adekbadek: **Nit:** Missing type declarations against the PR's own PHP 8.3 typing standard — `Feature_Flag::is_enabled()` lacks a `: bool` return type, and `Renderer_Controller::get_post_renderer()` / `stamp_renderer()` lack their `int`/`string` param and return types. The rest of the new email-renderer code is fully typed, so these read as oversights.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
```diff
@@ -0,0 +1,375 @@
...
+ * Builds a per-newsletter theme.json array from existing post meta.
+ *
+ * @package Newspack_Newsletters
+ */
+
+namespace Newspack\Newsletters\Email_Renderers;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
```
adekbadek, line 12: **Nit:** @package docblock tag is inconsistent across the new files — `class-theme-json-builder.php`, `class-fonts.php`, and `class-email-defaults.php` tag `@package Newspack_Newsletters`; every other new file in `email-renderers/` (and `blocks/`) uses `@package Newspack`. Pick one for consistency.

plugins/newspack-newsletters/includes/service-providers/class-newspack-newsletters-service-provider.php
```diff
@@ -459,6 +459,16 @@ public function send_newsletter( $post ) {
 
 		if ( true === $result ) {
 			Newspack_Newsletters::set_newsletter_sent( $post_id );
+			// Stamp the engine active at send time as the producing engine. Send time is
+			// the source of truth: the dispatched HTML reflects the engine resolved here,
+			// and for a scheduled send the flag is read on dispatch (not at authoring), so
+			// a flag flip between authoring and send is recorded as the send-time engine.
+			// The write is intentionally lossy toward MJML: if it fails, the resolver's
+			// default (Renderer_Controller::get_post_renderer) reports mjml.
+			\Newspack\Newsletters\Email_Renderers\Renderer_Controller::stamp_renderer(
```
adekbadek, line 468: **Nit:** stamp_renderer writes meta on every send, including legacy sends — `stamp_renderer()` runs on every successful send regardless of the flag, writing `newspack_newsletter_renderer = 'mjml'` on legacy sends that previously wrote nothing. Harmless (useful provenance) and cheap, but it's another small deviation from a strict no-op and adds an `update_post_meta` write to every send — worth being a conscious choice.

plugins/newspack-newsletters/tests/email-renderers/test-rest-post-html.php
```diff
@@ -0,0 +1,163 @@
...
+		$this->assertSame( 500, $response->get_status() );
+	}
+
+	/**
+	 * A request from an unauthorized user is rejected (401/403) — the route is gated by api_authoring_permissions_check.
+	 *
+	 * @return void
+	 */
+	public function test_post_html_route_requires_authorization() {
+		wp_set_current_user( 0 );
```
adekbadek, line 153: **Nit:** Permission test only covers the logged-out user — The permission test only exercises the logged-out user (`wp_set_current_user( 0 )`). The boundary being asserted is `edit_others_posts`, so the meaningful case — an authenticated but under-privileged user (e.g. a contributor/author who can edit their own posts but not others') — is untested. Adding it would pin the endpoint's real authorization contract.

adekbadek (changes requested):
Nice! Great to see this coming into shape. One big blocker and a slew of other comments.

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
chickenn00dle: Fixed in c4a1317dc — guarded blocks are now wrapped in their `conditionalBefore`/`conditionalAfter` merge tags via a `render_block` filter in `render_wc()`, mirroring the MJML `<mj-raw>` wrapping. It has to run at priority 20 (the package's own render_block callback at priority 10 replaces a structural block's content with its composed email table, so a same-priority wrap gets discarded). I also verified the guards survive the full-bleed DOM transform, so a segment-only *and* full-bleed band stays guarded. Covered by `test_render_wc_wraps_conditional_content_blocks` + the full-bleed variant.

plugins/newspack-newsletters/composer.json
chickenn00dle: Documented in 1c06827f4 — the `email-renderers/README.md` now spells out the pinned `2.14.0`, the "Woo core ships the same package and the autoloader loads the single highest version" risk, and the graceful-degradation behavior, with a note to smoke-test full-bleed + the overridden blocks whenever the pin (or Woo core) moves ahead. Went with documenting over a hard version assertion since the render path already fails soft (structure mismatch → no-op band, bad override → fail closed).

plugins/newspack-newsletters/includes/email-renderers/blocks/class-ad.php
chickenn00dle: Fixed in c4a1317dc — the ad renderer now runs its rendered HTML through `Newspack_Newsletters_Renderer::process_links( $html, $ad_post )` with the ad's *own* post context, so ad links get UTM + click-proxying (click tracking only proxies ad-CPT links, so it has to be the ad post, not the newsletter). The newsletter-context `process_links()` pass in `render_wc()` then skips them via its dedup. Covered by `test_ad_links_are_click_tracked`.

plugins/newspack-newsletters/includes/email-renderers/class-email-defaults.php
chickenn00dle: Fixed in 9202b2986 — `Email_Defaults::init()` now registers the `wp_theme_json_data_default` callbacks unconditionally. Each callback already re-resolves `Feature_Flag::is_enabled()` and the email-editor request context per request, so filter-based enablement on a later hook no longer misses the canvas defaults, and the filters stay inert until both are true.

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
chickenn00dle: Fixed in c4a1317dc — the WC preheader now routes through `Newspack_Newsletters_Renderer::get_preview_text( $post )`, so an empty `preview_text` meta falls back to the trimmed newsletter text and strips merge tags, matching MJML.

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
chickenn00dle: Fixed in c4a1317dc — `render_wc()` now runs a `finalize_html()` pass after the full-bleed transform that mirrors what MJML bakes in: `process_links()` (UTM + click rewriting), `custom_css` injected into `<head>`, and the open pixel injected before `</body>` (gated on `is_tracking_pixel_enabled()`, via a new shared `Pixel::get_pixel_markup()` the MJML body hook now reuses too). I also set `Newspack_Newsletters_Renderer::$newsletter_id` during the render so ad clicks attribute to the source newsletter. Covered by new UTM / custom-css / pixel-on / pixel-off tests.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
chickenn00dle: Left as-is for now. The legacy source (`get_font_size()` / `get_spacing_value()`) is private and *logic*-based, not a shared slug→size map, so deduping means refactoring the load-bearing legacy renderer — and that regression risk on the critical send path outweighs the low-probability drift of these scales. Kept the cross-reference docblocks pointing at the source of truth. Happy to extract a shared constant map as a follow-up if you'd prefer to close the gap properly.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
chickenn00dle: Fixed in 9202b2986 — dropped the inner `Feature_Flag` check so `build()` is a pure meta→theme.json translator whose output depends only on `$post`. It's only reached inside a WC render/editor context anyway, so this makes the `render_wc` tests exercise the same branch as production (flag on). Updated the test that asserted the old flag-coupled behavior.

plugins/newspack-newsletters/newspack-newsletters.php
chickenn00dle: Documented in 1c06827f4 rather than gated. The boot only registers package plumbing (container, CPT opt-in, template, theme.json/render filters); the actual editor takeover — asset enqueuing, theme.json overrides, the block allow-list, the `use_woo_renderer` editor flag — is all gated on `Feature_Flag::is_enabled()` in `Newspack_Newsletters_Editor`, so on a flag-off site the WC canvas never engages and the boot is behaviorally inert. Gating the boot itself would skip the load-bearing `init:11` CPT re-registration that keeps Newspack's CPT args authoritative after the package re-registers on `init:10`. Added a comment at the call site spelling this out.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
chickenn00dle: Left as-is — as you noted, the WC engine always emits the template shell, so `''` only means failure today. Kept the single error channel rather than adding a second signal; worth revisiting only if package output ever changes to allow a genuinely empty render.

plugins/newspack-newsletters/includes/email-renderers/class-feature-flag.php
chickenn00dle: Fixed — `: bool` on `Feature_Flag::is_enabled()` (9202b2986), and the `int`/`string` params + `string`/`void` returns on `Renderer_Controller::get_post_renderer()` / `stamp_renderer()` (c4a1317dc).

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
chickenn00dle: Fixed in 9202b2986 — `class-theme-json-builder.php`, `class-fonts.php` and `class-email-defaults.php` now all tag `@package Newspack`, matching the rest of `email-renderers/`.

plugins/newspack-newsletters/includes/service-providers/class-newspack-newsletters-service-provider.php
chickenn00dle: Kept intentionally, flagging it as the conscious choice you suggested. Writing `mjml` provenance on legacy sends is useful — it lets `get_post_renderer()` distinguish produced-by-MJML from unstamped/pre-feature — and the extra `update_post_meta` per send is cheap. Easy to gate behind the flag if you'd rather keep legacy sends a strict no-op; let me know.

plugins/newspack-newsletters/tests/email-renderers/test-rest-post-html.php
chickenn00dle: Fixed in 9202b2986 — added `test_post_html_route_rejects_under_privileged_user`, which exercises an authenticated contributor (has `edit_posts`, lacks `edit_others_posts`) to pin the endpoint's real `edit_others_posts` boundary rather than just the logged-out case.

chickenn00dle:
Thanks for the thorough pass, @adekbadek — pushed a set of commits addressing all of it. Quick rundown (details in each thread):

**The blocker + the other parity gaps are fixed in this PR (not deferred):**
- **Conditional content guards** — guarded blocks are now wrapped in their `conditionalBefore`/`conditionalAfter` merge tags in the WC render, verified to survive the full-bleed transform too.
- **`process_links` / UTM / click tracking / open pixel / `custom_css`** — a `finalize_html()` pass in `render_wc()` now bakes these in, matching what MJML saves.
- **Ad-link click tracking** — ad HTML runs through `process_links()` with the ad's own context.

**Review nits addressed:** flag-consistent `Email_Defaults`, `Theme_Json_Builder::build()` as a pure translator, preview-text fallback, type declarations, `@package` consistency, under-privileged permission test, and docs for the package-version assumption + the always-on bootstrap.

**Kept as conscious choices (reasoning in-thread):** the font/spacing scale duplication (deduping means refactoring the load-bearing legacy renderer), the render-failed-vs-empty error channel, and the send-time renderer stamp on legacy sends — all easy to change if you'd rather.

Full PHPUnit suite is green (657 tests) and PHPCS is clean. Ready for another look when you have a sec — WDYT?

plugins/newspack-newsletters/includes/email-renderers/class-renderer-controller.php
adekbadek: Verified against the branch, not just the description. The priority ordering holds: the package registers its own `render_block` callback at priority 10 (`class-content-renderer.php:163`), so the wrap at 20 receives the composed email table rather than the pre-render block markup, and it is removed in the `finally` block. The guard only fires when both `conditionalBefore` and `conditionalAfter` are present.

`test_render_wc_wraps_conditional_alignfull_background_section` is the one that matters most and it was not asked for: a guarded full-bleed background section goes through the `Full_Bleed_Sections` DOM surgery, which is exactly where the merge tags could have been dropped, and the test asserts position rather than mere presence. Full newsletters suite is green locally (657 tests, 2827 assertions, 3 skipped).

Resolving.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
adekbadek: Worth doing now rather than as a follow-up. The concern about refactoring the send path applies to the *methods*, but not to the scales: `get_font_size()` (`class-newspack-newsletters-renderer.php:162-176`) and `get_spacing_value()` (`:344-352`) each hold a plain array literal, and those literals are byte-identical to the new `FONT_SIZES` / `SPACING_SIZES`. Lifting them to two class constants and having both methods read the constant leaves every branch, fallback and return path untouched, so the load-bearing logic is not being refactored.

The reason for preferring it now: drift here fails silently and visually. A scale that diverges produces wrong-size text or spacing in a real send, with nothing to catch it.

plugins/newspack-newsletters/newspack-newsletters.php
adekbadek: The decision is acceptable for a first ship, but the documented reasoning does not hold and should be corrected, since it is what the next maintainer will rely on.

The README argues that gating `init()` would skip the load-bearing `init:11` CPT re-registration. That re-registration only exists because the bootstrap opts the newsletter CPT into the package in the first place: the package re-registers exactly the post types returned by the `woocommerce_email_editor_post_types` filter (`class-email-editor.php:187`), and that filter is added by `Editor_Bootstrap::init()` itself. Gate the bootstrap and the package never touches the CPT, so there is nothing left to re-assert.

No change requested to the code. Please reword the README so it states the actual trade (booting the package unconditionally keeps one boot path for both flag states) instead of a dependency that only exists when the bootstrap runs.

plugins/newspack-newsletters/includes/service-providers/class-newspack-newsletters-service-provider.php
adekbadek: Keep it unconditional. One `update_post_meta` per newsletter send, not per recipient, is not a meaningful cost, and stamping every send from here on means the absence of a stamp comes to mean only "pre-feature" rather than doubling as "flag was off". That provenance is worth more than strict flag-off-is-a-no-op purity, and it is what makes a later migration or a post-send investigation tractable.

Resolving as intentional.

plugins/newspack-newsletters/composer.json
adekbadek: Documenting the jetpack-autoloader highest-version-wins behaviour in `email-renderers/README.md` is an acceptable resolution. Asserting the version at runtime would trade a documented, diagnosable surprise for a hard failure on a site that is otherwise fine, which is the worse deal for a first ship. Resolving.

plugins/newspack-newsletters/newspack-newsletters.php
adekbadek: Verified the flag-off inertness claim rather than leaving it as an assertion, since the whole rollout rests on it. Both halves hold.

**Editor takeover is gated, and every gate fails safe toward legacy.** All five paths early-return to the legacy branch when the flag is off (`class-newspack-newsletters-editor.php:229, 306, 356, 535, 600`), and the stylesheet gate at `:636` additionally defaults to loading the legacy styles if the flag class is unavailable, so a fatal in the new subsystem degrades to MJML rather than to a broken canvas.

**The init:10 / init:11 CPT dance nets to zero.** Dumped the full `newspack_nl_cpt` post-type object on this branch with the flag off and diffed it against a baseline env running the plugin without this branch. The only difference is the `menu_icon` SVG, which comes from an unrelated icon-package change between the two refs. Capabilities, `show_in_rest`, `rest_base`, `rest_namespace`, rewrite rules, taxonomies, template and every other arg are identical, so the package re-registration followed by the re-assert leaves no drift.

One qualification worth putting in the README instead of "inert": the boot is not absent. `Bootstrap::init()` registers `block_type_metadata_settings` unconditionally, which adds `supports.email` and `render_email_callback` to core block type settings on every request site-wide, flag or no flag. Both are additive and unused without the renderer, and the WooCommerce coupon generator only hooks an email-render-time filter, so nothing here changes behaviour. But "behaviourally inert boot" is the accurate phrase, not "no-op".

adekbadek:
@chickenn00dle – just flagging the I pushed back in the two unresolved threads, but the rest is fine.

plugins/newspack-newsletters/includes/email-renderers/class-theme-json-builder.php
chickenn00dle: You're right — did it now. Lifted both literals to `Newspack_Newsletters_Renderer::FONT_SIZES` / `SPACING_SIZES` constants; `get_font_size()` / `get_spacing_value()` read the constants, and `Theme_Json_Builder` references them (`const FONT_SIZES = \Newspack_Newsletters_Renderer::FONT_SIZES`), so there's one source of truth and drift is structurally impossible. No branch/fallback/return path in either method changed. Added `test_scales_match_legacy_renderer_single_source` as a guard against a future re-inlined copy, and verified in a live env that both engines resolve the constants identically. Full suite green (658 tests). 5b775c539.

plugins/newspack-newsletters/newspack-newsletters.php
chickenn00dle: Good catch — the reasoning was circular and I've corrected it in 4f610f34c. You're right that the `init:11` re-assertion only exists because the boot adds the `woocommerce_email_editor_post_types` opt-in filter in the first place, so gating the boot would remove the opt-in and the re-assertion together — it's not a reason to keep the boot unconditional. Reworded both the call-site comment and the README to state the actual trade (one boot path for both flag states) and to describe the flag-off state as a **behaviourally inert boot**, not a no-op — explicitly noting that `Bootstrap::init()` still registers `block_type_metadata_settings` site-wide (adds `supports.email` + `render_email_callback`, both additive and unused until the renderer runs). Thanks also for the CPT-object diff — good to have that inertness confirmed rather than asserted.

chickenn00dle:
Thanks @adekbadek! I addressed these two points. This is ready for another round of review.

adekbadek (approved):
Both open threads are addressed.

**Font/spacing drift** (5b775c539) — the extraction is behaviour-preserving: all 13 `FONT_SIZES` and all 7 `SPACING_SIZES` entries match the deleted inline arrays exactly, and neither call site changed shape (`get_font_size()` keeps its `?? null`, `get_spacing_value()` keeps the `strpos( $value, 'var' )` guard and the `$preset . 'px'` fallback). One source of truth plus `test_scales_match_legacy_renderer_single_source` as a tripwire is exactly the fix. Small inconsistency, not worth another pass: `class-ad.php` guards its use of `\Newspack_Newsletters_Renderer` with `class_exists()` while `Theme_Json_Builder` references its constant directly — same-plugin class, so the direct reference is fine, but the two spellings imply different assumptions about its availability.

**Unconditional `Editor_Bootstrap` boot** (4f610f34c) — the documentation route was the alternative I offered, and the rewrite is the honest version of it: it retracts the circular `init:11` reasoning, states the actual trade, and describes flag-off as a behaviourally inert boot rather than a no-op, naming the residual site-wide `block_type_metadata_settings` registration instead of glossing it. That residue is worth remembering if a flag-off site ever behaves oddly, which is now what the README buys.

PR #570, opened 2026-07-08
fix(author-profile): hide default avatars when the toggle is enabled

wil-gerken:
### All Submissions:

Reference: NPPM-2626 (also resolves NPPM-163)

The Author Profile block includes a **Hide default avatar** toggle, but default avatars still appeared for authors without a photo. The same toggle on the Author List block behaves as expected.

## What this does

The Author Profile toggle now behaves exactly like Author List's. When it's on, avatars that come from Gravatar's generated fallback are hidden; avatars a publisher uploaded (the usual case for guest authors) always show. The same detection now drives the editor preview, so the editor matches the front end — which also resolves NPPM-163.

One behavior to be aware of: Gravatar serves real photos and generated placeholders from identical URLs, so telling them apart would require a remote lookup on Gravatar's servers. Matching Author List's long-standing behavior, the toggle treats any Gravatar-served avatar as hideable — in practice, "only show avatars the publisher uploaded." If per-address Gravatar detection is needed later, it can be added for both blocks together; it would introduce remote lookups and caching, so it's out of scope for this fix.

1. Create a user with no Gravatar (any made-up email address) and add an Author Profile block for them.
2. Enable **Hide default avatar** under the block's Avatar settings.
3. View the page — the gray placeholder avatar still shows.

**Verify the fix:**
1. Load the same page with this branch — the avatar is gone.
2. Turn the toggle off — the avatar shows again.
3. Add an Author List block with its own **Hide default avatar** enabled — the two blocks now match.
4. In the editor, the block preview reflects the toggle (NPPM-163).

### Other information:

## Notes for reviewers
- Root cause: detection relied on core's `avatar-default` CSS class, which core never emits when Gravatar is in play (the Gravatar branch always reports the avatar as "found"), so the check could not fire.
- The fix mirrors Author List's approach: fetch with the `blank` Gravatar fallback and match `avatar-default` OR `d=blank`, via a new shared helper `Newspack_Blocks\is_avatar_displayable()` applied at all 8 call sites (2 front-end in `author-profile/view.php`, 6 in the editor REST controller).
- The existing `avatarHideDefault` attribute already reaches every render path; only the detection changes.
- Five tests cover both render paths: the front-end helper with the toggle on and off; the `/newspack-blocks/v1/authors` route through its `author_id` and contextual `post_id` branches, which is the editor-preview path NPPM-163 reported; and a locally supplied avatar, which stays visible when the toggle is on. Full blocks suite green.
- Author List keeps its own inline copies of the same detection (`author-list/class-wp-rest-newspack-author-list-controller.php:315` and `:351`). Leaving those unchanged keeps this hotfix scoped to the Author Profile behavior NPPM-2626 reported; consolidating both blocks onto the shared helper is worth tracking separately.

### Release risk

**Medium.** This changes render output on a widely-used block, but only where the per-block toggle is enabled — and because the toggle has had no front-end effect, no publisher currently depends on it working. The one publisher-visible behavior (Gravatar-served real photos hide too when the toggle is on) matches Author List and is described above. The new helper is additive and plugin-internal; rollback is a plain revert.


plugins/newspack-blocks/tests/test-author-profile-block.php
```diff
@@ -0,0 +1,40 @@
...
+class Newspack_Blocks_Author_Profile_Test extends WP_UnitTestCase {
+	/**
+	 * When "Hide default avatar" is enabled, a user whose avatar is served by
+	 * Gravatar's generated fallback (i.e. no uploaded avatar) must not get an
+	 * avatar. Gravatar URLs never carry the `avatar-default` class core only
+	 * emits when no avatar was found, so detection must force the `blank`
+	 * fallback and match on it — mirroring the Author List block.
+	 */
+	public function test_hide_default_avatar_excludes_gravatar_fallback() {
+		$user_id = self::factory()->user->create( [ 'user_email' => 'nppm2626-no-gravatar@example.com' ] );
```
dkoo, line 48: **Suggestion:** The two new tests exercise only the `view.php` path (`newspack_blocks_get_author_or_guest_author()`) for a standard WP user. The six REST-controller call sites in `class-wp-rest-newspack-authors-controller.php` — which feed the editor preview that NPPM-163 is specifically about — and the Co-Authors Plus guest-author branch (`coauthors_get_avatar`, 6 of the 8 changed sites) are not asserted. The shared helper is transitively covered, but the controller wiring and the highest-risk CAP path are not. A single controller-level test (avatar omitted for a Gravatar-fallback user when `avatarHideDefault` is set) plus a guest-author case (the base test class already provides `create_guest_author()`) would harden exactly the paths this PR is meant to fix. Not a blocker — the core regression is covered.

plugins/newspack-blocks/src/shared/authors.php
```diff
@@ -61,3 +61,29 @@ function( $role ) {
...
+ * avatar by class alone. When hiding defaults, callers must fetch the avatar
+ * with Gravatar's `blank` fallback so it can be detected via the `d=blank`
+ * param — mirroring the Author List block's long-standing behavior.
+ *
+ * @param string|false $avatar       Avatar HTML as returned by get_avatar()/coauthors_get_avatar().
+ * @param bool         $hide_default Whether default (fallback) avatars should be hidden.
+ *
+ * @return bool True if the avatar should be displayed.
+ */
+function is_avatar_displayable( $avatar, $hide_default ) {
```
dkoo, line 80: **Suggestion:** The new `is_avatar_displayable()` helper is explicitly modeled on Author List's inline detection — its docblock cites Author List as the canonical behavior — yet Author List keeps its own copy of the `$is_default = strpos( … 'avatar-default' ) || strpos( … 'd=blank' )` logic at `class-wp-rest-newspack-author-list-controller.php:315` and `:351`. Two copies of the same detection is precisely the drift risk a shared helper is meant to remove. Consider routing those two Author List sites through `\Newspack_Blocks\is_avatar_displayable()` in this PR so detection lives in one place — they would also gain the falsy-`$avatar` guard the inline form lacks (the inline version calls `strpos()` on a possibly-`false` value). If this is intentionally deferred to keep the PR scoped, a one-line note in the description would make that explicit.

plugins/newspack-blocks/src/shared/authors.php
```diff
@@ -61,3 +61,29 @@ function( $role ) {
...
+	if ( ! $avatar ) {
+		return false;
+	}
+	if ( ! $hide_default ) {
+		return true;
+	}
+	$is_default = false !== strpos( $avatar, 'avatar-default' ) || false !== strpos( $avatar, 'd=blank' );
```
dkoo, line 87: **Nit:** Core always injects the `d=` param into the Gravatar URL and sets `found_avatar = true` whenever an email hash exists — it never checks whether Gravatar actually holds an image for that address. So with `default='blank'` forced, *every* Gravatar-served URL contains the literal `d=blank`, including for an author who has a genuine Gravatar photo. Matching the substring therefore hides real Gravatar photos too, not just generated placeholders. This is identical to Author List's long-standing behavior and the PR body documents it ("the toggle treats any Gravatar-served avatar as hideable"), so it is intentional parity rather than a regression — flagged only for awareness. Locally uploaded avatars are unaffected, since their URLs carry no `d=blank`.

plugins/newspack-blocks/tests/test-author-profile-block.php
```diff
@@ -0,0 +1,40 @@
...
+	 * With the toggle off, the avatar renders as before.
+	 */
+	public function test_avatar_present_when_hide_default_off() {
+		$user_id = self::factory()->user->create( [ 'user_email' => 'nppm2626-control@example.com' ] );
+
+		$author = newspack_blocks_get_author_or_guest_author( $user_id, 128, false, false );
+
+		$this->assertNotFalse( $author );
+		$this->assertArrayHasKey( 'avatar', $author );
+		$this->assertStringContainsString( '<img', $author['avatar'] );
```
dkoo, line 66: **Nit:** The test asserts the `avatar` key is present, but `get_avatar()` returns `false` when the `show_avatars` option is off. It defaults to on in core's schema and the WP test suite installs with schema defaults, so this passes today — but an explicit `update_option( 'show_avatars', 1 )` in the test (or a `setUp()`) would make it self-contained and immune to any suite that toggles the option off.

plugins/newspack-blocks/src/shared/authors.php
dkoo, line 80: **Nit:** The helper has no parameter or return types. A typed signature (`function is_avatar_displayable( string|false $avatar, bool $hide_default ): bool`) would be marginally stricter, but Newspack hasn't broadly adopted typed signatures, the surrounding functions in this file are untyped, and PHPCS won't flag it — so matching the file's prevailing untyped style is equally reasonable. Optional.

dkoo (approved):
**Approve with nits.** This is a clean, correctly-diagnosed fix: core omits the `avatar-default` class on the Gravatar branch, so the old check could never fire; forcing Gravatar's `blank` fallback and matching `d=blank` is the right mechanism, and it mirrors the Author List block. The new `is_avatar_displayable()` helper is boolean-equivalent to the checks it replaces, the `get_avatar()` / `coauthors_get_avatar()` `default` argument is used correctly, real uploaded avatars are unaffected, and every downstream consumer already guards a missing `avatar` key by design — no blockers.

Two things worth a glance before merge: (1) the fix's own target path — the REST/editor preview (NPPM-163) and the Co-Authors Plus guest-author branch — has no test; the two new tests only cover the WP-user `view.php` path. (2) Author List still carries its own copy of the detection logic the new helper was modeled on; folding it into the shared helper now would keep the two from drifting.

One product call to confirm: NPPM-2626 floated *removing* this toggle rather than fixing it — this PR fixes it to match Author List, which also resolves NPPM-163. Fixing is the more conservative choice, but it's worth a nod from the reporter/PM since the ticket suggested the opposite.

plugins/newspack-blocks/tests/test-author-profile-block.php
wil-gerken: Thanks @dkoo!

I added two controller-level tests. Each calls the `/newspack-blocks/v1/authors` route as an editor and asserts the toggle both on and off. `test_rest_authors_honors_hide_default_avatar` covers the `author_id` branch; `test_rest_authors_honors_hide_default_avatar_in_post_context` covers the contextual `post_id` branch, which contains three of the six changed call sites in this file. Neither needs mocks: without Co-Authors Plus, `get_post_authors()` falls back to the `post_author` lookup and the plain `get_avatar()` path at line 504. Together they cover the editor-preview path from NPPM-163 directly rather than transitively.

The guest-author case stays outside this hotfix. `coauthors_get_avatar()` is not defined in the test suite: `tests/wp-unittestcase-blocks.php` mocks the `CoAuthors_Plus` and `CoAuthors_Guest_Authors` classes, but not that function, and every CAP call site is gated on `function_exists( 'coauthors_get_avatar' )`. That branch never runs under PHPUnit, so a guest-author test would report no `avatar` key both before and after this change, and would not provide regression coverage.

CAP's source supports the expected behavior: `coauthors_get_avatar( $coauthor, $size, $default, $alt, $class )` takes `$default` third, matching the core positional contract; a guest author with an uploaded thumbnail returns from `get_guest_author_thumbnail()` before `$default` is consulted, so the uploaded image still displays; otherwise `$default` passes into `get_avatar()`. Executable regression coverage for that path still needs a `coauthors_get_avatar()` mock, which is worth tracking separately from this hotfix.

plugins/newspack-blocks/src/shared/authors.php
wil-gerken: The checks at lines 315 and 351 are verbatim copies of this detection, and the only other copies in the plugin.

Taking the documented deferral and leaving them unchanged here. Sharing the helper would change a second block's implementation on a release branch without dedicated regression coverage for that block. I have added the scope note to the description; consolidating both blocks onto the shared helper is worth tracking separately from this hotfix.

The missing falsy guard is valid, but it does not change the result today: the `if ( $avatar && ... )` on the next line short-circuits when no avatar is returned, and without strict types the `strpos()` call is coerced rather than fatal. Worth adding when the two copies are consolidated.

plugins/newspack-blocks/src/shared/authors.php
wil-gerken: This PR deliberately preserves Author List parity: the toggle hides the Gravatar fallback and keeps locally supplied avatars. Detecting whether a Gravatar actually exists for an address would need a `d=404` request plus caching across both blocks, so it stays outside this hotfix.

That comment also exposed an untested boundary. I added `test_locally_supplied_avatar_survives_hide_default`, which hooks `pre_get_avatar` to return a locally supplied image and checks that the toggle preserves it. It pins the distinction between a Gravatar fallback and a locally supplied avatar, so a later change that also hid local avatars would fail there.

plugins/newspack-blocks/tests/test-author-profile-block.php
wil-gerken: Pinned `show_avatars` in `set_up()` so it applies to both tests.

It matters most on the hide-default test: with `show_avatars` off, `get_avatar()` returns `false`, so `assertArrayNotHasKey( 'avatar', ... )` would pass without exercising the detection at all, and would pass against the pre-fix code too. The toggle-off assertion would have failed instead. Both now set the condition explicitly.

plugins/newspack-blocks/src/shared/authors.php
wil-gerken: Leaving it untyped for consistency with the file. `get_authors_roles()` and `get_authors_roles_slugs()` directly above it are also untyped, so a typed signature here would be the only one of its kind. Typing these helpers together fits a broader cleanup better than changing one function in a hotfix.

PR #572, opened 2026-07-08
fix(guest-contributors): don't email generated placeholder addresses

wil-gerken:
### All Submissions:

Reference: NPPM-2961

Guest Contributors created without an email address get a generated placeholder (`name@example.com`). Placeholder addresses can't receive mail, so anything WordPress sends them bounces.

## What this does

Email to generated placeholder addresses is no longer dispatched. The guard sits at the mail layer, so every send path is covered:

- Mail addressed **only** to placeholder addresses is skipped, and reported to the calling code as sent — the code that requested the email behaves normally; nothing is delivered.
- Mail addressed to a **mix** of real and placeholder recipients goes out to the real recipients only.
- Mail that carries **Cc or Bcc recipients** is never skipped outright, because those recipients may be real people.
- The guard decides **per send**: on by default everywhere except local and development environments (`wp_get_environment_type()`), with a `newspack_guest_author_mail_guard_active` filter — reachable from any plugin — that can force it either way. Dev containers ship with the filter set off via the Docker mu-plugin, so MailHog capture keeps working there.
- The Emails wizard's **Send test** rejects generated placeholder recipients with a clear message while the guard is active, instead of reporting a success that would never deliver.
- Real Guest Contributor emails are untouched. Only generated `@example.com` placeholder addresses are suppressed.

The placeholder check is also stricter now: only addresses that end in the placeholder domain match, so an address like `user@example.company.com` is never mistaken for a placeholder. The match is case-insensitive (`Foo@EXAMPLE.COM` counts as a placeholder). One consequence of the stricter match: a real address on a domain that merely starts with the placeholder domain (`user@example.com.au`) was previously over-hidden by the author-email display paths; it now displays where a publisher has opted into showing author emails, and receives mail. That is the intended direction — the old behavior was accidental over-hiding.

**Prerequisite:** Co-Authors Plus active, with the guest contributor assigned as the post's author through Co-Authors Plus. CAP sends the author notification to the placeholder address; without it, WordPress core excludes the guest contributor and the problem won't reproduce.

**Reproduce the problem** (works on this branch — no checkout needed):
1. Turn on Settings → Discussion → **Email me whenever anyone posts a comment**.
2. Create a Guest Contributor without an email. Confirm Newspack generates an `@example.com` placeholder address.
3. Assign the guest contributor as a post's author through Co-Authors Plus.
4. Switch the guard off, simulating the pre-fix behavior: `add_filter( 'newspack_guest_author_mail_guard_active', '__return_false' );` (a code snippet, or the same line ahead of a `wp eval` step).
5. Comment on the post. The outbox (MailHog locally) shows a notification to the `@example.com` placeholder address — in production, this is the message that bounces.

**Verify the fix:**

1. Remove the override from the previous step. On a dev container running the updated tooling the guard is off by default (the Docker mu-plugin line) — force it on for testing with `add_filter( 'newspack_guest_author_mail_guard_active', '__return_true', 20 );`. Outside local containers, no step is needed.
2. Comment on the guest contributor's post again. No notification reaches the placeholder address, and the comment posts normally.
3. Comment on a post whose author has a real email. The author still receives the notification.
4. Call `wp_mail()` (via WP-CLI) with one real recipient and one placeholder recipient. Check the outbox: only the real address receives the message.
5. Each suppression records a `newspack_log` entry under `newspack_guest_author_mail_suppressed`, with a `mode` naming the case and the removed addresses.
6. In the Emails wizard, **Send test** to an `@example.com` address is rejected with a clear message while the guard is active.

### Other information:

## Notes for reviewers
- Why the mail layer: Co-Authors Plus replaces the pluggable `wp_notify_postauthor()` with a version that applies no recipient filters and skips core's membership check — meaning there is no hook inside the notification path itself to intercept, and GCs get notified even though core would have excluded them. `pre_wp_mail` plus the `wp_mail` filter cover that path and every other send path.
- Ordering matters and is documented in the code: core applies the `wp_mail` filter **before** `pre_wp_mail`, so the stripping filter deliberately leaves an all-placeholder list intact (absent Cc/Bcc) for the short-circuit to recognize and suppress.
- The recipient value's type is left untouched unless a placeholder was actually removed. Entries that are empty after trimming are dropped — previously a trailing comma in the address list caused `wp_mail()` to return a hard failure instead of the intended silent success.
- Documented limitations: placeholder addresses inside Cc/Bcc headers are not scrubbed. Mailer plugins that deliver from their own `pre_wp_mail` callback receive the already-stripped list (core applies the `wp_mail` filter first); the all-placeholder short-circuit registers at priority 1 so such mailers cannot dispatch an all-placeholder send before it runs — effective against callbacks that honor a non-null incoming value, and pinned by a test. Plugins that replace the pluggable `wp_mail()` itself bypass both filters.
- Out of scope: guest contributors migrated before Automattic/newspack-plugin#4520 whose `user_login` contained an `@` hold a mangled placeholder (`name@domain.comexample.com`) that no suffix match can safely catch — `myexample.com` and `texample.com` are registrable domains, so any loosened match risks real addresses. Deliberately not pursued further: hand-typed fake addresses on other domains are untouchable by any placeholder-shaped repair, so a cleanup would cover only a small slice of the bad-address population.
- Local environments: the Docker mu-plugin — copied to every site root at container start — sets the activity filter off, so dev-container MailHog capture is unaffected, with no constant written and no side effects on other plugins. One-time step after updating: restart (`n start`) so containers pick up the refreshed mu-plugin.
- Each suppression emits `Logger::newspack_log()` under `newspack_guest_author_mail_suppressed` (on by default, consumed by Newspack Manager): a `mode` key naming which case fired, a count, and the removed addresses (bare, capped at 20 entries). No message content is logged.
- The placeholder domain has always been filterable via `newspack_guest_author_email_domain` (default `example.com`); the guard honors whatever domain the site generates because it reuses the existing `is_dummy_email_address()` helper.
- Thirteen mail-guard tests cover: placeholder-only suppressed and reported as sent; mixed lists stripped; real addresses untouched; trailing comma; all-placeholder with a real Cc (and Bcc) still delivering to those recipients; an empty-value Cc header not counting as recipients; end-anchored domain matching; display-name (`Name <address>`) recipients, including core-parity for bracket-bearing display names; case-insensitivity; the filterable-domain axis; the filter-registration priorities; the environment predicate; and the activity filter controlling suppression at send time. The test-send rejection has its own test in the emails suite. Full plugin suite green (2,360 tests).

### Release risk

**High.** This intercepts the outbound-mail path of the main Newspack plugin, so the code runs for every email on every publisher — that reach is what drives the tier. Mitigations: suppression is keyed to the reserved `example.com` placeholder domain with an end-anchored match; each named failure mode (over-suppression, recipient-type change, Cc suppression, trailing comma) has a dedicated test; behavior for real addresses is unchanged and covered by tests. Rollback is a plain revert — suppression stops and bounces resume.

The guard is additionally inert on dev containers via the mu-plugin override, so developer mail capture is unaffected. Production behavior differs from the originally reviewed design in four reviewed ways: the earlier `pre_wp_mail` priority, per-send gate evaluation, on-by-default suppression logging, and the test-send recipient check.

This ships as a hotfix to `release`: the bounce problem is active and every named failure mode has a dedicated test.


plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -87,6 +87,12 @@ public static function initialize() {
 		\add_filter( 'theme_mod_show_author_email', [ __CLASS__, 'should_display_author_email' ] );
 		\add_filter( 'newspack_show_coauthor_email', [ __CLASS__, 'should_display_coauthor_email' ], 10, 2 );
 
+		// Never send email to generated placeholder addresses — they bounce and
+		// can get the site's outbound email blocked (e.g. comment notifications
+		// to Guest Contributors created without a real email).
+		\add_filter( 'pre_wp_mail', [ __CLASS__, 'short_circuit_dummy_only_email' ], 10, 2 );
```
chickenn00dle: **Suggestion:** Silently blackholes mail on every local dev environment — Newspack's own tooling provisions everything on the default dummy domain: `bin/site-setup.sh:205` (the admin), `:348` (all generated users), `:460` (WooCommerce customers), plus `bin/env.sh:527`, `default.env:29`, `e2e/site-setup.sh` and `e2e/tests/utils.ts:8`.

Once this ships, every local and isolated environment stops delivering mail to MailHog entirely — and because the send is reported as successful there is no error, no bounce and no log line indicating anything was withheld. Reader Activation magic links, newsletter test sends and WooCommerce order mail against fixture users all go quiet at once. The symptom ("mail stopped arriving in the container") points nowhere near the cause.

CI is fine: the e2e suite installs its own `pre_wp_mail` capture that runs ahead of this guard and records every message regardless, and no spec asserts on mail today. So this is a developer-experience regression rather than a correctness one — but a patch release gets less bake time than usual, and this is the sort of thing that costs someone an afternoon.

Pointing the fixture defaults at a non-reserved domain in the setup scripts is the cheapest fix.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -279,10 +285,157 @@ public static function get_dummy_email_domain() {
 	/**
 	 * Is a dummy email address?
 	 *
+	 * The match is end-anchored so addresses on domains that merely contain
+	 * the dummy domain as a prefix (e.g. user@example.company.com) are not
+	 * mistaken for placeholders.
+	 *
 	 * @param string $email_address Email address to check.
 	 */
 	public static function is_dummy_email_address( $email_address ) {
```
chickenn00dle, line 311: **Suggestion:** Legacy placeholder addresses aren't matched — and they're the ones still bouncing — Before #4520 (`cec5c6893`, Feb 2026), a guest contributor whose `user_login` contained an `@` — common in migrations, which is why that fix exists — got a double-`@` placeholder that `sanitize_email()` then mangles. Verified against core: `john.doe@gmail.com@example.com` becomes `john.doe@gmail.com` + `example.com` concatenated as `john.doe@gmail.comexample.com`.

That matches **neither** the old `strpos` check nor the new end-anchored one. So those users are still emailed — to a nonexistent domain, which is precisely the bounce this PR sets out to stop — and their fake address is still displayed publicly wherever the site opts into showing author emails.

#4520 fixed generation but shipped no backfill, so existing sites still hold this data. Not introduced here, and reasonable to treat as out of scope — but a change titled "never send email to placeholder addresses" arguably owns it. Since username generation strips those characters now, the affected accounts can only have come from migrations, so a one-pass repair is feasible: regenerate the placeholder for any guest contributor whose address doesn't parse as one but was clearly meant to be. Either file that follow-up or note the cohort as out of scope in the PR description.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -279,10 +285,157 @@ public static function get_dummy_email_domain() {
...
+	 * Core applies the 'wp_mail' filter before this one, so the recipient
+	 * list seen here has already been stripped by
+	 * remove_dummy_email_recipients() — which deliberately leaves an
+	 * all-dummy list intact (absent Cc/Bcc) so it can be recognized and
+	 * suppressed here.
+	 *
+	 * Scope notes: mail whose headers carry Cc/Bcc recipients is never
+	 * short-circuited — suppressing it would also suppress delivery to those
+	 * (possibly real) header recipients. Dummy addresses inside Cc/Bcc headers
+	 * are not scrubbed by this guard. Mailer plugins that take over delivery
```
chickenn00dle, line 391: **Suggestion:** Guard is pre-empted by mailer plugins — and this docblock has the mechanism backwards — Two related things here.

The closing sentence — *"Mailer plugins that take over delivery from their own `pre_wp_mail` callback bypass the `wp_mail` filter below, so mixed-list stripping does not apply there"* — is inverted. Core applies `wp_mail` at `pluggable.php:209` **unconditionally, before** any `pre_wp_mail` callback runs, so mixed-list stripping *does* apply and such a mailer receives an already-stripped `$atts`.

The real caveat is the opposite one, and it isn't documented. The **all-dummy** case depends entirely on this plugin's own `pre_wp_mail` callback, because `remove_dummy_email_recipients()` deliberately leaves the full dummy list intact for it. Any mailer plugin returning non-null from `pre_wp_mail` at a lower priority — or at 10 but registered earlier, which is plugin load order and not something this controls — pre-empts `short_circuit_dummy_only_email()`, which then correctly defers via its `null !== $return` guard. The placeholder is mailed anyway. That is the reported bug, unfixed, on exactly the sites most likely to have hit it.

Registering the `pre_wp_mail` filter at priority `1` closes this for a one-character change — for an all-dummy list there is no legitimate delivery for any mailer to perform. Given how much of the correctness argument rests on this docblock, the wrong sentence is worth fixing too, along with a note that plugins redeclaring the pluggable `wp_mail()` outright bypass both filters and leave the guard inert.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -279,10 +285,157 @@ public static function get_dummy_email_domain() {
...
+		}
+		if ( empty( $atts['to'] ) ) {
+			return $return;
+		}
+		if ( self::has_cc_or_bcc_headers( $atts['headers'] ?? '' ) ) {
+			return $return;
+		}
+		$recipients = self::parse_recipients( $atts['to'] );
+		if ( ! empty( $recipients ) && empty( self::remove_dummy_addresses( $recipients ) ) ) {
+			return true;
```
chickenn00dle, line 430: **Suggestion:** Registration is unconditional and suppression is domain-only — `initialize()` runs at file load, included unconditionally, so both mail filters are live on every Newspack site — whether or not Co-Authors Plus is active, and whether or not the site has ever had a guest contributor. The class already tracks the latter via its `SITE_HAS_GUEST_AUTHORS_OPTION_NAME` option.

The suppression decision then rests solely on the address ending. Nothing checks that the recipient is actually a guest contributor. Under the default that's harmless — `example.com` is RFC 2606 reserved and undeliverable, which is genuinely what makes this defensible as a hotfix.

But `newspack_guest_author_email_domain` is a public filter, and its meaning has quietly widened: it used to answer "what domain do we generate placeholders on", and now it also answers "what domain does this site refuse to email". Pointing it at a domain the publisher owns is a natural thing to try — a real domain contains bounces better than a reserved one — and at that point all mail to that domain is dropped site-wide and reported as sent, including to real mailboxes. Nothing in the monorepo or `repos/` filters it today, so this is latent rather than live.

Two cheap options: gate registration on Co-Authors Plus being active (proportionality), and/or only suppress when the address both matches the dummy domain **and** belongs to a user with the guest-contributor role — that lookup only runs for addresses already matching, so it costs nothing on a default site.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
chickenn00dle, line 311: **Nit:** Two cross-repo callers now diverge from this implementation — `is_dummy_email_address()` is called from `newspack-manager` in two places that still use the old loose `strpos` logic and so no longer agree with this one:

- `includes/pugpig/modules/class-feed.php:229` uses it to decide whether an author's address is printed **into a public feed**. Worth knowing that unlike the two display paths in this plugin, that one has no guest-contributor gate — it applies to ordinary WordPress users too. The narrowing is still the correct direction, and the feed only prints addresses when the publisher has opted in, but a `.com.au`-shaped address that was previously withheld would now appear.
- `tests/mocks/class-guest-contributor-role.php:35` mirrors the old behaviour, so manager's tests now assert against a stale contract.

Both belong in follow-ups against that repo rather than in this PR — flagging them so the divergence is deliberate rather than discovered later.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
chickenn00dle, line 430: **Nit:** Suppression leaves no trace — and the send log says "Sent" — Returning `true` means core fires neither `wp_mail_succeeded` nor `wp_mail_failed`, so a suppressed message is invisible to everything downstream. `newspack-manager`'s outbound-mail alarm listens on `wp_mail_failed` (`includes/class-mail.php:40`) and never sees it, while `Emails::send()` logs `Sent "…" email to: …` (`includes/emails/class-emails.php:710`) — so the only trace that exists actively asserts the opposite of what happened.

A `Logger::log()` breadcrumb at the suppression point costs nothing in production (it's gated by `NEWSPACK_LOG_LEVEL`) and turns "mail vanished" into a one-line answer. It also largely defuses the local-environment problem above.

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
```diff
@@ -338,4 +338,90 @@ public function test_is_dummy_email_address() {
 			'Should not identify real email as dummy.'
 		);
 	}
+
+	/**
+	 * Outbound mail to generated dummy addresses must be suppressed entirely.
+	 */
+	public function test_mail_to_dummy_address_is_blocked() {
```
chickenn00dle, line 352: **Nit:** Test gaps on the riskiest parsing paths — The six new tests cover the important ground — all-dummy suppression, mixed lists, real mail untouched, end-anchoring, empty entries, and the Cc branch. Three gaps, roughly in risk order:

1. The `<...>` display-name extraction in `remove_dummy_addresses()` is untested. `Guest Author <guest@example.com>` is the form most likely to appear in real notification code, and that regex is the most likely thing to be subtly wrong.
2. `Bcc:` is untested even though the header regex is `b?cc` and Bcc carries different semantics from Cc.
3. Case-insensitivity is untested, and it's a behaviour *widening* introduced here — `Foo@EXAMPLE.COM` was not a placeholder before and is now.

Also worth acting on the point Copilot raised: the tests hard-code `example.com` rather than using `get_dummy_email_domain()`. That matters more than it looks, because the filtered-domain path is exactly the axis the registration concern above turns on — as written, nothing exercises it at all.

chickenn00dle (commented):
**Release risk: ELEVATED** — no blockers. The design is sound and I'd ship it; two things below are worth resolving first, and neither is a large edit.

Starting with what holds up, since a lot of this design's correctness argument lives in its docblocks and I wanted to check rather than take them on trust. The ordering claim is right: core applies the `wp_mail` filter at `pluggable.php:209`, before `pre_wp_mail` at `:233`, so leaving an all-dummy list intact for the short-circuit to recognise genuinely works. The Cc branch is safe too — core's recipient loop skips empty address lists, so emptying `to` with a Cc present sends to the Cc rather than erroring. Recipient parsing only comma-splits strings, which is exactly what core does, so a display name containing a comma inside an array entry is never corrupted. And "report as sent" breaks no caller: Guest Contributors already have password reset and authentication disabled, so there's no flow waiting on mail that will now never arrive.

I also checked whether narrowing `is_dummy_email_address()` could expose an address that used to be hidden. It can't meaningfully: the only addresses that change classification are real ones on domains that merely start with the dummy domain (`@example.com.au`, `@example.community`), and both display paths additionally require the guest-contributor role and the publisher's explicit opt-in. The old behaviour was accidental over-hiding, so this is a fix. Worth a line in the PR description so it isn't a surprise.

**The guard can be pre-empted, and that's where the reported bug lives.** `pre_wp_mail` is registered at priority 10, so any mailer plugin whose own `pre_wp_mail` callback runs first — lower priority, or the same priority registered earlier, which is load order and not something this plugin controls — returns non-null and this guard correctly defers. The placeholder then gets mailed anyway. That's NPPM-2961, unfixed, on sites running an SMTP plugin. Priority `1` closes it, and it's safe: the only other code in this repo on that hook records unconditionally and ignores the return value, so nothing is trampled. Worth also noting in the docblock that plugins which redeclare the pluggable `wp_mail()` outright bypass both filters entirely.

While there — the docblock's closing sentence has the mechanism backwards. It says mailer plugins short-circuiting from `pre_wp_mail` "bypass the `wp_mail` filter below, so mixed-list stripping does not apply there." Core applies `wp_mail` unconditionally first, so stripping *does* apply and those mailers receive an already-stripped list. The real caveat is the pre-emption above.

**This silently disables mail on every local environment.** Newspack's own tooling provisions the admin, every generated user and every WooCommerce customer on `@example.com`. Once this ships, local envs stop delivering to MailHog while `wp_mail()` keeps returning `true` — no error, no log line, no signal at all. CI is unaffected (the e2e suite's own capture runs ahead of this guard and records everything regardless), so this is a developer-experience problem rather than a correctness one, but it's a confusing one to debug from the symptom. Pointing the fixtures at a non-reserved domain in the setup scripts is the cheap fix, and worth landing alongside this given a patch release gets less bake time.

The remaining points are smaller and inline: a legacy address shape this guard doesn't reach (and which is the half that actually bounces today), the filter whose meaning has quietly widened from "what domain do we generate" to "what domain do we refuse to email", the absence of any trace when a message is dropped, and a few gaps in otherwise good test coverage.

One note for the ticket rather than the code: the reported publisher's authors have arbitrary fake addresses, not placeholder-domain ones, so this doesn't stop their bounces by itself — it makes the recommended conversion to Guest Contributors safe to do. Probably worth not closing NPPM-2961 for that site until the conversion happens.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Fixed. The short-circuit registers at priority 1. The scope notes now state the correct direction: the strip reaches pre-empting mailers, and the all-placeholder short-circuit is what has to run first. The docblock also notes that plugins replacing the pluggable `wp_mail()` implementation bypass both filters. A test asserts both callback priorities, so the ordering stays an explicit contract.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Proposal: gate the guard on environment type instead of changing fixture domains.

Why not the fixture swap: it only helps environments created after the change. Existing local sites already have `@example.com` users in their databases, and their mail would stay suppressed until each site was recreated. So the guard now skips registration when `wp_get_environment_type()` is `local` or `development`. Staging keeps protection, since it delivers real mail. In the standard local setup mail terminates in MailHog, so there are no bounces to prevent — suppression only hides mail developers are trying to inspect.

The tooling half: the main site and isolated envs never set `WP_ENVIRONMENT_TYPE` (additional sites already set it at creation). `init-wp.sh` now writes it on every container start, only when unset; `n pull` writes it to already-running containers under the same condition. An explicit value is never overridden, so if you flip an env to `production` to test the guard, that flip sticks.

The trade-off of declaring `local`: every plugin that reads the environment type gets the truth, and a few change behavior on dev sites — production and staging are untouched, since nothing writes the constant there. Checked against the plugins actually installed:

- **This guard** — off locally; MailHog testing works again (the point).
- **Manager's Perfmatters restriction** — settings screens unlock for all local admins, the carve-out manager already ships for local envs. Perfmatters' own behavior is unchanged.
- **WooCommerce Subscriptions** — its opt-in customer notification reminders (renewal/expiry/trial-ending) stop generating off-production; other subscription mail is unaffected.
- **Site Kit** — stops printing GA/AdSense tags locally, so dev traffic stops polluting real analytics; local tag-output testing needs the production flip.
- **Stripe gateway** — deleting a saved payment method locally no longer detaches it from the live Stripe account (its own clone-site protection).
- **Jetpack** — counts the site as local in its status heuristics.
- Query Monitor, Health Check, and Plugin Check only display the value.

Not writing it keeps the guard active in every existing local env, where placeholder-domain mail reports success and never reaches MailHog. Escape hatches both ways: `wp config set WP_ENVIRONMENT_TYPE production` flips a whole site back for testing (an explicit value is never overridden), and the `newspack_guest_author_mail_guard_active` filter flips just the guard. WDYT?

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: The description now names the cohort created before the placeholder format was corrected in Automattic/newspack-plugin#4520 as out of scope, and we're deliberately stopping at the note rather than filing the repair. Two reasons. The affected accounts' siblings — hand-typed fake addresses on arbitrary other domains — are untouchable by any placeholder-shaped repair, so a cleanup would cover only a small slice of the bad-address population. And the natural repair predicate (address ends with `example.com` but not `@example.com`) can match real domains — `user@texample.com` — so a safe pass would need exact reconstruction from `user_login`, more machinery than the slice justifies.

Loosening the runtime match fails the same way (`myexample.com` is registrable), so the matcher stays end-anchored.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Kept domain-only suppression. Two findings drove that, and the filter widening is answered separately below.

**The CAP gate.** The role system and placeholder generation both run without Co-Authors Plus (CAP). A site that deactivates CAP keeps its placeholder-emailed users, so the gate would switch protection off while the risky data persists. `SITE_HAS_GUEST_AUTHORS_OPTION_NAME` also tracks CAP's legacy guest-author CPTs, not `contributor_no_edit` users, so a gate built on it keys on a different population than it needs.

**The role lookup.** Mail to `@example.com` is undeliverable regardless of the account's role — a placeholder admin address is equally undeliverable, and a deleted Guest Contributor's lingering address would evade a role check.

**The filter widening.** The docblock now states the contract at the point where a developer would set the filter: the domain is treated as unroutable and outbound mail to it is suppressed, so point it only at a reserved domain. Every suppression also writes a log line. A search of the monorepo and the current `repos/` checkouts found no uses of the filter; that doesn't rule out external consumers, so the documented contract plus logging match the currently observed scope. Open to revisiting if you see a gap.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Deliberate, and now tracked: NPPM-3100 covers syncing the test mock and the feed's plugin-absent fallback branch to the new match. The feed's primary path calls this class directly, so a site picks up the new behavior once the newspack-plugin release carrying this PR reaches it; only the mock and the fallback stay divergent.

The manager sync PR (tracked in that ticket) will document the feed-output change with the concrete case: the old substring match treated a real address like `author@example.com.au` as a placeholder and withheld it from `<author_email>`. The end-anchored match does not, so the feed now includes that address when a publisher has enabled author emails.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Added log lines at all three decision points: short-circuiting an all-placeholder recipient list, removing placeholder recipients from a mixed list, and stripping a recipient list down to Cc/Bcc-only delivery. Each entry records the removed recipients — placeholder addresses by construction — under a dedicated `NEWSPACK-GC-MAIL-GUARD` header so the set is greppable. Subjects stay out of the logs, so no message content lands there.

I kept it at `Logger::log()`; the `newspack_log` action would give manager fleet visibility, and if we want suppression-rate metrics later, a structured event can be added separately as follow-up work.

In production the entries are gated by `NEWSPACK_LOG_LEVEL`, the usual switch for Newspack debugging aids. If the environment gate proposed in the local-env thread holds, day-to-day local development never reaches any of this, since the guard doesn't register there.

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
wil-gerken: All four gaps are covered:

- Display-name extraction (`Name <address>`), for suppression and mixed-list stripping.
- Bcc, mirroring the Cc test.
- Case-insensitivity, pinned as intended widening.
- The filter axis: a new test filters the domain to `placeholder.invalid`, asserts suppression follows it, and asserts `@example.com` mail flows again.

On Copilot's version of the last one: I kept the existing addresses hard-coded so those tests independently pin the shipped `example.com` default — deriving them from `get_dummy_email_domain()` would validate the code against itself. The new test owns the filter axis. (A wiring test was added too: a false `newspack_guest_author_mail_guard_active` filter must prevent registration, and re-registration must restore both contract priorities.)

wil-gerken:
Thanks for the detailed review, @chickenn00dle! I've resolved five threads below, but two are open for discussion:

- **Local-environment mail**: I took a different approach than the fixture-domain swap — the guard is now environment-gated. This one is undecided and needs your take; details in the local-env thread.
- **Suppression scope**: stays domain-only rather than role-gated, and the filter contract is now documented. Decided on my end, but the thread stays open in case you see a gap in the reasoning.

Neither open thread blocks the changes as implemented; if either discussion lands differently, we'll adjust before merge.

Ordering: the short-circuit now registers at priority 1, with a test asserting the callback order.

The newspack-plugin suite passes (1,653 tests) and PHPCS is clean.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: A data point that arrived on its own while this thread was open: CI failed on this branch's merge with current release, because a test suite that landed on release uses `tester@example.com` as a deliverable fixture — the guard suppressed the send, which is exactly the collision class this thread is about. Fixed in 0762923f2 by moving those fixtures to `tester@example.org`: delivery-expecting fixtures can't live on the domain the guard makes unsendable.

bin/init-wp.sh
```diff
@@ -48,6 +48,17 @@ if [ ! -f /var/www/html/wp-config.php ]; then
...
 fi
 
+# Declare the environment type (set only when absent — an explicit value in
+# wp-config.php or the container environment is never overridden) so
+# environment-gated behavior, e.g. the newspack-plugin outbound-mail guard,
+# can tell local development from production. Runs on every container start
+# so existing installs pick it up too. To exercise environment-gated
+# production behavior locally, set the constant to production for the site
+# and restore local afterward.
+if [ -f /var/www/html/wp-config.php ] && [ -z "${WP_ENVIRONMENT_TYPE:-}" ] && ! wp --allow-root config has WP_ENVIRONMENT_TYPE 2>/dev/null; then
```
chickenn00dle: **Suggestion:** The workspace already has a scoped mechanism for this — no need to redeclare the environment type — Answering the question directly, since you flagged this as proposed-not-settled: I'd take neither the fixture swap nor this.

`bin/newspack-docker-mu.php` is the workspace's designated "make this container behave like a dev box" mu-plugin, and it already carries exactly this class of override — `googlesitekit_allowed_tag_environment_types`, `woocommerce_subscriptions_is_duplicate_site` → `__return_false`, `auto_update_plugin`/`auto_update_theme` → `__return_false`. Two of those are for the very plugins whose behavior you listed as collateral from declaring `local`. A single line there:

```php
add_filter( 'newspack_guest_author_mail_guard_active', '__return_false' );
```

makes the guard inert on every dev container with none of the blast radius.

It is also strictly more coverage than this branch achieves. I checked: that file is copied unconditionally on every container start — not inside a wp-config-creation branch — to all three site roots (`bin/init-wp.sh:69`, `bin/init-wp-manager.sh:48`, `bin/init-additional-site.sh:50`), so it reaches manager-html and additional sites, which the `WP_ENVIRONMENT_TYPE` backfill does not (see the next comment). It applies retroactively to existing envs on the next container start, same as this hook. And because it's mu-plugin scope, the filter is actually reachable there — which is not true from most other places (see the comment on `register_mail_guard()`).

The payoff is that both this hunk and the `n pull` loop can be dropped, making this a plugin-only hotfix. On a patch release to `release`, that's worth a lot: the tooling half has a delivery path entirely separate from the plugin half, so today there's a window where a developer has the new plugin but not the new `n`, and their local mail goes quiet — the exact problem the gate exists to prevent.

There is a fair counter-argument that an undeclared environment type resolving to `production` is simply wrong for a Docker container, and that `bin/init-additional-site.sh:39` already declares `local` so this brings the main site to parity. I agree with that — but it's a standalone tooling improvement that deserves its own change, not a dependency of a hotfix.

n
```diff
@@ -467,6 +467,18 @@ case $1 in
...
+        # Declare the environment type on running containers (set only when
+        # absent — an explicit value in wp-config.php or the container
+        # environment is never overridden) so environment-gated behavior,
+        # e.g. the newspack-plugin outbound-mail guard, recognizes local envs
+        # without waiting for the next container restart.
+        for c in $(docker ps --format '{{.Names}}' 2>/dev/null | grep -E '^newspack_dev$|^newspack_env_'); do
+            if docker exec "$c" sh -c "[ -n \"\${WP_ENVIRONMENT_TYPE:-}\" ] || wp --allow-root config has WP_ENVIRONMENT_TYPE" > /dev/null 2>&1; then
+                continue
+            fi
+            docker exec "$c" sh -c "wp --allow-root config set WP_ENVIRONMENT_TYPE local --type=constant" > /dev/null 2>&1 \
```
chickenn00dle: **Suggestion:** The environment-type backfill only reaches the main site — If the environment gate stays as-is, the tooling half doesn't cover the whole local surface. Both this loop and the `bin/init-wp.sh` block run at the container's `WORKDIR`, which I confirmed is `/var/www/html` in every container — so `wp config has` / `wp config set` resolve the main site and nothing else.

That leaves two gaps:

- **manager-html** — `bin/init-wp-manager.sh` never sets `WP_ENVIRONMENT_TYPE` at all (zero occurrences), so it resolves to `production` and will suppress placeholder mail locally.
- **Pre-existing additional sites** — `bin/init-additional-site.sh:39` does set it, but only inside the `if [ ! -f .../wp-config.php ]` creation branch (lines 11–41). Any additional site already created never picks it up. That's precisely the limitation this PR just fixed for the main site by hoisting the assignment out of the creation branch.

Adding the constant to `init-wp-manager.sh` and hoisting the additional-site one out of its creation branch would make local behavior uniform. Adopting the mu-plugin approach from the previous comment makes all of this moot, since that file is already copied to all three site roots.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -273,16 +278,235 @@ public static function setup_custom_role_and_capability() {
...
+		 *
+		 * Defaults to active everywhere except local and development
+		 * environments (see is_mail_guard_active_for_environment()). Hook from
+		 * a must-use plugin or a wp-config-included file to force the guard on
+		 * or off — for example, to re-enable it on a site whose declared
+		 * environment type would otherwise switch it off.
+		 *
+		 * @param bool $active Whether the guard registers its mail filters.
+		 */
+		if ( ! \apply_filters( 'newspack_guest_author_mail_guard_active', self::is_mail_guard_active_for_environment( \wp_get_environment_type() ) ) ) {
```
chickenn00dle: **Suggestion:** The escape hatch is only reachable from a must-use plugin, and the docblock's wp-config route is impossible — `initialize()` is invoked at the bottom of this file, which `class-newspack.php` includes during plugin bootstrap, so this `apply_filters()` fires before ordinary plugins and themes load.

The docblock offers two routes, and one of them can't work: `add_filter()` is undefined at `wp-config.php` time, because `wp-config.php` runs before `wp-settings.php` loads `wp-includes/plugin.php`. A file included from wp-config would fatal, not filter. (The only wp-config-time option is hand-populating `$wp_filter` via `WP_Hook::build_preinitialized_hooks()`, which nobody is going to reach for.) So the realistic audience is mu-plugins only.

That's a steep requirement for the scenario the docblock itself names as the failure mode — a production site whose declared environment type silently switches the protection off. Two ways to soften it, either of which preserves current behavior exactly:

- Defer registration to `plugins_loaded`, so a normal plugin or `functions.php` can reach the filter; or
- Register unconditionally and evaluate the gate inside `short_circuit_dummy_only_email()` / `remove_dummy_email_recipients()`. The per-call cost is negligible next to a mail send, and `wp_get_environment_type()` memoizes.

At minimum, trim the docblock so it doesn't point people at a route that fatals. Worth noting a wp-config *constant* would fit this shape well and matches the convention already used in this file for `NEWSPACK_ENABLE_CAP_GUEST_AUTHORS`.

**Risk if not addressed:** If a publisher points the placeholder domain at a domain they actually own — a natural thing to try, since bounces to your own domain are easier to contain than bounces to a reserved one — then every address on that domain silently stops receiving mail site-wide, including magic-link logins and subscription notices, and switching it back off requires deploying a must-use plugin because the escape hatch runs too early for any other code to reach it.

**Consider:** Keying suppression on something narrower than the domain alone — confirming the recipient actually holds the guest-contributor role, or that the domain is one of the RFC 2606 reserved ones that can never receive mail — would stop a repointed filter from being a site-wide mail outage waiting to happen, and would mean the docblock warning is no longer the only thing standing between a publisher and silent mail loss.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -273,16 +278,235 @@ public static function setup_custom_role_and_capability() {
...
+		}
+		if ( empty( $atts['to'] ) ) {
+			return $return;
+		}
+		if ( self::has_cc_or_bcc_headers( $atts['headers'] ?? '' ) ) {
+			return $return;
+		}
+		$recipients = self::parse_recipients( $atts['to'] );
+		if ( ! empty( $recipients ) && empty( self::remove_dummy_addresses( $recipients ) ) ) {
+			Logger::log( 'Suppressing outbound email to placeholder-only recipient list: ' . implode( ', ', $recipients ), 'NEWSPACK-GC-MAIL-GUARD' );
```
chickenn00dle: **Suggestion:** The audit trail is switched off on exactly the sites where the guard fires — These three log lines are the right idea, but `Logger::log()` returns immediately unless `NEWSPACK_LOG_LEVEL` is defined and greater than zero, and that constant isn't defined by any tracked tooling in this repo — I checked. It's absent in isolated envs and set by hand only in my own main dev container's `wp-config.php`. The codebase's own comment at `class-newspack.php:553` describes the gate as what keeps "production quiet."

So on the sites where the guard actually fires, these lines produce nothing, and the first symptom a publisher reports — "the contributor never got the notification" — still has no trace to correlate against. The message never enters the mail system, so it's absent from SMTP and delivery logs too.

There's a pointed precedent a few files over. The comment above `Emails::dispatch_email()`'s logging (`includes/emails/class-emails.php:704`) explains that the old unconditional "Sending…" log "produced misleading success-shaped lines for wp_mail failures and routinely led incident investigation down the wrong path." Suppression reintroduces exactly that condition for placeholder recipients: line 710 logs `Sent "…" email to: …` off a `true` that means "withheld."

If the trail is meant for support triage rather than only local debugging, it needs a channel that's on by default — `Logger::newspack_log()`, or a `do_action` other code can subscribe to.

Two smaller things while you're here. The payload interpolates the raw recipient entries, so `Guest Author <someuser@example.com>` writes the contributor's display name into the log; that's a public byline rather than sensitive data, and the mixed-list branch correctly logs only the removed placeholders via `array_diff` so real recipients are never logged — but stripping to the bare address would cost nothing. And the `'NEWSPACK-GC-MAIL-GUARD'` header is an inline literal repeated at three call sites; the convention here (`AGENTS.md`, and `Data_Events::LOGGER_HEADER`, `Idempotent_Send::LOGGER_HEADER`, `Contact_Cron::LOGGER_HEADER`) is a `LOGGER_HEADER` class constant, which also gives operators a symbol to reference.

**Risk if not addressed:** If a publisher reports that a contributor never received a notification, there is nothing to correlate against: the message never enters the mail system so it is absent from SMTP and delivery logs, the calling code records it as sent, and the guard's own breadcrumb is switched off unless someone had already enabled debug logging before the fact. The delivery outcome for these addresses is unchanged — they bounced before — but the bounce was the last remaining signal that anything was wrong, and suppression removes it.

**Consider:** Emitting the suppression through a channel that is on by default — `Logger::newspack_log()`, or a `do_action` other code can subscribe to — would turn an otherwise undiagnosable support report into a one-grep answer.

plugins/newspack-plugin/tests/unit-tests/emails.php
```diff
@@ -324,15 +324,15 @@ public function test_test_send_works_for_draft() {
 
 		$admin = self::login_as_admin();
 		try {
-			$result = Emails::send_test_email( $test_email['post_id'], 'tester@example.com' );
+			$result = Emails::send_test_email( $test_email['post_id'], 'tester@example.org' );
```
chickenn00dle, line 327: **Suggestion:** The Emails wizard's test-send now reports success for a placeholder address and delivers nothing — Moving these fixtures to `example.org` is the right fix, and the CI failure that prompted it was a useful signal. Worth naming what it demonstrates, though, because there's a live surface with the same shape.

`Emails::send_test_email()` returns success whenever `dispatch_email()` returns truthy, so an admin who types the canonical throwaway address into the wizard's "Send test" field sees a success notice and receives no mail — with no log line by default, per the observability comment above.

That address would have bounced without the guard too, so this isn't a deliverability regression, only a feedback one. But the test-send is precisely the surface an admin or a support engineer reaches for to prove outbound mail is working, and reporting success there certifies a broken configuration as healthy. A note in the release notes or on the ticket would head off the support round-trip; a recipient check in the test-send path that reports "suppressed, not delivered" would close it properly.

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
```diff
@@ -338,4 +338,227 @@ public function test_is_dummy_email_address() {
...
+	/**
+	 * The guard's filter registrations are a contract: the pre_wp_mail
+	 * short-circuit runs at priority 1, before mailer plugins' own callbacks
+	 * (typically priority 10) could dispatch an all-dummy send.
+	 *
+	 * Precondition (shared by every mail-guard test in this file): the test
+	 * bootstrap defines no WP_ENVIRONMENT_TYPE, so wp_get_environment_type()
+	 * resolves to 'production' and initialize() registered the guard.
+	 */
+	public function test_mail_guard_filter_priorities() {
```
chickenn00dle, line 442: **Suggestion:** The stated precondition depends on ambient environment state — The docblock's precondition — "the test bootstrap defines no `WP_ENVIRONMENT_TYPE`, so `wp_get_environment_type()` resolves to `production`" — is true for the *constant*, but core also reads the value from `getenv( 'WP_ENVIRONMENT_TYPE' )`. This very branch adds a `[ -z "${WP_ENVIRONMENT_TYPE:-}" ]` check in both `bin/init-wp.sh` and `n` precisely because that environment variable may be set on a container.

So if anyone exports it — a `docker-compose.override.yml`, a CI runner, a shell profile — the guard stops registering and this test plus the seven behavioural mail tests in this file all fail with an error that points nowhere near the cause.

Pinning the guard explicitly would make the block independent of ambient state: add `newspack_guest_author_mail_guard_active` → `__return_true` in `set_up()` and call `register_mail_guard()`. Nothing is lost by doing so, because `test_mail_guard_environment_gate` already covers the gate itself directly.

(For what it's worth, I confirmed the constant half holds today — the tests pass with the site's `wp-config.php` set to `local`, because the PHPUnit bootstrap doesn't inherit the site config. This is about the `getenv` half.)

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
```diff
@@ -338,4 +338,227 @@ public function test_is_dummy_email_address() {
...
+		$this->assertSame( 1, has_filter( 'pre_wp_mail', [ Guest_Contributor_Role::class, 'short_circuit_dummy_only_email' ] ) );
+		$this->assertSame( 10, has_filter( 'wp_mail', [ Guest_Contributor_Role::class, 'remove_dummy_email_recipients' ] ) );
+	}
+
+	/**
+	 * The registration wiring honors the activity filter: forced inactive,
+	 * register_mail_guard() adds nothing; forced active again, it re-registers
+	 * both filters at their contract priorities.
+	 */
+	public function test_mail_guard_registration_honors_activity_filter() {
```
chickenn00dle: **Suggestion:** A failure inside this test poisons the four that follow it — This test removes both guard callbacks and adds `__return_false`, then unwinds inline rather than in a `finally`. If either of the two assertions between the `add_filter` and the matching `remove_filter` fails, PHPUnit throws and the guard is left unregistered with the activity filter still forcing it off — so the four mail-guard tests declared after this one fail as collateral, and the real failure is buried in the noise.

That matters more than usual here because the class overrides `set_up()` without calling `parent::set_up()`, which is what populates `WP_UnitTestCase_Base::$hooks_saved` — so `_restore_hooks()` in `tear_down()` is a no-op whenever this file runs on its own (`--filter Newspack_Test_Guest_Contributor_Role`, which is exactly how anyone iterating on this change would run it).

The neighbouring `test_mail_guard_follows_filtered_dummy_domain` already uses `try`/`finally`. Matching that pattern here would make the test self-contained.

n
```diff
@@ -467,6 +467,18 @@ case $1 in
...
+                || echo "  Could not set WP_ENVIRONMENT_TYPE on $c — environment-gated behavior may act as production there."
```
chickenn00dle: **Nit:** This warning names the wrong cause, and the loop is silent about what it did change — Two small things.

The `|| echo` fires whenever `wp config set` fails, and the most common reason won't be "the guard will misbehave" — it'll be "this container has no WordPress installed yet", i.e. a freshly created env before `n setup`. `config has` on line 476 fails the same way and falls through to the `set`, so every `n pull` will print this once per unprovisioned env, pointing the reader at the wrong problem. Skipping containers where `wp core is-installed` fails, or mentioning an unprovisioned site in the message, would keep the signal meaningful.

Conversely, the success path is entirely silent. I dry-ran the detection half against my five running containers: four of them — including the main dev site and three active PR-review envs — would be written to, changing Site Kit, Stripe, WooCommerce Subscriptions and Jetpack behavior in each as a side effect of a command called "pull", with no output saying so. Echoing which containers were changed would make that visible.

Also minor: `for c in $(docker ps … | grep -E …)` relies on word splitting. Container names can't contain whitespace so it works, but `… | while read -r c` avoids the pattern.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -273,16 +278,235 @@ public static function setup_custom_role_and_capability() {
...
+	 *
+	 * @return string[] Recipients that are not dummy addresses.
+	 */
+	private static function remove_dummy_addresses( array $recipients ): array {
+		return array_values(
+			array_filter(
+				$recipients,
+				function ( $recipient ) {
+					$address = $recipient;
+					if ( preg_match( '/<([^>]+)>/', $address, $matches ) ) {
```
chickenn00dle: **Nit:** Display-name extraction diverges from core's own parse — This uses `/<([^>]+)>/` while core's `wp_mail()` parses the same recipient with `/(.*)<(.+)>/` (`wp-includes/pluggable.php:477`) — greedy, so core takes the *last* bracketed segment.

They diverge only when a quoted display name itself contains angle brackets: for `'"a<b>c" <foo@example.com>'`, core dispatches to `foo@example.com` while this evaluates `b` and leaves the placeholder in place. That's pathological enough not to matter in practice. Still, since the whole point is to decide about the address core will actually dispatch to, matching core's pattern removes any chance of the two disagreeing.

chickenn00dle (approved):
Release risk: ELEVATED — no blockers, and everything I could exercise at runtime behaves as documented. The tier reflects two things rather than any defect: this sits on the outbound-mail path of the foundation plugin so the code runs for every email on every publisher, and a suppressed send currently leaves no trace on the sites where the guard is actually active.

On the question you asked directly: I'd take neither the fixture swap nor the environment gate. `bin/newspack-docker-mu.php` is already copied into every container on every start — and to all three site roots, not just the main site — and it already carries exactly this class of dev-box override, including ones for two of the plugins whose behavior you listed as collateral. A single `add_filter( 'newspack_guest_author_mail_guard_active', '__return_false' )` there is inert-everywhere-local with none of the blast radius, covers more surface than the current approach, and lets both the `init-wp.sh` hunk and the `n pull` loop be dropped — which turns this into a plugin-only hotfix, worth a lot on a patch release to `release`.

The other thing I'd resolve before merge is the audit trail. `NEWSPACK_LOG_LEVEL` isn't defined by any tracked tooling — I confirmed it's absent in isolated envs and set by hand only in my main dev container — so on the sites where the guard fires, the three new log lines produce nothing. That matters more than it first looks: the comment above `Emails::dispatch_email()`'s own logging explains that the old unconditional success-shaped log "routinely led incident investigation down the wrong path", and suppression reintroduces exactly that condition for placeholder recipients.

Everything else is small. Two test-hygiene items are worth taking because they'd bite whoever iterates on this next, and the rest are nits.

bin/init-wp.sh
wil-gerken: Thank you for this one — the mu-plugin approach was new to me.

Adopted in full: both tooling hunks are gone, and `bin/newspack-docker-mu.php` holds the single filter line. It sits next to the existing overrides for two of the plugins the constant would have touched.

We're tracking the standalone improvement of declaring `local` separately, as its own tooling change.

n
wil-gerken: Both tooling hunks dropped with the mu-plugin approach adopted this round, which reaches all three site roots, including the two this backfill missed.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Adopted the second option. The gate now evaluates inside both callbacks, so the filter is reachable from any plugin at the moment recovery needs it. Registration became deterministic as a side effect: the priority assertions no longer depend on the environment.

The docblock's wp-config route is gone; you're right that it would have fataled.

On the narrower-keying suggestion: domain-only stands, per the round-1 suppression-scope thread. The per-call evaluation answers the recovery risk — switching the guard on or off is one `add_filter` line from any plugin.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Adopted, reversing the round-1 call: the `dispatch_email()` precedent comment plus "the bounce was the last remaining signal" earned it.

All three decision points now emit `Logger::newspack_log()` under `newspack_guest_author_mail_suppressed`, in the `newspack_` prefix family. The payload carries a `mode` key naming which of the three cases fired, a `count`, and bare addresses capped at 20 entries, per the reader-registration precedent.

Payload composition is deliberate. The channel relays off-site through Manager by default, so the same core-parity parse the matcher uses strips display names, and the cap bounds pathological lists.

plugins/newspack-plugin/tests/unit-tests/emails.php
wil-gerken: Adopted the recipient check, gated on the guard being active. Dev test-sends into MailHog keep working, and the message stays true everywhere it appears:

`newspack_emails_test_placeholder_recipient` (400)
> This is a generated placeholder address, and mail sent to it is suppressed. Please use a real inbox for test sends.

A test pinned to guard-active covers it.

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
wil-gerken: Adopted. The mail tests pin the guard active via the activity filter in `set_up()`. That method now calls `parent::set_up()`, the missing call you flagged twice, so hook restoration is armed and the pin cannot leak.

With registration unconditional under the per-call gate, the `has_filter` assertions no longer depend on the environment.

plugins/newspack-plugin/tests/unit-tests/guest-contributor-role.php
wil-gerken: Adopted. The wiring test became a behavior test for the per-call design: forced inactive, placeholder mail delivers; pin restored, placeholder mail is suppressed. The flip sits inside try/finally, and `parent::set_up()` arms `_restore_hooks()` for solo runs.

n
wil-gerken: The loop is gone with the tooling revert adopted this round.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Adopted. Extraction now uses core's greedy pattern, via a shared `extract_address()` helper that both the matcher and the log payload call. A test asserts the bracket-in-display-name case resolves to the address core dispatches to.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Closing: the design landed on the mu-plugin approach from the round-2 review. No environment declaration, no fixture swap, one filter line in `bin/newspack-docker-mu.php`. Both tooling hunks are reverted.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
wil-gerken: Closing: domain-only stands, per the round-1 suppression-scope thread. Round 2's per-call gate answers the recovery concern, since the override filter is now reachable from any plugin.

wil-gerken:
Round 2 is in. What changed:

- The design landed on the mu-plugin approach, so both tooling hunks are reverted.
- The gate evaluates per send, which makes the override filter reachable from any plugin.
- Suppressions log through the manager-consumed channel, with a `mode` key naming which case fired.
- The wizard's test-send rejects placeholder recipients where the guard is active.

The newspack-plugin suite passes (2,359 tests) and PHPCS is clean. All threads resolved; ready for another look.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-guest-contributor-role.php
```diff
@@ -273,16 +284,292 @@ public static function setup_custom_role_and_capability() {
...
+			Logger::newspack_log(
+				self::MAIL_GUARD_LOG_CODE,
+				'Suppressed an outbound email: every recipient is a generated placeholder address.',
+				[
+					'mode'       => 'suppressed_all',
+					'count'      => count( $recipients ),
+					'recipients' => array_slice( array_map( [ __CLASS__, 'extract_address' ], $recipients ), 0, 20 ),
+				],
+				'info'
+			);
```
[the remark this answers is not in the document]
wil-gerken: This channel is deliberate and was requested in review: the earlier `NEWSPACK_LOG_LEVEL`-gated `Logger::log()` produced no trace on exactly the sites where the guard fires, and the bounce this replaces was the last remaining signal that anything was wrong (see the audit-trail thread).

On the specifics: the logged recipients are only the generated placeholder addresses the guard removed — real recipients are never in the payload — and the entry fires once per suppression event, not per send. The list is capped at 20 entries alongside a `count` and a `mode` key, a composition chosen with the off-site relay in mind.

PR #573, opened 2026-07-08
fix(active-campaign): keep signups working after field tag edits

wil-gerken:
### All Submissions:

Reference: NPPM-283

When a custom field's personalization tag (perstag) is renamed in ActiveCampaign, newsletter signups that sync reader data start failing with "There is already a field with this title". Because Newspack attaches signup metadata to every signup, on an affected site that means all signups fail.

## What this does

Signup field-matching is now resilient to field edits made on the ActiveCampaign side:

- Fields are matched by their tag first (as before), and now **fall back to matching by title** — case-insensitively, since ActiveCampaign treats titles as duplicates regardless of case.
- When a field is matched by title, the contact data is written to that field's **actual** tag, so the value lands in the right place.
- If a field can't be registered at all, the signup **no longer fails**: the problem is logged, that one field is skipped, and the reader is still subscribed.

**Prerequisite:** a site with ActiveCampaign connected. Run at least one signup first so the Newspack-generated contact fields exist.

1. In ActiveCampaign, under **Contacts → Fields → Manage**, change the personalization tag (`perstag`) of a Newspack-generated field (for example `NEWSLETTER_SUBSCRIPTION_METHOD` → `NEWSLETTERSSUBSCRIPTIONMETHOD`).
2. Register through Reader Activation so the signup syncs reader data. An email-only subscribe doesn't send this field and won't reproduce the problem.
3. The signup fails with "There is already a field with this title", and ActiveCampaign does not add the reader.

**Verify the fix:**
1. Keep the renamed personalization tag and sign up again on this branch. The signup succeeds, and ActiveCampaign saves the contact's data under the renamed field.
2. Sign up on a site with no renamed tags. The signup succeeds, and the contact data lands in the existing fields.

### Other information:

## Notes for reviewers
- The change is confined to the field-registration block of `add_contact()` in the AC provider; the method's signature and everything downstream are untouched.
- Matching uses index-preserving loops rather than `array_search` over `array_column` — AC field rows can be missing keys, and `array_column` reindexes around them, which could map a match to the wrong field's perstag. A title match also requires the row to carry a usable perstag; otherwise it falls through to field creation.
- A failed field-create now logs via `Newspack_Newsletters_Logger` and continues; previously the raw `WP_Error` aborted the signup before any contact sync ran.
- Four tests using the suite's established `pre_http_request` mock pattern: renamed perstag matches by title and uses the actual perstag; a failed create doesn't block the signup; the perstag fast path is unchanged; malformed rows (missing title or perstag) and re-cased, whitespace-padded titles are handled. Full newsletters suite green (486 tests).

### Release risk

**Medium.** The change sits on the reader-signup path, but it only executes for publishers using ActiveCampaign as their ESP, and the common case — tag matches, no admin edits — is verified unchanged by tests. One deliberate trade-off: a field that can't be registered is now skipped with a log entry instead of failing the whole signup loudly; keeping signups working outweighs one metadata field. Rollback is a plain revert, which restores the previous hard-fail behavior.

### DIFF
--- a/plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -1999,8 +1999,40 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 			$existing_fields = $this->get_all_contact_fields();
 			foreach ( $contact['metadata'] as $field_title => $value ) {
 				$field_perstag = strtoupper( str_replace( '-', '_', sanitize_title( $field_title ) ) );
-				/** For optimization, don't add the field if it already exists. */
-				if ( is_wp_error( $existing_fields ) || false === array_search( $field_perstag, array_column( $existing_fields, 'perstag' ) ) ) {
+
+				/**
+				 * For optimization, don't add the field if it already exists.
+				 * Match by perstag first, then by title — an ActiveCampaign
+				 * admin may have renamed a field's perstag, and creating a
+				 * field with a duplicate title fails.
+				 */
+				$existing_index = false;
+				if ( ! is_wp_error( $existing_fields ) ) {
+					// Index-preserving lookups: array_column() would reindex and skip
+					// rows missing the key, mapping a match to the wrong field.
+					foreach ( $existing_fields as $index => $existing_field ) {
+						if ( isset( $existing_field['perstag'] ) && $existing_field['perstag'] === $field_perstag ) {
+							$existing_index = $index;
+							break;
+						}
+					}
+					if ( false === $existing_index ) {
+						foreach ( $existing_fields as $index => $existing_field ) {
+							if (
+								isset( $existing_field['title'], $existing_field['perstag'] ) &&
+								0 === strcasecmp( trim( $existing_field['title'] ), trim( $field_title ) )
+							) {
+								$existing_index = $index;
+								break;
+							}
+						}
+					}
+				}
+
+				if ( false !== $existing_index ) {
+					/** Use the existing field's actual perstag — it may differ from the generated one. */
+					$field_perstag = $existing_fields[ $existing_index ]['perstag'];
+				} else {
 					$field_res = $this->api_v3_request(
 						'fields',
 						'POST',
@@ -2018,7 +2050,9 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 						]
 					);
 					if ( \is_wp_error( $field_res ) ) {
-						return $field_res;
+						/** A field that can't be registered must never block the signup — sync the contact without it. */
+						Newspack_Newsletters_Logger::log( 'Error creating ActiveCampaign field "' . $field_title . '": ' . $field_res->get_error_message() );
+						continue;
 					}
 					/** Set list relation. */
 					$this->api_v3_request(
--- a/plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
+++ b/plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
@@ -2010,8 +2010,10 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 				if ( ! is_wp_error( $existing_fields ) ) {
 					// Index-preserving lookups: array_column() would reindex and skip
 					// rows missing the key, mapping a match to the wrong field.
+					// A row with an empty perstag can't supply a usable payload key, so
+					// it must never match — `! empty()` rather than `isset()`.
 					foreach ( $existing_fields as $index => $existing_field ) {
-						if ( isset( $existing_field['perstag'] ) && $existing_field['perstag'] === $field_perstag ) {
+						if ( ! empty( $existing_field['perstag'] ) && $existing_field['perstag'] === $field_perstag ) {
 							$existing_index = $index;
 							break;
 						}
@@ -2019,7 +2021,7 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 					if ( false === $existing_index ) {
 						foreach ( $existing_fields as $index => $existing_field ) {
 							if (
-								isset( $existing_field['title'], $existing_field['perstag'] ) &&
+								! empty( $existing_field['title'] ) && ! empty( $existing_field['perstag'] ) &&
 								0 === strcasecmp( trim( $existing_field['title'] ), trim( $field_title ) )
 							) {
 								$existing_index = $index;
@@ -2032,6 +2034,14 @@ final class Newspack_Newsletters_Active_Campaign extends \Newspack_Newsletters_S
 				if ( false !== $existing_index ) {
 					/** Use the existing field's actual perstag — it may differ from the generated one. */
 					$field_perstag = $existing_fields[ $existing_index ]['perstag'];
+				} elseif ( empty( $field_perstag ) ) {
+					/**
+					 * The title sanitized down to nothing, so there's no perstag to create the
+					 * field under and no valid payload key to write to. Skip rather than emit a
+					 * malformed `field[%%,0]` key that two such fields would collide on.
+					 */
+					error_log( '[NEWSPACK-NEWSLETTERS]: Skipped ActiveCampaign field "' . $field_title . '": title produced an empty perstag.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+					continue;
 				} else {
 					$field_res = $this->api_v3_request(
 						'fields',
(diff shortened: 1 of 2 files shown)


plugins/newspack-newsletters/tests/test-active-campaign-field-matching.php
```diff
@@ -0,0 +1,245 @@
...
+			return $respond(
+				422,
+				[
+					'errors' => [
+						[
+							'code'  => 'duplicate',
+							'title' => 'There is already a field with this title',
+						],
+					],
+				]
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 80a9ca010 — trailing whitespace stripped (a leftover from the phpcbf reformat).

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1999,8 +1999,40 @@ public function add_contact( $contact, $list_id = false ) {
...
+					foreach ( $existing_fields as $index => $existing_field ) {
+						if ( isset( $existing_field['perstag'] ) && $existing_field['perstag'] === $field_perstag ) {
+							$existing_index = $index;
+							break;
+						}
+					}
+					if ( false === $existing_index ) {
+						foreach ( $existing_fields as $index => $existing_field ) {
+							if (
+								isset( $existing_field['title'], $existing_field['perstag'] ) &&
```
dkoo: **Suggestion:** Field-match guards check isset() not !empty() — an empty perstag can match — The perstag and title-match guards check `isset( $existing_field['perstag'] )` rather than `! empty( ... )`, so a field row whose `perstag` is an empty string can still match. When it does, `$field_perstag` is set to `''` (line 2034) and the payload emits a malformed `field[%%,0]` key — and two such fields would collide on it. There's **no cross-field data-leak risk**: a value is never routed into a different real named field, and in practice ActiveCampaign auto-assigns perstags, so an empty one is effectively unreachable. Still, tightening both guards to `! empty( $existing_field['perstag'] )` (and skipping the field, or falling through to create, when the *generated* perstag is empty) is cheap defense and would round out the existing malformed-row test coverage — the current `test_title_match_survives_field_rows_without_title` covers a *missing* perstag key but not an empty-string one.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -2018,7 +2050,9 @@ public function add_contact( $contact, $list_id = false ) {
 						]
 					);
 					if ( \is_wp_error( $field_res ) ) {
-						return $field_res;
+						/** A field that can't be registered must never block the signup — sync the contact without it. */
+						Newspack_Newsletters_Logger::log( 'Error creating ActiveCampaign field "' . $field_title . '": ' . $field_res->get_error_message() );
```
dkoo: **Suggestion:** Field-create failure log is level-gated, so a dropped RAS field is silent on a default site — Making a failed field-create non-fatal is the right call — one unregisterable field should never abort a reader's whole signup. Worth weighing, though: the metadata synced here is Reader Activation contact data (registration, engagement, subscription, donation, content-gate status) that ActiveCampaign segments and automations key on, so a *silently* dropped field isn't purely cosmetic. The common real failure — the duplicate-title 422 — is now caught by the new title match before any create is attempted, and the remaining path is self-healing (retried on the next signup), which keeps this low-risk. The catch: `Newspack_Newsletters_Logger::log()` no-ops entirely unless `NEWSPACK_LOG_LEVEL` is defined (`class-newspack-newsletters-logger.php:20`), which it isn't on a default production site — so a field that keeps failing to create leaves no trace at all. Consider making just this one failure surface unconditionally (e.g. a direct `error_log`) so an under-populated segment is diagnosable.

**Risk if not addressed:** If a field-create keeps failing for a non-title reason on a site that has not defined NEWSPACK_LOG_LEVEL, a load-bearing Reader Activation field (donation, subscription, or membership status) is silently dropped on every signup with no log trace, so readers who should enter an AC segment or automation keyed on that field are never added to it — invisible until a publisher notices the segment is under-populated.

dkoo (approved):
Tested end-to-end with a live ActiveCampaign account. Approve with comments. The core logic is a clear improvement over the previous `array_search`/`array_column` approach and the tests genuinely cover the claimed behaviors, including malformed rows. No blockers. Two low-severity suggestions before merge: harden the field-match guards against an empty-string perstag (defensive; no data-leak risk), and make the field-create failure log unconditional so a silently-dropped RAS metadata field is diagnosable on a default production site.

Making a failed field-create non-fatal is the correct direction — a single unregisterable field must never block a reader signup. The common real failure (a duplicate-title 422) is now caught by the new title match before any create is attempted, so this silent-drop path is only reached for genuinely-new fields that fail to create for some other reason: a narrow, self-healing window rather than a standing state.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -1999,8 +1999,50 @@ public function add_contact( $contact, $list_id = false ) {
...
+				if ( false !== $existing_index ) {
+					/** Use the existing field's actual perstag — it may differ from the generated one. */
+					$field_perstag = $existing_fields[ $existing_index ]['perstag'];
+				} elseif ( empty( $field_perstag ) ) {
+					/**
+					 * The title sanitized down to nothing, so there's no perstag to create the
+					 * field under and no valid payload key to write to. Skip rather than emit a
+					 * malformed `field[%%,0]` key that two such fields would collide on.
+					 */
+					error_log( '[NEWSPACK-NEWSLETTERS]: Skipped ActiveCampaign field "' . $field_title . '": title produced an empty perstag.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
```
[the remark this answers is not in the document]
wil-gerken: Sanitized in `3fcc2f098`.

On routing through the logger: `Newspack_Newsletters_Logger::log()` delegates to `\Newspack\Logger::log()`, which itself terminates in `error_log()` (`newspack-plugin/includes/class-logger.php:68`). The sink is the same either way, so a site with `NEWSPACK_LOG_LEVEL` set still gets this line in the same file. What it loses is the PID and caller prefix.

The unconditional call is deliberate. The logger no-ops entirely when the constant is undefined, which is the default in production, and that silence is what this change exists to remove.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
```diff
@@ -2018,7 +2060,17 @@ public function add_contact( $contact, $list_id = false ) {
...
+						/**
+						 * A field that can't be registered must never block the signup — sync the
+						 * contact without it.
+						 *
+						 * Logged with error_log() rather than Newspack_Newsletters_Logger, which
+						 * no-ops unless NEWSPACK_LOG_LEVEL is defined — undefined on a default
+						 * production site. The dropped field carries Reader Activation data that
+						 * segments and automations key on, so it must stay diagnosable.
+						 */
+						error_log( '[NEWSPACK-NEWSLETTERS]: Error creating ActiveCampaign field "' . $field_title . '": ' . $field_res->get_error_message() ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
```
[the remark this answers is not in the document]
wil-gerken: Sanitized in `3fcc2f098`. Worth flagging on this line specifically: `get_error_message()` carries ActiveCampaign's API text, so it's genuinely remote data rather than anything we control, which makes it the stronger of the two cases.

Same note as the other thread on the logger: `Newspack_Newsletters_Logger::log()` bottoms out in `error_log()` by way of `\Newspack\Logger` (`newspack-plugin/includes/class-logger.php:68`), so nothing is lost in delivery when `NEWSPACK_LOG_LEVEL` is set, only the PID and caller prefix. Logging unconditionally here is the intended behavior.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
wil-gerken: Great catch, fixed in `7b40713ab`. Both guards now use `! empty()`, and a metadata key whose generated perstag sanitizes to empty is skipped rather than sent through field creation. ActiveCampaign's behavior on an empty-perstag create isn't something I verified, so skipping avoids depending on it either way. Two new tests cover a field row with an empty-string perstag and a generated perstag that sanitizes to empty. Both fail without the change.

plugins/newspack-newsletters/includes/service-providers/active_campaign/class-newspack-newsletters-active-campaign.php
wil-gerken: Fixed in `bb5f2a707`. `\Newspack\Logger` has the same log-level gate, so the failure now goes straight to `error_log()` with the existing `[NEWSPACK-NEWSLETTERS]` prefix, and a comment explains the bypass. Not a `Logger::error()` wrapper, because that name would imply the method still respects the configured log level.

Follow-up in `3fcc2f098`: Copilot flagged that both interpolated values reach the log unsanitized, and the error message is ActiveCampaign API text, so a newline in it would forge a log line. Both are wrapped in `sanitize_text_field()` now.

PR #575, opened 2026-07-08
feat(gates): capture gate id on landing-page conversions (NPPD-1887)

kmwilkerson:
Refs [NPPD-1887](https://linear.app/a8c/issue/NPPD-1887/insights-credit-gates-and-prompts-for-conversions-that-happen-on-a). **Targets `release`.**

## Why this exists separately, and why it targets `release`

The full NPPD-1887 work is [workspace#571](https://github.com/Automattic/newspack-workspace/pull/571) → `feat/insights-rsm`. That branch also carries the Insights *reading* side, the prompts half, and the hub denominator — none of which is released.

But one piece of #571 is **forward-only and unrecoverable**: the `_gate_post_id` order meta. It's written at order time, so any gate→landing-page conversion that happens before the capture code ships is unattributed *forever* — there's no backfill. Insights isn't released yet (0 files on `main`/`alpha`/`release`), but when it does launch it can only show history that was captured in the meantime.

So this PR extracts **just the gate-side capture** — the part that banks data — and targets `release` so orders start carrying `_gate_post_id` now, independent of the Insights UI. The reading side ships later with Insights and finds the history waiting.

**This is capture, not Insights.** Every changed file references Insights only in *comments*. `_gate_post_id` is plain WooCommerce order meta; the code that writes it (`GATE_CPT`, `Donations`, `data-events/Memberships`) already exists on `release`. Verified: the full capture test suite passes against the `release` baseline with no Insights and no NPPD-1837 present (29 tests).

## What it does

A reader clicks a paid-intent CTA inside a gate that links to a subscription/donation landing page, converts there, and the order now carries `_gate_post_id` — exactly as it does for a checkout block rendered inside the gate.

1. **`Newspack\CTA_Intent_Classifier`** — classifies a CTA href as `donation` / `subscription` / non-conversion. (This is NPPD-1837's classifier; on `release` it's brand-new, self-contained, no dependency on Campaigns or Insights.)
2. **Stamp at gate render** — `data-newspack-cta` on paid-intent `wp-block-button__link` anchors, at the two gate-HTML producers only (never the article excerpt).
3. **Persist on click** — `gate.js` writes `sessionStorage.newspack_cta_attribution` and fires a new `np_gate_interaction / action:'clicked'` event (gates never had a click event).
4. **Replay at submit** — `modal.js` injects the hidden field, guarded so a form already carrying the id wins and a form inside a gate is never touched.
5. **Validate before writing meta** — client-supplied ids are checked against the gate CPT and persisted as integers.

## The gtag fix — a live bug this also carries

`gate.js::addFormInputs()` — which stamps `gate_post_id` onto the gate's own checkout form, i.e. the entire in-gate attribution chain — sat **behind** an early `if ( 'function' !== typeof window.gtag ) return;`. So on `release` today, **every reader with GA blocked who converts through a gate's embedded checkout produces an order with no gate on it** — silent, ongoing, unrecoverable data loss on every gated site, unrelated to landing pages. Moved above the guard. Verified live with `window.gtag` forced `undefined`: the order carried `_gate_post_id`.

## Not included (stays on #571 → insights-rsm)

- Prompts capture (depends on NPPD-1837, which isn't on `release`)
- The popups classifier delegation and prompt-id validation hardening
- Everything Insights reads (`Subscribers_Metric`, Gates tab, etc.)
- The hub denominator PR

## Verified end-to-end on real orders (on the rsm build; identical code)

- gated post → stamped CTA → `/subscribe/` → real WooCommerce subscription → order carried **`_gate_post_id`**
- in-gate checkout with `window.gtag` forced undefined → order carried `_gate_post_id` (the gtag fix)
- editorial link click in the same gate → persists nothing

## Testing

- Capture PHP suite: **29 pass against the `release` baseline** (classifier, gate annotation, order-meta validation)
- JS byte-identical to the ESLint-clean #571 versions
- Root PHPCS clean

## Reviewer notes

- **Touches the live checkout path** (`modal.js` replay runs on every modal-checkout submit; classifier runs at gate render). All defensively guarded (try/catch on storage, null guards, two replay guards), but it warrants a careful human read before merge.
- This deliberately duplicates #571's gate-capture commits. When this reaches `main` via `release → main` and `feat/insights-rsm` next syncs `main`, the identical hunks reconcile. #571 can then drop its capture portion, or be merged as-is.
- Copilot review response on #571 applies here too: `gate_has_checkout_link` is stamped for **any** paid intent (not just subscription) on purpose — `classify_href()` reads `/membership/` as `donation`, and narrowing it would drop those conversions. See NPPD-1888.

### DIFF
--- a/plugins/newspack-blocks/src/modal-checkout/modal.js
+++ b/plugins/newspack-blocks/src/modal-checkout/modal.js
@@ -25,6 +25,7 @@ import {
 	getFormattedAmount,
 } from './utils';
 import { resolveCheckoutButtonForm, readCheckoutData } from './checkout-button-trigger';
+import { applyCtaAttribution } from '../shared/js/cta-attribution';
 
 const CLASS_PREFIX = newspackBlocksModal.newspack_class_prefix;
 const IFRAME_NAME = 'newspack_modal_checkout_iframe';
@@ -300,6 +301,13 @@ domReady( () => {
 		}
 		const form = ev.target;
 		form.classList.add( 'modal-processing' );
+
+		// NPPD-1887: if the reader arrived here by clicking a paid-intent CTA in a gate
+		// or prompt, replay that surface's id as a hidden field so the order carries
+		// `_gate_post_id` / `_newspack_popup_id`. No-ops when the form already has one
+		// (a form rendered inside the surface itself always wins) or when the form is
+		// inside a gate. Must run BEFORE getCheckoutData(), which snapshots the form.
+		applyCtaAttribution( form );
 
 		const checkoutData = getCheckoutData( form );
 
--- a/plugins/newspack-blocks/src/shared/js/cta-attribution.js
+++ b/plugins/newspack-blocks/src/shared/js/cta-attribution.js
@@ -19,6 +19,7 @@
  * plugins ship independently and share no JS package. Change one, change the other.
  *
  *   sessionStorage[ 'newspack_cta_attribution' ] = JSON.stringify( {
+ *     v:    1,
  *     type: 'gate' | 'prompt',
  *     id:   '123',
  *     ts:   1751932800000,
@@ -27,6 +28,14 @@
 
 /** Storage key. Shared with newspack-plugin. */
 export const STORAGE_KEY = 'newspack_cta_attribution';
+
+/**
+ * Storage-record schema version. Must match the writer's in newspack-plugin. A record
+ * whose `v` differs is treated as absent rather than trusted, so a drift between the two
+ * duplicated copies (e.g. one changes the record shape or TTL) fails safe instead of
+ * silently honoring a stale record. (dkoo review, #575.)
+ */
+export const SCHEMA = 1;
 
 /** Attribution lifetime, in ms. Kept in sync with newspack-plugin. */
 export const TTL_MS = 30 * 60 * 1000;
@@ -62,7 +71,8 @@ export function readCtaAttribution() {
 	} catch ( e ) {
 		return null;
 	}
-	if ( ! record || ! FIELD_BY_TYPE[ record.type ] || ! record.id ) {
+	// Reject records written by a different schema version (drift fail-safe).
+	if ( ! record || record.v !== SCHEMA || ! FIELD_BY_TYPE[ record.type ] || ! record.id ) {
 		return null;
 	}
 	// Session-scoped AND time-bounded (NPPD-1887 product decision). A reader who clicks
--- a/plugins/newspack-blocks/src/shared/js/cta-attribution.js
+++ b/plugins/newspack-blocks/src/shared/js/cta-attribution.js
@@ -1,0 +1,120 @@
+/**
+ * CTA attribution replay (NPPD-1887).
+ *
+ * The consumer half of the hand-off written by newspack-plugin when a reader clicks a
+ * paid-intent CTA inside a gate or prompt and is sent to a landing page. We re-inject
+ * the surface's id as a hidden form field, so the existing server chain
+ * (`INPUT_GET` -> cart item data -> `_gate_post_id` / `_newspack_popup_id` order meta)
+ * fires exactly as it does for a form rendered inside the gate itself.
+ *
+ * Two cart paths exist, and they apply DIFFERENT filters. Checkout buttons go through
+ * `Modal_Checkout::add_to_cart_and_redirect()` -> `newspack_blocks_modal_checkout_cart_item_data`;
+ * donations go through `Donations::process_donation_request()` -> `newspack_donations_cart_item_data`.
+ * Injecting the field here is necessary but not sufficient — a listener must exist on
+ * whichever filter the target form uses, or the id rides all the way into `INPUT_GET`
+ * and is then dropped.
+ *
+ * ── Storage contract ────────────────────────────────────────────────────────────
+ * DUPLICATED from newspack-plugin (`src/shared/js/cta-attribution.js`). The two
+ * plugins ship independently and share no JS package. Change one, change the other.
+ *
+ *   sessionStorage[ 'newspack_cta_attribution' ] = JSON.stringify( {
+ *     type: 'gate' | 'prompt',
+ *     id:   '123',
+ *     ts:   1751932800000,
+ *   } )
+ */
+
+/** Storage key. Shared with newspack-plugin. */
+export const STORAGE_KEY = 'newspack_cta_attribution';
+
+/** Attribution lifetime, in ms. Kept in sync with newspack-plugin. */
+export const TTL_MS = 30 * 60 * 1000;
+
+/** Surface type -> the form field the server reads. */
+const FIELD_BY_TYPE = {
+	gate: 'gate_post_id',
+	prompt: 'newspack_popup_id',
+};
+
+/**
+ * Read a fresh attribution record, or null.
+ *
+ * Never throws: `sessionStorage` access raises in Safari private mode, and the stored
+ * value is user-writable so `JSON.parse` can fail. Attribution is best-effort — a bad
+ * record must never block a checkout.
+ *
+ * @return {{type: string, id: string, ts: number}|null} The record, or null.
+ */
+export function readCtaAttribution() {
+	let raw;
+	try {
+		raw = window.sessionStorage.getItem( STORAGE_KEY );
+	} catch ( e ) {
+		return null;
+	}
+	if ( ! raw ) {
+		return null;
+	}
+	let record;
+	try {
+		record = JSON.parse( raw );
+	} catch ( e ) {
+		return null;
+	}
+	if ( ! record || ! FIELD_BY_TYPE[ record.type ] || ! record.id ) {
+		return null;
+	}
+	// Session-scoped AND time-bounded (NPPD-1887 product decision). A reader who clicks
+	// the CTA, wanders off, and converts two hours later in the same tab gets no credit.
+	if ( 'number' !== typeof record.ts || Date.now() - record.ts > TTL_MS ) {
+		return null;
+	}
+	return record;
+}
+
+/**
+ * Inject the persisted surface id into a checkout/donate form, if it needs one.
+ *
+ * Two guards, both load-bearing:
+ *
+ *   1. A form that ALREADY carries the field wins. That is a form rendered inside the
+ *      gate or prompt itself, whose id was stamped server-side (prompts) or by
+ *      `gate.js::addFormInputs()` (gates). A replayed value must never overwrite the
+ *      surface the reader is actually looking at. Same precedence rule that
+ *      `copyContextFields()` applies to PICKER_CONTEXT_FIELDS.
+ *
+ *   2. A form inside a gate is never replayed into. `gate.js` owns attribution there,
+ *      and it binds on the gate's 'seen' event — which may not have fired yet when
+ *      this runs. Without this guard, a stale id from a gate on a PREVIOUS page could
+ *      beat the current page's gate to the field.
+ *
+ * Called from the submit handler rather than at DOM-ready, so it observes the final
+ * state of the form after every other script has had its turn.
+ *
+ * @param {HTMLFormElement} form The form about to be submitted.
+ *
+ * @return {boolean} Whether a field was injected.
+ */
+export function applyCtaAttribution( form ) {
+	if ( ! form || 'function' !== typeof form.querySelector ) {
+		return false;
+	}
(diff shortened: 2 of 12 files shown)


plugins/newspack-blocks/src/modal-checkout/modal.js
```diff
@@ -301,6 +302,13 @@ domReady( () => {
 		const form = ev.target;
 		form.classList.add( 'modal-processing' );
 
+		// NPPD-1887: if the reader arrived here by clicking a paid-intent CTA in a gate
+		// or prompt, replay that surface's id as a hidden field so the order carries
+		// `_gate_post_id` / `_newspack_popup_id`. No-ops when the form already has one
+		// (a form rendered inside the surface itself always wins) or when the form is
+		// inside a gate. Must run BEFORE getCheckoutData(), which snapshots the form.
+		applyCtaAttribution( form );
```
dkoo, line 310: **Severity:** Suggestion

**Suggestion:** Last-touch replay can attribute an unrelated purchase to a gate within the TTL — `applyCtaAttribution( form )` fires on **every** modal-checkout submit and injects the persisted `gate_post_id` whenever a fresh record (&lt;30 min, same session) exists and the form isn't inside a gate. Scenario: a reader clicks a gate's paid CTA (persist), abandons the landing page, then buys an unrelated product through a checkout-button block elsewhere in the same tab 10 minutes later — that order gets `_gate_post_id` stamped even though the gate had nothing to do with it. Server validation only confirms the id is a real gate, not that the purchase was related. This is the accepted last-touch / session / 30-minute product model, so no code change is strictly required — but it's the highest-impact behavioral change shipping to `release` and deserves an explicit product sign-off before merge.

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
```diff
@@ -0,0 +1,456 @@
...
+
+		// 3) Path / substring patterns. Substring (not slash-anchored) on purpose, so
+		// membership fragments and newslettersignup slugs are caught (NPPD-1836).
+		// Order matters: donation, then newsletter, then subscription.
+		// (?<![a-z]) applies only to the keyword branch so mid-word matches (e.g. "member"
+		// inside "remember") don't false-positive; the href is already lowercased. Digits,
+		// "-", "/", "#" and start-of-string are all valid boundaries, so "/donate",
+		// "-membership" and "#...-membership" still match. /give and /support stay
+		// slash-anchored exactly as before.
+		if ( preg_match( '/(?<![a-z])(?:donate|donation|contribute|donor|member|membership)|\/give\b|\/support\b/', $href ) ) {
```
dkoo, line 328: **Severity:** Suggestion

**Suggestion:** Same-site editorial URLs with conversion tokens are stamped as paid CTAs — The donation/subscription substring patterns (lines 308, 321) run **before** the same-host editorial guard, so a same-site editorial URL whose slug happens to contain a conversion token is classified as a paid CTA. Verified against the actual regexes: `/2026/06/12/school-board-member-profile/` → `donation` (the `member` token), and `/digital-divide/` → `subscription` (the `/digital` token). If a gate mixes a real paid CTA with a button linking to such a same-site article, that editorial button also gets `data-newspack-cta`; clicking it persists gate attribution and writes `_gate_post_id` onto a later order, crediting the gate for a conversion the reader didn't make through it. The tension is genuine — `/become-a-member/` legitimately needs `donation` to win over the editorial pattern — so a blanket reorder isn't safe. A targeted guard (e.g. short-circuit to abstain when the same-host href matches a strong editorial signal like the dated `/YYYY/MM/DD/` path prefix) plus one test asserting a dated same-host article slug containing a token abstains would close the most common case without regressing real conversion pages.

plugins/newspack-blocks/src/shared/js/cta-attribution.js
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Read a fresh attribution record, or null.
+ *
+ * Never throws: `sessionStorage` access raises in Safari private mode, and the stored
+ * value is user-writable so `JSON.parse` can fail. Attribution is best-effort — a bad
+ * record must never block a checkout.
+ *
+ * @return {{type: string, id: string, ts: number}|null} The record, or null.
+ */
+export function readCtaAttribution() {
```
dkoo, line 58: **Severity:** Nit

**Nit:** Duplicated JS storage contract has no version field — `STORAGE_KEY`, `TTL_MS`, and the record shape are duplicated across newspack-plugin and newspack-blocks (no shared JS package — a documented, reasonable call given the two plugins ship independently). Neither side writes a schema/version field, so if the two copies ever drift (e.g. one bumps `TTL_MS` or changes the shape) a stale record from the other's version would be silently honored or dropped with no signal. A one-byte `v: 1` field checked in `readCtaAttribution()` would make drift fail safe. Minor, since last-touch + the 30-minute TTL already bound the damage.

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
```diff
@@ -0,0 +1,456 @@
...
+ * Newspack CTA intent classifier.
+ *
+ * Classifies a call-to-action link target (an href) into a conversion intent —
+ * donation, newsletter, subscription — or into a clearly non-conversion tell
+ * (event, sponsor, editorial), or abstains.
+ *
+ * Originally written for NPPD-1837 as a display-only labeller for block-less
+ * button prompts, living in newspack-popups. Moved here (NPPD-1887) because
+ * content gates need the same classification and must not depend on Campaigns
+ * being active. `Newspack_Popups_Data_Api` now delegates to this class, and
```
dkoo: **Severity:** Nit

**Nit:** Docblock claims a popups delegation that doesn't exist on `release` — The class docblock states "`Newspack_Popups_Data_Api` now delegates to this class." On `release`, newspack-popups has no reference to the classifier (that delegation ships with #571), so the comment describes the full-feature end state, not current behavior on this branch. Harmless — the classifier is correctly self-contained here — but the doc reads as describing behavior that doesn't exist yet on `release`. Consider softening the tense until #571 lands.

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
```diff
@@ -0,0 +1,456 @@
...
+		return $host . ( $parts['path'] ?? '' );
+	}
+
+	/**
+	 * Is this href pointing off the current site's host?
+	 *
+	 * @param string $href Lowercased href.
+	 * @return bool
+	 */
+	private static function is_external_host( $href ) {
```
dkoo, line 473: **Severity:** Nit

**Nit:** `is_external_host` only checks the literal home_url() host — Host comparison uses only `home_url()`'s host (minus `www.`). On multisite / multibrand installs served on more than one hostname, a genuinely on-site editorial link on a secondary brand host is treated as external, so the editorial-abstain branch doesn't fire and the link abstains (returns null). The failure mode is *under*-attribution, which is safe under the module's precision-first contract — worth only a one-line note that host detection is single-host.

dkoo (approved):
Tested manually in browser; no blockers. The security-sensitive path — a client-supplied gate id written to order meta — is properly hardened: the id is cast to an integer and confirmed to be a real gate CPT before it's persisted, so a malformed or arbitrary value can't land in `_gate_post_id`. The HTML annotation uses `WP_HTML_Tag_Processor` rather than regex, the gtag-guard reordering closes a genuine GA-blocked data-loss path, and the double-count invariant is covered by a regression test.

Worth a look before merge: the classifier stamps some same-site editorial URLs (those whose slug contains a conversion token like `member` or `digital`) as paid CTAs, which can over-attribute a gate; and the last-touch/30-minute replay can attach a gate id to an unrelated purchase in the same session. Both are attribution-quality issues — nothing charges or mutates state on the value and it's recomputable from order history — but since this capture is forward-only they're worth a quick product/eng glance now.

- `plugins/newspack-plugin/includes/data-events/class-memberships.php:133` — **Suggestion:** Registration/auth metadata path writes the gate id without CPT validation — The order path was hardened with `get_valid_gate_post_id()`, but `register_reader_metadata()` still writes `filter_input( INPUT_POST, 'gate_post_id', FILTER_SANITIZE_NUMBER_INT )` straight into the `reader_registered` Data Event metadata with no CPT check. **This path is preexisting and this PR does not widen it** — the sessionStorage replay is imported only by `modal.js` and fires only on modal-checkout submit, so it never reaches registration/auth forms. So this is a defense-in-depth / consistency suggestion rather than a hole this PR introduces: routing this path through `get_valid_gate_post_id()` too would keep both attribution writes uniformly integer- and CPT-validated. Minor aside: `FILTER_SANITIZE_NUMBER_INT` lets non-canonical strings like `+9` / `12-3` through; `absint()` is the more idiomatic choice here.
- `plugins/newspack-plugin/src/content-gate/gate.js:132` — **Nit:** Dead `|| '1'` fallback in addFormInputs; diverges from the persist path — `addFormInputs()` defaults a missing `gate_post_id` to `'1'`, whereas the new click-persist path (`persistCtaAttribution( 'gate', gateInfo.gate_post_id )`, line 161) has no fallback and simply no-ops on a falsy id. With the new server-side validation, `get_valid_gate_post_id(1)` rejects post 1 unless it happens to be a gate CPT — so the `'1'` fallback is now effectively dead, but on the rare install where post 1 *is* a gate it would mis-attribute, and the two paths behaving differently is confusing. Prefer bailing out (no hidden input, no persist) when the real id is absent, consistently in both places.

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
kmwilkerson: Fixed in f5378e6. Added a strong-editorial short-circuit in `classify_href()`: a same-host dated path (`/YYYY/MM/`) returns `editorial` **before** the donation/subscription patterns run. So `/2026/06/12/school-board-member-profile/` now abstains instead of matching `donation` on the `member` token. Scoped exactly as you suggested — only the dated prefix — so real conversion pages like `/become-a-member/` (no date) still reach the donation pattern and win. Added `test_dated_article_with_token_abstains_to_editorial` (member/donate/subscribe tokens). The non-dated residual (`/digital-divide/`) is noted inline as a known, bounded edge under the precision-first contract.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-blocks/src/shared/js/cta-attribution.js
kmwilkerson: Fixed in f5378e6. Added a `SCHEMA = 1` constant to both copies; the writer stamps `v` into the record and `readCtaAttribution()` rejects any record whose `v` !== the reader's `SCHEMA` — so drift fails safe (ignored, not silently honored). Added `rejects a record written by a different schema version` to the blocks JS suite. Documented the bump-on-change rule in both files.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
kmwilkerson: Fixed in f5378e6. Softened the docblock: it now says the popups delegation ships with #571, and that on this branch the class is self-contained and used only by the content gate — matching current `release` behavior.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-plugin/includes/class-cta-intent-classifier.php
kmwilkerson: Fixed in f5378e6. Added a note to `is_external_host()` documenting single-host detection and that the multibrand failure mode is under-attribution (a secondary-host on-site link abstains), which is safe under the precision-first contract.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/newspack-blocks/src/modal-checkout/modal.js
kmwilkerson: No code change — this is the intended model, and you've described it exactly. Katie explicitly signed off on last-touch / session / 30-minute TTL, **including** the same-window unrelated-purchase case you describe: her words were that a second purchase inside the 30-minute window inheriting the credit is 'low stakes... totally fine.' Clear-on-`woocommerce_thankyou` was considered and deferred as a known follow-up, not a blocker.

Leaving this thread open rather than resolving it myself, since it's a product sign-off gate — @kmwilkerson to close when she's comfortable it's on record here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #576, opened 2026-07-08
fix(newspack-blocks): skip article-block image crops on wpcom uploads

chickenn00dle:
### All Submissions:

Uploading an image through the editor's Image block generated a full set of ~16 `newspack-article-block-*` crop derivatives, while uploading the same image through the Media Library generated only the core sizes. On WordPress.com sites this is inconsistent because `newspack-blocks` (and its image-size registration) is loaded by jetpack-mu-wpcom only for editor/REST requests: the REST upload path from the Image block (`is_admin()` is false) registers the crop sizes and generates the files, whereas the Media Library's `async-upload.php` path does not.

This skips generating the **physical** `newspack-article-block-*` crop files on WordPress.com platform sites (Simple and Atomic), where an image CDN crops on the fly from the source image. The sizes stay registered, so the blocks still resolve correctly-cropped URLs via the CDN — only the redundant physical files are skipped. The result is consistent across both upload paths (neither generates the extra files on wpcom).

* Adds an `intermediate_image_sizes_advanced` filter that removes the `newspack-article-block-*` sub-sizes from physical generation when the site is on the wpcom platform.
* Leaves the sizes registered so `image_size_for_orientation()` still resolves the cropped CDN URL (and falls back to `large` if a crop can't be resolved).
* Self-hosted sites are unaffected — they keep generating the physical crops. A new `newspack_blocks_skip_article_image_subsizes` filter lets self-hosted sites opt in (e.g. when using the Jetpack Image CDN).

Note: on wpcom this fix takes effect once the change is synced into `jetpack-mu-wpcom` (the blocks there are auto-synced from this plugin).

Closes NPPM-2798.

**Unit tests (local):**

1. `cd plugins/newspack-blocks && n test-php --filter ImageSizesTest` — both tests pass (crops retained by default; removed when skipping is enabled).
2. `n test-php` — full suite is green.

**wpcom / Atomic (full behavior — needs a WP Cloud site with this change synced into jetpack-mu-wpcom, and without the standalone `newspack-blocks` plugin):**

3. Upload an image via the editor's **Image block**. Confirm no `newspack-article-block-*` derivatives are created (e.g. via EWWW's file listing, or by inspecting the uploads dir).
4. Upload the same image via the **Media Library**. Confirm the result matches step 3 — both paths now produce the same (core-only) derivatives.
5. Add a Homepage Posts / Blog Posts block using that image and confirm the thumbnail still renders correctly cropped (served via the image CDN), not stretched or full-size.

**Self-hosted regression:**

6. On a self-hosted site (no wpcom platform), upload an image and confirm the `newspack-article-block-*` crops are still generated as before — behavior is unchanged.

### Other information:

### DIFF
--- a/plugins/newspack-blocks/includes/class-newspack-blocks.php
+++ b/plugins/newspack-blocks/includes/class-newspack-blocks.php
@@ -25,6 +25,7 @@ class Newspack_Blocks {
 	 */
 	public static function init() {
 		add_action( 'after_setup_theme', [ __CLASS__, 'add_image_sizes' ] );
+		add_filter( 'intermediate_image_sizes_advanced', [ __CLASS__, 'maybe_skip_article_block_image_subsizes' ] );
 		add_post_type_support( 'post', 'newspack_blocks' );
 		add_post_type_support( 'page', 'newspack_blocks' );
 		add_action( 'jetpack_register_gutenberg_extensions', [ __CLASS__, 'disable_jetpack_donate' ], 99 );
@@ -540,6 +541,57 @@ class Newspack_Blocks {
 		add_image_size( 'newspack-article-block-square-tiny', 200, 200, true );
 
 		add_image_size( 'newspack-article-block-uncropped', 1200, 9999, false );
+	}
+
+	/**
+	 * Optionally skip generating the physical `newspack-article-block-*` crop
+	 * files when an image is uploaded.
+	 *
+	 * These sizes stay registered (via `add_image_sizes()`) so the blocks keep
+	 * resolving correctly-cropped URLs, but on platforms where an image CDN
+	 * resizes/crops on the fly (WordPress.com Simple and Atomic), the physical
+	 * crop files are redundant. Skipping them also keeps derivative generation
+	 * consistent across upload paths: on wpcom, `newspack-blocks` is loaded via
+	 * jetpack-mu-wpcom only for editor/REST requests, so an image uploaded through
+	 * the editor's Image block (REST) would otherwise get the crops while the same
+	 * image uploaded through the Media Library (async-upload.php) would not.
+	 *
+	 * @param array $sizes Associative array of image sub-sizes to be generated,
+	 *                     keyed by size name.
+	 * @return array Filtered sizes.
+	 */
+	public static function maybe_skip_article_block_image_subsizes( $sizes ) {
+		/**
+		 * Filters whether to skip generating the physical `newspack-article-block-*`
+		 * crop files on upload. Defaults to true on WordPress.com platform sites
+		 * (Simple and Atomic), where an image CDN handles cropping on the fly.
+		 * Self-hosted sites can opt in, e.g. when using the Jetpack Image CDN.
+		 *
+		 * @param bool $skip Whether to skip physical crop generation.
+		 */
+		$skip = apply_filters( 'newspack_blocks_skip_article_image_subsizes', self::is_wpcom_platform() );
+		if ( ! $skip ) {
+			return $sizes;
+		}
+
+		foreach ( array_keys( $sizes ) as $size_name ) {
+			if ( is_string( $size_name ) && 0 === strpos( $size_name, 'newspack-article-block-' ) ) {
+				unset( $sizes[ $size_name ] );
+			}
+		}
+		return $sizes;
+	}
+
+	/**
+	 * Whether the current site runs on the WordPress.com platform (Simple or Atomic).
+	 *
+	 * @return bool
+	 */
+	private static function is_wpcom_platform() {
+		if ( ! class_exists( '\Automattic\Jetpack\Status\Host' ) ) {
+			return false;
+		}
+		return ( new \Automattic\Jetpack\Status\Host() )->is_wpcom_platform();
 	}
 
 	/**
--- a/plugins/newspack-blocks/includes/class-newspack-blocks.php
+++ b/plugins/newspack-blocks/includes/class-newspack-blocks.php
@@ -544,28 +544,21 @@ class Newspack_Blocks {
 	}
 
 	/**
-	 * Optionally skip generating the physical `newspack-article-block-*` crop
-	 * files when an image is uploaded.
+	 * Skip generating the physical `newspack-article-block-*` crop files on upload.
 	 *
-	 * These sizes stay registered (via `add_image_sizes()`) so the blocks keep
-	 * resolving correctly-cropped URLs, but on platforms where an image CDN
-	 * resizes/crops on the fly (WordPress.com Simple and Atomic), the physical
-	 * crop files are redundant. Skipping them also keeps derivative generation
-	 * consistent across upload paths: on wpcom, `newspack-blocks` is loaded via
-	 * jetpack-mu-wpcom only for editor/REST requests, so an image uploaded through
-	 * the editor's Image block (REST) would otherwise get the crops while the same
-	 * image uploaded through the Media Library (async-upload.php) would not.
+	 * The sizes stay registered, so blocks still resolve correctly-cropped URLs;
+	 * we only skip writing the files where an image CDN crops on the fly (wpcom).
+	 * This also makes the Image block (REST) and Media Library upload paths behave
+	 * the same on wpcom, where they otherwise differ.
 	 *
-	 * @param array $sizes Associative array of image sub-sizes to be generated,
-	 *                     keyed by size name.
+	 * @param array $sizes Image sub-sizes to generate, keyed by size name.
 	 * @return array Filtered sizes.
 	 */
 	public static function maybe_skip_article_block_image_subsizes( $sizes ) {
 		/**
-		 * Filters whether to skip generating the physical `newspack-article-block-*`
-		 * crop files on upload. Defaults to true on WordPress.com platform sites
-		 * (Simple and Atomic), where an image CDN handles cropping on the fly.
-		 * Self-hosted sites can opt in, e.g. when using the Jetpack Image CDN.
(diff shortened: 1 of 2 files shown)


plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -542,6 +543,50 @@ public static function add_image_sizes() {
...
+	 */
+	public static function maybe_skip_article_block_image_subsizes( $sizes ) {
+		/**
+		 * Filters whether to skip physical `newspack-article-block-*` crop generation.
+		 * Defaults to true on WordPress.com (Simple/Atomic); self-hosted sites can
+		 * opt in, e.g. when using the Jetpack Image CDN.
+		 *
+		 * @param bool $skip Whether to skip physical crop generation.
+		 */
+		$skip = apply_filters( 'newspack_blocks_skip_article_image_subsizes', self::is_wpcom_platform() );
```
adekbadek: **Suggestion:** The skip defaults to `is_wpcom_platform()`, but the behavior the fix actually depends on is an on-the-fly image CDN that crops from the still-registered sizes — not merely running on the wpcom platform. An Atomic site with the Jetpack Image CDN disabled would satisfy this check, skip the physical crops, and then have no way to produce them: every article-block image from new uploads would fall back to the uncropped `large` size (wrong aspect ratios, heavier payloads) site-wide until noticed. The default keeps self-hosted safe and this degrades gracefully and is recoverable, so it's non-blocking — but consider detecting CDN availability at the point the filter decides to skip, or at least documenting that the effective precondition is "crop-synthesizing CDN active", so a CDN-disabled Atomic site keeps generating files.

**Consider:** gate the skip on the actual precondition (an on-the-fly image CDN is active and will crop from the registered sizes) rather than on the platform, so an Atomic site with the CDN turned off keeps generating physical crop files.

plugins/newspack-blocks/tests/test-image-sizes.php
```diff
@@ -0,0 +1,101 @@
...
+					return 0 === strpos( $key, 'newspack-article-block-' );
+				}
+			)
+		);
+	}
+
+	/**
+	 * Default (not skipping): all article block crops are left in place.
+	 */
+	public function test_article_block_subsizes_kept_by_default() {
```
adekbadek, line 85: **Suggestion:** `test_article_block_subsizes_kept_by_default` forces the filter to `__return_false`, so it exercises the *not-skipping* branch rather than the real default (`self::is_wpcom_platform()`). Since `\Automattic\Jetpack\Status\Host` isn't vendored in the test env, `is_wpcom_platform()` returns false locally — so the true default is observable and could be asserted directly with **no** filter attached, which would also cover the `class_exists` guard (the branch most likely to regress). At minimum, rename the test to reflect that it pins the override, not the default.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -542,6 +543,50 @@ public static function add_image_sizes() {
 		add_image_size( 'newspack-article-block-uncropped', 1200, 9999, false );
 	}
 
+	/**
+	 * Skip generating the physical `newspack-article-block-*` crop files on upload.
```
adekbadek: **Nit:** The docblock and inner-filter comment describe skipping "crop files" / where "an image CDN crops on the fly", but the `newspack-article-block-` prefix match also removes `newspack-article-block-uncropped` (registered with `crop => false` — a plain downscale, not a crop). This is very likely intended (the test asserts its removal, and the CDN resizes as well as crops), but a one-line note that *all* `newspack-article-block-*` sub-sizes are skipped — including the uncropped resize — would stop a future reader assuming that file is still written.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -542,6 +543,50 @@ public static function add_image_sizes() {
...
+		 *
+		 * @param bool $skip Whether to skip physical crop generation.
+		 */
+		$skip = apply_filters( 'newspack_blocks_skip_article_image_subsizes', self::is_wpcom_platform() );
+		if ( ! $skip ) {
+			return $sizes;
+		}
+
+		foreach ( array_keys( $sizes ) as $size_name ) {
+			if ( is_string( $size_name ) && 0 === strpos( $size_name, 'newspack-article-block-' ) ) {
```
adekbadek: **Nit:** The workspace targets PHP 8.3, so `str_starts_with( $size_name, 'newspack-article-block-' )` is the idiomatic, more legible form here (and in the mirrored test helper). Purely a modernization nit.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -542,6 +543,50 @@ public static function add_image_sizes() {
...
+	 *
+	 * The sizes stay registered, so blocks still resolve correctly-cropped URLs;
+	 * we only skip writing the files where an image CDN crops on the fly (wpcom).
+	 * This also makes the Image block (REST) and Media Library upload paths behave
+	 * the same on wpcom, where they otherwise differ.
+	 *
+	 * @param array $sizes Image sub-sizes to generate, keyed by size name.
+	 * @return array Filtered sizes.
+	 */
+	public static function maybe_skip_article_block_image_subsizes( $sizes ) {
```
adekbadek: **Nit:** `maybe_skip_article_block_image_subsizes( array $sizes ): array` and `is_wpcom_platform(): bool` would match the PHP 8.3 typing standard. The surrounding class is legacy and untyped, so adding types only here is a judgment call for consistency — low priority.

adekbadek (approved):
Pre-approving. Two suggestions, but the first one should really be addressed.

plugins/newspack-blocks/includes/class-newspack-blocks.php
chickenn00dle: Good catch — keyed the default on an active on-the-fly image CDN instead of the platform. New \`is_wpcom_image_cdn_active()\`: true on Simple (the platform image CDN is always on), and on Atomic only when the Jetpack Image CDN (\`photon\`) module is active — so an Atomic site with it disabled keeps generating the physical crops rather than dropping them. Self-hosted stays opt-in via the filter. (04e3c631b)

plugins/newspack-blocks/tests/test-image-sizes.php
chickenn00dle: Dropped the \`__return_false\` — \`test_article_block_subsizes_kept_by_default\` now runs with no filter attached, so it exercises the real default (\`is_wpcom_image_cdn_active()\` → false in the test env, since neither the Jetpack Status\Host nor the Jetpack class is loaded) and covers the \`class_exists\` guard. (04e3c631b)

plugins/newspack-blocks/includes/class-newspack-blocks.php
chickenn00dle: Added a note to the docblock: the \`newspack-article-block-\` prefix match removes every sub-size including \`newspack-article-block-uncropped\` (a \`crop => false\` downscale, not a crop) — the CDN resizes as well as crops, so none of them need a physical file. (04e3c631b)

plugins/newspack-blocks/includes/class-newspack-blocks.php
chickenn00dle: Switched to \`str_starts_with()\` here and in the mirrored test helper. (04e3c631b)

plugins/newspack-blocks/includes/class-newspack-blocks.php
chickenn00dle: Added return types: \`maybe_skip_article_block_image_subsizes( array $sizes ): array\` and \`is_wpcom_image_cdn_active(): bool\`. (04e3c631b)

adekbadek (approved):
All five threads are addressed by 04e3c631b — verified against the files, not just the replies.

The one that mattered: the skip is now keyed on `is_wpcom_image_cdn_active()` rather than the platform, so Atomic with the Jetpack Image CDN switched off keeps writing the physical crops instead of silently dropping them. The Atomic branch (`class_exists( 'Jetpack' ) && \Jetpack::is_module_active( 'photon' )`) matches the existing house check in `newspack-plugin/includes/plugins/class-jetpack.php:185`, and Simple short-circuits to true ahead of the `is_wpcom_platform()` branch, so the ordering is right.

The test change is real rather than cosmetic: with no filter attached, `test_article_block_subsizes_kept_by_default` exercises the actual default and the `class_exists` guard — `Automattic\Jetpack\Status\Host` is genuinely absent in this test env (no Jetpack packages in the plugin's `composer.json`, bootstrap loads only WP core plus the plugin) — and the companion skip-enabled test proves the filter is attached, so the default case can't pass vacuously. `str_starts_with()` is safe under the root `phpcs.xml` `testVersion 8.0-`, and the docblock note on `newspack-article-block-uncropped` plus the return types are in.

adekbadek (approved):
Re-checked at 04e3c631b — unchanged since the last pass, and all five threads are now resolved.

The one that mattered: the skip is keyed on `is_wpcom_image_cdn_active()` rather than the platform, so an Atomic site with the Jetpack Image CDN switched off keeps writing the physical crops instead of silently dropping them. Simple short-circuits to true ahead of the `is_wpcom_platform()` branch, so the ordering is right, and the Atomic check matches the existing house pattern.

The default-case test runs with no filter attached, so it exercises the real default and the `class_exists` guard rather than passing vacuously; the docblock note on `newspack-article-block-uncropped`, `str_starts_with()` and the return types are all in.

PR #578, opened 2026-07-08
fix(ads): remove colored gaps from empty Broadstreet ad slots

wil-gerken:
### All Submissions:

Reference: NPPM-2975

With Broadstreet as the ads provider and a custom ad background color enabled (Customize → Colors → Ad background), empty ad slots rendered as colored gaps — most visibly in the footer, where a disabled placement still produced a padded, colored strip.

## What this does

Only Broadstreet slots that actually contain an ad get the background and padding treatment. Empty slots — disabled zones, or zones that served no creative — get no styling from these rules.

Broadstreet's loader keeps its slot element in the page even when a zone is empty. The styling applied to every slot, filled or not, so those retained empty slots showed up as visible colored boxes. The rules now test for real ad content before applying, in both places that style these slots:

- newspack-theme (`inc/color-patterns.php`) — the background color;
- newspack-ads (`src/frontend/style.scss`) — the flex layout and padding.

Both changes must release together: if only the theme ships, empty slots lose the color but the ads plugin's padding still reserves a transparent gap. Shipping both keeps the two rules gating on the identical content test.

1. Enable a custom ad background color: Customize → Colors → Ad background → Custom, pick a color.
2. Add this markup — the structure Broadstreet's script leaves in the page for an empty zone — to a post (Custom HTML block): `<broadstreet-zone-container><broadstreet-zone zone-id="0" tracked="true"><div></div></broadstreet-zone></broadstreet-zone-container>`
3. View the post — an empty colored strip renders where nothing should be.

**Verify the fix:**
1. Load the same content with this branch — the empty skeleton renders nothing: no color, no padding, zero height.
2. Add this filled-zone skeleton to the same post: `<broadstreet-zone-container><broadstreet-zone zone-id="1" tracked="true"><div>test creative</div></broadstreet-zone></broadstreet-zone-container>` — it renders with the custom background color and padding around the image.
3. On a site with live Broadstreet zones: filled ads keep the background treatment; empty zones show no gap.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? (CSS-only mechanics — verified via compiled output, rendered pages, and live production Broadstreet DOM)

## Notes for reviewers
- The hydrated DOM shape is the crux: street.js always leaves a `<broadstreet-zone><div>` skeleton inside `<broadstreet-zone-container>`, including for empty zones, so a plain `:has(*)` gate cannot tell filled from empty. The gating selector is `broadstreet-zone-container:has(broadstreet-zone div *, a, iframe, img, video)`: the structural clause (elements inside the loader's div wrapper) is creative-type-agnostic, and the listed elements match anywhere beneath the container as a fallback for creatives rendered outside that wrapper.
- Validated against 7 of 7 hydrated containers (1 filled, 6 empty, 5 distinct zones) on a live production Broadstreet site, including an in-browser trial of these exact rules against that page's DOM: the filled zone painted and padded, the empty zones rendered at zero height with no styling.
- Old-browser fallback: browsers without `:has()` support drop each gated rule entirely, so filled ads lose the background and padding there (decorative loss only), and empty slots get no styling from these rules. Each rule is emitted standalone because an unrecognized selector invalidates an entire comma-separated rule — joining them to the existing ad-selector lists would take the other ad-container styling down with them.
- The previously duplicated `ads_color_hex` theme-mod default is hoisted into a variable so the two rules can't drift.
- CSS-only mechanics (the theme side is PHP-generated CSS). Verified in compiled output and on rendered pages against faithful hydration skeletons.

- Known coupling (flagged in review): the content test assumes street.js's empty skeleton stays exactly `<broadstreet-zone><div></div>` and that creatives render as elements. If street.js ever adds an element inside empty zones, the gap quietly returns; an element-less creative (a text node, or a CSS background image) would lose the decoration. Neither is testable from our side — a street.js rendering change is the thing to re-check if this regresses.

### Release risk

**Low-medium.** The change narrows existing decoration to filled slots only, and runs wherever Broadstreet plus a custom ad background color are enabled together. The content test was verified against live production Broadstreet DOM rather than assumptions, and old-browser degradation is decorative only. The theme and ads plugin each carry one of the two rules, so the releases must be coordinated. Rollback is a plain revert of both.


themes/newspack-theme/newspack-theme/inc/color-patterns.php
```diff
@@ -195,15 +195,29 @@ function newspack_custom_colors_css() {
 
 	// Set ads background color
 	if ( 'default' !== get_theme_mod( 'ads_color', 'default' ) ) {
-		$theme_css .= '
+		$ads_color_hex = get_theme_mod( 'ads_color_hex', '#ffffff' );
+		$theme_css    .= '
```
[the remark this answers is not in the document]
wil-gerken: Fixed in e3a2a5818 — spacing matched to the file's convention. (For the record: phpcs accepts both forms here, so this is consistency, not a standards fix.)

themes/newspack-theme/newspack-theme/inc/color-patterns.php
```diff
@@ -195,15 +195,29 @@ function newspack_custom_colors_css() {
 
 	// Set ads background color
 	if ( 'default' !== get_theme_mod( 'ads_color', 'default' ) ) {
+		$ads_color_hex = get_theme_mod( 'ads_color_hex', '#ffffff' );
 		$theme_css .= '
 			.site .entry .entry-content .scaip .newspack_global_ad,
 			.site .entry .entry-content .scaip .widget_newspack-ads-widget,
 			.newspack_global_ad,
```
adekbadek, line 202: **Suggestion:** The `:has()` gate zeroes the inner `broadstreet-zone-container`, but for a Broadstreet placement rendered through Newspack's placement system the visible gap is emitted one level up: `Placements::render_placement_ad()` wraps the container in `.newspack_global_ad`, which this first (untouched) rule still paints unconditionally, and under `.custom-ad-bg` its `.newspack-broadstreet-ad` child carries 8px top/bottom margins. Reproduced in-browser: an empty enabled placement still renders a ~16px colored strip. The reported NPPM-2975 case is a *disabled* footer placement (a bare container, no wrapper) and is fully fixed — so this is a distinct, arguably out-of-scope path, but please confirm against the reporting site's footer DOM or note it as a scoped follow-up, since a publisher enabling a Broadstreet placement that serves no creative would report the same symptom.

image

themes/newspack-theme/newspack-theme/inc/color-patterns.php
```diff
@@ -195,15 +195,29 @@ function newspack_custom_colors_css() {
...
+		// serves no creative, and street.js always leaves a
+		// <broadstreet-zone><div> skeleton inside them — so :has(*) cannot
+		// tell filled from empty. A zone is filled when elements exist inside
+		// that div wrapper (or a creative element appears anywhere in the
+		// container); empty zones carry only the empty skeleton and must not
+		// render as a colored gap. Kept as a standalone rule: in browsers
+		// without :has() support an unknown selector invalidates the entire
+		// rule, which must not take the selectors above down with it.
+		$theme_css .= '
+			broadstreet-zone-container:has(broadstreet-zone div *, a, iframe, img, video) {
```
adekbadek, line 221: **Nit:** The content test relies on street.js's empty skeleton being exactly `<broadstreet-zone><div></div></broadstreet-zone>`. It's a two-sided coupling to a third-party loader with no test guarding it: if street.js ever seeds the empty skeleton with a nested element, `broadstreet-zone div *` false-positives and the gap returns; a creative rendered as element-less content (a text node, or a CSS background-image on the wrapper div) false-negatives and loses its decoration. Both are unlikely for real Broadstreet creatives and strictly better than today — flagging so the assumption is on record and a street.js change is the known failure mode to watch.

plugins/newspack-ads/src/frontend/style.scss
```diff
@@ -211,9 +211,16 @@
...
+	// persist in the DOM even when the zone is unfilled, and street.js always
+	// leaves a <broadstreet-zone><div> skeleton inside them (so :has(*) can't
+	// tell filled from empty). A filled zone has elements inside that div
+	// wrapper or a creative element in the container; empty zones must not
+	// render as a padded colored gap. In browsers without :has() support the
+	// unknown selector drops just this rule (filled ads lose the decoration
+	// there; cosmetic degradation only).
 	/* stylelint-disable selector-type-no-unknown */
-	broadstreet-zone-container {
+	broadstreet-zone-container:has(broadstreet-zone div *, a, iframe, img, video) {
```
adekbadek, line 226: **Nit:** The identical `:has(broadstreet-zone div *, a, iframe, img, video)` test is now hand-maintained in two independently-versioned deployables — this SCSS file (plugin) and a PHP string in the theme (color-patterns.php:219). Unavoidable given the plugin/theme split, but a one-line cross-reference comment in each (pointing at the other file and NPPM-2975) would make the coupling — and the "must ship together" requirement — discoverable from either side.

adekbadek (changes requested):
One significant issue – no creative results in an empty padded div.

themes/newspack-theme/newspack-theme/inc/color-patterns.php
wil-gerken: Thanks @adekbadek! Confirmed — the reported strip is the bare-container case this PR gates. On the reporting site's homepage and an article page, the empty footer `broadstreet-zone-container` is a direct child of `div.footer-branding`, with no `.newspack_global_ad` or `.newspack-broadstreet-ad` ancestor. Every empty container inspected on those two pages renders bare; the only wrapped container is the filled above-header ad.

The enabled-placement case is real, and distinct from this one. In the `nppm2975` env with this fix applied, I injected the enabled-placement DOM structure to isolate the CSS: the 16px strip appears, and computed styles attribute it entirely to the untouched `.newspack_global_ad` background rule and the `.newspack_global_ad > *` margins. The gated container inside stays collapsed. It's out of scope here, so I filed NPPM-3096 to gate the placement wrapper the same way.

themes/newspack-theme/newspack-theme/inc/color-patterns.php
wil-gerken: On the record now — added a "Known coupling" bullet to the notes for reviewers naming the skeleton assumption and both failure modes.

plugins/newspack-ads/src/frontend/style.scss
wil-gerken: Added in 8aaed3df2 — each comment block now points at its twin and NPPM-2975.

wil-gerken:
Addressed the review: the reported case is confirmed as the bare-container path this PR fixes (evidence on the thread), the enabled-placement case is filed as NPPM-3096, the street.js assumption is now a "Known coupling" bullet in the notes for reviewers, and the two files cross-reference each other (8aaed3df2). Ready for another look.

themes/newspack-theme/newspack-theme/inc/color-patterns.php
adekbadek: Confirmed — the code comments now carry the skeleton assumption on both sides too, which is where someone debugging a returned gap will actually be looking. Resolving.

plugins/newspack-ads/src/frontend/style.scss
adekbadek: Verified — both comment blocks point at their twin and NPPM-2975. Resolving.

themes/newspack-theme/newspack-theme/inc/color-patterns.php
adekbadek: That settles it — the reporting DOM is the bare-container case, and splitting the placement-wrapper gap into NPPM-3096 is the right call rather than widening this hotfix. Resolving.

adekbadek (approved):
The empty-strip case I raised is confirmed as the untouched placement wrapper, correctly scoped out to NPPM-3096, and the duplicated content test is now cross-referenced on both sides. Splitting the theme rule so a dropped `:has()` selector cannot invalidate the other ad-background selectors is the right shape. All three threads resolved.

PR #579, opened 2026-07-09
fix: Contributor/Author newsletter + content-gate permission errors (NPPM-2982)

jason10lee:
### All Submissions:

Contributor and Author roles (commonly used for freelance and junior writers) can't properly use the newsletter editor, and on sites with content gating enabled they can't save posts at all. Opening a newsletter shows "You cannot use this resource." notices, the email HTML fails to refresh ("Failed to refresh email HTML"), and a send can even go out with stale content; saving a post errors with "…not allowed to edit the `newspack_content_restriction_is_exempt` custom field." The common cause is a set of REST permission checks that require `edit_others_posts` — a capability these roles lack — even when the user is only acting on their own content. This PR makes two independent fixes (one per plugin; no cross-plugin dependency, so each can release independently):

**newspack-newsletters** — "You cannot use this resource." / "Failed to refresh email HTML":
- `post-mjml` render now checks `edit_post( $post_id )` on the actual newsletter (scoped on `post_id` only).
- `layouts` list now allows `edit_posts`.
- `color-palette` (auto-POSTed by the editor on every load) now returns success at `edit_posts` but only **writes** the site-wide palette option for `edit_others_posts` — so a Contributor's editor stops erroring on every load without gaining a new write. Filterable via `newspack_newsletters_color_palette_capability`; the UI/effect divergence is documented in code.
- The shared `api_authoring_permissions_check` (16 routes incl. `send-lists`, `layouts/{id}/test`, Ads, and all ESP controllers) is left untouched at `edit_others_posts` — send/audience/ESP surfaces are deliberately not broadened.

**newspack-plugin** — "…not allowed to edit the `newspack_content_restriction_is_exempt` custom field":
- The content-gate exemption meta is registered only when the content-gate feature is enabled.
- Exemption stays editor-only (`auth_callback` = `edit_others_posts`).
- A `rest_pre_insert_*` filter strips the exempt key from an unauthorized save instead of 403-ing the whole post, so lower-role saves succeed and the value is simply left unchanged.

Addresses NPPM-2982 — Contributors and Authors hit permission errors authoring newsletters and saving content.

1. Ensure `newspack-newsletters` is active; create a Contributor and an Author user.
2. As the Contributor, open **Newsletters → Add New**. Confirm the editor loads and saves with no "You cannot use this resource." / "Failed to refresh email HTML" error, and that Update/preview refreshes the email HTML. (With a configured ESP, one `send-lists` notice remains — expected/by design.)
3. Confirm the Contributor can only act on their **own** newsletter drafts (open another user's newsletter → refresh/edit is denied).
4. Enable content gates (`define( 'NEWSPACK_CONTENT_GATES', true )`). As the Contributor, edit and save any post — the save succeeds with no "…not allowed to edit `newspack_content_restriction_is_exempt`…" error, and the exemption value is unchanged.
5. As an Editor, confirm the access-control exemption toggle still works and writes.
6. Run PHP tests in both `newspack-newsletters` and `newspack-plugin`.

### Other information:


plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-restriction-control.php
```diff
@@ -0,0 +1,101 @@
...
+ */
+class Test_Content_Restriction_Control extends WP_UnitTestCase {
+
+	/**
+	 * Reset registered meta between tests.
+	 */
+	public function tear_down() {
+		unregister_meta_key( 'post', Content_Restriction_Control::IS_EXEMPT_META_KEY, 'post' );
+		parent::tear_down();
+	}
```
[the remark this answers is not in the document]
jason10lee: Thanks — addressed in 7da7e84dd. `tear_down()` now unregisters the exempt meta for every supported post type (looping `get_available_post_types()`), not just `post`. Relatedly, the `rest_pre_insert_*` strip filter is now registered unconditionally at `init` (in `register_meta_guards()`) rather than inside the feature-gated `register_meta()`, so its lifetime no longer depends on meta registration. The full content-gate test group (458 tests) now passes with no ordering sensitivity.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -889,7 +889,7 @@ public static function rest_api_init() {
 			[
 				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ __CLASS__, 'api_get_layouts' ],
-				'permission_callback' => [ __CLASS__, 'api_authoring_permissions_check' ],
+				'permission_callback' => [ __CLASS__, 'api_edit_posts_permissions_check' ],
```
adekbadek, line 892: **Suggestion:** layouts route now exposes send/audience defaults to Contributors/Authors — Moving the `layouts` list from `edit_others_posts` to `edit_posts` also exposes the payload `get_layouts()` returns, whose per-layout `campaign_defaults` meta includes `senderEmail`, `send_list_id`, and `send_sublist_id` — previously readable only by editors. The layout picker doesn't just display these: picking a layout copies `campaign_defaults` into the draft, and the editor sidebar writes `senderEmail`/`send_list_id`/`send_sublist_id` into the newsletter's own send meta and calls `savePost()`, so a lower-role author's draft silently inherits a sender identity and audience list it never chose. The PR frames this route as "editor-support content, not send/audience configuration," which isn't quite accurate given `campaign_defaults`. Decide whether this is acceptable or whether `campaign_defaults` should be stripped from the layouts payload for users below `edit_others_posts`.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -962,8 +962,21 @@ public static function rest_api_init() {
...
+		 * reports success while the option write is a no-op. The write capability is
+		 * filterable via `newspack_newsletters_color_palette_capability`.
+		 */
+		$capability = apply_filters( 'newspack_newsletters_color_palette_capability', 'edit_others_posts' );
+		if ( current_user_can( $capability ) ) {
+			self::update_color_palette( json_decode( $request->get_body(), true ) );
+		} else {
+			Newspack_Newsletters_Logger::log( 'Color palette write skipped: current user lacks the "' . $capability . '" capability.' );
+		}
 		return \rest_ensure_response( [] );
```
adekbadek: **Suggestion:** color-palette always returns 200, masking real write failures for authorized users — Returning `rest_ensure_response( [] )` unconditionally is the right call for the permission divergence (it avoids the spurious per-load editor notice for lower roles, and the skip is logged). But it now discards `update_color_palette()`'s outcome for *authorized* users too, so a genuine `update_option` failure reads as success. The route's contract is now explicitly "always 200"; consider returning a small body distinguishing written from skipped (e.g. `[ 'updated' => $did_write ]`) so a client and future maintainer can tell a no-op apart from a real write. Non-blocking.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -488,5 +492,45 @@ public static function register_meta() {
...
+	 * @param stdClass        $prepared_post Prepared post object (returned unchanged).
+	 * @param WP_REST_Request $request       Incoming request.
+	 * @return stdClass
+	 */
+	public static function strip_unauthorized_exempt_meta( $prepared_post, $request ) {
+		$meta = $request['meta'];
+		if (
+			is_array( $meta ) &&
+			array_key_exists( self::IS_EXEMPT_META_KEY, $meta ) &&
+			! current_user_can( 'edit_others_posts' )
```
adekbadek: **Suggestion:** exempt-meta write capability duplicated in two places — The exemption's write capability `edit_others_posts` is now encoded independently in the meta's `register_meta` `auth_callback` (line 489) and in `strip_unauthorized_exempt_meta` (line 528). They must stay in lockstep or the strip guard silently desyncs from the field it guards — e.g. if the `auth_callback` were ever made filterable (as newsletters did for the palette via `newspack_newsletters_color_palette_capability`), the strip would keep enforcing the old capability. Consider extracting a shared private helper so the strip is defined as "the inverse of what the field authorizes" rather than a re-derivation of the same literal.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -470,6 +471,9 @@ public static function get_gate_layout_id( $post_id = null ) {
 	 * Register post meta for the exemption flag.
 	 */
 	public static function register_meta() {
+		if ( ! Content_Gate::is_newspack_feature_enabled() ) {
```
adekbadek, line 489: **Nit:** register_meta feature-gate diverges from the editor-panel enqueue condition — `register_meta` now early-returns unless the content-gate feature is enabled, but the editor panel that reads/writes this meta (`Content_Gate::enqueue_block_editor_assets`) gates on `edit_others_posts` + a non-empty `get_gates()` count, not the feature flag. If a site ever had gate posts while the flag were off, the panel would render but the exempt key would be absent from the REST schema and the toggle couldn't persist. In practice this is essentially unreachable (gate management is itself flag-gated, so `get_gates()` is empty when the flag is off) and runtime enforcement is unaffected (the exemption is read via `get_post_meta()` directly, not REST) — but the two conditions would ideally share one gate for robustness.

adekbadek (approved):
Sound, well-scoped, well-tested permission fix with no blockers — pre-approving with suggestions.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
jason10lee: Confirmed the chain: the layouts route at `edit_posts` hands `campaign_defaults` to sub-editors, `insertLayout` copies it into the draft, and the sidebar effect writes `senderEmail`/`send_list_id`/`send_sublist_id` into the newsletter's own meta and auto-saves. `api_get_layouts` now strips `campaign_defaults` from the payload when the user lacks `edit_others_posts`, so the send/audience surface stays editor-only; the picker still applies content, colors, and fonts (`insertLayout` already guards on the key). Covered by `test_layouts_payload_strips_campaign_defaults_for_contributor` and `..._keeps_..._for_editor`. Also reworded the `api_edit_posts_permissions_check` docblock so its "not send/audience configuration" claim points at the strip instead of overstating. 9f77066.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
jason10lee: The callback now returns `[ 'updated' => $did_write ]`. For an authorized user that is `update_color_palette()`'s result, so a genuine `update_option` failure no longer reads as success; for a permission-skipped role it is false. A comment notes `update_option` also returns false on an unchanged value, so `updated` means the stored palette changed rather than "no error". Tests assert both the editor (true) and lower-role (false) cases. 9f77066.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
jason10lee: Extracted `current_user_can_edit_exemption()` as the single source of truth. Both `register_meta`'s `auth_callback` and `strip_unauthorized_exempt_meta` now call it, so the strip is the inverse of what the field authorizes rather than a re-derivation of the literal; if the capability is ever made filterable, the two stay in lockstep. The existing strip/auth tests still pass. aaf943b.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
jason10lee: `enqueue_block_editor_assets` now early-returns on `! is_newspack_feature_enabled()`, the same gate `register_meta` uses, so the panel and the meta schema share one condition. As you noted it is essentially unreachable in practice (gate management is itself flag-gated, so `get_gates()` is empty when the flag is off), so we did not add a dedicated test: exercising the enqueue needs full editor context and a built asset, and the guard mirrors `register_meta`'s already-tested gate. aaf943b.

PR #589, opened 2026-07-09
fix(reader-revenue): payment method merge field shows card details

wil-gerken:
The `*PAYMENT_METHOD*` placeholder in reader revenue receipt and welcome emails printed the word "Card" plus the payment gateway's internal ID — donors saw "Card – stripe" on their receipts, while the placeholder's description promised the last four digits of the card used.

Reference: NPPM-2903

### What this does

Builds the payment method label from what the site knows about the payment:

| Donation type | What the site has stored | Receipt shows |
|---|---|---|
| Recurring (signup and renewal) | Saved card on the customer + the gateway's reference on the order | "Visa ending in 4242" |
| One-time, customer saved their card | Saved card on the customer + the gateway's reference on the order | "Visa ending in 4242" |
| One-time, plain checkout | Nothing about the card | "Credit / Debit Card" |

Two lookups feed the label. Payment tokens attached to the order itself are read first (how WooPayments stores cards). When there are none — the Stripe gateway saves cards against the *customer*, leaving only its payment reference in order meta — the label matches that reference (`_stripe_source_id`) against the customer's saved cards. Review feedback caught that the first round read only the order tokens, which the Stripe gateway never writes; delivered-mail checks in an isolated environment now confirm all three rows above, including a subscription renewal order.

On a plain one-time checkout the card number never touches the site — it goes straight to the payment processor — so the gateway's customer-facing title is the honest ceiling there. (The fix avoids live processor lookups during email delivery; they would add latency and a failure point.) Because the fallback is the gateway's own title, a PayPal order says "PayPal", never "Card – paypal". A saved card missing its brand or last four falls back to the title rather than printing a partial label.

The placeholder description in the email editor now reads "card brand and last four digits, when available", matching what the field actually does.

### Behavior changes

Three deliberate changes ride along, all in the same send path:

1. **Sites with the receipt email but no donation products configured stop sending Newspack-styled emails for shop orders.** `Donations::get_order_donation_product_id()` now returns its documented `false` instead of null when no donation products exist, so the donation-order guard holds. Previously, every completed order on such a site got the donation-styled receipt, and an order with no line items crashed the request. **This includes the welcome email**: a reader who registers during a modal checkout on such a site — a subscription or membership purchase, for example — no longer receives the Newspack welcome email, since it shares the donation-gated send path and always has on sites with donations configured. Readers still receive WooCommerce's default order email; verified live on a deconfigured site (only the WooCommerce email arrives, no crash). Worth a changelog line so support recognizes "my branded receipt stopped" reports — and the welcome email is the part to name when that conversation happens.
2. **The receipt/welcome resend guard now works.** The already-sent marker was written to the in-memory order but never persisted, so moving an order out of `completed` and back in resent the email. The marker is now saved; each order gets its receipt (or welcome email) once.
3. **The `newspack_donation_order_processed` hook no longer fires for non-donation orders on unconfigured sites.** Same root cause: `order_paid()`'s `false ===` check now behaves per its docblock. The hook's only listener in this codebase already discarded those fires.

### How to test

Reproduce the problem (on `release`, without this branch):

1. On a dev site with donations configured and the Receipt email enabled (Engagement → Emails), create a completed donation order paid by card.
2. Open the receipt email (MailHog at `localhost:8025` on a dev site). The "Payment method" block reads "Card – stripe".

Verify the fix (on this branch):

1. Create the same completed donation order — a plain one-time checkout, no saved card. The receipt's "Payment method" block reads "Credit / Debit Card".
2. Save a card for the customer (a `WC_Payment_Token_CC` with gateway `stripe`, card type `visa`, last4 `4242`, and a token value) and set the same value as the order's `_stripe_source_id` meta before completing — how the Stripe gateway stores a saved-card payment. The receipt reads "Visa ending in 4242".
3. In Engagement → Emails → Receipt, the placeholder list describes `*PAYMENT_METHOD*` as "card brand and last four digits, when available".
4. Regression (the guard fix): with the Receipt email enabled and no donation products configured, complete an order. Only WooCommerce's default email sends and nothing crashes.
5. Resend guard: move a completed donation order to processing and back to completed. The receipt does not send a second time.
6. Unit coverage for all of the above: `n test-php --group payment_method_label` in `plugins/newspack-plugin` (14 tests).

### What it looks like

Delivered receipts from the verification run (real sends through the completed-order path, read from MailHog):

Saved card, stored as the Stripe gateway stores it — the card details recovered from the customer's token:

Receipt for a saved-card donation showing Payment method: Visa ending in 4242

Plain one-time checkout — the gateway's customer-facing title, since the card never touches the site:

Receipt for a plain one-time donation showing Payment method: Credit / Debit Card

The updated placeholder description in the email editor:

Email editor sidebar describing the PAYMENT_METHOD placeholder as card brand and last four digits, when available

### Notes for reviewers

- `WooCommerce_Connection::get_payment_method_label( $order )`: order-attached tokens first (WooPayments), then the customer-token match on `_stripe_source_id` (Stripe; exact token-string match, scoped to the order's customer and gateway, so an unmatched or deleted card degrades to the gateway title and never borrows another card's digits), then `get_payment_method_title()`, then "Card". Guest orders skip the customer lookup.
- Renewal orders: WooCommerce Subscriptions replaces the standard completed-order notification with its own renewal email, so the customizable receipt fires for a renewal only when that notification chain runs; the label handles the renewal order shape either way (test covers it).
- `Donations::get_order_donation_product_id()` is fixed at the source (bare `return` → `return false`, per its `int|false` docblock), which makes every caller correct without call-site null handling.
- `send_customizable_receipt_email()` returned `false` after a successful send and `true` on failure — inverted from its docblock. Fixed; the sole caller is a WooCommerce action dispatched via `do_action_ref_array`, which discards the return.
- The label passes through `wp_strip_all_tags()` (the same treatment `*AMOUNT*` gets) — the gateway title and the card-brand label both pass through public filters.
- Test mocks: customer-scoped token lookups faithful to `WC_Payment_Tokens::get_customer_tokens` (including the guest guard), a `wc_get_credit_card_type_label` mock faithful to the real label map (Stripe's `amex` slug renders "Amex ending in 0005" — the real WooCommerce output, pinned by a test), and `WC_Order` re-hydration. The mock cannot model persistence (lookups return the same instance), so the sent-marker save is pinned via a save-call counter with a comment explaining why.

### Release risk

Medium — the email path is reader-facing and the guard fix changes behavior for the receipt-without-donations sites described above. Implementation risk is limited: the change is additive (a new label helper, no signature changes), test-covered (14 tests; full plugin suite green), and every fallback degrades to the gateway title. The welcome-email scoping on unconfigured sites is the change support should know about.

### All Submissions:

### Other information:


plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -495,6 +495,50 @@ public static function force_allow_failed_payment_retry( $should_retry ) {
...
+					continue;
+				}
+				$last4 = $token->get_last4();
+				$type  = $token->get_card_type();
+				// Both parts are needed for a meaningful label — some gateways store
+				// tokens without a brand or last4; fall through to the gateway title.
+				if ( ! $last4 || ! $type ) {
+					continue;
+				}
+				$brand = function_exists( 'wc_get_credit_card_type_label' ) ? \wc_get_credit_card_type_label( $type ) : ucwords( (string) $type );
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 5e4c5feb2 — the fallback now normalizes `-`/`_` to spaces, matching the test mock's behavior.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-connection-payment-method.php
```diff
@@ -0,0 +1,183 @@
...
+	 * which the guard's `!== false` comparison let through, and the code then
+	 * fataled on `$item->get_product_id()` with an empty items array.
+	 */
+	public function test_receipt_email_bails_on_order_without_donation_items() {
+		// Ensure the receipt email is enabled, so the guard is what stops the send.
+		add_filter(
+			'newspack_email_configs',
+			function ( $configs ) {
+				$configs[ Reader_Revenue_Emails::EMAIL_TYPES['RECEIPT'] ] = [
+					'name'        => Reader_Revenue_Emails::EMAIL_TYPES['RECEIPT'],
```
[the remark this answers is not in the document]
wil-gerken: `WP_UnitTestCase` snapshots `$wp_filter` in `set_up()` and restores it in `tear_down()`, so hooks added inside a test are discarded before the next test runs — order-dependent leakage can't occur here. Same pattern as the existing `emails.php` test in this suite.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -495,6 +495,50 @@ public static function force_allow_failed_payment_retry( $should_retry ) {
...
+	 * which is available for saved-card and subscription orders. Falls back to
+	 * the payment gateway's customer-facing title (e.g. "Credit / Debit Card"),
+	 * since the card details of a one-time payment are not stored locally.
+	 *
+	 * @param \WC_Order $order The order.
+	 *
+	 * @return string Payment method label.
+	 */
+	public static function get_payment_method_label( $order ) {
+		if ( class_exists( 'WC_Payment_Tokens' ) && method_exists( $order, 'get_payment_tokens' ) ) {
```
adekbadek: **Blocker:** The token branch never executes on the Stripe gateway, so the last four digits never render — `WC_Order::get_payment_tokens()` returns only the tokens a gateway explicitly attached to *that order* via `$order->add_payment_token()`. The WooCommerce Stripe Gateway never does this: it creates **customer-scoped** tokens and records the gateway reference on the order as `_stripe_source_id` instead. There is not a single `add_payment_token()` call in WooCommerce core (10.6.1), WooCommerce Subscriptions, or the Stripe gateway (10.5.3) — so on a Stripe site this loop iterates over an empty array for one-time, subscription-parent and renewal orders alike, and every receipt falls through to `get_payment_method_title()`.

Verified end-to-end in an isolated env by sending real receipts through `woocommerce_order_status_completed_notification` and reading the delivered mail:

| Order shape | Delivered receipt |
|---|---|
| One-time, no saved card | `Credit / Debit Card` (matches the PR) |
| **Customer has a saved Visa •4242** (stored as Stripe stores it) | `Credit / Debit Card` — the PR's table promises `Visa ending in 4242` |
| Token attached to the order (as WooPayments does) | `Visa ending in 4242` |
| Subscription renewal order | `Card` |

So the branch works, but only for gateways that attach tokens to orders — which is the minority. For the recurring and saved-card donors [NPPM-2903](https://linear.app/a8c/issue/NPPM-2903) is actually about, the merge field still does not show the card digits, and the re-worded description ("card brand and last four digits, when available") continues to overpromise.

The details *are* recoverable: the order carries `_stripe_source_id`, which matches `$token->get_token()` on the customer's saved tokens. Resolving that recovered `Visa ending in 4242` in the same env:

```php
$source_id = $order->get_meta( '_stripe_source_id', true );
foreach ( \WC_Payment_Tokens::get_customer_tokens( $order->get_customer_id(), $order->get_payment_method() ) as $token ) {
 if ( $token instanceof \WC_Payment_Token_CC && $token->get_token() === $source_id ) {
 // brand + last4 available here.
 }
}
```

Keeping the existing order-token lookup as the first attempt (it is what WooPayments populates) and adding this as the second would deliver the fix for the gateway most publishers run. Failing that, the field's description and the PR's behaviour table should be re-scoped to what actually ships.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -510,7 +554,9 @@ public static function send_customizable_receipt_email( $order_id ) {
 		}
 
 		// If there are no donation products in the order, do not override the default WC receipt email.
-		$has_donation_product = \Newspack\Donations::get_order_donation_product_id( $order->get_id() ) !== false;
+		// Note: get_order_donation_product_id() returns null (not false) when no donation
+		// products are configured at all, so an emptiness check is required here.
+		$has_donation_product = ! empty( \Newspack\Donations::get_order_donation_product_id( $order->get_id() ) );
```
adekbadek: **Blocker:** The bundled guard change silently stops receipt and welcome emails on sites with no donation products — The `! empty()` guard is correct in isolation — `get_order_donation_product_id()` returns a positive product ID, `false`, or `null`, so no legitimate donation order can be filtered out by an emptiness check, and it does fix a real fatal on an order with no line items.

The concern is blast radius, not correctness. Because the helper bare-returns `null` when a site has **no donation products configured at all**, the old `!== false` test passed for *every* order on such a site. Those sites are therefore sending the Newspack donation-styled receipt for every completed WooCommerce order today — and, for a checkout that also registers an account, the Newspack **welcome** email. After this patch they send neither, falling back to WooCommerce's defaults. Both email posts are auto-created with a published status, so this is not an opt-in population: it is any Newspack store selling subscriptions, memberships or merch without a donation product, plus any site whose donation product was later deleted or un-flagged.

The narrower behaviour is arguably the right one — a donation-styled receipt for a t-shirt order is its own bug — so the ask is not to drop the change but to land it deliberately: split it into its own PR, or keep it here with a release note and a row in the PR's behaviour table. As written it rides a patch release titled after a card-digits fix, and the first signal will be a publisher asking why their customised receipt template stopped going out.

**Risk if not addressed:** The welcome email is the sharper edge — it is the one-shot message pointing a brand-new reader at their account, gated by a sent-once meta flag. A reader who registers during checkout in the window between this shipping and anyone noticing never receives it, and rolling the release back does not send it retroactively.

**Consider:** The cleanest fix is at the source rather than at this call site: `Donations::get_order_donation_product_id()` (class-donations.php:379) does a bare `return;` when no donation products exist, even though its own docblock promises `int|false`. Returning `false` there makes every caller correct without any of them having to know about the null case, and removes the need for the explanatory comment added here.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -495,6 +495,50 @@ public static function force_allow_failed_payment_retry( $should_retry ) {
...
+					sprintf(
+						/* translators: 1: card brand, e.g. "Visa". 2: the card's last four digits. */
+						__( '%1$s ending in %2$s', 'newspack-plugin' ),
+						$brand,
+						$last4
+					)
+				);
+			}
+		}
+		$title = $order->get_payment_method_title();
```
adekbadek, line 542: **Suggestion:** Gateway title reaches the email HTML unescaped — `Emails::get_email_html()` substitutes each placeholder with a bare `str_replace()`, and the receipt template renders `*PAYMENT_METHOD*` inside a `<strong>` tag — so whatever this returns lands in an HTML context unescaped. The value now originates from `get_payment_method_title()` (order data written by the gateway from an admin-configured title, and passed through the `woocommerce_gateway_title` filter) and from `wc_get_credit_card_type_label()` (filterable via `woocommerce_get_credit_card_type_label`). Neither is reader-controlled, so there is no practical XSS vector here and this is no worse than the neighbouring `*PRODUCT_NAME*`. But the `*AMOUNT*` placeholder right above already defends itself with `wp_strip_all_tags()`, and giving the returned label the same treatment costs nothing and keeps a gateway or a third-party filter from injecting markup into a transactional email.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
adekbadek: **Nit:** Redundant defensive guards — `method_exists( $order, 'get_payment_tokens' )` is redundant — the method has lived on `WC_Abstract_Order` since WooCommerce 2.6, this class only loads when WooCommerce is active, and the sole caller has already confirmed `is_a( $order, 'WC_Order' )`. The `class_exists( 'WC_Payment_Tokens' )` check already covers the "WooCommerce isn't loaded" case the guard appears to defend against. Dropping the extra check would let the happy path read straight through.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -495,6 +495,50 @@ public static function force_allow_failed_payment_retry( $should_retry ) {
...
+				}
+				$last4 = $token->get_last4();
+				$type  = $token->get_card_type();
+				// Both parts are needed for a meaningful label — some gateways store
+				// tokens without a brand or last4; fall through to the gateway title.
+				if ( ! $last4 || ! $type ) {
+					continue;
+				}
+				$brand = function_exists( 'wc_get_credit_card_type_label' ) ? \wc_get_credit_card_type_label( $type ) : ucwords( str_replace( [ '-', '_' ], ' ', (string) $type ) );
+				return trim(
```
adekbadek: **Nit:** Dead `trim()` — The `continue` four lines above guarantees both `$brand` and `$last4` are non-empty, and the format string `'%1$s ending in %2$s'` has no leading or trailing whitespace, so `trim()` can never remove anything. It also signals to a future maintainer that a brandless `"ending in 4242"` is a state this code can produce — which the guard directly above prevents and `test_label_ignores_token_without_brand` pins. Removing it makes the invariant self-evident.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-connection-payment-method.php
```diff
@@ -0,0 +1,183 @@
...
+		$order = new WC_Order(
+			[
+				'status'        => 'completed',
+				'customer_id'   => 1,
+				'total'         => 20,
+				'billing_email' => 'reader@example.com',
+				'items'         => [],
+			]
+		);
+		self::assertFalse(
```
adekbadek, line 429: **Nit:** The headline assertion cannot fail for the reason its message claims — `send_customizable_receipt_email()` returns `false` on every bail *and also* after a **successful** send (`if ( $sent ) { …; return false; }`), returning `true` only when `Emails::send_email()` fails — the inverse of what its docblock promises. So `assertFalse( …, 'Orders without donation items should never send the customizable receipt.' )` is not a discriminating assertion: `false` is exactly what a successful send returns. What actually turns this test red without the fix is the fatal on the null `$item`, plus the `meta_exists()` assertion on the next line, which is the one doing the work. Either drop the `assertFalse` in favour of the observable, or fix the inverted return and its docblock so the assertion means what it says. (The inverted return is pre-existing and harmless in production — WooCommerce fires the notification with `do_action_ref_array()` and discards the return.)

Two related gaps in the same suite: the mock `wc_get_credit_card_type_label()` implements only the `ucwords()` fallback, so real WooCommerce's label map is never exercised — worth knowing that Stripe stores `card_type` as `amex` while real WC keys its map on `american express`, so an Amex card renders `Amex ending in 4242`, not `American Express ending in 4242`. And for a suite whose subject is recurring-donor receipts, no test covers a subscription parent or renewal order — the exact case where the token assumption above breaks down.

adekbadek (changes requested):
The label helper is well-built and its fallback chain is correctly ordered, but it reads the card details from a place the Stripe gateway never writes to, so the fix does not reach the donors the ticket is about.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
wil-gerken: Adopted (5feee82e5; restructured in ceab0f101 to keep the order-token read first). The order-attached token is the first lookup, as WooPayments attaches the card token to the order. A customer-token match on `_stripe_source_id` is the second: an exact match on the token string, scoped to the order's customer and gateway. An unmatched or deleted card falls back to the gateway title rather than borrowing another card's digits.

Re-ran your table through real sends in an isolated env:

- one-time, no saved card → "Credit / Debit Card"
- saved Visa, customer-scoped → "Visa ending in 4242"
- order-attached token → "Visa ending in 4242"
- subscription renewal → "Visa ending in 4242"

One note from that run: WooCommerce Subscriptions replaces the standard completed notification with its own renewal email, so to reach the renewal row I fired `woocommerce_order_status_completed_notification` directly. Did you drive it the same way?

Thanks for mapping the token storage model; the table made the fix direction clear.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
wil-gerken: The welcome email is now named, not implicit: the PR body calls it out in a Behavior changes section, along with the changelog entry you asked for. Kept the guard change itself, and adopted the source fix you suggested (5feee82e5): `get_order_donation_product_id()` returns `false` per its docblock, the call-site note is gone, and the guard is a plain falsy check.

Also ran the check in the same isolated env with the donation products setting cleared: a completed order carrying the registration meta produced only WooCommerce's default completed-order email. No Newspack receipt, no welcome send, no crash.

The same source fix makes the existing `false ===` comparison in `order_paid()` match. The helper had been returning null rather than the documented `false` there too, so that comparison never matched, and `newspack_donation_order_processed` fired with a null product ID for every paid order on every site with no donation products configured. It now fires only for donation orders. The hook's only in-repo listener already discarded the null-product events, so nothing in this repo changes; a listener outside it would stop receiving those fires.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
wil-gerken: Adopted (5feee82e5): both the token label and the title fallback pass through `wp_strip_all_tags()`, with a test asserting a gateway title containing markup comes back tag-free.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
wil-gerken: Removed (5feee82e5).

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
wil-gerken: Removed (5feee82e5).

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-connection-payment-method.php
wil-gerken: All three adopted (5feee82e5):

- Return semantics now match the docblock: `true` on a sent email. A success-path test runs the full send and asserts the return value, the sent marker, and the delivered body containing the card digits.
- The label-map mock now matches the real map plus the ucwords fallback. Stripe's `amex` produces "Amex ending in 0005", pinned by a test. I kept "Amex" rather than normalizing to "American Express": it's the brand name a reader recognizes, and normalizing would mean maintaining a lookup table of gateway spellings.
- A renewal-order test covers the renewal row from your table.

The corrected sent-marker assertion also exposed that the marker was never persisted. Added `$order->save()` (ceab0f101), so the resend guard now holds across requests.

adekbadek (approved):
The card details now come from where the Stripe gateway actually writes them: the order's `_stripe_source_id` matched against the customer's saved tokens, scoped to that customer and gateway, with guest orders and stale references falling through to the gateway title. The `save()` on the sent marker and the `int|false` contract on `get_order_donation_product_id()` are both correct — the latter also stops `order_paid()` firing donation events on sites with no donation products.

PR #591, opened 2026-07-09
Improved handling/logging for Integrations sync errors

claudiulodro:
### All Submissions:

Closes NPPD-1818

This PR improves the error handling by differentiating between different types of errors. We don't want alerts about every single kind of error, and we don't want to retry on errors that we know will always fail. We can add more errors to the appropriate classification as we get alerts about them. The ones we've received over the last 48 hours are already in there.

1. Verify syncing continues working correctly. It's difficult and finicky to force the failures (MC doesn't always categorize emails as obviously fake for example). The unit tests should prove the failure case.

### Other information:


plugins/newspack-plugin/tests/unit-tests/reader-activation-sync.php
```diff
@@ -618,6 +618,88 @@ function ( $data ) use ( &$hook_fired, &$hook_data ) {
...
+			'newspack_sync_permanent_failure',
+			function ( $d ) use ( &$fired, &$data ) {
+				$fired = true;
+				$data  = $d;
+			}
+		);
+
+		Failing_Sample_Integration::reset();
+		Failing_Sample_Integration::$should_fail  = true;
+		Failing_Sample_Integration::$fail_message = 'saraeschwartz@icloud.co looks fake or invalid, please enter a real email address.';
```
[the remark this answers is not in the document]
claudiulodro: https://github.com/Automattic/newspack-workspace/pull/591/commits/fc333bbce5814a71a9666a6b52c65c321a43fedd Should be resolved from the git history once this is squashed and branch is deleted

plugins/newspack-plugin/includes/class-alert-manager.php
```diff
@@ -278,6 +279,39 @@ public static function handle_sync_retry_exhausted( $payload ) {
...
+			'newspack_alert',
+			[
+				'type'      => 'sync_permanent_failure',
+				'severity'  => 'config' === $class ? 'error' : 'warning',
+				'message'   => $message,
+				'context'   => $payload,
+				'timestamp' => time(),
+			]
+		);
+	}
```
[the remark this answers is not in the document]
claudiulodro: cleaner fix at https://github.com/Automattic/newspack-workspace/pull/591/changes/0c00a6ec4eac910b352848cddf608b362fd94744

plugins/newspack-plugin/includes/class-alert-manager.php
```diff
@@ -278,6 +285,36 @@ public static function handle_sync_retry_exhausted( $payload ) {
...
+	 */
+	public static function handle_sync_permanent_failure( $payload ) {
+		$message = sprintf(
+			'Permanent config sync failure for integration "%s" (no retry). Last error: %s',
+			$payload['integration_id'] ?? 'unknown',
+			$payload['reason'] ?? 'unknown'
+		);
+
+		/** This action is documented in includes/class-alert-manager.php */
+		do_action(
```
chickenn00dle, line 362: **Severity:** suggestion

During an account-wide config failure (disabled or unpaid ESP account), every contact sync classifies as `permanent_config` and fires this `error`-severity alert immediately, with no dedup, so a busy site could burst a large number of Slack alerts for what is really a single site-level problem. The existing retry-exhausted path is naturally rate-limited by the 5-retry backoff; this path fires on the first failure per contact. Since a config failure is site-wide rather than per-contact, consider deduping per `integration_id` over a short window (a transient, like the existing `HEALTH_CHECK_DEDUP_INTERVAL` / pattern-scan dedup already in this class) so one alert conveys the signal without the flood.

chickenn00dle (commented):
Release risk: LOW. Clean, well-tested change. The error-classification approach is sound and, importantly, fail-safe: anything the signature lists don't match falls back to `transient` and still retries, so drift in ESP error copy can never silently drop a contact. Tests cover each class and both the sync and deletion paths. Nice touch keeping the contact email out of the alert message and routing it through the structured `user_email` param, with a test asserting it doesn't leak. One consideration before merge around alert volume during an account-wide ESP outage, inline.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -482,6 +535,67 @@ private static function schedule_integration_retry( $integration_id, $user_id, $
 		$error_message = $error instanceof \WP_Error ? $error->get_error_message() : (string) $error;
 		$user_email    = $user ? $user->user_email : 'unknown';
 
+		$error_class = self::classify_error( $error );
+		if ( 'benign' === $error_class ) {
```
miguelpeixe, line 676: **Suggestion:** Never-fixable failures still reach the Slack pattern-alert channel — The retry short-circuit stops `sync_retry_exhausted` alerts, but every `WP_Error` — including `benign` and `permanent_contact` results — still fires `newspack_sync_contact_failed` (line 266, before classification runs), which `Alert_Manager::record_failure()` writes to the failure log the hourly pattern scanner reads. The `same_integration` / `same_message` rules (threshold 5/hour) then emit an error-severity `failure_pattern` alert that `forward_alert_to_log()` routes to Slack — deduped per rule+group to one per hour, but recurring for as long as the condition persists. Per the NPPD-1818 census (63 permanently-deleted + 53 invalid-merge-field + 59 fake/invalid emails in 48h), a busy site clears 5/hour for a single ESP from never-fixable failures alone, so the exact classes this PR de-noises keep paging through this side channel. Consider carrying the error class in the `newspack_sync_contact_failed` payload so `record_failure()` or the pattern rules can skip or segregate non-transient classes.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -482,6 +535,67 @@ private static function schedule_integration_retry( $integration_id, $user_id, $
 		$error_message = $error instanceof \WP_Error ? $error->get_error_message() : (string) $error;
 		$user_email    = $user ? $user->user_email : 'unknown';
 
+		$error_class = self::classify_error( $error );
```
miguelpeixe: **Suggestion:** User-less syncs never classify — config alerts can be missed — `schedule_integration_retry()` returns at the `! $user` bail (lines 529-533) before `classify_error()` runs, so syncs without a resolvable WP user (guest checkouts, users deleted mid-flight) exit before classification. A site-level `permanent_config` failure ("API access has been disabled", "Payment Required") on those syncs never fires `newspack_sync_permanent_failure` — the actionable alert only goes out once a user-backed sync happens to fail. Since `classify_error()` needs only `$error`, hoisting the classification (and the config-alert dispatch) above the user bail would make the alert independent of user resolution; `Alert_Manager::get_alert_user_email()` already tolerates a missing email. (The deletion path has no such gate, so the two paths currently behave differently.) Related: past the bail, the `$user ? $user->user_email : 'unknown'` ternary on line 536 is dead code — `$user` is always truthy there.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -693,7 +807,58 @@ private static function schedule_deletion_retry( $integration_id, $mode, $email,
 		}
 
 		$error_message = $error instanceof \WP_Error ? $error->get_error_message() : (string) $error;
-		$next_retry    = $retry_count + 1;
+
+		$error_class = self::classify_error( $error );
+		if ( 'benign' === $error_class ) {
```
miguelpeixe, line 974: **Suggestion:** Push-oriented error classes invert on the deletion path — The signature lists are written for the push/upsert direction, but `schedule_deletion_retry()` applies them verbatim to the removal direction, where two classes invert their meaning: a `benign` "Member Exists" response to a `flag`-mode push means the ESP contact still exists *without* the `account_deleted`/`membership_status` flags — skipping the retry permanently drops the deletion signal for a flow whose own docblock frames retries as protection against stranding contacts in an undeleted state (and the WP user is already gone, so nothing will ever re-trigger it). Conversely, "was permanently deleted" is the *success* end-state for `delete` mode, yet it classifies as `permanent_contact` and logs as "Permanent permanent_contact failure". The log copy "ESP reports contact already synced" (line 815) is also inaccurate in the deletion context. Consider mode-aware classification (a deletion-specific signature map), or at minimum not short-circuiting `benign` on this path and rewording the log copy.

**Risk if not addressed:** If a flag-mode deletion (the GDPR marker push) fails for a transient reason whose text matches a benign or invalid-contact phrase, its retry is skipped with no alert and no possible re-trigger, so the ESP-side deletion/flag signal can stay unpropagated indefinitely; because the pre-existing behavior would have eventually fired a retry-exhaustion alert, this class of deletion failure goes from "alerts after 5 retries" to surfacing only if the hourly failure-log pattern threshold is crossed.

**Consider:** The deletion path has no natural re-trigger — the WordPress user is already gone — so silently skipping its retry for a benign or invalid-contact classification can leave the ESP-side deletion or flag signal unpropagated for good, whereas the contact path re-syncs on the reader's next event. Keeping a surfaced notice (or the existing retry-exhaustion alert) for skipped deletion retries would keep these GDPR-relevant drops from going unobserved unless the hourly pattern threshold happens to be crossed.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -458,6 +490,27 @@ public static function handle_account_deletion( $email, $contact, $context = ''
...
 	}
 
+	/**
+	 * Classify an ESP sync error to decide retry behavior.
+	 *
+	 * @param string|\WP_Error $error The error from a failed push.
+	 * @return string One of 'permanent_contact', 'permanent_config', 'benign', or 'transient'.
+	 */
+	private static function classify_error( $error ) {
+		$message  = $error instanceof \WP_Error ? $error->get_error_message() : (string) $error;
```
miguelpeixe: **Nit:** classify_error() sees only the first message of aggregate WP_Errors — `classify_error()` reads `$error->get_error_message()` — the first message only — but the ESP layer it classifies returns aggregates: `Newspack_Newsletters_Contacts::upsert()` builds one `WP_Error` and `->add()`s invalid-list errors (contacts.php:209) and exception messages (:218) *before* the provider's own error (:222). With a stale list ID configured, the invalid-list message comes first and a real signature in a later message is missed → classified `transient` → retried. Fail-safe (extra retries, no data loss), but it silently defeats classification exactly when a site is misconfigured — and it's inconsistent with `execute_integration_retry()` / `execute_deletion_retry()`, which already `implode()` `get_error_messages()` for their own logging. Matching against all messages closes the gap.

**Risk if not addressed:** If a genuinely retryable error returns a message that happens to contain one of the "member exists" / "already a list member" or invalid-contact phrases, that contact's retry chain is dropped with no alert; on the contact path it re-syncs on the reader's next event, but if the triggering condition is account-wide, every contact hitting it during that window is dropped until each reader's next natural sync.

**Consider:** The benign and permanent-contact classes skip the retry with no alert, so a false substring match here is the one direction that is not fail-safe: a genuinely retryable error whose text merely contains one of these phrases would be silently dropped. Anchoring these signatures more tightly (or requiring a more specific match before skipping without any alert) keeps the classifier from turning a transient error into a silent drop.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -482,6 +535,67 @@ private static function schedule_integration_retry( $integration_id, $user_id, $
...
+			static::log(
+				sprintf(
+					'Skipping retry for integration "%s" sync of user %d (%s); ESP reports contact already synced. Detail: %s',
+					$integration_id,
+					$user_id,
+					$user_email,
+					$error_message
+				)
+			);
+			return;
```
miguelpeixe: **Nit:** Benign skip: invisible in ActionScheduler, and still marked failed at max retries — Two observability asymmetries with the permanent branch: (1) the permanent path writes an `ActionScheduler_Logger` note ("Permanent failure (%s); not retrying.", lines 562-567) but the benign skip writes none, so an AS action whose chain intentionally ended shows only "Retry N failed …"; (2) when a benign result lands exactly at `retry_count == MAX_RETRIES`, `execute_integration_retry()` (lines 771-772) and `execute_deletion_retry()` (1019-1020) still throw after the short-circuit returns, marking the AS action *failed* for what this PR itself defines as an effectively-synced outcome. A benign AS-log note, plus letting the executors know the chain was deliberately ended before the throw, would make both surfaces accurate.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -693,7 +807,58 @@ private static function schedule_deletion_retry( $integration_id, $mode, $email,
...
+			);
+			if ( self::$current_as_action_id ) {
+				\ActionScheduler_Logger::instance()->log(
+					self::$current_as_action_id,
+					sprintf( 'Permanent failure (%s); not retrying.', $error_class )
+				);
+			}
+			if ( 'permanent_config' === $error_class ) {
+				/**
+				 * This action is documented in schedule_integration_retry(). The
```
miguelpeixe: **Nit:** Deletion hook doc: method-name reference and undocumented payload divergence — The duplicate-hook reference reads "This action is documented in schedule_integration_retry()" — every other cross-reference in this file (lines 371, 721) and in `Alert_Manager` uses the `includes/…` file-path form that hook-doc parsers key on. More substantively, the deletion payload diverges from the canonical docblock (drops `user_id`, adds `mode`) and that shape is only described in prose; the in-file precedent for exactly this situation — the deletion variant of `newspack_sync_retry_exhausted` (lines 879-895) — re-documents its divergent shape with full `@type` entries. A consumer reading the canonical docblock at line 575 would code against `user_id` and miss `mode`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
```diff
@@ -83,6 +83,38 @@ class Contact_Sync extends Sync {
...
+	 * layer (only a "{Title}: {detail}" string survives), so classification is
+	 * necessarily string-based. Extend the lists as ESP error copy evolves.
+	 *
+	 *   - permanent_contact: bad contact data; retrying can never succeed.
+	 *   - permanent_config:  site-level ESP account problem; actionable.
+	 *   - benign:            the contact already exists; no retry needed.
+	 *
+	 * Anything not matched here is treated as 'transient' and retried.
+	 */
+	const ERROR_SIGNATURES = [
```
miguelpeixe: **Nit:** ERROR_SIGNATURES could be a private const — `ERROR_SIGNATURES` is public but consumed only by the private `classify_error()` (tests reach it via reflection anyway). The docblock anticipates the lists evolving "as ESP error copy evolves"; keeping them off the public class API preserves that freedom, matching the in-file precedent of `Alert_Manager::HEALTH_CHECK_DEDUP_INTERVAL` (private, with a "keep the surface minimal" rationale) — unlike `RETRY_HOOK` / `MAX_RETRIES`, which have external consumers.

miguelpeixe (approved):
No blockers — the core design is sound and deliberately fail-safe: unmatched error copy falls back to `transient` and keeps retrying, so signature drift can only ever cost extra retries, never a silently dropped contact. Three things are worth a close look before merge: (1) never-fixable failures still enter the failure log and keep tripping hourly `same_integration`/`same_message` Slack pattern alerts, which works against this PR's stated goal; (2) the deletion path applies push-oriented semantics to removal flows, where "Member Exists" actually means the deletion flags never landed — that one has GDPR texture; (3) classification sits below the user-existence bail, so guest/user-less syncs on a broken ESP account never fire the new config alert. All three are contained, follow-up-sized changes; the rest is nits.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: `newspack_sync_contact_failed` now carries an `error_class` field (classified at the failure site, deletion-aware on the deletion branches), and `Alert_Manager::record_failure()` keeps non-transient classes out of the failure log — so benign/permanent failures can no longer trip the hourly `same_integration`/`same_message` pattern rules. Data-event handler payloads (which carry no classification) still record. Covered by `test_record_failure_skips_non_transient_classes`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: classification (including the `permanent_config` alert dispatch) now runs before the user-existence bail, so user-less syncs classify and alert. The alert payload `email` is empty when no WP user resolves (`Alert_Manager::get_alert_user_email()` tolerates it), and the previously-dead `$user ? … : 'unknown'` ternary is now live above the bail. Covered by `test_permanent_config_alert_fires_without_user`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6 with a deletion-specific signature map (`DELETION_ERROR_SIGNATURES`): the "member exists" phrases are deliberately absent from it, so on the deletion path they classify as transient and retry (the flags have not landed); "was permanently deleted" is benign there (the deletion end-state), and the skip log copy now reads "the contact is already gone". Permanent contact-data errors on the deletion path now fire `newspack_sync_permanent_failure` with their `error_class`, which `Alert_Manager` surfaces at warning severity (Watch) — the skipped-deletion drop stays observable without paging. Tests: `test_classify_error_deletion_direction`, `test_deletion_member_exists_is_retried`, `test_deletion_already_deleted_is_benign`, `test_deletion_permanent_contact_error_fires_hook`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: `classify_error()` now matches against all messages (`implode( ' ', $error->get_error_messages() )`), consistent with how the retry executors already log. Covered by `test_classify_error_reads_all_messages`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: the benign skip now writes an AS-log note ("Benign result …; retry chain deliberately ended.") on both the sync and deletion paths, and the schedule methods return the classification so the executors skip the final-retry throw for benign results — the AS action completes instead of being marked failed. Covered by `test_benign_error_at_max_retries_does_not_throw`.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: the deletion-path hook is now fully re-documented with its own `@type` entries (the `email` + `mode` payload shape, plus the new `error_class`), following the in-file precedent of the deletion `newspack_sync_retry_exhausted` docblock.

plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
miguelpeixe: Addressed in 6e16934e6: `ERROR_SIGNATURES` (and the new `DELETION_ERROR_SIGNATURES`) are now `private const`.

plugins/newspack-plugin/includes/class-alert-manager.php
miguelpeixe: Addressed in 6e16934e6: config-class permanent-failure alerts are now deduped per `integration_id` via a transient for `PERMANENT_FAILURE_DEDUP_INTERVAL` (1 hour, matching the pattern-scan cadence), with the transient set before dispatch per the health-check precedent so a throwing handler cannot defeat it. Contact-class alerts (deletion path, added in this same commit) are warning-severity and not deduped, since each concerns a distinct contact. Covered by `test_permanent_config_failure_alert_deduped_per_integration`.

miguelpeixe:
Took this over to get it merge-ready: merged `main` into the branch — which brings in the RTL-suite stabilization from #699 that the failing `JS / newspack` check was hitting — and addressed the outstanding review feedback directly in 6e16934e6.

Point by point on the review:

1. **Never-fixable failures still reaching the hourly pattern alerts** — `newspack_sync_contact_failed` now carries an `error_class`, and `Alert_Manager::record_failure()` keeps non-transient classes out of the failure log, so the `same_integration`/`same_message` rules only ever see transient failures. Unclassified payloads (data-event handler failures) still record.
2. **Push-oriented semantics on the deletion path** — deletion now classifies against its own signature map: "member exists" retries (the deletion flags did not land), "was permanently deleted" is the benign end-state, and the skip log copy is reworded. Permanent contact-data errors on the deletion path — which have no natural re-trigger — now fire `newspack_sync_permanent_failure` and surface at warning severity (Watch), so those GDPR-relevant drops stay observable without paging.
3. **Classification below the user-existence bail** — classification and the config-alert dispatch now run before the bail, so guest/user-less syncs on a broken ESP account fire the config alert too.

Also folded in @chickenn00dle's suggestion: config-class permanent-failure alerts now dedup per integration (1-hour transient, set before dispatch), so an account-wide ESP outage pages once instead of once per contact. The inline nits landed in the same commit: all-messages matching in `classify_error()`, a benign AS-log note plus no failed-marking when a benign result lands on the final retry, a fully re-documented deletion-path hook, and `private const` signature maps.

All 8 inline threads are replied to and resolved. The full newspack-plugin PHPUnit suite passes locally (2308 tests).

PR #593, opened 2026-07-09
fix(rtt): filter bots and dedup repeat views in pixel counter

wil-gerken:
Republication Tracker can now measure what its view counts would look like with non-reader traffic filtered out. No new number is put in front of publishers. Today the "views" number counts every request to the tracking pixel. Link-preview bots (Slack, Facebook), crawlers, prefetches, and the same reader reloading all count as views. A publisher report has the tracker reading two to three times higher than the bot-filtered analytics publishers compare it against.

Reference: NPPM-2603

### What this does

Adds counting guards to the pixel request handler, **behind a default-off flag (`WPRTT_COUNTING_GUARDS_ENABLED`), in shadow mode**. The plugin records a second, guarded count alongside the existing one, so the two can be compared on real traffic before any displayed count changes. The guarded count is never displayed.

- **Raw counter (the displayed "views" number):** counts the same way it always has, with one change under the flag. Pixel responses stop being cacheable, described in the third bullet, so requests that page and edge caches used to absorb now reach the counter. A busy story can read higher. The rollout below puts a number on that effect.
- **Guarded count** (`republication_tracker_tool_sharing_guarded` post meta, written only with the flag on): excludes known crawlers, link-preview agents, and scripts. Those requests still receive the image. A reader who reloads the same story on the same republisher site inside 30 minutes counts once. The same reader opening the story on a second republisher counts there too, because attribution is per republisher. The 30-minute window matches the session windows GA4 and Parse.ly use, and those are the numbers publishers compare against.
- **Baseline snapshot** (first guarded view per post): records the raw counter as it stood when guarded counting started, plus a start timestamp (`republication_tracker_tool_sharing_guarded_baseline` post meta). The raw counter carries a story's full history while the guarded count starts at zero, so without this the pilot would compare different time windows. The like-for-like comparison is guarded vs. raw minus baseline. The README documents the WP-CLI commands for reading both during the pilot.
- **Uncacheable responses** (flag on): pixel responses carry `Cache-Control: no-store, no-cache, must-revalidate, max-age=0` plus a batcache cancel, so every request reaches the counter instead of being absorbed or replayed by page and edge caches.

Rollout: a pilot enables the flag on the reporting publication first, then compares raw, guarded, and Parse.ly counts on real traffic. Whether the displayed count switches to the guarded count stays open. It waits on the pilot data, carries its own publisher communication, and is tracked in NPPM-3106 along with the pre-fleet cache-relocation change. Because the guards run server-side, they cover every embed already copied to republisher sites.

**Also fixed, in both flag states:** a request whose post ID is anything but a strictly positive integer no longer records anything. That covers a missing ID, the settings snippet's literal `YOUR-POST-ID` placeholder, an array, a negative, and trailing characters like `47abc`. PHP's coercing casts map several of those to real post IDs (`get_post( 0 )` falls back to the global post; an array casts to 1; `47abc` casts to 47), so the resolver validates with `FILTER_VALIDATE_INT` before any lookup.

### How to test

Every request below needs a new `cb=` value. Page caching absorbs repeats otherwise, and the counts will not match.

Reproduce the baseline counting behavior (flag off; matches `release` apart from the two divergences listed under Notes for reviewers):

1. Publish a post and note its ID. Request the pixel as a reader: `curl -A "Mozilla/5.0 (Macintosh) Chrome/126" -e "https://republisher.example/story" "https://<site>/?republication-pixel=true&post=<ID>&ga4=test&cb=1"`.
2. Send the same request again with a new `cb=`. Send it a third time with a bot user agent (`-A "Slackbot-LinkExpanding 1.0"`).
3. Read the counter: `wp eval 'print_r( get_post_meta( <ID>, "republication_tracker_tool_sharing", true ) );'`. It reads 3. Every request counted.

Verify this branch (guards enabled unless a step says otherwise):

1. Enable the guards: `wp config set WPRTT_COUNTING_GUARDS_ENABLED true --raw`.
2. Run the same sequence against a fresh post: reader, same reader again, bot.
3. Read the raw counter. It reads 3, the same as the baseline.
4. Read the guarded count (`republication_tracker_tool_sharing_guarded` post meta). It reads 1. The guards excluded the repeat and the bot. The baseline meta (`republication_tracker_tool_sharing_guarded_baseline`) holds the raw counter as it stood before the first guarded view, plus a start timestamp.
5. Check the response headers. They carry `Cache-Control: no-store, no-cache, must-revalidate, max-age=0`.
6. Turn the flag off and request the pixel again. The response is cacheable, as before.
7. Request with `post=YOUR-POST-ID`, in either flag state. The image is served. No post gains counter meta, including the site's newest.
8. Run the unit tests: `n test-php --group pixel_counting` in `plugins/republication-tracker-tool`.

Every step above was run as written against an isolated environment on 75589ed0e: baseline read 3, shadow read raw 3 / guarded 1 with the baseline snapshot present, headers matched in both flag states, `YOUR-POST-ID` served the image and recorded nothing in either state, and the test group ran green. The admin posts list keeps reading the raw counter while the guards are on:

Admin posts list: displayed Total Views reads the raw counter (5 and 3) while the guards are on

### Notes for reviewers

- `pixel.php` is now a thin request handler. Every counting decision lives in named, tested functions in `pixel-functions.php` (`wprtt_resolve_shared_post()`, `wprtt_record_view()`, plus the guard helpers), so tests cover the flag-off behavior directly.
- The counting path reads cookies through the read-only `wprtt_read_cid_from_cookies()` only. `newspack-cid`, the shared Newspack reader-identity cookie, is minted only inside the fully-configured GA4 branch, which is where it sits on `release` too. A test pins this: under PHPUnit any `setcookie()` attempt fails the test.
- Dedup identity: cookie client ID when present, else `wp_hash( IP | UA | Accept-Language )`, an HMAC keyed on the auth salt. Rotating salts resets the window and fails open. Accept-Language adds entropy where Chrome's UA reduction makes IP + UA coarse behind shared egress IPs. The dedup key also includes the referrer, so per-republisher attribution survives dedup.
- Dedup storage: `wp_cache_add()` when a persistent object cache exists. That call is atomic, so it closes the simultaneous-hit race. When the add fails, a follow-up read distinguishes "already counted" from "cache backend down," and a broken backend counts the view instead of silently zeroing the guarded count. Installs without a persistent cache fall back to transients, a documented trade-off for WordPress.org: two `autoload=no` options rows per counted view, cleared by the daily cleanup.
- The bot pattern matches Pinterest's crawler token (`pinterest/0.`), not the bare brand name, so in-app browsers for Pinterest, Facebook, and Instagram all count as readers (test-covered). The pattern is filterable (`wprtt_bot_request_pattern`), so a false positive found in production can be fixed at the site level without waiting for a plugin release.
- Two deliberate flag-off divergences from `release`, both narrow. (1) The post ID validation described above: an invalid ID also no longer fires a stray GA4 event built from the global post. (2) The referrer title fetch runs only for the GA4 event, through `wp_safe_remote_get()` after a `wp_http_validate_url()` gate, so redirect hops are validated too. A referrer that fails validation yields an empty GA4 `page_title` instead of a server-side fetch of an unvalidated URL. Both outbound GA4-branch requests (title fetch, Measurement Protocol event) now carry a 2-second timeout, which bounds how long a public pixel request can hold a worker.
- Test infrastructure: the PHPUnit bootstrap loads the composer autoloader unconditionally.

### Release risk

Low at the default setting: the flag is off, and the flag-off path matches `release` apart from the two disclosed divergences above. That match was verified live (same request sequence, same counts, same cacheable response) and is pinned by tests that exercise the counting decisions directly.

With the flag on, no new number reaches publishers, because the guarded count is never displayed. The one effect on a displayed number is the cache change: requests the caches used to absorb now reach the raw counter, so a busy story can read higher during the pilot.

Remaining code-level risks affect the guarded count. A bot-pattern false positive silently lowers it; the in-app browser cases are test-covered, and the filter allows a site-level fix. Memcached eviction or salt rotation weakens dedup toward legacy counting, and dedup fails open in both cases.

### All Submissions:

### Other information:


plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -109,58 +91,69 @@ function wprtt_extract_cid_from_cookies() {
...
+	// within the window counts once, for both the counter and the GA4 event.
+	// The dedup identity falls back to IP + user agent, since cross-site pixel
+	// requests usually arrive without cookies; the GA4 client ID below keeps
+	// the original cookie-or-generated behavior.
+	if ( ! $wprtt_guards_enabled || wprtt_should_count_view( $shared_post_id, wprtt_get_dedup_identity() ) ) {
+		$wprtt_client_id = wprtt_extract_cid_from_cookies();
+		// The title fetch is a blocking outbound request only the GA4 event needs,
+		// so it runs only for views that actually count.
+		$url_title = '' !== $url ? wprtt_get_referring_page_title( $url, $shared_post_id ) : '';
+		$value = get_post_meta( $shared_post_id, 'republication_tracker_tool_sharing', true );
```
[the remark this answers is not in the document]
wil-gerken: Removed in 353307ede. It was pre-existing and harmless — PHP ignores extra arguments to userland functions; the warning/fatal behavior applies to internal functions — but the line moved in this PR, so it was worth cleaning up.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,140 @@
...
+			list( $version, $domain_depth, $cid ) = $cookie_pieces;
+		}
+		return $cid;
+	}
+
+	if ( isset( $_COOKIE['newspack-cid'] ) ) {
+		return sanitize_text_field( wp_unslash( $_COOKIE['newspack-cid'] ) ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+	}
+	return wprtt_create_cid_cookie_if_not_set();
+}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 353307ede — the parser now takes the last cookie piece, so a malformed value yields a stable client ID instead of null, with a regression test.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,137 @@
...
+		// A well-formed cookie (GA1.2.<cid>) yields the third piece; malformed
+		// values with fewer pieces still yield their last piece, never null.
+		return end( $cookie_pieces );
+	}
+
+	if ( isset( $_COOKIE['newspack-cid'] ) ) {
+		return sanitize_text_field( wp_unslash( $_COOKIE['newspack-cid'] ) ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+	}
+	return wprtt_create_cid_cookie_if_not_set();
+}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in b07480e76 — an empty cookie value now falls through to the `newspack-cid` cookie or a generated ID, with a regression test.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,46 +38,31 @@ function wprtt_get_referring_page_title( $url ) {
...
+	if ( ! defined( 'DONOTCACHEPAGE' ) ) {
+		define( 'DONOTCACHEPAGE', true );
 	}
-	return wprtt_create_cid_cookie_if_not_set();
+	header( 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0' );
 }
 
+// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__ -- Only used when the guards are on, which makes the pixel response uncacheable (no-store + batcache_cancel above).
+$wprtt_user_agent = isset( $_SERVER['HTTP_USER_AGENT'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) ) : '';
+
```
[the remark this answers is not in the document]
wil-gerken: Fixed in b07480e76 — the user agent is now read only when the guards are enabled, so the flag-off path touches no cache-sensitive request data and the justification is accurate.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,50 +38,42 @@ function wprtt_get_referring_page_title( $url ) {
...
 
+$wprtt_user_agent = '';
+if ( $wprtt_guards_enabled ) {
+	// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__, WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders -- Only read when the guards are on, which makes the pixel response uncacheable (no-store + batcache_cancel above).
+	$wprtt_user_agent = isset( $_SERVER['HTTP_USER_AGENT'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) ) : '';
+}
+
+// The pixel endpoint is public: unknown or deleted post IDs still get the
+// image below, but there is nothing to count.
+$wprtt_shared_post = isset( $_GET['post'] ) ? get_post( absint( $_GET['post'] ) ) : null; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 83949480b — the lookup now requires both `post` and `ga4` params, restoring the original precondition.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -109,58 +98,69 @@ function wprtt_extract_cid_from_cookies() {
...
-			[
-				'api_secret'     => $ga4_secret,
-				'measurement_id' => $ga4_id,
-			],
-			'https://www.google-analytics.com/mp/collect'
-		);
-		$payload  = [
-			'client_id' => wprtt_extract_cid_from_cookies(),
-			'events'    => [
+		$update = update_post_meta( $shared_post_id, 'republication_tracker_tool_sharing', $value );
```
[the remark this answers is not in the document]
wil-gerken: Dropped in 83949480b.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,141 @@
...
+	setcookie( 'newspack-cid', $cid, time() + 30 * DAY_IN_SECONDS, COOKIEPATH, COOKIE_DOMAIN, true );
+
+	return $cid;
+}
+
+/**
+ * Extracts the Client ID from the _ga cookie
+ *
+ * @return ?string
+ */
```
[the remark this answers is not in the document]
wil-gerken: Updated in 83949480b — it now documents the fallback chain and the always-string return.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Generate a random client ID string and set the newspack-cid fallback cookie if not set.
+ *
+ * @return string Randomly generated client ID.
+ */
+function wprtt_create_cid_cookie_if_not_set() {
+	$cid = (string) wp_rand( 100000000, 999999999 );
+
+	// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.cookies_setcookie
+	setcookie( 'newspack-cid', $cid, time() + 30 * DAY_IN_SECONDS, COOKIEPATH, COOKIE_DOMAIN, true );
```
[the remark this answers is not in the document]
wil-gerken: Leaving the cookie attributes unchanged in this PR: the `setcookie()` call moved files verbatim, and changing Set-Cookie behavior for every site deserves its own change and testing. The dedup path deliberately doesn't depend on this cookie in the cross-site case. Happy to see it as a follow-up.

plugins/republication-tracker-tool/tests/test-pixel-functions.php
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	/**
+	 * The filter enables the counting guards.
+	 */
+	public function test_counting_guards_filter_enables() {
+		add_filter( 'wprtt_counting_guards_enabled', '__return_true' );
+		self::assertTrue( wprtt_counting_guards_enabled() );
+		remove_filter( 'wprtt_counting_guards_enabled', '__return_true' );
+	}
```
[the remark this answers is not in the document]
wil-gerken: The WP test framework restores hooks unconditionally: `tear_down()` calls `_restore_hooks()`, which rebuilds `$wp_filter` from clones taken at set-up (abstract-testcase.php), so a failing assertion can't leak this filter into other tests. The `remove_filter()` call is tidiness, not load-bearing isolation.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -91,16 +83,13 @@ function wprtt_extract_cid_from_cookies() {
 			$url = esc_url_raw( $_SERVER['HTTP_REFERER'] );
 		}
 
-		$url_title = wprtt_get_referring_page_title( $url, $shared_post_id );
-
 		$url_host = wp_parse_url( $url, PHP_URL_HOST );
 		$url_path = wp_parse_url( $url, PHP_URL_PATH );
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 22604f9f3 — dropped the unused variables and collapsed the referrer read to a single `isset()`.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,50 +38,42 @@ function wprtt_get_referring_page_title( $url ) {
...
+
+// The pixel endpoint is public: unknown or deleted post IDs still get the
+// image below, but there is nothing to count.
+$wprtt_shared_post = isset( $_GET['post'], $_GET['ga4'] ) ? get_post( absint( $_GET['post'] ) ) : null; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
 // Non-ga4 hits (bots, crawlers) skip this block entirely. No counter update, no DB writes.
 // The wp-admin referrer bailout below is therefore only needed within this block.
 // Only update share tracking when a ga4 param is present (real pixel fires from configured republishers).
-if ( isset( $_GET['post'] ) && isset( $_GET['ga4'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+// Bot, crawler, and link-preview requests are served the image but never counted.
```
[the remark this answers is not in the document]
wil-gerken: Rewritten in 22604f9f3 — the comment now describes the current gate: tracking requires a valid post plus the `ga4` param, everything else just gets the image, and the guards additionally exclude bots when enabled.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,142 @@
...
+		if ( '' !== $cid ) {
+			return $cid;
+		}
+	}
+
+	if ( isset( $_COOKIE['newspack-cid'] ) ) {
+		return sanitize_text_field( wp_unslash( $_COOKIE['newspack-cid'] ) ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+	}
+	return wprtt_create_cid_cookie_if_not_set();
+}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in dfeb1f4a2 — cookie reading is now a side-effect-free helper that treats present-but-empty values as missing: the GA4 client ID falls through to a generated one, and the dedup identity falls back to the IP + user agent hash (better than a per-request random ID for that client). Regression test added.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,116 @@ function wprtt_get_referring_page_title( $url ) {
...
 	}
-
-	if ( isset( $_COOKIE['newspack-cid'] ) ) {
-		return sanitize_text_field( wp_unslash( $_COOKIE['newspack-cid'] ) ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
+	if ( ! defined( 'DONOTCACHEPAGE' ) ) {
+		define( 'DONOTCACHEPAGE', true );
 	}
-	return wprtt_create_cid_cookie_if_not_set();
+	header( 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0' );
+}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 6323bab84 — the cache header is now guarded with `headers_sent()`.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,116 @@ function wprtt_get_referring_page_title( $url ) {
...
+	// within the window counts once, for both the counter and the GA4 event.
+	// The dedup identity falls back to IP + user agent, since cross-site pixel
+	// requests usually arrive without cookies; the GA4 client ID below keeps
+	// the original cookie-or-generated behavior.
+	if ( ! $wprtt_guards_enabled || wprtt_should_count_view( $shared_post_id, wprtt_get_dedup_identity() ) ) {
+		$wprtt_client_id = wprtt_extract_cid_from_cookies();
+		// The title fetch is a blocking outbound request only the GA4 event needs,
+		// so it runs only for views that actually count.
+		$url_title = '' !== $url ? wprtt_get_referring_page_title( $url ) : '';
+		$value = get_post_meta( $shared_post_id, 'republication_tracker_tool_sharing', true );
```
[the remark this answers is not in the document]
wil-gerken: Fixed in 6323bab84 — the referrer is validated with `wp_http_validate_url()` before any server-side fetch, so non-http(s) and local/private targets are refused and the title falls back to empty. Good catch: the fetch itself is long-standing behavior, but this closes the request-forgery exposure on hosts without a safe fetcher.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
-		$url_path = wp_parse_url( $url, PHP_URL_PATH );
-
-	} else {
-
-		$url       = '';
-		$url_title = '';
-		$url_host  = '';
-
-	}
+	$url = isset( $_SERVER['HTTP_REFERER'] ) ? esc_url_raw( $_SERVER['HTTP_REFERER'] ) : '';
```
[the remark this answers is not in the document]
wil-gerken: Fixed in `6f8913702` — the referrer is now `wp_unslash()`'d before `esc_url_raw()`, matching the `REMOTE_ADDR`/`HTTP_USER_AGENT` reads elsewhere in this file. No PHPCS ignore was needed; `esc_url_raw()` already satisfies the sanitization sniff and package lint stays clean.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,161 @@
...
+function wprtt_read_cid_from_cookies() {
+	if ( isset( $_COOKIE['_ga'] ) ) {
+		$cookie_pieces = explode( '.', $_COOKIE['_ga'], 3 ); // phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		// A well-formed cookie (GA1.2.<cid>) yields the third piece; malformed
+		// values with fewer pieces still yield their last piece.
+		$cid = trim( (string) end( $cookie_pieces ) );
+		if ( '' !== $cid ) {
+			return $cid;
+		}
+	}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in `6f8913702` — the `_ga` value is now `sanitize_text_field( wp_unslash() )`'d before parsing, consistent with the `newspack-cid` branch below it. The `InputNotSanitized` ignore is dropped as a result; only the VIP `_COOKIE` cache-constraint ignore remains.

plugins/republication-tracker-tool/tests/bootstrap.php
```diff
@@ -10,6 +10,11 @@
 	$_tests_dir = '/tmp/wordpress-tests-lib';
 }
 
+// Load the composer autoloader (provides the PHPUnit Polyfills required by the WP test suite).
+if ( file_exists( dirname( __DIR__ ) . '/vendor/autoload.php' ) ) {
+	require_once dirname( __DIR__ ) . '/vendor/autoload.php';
+}
```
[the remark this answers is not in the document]
wil-gerken: Fixed in `6f8913702` — the autoloader is now required unconditionally, so a missing `vendor/` fails loudly rather than surfacing as a confusing "class not found" later. Suite still green (30/30).

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
-if ( isset( $_GET['post'] ) && isset( $_GET['ga4'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+$wprtt_user_agent = '';
+if ( $wprtt_guards_enabled ) {
+	// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__, WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders -- Only read when the guards are on, which makes the pixel response uncacheable (no-store + batcache_cancel above).
+	$wprtt_user_agent = isset( $_SERVER['HTTP_USER_AGENT'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) ) : '';
+}
+
+// The pixel endpoint is public: unknown or deleted post IDs still get the
+// image below, but there is nothing to count.
+$wprtt_shared_post = isset( $_GET['post'], $_GET['ga4'] ) ? get_post( absint( $_GET['post'] ) ) : null; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
adekbadek: **Blocker:** Missing or non-numeric `post` records a view against the site's newest story (flag-off path) — `absint()` maps anything non-numeric — `post=`, `post=abc`, `post=YOUR-POST-ID` — to `0`, and WordPress core's `get_post( 0 )` does not return `null`: it falls back to `$GLOBALS['post']`. The pixel is included on `template_include`, by which point the main query has run, so on `/?republication-pixel=true&post=&ga4=…` that global is the site's most recent published post. The `instanceof WP_Post` gate therefore passes, and line 78 takes `$shared_post_id` from that fallback object, so the counter is written against a story that was never republished. On `release` this could not happen: `$shared_post_id` came from `absint( $_GET['post'] )` = `0`, and `update_post_meta( 0, … )` is a no-op.

Reproduced with the guards **off** — the state every publisher ships in. One request with an empty `post=` wrote `{"https://republisher.example.com/reprinted-story/": 1}` onto post 6, the newest post; the identical request against a clean `origin/release` checkout wrote nothing. Three such requests render as **"Total Views: 3"** in wp-admin → Posts on a story that was never republished anywhere.

This is not only a crawler edge case. The plugin's own settings page builds its copy-paste embed snippet with the literal string `YOUR-POST-ID` (`includes/class-settings.php:124`) and tells the publisher to remember to replace it — and `absint( 'YOUR-POST-ID' )` is `0`. Any page carrying that snippet unmodified credits a phantom view, on every pageview, to whatever story the newsroom published most recently.

Resolve the ID before the lookup and refuse zero:
```php
$wprtt_post_id     = isset( $_GET['post'], $_GET['ga4'] ) ? absint( $_GET['post'] ) : 0;
$wprtt_shared_post = $wprtt_post_id ? get_post( $wprtt_post_id ) : null;
```

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
-	$value = get_post_meta( $shared_post_id, 'republication_tracker_tool_sharing', true );
-	if ( $value ) {
-		if ( isset( $value[ $url ] ) ) {
-			$value[ $url ]++;
+	// Deduplicate repeat views: the same client viewing the same shared post
+	// within the window counts once, for both the counter and the GA4 event.
+	// The dedup identity falls back to IP + user agent, since cross-site pixel
+	// requests usually arrive without cookies; the GA4 client ID below keeps
+	// the original cookie-or-generated behavior.
+	if ( ! $wprtt_guards_enabled || wprtt_should_count_view( $shared_post_id, wprtt_get_dedup_identity() ) ) {
```
adekbadek: **Blocker:** Dedup does not dedup: the identity flips key once the fallback cookie is set — The dedup identity and the cookie side effect fight each other. On a reader's first view, `wprtt_get_dedup_identity()` finds no cookie and returns the `ipua_…` IP+user-agent hash, so the transient is stored under that key — and then, further down, `wprtt_extract_cid_from_cookies()` *sets* a `newspack-cid` cookie. On the next view the browser sends that cookie back, so `wprtt_get_dedup_identity()` now returns the cookie's client ID instead: a **different** transient key, nothing found, and the repeat view is counted.

Verified with the guards on and a cookie jar (i.e. a browser that accepts and returns cookies, which is the whole point of the fallback cookie): one reader, one story, three views in a row counted **2**. Expected 1. View 1 counted under the IP+UA key, view 2 counted again under the newly-minted cookie key, view 3 was finally deduplicated. So every client gets one free double-count — precisely the repeat-view inflation this PR sets out to remove.

The fix is the same as the finding above: keep the counting path read-only with respect to cookies, so the dedup identity is stable across requests rather than changing the moment a cookie is issued.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
+		$wprtt_client_id = wprtt_extract_cid_from_cookies();
```
adekbadek: **Blocker:** Pixel now sets the shared `newspack-cid` cookie on sites that send no GA4 event (flag-off path) — `wprtt_extract_cid_from_cookies()` is not a read — when no usable cookie exists it calls `wprtt_create_cid_cookie_if_not_set()`, which issues a `setcookie()`. On `release` that call sat **inside** the GA4 branch (`if ( $ga4_id && $ga4_secret && $_GET['ga4'] === $ga4_id )`), so the cookie was only ever set on sites fully configured to send Measurement Protocol events. This PR hoists it to line 96, 26 lines above that gate, so it now runs for every counted view — including on sites that never finished GA4 setup, since the API secret is a separate field that has to be created by hand in the GA4 admin.

Reproduced with the guards **off**, a Measurement ID set and no secret: the response carries `Set-Cookie: newspack-cid=…; Max-Age=2592000`. The identical request against `origin/release` sets no cookie.

What makes this more than cosmetic is whose cookie it is. `newspack-cid` is not this plugin's — it is `NEWSPACK_CLIENT_ID_COOKIE_NAME` (`newspack-plugin/includes/class-newspack.php:82`), the shared Newspack reader-identity cookie that Reader Activation's session hydration reads and that gets attached to orders for donation attribution. A view-counting endpoint should not be minting the identifier the revenue path reads.

Use the pure read-only helper this PR already adds (`wprtt_read_cid_from_cookies()`) on the counting path, and move the cookie-setting variant back inside the GA4 block where it was. That one move also fixes the dedup defect below, which shares this root cause.

plugins/republication-tracker-tool/tests/test-pixel-functions.php
```diff
@@ -0,0 +1,213 @@
...
+ */
+
+// phpcs:disable WordPressVIPMinimum.Variables.RestrictedVariables, WordPressVIPMinimum.Variables.ServerVariables -- Tests simulate pixel requests by manipulating $_COOKIE / $_SERVER directly.
+
+/**
+ * Test bot filtering and per-client view deduplication for the tracking pixel.
+ *
+ * @group pixel_counting
+ */
+class PixelFunctionsTest extends WP_UnitTestCase {
```
adekbadek, line 15: **Blocker:** Nothing tests the pixel endpoint, so "flag-off is behavior-identical" is asserted but never checked — All 30 tests target the four pure helpers in `pixel-functions.php`. Nothing exercises `pixel.php`, where every flag check lives, where the counted/not-counted decision is actually made, and where all three regressions above landed. That is exactly why a change described as "the flag-off path behaves exactly as it always has" ships three flag-off behavior changes with a fully green build — and this is a hotfix onto `release`, where that claim is the entire safety argument.

The structural fix is small: `pixel.php` is procedural top-level code ending in `exit`, so it is unreachable from any test. Lift the request-level decision into a named function in `pixel-functions.php` — something like `wprtt_resolve_shared_post( array $get )` returning the post or `null`, and `wprtt_should_track_request( $post, $guards_enabled, $user_agent )` — and leave `pixel.php` as a thin caller. That turns the behavior-identity claim into an assertable test: same request, flag on versus flag off, same decision.

At minimum, before this merges, add tests pinning the two flag-off regressions: no view recorded for a missing or non-numeric `post`, and no cookie set when GA4 is not fully configured.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * @param string $user_agent The request's user agent string.
+ * @return bool True if the request should be treated as a bot.
+ */
+function wprtt_is_bot_request( $user_agent ) {
+	$user_agent = trim( (string) $user_agent );
+	if ( '' === $user_agent ) {
+		return true;
+	}
+	$bot_pattern = '/(?<!cu)bot|crawl|spider|slurp|preview|externalhit|feedfetcher|embedly|quora link|outbrain|pinterest|vkshare|validator|whatsapp|telegram|skypeuripreview|nuzzel|discordapp|qwantify|bitlybot|scanner|scrape|curl|wget|python|libwww|httpunit|nutch|go-http-client|java\/|okhttp|phantomjs|headlesschrome|lighthouse|pingdom|gtmetrix|uptimerobot|statuscake|newspaper|monitor/i';
```
adekbadek: **Suggestion:** The bot pattern filters the Pinterest in-app browser, silently dropping real readers — The `pinterest` token is aimed at Pinterest's link-preview crawler, but Pinterest's mobile apps ship a real in-app browser whose WebView user agent carries the same brand string — `… Mobile/15E148 [Pinterest/iOS]` and `… Mobile Safari/537.36 [Pinterest/Android]`. Both match, so a person who taps a Pin and reads the republished story is never counted, with no error and no log line — just a lower number. Pinterest is a meaningful referral source for exactly the local, food and lifestyle coverage that tends to get republished.

Confirmed in the env: same device, same IP, guards on — plain iOS Safari counts, the `[Pinterest/iOS]` UA does not.

Narrow the token to the crawler (`pinterestbot`, or `pinterest/0.` which is what the crawler actually sends) rather than the bare brand name. The rest of the pattern held up against a corpus of real human user agents — Facebook and Instagram in-app browsers, reduced-UA Chrome, iOS Safari and Cubot devices (the `(?<!cu)` lookbehind does work) all pass — so this is a one-token fix, not a rewrite. Worth adding these in-app UAs to the `browser_user_agents` provider so a brand-name token can't regress this again, and worth a `apply_filters()` escape hatch on the pattern so a false positive is a support-desk fix rather than a plugin release.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,162 @@
...
+		return $cid;
+	}
+	// phpcs:disable WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__REMOTE_ADDR__, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__, WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders -- The pixel response is explicitly uncacheable (no-store + batcache_cancel), and these values only feed a salted dedup hash: spoofing them merely weakens dedup for that client (fails open, same as no dedup).
+	$ip = isset( $_SERVER['REMOTE_ADDR'] ) ? sanitize_text_field( wp_unslash( $_SERVER['REMOTE_ADDR'] ) ) : '';
+	$ua = isset( $_SERVER['HTTP_USER_AGENT'] ) ? sanitize_text_field( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) ) : '';
+	// phpcs:enable
+	if ( '' === $ip && '' === $ua ) {
+		return '';
+	}
+	return 'ipua_' . md5( wp_salt( 'auth' ) . '|' . $ip . '|' . $ua );
```
adekbadek: **Suggestion:** The IP+user-agent dedup identity has little entropy, so shared networks collapse into one view — The docblock is right that browsers withhold `SameSite=Lax` cookies on third-party image loads — and since republisher pages are the only place this pixel lives, that means the cookie branch almost never fires and `md5( salt | REMOTE_ADDR | User-Agent )` is not a fallback, it is *the* dedup mechanism for nearly every real hit.

That identity is coarser than it looks. Chrome's User-Agent Reduction (fully rolled out since Chrome 110) freezes the Android UA to a literal `Android 10; K` with no device model, and iOS Safari's varies only by OS version — so the identity degrades to roughly "IP address + browser major version". Behind any shared egress IP (mobile-carrier CGNAT, an office or school NAT, public wifi) a large group of readers hashes to one identity and the whole group counts as **one** view per story per 30 minutes.

That is an undercount — an error in the opposite direction from the one being fixed, and an invisible one. Two things would de-risk it: mix in headers that still carry entropy and are sent on image subresource requests (`Accept-Language` is the strongest, plus any `Sec-CH-UA-*` present); and, more valuably, **shadow-count during the pilot** — keep writing the existing raw counter and write the guarded count to a second meta key. Without a control number, an over-filter is indistinguishable from the fix working, and once the guards are on the pre-guard number is gone for good.

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,162 @@
...
+ * Without a client ID there is nothing to key on, so the view counts —
+ * bot filtering is the guard for cookie-less clients.
+ *
+ * @param int    $post_id   The shared post ID.
+ * @param string $client_id The client ID extracted from cookies.
+ * @param int    $window    Dedup window in seconds. Defaults to 30 minutes,
+ *                          matching the session windows of GA4 and Parse.ly.
+ * @return bool True if the view should be counted.
+ */
+function wprtt_should_count_view( $post_id, $client_id, $window = 30 * MINUTE_IN_SECONDS ) {
```
adekbadek: **Suggestion:** Dedup writes two `wp_options` rows per view on hosts with no object cache, and the get/set is not atomic — `get_transient()`/`set_transient()` resolve to two very different systems. On Newspack (memcached present) each counted view is a cheap object-cache write — though memcached evicts under memory pressure, so a dedup entry can vanish early and the reader is counted again; that fails open, which is the right direction, but it means dedup effectiveness is a function of cache pressure rather than of the code. This plugin is also distributed on WordPress.org to hosts with no persistent object cache, where the same code writes **two `wp_options` rows per counted view** (value + timeout) on an endpoint that now bypasses page caching entirely, reaped only by WP's daily expired-transient cron. They're `autoload = no`, so `alloptions` is unaffected and this stays bounded — but a widely republished story could add rows faster than the reaper drains them.

Consider `wp_cache_add()`, which is also **atomic** — it closes the small race where two simultaneous pixel loads both pass the `get_transient()` check and both count — with a "no persistent object cache → skip dedup" fallback. If the DB path is meant to be supported, that's worth making a conscious, documented decision. Also note the transient is set *before* the counter is written, so if `update_post_meta()` fails the view is lost for 30 minutes rather than retried.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
-		} else {
-			list( $version, $domain_depth, $cid ) = $cookie_pieces;
-		}
-		return $cid;
+// Counting guards (bot filtering, dedup, uncacheable responses) are gated for
+// a gradual rollout — see wprtt_counting_guards_enabled(). When off, the pixel
+// behaves exactly as it always has.
+$wprtt_guards_enabled = wprtt_counting_guards_enabled();
+
+if ( $wprtt_guards_enabled ) {
```
adekbadek, line 17: **Suggestion:** Enabling the guards ends page-cache absorption, which both adds origin load and pushes counts back up — Newspack's batcache keys on the full query string and caches a URL once it's requested twice within 120s, serving it for 5 minutes. A cross-site `` load sends no origin-site cookies, so **today the pixel URL for a busy republished story is batcached and served without booting PHP for 5-minute stretches** — those hits never reach the counter at all. `batcache_cancel()` + `no-store` deliberately end that, and the placement is correct (batcache decides whether to store at shutdown, so cancelling from `template_include` works).

Two consequences worth settling before the flag is switched on anywhere. **Load:** the origin goes from serving cached image responses to a full WordPress bootstrap per reader per republished pageview — and because the handler runs on `template_include`, WordPress has already executed the main query before the pixel code runs. Moving the handler to a pre-query hook (`parse_request`, or an early `send_headers`) would skip that entirely, make the now-uncacheable endpoint far cheaper per hit, and as a bonus remove the `$GLOBALS['post']` trap behind the first blocker. **Direction:** hits batcache used to absorb will now reach the counter, pushing counts *up* and partially offsetting the drop from bot filtering and dedup. So the same commit both removes and adds hits, and the net effect on the 3x gap against Parse.ly genuinely is not predictable from the code — it needs measuring on a real republished story rather than assuming.

(Also worth a one-line comment that `DONOTCACHEPAGE` is there for non-batcache stacks — the batcache Newspack ships never reads it, so on a Newspack site the whole protection rests on `batcache_cancel()`.)

plugins/republication-tracker-tool/includes/pixel-functions.php
```diff
@@ -0,0 +1,162 @@
...
+ * responses) are enabled.
+ *
+ * Off by default for a gradual rollout: enabling changes what the "views"
+ * number means, and counts visibly drop. Enable per site (or fleet-wide via a
+ * managed define) with the WPRTT_COUNTING_GUARDS_ENABLED constant, or in code
+ * via the filter. When off, the pixel counts exactly as it always has.
+ *
+ * @return bool True if the counting guards are enabled.
+ */
+function wprtt_counting_guards_enabled() {
```
adekbadek: **Nit:** New functions carry no parameter or return types — These are brand-new functions with unambiguous signatures on a PHP 8.3 codebase, so the types are free to add and remove a class of caller bug: `wprtt_counting_guards_enabled(): bool`, `wprtt_is_bot_request( string $user_agent ): bool`, `wprtt_get_dedup_identity(): string`, `wprtt_should_count_view( int $post_id, string $client_id, int $window = 30 * MINUTE_IN_SECONDS ): bool`. The defensive `(string)` / `absint()` casts inside the bodies then become belt-and-braces rather than the only thing standing between a bad caller and a wrong count. (The plugin ships no `Requires PHP` header — scalar types have been available since PHP 7.0, so this is safe regardless of the intended floor.)

plugins/republication-tracker-tool/includes/pixel-functions.php
adekbadek: **Nit:** Prefer `wp_hash()` over hand-rolling `md5( wp_salt( 'auth' ) . '|' . … )` — Using the auth salt as a keying secret here isn't a security problem — nothing derived from it reaches the client — but `salt . '|' . $data` + MD5 is a bespoke reinvention of `wp_hash()`, which is `hash_hmac( 'md5', $data, wp_salt( 'auth' ) )`: same secret, standard HMAC construction, one WP API call. `'ipua_' . wp_hash( $ip . '|' . $ua )`. Worth a line of comment either way noting that rotating the site's salts resets every dedup identity and hence the window — harmless (it fails open), but it would save a future reader the analysis.

plugins/republication-tracker-tool/includes/pixel.php
```diff
@@ -38,129 +38,120 @@ function wprtt_get_referring_page_title( $url ) {
...
+	// The dedup identity falls back to IP + user agent, since cross-site pixel
+	// requests usually arrive without cookies; the GA4 client ID below keeps
+	// the original cookie-or-generated behavior.
+	if ( ! $wprtt_guards_enabled || wprtt_should_count_view( $shared_post_id, wprtt_get_dedup_identity() ) ) {
+		$wprtt_client_id = wprtt_extract_cid_from_cookies();
+		// The title fetch is a blocking outbound request only the GA4 event needs,
+		// so it runs only for views that actually count. The referrer is
+		// client-controlled: validate it (blocks non-http(s) and local/private
+		// targets) before fetching anything server-side.
+		$url_title = '' !== $url && wp_http_validate_url( $url ) ? wprtt_get_referring_page_title( $url ) : '';
```
adekbadek: **Nit:** The `wp_http_validate_url()` gate is a second, smaller flag-off divergence — Validating the referrer before fetching it server-side is the right call and should stay — it closes the request-forgery exposure. But it isn't flag-gated, so it changes behavior for every publisher on the next release, not just the pilot site. `wp_http_validate_url()` performs a blocking `gethostbyname()` and returns false for hosts with no A record (an IPv6-only republisher) or on ports other than 80/443/8080, so in those cases the GA4 event's `page_title` now arrives empty where it previously carried the referring page's title. The view counter itself is unaffected — the meta array is still keyed on the raw `esc_url_raw()` referrer. Worth a line in the PR description so it isn't a surprise, since it is a real, if narrow, divergence from "the flag-off path is identical".

adekbadek (changes requested):
Three of the four blockers land on the **flag-off** path — the path every publisher gets on the next release — so the PR's central safety claim, that flag-off is behavior-identical to today, is the thing to fix before this merges.

plugins/republication-tracker-tool/includes/pixel.php
wil-gerken: Fixed in 770e11f7e. Missing, non-numeric, and non-positive `post` values are now rejected before any lookup: the resolver is `wprtt_resolve_shared_post()` in pixel-functions.php, validating with `FILTER_VALIDATE_INT` (minimum 1).

Rechecking this finding turned up three more shapes `absint()` coerces into real IDs: arrays (`?post[]=x` casts to 1), negatives (`-47` becomes 47), and trailing characters (`47abc` becomes 47). The resolver refuses all three.

The tests cover both false-positive targets. A post exists at ID 1, and a decoy sits in `$GLOBALS['post']` so a `get_post( 0 )` fallback would resolve to something. A refusal therefore can't pass for the wrong reason. The `YOUR-POST-ID` placeholder from the settings page's embed snippet is an explicit case in the data provider.

plugins/republication-tracker-tool/includes/pixel.php
wil-gerken: Fixed in 770e11f7e, by the same move as the `newspack-cid` cookie thread (pixel.php:96): the counting path reads existing cookies but never creates one, so the identity no longer flips keys because of a cookie the pixel itself set.

One scope note, reasoning from SameSite semantics rather than from logs: on real cross-site image loads the browser withholds the cookie, so the double-count would have reached cookie-returning clients. The fix is the same either way.

`test_guarded_counter_dedups_repeat_views` confirms three same-client views count as one.

plugins/republication-tracker-tool/includes/pixel.php
wil-gerken: Fixed in 770e11f7e. The minting variant (`wprtt_extract_cid_from_cookies()`) is back inside the fully-configured GA4 branch, the position it holds on `release`. The counting and dedup paths call the read-only `wprtt_read_cid_from_cookies()` only.

`test_counting_path_never_mints_the_cid_cookie` pins it: headers are already sent under PHPUnit, so the test fails if the counting path tries to set a cookie. That depends on PHPUnit 9's `convertWarningsToExceptions`, and a docblock note flags re-verification on a toolchain upgrade.

plugins/republication-tracker-tool/tests/test-pixel-functions.php
wil-gerken: Fixed in 770e11f7e with the refactor this thread sketched. pixel.php is a thin handler; the request-level decisions live in `wprtt_resolve_shared_post()` and `wprtt_record_view()` in pixel-functions.php. Both flag-off regressions are pinned: no view recorded for an unresolvable post ID, and no cookie without full GA4 config. The suite went from 30 to 64 tests.

One design note: the rework also reshaped the flag, and the PR body is updated to match. The flag now runs in shadow mode, so the raw counter shares one counting path in both states and the guarded count is written alongside it. "Flag-off is identical" became "both states share the counting code," which is a stronger claim to test. The flag changes how many requests reach that shared path, not what it does — that volume shift is the pixel.php:46 thread.

plugins/republication-tracker-tool/includes/pixel-functions.php
wil-gerken: Adopted in 770e11f7e:

- The token is now `pinterest\/0\.`, the crawler's product token. `bot` separately matches Pinterestbot.
- The in-app UAs for Pinterest, Facebook, and Instagram are test cases in the browser corpus.
- The pattern gained a `wprtt_bot_request_pattern` filter, so a false positive found in production can be overridden per site while a default-pattern change waits for a release.

plugins/republication-tracker-tool/includes/pixel-functions.php
wil-gerken: Both adopted in 770e11f7e. Accept-Language is mixed into the identity hash, now `wp_hash()` per the pixel-functions.php:132 hashing thread. The shadow-count suggestion became the PR's design: the raw counter stays the displayed number in both flag states as the control, and the guarded count writes to a second meta key.

Extending the control-number point one step: the first guarded view for a post also snapshots the raw counter and a start timestamp in `republication_tracker_tool_sharing_guarded_baseline` meta. The raw number carries lifetime history the guarded one doesn't, so the pilot compares guarded against raw-minus-baseline over the same window. The README documents the WP-CLI reads.

plugins/republication-tracker-tool/includes/pixel-functions.php
wil-gerken: Adopted in 770e11f7e: `wp_cache_add()` when a persistent object cache exists, which is atomic and closes the simultaneous-hit race. Transients remain the documented WordPress.org fallback.

One addition beyond the suggestion. `wp_cache_add()` also returns false when the backend is down or flapping, which would have silently stopped the guarded count from advancing. A follow-up `wp_cache_get()` miss now separates "already counted" from "broken backend," and the broken case counts the view. The comment notes eviction as the same fail-open direction.

plugins/republication-tracker-tool/includes/pixel.php
wil-gerken: Deferred by design, with the analysis folded into the PR body rather than dropped. The relocation to a pre-query hook is its own PR, tracked in NPPM-3106. It's required before fleet enablement, which is when the per-hit cost arrives, but not before this merge or the single-site pilot.

The direction point shaped the PR body, which now discloses that raw counts can rise while the flag is on. One refinement: on a fully GA4-configured site the pixel response mints a cookie, and batcache declines to cache Set-Cookie responses, so those hits were already reaching PHP. The drift concentrates on sites with a measurement ID but no API secret. The `DONOTCACHEPAGE`-for-non-batcache-stacks comment is in as of 91db1a619.

If you read the relocation as a merge blocker rather than a fleet-enablement one, say so and I'll pull it into this PR.

plugins/republication-tracker-tool/includes/pixel-functions.php
wil-gerken: Adopted in 770e11f7e: every function in pixel-functions.php carries parameter and return types (`?WP_Post`, `bool`, `string`, `void`, typed defaults).

plugins/republication-tracker-tool/includes/pixel-functions.php
wil-gerken: Adopted in 770e11f7e: `'ipua_' . wp_hash( $ip . '|' . $ua . '|' . $lang )`. The comment records the salt-rotation consequence: rotating resets every dedup identity and restarts the window, failing open.

plugins/republication-tracker-tool/includes/pixel.php
wil-gerken: Disclosed in the PR body's divergence list, and the gate grew while rechecking it in 770e11f7e. The fetch now goes through `wp_safe_remote_get()`, so redirect hops are validated too; plain `wp_remote_get()` follows a 302 without re-validating. Both outbound GA4-branch requests carry a 2-second timeout.

One accepted limitation, noted in the body and tracked in NPPM-3106: `wp_http_validate_url()` does not block 169.254.0.0/16, so a link-local address remains reachable on the first hop. The exposure is narrow — no response body reaches the requester, and the only output is the page title, which lands in the site's own GA4 property.

plugins/republication-tracker-tool/includes/pixel.php
adekbadek: Not a merge blocker — the relocation is a fleet-enablement gate, and NPPM-3106 is the right place for it. The guards ship default-off, so nothing on the next release pays the per-hit cost; the bill only arrives when `WPRTT_COUNTING_GUARDS_ENABLED` is defined somewhere, and a single pilot site's republished traffic is a bounded amount of it. The condition worth holding to is that NPPM-3106 lands before the constant goes anywhere beyond the pilot site.

The Set-Cookie refinement is right, and stronger than stated: `newspack-cid` is set with no SameSite attribute, so browsers apply Lax and withhold it on the cross-site `` load. A fully GA4-configured site therefore mints a fresh cid on nearly *every* pixel request, not just the first — those responses have always carried Set-Cookie and were never batcached. So the drift really is confined to sites with a measurement ID but no API secret, and the README's "can read somewhat higher" wording covers it.

`DONOTCACHEPAGE` comment confirmed at pixel.php:23-28.

adekbadek (approved):
Flag-off is now behavior-identical on the counting path — strict post-ID resolution closes the `get_post( 0 )` global-post trap, the raw counter keeps its rules, and no guarded meta, baseline, or cache header is written — so the guarded count is a pure shadow and this is safe to ship default-off. Two non-blocking notes: `tests/bootstrap.php` now requires `vendor/autoload.php` twice, since release's guarded loader arrived in the merge and left the branch's unconditional require dead; and the "concurrent first views can't write a second baseline" comment overstates `add_post_meta( …, true )`, which is check-then-act rather than an atomic insert (harmless here — a duplicate row is never read). Relocation to a pre-query hook stays a fleet-enablement gate under NPPM-3106, per the batcache thread.

PR #597, opened 2026-07-09
fix(blocks): fix express checkout address validation in modal checkout

rbcorrales:
### All Submissions:

Express checkout wallets (Apple Pay, Google Pay) in the modal checkout submit to the Store API (`wc/store/v1/checkout`), whose address validation rejects any state value not in the country's list. Wallets can supply values the buyer never sees and cannot correct (for example a locality in the state slot), hard-failing the payment on a field the site is configured not to collect.

The fix is scoped strictly to the modal checkout, so standard Woo checkout flows keep stock behavior (per review discussion, the modal-only scoping of the billing fields setting is deliberate):

* A `rest_pre_dispatch` scrub drops configured-off state/postcode values from modal-originated `wc/store/v1/checkout` requests, and only when the value would fail validation anyway; valid wallet values pass through untouched and are stored on the order. Modal origin is detected via the referer (the modal iframe URL carries `modal_checkout=1`), since express submissions carry no request params.
* A `woocommerce_get_country_locale` filter relaxes the required flag for configured-off state/postcode on modal-originated requests, so the scrubbed value's absence is accepted. It never applies when the cart needs a shipping address.
* The existing modal-only field removal is unchanged; its docblock now records that the modal-only scoping is by design.

Everything outside the modal is byte-for-byte stock: the standard Woo checkout page, the blocks checkout, express buttons on product/cart pages, My Account flows, and any cart needing a shipping address.

1. In Audience > Configuration > Checkout & Payment > Billing Fields, uncheck the state and address fields (keep name and email) and save.
2. Open the modal checkout via a donate block and confirm the form shows only the configured fields.
3. Complete an express payment (Apple Pay / Google Pay) in the modal using a wallet address whose state the store rejects (to simulate, temporarily filter `woocommerce_states` to replace the state list for your wallet's country). Confirm checkout completes with the invalid state dropped, instead of failing with "The provided state (...) is not valid".
4. Open the standard checkout page with a virtual product in the cart. Confirm the full address field set renders despite the reduced configuration.
5. On that standard checkout page, attempt the same express payment. Confirm stock behavior (the address validation error appears), since the fix must not apply outside the modal.
6. Add a physical (shippable) product and confirm standard checkout still renders and requires the full address field set.
7. Pay for an order from My Account and confirm those fields are unchanged.
8. Clear the billing fields configuration (default field set) and confirm modal express checkout validates addresses as stock.

### Other information:

### DIFF
--- a/plugins/newspack-blocks/includes/class-modal-checkout.php
+++ b/plugins/newspack-blocks/includes/class-modal-checkout.php
@@ -1402,8 +1402,9 @@ final class Modal_Checkout {
 	 * Express checkout wallets submit to wc/store/v1/checkout and can supply
 	 * values the buyer never sees or corrects (e.g. Apple Pay sending a suburb
 	 * as the state), which hard-fails Store API address validation. Values the
-	 * validation would accept are left untouched. Runs on rest_pre_dispatch
-	 * because the validation happens during dispatch.
+	 * validation would accept are left untouched, and only the billing address
+	 * is scrubbed. Runs on rest_pre_dispatch because the validation happens
+	 * during dispatch.
 	 *
 	 * @param mixed            $result  Response to replace the requested version with.
 	 * @param \WP_REST_Server  $server  Server instance.
@@ -1434,17 +1435,21 @@ final class Modal_Checkout {
 			return $result;
 		}
 
-		foreach ( [ 'billing_address', 'shipping_address' ] as $param ) {
-			$address = $request->get_param( $param );
+		// Physical-goods flows are never modified. The cart is not always
+		// initialized this early in Store API requests; when it is, bail for
+		// carts that need a shipping address. Only the billing address is
+		// scrubbed, so shipping data is never touched either way.
+		if ( \WC()->cart && \WC()->cart->needs_shipping_address() ) {
+			return $result;
+		}
 
-			if ( ! is_array( $address ) ) {
-				continue;
-			}
+		$address = $request->get_param( 'billing_address' );
 
+		if ( is_array( $address ) ) {
 			$scrubbed = self::scrub_invalid_address_values( $address, $billing_fields );
 
 			if ( $scrubbed !== $address ) {
-				$request->set_param( $param, $scrubbed );
+				$request->set_param( 'billing_address', $scrubbed );
 			}
 		}
 
--- a/plugins/newspack-blocks/includes/class-modal-checkout.php
+++ b/plugins/newspack-blocks/includes/class-modal-checkout.php
@@ -1345,21 +1345,20 @@ final class Modal_Checkout {
 	}
 
 	/**
-	 * Remove the configured-off billing fields from checkout.
+	 * Remove the configured-off billing fields from modal checkout requests.
 	 *
-	 * Not limited to modal checkout requests (NPPM-2937): the fields must be
-	 * unregistered wherever WooCommerce renders or validates them. When they were
-	 * only removed from the modal render, they stayed registered on the standard
-	 * checkout page, and express checkout wallets (e.g. Apple Pay) would populate
-	 * and fail validation on fields the buyer could not see or edit.
+	 * Deliberately scoped to modal checkout: some publishers rely on standard
+	 * Woo checkout flows that predate Audience Management, so those keep the
+	 * stock field set. The express checkout gap this leaves (wallets submitting
+	 * values for fields the buyer cannot see) is handled for modal-originated
+	 * requests by scrub_store_api_checkout_address() below.
 	 *
 	 * @param array $fields Checkout fields.
 	 *
 	 * @return array
 	 */
 	public static function woocommerce_checkout_fields( $fields ) {
-		// My Account checkouts relax required flags instead of removing fields.
-		if ( method_exists( 'Newspack\WooCommerce_My_Account', 'is_from_my_account' ) && \Newspack\WooCommerce_My_Account::is_from_my_account() ) {
+		if ( ! self::is_modal_checkout() ) {
 			return $fields;
 		}
 		$cart = \WC()->cart;
@@ -1397,7 +1396,7 @@ final class Modal_Checkout {
 
 	/**
 	 * Scrub invalid address values from Store API checkout requests when the
-	 * corresponding billing fields are configured off (NPPM-2937).
+	 * corresponding billing fields are configured off.
 	 *
 	 * Express checkout wallets submit to wc/store/v1/checkout and can supply
 	 * values the buyer never sees or corrects (e.g. Apple Pay sending a suburb
@@ -1405,6 +1404,10 @@ final class Modal_Checkout {
 	 * validation would accept are left untouched, and only the billing address
 	 * is scrubbed. Runs on rest_pre_dispatch because the validation happens
 	 * during dispatch.
+	 *
+	 * Scoped to requests originating from the modal checkout, so any Store API
+	 * checkout outside the modal (e.g. the blocks checkout page or express
+	 * buttons on product pages) keeps stock behavior.
 	 *
 	 * @param mixed            $result  Response to replace the requested version with.
 	 * @param \WP_REST_Server  $server  Server instance.
@@ -1422,6 +1425,10 @@ final class Modal_Checkout {
 			'POST' !== $request->get_method() ||
 			0 !== strpos( $request->get_route(), '/wc/store/v1/checkout' )
 		) {
+			return $result;
+		}
+
+		if ( ! self::is_modal_checkout_referer() ) {
 			return $result;
 		}
 
@@ -1454,6 +1461,23 @@ final class Modal_Checkout {
 		}
 
 		return $result;
+	}
+
+	/**
+	 * Whether the current request originates from the modal checkout, based on
+	 * the referer.
+	 *
+	 * Express checkout submissions go to the Store API as JSON, so the usual
+	 * modal_checkout request params are absent. The requests are made from the
+	 * modal checkout iframe, whose URL carries modal_checkout=1, so it shows up
+	 * as the (same-origin) referer.
+	 *
+	 * @return bool
+	 */
+	private static function is_modal_checkout_referer() {
+		$referer = isset( $_SERVER['HTTP_REFERER'] ) ? esc_url_raw( wp_unslash( $_SERVER['HTTP_REFERER'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
+		return false !== strpos( $referer, 'modal_checkout=1' );
 	}
 
 	/**
@@ -1496,7 +1520,7 @@ final class Modal_Checkout {
 	}
 
 	/**
-	 * Relax locale-required flags for configured-off billing fields (NPPM-2937).
+	 * Relax locale-required flags for configured-off billing fields.
 	 *
 	 * After an invalid wallet value is scrubbed (see
 	 * scrub_store_api_checkout_address), the Store API order validation would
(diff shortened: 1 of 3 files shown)


plugins/newspack-blocks/tests/test-modal-checkout.php
```diff
@@ -43,14 +43,73 @@ function wcs_get_product_limitation( $product_id ) { // phpcs:ignore WordPress.N
...
 	public function tear_down() {
-		global $newspack_blocks_test_limited_product_id, $newspack_blocks_test_limited_user_id;
+		global $newspack_blocks_test_limited_product_id, $newspack_blocks_test_limited_user_id, $newspack_blocks_test_wc;
 		$newspack_blocks_test_limited_product_id = null;
 		$newspack_blocks_test_limited_user_id    = null;
+		$newspack_blocks_test_wc                 = null;
+		\Newspack\WooCommerce_My_Account::$is_from_my_account = false;
 		remove_all_filters( 'woocommerce_cart_item_removed_message' );
 		unset( $_POST['billing_email'], $_POST['post_data'], $_REQUEST['modal_checkout'], $_REQUEST['post_data'] );
 		parent::tear_down();
```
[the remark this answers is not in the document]
rbcorrales: Added the explicit remove_all_filters for the hook in tear_down (bec03d98), matching the file's existing cleanup convention. Worth noting WP_UnitTestCase also backs up and restores all hooks between tests, so the closure could not actually leak, but the explicit reset makes the intent clear.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1395,148 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+		if ( ! function_exists( 'WC' ) || ! class_exists( 'WC_Validation' ) ) {
+			return $result;
+		}
+
+		$billing_fields = apply_filters( 'newspack_blocks_donate_billing_fields_keys', [] );
+
+		if ( empty( $billing_fields ) ) {
+			return $result;
+		}
+
```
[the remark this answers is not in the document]
rbcorrales: Addressed in afb599d. The scrub now bails when the cart is available and needs a shipping address, and only ever touches billing_address, so shipping data is never modified. Worth noting the failure mode couldn't actually be created by the scrub (an invalid state 400s the checkout with or without it, only the error message differed), but restricting to billing plus the cart bail removes the concern entirely.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1395,148 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+	 * @param mixed            $result  Response to replace the requested version with.
+	 * @param \WP_REST_Server  $server  Server instance.
+	 * @param \WP_REST_Request $request Request used to generate the response.
+	 *
+	 * @return mixed
+	 */
+	public static function scrub_store_api_checkout_address( $result, $server, $request ) {
+		if ( null !== $result ) {
+			return $result;
+		}
```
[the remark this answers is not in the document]
rbcorrales: Added in c0f3ca4: 8 tests covering the Store API scrub (invalid state dropped when configured off, valid code/name values kept, invalid postcode dropped, no-op for configured-on fields, default config, and shipping carts) and the locale relaxation (configured-off state/postcode relaxed, city untouched, no-op for default config, configured-on fields, and shipping carts). Suite passes: 45 tests, 77 assertions.

plugins/newspack-blocks/tests/test-modal-checkout.php
```diff
@@ -43,15 +43,75 @@ function wcs_get_product_limitation( $product_id ) { // phpcs:ignore WordPress.N
...
 	/**
 	 * Clean up request data.
 	 */
 	public function tear_down() {
-		global $newspack_blocks_test_limited_product_id, $newspack_blocks_test_limited_user_id;
+		global $newspack_blocks_test_limited_product_id, $newspack_blocks_test_limited_user_id, $newspack_blocks_test_wc;
 		$newspack_blocks_test_limited_product_id = null;
 		$newspack_blocks_test_limited_user_id    = null;
+		$newspack_blocks_test_wc                 = null;
+		\Newspack\WooCommerce_My_Account::$is_from_my_account = false;
```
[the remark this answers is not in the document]
rbcorrales: Guarded in c0f3ca4: the tear_down reset and the My Account test now check property_exists first, and the test skips with a message when the real class is loaded instead of the mock.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1395,153 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+
+		if ( empty( $off ) ) {
+			return $locale;
+		}
+
+		foreach ( array_keys( $locale ) as $country ) {
+			foreach ( $off as $address_key ) {
+				$locale[ $country ][ $address_key ]['required'] = false;
+			}
+		}
```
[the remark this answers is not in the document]
rbcorrales: Intentional, and restricting to existing keys would under-relax. Store API required validation (OrderController::validate_address_fields) merges the default locale into each country via wp_parse_args, so countries without their own state/postcode entry inherit required from the default (the US entry, for example, only defines a label for state). The loop covers the default key too, which is what relaxes countries without explicit entries, and the per-country writes override locales that explicitly set required true. The created overrides contain only required=false and are gated to the intended cases (non-empty config, field configured off, cart without shipping), so no other field behavior changes.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1394,185 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+
+		if (
+			! in_array( 'billing_state', $billing_fields, true ) &&
+			! empty( $address['state'] ) &&
+			$country
+		) {
+			$states = \WC()->countries->get_states( $country );
+
+			if ( is_array( $states ) && ! empty( $states ) ) {
+				$code_match = array_key_exists( strtoupper( $address['state'] ), array_change_key_case( $states, CASE_UPPER ) );
```
dkoo, line 1515: **Suggestion:** The scrub runs `strtoupper()` / `strtolower()` / `WC_Validation::is_postcode()` on the raw address value here, on `rest_pre_dispatch` — before the Store API schema sanitizes or rejects the request. If a modal checkout request sends `billing_address.state` (or `.postcode`) as a JSON array instead of a string, these throw a PHP `TypeError` and the endpoint returns HTTP 500 instead of the clean 400 the schema would have returned. I reproduced this on PHP 8.3 (`strtoupper(): Argument #1 ($string) must be of type string, array given`). Only reachable for modal-originated requests on sites with the field configured off, and the request was going to fail either way — but a one-line `is_string()` guard on the value before transforming turns the 500 back into the intended 400.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1394,185 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+
+		// Physical-goods flows are never modified. The cart is not always
+		// initialized this early in Store API requests; when it is, bail for
+		// carts that need a shipping address. Only the billing address is
+		// scrubbed, so shipping data is never touched either way.
+		if ( \WC()->cart && \WC()->cart->needs_shipping_address() ) {
+			return $result;
+		}
+
+		$address = $request->get_param( 'billing_address' );
```
dkoo, line 1458: **Suggestion:** The scrub only touches `billing_address` and bails entirely when the cart needs shipping. WooCommerce's `CheckoutSchema` registers a `validate_callback` on the `shipping_address` argument itself, which runs whenever a `shipping_address` payload is submitted regardless of cart shipping state — so a wallet that attaches a `shipping_address` with an invalid state on a virtual/donation cart (shipping not needed) could still hard-fail the same validation this fix targets, unscrubbed. Likelihood is low (wallets generally return only billing contact when shipping isn't requested), so this is a consider-scrubbing-symmetrically note rather than a blocker — or document why `shipping_address` is intentionally out of scope.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1375,6 +1394,185 @@ public static function woocommerce_checkout_fields( $fields ) {
...
+	 * the referer.
+	 *
+	 * Express checkout submissions go to the Store API as JSON, so the usual
+	 * modal_checkout request params are absent. The requests are made from the
+	 * modal checkout iframe, whose URL carries modal_checkout=1, so it shows up
+	 * as the (same-origin) referer.
+	 *
+	 * @return bool
+	 */
+	private static function is_modal_checkout_referer() {
```
dkoo, line 1482: **Nit:** `is_modal_checkout_referer()` matches with a raw `false !== strpos( $referer, 'modal_checkout=1' )` against the whole referer URL, whereas this class already parses the referer properly a few lines down in `is_express_checkout()` via `wp_parse_url( …, PHP_URL_QUERY )` + `wp_parse_str`. The `strpos` form also matches `modal_checkout=1` appearing in a path segment, fragment, or another param's value. It's functionally fine for the real iframe URL and the security impact is nil (see the review summary), but reusing the existing `wp_parse_url`/`wp_parse_str` idiom would keep the two modal-detection paths consistent and slightly harder to trip accidentally.

dkoo (approved):
Tested end-to-end on a dev site with Apple Pay enabled. All testing steps pass as described—the invalid address fields are scrubbed on an express checkout request during a modal checkout where those fields are disabled, allowing the checkout to complete, while default validation prevails when address fields are required.

Approved with comments. This is a careful, well-tested hotfix. The two new Store-API hooks are correctly wired (right `rest_pre_dispatch` signature and null-return contract, `set_param` writes to the bucket the Store API reads, cart/shipping guards in place), and the scrub only ever drops values that would fail WooCommerce validation anyway — valid wallet values pass through and are stored. I verified the core logic end-to-end against a live WooCommerce env (invalid state dropped, valid code kept, configured-on field untouched, `required` relaxed only under a modal referer) and confirmed no regression in standard checkout. The referer-trust surface is benign: a forged referer can only loosen `required` on two optional fields for the caller's own non-shipping cart — no control is bypassed and there's no cross-user effect. All prior Copilot comments are addressed.

One thing worth fixing before merge, plus two minor notes below. Not blocking.

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Fixed in fb52290. Added an `is_string()` guard on `state` and `postcode` before the case transforms and `is_postcode()`, so an array-valued field now falls through to the schema's clean 400 instead of a `TypeError` 500. Covered by a new test (`test_store_api_scrub_ignores_non_string_values`).

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Documented rather than expanded scope, in fb52290. Added a docblock note that the shipping address is intentionally out of scope: carts that need shipping bail before the scrub, and wallets return only billing contact when shipping is not requested, so a virtual-cart request carries no `shipping_address` to scrub. Kept the fix billing-only to avoid touching shipping data on a hotfix.

plugins/newspack-blocks/includes/class-modal-checkout.php
rbcorrales: Done in fb52290. `is_modal_checkout_referer()` now reuses the `wp_parse_url( …, PHP_URL_QUERY )` + `wp_parse_str` idiom from `is_express_checkout()` and matches on the actual `modal_checkout` query param instead of `strpos` on the whole URL.

PR #604, opened 2026-07-13
fix(group-subscription): hide add-member fields at member limit in admin metabox (NPPD-1483)

adekbadek:

In the WP-admin WooCommerce Subscriptions **Group subscription** metabox, the "Add new group members" reader-search select and the invite-by-email field stayed visible even when the group was already at its member limit. An admin only discovered the limit *after* attempting an add, which the server rejected with a 409 — the poor UX reported in [NPPD-1483](https://linear.app/a8c/issue/NPPD-1483).

https://github.com/user-attachments/assets/5fc40454-b710-4932-9706-ea97048e5ef1

**Metabox at-limit state:**
- At-limit = raw `Group_Subscription::get_members()` + pending (non-expired) invites `>=` `Group_Subscription::get_member_seat_limit()` — the same threshold the server enforcement and the reader-facing My Account view use. That is the *member-seat* limit: the configured limit is owner-inclusive, so the seat of every manager who holds no member meta (in practice the owner) is reserved out of it, which keeps it in step with `get_members()`, which excludes the owner (limit `0` = unlimited). Comparing against the raw configured limit would be one seat too generous and would still render the add form at the boundary. When reached, the add form is hidden and a warning notice (`role="status"`) shows instead. Members the server counts but that aren't rendered as removable rows (a non-reader member, or a manager who also carries member meta) are folded into the client tally via `data-spots-offset`.
- The admin JS re-evaluates on add / remove / invite / cancel (`refreshLimitState()`), so it stays in sync without a reload; `data-member-limit` carries the same member-seat limit, so the live updates inherit the server threshold. The "invitation sent" confirmation renders in the always-visible members section so reaching the limit doesn't hide its own success message.

**Server enforcement made consistent** (previously deferred, now folded in):
- `Group_Subscription::update_members()` now counts pending invites toward the limit too, so a direct admin add can't overfill a group that already has invites out — the admin metabox and the server now enforce the same ceiling. An invite addressed to a user being added is excluded from that count (adding them fulfils it), so an invite can still be accepted at the limit.
- The additions are narrowed to the IDs that would genuinely become members (non-readers and users who already hold this subscription's member meta were already skipped by the add loop) *before* the limit is projected, so a batch carrying a no-op ID can't be rejected for an overshoot it would never have caused.
- A direct add now **cancels** any matching pending invite (case-insensitively, in `cancel_invites()` as well as in the limit tally), mirroring what the acceptance path already did — otherwise the new member and their stale invite would both linger and both reserve a spot. `accept_invite()` correspondingly treats an already-fulfilled invite as a success for the invited email's user, not just the current user, so a reader who was added directly and then clicks their old link still lands on the friendly "you're already a member" path.
- The limit check now runs only when there are additions, fixing a latent case where removing a member from an over-capacity group (e.g. after lowering the limit) could return a spurious "limit reached" error.

**Hardening in touched files:**
- The select2 error handler no longer interpolates a server message into `.before()` HTML — it uses the skeleton + `.text()` pattern like the other handlers.
- JS-appended row screen-reader labels ("Remove"/"Cancel") and the "Select a reader…" option are now localized.

Seven unit tests cover the new server behavior (a pending invite counts toward the limit; adding the invited user isn't double-counted; an ID the add would skip anyway takes no seat in the projection; a direct add cancels the matching invite, including a case-mismatched one; an expired invite reserves nothing; a removal-only call succeeds on an over-limit group), plus a data-provider test that renders the metabox itself and asserts the `is-at-limit` class, the `data-member-limit` value and the hidden add form across the limit boundary (including the minimum real limit of 2). That render test is the only coverage of the metabox callback, where a fatal would take down the whole subscription edit screen.

Closes [NPPD-1483](https://linear.app/a8c/issue/NPPD-1483).

1. On a site with `NEWSPACK_CONTENT_GATES` on + RAS enabled + WooCommerce Subscriptions, create a subscription product with **Group subscription** enabled and a member limit (e.g. 1), and a subscription for it.
2. Open the subscription in WP admin → **Group subscription** metabox. Fill it to the limit (add a member or invite). The add select + invite field disappear and a "member limit reached" notice appears — without saving/reloading.
3. Remove a member (or cancel an invite): the add form reappears live.
4. From below the limit, invite an email that reaches the limit: the pending invite shows in the list, the form hides, and the "Invitation sent successfully." confirmation stays visible.
5. Edge case: make a member a non-reader (e.g. an editor) at the limit — the metabox still correctly reports at-limit (previously it undercounted and left the form open).
6. Boundary: on a group with a limit of 2 (owner + one member seat), add a single member — the add form hides right away, matching the 409 `update_members()` would return for a second one.
7. Server parity: with a pending invite filling the last spot, a direct add via the members REST endpoint returns a 409 (the invite counts); accepting that invite still succeeds.
8. Invite cleanup: invite an email (in mixed case), then add that same reader directly from the metabox — the pending invite disappears from the list (it no longer reserves a spot), and opening the old invite link still shows the "already a member" success rather than an error.

### Other information:

* Verified end-to-end in an isolated env (Playwright) — at-limit hide/show on load and live, the non-reader edge case, and the invite-confirmation-survives case. The full `newspack-plugin` PHPUnit suite passes (2357 tests, 6684 assertions; the 7 skips are pre-existing and unrelated), including the group-subscription classes. PHPCS / ESLint clean.
* Reviewed via multi-reviewer code review across two rounds. The one nit that had been deferred — a pending invite for a reader added via a manual/admin path kept reserving a spot until the reader engaged or it expired — is now fixed in this PR (see the invite-cancellation bullet above), so nothing is left as a follow-up.


plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -93,7 +113,7 @@ import './admin.scss';
 						.closest( '.newspack-group-subscription__container' )
 						.find( '.newspack-group-subscription__members-count' );
 					$membersList.append(
-						`<li><a class="newspack-group-subscription__member-user-link" href="#"></a><a href="#" class="newspack-group-subscription__remove-member">&#215; <span class="screen-reader-text">Remove</span></a></li>`
+						`<li data-consumes-spot="1"><a class="newspack-group-subscription__member-user-link" href="#"></a><a href="#" class="newspack-group-subscription__remove-member">&#215; <span class="screen-reader-text">Remove</span></a></li>`
```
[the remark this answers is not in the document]
adekbadek: Fixed in 9e5791e. `addMember()` now derives `$container` from the select once and scopes both `$membersList` and `$membersCount` to it, so the append, the re-tally and `refreshLimitState()` all operate on the same container.

I also applied the same fix to `removeMember()`, which had the identical global `$( '.newspack-group-subscription__members-list' )` lookup (pre-existing, not flagged), and to the `$email` input in `inviteMember()`, which was scoped to the whole metabox rather than the container. Every DOM lookup in the four handlers is now container-scoped.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -191,23 +216,24 @@ import './admin.scss';
...
-				$this.parent().find( '.message' ).text( newspackGroupSubscriptions.success_message );
 				const $membersList = $( '.newspack-group-subscription__members-list' );
-				const $membersCount = $( '#_newspack_group_subscription_member_ids' )
-					.closest( '.newspack-group-subscription__container' )
-					.find( '.newspack-group-subscription__members-count' );
+				const $membersCount = $container.find( '.newspack-group-subscription__members-count' );
 				$membersList.find( `li[data-email="${ data.email }"]` ).remove();
 				$membersList.append(
-					`<li data-email="${ data.email }"><span class="newspack-group-subscription__pending-invite"></span> <span class="newspack-group-subscription__pending-invite-label"></span><a href="#" class="newspack-group-subscription__cancel-invite">&#215; <span class="screen-reader-text">Delete</span></a></li>`
+					`<li data-email="${ data.email }" data-consumes-spot="1"><span class="newspack-group-subscription__pending-invite"></span> <span class="newspack-group-subscription__pending-invite-label"></span><a href="#" class="newspack-group-subscription__cancel-invite">&#215; <span class="screen-reader-text">Delete</span></a></li>`
```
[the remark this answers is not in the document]
adekbadek: Fixed in 9e5791e. `inviteMember()` now takes `$membersList` from `$container`, matching `$membersCount` and `refreshLimitState()`, so the stale-invite removal and the new row both land in the right list. The `$email` input lookup in the same handler was also scoped to the metabox rather than the container, so I moved that onto `$container` too.

Verified against a running env: inviting a second member on a limit-2 group appends the pending row, bumps the count to 3, adds `is-at-limit` and hides the add form.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -245,11 +272,10 @@ import './admin.scss';
 					throw new Error( data.message || 'Failed to cancel invite' );
 				}
 				const $membersList = $( '.newspack-group-subscription__members-list' );
-				const $membersCount = $( '#_newspack_group_subscription_member_ids' )
-					.closest( '.newspack-group-subscription__container' )
-					.find( '.newspack-group-subscription__members-count' );
+				const $membersCount = $container.find( '.newspack-group-subscription__members-count' );
 				$listItem.remove();
 				$membersCount.text( $membersList.find( 'li' ).length );
+				refreshLimitState( $container );
```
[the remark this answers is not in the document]
adekbadek: Fixed in 9e5791e. `cancelInvite()` now scopes `$membersList` to `$container`, consistent with `$membersCount` and `refreshLimitState()`.

Verified against a running env: cancelling the pending invite on an at-limit group removes the row, drops the count back to 2, clears `is-at-limit` and re-reveals the add form.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -367,14 +367,33 @@ public static function update_members( $subscription, $members_to_add, $members_
...
+		// The limit only bounds additions, so skip the check on removal-only calls: a removal can
+		// never push a group over its limit, and running it there would spuriously 409 an admin
+		// removing a member from an already-over-capacity group (e.g. after lowering the limit).
+		// Removals are persisted above; no shipped caller batches add + remove in one call (the admin
+		// JS and admin-post handlers split them), so a combined over-limit add can't strand a removal.
+		if ( ! empty( $members_to_add ) && $subscription_settings['limit'] > 0 ) {
+			// Pending (non-expired) invites reserve a spot, so count them alongside existing members --
+			// this matches the invite path's own limit check, so a direct add can't overfill a group
+			// that already has invites out. Exclude any invite addressed to a user being added: the
+			// add fulfils that invite (it is cancelled below), so it must not double-count against them.
```
chickenn00dle, line 635: **Severity:** suggestion

A direct admin add via the `/members` endpoint doesn't cancel a matching pending invite the way `accept_invite()` does. Since this PR now counts pending non-expired invites toward the limit, adding a reader by ID who still has an outstanding email invite leaves both the member row and the pending-invite row in place — that reader is then double-counted, which can prematurely trip the at-limit state (hiding the add form) and `409` later legitimate adds. It fails safe (over-strict, never over-permissive), so it's not a correctness or security problem, but it's a reachable UX inconsistency the new counting introduces. Consider cancelling any matching-email invite in the direct-add loop, or documenting the limitation. Also, the comment saying the invite "is cancelled below" is inaccurate for this method — the cancellation lives in the caller (`accept_invite()`), not here.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -221,4 +242,41 @@ public function test_member_capacity_excludes_phantom_owner_when_ownerless() {
 			'An ownerless group should not add a phantom owner to capacity (10, not 11), keeping numerator and denominator consistent.'
 		);
 	}
+
+	/**
+	 * A pending (non-expired) invite reserves a spot, so a direct member add that would push
+	 * members + pending invites past the limit is rejected -- matching the invite path's own check.
+	 */
+	public function test_update_members_counts_pending_invites_toward_limit() {
```
chickenn00dle, line 291: **Severity:** suggestion

The two new tests cover invite-counting and invitee-exclusion well. Two other behavioral changes in the guard refactor are untested and easy to regress: (a) the limit check is now skipped for removal-only calls — fixing a latent bug where a removal from an over-capacity group would spuriously `409` after already persisting the removal; and (b) expired invites are excluded from the count via `get_invites( $sub, false )`. A test that a removal-only call succeeds on an over-limit group, and one that an expired invite doesn't consume a spot, would lock both in.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -191,23 +214,25 @@ import './admin.scss';
...
-				const $membersList = $( '.newspack-group-subscription__members-list' );
-				const $membersCount = $( '#_newspack_group_subscription_member_ids' )
-					.closest( '.newspack-group-subscription__container' )
-					.find( '.newspack-group-subscription__members-count' );
+				const $membersList = $container.find( '.newspack-group-subscription__members-list' );
+				const $membersCount = $container.find( '.newspack-group-subscription__members-count' );
 				$membersList.find( `li[data-email="${ data.email }"]` ).remove();
 				$membersList.append(
-					`<li data-email="${ data.email }"><span class="newspack-group-subscription__pending-invite"></span> <span class="newspack-group-subscription__pending-invite-label"></span><a href="#" class="newspack-group-subscription__cancel-invite">&#215; <span class="screen-reader-text">Delete</span></a></li>`
+					`<li data-email="${ data.email }" data-consumes-spot="1"><span class="newspack-group-subscription__pending-invite"></span> <span class="newspack-group-subscription__pending-invite-label"></span><a href="#" class="newspack-group-subscription__cancel-invite">&#215; <span class="screen-reader-text">Delete</span></a></li>`
```
chickenn00dle: **Severity:** nit

The invite-append template hardcodes `<span class="screen-reader-text">Delete</span>` before overwriting it with the localized `cancel_label` ("Cancel") a few lines down. It's never user-visible, but the PHP renders this same control as "Cancel" — worth aligning the placeholder to avoid confusion.

chickenn00dle (commented):
Nicely scoped fix. The UI hiding is correctly treated as defense-in-depth: the server-side guard in `Group_Subscription::update_members()` still enforces the limit, and this PR strengthens it by counting pending non-expired invites and excluding the invitee's own invite. The `>=` used to hide the form pairs correctly with the server's `>` on the projected total, unlimited (limit `0`) is handled on all three surfaces (PHP metabox, PHP guard, JS), and the new metabox output is properly escaped (`data-*` via `esc_attr`, notice text via `esc_html_e`, localized JS labels inserted via `.text()`).

One caveat: admin-UI / visual behavior was not verified at runtime — the metabox is on a WooCommerce Subscription edit screen that couldn't be loaded in this environment, so the hide/notice behavior should be confirmed on a running site before merge.

A few non-blocking notes below — the main one is a fail-safe double-count edge when an already-invited reader is added directly by ID.

Release risk: LOW.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -191,23 +214,25 @@ import './admin.scss';
...
 				$added.find( '.newspack-group-subscription__pending-invite-label' ).text( newspackGroupSubscriptions.pending_label );
+				$added.find( '.screen-reader-text' ).text( newspackGroupSubscriptions.cancel_label );
 				$membersCount.text( $membersList.find( 'li' ).length );
+				refreshLimitState( $container );
+				// Show the confirmation in the members section, not the add-member form: when this
+				// invite reaches the limit the form is hidden, which would otherwise swallow its own
+				// "invitation sent" message.
+				const $members = $container.find( '.newspack-group-subscription__members' );
+				$members.append( `<mark class="success"><span class="dashicons dashicons-yes-alt"></span><span class="message"></span></mark>` );
+				$members.find( 'mark.success .message' ).text( newspackGroupSubscriptions.success_message );
```
chickenn00dle, line 275: **Severity:** nit

The invite success `mark.success` appended to `.newspack-group-subscription__members` is only cleared at the start of the next invite; after an add/remove/cancel it lingers. Harmless, but a stale "invitation sent" notice can read oddly following an unrelated action.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Took the real fix: the direct-add loop now cancels any matching pending invite (via `Group_Subscription_Invite::cancel_invite`, guarded by a pre-collected list of pending emails so we don't save when there's no match). That eliminates the double-count and the premature at-limit/409 you flagged, and it mirrors the invite-acceptance path. With the cancellation now living in this method, the "it is cancelled below" note in the limit-check comment is accurate. Added a test (`test_update_members_direct_add_cancels_matching_invite`) locking it in.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Added both tests: `test_update_members_removal_only_succeeds_on_over_limit_group` (a removal-only call succeeds on a group seeded over its lowered limit and reports the removal) and `test_update_members_expired_invite_does_not_consume_a_spot` (an expired invite doesn't reserve a spot, so the direct add succeeds). Extended the `add_invite` helper with an `$expired` flag for the latter.

plugins/newspack-plugin/src/group-subscription/admin.js
adekbadek: Aligned the placeholder to "Cancel" to match the PHP-rendered control and the localized `cancel_label` that overwrites it a few lines down.

plugins/newspack-plugin/src/group-subscription/admin.js
adekbadek: Fixed: added a `clearInviteSuccess( $container )` helper and call it at the start of the add, remove, and cancel handlers (alongside the existing invite handler), so the "invitation sent" notice is cleared before any unrelated action rather than lingering until the next invite.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -449,8 +451,23 @@ public static function add_group_subscription_options( $subscription ) {
...
+		// is excluded because the limit counts members "in addition to the owner". The count uses
+		// the raw $members set (not the display-filtered $member_users) so it matches the server-side
+		// enforcement (Group_Subscription::update_members and the invite gate) and the reader-facing
+		// My Account view (templates/v1/group-subscription-members.php) exactly, so the admin is
+		// never shown an add form the server would then reject with a 409. A limit of 0 means
+		// unlimited. The admin JS re-evaluates on add/remove/invite/cancel to keep the form and
+		// notice in sync.
+		$member_limit    = (int) $settings['limit'];
+		$pending_invites = Group_Subscription_Invite::get_invites( $subscription, false );
+		$is_at_limit     = $member_limit > 0 && ( count( $members ) + count( $pending_invites ) ) >= $member_limit;
```
chickenn00dle: **Blocker:** At-limit uses the owner-inclusive limit, not the seat limit — off by one, so the reported bug survives — `$member_limit = (int) $settings['limit']` is the **owner-inclusive** capacity (`get_member_capacity()`), but `$members` comes from `get_members()`, which returns member-meta holders and therefore excludes the owner (the owner holds manager meta). Comparing the two is off by exactly the owner's seat.

Every enforcement point uses `get_member_seat_limit()` (`limit` − managers without member meta, i.e. `limit - 1` in the normal owned case) instead: `update_members()` at `class-group-subscription.php:630`, `generate_invite()` at `class-group-subscription-invite.php:429`, and the reader-facing My Account template at `templates/v1/group-subscription-members.php:35`, which computes the identical expression from the seat limit and says so in a comment.

**Verified at runtime** on a group with the minimum real limit of 2 (`normalize_limit()` floors positive limits to 2) and one member: `get_member_seat_limit()` returns 1, `update_members()` rejects the next add with a 409 — while this expression evaluates `1 + 0 >= 2 → false`, leaving `is-at-limit` off and the add form rendered. That is precisely the "warning only after you try" behaviour the ticket asks us to remove. The same value is emitted as `data-member-limit` at line 470, so `refreshLimitState()` in `admin.js:67` inherits the off-by-one on live updates too.

The comment at lines 454-461 asserts this matches the My Account view exactly — it doesn't, and should be corrected alongside the fix.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -449,8 +451,23 @@ public static function add_group_subscription_options( $subscription ) {
...
+		// Members in the raw set but not rendered as spot-marked rows (a manager who also carries
+		// member meta, or a non-reader member) still consume a spot; the JS adds this offset to
+		// its rendered-row tally so its live count matches this server-side one.
+		$spots_offset = count( $members ) - count( $member_users );
```
chickenn00dle: **Blocker:** Undefined $member_users fatals every subscription edit screen (HTTP 500) — `count( $member_users )` references a variable that is never assigned in this scope — the rows array built at line 441 is `$member_rows` (used at 537 and 556). `origin/main` contains only `$member_rows`; this branch was written against an older revision that used `$member_users`, and the Jul 19 merge auto-merged the file without flagging a conflict.

On PHP 8 `count( null )` throws `TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given`, which is uncaught inside the `add_group_subscription_options()` metabox callback. **Verified at runtime:** the WooCommerce Subscriptions edit screen returns HTTP 500 with WordPress's "critical error" page, and `debug.log` records the fatal at this line with a stack trace through `Automattic\WooCommerce\Internal\Admin\Orders\Edit->render_meta_boxes()`. It reproduces on plain, non-group subscriptions as well (three checked, all 500), so the blast radius is every subscription edit screen on any site with `NEWSPACK_CONTENT_GATES` enabled.

PHPCS can't see undefined variables and no test invokes the metabox callback, which is why CI is green. The comment at line 456 carries the same stale name and should be updated with the fix.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -516,11 +516,17 @@ public static function accept_invite( $subscription, $key, $email ) {
 		}
 		$invite = self::get_invite_by_key( $subscription, $key );
 		if ( ! $invite || $invite['email'] !== $email ) {
-			// No need to display an error if the user is already a member: just give a success message.
+			// No need to display an error if the invite is already fulfilled: just give a success
+			// message. This covers a direct member add cancelling the invite before it is accepted.
```
chickenn00dle, line 522: **Suggestion:** New accept_invite() clause looks unreachable for shipped callers, and is untested — Both production callers constrain `$email` to the acting user: `process_invite_request()` case 1 rejects unless `$current_user->user_email === $email` before calling, and case 3 creates the account for `$email` and calls `set_current_reader()` first (behind an `is_valid_invite()` gate, so the `! $invite` branch isn't taken there anyway). In both, `$invited_user_id === $current_user_id`, so the new `|| ( $invited_user_id && user_is_member( $invited_user_id, … ) )` is already subsumed by the existing `user_is_member( $current_user_id, … )` check on the line above — including the exact scenario the comment describes.

I could not construct a reachable path where it changes the outcome, so I'm flagging this as uncertain rather than asserting it's dead code. **What would settle it:** a caller-level test demonstrating a case where the current user differs from the invitee and the clause changes the result. If there isn't one, it's worth removing — `$email` originates from `$_GET`, and this clause turns `accept_invite()` into a "does this address belong to this group?" oracle for any future caller that doesn't first bind `$email` to the session.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -566,11 +583,14 @@ public static function add_group_subscription_options( $subscription ) {
 					?>
 				</ul>
 			</div>
+			<div class="newspack-group-subscription__limit-notice notice notice-warning inline" role="status">
```
chickenn00dle: **Suggestion:** role="status" toggled via display:none won't announce — The notice is rendered up-front with static text and revealed only by `.is-at-limit` flipping `display: none` to `block` (`admin.scss:22-29`). An element with `display: none` isn't in the accessibility tree, so it isn't a live region while hidden — and when it becomes visible its *content* hasn't changed, which is an unreliable announcement trigger across NVDA/JAWS/VoiceOver. The net effect is that on exactly the JS-driven transitions the live region was added for (`refreshLimitState()` after add/remove/invite/cancel), a screen reader user gets nothing.

Keep the container in the accessibility tree — an empty `<div role="status">` that is never `display: none` — and have the JS write and clear the sentence as the state flips. The `notice notice-warning inline` classes themselves are the right convention for a metabox notice (`inline` is what stops core from hoisting it under the page `<h1>`).

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
```diff
@@ -566,11 +583,14 @@ public static function add_group_subscription_options( $subscription ) {
...
+				<p><?php \esc_html_e( 'This group subscription has reached its member limit. Remove a member, or raise the member limit above and save, to add more.', 'newspack-plugin' ); ?></p>
```
chickenn00dle: **Suggestion:** Notice hardcodes "group subscription" instead of the publisher-configurable label — Every other user-facing string in this feature interpolates `Group_Subscription::get_label_lower( 'singular' )` so a publisher who renames "Group" to "Team" or "Household" gets consistent copy — see `class-group-subscription-invite.php:393-397` and `:510-514`, `class-group-subscription-api.php:283-287`, and the equivalent My Account limit notice at `templates/v1/group-subscription-members.php:315-318`. This new admin notice bakes the label in. Text domain is correct on all the new strings, and the `remove_label`/`cancel_label` additions are wired up properly.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -624,12 +657,17 @@ public static function update_members( $subscription, $members_to_add, $members_
...
 			if ( \add_user_meta( $member_id, self::GROUP_SUBSCRIPTION_USER_META_KEY, $subscription->get_id() ) ) {
 				\update_user_meta( $member_id, self::get_member_joined_meta_key( $subscription->get_id() ), time() );
+				$member_email                = \get_userdata( $member_id )->user_email;
 				$members_added[ $member_id ] = [
-					'email' => \get_userdata( $member_id )->user_email,
+					'email' => $member_email,
 					'url'   => \get_edit_user_link( $member_id ),
 				];
+				if ( in_array( strtolower( $member_email ), $pending_invite_emails, true ) ) {
+					Group_Subscription_Invite::cancel_invite( $subscription, $member_email );
```
chickenn00dle: **Suggestion:** Case-mismatched invite is excluded from the limit tally but never cancelled — The limit check filters pending invites case-insensitively (`strtolower()` on both sides at lines 621 and 627) and the add loop decides to cancel on the same case-insensitive comparison at line 665 — but `cancel_invite()` matches with a strict, case-sensitive `$invite['email'] === $email` (`class-group-subscription-invite.php:927`). `sanitize_email()` preserves case and `wp_insert_user()` does not lowercase `user_email`, so the two can legitimately differ.

**Verified at runtime with a control:** an invite stored as `pr604.mixed@example.com` for a user whose `user_email` is `PR604.Mixed@Example.com` survives the add (1 pending invite before, 1 after), leaving the reader a member *and* a phantom seat still consumed; the identical flow with matching casing cancels correctly (1 → 0). The group then sits one seat over its limit with a stale invite still rendered — the precise leak this change exists to close.

Either make `cancel_invite()` compare case-insensitively, or have the add loop pass the stored invite email rather than the member's.

**Consider** two smaller things at this call site while you're in it: `cancel_invite()` ends in `update_meta_data()` + `save()`, so a batch add of N invited users performs N full subscription writes and fires the `woocommerce_update_subscription` cascade N times inside what is otherwise a pure user-meta loop — collecting the matches and doing a single write after the loop would be cheaper and keeps the persistence boundary atomic. And its `WP_Error` return is discarded here; since `cancel_invite()` guards on `is_group_subscription()`, which is documented to return `false` in some My Account contexts (`class-group-subscription.php:150-155`), a silent failure would leave the stale invite in place. Surfacing or logging it would make that visible.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -88,20 +118,20 @@ import './admin.scss';
...
 					);
 					const $added = $membersList.find( 'li' ).last();
 					$added
 						.find( '.newspack-group-subscription__member-user-link' )
 						.text( data.members_added[ memberToAdd ].email )
 						.attr( 'href', data.members_added[ memberToAdd ].url );
 					$added.find( ' .newspack-group-subscription__remove-member' ).data( 'user-id', memberToAdd );
+					$added.find( '.screen-reader-text' ).text( newspackGroupSubscriptions.remove_label );
 					$membersCount.text( $membersList.find( 'li' ).length );
+					refreshLimitState( $container );
```
chickenn00dle, line 162: **Suggestion:** Direct add leaves the now-cancelled invite row in the list — `update_members()` now cancels a matching pending invite server-side, but the `members_added` response says nothing about it and `addMember()` only appends the new member row. The stale `li[data-email=…][data-consumes-spot="1"]` stays in the DOM, so `$membersCount` at line 133 is one too high and `refreshLimitState()` counts a spot that no longer exists — the form can be hidden while a seat is genuinely free, until the page is reloaded.

This is reachable: `api_search_users` excludes existing members and the owner (`class-group-subscription-api.php:196-197`) but not invited addresses, so an admin can invite an existing reader and then add that same reader from the select. Simplest fix is to return the cancelled email(s) from `update_members()` and drop the matching row, or have the JS remove any `li[data-email]` matching the added member's email.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -600,15 +600,48 @@ public static function update_members( $subscription, $members_to_add, $members_
...
+					$adding_emails[] = strtolower( $add_user->user_email );
+				}
+			}
+			$pending_invites = array_filter(
+				Group_Subscription_Invite::get_invites( $subscription, false ),
+				function ( $invite ) use ( $adding_emails ) {
+					return empty( $invite['email'] ) || ! in_array( strtolower( $invite['email'] ), $adding_emails, true );
+				}
+			);
+			if ( count( $existing_members ) + count( $pending_invites ) + count( $members_to_add ) > $seat_limit ) {
```
chickenn00dle, line 650: **Nit:** TOCTOU between the count and the add — There's no lock between reading `get_members()`/`get_invites()` and writing the member meta, so two concurrent adds (two admins, or an admin add racing an invite acceptance) can both pass a check that leaves exactly one seat and both land. This is pre-existing behaviour the PR inherits rather than introduces, and the exposure is small — admin-only, low concurrency — so I wouldn't block on it, but the new invite-counting widens the window slightly and it's worth a comment acknowledging the assumption.

plugins/newspack-plugin/src/group-subscription/admin.js
```diff
@@ -191,23 +223,25 @@ import './admin.scss';
...
-					.parent()
-					.append( `<mark class="success"><span class="dashicons dashicons-yes-alt"></span><span class="message"></span></mark>` );
-				$this.parent().find( '.message' ).text( newspackGroupSubscriptions.success_message );
-				const $membersList = $( '.newspack-group-subscription__members-list' );
-				const $membersCount = $( '#_newspack_group_subscription_member_ids' )
-					.closest( '.newspack-group-subscription__container' )
-					.find( '.newspack-group-subscription__members-count' );
+				const $membersList = $container.find( '.newspack-group-subscription__members-list' );
+				const $membersCount = $container.find( '.newspack-group-subscription__members-count' );
 				$membersList.find( `li[data-email="${ data.email }"]` ).remove();
```
chickenn00dle: **Nit:** data.email interpolated into an HTML string — `data.email` is interpolated into both a selector and an HTML string handed to `.append()` here and on the member row at line 124. The REST layer applies `sanitize_email()` (`class-group-subscription-api.php:100-104`), which strips quotes and angle brackets, so I could not construct a breakout and don't believe this is exploitable today. It's still the only remaining raw interpolation in a file this PR otherwise converted to `.text()`-based injection (nice fix at lines 43-44), and the row is already built-then-populated with `.text()`/`.attr()` on the following lines — so `.attr( 'data-email', data.email )` would make the safety local rather than dependent on a sanitizer three layers away. While nearby: the `'Failed to cancel invite'` fallback at line 281 is an untranslated user-facing string.

plugins/newspack-plugin/src/group-subscription/admin.scss
```diff
@@ -13,6 +13,20 @@
 			display: block;
 		}
 	}
+	// The limit warning is hidden until the group is enabled and at capacity; when it shows,
+	// the add-member form is hidden so an admin can't attempt an add the server would reject.
+	.newspack-group-subscription__limit-notice {
+		display: none;
+		margin: 0 0 16px;
+	}
+	&.enabled .newspack-group-subscription__container.is-at-limit {
```
chickenn00dle: **Nit:** Hiding the add form purely in CSS, and the notice is a bare flex item — Both `.is-at-limit` rules are nested under `&.enabled`, which is applied only by JS. If the JS fails the notice never appears (the form is separately hidden by `.show_if_newspack_group_subscription_enabled`, so that direction fails closed); if the stylesheet fails, the form and the notice render together. Neither is a security problem — `update_members()` remains the authority — but since `$is_at_limit` is already computed server-side, wrapping the `__add-member` block in `if ( ! $is_at_limit )` would make correctness independent of the asset pipeline and leave CSS handling only the live transitions.

Separately, `.newspack-group-subscription__container` becomes `display: flex; flex-wrap: wrap` above 768px and the notice is a direct child without a `flex-basis`, unlike its siblings — it will render as a narrow column wedged beside the member list rather than spanning the metabox.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -223,6 +245,64 @@ public function test_member_capacity_is_the_limit_when_ownerless() {
 		);
 	}
 
+	/**
+	 * A pending (non-expired) invite reserves a spot, so a direct member add that would push
+	 * members + pending invites past the limit is rejected -- matching the invite path's own check.
+	 */
+	public function test_update_members_counts_pending_invites_toward_limit() {
```
chickenn00dle, line 291: **Nit:** No coverage for the metabox render — the one path both blockers live on — The new tests cover `update_members()` thoroughly (pending invites, no double-count, invite cancellation, removal-only, expired invites). Nothing calls `Group_Subscription_Settings::add_group_subscription_options()`, which is where both blockers above live — a smoke test that renders the metabox for a full group and asserts the `is-at-limit` class and the `data-member-limit` value would have caught each of them, and is cheap given the `create_group_subscription()` / `add_invite()` helpers you already have.

Minor fixture note: `add_invite()` calls `update_meta_data()` without `save()`, so it only works because every assertion path happens to pass the same in-memory `WC_Subscription` instance through `sanitize_subscription()`. A future test that re-fetches by ID would silently see no invites — adding `$subscription->save()` makes the fixture honest.

chickenn00dle (changes requested):
Thanks for the thorough server-side work here — the invite-aware limit check, the cancel-on-add, and the add-only gating are well constructed and well covered by the new tests.

I spun this up in an isolated env (WooCommerce + Subscriptions, HPOS, `NEWSPACK_CONTENT_GATES` on, branch assets freshly built) and hit two blockers, both reproduced at runtime and both introduced by this branch.

**1. The metabox fatals on every subscription edit screen.** `class-group-subscription-settings.php:468` reads `$member_users`, which doesn't exist in that scope — `main` renamed it to `$member_rows`, and the Jul 19 merge auto-merged the file without flagging a conflict. On PHP 8 that's a `TypeError`, so the edit screen returns HTTP 500 with WordPress's "critical error" page. I confirmed it on three plain (non-group) subscriptions too, so it isn't limited to group-enabled ones. One-word fix, but nothing in CI exercises the render path.

**2. The at-limit threshold is off by one seat, so the reported bug survives at the boundary.** The metabox compares against the owner-inclusive configured limit while `get_members()` excludes the owner; every enforcement point uses `get_member_seat_limit()` instead. On the minimum real limit of 2 (owner + 1 member), `update_members()` returns a 409 while the metabox computes `1 + 0 >= 2 → false` and still renders the add form — exactly the 409-after-the-fact this PR set out to remove. `data-member-limit` carries the same number, so the live JS updates inherit it.

One more worth fixing before merge: a mixed-case invite is excluded from the limit tally case-insensitively but `cancel_invite()` matches with a strict `===`, so it survives the add and keeps consuming a seat. I reproduced that with a control (matching casing cancels correctly, mismatched doesn't).

The rest are suggestions and nits inline. A metabox render smoke test asserting the `is-at-limit` class and `data-member-limit` value would have caught both blockers and is cheap given the fixtures you already have.

Small housekeeping: the description still lists the un-cancelled invite as a known follow-up, but a later commit implemented that cancellation — worth updating, since it becomes the squash commit message.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
adekbadek: Fixed: the variable is `$member_rows`. Confirmed at runtime that the subscription edit screen renders again (HTTP 200, no fatal) and the stale name in the comment above is gone.

Added the render-path coverage that was missing: `test_metabox_gates_on_the_member_seat_limit` renders `add_group_subscription_options()` and asserts on the markup, so a fatal in this callback now fails CI rather than every edit screen. It needed a `woocommerce_wp_text_input()` stand-in in `tests/mocks/wc-mocks.php`.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
adekbadek: Fixed: it now uses `Group_Subscription::get_member_seat_limit()`, so the metabox gates on the same threshold as `update_members()`, `generate_invite()` and the My Account view. `data-member-limit` carries the seat limit too, so `refreshLimitState()` inherits the corrected value.

Two details worth flagging. Unlimited is now signalled by an empty `data-member-limit` rather than `0`, because `get_member_seat_limit()` returns `null` for unlimited and a seat limit of `0` is a legitimate at-limit state that a `> 0` guard would have read as unlimited (the PHP uses `null !== $seat_limit` for the same reason; the My Account template's `> 0` still has that edge, which `normalize_limit()`'s floor of 2 keeps unreachable in practice). And the stale comment is rewritten to describe what the expression actually does.

Verified at runtime on a limit-2 group with one member: seat limit 1, metabox renders `is-at-limit` and hides the add form, matching the 409 `update_members()` would return. The new `test_metabox_gates_on_the_member_seat_limit` data provider pins the boundary (empty and full two-seat, roomy five-seat, unlimited).

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Removed. Walking both callers confirms your reading: case 1 rejects unless `$current_user->user_email === $email`, and case 3 calls `set_current_reader()` for `$email` before accepting, so the invitee is always the current user and the clause is subsumed by the `user_is_member( $current_user_id, … )` check above it. Rather than keep an oracle over a `$_GET` address for a path nothing reaches, the comment now states the invariant the callers guarantee, so a future caller that skips the binding is a visible violation.

`test_email_invite_acceptance_succeeds_for_existing_member` was the one thing exercising it, and only because it accepted as nobody. It now sets the current user to the invitee first, which is what both callers do.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
adekbadek: Done, as suggested. `.newspack-group-subscription__limit-notice` is now an empty `role="status"` wrapper that is never `display: none` while the group is enabled; the `notice notice-warning inline` box lives inside it and is written and cleared as the state flips -- server-side on render, by `refreshLimitState()` on live transitions. So the announcement trigger is a content change in a region that was already in the accessibility tree.

Verified in the browser: removing the last member over the limit empties the wrapper and reveals the form; the reverse writes the notice back.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-settings.php
adekbadek: Fixed: the sentence now interpolates `Group_Subscription::get_label_lower( 'singular' )`. Since the JS rewrites the same notice on live transitions, the string lives in one place -- `Group_Subscription_Settings::get_limit_notice_text()` -- used by both the render and the `wp_localize_script()` payload, so the two can't drift.

Confirmed in the browser that the rendered notice reads "This group has reached its member limit…" off the label rather than a baked-in string.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Comment added at the check, noting that nothing locks between reading members/invites and writing the member meta, that two adds racing for the last seat can therefore both land, and that this is accepted rather than locked against given the admin-only, low-concurrency exposure.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: All three fixed.

Matching is case-insensitive now, in `cancel_invite()` rather than at the call site: the mismatch is a property of stored invites, so every caller benefits (the admin "cancel" button had the same blind spot). `generate_invite()`'s dedupe loop compared the same way, so a re-invite differing only in case would have left two invites for one address -- that is case-insensitive now too.

Batched: the add loop collects matches and `cancel_invites( $subscription, $emails )` performs one write after the loop, so a bulk add is a single `save()` and one `woocommerce_update_subscription` cascade. `cancel_invite()` delegates to it.

The `WP_Error` is no longer discarded -- it is logged via `newspack_log` with the subscription ID and the affected emails, and the emails are dropped from `invites_cancelled` so the admin JS doesn't remove rows that are still there.

Verified in the browser with your scenario: an invite stored as `GS-Member@Test.com` for an account whose email is `gs-member@test.com` is cancelled by the direct add (1 pending before, 0 after, member count 2 not 3). `test_update_members_cancels_a_case_mismatched_invite` covers it.

plugins/newspack-plugin/src/group-subscription/admin.js
adekbadek: Fixed the first way you suggest: `update_members()` now returns `invites_cancelled` (the emails whose pending invites the add fulfilled), and `addMember()` drops the matching `li[data-email]` rows before re-tallying, so the count and `refreshLimitState()` stay honest without a reload. Row lookup is case-insensitive, matching how the server now cancels.

Verified with exactly your path -- invite an existing reader, then add that same reader from the select: the invite row disappears, the count reads 2 rather than 3, and the invite is gone server-side.

plugins/newspack-plugin/src/group-subscription/admin.js
adekbadek: Both fixed. The invite row is built empty and populated with `.attr( 'data-email', … )` / `.text()` like the member row above it, so nothing is interpolated into markup and the escaping is local rather than dependent on `sanitize_email()`. The two selector lookups went the same way -- a shared `findInviteRows()` helper filters on the attribute value instead of interpolating the address into a selector string.

`'Failed to cancel invite'` is now `newspackGroupSubscriptions.cancel_error_message`, localized alongside the other strings.

plugins/newspack-plugin/src/group-subscription/admin.scss
adekbadek: Both fixed, though the first differs slightly from the suggestion. Rendering the add-member block only `if ( ! $is_at_limit )` would make correctness stylesheet-independent but would also strand the live path: once the block is absent, removing a member could not bring the form back without a reload. So the gate is the `hidden` attribute instead -- set server-side from `$is_at_limit`, toggled by `refreshLimitState()`. The initial state is then right with no stylesheet at all (`[hidden]` is UA-level), and the live transitions still work. The two `.is-at-limit` display rules are gone; the reveal rule is now `.show_if_newspack_group_subscription_enabled:not([hidden])` so it doesn't override the attribute. The `is-at-limit` class survives as a state hook only, which is what the new render test asserts on.

The notice gets `flex: 0 0 100%` in the ≥768px block so it spans the metabox -- screenshot in the browser check shows it full-width under the settings and member columns.

Verified without a reload: at limit the form is `hidden` and `display: none`; removing the last over-limit member clears the notice and returns it to `display: block`.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Added `test_metabox_gates_on_the_member_seat_limit`: it renders `add_group_subscription_options()` and asserts the `data-member-limit` value, the `is-at-limit` class and the server-side `hidden` on the add form, over a data provider covering the boundary (empty and full two-seat, roomy five-seat, unlimited). It fails on both blockers as written before the fix -- fatal on one, wrong `is-at-limit` on the other. It needed one `woocommerce_wp_text_input()` stand-in in the WC mocks.

Fixture note taken: `add_invite()` now calls `$subscription->save()`, so it no longer depends on the same in-memory instance being threaded through every call.

adekbadek:
Thanks -- both blockers were real and both are fixed, along with the mixed-case cancellation and every suggestion and nit; details are in the individual threads.

Summary of the shape of the fix: the metabox now gates on `get_member_seat_limit()` (unlimited signalled by an empty `data-member-limit`, since `0` is a legitimate at-limit seat count), the undefined variable is `$member_rows`, and `cancel_invite()` matches case-insensitively -- as does `generate_invite()`'s dedupe, which had the same blind spot. The add-member form is gated by the `hidden` attribute rather than a stylesheet rule, so the initial state holds without CSS while live transitions still work, and the limit notice is an always-present `role="status"` region whose content is written and cleared instead of being toggled with `display`. `update_members()` batches the invite cancellation into one write, logs the `WP_Error` it previously discarded, and returns `invites_cancelled` so the JS can drop the stale rows.

Your suggested render smoke test is in: it fails on both blockers as they were written. Full suite green (2097 tests), and I re-ran your runtime scenarios in an isolated env -- edit screen renders, the limit-2 boundary gates correctly, the mixed-case invite is cancelled, and the invite-then-direct-add path leaves the count right without a reload.

adekbadek:
- fix(group-subscription): gate admin metabox on the member-seat limit (NPPD-1483, #604)

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -600,17 +601,57 @@ public static function update_members( $subscription, $members_to_add, $members_
...
```
chickenn00dle, line 650: **Nit:** The pre-add limit check counts `count( $members_to_add )` in full, but the add loop below skips IDs that are already members or aren't readers. A batch containing an already-member or non-reader ID inflates this projection and can return a spurious 409 for an add that would have been a no-op. All shipped callers add a single, genuinely-addable reader (the admin search excludes existing members; `accept_invite`/link-invite add one), so this is only reachable via a hand-forged REST request and yields a wrong status, not an over-fill. Optional: filter `$members_to_add` to genuinely-addable IDs before the count.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -624,15 +665,45 @@ public static function update_members( $subscription, $members_to_add, $members_
...
+				// false in some My Account contexts, which is the likeliest cause.
+				\do_action(
+					'newspack_log',
+					'newspack_group_subscription_invite_cancel_failed',
+					$cancelled->get_error_message(),
+					[
+						'type' => 'error',
+						'data' => [
+							'subscription_id' => $subscription->get_id(),
+							'emails'          => $invites_to_cancel,
```
chickenn00dle: **Nit:** This new `newspack_group_subscription_invite_cancel_failed` log nests the recipient addresses under `data.emails`, whereas the other `newspack_log` calls in this feature pass the recipient via a top-level `user_email` key that the Logger/Manager pipeline may treat specially (e.g. hashing/redaction). It's batched here so it can't reuse the singular key as-is, but worth being aware the addresses land in a plain `data` key — if log-side PII handling keys off `user_email`, these won't get the same treatment.

chickenn00dle (approved):
Re-reviewed after the fixes. Both blockers from last round are resolved and I verified them against a live env: the metabox no longer fatals on any subscription edit screen (it now renders cleanly where `$member_users` previously threw a TypeError), and the at-limit gate is correct at the boundary — a configured limit of 2 (one member seat) with one member now hides the add form and shows the notice, instead of leaving the form open and 409ing on the next add. The mixed-case invite cancellation, the always-present `role="status"` live region, the configurable label, the `invites_cancelled` row drop, and the removal of the unreachable `accept_invite()` clause all check out; the batched case-insensitive `cancel_invites()` cancels a lowercase invite when the mixed-case user is added directly. PHP tests pass (212/505) and ESLint is clean. Nice work — approving with some non-blocking nits.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/604#pullrequestreview-4746978693)

Fixed, in the current head `dd541b66d1`. The render callback now builds and consumes one `$member_rows` array — `class-group-subscription-settings.php:460` builds it, :558 counts it for the header, :577 renders it, and :489 derives `data-spots-offset` from it. `$member_users` no longer appears anywhere in the plugin.

Confirmed not by reading but by rendering: `test_metabox_gates_on_the_member_seat_limit()` calls `add_group_subscription_options()` for real and asserts the markup is non-empty, so the fatal you hit would now fail CI rather than the edit screen.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/604#pullrequestreview-4746978693)

Agreed that the one-word fix on its own was the weak part — the gap was that nothing invoked the render callback. `test_metabox_gates_on_the_member_seat_limit()` (tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php:457) now does: it seeds a real group, calls `Group_Subscription_Settings::add_group_subscription_options()`, and asserts on the captured markup. The first assertion is deliberately just "the callback rendered something", since an undefined variable there is a fatal that takes down the whole subscription edit screen.

```
✔ Metabox gates on the member seat limit with data set "empty two-seat group"
✔ Metabox gates on the member seat limit with data set "full two-seat group"
✔ Metabox gates on the member seat limit with data set "roomy five-seat group"
✔ Metabox gates on the member seat limit with data set "unlimited group"
OK (4 tests, 16 assertions)
```

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/604#pullrequestreview-4746978693)

Both fixed in `dd541b66d1`.

**Off-by-one:** the metabox now gates on `Group_Subscription::get_member_seat_limit()` rather than the raw configured limit (`class-group-subscription-settings.php:483-485`), which is the same expression `update_members()` and `generate_invite()` enforce and the reader-facing My Account view computes. `data-member-limit` carries that same seat limit (:491), so the live JS tally inherits the server threshold instead of the owner-inclusive one. The `full two-seat group` data set in the new render test pins exactly the boundary you named: limit 2, one member, expected at-limit — it fails against the old comparison.

**Mixed-case invite:** `cancel_invites()` now lowercases both sides (`class-group-subscription-invite.php:947` builds lowercased needles, :955 matches lowercased), so a strict `===` can no longer leave a phantom invite consuming a seat after the add. `test_update_members_cancels_a_case_mismatched_invite()` covers it by seeding the invite with `strtoupper()` of the member's account email and asserting no invites survive the add. The admin JS's `findInviteRows()` matches case-insensitively too, so the row disappears without a reload.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/604#pullrequestreview-4746978693)

Both done.

**Smoke test:** `test_metabox_gates_on_the_member_seat_limit()` renders the metabox and asserts the `is-at-limit` class, the `data-member-limit` value and the `hidden` add form, over a data provider covering the empty/full two-seat group, a roomy five-seat group and an unlimited group. It uses the existing fixtures as you suggested, and it does catch both blockers: the `full two-seat group` case fails on the old owner-inclusive comparison, and any case fails outright on the undefined variable.

**Description:** updated. Beyond the follow-up line you flagged, two other claims had gone stale and are now corrected — the at-limit bullet still described the threshold as the raw configured limit, and the test-count figure no longer matched. Since GitHub uses this as the squash body, it now states the member-seat basis and the verified suite result (2097 tests, 6003 assertions).

Full `newspack-plugin` PHPUnit suite is green; no new commit was needed for the code, all of it is already in the pushed head `dd541b66d1`.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Took the real fix rather than leaving the projection optimistic. `$members_to_add` is now narrowed to the IDs that would genuinely become members — readers who do not already hold this subscription's member meta — immediately after the removal loop and before the seat check, so the projection counts only additions that will actually take a seat. That also lets the add loop drop its own reader/duplicate guards, since the filter is the single place that decides addability; the two can no longer drift apart, which is what allowed the spurious 409 in the first place.

Covered by `test_update_members_ignores_non_addable_ids_when_projecting_the_limit`: a group with one free seat accepts a batch of `[existing member, non-reader, new reader]` and reports only the new reader as added. Fixed in 466a1537b.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Good catch — the addresses are out of the payload entirely now. `Logger::newspack_log()` lifts a *scalar* top-level `user_email` out of `$data` and promotes it to its own field (`class-logger.php:109-112`), so a batch genuinely cannot reuse the key, and leaving the addresses under `data.emails` meant they would land as plain payload with none of that handling. Rather than accept that, the log now names the affected people by user ID: `$invites_to_cancel` is keyed by member ID (the addresses are still passed positionally to `cancel_invites()` and returned to the caller via `array_values()`), and the failure log carries `data.member_ids`. That also matches the sibling calls in `class-group-subscription-invite.php`, which pair a scalar `user_email` with `member_id` in `data`.

The IDs resolve to the same readers for anyone triaging, and the stale invites themselves remain on the subscription, so nothing is lost by dropping the addresses. A comment at the call site records the reasoning. Fixed in 466a1537b.

adekbadek:
@chickenn00dle Re-requesting review – since your approval the reader and duplicate-member guards moved ahead of the limit projection, which also shifts which pending invites count toward the tally, and the cancel-failure log now records member IDs instead of addresses:
- fix(group-subscription): count addable IDs only (NPPD-1483, #604)

PR #605, opened 2026-07-13
fix(content-gate): derive new gate priority from its own bucket (NPPD-2035)

adekbadek:
### All Submissions:

`Content_Gate::create_gate()` set a new gate's `gate_priority` to `count( self::get_gates() )`. That's wrong in two ways:

1. **Wrong bucket for premium newsletter gates.** `get_gates()` defaults to the *content*-gate bucket, so a new premium newsletter gate was numbered against however many content gates the site happened to have. Newsletter priorities came out meaningless, and two newsletter gates created on the same site collided with each other.
2. **A count collides after a mid-list delete.** Priorities are positions, not a counter, so once any gate is deleted from the middle of the list the count returns a number already in use and the new gate ties with an existing one.

Gate priority is what orders overlapping gates, so a tie leaves an arbitrary gate deciding what a reader actually sees.

This derives the priority from the highest one in use within the gate's *own* bucket, via a new `get_next_gate_priority( $is_newsletter )` helper. Only affects gates created going forward; existing gates keep their stored priority.

Found while reviewing the duplicate-gate work (NPPD-1693), whose `duplicate_gate()` already derives priority this way — this brings `create_gate()` in line so the two paths agree.

Closes NPPD-2035.

Both bugs are covered by new regression tests (red before this change, green after):

```
cd plugins/newspack-plugin && n test-php --filter test_create_gate_priority_survives_a_mid_list_delete
cd plugins/newspack-plugin && n test-php --filter test_create_newsletter_gate_priority_uses_the_newsletter_bucket
```

Manually, with `NEWSPACK_CONTENT_GATES` defined:

1. Under **Audience > Access control**, create three gates, delete the middle one, then create another. The new gate should land at the end of the list, and no two gates should share a `gate_priority`.
2. Under **Newsletters > Premium**, on a site that already has content gates, create two premium newsletter gates. They should be numbered 0 and 1 within their own list, not tied at the content-gate count.

### Other information:

* [x] Have you successfully run tests with your changes locally? (full `newspack-plugin` PHP suite green)

https://claude.ai/code/session_01GPFS1FtRWDaQi2GJQs5NNy


plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -680,6 +680,25 @@ public static function is_post_restricted( $post_id = null ) {
...
+	 * deciding what a reader sees.
+	 *
+	 * @param bool $is_newsletter Whether the new gate is a premium newsletter gate.
+	 *
+	 * @return int
+	 */
+	public static function get_next_gate_priority( $is_newsletter = false ) {
+		$bucket_gates = self::get_gates( self::GATE_CPT, null, $is_newsletter );
+		return $bucket_gates ? max( wp_list_pluck( $bucket_gates, 'priority' ) ) + 1 : 0;
+	}
```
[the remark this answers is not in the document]
adekbadek: Good catch — made `$post_type` an explicit parameter of `get_next_gate_priority()` (defaulting to `self::GATE_CPT` to keep existing call sites working) and passed it through to `get_gates()`, so priority is now derived from the correct CPT bucket rather than always assuming the content-gate CPT.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -690,14 +709,13 @@ public static function is_post_restricted( $post_id = null ) {
...
-		$args      = [
+		$args = [
 			'post_title'   => $gate['title'] ?? __( 'Untitled Content Gate', 'newspack-plugin' ),
 			'post_type'    => $post_type,
 			'post_status'  => isset( $gate['status'] ) && in_array( $gate['status'], self::get_post_statuses(), true ) ? $gate['status'] : 'publish',
 			'post_content' => '',
 			'meta_input'   => [
-				'gate_priority' => count( $all_gates ),
+				'gate_priority' => self::get_next_gate_priority( $is_newsletter ),
 			],
```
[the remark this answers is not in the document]
adekbadek: Done — `create_gate()` now passes its own `$post_type` through to `get_next_gate_priority()` so non-default gate CPTs (e.g. WC Memberships' `np_memberships_gate`) get priorities from their own bucket.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -680,6 +680,26 @@ public static function is_post_restricted( $post_id = null ) {
...
+	 * deciding what a reader sees.
+	 *
+	 * @param string $post_type     Post type whose bucket the new gate belongs to. Defaults to self::GATE_CPT.
+	 * @param bool   $is_newsletter Whether the new gate is a premium newsletter gate.
+	 *
+	 * @return int
+	 */
+	public static function get_next_gate_priority( $post_type = self::GATE_CPT, $is_newsletter = false ) {
+		$bucket_gates = self::get_gates( $post_type, null, $is_newsletter );
+		return $bucket_gates ? max( wp_list_pluck( $bucket_gates, 'priority' ) ) + 1 : 0;
```
chickenn00dle: **Severity:** nit — The `max + 1` derivation has a check-then-act race: two concurrent `create_gate()` calls can both read the same max and both write `max + 1`, reintroducing the priority tie this fix removes. It's the same class of race the old `count()` had (not a regression), and gate creation is a one-at-a-time admin action, so the practical risk is near-zero — a short code comment acknowledging it is plenty, no locking needed.

chickenn00dle: **Severity:** nit — `get_next_gate_priority()` fully hydrates every gate in the bucket via `get_gates()` (per-gate meta reads + content-rule resolution + a `usort`) only to pluck one `priority` value. This is the same cost the old `count( self::get_gates() )` paid, so it's not a regression — but if you're in here, a scoped query for just the `gate_priority` meta (or a small `$wpdb` `MAX`) would compute the same result more cheaply. Optional; gate counts are small.

chickenn00dle (approved):
Approve — no blockers. This is a correct, well-scoped fix: deriving a new gate's priority from `max + 1` within its own bucket (content / premium-newsletter / memberships CPT) resolves both the mid-list-delete collision and the newsletter gates being numbered against the content-gate count, and it brings `create_gate()` in line with `duplicate_gate()`. The two new regression tests are genuine (each fails under the old `count()` logic), and I confirmed at runtime that a create after a mid-list delete lands at the next free slot and newsletter gates number 0,1 in their own bucket. Only two cosmetic notes below, neither blocking.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Added a comment on `get_next_gate_priority()` noting the check-then-act read of max + 1 isn't atomic, and that it's benign because gate creation is a one-at-a-time admin action, so no lock is warranted.

adekbadek: Done. Replaced the full `get_gates()` hydration with a scoped query that fetches only the single highest-priority gate in the bucket (its ID + `gate_priority` meta), ordered by the numeric priority meta with the same is_newsletter EXISTS/NOT EXISTS filter, then returns that + 1. No more per-gate content-rule/settings reads or usort just to pluck one value. Guard tests (mid-list-delete and newsletter-bucket) still pass.

PR #606, opened 2026-07-13
feat(content-gate): add a duplicate action to the gates list (NPPD-1693)

adekbadek:
### All Submissions:

Publishers currently recreate near-identical gates by hand — AND/OR restriction logic needs more gates than anticipated, and the highest-access products have to be added to every gate. This adds a **Duplicate** action to the three-dots menu of the gates list, on both surfaces that share the row component: **Audience › Access control** and **Newsletters › Premium**.

One click, no naming dialog. The server generates the title (`X copy`, `X copy 2`, …) and the success notice offers an **Edit** action.

**The copy is always created inactive**, regardless of the site's default status for new gates — a copy of a live gate silently going live would change site-wide access behaviour.

https://github.com/user-attachments/assets/5f6d13d1-fe17-4291-b361-4379ea9b748e

Notes on the implementation:

* **Gate settings are copied generically** from post meta, so settings added later are carried over without a list to maintain here.
* **Layouts are deep-copied.** The reader-facing gate content lives in separate `np_gate_layout` posts whose IDs are stored *inside* the gate's `registration` / `custom_access` meta. Sharing those posts between two gates would let `delete_gate_layouts()` destroy the surviving gate's reader-facing content when the other is deleted. The copy gets its own layout posts, carrying the source's customized content **and** its presentation settings (`style`, `visible_paragraphs`, `overlay_position`, …) — those decide how much of a restricted article a reader sees, so dropping them would not just look different, it would reveal a different amount of the gated content.
* **Failure is rolled back.** The source's layout IDs are never persisted on the copy, not even briefly, and a copy whose layout can't be created is discarded rather than left in the list — otherwise deleting that half-built copy would take the *source's* live layouts with it.
* Each wizard only duplicates gates from its own bucket (400 otherwise), so a copy can't be created into a list that couldn't show it.

Closes NPPD-1693.

Requires `NEWSPACK_CONTENT_GATES` defined.

1. Under **Audience › Access control**, create a gate with content rules, registered + paid access and metering. Edit its registered access layout so it's visibly customized (and, in the layout's sidebar, set a non-default presentation — e.g. Overlay style).
2. Open the gate's three-dots menu → **Duplicate**. The copy should appear at the end of the list, badged **Inactive**, titled "… copy", with a summary identical to the source. The notice offers **Edit**.
3. Duplicate again → "… copy 2".
4. From the copy's menu, **Edit registered access layout**. It should carry your customized content and presentation settings — and be a *distinct* post from the source's layout.
5. Permanently **delete the original** gate. The copy's layouts must still exist and still render.
6. Repeat 2–3 on **Newsletters › Premium**. Copies stay in that list and don't appear under Access control.

Unit tests (11 new, covering settings copy, layout deep-copy incl. presentation meta, empty-layout preservation, rollback on failure, title numbering, the newsletter bucket, and the cross-wizard REST guards):

```
cd plugins/newspack-plugin && n test-php --filter test_duplicate
```

### Other information:

* [x] Have you successfully run tests with your changes locally? (`Test_Content_Gates` 81 tests green; full `newspack-plugin` PHP suite green)

**Related:** #605 (NPPD-2035) fixes a pre-existing `create_gate()` priority bug surfaced while reviewing this. It touches `create_gate()` in the same file, so expect a small conflict for whichever merges second.

https://claude.ai/code/session_01GPFS1FtRWDaQi2GJQs5NNy


plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -755,6 +757,198 @@ public static function create_gate( $gate, $post_type = self::GATE_CPT, $is_news
...
+		if ( is_wp_error( $new_gate_id ) ) {
+			return $new_gate_id;
+		}
+
+		$layout_titles = self::get_gate_mode_layout_titles();
+
+		// Copy the settings generically, so gate settings added later are carried over without
+		// a list to maintain here.
+		foreach ( \get_post_meta( $gate_id ) as $key => $values ) {
+			if ( 'gate_priority' === $key || str_starts_with( $key, '_' ) ) {
```
chickenn00dle, line 909: **Severity:** suggestion — The meta-copy loops here (and in `duplicate_gate_layout()`) correctly skip `_`-prefixed/protected keys, which is the right WP convention, but no test pins that behavior. A test seeding a `_private` meta on the source gate/layout and asserting it does *not* land on the copy would lock in the exclusion and guard against a future refactor accidentally copying protected keys like `_edit_lock`. Low priority — the intended meta was confirmed to copy through at runtime.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -755,6 +757,198 @@ public static function create_gate( $gate, $post_type = self::GATE_CPT, $is_news
...
+		}
+
+		$is_newsletter = (bool) get_post_meta( $gate_id, 'is_newsletter', true );
+
+		// Content gates and premium newsletter gates are prioritized in separate buckets, so
+		// the copy goes after the last gate of its own. Derived from the highest priority in
+		// use rather than the gate count, which would collide with an existing gate whenever
+		// one has been deleted from the middle of the list.
+		$bucket_gates = self::get_gates( self::GATE_CPT, null, $is_newsletter );
+		$priority     = $bucket_gates ? max( wp_list_pluck( $bucket_gates, 'priority' ) ) + 1 : 0;
```
chickenn00dle, line 884: **Severity:** nit — Priority (`max existing + 1`) and the unique title are both computed from a `get_gates()` snapshot taken before the insert, with no locking. Two concurrent duplicate requests would compute the same priority and title, colliding on both. In practice this is a single-admin `manage_options` screen so the probability is near zero and the consequence is cosmetic (a tie broken arbitrarily by the sort) — a mental note, not necessarily a fix.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -755,6 +757,198 @@ public static function create_gate( $gate, $post_type = self::GATE_CPT, $is_news
...
+
+		$new_gate_id = \wp_insert_post(
```
chickenn00dle, line 886: **Severity:** nit — When the gate insert itself fails, `duplicate_gate()` returns the raw `wp_insert_post` `WP_Error`, which carries no `['status' => …]`, so the REST layer falls back to a generic 500 rather than a controlled code (the two validation failures do set 400). Defensible — a failed insert is a genuine server error — flagging only for consistency with the validation paths.

chickenn00dle (approved):
Approve — no blockers. This is a clean, well-defended change. The tricky parts are all correct: the copy is forced inactive, layout IDs are stripped before being persisted on the copy (so deleting the copy can't destroy the source's reader-facing layouts), layouts are deep-copied with rollback on failure, and priority/title are derived per bucket. Both `/duplicate` routes gate on `manage_options` and reject the other wizard's bucket; the `id` param is `absint`-sanitized. I verified at runtime against an isolated env DB: every gate meta key is carried to the copy with none dropped, the copy is `draft` with its own max-in-bucket + 1 priority and distinct layout posts, a lower-privilege (editor) and anonymous request both get 403 while admin gets 200, and the source gate is untouched after duplication. New duplicate tests (11) and the full `Test_Content_Gates` class (81) pass. Only a few low-severity, non-blocking notes below.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Added `test_duplicate_gate_skips_protected_meta`: it seeds a `_private` meta on both the source gate and its registration layout, duplicates, and asserts neither carried over — pinning the `_`-prefixed skip on both copy loops so a future refactor can't quietly start copying protected keys like `_edit_lock` through.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Fixed for consistency: the failed-insert branch now sets an explicit `['status' => 500]` on the error (via `add_data`, so the underlying insert error is preserved for a maintainer) rather than leaning on the REST layer's generic 500 fallback. Kept 500 rather than 400 since a failed insert is a genuine server error, not bad input.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Left as-is deliberately. The only real fix here is locking (or a unique DB constraint), which is over-engineering for a single-admin `manage_options` screen where the window is a few milliseconds and the worst case is cosmetic — a priority tie broken arbitrarily by the sort, or two identically-named copies. The title while-loop already de-dupes within a single request; a lock to cover the cross-request race isn't worth the complexity. Flag if you'd rather I add one.

PR #608, opened 2026-07-13
feat(subscribers): group manager roles + peer-removal guard (NPPD-1753)

adekbadek:

First slice of the admin Subscribers work ([NPPD-1753](https://linear.app/a8c/issue/NPPD-1753)): the **group "manager" role backend**, plus the [NPPD-1815](https://linear.app/a8c/issue/NPPD-1815) multi-manager regression folded in. A group has exactly one **owner** (the subscription customer, holds billing, never removable), any number of **managers** (maintenance, no billing), and plain **members**.

https://github.com/user-attachments/assets/4d287e7c-5125-4a68-98ba-8c8c553da758

**Data layer** (`class-group-subscription.php`)
- Per-manager user meta + `get_managers()` reverse lookup (owner + promoted managers), with a per-request static cache busted by the existing `added/updated/deleted_user_meta` invalidation hook.
- `add_manager()` / `remove_manager()` (owner rejected; only existing members promotable; demoting a non-manager errors).
- `get_managed_subscriptions_for_user()` now includes groups a user manages without owning — this is what lights up a promoted manager's My Account access.
- Removing a member also clears their manager meta (no orphaned managers).

**Authorization (the folded-in blocker)**
- New `can_actor_remove_member( $actor, $target, $subscription )` is the single server-side authority for member removal: owner / `manage_woocommerce` may remove anyone but the owner; a **manager may remove plain members only, never a peer manager**; plain members and outsiders remove no one.
- Enforced server-side in **both** the My Account admin-post handler (`handle_remove_member`) **and** the REST endpoint (`api_update_members`) — previously the peer-manager rule was enforced client-side only, so a forged request could bypass it.
- `handle_set_manager_role()` (My Account) gates promote/demote to the owner or a `manage_woocommerce` admin, so managers can't change peer roles.

**UI**
- My Account members table: role-aware kebab (owner-only "Make/Remove manager" form); "View subscription" (billing) hidden from non-owners.
- wp-admin subscription metabox: promoted managers now render as removable **"(manager)"** rows instead of being mislabeled "(owner)" and locked (regression surfaced in review once `get_managers()` began returning managers).

**Seat accounting**
- `get_member_capacity()` no longer double-counts a promoted manager (who keeps a member seat): capacity = `limit + owner`, keeping the "X / Y" denominator in step with the headcount and the add/invite limit check.
- ⚠️ **Open question for @thomasguillot:** this fix keeps the *current* enforcement (a manager consumes a member seat) and only makes the display consistent. If the intent is instead that managers should sit *outside* the member limit (so promoting frees a seat), that's a deliberate change to the `update_members()` enforcement too — flagging rather than deciding it here.

**Tests / infra**
- New `class-group-subscription-managers.php`: promote/demote data layer, cache invalidation, the owner/admin/manager/member/outsider removal matrix, peer-removal refused at both enforcement sites (REST + admin-post), handler role-gating, and seat-count consistency.
- Extended the PHPUnit `filter_input` mock to support `INPUT_POST` (was GET-only) so admin-post handlers are testable.

Ships live for admins as soon as it lands (no feature gate beyond the existing Access Control flag): My Account manager promote/demote goes live.

Part of [NPPD-1753](https://linear.app/a8c/issue/NPPD-1753) · [NPPD-1815](https://linear.app/a8c/issue/NPPD-1815).

1. On a WooCommerce + Subscriptions site with the Access Control feature on (`define( 'NEWSPACK_CONTENT_GATES', true );`) and WC Memberships off, create a group-enabled subscription owned by a reader ("owner") with two member readers.
2. As the owner, visit **My Account → Group**. Confirm the owner row has no actions, member rows show a "Make manager" kebab, and "View subscription" is visible. Promote a member → row flips to "Manager" / "Remove manager".
3. Log in as the promoted manager → **My Account → Group**: the group is reachable, "View subscription" is **gone**, plain members show only "Remove member", and there is **no** role-change control anywhere.
4. In wp-admin, open the subscription's edit screen → **Group members** metabox: the owner is "(owner)" and non-removable; the promoted manager is "(manager)" and removable; the count is consistent.
5. Run the suite: `n test-php --group WooCommerce_Subscriptions_Integration` (89 green) and `n test-php --filter Test_Group_Subscription_Managers` (20 green).

### Other information:


plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-managers.php
```diff
@@ -0,0 +1,589 @@
...
+		$_POST['member_id']        = (string) $member_id;
+		$_POST['role']             = $role;
+		$_POST['_wpnonce']         = wp_create_nonce( Group_Subscription_MyAccount::SET_MANAGER_ROLE_NONCE_ACTION );
+		$_REQUEST                  = $_POST;
+		$_SERVER['REQUEST_METHOD'] = 'POST';
+		// phpcs:enable WordPress.Security.NonceVerification.Missing
+
+		$capture    = function () {
+			throw new \Exception( 'redirect_intercepted' );
+		};
```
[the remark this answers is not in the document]
adekbadek: Good catch — fixed in a85275553: the redirect-filter callbacks now accept the `$location` argument to match the `wp_redirect` hook signature (it's unused, since the tests assert only that a redirect fired).

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-managers.php
```diff
@@ -0,0 +1,589 @@
...
+		$_POST['subscription_id']  = (string) $subscription_id;
+		$_POST['member_id']        = (string) $member_id;
+		$_POST['_wpnonce']         = wp_create_nonce( Group_Subscription_MyAccount::REMOVE_MEMBER_NONCE_ACTION );
+		$_REQUEST                  = $_POST;
+		$_SERVER['REQUEST_METHOD'] = 'POST';
+		// phpcs:enable WordPress.Security.NonceVerification.Missing
+
+		$capture    = function () {
+			throw new \Exception( 'redirect_intercepted' );
+		};
```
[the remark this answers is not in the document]
adekbadek: Good catch — fixed in a85275553: the redirect-filter callbacks now accept the `$location` argument to match the `wp_redirect` hook signature (it's unused, since the tests assert only that a redirect fired).

thomasguillot:
LGTM

> a manager consumes a member seat

↑ this

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -325,7 +546,13 @@ public static function get_member_capacity( $subscription ) {
...
 			return null;
 		}
-		return $limit + count( array_filter( array_map( 'intval', self::get_managers( $subscription ) ) ) );
+		// Managers who hold no member seat (in practice just the owner) occupy a seat
+		// within the limit; promoted managers already count via their member meta.
+		$managers_without_member_seat = array_diff(
+			array_filter( array_map( 'intval', self::get_managers( $subscription ) ) ),
+			array_map( 'intval', self::get_members( $subscription ) )
+		);
+		return max( 0, $limit - count( $managers_without_member_seat ) );
```
dkoo, line 559: **Blocker:** Owner-inclusive limit floor is applied on write only, not on read — The 2-seat minimum floor is applied only when settings are saved (`update_subscription_settings` runs `max( 2, $value )`), but `get_member_seat_limit()` here subtracts the owner's seat from whatever the read paths return — and both `get_subscription_settings` and `get_product_settings` read the raw stored `limit` with no floor, while the product limit field still allows `min=0`. So a group whose product or subscription stores `limit=1` — a valid, common configuration before this change redefined the limit from “members in addition to the owner” to “total including the owner” — computes `get_member_seat_limit()` = 0. Every gate that compares against it then treats 0 as a hard ceiling: the direct-add gate (`update_members`), the send-invite gate, and the invite-link acceptance gate all reject the *first* member on an otherwise-empty group. I reproduced this against a live env: a group stored with `limit=1` returned “Member limit reached” on the first add. Because the failure surfaces as a normal limit message rather than an error, it can go unnoticed while blocking onboarding for every affected group until each product is manually re-saved. Floor the limit on read as well as on write — or migrate existing stored `1`→`2` and set the product field's `min` to 2.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -167,24 +194,164 @@ public static function get_label_lower( $variant = 'singular' ) {
...
+	public static function add_manager( $subscription, $user_id ) {
+		$subscription = WooCommerce_Subscriptions::sanitize_subscription( $subscription );
+		$user_id      = absint( $user_id );
+		if ( ! self::is_group_subscription( $subscription ) ) {
+			return new \WP_Error( 'newspack_group_subscription_add_manager', __( 'Subscription not found.', 'newspack-plugin' ), [ 'status' => 404 ] );
+		}
+		if ( $user_id === (int) $subscription->get_user_id() ) {
+			return new \WP_Error( 'newspack_group_subscription_add_manager', __( 'The owner already manages this subscription.', 'newspack-plugin' ), [ 'status' => 400 ] );
+		}
+		if ( ! self::user_is_member( $user_id, $subscription ) ) {
```
dkoo: **Suggestion:** add_manager() routes membership check through the side-effecting is_group_subscription() — `add_manager()` checks membership via `user_is_member()`, which calls `is_group_subscription()` — the method with a documented My Account side effect (it returns `false` when Woo Memberships is active on an account page), which is exactly why `can_actor_remove_member()` deliberately reads `get_managers()` directly with a comment to that effect. This handler runs on `admin-post.php`, not an account page, so the side effect doesn't fire and there's no live bug. But this is now the one manager operation that still routes through the gate the rest of the PR was written to avoid, which is a latent fragility if the promote path is ever reached from another context. Consider having `add_manager()` verify membership via the canonical data-layer read for consistency with the hardening applied elsewhere.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
```diff
@@ -602,5 +624,57 @@ public static function handle_remove_member() {
...
+				new \WP_Error( 'newspack_group_subscription_role_permission', $error_message ),
+				$redirect_url,
+				'members',
+				$error_message
+			);
+		}
+		self::verify_manageable( $subscription_id, $redirect_url, 'members' );
+
+		$member_id = absint( filter_input( INPUT_POST, 'member_id', FILTER_VALIDATE_INT ) );
+		$role      = 'manager' === filter_input( INPUT_POST, 'role', FILTER_SANITIZE_SPECIAL_CHARS ) ? 'manager' : 'member';
```
dkoo, line 715: **Nit:** Handler input coercion is validated against the namespaced filter_input mock, not production — The admin-post handlers call `filter_input()` unqualified inside the `Newspack` namespace. In production that resolves to PHP's global `filter_input()`, but under PHPUnit the required `mocks/filter-input-mock.php` defines `Newspack\filter_input()`, which PHP prefers over the global — so the handler tests exercise a shim rather than the real `FILTER_VALIDATE_INT` / `FILTER_SANITIZE_SPECIAL_CHARS` coercion that runs live. This is an established pattern here (the invite handlers already relied on it), so it's not a defect, just worth noting so the input-coercion coverage isn't over-read as validating the shipped code path.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/group-subscription-members.php
```diff
@@ -38,6 +38,8 @@
 $invite_link_url      = $invite_link ? Group_Subscription_Invite::get_link_invite_url( $subscription->get_id(), $current_user_id, $invite_link['key'] ) : '';
 $is_at_limit         = $member_limit > 0 && ( count( $members ) + count( $pending_invites ) ) >= $member_limit;
```
dkoo, line 41: **Nit:** My Account invite affordance keys off the raw owner-inclusive limit — `$is_at_limit` compares the raw owner-inclusive `limit` against `count( $members ) + count( $pending_invites )`, where `$members` excludes the owner. The server now gates additions on `get_member_seat_limit()` (`limit - 1`), so this display flag trips one seat *behind* what the server enforces: for a 2-seat group (owner + 1 member) the modal still offers the invite form even though the server will reject it. It's informational only (drives the invite-modal message), but a publisher can be shown an available invite affordance for a seat the server then refuses. Aligning it with `get_member_seat_limit()` keeps the UI honest.

dkoo (changes requested):
Nicely built manager-role backend, and the security core is sound: `can_actor_remove_member()` is a single, context-independent authority correctly enforced at both the REST endpoint and the My Account admin-post handler, every mutating path verifies a nonce and a capability, and the peer-manager removal guard holds under a forged request. I exercised the full removal matrix and the promote/demote guards against a live env and they behave exactly as specified; both named test suites and PHPCS are green.

One thing to resolve before merge: the owner-inclusive limit change floors the limit to a 2-seat minimum only when settings are *saved*, but both settings read paths (and the product limit field, which still allows `min=0`) take the raw stored value. A group whose product or subscription still stores `limit=1` — a valid configuration under the previous “owner + 1” meaning — then computes zero usable member seats, and silently rejects every add, invite, and invite-link acceptance on an otherwise-empty group with a plain “Member limit reached” message. I reproduced this at runtime. It's data-conditional (only bites groups that carry a stored `1`, but that value was both common historically and still enterable today), so please confirm whether any production groups carry it and either floor the limit on read or migrate stored `1`→`2` before this ships.

The two smaller items below are non-blocking.

- `plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php:260` — **Nit:** get_subscription_context() only normalizes the absent subscription_id — `filter_input( INPUT_POST, 'subscription_id', FILTER_VALIDATE_INT ) ?? 0` maps only the *absent* case (`null`) to `0`; a present-but-invalid value yields `false`, which then flows into the downstream `get_group_url()` / `verify_permission()` calls. Both handle it safely (`absint( false )` = 0, `sanitize_subscription( false )` = permission-denied redirect), so there's no reachable defect — but wrapping in `absint(...)` (as already done at line 659 for `member_id`) would make the `int` contract of the returned `$subscription_id` unambiguous.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Fixed in fc07463c2 — floored on read, as you suggested.

`Group_Subscription_Settings::normalize_limit()` (`0` stays unlimited, any positive value floors to 2) is now applied at the end of both `get_product_settings()` and `get_subscription_settings()`, after their filters, so the "0 or >= 2" invariant holds for every consumer regardless of what is stored or what a filter returns. `update_subscription_settings()` calls the same helper instead of its own inline `max( 2, ... )`, so write and read now share one definition.

A group carrying a stored `1` therefore reads as 2 and gets its one usable member seat back with no re-save and no migration.

One deliberate deviation: I left the product field at `min=0` rather than `min=2`. `0` is the "unlimited" value, so `min=2` would make it unreachable from the field. The read floor covers the `1` case anyway, and a typed `1` still floors to 2 on save.

Covered by a regression test (`test_stored_limit_of_one_is_floored_on_read`) that writes the meta directly to mimic a legacy group, and asserts the first add on an empty group is no longer rejected. I confirmed it fails without the read floor ("Failed asserting that 1 is identical to 2") and passes with it. Since your repro was at runtime and the unit tests use the WC mocks, I also verified against real WooCommerce on an env — a real product with stored meta `'1'` returns `limit=2`, while `0` and `10` are untouched.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Fixed in fc07463c2. `add_manager()` now reads membership straight from `get_members()` instead of going through `user_is_member()` → `is_group_subscription()`, with a comment pointing at the same reasoning as `can_actor_remove_member()`: a role decision should not depend on the context it is made from. That leaves no manager operation routing through the gate the rest of the PR was written to avoid, so the latent fragility goes away rather than staying dependent on the promote path never being reached from an account page.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
adekbadek: Acknowledged — no code change. You are right that the handler tests exercise `Newspack\filter_input()` from the mock rather than the real `FILTER_VALIDATE_INT` / `FILTER_SANITIZE_SPECIAL_CHARS` coercion, so that coverage should be read as testing the handler logic given already-coerced input, not the coercion itself. Leaving the established pattern alone here rather than reworking the shim in this PR; the note is useful and worth keeping in mind if the handler input contract ever carries more weight.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/group-subscription-members.php
adekbadek: Fixed in fc07463c2. `$member_limit` now comes from `Group_Subscription::get_member_seat_limit()` instead of the raw owner-inclusive `limit`, so the display flag is computed against the same threshold the server enforces and the affordance no longer runs a seat ahead of it. The unlimited case still behaves — `get_member_seat_limit()` returns `null` there, and the existing `$member_limit > 0` guard treats that as "not at limit".

adekbadek:
Thanks @dkoo — all addressed in fc07463c2, details in the inline replies.

One point had no inline thread to reply on, since it's outside this PR's changed lines: `get_subscription_context()` now uses `absint( filter_input( ... ) )`, matching line 659's `member_id` handling, so a present-but-invalid value lands on the `int` the contract promises rather than relying on downstream tolerance.

dkoo (approved):
Re-reviewed the fix commit (`fc07463c`). The blocker from my prior review is correctly resolved and the three secondary points are addressed as well.

- **Read-path floor (was the blocker):** `normalize_limit()` is idempotent (`absint` is identity on an int; `max(2, n)` and the `0` branch are fixed points) and is now applied after the filter in both `get_product_settings()` and `get_subscription_settings()`, and reused on the write path. A group whose limit was stored as `1` under the older "members besides the owner" meaning now reads as `2`, leaving exactly one member seat via `get_member_seat_limit()` — so its first member is no longer rejected. This also corrects the `X / Y` denominator, since `get_member_capacity()` now returns the floored, owner-inclusive value in step with the owner-inclusive count. The new test writes the raw `1` meta directly and asserts both the floored read and that the first add succeeds — real coverage of the read path.
- **`add_manager()` membership source:** now reads via `get_members()` (data layer) instead of `user_is_member()`, dropping the `is_group_subscription()` My Account side effect from a role decision. Both resolve the same underlying member-meta relationship; `add_manager()` runs only after the group is already validated, so no regression.
- **`get_subscription_context()` coercion:** `absint( filter_input( INPUT_POST, 'subscription_id', FILTER_VALIDATE_INT ) )` correctly folds both absent (null) and present-but-invalid (false) to the `0` the method's contract promises — strictly better than the old `?? 0`, which only caught the absent case.
- **My Account invite affordance:** the template now gates on the owner-excluded `get_member_seat_limit()`, matching the server-side `update_members` gate; `null` (unlimited) is handled safely since `null > 0` is `false`.

No new issues found. Approving.

PR #614, opened 2026-07-13
fix(content-gate): keep institutional-access check URL host-relative

dkoo:
### All Submissions:

Institutional readers reaching a site through a library off-campus proxy (e.g. EZproxy) were rejected at the institutional-access page with "We couldn't verify your location," even though the proxy's IP is whitelisted.

The dedicated `/institutional-access/<slug>/` loading page verifies access with a client-side `fetch()` to the check REST endpoint, and that URL was emitted as an absolute canonical URL. A rewriting reverse proxy like EZproxy leaves the (slash-escaped, JS-embedded) absolute URL unrewritten, so the browser sends the check request **directly to the canonical host, bypassing the proxy** — and the origin then sees the reader's real off-campus IP instead of the whitelisted proxy IP.

This emits the check URL **host-relative** so it resolves against the document origin. When the page is served through the proxy, the fetch stays on the proxy host and is proxied, so the origin sees the proxy's whitelisted IP and access is granted. Direct (on-campus / no-proxy) visitors are unaffected — the relative URL still resolves to the canonical host.

No REST endpoints change; the server-to-server `POST` check used by mobile apps (Pugpig Bolt, NPPD-1395) is untouched.

Closes NPPD-2039.

1. Load `/institutional-access/` (or `/institutional-access/<institution-slug>/`) directly. View source and confirm the inline `fetch(...)` URL is now **host-relative** — `"/wp-json/newspack/v1/institutional-access/check"` — not an absolute `https://<host>/wp-json/...` URL.
2. Confirm normal (non-proxy) verification still works: with an Institution (Audience → Access Control) whitelisting your IP, the page grants access and sets the `wp_nocache_ip` cookie.
3. Put a reverse proxy in front of the site under a **different host** (simulating EZproxy) and load the institutional-access page through it. In the browser Network panel, confirm the `/wp-json/newspack/v1/institutional-access/check` request goes to the **proxy** host (not the canonical host).
4. With the proxy's egress IP whitelisted on an Institution, confirm the proxied check returns `{"valid":true,...}` and sets `wp_nocache_ip` — i.e. the reader is granted access through the proxy.
5. Confirm the server-to-server `POST /wp-json/newspack/v1/institutional-access/check` (body `{"ip":"..."}`) still returns `show_paywall` as before (unchanged).
6. Run the PHPUnit `Access_Rules` group — `test_check_url_is_host_relative` and `test_check_url_includes_institution_id` pass.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
@@ -439,8 +439,13 @@ class IP_Access_Rule {
 	 * Get the URL to redirect to from the dedicated endpoint.
 	 *
 	 * Checks redirect_to param, then Referer header, then falls back to homepage.
+	 * Returned host-relative (see get_check_url()) so the loading page's
+	 * client-side redirect resolves against the document (proxy) origin. Under a
+	 * rewriting proxy (e.g. a library EZproxy) an absolute URL would send the
+	 * just-verified reader back to the canonical host — off the proxy, without
+	 * the proxied IP — and re-lock the content. See NPPD-2039.
 	 *
-	 * @return string The redirect URL.
+	 * @return string Host-relative redirect URL (path and query, without scheme or host).
 	 */
 	private static function get_dedicated_redirect_url() {
 		$home = home_url( '/' );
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
Tests well!

The `$redirect_url`, which is used as a post-success redirect, is also an absolute URL and added to js as `var redirectUrl = <?php echo wp_json_encode( $redirect_url ); ?>;`.

This is likely to also share the same bug in EZproxy and would warrant the relative URL fix, otherwise the reader would face a locked article right after getting authorized due to the unproxied IP. Mind that this is an assumption as I don't know the exact behavior of EZproxy.

dkoo:
Good catch — fixed in 3b560c8, and I was able to confirm your EZproxy assumption directly rather than leave it an assumption.

Applied the same host-relative treatment to `$redirect_url`: `get_dedicated_redirect_url()` now returns via `wp_make_link_relative()`, covering both the JS `redirectUrl` and the "Continue to site" link. Its three unit tests were updated to expect relative; the `Access_Rules` group (66 tests) + PHPCS pass.

Verified in the reverse-proxy env: after a successful check the page now emits `redirectUrl = "/…"` (and `href="/…"`), so `new URL(redirectUrl, location.origin)` resolves to the **proxy** host — the just-authorized reader stays in the proxied session (proxy IP + cookie) instead of being bounced to the canonical host and re-locked. So your assumption held: with the absolute URL they'd have hit a locked article immediately after authorizing.

miguelpeixe (approved):

PR #615, opened 2026-07-14
fix(newsletters): keep shared list on membership switch (NPPM-3000)

dkoo:
### All Submissions:

When a reader switches between two WooCommerce Memberships plans that grant a **shared** newsletter Subscription List (e.g. a monthly and an annual premium newsletter that both apply a common "member" list/tag), the shared list was being lost from the reader's ESP contact (observed on Mailchimp for HKFP).

Root cause: on a switch, activating the new plan adds its lists (diffed against the reader's current lists, so the already-present shared list is skipped) while deactivating the old plan removes **all** of that plan's lists — including the shared one — with no check for whether the reader still qualifies for it via the now-active plan. Depending on the order the two membership hooks fire, the shared list gets added then immediately removed.

This PR makes `remove_user_from_lists()` exclude any list the reader still has access to through **another active membership**, so switching between plans that share a list no longer strips the shared list. The outcome is now independent of the order in which the add/remove membership hooks fire.

Closes NPPM-3000.

#### Automated:
1. Run the newsletters PHP tests: from `plugins/newspack-newsletters`, `n test-php --filter NPPM_3000_Membership_Switch_Ordering_Test`.
   - The new test `test-nppm-3000-membership-switch-ordering.php` sets up two membership plans that share a Subscription List and drives a monthly→annual switch in the reported (add-then-remove) order.
   - It asserts the shared list survives and passes only with this fix. A companion test confirms the remove-then-add ordering already preserved the list.
2. Run the full suite from `plugins/newspack-newsletters`: `n test-php` — should be green (no regressions).

#### Manual (Mailchimp + WC Memberships + WC Subscriptions site):

3. Configure two premium-newsletter membership plans (e.g. monthly and annual) that each tie the reader to a **shared** Subscription List plus one unique list per plan.
4. As a reader, purchase the monthly product; confirm in Mailchimp you receive the shared list/tag and the monthly-only one.
5. Switch the subscription to the annual product.
6. Confirm in Mailchimp the reader now has the annual-only list/tag **and still has the shared list/tag**, and no longer has the monthly-only one.
7. Regression: cancel a lone membership (reader has no other active membership granting the shared list) and confirm the shared list is correctly removed.

### Other information:


dkoo:
Self-review turned up a regression the original fix introduced, so I've pushed three follow-up commits.

**The problem.** The first version excluded still-granted lists from `$lists_to_remove` *before* the deactivation-history snapshot was taken. When a shared list was the only one of a plan's lists the reader was subscribed to, that snapshot got saved empty — and `add_user_to_lists()` treats an empty history as "no restriction," re-adding every plan list on the next reactivation. So a reader who had opted out of a plan's secondary newsletter would be silently re-subscribed to it on a subscription renewal (pause → reactivate). Reproduced against a local env: moving the snapshot back to its old position brings the bug back.

**The fix** (`83f3676`): take the subscribed-lists snapshot from the full plan list first, then apply the still-granted exclusion only to what's actually sent to the ESP. That keeps both the shared-list fix and an accurate deactivation history. Added a regression test (`test_optout_list_survives_membership_pause_reactivate`) that fails on the old ordering and passes now.

Two smaller follow-ups:
- `1326874` — documents the ordering assumption: the exclusion is order-independent when the new plan is active as the old one deactivates (the reported flow); a deactivate-first ordering still leans on the re-add path, which doesn't cover post-checkout signup-modal lists.
- `e66e4fa` — review nits: an `instanceof` guard in the new helper, the test's `set_post_status_raw` moved from a global function to a private method, a mock default so `get_content_restriction_rules()` always returns an array, and a clearer docblock on the remove-before-add control test.

Full newsletters suite green (494 tests).

dkoo:
Passed multiple rounds of agentic review, fixes an active bug in production now. Merging.

PR #626, opened 2026-07-14
feat(subscribers): membership migration CLI + manager backfill (NPPD-1753)

adekbadek:

Second slice of the admin Subscribers work ([NPPD-1753](https://linear.app/a8c/issue/NPPD-1753)): the WooCommerce Memberships → Access Control **migration tooling**, ported from the standalone `migrate-memberships` drop-in into the plugin as proper WP-CLI classes, plus a new **`backfill-team-managers`** command that restores the group-manager designation the earlier migrator dropped.

Stacked on [#608](https://github.com/Automattic/newspack-workspace/pull/608) (PR 1 — the manager backend); the diff base is that branch and will retarget to `main` once it merges.

**New class** `Newspack\CLI\Teams_Migration` (`includes/cli/class-teams-migration.php`), registered in `class-initializer.php`. CLI classes load only under `WP_CLI`, so there is zero web-request cost. Commands:

- `wp newspack migrate-teams` — migrate active team memberships to group subscriptions. Now **manager-aware**: reads each member's Teams role (`_wc_memberships_for_teams_team_{id}_role`) and promotes `manager`-role members to group managers in the same pass. The summary gains a "Number of Managers" column.
- `wp newspack migrate-team-products` — enable group-subscription settings on team-membership products (verbatim port).
- `wp newspack migrate-manual-members` — create free subscriptions for manually-assigned members.
- `wp newspack backfill-team-managers` **(new)** — for sites already migrated (where the old migrator flattened every member to a plain member): resolves each team's group subscription (its linked active subscription, else an active group subscription owned by the team owner) and promotes each `manager`-role member who is already a group member. Dry-run by default; `--live` to apply. Idempotent.

**Data-layer refactor.** Member adds now route through `Group_Subscription::update_members()` (records the joined-at timestamp the raw drop-in omitted, and auto-enables the group) and manager promotion through `add_manager()`, instead of raw `add_user_meta`. Confirmed these fire only the group cache-reset hook — no data events, ESP sync, or emails. The drop-in duplicated its member-add helper three times; the port dedupes it to one. WooCommerce transactional emails are suppressed during runs that create subscriptions (activating created subscriptions still fires the standard subscription-status actions — noted in the class docblock; prefer a low-traffic window on large migrations).

Reader-guard behaviour: `update_members()` adds readers only, so administrators/editors (who already have full access) are skipped and surfaced in the run output rather than silently written.

⚠️ **Flag for @thomasguillot (seat policy):** the migration maps a team's `_seat_count` to the group `limit` 1:1. WooCommerce Teams counts the owner within `_seat_count`, whereas the group `limit` counts members *in addition to* the owner (capacity = `limit + owner`), so the migrated group ends up one seat more generous than the original team. A tighter `seat_count - 1` mapping was rejected because it collides with `0 = unlimited` at a single seat and would diverge from the already-migrated sites. Kept the generous mapping (it never locks out an existing member) and documented it in-code — but the intended seat semantics are your call, same axis as the seat question flagged in #608.

`migrate-membership-gates` (the Memberships-plan → content-gate migration in the drop-in) is intentionally **not** ported here — it splits into a follow-up PR.

1. On a WooCommerce + Subscriptions + WC Memberships for Teams site, confirm the commands register: `wp newspack --help | grep migrate` shows all four, and `wp newspack backfill-team-managers --help` renders.
2. On a site that has already been migrated to group subscriptions, pick a group subscription whose team had a `manager`-role member. Run `wp newspack backfill-team-managers` (dry-run) — the summary lists the manager under "Would Add" and writes nothing. Run `wp newspack backfill-team-managers --live` — the member is promoted; confirm with `Group_Subscription::get_managers()`. Re-run `--live`: "Already Managers" is counted, "Added" is 0 (idempotent).
3. For a team still linked to an active group subscription, `wp newspack migrate-teams --skip-unlinked --dry-run` reports the members it would add and the managers it would promote (new "Number of Managers" column), writing nothing.
4. Run the suite from `plugins/newspack-plugin`: `n test-php --filter Test_Teams_Migration` (11 green) and `n test-php --group WooCommerce_Subscriptions_Integration` (100 green).

A narrated terminal walkthrough of the backfill (dry-run → live → idempotent re-run) and the manager-aware `migrate-teams` dry-run, recorded against a local env, is attached below.

### Other information:

Part of [NPPD-1753](https://linear.app/a8c/issue/NPPD-1753).


thomasguillot:
> the migration maps a team's _seat_count to the group limit 1:1. WooCommerce Teams counts the owner within _seat_count, whereas the group limit counts members in addition to the owner (capacity = limit + owner), so the migrated group ends up one seat more generous than the original team. A tighter seat_count - 1 mapping was rejected because it collides with 0 = unlimited at a single seat and would diverge from the already-migrated sites. Kept the generous mapping (it never locks out an existing member) and documented it in-code — but the intended seat semantics are your call, same axis as the seat question flagged in https://github.com/Automattic/newspack-workspace/pull/608.

Thanks for flagging this. To be clear on intended semantics: the owner should count against the `limit`, not be free on top of it.

So for a group with `limit = 10`:
- 1 owner + 2 managers + 6 members = 9/10 (1 seat free)
- 1 owner + 4 members = 5/10 (5 seats free)

This matches how WooCommerce Teams' `_seat_count` worked, and it's the model I want going forward.

On the `0 = unlimited` collision: rather than changing the unlimited sentinel (which would mean auditing every `limit === 0` check across the codebase, backfilling already-migrated groups, and checking for breaking changes on anything exposing `limit` externally), I'd rather solve it with a floor at migration time:

```
limit = max(seat_count - 1, 1)
```

A solo-owner team (`seat_count = 1`) migrates to `limit = 1` (room for one more person) instead of colliding with `0`/unlimited. This is a much smaller, contained fix — just the migration script, no systemic sentinel change, no backfill, no external breaking-change risk. Worth a code comment explaining why the floor exists so it doesn't get "simplified" back to `seat_count - 1` later.

Separately: a group should never realistically have only 1 seat (owner-only, no room for anyone else) — that's not a meaningful product state. I'd like to enforce a minimum of 2 seats (i.e. `limit >= 1`, since capacity = `limit + 1` for the owner) going forward, so we can't end up with a group that has no room for a single other person.

Can we scope this as: (1) agree `limit` includes the owner, (2) apply the `max(seat_count - 1, 1)` floor in the migration mapping, (3) add minimum-2-seats enforcement going forward? This keeps migrated and native groups consistent without touching the unlimited sentinel.

adekbadek:
@thomasguillot  –  Agreed on the model — owner counts against `limit` (capacity = `limit`), and the floor/min-seats direction is right.

**Mapping.** Whether `_seat_count` includes the owner isn't fixed — WC Teams makes it configurable (`wc_memberships_for_teams_owners_must_take_seat` + a per-purchase "owner takes a seat" toggle). A blanket `seat_count - 1` therefore under-provisions every group — two seats short in the common "owner doesn't take a seat" case, pushing a full team over its new limit. The migration can read it exactly per team (owner is/isn't in the member list):

```
limit = seat_count === 0                          // 0 = unlimited (same sentinel both sides)
      ? 0
      : max( seat_count + ( owner_is_team_member ? 0 : 1 ), 2 );
```

Preserves real capacity, keeps min-2, never strands a member; solo-owner → `limit = 2`. (Your 9/10 counting the 2 managers also settles the open question in #608 — managers stay *inside* the limit, so its enforcement is right as-is.)

**Scope.** I'll put (1) and (3) in #608 — same seat-accounting code it already touches (`get_member_capacity()`, `update_members()`, the limit save), so the model lands in one place. (2), the mapping, stays here — #626 is stacked on #608, so it inherits (1) automatically and lands after. Shout if you'd rather split them out.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1436 @@
...
+				$created_new = true;
+				if ( ! $dry_run ) {
+					$new_sub = self::create_migration_subscription( $owner_id, $migration_product, $billing_period, $billing_interval, $start_date, $end_date, $errors, $team_id );
+					if ( ! $new_sub ) {
+						$summary[] = self::summary_row( $team_id, 'ERROR', 0, 0, $seat_count, true, $errors );
+						continue;
+					}
+					$subscription = $new_sub;
+				}
+			}
```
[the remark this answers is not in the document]
adekbadek: Fixed in 67bdc7f. Added a guard before the create-new branch: when no reusable subscription resolves and `$migration_product` is null (the --skip-unlinked-without-product-id case, where a team whose linked subscription is inactive/missing isn't skipped), the team is now recorded as an error and skipped instead of fataling on `$migration_product->get_id()`. The guard runs in dry-run too, so the preview no longer claims it would create subscriptions it couldn't.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1436 @@
...
+			? array_filter( array_map( 'trim', explode( ',', strtolower( $skip_domains ) ) ) )
+			: [];
+
+		if ( ! function_exists( 'wcs_create_subscription' ) ) {
+			WP_CLI::error( 'WooCommerce Subscriptions is not active. Aborting.' );
+		}
+
+		if ( ! $product_id ) {
+			WP_CLI::error( 'Missing required argument: product-id' );
+		}
```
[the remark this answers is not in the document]
adekbadek: Fixed in 67bdc7f. Reworded to 'Missing required option: --product-id=<id>.' to match the named-option contract (and the equivalent message in migrate_teams).

plugins/newspack-plugin/includes/cli/class-initializer.php
```diff
@@ -92,6 +93,12 @@ public static function register_comands() {
 			);
 		}
 
+		// WooCommerce Memberships → Access Control group subscription migration commands.
+		WP_CLI::add_command( 'newspack migrate-teams', [ 'Newspack\CLI\Teams_Migration', 'migrate_teams' ] );
+		WP_CLI::add_command( 'newspack migrate-team-products', [ 'Newspack\CLI\Teams_Migration', 'migrate_team_products' ] );
+		WP_CLI::add_command( 'newspack migrate-manual-members', [ 'Newspack\CLI\Teams_Migration', 'migrate_manual_members' ] );
+		WP_CLI::add_command( 'newspack backfill-team-managers', [ 'Newspack\CLI\Teams_Migration', 'backfill_team_managers' ] );
```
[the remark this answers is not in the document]
adekbadek: No change needed — this is a false positive for WP-CLI command registration. WP-CLI instantiates the class for a `[Class, method]` array callable (its CommandFactory reflects the method and calls it on an instance), so no static call happens and no deprecation is emitted. Verified empirically on PHP 8.3: `wp newspack migrate-team-products --dry-run` runs cleanly with no 'Non-static method ... cannot be called statically' fatal (which PHP 8 would throw, not merely deprecate, if it were called statically). This also matches every sibling CLI class in the plugin (Co_Authors_Plus, WooCommerce_Subscriptions, Teams_For_Memberships_Diagnostics), which all register non-static methods the same way; making these four static would diverge from that established convention for no runtime benefit.

adekbadek:
Landed, per the split above:

- **#608** — (1) the owner now counts *against* the limit (capacity = `limit`, not `limit + owner`). `get_member_capacity()` returns the limit and the add/invite gates route through a new `get_member_seat_limit()` helper (`limit − owner_seat`). (3) the 2-seat floor is enforced at the limit-save path (`update_subscription_settings()`), and the product option is relabelled "including owner".
- **#626** — (2) the per-team mapping: `limit = seat_count === 0 ? 0 : max( seat_count + ( owner_is_team_member ? 0 : 1 ), 2 )`, with `owner_is_team_member` read from whether the owner holds a `_member_id` seat. Verified against WC Teams source — `Orders.php` adds the owner to `_member_id` only when they take a seat — so the `+1` fires exactly for owner-takes-no-seat teams, and never strands a member.

Full PHPUnit suites green on both (#608 standalone and #626 stacked); PHPCS clean. #626 was rebased onto the updated #608, so its commits were force-updated.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+				'subtotal'   => 0,
+				'total'      => 0,
+			]
+		);
+
+		$subscription->set_billing_period( $billing_period );
+		$subscription->set_billing_interval( $billing_interval );
+		$dates_to_set = [
+			'start'        => $start_date,
+			'next_payment' => gmdate( 'Y-m-d H:i:s', strtotime( "+$billing_interval $billing_period", strtotime( $start_date ) ) ),
```
dkoo: **Severity:** Suggestion

When `--product-id` is supplied, `replace_subscription_product()` runs on every reused active subscription and sets `next_payment` to one billing interval after the team’s original start date. For any team older than a single interval, that stores a past-due `next_payment` on a live, active subscription — the only guard here is `next_payment >= end_date`, with no floor against a past date. Depending on how WooCommerce Subscriptions handles a past `next_payment`, the subscription can be treated as overdue or have a renewal scheduled/processed immediately instead of preserving its billing schedule. Preserve the existing `next_payment` when reusing the subscription, or compute the next *future* interval, before saving.

**Consider:** the same applies if the site runs Newspack Network as a Node or has webhook endpoints configured — a run-scoped guard that also unhooks the `woo_subscription_updated` dispatcher and the network subscription listener (mirroring the email suppression) would keep the backfill from masquerading as real new-subscription activity.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+					'email'      => $owner->user_email,
+					'country'    => \get_user_meta( $owner_id, 'billing_country', true ),
+				],
+				'billing'
+			);
+		}
+
+		$dates_to_set = [
+			'start'        => $start_date,
+			'next_payment' => gmdate( 'Y-m-d H:i:s', strtotime( "+$billing_interval $billing_period", strtotime( $start_date ) ) ),
```
dkoo: **Severity:** Suggestion

`create_migration_subscription()` uses the same `start + interval` formula off the team’s original start date, so a newly-created subscription for an old team also gets a past `next_payment`. Relatedly, `normalise_date()` on `_membership_end_date` does not filter past dates, so an already-expired team is activated with an end date in the past — whereas `migrate-manual-members` guards this (`$has_end_date = ... && strtotime( $end_date ) > time()`). Worth aligning the two creation paths: apply the same future-date guard, or explicitly decide expired teams migrate as expired. As written this surfaces as noisy caught `update_dates` errors on real data.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+			]
+		);
+	}
+
+	/**
+	 * Suppress WooCommerce transactional emails during migration.
+	 *
+	 * @return void
+	 */
+	private static function suppress_woocommerce_emails() {
```
dkoo, line 1447: **Severity:** Suggestion

`suppress_woocommerce_emails()` unhooks four WooCommerce email filters, but `update_status('active')` on every created subscription still fires `woocommerce_subscription_status_updated`/`_changed`. Newspack’s data-events listener turns those into dispatched `woo_subscription_updated` events (one per line item), which reach the Webhooks dispatcher and — on a Network Node — the Hub subscription-sync listener. On a bulk run this emits a burst of external activity for what is really a data backfill. The class docblock already discloses this and recommends a low-traffic window; flagging it so the external-sync dimension is explicit.

**Consider:** if `migrate-manual-members` is run more than once against the same plans, every member gains a duplicate active $0 subscription because creation is unconditional — beyond the dry-run-first guidance, skipping a member who already owns an active migration-created subscription for the same product would make it safe to re-run.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+		}
+
+		// Suppress WooCommerce emails to avoid spamming customers during migration.
+		self::suppress_woocommerce_emails();
+
+		$teams = \get_posts(
+			[
+				'post_type'      => 'wc_memberships_team',
+				'post_status'    => 'publish',
+				'posts_per_page' => -1,
```
dkoo, line 135: **Severity:** Suggestion

Every migration query loads its full result set at once (`posts_per_page => -1`) — teams, products, plans, and most exposed, all `wc_user_membership` posts under a plan — and the per-item loops save subscriptions without clearing the object cache between iterations. The sibling `WooCommerce_Subscriptions` CLI class in the same directory paginates (`get_subscriptions( $page )` in a `while` loop) for exactly this reason. On a large site this risks memory exhaustion. Batched pagination with `\WP_CLI\Utils\wp_clear_object_cache()` (or `stop_the_insanity()`) between batches — at least for the unbounded `migrate_manual_members` membership loop — would match the established pattern. Since migrations run once in a controlled window, this is robustness rather than a defect.

dkoo (commented):
Reviewed the migration CLI end-to-end against an isolated env (WooCommerce + Subscriptions + Memberships for Teams). No blockers — recommending approve with comments.

This is a careful, well-documented port. Member and manager writes correctly route through the real `Group_Subscription` data layer rather than raw meta, idempotency is handled at the data layer (deferred limit write, owner-owned-subscription reuse, membership checks before add/promote), and I verified the manager backfill end-to-end at runtime: dry-run wrote nothing, `--live` promoted the manager, and a live re-run was idempotent ("Already: 1, Added: 0"). Command registration and the manager-aware `migrate-teams` dry-run both ran clean; `Test_Teams_Migration` (12) and the `WooCommerce_Subscriptions_Integration` group (101) are green.

The one thing worth a close look before merge is the renewal-date handling when reusing an existing subscription with `--product-id`: `next_payment` is computed as one interval after the team’s *original* start date, so for teams older than a billing interval it stores a past-due renewal on a live subscription. Two related notes: the same past-date pattern exists in the create path, and the bulk `update_status('active')` still fires subscription-status data events / webhooks / network sync that the email-only suppression does not cover (already disclosed in the class docblock). The rest are memory-footprint and consistency polish.

Note on the three earlier Copilot comments: the non-static-callable one is a verified false positive (all four commands run with no fatal on PHP 8.3 — WP-CLI instantiates the class), and the null-product and error-wording ones were fixed in 67bdc7f. Not re-raising those.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+		$migration_product = $product_id ? \wc_get_product( $product_id ) : null;
+		$billing_period    = 'month';
+		$billing_interval  = 1;
+		if ( $product_id && ! $migration_product ) {
+			WP_CLI::error( sprintf( 'Product ID %d not found. Aborting.', $product_id ) );
+		}
+
+		if ( $migration_product ) {
+			WP_CLI::line( sprintf( 'Using product: "%s" (ID: %d)', $migration_product->get_name(), $migration_product->get_id() ) );
+			$billing_period   = \get_post_meta( $product_id, '_subscription_period', true );
```
dkoo: **Severity:** Suggestion

`$billing_period` / `$billing_interval` come from the product’s `_subscription_period` / `_subscription_period_interval` post meta. For a *variable* subscription product those live on the variation, not the parent, so a `--product-id` pointing at a variable-subscription parent returns empty strings that flow into `wcs_create_subscription()` as `billing_period => ''`. `create_group_subscription()` and `create_individual_subscription()` guard this with `'' !== $meta ? $meta : 'month'`; `migrate_teams` does not. Mirror the same defaulting here.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+				}
+			}
+			if ( $non_reader_skips ) {
+				WP_CLI::warning( sprintf( 'Team %d: %d team member(s) skipped — not readers (e.g. administrators/editors), who already have full access.', $team_id, $non_reader_skips ) );
+			}
+
+			// Set the seat limit now that members are in, using the owner-inclusive
+			// $group_limit derived from the team's seat count above.
+			if ( ! $dry_run ) {
+				$subscription->update_meta_data( '_newspack_group_subscription_limit', $group_limit );
```
dkoo, line 326: **Severity:** Suggestion

For a reused subscription that already carried members, the seat limit is written unconditionally to the team-derived `$group_limit` after members are added. If `$group_limit` is smaller than the resulting membership count, the group is saved in an over-limit state (limit < actual members). This is consistent with treating the team’s seat count as authoritative, but a one-line warning in the summary when `$group_limit < members-in-group` would let an operator notice an unintended shrink. Low priority.

plugins/newspack-plugin/includes/cli/class-initializer.php
```diff
@@ -92,6 +93,12 @@ public static function register_comands() {
 			);
 		}
 
+		// WooCommerce Memberships → Access Control group subscription migration commands.
+		WP_CLI::add_command( 'newspack migrate-teams', [ 'Newspack\CLI\Teams_Migration', 'migrate_teams' ] );
```
dkoo: **Severity:** Nit

The four new `migrate-*` / `backfill-team-managers` commands register unconditionally, whereas the immediately-preceding Teams-for-Memberships diagnostics command is gated behind a `class_exists()` check so it doesn’t surface in `wp help` on sites without the plugin. These are equally Teams-specific (and each `WP_CLI::error()`s early when WC Subscriptions is inactive), so the same guard would keep `wp help` uncluttered on non-Teams sites. Cosmetic.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -0,0 +1,1477 @@
...
+
+				$user = \get_userdata( $user_id );
+				if ( ! $user ) {
+					WP_CLI::warning( sprintf( '  Membership %d: user %d not found — skipping.', $membership_id, $user_id ) );
+					continue;
+				}
+
+				// Skip users whose email domain is in the skip-domains list.
+				if ( ! empty( $skip_domains ) ) {
+					$user_domain = strtolower( substr( strrchr( $user->user_email, '@' ), 1 ) );
```
dkoo: **Severity:** Nit

`substr( strrchr( $user->user_email, '@' ), 1 )` returns `''` with a PHP 8 deprecation for `substr(false, …)` if an email lacks `@`. Defensive-only (emails are validated at registration), but `( strrchr( $user->user_email, '@' ) ?: '' )` would silence the edge case.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-teams-migration.php
```diff
@@ -0,0 +1,397 @@
...
+use Newspack\CLI\Teams_Migration;
+use Newspack\Group_Subscription;
+use Newspack\Group_Subscription_Settings;
+
+/**
+ * Test the migration data-layer helpers and the manager backfill.
+ *
+ * @group WooCommerce_Subscriptions_Integration
+ */
+class Test_Teams_Migration extends WP_UnitTestCase {
```
dkoo, line 25: **Severity:** Nit

The suite thoroughly covers the data-layer helpers and the backfill resolution/idempotency, and reasonably defers subscription-creation to live testing (the WC mocks don’t model line items). But the reuse branch of `find_group_subscription_for_owner()` (ownership + active + `is_group_subscription` filtering) and the multi-team merge-warning path are pure logic the mocks *can* exercise and are currently untested. A couple of tests there would lock in the idempotency guarantee the `migrate-teams` docblock promises. Optional given the acknowledged live-test plan.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed. Both creation paths (`create_migration_subscription` and `replace_subscription_product`) now build their dates via a shared `build_subscription_dates()` → `next_future_payment_date()`, which rolls next_payment forward from the team's start date to the first future occurrence instead of blindly `start + one interval`, so a live subscription can never be saved with a past-due next_payment. On the external-sync half: added `suppress_data_events()` (a run-scoped `newspack_data_events_dispatch_body` filter returning WP_Error, the documented cancel path), hooked alongside the email suppression in migrate-teams and migrate-manual-members. That cancels the dispatch at source, so the `woo_subscription_updated` events never reach the Webhooks dispatcher or the Network Node's Hub sync listener.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed and aligned both creation paths. `create_migration_subscription` now goes through the same `build_subscription_dates()` helper as the reuse path: next_payment rolls forward to the first future occurrence, and the end date is set only when `strtotime( $end_date ) > time()` — mirroring the `$has_end_date` guard in migrate-manual-members. So an already-expired team migrates as an ongoing subscription rather than triggering a caught `update_dates` error on a past end date.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Addressed both points. External sync: added `suppress_data_events()` (run-scoped `newspack_data_events_dispatch_body` → WP_Error) hooked next to the email suppression in migrate-manual-members, so the activation events are cancelled at source rather than only disclosed in the docblock. Re-run safety: individual mode now skips a member who already owns an active migration-created subscription (created_via 'manual migration') for the same product, via a new `member_has_migration_subscription()` check that runs before both the dry-run and live branches — so a re-run no longer stacks duplicate $0 subscriptions.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Added `\WP_CLI\Utils\wp_clear_object_cache()` to bound memory across all three per-item loops — the teams loop, the products loop, and (as flagged) the unbounded membership loop in migrate_manual_members, cleared at the top of each membership iteration so the held $group_subscription/$product objects are unaffected. I kept the `fields => 'ids'` `get_posts` calls rather than converting to full `get_subscriptions( $page )` pagination: the ID arrays are memory-light, and the real growth was the per-save object cache, which the flush now caps. That keeps the diff minimal for a one-off controlled run while matching the sibling command's intent.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed. After writing the seat limit, migrate-teams now compares it against `Group_Subscription::get_member_count()` (owner-inclusive) and emits a `WP_CLI::warning` when a reused group ends up over its limit, so an operator sees the unintended shrink.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed. `migrate_teams` now defaults the billing period/interval the same way `create_group_subscription()`/`create_individual_subscription()` do — `'' !== $meta ? $meta : 'month'` / `1` — so a --product-id pointing at a variable-subscription parent (whose `_subscription_period` meta lives on the variation and comes back empty) no longer passes an empty `billing_period` into `wcs_create_subscription()`.

plugins/newspack-plugin/includes/cli/class-initializer.php
adekbadek: Gated all four commands behind `\Newspack\Memberships::is_active()` so they no longer clutter `wp help` on sites without WooCommerce Memberships. I used the Memberships check rather than the diagnostics command's `WC_Memberships_For_Teams_Loader` guard because migrate-manual-members operates on plain Memberships plans (no Teams), so a Teams-only gate would wrongly hide it on a Memberships-but-not-Teams site; the whole suite targets WC Memberships, and Teams implies Memberships anyway.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed. `strrchr` is now captured first and `substr` only runs when it returned a non-false match (`$at = strrchr(...); $user_domain = $at ? strtolower( substr( $at, 1 ) ) : ''`), silencing the PHP 8 `substr( false, … )` deprecation. (Went with an if-guard rather than `?:` since the WP coding standards ban short ternaries.)

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-teams-migration.php
adekbadek: Added three tests locking in the `find_group_subscription_for_owner()` filtering that the migrate-teams idempotency guarantee rests on: an inactive (cancelled) owner-owned group sub is not resolved; an active but non-group owner-owned sub is not resolved; and among a mix (cancelled group + active non-group + active group) the fallback selects the active group one and promotes into it. I left the multi-team merge-warning path to the acknowledged live-test plan — it lives inside migrate_teams' full command loop, which needs the line-item machinery the WC mocks deliberately don't model.

nppd1753-pr-series-plan.md
```diff
@@ -0,0 +1,107 @@
+# NPPD-1753 — Implement admin-side Subscribers management
```
dkoo: **Severity:** Suggestion

This fix commit adds two files at the workspace root that look accidental and unrelated to the migration hardening: `nppd1753-pr-series-plan.md` (a planning doc) and `subscribers-search.yml` (a Playwright/accessibility snapshot — `rowgroup`/`row`/`cell` refs). Neither belongs in the plugin change; worth removing before merge so they don't land in the tree.

dkoo (commented):
Re-reviewed the fix commit (code-only, static — I diffed 35d80e4...5a779ed and read the reworked functions at head; I did not re-run the env, so the prior full-env verification stands for the unchanged surface).

All of the earlier points are addressed, and cleanly:

- **Past-due `next_payment`** — both the reuse path (`replace_subscription_product`) and the create path (`create_migration_subscription`) now go through the shared `build_subscription_dates()` → `next_future_payment_date()`, which rolls `next_payment` forward to the first occurrence after `now()` instead of `start + one interval`. The `while` loop has a sensible defensive floor against a period that fails to advance the timestamp.
- **Expired-team end date** — `build_subscription_dates()` sets `end` only when `strtotime( $end_date ) > time()`, so an already-expired team migrates as ongoing rather than erroring on a past end date, matching `migrate-manual-members`.
- **Bulk activation side effects** — `suppress_data_events()` is added to both bulk commands and cancels the dispatch at `newspack_data_events_dispatch_body` (WP_Error being the documented cancel path), so the backfill no longer emits a burst of ESP/webhook/network-sync traffic. The re-run guard `member_has_migration_subscription()` correctly filters on `created_via === 'manual migration'` — the exact value the individual-create path stamps — so a re-run no longer stacks duplicate $0 subscriptions.
- **Variable-subscription billing** — `migrate_teams` now defaults an empty period/interval to `month`/`1`, mirroring the create helpers.
- **Over-limit reused group** — a `WP_CLI::warning()` now fires when the derived seat limit is below `get_member_count()`.
- **Command registration** — all four commands are gated behind `Memberships::is_active()`.
- **`strrchr` deprecation** — captured to a variable first with a false-guard.
- **Test coverage** — three new tests exercise the owner-subscription fallback's active / non-group / mixed-selection filtering.

One finding is addressed only partway, which I think is fine: the `posts_per_page => -1` queries are still not paginated, but `wp_clear_object_cache()` is now called in all three per-item loops, which covers the memory-accumulation risk that mattered for a controlled one-time migration.

One net-new, non-blocking thing: this commit also adds two files at the repo root — `nppd1753-pr-series-plan.md` and `subscribers-search.yml` (a Playwright/accessibility snapshot) — that look accidental and unrelated to the migration hardening. Worth dropping them before merge.

Nothing here is wrong and no new blocker appeared — still approve with comments.

dkoo (approved):
@adekbadek Thanks for addressing the prior review feedback! I've manually run through all testing steps and ran into zero issues, so pre-approving. One non-blocking nit that I wanted to mention: some commands run in dry-run mode by default and require a `--live` flag to run live. Others run live by default and require a `--dry-run` flag to run in dry run mode. Both conflicting patterns already exist for various CLI commands across this repo, but since the commands in this PR are part of a suite that are likely to be used in tandem, maybe we should pick one approach and settle all of them: In general I like the dry-run-default requiring `--live` for live mode since it's a safer default. But your call.

nppd1753-pr-series-plan.md
adekbadek: Good catch, both removed in adaacee. Confirmed accidental: they were swept into 5a779ed13 alongside the real CLI changes and are referenced nowhere in the tree. `subscribers-search.yml` is exactly what you guessed, a raw Playwright a11y snapshot (`[ref=…]` handles and all), and the .md is a working plan doc for the PR series. Full suite still green (2018 tests, 7 pre-existing skips).

One thing worth flagging beyond the removal: the plan doc names a specific publisher six times, and 5a779ed13 is already pushed here. Deleting it in a follow-up commit takes it out of the tree but not out of the pushed history, so it stays reachable via the commit. I've escalated that to a human to decide whether this warrants a history rewrite rather than acting on it myself.

adekbadek:
Addressed the flag inconsistency from @dkoo's approval — thanks, agreed that dry-run-by-default is the safer convention to settle on.

All four commands now run **dry-run by default and require `--live` to write**; `migrate-teams`, `migrate-team-products` and `migrate-manual-members` previously wrote live with no flags and have been flipped to match `backfill-team-managers`. The internal `$dry_run` variable is preserved and simply inverted at the flag read, so no downstream logic moved.

One related fix that runtime testing surfaced: `migrate-team-products` printed `set enabled=yes` and `1 product(s) updated` in past tense during a dry-run, claiming writes it hadn't made. That was tolerable when dry-run was opt-in, but it's now the default path, so the wording is conditional (`would set` / `would be updated`), matching the other commands.

Verified against a local env (product with the Team membership option enabled, `_wc_memberships_for_teams_max_member_count = 7`):

- `wp newspack <cmd> --help` renders `[--live]` and no `--dry-run` for all four commands.
- Bare `migrate-team-products` → dry-run banner, `would set enabled=yes, limit=7`, and the group meta stays unwritten.
- `migrate-team-products --live` → no banner, `set enabled=yes, limit=7`, meta written (`enabled=yes`, `limit=7`).

`Test_Teams_Migration` (15) and the `WooCommerce_Subscriptions_Integration` group (104) are green; PHPCS clean.

Labelled **Do not merge**: this is stacked on #608 and should only be merged once that lands (the diff base is that branch and still needs to retarget to `main`).

PR #628, opened 2026-07-15
feat(audience): per-field operators for incoming ESP field criteria

miguelpeixe:
### All Submissions:

Publishers can now assign a per-field **segmentation operator** to each incoming ESP field (in **Audience → Connections → _integration_**), so those fields become usable Campaigns segment criteria instead of a free-text / exact-match box. The chosen operator drives both the control the segment editor renders and how readers are matched:

- **Text** (`default`) — free text, exact match (the prior default)
- **Number** (`range`) — numeric min/max; decimals allowed
- **Single value** (`default`) / **Multiple values** (`list__in`) — for enumerated fields; "Multiple values" matches ActiveCampaign's `||A||B||` multi-value format **client-side** (segments are matched in the browser; only the server side parsed this before)

**Operator choices are constrained by the field's `value_type`** so a conflicting type can't be picked — a `date`/`datetime` field offers Text only (real date matching is Phase 2), a multi-value field offers Multiple only, and so on. Integrations declare the `value_type`; the framework falls back to a has-options heuristic for fields left as plain `string`.

The `{ field_key => operator }` choice is persisted per field and flows consistently: admin save → REST sanitize → storage (in the field's `raw_data`) → the settings API → the segment editor's control selection → the client-side matchers. It works for **any** reader-activation integration — the stored operator is re-applied after each integration's `configure_incoming_field()`, so a non-ESP integration (e.g. newspack-manager's ActiveCampaign) that derives its own defaults doesn't clobber the publisher's choice.

**The built-in ESP integration now declares `value_type` too.** newspack-plugin's `ESP` class (fed by newspack-newsletters) already consumed `value_type` / `matching_function` / `options` from each provider's field schema, but both providers emitted `value_type => 'string'` for every field. This PR teaches them to map their own field types, so a reader synced through the **standard** ESP path gets the same type-constrained operators as the managed path:

- **ActiveCampaign** — `dropdown`/`radio`/`listbox` → `select`, `checkbox`/`multiselect` → `multiselect`, `date` → `date`, `datetime` → `datetime`. This mirrors newspack-manager's AC integration (https://github.com/Automattic/newspack-manager/pull/657) so the **same AC field types identically** whether a contact syncs via newspack-newsletters or the managed integration.
- **Mailchimp** — `dropdown`/`radio` → `select`, `date`/`birthday` → `date`, `number` → `number` (range-matched, default operator `range`); text/phone/url/zip/etc. stay `string`.

Also included: the numeric `range` value schema now permits decimals; the client `list__in`/`list__not_in` matchers parse AC's `||`-delimited values; plus an `n test-js` tooling fix (newspack-plugin's pnpm package name diverges from its directory) and a PHP 8.0-safe list check (avoids `array_is_list()`, which is 8.1+).

**Companion PR:** newspack-manager's **managed** ActiveCampaign integration — a separate reader-activation integration from the newspack-newsletters ESP path above — gets the same treatment in [Automattic/newspack-manager#657](https://github.com/Automattic/newspack-manager/pull/657): the same `value_type` mapping plus fetching its option choices. The two **depend on each other — land together.**

**Out of scope (Phase 2):** date-range matching. Date and datetime fields are selectable and correctly typed (as distinct `value_type`s), but offer only Text for now.

Ref: NPPD-2047.

**Automated** (green in an isolated env):
- newspack-plugin: `n test-php --filter Test_Integrations` (68) and `--filter Test_ESP` (22); `n test-js` (335)
- newspack-popups: `n test-php --filter SchemasTest` (21); `n test-js` (61)
- newspack-newsletters: `n test-php --filter IntegrationsSchema` (14 — AC + Mailchimp `value_type` / `matching_function` mapping)

**Manual** (with [newspack-manager#657](https://github.com/Automattic/newspack-manager/pull/657) active + an ActiveCampaign connection):
1. In **Audience → Connections → ActiveCampaign**, enable incoming fields of several AC types (a text/amount field, a `date`, a `dropdown`, a `checkbox`).
2. Confirm the per-field operator dropdown is **constrained by type**: date → Text only; checkbox/multiselect → Multiple values only; dropdown/radio/listbox → Single/Multiple; text → Text/Number; score → Number.
3. In **Audience → Campaigns → Segments**, add each as a criterion and confirm the rendered control matches: min/max for **Number**; checkboxes (with the field's real option choices) for **Multiple**; a dropdown for **Single**; a text box for **Text**.
4. Confirm matching: with **Number** `max = 100`, a reader whose value is `50` matches and `150` doesn't; a decimal like `49.99` saves without a validation error; a **Multiple** criterion against a reader value `||A||C||` matches on `A`.
5. Regression: existing default criteria (Articles read range, Subscribed lists), plain text fields, and outbound metadata fields are unchanged.
6. **Standard ESP path** (no newspack-manager): with a **Mailchimp** or **ActiveCampaign** connection via newspack-newsletters, enable a `date`, a `dropdown`/`radio`, and (Mailchimp) a `number` merge field, then confirm the same type constraints — date → Text only; dropdown/radio → Single/Multiple; Mailchimp `number` → Number.

### Other information:


plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -637,11 +637,11 @@ public function get_enabled_outgoing_fields() {
...
 	 *
-	 * @param string[] $keys Array of field keys to enable.
+	 * @param array $fields Array of field keys, or a map of key => matching_function.
 	 *
 	 * @return bool True if updated, false otherwise.
 	 */
-	public function update_enabled_incoming_fields( $keys ) {
+	public function update_enabled_incoming_fields( $fields ) {
 		$available = $this->get_available_incoming_fields();
 		if ( is_wp_error( $available ) ) {
```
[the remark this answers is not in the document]
miguelpeixe: Good catch. Added an is_array() guard as the first statement of update_enabled_incoming_fields(), so non-array input no-ops (restoring the pre-change tolerance) instead of fataling on array_values(). Covered by a new test_update_enabled_incoming_fields_tolerates_non_array. Fixed in b16ccc61.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1136,6 +1161,24 @@ protected function sanitize_settings_field_value( $field, $value ) {
...
+						// Legacy: a plain list of enabled keys, no operator override.
+						foreach ( $value as $key ) {
+							$sanitized[ \sanitize_text_field( (string) $key ) ] = 'default';
+						}
+					} else {
+						foreach ( $value as $key => $operator ) {
+							$operator = is_string( $operator ) ? $operator : 'default';
+							$sanitized[ \sanitize_text_field( (string) $key ) ] = in_array( $operator, $allowed_matching_functions, true ) ? $operator : 'default';
+						}
+					}
```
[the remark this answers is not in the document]
miguelpeixe: Fixed. The legacy plain-list path now stays a list instead of a {key => default} map, so update_enabled_incoming_fields() preserves each field's provider-default matching_function (list entries map to a null = no override). Covered by a new test_sanitize_incoming_metadata_keeps_legacy_list. Fixed in b16ccc61.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1136,6 +1172,26 @@ protected function sanitize_settings_field_value( $field, $value ) {
...
+							$sanitized[] = \sanitize_text_field( (string) $key );
+						}
+					} else {
+						foreach ( $value as $key => $operator ) {
+							$operator = is_string( $operator ) ? $operator : 'default';
+							$sanitized[ \sanitize_text_field( (string) $key ) ] = in_array( $operator, $allowed_matching_functions, true ) ? $operator : 'default';
+						}
+					}
+					return $sanitized;
+				}
```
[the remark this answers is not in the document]
miguelpeixe: Both incoming-field write paths now drop empty keys: `sanitize_settings_field_value()` skips any key that sanitizes to `""`, and `update_enabled_incoming_fields()` does the same while building the operator map, so a `""` key can no longer be persisted — matching the read-side skip already in `get_enabled_incoming_fields()`. Fixed in 1a192d6.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -259,21 +332,47 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
...
+									const optionLabel = option.label || option.value;
+									// The stored value for this field is a { key => operator } map, not an array:
+									// a key present means enabled, and its value is the chosen matching operator.
+									const currentMap = getFieldValue( inboundField ) || {};
+									const checked = Object.prototype.hasOwnProperty.call( currentMap, optionValue );
+									const operatorOptions = operatorOptionsForField( option );
+									// If the stored operator isn't among the options offered for this field's
+									// current value_type (e.g. a field enabled before it declared a type), fall
+									// back to the first option so the control never shows a value with no option.
+									const selectedOperator = operatorOptions.some( o => o.value === currentMap[ optionValue ] )
```
dkoo, line 423: **Suggestion:** For a field enabled before its `value_type` was declared (stored operator `'default'`), `operatorOptionsForField()` now offers only the typed operator (e.g. `range` for a number field). `selectedOperator` **displays** `operatorOptions[0]`, but `currentMap[optionValue]` keeps `'default'`, and with a single option the `SelectControl`'s `onChange` can never fire to persist the fallback — so the effective operator stays `'default'` while the UI shows 'Number'. Reader matching for these fields is unchanged from before this release (the segment editor and content-gate read the same stored operator), so this is a config-screen label mismatch rather than a matching regression — but a numeric field mislabeled 'Number' can mislead a publisher into leaving a range gate that (as exact-match) admits no one. Consider persisting the reconciled operator, e.g. a one-time on-read backfill, when the stored operator isn't valid for the field's declared `value_type`.

plugins/newspack-popups/src/criteria/matching-functions.js
```diff
@@ -63,7 +84,10 @@ export default {
 			return false;
 		}
 		const { min, max } = config.value;
-		if ( ( min && criteria.value < min ) || ( max && criteria.value > max ) ) {
+		// Treat only genuinely-absent bounds ( undefined / null / '' ) as unbounded, so a
+		// min or max of 0 is still enforced — matching the server's (float) min/max compare.
+		const hasBound = value => undefined !== value && null !== value && '' !== value;
```
dkoo, line 95: **Suggestion:** The comment says `hasBound` matches "the server's (float) min/max compare," but the server floors an absent `min` at `0` and caps an absent `max` at `PHP_INT_MAX` (`class-promoted-fields.php`), whereas the client treats absent / `null` / `''` bounds as fully unbounded. They diverge for a negative reader value with an unset lower bound (server excludes, client includes) and for non-numeric reader values (server coerces to `0.0` and can match; the client's `isNaN` guard fails closed). ESP numeric fields are typically non-negative so impact is narrow — either default the client's absent lower bound to `0`, or soften the comment to note parity holds only when the bound is present.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1197,6 +1253,32 @@ protected function sanitize_settings_field_value( $field, $value ) {
...
+							}
+							$sanitized[] = $key;
+						}
+					} else {
+						foreach ( $value as $key => $operator ) {
+							$key = \sanitize_text_field( (string) $key );
+							if ( '' === $key ) {
+								continue;
+							}
+							$sanitized[ $key ] = ( is_string( $operator ) && in_array( $operator, self::ALLOWED_INCOMING_MATCHING_FUNCTIONS, true ) ) ? $operator : 'default';
```
dkoo: **Suggestion:** In the map branch of the incoming-fields sanitizer, an operator not on the allowlist is coerced to `'default'` — itself a valid allowlist member — which then overrides the field's provider-default operator in `update_enabled_incoming_fields()`. For a `multiselect`/`select` field whose real default is `list__in`, a malformed operator silently downgrades matching to exact-match (which never matches a typed field). Only reachable via a crafted REST payload from an already-capability-gated admin, since the UI constrains the dropdown per `value_type` — so risk is minimal. Consider dropping the key (leaving the provider default intact) instead of coercing to `'default'`.

dkoo (approved):
Manually tested end-to-end with an ActiveCampaign contact seeded with custom field values—the corresponding reader account matches segmentation criteria. 

Approve with comments — release risk: LOW. Solid, well-tested feature with no blockers. The `{field_key => operator}` choice flows consistently through save → REST sanitize → storage → settings API → segment editor → client matchers, with a single operator allowlist enforced on every write path and re-applied on read. The map-shape settings change is contained to newspack-plugin's own REST config builder and React UI; all cross-plugin, sync and content-gate consumers read `Incoming_Field` objects, which are unaffected. PHP and JS suites pass in an isolated env, and the Segments editor renders cleanly.

A few things worth a look before merge, none blocking:

1. A field enabled before its `value_type` was declared can **display** a typed fallback operator the UI never persists (config-screen label only — reader matching for those fields is unchanged from before this release, but a numeric field mislabeled 'Number' can mislead a publisher). A one-time on-read reconciliation would resolve it.
2. `valuesMatch()` doesn't compare the new `{key => operator}` map shape, so a net-zero inbound edit reads as dirty and keeps the Save action and unsaved-changes guard armed.
3. The client range matcher's parity comment overstates agreement with the server for absent / negative / non-numeric bounds.
4. An invalid operator in a crafted REST payload coerces to `'default'`, overriding a typed field's provider default (capability-gated admin only).

Four additional nits are noted inline. Verification was partial: the live-ESP operator UI in Audience → Connections needs an ActiveCampaign/Mailchimp connection the review env can't provision, so that surface was reviewed by code + passing schema/mapping tests rather than exercised. Companion newspack-manager#657 must land together.

- `plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js:102` — **Suggestion:** `incoming_metadata_fields` values are now `{key => operator}` objects, but `valuesMatch()` handles only arrays and booleans and falls through to reference equality (`a === b`) for objects. A net-zero inbound edit (toggle a field on then off, or change an operator then change it back) leaves `hasPending` true, keeping the Save action armed and the unsaved-changes navigation guard active for semantically-unchanged settings — the same stuck-dirty problem the array branch was added to avoid. Consider a shallow key-by-key object comparison for the map shape.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
miguelpeixe: Addressed in 5bc0337.

Rather than a write-on-read backfill, the reconciliation is folded into the next save. A new `reconcileOperators()` helper rewrites any **enabled** field whose stored operator isn't valid for its declared `value_type` to that field's first valid option, and the Save action submits the reconciled map alongside the draft.

That repairs the stored operator without the two side effects a true on-read backfill would have had: it never writes during a plain GET, and it never arms the unsaved-changes guard on page load for settings nobody touched. It returns the *same map reference* when nothing needs repair, so a healthy integration still submits only what the user actually edited.

Covered by `reconciles a stored operator that is invalid for the declared value_type`, `returns the same map when every stored operator is already valid`, `never enables a field that is absent from the map`, plus two end-to-end save-payload tests.

plugins/newspack-popups/src/criteria/matching-functions.js
miguelpeixe: Addressed in 5bc0337 — softened the comment rather than changing the matcher.

It now records that parity with the server holds only *while a bound is present*: the server floors an absent `min` at 0 and caps an absent `max` at `PHP_INT_MAX` (`class-promoted-fields.php`), whereas an absent bound here is fully unbounded, so the two diverge for a negative reader value with no lower bound. It also notes the `isNaN` guard fails closed where the server coerces a non-numeric reader value to `0.0`.

Went with the comment rather than defaulting the client's absent lower bound to 0 deliberately: that would change live segment matching on existing sites, which felt like more than a comment-accuracy fix should carry. Happy to switch if you'd rather have true parity here.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Addressed in 5bc0337.

An operator outside the allowlist now maps to `null` (no override) instead of `'default'`, so `update_enabled_incoming_fields()` leaves the field's provider default intact — a `multiselect`/`select` field keeps its `list__in` rather than being silently downgraded to exact match.

Worth noting the field stays *enabled*; only the operator override is dropped, since dropping the whole key would have discarded the admin's actual intent to enable it. Updated `test_sanitize_incoming_metadata_preserves_operators` to assert both halves of that (`assertArrayHasKey` + `assertNull`).

miguelpeixe:
Thanks for the thorough pass @dkoo — all four points are addressed in 5bc0337. `main` was also merged in first (0792dc0) to clear a conflict in `configure-view.js`/`.scss`, so line numbers have shifted since your review.

**1. Field enabled before its `value_type` was declared displays an operator it never persists**
Fixed. New `reconcileOperators()` helper rewrites any enabled field whose stored operator isn't valid for its declared `value_type` to the first valid option, folded into the next save rather than a write-on-read backfill — so it never writes on a GET and never arms the unsaved-changes guard on page load. Details in the inline thread.

**2. `valuesMatch()` doesn't compare the `{key => operator}` map shape**
Fixed. Added a map branch that compares key-by-key (with an `isMap()` guard that excludes arrays and `null` so the existing branches are untouched), so a net-zero inbound edit no longer reads as dirty. Two regression tests: `treats an equivalent inbound operator map as unchanged` and `still flags a genuinely changed inbound operator map as pending`.

Worth flagging that the *stuck-after-save* case was already benign — `handleSave()` stores the draft by reference, so the reconcile effect cleared the key via identity. What was actually broken is the case you described: toggling a field on then off, or changing an operator and changing it back, left Save and the nav guard armed.

**3. Range matcher parity comment overstates agreement with the server**
Fixed as a comment change. It now says parity holds only while a bound is present, and spells out the absent-`min`-floors-at-0 / absent-`max`-caps-at-`PHP_INT_MAX` divergence plus the non-numeric `isNaN` difference. I deliberately did *not* default the client's lower bound to 0, since that would change live segment matching — say the word if you'd rather have true parity.

**4. Invalid operator coerces to `'default'`, overriding a typed field's provider default**
Fixed. Out-of-allowlist operators now map to `null` (no override) instead of `'default'`, so the provider default survives. The field stays enabled — only the override is dropped.

**Verification:** newspack-plugin JS suite 428/428 (45 suites, 7 new tests); `Test_Integrations` 79, `Test_ESP` 25, `Test_Promoted_Fields` 12 — all green; prettier, ESLint and PHPCS clean on every changed file.

One note for reviewers: the `@wordpress/components` mock in `configure-view.test.js` had no `SelectControl` stub, so no test could render an *enabled* inbound field. Added one — that's what unblocked the new save-payload coverage.

PR #630, opened 2026-07-15
fix(subscriptions): keep group owner sub on member delete (NPPM-3021)

dkoo:
### All Submissions:

Fixes a data-loss bug: a **member** of a group subscription who deletes their own reader account (self-service **My Account → Delete Account**) would also permanently delete the **group owner's** subscription — cutting off access for the entire group. The owner's subscription was force-deleted (not trashed), leaving orphaned renewal orders.

Root cause: Newspack surfaces a group subscription to its members by filtering `wcs_get_users_subscriptions()`, guarded only by `is_account_page()`. WooCommerce Subscriptions' `delete_user` cascade (`WC_Subscriptions_Manager::trash_users_subscriptions()`) calls that same function and `->delete( true )`s everything it returns. Because self-service deletion runs on the My Account page, the member's group-owner subscription was injected into the cascade and hard-deleted.

This PR:

- **(primary fix)** Removes a departing user from their group memberships on `delete_user` / `wpmu_delete_user` (priority 5, ahead of WCS's cascade), so the cascade only ever deletes subscriptions the user actually owns.
- Hardens the member-injection filter to bail during a user-deletion cascade and to only augment the **current viewer's own** account, so cross-user reads (admin views, contact syncs) see owner-only subscriptions.
- Excludes group subscriptions the user is only a **member** of from the ESP contact-sync subscription metadata and "current product" resolution, so a member's ESP subscription fields reflect only subscriptions they own. Group access continues to sync via the **Content Access Group / Content Access Source** fields. Gifted subscriptions are preserved.

No public API/hook signatures change.

Closes NPPM-3021.

Prerequisites: Access Control feature flag on, v1 My Account active, WooCommerce + WooCommerce Subscriptions active, WC Memberships **not** active.

1. As an admin, create a group subscription owned by **User A** and add **User B** as a member (e.g. via `Group_Subscription::update_members()` or the group-management UI). Confirm User B has group access.
2. Log in as **User B** → **My Account → Edit Account → Delete Account**; complete the email-confirmation flow to delete User B's account.
3. Verify **User A's group subscription still exists** (WooCommerce → Subscriptions) and is unchanged, and that other group members still have access. (Before this fix, User A's subscription would be gone.)
4. Confirm User B's own subscriptions (if any) are still cleaned up when their account is deleted.
5. Delete a group **member** from **wp-admin → Users** — confirm the owner's subscription is likewise unaffected.
6. ESP metadata: for a reader who is only a **member** of a group subscription (owning none of their own), trigger a contact sync and confirm the subscription metadata fields (Subscriber Status, Active Subscription Count, etc.) are empty/zero — the group is reflected via **Content Access Group / Content Access Source**, not the subscription fields.
7. Regression: a group **owner** viewing **My Account → Subscriptions** still sees and can manage their group subscription; a member still sees the group subscription in their account per existing behavior.

Automated: `n test-php` for `newspack-plugin` passes (1678 tests). New regression tests cover the delete cascade, the current-user allowlist, the membership-cleanup ordering, and the ESP-metadata exclusion.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-myaccount.php
@@ -63,6 +63,12 @@ class Group_Subscription_MyAccount {
 		add_action( 'template_redirect', [ __CLASS__, 'redirect_legacy_manage_members' ] );
 		add_filter( 'wcs_get_users_subscriptions', [ __CLASS__, 'inject_member_group_subscriptions' ], 15, 2 );
 		add_filter( 'map_meta_cap', [ __CLASS__, 'grant_group_member_view_order_cap' ], 15, 4 );
+		// Remove a departing user from their group memberships before WooCommerce
+		// Subscriptions' `delete_user` cascade (priority 10) force-deletes their
+		// subscriptions, so the cascade never sees a group owner's subscription as
+		// belonging to the member being deleted. See NPPM-3021.
+		add_action( 'delete_user', [ __CLASS__, 'remove_deleted_user_from_groups' ], 5 );
+		add_action( 'wpmu_delete_user', [ __CLASS__, 'remove_deleted_user_from_groups' ], 5 );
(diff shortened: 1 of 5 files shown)


dkoo:
Fixes a highly destructive bug known to affect production sites, so merging without human review.

PR #631, opened 2026-07-16
feat(subscribers): audience wizard shell + read endpoints (NPPD-1753)

adekbadek:
### All Submissions:

Part of NPPD-1753 (Subscribers management surface). **PR 3 of the series** – the first user-visible slice. Adds the Audience → Subscribers wizard shell, its read-only REST endpoints, and both level-0 list screens.

Depends on #608 (group-subscription backend), which has merged; this branch is rebased onto `main` so the diff is just this slice. Later slices – the in-wizard group detail and person profile, the plans endpoint, and the tags/newsletters columns – follow in their own PRs, so NPPD-1753 stays open after this merges.

**Backend (`class-subscribers-wizard.php`)**
- Registers the `newspack-subscribers` wizard under the Audience menu, riding the existing Access Control feature flag (no new flag).
- Three read endpoints under `newspack/v1/wizard/newspack-subscribers/`: `subscribers` (server-side paginated / filtered / sorted), `groups` (full set), `avatars` (batched email→avatar lookup). All gated by `manage_options`.
- Subscription status / plan filters resolve to customer IDs HPOS-safely and invert into a capped `WP_User_Query` include. The `cancelled` status filter matches only fully-churned readers – a reader who still holds a live plan is excluded.
- Per-row group memberships are built once per request from the site's groups (no N+1).

**Frontend (`src/wizards/subscribers/`)**
- A ~1:1 port of the signed-off prototype's presentation layer; only the `data/` hooks are new, wired to the endpoints above.
- **Subscribers** list: controlled server-side DataViews – pagination / sort / status filter round-trip to the endpoint.
- **Groups** list: loads the full (small) set and filters / sorts / paginates client-side.
- Interim click-through to the native admin edit screens (via server-provided `editUrl`s) until the in-wizard detail views land in a later slice. Both tabs follow the same rule: a person's name (and the row around it) opens that person's user-edit screen, while a plan name opens that subscription.

1. On a site with WooCommerce Subscriptions and the Access Control feature enabled, build the plugin: `n build newspack-plugin`.
2. Seed several subscribers and at least one group/team subscription (owner + managers + members).
3. Go to **Audience → Subscribers**. Confirm the Subscribers list paginates, sorts, searches, and filters by status/plan **server-side** (network tab hits `.../wizard/newspack-subscribers/subscribers` with the matching params).
4. Tick **Cancelled** in the status filter – only fully-churned readers should appear (a reader holding a live plan alongside a cancelled one must *not* show).
5. Open the **Groups** tab; confirm owner-inclusive member counts and seat limits render, and the status/plan filters work. A group awaiting its first payment (WCS status `pending`) stays visible in the default view and carries a **Pending** badge.
6. Check the click targets in both tabs: clicking a row (or the subscriber / group-owner name in it) opens that person's user-edit screen, while clicking a plan name – in the Subscribers tab's Subscription column, or under the owner in the Groups tab – opens that subscription's edit screen.
7. Run the tests: from `plugins/newspack-plugin`, `n test-php` (subscribers endpoint + groups endpoint suites) and `n test-js` (the `use-subscribers` / `use-groups` hook tests).

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
+++ b/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
@@ -49,6 +49,21 @@ class Subscribers_Wizard extends Wizard {
 	const FILTER_INCLUDE_CAP = 10000;
 
 	/**
+	 * Upper bound on the number of avatars one /avatars request resolves. A single
+	 * list page needs at most `per_page` (≤100); the client batches larger sets.
+	 */
+	const AVATAR_BATCH_CAP = 200;
+
+	/**
+	 * How many subscriptions the filter scan hydrates at a time. The scan itself
+	 * is bounded by FILTER_INCLUDE_CAP, but WooCommerce returns fully hydrated
+	 * WC_Subscription objects and only their customer ID is needed here — walking
+	 * the set a page at a time keeps peak memory at one chunk rather than the
+	 * whole cap. See customer_ids_for_raw_statuses().
+	 */
+	const FILTER_SCAN_CHUNK = 500;
+
+	/**
 	 * Per-request memo of the site's group subscriptions (each with its resolved
 	 * settings). A single request can resolve these several times — once per active
 	 * filter set plus the group list itself — and each resolution hydrates every
@@ -66,6 +81,14 @@ class Subscribers_Wizard extends Wizard {
 	 * @var array<int,array<int,array>>|null
 	 */
 	private $group_membership_index = null;
+
+	/**
+	 * Per-request memo of customer IDs keyed by the sorted status set they were
+	 * resolved for. See customer_ids_for_raw_statuses().
+	 *
+	 * @var array<string,int[]>
+	 */
+	private $raw_status_ids_cache = [];
 
 	/**
 	 * Constructor.
@@ -133,9 +156,14 @@ class Subscribers_Wizard extends Wizard {
 				'permission_callback' => [ $this, 'api_permissions_check' ],
 				'args'                => [
 					'emails' => [
-						'type'     => 'array',
-						'required' => true,
-						'items'    => [ 'type' => 'string' ],
+						'type'              => 'array',
+						'required'          => true,
+						'items'             => [ 'type' => 'string' ],
+						// Sanitized declaratively, like the sibling `size` arg. Invalid
+						// entries are dropped rather than failing the whole request:
+						// one malformed address shouldn't blank a page of avatars.
+						'sanitize_callback' => [ __CLASS__, 'sanitize_emails_arg' ],
+						'validate_callback' => 'rest_validate_request_arg',
 					],
 					'size'   => [
 						'type'              => 'integer',
@@ -211,10 +239,27 @@ class Subscribers_Wizard extends Wizard {
 						'type'              => 'array',
 						'items'             => [ 'type' => 'string' ],
 						'sanitize_callback' => 'rest_sanitize_request_arg',
+						'validate_callback' => 'rest_validate_request_arg',
 					],
 				],
 			]
 		);
+	}
+
+	/**
+	 * Sanitize the /avatars `emails` arg: valid addresses only, deduplicated and
+	 * bounded to AVATAR_BATCH_CAP.
+	 *
+	 * Invalid entries are dropped rather than rejected so a single malformed
+	 * address can't fail the whole batch.
+	 *
+	 * @param mixed $emails The raw `emails` request arg.
+	 *
+	 * @return string[] Sanitized, unique, capped email addresses.
+	 */
+	public static function sanitize_emails_arg( $emails ) {
+		$sanitized = array_filter( array_map( 'sanitize_email', (array) $emails ) );
+		return array_values( array_slice( array_unique( $sanitized ), 0, self::AVATAR_BATCH_CAP ) );
 	}
 
 	/**
@@ -257,6 +302,7 @@ class Subscribers_Wizard extends Wizard {
 	public function api_get_groups() {
 		$this->group_subscriptions_cache = null;
 		$this->group_membership_index    = null;
+		$this->raw_status_ids_cache      = [];
 		$items                           = [];
 		foreach ( $this->get_group_subscriptions() as $group ) {
 			$items[] = $this->prepare_group( $group['subscription'], $group['settings'] );
@@ -363,9 +409,10 @@ class Subscribers_Wizard extends Wizard {
 	public function api_get_subscribers( $request ) {
 		$this->group_subscriptions_cache = null;
 		$this->group_membership_index    = null;
+		$this->raw_status_ids_cache      = [];
 		$per_page                        = (int) $request->get_param( 'per_page' );
-		$page     = (int) $request->get_param( 'page' );
-		$search   = trim( (string) $request->get_param( 'search' ) );
+		$page                            = (int) $request->get_param( 'page' );
+		$search                          = trim( (string) $request->get_param( 'search' ) );
 
 		$query_args = [
 			'role__in'    => Reader_Activation::get_reader_roles(),
@@ -418,6 +465,19 @@ class Subscribers_Wizard extends Wizard {
 
 	/**
 	 * Shape a reader user for the admin subscriber list.
+	 *
+	 * Cost note: group memberships are indexed once per request
+	 * (get_group_membership_index), but two lookups here are inherently per-row —
+	 * wcs_get_users_subscriptions() and last_payment_date()'s wc_get_orders() —
+	 * so a page costs ~2 × `per_page` (≤100) WooCommerce queries. Both are
+	 * "newest/owned rows for one customer" lookups that WooCommerce can't answer
+	 * per-customer in a single batched query: batching last_payment_date() over
+	 * the page's customer IDs would need either an unbounded `limit => -1` (worse
+	 * peak memory than the bounded per-row queries) or a truncated scan that
+	 * silently blanks the last payment of a reader whose most recent order falls
+	 * outside it. The bounded per-row shape is the correct trade-off here, and
+	 * matches the existing convention elsewhere in the plugin
+	 * (Sync\Contact_Metadata\Engagement::get_latest_order).
 	 *
 	 * @param \WP_User $user The reader user.
 	 *
@@ -567,6 +627,22 @@ class Subscribers_Wizard extends Wizard {
 	 * dropped whenever any live status remains. Empty when they have no
 	 * subscription at all (a free reader shows no status badge).
 	 *
+	 * SOURCE OF TRUTH for the status-reduction rule. The rule is:
+	 *
+	 *   Rank statuses active → pending → on-hold → cancelled, and drop
+	 *   `cancelled` entirely whenever the reader still holds any live
+	 *   (non-cancelled) subscription; a reader with no subscription at all has
+	 *   no status.
+	 *
+	 * It is re-encoded in three other places, each of which points back here and
+	 * must be changed in step or the endpoint's filter will contradict the badge
+	 * it filters on:
+	 *   - customer_ids_for_statuses() below — the endpoint's `status` filter.
+	 *   - displayStatuses() in src/wizards/subscribers/status.js — the row's
+	 *     status badges.
+	 *   - visiblePlanEntries in src/wizards/subscribers/screens/SubscriberList.jsx
+	 *     — the Subscription column.
+	 *
 	 * @param array $subscriptions Individual subscription entries.
 	 * @param array $groups        Group membership entries.
 	 *
@@ -667,8 +743,9 @@ class Subscribers_Wizard extends Wizard {
 
 	/**
 	 * Customer IDs whose displayed status matches any of the given prototype
-	 * statuses, mirroring the list's badge reduction (see reduced_status /
-	 * displayStatuses): a live status always qualifies, but `cancelled` matches
+	 * statuses, mirroring the list's badge reduction (reduced_status() above is
+	 * the documented source of truth for that rule): a live status always
+	 * qualifies, but `cancelled` matches
 	 * only fully-churned readers — anyone who also holds a live (active/pending/
 	 * on-hold) subscription is dropped, since the badge hides cancelled while a
 	 * live plan remains. Without this the Cancelled filter would surface readers
@@ -711,21 +788,39 @@ class Subscribers_Wizard extends Wizard {
 		if ( empty( $prototype_statuses ) ) {
 			return [];
 		}
+		// Memoized per status set: customer_ids_for_statuses() asks for up to three
+		// overlapping sets per request (and a filter like ['active','cancelled']
+		// asks for `active` twice), each of which would otherwise repeat the full
+		// subscription scan below.
+		sort( $prototype_statuses );
+		$cache_key = implode( ',', $prototype_statuses );
+		if ( isset( $this->raw_status_ids_cache[ $cache_key ] ) ) {
+			return $this->raw_status_ids_cache[ $cache_key ];
+		}
 		$ids = [];
 
 		// Individual + owned subscriptions in a matching WCS status (HPOS-safe).
 		// Bounded to the same ceiling the include set is capped at, so the filter
-		// path can't load an unbounded number of subscription objects on a large store.
+		// path can't load an unbounded number of subscription objects on a large
+		// store, and walked a chunk at a time: WooCommerce hands back fully
+		// hydrated WC_Subscription objects and only the customer ID is read here,
+		// so paging keeps peak memory at one chunk instead of the whole cap.
 		$wcs_statuses = $this->wcs_statuses_for( $prototype_statuses );
 		if ( ! empty( $wcs_statuses ) && function_exists( 'wcs_get_subscriptions' ) ) {
-			$subs = \wcs_get_subscriptions(
-				[
-					'subscriptions_per_page' => self::FILTER_INCLUDE_CAP,
-					'subscription_status'    => $wcs_statuses,
-				]
-			);
-			foreach ( $subs as $subscription ) {
-				$ids[] = (int) $subscription->get_customer_id();
+			for ( $page = 1; ( $page - 1 ) * self::FILTER_SCAN_CHUNK < self::FILTER_INCLUDE_CAP; $page++ ) {
+				$subs = \wcs_get_subscriptions(
+					[
+						'subscriptions_per_page' => self::FILTER_SCAN_CHUNK,
+						'paged'                  => $page,
+						'subscription_status'    => $wcs_statuses,
+					]
+				);
+				foreach ( $subs as $subscription ) {
+					$ids[] = (int) $subscription->get_customer_id();
+				}
+				if ( count( $subs ) < self::FILTER_SCAN_CHUNK ) {
+					break;
+				}
 			}
 		}
 
@@ -736,7 +831,8 @@ class Subscribers_Wizard extends Wizard {
 			}
 		}
 
-		return array_values( array_unique( array_filter( $ids ) ) );
+		$this->raw_status_ids_cache[ $cache_key ] = array_values( array_unique( array_filter( $ids ) ) );
+		return $this->raw_status_ids_cache[ $cache_key ];
 	}
 
 	/**
@@ -858,14 +954,10 @@ class Subscribers_Wizard extends Wizard {
 		// displays (list: 32px → 64, profile: 64px → 128).
 		$size    = $request->get_param( 'size' );
 		$avatars = [];
-		// A single list page resolves at most `per_page` (≤100) avatars; bound the
-		// batch so an oversized payload can't fan out into unbounded get_avatar_url() calls.
-		$emails = array_slice( (array) $request->get_param( 'emails' ), 0, 200 );
-		foreach ( $emails as $email ) {
-			$email = sanitize_email( $email );
-			if ( '' === $email ) {
-				continue;
-			}
+		// Already sanitized, deduplicated and capped at AVATAR_BATCH_CAP by the
+		// arg's sanitize_callback (sanitize_emails_arg), so an oversized payload
+		// can't fan out into unbounded get_avatar_url() calls.
+		foreach ( (array) $request->get_param( 'emails' ) as $email ) {
 			$avatars[ $email ] = get_avatar_url( $email, [ 'size' => $size ] );
 		}
 		return rest_ensure_response(
--- a/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
+++ b/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
@@ -1,0 +1,932 @@
+<?php
+/**
+ * Newspack Subscribers management wizard.
+ *
+ * The admin-side, people-first view of the site's subscribers and group
+ * subscriptions. Ported from the signed-off i2 design prototype; it lives under
+ * the Audience menu and is visible to any admin who can `manage_options`.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+require_once NEWSPACK_ABSPATH . '/includes/wizards/class-wizard.php';
+
+/**
+ * Subscribers wizard.
+ */
+class Subscribers_Wizard extends Wizard {
+
+	/**
+	 * The slug of this wizard.
+	 *
+	 * @var string
+	 */
+	protected $slug = 'newspack-subscribers';
+
+	/**
+	 * The capability required to access this wizard.
+	 *
+	 * @var string
+	 */
+	protected $capability = 'manage_options';
+
+	/**
+	 * The parent menu slug this wizard hangs under.
+	 *
+	 * @var string
+	 */
+	protected $parent_slug = 'newspack-audience';
+
+	/**
+	 * Upper bound on the number of customer IDs a subscription-status/plan filter
+	 * resolves to. Keeps both the subscription scan and the WP_User_Query IN()
+	 * clause bounded on very large stores. See resolve_filter_include().
+	 */
+	const FILTER_INCLUDE_CAP = 10000;
+
+	/**
+	 * Per-request memo of the site's group subscriptions (each with its resolved
+	 * settings). A single request can resolve these several times — once per active
+	 * filter set plus the group list itself — and each resolution hydrates every
+	 * group subscription, so it's cached for the life of the (per-request) wizard.
+	 *
+	 * @var array<int,array{subscription:\WC_Subscription,settings:array}>|null
+	 */
+	private $group_subscriptions_cache = null;
+
+	/**
+	 * Per-request memo mapping each user ID to their group memberships
+	 * ({ id, plan, status, role }), built once from the site's groups so the
+	 * subscriber list doesn't re-resolve group roles per row.
+	 *
+	 * @var array<int,array<int,array>>|null
+	 */
+	private $group_membership_index = null;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param array $args Optional arguments.
+	 */
+	public function __construct( $args = [] ) {
+		parent::__construct( $args );
+		add_action( 'rest_api_init', [ $this, 'register_api_endpoints' ] );
+	}
+
+	/**
+	 * Whether this wizard is available.
+	 *
+	 * The subscribers surface is the admin face of the group-subscription / Access
+	 * Control feature, so it rides the same flag the rest of that code gates on
+	 * rather than introducing a new one — a site without Access Control has no
+	 * group data to manage here.
+	 *
+	 * @return bool
+	 */
+	public function is_feature_enabled() {
+		return Content_Gate::is_newspack_feature_enabled();
+	}
+
+	/**
+	 * Get the name for this wizard.
+	 *
+	 * @return string The wizard name.
+	 */
+	public function get_name() {
+		return esc_html__( 'Audience Management / Subscribers', 'newspack-plugin' );
+	}
+
+	/**
+	 * Add the Subscribers subpage under the Audience menu.
+	 */
+	public function add_page() {
+		if ( ! $this->is_feature_enabled() ) {
+			return;
+		}
+		add_submenu_page(
+			$this->parent_slug,
+			$this->get_name(),
+			esc_html__( 'Subscribers', 'newspack-plugin' ),
+			$this->capability,
+			$this->slug,
+			[ $this, 'render_wizard' ]
+		);
+	}
+
+	/**
+	 * Register REST endpoints.
+	 */
+	public function register_api_endpoints() {
+		if ( ! $this->is_feature_enabled() ) {
+			return;
+		}
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/avatars',
+			[
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'callback'            => [ $this, 'api_get_avatars' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
+					'emails' => [
+						'type'     => 'array',
+						'required' => true,
+						'items'    => [ 'type' => 'string' ],
+					],
+					'size'   => [
+						'type'              => 'integer',
+						'default'           => 64,
+						'minimum'           => 16,
+						'maximum'           => 512,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+				],
+			]
+		);
+
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/groups',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ $this, 'api_get_groups' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+			]
+		);
+
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/subscribers',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ $this, 'api_get_subscribers' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
+					'page'     => [
+						'type'              => 'integer',
+						'default'           => 1,
+						'minimum'           => 1,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'per_page' => [
+						'type'              => 'integer',
+						'default'           => 20,
+						'minimum'           => 1,
+						'maximum'           => 100,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'search'   => [
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'orderby'  => [
+						'type'              => 'string',
+						'enum'              => [ 'name', 'memberSince' ],
+						'default'           => 'memberSince',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'order'    => [
+						'type'              => 'string',
+						'enum'              => [ 'asc', 'desc' ],
+						'default'           => 'desc',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'status'   => [
+						'type'              => 'array',
+						'items'             => [
+							'type' => 'string',
+							'enum' => [ 'active', 'pending', 'on-hold', 'cancelled' ],
+						],
+						'sanitize_callback' => 'rest_sanitize_request_arg',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'plan'     => [
+						'type'              => 'array',
+						'items'             => [ 'type' => 'string' ],
+						'sanitize_callback' => 'rest_sanitize_request_arg',
+					],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Map a WooCommerce Subscriptions status onto the prototype's status
+	 * vocabulary (active / pending / on-hold / cancelled).
+	 *
+	 * The admin UI was designed against these four values; WCS carries a handful
+	 * more that collapse cleanly onto them:
+	 *   - active, pending-cancel → active (still entitled)
+	 *   - pending               → pending (awaiting first payment)
+	 *   - on-hold               → on-hold (failed/paused renewal)
+	 *   - cancelled, expired    → cancelled (no longer entitled)
+	 * Anything unrecognised is treated as on-hold, the safe "needs attention"
+	 * bucket, rather than surfacing a raw WCS slug the UI has no label for.
+	 *
+	 * @param string $wcs_status The WooCommerce Subscriptions status slug.
+	 *
+	 * @return string One of 'active' | 'pending' | 'on-hold' | 'cancelled'.
+	 */
+	public static function map_subscription_status( $wcs_status ) {
+		return match ( $wcs_status ) {
+			'active', 'pending-cancel' => 'active',
+			'pending'                  => 'pending',
+			'cancelled', 'expired'     => 'cancelled',
+			default                    => 'on-hold',
+		};
+	}
+
+	/**
+	 * GET the site's group subscriptions, hydrated for the admin group list.
+	 *
+	 * Returns the `{ items, total, pages }` envelope the wizard's data hooks
+	 * expect. Groups are returned in full (the group list paginates client-side),
+	 * so `pages` is always 1. Each group carries the owner-inclusive member count
+	 * and the seat limit so the list can render "used / limit" without a second
+	 * round-trip.
+	 *
+	 * @return \WP_REST_Response
+	 */
+	public function api_get_groups() {
+		$this->group_subscriptions_cache = null;
+		$this->group_membership_index    = null;
+		$items                           = [];
+		foreach ( $this->get_group_subscriptions() as $group ) {
+			$items[] = $this->prepare_group( $group['subscription'], $group['settings'] );
+		}
+
+		return rest_ensure_response(
+			[
+				'items' => $items,
+				'total' => count( $items ),
+				'pages' => 1,
+			]
+		);
+	}
+
+	/**
+	 * Every group-enabled subscription on the site, each paired with its resolved
+	 * settings, keyed by subscription ID. Memoized for the request.
+	 *
+	 * The HPOS-safe site-wide query get_group_subscription_ids() can over-report
+	 * under product inheritance, so each candidate is re-checked against its own
+	 * settings — the authority on group membership.
+	 *
+	 * @return array<int,array{subscription:\WC_Subscription,settings:array}>
+	 */
+	private function get_group_subscriptions() {
+		if ( null !== $this->group_subscriptions_cache ) {
+			return $this->group_subscriptions_cache;
+		}
+		$groups = [];
+		if ( class_exists( '\Newspack\Group_Subscription_Settings' ) && function_exists( 'wcs_get_subscription' ) ) {
+			foreach ( Group_Subscription_Settings::get_group_subscription_ids() as $subscription_id ) {
+				$subscription = \wcs_get_subscription( $subscription_id );
+				if ( ! $subscription ) {
+					continue;
+				}
+				$settings = Group_Subscription_Settings::get_subscription_settings( $subscription );
+				if ( empty( $settings['enabled'] ) ) {
+					continue;
+				}
+				$groups[ (int) $subscription->get_id() ] = [
+					'subscription' => $subscription,
+					'settings'     => $settings,
+				];
+			}
+		}
+		$this->group_subscriptions_cache = $groups;
+		return $groups;
+	}
+
+	/**
+	 * Shape a group subscription for the admin group list.
+	 *
+	 * @param \WC_Subscription $subscription The group subscription.
+	 * @param array            $settings     Its resolved group settings (name + limit).
+	 *
+	 * @return array
+	 */
+	private function prepare_group( $subscription, $settings ) {
+		$owner_id   = (int) $subscription->get_user_id();
+		$owner      = $owner_id ? get_userdata( $owner_id ) : false;
+		$created    = $subscription->get_date_created();
+		$created_at = $created ? gmdate( 'Y-m-d', $created->getTimestamp() ) : null;
+
+		return [
+			'id'          => (int) $subscription->get_id(),
+			'ownerId'     => $owner_id,
+			'owner'       => $owner ? [
+				'id'    => $owner_id,
+				'name'  => $owner->display_name,
+				'email' => $owner->user_email,
+			] : null,
+			'plan'        => (string) $settings['name'],
+			'status'      => self::map_subscription_status( $subscription->get_status() ),
+			// The configured limit is owner-inclusive (0 = unlimited); the member
+			// count is likewise owner-inclusive, so "members / seatLimit" reads true.
+			'seatLimit'   => (int) $settings['limit'],
+			'members'     => Group_Subscription::get_member_count( $subscription ),
+			'createdAt'   => $created_at,
+			// Interim click-through target: the WooCommerce subscription edit
+			// screen (HPOS-safe), until the in-wizard group detail lands (PR 4).
+			'editUrl'     => method_exists( $subscription, 'get_edit_order_url' ) ? $subscription->get_edit_order_url() : '',
+			// Seat requests are surfaced in a later slice (NPPD-1753 PR 7).
+			'seatRequest' => null,
+		];
+	}
+
+	/**
+	 * GET a paginated page of subscribers (reader-role users), hydrated for the
+	 * admin subscriber list.
+	 *
+	 * The list is server-paginated: filtering, sorting and paging all run against
+	 * the database rather than a client-side array. Returns the
+	 * `{ items, total, pages }` envelope the wizard's data hooks expect.
+	 *
+	 * Subscription-status and plan filters can't be expressed as user-table
+	 * columns, so they run inverted: an HPOS-safe subscription query resolves the
+	 * matching customer IDs, which are then intersected into the user query's
+	 * `include` set. See resolve_filter_include().
+	 *
+	 * @param \WP_REST_Request $request The request.
+	 *
+	 * @return \WP_REST_Response
+	 */
+	public function api_get_subscribers( $request ) {
+		$this->group_subscriptions_cache = null;
+		$this->group_membership_index    = null;
+		$per_page                        = (int) $request->get_param( 'per_page' );
+		$page     = (int) $request->get_param( 'page' );
+		$search   = trim( (string) $request->get_param( 'search' ) );
+
+		$query_args = [
+			'role__in'    => Reader_Activation::get_reader_roles(),
+			'number'      => $per_page,
+			'paged'       => $page,
+			'orderby'     => 'name' === $request->get_param( 'orderby' ) ? 'display_name' : 'registered',
+			'order'       => 'asc' === $request->get_param( 'order' ) ? 'ASC' : 'DESC',
+			'count_total' => true,
+			'fields'      => 'all',
+		];
+
+		if ( '' !== $search ) {
+			$query_args['search']         = '*' . $search . '*';
+			$query_args['search_columns'] = [ 'user_login', 'user_email', 'display_name' ];
+		}
+
+		// Invert subscription-status / plan filters into an `include` set. Tri-state:
+		// null = no filter applied; [] = filter active but matched nobody (short-circuit
+		// to an empty page); a populated list = restrict the query to those users.
+		$include = $this->resolve_filter_include( $request );
+		if ( is_array( $include ) ) {
+			if ( empty( $include ) ) {
+				return rest_ensure_response(
+					[
+						'items' => [],
+						'total' => 0,
+						'pages' => 0,
+					]
+				);
+			}
+			$query_args['include'] = $include;
+		}
+
+		$user_query = new \WP_User_Query( $query_args );
+		$total      = (int) $user_query->get_total();
+
+		$items = [];
+		foreach ( $user_query->get_results() as $user ) {
+			$items[] = $this->prepare_subscriber( $user );
+		}
+
+		return rest_ensure_response(
+			[
+				'items' => $items,
+				'total' => $total,
+				'pages' => (int) ceil( $total / $per_page ),
+			]
+		);
+	}
+
+	/**
+	 * Shape a reader user for the admin subscriber list.
+	 *
+	 * @param \WP_User $user The reader user.
+	 *
+	 * @return array
+	 */
+	private function prepare_subscriber( $user ) {
+		$user_id = (int) $user->ID;
+		// Resolve the user's own subscriptions once and reuse them, so hydration
+		// doesn't fetch the same list twice per row.
+		$owned_subscriptions = function_exists( 'wcs_get_users_subscriptions' ) ? \wcs_get_users_subscriptions( $user_id ) : [];
+		$subscriptions       = $this->get_individual_subscriptions( $user_id, $owned_subscriptions );
+		$groups              = $this->get_group_memberships( $user_id );
+
+		// user_registered can be a zero date ('0000-00-00 …'), which is truthy but
+		// unparseable; guard on the parsed timestamp so it degrades to null, not 1970.
+		$registered = $user->user_registered ? strtotime( $user->user_registered ) : false;
+
+		return [
+			'id'            => $user_id,
+			'name'          => $user->display_name,
+			'email'         => $user->user_email,
+			// Interim click-through target: the native user-edit screen (self edits
+			// resolve to profile.php), until the in-wizard person profile lands (PR 5).
+			'editUrl'       => get_edit_user_link( $user_id ),
+			'status'        => $this->reduced_status( $subscriptions, $groups ),
+			'memberSince'   => $registered ? gmdate( 'Y-m-d', $registered ) : null,
+			'lastPayment'   => $this->last_payment_date( $user_id ),
+			// Wired to reader activity in a later slice; the column is hidden by default.
+			'lastSeen'      => null,
+			'subscriptions' => $subscriptions,
+			'groups'        => $groups,
+			// Tags and newsletters are populated in a later slice (NPPD-1753 PR 7).
+			'tags'          => [],
+			'newsletters'   => [],
+		];
+	}
+
+	/**
+	 * A reader's own (non-group) subscriptions, shaped as { id, plan, status }.
+	 *
+	 * The wcs_get_users_subscriptions() list is filtered to also surface subs the
+	 * user is only a member of; those are excluded here (they belong to
+	 * get_group_memberships()) by keeping only subs the user actually owns and
+	 * that are not group-enabled.
+	 *
+	 * @param int                $user_id      The reader user ID.
+	 * @param \WC_Subscription[] $owned_subscriptions The user's subscriptions, already fetched by the caller.
+	 *
+	 * @return array<int,array>
+	 */
+	private function get_individual_subscriptions( $user_id, $owned_subscriptions ) {
+		$out = [];
+		foreach ( $owned_subscriptions as $subscription ) {
+			if ( ! $subscription instanceof \WC_Subscription || (int) $subscription->get_customer_id() !== $user_id ) {
+				continue;
+			}
+			$settings = Group_Subscription_Settings::get_subscription_settings( $subscription );
+			if ( ! empty( $settings['enabled'] ) ) {
+				continue; // Group subscriptions are surfaced via get_group_memberships().
+			}
+			$out[] = [
+				'id'     => (int) $subscription->get_id(),
+				'plan'   => $this->individual_plan_name( $subscription ),
+				'status' => self::map_subscription_status( $subscription->get_status() ),
+			];
+		}
+		return $out;
+	}
+
+	/**
+	 * Resolve the display name of an individual subscription's plan (its product name).
+	 *
+	 * @param \WC_Subscription $subscription The subscription.
+	 *
+	 * @return string The product name, or '' when it can't be resolved.
+	 */
+	private function individual_plan_name( $subscription ) {
+		if ( ! class_exists( '\Newspack\WooCommerce_Subscriptions' ) || ! function_exists( 'wc_get_product' ) ) {
+			return '';
+		}
+		$product_id = WooCommerce_Subscriptions::get_subscription_product_id( $subscription );
+		$product    = $product_id ? \wc_get_product( $product_id ) : false;
+		return $product ? (string) $product->get_name() : '';
+	}
+
+	/**
+	 * A reader's group memberships, shaped as { id, plan, status, role }.
+	 *
+	 * Read from the per-request membership index rather than re-querying the
+	 * user's owned/managed/member subscriptions per row.
+	 *
+	 * @param int $user_id The reader user ID.
+	 *
+	 * @return array<int,array>
+	 */
+	private function get_group_memberships( $user_id ) {
+		$index = $this->get_group_membership_index();
+		return $index[ $user_id ] ?? [];
+	}
+
+	/**
+	 * Build (once per request) a map of user ID → their group memberships, each
+	 * shaped as { id, plan, status, role }.
+	 *
+	 * Iterating the site's (few, memoized) groups and reading each group's people
+	 * once is far cheaper than resolving every subscriber's owned/managed/member
+	 * subscriptions individually while paginating a large reader list. Role
+	 * precedence matches the per-group display: owner (the subscription customer),
+	 * else manager (in get_managers), else member.
+	 *
+	 * @return array<int,array<int,array>> User ID → list of membership entries.
+	 */
+	private function get_group_membership_index() {
+		if ( null !== $this->group_membership_index ) {
+			return $this->group_membership_index;
+		}
+		$index = [];
+		if ( class_exists( '\Newspack\Group_Subscription' ) ) {
+			foreach ( $this->get_group_subscriptions() as $group ) {
+				$subscription = $group['subscription'];
+				$owner_id     = (int) $subscription->get_user_id();
+				$managers     = array_map( 'intval', Group_Subscription::get_managers( $subscription ) );
+				$entry        = [
+					'id'     => (int) $subscription->get_id(),
+					'plan'   => (string) $group['settings']['name'],
+					'status' => self::map_subscription_status( $subscription->get_status() ),
+				];
+				foreach ( array_map( 'intval', Group_Subscription::get_all_members( $subscription ) ) as $member_id ) {
+					if ( $member_id === $owner_id ) {
+						$role = 'owner';
+					} elseif ( in_array( $member_id, $managers, true ) ) {
+						$role = 'manager';
+					} else {
+						$role = 'member';
+					}
+					$index[ $member_id ][] = array_merge( $entry, [ 'role' => $role ] );
+				}
+			}
+		}
+		$this->group_membership_index = $index;
+		return $index;
+	}
+
+	/**
+	 * Reduce a reader's many subscription statuses to a single subscriber-level
+	 * status, mirroring the list's badge logic: active-first, with cancelled
+	 * dropped whenever any live status remains. Empty when they have no
+	 * subscription at all (a free reader shows no status badge).
+	 *
+	 * @param array $subscriptions Individual subscription entries.
+	 * @param array $groups        Group membership entries.
+	 *
+	 * @return string One of 'active' | 'pending' | 'on-hold' | 'cancelled' | ''.
+	 */
+	private function reduced_status( $subscriptions, $groups ) {
+		$statuses = array_values(
+			array_unique(
+				array_filter(
+					array_merge(
+						array_column( $subscriptions, 'status' ),
+						array_column( $groups, 'status' )
+					)
+				)
+			)
+		);
+		if ( empty( $statuses ) ) {
+			return '';
+		}
+		$rank = [
+			'active'    => 0,
+			'pending'   => 1,
+			'on-hold'   => 2,
+			'cancelled' => 3,
+		];
+		$live = array_filter( $statuses, fn( $status ) => 'cancelled' !== $status );
+		if ( ! empty( $live ) ) {
+			$statuses = $live;
+		}
+		usort( $statuses, fn( $a, $b ) => ( $rank[ $a ] ?? 99 ) - ( $rank[ $b ] ?? 99 ) );
+		return reset( $statuses );
+	}
+
+	/**
+	 * The date of a reader's most recent completed/processing order, or null.
+	 *
+	 * @param int $user_id The reader user ID.
+	 *
+	 * @return string|null 'YYYY-MM-DD', or null when they have no paid order.
+	 */
+	private function last_payment_date( $user_id ) {
+		if ( ! function_exists( 'wc_get_orders' ) ) {
+			return null;
+		}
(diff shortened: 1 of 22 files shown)


plugins/newspack-plugin/src/wizards/subscribers/labels.js
```diff
@@ -0,0 +1,25 @@
...
+ * enqueue. Falls back to the default "Group" / "Groups" when unset.
+ */
+
+import { __ } from '@wordpress/i18n';
+
+const cfg = ( typeof window !== 'undefined' && window.newspackSubscribers ) || {};
+
+export const GROUP_LABEL = cfg.groupLabel || __( 'Group', 'newspack-plugin' );
+export const GROUP_LABEL_PLURAL = cfg.groupLabelPlural || __( 'Groups', 'newspack-plugin' );
+export const GROUP_LABEL_PLURAL_LOWER = GROUP_LABEL_PLURAL.toLowerCase();
```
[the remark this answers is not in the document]
adekbadek: Removed the lowercasing entirely rather than working around it. `GROUP_LABEL_PLURAL_LOWER` had exactly one consumer (the aria-label you flagged), so the aria-label now uses `GROUP_LABEL_PLURAL` as-is and the export is gone — no locale-unsafe case transform left in the module.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupList.jsx
```diff
@@ -0,0 +1,224 @@
...
+				label: __( 'Members', 'newspack-plugin' ),
+				// The endpoint returns the owner-inclusive member count directly.
+				getValue: ( { item } ) => item.members,
+				render: ( { item } ) => (
+					<span>
+						{ item.members } / { item.seatLimit }
+					</span>
+				),
+				enableSorting: true,
+			},
```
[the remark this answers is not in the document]
adekbadek: Good catch — confirmed against the backend: `prepare_group()` returns `seatLimit` as the raw `(int) $settings['limit']`, and `get_member_capacity()` treats `limit <= 0` as uncapped (returns null), so an unlimited group really did render as "5 / 0".

The Members column now renders `5 / Unlimited` (translatable) when `seatLimit` is 0, and `5 / 20` otherwise. Sorting is unaffected — `getValue` returns the member count, not the ratio.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupList.jsx
```diff
@@ -0,0 +1,228 @@
...
+const DEFAULT_VIEW = {
+	type: 'table',
+	page: 1,
+	perPage: 20,
+	sort: { field: 'createdAt', direction: 'desc' },
+	search: '',
+	fields: [ 'members', 'status', 'createdAt' ],
+	// Hide cancelled groups by default: they add noise with little value. Still
+	// reachable by ticking "Cancelled" in the Status filter (or clearing it).
+	filters: [ { field: 'status', operator: 'isAny', value: [ 'active', 'on-hold' ] } ],
```
dkoo: **SUGGESTION: Pending group subscriptions are hidden by default and render an unlabeled badge**

`DEFAULT_VIEW` filters the group list to `filters: [ { field: 'status', operator: 'isAny', value: [ 'active', 'on-hold' ] } ]` — which excludes both `cancelled` (intended) and `pending` (apparently not). Paired with `status.js`, whose `GROUP_STATUS_LABELS` deliberately omits `pending` on the premise that *"Group subscriptions never sit in the 'pending' (awaiting-first-payment) bucket"*, this produces two effects for any group awaiting first payment: it's hidden from the default view, and if the admin adjusts the Status filter to include it, the badge renders with no label (no `pending` key).

That premise is empirically false: I created a group subscription left in `pending` (awaiting first payment) and the `/groups` endpoint returned it as `status: pending` (via `map_subscription_status()`, which maps `pending` → `pending`). A freshly-purchased or admin-created group sits in `pending` until first payment clears. Either include `pending` in `GROUP_STATUS_LABELS`/`GROUP_STATUS_BADGE_LEVEL` and the default filter, or intentionally collapse a pending group onto a displayed status — but the current "never pending" assumption doesn't hold.

**Manual QA independently reproduced this:** seeding a group subscription with a `pending` status, it isn't visible in the Groups tab unless all status filters are cleared, and its Status column then renders empty instead of showing "Pending".

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+	 *
+	 * @param \WP_User $user The reader user.
+	 *
+	 * @return array
+	 */
+	private function prepare_subscriber( $user ) {
+		$user_id = (int) $user->ID;
+		// Resolve the user's own subscriptions once and reuse them, so hydration
+		// doesn't fetch the same list twice per row.
+		$owned_subscriptions = function_exists( 'wcs_get_users_subscriptions' ) ? \wcs_get_users_subscriptions( $user_id ) : [];
```
dkoo, line 508: **SUGGESTION: Per-row N+1 WooCommerce queries in prepare_subscriber()**

`prepare_subscriber()` runs once per row in the paginated result, and each call issues two per-row WC queries: `wcs_get_users_subscriptions( $user_id )` (line 430) and, via `last_payment_date()`, `wc_get_orders( [ 'customer_id' => $user_id, 'limit' => 1, ... ] )` (line 614). Group memberships are correctly indexed once per request (`get_group_membership_index`), but these two are not — so a page of up to `per_page` (max 100) subscribers fans out into ~200 uncached WC/HPOS queries on every list load, sort, page change, or search keystroke. It's bounded, so not a blocker, but `last_payment_date` in particular is a clean candidate for a single batched `wc_get_orders` over the page's customer-ID set, grouped in PHP — mirroring the group-index approach already used here. Worth at least a comment acknowledging the cost scales with `per_page`.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+			return [];
+		}
+		$ids = [];
+
+		// Individual + owned subscriptions in a matching WCS status (HPOS-safe).
+		// Bounded to the same ceiling the include set is capped at, so the filter
+		// path can't load an unbounded number of subscription objects on a large store.
+		$wcs_statuses = $this->wcs_statuses_for( $prototype_statuses );
+		if ( ! empty( $wcs_statuses ) && function_exists( 'wcs_get_subscriptions' ) ) {
+			$subs = \wcs_get_subscriptions(
```
dkoo: **SUGGESTION: Filter resolution hydrates up to 10k full subscription objects just to read customer IDs**

`customer_ids_for_raw_statuses()` calls `wcs_get_subscriptions( [ 'subscriptions_per_page' => self::FILTER_INCLUDE_CAP /* 10000 */, 'subscription_status' => $wcs_statuses ] )` and then loops only to read `$subscription->get_customer_id()`. `FILTER_INCLUDE_CAP` bounds the count, but `wcs_get_subscriptions` returns fully hydrated `WC_Subscription` objects, so a large store instantiates up to 10k objects per filtered request purely for their customer IDs — a real memory/time cost on the admin request path. The same pattern recurs in `customer_ids_for_plans()` (`wcs_get_subscription` in a per-ID loop, line 774). If WCS exposes an IDs-only / customer-column query (or a lighter HPOS query), prefer it; otherwise a smaller cap for the status-scan path would reduce the hydration blast radius. Scale-hardening, not a correctness gap.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+	 * The live-subscription scan backing that exclusion is itself bounded by
+	 * FILTER_INCLUDE_CAP, so on a store past that ceiling a reader whose only live
+	 * plan sits beyond the cap could slip into the Cancelled set — an accepted
+	 * edge at a scale where exact filtering matters less than a bounded query.
+	 *
+	 * @param string[] $prototype_statuses Prototype statuses (active/pending/on-hold/cancelled).
+	 *
+	 * @return int[]
+	 */
+	private function customer_ids_for_statuses( array $prototype_statuses ) {
```
dkoo, line 783: **SUGGESTION: Redundant raw-status scans when the cancelled filter is active**

`customer_ids_for_statuses()` calls `customer_ids_for_raw_statuses()` up to three times when the filter includes `cancelled` — once for the non-cancelled statuses (line 689), once for `cancelled` (693), and once for the live set `['active','pending','on-hold']` (694). Each call issues a fresh `wcs_get_subscriptions( subscriptions_per_page => 10000, ... )` hydrating up to 10k objects, and a filter like `['active','cancelled']` scans `active` twice. Memoizing `customer_ids_for_raw_statuses()` by its status-set (the groups are already memoized, but the WCS scan is not) would cut the repeated large scans. Bounded, so a suggestion — but it compounds with the hydration cost above.

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscribers.js
```diff
@@ -0,0 +1,99 @@
...
+				if ( cancelled ) {
+					return;
+				}
+				setResult( {
+					items: response?.items || [],
+					total: response?.total || 0,
+					pages: response?.pages || 0,
+				} );
+			} )
+			.catch( () => {
```
dkoo: **SUGGESTION: A failed fetch is indistinguishable from an empty result — no error state or retry**

Both `use-subscribers.js` (line 83) and `use-groups.js` (line 35) `.catch()` a failed request into an empty `{ items: [], total: 0, pages: 0 }` envelope, and the screens then render the normal empty DataViews. A transient REST failure (a WCS fatal on some store, a timeout) therefore looks identical to "this site has no subscribers" — no error banner, no retry affordance. For a read-only admin surface whose whole job is to reflect reality, that's a real diagnosability gap; surfacing an error state distinct from the empty state would prevent an admin mis-reading a fetch failure as an empty audience.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+						aria-label={ sprintf( __( '%s subscribers total', 'newspack-plugin' ), total.toLocaleString() ) }
+					>
+						{ `(${ total.toLocaleString() })` }
+					</span>
+				</>
+			),
+		} );
+	}, [ setHeaderData, total ] );
+
+	if ( subscribersLoading || avatarsLoading ) {
```
dkoo: **SUGGESTION: Whole table blanks to a centered spinner on every page/sort/filter change**

The list is gated behind `if ( subscribersLoading || avatarsLoading )` (line ~307). Because `avatarsLoading` flips back to `true` whenever the email set changes (`use-avatars.js` `setLoading( true )` on each new key), every pagination/sort/filter interaction replaces the full table with `<Waiting isCenter />` until avatars re-resolve — a heavy full-page flash to avoid a comparatively minor avatar flash-in. Rendering the table with avatar *placeholders* (the `.newspack-subscribers__avatar` fallback already exists) and letting avatars fill in asynchronously would be a smoother trade.

plugins/newspack-plugin/src/wizards/subscribers/data/use-avatars.js
```diff
@@ -0,0 +1,98 @@
...
+		}
+		let cancelled = false;
+		// Reset so a previous result doesn't linger across hops, and keep the
+		// spinner up until this set resolves (no flash-in).
+		setLoading( true );
+		setAvatars( {} );
+		// Batch to the endpoint's per-request cap and merge, so a set larger than
+		// one batch still resolves fully instead of dropping the overflow.
+		const batches = [];
+		for ( let i = 0; i < list.length; i += AVATAR_BATCH_SIZE ) {
```
dkoo: **SUGGESTION: The most intricate client hook (use-avatars) has no test**

`use-groups.js` and `use-subscribers.js` each ship a `.test.js`, but `use-avatars.js` — materially more complex — has none. Its untested logic includes the batching loop (`for ( let i = 0; i < list.length; i += AVATAR_BATCH_SIZE )`), the multi-response merge (`responses.forEach( response => { if ( response?.avatars ) Object.assign( merged, response.avatars ); } )`), and the server-authority short-circuit (`if ( responses.some( response => response?.show === false ) )`). The hook exists specifically to batch past the endpoint's 200-email cap, so the batch-and-merge path is exactly what a test should pin.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+	 * status, mirroring the list's badge logic: active-first, with cancelled
+	 * dropped whenever any live status remains. Empty when they have no
+	 * subscription at all (a free reader shows no status badge).
+	 *
+	 * @param array $subscriptions Individual subscription entries.
+	 * @param array $groups        Group membership entries.
+	 *
+	 * @return string One of 'active' | 'pending' | 'on-hold' | 'cancelled' | ''.
+	 */
+	private function reduced_status( $subscriptions, $groups ) {
```
dkoo, line 671: **SUGGESTION: Status-reduction rule is duplicated across the PHP/JS boundary**

The "cancelled is dropped whenever a live status remains" invariant is encoded in four places across two languages: `reduced_status()` (line 575) and `customer_ids_for_statuses()` (line 686) in PHP, and `displayStatuses()` (`status.js`) plus `visiblePlanEntries` (`SubscriberList.jsx`) in JS. The endpoint's `cancelled` *filter*, the row's single `status` field, and the badge *display* must all stay in lockstep, but the rule has no shared anchor — a future change (e.g. adding a `pending` nuance, which the Codex finding shows is already relevant) has to land in all four or the filter will contradict the badge. The code partially cross-references; a single documented source-of-truth note (or a shared constant/comment pointing each copy at the others) would make the invariant discoverable. Lowest-effort mitigation for the highest-drift-risk area in the slice.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+				render: ( { item } ) => <div>{ ( item.newsletters || [] ).join( ', ' ) }</div>,
+			},
+		],
+		[ avatars ]
+	);
+
+	// DataViews only makes the title cell clickable; delegate clicks from the
+	// rest of the row to the same target, ignoring genuinely interactive elements
+	// (the title button, selection checkbox, links).
+	const onRowClick = event => {
```
dkoo, line 297: **SUGGESTION: Normalize row link targets across both tabs (manual QA)**

From manual QA — a UX consistency suggestion, non-blocking. Today the whole row click-through resolves to a single `editUrl` (the native user/subscription edit screen). Consider normalizing the clickable targets across both the Subscribers and Groups tabs so distinct affordances lead to distinct destinations: the subscriber / group-owner **name** links to that user's profile, while the subscription name in the Subscription column (Subscribers tab) and the group name under the owner (Groups tab) link to that specific subscription's edit page. This is interim behavior until the in-wizard detail views land (later slices), so it can be tuned then — but making the link targets predictable now would help.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+
+	if ( subscribersLoading || avatarsLoading ) {
+		return (
+			<div className="newspack-subscribers__loading">
+				<Waiting isCenter />
+			</div>
+		);
+	}
+
+	return (
```
dkoo, line 350: **SUGGESTION: Consider a per-row "Actions" dropdown (manual QA)**

From manual QA — a UX suggestion, non-blocking. Instead of (or alongside) whole-row click-through, consider an "Actions" dropdown menu per row exposing the relevant links: "View subscriber", "View [subscription name]", "View group owner", "View group subscription", etc. The available actions can be tuned per row and expanded in future PRs as the in-wizard detail views land. This also gives keyboard users a discrete, labeled control (see the related row-click accessibility nit).

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+				getValue: ( { item } ) => subscriberStatuses( item, groupEntriesOf( item ) ),
+				render: ( { item } ) => (
+					<HStack spacing={ 2 } justify="flex-start" alignment="center" wrap>
+						{ subscriberStatuses( item, groupEntriesOf( item ) ).map( status => (
+							<Badge key={ status } level={ STATUS_BADGE_LEVEL[ status ] } text={ STATUS_LABELS[ status ] } />
+						) ) }
+					</HStack>
+				),
+			},
+			{
```
dkoo, line 176: **SUGGESTION: Add missing filters/options across both tabs (manual QA)**

From manual QA — feature-completeness suggestions, non-blocking (some are already scoped to later slices in code comments here). Across the two tabs:

- **Subscribers tab — product/plan filter.** The Groups tab already offers a Subscription filter (Add filter → Subscription); the Subscribers tab has none. The `plans` column comment here notes plan filtering waits on the plans endpoint (NPPD-1753 PR 6), so this is a known follow-up — flagging for parity.
- **Subscribers tab — Group role filter.** The `groupRole` column already exists (line ~188) but is hidden and unfilterable. Adding a filter with options Owner / Manager / Member / None (None = non-group subscribers), all selected by default to match the current view, would make it useful.
- **Groups tab — member-count filters.** Consider number-based filters for current and max members (seats).

(The Groups-tab pending status-filter gap is tracked in the pending-group finding above.)

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+	sort: { field: 'memberSince', direction: 'desc' },
+	search: '',
+	fields: [ 'status', 'plans', 'lastPayment', 'memberSince' ],
+	filters: [],
+	layout: {},
+	titleField: 'name',
+};
+
+export default function SubscriberList() {
+	const [ view, setView ] = useState( DEFAULT_VIEW );
```
dkoo, line 94: **SUGGESTION: Persist DataViews settings per tab across refresh/remount (manual QA)**

From manual QA — a future-enhancement suggestion, non-blocking. The DataViews `view` state (column visibility, filters, sorting) is held only in component state (`useState( DEFAULT_VIEW )`), so it resets on every page refresh or view remount. Consider caching each tab's view settings in `localStorage`/`sessionStorage` so a reviewer doesn't lose their column/filter/sort setup. There's precedent for this in the `newspack-story-budget` plugin that could be reused here.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,331 @@
...
+	);
+
+	// DataViews only makes the title cell clickable; delegate clicks from the
+	// rest of the row to the same target, ignoring genuinely interactive elements
+	// (the title button, selection checkbox, links).
+	const onRowClick = event => {
+		if ( event.target.closest( 'a, button, input, label, [role="button"], [role="checkbox"]' ) ) {
+			return;
+		}
+		const row = event.target.closest( 'tbody tr.dataviews-view-table__row' );
```
dkoo, line 301: **NIT: Whole-row click couples to DataViews-internal DOM and isn't keyboard-reachable**

Row click-through (`SubscriberList.jsx:278`, `GroupList.jsx:173`) locates the row via `event.target.closest( 'tbody tr.dataviews-view-table__row' )` and resolves the item by a stamped `data-subscriber-id` / `data-group-id`, coupling the interaction to a DataViews-internal class name — if that markup changes on a DataViews upgrade, click-through breaks silently. Separately, the row wrapper is a `<div onClick>` with `jsx-a11y` rules disabled, so the row itself isn't keyboard-reachable; this degrades acceptably because the DataViews title cell is a real button wired to the same handler, giving keyboard users an equivalent path. Both are reasonable interim choices — worth a `depends on DataViews internal markup` comment so it's grep-able when DataViews is upgraded.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+					'status'   => [
+						'type'              => 'array',
+						'items'             => [
+							'type' => 'string',
+							'enum' => [ 'active', 'pending', 'on-hold', 'cancelled' ],
+						],
+						'sanitize_callback' => 'rest_sanitize_request_arg',
+						'validate_callback' => 'rest_validate_request_arg',
+					],
+					'plan'     => [
```
dkoo, line 238: **NIT: `plan` REST arg lacks a validate_callback (consistency)**

Unlike `status` (which carries both `sanitize_callback` and `validate_callback => rest_validate_request_arg`), the `plan` array arg (lines 210-214) declares only `sanitize_callback => rest_sanitize_request_arg`. Not a security issue — values are re-sanitized via `array_map( 'sanitize_text_field', ... )` in `resolve_filter_include()` and only used as `WP_Query` title matches / configured-name comparisons — purely a consistency nit with the sibling `status` arg.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,932 @@
...
+		}
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/avatars',
+			[
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'callback'            => [ $this, 'api_get_avatars' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
+				'args'                => [
+					'emails' => [
```
dkoo, line 158: **NIT: `/avatars` `emails` arg has no registration-level sanitize/validate (consistency)**

The `emails` array arg (lines 135-139) declares `type`/`items` but no `sanitize_callback`/`validate_callback`, unlike the sibling `size` arg on the same route. Sanitization is instead deferred to `sanitize_email()` inside `api_get_avatars()` (line 865) and the batch is capped at 200 (line 863) — so there's no exposure (route is `manage_options`-gated, `get_avatar_url()` builds Gravatar URLs with no outbound fetch, so no SSRF/enumeration beyond what an admin already has). Registering a `sanitize_callback` on `emails` would just match the `size` arg and keep sanitization declarative.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
```diff
@@ -0,0 +1,543 @@
...
+				'page'     => 3,
+			] 
+		)->get_data();
+		$this->assertCount( 1, $page3['items'] );
+	}
+
+	/**
+	 * Sorting by name is honoured server-side.
+	 */
+	public function test_sorts_by_name() {
```
dkoo, line 243: **NIT: Minor endpoint-test coverage gaps**

`memberSince` is the default sort yet only the `name` sort is asserted (line ~235); the avatar `size` clamping (16–512, registration line 140) and the 200-email `array_slice` cap (line 863) are also unexercised, as is `status.js` `displayStatuses` (which re-encodes the cancelled-reduction the PHP side tests heavily). All low-value, but the `memberSince` ordering being the *default* makes its absence the most notable.

dkoo (approved):
Approving. Reviewed the wizard-shell slice (`fc07463c2..HEAD` — the group-subscription backend it carries is #608, already merged, so I scoped to the net-new files) and exercised it in an isolated env with seeded WooCommerce Subscriptions data. No blockers, and manual QA found none.

Every REST route is `manage_options`-gated through the inherited `api_permissions_check()`, admin-supplied params reach the DB only through `WP_User_Query`/`WP_Query`/`wcs_*` after enum/sanitize validation, and the customer-ID resolution is HPOS-safe and bounded. Runtime checks confirmed the `cancelled`-only-if-fully-churned logic, group-status inheritance, plan/search filters, and both permission gates (403 for non-admin and logged-out); the debug log is clean.

One functional gap worth folding into a follow-up (or here if quick): a `pending` group subscription (awaiting first payment) is real — I created one and `/groups` returned it as `status: pending`, and manual QA reproduced it — but the default group view filters to `['active','on-hold']` (hiding it) and `GROUP_STATUS_LABELS` has no `pending` entry (so its status renders empty). The remaining inline comments are non-blocking: scale-hardening (per-row N+1, up-to-10k subscription hydration on the filter path) and UX/feature-completeness suggestions from manual QA (row link normalization, a per-row actions menu, additional filters, DataViews state persistence).

Housekeeping: the branch still carries #608's merged commits — a rebase onto current `main` would trim the PR diff to just this slice.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupList.jsx
adekbadek: Fixed — the "never pending" premise was wrong, so I dropped it rather than working around it. `GROUP_STATUS_LABELS`/`GROUP_STATUS_BADGE_LEVEL` are gone (they only existed to encode that premise) and the group list now uses the same `STATUS_LABELS`/`STATUS_BADGE_LEVEL` as the subscriber list, so every status the endpoint can return has a label. The default filter is now `['active','pending','on-hold']` — cancelled stays hidden by default (still intended), pending stays visible.

Verified on the review env with your repro: flipped the group subscription to `pending`, and it now shows in the default Groups view with a labeled "Pending" badge (previously: filtered out entirely). Also added `status.test.js`, which pins the reduction rule and asserts every status in the vocabulary carries a label, so a future status can't reintroduce the unlabeled badge.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Took the comment option here, deliberately — flagging that plainly so you can push back.

I tried the batched `wc_get_orders` and couldn't make it both correct and bounded. "Newest paid order per customer" isn't a query WooCommerce can answer for a set of customers in one pass: batching over the page's customer IDs needs either `limit => -1` (peak memory strictly worse than the 100 bounded single-row queries it replaces) or a truncated scan ordered by date, which silently blanks Last payment for any reader whose most recent order falls outside the truncation — wrong data in a column whose whole job is to be true. `wcs_get_users_subscriptions()` has the same shape. The existing convention in the plugin (`Sync\Contact_Metadata\Engagement::get_latest_order`) is the same bounded per-row query.

So instead of a batch I documented the cost on `prepare_subscriber()`: ~2 × `per_page` (≤100) WC queries per page, why each is per-row, and what the batched alternative trades away — so the next person doesn't re-derive this. If you know of a WC/HPOS query that returns newest-paid-per-customer in one hit, I'll happily take it.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Reduced the blast radius without moving the cap. WCS has no IDs-only / customer-column query (`wcs_get_subscriptions()` always returns hydrated objects, and there's no `fields` arg), so the scan in `customer_ids_for_raw_statuses()` now walks the set a page at a time — `subscriptions_per_page => FILTER_SCAN_CHUNK` (500) with `paged`, still stopping at `FILTER_INCLUDE_CAP`. Same results, same query ceiling, but peak memory is one chunk (500 objects) instead of 10k. Combined with the memoization in the sibling thread, a filtered request now hydrates one chunk at a time, once per distinct status set.

I left the cap where it is rather than lowering it for the status path: a smaller cap silently drops matching readers from the results, which is a correctness cost, whereas chunking is free.

`customer_ids_for_plans()` is already one-at-a-time (`wcs_get_subscription` per ID in a loop, with the ID list capped), so each object is freed on the next iteration — peak memory there is one subscription, not the cap.

Extended the `wcs_get_subscriptions` mock to honour `subscriptions_per_page`/`paged` so the paging path is actually exercised by the endpoint tests rather than degenerating to a single pass.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed — `customer_ids_for_raw_statuses()` is now memoized per status set (sorted, comma-joined key) in `$raw_status_ids_cache`, reset per request alongside the two existing memos. A filter of `['active','cancelled']` now scans `active` once instead of twice, and the three calls `customer_ids_for_statuses()` makes collapse to at most three *distinct* scans, each of which is also chunked now (see the hydration thread).

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscribers.js
adekbadek: Fixed in both hooks. `useSubscribers` and `useGroups` now return `error` (the server's message, or a generic fallback) and a `reload` retry, instead of collapsing a failure into an empty envelope. Both screens render a `Notice isError` with the message and a Retry button in place of the empty table, so a WCS fatal or timeout can no longer read as "this site has no subscribers".

Verified on the review env by forcing the subscribers route to return a 500: the list shows "Could not load subscribers: Simulated WooCommerce failure." with a Retry button, and after removing the fault Retry refetches and renders the table without a page reload.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Fixed — took the placeholder trade. Neither screen gates on `avatarsLoading` any more (the flag is no longer read at all): the table renders as soon as the data resolves, each row showing the existing `.newspack-subscribers__avatar" placeholder, and avatars fill in as they arrive. Pagination, sorting and filtering no longer blank the page.

The spinner still covers the initial data fetch in both screens, which is the load that actually has nothing to show.

plugins/newspack-plugin/src/wizards/subscribers/data/use-avatars.js
adekbadek: Added `use-avatars.test.js`, pinning exactly the paths you named:

- the batch-and-merge path — 250 emails fan out into 2 requests (200 + 50) and the merged map holds all 250, asserting merge rather than last-batch-wins (this is the hook's entire reason to exist);
- the server-authority short-circuit — one batch replying `show: false` clears the whole map, even when a sibling batch returned avatars;
- partial failure — a rejected batch keeps the batches that did resolve;
- the no-op path (empty/falsy email set issues no request) and `size` pass-through.

Also added `status.test.js` for `displayStatuses` while I was here (it was the other untested piece of shared logic, and it re-encodes the cancelled-reduction rule).

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Added the anchor. `Subscribers_Wizard::reduced_status()` is now the documented SOURCE OF TRUTH: it states the rule in prose (rank active → pending → on-hold → cancelled; drop cancelled whenever any live subscription remains; no subscription means no status) and names the three copies that must move with it — `customer_ids_for_statuses()`, `displayStatuses()` in `status.js`, and `visiblePlanEntries` in `SubscriberList.jsx`. Each of those three now carries a pointer back, so the invariant is discoverable from any of the four places and grep-able from all of them.

A genuinely shared implementation isn't available across the PHP/JS boundary here (the endpoint has to express the rule as a query, the client as a render), so a documented single statement plus back-pointers is the strongest anchor available short of code-generating one from the other. `status.test.js` now pins the JS copy's behaviour, and the PHP copies are covered by the existing `reduced_status`/cancelled-filter tests, so a one-sided change fails a test rather than only drifting silently.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Added the grep-able comment on both handlers (`SubscriberList.jsx`, `GroupList.jsx`): "DEPENDS ON DATAVIEWS INTERNAL MARKUP", naming `dataviews-view-table__row` as DataViews-owned, the failure mode (whole-row click-through silently stops working), and the instruction to grep that string when bumping `@wordpress/dataviews`. The comment also records why the `<div onClick>` degrades acceptably — the title cell is a real button on the same handler, so keyboard users keep an equivalent path.

I left the mechanism itself alone: DataViews exposes no per-row hook, so locating the row from the event is the only way to extend click-through beyond the title cell today, and both of these are interim targets until the in-wizard detail views land.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Added `'validate_callback' => 'rest_validate_request_arg'` to `plan`, matching the sibling `status` arg.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Made it declarative rather than just symmetrical. `emails` now carries `validate_callback` plus a `sanitize_callback` (`Subscribers_Wizard::sanitize_emails_arg`) that runs `sanitize_email`, drops invalid entries, dedupes, and applies the batch cap — which is the whole of the input handling that used to live inside `api_get_avatars()`, so the callback is now just the `get_avatar_url()` loop.

Two deliberate choices: invalid addresses are dropped rather than rejected (a `format => email` + `rest_validate_request_arg` combination would 400 the whole request over one malformed address and blank a page of avatars, which is worse than skipping it), and the 200 cap is now the named `AVATAR_BATCH_CAP` constant rather than a literal, so the endpoint and the client's `AVATAR_BATCH_SIZE` are visibly the same number.

Added a test covering both bounds: `size` outside 16–512 is rejected with a 400 while 128 is accepted, and a request one email over the cap drops only the overflow.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
adekbadek: Closed all four gaps:

- **`memberSince` ordering** — `test_sorts_by_member_since_by_default()` creates three readers with explicit registration dates and asserts the default (no `orderby`/`order` params) is newest-first, that `order=asc` reverses it, and that `memberSince` hydrates from the same column. `create_reader()` grew an optional `$registered` argument, since fixture users otherwise all register in the same second and the ordering isn't observable.
- **avatar `size` clamping** and **the email cap** — both in `test_avatars_endpoint_bounds_its_inputs()`: 8 and 1024 are rejected with a 400, 128 is accepted, and a request of `AVATAR_BATCH_CAP + 1` emails resolves exactly the cap with the overflow dropped.
- **`displayStatuses`** — new `status.test.js` covering the ordering, dedup, the cancelled-hidden-while-live rule, the fully-churned case, and the fallback/no-badge paths.

Full PHP suite (2027 tests) and JS suite (353 tests) green.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Done — both tabs now follow the rule you described: a person's name goes to that person, a plan name goes to that plan.

Endpoint side, every affordance now carries its own target: a group's `owner` gained `editUrl` (the user-edit screen), and individual subscriptions and group memberships each gained `editUrl` (the HPOS-safe subscription edit screen), all via one `subscription_edit_url()` helper.

UI side, a small shared `SubscriptionLink` (`src/wizards/subscribers/links.jsx`) renders a plan name as a real link to its subscription — used by the Subscribers tab's Subscription column, the Groups tab's Subscription column, and the plan line under the group owner. It stops the click from bubbling so the row delegate can't fire as well, and falls back to plain text when no URL resolves (or the plan has no name, e.g. a deleted product — otherwise that row rendered an empty, invisible link). The Groups tab's row and title click now open the **owner** rather than the subscription, so it matches the Subscribers tab; a group whose owner no longer exists is marked not clickable rather than silently doing nothing.

Verified in the browser on the PR env: owner name → `user-edit.php?user_id=10`, group plan name → subscription #63, subscriber plan name → subscription #65, row click → `user-edit.php?user_id=19`. Full PHPUnit (2027) and jest (353) suites, eslint and phpcs are green; endpoint tests assert the new fields, and the PR description's click-through notes are updated.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Captured as a design ask in [NPPD-2111](https://linear.app/a8c/issue/NPPD-2111/subscribers-lists-filter-persistence-and-row-action-polish).

The proposal there is to fold it into the detail-view design work rather than specify it standalone: the menu's contents depend on what the in-wizard detail views expose (so specifying it before those exist means specifying it twice), and it changes a row affordance ported ~1:1 from the signed-off i2 prototype. Once there is a signed-off shape the implementation is small — DataViews supports row actions natively.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Split across two places:

- **Subscribers-tab plan filter** — unchanged, still scoped to the later slice of [NPPD-1753](https://linear.app/a8c/issue/NPPD-1753/implement-admin-side-subscribers-management) that adds the plans endpoint, as the `plans` column comment says.
- **Group-role filter** and **Groups-tab member-count filters** — items 1 and 2 of [NPPD-2111](https://linear.app/a8c/issue/NPPD-2111/subscribers-lists-filter-persistence-and-row-action-polish).

Worth noting on the role filter: it is backend work, not just a column tweak. The subscribers endpoint accepts only `status` today, so it needs a new query parameter plus the same group-membership → customer-ID inversion (and include-cap handling) the status filter uses.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Tracked as item 3 of [NPPD-2111](https://linear.app/a8c/issue/NPPD-2111/subscribers-lists-filter-persistence-and-row-action-polish), with the `newspack-story-budget` implementation noted there as the prior art to reuse.

adekbadek:
@dkoo Re-requesting review – commits landed after your approval that need a look before this merges:
- fix(subscribers): pending groups + list error state (NPPD-1753, #631)
- fix(subscribers): normalize list click targets (NPPD-1753, #631)
- chore(subscribers): point the group-role comment at NPPD-2111 (#631)

adekbadek:
Housekeeping done — rebased onto current `main`, so the diff is now just this slice.

| | files | insertions |
|---|---|---|
| before | 34 | 4,970 |
| after | 22 | 3,789 |

#608 was squash-merged, so its four original commits on this branch couldn't be matched against the squash and the merge base was still stuck at `3ab5e5756f` (pre-#608). Dropping them (`git rebase --onto origin/main fc07463c2a`) rather than replaying them was the right move for a second reason: `main` has since relocated the group-subscription classes to `includes/plugins/woocommerce-subscriptions/group-subscription/`, so a plain rebase tried to re-apply #608's changes at their old paths and conflicted. The wizard code is namespace-based (`Newspack\Group_Subscription`), so it needed no changes for the move.

Re-verified after the rebase:

- `n test-php` — 2331 tests / 6583 assertions. The 11 errors are pre-existing on `main` (identical set at 2308 tests on a pristine `origin/main` checkout in the same env: Collections, Reader_Segment_Eligibility, Emails, RAS registration). Nothing in the subscribers suites: 23/23 green, 95 assertions.
- jest — 49 suites / 452 tests green, including all four `src/wizards/subscribers` suites.
- ESLint and PHPCS clean.

All review threads survived the force-push and remain resolved.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/631#pullrequestreview-4737618265)

Fixed here rather than in a follow-up. `pending` is now in the group list's default status filter, and `GROUP_STATUS_LABELS`/`GROUP_STATUS_BADGE_LEVEL` are gone: they existed only to drop `pending` from the shared maps, which was the wrong premise, so the group list now uses the same `STATUS_LABELS`/`STATUS_BADGE_LEVEL` as the subscriber list (05448d71a8).

Also added the regression tests that were missing on both sides: `status.test.js` asserts every status a subscription can hold has a non-empty label (an unlabeled badge renders empty, which is exactly the symptom you saw), and the groups endpoint test now pins the whole WCS -> prototype mapping — `pending` -> `pending`, plus `pending-cancel`/`expired`/unknown-slug — instead of the single `cancelled` case it checked before.

Verified in the review env: created a `pending` group subscription, and it shows in the default Groups view with a `Pending` badge (that group is still on https://nppd1753-pr3.test if you want to look). Full `n test-php` is green: 2331 tests, 7 documented skips.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/631#pullrequestreview-4737618265)

Done — the branch is rebased onto `main`. `origin/main` is now an ancestor of HEAD, and `git diff --stat origin/main...HEAD` is just this slice: the wizard class, `src/wizards/subscribers/`, the two endpoint test suites, the WC mock additions and the `.gitignore` anchor fix. #608's commits are no longer in the PR diff.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/feat/nppd1753-wizard-shell` (stash local work first).

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,1044 @@
...
+
+		// Individual + owned subscriptions in a matching WCS status (HPOS-safe).
+		// Bounded to the same ceiling the include set is capped at, so the filter
+		// path can't load an unbounded number of subscription objects on a large
+		// store, and walked a chunk at a time: WooCommerce hands back fully
+		// hydrated WC_Subscription objects and only the customer ID is read here,
+		// so paging keeps peak memory at one chunk instead of the whole cap.
+		$wcs_statuses = $this->wcs_statuses_for( $prototype_statuses );
+		if ( ! empty( $wcs_statuses ) && function_exists( 'wcs_get_subscriptions' ) ) {
+			for ( $page = 1; ( $page - 1 ) * self::FILTER_SCAN_CHUNK < self::FILTER_INCLUDE_CAP; $page++ ) {
```
dkoo, line 837: **BLOCKER: Chunked filter scan pages with `paged`, which wcs_get_subscriptions() ignores**

The new chunked scan pages with `'paged' => $page`, but `wcs_get_subscriptions()` does not honor that argument. `paged` is declared in the function's own `$default_args` (WCS 9.x `wcs-functions.php:475`), so `$extra_args = array_diff_key( $provided_args, $default_args )` strips it, and the `WC_Order_Query` it builds receives only `'limit' => $working_args['limit'] ?? $working_args['subscriptions_per_page']` and `'offset' => $working_args['offset'] > 0 ? $working_args['offset'] : null` (`:522-523`). There is no `paged` key in the query at all.

**Confirmed at runtime, not just by reading:** calling the real function in a live env with `subscriptions_per_page=2` returns the identical rows for `paged=1`, `paged=2` and `paged=3`, while `offset=2` and `offset=4` return correctly different rows.

Consequences on a store with 500+ subscriptions in a matched status: every iteration of `for ( $page = 1; ( $page - 1 ) * self::FILTER_SCAN_CHUNK < self::FILTER_INCLUDE_CAP; $page++ )` re-runs the same page-1 query; `count( $subs )` is therefore always `500`, so `if ( count( $subs ) < self::FILTER_SCAN_CHUNK ) break;` never fires and all 20 iterations run — up to 60 identical queries per request, since the cancelled path resolves three status sets. The effective scan ceiling silently drops from `FILTER_INCLUDE_CAP` (10,000) to 500, so the Active/Pending/On-hold filters under-report, and because `$live_ids` is likewise capped at 500 a reader whose live plan sits past row 500 is wrongly returned by the Cancelled filter — regressing the fully-churned invariant verified in the previous review. Below 500 matching subscriptions behaviour is unchanged, which is why this doesn't show up locally.

`'offset' => ( $page - 1 ) * self::FILTER_SCAN_CHUNK` is honored and expresses the same intent. Note that `tests/mocks/wc-mocks.php` was extended in this same diff to implement `paged`, so the mock masks the defect — it should move to `offset` alongside the fix, or no test can ever fail on this.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupList.jsx
```diff
@@ -0,0 +1,261 @@
...
+										text={
+											item.seatRequest.status === 'awaiting-payment'
+												? __( 'Awaiting payment', 'newspack-plugin' )
+												: __( 'Seat increase requested', 'newspack-plugin' )
+										}
+									/>
+								) }
+							</HStack>
+							<div className="newspack-subscribers__email">
+								<SubscriptionLink href={ item.editUrl }>{ item.plan }</SubscriptionLink>
```
dkoo: **BLOCKER: Subscription-column links don't navigate — the row click swallows them**

**Confirmed in-browser during manual QA:** clicking a subscription/plan link in the Subscription column does not open that subscription — the row-level click-through takes precedence and you land on the person instead. Since normalizing these click targets is this commit's stated purpose, the feature doesn't currently do its job. Worth noting the endpoint side is fine: `editUrl` is populated on every individual-subscription and group-membership entry (verified against a live env), so this is a front-end event/DOM issue, not a missing href.

Two structural facts that are confirmed from the `@wordpress/dataviews` v10 source and are likely where this lives — the `owner` field is the `titleField`, and its render nests a real link inside it (`<SubscriptionLink href={ item.editUrl }>{ item.plan }</SubscriptionLink>`), while `ColumnPrimary` wraps the title field in `ItemClickWrapper`, which for a clickable item renders `<div role="button" tabIndex={0} onClick={…} onKeyDown={…}>`:

1. **Keyboard definitely mis-routes.** The wrapper's handler is `onKeyDown: (event) => { if (event.key === "Enter" || event.key === " ") { event.stopPropagation(); onClickItem(item); } }`, with no `event.target === event.currentTarget` check. `SubscriptionLink` guards only `onClick`, so nothing stops a keydown originating on the link — pressing Enter on the focused plan link fires `openOwner( item )` alongside the browser's default navigation.
2. **ARIA.** Descendants of `role="button"` are treated as presentational, so assistive tech doesn't reliably expose the nested link at all.

The standalone `plan` column (`:139`) sits outside the wrapper and would normally be the mitigation, but `DEFAULT_VIEW.fields` is `[ 'members', 'status', 'createdAt' ]` — `plan` is hidden by default, so the title cell is the only place a group's subscription is reachable out of the box.

Suggested direction: add `onKeyDown={ event => event.stopPropagation() }` to `SubscriptionLink`'s anchor (harmless in `SubscriberList`, where the link sits in an ordinary cell), and move the link out of the title wrapper — e.g. render the title-cell copy as plain text and add `plan` to the group `DEFAULT_VIEW.fields` — which removes the nesting and the ARIA problem together. Please re-verify the mouse path in a browser after changing it, since that's where the failure was observed.

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscribers.test.js
```diff
@@ -0,0 +1,114 @@
...
+		expect( path ).toContain( '/newspack/v1/wizard/newspack-subscribers/subscribers' );
+		expect( path ).toContain( 'orderby=memberSince' );
+		expect( path ).toContain( 'per_page=20' );
+
+		expect( result.current.items ).toEqual( [ { id: 7 } ] );
+		expect( result.current.total ).toBe( 42 );
+		expect( result.current.pages ).toBe( 3 );
+	} );
+
+	it( 'degrades to an empty page when the request fails', async () => {
```
dkoo: **SUGGESTION: Hook tests weren't updated with the error-state fix, so they can't catch its regression**

`use-subscribers.test.js` and `use-groups.test.js` were not updated alongside the error-state change. Both are still named "degrades to an empty list/page when the request fails" and assert only the empty envelope; neither asserts that `result.current.error` is truthy, and neither exercises `reload()`. A future regression back to `.catch( () => ({ items: [], total: 0, pages: 0 }) )` — precisely the bug these commits fix — would leave both tests green. One `expect( result.current.error ).toBeTruthy()` per file plus a `reload()` re-fetch case would pin the new contract.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
```diff
@@ -0,0 +1,611 @@
...
+		$this->assertContains( $acme_member, $ids );
+		$this->assertNotContains( $beta_member, $ids, 'A Beta-Team member is active but not on the Acme plan, so the AND filter excludes them (plan axis narrows).' );
+		$this->assertNotContains( $cancelled_acme_member, $ids, 'A cancelled Acme member is on the plan but not active, so the AND filter excludes them (status axis narrows).' );
+	}
+
+	/**
+	 * A filter that matches nobody short-circuits to an empty envelope (total 0,
+	 * pages 0) rather than falling through to an unfiltered query.
+	 */
+	public function test_filter_matching_nobody_returns_empty_envelope() {
```
dkoo, line 475: **SUGGESTION: The filter-side status mapping is never asserted positively**

The commit is titled "pin the WCS status mapping incl. pending groups", but only the *display* direction (`map_subscription_status`) is pinned. The filter-side inverse, `wcs_statuses_for()`, is never asserted positively — the sole `pending` filter test asserts absence (`$data = $this->dispatch( [ 'status' => [ 'pending' ] ] )->get_data(); $this->assertSame( [], $data['items'] );`). Nothing fails if someone deletes `'pending' => [ 'pending' ]`, `'switched'`, `'pending-cancel'` or `'expired'` from the `$map`, even though that map's own comment says it exists "to keep display and filter in step". Two short cases would close it: a reader on a `pending` individual subscription (and a member of a `pending` group) returned by `status=['pending']`, and a reader on a `switched` subscription returned by `status=['on-hold']`.

plugins/newspack-plugin/src/wizards/subscribers/data/use-avatars.js
```diff
@@ -0,0 +1,98 @@
...
+import { useEffect, useState } from '@wordpress/element';
+import apiFetch from '@wordpress/api-fetch';
+
+const config = ( typeof window !== 'undefined' && window.newspackSubscribers ) || {};
+
+export const SHOW_AVATARS = config.showAvatars !== false;
+
+// The endpoint caps each request at this many emails; the hook batches larger
+// sets so nothing is silently dropped (e.g. a site with many groups).
+const AVATAR_BATCH_SIZE = 200;
```
dkoo: **SUGGESTION: Avatar batch size is hard-coded in three places and the server truncates silently**

The batch size now lives independently in `AVATAR_BATCH_SIZE = 200` (client), `AVATAR_BATCH_CAP = 200` (server, `class-subscribers-wizard.php:55`), and `BATCH_SIZE = 200` (test). The server's `sanitize_emails_arg()` *silently truncates* via `array_slice( array_unique( $sanitized ), 0, self::AVATAR_BATCH_CAP )` rather than erroring, so if the client constant ever drifts above the server's, the overflow avatars come back blank with no signal anywhere — and the test pins the client's copy, not the agreement between the two. The wizard already localizes config onto `window.newspackSubscribers` (that's where `showAvatars` comes from); pushing `avatarBatchCap` through the same channel would collapse this to a single authority.

plugins/newspack-plugin/src/wizards/subscribers/links.jsx
```diff
@@ -0,0 +1,33 @@
...
+ * goes to that plan. Names are the DataViews title cell, so their target is the
+ * list's `onClickItem`; plan names are ordinary cells and carry their own link.
+ */
+
+/**
+ * Link a plan/subscription name to that subscription's admin edit screen.
+ *
+ * Renders the name unlinked when the endpoint couldn't resolve an edit URL (no
+ * WooCommerce, or an admin without the capability), so the column never shows a
+ * dead link — and likewise when the name itself is empty (a subscription whose
```
dkoo: **NIT: Docblock claims the edit URL is capability-filtered; it isn't**

The docblock says the name renders unlinked "when the endpoint couldn't resolve an edit URL (no WooCommerce, or an admin without the capability)", and the PHP `subscription_edit_url()` docblock says the same. `WC_Order::get_edit_order_url()` performs no capability check — it returns `apply_filters( 'woocommerce_get_edit_order_url', get_admin_url( … ), $this )` unconditionally, so the only real empty case is the `method_exists()` miss (no WooCommerce). Harmless while the route is gated at `manage_options`, but the comment sets a false expectation for anyone who later relaxes that gate.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -0,0 +1,361 @@
...
+		<div className="newspack-subscribers__clickable-rows" onClick={ onRowClick }>
+			<DataViews
+				data={ items }
+				fields={ fields }
+				view={ view }
+				onChangeView={ setView }
+				paginationInfo={ { totalItems: total, totalPages: pages } }
+				defaultLayouts={ { table: {} } }
+				getItemId={ item => item.id }
+				onClickItem={ openSubscriber }
```
dkoo, line 361: **NIT: SubscriberList lacks the isItemClickable guard GroupList has**

`GroupList` guards its row clicks with `isItemClickable={ hasOwnerLink }`, but `SubscriberList` passes `onClickItem={ openSubscriber }` with no `isItemClickable`. `prepare_subscriber()` sets `'editUrl' => get_edit_user_link( $user_id )`, which returns `''` when `! current_user_can( 'edit_user', $user_id )` — reachable on multisite. Those rows get a focusable, clickable title that silently no-ops on `if ( item?.editUrl )`. Mirroring GroupList with `isItemClickable={ item => !! item?.editUrl }` would keep the two tabs consistent.

plugins/newspack-plugin/src/wizards/subscribers/status.js
```diff
@@ -0,0 +1,59 @@
...
+export const displayStatuses = ( statuses, fallback ) => {
+	const present = ( statuses || [] ).filter( Boolean );
+	if ( ! present.length ) {
+		return fallback ? [ fallback ] : [];
+	}
+	let distinct = [ ...new Set( present ) ];
+	if ( distinct.some( s => s !== 'cancelled' ) ) {
+		distinct = distinct.filter( s => s !== 'cancelled' );
+	}
+	return distinct.sort( ( a, b ) => STATUS_RANK[ a ] - STATUS_RANK[ b ] );
```
dkoo: **NIT: JS rank sort lacks the unknown-status guard PHP has**

Now that the PR explicitly claims the four copies of the reduction rule stay in lockstep, one asymmetry is worth closing: PHP guards unknown statuses in its sort (`usort( $statuses, fn( $a, $b ) => ( $rank[ $a ] ?? 99 ) - ( $rank[ $b ] ?? 99 ) )`), while JS does not — `distinct.sort( ( a, b ) => STATUS_RANK[ a ] - STATUS_RANK[ b ] )` yields `NaN` for any status outside the four, giving an implementation-defined order (same at `SubscriberList.jsx:59`). Unreachable today because the server maps everything into the four-value vocabulary, but a `?? 99` would make the parity claim literally true.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -0,0 +1,1044 @@
...
+				$subs = \wcs_get_subscriptions(
+					[
+						'subscriptions_per_page' => self::FILTER_SCAN_CHUNK,
```
dkoo, line 840: **NIT: Chunked scan has no deterministic tiebreak**

Once the paging argument is corrected, the scan will still page with WCS's default `'orderby' => 'start_date'` (mapped to `orderby => 'date'`), which has no deterministic tiebreak. Rows sharing a start date can shuffle between chunk boundaries, silently skipping a customer ID. Passing `'orderby' => 'ID'` makes the walk stable and is cheaper than a date sort.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
```diff
@@ -0,0 +1,611 @@
...
+	/**
+	 * Create a reader user (display name carries the scope token) and track it.
+	 *
+	 * @param string $name       Human name, prefixed with the scope token in display_name.
+	 * @param string $role       Optional role.
+	 * @param string $registered Optional 'Y-m-d H:i:s' registration date. Fixture users
+	 *                           otherwise all register in the same second, which makes
+	 *                           member-since ordering unobservable.
+	 *
+	 * @return int The new user ID.
```
dkoo, line 88: **NIT: Test helper's array_filter can drop a deliberately-empty role**

`wp_insert_user( array_filter( [ … 'role' => $role, 'user_registered' => $registered ] ) )` applies `array_filter`'s default truthiness callback across the whole arg array purely to drop an empty `user_registered`. It would equally drop a deliberately-empty `role` — the documented way to create a role-less user, and exactly the fixture a future reader-role test might want. `$args = [ … ]; if ( $registered ) { $args['user_registered'] = $registered; }` is the same length without the booby trap.

dkoo (changes requested):
Re-reviewed the 4 follow-up commits (`5d5d8c653..6f4e07ddd`). First: thank you — all six findings from my previous review are genuinely fixed, and the fixes are substantive rather than gestural. The `GROUP_STATUS_*` maps being deleted outright (rather than patched) is the right call, the error/retry path is properly cancel-guarded, and the new tests are real behavioural pins. Adding `pending` to the live set in `customer_ids_for_statuses()` is a correctness gain I missed last round: without it, a reader holding a pending plus a cancelled subscription would have matched the Cancelled filter while their badge read Pending.

Two things need another pass.

**1. The chunked filter-scan refactor pages with an argument WooCommerce Subscriptions ignores.** `wcs_get_subscriptions()` does not honor `paged` — it sits in the function's own `$default_args`, so `array_diff_key()` strips it from the extra args, and the `WC_Order_Query` it builds receives only `limit` and `offset`. I confirmed this against the running plugin rather than just reading it: with `subscriptions_per_page=2`, `paged=1`, `paged=2` and `paged=3` all return the identical rows, while `offset=2` and `offset=4` return correctly different ones.

I also reproduced the user-visible symptom end-to-end through the REST endpoint, by shrinking the constants so the chunk boundary was reachable on a small dataset. Same data, same query, only the paging argument differing: with `offset` the Cancelled filter returned just the one genuinely churned reader; with `paged` it also returned a reader whose own status badge read **Active**. So the filter contradicts the badge it filters on. A second reader was wrongly *omitted* from the same filter, because the cancelled-side scan misses rows past chunk 1 too.

On a real store this needs 500+ subscriptions in a matched status: the loop then re-runs the same page-1 query for all 20 iterations (up to 60 queries per request, since the cancelled path resolves three status sets) and the effective scan ceiling silently drops from `FILTER_INCLUDE_CAP` (10,000) to 500. Below that threshold behaviour is unchanged, which is why nothing catches it locally. `'offset' => ( $page - 1 ) * self::FILTER_SCAN_CHUNK` expresses the same intent and is honored.

Worth flagging: `tests/mocks/wc-mocks.php` was extended in this same diff to implement `paged`, so the mock makes the broken path look correct and no test can fail on it. The mock should move to `offset` alongside the fix.

**2. The new Subscription-column links don't actually navigate.** Verified in a browser: clicking a subscription/plan link lands you on the person instead — the row click-through swallows it. Since normalizing these click targets is what this commit set out to do, that's worth sorting before merge. The endpoint side is fine (`editUrl` is populated on every entry), so it's a front-end event/DOM issue. Details and a suggested direction are in the inline comment, including a keyboard mis-routing path in the same component that's confirmed from the DataViews source.

The remaining comments are non-blocking.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in 3aa48e08d — `'offset' => ( $page - 1 ) * self::FILTER_SCAN_CHUNK`, and confirmed against the WCS source in the env: `paged` sits in `$default_args` so `array_diff_key()` strips it, while `'offset' => $working_args['offset'] > 0 ? … : null` is passed straight through to the `WC_Order_Query`.

The mock moved to `offset` in the same commit, with a comment saying `paged` is deliberately not implemented so it can't mask this again. And there's now a test that actually fails on the defect — `test_status_filter_scan_advances_past_the_first_chunk` seeds a full `FILTER_SCAN_CHUNK` of active filler (bare customer IDs, no user rows, so it stays cheap) plus two readers whose live plans can only be found in chunk 2. Verified red before green: on `paged` it fails with *"A reader whose only subscription sits past the first chunk still matches Active. Failed asserting that an array contains 3."*, and the second assertion covers the case you flagged as worse — a reader whose live plan sits past the boundary being wrongly returned by Cancelled.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupList.jsx
adekbadek: Fixed in 3aa48e08d, taking the structural half of your suggestion: the plan link is out of the title cell entirely. The `owner` render now shows the owner's **email** as its secondary line (mirroring the subscriber list's title cell), and `plan` joins `DEFAULT_VIEW.fields`, so the group's subscription is reachable by default from an ordinary cell. That removes the nesting and the ARIA problem together, and it avoids showing the plan name twice in the same row.

Two corrections to the record, both from re-testing in the env rather than reasoning about it:

**The mouse symptom didn't reproduce on a fresh build of the branch tip.** Clicking the nested title-cell link opened the subscription, not the person. Worth checking your env's bundle: `dist/` there may predate `0401e5222` ("normalize list click targets") — mine did, and rebuilding changed the behaviour. Separately, the admin page HTML caches hard in the browser, so a plain reload can keep serving the old `?ver=` and the old JS with it; `&cb=<n>` on the admin URL is what finally showed me the current build.

**The keyboard path was real, and I confirmed it directly.** Suppressing only the anchor's own default (`preventDefault` on its keydown) and pressing Enter navigated to `user-edit.php?user_id=10` — i.e. `onClickItem` had fired alongside the link. So both destinations were being triggered; the anchor's default just happened to win the race. Re-running that same probe after the fix: no navigation at all, so the double-fire is gone.

**I did not add the `onKeyDown` stopPropagation.** With the link outside `ItemClickWrapper` there's no ancestor keydown handler left to guard against, so it would be dead code — `links.jsx` now carries a comment saying why, and that the component must stay out of title cells. (I briefly had it in and thought it broke Enter activation; that turned out to be my own leftover `preventDefault` probe on a page that had never navigated away, not the guard. Flagging it in case the same trap catches you.)

Verified on the current build, each from a fresh cache-busted load: plan link by mouse → subscription; plan link by Enter → subscription (both rows); the suppression probe → no stray navigation; row click and title-cell click → the owner. Subscribers-tab rows still click through with the new `isItemClickable` guard.

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscribers.test.js
adekbadek: Fixed in 3aa48e08d. Both hook tests now assert the contract rather than the empty envelope beside it: renamed to "surfaces the failure instead of passing an empty list/page off as the answer" with `expect( result.current.error ).toBeTruthy()`, plus a new `reload()` case in each file that fails the first fetch, asserts the error, then resolves the retry and asserts the error clears and the good data lands (`toHaveBeenCalledTimes( 2 )`). A regression to `.catch( () => ({ items: [], total: 0, pages: 0 }) )` now fails four assertions.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
adekbadek: Fixed in 3aa48e08d — `test_status_filter_maps_wcs_statuses_positively` pins the filter direction positively, covering every entry in the map: `pending` via an individual subscription *and* via group inheritance, `switched` reachable through the On hold filter, `pending-cancel` through Active, and `expired` through Cancelled. Deleting any one of those from `wcs_statuses_for()`'s `$map` now fails a test instead of silently narrowing the results.

plugins/newspack-plugin/src/wizards/subscribers/data/use-avatars.js
adekbadek: Fixed in 3aa48e08d — collapsed to one authority. The wizard now localizes `avatarBatchCap => self::AVATAR_BATCH_CAP` alongside `showAvatars`, and the hook reads it. It reads at call time rather than import time, so the value can't depend on this module loading after the inline config script; the literal `200` survives only as a fallback for a config that never loaded.

The test pins the agreement rather than the client's copy: it sets `avatarBatchCap: 5` on `window`, passes 12 emails and asserts three requests of 5/5/2. A hook holding its own hard-coded 200 sends one batch and fails.

plugins/newspack-plugin/src/wizards/subscribers/links.jsx
adekbadek: Fixed in 3aa48e08d — you're right, `get_edit_order_url()` returns the filtered URL unconditionally. The docblock now says the empty case in practice means WooCommerce is absent (the `method_exists()` miss), with no claim about capabilities.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
adekbadek: Fixed in 3aa48e08d — `SubscriberList` now passes `isItemClickable={ hasSubscriberLink }`, mirroring `hasOwnerLink` in `GroupList`, with a comment naming the multisite `edit_user` case that makes `get_edit_user_link()` return `''`. Confirmed in the browser that ordinary rows still click through to the person.

plugins/newspack-plugin/src/wizards/subscribers/status.js
adekbadek: Fixed in 3aa48e08d — `status.js` exports `statusRank = status => STATUS_RANK[ status ] ?? 99`, and both comparators use it (`displayStatuses` and `planEntries` in `SubscriberList.jsx`). The parity claim with `reduced_status()`'s `?? 99` is now literally true rather than true-by-luck.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in 3aa48e08d, in the same edit as the paging fix — the scan now passes `'orderby' => 'ID'`. Confirmed it survives WCS's orderby mapping: `start_date` is the only value rewritten (to `date`), and the `default:` branch passes anything else through to the `WC_Order_Query` untouched.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
adekbadek: Fixed in 3aa48e08d — `create_reader()` builds `$args` and adds `user_registered` only when given, exactly as you wrote it. Comment notes that the blanket `array_filter()` would equally drop a deliberately-empty `role`.

adekbadek:
> Re: [this review](https://github.com/Automattic/newspack-workspace/pull/631#pullrequestreview-4768875236)

All ten comments addressed in 3aa48e08d — both blockers, all three suggestions and all four nits. Per-point detail is on the threads; the two headline items:

**1. The chunked filter-scan pages with an argument WooCommerce Subscriptions ignores.** Correct, and confirmed against the WCS source in the env: `paged` sits in `wcs_get_subscriptions()`'s own `$default_args`, so `array_diff_key()` strips it, while `offset` is passed through to the `WC_Order_Query`. Now `'offset' => ( $page - 1 ) * self::FILTER_SCAN_CHUNK`, with `'orderby' => 'ID'` for the deterministic tiebreak you flagged separately. The mock moved to `offset` in the same commit and carries a comment saying `paged` is deliberately unimplemented so it can't mask this again — and `test_status_filter_scan_advances_past_the_first_chunk` now fails outright on the defect (verified red, then green), covering both the under-reporting and the worse case of a live plan past the boundary being reported as churned.

**2. The Subscription-column links don't navigate.** Fixed structurally — the plan link is out of the DataViews title cell entirely. The owner cell's secondary line is now the owner's email (mirroring the subscriber list), and `plan` is in `DEFAULT_VIEW.fields`, so the subscription is reachable by default from an ordinary cell. That drops the nesting and the ARIA problem together, without printing the plan name twice per row.

Two corrections there, both from re-testing rather than reasoning:

- **The mouse symptom didn't reproduce on a fresh build of the branch tip** — the nested link opened the subscription. Your env's `dist/` may predate `0401e5222` ("normalize list click targets"); mine did, and rebuilding changed the behaviour. The admin page HTML also caches hard in the browser, so a plain reload can keep serving the old `?ver=` and the old JS with it — `&cb=<n>` on the admin URL is what surfaced the current build for me.
- **The keyboard path was real and I confirmed it directly.** Suppressing only the anchor's own default and pressing Enter navigated to `user-edit.php?user_id=10`, so `onClickItem` was firing alongside the link — the anchor's default just won the race. Same probe after the fix: no navigation.

I did **not** add the `onKeyDown` stopPropagation. With the link outside `ItemClickWrapper` there's no ancestor keydown handler left to guard against, so it would be dead code; `links.jsx` documents that and that the component must stay out of title cells.

Verification, each from a fresh cache-busted load on `https://nppd1753-pr3.test`: plan link by mouse and by Enter → the subscription (both rows); suppression probe → no stray navigation; row click and title-cell click → the owner; Subscribers-tab rows still click through with the new `isItemClickable` guard.

- `n test-php` — 2333 tests, 6619 assertions, 7 documented skips, no errors or failures (2331 before; the two new endpoint tests account for the difference).
- jest — 49 suites / 455 tests green.
- ESLint and PHPCS clean on every touched file.

dkoo (approved):
Approving — both blockers from my previous review are verified resolved, and every suggestion and nit was addressed. Thorough follow-up, thank you.

**Blocker 1 (filter scan paged → offset).** Fixed by paging with `offset` and adding `orderby => ID` for a deterministic tiebreak. I confirmed this is a real fix, not just a rewrite: the new `test_status_filter_scan_advances_past_the_first_chunk` fails against the old `paged` code and passes against `offset`, and — critically — the mock now implements `offset` and deliberately refuses `paged`, so this class of bug can no longer hide behind the mock. I also re-ran the live reproduction against real WooCommerce Subscriptions in an isolated env: with a shrunk chunk size, the Cancelled filter now returns only the genuinely-churned reader and correctly excludes one holding a live plan alongside a cancelled one (the exact contradiction from last round is gone).

**Blocker 2 (Subscription-column links swallowed).** Fixed by unnesting the link from the DataViews title cell — the owner cell now renders the email, and the plan link lives in its own default-visible `plan` column outside the `ItemClickWrapper`. Confirmed clickable in-browser against the updated env: the Subscription-column links now navigate to the subscription rather than being intercepted by the row click-through. The keyboard/ARIA nesting is eliminated by the same change.

Everything else landed too: positive `wcs_statuses_for()` assertions (pending via individual + group inheritance, switched→on-hold, pending-cancel→active, expired→cancelled), error-state + `reload()` hook tests, the avatar batch cap unified to a single server-localized authority (with a test that pins agreement rather than a client-side copy), the `statusRank` `?? 99` guard mirroring PHP, `isItemClickable` on the subscriber list, the corrected `links.jsx` docblock, and the test-helper `array_filter` fix. Full PHP suite (2333) and CI are green.

Nice work — LGTM.

PR #632, opened 2026-07-16
refactor(audience): split the configure-view draft from its retry buffer

thomasguillot:
### All Submissions:

Separates the Configure form's live draft from the save retry buffer in the Audience → Integrations wizard, and fixes the unsaved-changes edge cases that separation surfaces.

- The Configure form's unsaved edits now belong to the view and reset when you leave or switch integrations. The retry buffer — the copy that survives a failed save — stays in the request layer, so a failed edit is never lost: returning to the integration re-shows it with Save enabled to retry.
- A field the backend normalizes on save no longer stays stuck "unsaved" after a successful save (clearing the Metadata prefix stores `NP_`; a number field's text coerces to a number), and no longer fires a redundant save on the next click.
- Editing a field back to its saved value now clears the unsaved-changes state instead of leaving it dirty.
- When enabling an integration fails after its settings already saved, the enable dialog keeps the fields you filled in for an in-place retry, and an unrelated pending Configure edit is preserved.

Follows #258 (now merged), which added the unsaved-changes guard this refactor rebuilds on. Carries #641 (components more-menu fix) and #657 (integrations settings polish); both land together when this merges.

1. Go to **Audience → Integrations** and open an ESP integration's Configure view. Change a field and Save — confirm the success notice, that Save disables afterward, and that leaving the view prompts no unsaved-changes dialog.
2. Clear the **Metadata field prefix** and Save. Confirm the field reconciles to `NP_`, Save disables, and clicking Save again does **not** fire another request.
3. Edit a field, then edit it back to its original value. Confirm Save disables and leaving the view prompts no unsaved-changes dialog.
4. Simulate a failed save (e.g. go offline) and Save. Confirm an error notice, the value is retained, and Save stays enabled. Leave the view in-app and return — confirm the edit is re-shown with Save enabled. Go back online and Save — confirm success and that the retry state clears.
5. From the Integrations grid, **Enable** an integration that needs required fields. Fill the dialog fields and trigger an enable failure (e.g. offline). Confirm the dialog keeps your filled fields (does not collapse to a bare error) so you can retry.
6. With a pending failed Configure edit on one field, use the grid Enable dialog to save a **different** required field. Return to Configure and confirm your pending edit is still there.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
@@ -9,7 +9,7 @@ import { useEffect, useMemo, useRef, useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { Accordion, Divider, Grid, SectionHeader, useUnsavedChangesDialog } from '../../../../../packages/components/src';
+import { Accordion, AccordionPanel, Divider, Grid, SectionHeader, useUnsavedChangesDialog } from '../../../../../packages/components/src';
 import { WIZARD_STORE_NAMESPACE } from '../../../../../packages/components/src/wizard/store';
 import WizardsTab from '../../../wizards-tab';
 import { SettingsField } from './settings-field';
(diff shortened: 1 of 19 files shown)


plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -93,7 +93,11 @@ export const ConfigureView = ( { integrations, loading, pendingChanges, saving,
 				{
 					type: 'primary',
 					label: __( 'Save', 'newspack-plugin' ),
-					action: () => onSave( integrationId ),
+					action: () => {
+						onSave( integrationId, draftRef.current )
+							.then( () => setDraft( {} ) )
+							.catch( () => {} );
+					},
```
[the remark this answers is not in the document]
thomasguillot: Addressed in the reconcile commit (7f29edce3). The Save action no longer does `.then(() => setDraft({}))` — it is now `onSave(id, draftRef.current).catch(() => {})`, and the draft clears via a reconcile effect on the mounted instance when the parent reflects the saved values. So there is no setState on an unmounted component, and it also fixes the phantom dirty-draft case when a save completes after the view has remounted.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -16,38 +16,58 @@ import { SettingsField } from './settings-field';
...
+// are compared as sets and booleans are coerced, else a saved field would stay
+// stuck "dirty".
+const valuesMatch = ( a, b ) => {
+	if ( Array.isArray( a ) && Array.isArray( b ) ) {
+		return a.length === b.length && a.every( value => b.includes( value ) );
+	}
+	if ( typeof a === 'boolean' || typeof b === 'boolean' ) {
+		return toBool( a ) === toBool( b );
+	}
+	return a === b;
```
miguelpeixe, line 36: **Value-equality reconcile leaves backend-normalized fields stuck "unsaved"** — this is the change I'd like addressed before merge.

The reconcile effect (and `hasPending`) clear the draft only when a key's value *equals* the saved server value. That breaks whenever the backend normalizes a value on save:

- **Live:** the **Metadata field prefix** is an editable text field; clearing it and saving submits `''`, but the backend forces the default — `update_metadata_prefix( '' )` stores `'NP_'` (`class-integration.php:937-942`). The response carries `value: 'NP_'`, so `valuesMatch( 'NP_', '' )` is `false`, the draft `{ metadata_prefix: '' }` is retained, and the field stays stuck **dirty after a successful save** — Save stays enabled, the nav-guard stays armed, and re-clicking Save fires a redundant POST.
- **Latent:** the framework's `number` field type emits a string from `TextControl` while the sanitizer coerces to a number (`$value + 0`), so `valuesMatch( 5, '5' )` is `false` — any saved number field would be permanently phantom-dirty. No integration declares one today, but booleans *were* explicitly handled, so it's a real gap.
- Typing a field **back to its original value** also leaves the key present (`hasPending` checks key *presence*, not a value diff), the same stuck-dirty result without a round-trip.

Since the old model cleared all pending changes unconditionally on save-success, these are narrow regressions (not data loss). Consider reconciling by "this key was submitted and the save succeeded" (robust to server normalization) rather than pure value-equality, and/or value-diffing on edit — the sibling `product-form.tsx:150-168` diffs a live snapshot against a captured baseline. Worth a reconcile test for the normalized / number case; it's currently unpinned.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.test.js
```diff
@@ -1,20 +1,30 @@
...
 import { useUnsavedChangesDialog } from '../../../../../packages/components/src';
-import registerWizardStore from '../../../../../packages/components/src/wizard/store';
 
+const mockSetHeaderData = jest.fn();
+
+jest.mock( '@wordpress/data', () => ( {
+	useDispatch: () => ( { setHeaderData: mockSetHeaderData } ),
+} ) );
+// Stub the components barrel: with @wordpress/data mocked, the real barrel eagerly loads @wordpress/rich-text, whose module-load combineReducers() call throws.
+jest.mock( '@wordpress/components', () => ( {
```
miguelpeixe, line 19: **[nit] Test barrel mocks only cover the current field types**

Leaving `SettingsField` unmocked is the right call (it's what wires the mocked `TextControl` to the draft). But the barrel mocks supply only a subset of what `SettingsField` imports: `@wordpress/components` provides only `CheckboxControl` (not `ExternalLink` / `TextareaControl`), and the internal barrel omits `Button` / `SelectControl`. Fine for the current text/checkbox/metadata fixtures, but a future fixture using `select` / `oauth` / `textarea` or `help_url` would render an `undefined` component and fail with a cryptic error rather than an obvious one. (The `@wordpress/components` stub-with-comment workaround for the `@wordpress/rich-text` `combineReducers()` module-load throw is sound and durable.)

miguelpeixe (changes requested):
Requesting changes on one item — the refactor itself is clean and well-tested, and I'd happily approve once the phantom-dirty below is sorted.

The load-bearing invariant (a failed edit is never silently lost) holds on every path I traced, the `savingRef` fragility from #258 is structurally eliminated rather than patched, and the setState-after-unmount concern from the earlier thread is properly resolved by the reconcile-effect approach.

**The one thing to fix before merge** — the draft reconciles by *value-equality*, so a field the backend normalizes on save stays stuck "unsaved" afterward. Concretely: clear the **Metadata field prefix** and save — the backend forces the default `NP_` (`class-integration.php:937-942`), the response comes back `NP_`, `valuesMatch( 'NP_', '' )` is `false`, so the draft is retained and the field stays dirty **after a successful save** (Save stays enabled, the nav-guard stays armed, and re-saving fires a redundant POST). Same shape is latent for `number` fields (`'5' !== 5`). Since the old model cleared pending changes unconditionally on success, this is a narrow regression. Reconciling by "key submitted + save succeeded" (rather than pure value-equality) would be robust to server normalization — the sibling `product-form.tsx:150-168` diffs against a baseline snapshot. Inline detail below.

Two non-blocking notes:

- **"No user-facing behavior change" undersells a real improvement.** Failed edits now survive navigation — the parent retains `inFlightChanges[id]` on failure and `ConfigureViewInner` re-seeds from it on return, where the old model discarded the edit on leave. That's the actual #258 fix; worth naming as intentional rather than folding under "no behavior change."
- A small test-mock robustness nit, inline.

plugins/newspack-plugin/src/wizards/audience/views/integrations/index.js
```diff
@@ -1,98 +1,177 @@
...
+				apiFetch( {
+					path: `${ API_PATH }/${ integrationId }/enabled`,
+					method: 'POST',
+					data: { enabled: true },
+				} )
+					// The settings save succeeded even if enabling failed — reflect
+					// the saved values so the UI doesn't offer the modal again for
+					// settings that are already stored.
+					.catch( error => {
+						setIntegrations( savedData );
```
chickenn00dle: **Severity:** suggestion

On a partial setup-and-enable — the settings POST succeeds but the enable POST fails — this catch does `setIntegrations( savedData )` before rethrowing. That updates `integrations[ integrationId ]`, which is the enable modal's `integration` prop, so `getMissingRequiredFields()` (`enable-modal.js:46`) recomputes to empty and the modal re-renders with **no form fields** — just the generic error notice and the footer buttons. The modal stays open (the rejection skips the close), and a second Enable click self-heals by retrying with the retained `values`, so this is UX-only, not data loss — but a user who hits a transient enable error watches the inputs they just filled disappear under an error message. Consider keeping the fields rendered (filled) on retry, or not swapping the `integration` state until enable also succeeds.

plugins/newspack-plugin/src/wizards/audience/views/integrations/index.js
```diff
@@ -1,98 +1,177 @@
...
+					} )
+					.then( data => {
+						setIntegrations( data );
+						addEnabledNotice( integrationId, true, data );
+						return data;
+					} )
+					.finally( () => {
+						setInFlightChanges( prev => {
+							const next = { ...prev };
+							delete next[ integrationId ];
```
chickenn00dle: **Severity:** nit

This `finally` deletes `inFlightChanges[ integrationId ]` whenever the settings save succeeded. If a *prior* failed ConfigureView save had left a recovery copy in `inFlightChanges` for the same integration (the retry buffer #258 added), enabling via the grid modal silently discards it — the modal only submits the missing-required fields, so any failed edits to already-set fields in that buffer are dropped without being saved, and a later ConfigureView mount seeds an empty draft. Very edge (needs a stale failed buffer to coexist with the enable-modal path) and the clear may well be intentional — worth a quick confirm that the cross-flow discard is deliberate.

chickenn00dle (commented):
Release risk: LOW. Focused on the draft / retry-buffer split the PR is named for, the core mechanism traces clean — the per-integration remount, the ref-through-closure Save action, the seed-from-retry-buffer recovery, and the fail-then-retain path all interlock correctly, and I couldn't find a lost-update or stale-read regression across the in-flight-edit / cross-integration-refetch / double-save matrix. The #258 recovery (a failed edit surviving navigation) works as intended.

Two non-blocking observations on the newer setup-and-enable (enable modal) flow, left inline — neither is a blocker, and the second is an edge case worth a quick confirm that the behavior is intentional. I'm not restating the reconcile / phantom-dirty concern already raised on this PR.

Note: UI/visual behavior was not verified in a browser (no env).

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
thomasguillot: Fixed in 2019322b6. The reconcile no longer compares against the server value — it drops the keys the draft submitted once the parent clears them from the retry buffer, so a backend-normalized field (`'' → NP_`, number `'5' → 5`) reconciles cleanly. `hasPending` is a value diff now too, so editing a field back to its saved value disarms the guard. Added reconcile tests for the normalized-prefix and number cases.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.test.js
thomasguillot: Fixed in 2019322b6. The barrel now stubs everything `SettingsField` imports (`ExternalLink`/`TextareaControl` from `@wordpress/components`, `Button`/`SelectControl` from the internal barrel), so a future select/oauth/textarea fixture renders a stub instead of `undefined`.

plugins/newspack-plugin/src/wizards/audience/views/integrations/index.js
thomasguillot: Fixed in 60a7e0432. `handleSetupAndEnable` no longer swaps integration state on enable failure — state changes only once both steps succeed, so the enable modal keeps its filled fields for an in-place retry.

plugins/newspack-plugin/src/wizards/audience/views/integrations/index.js
thomasguillot: Good catch, not deliberate. Fixed across 60a7e0432 and 8186af72a: the flow no longer clears the whole buffer — it drops only the keys it just saved, so a prior failed ConfigureView edit on other fields survives.

miguelpeixe (approved):

PR #634, opened 2026-07-16
fix(cli): key migrate-teams group reuse on team ID, not owner (NPPD-2060)

adekbadek:
> **Stacked on #626** (base branch `feat/nppd1753-migration-cli`). GitHub will retarget this PR to `main` when #626 merges.

Ref: NPPD-2060 (carries the private specifics).

### All Submissions:

**Problem.** In `migrate-teams`, when a team's linked subscription wasn't reusable, the target group subscription was resolved **by owner**. An owner holding several teams therefore got a **single merged group**: members of distinct teams combined, and the group's name and seat limit took whichever team was processed last. Because the merged group inherits the last team's seat limit as its member cap, members beyond that limit can **lose access** — and distinct company groups lose their identity. This is a real shape in the field (e.g. a site with a few staff/agency accounts each owning dozens of teams and hundreds of members). The Access Control runtime already supports multi-group owners; the merge was a migration-only defect.

**Fix.** Key the reuse lookup on the team, not the owner:

- Stamp each migrated group subscription with its source team via a `_newspack_migrated_team_id` subscription meta marker.
- On re-runs, reuse by that marker — one group subscription per team, always, regardless of owner. The command stays idempotent. A marker match wins whatever the subscription's status: a team whose group has since expired or gone on hold reuses that group rather than getting a second, duplicate one. When the reused group's status actually withholds access (i.e. anything outside `WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES` — note `pending-cancel` still grants group access), the team is recorded as a run issue, so it appears in the end-of-run ERRORS section and the summary tally rather than reading as a clean success.
- The owner lookup survives only as a **fallback for groups created by migrator runs predating the marker** (unmarked). It skips already-marked groups (so a sibling team never merges into another team's group), requires the subscription's **own** `_newspack_group_subscription_enabled` meta — not just `is_group_subscription()`, which reads through to product settings that `migrate-team-products` stamps on every team product, so a sibling team's live linked subscription would otherwise look adoptable — and emits a `WP_CLI::warning` naming any team the adopted subscription is currently linked to.
- The resolver does this in a **single pass** over the owner's subscriptions (marker match wins outright, even when an unmarked group is iterated first) and returns a structured result rather than a by-ref out-param.
- **A group the publisher disabled is refused, not routed around.** The marker is read *before* the group-enabled gate, so a group this team was migrated to that a publisher has since set `_newspack_group_subscription_enabled => 'no'` on stays visible to the resolver. Reusing it would silently re-enable a flag the publisher set deliberately (the migrator stamps `'yes'` unconditionally); skipping it would create a second group carrying the same team ID. So neither happens: the team is reported as a run issue naming every such group, and left untouched for the operator to either re-enable the group or clear its marker. A team that also has a usable marked group is migrated into that one instead of refused. `backfill-team-managers` reports the same teams as unresolved, with the group named rather than a generic "nothing found".
- **Backfill unified.** `backfill-team-managers` resolved by owner too, so for a multi-team owner it could promote one team's managers into another team's group — the same cross-team merge, in the sibling command. `resolve_backfill_subscription()` now routes through the same team-aware resolver, closing that path; the now-redundant owner-only finder was removed.

Net result: multi-team owners migrate to **multiple groups** (correct names + seat limits), and both the migration and the manager backfill are merge-safe.

Unit tests (run from `plugins/newspack-plugin` in the worktree/env):

1. `n test-php --filter Test_Teams_Migration` -> `OK (25 tests, 135 assertions)`. New tests cover: per-team keying + idempotency, the multi-team merge regression, the warned legacy fallback, marker-wins over an earlier-iterated unmarked group, reuse of a non-active marked group, a `pending-cancel` marked group not being reported as access-less, the fallback ignoring a subscription group-enabled only via its product, reuse ignoring other-team/inactive/non-group subs, and the two backfill-resolution cases (multi-team marker resolution; legacy unmarked fallback).
2. `n test-php --group WooCommerce_Subscriptions_Integration` -> `OK (114 tests, 318 assertions)` (no regression in the shared group-subscription/backfill code).
3. PHPCS on both changed files -> clean.

End-to-end (WordPress + WooCommerce + Subscriptions + Memberships + Teams in a disposable env):

4. Create an owner with two unlinked teams of different seat counts and member sets (e.g. seat 5 / 3 members and seat 10 / 4 members).
5. `wp newspack migrate-teams --product-id=<id>` (dry-run), then `--live`: two **distinct** group subscriptions are created — correct distinct names, seat limits (6 and 11), and `_newspack_migrated_team_id` markers. No merge.
6. Re-run `--live`: both subscriptions are reused by marker; zero new subscriptions; owner still holds exactly two groups (idempotent).
7. Legacy path: give an owner a pre-existing unmarked group (with subscription-level group meta) + two teams; the first team adopts the legacy group (with a warning), the second gets a fresh group — no merge.
8. `backfill-team-managers` for the second team resolves to that team's own group, never the first team's.

### Other information:

**Notes for reviewers:**

- A loop-level `migrate_teams()` CLI test is deliberately omitted: `WP_CLI` isn't stubbed in this unit suite (the test file's own convention is that subscription-creation machinery is exercised on a real site, not in unit tests). The command-level behavior was instead verified end-to-end in a live env (steps 4-8 above).
- Pre-existing follow-up (out of scope here): the `backfill-team-managers` loop lacks a `wp_clear_object_cache()` between teams, unlike its sibling commands. This diff makes each iteration slightly heavier (a full owner-subscription pass), so it's worth a separate small fix, but the loop internals aren't touched here.

Closes NPPD-2060.


plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1178,10 +1226,27 @@ private static function find_group_subscription_for_owner( $owner_id ) {
...
+			}
+			$marked_team_id = (int) $subscription->get_meta( self::MIGRATED_TEAM_ID_META_KEY );
+			// A group already migrated from this team wins outright — reuse keys on the
+			// source team, so a sibling team of the same owner never merges into it.
+			if ( $marked_team_id === $team_id ) {
 				return $subscription;
 			}
+			// Remember the first unmarked group as the legacy fallback. A group already
+			// marked for a different team is skipped so it is never merged into.
+			if ( null === $unmarked_group_sub && 0 === $marked_team_id ) {
```
miguelpeixe: **Blocker:** Legacy fallback can adopt a never-migrated, product-level group-enabled subscription — The fallback's eligibility test is `Group_Subscription::is_group_subscription()`, which treats a subscription as group-enabled when only its *product* carries `_newspack_group_subscription_enabled` (subscription-level meta overrides product settings only when non-empty). The sibling `migrate-team-products` command stamps that product meta on every team product and its variations — so after the prescribed product migration, any active subscription on a team product reads as a group subscription while carrying neither subscription-level meta nor the team marker. "Unmarked" therefore doesn't reliably mean "created by a pre-marker migrator run", as the docblock and the adoption warning assume.

Concrete first-live-run case: a multi-team owner whose broken-link team B has a lower post ID than their healthy team A. B's fallback scan encounters A's still-unmarked linked subscription (group-enabled via product meta), adopts it, stamps it with B's team ID, renames it, sets B's seat limit, and adds B's members into it; A then reclaims and re-stamps it via the linked path — leaving B's members stranded in A's group. Re-runs don't clean this up: B gets a fresh group, but the mis-added members stay in A's.

**Risk if not addressed:** If migrate-team-products has already run (the expected companion step) and a multi-team owner has an unlinked team whose post ID is lower than a sibling team with a healthy linked subscription, the first live migrate-teams run adopts that sibling's live subscription for the lower-ID team — stamping it, renaming it, adding that team's members and setting its seat limit — before the sibling reclaims it. The sibling's group ends up holding the wrong team's members (possibly over its seat limit, so those members lose content access) and the lower-ID team has no group until a second run; the second run does not remove the mis-added members, so the cross-team merge persists for exactly the multi-team-owner sites this change exists to protect.

**Consider:** Require the subscription itself to carry `_newspack_group_subscription_enabled` before adopting it in this fallback, so the branch can only reuse a group an earlier migrator run actually created — never a sibling team's subscription that looks group-enabled only through its product. Because a re-run does not remove members mis-added this way, a wrong adoption is not self-correcting; keeping the fallback strict is what prevents a persistent cross-team merge.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1178,10 +1226,27 @@ private static function find_group_subscription_for_owner( $owner_id ) {
...
 				continue;
 			}
-			if ( Group_Subscription::is_group_subscription( $subscription ) ) {
+			if ( ! Group_Subscription::is_group_subscription( $subscription ) ) {
+				continue;
+			}
+			$marked_team_id = (int) $subscription->get_meta( self::MIGRATED_TEAM_ID_META_KEY );
+			// A group already migrated from this team wins outright — reuse keys on the
+			// source team, so a sibling team of the same owner never merges into it.
+			if ( $marked_team_id === $team_id ) {
```
miguelpeixe, line 1327: **Suggestion:** Marked group that is no longer active is invisible to reuse — The `'active' !== get_status()` gate runs before the marker check, so a team's own marked group that has expired (or was put on hold or cancelled) is never matched. This is reachable in normal operation: `build_subscription_dates()` sets an `end` date whenever the team's membership end is in the future, and WooCommerce Subscriptions transitions the subscription to `expired` once that date passes. On a later `migrate-teams --live --product-id=…` re-run, the resolver returns null and the loop creates a new active group and stamps it with the same team ID — duplicating the team's group and splitting members/managers across the expired and new subscriptions. In `backfill-team-managers`, the same team reports "unresolved" even though a real (expired) group exists.

**Risk if not addressed:** A team whose group carries a future end date that has since passed (or whose owner changed since migration, per the resolver's own note) is invisible to reuse; each re-run creates another active group for it and backfill-team-managers reports it unresolved, leaving two or more groups per team to reconcile by hand while the original members sit in the now-expired group.

**Consider:** Check the team marker before the active-status gate (or, when no active group resolves but a marked inactive one exists for the team, report it instead of creating a new one), so an expired or on-hold group is reused or surfaced rather than silently duplicated on the next run and hidden from backfill.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-teams-migration.php
```diff
@@ -483,4 +484,175 @@ public function test_backfill_selects_active_group_subscription_among_mixed() {
...
+		$subscription->save();
+		return $subscription;
+	}
+
+	/**
+	 * Reuse keys on the source team: each team resolves back to the group subscription
+	 * stamped with its own team ID, even when one owner owns both — so re-running the
+	 * migration updates each team's group in place rather than duplicating it.
+	 */
+	public function test_reuse_keys_on_team_marker_not_owner() {
```
miguelpeixe, line 595: **Suggestion:** No test pins marker-wins over an earlier-iterated unmarked group — The resolver's load-bearing property is that a marker match wins even when an unmarked (fallback-eligible) group appears **earlier** in `wcs_get_users_subscriptions()` iteration order — the unmarked group is only remembered, never returned early. None of the six new tests constructs that mixed case for one owner: `test_reuse_keys_on_team_marker_not_owner` uses only marked groups, and the fallback/ignore tests have no marked-for-the-queried-team group at all. A refactor to "return the first eligible subscription" would pass every current test while reintroducing cross-team adoption. The mock's `wcs_get_users_subscriptions()` iterates in insertion order, so the case is trivially controllable: create an unmarked group first, then a group marked for the queried team, and assert the marked one is returned with `$used_owner_fallback === false`.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1155,21 +1178,46 @@ private static function resolve_backfill_subscription( $team_id, $owner_id ) {
...
+	 * @param int  $team_id             The team post ID.
+	 * @param int  $owner_id            The team owner user ID.
+	 * @param bool $used_owner_fallback Set true when the unmarked owner fallback matched. Passed by reference.
 	 *
 	 * @return \WC_Subscription|null
 	 */
-	private static function find_group_subscription_for_owner( $owner_id ) {
-		$owner_id = absint( $owner_id );
-		if ( ! $owner_id || ! function_exists( 'wcs_get_users_subscriptions' ) ) {
+	public static function find_reusable_group_subscription( $team_id, $owner_id, &$used_owner_fallback = false ) {
```
miguelpeixe: **Nit:** By-ref out-param on a now-public resolver — `find_reusable_group_subscription( $team_id, $owner_id, &$used_owner_fallback = false )` is correct for all current call sites (the callee re-initializes the flag at entry, so the undeclared variable at the `migrate_teams()` call site and cross-iteration staleness are both safe), but an out-param is an awkward shape for a public method: dynamic invocation (`call_user_func()`, hook machinery, spread args) can't propagate the reference — PHP 8 warns and the flag silently stays false while a subscription is still returned, which would suppress the legacy-adoption warning. Returning a small structured result (e.g. `[ 'subscription' => …, 'used_owner_fallback' => … ]`) would be sturdier. Fine to defer while all callers are direct static calls.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -215,22 +228,24 @@ public function migrate_teams( $args, $assoc_args ) {
...
+				$reusable_sub = self::find_reusable_group_subscription( $team_id, $owner_id, $used_owner_fallback );
+				if ( $reusable_sub ) {
+					$subscription = $reusable_sub;
+					if ( $used_owner_fallback ) {
+						// The only groups lacking the team marker are from migrator runs
+						// predating per-team marking. Adopt one for this team (it is marked
+						// below), and warn: for a multi-team owner only the first team can
+						// adopt the legacy group, and if it was another team's group the
+						// name changes.
+						WP_CLI::warning( sprintf( 'Team %d ("%s"): no group subscription marked for this team; adopting unmarked group subscription %d owned by owner %d (legacy migrator run) and marking it for this team.', $team_id, $team->post_title, $reusable_sub->get_id(), $owner_id ) );
```
miguelpeixe: **Suggestion:** Consider rewording this warning so it does not assert "legacy migrator run" unconditionally: the same fallback fires when the adopted subscription is a sibling team's currently-linked group (group-enabled via its product), and an operator who reads "legacy" may wave through a real cross-team merge in the dry-run preview. Naming the adopted subscription's own linked team here would let the operator catch that case before going live.

miguelpeixe (changes requested):
The team-marker design is right and the core resolver logic is sound — marker-wins ordering holds regardless of iteration order, meta handling is HPOS-safe, making the resolver public opens no web surface, and the six new tests pin the headline regression. One thing needs closing before this runs on a multi-team-owner site: the legacy fallback treats any unmarked, active, group-enabled subscription as a pre-marker migrator group, but after `migrate-team-products` stamps product-level meta, a sibling team's live linked subscription also qualifies — so a broken-link team can adopt it on the first live run, recreating the cross-team merge this fix exists to prevent (and re-runs don't fully self-correct). Requiring subscription-level `_newspack_group_subscription_enabled` in the fallback closes it in one line. Two smaller items worth a look: an expired/on-hold marked group is invisible to reuse (a re-run duplicates the team's group), and no test pins that a marker match beats an earlier-iterated unmarked group. Nits: stale owner-search wording in the inactive-linked warning, and the by-ref out-param on a now-public method.

- `plugins/newspack-plugin/includes/cli/class-teams-migration.php:227` — **Nit:** Inactive-linked warning still describes the removed owner-keyed search — The warning for an inactive linked subscription still says "— searching for an existing group subscription owned by the team owner.", carried over from the owner-keyed resolver this PR removes. On a live migration transcript it primes the operator with exactly the owner-merge semantics the fix eliminates, right before the corrected marker-first lines. Rewording to e.g. "searching for the group subscription this team was previously migrated to" keeps the transcript accurate.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed. The owner fallback now requires the subscription's *own* `_newspack_group_subscription_enabled` meta (`has_own_group_enabled_meta()`), so a subscription that only reads as group-enabled through its product — which is every team product after `migrate-team-products` — is no longer adoptable. Only a group an earlier migrator run (or a publisher) explicitly enabled on the subscription itself can be reused, which is what "unmarked" was meant to mean. New test `test_fallback_ignores_group_enabled_only_via_product()` pins it (verified red without the extra condition).

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Fixed by moving the marker check ahead of the status gate: a group marked for this team is now reused whatever its status. A non-active match is flagged (`reused_inactive`) and migrate-teams warns that members added get no access until it is active again; an active marked group still wins if the team somehow has both. That keeps one group per team instead of duplicating on re-run, and makes backfill-team-managers resolve the expired group rather than reporting the team unresolved. The unmarked owner fallback still requires an active subscription — that one is not known to belong to the team. New test `test_inactive_marked_group_is_reused_rather_than_duplicated()`.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-teams-migration.php
adekbadek: Added `test_marked_group_wins_over_earlier_iterated_unmarked_group()`: creates the unmarked group first (so it is iterated first), then the group marked for the queried team, and asserts the marked one is returned with no owner fallback. A "return the first eligible subscription" refactor now fails.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Done — the out-param is gone. `find_reusable_group_subscription( $team_id, $owner_id )` now returns `[ 'subscription' => …, 'used_owner_fallback' => …, 'reused_inactive' => … ]`; the third key carries the marked-but-inactive case from the sibling thread, which the by-ref shape would have made worse. Callers and tests updated.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Reworded: the warning no longer asserts "legacy migrator run", and it now names the team the adopted subscription is currently linked to when there is one — "It is currently linked to team %d (\"%s\") — check this is not a cross-team merge before running --live." (via a new `find_team_linked_to_subscription()` lookup). With the fallback also requiring subscription-level group meta (see the blocker thread), a sibling team's live linked subscription no longer qualifies at all, so this is now belt-and-braces for the dry-run reader.

adekbadek:
All five addressed in 269cdde: fallback now requires subscription-level `_newspack_group_subscription_enabled` (blocker), marker match wins over the active-status gate with a warning on a non-active reuse, out-param replaced by a structured return, adoption warning no longer claims "legacy" and names any team the subscription is linked to, and the inactive-linked warning at line 227 now says it is searching for the group this team was previously migrated to. Three new tests (product-only enablement, inactive marked reuse, marker-wins-over-earlier-unmarked); `n test-php` full suite green (2111 tests), PHPCS clean. PR description updated for the new fallback rule and test counts.

adekbadek:
@miguelpeixe Re-requesting review – the changes you asked for have been pushed and need a look before this merges:
- fix(cli): tighten migrate-teams group reuse fallback (NPPD-2060, #634)

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/634#pullrequestreview-4748298205)

All five items are addressed in 269cdde (already on the branch).

**The blocker — fallback adopting a sibling team's live linked subscription.** Taken as proposed: the fallback now additionally requires the subscription's *own* `_newspack_group_subscription_enabled` meta via a new `has_own_group_enabled_meta()`, rather than trusting `Group_Subscription::is_group_subscription()`, which reads through to the product settings `migrate-team-products` stamps. So a sibling team's not-yet-processed linked subscription (marker-less, group-enabled only via its product) is no longer adoptable. Pinned by `test_fallback_ignores_group_enabled_only_via_product()`, which enables the group through `newspack_group_subscription_settings` — the product-level path — and asserts nothing is adopted.

**Expired/on-hold marked group invisible to reuse.** Fixed: the status filter moved off the loop head, so a marker match now wins whatever its status. An active marked group still returns immediately; a non-active one is held and returned only if no active marked group turns up later in the pass, flagged back to the caller as `reused_inactive` so `migrate-teams` warns that members added get no access until the group is active again. It is not silently reactivated. Covered by `test_inactive_marked_group_is_reused_rather_than_duplicated()`.

**Missing ordering test.** Added `test_marked_group_wins_over_earlier_iterated_unmarked_group()`: an unmarked, fallback-eligible group is created *before* the team's own marked one, so it is iterated first, and the test asserts the marked one still wins. That is the load-bearing property — the fallback is only ever remembered, never returned early — so a "return the first eligible subscription" refactor can't creep back in.

**Nits.** Both done: the inactive-linked warning is reworded (see the separate thread), and `find_reusable_group_subscription()` no longer takes a by-ref out-param — it returns `[ 'subscription', 'used_owner_fallback', 'reused_inactive' ]`, which is also what carries the new inactive-reuse signal to the caller.

Verified: `n test-php --filter Test_Teams_Migration` -> OK (24 tests, 131 assertions); `--group WooCommerce_Subscriptions_Integration` -> OK (113 tests, 314 assertions); PHPCS clean on both changed files. The PR description was updated to match.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/634#pullrequestreview-4748298205)

Reworded — the warning at `class-teams-migration.php:227` now reads "— searching for the group subscription this team was previously migrated to.", matching the team-keyed resolver that replaced the owner-keyed search. Fixed in 269cdde.

adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/634#pullrequestreview-4748298205)

Used your wording verbatim: `class-teams-migration.php:227` now reads "Team %d: linked subscription %d is not active (status: %s) — searching for the group subscription this team was previously migrated to." Fixed in 269cdde.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2060-migration-tooling-migrate-teams-merges-multi-team-owners` (stash local work first).

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1155,34 +1190,157 @@ private static function resolve_backfill_subscription( $team_id, $owner_id ) {
...
 				continue;
 			}
-			if ( Group_Subscription::is_group_subscription( $subscription ) ) {
-				return $subscription;
+			$marked_team_id = (int) $subscription->get_meta( self::MIGRATED_TEAM_ID_META_KEY );
+			// A group already migrated from this team wins outright — reuse keys on the
+			// source team, so a sibling team of the same owner never merges into it, and
+			// an unmarked group appearing earlier in the iteration never pre-empts it.
+			if ( $marked_team_id === $team_id ) {
+				if ( 'active' === $subscription->get_status() ) {
```
miguelpeixe, line 1337: **Nit:** `pending-cancel` is flagged inactive but still grants access — The marker branch treats any status other than `active` as `reused_inactive`, but group membership reads through `WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES`, which is `[ 'active', 'pending-cancel' ]` (see `Group_Subscription::get_subscriptions_for_user()`). So for a `pending-cancel` group the caller's warning tells the operator "members added below get no access until it is active again" when those members in fact have access — which could prompt reactivating a subscription a reader deliberately cancelled.

The resolution itself is unaffected: a `pending-cancel` marked group is held as `$inactive_marked_sub` and still returned after the scan, and an active marked group still wins. It is purely the flag, and therefore the message.

**Consider:** setting `reused_inactive` from `! $subscription->has_status( WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES )` rather than `'active' !== get_status()`, so the warning fires only for statuses that actually withhold access.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -211,26 +224,40 @@ public function migrate_teams( $args, $assoc_args ) {
...
+						if ( $linked_team_id && $linked_team_id !== $team_id ) {
+							$linked_team = \get_post( $linked_team_id );
+							$linked_note = sprintf( ' It is currently linked to team %d ("%s") — check this is not a cross-team merge before running --live.', $linked_team_id, $linked_team ? $linked_team->post_title : '?' );
+						}
+						WP_CLI::warning( sprintf( 'Team %d ("%s"): no group subscription marked for this team; adopting unmarked group subscription %d owned by owner %d and marking it for this team.%s', $team_id, $team->post_title, $reusable_sub->get_id(), $owner_id, $linked_note ) );
+					} elseif ( $reuse['reused_inactive'] ) {
+						// Reusing the team's own group even though it is no longer active
+						// keeps one group per team; creating a second one would split the
+						// team's members across two groups. It is not reactivated here.
+						WP_CLI::warning( sprintf( 'Team %d: reusing group subscription %d migrated from this team, but its status is "%s" — members added below get no access until it is active again.', $team_id, $reusable_sub->get_id(), $reusable_sub->get_status() ) );
```
miguelpeixe: **Suggestion:** The inactive-reuse outcome never reaches the summary — This is a `WP_CLI::warning` only; it does not enter `$errors`, so the team still gets a `WP_CLI::success` line immediately below and an ordinary `New? = N` row in the end-of-run summary table, indistinguishable from a clean reuse. On a run over hundreds of teams the summary is what actually gets read, and this is the one outcome where the members were written into a group that grants nobody access.

Worth surfacing because this PR is what makes the outcome reachable at all — before it, a non-active group was skipped and the team got a fresh active one instead, so no operator has had to look for this before.

**Consider:** pushing it into `$errors` (the summary already has an errors section and downgrades the success line when it is non-empty), or carrying it as a separate flag into `summary_row()` so the table can show it. Either way the team stops reading as a clean success.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1155,34 +1190,157 @@ private static function resolve_backfill_subscription( $team_id, $owner_id ) {
...
+
+		$inactive_marked_sub = null;
+		$unmarked_group_sub  = null;
 		foreach ( \wcs_get_users_subscriptions( $owner_id ) as $subscription ) {
 			// wcs_get_users_subscriptions is filtered to include member-only groups; require ownership.
 			if ( (int) $subscription->get_user_id() !== $owner_id ) {
 				continue;
 			}
-			if ( 'active' !== $subscription->get_status() ) {
+			if ( ! Group_Subscription::is_group_subscription( $subscription ) ) {
```
miguelpeixe: **Nit (design call, fine to leave as-is):** A marked group with groups disabled still duplicates — This gate runs before the marker check, so if a publisher sets subscription-level `_newspack_group_subscription_enabled => 'no'` on a group this migrator created, the marker is ignored, the resolver returns null, and the next run creates a second subscription stamped with the same team ID — the same duplication the inactive-status fix just closed, through a different door. It then stabilises at two groups per team with members split across them, since the disabled one stays invisible.

Not obviously worth fixing: moving the marker check above this gate would close it, but `migrate_teams()` re-stamps `'yes'` unconditionally afterwards, so the migrator would silently re-enable a flag the publisher turned off deliberately. Both outcomes have a cost, and it is narrow either way.

Raising it only so the "one group subscription per team, always" invariant in the PR description has a known, deliberate exception rather than an accidental one — a sentence in the resolver docblock would be enough.

miguelpeixe (approved):
All five threads are genuinely closed, and the follow-up went further than asked in two places. Approving.

**What I verified**

| Check | Result |
| --- | --- |
| `n test-php --filter Test_Teams_Migration` | OK (24 tests, 131 assertions) |
| `n test-php --group WooCommerce_Subscriptions_Integration` | OK (162 tests, 402 assertions) |
| PHPCS on both changed files | clean |

I mutation-tested the three new tests rather than just reading them, and each is load-bearing:

- Making the fallback `return` early instead of remembering it → `test_marked_group_wins_over_earlier_iterated_unmarked_group` fails. That is exactly the refactor-guard the thread asked for.
- Restoring the `'active' !==` gate ahead of the marker check → `test_inactive_marked_group_is_reused_rather_than_duplicated` errors.
- Dropping `has_own_group_enabled_meta()` from the fallback condition → `test_fallback_ignores_group_enabled_only_via_product` fails.

**On the blocker.** `has_own_group_enabled_meta()` mirrors `Group_Subscription_Settings::get_subscription_settings()`'s own-meta branch exactly (`'' !== $meta && wc_string_to_bool( $meta )`), so it is a faithful subscription-level-only test rather than an approximation. It also does not over-tighten: `migrate_teams()` stamps `_newspack_group_subscription_enabled => 'yes'` itself, so real pre-marker migrator groups still qualify for the fallback and the legacy path stays alive.

**On the inactive-reuse path**, which this PR makes newly reachable — I checked the second-order effects and they hold up: `Group_Subscription::update_members()` has no status gate so the adds succeed mechanically; access really does gate on `WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES`, so the new warning's claim is accurate; and `replace_subscription_product()`'s `update_dates()` is already wrapped in try/catch, so an expired reuse degrades to a recorded error in the summary rather than fataling.

**On the out-param removal**, I grepped `repos/` as well as the monorepo — nothing outside this file calls `find_reusable_group_subscription()` or reads the marker, so the public signature change breaks no consumer.

**On the reworded warning**, it turned out better than I expected: `migrate_teams()` never writes `_subscription_id` back onto the team, so `find_team_linked_to_subscription()` can only match a link WC Teams itself created — which is precisely the sibling-team case that made the old wording dangerous.

Three non-blocking items inline: a wrong warning for `pending-cancel`, the `reused_inactive` outcome not reaching the end-of-run summary, and one design call worth making deliberately rather than by accident. None need to hold the merge.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Good catch — fixed in eeb7b79. The flag is now \`! $sub->has_status( WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES )\`, so a `pending-cancel` group is reused silently and only statuses that genuinely withhold access are reported.

Two things came along with it, since the old name would have become a trap: the return key is now `reused_without_access` (a `pending-cancel` group *is* inactive, so `reused_inactive` would read as false for it), and the held candidate is `$non_active_marked_sub`. Resolution is untouched — a truly `active` marked group still returns early and wins over a `pending-cancel` sibling. Added `test_pending_cancel_marked_group_is_not_reported_as_access_less`, confirmed red before the change.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Agreed — fixed in eeb7b79 by taking the first option: the outcome now goes into `$errors` instead of being a bare warning. That gets it into all four places an operator looks — the per-team report downgrades from success to warning, the message prints under the team, the end-of-run ERRORS section lists it with team + sub, and it counts in the final `%d had error(s)` tally.

`$errors` already holds operator-actionable conditions rather than strictly code failures (the "no reusable subscription and no --product-id" skip is the existing precedent), and the report line calls them "issue(s)", so the semantics fit without adding a summary column that would be empty for almost every row.

It fires only when the status genuinely withholds access, per the `pending-cancel` thread.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
adekbadek: Good catch — closed it rather than documenting it, in 653e84a.

You framed it as a choice between re-enabling the flag and leaving the duplication open, but the status quo did not really preserve the publisher's intent either: with the marked group invisible, the migrator created a **brand-new** group, enabled it, and stamped the same team ID. So both doors overrode the publisher; one just left extra mess. That made a third option worth taking.

The marker is now read before the group-enabled gate, and a marked group with group subscriptions disabled is neither reused nor ignored — the resolver reports it via `disabled_marked_group_ids` and `migrate_teams()` refuses the team (run issue + ERROR summary row, nothing written), leaving the operator to re-enable the group or clear its marker. A team that also has a usable marked group is migrated into that one instead of refused, so the refusal only fires when there is genuinely nowhere valid to go. All the disabled groups are named at once, so recovering a team that has more than one does not cost a run per group.

Two knock-ons:

- `backfill-team-managers` reported these teams as `No active group subscription found (linked or owner-owned).`, which was false twice over — a group *was* found and it *is* active. It now names the group and says to re-enable it.
- The refusal path calls `wp_clear_object_cache()` before `continue`, since it has just loaded every subscription the owner has (the same reason as #643).

Five tests added: the report-not-reuse case, the multiple-disabled case, precedence over the unmarked owner fallback, and precedence *under* a usable marked group (active and expired). The last two are the ones that matter for the hold-and-scan branch — mutating that `continue` into an early return makes both fail, so the branch is pinned rather than incidental. Full suite green (2118), PHPCS clean.

The one thing deliberately left: when a usable marked group wins, the inert disabled duplicate is not surfaced. Reporting it would mean decoupling the field from the "refuse this team" directive it currently carries, and the migration has gone somewhere correct, so the resolver docblock states that exception explicitly instead.

PR #635, opened 2026-07-16
feat(cli): port migrate-membership-gates into in-plugin migration CLI (NPPD-2059)

adekbadek:
> **#626 has landed** (as `ad8720483` on `main`), so this PR is now based on `main` and its diff shows only this PR's changes.

### All Submissions:

Part of NPPD-2059.

**Why.** The `wp newspack migrate-membership-gates` command (reads a site's WooCommerce Memberships `np_memberships_gate` layout + plan restriction rules and generates the equivalent Access Control gate(s), layouts, and content rules) existed only as an un-hosted, per-site drop-in plugin — not version-controlled anywhere, with no code review, no history, and no canonical home. #626 ported the other three migration commands (`migrate-teams`, `migrate-team-products`, `migrate-manual-members`) into the in-plugin CLI suite but left this one out. This PR completes that set.

**What.** A faithful port of the drop-in's gates command into `Newspack\CLI\Membership_Gates_Migration`, following exactly the conventions #626 established in `class-teams-migration.php`:

- New WP-CLI command `wp newspack migrate-membership-gates`, registered in `class-initializer.php` under the existing `Memberships::is_active()` gate (like the other CLI classes the file is included on every request, but only the command registration is gated on `WP_CLI`, so the cost outside CLI is a class definition).
- **Dry-run by default; pass `--live` to write** (the drop-in's convention was the inverse — write by default with a `--dry-run` opt-in; this port flips it to match #626, so an accidental invocation is a no-op).
- The rule-mapping, layout-extraction, and fingerprint-grouping logic are kept as separate, testable seam methods (`map_rules_to_ac_format()`, `extract_gate_layouts()` / `serialize_gate_inner_blocks()`, `compute_rules_fingerprint()` / `group_plans_by_fingerprint()`) so the follow-up fixes below can land surgically.
- Characterization tests pin the current behavior (including the known-buggy behaviors, marked with their follow-up issue IDs) so the stacked fixes get a clear red→green signal.

**Known bugs preserved by design.** This is a faithful port, not a rewrite — the drop-in's known defects are reproduced as-is and fixed in separate stacked PRs, each pinned or documented here:

- **NPPD-2058** — `extract_gate_layouts()` only inspects *top-level* wrapper blocks, so a gate whose `non-member-content` / `member-content` wrapper is nested inside another block (a group, or a reusable `core/block`) extracts as *empty*. Pinned red by the `extract_gate_layouts` / `serialize_gate_inner_blocks` characterization tests. (An empty extraction no longer *blanks* the layout the gate points at — see the departures below — so the consequence is a default layout rather than a blank one, but the extraction itself is still 2058's to fix.)
- **NPPD-2063** — `map_rules_to_ac_format()` uses the raw WooCommerce content-type name (e.g. `post`, `page`) verbatim as the Access Control rule slug, instead of remapping to AC's `post_types` / `specific_posts` keys. Pinned red by the `map_rules_to_ac_format` tests.
- **NPPD-2064** — fingerprint-based gate splitting/grouping lives in `group_plans_by_fingerprint()` and the merged-product consolidation. That logic depends on `WC_Memberships_Membership_Plan` and is exercised end-to-end (not unit-tested); the `compute_rules_fingerprint()` tests pin only the fingerprint's *canonicality* (order-independence), which the fix preserves, so the NPPD-2064 PR must add net-new grouping/split tests. The seam docstrings and test-file header state this precisely.

**`get_gates()` array-access adaptation.** The drop-in indexed existing gates with object access (`$gate_post->post_title` / `->ID`), but on this base branch `Content_Gate::get_gates()` returns associative arrays keyed `id` / `title` (never `WP_Post`). Porting the object access verbatim would have collapsed every existing gate under an empty key (`trim(strtolower(null))`), silently defeating the "update the existing gate matched by title" path and creating a duplicate gate on every re-run. The port uses `$gate['title']` / `$gate['id']`, which restores the intended dedup — so this port quietly corrects a latent bug in the drop-in. Idempotency is verified below.

**Drop-in collision guard.** The standalone drop-in registers the same command name (with the opposite, write-by-default flag convention). Empirically, this port's `init` registration overrides the drop-in's earlier `plugins_loaded` one, so the in-plugin (dry-run-default) command wins even when both are active. When `class-initializer.php` detects the drop-in class `Newspack_Migrate_Membership_Gates_Command`, it registers a `before_invoke:newspack migrate-membership-gates` hook that emits a loud `WP_CLI::warning` telling the operator to deactivate it. Hooking the command name rather than putting the notice in the command callback means the operator is told even if the registration order ever flips and the drop-in's write-by-default command is the one dispatched — and scoping it to that command keeps the notice out of every unrelated `wp` invocation on the site.

**Departures from the drop-in (review round 1).** Beyond the three ticketed bugs above, review surfaced defects that are fixed here rather than preserved, because leaving them would silently mis-migrate:

- `content_rules_match` is now written as `any`. WooCommerce Memberships restricts the *union* of a plan's restriction rules, while the gate evaluator defaults an unset match mode to AND — without this, every multi-rule plan under-gates at cutover.
- `extract_gate_layouts()` concatenates repeated top-level wrappers of the same type instead of keeping the first `non-member-content` and the last `member-content` (pinned by a new test). Distinct from NPPD-2058, which is about *nested* wrappers and stays preserved.
- An empty extracted layout no longer overwrites the layout the gate already points at. `Content_Gate::create_gate()` seeds both layout posts with a default block pattern, so the update path is the normal one for a freshly created gate, and writing `''` into it blanked that default — leaving the reader a truncated article with an empty gate beneath it. `create_gate_layout()` substitutes the default for empty content; the update path now agrees with it and keeps what is there, with a warning naming the layout to review.
- Plans with no content restriction rules are now skipped regardless of access method — an empty-rules gate is skipped for every post by the evaluator, so migrating one only published an inert gate the summary reported as created.
- Operator-safety hardening: `--plan` must parse to a positive ID (a mistyped value no longer widens a `--live` run to every plan); the existing-gate dedup only considers `publish` gates; a duplicate gate title within one run warns; layout writes bypass kses so admin-authored embeds/iframes survive; a mode is no longer activated when its layout could not be written (that produced a gate that never restricts) and the failure is surfaced as an `ERROR` row; the Primary gate is resolved from the canonical `newspack_memberships_gate_post_id` option; plans are loaded through `wc_memberships_get_membership_plan()`.

**Post-write verification (review rounds 2 and 3).** On `--live` each written gate is re-read and checked against the conditions `Content_Restriction_Control::get_post_gates()` and `is_post_restricted()` apply — published status, content rules the evaluator can resolve, at least one active mode, and every active mode pointing at a layout post that exists and is not blank. A gate that fails is reported as a `WARN:` row (distinct from the `ERROR:` rows for write failures) plus a count after the success line, instead of being counted as a plain success. This matters because migrated gates stay dormant until WooCommerce Memberships is deactivated, so an unenforceable gate would otherwise look migrated until someone noticed at cutover. Two cases are worth calling out:

- **Paid plans that lost their paywall.** Registration mode stops nobody who has an account (the migration never writes `require_verification`), so a gate derived from a plan that required a *purchase* is only a paywall if its paid-access mode is active *and* carries access rules. Neither is guaranteed: no extracted `member-content` markup leaves the mode inactive, and a plan whose products are all variations (dropped, since gates reference parent products) leaves it active with an empty rule set, which the evaluator skips. Verification now takes the group's purchase flag and flags both — otherwise the summary would print `created` / `purchase` for content any reader can unlock by registering a free account.
- **Partially dead rule sets.** Rules combine with `any`, so a gate is flagged whenever *any* content rule carries a slug the evaluator cannot resolve, not only when all of them do. A plan restricting all posts plus a category (a common WCM shape) maps to `post` (dead under NPPD-2063) plus `category` (live), gates the category and leaves everything else readable — a partial leak that previously verified clean. The message names the dead slugs, which also gives the NPPD-2063 PR a signal to flip.

**Not addressed here (ticketed).** Purchase plans emit only a `subscription` access rule, so a WCM plan granted by a one-time/prepaid/lifetime purchase of a simple product migrates to a gate that would paywall those buyers at cutover. Access Control has no rule that can express a completed purchase; [NPPD-2053](https://linear.app/a8c/issue/NPPD-2053) adds the `one_time_purchase` rule and [NPPD-2106](https://linear.app/a8c/issue/NPPD-2106) covers mapping WCM simple-product plans onto it.

**Mixed-group mode selection (review round 3).** A plan group is now purchase-gated only when *every* plan in it requires a purchase (`group_requires_purchase()`), not when any does. The two gate modes compose with AND for a logged-in reader, so activating paid access on a group that also holds a signup plan demanded the subscription from everyone — and since WCM grants access to a holder of *either* plan (OR semantics), the signup plan's free-registration members would have silently lost access at cutover. A mixed group now migrates registration-gated, the faithful most-permissive-plan mapping. This is distinct from NPPD-2064 / #642, which reworks *which* plans group together; this is the per-group mode decision and composes with that work.

> **Operational note:** deactivate/delete the standalone `migrate-memberships` drop-in before running the in-plugin CLI, to remove any ambiguity. Re-running the command is not edit-preserving: a matched gate's rules and layout content are overwritten.

Closes NPPD-2059.

Requires WooCommerce + WooCommerce Subscriptions + WooCommerce Memberships + newspack-plugin active on an isolated env (the standalone `migrate-memberships` drop-in should stay **deactivated**).

**Unit tests:**

1. From `plugins/newspack-plugin`: `n test-php --filter Test_Membership_Gates_Migration` → `OK (19 tests, 45 assertions)`.
2. Regression: `n test-php --filter Test_Content_Gates` → `OK (93 tests, 330 assertions)`, and the full suite → `Tests: 2439, Assertions: 6950, Skipped: 7`.
3. `./vendor/bin/phpcs` on the changed files → clean.

**End-to-end (WCM fixture):**

4. Create a signup Membership plan restricting a category, plus a Primary `np_memberships_gate` post with top-level `non-member-content` / `member-content` wrapper blocks.
5. Dry-run: `wp newspack migrate-membership-gates` → reports `DRY RUN MODE`, the plan as `created (dry-run)`, gate `(pending)`, and writes **no** gate.
6. Live: `wp newspack migrate-membership-gates --live` → creates the content gate (its `content_rules` mirror the plan's restriction), a registration layout carrying the extracted `non-member-content` inner block (wrapper stripped), and leaves the paid-access layout at its default (signup plan, so `custom_access` is not written).
7. Re-run `--live` → reports `updated`, no duplicate gate/layouts (confirms the `get_gates()` dedup).
8. Activate the drop-in and re-run → confirm the collision `WP_CLI::warning` fires and the dry-run-default (port) behavior still wins. Run an unrelated command (`wp cache flush`) → no warning, since the notice is hooked to `before_invoke` on this command only. Deactivate the drop-in again.
9. Purchase-plan check: create a plan granted by purchasing a product, with a `member-content` wrapper in its gate, and run `--live` → the gate's paid-access mode is active with a `subscription` rule for the parent product. Remove the `member-content` wrapper from the gate post and re-run against a second plan → that gate is reported as a `WARN` row saying its paid access mode is not active, rather than a plain `created`.

### Other information:


plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+					array_merge( ...array_column( $group, 'product_ids' ) )
+				)
+			);
+			// Drop product variations — gates should reference parent products only.
+			$merged_product_ids = array_values(
+				array_filter(
+					$merged_product_ids,
+					fn( $id ) => 'product_variation' !== \get_post_type( $id )
+				)
+			);
```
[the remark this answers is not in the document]
adekbadek: This mirrors the standalone drop-in verbatim — variations are dropped so gates reference parent products only. Consolidating the merged product IDs (including mapping variations to their parent so a variation-only purchase plan doesn't end up with empty access rules) is deliberately out of scope for this faithful port: it's owned by the follow-up **NPPD-2064** (the merged-product consolidation), per the "known bugs preserved by design" section of the PR description. Leaving it as-is keeps the characterization baseline stable for that stacked fix.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+				// Set content rules.
+				\Newspack\Content_Rules::update_gate_content_rules( $gate_id, $ac_rules );
+
+				// Registration layout (always).
+				self::apply_layout( $gate_id, $gate_title, 'registration', $layouts['registration'] );
+
+				// Custom access layout (purchase plans only).
+				if ( $has_purchase && null !== $layouts['custom_access'] ) {
+					self::apply_layout( $gate_id, $gate_title, 'custom_access', $layouts['custom_access'], $merged_product_ids );
+				}
```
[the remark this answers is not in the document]
adekbadek: Correct — and this is the deliberately-preserved **NPPD-2058** behavior called out in the PR description. `extract_gate_layouts()` only inspects top-level wrapper blocks, so a nested/reusable `member-content` wrapper (or a missing one) yields a `null` custom_access and the purchase restriction isn't written. The fix — walking nested/reusable blocks so the wrapper is found — lands in the stacked NPPD-2058 PR; porting the behavior as-is here is intentional so that PR gets a clean red→green signal from the characterization tests.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+							[
+								[
+									'slug'  => 'subscription',
+									'value' => $product_ids,
+								],
+							],
+						]
+						: [],
+				]
+			);
```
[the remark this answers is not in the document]
adekbadek: This is faithful to the drop-in (`active => true` with `access_rules` defaulting to `[]` when there are no product IDs). For a purchase plan, `$product_ids` goes empty only via the variation-dropping in the merged-product consolidation, which is owned by the follow-up **NPPD-2064** (see the PR description's known-bugs section); its fix maps variations to their parent so the product IDs — and thus `access_rules` — are populated. Since this is a faithful port I'm not diverging the active-flag behavior here; the stacked PR addresses the root cause.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+							'access_type'   => $access_type,
+						];
+						continue;
+					}
+					$gate_id = $result;
+				}
+				$existing_gates[ $gate_key ] = $gate_id;
+
+				// Set content rules.
+				\Newspack\Content_Rules::update_gate_content_rules( $gate_id, $ac_rules );
```
miguelpeixe, line 240: **Blocker:** content_rules_match is never written — AC's default AND inverts WCM's union semantics for multi-rule plans — WC Memberships restricts the **union** of a plan's content-restriction rules: content matching *any* rule is members-only. The migrated gate gets its rules via `Content_Rules::update_gate_content_rules()` but `content_rules_match` is never set, so `get_gate_content_rules_match()` returns the default `'all'` and `Content_Restriction_Control::get_post_gates()` combines inclusion rules with AND. A plan restricting category X and tag Y (both slugs survive as valid taxonomy rules even before the NPPD-2063 remap) migrates into a gate that only applies to posts carrying **both** — everything else silently un-gates at cutover. The AND default also compounds the known slug mistranslation: one dead `post`/`page` rule in a gate makes the *whole* gate apply nowhere, instead of just that one rule going inert. Writing `Content_Rules::update_gate_content_rules_match( $gate_id, 'any' )` next to the rules write restores WCM's semantics — or, if this too is meant to be preserved-and-fixed-later, it needs an owning ticket like the other three (none of NPPD-2058/2063/2064 covers it).

**Risk if not addressed:** If a plan restricts all posts, or an all-posts rule combined with a category rule, and the site cuts over by deactivating WooCommerce Memberships, the migrated gate matches no content, so the whole set of articles that plan used to paywall becomes freely readable until someone notices and rebuilds the gate by hand.

**Consider:** WooCommerce Memberships combines a plan's restriction rules as a union — content is restricted if it matches any rule — but the gate evaluator defaults an unset match mode to AND (`content_rules_match ?? 'all'`). Writing `content_rules_match => 'any'` alongside the content rules here would reproduce the source semantics, so a multi-rule gate no longer requires a post to match every rule at once. Note this only fixes the combination logic: a gate whose sole rule uses a slug the evaluator can't resolve (the raw 'post'/'page' names) still matches nothing, so it does not remove the need for the slug remapping in the stacked fix.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			);
+			// Drop product variations — gates should reference parent products only.
+			$merged_product_ids = array_values(
+				array_filter(
+					$merged_product_ids,
+					fn( $id ) => 'product_variation' !== \get_post_type( $id )
+				)
+			);
+
+			$action  = isset( $existing_gates[ $gate_key ] ) ? 'updated' : 'created';
```
miguelpeixe: **Suggestion:** Title-keyed gate identity vs fingerprint grouping — same-name plans clobber each other on --live — Plans are grouped by rule *fingerprint*, but the written gate's identity is its lower-cased *title*. Two plans that share a name but have different rules form two groups with the same `$gate_key`: the first creates the gate and writes its ID into `$existing_gates` (line 193); the second finds it, takes the `updated` branch, and `update_gate_content_rules()` + `apply_layout()` overwrite the first group's rules and layouts — last writer wins, no warning, one plan's restriction config silently lost. (The "theoretically impossible given unique fingerprints" comment at line 153 doesn't hold: titles derive from plan names, not fingerprints.) In dry-run, the second group also prints the `-1` sentinel in the Gate ID column. Duplicate plan names are uncommon but WC permits them — consider a `WP_CLI::warning` when a key seen earlier in the same run recurs, and printing `(pending)` instead of `-1`.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+					'plan_name'     => $plan_name,
+					'action'        => 'skipped (no restrictions)',
+					'gate_id'       => '—',
+					'content_rules' => '0',
+					'access_type'   => $access_method,
+				];
+				continue;
+			}
+
+			// Note: signup plans with no content rules all share the same empty
```
miguelpeixe: **Suggestion:** Rules-less signup plans migrate to gates the evaluator skips — the "applies site-wide" comment is wrong — `group_plans_by_fingerprint()` deliberately lets a `signup` plan with zero content rules through, and the comment asserts the resulting gate "applies site-wide". The evaluator does the opposite: in `Content_Restriction_Control::get_post_gates()`, a gate whose `content_rules` is empty sets no `$has_non_specific_rule` and is skipped for every post — the migrated gate is published with `registration.active => true` yet enforces nothing, while the summary reports it as created. The outcome incidentally matches WCM (a plan with no restriction rules restricts nothing), but the comment documents the opposite semantics and the summary gives the operator no hint the gate is inert. Worth either skipping these plans like non-signup ones, or fixing the comment and marking the summary row (e.g. `created (no rules — inert)`).

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			WP_CLI::line( $plan_id ? sprintf( 'No published plan found with ID %d.', $plan_id ) : 'No published membership plans found.' );
+			return;
+		}
+
+		WP_CLI::line( sprintf( 'Found %d membership plan(s). Starting migration…', $total ) );
+		WP_CLI::line( '' );
+
+		// Pre-load existing gates indexed by lower-cased title.
+		$existing_gates = [];
+		foreach ( \Newspack\Content_Gate::get_gates() as $gate ) {
```
miguelpeixe: **Suggestion:** Existing-gate dedup indexes gates in all statuses — a draft/trashed title match is "updated" but never enforces — The dedup preload calls `Content_Gate::get_gates()` with no status argument, which fetches `['publish','draft','pending','future','private','trash']`. A pre-existing gate in any of those statuses whose lower-cased title matches becomes the write target: `--live` takes the `updated` path, writes rules and layouts into it, and creates nothing new — but the frontend only enforces `publish` gates, so the "updated" gate enforces nothing (a trashed gate written to this way also stays trashed). Scope the preload to `Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish' )` to match what the command creates.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+	private static function extract_gate_layouts( \WP_Post $gate_post ): array {
+		$blocks               = \parse_blocks( $gate_post->post_content );
+		$registration_markup  = null;
+		$custom_access_markup = null;
+
+		foreach ( $blocks as $block ) {
+			if ( 'woocommerce-memberships/non-member-content' === $block['blockName'] && null === $registration_markup ) {
+				$registration_markup = self::serialize_gate_inner_blocks( $block['innerBlocks'] );
+			}
+			if ( 'woocommerce-memberships/member-content' === $block['blockName'] ) {
```
miguelpeixe, line 956: **Suggestion:** extract_gate_layouts() is first-wins for non-member-content but last-wins for member-content — The `non-member-content` branch is guarded by `null === $registration_markup` (first top-level occurrence wins); the `member-content` branch has no guard, so each occurrence overwrites the previous (last wins). A gate post interleaving multiple top-level wrapper blocks of the same type — a realistic WCM authoring pattern for posts mixing public and members-only sections — silently loses all but one block's content, differently per wrapper type, and nothing surfaces the loss before `--live`. This is distinct from NPPD-2058 (nested wrappers): a fix that walks nested blocks won't resolve the asymmetry. Apply the same null guard to both branches (or make both concatenate), and pin the choice with a two-wrapper fixture — no current test covers multiple top-level wrappers of the same type.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+		if ( ! empty( $plan_gates ) ) {
+			return $plan_gates[0];
+		}
+
+		if ( ! $primary_fallback ) {
+			return null;
+		}
+
+		// Fall back to the Primary gate (no `plans` meta).
+		$primary_gates = \get_posts(
```
miguelpeixe, line 882: **Suggestion:** Primary-gate fallback bypasses the canonical newspack_memberships_gate_post_id option — The plugin identifies the Primary memberships gate by option: `Memberships::get_gate_post_id()` reads `newspack_memberships_gate_post_id`. "Has no `plans` meta" is only an incidental property of how that gate is created, and the `NOT EXISTS` meta query returns the *newest* plan-less published gate — which diverges whenever more than one plan-less gate post exists (e.g. untrashed leftovers) or the recorded primary gate later acquired meta. Now that this code lives inside the plugin, prefer the option (falling back to the meta query only when the option is unset or stale).

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+		}
+
+		if ( $layout_id ) {
+			// The overwrite is unconditional even when $content is '' — mirroring the
+			// drop-in. Content_Gate::create_gate() seeds a default block-pattern layout,
+			// which this replaces with the migrated markup; an empty $content (no
+			// np_memberships_gate found for the group, or a nested/reusable wrapper) thus
+			// blanks that default. Preserving those defaults on empty content is the
+			// empty-layout fix tracked in NPPD-2058; kept faithful here.
+			\wp_update_post(
```
miguelpeixe: **Suggestion:** Without --user, kses strips embeds/iframes/custom HTML from migrated layout content — `wp_update_post()` / `create_gate_layout()` → `wp_insert_post()` run `post_content` through `content_save_pre`; in a WP-CLI run with no `--user` there is no `unfiltered_html` capability, so `wp_filter_post_kses` is active. Gate content authored by admins (who saved unfiltered) can legitimately contain iframes/embeds/Custom-HTML blocks that this re-save silently strips — the migrated layout won't match the source `np_memberships_gate` content. Either hard-require (or at least document) `--user=<admin>` in the preflight, or wrap the writes in `kses_remove_filters()` / `kses_init_filters()`.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+	 *     wp newspack migrate-membership-gates --plan=711923
+	 *
+	 * @param array $args       Positional args (unused).
+	 * @param array $assoc_args Named args.
+	 *
+	 * @return void
+	 */
+	public function migrate_membership_gates( $args, $assoc_args ) {
+		$dry_run = ! (bool) \WP_CLI\Utils\get_flag_value( $assoc_args, 'live', false );
+		$plan_id = (int) \WP_CLI\Utils\get_flag_value( $assoc_args, 'plan', 0 );
```
miguelpeixe: **Suggestion:** Non-numeric --plan silently widens a --live run from one plan to all plans — `(int) get_flag_value( $assoc_args, 'plan', 0 )` turns `--plan=abc` (or a stray character from copy-paste) into `0`, which `get_plans()` treats as "no filter" — the one flag meant to *narrow* a `--live` run instead migrates every plan, with no warning. Validate that the raw value is numeric when the flag is present and `WP_CLI::error()` otherwise (the teams commands error on an invalid `--product-id`).

**Consider:** `(int) get_flag_value( ..., 'plan', 0 )` turns any non-numeric `--plan` value into 0, which get_plans() treats as 'no filter' — so a `--live` run an operator meant to scope to a single plan would process every plan instead. Validating that `--plan`, when present, parses to a positive integer that resolves to an existing plan (and erroring otherwise) would keep a mistyped scope from silently widening the write.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+	 * @param array $skipped  Skipped-plan summary rows, appended to by reference.
+	 *
+	 * @return array<string,array> Map of fingerprint => list of plan descriptors, each
+	 *                             [ 'pid', 'name', 'access_method', 'ac_rules', 'product_ids' ].
+	 */
+	private static function group_plans_by_fingerprint( array $plan_ids, array &$skipped ): array {
+		$plan_groups = [];
+
+		foreach ( $plan_ids as $pid ) {
+			$plan          = new \WC_Memberships_Membership_Plan( $pid );
```
miguelpeixe: **Suggestion:** Prefer the wc_memberships_get_membership_plan() factory over direct construction — The rest of the plugin resolves plans through the factory (e.g. `Memberships::get_gate_plans()`), which validates the post and lets WC Memberships' integrations substitute subclasses — the WC Subscriptions integration swaps in its subscription-aware plan class. Direct `new \WC_Memberships_Membership_Plan( $pid )` happens to work for the four methods used here, but the factory returns a falsy value for an invalid post instead of a half-built object and keeps this file consistent with the plugin's other WC Memberships call sites.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			$layout_id = \Newspack\Content_Gate::create_gate_layout(
+				sprintf( '%s — %s', $gate_title, $label ),
+				$content
+			);
+			if ( \is_wp_error( $layout_id ) ) {
+				WP_CLI::warning( sprintf( 'Could not create %s for "%s": %s', strtolower( $label ), $gate_title, $layout_id->get_error_message() ) );
+			}
+		}
+
+		$resolved_layout_id = \is_wp_error( $layout_id ) ? 0 : $layout_id;
```
miguelpeixe: **Suggestion:** Layout-creation failure still activates the mode with gate_layout_id => 0 — a gate that never restricts — When `create_gate_layout()` returns a `WP_Error`, `$resolved_layout_id` becomes `0` but the settings write still happens with `active => true`. `Content_Restriction_Control::is_post_restricted()` requires a truthy `gate_layout_id`, so the gate silently never restricts — and only a mid-run `WP_CLI::warning` (not the summary table) records why. The `wp_update_post()` return on the overwrite path is likewise unchecked (a stale layout ID pointing at a deleted post becomes a silent no-op). Consider surfacing layout failures as ERROR rows in the summary and skipping the `active => true` write when there is no layout to point at.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+		if ( ! function_exists( 'wc_memberships' ) ) {
+			WP_CLI::error( 'WooCommerce Memberships is not active. Aborting.' );
+		}
+
+		// The standalone `migrate-memberships` drop-in registers the same command
+		// name (defaulting to WRITE — the inverse of this port's dry-run default).
+		// This port's `init` registration overrides the drop-in's `plugins_loaded`
+		// one, so this callback is the one running — but the ambiguity is a footgun,
+		// so warn loudly and tell the operator to deactivate the drop-in.
+		if ( class_exists( 'Newspack_Migrate_Membership_Gates_Command' ) ) {
```
miguelpeixe: **Nit:** Drop-in collision warning only fires when the port already won the registration race — The warning runs inside `migrate_membership_gates()` — i.e. only when this port's callback is the one WP-CLI actually dispatched. `WP_CLI::add_command()` is last-registration-wins with no error, and `init` fires after `plugins_loaded`, so the port does win today; but if that ordering ever changes, the drop-in's write-by-default command runs and this guard never fires — the exact footgun it exists to catch. A registration-time guard (refuse, or actively unregister, when `Newspack_Migrate_Membership_Gates_Command` exists) fails safe regardless of order.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			$settings  = \Newspack\Content_Gate::get_custom_access_settings( $gate_id );
+			$layout_id = $settings['gate_layout_id'] ?? 0;
+			$label     = 'Paid Access Layout';
+		} else {
+			$settings  = \Newspack\Content_Gate::get_registration_settings( $gate_id );
+			$layout_id = $settings['gate_layout_id'] ?? 0;
+			$label     = 'Registration Layout';
+		}
+
+		if ( $layout_id ) {
```
miguelpeixe: **Nit:** Re-running --live overwrites operator-edited layouts without warning — `apply_layout()` unconditionally replaces the existing layout's `post_content` on every run. An operator who migrates, customizes the generated Registration/Paid-Access layout in the admin, then re-runs (say, after adding a plan) loses those edits to freshly re-extracted markup — or to `''` when no `np_memberships_gate` is found. Inherent to a re-runnable migration and matching the drop-in, but worth an explicit "re-running is not edit-preserving" note in the `## OPTIONS` docblock.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			// fingerprint and are merged into a single gate with a combined title.
+			// This is intentional — a free-registration gate that restricts no
+			// specific content applies site-wide.
+			$fingerprint                   = self::compute_rules_fingerprint( $ac_rules );
+			$plan_groups[ $fingerprint ][] = [
+				'pid'           => $pid,
+				'name'          => $plan_name,
+				'access_method' => $access_method,
+				'ac_rules'      => $ac_rules,
+				'product_ids'   => 'purchase' === $access_method ? array_values( $plan->get_product_ids() ) : [],
```
miguelpeixe, line 607: **Nit:** Cast merged product IDs with absint for parity with the REST sanitizer — The REST write path stores subscription access-rule values as ints (`Content_Gate_API::sanitize_access_rule()` int-casts numerics), while `$plan->get_product_ids()` returns raw `_product_ids` meta that can hold strings depending on WCM version. Evaluation is safe (`WC_Subscription::has_product()` compares loosely), but string IDs can strict-mismatch the wizard's product multi-select options until the gate is re-saved through REST. `array_map( 'absint', … )` in the merge closes the gap.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
+<?php
+/**
+ * WP-CLI command to migrate WooCommerce Membership plans and their content gate
+ * layouts to Newspack Access Control content gates.
+ *
+ * Ported from the standalone `migrate-memberships` drop-in so the tooling ships
+ * with the plugin (the CLI class loads only under WP_CLI, so there is no
```
miguelpeixe: **Nit:** File docblock claims CLI classes load only under WP_CLI — they load on every request — `Newspack::includes()` calls `CLI\Initializer::init()` unconditionally, and `init()` `include_once`s every CLI class file on all requests — only `register_comands()` early-returns without `WP_CLI`. The cost is just a class definition, but the header asserts a load gate that doesn't exist; worth tweaking the wording so future readers don't rely on it.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+		);
+
+		WP_CLI::line( '' );
+		$processed = count(
+			array_filter(
+				$summary,
+				fn( $r ) => ! str_starts_with( $r['action'], 'ERROR' )
+			)
+		);
+		WP_CLI::success(
```
miguelpeixe, line 329: **Suggestion:** Consider a post-write verification step on `--live`: after each gate is written, re-read it and confirm it is actually enforceable — published status, at least one content rule the evaluator can resolve, and every active mode pointing at a non-zero layout id — then surface any gate that fails those checks as an explicit WARN or FAIL row. As written, the action column and the final success count reflect only that the writes did not error, not that any gate will restrict once WooCommerce Memberships is deactivated. Because the migrated gates stay dormant until that cutover, a gate that enforces nothing can be reported as a success today and go unnoticed for weeks.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,625 @@
...
+			\Newspack\Content_Gate::update_custom_access_settings(
+				$gate_id,
+				[
+					'active'         => true,
+					'gate_layout_id' => $resolved_layout_id,
+					'access_rules'   => ! empty( $product_ids )
+						? [
+							[
+								[
+									'slug'  => 'subscription',
```
miguelpeixe, line 742: **Suggestion:** The only access rule emitted for purchase plans is 'subscription', which is satisfied only by an active WooCommerce Subscription (or group subscription). WooCommerce Memberships can also grant a plan through a one-time purchase of a simple product, and those members hold no subscription — so at cutover this gate would treat them as non-payers and paywall content they already bought. Before relying on this at cutover, either confirm the target site grants membership only via subscription products, or extend the access rules to also recognize a completed purchase of the plan's product(s).

miguelpeixe (changes requested):
Request changes, narrowly: this is a careful, honestly-annotated port — dry-run purity holds, the data shapes written match what the content-gate subsystem reads, and the characterization tests genuinely pin the seams to their tickets. But the migration never writes `content_rules_match`, and Access Control's default combines inclusion rules with AND while WC Memberships restricts by the union of a plan's rules — so any multi-rule plan under-gates after cutover, and a single dead rule (the known slug mistranslation) silences its whole gate. That's a one-line fix, or it needs an owning ticket like the other preserved bugs. The remaining suggestions are robustness hardening around gate identity (title collisions, non-publish dedup targets) and operator safety (a `--plan` typo widens to all plans; kses stripping without `--user`); none block on their own.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: `Content_Rules::update_gate_content_rules_match( $gate_id, 'any' )` is now written alongside the rules, so the migrated gate reproduces WCM's union semantics instead of the evaluator's AND default. Went with the fix rather than a fourth ticket — the AND default is a mistranslation of the source data, not a drop-in behavior worth preserving, and as you note it also compounds the NPPD-2063 slug bug.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed both halves. A `$titles_written` set now tracks gate keys claimed earlier in the same run and emits a `WP_CLI::warning` naming the colliding title when one recurs. The `-1` sentinel is gone: the claim is recorded as `null` and the lookup switched from `isset()` to `array_key_exists()`, so the summary's existing `?? '(pending)'` renders correctly in dry-run instead of printing `-1`. The stale "theoretically impossible given unique fingerprints" comment is replaced with what the null actually means.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Took the skip option: the `'signup' !== $access_method` exception is gone, so any plan with zero content rules lands in $skipped as `skipped (no restrictions)`. The false "applies site-wide" comment is replaced with why — an empty-rules gate sets no \$has_non_specific_rule and is skipped for every post, so migrating one only published an inert gate the summary reported as created, and WCM restricted nothing for that plan anyway. Flagging the behavior change explicitly in case the exception was deliberate for a publisher case I can't see: the alternative (create it, mark the row inert) is a one-line change if you'd rather keep the gate as a starting point.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: the preload is now `Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish' )`, matching the status the command creates, so a draft/pending/trashed title match no longer becomes the write target.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed by concatenating rather than first-wins, for both branches — repeated top-level wrappers of either type are appended in document order, so no authored content is dropped (first-wins would still have lost the later wrappers). Pinned by a new test with two wrappers of each type interleaved, which also asserts document order. Left the NPPD-2058 nested-wrapper test untouched.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: the fallback now reads `Memberships::get_gate_post_id()` first and uses that post when it is a published `np_memberships_gate`, dropping to the `NOT EXISTS` meta query only when the option is unset or stale. Docblock updated to describe the option as the primary source.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Wrapped the writes instead of requiring `--user`: `apply_layout()` checks `has_filter( 'content_save_pre', 'wp_filter_post_kses' )`, calls `kses_remove_filters()` around the create/update, and restores with `kses_init_filters()` only if it was active. That keeps admin-authored embeds/iframes/Custom HTML intact without depending on the operator remembering a flag.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: `--plan` is read raw (default `null`) and, when present, must be numeric and positive or the command `WP_CLI::error`s. A non-numeric value can no longer degrade to 0 and widen a `--live` run to every plan. A plan ID that parses but resolves to nothing still hits the existing "No published plan found with ID %d." early return.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: switched to `wc_memberships_get_membership_plan( $pid )`. A falsy return now produces a `skipped (not a valid membership plan)` summary row instead of a half-built object, and integrations that substitute plan subclasses are respected.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: `apply_layout()` now returns bool and skips the settings write entirely when there is no usable layout post, so a mode is never activated with `gate_layout_id => 0`. The overwrite path passes `true` to `wp_update_post()` and treats a WP_Error or 0 (stale ID pointing at a deleted post) as a failure, warning with the post ID. The caller collects failures per gate and renders the row as `ERROR: could not write registration layout` / `… + paid access layout`, which the existing `str_starts_with( 'ERROR' )` filter already excludes from the processed count.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Moved the guard to registration time in `class-initializer.php`: the warning now fires when the command is registered, so the operator is told regardless of which registration wins the race — including the failure mode you describe, where the drop-in's write-by-default callback is the one dispatched and the in-command check would never run. Removed the in-command copy so it isn't emitted twice. Didn't attempt to unregister the drop-in's command: WP_CLI has no supported removal path, and warning at registration covers the ordering risk.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Added to the command docblock, just above `## OPTIONS`: re-running is not edit-preserving — a title-matched gate has its content rules and layout content overwritten with freshly extracted markup, so post-migration admin edits are lost. Also called out in the PR description's operational note.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed: `array_map( 'absint', … )` now runs inside the merge, before the dedupe and the variation filter, so the stored subscription rule values are ints and match the wizard's product options without a REST re-save.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed the wording: the header now says the file is included on every request like the other CLI classes, and that only the command registration is gated on WP_CLI, so the cost outside CLI is a class definition. Same correction applied to the equivalent claim in the PR description.

adekbadek:
All addressed in 7e47e19. The blocker is fixed at the source: `content_rules_match => 'any'` is written alongside the rules, so the migrated gate reproduces WCM's union semantics rather than needing a fourth preserved-bug ticket. The robustness and operator-safety items are all fixed too — `--plan` validation, publish-only dedup, duplicate-title warning, kses bypass on layout writes, no mode activation without a layout (surfaced as an ERROR row), the canonical Primary-gate option, and the plan factory.

Two departures beyond what you flagged are worth your eye: `extract_gate_layouts()` now concatenates repeated top-level wrappers rather than first-wins (no authored content dropped; new test pins it), and rules-less `signup` plans are now skipped rather than migrated into an inert gate — I took the skip option there, so push back if that exception was deliberate. NPPD-2058/2063/2064 remain preserved and pinned as before. Full suite green: 2111 tests.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Added in 58b26c82cd. On `--live` each written gate is re-read by `verify_migrated_gate()`, which checks the same conditions `Content_Restriction_Control::get_post_gates()` applies: published status, at least one content rule the evaluator can resolve, at least one active mode, and no active mode with a zero `gate_layout_id`. Failures are surfaced as a `WARN:` row (distinct from the existing `ERROR:` rows for write failures), a per-gate `WP_CLI::warning`, and a count after the success line.

The rule-resolvability check turned out to be the part that earns its keep, because it makes the deliberately-preserved NPPD-2063 slug mistranslation loud instead of silent: `rule_matches_post()` handles `post_types` / `specific_posts` / `newsletters` by name and treats everything else as a taxonomy, so a migrated rule with slug `post` falls to `get_taxonomy( 'post' )` → null and matches nothing. Verified end-to-end against a WCM fixture:

- category restriction (resolvable) → `updated`, no warning
- post-type restriction with object IDs → `WARN: none of its content rules resolve to a post type or taxonomy the evaluator can match (post)`
- post-type restriction with no object IDs → `WARN: none of its content rules select any content` (the empty value is dropped by `get_gate_content_rules()`, so the gate evaluates as having no rules at all while the summary's Content Rules column still reads 1 – the message distinguishes that from a gate genuinely written with none)

Four unit tests cover the branches (`n test-php --filter Test_Membership_Gates_Migration` → OK, 14 tests, 34 assertions).

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Correct, and it's already ticketed – leaving it out of this PR deliberately rather than papering over it here.

The gap is structural, not a mapping mistake in this CLI: `Access_Rules::register_default_rules()` registers only `subscription`, `email_domain`, `reader_data` and `institution`, and `has_active_subscription()` requires a real WooCommerce Subscription. There is no rule that can express "completed a one-time purchase of product X", so `subscription` is the only thing this command has to emit today.

That parity gap is [NPPD-2053](https://linear.app/a8c/issue/NPPD-2053) (in review – adds a `one_time_purchase` paid-access rule: product IDs plus a duration anchored on the order date, covering prepaid and lifetime plans as well as plain one-offs). The migration side is [NPPD-2106](https://linear.app/a8c/issue/NPPD-2106), which is specifically about translating a WCM simple-product plan – products plus the plan's membership length, unlimited mapping to `forever` – onto that new rule, and it names this NPPD-2059 stack as one of the two candidate homes for the work. Both were opened off the field parity audits that surfaced exactly the cohort you describe, so the sizing evidence is on the ticket rather than being re-derived here.

Emitting a warning for it in the meantime would be guesswork about the fix's eventual shape, so it stays with NPPD-2106.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2059-migration-tooling-port-migrate-membership-gates-into-the-in` (stash local work first).

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,851 @@
...
+				\Newspack\Content_Rules::update_gate_content_rules( $gate_id, $ac_rules );
+				\Newspack\Content_Rules::update_gate_content_rules_match( $gate_id, 'any' );
+
+				// Registration layout (always).
+				if ( ! self::apply_layout( $gate_id, $gate_title, 'registration', $layouts['registration'] ) ) {
+					$layout_errors[] = 'registration layout';
+				}
+
+				// Custom access layout (purchase plans only).
+				if ( $has_purchase && null !== $layouts['custom_access'] ) {
```
miguelpeixe, line 250: **Blocker:** A purchase plan can migrate to a gate that any registered reader passes — and verification reports it as a clean success.

Paid access is only applied when `$has_purchase && null !== $layouts['custom_access']`. Two reachable paths defeat that:

1. **No top-level `member-content` wrapper** — no `np_memberships_gate` post for the group at all, or the NPPD-2058 nested/reusable case. `custom_access` is never activated, so the gate is registration-only.
2. **`$merged_product_ids` is empty** — every product ID was a variation (the deferred Copilot thread) or the plan has no products. The mode *is* activated, but with `'access_rules' => []`, and `Content_Restriction_Control:412` guards on `! empty( $access_rules )` — an active paid mode with no rules restricts nobody.

Either way, a logged-in reader passes registration mode (`require_verification` is never written, so the `elseif` at `Content_Restriction_Control:399` never fires), custom access does not stop them, and `is_post_restricted()` returns false.

`verify_migrated_gate()` checks published / has-rules / has-an-active-mode / no-active-mode-with-zero-layout. All four pass. The summary row prints `created` with Access Type `purchase`, which actively reassures the operator that the paid plan migrated.

**Risk if not addressed:** at cutover, content that previously required an actual purchase becomes readable by anyone who registers a free account — and the migration report said it was fine. That is the failure mode this verification step exists to prevent, and it is worse than the inert-gate case it does catch, because an inert gate at least produces visible complaints.

**Consider:** pass `$has_purchase` into `verify_migrated_gate()` and add two checks — a purchase-derived gate whose paid-access mode is inactive, and one whose `access_rules` is empty. Both are one-liners against data already in scope here.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,851 @@
...
+		);
+		if ( empty( $content_rules ) ) {
+			// get_gate_content_rules() drops empty-value rules, so a gate can be written
+			// with rules and still evaluate as having none — say which of the two it is,
+			// because the summary's Content Rules column reports the pre-write count.
+			$written_rules = \get_post_meta( $gate_id, 'content_rules', true );
+			$issues[]      = empty( $written_rules )
+				? 'it has no content rules'
+				: 'none of its content rules select any content';
+		} elseif ( count( $unresolvable ) === count( $content_rules ) ) {
```
miguelpeixe, line 386: **Suggestion:** Only flagging when *every* rule is unresolvable misses the common case.

`count( $unresolvable ) === count( $content_rules )` means a gate warns only when it matches nothing at all. A plan restricting all posts *and* a category maps to `post` (dead under NPPD-2063) plus `category` (live). Under the new `any` match mode the gate covers the category and everything else un-gates — a partial leak, reported clean.

That shape (a post-type rule alongside a taxonomy rule) is one of the most common ways WCM plans are configured, so a large share of real migrations will pass verification while under-gating. Since these gates lie dormant until cutover, nobody finds out until then.

Warning whenever *any* slug is unresolvable — still a `WARN`, not an `ERROR`, and naming the dead slugs — would make the NPPD-2063 blast radius visible per gate instead of only in the all-or-nothing case. It also gives the NPPD-2063 PR a natural signal to flip.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,851 @@
...
+
+		// Gate content authored by an admin with unfiltered_html can legitimately
+		// contain iframes/embeds/Custom HTML. A WP-CLI run has no user, so kses would
+		// strip those on re-save and the migrated layout would not match its source.
+		$kses_was_active = \has_filter( 'content_save_pre', 'wp_filter_post_kses' );
+		if ( $kses_was_active ) {
+			\kses_remove_filters();
+		}
+
+		if ( $layout_id ) {
```
miguelpeixe: **Suggestion:** An empty extracted layout blanks the seeded default, and verification passes it.

Because `Content_Gate::create_gate()` always seeds both layout posts, `$layout_id` is truthy for every freshly created gate — so this overwrite branch is the normal path, not the exception, and empty `$content` blanks the default block-pattern layout. Note `create_gate_layout()` does the opposite for the same input: it substitutes `get_default_gate_content()` when content is empty. So the create and update paths disagree, and in practice only the blanking one runs.

`verify_migrated_gate()` only checks that `gate_layout_id` is truthy, not that the layout has content, so this passes. The reader gets a truncated article with an empty gate underneath — no registration form, no upsell, no explanation.

NPPD-2058 owns the *extraction* fix, but the "no `np_memberships_gate` found for this group" path is not a nested-wrapper problem and stays broken after it lands. Either skip the content overwrite when `'' === $content` (keeping the seeded default as a usable fallback), or add the empty-layout check to `verify_migrated_gate()`.

plugins/newspack-plugin/includes/cli/class-initializer.php
```diff
@@ -101,6 +102,15 @@ public static function register_comands() {
 			WP_CLI::add_command( 'newspack migrate-team-products', [ 'Newspack\CLI\Teams_Migration', 'migrate_team_products' ] );
 			WP_CLI::add_command( 'newspack migrate-manual-members', [ 'Newspack\CLI\Teams_Migration', 'migrate_manual_members' ] );
 			WP_CLI::add_command( 'newspack backfill-team-managers', [ 'Newspack\CLI\Teams_Migration', 'backfill_team_managers' ] );
+			// The standalone `migrate-memberships` drop-in registers the same command
+			// name with the opposite, write-by-default flag convention. Registration is
+			// last-wins with no error, so warn here — at registration time, before any
+			// command is dispatched — rather than inside the callback, which only runs
+			// when this registration happens to be the one that won.
+			if ( class_exists( 'Newspack_Migrate_Membership_Gates_Command' ) ) {
```
miguelpeixe, line 117: **Nit:** The registration-time guard now fires on every WP-CLI invocation, not just this command.

`register_comands()` runs on `init` for every `wp` call, so on a site with the drop-in active this warns on `wp cron event run`, `wp cache flush`, deploy scripts — everything. That is the cost of the move I asked for and probably still the right trade over the ordering footgun, but it is worth bounding: checking the runner's arguments for `migrate-membership-gates` before warning keeps the fail-safe property while confining the noise to the command it is about.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -63,6 +63,21 @@ public static function init() {
...
 
+	/**
+	 * Discard the request-scoped caches.
+	 *
+	 * The caches are keyed by post ID and assume the gate configuration does not
+	 * change for the lifetime of the request. Callers that write gates and then
+	 * evaluate them in the same process — or that reuse post IDs, as the test
+	 * suite does when it rolls back each case — must discard them first.
+	 */
+	public static function flush_request_cache() {
```
miguelpeixe: **Nit:** New public production API with zero production callers.

`flush_request_cache()` is called only from two test `set_up()` methods, in a commit typed `test(...)`. Its docblock states an invariant — "callers that write gates and then evaluate them in the same process ... must discard them first" — that nothing enforces: none of the gate write paths call it, and neither does the CLI in this PR (it happens not to need it, since it never goes through `get_post_gates()`).

The precedent for keeping this out of the production surface is four lines below the call site in `class-premium-newsletters.php`, which resets a comparable private static via `ReflectionProperty`. If the invariant is real, the write paths should flush; if it is test isolation, reflection keeps it in the test layer.

miguelpeixe (changes requested):
Re-reviewed at `790569f3`. All 16 threads from the last round are genuinely addressed — I checked each against the code and the APIs it calls, not just the replies. A few I want to explicitly confirm because they depended on details outside the diff:

- `content_rules_match => 'any'` is read by the evaluator at `Content_Restriction_Control:208`, and it is a first-class wizard field, so an admin opening the migrated gate and saving will not silently reset it to AND.
- `Memberships::get_gate_post_id()` degrades to the plain option in CLI (`is_singular()` is false there), so the Primary-gate fallback resolves the way the docblock says.
- The kses save/restore is faithful in both the `--user` and no-user cases — `kses_init_filters()` re-adds `pre_comment_content` with the same handler the CLI context had.

The new post-write verification is a real improvement. But it answers **"will this gate restrict anything?"** when the question the operator needs answered at cutover is **"will it restrict what WooCommerce Memberships restricted?"** Three cases pass `verify_migrated_gate()` clean, land in the success count, and under-gate. The first is a paid-content leak, so flagging as changes requested; the other two are cheap additions to the same method.

Everything else below is a nit.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in 01fefc4. `verify_migrated_gate()` now takes the group's `$has_purchase` and flags both paths: an inactive paid-access mode ("it migrates a plan that requires a purchase, but its paid access mode is not active — any registered reader would get in") and an active one with an empty `access_rules` ("…so it asks for no purchase…"). Both surface as WARN rows and as per-gate warnings, so the summary can no longer print a clean `created` / `purchase` for content a free account unlocks.

I kept this at verification rather than auto-activating paid access with the seeded default layout: which upsell a publisher shows behind a paywall is their call, and the operator has to author it either way — so the useful output is a loud, itemised list of the gates that need it, not a gate that silently restricts with generic Newspack copy.

Three tests cover it: the inactive-mode case (asserting the same gate verifies clean when `$has_purchase` is false, so it is the purchase requirement that makes it a leak), the empty-`access_rules` case, and the fully-migrated paid gate passing.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in 01fefc4. Any unresolvable slug now warns, not just the all-or-nothing case: "1 of its 2 content rules do not resolve to a post type or taxonomy the evaluator can match (post), so the content they select stays ungated". The all-unresolvable message is unchanged, so the existing assertion still pins that wording, and a new test covers the mixed `post` + `category` shape. The dead slugs are named in both, which is the per-gate NPPD-2063 signal you were after.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in 01fefc4 — took both options, since they cover different failure modes.

`apply_layout()` no longer overwrites an existing layout when `'' === $content`: the update path now agrees with `create_gate_layout()` and keeps what is there (for a fresh gate, the seeded default), with a warning naming the layout post to review before cutover. And `verify_migrated_gate()` now checks each active mode's layout post for real — flagging one that is blank ("points at an empty layout (post %d), so the reader would see a blank gate") or that no longer exists — which catches a pre-existing empty layout on a re-run into an operator-edited gate, where skipping the write alone would not.

Good catch that the two paths disagreed on the same input; the seeded-default behaviour is now the same whichever one runs.

plugins/newspack-plugin/includes/cli/class-initializer.php
adekbadek: Bounded in 01fefc4. When the drop-in class is present, registration now adds a `before_invoke:newspack migrate-membership-gates` hook instead of warning inline, so the notice fires only when that command is actually dispatched. It keeps the fail-safe property you asked for — the hook is keyed on the command name, not on which registration won, so it fires even if the ordering flips and the drop-in's write-by-default callback is the one that runs.

Verified in a container: a `before_invoke` hook registered on `init` fires for its command and stays silent on `wp cache flush`.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
adekbadek: Removed in 01fefc4 — you are right that it was test isolation dressed up as an invariant, so it does not belong on the production surface.

The reflection reset already existed twice (`content-gates.php` and `feed-restriction.php` each had their own copy of the same four-property loop), so rather than add a third I pulled it into `tests/unit-tests/content-gate/traits/trait-restriction-cache-test.php` and used it from all three classes. The trait's docblock carries the "why" — the caches assume a request-stable gate configuration, which is true in production and false across rolled-back test cases that reuse post IDs.

adekbadek:
@miguelpeixe Re-requesting review – the changes you asked for have been pushed and need a look before this merges:
- fix(cli): flag under-gating migrated gates and preserve seeded layouts (NPPD-2059, #635)
- chore(cli): annotate unbounded plan query for vipwpcs NoPaging sniff (NPPD-2059)

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,912 @@
...
+		$progress = \WP_CLI\Utils\make_progress_bar( 'Migrating gates', $group_count );
+
+		foreach ( $plan_groups as $group ) {
+			$progress->tick();
+
+			$layout_errors = [];
+			$ac_rules      = $group[0]['ac_rules'];
+			$gate_title    = implode( ' | ', array_column( $group, 'name' ) );
+			$gate_key      = trim( strtolower( $gate_title ) );
+			$has_purchase  = ! empty(
```
miguelpeixe: **Suggestion:** `$has_purchase` is *any*, but the two gate modes compose with AND — a mixed group locks out the signup plan's members.

For a logged-in reader the evaluator requires **both** active modes to pass, not either. `tests/unit-tests/content-gate/content-gates.php:2193` proves it directly: with registration and custom access both active, a logged-in, email-verified subscriber is still restricted because the `custom_access` rule fails.

So a group holding one `signup` plan and one `purchase` plan produces a gate that demands the subscription from everyone. WooCommerce Memberships grants access to a holder of *either* plan, so the signup plan's members lose access at cutover — silently, since nothing in the summary distinguishes this gate from a correctly migrated paid one.

The new verification inverts for the same group: when there is no `member-content` wrapper, it reports "any registered reader would get in", which is exactly what the signup plan was for. So the WARN pushes the operator toward adding a paywall that should not be there.

The trigger is narrow — the plans have to share an identical rule fingerprint to group at all — so this is a suggestion rather than a blocker. But the migration already models a signup plan as "registration active", and the consistent treatment for a mixed group is that same mapping: treat the group as purchase-gated only when *every* plan requires a purchase (`! array_filter( $group, fn( $g ) => 'purchase' !== $g['access_method'] )`), and use that flag both for the layout branch and for the verification argument. If you would rather this rides with NPPD-2064, which already owns grouping, that seems reasonable — it just needs to be written down somewhere, because as it stands the group is silently lossy.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -0,0 +1,912 @@
...
+			}
+
+			// Post-write verification. The writes above only report that nothing
+			// errored; they say nothing about whether the resulting gate will
+			// actually restrict once WooCommerce Memberships is deactivated. Because
+			// migrated gates lie dormant until that cutover, an unenforceable gate
+			// would otherwise be reported as a success today and go unnoticed for
+			// weeks.
+			$verification_issues = [];
+			if ( ! $dry_run && empty( $layout_errors ) && $gate_id ) {
```
miguelpeixe, line 265: **Suggestion:** The dry run reports clean for exactly the cases `--live` now warns about.

Gating verification on `! $dry_run` is right for the checks that re-read a written gate — status, layout posts, resolved rules. But all three of the new signals are computable before any write: `$has_purchase && null === $layouts['custom_access']`, `$has_purchase && empty( $merged_product_ids )`, and the unresolvable slugs in `$ac_rules`.

Dry run is the default mode and the one an operator uses to plan a cutover, and today it prints `created (dry-run)` with Access Type `purchase` for a plan that will migrate into a gate any registered reader passes. The warnings do still appear on `--live`, so nothing is ultimately missed — this is about the planning pass being the one place the problem is invisible, which is a shame given the checks now exist.

A pre-write subset — the two purchase cases plus the slug check, run off the group data rather than off a written gate — would make the dry-run summary predictive rather than optimistic.

miguelpeixe (commented):
Re-reviewed at `82d73d35`. All five items from the last round landed and are correct — the purchase blocker is resolved, so **nothing here blocks merge**. Two things I checked specifically because they could have made a fix silently inert:

- `get_custom_access_settings()` runs `access_rules` through `Access_Rules::normalize_rules()`, which returns `[]` for empty input — so the new `empty( $custom_access['access_rules'] )` check does fire rather than being defeated by a normalised `[[]]`.
- `apply_layout()`'s empty-content branch keeps `$layout_id` set, so the mode is still activated against the layout it left in place; the create and update paths now agree on empty content, as intended.

The trait replaced three copies of the reflection reset, not the two I'd found, and it sits in the directory the bootstrap already globs — good.

Two new observations below, both narrower than the last round. Neither is a blocker; I am leaving this as a comment rather than an approval only because the first one is about grouping semantics you may want to reason about (or hand to NPPD-2064) before this lands.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Done — added `compute_pre_write_issues()` (a private static method) that computes the three signals you called out from the in-memory group data:

- Unresolvable content-rule slugs (all or partial)
- Purchase plan with no extracted `custom_access` layout → paid access mode will not activate
- Purchase plan with empty `merged_product_ids` → paid access mode will have no purchase constraint

The verification block now has an `elseif ( $dry_run )` branch that calls it and emits `WP_CLI::warning()` lines in the same shape as the live path, so the summary table gets `WARN:` rows for those cases in dry-run too. The live post-write path is unchanged.

Six new unit tests cover the method directly (all-unresolvable slugs, partial slugs, purchase with null layout, purchase with empty product IDs, clean signup plan, clean purchase plan). 25/25 green.

77deb09

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in a94353220 — treating the group as purchase-gated only when every plan requires a purchase, exactly as you suggested. Extracted the predicate to `group_requires_purchase()` so it's the single source of truth for both the custom_access layout branch and the verification argument:

```php
private static function group_requires_purchase( array $group ): bool {
    return ! array_filter( $group, fn( $g ) => 'purchase' !== $g['access_method'] );
}
```

Confirmed the AND composition you cited against the evaluator: for a logged-in reader with both modes active and `require_verification` unset (which this migration never writes), `is_post_restricted()` passes them on registration, then restricts them on custom_access unless they hold a subscription — so a mixed group did demand the subscription from everyone, dropping the signup plan's free members at cutover. A mixed group now migrates registration-gated (custom_access stays inactive), which is the most-permissive-plan mapping WCM's OR semantics call for.

Verified end-to-end on a WCM fixture (one signup + one purchase plan sharing a category rule → one group): the gate writes `registration` active + `custom_access` inactive, Access Type `signup`, and a fresh-read `verify_migrated_gate()` returns no issues. Red→green unit test added (`test_group_requires_purchase_only_when_every_plan_is_purchase`) — it fails against the old `any` logic on exactly the mixed case. Full class: `n test-php --filter Test_Membership_Gates_Migration` → OK (26 tests).

Kept this in #635 rather than deferring to NPPD-2064: the fix is the verification's own `has_purchase` argument, so it belongs with the code that introduced it, and it's small and self-contained. NPPD-2064's consolidation (PR #642) reworks which plans group together — a layer above this mode-selection — so the two compose rather than collide.

miguelpeixe (approved):

PR #636, opened 2026-07-16
feat(network): product network ID assignment + verification CLI (NPPD-2057)

adekbadek:
### All Submissions:

Adds migration tooling for the Access Control rollout on a Newspack Network (hub + nodes).

**Why.** Access Control's cross-site paid access only grants something when a gate's products carry a product Network ID (the `_newspack_network_product_id` postmeta that `Content_Gate\Access` reads). The only writer today is a manual per-product metabox, so a network migrated from Woo Memberships routinely ends up with healthy membership/subscription sync but **zero tagged products** — and flipping a node in that state silently strips its network-synced members, because nothing links its gate products to the matching products on other sites. This PR makes tagging assignable in bulk and verifiable before a flip.

**What.** Two WP-CLI commands in the `newspack-network` namespace:

- `wp newspack-network assign-product-network-ids` — assigns product Network IDs across the network's products. By default it **derives** them from the shared plan-level `_newspack_network_id` (writing each membership plan's Network ID onto every product the plan links, read from the plan's `_product_ids`); alternatively `--map` takes an explicit operator mapping (inline JSON `{"<product_id>":"<network_id>"}` or a path to a JSON file). Dry-run by default; `--apply` writes. When a product is claimed by two plans with **different** Network IDs the assignment is **withheld** and reported as a conflict rather than guessed. Existing differing values are protected unless `--overwrite`, and a Network ID that sanitizes to empty is never written whatever the flags — with `--overwrite` a whitespace-only plan value would otherwise blank a correct ID and propagate the blank network-wide. Everything the command **could not** resolve (an untagged plan, a conflict, a skipped product, an unusable `--map` entry) rolls into a **non-zero exit**, so a scripted migration cannot record a green step for a partial run. On write it fires the existing `newspack_network_save_product` action — the same one the product metabox fires — so the shipped emitter/propagation path runs; `--repropagate` re-emits for products that already carry the target ID (which otherwise fire no event), which is the way to re-sync a tagged-but-desynced product.
- `wp newspack-network verify-product-network-ids` — a per-site pre-flip gate. It reports, per product, whether it carries a Network ID and which **other sites** in the synced products map share that ID (without which the cross-site grant resolves to nothing). The default product set is every product a membership plan links **plus** every already-tagged product, so the products `assign` withheld are checked rather than defined away; `--products` narrows it to a gate's products. `--expect-sites=<count|all>` raises the bar from "linked somewhere" to real coverage (`all` = every Node registered on this Hub), since a reader is only granted from the site their own subscription lives on. `--format=json` emits one machine-readable report for aggregating the per-site gate across a network. It **exits non-zero** on any issue so a flip can be gated on it.

**Hub self-inclusion subtlety.** `verify` excludes the current site from the linkage count. This is load-bearing on the **Hub**: a Hub self-includes its own products under its own URL key in the synced map (via `always_process_in_hub`), so without the exclusion a Hub-only Network ID would count as "linked to itself" and produce a false "ready to flip". On a **Node** there is no self-entry (a node never receives its own events back — the Hub excludes the requesting node when serving pulls), so the exclusion is simply a no-op there. Checking only cross-site linkage is what makes the command correct on both roles.

**Documented limitations** (in the command docblocks):

- A default `verify` run only proves each product is linked to *some* other visible site. Full coverage needs `--expect-sites=all`, which is only available on a Hub (a Node cannot read the network's membership — pass a count there).
- The synced products option is **append-only** (there is no product-deleted listener), so a stale entry for a since-removed product or site can still report as "linked".
- Linkage matching relies on the site's URL staying stable as the map key (compared case- and trailing-slash-insensitively, but still the usual URL-key caveat).

This is a **hard prerequisite step for Access Control migration on networked sites**; runbook integration follows separately.

Closes NPPD-2057.

1. On a Newspack Network (hub + nodes) with membership plans carrying `_newspack_network_id` and linked products, run `wp newspack-network assign-product-network-ids` (dry-run) and confirm it reports the plan-derived product => Network ID assignments, withholds any conflicting product, and protects already-tagged products; then `--apply` and confirm the meta is written and the save-product event fires.
2. Run `wp newspack-network verify-product-network-ids` and confirm a plan-linked but untagged product is flagged and the command exits non-zero; then `--products=<gate product IDs>` for the gate-specific check, and `--expect-sites=all` on the Hub to confirm a product missing from one Node fails even though it is linked on another.
3. Unit suite: from `plugins/newspack-network`, `n test-php --filter TestProductNetworkIds` — green. PHPCS clean on the changed files.

E2E-verified locally in an isolated env: dry-run/apply/`--overwrite`/`--map` paths; conflict withholding; the save-product propagation action firing on `--apply`; verify exit codes (0 green / non-zero on untagged or unlinked); and specifically that a product self-included in the map under the current site's own URL (with no other site sharing the ID) is correctly flagged **unlinked**, proving the Hub self-exclusion is load-bearing.

### Other information:

### DIFF
--- a/plugins/newspack-network/includes/class-initializer.php
+++ b/plugins/newspack-network/includes/class-initializer.php
@@ -58,6 +58,7 @@ class Initializer {
 		Data_Backfill::init();
 		Membership_Dedupe::init();
 		CLI\Integrity_Check::init();
+		CLI\Product_Network_Ids::init();
 
 		Woocommerce\Events::init();
 		Woocommerce\Product_Admin::init();
--- a/plugins/newspack-network/includes/cli/class-product-network-ids.php
+++ b/plugins/newspack-network/includes/cli/class-product-network-ids.php
@@ -719,8 +719,15 @@ class Product_Network_Ids {
 			WP_CLI::error( 'Verification found issues ( see above ). Not ready to flip.' );
 		}
 
+		if ( $is_json ) {
+			// JSON mode prints exactly one line, the report above, on both the passing and the failing
+			// path ( WP_CLI::error writes to stderr ), so an aggregator can json_decode stdout as-is.
+			// The verdict is carried by the exit code and by the report's own 'ready_to_flip'.
+			return;
+		}
+
 		WP_CLI::success( sprintf( 'All checked products carry a Network ID and are linked on %s.', $expectation['label'] ) );
-		if ( ! $is_json && ! $expectation['require_all'] && 1 === $expectation['minimum'] ) {
+		if ( ! $expectation['require_all'] && 1 === $expectation['minimum'] ) {
 			// Being linked *somewhere* is not the guarantee operators read into a green run: access is only
 			// granted from the site a reader's own subscription lives on.
 			WP_CLI::line( 'Note: this only proves each product is linked somewhere, not that every site a reader may subscribe on carries the ID. Re-run with --expect-sites=all ( on the Hub ) or --expect-sites=<count> to gate on real coverage.' );
@@ -745,6 +752,13 @@ class Product_Network_Ids {
 				'minimum'     => 1,
 				'label'       => 'at least 1 other site',
 			];
+		}
+
+		// A bare --expect-sites ( or --no-expect-sites ) arrives from WP-CLI as a boolean. (string) true
+		// is '1', which would pass the integer guard below and silently resolve to the 1-site floor: a
+		// green run for an operator who meant --expect-sites=all and fumbled the syntax.
+		if ( is_bool( $raw ) ) {
+			WP_CLI::error( '--expect-sites needs a value: pass "all" or a positive integer ( e.g. --expect-sites=all or --expect-sites=3 ). Omit the flag entirely for the default of 1 other site.' );
 		}
 
 		if ( 'all' === $raw ) {
@@ -966,9 +980,16 @@ class Product_Network_Ids {
 	 */
 	private static function get_products_to_check( $product_ids = null ) {
 		if ( null !== $product_ids ) {
+			$product_ids = array_map( 'intval', $product_ids );
+
+			// Prime up front, as on the default set below: get_network_id() reads the post and its meta,
+			// which would otherwise be a DB round trip per product on exactly the input a scripted flip
+			// gate passes.
+			self::prime_post_caches( $product_ids );
+
 			$products = [];
 			foreach ( $product_ids as $product_id ) {
-				$products[ (int) $product_id ] = Product_Admin::get_network_id( $product_id );
+				$products[ $product_id ] = Product_Admin::get_network_id( $product_id );
 			}
 			return $products;
 		}
--- a/plugins/newspack-network/includes/cli/class-product-network-ids.php
+++ b/plugins/newspack-network/includes/cli/class-product-network-ids.php
@@ -236,6 +236,9 @@ class Product_Network_Ids {
 		$to_write = 0;
 		foreach ( $assignments as $product_id => $network_id ) {
 			$product_id = (int) $product_id;
+			// Sanitize at the write path so the meta is consistent whatever the source ( plan meta or --map ),
+			// matching the product metabox's sanitize_text_field(). Done before the preview so dry-run matches apply.
+			$network_id = sanitize_text_field( (string) $network_id );
 
 			$post_type = get_post_type( $product_id );
 			if ( 'product' !== $post_type ) {
@@ -473,7 +476,11 @@ class Product_Network_Ids {
 
 		$assignments = [];
 		foreach ( $decoded as $product_id => $network_id ) {
-			$network_id = trim( (string) $network_id );
+			if ( ! is_scalar( $network_id ) ) {
+				WP_CLI::warning( sprintf( 'Skipping product #%d in --map: Network ID must be a string.', (int) $product_id ) );
+				continue;
+			}
+			$network_id = sanitize_text_field( (string) $network_id );
 			if ( '' === $network_id ) {
 				WP_CLI::warning( sprintf( 'Skipping product #%d in --map: empty Network ID.', (int) $product_id ) );
 				continue;
--- a/plugins/newspack-network/includes/cli/class-product-network-ids.php
+++ b/plugins/newspack-network/includes/cli/class-product-network-ids.php
@@ -34,11 +34,41 @@ class Product_Network_Ids {
 	const PLAN_PRODUCT_IDS_META_KEY = '_product_ids';
 
 	/**
-	 * Initialize this class and register the WP-CLI commands.
+	 * How many product writes to make before flushing the Data Events dispatch queue.
+	 *
+	 * Data_Events::dispatch() only queues; the queue is drained on shutdown. Flushing periodically
+	 * keeps the queued payloads from accumulating in memory for the whole run and means a run that
+	 * dies mid-way has already propagated everything up to the last flush.
+	 *
+	 * @var int
+	 */
+	const DISPATCH_FLUSH_INTERVAL = 100;
+
+	/**
+	 * How many posts to prime the meta cache for at a time.
+	 *
+	 * Priming loads every meta row of every listed post, so it is chunked to bound peak memory on
+	 * stores with many products.
+	 *
+	 * @var int
+	 */
+	const META_CACHE_CHUNK_SIZE = 500;
+
+	/**
+	 * Initialize this class and register hooks.
 	 *
 	 * @return void
 	 */
 	public static function init() {
+		add_action( 'init', [ __CLASS__, 'register_commands' ] );
+	}
+
+	/**
+	 * Register the WP-CLI commands.
+	 *
+	 * @return void
+	 */
+	public static function register_commands() {
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			// These are migration tooling, so they take the migration flow's --apply flag ( dry-run by default )
 			// rather than this plugin's older data-* commands' --live flag.
@@ -56,17 +86,31 @@ class Product_Network_Ids {
 	 * two plans with different Network IDs is ambiguous: it is withheld from the assignments and
 	 * reported as a conflict rather than guessed.
 	 *
+	 * Plan Network IDs are sanitized here, before the emptiness check, so a whitespace-only ( or
+	 * markup-only ) plan value is treated as "no Network ID" rather than surviving as an assignment
+	 * that later sanitizes down to '' and blanks a product.
+	 *
 	 * @param array $plans Array of [ 'network_id' => string, 'product_ids' => int[] ].
 	 * @return array {
 	 *     @type array $assignments Map of product ID => Network ID.
 	 *     @type array $conflicts   Map of product ID => list of the distinct Network IDs claiming it.
+	 *     @type array $stats       Per-plan diagnostics: total, without_network_id, without_products.
 	 * }
 	 */
 	public static function derive_assignments_from_plans( array $plans ) {
-		$claims = []; // Product ID => list of distinct Network IDs claiming it.
+		$claims             = []; // Product ID => list of distinct Network IDs claiming it.
+		$without_network_id = 0;
+		$without_products   = 0;
+
 		foreach ( $plans as $plan ) {
-			$network_id  = (string) ( $plan['network_id'] ?? '' );
+			$network_id  = sanitize_text_field( (string) ( $plan['network_id'] ?? '' ) );
 			$product_ids = $plan['product_ids'] ?? [];
+			if ( '' === $network_id ) {
+				$without_network_id++;
+			}
+			if ( empty( $product_ids ) ) {
+				$without_products++;
+			}
 			if ( '' === $network_id || empty( $product_ids ) ) {
 				continue;
 			}
@@ -94,6 +138,11 @@ class Product_Network_Ids {
 		return [
 			'assignments' => $assignments,
 			'conflicts'   => $conflicts,
+			'stats'       => [
+				'total'              => count( $plans ),
+				'without_network_id' => $without_network_id,
+				'without_products'   => $without_products,
+			],
 		];
 	}
 
@@ -170,6 +219,10 @@ class Product_Network_Ids {
 	 * written to every product the plan links ). Pass --map to assign an explicit operator mapping
 	 * instead. Runs in dry-run mode unless --apply is given.
 	 *
+	 * Exits non-zero when anything could not be assigned ( a plan carrying no Network ID, a product
+	 * claimed by two plans with different IDs, a product already carrying a different ID without
+	 * --overwrite ), so a scripted migration cannot record a green step for a partial run.
+	 *
 	 * ## OPTIONS
 	 *
 	 * [--map=<map>]
@@ -180,6 +233,10 @@ class Product_Network_Ids {
 	 * : Overwrite a product's existing Network ID when it differs from the derived/mapped value.
 	 * By default, existing differing values are left untouched and reported.
 	 *
+	 * [--repropagate]
+	 * : Also emit a product_updated event for products that already carry the target Network ID.
+	 * Use this to re-sync an already-tagged but desynced product, which otherwise fires no event.
+	 *
 	 * [--apply]
 	 * : Write the changes. Without this flag the command only reports what it would do.
 	 *
@@ -188,6 +245,7 @@ class Product_Network_Ids {
 	 *     wp newspack-network assign-product-network-ids
 	 *     wp newspack-network assign-product-network-ids --apply
 	 *     wp newspack-network assign-product-network-ids --map='{"123":"premium","456":"premium"}' --apply
+	 *     wp newspack-network assign-product-network-ids --apply --repropagate
 	 *
 	 * @param array $args       Positional arguments ( unused ).
 	 * @param array $assoc_args Associative arguments.
@@ -196,8 +254,10 @@ class Product_Network_Ids {
 	public static function assign( $args, $assoc_args ) {
 		self::require_network_site();
 
-		$apply     = isset( $assoc_args['apply'] );
-		$overwrite = isset( $assoc_args['overwrite'] );
+		$apply       = (bool) \WP_CLI\Utils\get_flag_value( $assoc_args, 'apply', false );
+		$overwrite   = (bool) \WP_CLI\Utils\get_flag_value( $assoc_args, 'overwrite', false );
+		$repropagate = (bool) \WP_CLI\Utils\get_flag_value( $assoc_args, 'repropagate', false );
+		$map         = \WP_CLI\Utils\get_flag_value( $assoc_args, 'map', null );
 
 		// Writing the meta only propagates across the network through the newspack_network_product_updated
 		// listener, which registers only when Newspack\Data_Events is loaded and emits only when wc_get_product
@@ -215,18 +275,48 @@ class Product_Network_Ids {
 
 		if ( ! $can_propagate ) {
 			WP_CLI::warning(
-				'Cross-site propagation is unavailable here ( newspack_network_product_updated needs Newspack\\Data_Events and WooCommerce active ). Network IDs will be written to this site only; propagate them afterwards with "wp newspack-network data-backfill newspack_network_product_updated --live".'
+				'Cross-site propagation is unavailable here ( newspack_network_product_updated needs Newspack\\Data_Events and WooCommerce active ). Network IDs will be written to this site only; once both are active, re-run this command with --apply --repropagate to emit the events.'
 			);
 			WP_CLI::line( '' );
 		}
 
-		if ( isset( $assoc_args['map'] ) ) {
-			$assignments = self::parse_map( $assoc_args['map'] );
+		// Everything this command cannot resolve. Reported per item below and rolled into a non-zero
+		// exit at the end so "assign, then verify" cannot record a green step on a partial run.
+		$unresolved  = 0;
+		$plans_found = 0;
+
+		if ( null !== $map ) {
+			$assignments = self::parse_map( $map );
 			WP_CLI::line( sprintf( 'Using explicit operator mapping ( %d product(s) ).', count( $assignments ) ) );
 		} else {
-			$derived     = self::derive_assignments_from_plans( self::get_plans() );
+			$plans       = self::get_plans();
+			$plans_found = count( $plans );
+			$derived     = self::derive_assignments_from_plans( $plans );
 			$assignments = $derived['assignments'];
+			$stats       = $derived['stats'];
+
+			WP_CLI::line(
+				sprintf(
+					'Found %d membership plan(s): %d carry no Network ID, %d link no products.',
+					$stats['total'],
+					$stats['without_network_id'],
+					$stats['without_products']
+				)
+			);
 			WP_CLI::line( sprintf( 'Derived %d assignment(s) from membership plans.', count( $assignments ) ) );
+
+			// A plan nobody ever tagged is the NPPD-2057 field condition, and this is the only place in
+			// the system positioned to see it -- so it is reported, not silently skipped.
+			if ( $stats['without_network_id'] > 0 ) {
+				WP_CLI::warning(
+					sprintf(
+						'%d membership plan(s) carry no Network ID, so their products cannot be derived. Set the plan Network ID on every site ( the same value on matching plans ), or assign the products via --map.',
+						$stats['without_network_id']
+					)
+				);
+				$unresolved += $stats['without_network_id'];
+			}
+
 			foreach ( $derived['conflicts'] as $product_id => $network_ids ) {
 				WP_CLI::warning(
 					sprintf(
@@ -235,37 +325,82 @@ class Product_Network_Ids {
 						implode( ', ', $network_ids )
 					)
 				);
+				$unresolved++;
 			}
 		}
 		WP_CLI::line( '' );
 
+		// One query for the whole working set: the loop below reads the post type and the Network ID
+		// meta of every assignment, which would otherwise be two DB round trips per product.
+		self::prime_post_caches( array_keys( $assignments ) );
+
+		// A plan may link variation IDs; the Network ID lives on the parent product ( that is where
+		// Product_Admin::get_network_id resolves it from ), so fold them in rather than skipping them.
+		$folded      = self::fold_variations_into_parents( $assignments );
+		$assignments = $folded['assignments'];
+		foreach ( $folded['conflicts'] as $product_id => $network_ids ) {
+			WP_CLI::warning(
+				sprintf(
+					'Product #%d is claimed with different Network IDs ( %s ) once its variations are resolved to it - skipped. Resolve manually or via --map.',
+					$product_id,
+					implode( ', ', $network_ids )
+				)
+			);
+			$unresolved++;
+		}
+
 		if ( empty( $assignments ) ) {
 			WP_CLI::warning( 'No assignments to make.' );
+			if ( $plans_found > 0 || null !== $map || $unresolved > 0 ) {
+				// Plans ( or a map ) exist but nothing could be derived from them: a no-op here reads as
+				// "already migrated" if it exits 0, so fail instead.
+				WP_CLI::error( 'Nothing could be assigned. Cross-site paid access will grant nothing here.' );
+			}
 			return;
 		}
 
-		$skipped  = 0;
-		$already  = 0;
-		$to_write = 0;
+		$skipped      = 0;
+		$already      = 0;
+		$to_write     = 0;
+		$inapplicable = 0;
+		$dispatched   = 0;
 		foreach ( $assignments as $product_id => $network_id ) {
 			$product_id = (int) $product_id;
 			// Sanitize at the write path so the meta is consistent whatever the source ( plan meta or --map ),
 			// matching the product metabox's sanitize_text_field(). Done before the preview so dry-run matches apply.
 			$network_id = sanitize_text_field( (string) $network_id );
 
-			$post_type = get_post_type( $product_id );
-			if ( 'product' !== $post_type ) {
-				// The Network ID lives on the parent product; variations resolve to it via Product_Admin::get_network_id.
-				$reason = 'product_variation' === $post_type ? 'a product variation ( set the Network ID on its parent product )' : 'not a product';
-				WP_CLI::warning( sprintf( 'Skipping #%d: %s.', $product_id, $reason ) );
+			// Never write a blank Network ID: it is indistinguishable from "untagged", and with --overwrite
+			// it would replace a correct value and propagate the blank to every site.
+			if ( '' === $network_id ) {
+				WP_CLI::warning( sprintf( 'Skipping #%d: the Network ID is empty - refusing to write a blank value.', $product_id ) );
 				$skipped++;
 				continue;
 			}
 
+			if ( 'product' !== get_post_type( $product_id ) ) {
+				WP_CLI::warning( sprintf( 'Skipping #%d: not a product.', $product_id ) );
+				$skipped++;
+				continue;
+			}
+
+			// Access only ever grants from a reader's synced subscriptions, and the product metabox only
+			// writes for these types, so tagging anything else would just bloat every site's synced map.
+			if ( ! self::is_taggable_product( $product_id ) ) {
+				WP_CLI::line( sprintf( '  #%d is not a subscription product; a Network ID would grant nothing - left alone.', $product_id ) );
+				$inapplicable++;
+				continue;
+			}
+
 			$current = (string) get_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, true );
-			if ( $current === (string) $network_id ) {
+			if ( $current === $network_id ) {
 				WP_CLI::line( sprintf( '  #%d already set to "%s".', $product_id, $network_id ) );
 				$already++;
+				if ( $apply && $repropagate ) {
+					do_action( 'newspack_network_save_product', $product_id );
+					$dispatched++;
+					self::maybe_flush_dispatch_queue( $dispatched );
+				}
 				continue;
 			}
 			if ( '' !== $current && ! $overwrite ) {
@@ -283,22 +418,63 @@ class Product_Network_Ids {
 				update_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, $network_id );
 				// Fire the same action the product metabox fires, so the existing emitter propagates the change.
 				do_action( 'newspack_network_save_product', $product_id );
+				$dispatched++;
+				self::maybe_flush_dispatch_queue( $dispatched );
 			}
+		}
+		$unresolved += $skipped;
+
+		if ( $dispatched > 0 ) {
+			self::flush_dispatch_queue();
 		}
 
 		WP_CLI::line( '' );
 		if ( $apply ) {
-			WP_CLI::success( sprintf( 'Wrote %d product Network ID(s), skipped %d, %d already set.', $to_write, $skipped, $already ) );
+			WP_CLI::line(
+				sprintf(
+					'Wrote %d product Network ID(s), skipped %d, %d already set, %d not applicable.',
+					$to_write,
+					$skipped,
+					$already,
+					$inapplicable
+				)
+			);
 			if ( ! $can_propagate ) {
-				WP_CLI::warning( 'These writes were NOT propagated ( see the warning above ). Once Data Events and WooCommerce are active, replay propagation with:' );
-			} else {
-				// A product already carrying the target Network ID fires no event, so this command cannot
-				// re-propagate an already-tagged-but-desynced product. Point operators at the backfill for that.
-				WP_CLI::line( 'Products already carrying the target Network ID fire no event; if propagation may have been missed ( including for already-set products ), replay it with:' );
+				WP_CLI::warning( 'These writes were NOT propagated ( see the warning above ). Once Data Events and WooCommerce are active, re-run with --apply --repropagate to emit the events.' );
+			} elseif ( ! $repropagate ) {
+				// A product already carrying the target Network ID fires no event, so a plain run cannot
+				// re-sync an already-tagged-but-desynced product.
+				WP_CLI::line( 'Products already carrying the target Network ID fired no event. To re-emit for every assigned product, re-run with --apply --repropagate.' );
+				WP_CLI::line( 'Note: "data-backfill newspack_network_product_updated --live" is not a reliable replay for this - on a Hub it dedupes against the event log on the product\'s post_modified_gmt timestamp, which writing this meta does not change.' );
 			}
-			WP_CLI::line( '  wp newspack-network data-backfill newspack_network_product_updated --live' );
 		} else {
-			WP_CLI::success( sprintf( 'Dry run: %d product Network ID(s) would be written, %d skipped, %d already set. Re-run with --apply.', $to_write, $skipped, $already ) );
+			WP_CLI::line(
+				sprintf(
+					'Dry run: %d product Network ID(s) would be written, %d skipped, %d already set, %d not applicable.',
+					$to_write,
+					$skipped,
+					$already,
+					$inapplicable
+				)
+			);
+		}
+		WP_CLI::line( '' );
+
+		if ( $unresolved > 0 ) {
+			// Exit non-zero so a scripted migration cannot treat a partial assignment as done: everything
+			// withheld here is a product ( or plan ) that will grant nothing across the network.
+			WP_CLI::error(
+				sprintf(
+					'%d item(s) could not be assigned ( see the warnings above ). Resolve them ( e.g. via --map ) and re-run; verify-product-network-ids will fail until they are.',
+					$unresolved
+				)
+			);
+		}
+
+		if ( $apply ) {
+			WP_CLI::success( 'Every derived product Network ID is assigned.' );
+		} else {
+			WP_CLI::success( 'Dry run complete: every derived product Network ID can be assigned. Re-run with --apply.' );
 		}
 		WP_CLI::line( '' );
 	}
@@ -312,6 +488,11 @@ class Product_Network_Ids {
 	 * or not tagged at all, so the cross-site grant resolves to nothing ). Run it on every site: each
 	 * site's linkage check confirms the other sites' products are present in its synced map.
 	 *
+	 * The default product set is every product the site's membership plans link, plus every product that
+	 * already carries a Network ID -- not just the tagged ones. Checking only tagged products would hide
+	 * exactly what assign-product-network-ids withholds ( conflicts, skipped products ), so a network
+	 * that assign could not finish would still verify green.
+	 *
 	 * Limitation: the synced products option is append-only ( there is no product-deleted listener ), so a
 	 * stale entry for a since-removed product or site can still report as "linked" -- the same class of
 	 * caveat as relying on the site's URL staying byte-identical as the map key.
@@ -319,9 +500,9 @@ class Product_Network_Ids {
 	 * ## OPTIONS
 	 *
 	 * [--products=<ids>]
-	 * : Comma-separated product IDs to check ( e.g. a gate's products ). Defaults to every local product
-	 * that carries a Network ID. To gate a flip, pass the gate's product IDs: the bare form only looks at
-	 * already-tagged products, so a site with zero tagged products ( the NPPD-2057 failure ) passes it.
+	 * : Comma-separated product IDs to check ( e.g. a gate's products ). Defaults to every product linked
+	 * by a membership plan plus every product that already carries a Network ID. Pass a gate's product IDs
+	 * to check exactly what the gate depends on.
 	 *
 	 * ## EXAMPLES
 	 *
@@ -341,10 +522,16 @@ class Product_Network_Ids {
 			$network_products = [];
 		}
 
-		$explicit_products = isset( $assoc_args['products'] );
-		$local_products    = self::get_products_to_check(
-			$explicit_products ? wp_parse_id_list( $assoc_args['products'] ) : null
-		);
+		$product_ids = null;
+		if ( isset( $assoc_args['products'] ) ) {
+			$product_ids = wp_parse_id_list( $assoc_args['products'] );
+			if ( empty( $product_ids ) ) {
+				// An unset shell variable ( --products="$GATE_PRODUCTS" ) must never read as "nothing to
+				// check, all good": this is the path a scripted flip gate runs.
+				WP_CLI::error( '--products was passed but no product IDs could be parsed from it. Pass a comma-separated list of product IDs, or omit --products to check every plan-linked and tagged product.' );
+			}
+		}
+		$local_products = self::get_products_to_check( $product_ids );
 
 		WP_CLI::line( '' );
 		WP_CLI::line( sprintf( 'Verifying product Network IDs for %s.', $current_site ) );
@@ -353,15 +540,13 @@ class Product_Network_Ids {
 		WP_CLI::line( '' );
 
 		if ( empty( $local_products ) ) {
-			WP_CLI::warning( 'No products carry a Network ID on this site. Cross-site paid access will grant nothing here. Run assign-product-network-ids first.' );
-			return;
+			WP_CLI::error( 'No products to check: no membership plan links a subscription product and no product carries a Network ID. Cross-site paid access will grant nothing here. Run assign-product-network-ids first, or pass --products with the gate\'s products.' );
 		}
 
 		$findings = self::verify_products( $local_products, $network_products, $current_site );
 		$untagged = 0;
 		$unlinked = 0;
 		foreach ( $findings as $product_id => $finding ) {
-			// An untagged product ( only reachable when passed explicitly via --products ) is the NPPD-2057 failure itself.
 			if ( '' === $finding['network_id'] ) {
 				WP_CLI::warning( sprintf( '#%d: no Network ID set ( cross-site access grants nothing ). Run assign-product-network-ids.', $product_id ) );
 				$untagged++;
@@ -383,8 +568,8 @@ class Product_Network_Ids {
 		WP_CLI::line( '' );
 		WP_CLI::line( sprintf( 'Checked %d product(s): %d untagged, %d unlinked.', count( $findings ), $untagged, $unlinked ) );
 		if ( $unlinked > 0 ) {
-			WP_CLI::line( 'For unlinked products, make sure every site has run assign-product-network-ids and its product_updated events have propagated:' );
-			WP_CLI::line( '  wp newspack-network data-backfill newspack_network_product_updated --live' );
+			WP_CLI::line( 'For unlinked products, make sure every site has run assign-product-network-ids, then re-emit the events from those sites with:' );
+			WP_CLI::line( '  wp newspack-network assign-product-network-ids --apply --repropagate' );
 		}
 		if ( $untagged > 0 || $unlinked > 0 ) {
 			// Exit non-zero so callers can gate a flip on this check.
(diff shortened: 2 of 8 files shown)


plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,485 @@
...
+			return;
+		}
+
+		$skipped  = 0;
+		$already  = 0;
+		$to_write = 0;
+		foreach ( $assignments as $product_id => $network_id ) {
+			$product_id = (int) $product_id;
+
+			$post_type = get_post_type( $product_id );
```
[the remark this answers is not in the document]
adekbadek: Good call — applied in de888d6e0. The Network ID is now run through `sanitize_text_field()` at the `assign()` write path (before the dry-run preview, so preview matches apply), keeping the meta consistent whatever the source.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,485 @@
...
+		$assignments = [];
+		foreach ( $decoded as $product_id => $network_id ) {
+			$network_id = trim( (string) $network_id );
+			if ( '' === $network_id ) {
+				WP_CLI::warning( sprintf( 'Skipping product #%d in --map: empty Network ID.', (int) $product_id ) );
+				continue;
+			}
+			$assignments[ (int) $product_id ] = $network_id;
+		}
+		return $assignments;
```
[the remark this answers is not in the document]
adekbadek: Applied in de888d6e0. `parse_map()` now skips non-scalar values with a warning (no more "Array to string conversion") and runs each value through `sanitize_text_field()`.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,485 @@
...
+		WP_CLI::line( '' );
+		WP_CLI::line( sprintf( 'Verifying product Network IDs for %s.', $current_site ) );
+		WP_CLI::line( 'Checked: whether this site\'s products carry a Network ID, and whether other sites in the synced map share it.' );
+		WP_CLI::line( 'Not checked from here: other sites\' databases ( only their synced map entries are visible ). Run this on every site.' );
+		WP_CLI::line( '' );
+
+		if ( empty( $local_products ) ) {
+			WP_CLI::warning( 'No products carry a Network ID on this site. Cross-site paid access will grant nothing here. Run assign-product-network-ids first.' );
+			return;
+		}
```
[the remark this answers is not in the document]
adekbadek: This is deliberate and documented. A regwall-only site legitimately has zero tagged products, and the bare command should not report a failure there — so it warns and exits 0. Flip-gating for paid sites is done by passing the gate's product IDs (`--products=<gate IDs>`), where an untagged product is flagged and the command exits non-zero; the `--products` docblock spells this out. Making the bare form exit non-zero would spuriously fail regwall-only sites.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,492 @@
...
+			if ( $apply ) {
+				update_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, $network_id );
+				// Fire the same action the product metabox fires, so the existing emitter propagates the change.
+				do_action( 'newspack_network_save_product', $product_id );
+			}
+		}
+
+		WP_CLI::line( '' );
+		if ( $apply ) {
+			WP_CLI::success( sprintf( 'Wrote %d product Network ID(s), skipped %d, %d already set.', $to_write, $skipped, $already ) );
```
chickenn00dle: **Severity:** suggestion

`--apply` prints `Wrote N product Network ID(s)` unconditionally, but propagation via `do_action( 'newspack_network_save_product' )` depends on the `Woocommerce\Events::product_updated` listener, which early-returns when `wc_get_product` is unavailable and isn't even registered without `Newspack\Data_Events`. In a migration state where WooCommerce or newspack-plugin isn't active, the meta is written locally with zero propagation, yet the output still reports success — and the `data-backfill` hint below is framed conditionally ("If propagation did not complete"), so an operator gets no clear signal it didn't. Consider a preflight check that warns loudly (or refuses `--apply`) when `wc_get_product` / `Data_Events` aren't present. Relatedly, a product whose meta already equals the target value is counted as "already set" and fires no event, so `assign` alone can't re-propagate a tagged-but-desynced product — a one-line note pointing operators at `data-backfill` would set expectations.

**Risk if not addressed:** if an operator reads "Wrote N" as done and flips a node before propagation actually ran, the flip strips network-synced members — the exact NPPD-2057 failure this tool exists to prevent. The `verify --products` gate does catch it (unlinked → non-zero), so this is a signal-clarity hardening, not a correctness bug.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,492 @@
...
+						'value'   => '',
+					],
+				],
+			]
+		);
+
+		// The meta_query already excludes empty/absent Network IDs, so every ID here is tagged.
+		$tagged = [];
+		foreach ( $tagged_ids as $product_id ) {
+			$tagged[ (int) $product_id ] = Product_Admin::get_network_id( $product_id );
```
chickenn00dle: **Severity:** suggestion

N+1 meta read in the bare (`--products`-less) path: `get_posts()` with `fields => 'ids'` doesn't prime the postmeta cache, so each `Product_Admin::get_network_id()` in the loop issues its own `get_post_meta()` DB hit — one query per tagged product. This is migration tooling and the set is bounded to already-tagged products, so impact is low, but selecting the meta in the query (or dropping `fields => 'ids'` so WP primes the meta cache) would collapse it to one round-trip.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	/**
+	 * Parse the --map value ( inline JSON or a path to a JSON file ) into a product ID => Network ID map.
+	 *
+	 * @param string $map The raw --map argument.
+	 * @return array
+	 */
+	private static function parse_map( $map ) {
+		if ( is_string( $map ) && is_readable( $map ) ) {
```
chickenn00dle: **Severity:** nit

`is_readable()` runs against the raw `--map` value, so for an inline-JSON argument it passes the whole JSON string to a filesystem call. It's harmless (returns false → falls through to `json_decode`), but a very long inline map can trip an `E_WARNING` ("File name too long") on some platforms. A cheap guard — e.g. only treat it as a path when it doesn't look like JSON, or `strlen($map) < PHP_MAXPATHLEN` — avoids it.

chickenn00dle (commented):
Nice, carefully-guarded migration command — approving with a few non-blocking comments.

Release risk: LOW. No blockers. It's dry-run by default, idempotent (a second run re-reads the meta and reports "already set", firing no event), withholds conflicting products rather than guessing, keeps `--overwrite` opt-in, and the verify gate exits non-zero so a flip can be gated on it. Queries use `fields => 'ids'`, so there's no full-catalog object load. The Hub self-exclusion in `verify_products()` is correct and unit-tested.

Three small things worth a look before merge (inline). None block.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Added a preflight: `assign` now computes `$can_propagate = class_exists( 'Newspack\\Data_Events' ) && function_exists( 'wc_get_product' )` and, when it's false, prints a loud upfront `WP_CLI::warning` that propagation is unavailable and the writes will stay local (with the data-backfill replay command). The apply summary is no longer unconditional: when propagation was unavailable it says the writes were NOT propagated; otherwise it notes that already-set products fire no event and points at data-backfill for a tagged-but-desynced product. Chose the loud-warning route over refusing `--apply` because writing locally then backfilling is a legitimate migration flow.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed the N+1: added `update_meta_cache( 'post', $tagged_ids )` right after the `fields => 'ids'` query, priming the whole postmeta cache in one round-trip so the per-product `get_network_id()` reads all hit cache. Kept `fields => 'ids'` (surgical, avoids hydrating full post objects) rather than dropping it.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Guarded it: `parse_map` now only calls `is_readable()` when `strlen( $map ) < PHP_MAXPATHLEN`, so a long inline-JSON map is never handed to the filesystem call and can't trip the "File name too long" E_WARNING. Added a test that a JSON map exceeding PHP_MAXPATHLEN still parses cleanly.

adekbadek:
- feat(network): harden product network ID CLI signal + query paths (NPPD-2057, #636)

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+				);
+				$skipped++;
+				continue;
+			}
+
+			WP_CLI::line( sprintf( '  #%d "%s" => "%s"', $product_id, $current, $network_id ) );
+			$to_write++;
+
+			if ( $apply ) {
+				update_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, $network_id );
```
chickenn00dle, line 483: **Blocker:** --overwrite can wipe a correct Network ID and propagate the blank network-wide — The emptiness guard at line 70 tests the **raw** plan meta, but `sanitize_text_field()` doesn't run until line 254. A plan `_newspack_network_id` of `"   "` (or `"<b></b>"`) passes the guard, is carried into `$assignments` unsanitized, then sanitizes to `''`. With `--overwrite` the "already set" protection at 271 is bypassed, so line 283 writes an empty string over a correct value — and line 285 fires `newspack_network_save_product`, propagating the blank to every site via `Woocommerce\Events::product_updated`.

**Reproduced at runtime.** With products 48/49 correctly tagged `golden-net` and plan 62's Network ID set to `"   "`, `assign --apply --overwrite` printed `#48 "golden-net" => ""`, wiped both to empty, and reported `Success: Wrote 2 product Network ID(s)` with exit 0. The output displays the destruction but frames it as success.

Two changes: sanitize inside `derive_assignments_from_plans()` (or before the check at line 70), and refuse to write an empty Network ID at line 283 regardless of flags. Without `--overwrite` the same root cause produces a milder bug — an untagged product is tallied as `already set to ""`, so the summary's "N already set" reads as confirmation those products are healthy when none carry an ID.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+	public static function verify( $args, $assoc_args ) {
+		self::require_network_site();
+
+		$current_site     = get_bloginfo( 'url' );
+		$network_products = get_option( Product_Updated::OPTION_NAME, [] );
+		if ( ! is_array( $network_products ) ) {
+			$network_products = [];
+		}
+
+		$explicit_products = isset( $assoc_args['products'] );
```
chickenn00dle: **Blocker:** verify --products= with an empty value exits 0 having checked nothing — `$explicit_products` is computed but never used, so the empty-list branch can't distinguish "no local products are tagged" from "you passed an explicit list that parsed to nothing". `wp_parse_id_list( '' )` returns `[]`, which falls into that branch, warns, and returns — exit code 0.

**Reproduced at runtime**, including the realistic form: `GATE_PRODUCTS=""; wp newspack-network verify-product-network-ids --products="$GATE_PRODUCTS"` exits **0**. The message is also wrong on that path — it printed "No products carry a Network ID on this site" while products 47/48/49 were tagged.

`--products` is the documented escape hatch for the bare-verify gap ("To gate a flip, pass the gate's product IDs"), so this is the exact path an automated flip gate depends on, and it fails green on the most likely operational mistake. When `isset( $assoc_args['products'] )` is true but the parsed list is empty, this should be `WP_CLI::error`.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+	private static function get_products_to_check( $product_ids = null ) {
+		if ( null !== $product_ids ) {
+			$products = [];
+			foreach ( $product_ids as $product_id ) {
+				$products[ (int) $product_id ] = Product_Admin::get_network_id( $product_id );
+			}
+			return $products;
+		}
+
+		$tagged_ids = get_posts(
```
chickenn00dle, line 997: **Blocker:** Everything assign withholds is invisible to verify — the gate greens a network that isn't ready — `assign` produces three classes of unresolved product — conflicts withheld (230-238), variations and non-products skipped (256-263), and "already set to X, would become Y" skipped without `--overwrite` (271-277) — and none leave a trace the second command can see. Bare `verify` builds its product set from a `meta_query` for a non-empty Network ID, so the products `assign` could not resolve are excluded from the check by construction.

**Reproduced at runtime.** Product 48 claimed by two plans with different Network IDs: `assign --apply` correctly withheld it, warned, and exited **0** with "Success". Bare `verify` then reported **"All checked products carry a Network ID and are linked to at least one other site"** and exited **0** — while product 48 sat untagged. An operator following the runbook (assign, then verify) gets a green gate on a network where a gate product grants nothing, which is the NPPD-2057 failure this tool exists to prevent.

The structural fix is to have `verify` default to the plan-derived set (`derive_assignments_from_plans( self::get_plans() )`) rather than "whatever is already tagged" — then the products assign couldn't resolve are precisely the ones the gate fails on. Making `assign` exit non-zero when anything was withheld would also break the chain.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+	 *     @type array $assignments Map of product ID => Network ID.
+	 *     @type array $conflicts   Map of product ID => list of the distinct Network IDs claiming it.
+	 * }
+	 */
+	public static function derive_assignments_from_plans( array $plans ) {
+		$claims = []; // Product ID => list of distinct Network IDs claiming it.
+		foreach ( $plans as $plan ) {
+			$network_id  = (string) ( $plan['network_id'] ?? '' );
+			$product_ids = $plan['product_ids'] ?? [];
+			if ( '' === $network_id || empty( $product_ids ) ) {
```
chickenn00dle, line 115: **Suggestion:** Plans with no Network ID, and products in no plan, are dropped without a word — Conflicts get a per-product warning, but plans skipped by the `'' === $network_id` guard get nothing, and products belonging to no plan are never mentioned. That silent case is the exact NPPD-2057 field condition — a plan nobody ever tagged — and `assign` is the only place in the system positioned to see it, having just enumerated every plan.

As written, an operator on such a site sees "Derived 0 assignment(s)" and exit 0, which is indistinguishable from "already migrated". Per-plan diagnostics (plans found / with a Network ID / with product IDs) plus a non-zero exit when plans exist but nothing was derived would stop a scripted migration recording a green step for a no-op.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+		$skipped  = 0;
+		$already  = 0;
+		$to_write = 0;
+		foreach ( $assignments as $product_id => $network_id ) {
+			$product_id = (int) $product_id;
+			// Sanitize at the write path so the meta is consistent whatever the source ( plan meta or --map ),
+			// matching the product metabox's sanitize_text_field(). Done before the preview so dry-run matches apply.
+			$network_id = sanitize_text_field( (string) $network_id );
+
+			$post_type = get_post_type( $product_id );
```
chickenn00dle: **Suggestion:** Variation IDs in a plan are warned about rather than resolved to the parent — If a plan's `_product_ids` contains variation IDs the loop skips them, telling the operator to set the Network ID on the parent — but the command already has what it needs to do that itself, and `Product_Admin::get_network_id()` performs exactly that resolution when reading. A plan linking only variations therefore produces zero assignments, and per the finding above bare `verify` won't catch it either.

Mapping the variation to `wp_get_post_parent_id()` and folding it into the assignment set (deduped, with a parent claimed by two Network IDs treated as a conflict like any other) would close the gap. I could not confirm from this repo how often Memberships stores variation IDs in `_product_ids`, so treat the frequency as uncertain — the handling gap is not.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+				continue;
+			}
+
+			WP_CLI::line( sprintf( '  #%d "%s" => "%s"', $product_id, $current, $network_id ) );
+			$to_write++;
+
+			if ( $apply ) {
+				update_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, $network_id );
+				// Fire the same action the product metabox fires, so the existing emitter propagates the change.
+				do_action( 'newspack_network_save_product', $product_id );
```
chickenn00dle, line 485: **Suggestion:** Every write's Data Event is queued in memory and flushed only on shutdown — `Data_Events::dispatch()` appends to a queue drained by `execute_queued_dispatches()` on `shutdown`. For a thousands-of-products migration that means the payload of every product accumulates in PHP memory for the whole run, all dispatches fire in one burst (under Action Scheduler as a single async action carrying the entire array), and if the process is killed or hits a time limit before shutdown the postmeta is written but **zero events propagate** — precisely the "wrote but didn't sync" state the `$can_propagate` preflight was added to prevent.

`execute_queued_dispatches()` is public, so flushing every N writes inside the loop is a small change; alternatively make the backfill the documented propagation path rather than a footnote.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+		if ( $apply ) {
+			WP_CLI::success( sprintf( 'Wrote %d product Network ID(s), skipped %d, %d already set.', $to_write, $skipped, $already ) );
+			if ( ! $can_propagate ) {
+				WP_CLI::warning( 'These writes were NOT propagated ( see the warning above ). Once Data Events and WooCommerce are active, replay propagation with:' );
+			} else {
+				// A product already carrying the target Network ID fires no event, so this command cannot
+				// re-propagate an already-tagged-but-desynced product. Point operators at the backfill for that.
+				WP_CLI::line( 'Products already carrying the target Network ID fire no event; if propagation may have been missed ( including for already-set products ), replay it with:' );
+			}
+			WP_CLI::line( '  wp newspack-network data-backfill newspack_network_product_updated --live' );
```
chickenn00dle: **Suggestion:** The recommended data-backfill remediation may be a no-op on a Hub — Both commands point operators at `wp newspack-network data-backfill newspack_network_product_updated --live` as the fix for missed propagation. On a Hub that path runs `Event_Log::persist()`, which dedupes on the exact payload and only reaches the option write when the insert succeeds. The backfiller stamps `timestamp = strtotime( $product->post_modified_gmt )`, which is stable across runs, so replaying it for products already in the event log is deduped and never re-writes the option.

Relatedly, `update_post_meta()` doesn't bump `post_modified_gmt`, so an operator scoping the backfill with `--start`/`--end` to "today" matches nothing. Worth confirming and documenting in the CLI output, since this is the advice printed on the failure path.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+		$unlinked = 0;
+		foreach ( $findings as $product_id => $finding ) {
+			// An untagged product ( only reachable when passed explicitly via --products ) is the NPPD-2057 failure itself.
+			if ( '' === $finding['network_id'] ) {
+				WP_CLI::warning( sprintf( '#%d: no Network ID set ( cross-site access grants nothing ). Run assign-product-network-ids.', $product_id ) );
+				$untagged++;
+				continue;
+			}
+
+			if ( empty( $finding['linked_sites'] ) ) {
```
chickenn00dle: **Suggestion:** Green means "linked to ≥1 other site", not "linked to the site the member subscribes on" — `Access::check_network_subscriptions()` requires the *specific* remote site's map entry for the *specific* product a user holds. In the 4-site Source Media shape, a product linked only hub↔node1 passes on both while node2 and node3 grant nothing. The comment in the code is honest about this, but the success string and the exit code are not — and a scripted per-site gate reads only the exit code.

On a Hub the registered `Hub\Nodes` URLs are available, so listing the network's known sites and reporting which ones share each ID (or an `--expect-sites=N` threshold) would make the green mean what operators will read it to mean.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+	 * @param string $map The raw --map argument.
+	 * @return array
+	 */
+	private static function parse_map( $map ) {
+		// Only treat the argument as a path when it's short enough to be one: passing a long inline-JSON
+		// map straight to is_readable() can trip an E_WARNING ( "File name too long" ) on some platforms.
+		if ( is_string( $map ) && strlen( $map ) < PHP_MAXPATHLEN && is_readable( $map ) ) {
+			$map = file_get_contents( $map ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents, WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown
+		}
+		$decoded = json_decode( (string) $map, true );
```
chickenn00dle, line 1100: **Suggestion:** parse_map accepts JSON lists and silently coerces non-numeric keys — `is_array( $decoded )` doesn't distinguish an object from a list, so `--map='["premium","basic"]'` is accepted and becomes assignments for product IDs 0 and 1. Non-numeric keys go the same way: `{"abc":"premium"}` → product 0, and `{"12abc":"premium"}` → **product 12**, with no warning. Product 0 is caught downstream by the post-type check, but `12abc` writes to a real product.

On PHP 8.3 this is a two-line fix: reject `array_is_list( $decoded )` and require each key to match a positive integer, warning on anything else. `json_validate()` would also let the error distinguish "not JSON at all" from "JSON but not an object", which matters when the argument was meant to be a file path. Relatedly, the skip warnings print `(int) $product_id`, so a bad key renders as `#0` and can't be traced back to the offending map entry.

plugins/newspack-network/tests/unit-tests/test-product-network-ids-cli.php
```diff
@@ -0,0 +1,243 @@
+<?php
```
chickenn00dle, line 1: **Suggestion:** Tests cover only the pure helpers — none of the four blockers would be caught — What's there is good: the verify test asserts the site's own entry is *dropped* rather than merely absent, which is the right assertion for the self-exclusion, and the conflict test checks the unambiguous sibling is still assigned. But all seven tests build plain arrays and call three pure statics — nothing creates a post, product, plan, or option despite extending `WP_UnitTestCase`. `assign()`, `verify()`, `get_plans()` and `get_products_to_check()` have zero coverage, which is where all four blockers live.

The stated obstacle is that `WP_CLI::warning`/`error` are undefined under PHPUnit; a small `WP_CLI` shim in `tests/bootstrap.php` plus post factories would solve it and let `assign()` run end-to-end against real postmeta. Three tests would pin the riskiest behavior: self-exclusion keyed by the real `get_bloginfo( 'url' )` producer; a `wc_membership_plan` fixture asserting the `_product_ids` parse; and a two-plan conflict asserting the withheld product carries no meta after `--apply`. For tooling whose failure mode is thousands of members losing paid access, that's worth the bootstrap change.

plugins/newspack-network/includes/class-initializer.php
```diff
@@ -58,6 +58,7 @@ public static function init() {
 		Data_Backfill::init();
 		Membership_Dedupe::init();
 		CLI\Integrity_Check::init();
+		CLI\Product_Network_Ids::init();
```
chickenn00dle, line 61: **Nit:** Unconditional class reference means an unregenerated classmap fatals the whole site — `newspack-network` autoloads via a composer **classmap**, and `Initializer::init()` references `CLI\Product_Network_Ids` unconditionally — so if the class file ships without a regenerated classmap, every request fatals with `Class "Newspack_Network\CLI\Product_Network_Ids" not found`, not just CLI invocations.

I hit this in my env because the isolated environment hardlinks the main checkout's prebuilt `vendor/`, and it cleared as soon as I ran `composer dump-autoload` — so this is a deploy-hygiene note, not a defect in the PR: the release build runs `composer install`. Flagging it only because a new class in a classmap-autoloaded plugin has a site-wide blast radius if any deploy path ever skips that step. Registration also happens at include time here, whereas sibling CLI classes defer `add_command` to the `init` hook — functionally fine, just inconsistent.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+
+		if ( empty( $assignments ) ) {
+			WP_CLI::warning( 'No assignments to make.' );
+			return;
+		}
+
+		$skipped  = 0;
+		$already  = 0;
+		$to_write = 0;
+		foreach ( $assignments as $product_id => $network_id ) {
```
chickenn00dle, line 432: **Nit:** assign path still has the N+1 the verify path fixed, and both queries are unbounded — The round-1 meta-cache fix landed on `get_products_to_check()` but not here: `get_post_type()` and `get_post_meta()` run per assignment with no priming, so a 3,000-product `--map` is ~6,000 round trips before a single write. `_prime_post_caches( array_keys( $assignments ), false, true )` before the loop covers both.

Separately, `update_meta_cache( 'post', $tagged_ids )` on the verify path loads *all* meta rows for every tagged product, not just the Network ID key — on a large WooCommerce store that's a memory spike, not just a slow query. Chunking with `array_chunk( $tagged_ids, 500 )` would bound it.

plugins/newspack-network/includes/cli/class-product-network-ids.php
chickenn00dle: **Nit:** assign tags product types the metabox would refuse — `Product_Admin::save_meta_box()` bails unless the product is `subscription` / `variable-subscription`, and the metabox only renders for those. The CLI accepts any `product` post type, and `Woocommerce\Events::product_updated()` has no type guard, so a simple or grouped product tagged by the CLI will emit and land in every site's synced map. It's inert for `Access` (which only matches products in a user's synced subscriptions), so this is map bloat rather than a correctness bug — but it diverges from "fires the same action the metabox fires under the same preconditions" and is worth either matching or documenting as deliberate.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,517 @@
...
+			if ( $apply ) {
+				update_post_meta( $product_id, Product_Admin::NETWORK_ID_META_KEY, $network_id );
+				// Fire the same action the product metabox fires, so the existing emitter propagates the change.
+				do_action( 'newspack_network_save_product', $product_id );
+			}
+		}
+
+		WP_CLI::line( '' );
+		if ( $apply ) {
+			WP_CLI::success( sprintf( 'Wrote %d product Network ID(s), skipped %d, %d already set.', $to_write, $skipped, $already ) );
```
chickenn00dle: **Nit:** assign always exits 0, and there is no machine-readable output on either command — Conflicts, type skips and overwrite-withheld products all warn and continue, and the command finishes with `WP_CLI::success` regardless — so a run where every product was withheld is indistinguishable from a clean one to a script. Exiting non-zero when anything was skipped (or a `--strict` flag) would make it composable with the verify gate.

Neither command offers `--format=json` / `--porcelain`. The docblock positions `verify` as a per-site flip gate across a 4-site network, and a gate that emits only human prose plus an exit code is awkward to aggregate. A progress bar on the apply path would also match `Integrity_Check`'s precedent.

chickenn00dle (changes requested):
Re-reviewed the round-1 fixes and exercised both commands end-to-end in an isolated env (WooCommerce + Memberships, `hub` role, 2 plans / 4 products) reproducing the NPPD-2057 starting state — plans tagged, zero products tagged.

**The hard part is right.** I confirmed the Hub self-exclusion is genuinely load-bearing: after `--apply` the Hub self-included its products under its own URL key, and `verify` correctly excluded them and reported *unlinked* rather than green. Dry-run/apply parity is exact, the derive path is correct, a product in no plan is left alone, and the real green path exits 0 only when another site actually shares the ID. All six round-1 points are properly addressed.

**Three blockers, all reproduced at runtime.**

1. **`verify --products=` exits 0 having checked nothing.** An unset shell variable in a runbook (`--products="$GATE_PRODUCTS"`) produces exit 0 and the message "No products carry a Network ID on this site" — which was false at the time, they were tagged. This is the documented escape hatch for the bare-verify gap, so it's the path a flip gate actually depends on.
2. **`--overwrite` can wipe correct IDs.** The emptiness guard runs on the raw plan meta but sanitization happens later, so a whitespace-only plan Network ID survives the guard and sanitizes to `""`. I watched two correct IDs get wiped (`#48 "golden-net" => ""`) and reported as "Success: Wrote 2 product Network ID(s)" — and the write fires the propagation event, so the blank travels network-wide.
3. **Anything `assign` withholds is invisible to bare `verify`.** I made a two-plan conflict; assign correctly withheld the product, warned, and still exited 0 — then bare `verify` reported "All checked products carry a Network ID and are linked" and exited 0 with that gate product untagged. Individually both commands are defensible; composed the way the runbook composes them, they green-light a flip that would strip members.

Smallest set that closes these: `get_flag_value()` for the three flags; error rather than warn when `--products` was passed but parsed empty; sanitize before the emptiness check and refuse to write an empty Network ID regardless of flags; and default `verify` to the plan-derived product set rather than "whatever is already tagged".

One thread connecting all four: the tests only exercise the three pure helpers against hand-built arrays, so `assign()` and `verify()` — where every blocker lives — have no coverage. A small `WP_CLI` shim in the bootstrap plus post factories would make these regressions catchable.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed both ways you suggested. `derive_assignments_from_plans()` now runs `sanitize_text_field()` on the plan value *before* the emptiness guard, so a `"   "` or `"<b></b>"` plan Network ID is treated as "no Network ID" (counted in the new per-plan diagnostics and rolled into the non-zero exit) instead of surviving into the assignment set. The write path additionally refuses an empty Network ID unconditionally, regardless of `--overwrite`, since a blank is indistinguishable from untagged and would propagate network-wide. New test `test_assign_never_writes_a_blank_network_id_over_a_correct_one` reproduces your exact scenario and asserts the correct value survives.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed: when `--products` is passed but `wp_parse_id_list()` yields nothing, the command now `WP_CLI::error`s ("--products was passed but no product IDs could be parsed from it...") instead of falling into the empty-set branch. The dead `$explicit_products` variable is gone. Verified against the env: `verify-product-network-ids --products=` now exits 1. Covered by `test_verify_errors_when_the_products_flag_parses_to_nothing`. The wrong message on that path is gone too, and the genuinely-empty default set is now an error rather than a warning+exit 0 (see the sibling thread on the default product set).

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Did both. Bare `verify` no longer builds its set from "whatever is tagged": it is now every product a membership plan links (via `get_plan_linked_product_ids()`, variations folded to parents, non-subscription and deleted products dropped) unioned with every product that already carries a Network ID — so the products `assign` withholds are exactly the ones the gate now fails on. And `assign` tallies everything it could not resolve (plans with no Network ID, conflicts, post-type skips, "already set to X" without `--overwrite`) and exits non-zero with a summary, so a runbook can't record a green step on a partial run. `test_verify_checks_plan_linked_products_that_assign_could_not_resolve` reproduces your two-plan conflict end-to-end and asserts verify fails on the untagged product while still reporting the healthy one.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Added the diagnostics and the non-zero exit. `derive_assignments_from_plans()` now returns per-plan stats and assign prints "Found N membership plan(s): X carry no Network ID, Y link no products.", warns explicitly when any plan carries no Network ID (with what to do about it), counts those plans as unresolved, and errors out when plans exist but nothing could be derived — so "Derived 0 assignment(s)" can no longer exit 0 and read as "already migrated".

On products belonging to no plan: assign deliberately still leaves them alone, because it has no way to know which Network ID they should get. They are no longer invisible, though — bare `verify` now checks plan-linked products too and `--products` covers a gate's specific products, so an untagged gate product fails the gate rather than passing silently.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Resolved rather than warned, as suggested. `fold_variations_into_parents()` maps any variation in the assignment set to its parent via `wp_get_post_parent_id()` (post parent, not `wc_get_product()`, so it also works with WooCommerce deactivated mid-migration), dedupes, and treats a parent claimed with two different Network IDs as a conflict like any other. So a plan linking only variations now produces real assignments. `test_assign_folds_a_linked_variation_into_its_parent_product` covers it. The same resolution runs on verify's default set, so a plan-linked variation is checked through its parent.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Took the flush option: the loop now calls `Data_Events::execute_queued_dispatches()` every 100 dispatches (`DISPATCH_FLUSH_INTERVAL`) and once more at the end, guarded by `method_exists()` so it is a no-op when the Newspack plugin isn't loaded. That bounds the in-memory queue, spreads the Action Scheduler work, and means a run killed mid-way has already propagated everything up to the last flush instead of nothing.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Confirmed both halves, and replaced the advice rather than documenting a broken remedy. On a Hub, `Abstract_Incoming_Event::process_in_hub()` only calls `always_process_in_hub()` (the option write) when `Event_Log::persist()` returns an insert ID, and `persist()` bails on an identical payload — with the backfiller stamping `timestamp = strtotime( post_modified_gmt )`, a replay for an already-logged product is deduped and never re-writes the option. And `update_post_meta()` doesn't touch `post_modified_gmt`, so `--start`/`--end` scoped to "today" matches nothing.

So the command now offers its own replay: `--repropagate` fires `newspack_network_save_product` for products that already carry the target Network ID as well, which dispatches with a fresh timestamp and therefore isn't deduped. All three places that used to point at `data-backfill` now point at `--apply --repropagate`, and the assign output states the backfill dedupe caveat explicitly.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed. `parse_map()` now rejects an empty mapping and a JSON list (keys `0..n-1`) with distinct errors, and each key must match `/^[1-9][0-9]*$/` — so `"abc"` and `"12abc"` are reported and skipped rather than becoming products 0 and 12. The skip warnings print the raw key (`Skipping --map entry "12abc"`) instead of `#0`, so a bad entry is traceable. Covered by `test_assign_rejects_a_json_list_map` and `test_assign_skips_map_entries_with_non_integer_keys` (which asserts product 12 is untouched).

Left `json_validate()` out: PHPCS pins `testVersion` at 8.0-, so it isn't available across the supported range. The "not JSON at all" vs "JSON but not an object" distinction is carried by the separate error messages instead.

plugins/newspack-network/tests/unit-tests/test-product-network-ids-cli.php
adekbadek: Did the bootstrap change. `tests/class-wp-cli.php` + `class-wp-cli-halt.php` + `wp-cli-utils.php` provide a minimal shim (output captured in memory, `WP_CLI::error()` throws `WP_CLI_Halt` so exit-non-zero is assertable, `\WP_CLI\Utils\get_flag_value()` stubbed; the `WP_CLI` constant is deliberately left undefined so command registration stays off). `tests/unit-tests/test-product-network-ids-cli-commands.php` then drives `assign()` and `verify()` end-to-end against real posts, postmeta and options with plan/product/variation factories — 10 tests, one per blocker plus the happy paths: blank-ID overwrite refusal, conflict withheld + non-zero exit, variation folded to parent, `--products=` empty erroring, bare verify catching a plan-linked untagged product, a green verify keyed on the real `get_bloginfo( 'url' )`, and the two `--map` rejection paths. 17 tests / 49 assertions green in this file.

plugins/newspack-network/includes/class-initializer.php
adekbadek: Fixed the inconsistency: `Product_Network_Ids::init()` now just does `add_action( 'init', [ __CLASS__, 'register_commands' ] )` and the `add_command` calls moved into `register_commands()`, matching `Integrity_Check` and `Data_Backfill`. Verified the commands still register in the env.

Left the unconditional `Initializer` reference as-is — every sibling there is referenced the same way, and as you say the release build runs `composer install`; making this one class conditional would be the odd one out without removing the class of risk.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Both done. The assign path primes once up front (`prime_post_caches( array_keys( $assignments ) )` → `_prime_post_caches( $chunk, false, true )`), so the per-product `get_post_type()` / `get_post_meta()` reads in the loop hit the cache. `get_plans()` primes its plan posts the same way. And all priming — including the verify path — goes through that helper, which chunks at 500 IDs, so the meta load no longer scales unbounded with the number of tagged products.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Matched the metabox. `is_taggable_product()` gates the write on `is_type( [ 'subscription', 'variable-subscription' ] )`; anything else is reported ("a Network ID would grant nothing - left alone") and counted separately as "not applicable" rather than as an unresolved item, since a simple product can never be the NPPD-2057 failure and shouldn't fail the run. Verify's default set applies the same filter, so a plan-linked simple product doesn't become a false flip blocker. When WooCommerce is deactivated mid-migration the type is unknowable, so nothing is filtered there — noted in the helper's docblock.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Agreed, and confirmed at the source: `check_network_subscriptions()` resolves the grant from `$network_products[$site][$product_id]`, i.e. the specific site the reader's subscription lives on, so hub↔node1-only linkage really does grant a node2 subscriber nothing while the exit code says otherwise. Fixed in a6bf10aa69:

- On a Hub, `verify` now reads the network's membership from `Hub\Nodes`, prints it up front, and names per product which registered sites do **not** carry the ID — including on a passing run, since that gap is what an operator has to act on.
- New `--expect-sites=<count|all>` gates the exit code on real coverage. `all` is refused on a Node (which cannot read the Nodes list) rather than silently degrading to "any one site"; a plain count works there.
- The default stays at 1 so existing runbooks do not change meaning underneath them, but the success line now states what it actually proves and points at `--expect-sites`.
- Site comparison is normalized (case, trailing slash) on both sides, so a self-entry differing only by a slash no longer counts as a link.

Covered by `test_verify_expect_sites_all_fails_on_a_node_missing_the_network_id` and `test_verify_expect_sites_all_errors_without_a_known_network_membership`.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Partly done already, partly taken, partly declined.

The exit code was fixed in cb4f8a433a, just before this review landed: `assign` now accumulates everything it could not resolve (plans with no Network ID, conflicts, skipped products) and ends in `WP_CLI::error`, so a fully-withheld run is no longer indistinguishable from a clean one. Pinned by `test_assign_withholds_a_conflicted_product_and_exits_non_zero`.

Machine-readable output: taken for `verify` in a6bf10aa69 — `--format=json` emits a single report (per-product status, linked sites, missing sites, counts, `ready_to_flip`) and suppresses the prose, which is what makes the per-site gate aggregatable across the network.

Declined for `assign`, deliberately: its script-facing signal is the exit code, and it is a write step run once per site rather than something a gate aggregates. Its per-product `#id "old" => "new"` lines are the dry-run audit trail an operator reads before `--apply`, and a second output mode for them would be scaffolding with no caller.

Progress bar declined on the same grounds: `Integrity_Check` walks every user and order, whereas this walks plan-linked plus already-tagged products (tens, not thousands), and a bar would replace exactly those per-product lines. Happy to revisit if a store turns up where the working set is large enough for the run to look hung.

adekbadek:
All three blockers are fixed, plus the coverage gap under them.

1. `verify --products=<empty>` is now `WP_CLI::error` rather than a green no-op, so an unset shell variable in a runbook fails the gate instead of passing it.
2. Plan Network IDs are sanitized in `derive_assignments_from_plans()` *before* the emptiness check, so a whitespace-only plan value is treated as "no Network ID" instead of surviving to blank a product. Belt and braces at the write path too: `assign()` refuses to write a Network ID that sanitizes to empty regardless of `--overwrite`, since a blank is indistinguishable from untagged and propagates network-wide.
3. Bare `verify` now defaults to every plan-linked subscription product **plus** every already-tagged product, rather than "whatever is already tagged" — so exactly what `assign` withholds is what gets checked. `assign` also exits non-zero on anything unresolved (untagged plan, conflict, skipped product), so the two commands compose without a green flip in between.

All four flags go through `\WP_CLI\Utils\get_flag_value()`.

On the coverage thread: there is now a `WP_CLI` shim in the bootstrap (with a `WP_CLI_Halt` exception standing in for the `error` exit) and `TestProductNetworkIdsCLICommands` drives `assign()`/`verify()` end to end against real posts, postmeta and options — including the whitespace-plan wipe, the withheld conflict reaching bare verify, and the empty `--products` error, so each of these regressions is now catchable.

One more of the same class while I was in there: `parse_map()` was dropping unusable entries with only a warning, so a partially-skipped `--map` still exited 0 — a product the operator listed, believing it covered, granting nothing. Skipped entries are now counted and folded into the same non-zero exit, with the existing map test updated to pin it.

plugins/newspack-network/tests/unit-tests/test-product-network-ids-cli.php
```diff
@@ -0,0 +1,243 @@
...
+	 * commands through the WP_CLI shim. )
+	 */
+	public function test_parse_map() {
+		$parse_map_method = new ReflectionMethod( Product_Network_Ids::class, 'parse_map' );
+		$parse_map_method->setAccessible( true );
+
+		$parsed = $parse_map_method->invoke( null, '{"5":"premium","6":" basic "}' );
+
+		// JSON object keys arrive as strings; they must become integer product IDs.
+		$this->assertSame( [ 5, 6 ], array_keys( $parsed ) );
```
chickenn00dle: **Blocker:** `parse_map()` now returns `[ 'assignments' => [...], 'skipped' => int ]` (commit `cb4f8a433`), but `test_parse_map` still asserts the old flat shape: `array_keys( $parsed )` is `['assignments','skipped']` (not `[5, 6]`) and `$parsed[5]` is undefined. `test_parse_map_long_inline_json_is_not_treated_as_a_path` (~line 240) fails the same way — `count( $parsed )` is 2, not `PHP_MAXPATHLEN`. So `PHPUnit / newspack-network` and the aggregate `CI OK` check are both **red**, and a squash-merge can't land. The production caller `assign()` (~line 349) already consumes `$parsed['assignments']` / `$parsed['skipped']` correctly, so this is test-only — update the two assertions to unwrap `$parsed['assignments']` (and assert `$parsed['skipped']` where relevant) and the suite goes green.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,1119 @@
...
+				WP_CLI::line( '  wp newspack-network assign-product-network-ids --apply --repropagate' );
+			}
+		}
+
+		if ( $has_issues ) {
+			// Exit non-zero so callers can gate a flip on this check.
+			WP_CLI::error( 'Verification found issues ( see above ). Not ready to flip.' );
+		}
+
+		WP_CLI::success( sprintf( 'All checked products carry a Network ID and are linked on %s.', $expectation['label'] ) );
```
chickenn00dle, line 729: **Suggestion:** On the passing path this `WP_CLI::success(...)` writes a human `Success: …` line to **stdout**, and it isn't guarded by `$is_json` the way the human-output blocks above it are. So a green `--format=json` run prints the JSON object *and* a trailing `Success:` line, while a failing run (which uses `WP_CLI::error` → stderr) prints only the JSON — stdout is inconsistent between pass and fail. `--format=json` exists to make the per-site gate aggregatable across the network, so a naive `json_decode(stdout)` in that aggregator breaks on healthy sites. It's fail-safe (a healthy site reads as unparseable/not-ready rather than falsely green), but worth guarding the success/note lines behind `! $is_json` so JSON mode emits exactly one report line. The existing `test_verify_json_format_reports_missing_sites` filters to lines starting with `{`, so it won't catch this.

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,1119 @@
...
+				WP_CLI::error( '--expect-sites=all needs the network\'s registered Nodes, which are only readable on a Hub with Nodes registered. Run it on the Hub, or pass an explicit count ( e.g. --expect-sites=3 ).' );
+			}
+			return [
+				'require_all' => true,
+				'minimum'     => count( $known_sites ),
+				'label'       => sprintf( 'every registered network site ( %s )', implode( ', ', $known_sites ) ),
+			];
+		}
+
+		if ( ! preg_match( '/^[1-9][0-9]*$/', (string) $raw ) ) {
```
chickenn00dle, line 776: **Suggestion:** A bare `--expect-sites` (flag passed with no `=value`) arrives from WP-CLI as boolean `true`; it falls past the `null` and `'all'` branches, `(string) true === '1'` matches this `/^[1-9][0-9]*$/` guard, and the coverage bar silently resolves to the 1-site floor — `verify` then greens and exits 0. It fails toward the documented default so impact is limited, but this is the one path that lands on the exact NPPD-2057 failure: an operator who *meant* `--expect-sites=all`, fumbles the syntax, doesn't read the effective-bar line, and flips a node believing full coverage was verified. A quick `is_bool( $raw )` rejection up front turns the fumble into a hard error instead of a falsely-green pass. (Bumped from nit given the member-harm direction — still not a blocker.)

plugins/newspack-network/includes/cli/class-product-network-ids.php
```diff
@@ -0,0 +1,1119 @@
...
+	 * that assign-product-network-ids could not resolve are checked rather than defined away.
+	 *
+	 * @param array|null $product_ids Explicit product IDs to look up; null builds the default set.
+	 * @return array
+	 */
+	private static function get_products_to_check( $product_ids = null ) {
+		if ( null !== $product_ids ) {
+			$products = [];
+			foreach ( $product_ids as $product_id ) {
+				$products[ (int) $product_id ] = Product_Admin::get_network_id( $product_id );
```
chickenn00dle: **Suggestion:** The explicit `--products` branch of `get_products_to_check()` is the one path in the file that doesn't prime caches before the loop — `Product_Admin::get_network_id( $product_id )` runs per product with no `prime_post_caches()`, so each iteration triggers a lazy single-post meta fill plus a `wc_get_product()` load. N+1 on exactly the input a scripted flip gate passes. Negligible for a handful of gate IDs, but inconsistent with the care taken on the default branch (line ~996). A `self::prime_post_caches( array_map( 'intval', $product_ids ) )` before the loop closes it.

chickenn00dle (approved):
Re-reviewed the round-3 changes and traced each fix. **All three blockers from my last review are genuinely closed** — plan Network IDs are sanitized before the emptiness guard (and the write path refuses an empty ID regardless of `--overwrite`), empty `--products` now errors instead of greening the gate, and bare `verify`'s default set is the plan-derived product set unioned with already-tagged products, so anything `assign` withholds is exactly what the gate now fails on. The new surface (`--expect-sites`, `--repropagate`, variation folding, map key validation, URL normalization) holds up, and the new command-level tests are exactly the coverage that was missing.

One thing blocks merge: **the test suite is red.** `parse_map()` was restructured to return `['assignments' => …, 'skipped' => …]`, but two reflection-based tests still assert the old flat `[product_id => network_id]` shape, so `PHPUnit / newspack-network` (and the aggregate `CI OK` check) fail. Production is fine — the caller already uses the new shape — so it's just the two assertions that need updating. A couple of smaller, non-blocking notes inline. Nice work on this round.

I left a handful of suggestions/nits that are non-blocking. I'm going to pre-approve assuming the blocking test failures will be fixed first.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2057-migration-tooling-assign-verify-product-network-ids-across` (stash local work first).

plugins/newspack-network/tests/unit-tests/test-product-network-ids-cli.php
adekbadek: Already fixed in 7a3dc10c5 — both assertions unwrap `$parsed['assignments']` now, and `test_parse_map` additionally asserts `$parsed['skipped']` with a whitespace-only value that sanitizes to `''`, so the counter that feeds the non-zero exit is pinned too. Full suite is green on the current head.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed in e09c45f: `verify` now returns straight after emitting the report in JSON mode, so stdout is exactly one line on the passing path as it already was on the failing one (`WP_CLI::error` goes to stderr) — a naive `json_decode(stdout)` works on healthy and unhealthy sites alike. The verdict is still carried by the exit code and by the report's own `ready_to_flip`.

`test_verify_json_format_reports_missing_sites` no longer filters to lines starting with `{` — it asserts the whole captured output is one line, which is what the aggregator actually consumes and what would have caught this.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Good catch — fixed in e09c45f with an `is_bool( $raw )` rejection ahead of the `all`/integer branches, so a bare `--expect-sites` (and `--no-expect-sites`, which arrives as `false`) is a hard error naming both valid forms rather than a silent fall-through to the 1-site floor. Covered by `test_verify_errors_on_a_valueless_expect_sites_flag`, which sets up a network the 1-site floor *would* have greened, so the test fails if the guard is removed.

plugins/newspack-network/includes/cli/class-product-network-ids.php
adekbadek: Fixed in e09c45f: the explicit branch now intval-maps the IDs once and calls `self::prime_post_caches()` before the loop, matching the default branch. The intval also moved out of the loop body, so the map key and the ID passed to `get_network_id()` are the same value.

PR #639, opened 2026-07-16
fix(cli): translate WCM content rules to valid AC slugs in migrate-membership-gates (NPPD-2063)

adekbadek:
### All Submissions:

Fixes a silent fail-OPEN in the `wp newspack migrate-membership-gates` command. `map_rules_to_ac_format()` copied the WooCommerce Memberships (WCM) rule's `content_type_name` verbatim into the Access Control (AC) rule `slug`. But AC enforcement only honours the slugs `post_types` (value = post-type slugs), `specific_posts` (value = post IDs), `newsletters`, and taxonomy slugs — any other slug resolves through `get_taxonomy()`, returns `false`, and the rule **never matches any post**. So:

- A specific page/post rule (`content_type_name = page|post`, non-empty object IDs) produced `['slug' => 'page', 'value' => [ids]]` → never matched → the gate gated nothing.
- An all-posts rule (`content_type_name = post`, empty object IDs) produced `['slug' => 'post', 'value' => []]` → filtered out as empty by `Content_Rules::get_gate_content_rules()` → an empty-rules gate that gated nothing.

Both are silent fail-OPEN on a live paywall. Only taxonomy rules (category, post_tag) translated correctly, by accident.

**Correct translation** — each WCM rule is now mapped using its own content-type discriminator (`WC_Memberships_Membership_Plan_Rule::get_content_type()`, either `post_type` or `taxonomy`), so custom post types map correctly without any hardcoded name list:

| WCM rule | AC rule |
| --- | --- |
| `content_type_name = <post_type>`, **empty** object IDs | `['slug' => 'post_types', 'value' => ['<post_type>']]` |
| `content_type_name = <post_type>` (incl. custom types), **non-empty** object IDs | `['slug' => 'specific_posts', 'value' => [object IDs]]` |
| `content_type_name = <taxonomy>` (category/post_tag/…) | unchanged — already correct |

Merge/dedup semantics across rules mapping to the same slug are preserved (post-type slugs merge under `post_types`; IDs merge under `specific_posts` / the taxonomy slug).

**`post_types` value-order canonicalization.** These `post_types` rules are the first AC content rules with non-numeric string values. `compute_rules_fingerprint()` (used to group plans with identical rules into one gate) orders each rule's value with `SORT_NUMERIC`, under which every non-numeric string compares as `0` and stays unordered — so `['post','page']` and `['page','post']` would fingerprint differently, splitting two plans that restrict the same post types (in a different rule order) into duplicate gates. The mapping now sorts each `post_types` value (`SORT_STRING`) so the output shape is canonical. `compute_rules_fingerprint()` itself is untouched.

Closes NPPD-2063.

1. On a site with WooCommerce Memberships active and `NEWSPACK_CONTENT_GATES` enabled, create three membership plans with content-restriction rules: (a) a single-page paywall (a specific page), (b) an all-posts rule (whole `post` type, no specific IDs), (c) a category rule.
2. Run `wp newspack migrate-membership-gates` (dry-run), then `wp newspack migrate-membership-gates --live`.
3. Inspect the generated gates' `content_rules` postmeta: the specific-page gate carries `specific_posts` with the page ID; the all-posts gate carries `post_types` with `['post']` (no longer an empty, silently-dropped rule); the category gate carries the taxonomy slug with the term ID.
4. In an incognito window (anonymous reader): the paywalled page, an all-posts article, and a post in the gated category are all gated; a free page stays open.
5. Run the unit suites:
   - `n test-php --filter Test_Membership_Gates_Migration` — includes the characterization tests, now asserting the correct translation table plus the `post_types` canonicalization.
   - `n test-php --filter Test_Content_Gates`.

Verified locally: unit suites green (`Test_Membership_Gates_Migration` 73/73, `Test_Content_Gates` 93/93), PHPCS clean, and an anonymous browser pass confirmed the specific-page / all-posts / taxonomy content gated and free content open.

### Other information:

**Whole-taxonomy rules are refused rather than translated.** Memberships spells "every term of this taxonomy" as a taxonomy rule with no terms, and keeps applying it to terms created after the plan was saved. An AC taxonomy rule names the term IDs it covers, so there is no faithful translation of a rule whose membership keeps growing. A plan carrying one is now skipped, and the summary names the taxonomy so the operator can build that gate by hand. Both the dry-run planning pass and the post-write verification also name any content rule that selects no content, so the two report the same gap. Translating the shape needs an AC slug that does not exist yet; follow-up ticket to come.


plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1343,27 +1352,58 @@ private static function map_rules_to_ac_format( array $wc_rules ): array {
...
 			// entering the rule fingerprint, splitting two plans that restrict
 			// identical content into two gates.
-			if ( self::get_newsletter_list_cpt() === $slug ) {
+			if ( self::get_newsletter_list_cpt() === $content_type_name ) {
 				continue;
 			}
+
+			$object_ids = array_map( 'strval', array_values( $rule->get_object_ids() ) );
+
+			if ( 'taxonomy' === $rule->get_content_type() ) {
```
dkoo, line 1469: **Suggestion:** The taxonomy branch sets the value to the term IDs unconditionally, so a Memberships rule that restricts a whole taxonomy with no specific terms maps to `['slug' => <taxonomy>, 'value' => []]`. `Content_Rules::get_gate_content_rules()` drops empty-value rules, so that restriction silently disappears from the migrated gate — a fail-open. The PR already documents this as a deferred limitation, which is the right call since AC has no "all content in taxonomy X" slug to map it to. The gap worth closing separately is the operator's pre-cutover signal: `compute_pre_write_issues()` (the dry-run planning pass) only checks slug *resolvability*, not empty values, and a taxonomy slug like `category` resolves fine — so a dry-run stays silent. `--live`'s `verify_migrated_gate()` warns only when the dropped rule was the gate's *only* rule; a term-less taxonomy rule sitting alongside a valid one escapes both. Consider having `compute_pre_write_issues()` flag any empty-value rule so dry-run and `--live` agree. (Whether this shape is even constructible depends on whether the Memberships UI permits saving a term-less taxonomy rule — worth confirming.)

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
```diff
@@ -550,11 +635,89 @@ public function test_map_rules_to_ac_format_merges_and_dedupes_object_ids_for_th
...
+			$this->invoke_private_static( 'compute_rules_fingerprint', [ $mapped_posts_first ] ),
+			$this->invoke_private_static( 'compute_rules_fingerprint', [ $mapped_pages_first ] ),
+			'Identical output yields identical fingerprints, so the plans group into one gate.'
+		);
+	}
+
 	/**
 	 * Rules with an empty content-type name are dropped entirely.
 	 */
 	public function test_map_rules_to_ac_format_skips_rules_with_empty_content_type() {
```
dkoo, line 754: **Nit:** Coverage on the mapping is strong. Two small edges aren't pinned: `map_rules_to_ac_format( [] )` returning `[]` (empty input), and two identical whole-post-type rules collapsing to one `post_types` value (the dedup path is exercised only indirectly, via the taxonomy-merge and mixed-set tests). Neither is load-bearing — adding them would just fully lock the contract.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1343,27 +1352,58 @@ private static function map_rules_to_ac_format( array $wc_rules ): array {
...
+				// Taxonomy rules key under the taxonomy slug; the value is the term IDs.
+				$slug  = $content_type_name;
+				$value = $object_ids;
+			} elseif ( empty( $object_ids ) ) {
+				// A whole-post-type rule keys under 'post_types'; the value is the
+				// post-type slug.
+				$slug  = 'post_types';
+				$value = [ $content_type_name ];
+			} else {
```
dkoo, line 1478: **Nit:** The `else` branch treats "not taxonomy" as post-type, which is correct because WC only returns `post_type` or `taxonomy` — but that two-value domain is stated only in the method docblock. A one-line comment at the branch would make the fall-through intent explicit where a future reader lands.

dkoo (approved):
Release risk: LOW. Works as described with no blockers: the mapping is faithful to how the paywall evaluator reads each slug (checked against `Content_Restriction_Control`: `post_types` matches on post-type slug, `specific_posts` on post ID, everything else via `get_taxonomy()`). The stacked characterization tests are rewritten cleanly and the stale "buggy on purpose" scaffolding is removed.

Verified end-to-end in an isolated env with real WooCommerce Memberships: four plans covering every branch (specific page, whole `post` type, `category` term, posts+pages) were migrated by the actual `wp newspack migrate-membership-gates` command (dry-run then `--live`, no fatals), and the generated gates carried exactly the corrected rules — `specific_posts`, `post_types:['post']`, the `category` taxonomy slug, and the merged/canonicalized `post_types:['page','post']`. The real evaluator gated the right content and left the uncovered free page open, and an anonymous browser pass showed the registration gate on the paywalled page and both articles while a free page rendered normally. Unit suites (67/67, 93/93) and PHPCS are green.

One open edge worth a follow-up: a term-less whole-taxonomy rule still maps to an empty-value rule that gets silently dropped. The PR deliberately defers the mapping side of that, which is reasonable — the piece worth folding in is a pre-cutover warning so a dry-run flags that case the way `--live` already partly does. Details inline. Approvable with comments.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
adekbadek: Both pinned in 99565b8: `test_map_rules_to_ac_format_maps_an_empty_rule_set_to_no_rules` (empty input) and `test_map_rules_to_ac_format_dedupes_identical_whole_post_type_rules` (two identical whole-post-type rules collapsing to a single `post_types` value).

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Stated at the branch in 99565b8: a comment above the `if` now says the WC content type is one of exactly two values, so everything below the taxonomy branch is a post-type rule.

adekbadek:
Folded in the pre-cutover warning for the term-less whole-taxonomy case (99565b8). `describe_valueless_rules()` names the rules whose value is empty — the ones `Content_Rules::get_gate_content_rules()` drops at read time — and `compute_pre_write_issues()` now calls it, so a dry run reports it before the operator commits to `--live`. The same helper runs in `verify_migrated_gate()`, which read the evaluated rules and so only caught the case where every rule was dropped; it reads the written meta now and reports a partial drop as a partial leak ("1 of its 2 content rules select no content (category), so the content they were meant to cover stays ungated"). Mapping the case to a real AC rule is still deferred. Covered by `test_compute_pre_write_issues_flags_rules_that_select_no_content` and `test_verify_migrated_gate_flags_a_written_rule_that_selects_no_content`, and the "deliberately not addressed" note in the description now says the gap is warned about rather than silent.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Confirmed constructible, and it is how Memberships spells "all terms": the rule's object field carries the placeholder `Search… or leave blank to apply to all`, and `WC_Memberships_Admin_Membership_Plan_Rules::save_rules()` only calls `set_object_ids()` when the posted value is non-empty, with nothing requiring terms. The rule also keeps applying to terms created after the plan was saved, so even a term-ID snapshot would drift.

That pushed the fix past the warning:

- `compute_pre_write_issues()` now flags any empty-value rule, through a `describe_valueless_rules()` helper that `verify_migrated_gate()` also uses, so the dry-run and `--live` report the same gap (99565b8).
- A plan carrying a term-less taxonomy rule is skipped outright rather than migrated, with the taxonomy named in the summary's skipped row (33e5d51). The empty-value check stays as the backstop for any future mapping path that produces one.

The skip is per plan, not per rule: a plan mixing a term-less taxonomy rule with valid ones migrates nothing, on the grounds that a partial gate enforcing less than the summary reports is worse than an untouched plan the operator handles by hand. Happy to switch to migrating the valid rules and warning loudly instead if you read that trade differently.

PR description updated to describe the refusal.

PR #641, opened 2026-07-16
fix(components): keep the more menu on enabled unsupported cards

thomasguillot:
Stacked on #632 (base `pr-a/split-pending-changes`).

## Problem

Enable ESP on Mailchimp, then switch the provider to Manual. The card correctly shows the red "Requires an API-based ESP" badge and a "Change provider" button, but its More menu vanishes, so there is no Disable and no Logs for an integration that is enabled and still operable.

Cause: in `card-feature/index.tsx` the More dropdown is gated on `isConfigureState` (`enabled && !requirements`), so any `requirements` string hides the menu.

## Why not just gate on `enabled`

The enabled+requirements state means opposite things across consumers:

- **ESP integrations** (unsupported provider): degraded but operable, sets `requirementsActionable` -> More menu should show.
- **experimental-tools** (`constant_active`, managed by a wp-config constant): locked, does not set `requirementsActionable` -> More ("Disable") must stay hidden.
- **content-gates** ("Requires metering"): prerequisite missing, no `requirementsActionable` -> unchanged.

A blanket `enabled` gate would expose a broken "Disable" on a constant-locked tool.

## Fix

Gate the dropdown on operability, reusing the existing `requirementsActionable` signal:

```js
const showMoreControls = enabled && !! moreControls?.length && ( ! requirements || requirementsActionable );
```

`isConfigureState` stays as-is for the primary button's label, variant, and click handler. Exactly one behavior row changes (enabled + requirements + actionable: hidden -> shown); every other card renders its More menu identically. No consumer files change.

Known edge: an enabled ESP that is unsupported with no remediation target (`requirementsActionable` false, e.g. empty `setup_url`) keeps the menu hidden. Rare, and the deliberate cost of tying the gate to the remediable signal.

## Testing

`packages/components` has no test runner and is not in the `--filter newspack` suite, so the test lives in `newspack-plugin` and renders the real `CardFeature`. It asserts the "More" toggle across the 5-row matrix; the enabled+actionable-requirement row is the discriminator (fails against the old gating), and the enabled+non-actionable row pins the locked case so a future "gate on `enabled`" regression is caught. Suite: 41 suites / 374 tests green.

Note: interactive browser QA (ESP card, experimental-tools locked) is pending a manual pass; the component behavior is covered by the real-component unit test above.

### DIFF
--- a/packages/components/src/card-feature/index.tsx
+++ b/packages/components/src/card-feature/index.tsx
@@ -58,7 +58,9 @@ type CardFeatureProps = {
 	requirements?: string;
 	/**
 	 * When `requirements` is set, keep the primary button clickable so the
-	 * user can remediate the unmet requirement from this card.
+	 * user can remediate the unmet requirement from this card, and keep the
+	 * "More" dropdown available — the feature is degraded but still operable
+	 * (e.g. can be disabled), unlike a hard-locked requirement.
 	 */
 	requirementsActionable?: boolean;
 	/** Primary button label when not enabled. Default: "Enable". */
@@ -71,7 +73,7 @@ type CardFeatureProps = {
 	onEnable?: () => void;
 	/** Called when the primary button is clicked and the feature is enabled. */
 	onConfigure?: () => void;
-	/** Controls rendered inside the "More" dropdown, shown only when enabled. */
+	/** Controls rendered inside the "More" dropdown, shown when enabled — including the unmet-requirements state when `requirementsActionable`. */
 	moreControls?: MoreControl[];
 	/** Badge text shown when enabled. Default: "Enabled". */
 	badgeText?: string;
@@ -118,6 +120,7 @@ const CardFeature = ( {
 
 	const isConfigureState = enabled && ! requirements;
 	const buttonLabel = isConfigureState ? configureLabel ?? __( 'Configure', 'newspack-plugin' ) : enableLabel ?? __( 'Enable', 'newspack-plugin' );
+	const showMoreControls = enabled && !! moreControls?.length && ( ! requirements || requirementsActionable );
 
 	const handleButtonClick = () => {
 		if ( isConfigureState ) {
@@ -174,7 +177,7 @@ const CardFeature = ( {
 								>
 									{ buttonLabel }
 								</Button>
-								{ isConfigureState && !! moreControls?.length && (
+								{ showMoreControls && (
 									<DropdownMenu
 										icon={ moreVertical }
 										label={ __( 'More', 'newspack-plugin' ) }
--- a/plugins/newspack-plugin/src/wizards/audience/views/integrations/card-feature-more-menu.test.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/integrations/card-feature-more-menu.test.js
@@ -41,6 +41,11 @@ describe( 'CardFeature More menu gating', () => {
 		expect( moreMenu() ).not.toBeInTheDocument();
 	} );
 
+	it( 'hides the More menu when not enabled even with an actionable requirement', () => {
+		renderCard( { enabled: false, requirements: 'Requires an API-based ESP', requirementsActionable: true } );
+		expect( moreMenu() ).not.toBeInTheDocument();
+	} );
+
 	it( 'hides the More menu when there are no controls', () => {
 		renderCard( { enabled: true, moreControls: [] } );
 		expect( moreMenu() ).not.toBeInTheDocument();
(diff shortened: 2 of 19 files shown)


plugins/newspack-plugin/src/wizards/audience/views/integrations/card-feature-more-menu.test.js
```diff
@@ -0,0 +1,48 @@
...
+		expect( moreMenu() ).toBeInTheDocument();
+	} );
+
+	it( 'hides the More menu when the requirement is not actionable (locked)', () => {
+		renderCard( { enabled: true, requirements: 'Managed by site configuration', requirementsActionable: false } );
+		expect( moreMenu() ).not.toBeInTheDocument();
+	} );
+
+	it( 'hides the More menu when not enabled', () => {
+		renderCard( { enabled: false } );
```
miguelpeixe, line 40: **[nit] Add a row locking `enabled` as a hard gate**

The matrix covers the important cases, but omits `enabled: false, requirements set, requirementsActionable: true` — the row that would prove `enabled` stays a hard gate independent of `requirementsActionable` (the current "not enabled" row only covers the *no requirements* case). One extra line locks that a future refactor can't let `requirementsActionable` alone surface the menu on a disabled card.

packages/components/src/card-feature/index.tsx
```diff
@@ -118,6 +120,7 @@ const CardFeature = ( {
 
 	const isConfigureState = enabled && ! requirements;
 	const buttonLabel = isConfigureState ? configureLabel ?? __( 'Configure', 'newspack-plugin' ) : enableLabel ?? __( 'Enable', 'newspack-plugin' );
+	const showMoreControls = enabled && !! moreControls?.length && ( ! requirements || requirementsActionable );
```
miguelpeixe, line 123: **[nit] `requirementsActionable` now couples two affordances**

This prop now drives two distinct things: the primary button's clickability (existing, line 173) and the More-menu visibility (new, here). For the single motivating consumer that coupling is exactly right, and the JSDoc reframing ("degraded but still operable") is coherent — reusing the prop rather than adding a dedicated one is a reasonable, pragmatic call. Flagging only so it's a conscious constraint: a future consumer can no longer keep the primary button actionable while hiding the More menu (or vice versa). No change needed now; split into a dedicated prop if that need ever materializes.

miguelpeixe (approved):
Approving — no blockers.

A tight, well-reasoned fix. The change is **strictly additive**: the new `showMoreControls` predicate is a superset of the old gate, so no card that previously showed its More menu now hides it — the only net-new exposure is `enabled + an actionable requirement`, which is exactly the bug (an enabled ESP switched to an unsupported Manual provider keeping its Logs/Disable). I traced all four consumers: only the ESP card sets `requirementsActionable`, and the wp-config-locked Experimental Tools case never does, so its always-present "Disable" stays correctly hidden — the broken-Disable-on-a-locked-tool scenario is structurally impossible. The new real-component test pins all five matrix rows.

One thing worth doing before merge (couldn't leave this inline — the file isn't in the diff):

**`packages/components/src/card-feature/README.md`** — the in-code JSDoc was updated for the new gating, but the co-located README (the canonical doc for this shared component) still describes the old behavior in three places:
- the **States** table still lists the dropdown as "Hidden" for the unmet-requirements state — it now shows on an enabled card with `requirementsActionable`;
- the `requirementsActionable` prop description says it only keeps the primary button clickable — it now also gates the More menu;
- the `moreControls` prop description says "shown only when enabled" — visibility now also depends on `requirements` / `requirementsActionable`.

Reconcile those with the new gating (matching the updated `index.tsx` JSDoc). Two small inline nits below.

plugins/newspack-plugin/src/wizards/audience/views/integrations/card-feature-more-menu.test.js
thomasguillot: Added — a new row asserts `enabled: false` + `requirements` + `requirementsActionable: true` keeps the menu hidden, locking `enabled` as a hard gate independent of `requirementsActionable`.

packages/components/src/card-feature/index.tsx
thomasguillot: Conscious call — reusing `requirementsActionable` for both the primary button's clickability and the More-menu visibility fits the single motivating consumer, and the "degraded but still operable" framing is intentional. Happy to split into a dedicated prop if a future consumer needs to decouple the two.

PR #642, opened 2026-07-16
fix(cli): consolidate overlapping plan rule sets and preserve OR semantics in migrate-membership-gates (NPPD-2064)

adekbadek:
### All Submissions:

Closes NPPD-2064.

WooCommerce Memberships admits a reader holding **any** plan that covers a piece of content. `migrate-membership-gates` grouped plans by an exact content-rule fingerprint, so two plans gating the same content under different rule sets became two gates with disjoint product lists, and a reader entitled through the plan whose gate lost on priority is denied.

- Groups merge by what they cover, not by identical rules. Terms expand the way the evaluator expands them, so a plan gating a parent category absorbs one gating its child. Merging never crosses the purchase/signup boundary, and overlaps where neither side covers the other stay separate.
- An overlap that merging cannot repair, and whose rule sets allow it, is repaired by an exclusion carve-out: the broad gate excludes the narrow group's content, and the narrow gate takes on the broad group's access products and the access length they need. Both sides must require a purchase.
- Merging widens access, a carve-out rewrites a gate's content rules and moves access products between gates, and an un-mergeable overlap leaves a reader liable to be denied, so `--live` asks about all three before writing. Declining exits non-zero, so a chained Memberships deactivation stops there. The overlap warning names each side's active membership count; the carve-out warning names how many posts it actually covers.
- Product variations reach the gate unchanged. A gate rule matches an order line item on its product ID or its variation ID, which is the test Memberships applies. Dropping them emptied the paid list for a plan listing only variations, leaving a gate that admits every registered reader.
- A paid group that would emit no access rule stops a `--live` run and is a WARN row on a dry run, so the whole set can be fixed in one pass. It was a warning before, and only on a dry run.
- A `--live` run exits non-zero when it wrote an error row or a gate flagged as not restricting as intended.

Because gates can now name a variation, both product pickers list them, and the three editor fields that resolve a selection back to a product share one helper so same-named products and sibling variations stay distinct. Docblocks in the subscription audit and premium-newsletters commands are brought in line, and the audit's picker comparison now asserts the parent-only subset it actually holds.

The base retargeted to `main` once the lower PRs in the stack merged, and the branch was rebuilt on it, so earlier inline threads no longer anchor.

Use a site with WooCommerce Memberships active.

1. Create two purchase plans, each with its own subscription product.
2. Gate one plan on a category. Gate the other on that category and all posts.
3. Run `wp newspack migrate-membership-gates`. One gate is listed, not two.
4. Confirm its paid rule carries both plans' products.
5. Re-gate the second plan on a child of the first plan's category. Re-run. One gate is still listed.
6. Add a plan whose only product is a subscription variation. Re-run. Its gate names the variation ID.
7. Run with `--live` and no `--yes`. The run stops and asks before writing.
8. Delete the product of a purchase plan. Re-run. The plan gets a WARN row and the table still prints.
9. Re-run with `--live`. The run refuses and writes nothing.
10. Open a migrated gate in Audience → Access Control. The variation is listed in the product field.
11. Give two products the same name. Both appear in that field with their IDs.

### Other information:

<details>

**Consolidation.** Phase 1b sits between fingerprint grouping and the pre-flight, so every later check and the write loop see the gate groups that get written. `plan_rule_set_consolidation()` decides from rule sets and purchase flags alone; `consolidate_plan_groups()` does the fold. Absorption chains resolve to their terminal root. Ties between equally sized covering roots, and mutual coverage, break on the rule set's own signature, so the partition does not move with the order the plans arrive in.

**Term expansion.** `Content_Restriction_Control::expand_hierarchical_terms()` is now public and called by the migration rather than a second copy. Only the comparison sees expanded values; the gate is written from the group's own rules.

**Variations.** `WC_Subscription::has_product()`, `customer_bought_product_after()` and `wc_customer_bought_product()` each test both `product_id` and `variation_id`, matching `wc_memberships_get_order_access_granting_product_ids()`. Substituting the parent would admit buyers of sibling variations the plan never covered.

**Editor.** `getProductTokens()` builds a bijection between IDs and token strings, and keeps a `#<id>` token for a stored ID the picker no longer offers, so editing a rule cannot silently drop it. It is now used by the access-rule control, block visibility, and the gate summary.

**Not addressed here.** A gate naming one variation drives the one-tier paywall pattern to sell that variation rather than the parent's tier picker, which is what unlocks that gate. On a network node, a variation without its own Network ID resolves to the parent's. Both predate this change and neither is altered by it.

</details>


plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -662,4 +670,314 @@ function ( $rule ) {
...
+	 */
+	private static function group_product_ids( array $group ): array {
+		$product_ids = array_values( array_unique( array_merge( [], ...array_column( $group, 'product_ids' ) ) ) );
+		return array_values(
+			array_filter(
+				$product_ids,
+				fn( $id ) => 'product_variation' !== \get_post_type( $id )
+			)
+		);
+	}
```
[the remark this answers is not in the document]
adekbadek: Done in 6c56ae52f — group_product_ids() now casts the union to int, so gate access rules only ever carry integer product IDs.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
```diff
@@ -382,6 +383,352 @@ public function test_compute_rules_fingerprint_differs_for_different_rules() {
...
+		$group = [
+			[ 'product_ids' => [ '103' ] ],
+			[ 'product_ids' => [ '101', '102' ] ],
+			[ 'product_ids' => [ '102' ] ],
+		];
+
+		$this->assertSame(
+			[ '103', '101', '102' ],
+			$this->invoke_private_static( 'group_product_ids', [ $group ] )
+		);
```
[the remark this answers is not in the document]
adekbadek: Done in 6c56ae52f — the test now uses int product IDs to match production (WooCommerce product IDs are ints) and the method's int[] contract.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1091,343 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+					self::gate_title( $groups[ $root ] )
+				)
+			);
+		}
+
+		foreach ( $plan['overlaps'] as $pair ) {
+			list( $i, $j ) = $pair;
+			// Overlap indices are always roots, so both keys survive in $merged, and any
+			// plan folded into a root is named in the warning.
+			WP_CLI::warning( self::format_overlap_warning( $merged[ $i ], $merged[ $j ] ) );
```
chickenn00dle: **Blocker:** Un-mergeable overlaps are written without confirmation, though they are the denial shape this fixes — When two roots overlap but neither covers the other, the run emits a `WP_CLI::warning` and carries on. Those pairs are never added to `$to_confirm`, so `--live` writes them unattended.

I reproduced this: with one overlapping pair and no merges, `--live` *without* `--yes` on non-interactive input exited 0, prompted for nothing, and wrote two gates with disjoint product lists (16 vs 15). Because `is_post_restricted()` returns restricted on the first gate that denies, that is exactly the NPPD-2064 shape — a reader entitled through one plan denied at the other gate. The commonest instance is a `post_types` group against a `specific_posts` group, which `rules_cover()` deliberately never merges.

To be fair to the PR: before this change those splits happened silently with no warning at all, so this is an improvement rather than a regression. The asymmetry is what stands out — the widening this PR *introduces* is fenced behind a prompt, while the denial it knowingly *cannot* fix is left as scrollback. Adding overlaps to `$to_confirm` alongside the merges would make the two risks symmetric, and naming the affected plans' member counts would meet the ticket's stated floor ("emit a loud warning listing the cross-plan subscribers who would be denied by the split").

One more thing sharpens this. Access Control gates do nothing while WooCommerce Memberships is still active — `Content_Restriction_Control::is_post_restricted()` returns early on `Memberships::is_active()`. So a run that writes split gates looks completely clean at the time it runs, exits 0, and prints a success line. The denial only becomes visible to readers after Memberships is deactivated at cutover, by which point whoever ran the migration has moved on and the warning is long out of scrollback.

**Risk if not addressed:** Because gates do nothing while WooCommerce Memberships is still active, a run that wrote split gates looks clean at the time it runs. The problem only becomes visible after Memberships is deactivated, at which point readers entitled through one plan are denied the other gate's content — the exact outcome this change exists to prevent — and neither the exit code nor the success line gave the person running the cutover any reason to stop.

**Consider:** Adding the un-mergeable overlap pairs to the same pre-flight confirmation the merges already use would mean a live run cannot write two gates with disjoint product lists without someone saying yes to it. The pairs are already computed by the time the prompt is built.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -247,6 +256,21 @@ public function migrate_membership_gates( $args, $assoc_args ) {
 				$needs_duration = array_merge( $needs_duration, $durations_by_group[ $fingerprint ]['plans'] );
 			}
 		}
+		// A purchase group with no usable product IDs writes an active paid mode with no
+		// access rule, which asks for no purchase at all — every registered reader gets
+		// in. Refused before the first write rather than reported after it: on --live
+		// the post-write verification is the only thing that would notice, by which
+		// point the open gate is already published.
+		$paid_without_products = self::find_paid_groups_without_products( $plan_groups, $products_by_group );
```
chickenn00dle: **Suggestion:** A dry run now aborts on the first unusable paid group, losing the rest of the planning table — `find_paid_groups_without_products()` raises `WP_CLI::error` unconditionally, in both modes. Refusing before the first write on `--live` is clearly right. On a dry run there is no write to refuse, and aborting costs the operator the thing a dry run is for: the full plan-by-plan table. They now fix one plan, re-run, and discover the next one.

I confirmed the behaviour — with one plan whose product had been deleted, the dry run exited 1 and printed no summary.

The docs have drifted with it: the command docblock at :68 still lists "no usable product IDs" among the gaps that "appear in dry-run" as WARN rows, and `compute_pre_write_issues()`'s docblock at :848 repeats it. A WARN row on dry-run and the hard error on `--live` would match both the docs and the intent.

**Consider:** In dry-run mode, collecting every group with no usable products and reporting them after the planning table — keeping the hard stop for live runs — would preserve the fail-closed behaviour while letting an operator fix the whole set in one pass.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -643,6 +678,29 @@ private static function find_colliding_gate_titles( array $plan_groups ): array
...
+	 * resolve, writes an active paid mode carrying no access rule — which demands no
+	 * purchase and admits every registered reader. The caller refuses the run on a
+	 * non-empty result.
+	 *
+	 * @param array[] $plan_groups       Gate groups, keyed as $products_by_group is.
+	 * @param array[] $products_by_group resolve_product_ids() results, keyed by group.
+	 *
+	 * @return string[] The gate titles at risk, in group order.
+	 */
+	private static function find_paid_groups_without_products( array $plan_groups, array $products_by_group ): array {
```
chickenn00dle: **Suggestion:** The unusable-products check and the rule builder decide on different things — `find_paid_groups_without_products()` tests `empty( $products_by_group[ $key ]['product_ids'] )`, but what actually determines whether an open gate gets published is whether `build_access_rules()` (:1587) emits anything — that is, whether `subscription_ids ∪ one_time_ids` is empty.

Those agree today only because `classify_product_ids()` partitions exhaustively: subscription or one-time, no third branch and no drop. That equivalence is load-bearing and stated nowhere. If anyone adds a bucket, or makes `classify_product_ids()` skip a type it cannot place, this check quietly stops catching the case it exists for and `verify_migrated_gate()` becomes the only thing that notices — after the open gate is published. Asserting on the classified buckets directly, or noting the invariant at both ends, would keep them from drifting.

**Risk if not addressed:** If a later change gives product classification a third outcome — a product neither paid rule can carry — the group still has a non-empty product list, so the guard passes, no rule is emitted, and a paid gate publishes that lets in every registered reader. That gate reads as a successful migration in the summary, and the plan's content becomes free to anyone with an account once Memberships is switched off.

**Consider:** Having the pre-flight test what would actually be written — whether any access rule would be emitted — rather than whether the product list is non-empty removes the dependency on every resolved product continuing to fall into one of the two buckets.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1091,343 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+				// Mutual coverage is ordinary once terms are expanded — {parent} and
+				// {parent, child} gate the same content under different fingerprints —
+				// and those two belong on one gate as much as a strict subset does.
+				// Absorbing only into a lower index keeps the relation antisymmetric, so
+				// the chain below still terminates.
+				if ( self::rules_cover( $rules_i, $rules_j ) && $j > $i ) {
+					continue;
+				}
+				$size = count( $rules_j );
+				if ( $size > $best_size ) {
```
chickenn00dle: **Suggestion:** Absorption tie-break counts rules rather than coverage, making the partition order-dependent — The docblock says a subset group is absorbed into "the largest covering group", but `$best_size = count( $rules_j )` measures how many rules a group has, not how much it covers. A one-rule `post_types: [post, page]` group covers far more than a narrow three-rule group and loses the tie.

This only bites among covering roots that are incomparable to each other — if one covered the other, chain resolution would forward the group outward anyway — but there the choice is positional. Fuzzing input-order permutations, about 1.6% of random catalogues produced a different partition purely from ordering.

`get_plans()` orders by ID ascending, so a fixed plan set is reproducible run to run. The exposure is narrower and more specific: a plan published or unpublished between the approved dry run and the `--live` run can reshuffle which gate an absorbed plan's products land on, and therefore who is denied at the other gate. Tie-breaking on the group's own fingerprint instead of array position would make it order-independent; the docblock should say "most rules" either way.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -461,25 +471,111 @@ function ( $rule ) {
...
+	 * is saved and does not regenerate when its attributes change afterwards. The
+	 * editor's token field resolves a selection back to a rule value by label, so two
+	 * options sharing a label select the wrong variation.
+	 *
+	 * @param \WC_Product $parent    The parent product.
+	 * @param \WC_Product $variation The variation (a \WC_Product_Variation in production).
+	 *
+	 * @return string
+	 */
+	private static function variation_option_label( \WC_Product $parent, \WC_Product $variation ): string {
```
chickenn00dle: **Suggestion:** `variation_option_label()` is typed wider than its call site, and the wider type fails silently — The signature takes `\WC_Product $variation`, but `build_product_options()` only ever passes objects from a `type => variation` query, which are `WC_Product_Variation`. The widening looks driven by the test mock, and it gives up a real guarantee: `wc_get_formatted_variation()` short-circuits on `is_a( $variation, 'WC_Product_Variation' )` and returns `''` for anything else, so a non-variation silently falls through to the `%1$s (variation #%2$d)` branch and produces a nonsense label instead of failing.

Typing the parameter `\WC_Product_Variation` and adjusting the mock would close it. Keeping `\WC_Product` is defensible too, but then the docblock's "a `\WC_Product_Variation` in production" caveat is documenting a gap rather than closing it.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -350,12 +350,17 @@ private static function rule_matches_post( $content_rule, $post_id ) {
...
+	 * two disagree about what a rule gates, and the migration would then merge, or
+	 * refuse to merge, on a reading the site never applies.
+	 *
 	 * @param array        $term_ids Term IDs from a content rule's value (may be stored as strings).
 	 * @param \WP_Taxonomy $taxonomy Taxonomy object the term IDs belong to.
 	 *
 	 * @return int[] De-duplicated term IDs including descendants.
 	 */
-	private static function expand_hierarchical_terms( array $term_ids, \WP_Taxonomy $taxonomy ): array {
+	public static function expand_hierarchical_terms( array $term_ids, \WP_Taxonomy $taxonomy ): array {
```
chickenn00dle, line 369: **Suggestion:** `expand_hierarchical_terms()` becomes public with no way to invalidate its memo — Making this public is the right call — sharing one expansion is what keeps coverage decisions and access decisions in agreement, and taking a `\WP_Taxonomy` self-guards the pre-`init` case because holding the object implies the taxonomy is registered.

What comes with it is that `self::$term_descendants_map` is now reachable from outside this class with no invalidation path: no flush method, and no hook on `edited_term` / `created_term` / `clean_term_cache`. Within one web request that is fine. From `newspack-manager`, or from a long-lived WP-CLI process that edits terms (a taxonomy remap, an importer), it will return a stale descendant set and therefore a stale access decision.

The asymmetry is what makes this worth raising: this same PR adds `Access_Rules::flush_product_options_memo()` for exactly this reason, on a memo that is *not* part of a shared contract. A matching `flush_term_descendants_memo()`, or hook-based invalidation, plus a docblock line saying the memo is request-scoped, would settle it.

**Consider:** Worth bounding rather than over-engineering: the descendant cache is only read within a single request or CLI process, and nothing else in the workspace calls the expansion today. A flush method mirroring the one added alongside it, callable from tests, covers the realistic exposure; a term-edit invalidation hook is not needed for the current consumers.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -277,11 +301,22 @@ public function migrate_membership_gates( $args, $assoc_args ) {
...
+		// One prompt for the whole run, so the write loop never reads STDIN. Both of
+		// these hand the operator something they cannot simply undo afterwards: a
+		// superseded gate stays published and keeps restricting, and a consolidated gate
+		// has already widened paid access by the time the summary is read.
+		$to_confirm = [];
+		if ( ! empty( $superseding ) ) {
+			$to_confirm[] = sprintf( 'create %d gate(s) that supersede the gates named above', count( $superseding ) );
+		}
+		if ( ! empty( $widened_by_consolidation ) ) {
+			$to_confirm[] = sprintf( 'merge %s, widening paid access as described above', implode( ', ', $widened_by_consolidation ) );
```
chickenn00dle: **Nit:** Signup-only merges are described as widening paid access — Both operator-facing strings assume the fold is a paid one — `$to_confirm` says "widening paid access as described above" and the per-merge warning says "one gate carries both plans' access products". But `plan_rule_set_consolidation()` also merges signup groups with signup groups (the flags only have to *match*, not be `true`), and those carry no access products at all.

I confirmed it: two `signup` plans with no products consolidated, and the warning still claimed both "access products" and widened access, while the summary row read Access Type `signup`. On `--live` that produces a prompt whose stated stake is wrong, which could just as easily make an operator abort a harmless merge as wave through a real one. Branching the copy on `group_requires_purchase()` would fix both directions.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
```diff
@@ -2311,4 +2291,281 @@ public function test_a_whole_taxonomy_rule_still_maps_to_a_value_the_reader_disc
...
+		$this->assertSame( [], $plan['absorbed_by'] );
+		$this->assertSame( [ [ 0, 1 ] ], $plan['overlaps'], 'The cross-boundary overlap is reported instead.' );
+	}
+
+	/**
+	 * A gate group of one plan, gating the given terms of one taxonomy.
+	 *
+	 * @param string $name        Plan name, which becomes part of the gate title.
+	 * @param int[]  $term_ids    Category term IDs the plan gates.
+	 * @param int    $product_id  The plan's access product.
```
chickenn00dle: **Nit:** New consolidation tests can leak cached term descendants between cases — Both new `consolidate_plan_groups()` tests exercise `Content_Restriction_Control::expand_hierarchical_terms()`, which populates the private static `$term_descendants_map`, and this class doesn't reset it. The repo already ships `Trait_Restriction_Cache_Test` for exactly this, and its docblock spells out why: cases roll back but reuse term IDs, so an earlier case's cached descendants get served to a later one.

`test_consolidate_plan_groups_folds_nested_tiers_into_one_group` compounds it by using bare term IDs 5, 6, 7 while its sibling creates real parent/child categories in the same process. Pulling in the trait and using factory-created terms in both would remove the ordering dependence. Not failing today, but it is the kind that fails once, in CI, on an unrelated PR.

chickenn00dle (changes requested):
Release risk: HIGH — two confirmed blockers, both about what a live run writes without asking, on one-shot tooling that rewrites entitlement against production data.

The core of this looks good. The consolidation relation is sound: coverage is a preorder, the purchase-flag filter partitions rather than crosses, and the mutual-coverage break on the lower index forces every edge inside an equivalence class to point strictly downward, so absorption is acyclic and always terminates at a real root. Fuzzing 400k random catalogues found no missing or self-rooted groups. Coverage is decided on expanded rule sets and the gate is written from the unexpanded ones, which is right because the evaluator re-expands at read time. I confirmed the whole path on a running site against real Memberships, including that a merged gate carries a variation ID.

Both blockers are small fixes. The rest are smaller and mostly about the operator's view. One question rather than a request: exclusion carve-outs are a first-class rule shape in the evaluator, the JS matcher and the wizard, and for the Asia Times case they would reproduce Memberships' semantics with zero widening — write the broad gate as "all posts except category 155170" beside the narrow one. They don't generalise (the wizard keys rules by slug, so nested same-taxonomy tiers can't express it, and `specific_posts` is an inclusion override evaluated before exclusions), and this migration would be the first producer of exclusion rules anywhere in the codebase, which is its own risk. Was that considered and ruled out?

- `plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php:104` — **Blocker:** `--yes` also waives the new access-widening prompt, which its documentation doesn't mention — The `--yes` text describes one prompt: the pre-flight confirmation "shown when gates would be created alongside gates the same plans were migrated to individually". This PR adds a second, materially different prompt at :313 — the one that confirms merging plans and widening paid access — and both are folded into a single `$to_confirm` list passed to one `confirm_or_error( ..., $assoc_args, ... )`. `WP_CLI::confirm()` returns immediately when `yes` is set, so `--yes` pre-approves both. I confirmed this on a running site: `migrate-membership-gates --live --yes` merged three plans and widened access with no prompt at all. What makes this worth fixing before merge rather than after is who it affects. The same paragraph tells operators `--yes` is *required* for non-interactive runs (`cron`, `ssh host "wp ..."`), so the people most likely to pass it are passing it for a reason unrelated to entitlement — and the code's own comment at :305-307 says these prompts exist because they "hand the operator something they cannot simply undo afterwards". This runs once against a publisher's production data. Naming both prompts in the option text would close the gap. A separate `--allow-widening` flag would be stronger, since it keeps `--yes` from approving an access change on its own. **Consider:** Either give the access-widening decision its own opt-in flag, or at minimum have the flag's description name both things it answers yes to. As written, the operators told they must pass it for non-interactive runs are exactly the ones who will never see the widening described.
- `plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php:492` — **Suggestion:** A `--live` run exits 0 even when gates failed to write or cannot enforce — The command returns success after a live run that produced ERROR rows (gate creation failed, layouts unwritable) or WARN rows for gates that will not restrict as intended. A sibling command in this same class already takes the opposite line, and says why in its own comment: a warning line is invisible to whatever chained this command with the Memberships deactivation that follows it, so it exits non-zero to stop the cutover there. That reasoning applies here unchanged — an unenforceable or unwritten gate discovered after Memberships has been deactivated is considerably worse than one discovered at the prompt. `WP_CLI::halt( 1 )` when `$layout_errors` or `$unenforceable` is non-zero on a live run would make the two commands consistent. **Consider:** Exiting non-zero when a live run produced error rows or gates flagged as not restricting as intended would match what the exemption command in this same class already does, for the reason its comment gives.
- `plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php:341` — **Nit:** Dead assignment left behind — `$merged_product_ids = $products['product_ids'];` no longer has a consumer — the `compute_pre_write_issues()` parameter it fed was removed in this PR, and `$products` is passed whole to `build_access_rules()` at :397.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Overlap pairs now go into `$to_confirm` alongside the merges, so `--live` cannot write two gates with disjoint product lists unattended — no terminal and no `--yes` hits the hard error `confirm_or_error()` already raises for the supersede case.

Each pair carries both sides' active membership count, summed across the group's plans, so the prompt names the population at risk and not just the gate titles. `group_plans_by_fingerprint()` reads it from `WC_Memberships_Membership_Plan::get_memberships_count( 'active' )` onto the plan descriptor; `format_overlap_warning()` reports it. It is a membership count, not a headcount — a reader holding both plans counts twice — and the docblock says so.

The comment above `$to_confirm` records why this belongs at the prompt rather than the summary: gates do nothing while Memberships is active, so the run that writes the split exits clean and the warning is out of scrollback by the time a reader is turned away.

Pinned by `test_consolidate_plan_groups_hands_unmergeable_overlaps_to_the_caller`.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Split by mode: `--live` still refuses before the first write; a dry run collects every such group and gives each a WARN row from `compute_pre_write_issues()`, so the table prints and the whole set can be fixed in one pass. That closes the docblock drift at :68 and :848.

One reorder came with it: the one-time-duration error now runs before this guard. The guard asks `build_access_rules()` what it would emit (see the other thread), and that returns empty for a one-time group with no duration too, so without the reorder that group would draw this message instead of the more actionable duration one.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Switched the pre-flight to ask `build_access_rules()` what it would emit rather than testing `product_ids`; renamed to `find_paid_groups_without_access_rules()`, which now takes the durations since the builder needs them.

No behaviour change today — `subscription_ids ∪ one_time_ids` is empty exactly when `product_ids` is — but the equivalence is no longer load-bearing: a third classification outcome fails the guard instead of slipping past it. Added a test group holding a product that reaches neither bucket, which the old check passed.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Tie-break is now the rule set's own canonical signature (`rule_set_signature()`: slugs and values sorted), not array position; the mutual-coverage guard uses the same order, so it stays a strict total order and the chain still terminates. Array position now only settles groups whose expanded rule sets are byte-identical, where the pick is arbitrary either way.

Docblock says "the covering group with the most rules", and states that rule count is a proxy for breadth rather than a measure of it.

`test_consolidate_plan_groups_partition_does_not_depend_on_input_order` asserts the same partition from three groups in two orders — a narrow group and two equally sized covering roots that cover neither each other nor the same content. It fails on the old positional tie-break with `Narrow` changing which root it lands behind.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Narrowed to `\WC_Product_Variation` and adjusted the mocks: `WC_Product_Variation extends WC_Product` in `wc-mocks.php`, plus a `wc_mock_product_class()` helper that picks it for `variation` / `subscription_variation`, so `wc_create_mock_product()` and the audit test's `register_product()` stage variations as the right class rather than each fixture remembering to.

A non-variation now fails at the boundary instead of getting a `(variation #12)` label from `wc_get_formatted_variation()`'s empty return. WooCommerce Subscriptions' `WC_Product_Subscription_Variation` extends `WC_Product_Variation`, so both pickers still reach it.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
adekbadek: Took the flush-method path. `Content_Restriction_Control::flush_term_descendants_memo()` mirrors `flush_product_options_memo()`, and `expand_hierarchical_terms()`'s docblock now states the memo is request-scoped with no term-edit invalidation, and points a caller that edits the hierarchy mid-process at the flush. No `edited_term` / `created_term` hook, per your note that it isn't needed for the current consumers.

`Trait_Restriction_Cache_Test` calls it instead of reaching into `$term_descendants_map` by reflection; the other three caches back nothing public, so those stay reflection, and the trait's docblock says which is which.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Both strings branch on `group_requires_purchase()` now. `format_consolidation_warning()` says "one gate carries both plans' access products" and "widens access" only for a purchase fold; a signup fold says one registration gate covers both plans and that no paid access widens. The `$to_confirm` line dropped "paid" and reads "widening access as described above", since the same list can hold folds of either kind.

`test_consolidate_plan_groups_does_not_claim_paid_widening_for_a_signup_fold` folds two signup plans and asserts the warning names neither access products nor paid widening.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
adekbadek: Pulled in `Trait_Restriction_Cache_Test` and reset in `set_up()`, and `test_consolidate_plan_groups_folds_nested_tiers_into_one_group` now builds three real categories through the factory instead of bare 5/6/7. The nesting in that case is in the rule values rather than the term hierarchy, so unrelated sibling terms keep the shape while removing the collision with the sibling case's parent/child pair.

adekbadek:
Exit code: a `--live` run now ends with `WP_CLI::halt( 1 )` when the summary holds an error row or a gate flagged as not restricting as intended, the same shape `migrate-post-exemptions` uses for an unwritten exemption. Dry runs stay at 0 — nothing was written, and surfacing those rows is what the pass is for. The comment at the call site names the reason: whatever chains this command with the Memberships deactivation cannot see a warning line, and the gate's failure is only visible to readers after that deactivation.

---

Dead assignment: `$merged_product_ids` is gone. `$products` is what the write loop reads, and nothing had used the extracted list since the classified result replaced it.

adekbadek:
Both review-body items. The inline threads are answered and resolved above.

**`--yes` and the widening prompt.** The description now names all three prompts (`e9aced8092`), and stops there rather than adding a separate flag. The widening is surfaced twice before anything is written: every warning it raises prints during the dry run, which is the default and the first step of the intended workflow, and again in the prompt. A second flag adds friction to every remote run to guard a case the dry run already covers.

**Exclusion carve-outs.** Not considered, and they do fit. In as of `859bfd87d1`.

They belong on the overlap path, not the merge path: `rules_cover()` decides coverage per slug and never across slugs, so a whole-post-type plan never covers a taxonomy plan. That pair was always an un-mergeable overlap, so the carve-out repairs the case that had no repair and leaves merging untouched.

Where the shape allows, the broad gate is now written as its own content minus the narrow group's, and the narrow gate takes on the broad plan's access products. Verified against Memberships on a running site: a post carrying the narrow plan's term used to match both gates, and `is_post_restricted()` stops at the first gate that denies, so both populations were turned away. It now matches the narrow gate alone, whose product list carries both plans.

Three shapes fall back to the previous behaviour:

- Both sides using one slug. The wizard renders one row per slug and applies an edit to every rule carrying it, so a gate holding both an inclusion and an exclusion of `category` could not be edited afterwards. Nested same-taxonomy tiers stay on the merge path.
- Either side carrying `specific_posts`, an inclusion override evaluated ahead of exclusions.
- A pair crossing the purchase boundary: carving a signup group out of its own gate hands that content to a paid gate, and a reader who reaches it by registering today is denied at cutover.

It is not zero widening, though. The broad plan's holders reach everything the narrow gate covers, including content outside the post types the broad plan gated. Nobody is denied, and the warning says so.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -231,7 +248,7 @@ public function migrate_membership_gates( $args, $assoc_args ) {
 		$durations_by_group = [];
 		foreach ( $plan_groups as $fingerprint => $group ) {
 			$gate_title                         = self::gate_title( $group );
-			$products_by_group[ $fingerprint ]  = self::resolve_product_ids( $group );
+			$products_by_group[ $fingerprint ]  = self::resolve_product_ids( $group, $carved_products[ $fingerprint ] ?? [] );
 			$durations_by_group[ $fingerprint ] = self::resolve_group_duration( $group, $duration_override );
```
chickenn00dle: **Blocker:** A carve-out transfers the product IDs but not the duration, so a one-time entitlement is written nowhere — A carve-out makes one promise to the operator: the narrow gate will take on the broad plan's access products, so nobody loses access. It hands over the product IDs, but not the access length that a one-time product needs in order to become a rule. Where the narrow group has no one-time plan of its own, there is no length to use, and the product is left off the gate entirely.

Walking one case through:

| | |
|---|---|
| Broad plan | gates all posts, grants a **one-time** product |
| Narrow plan | gates one category, grants a **subscription** |
| After the run | broad gate covers all posts *except* that category; narrow gate covers the category and grants the subscription alone |

A reader who bought the one-time product now has no gate that admits them on those posts. The broad gate would let them in, but the carve-out just took those posts off it.

I ran exactly that on the env:

```
gate 59 (broad) rules: post_types[post] + category[18] exclusion
 access: one_time_purchase{ product_ids:[36], forever }
gate 64 (narrow) rules: category[18]
 access: subscription[56] <- product 36 is absent
```

The same run printed "gains ... access products, so neither plan's members are denied" and exited 0.

Where it comes apart: line 251 resolves the narrow group's products *with* the carried-in IDs, line 252 resolves its access length *without* them. `resolve_group_duration()` (`includes/cli/trait-one-time-purchase-migration.php:74`) only ever looks at the group's own plans, so it returns `null` at `:99-105`, and `build_access_rules()` skips the one-time rule at `:1926`.

Two things that would normally catch this don't. The guard at `:263` does spot the condition, then merges `$durations_by_group[...]['plans']` — which is the empty array — so `$needs_duration` stays empty and its `WP_CLI::error()` at `:267` never runs. And `find_paid_groups_without_access_rules()` at `:748` passes, because the narrow group's own subscription rule *is* emitted. `--one-time-duration` doesn't help either: `:99` returns before the override is read at `:106`.

There's a quieter version that needs no subscription at all. When both sides have one-time products of different lengths, the carried-over product silently takes the narrow group's length, and `report_duration_conflict()` never sees the mismatch because it only reads the group's own plans.

**Risk if not addressed:** If the deactivation happens in the same maintenance window as the migration, as the runbook intends, then every holder of the dropped product loses access at cutover with no error anywhere in the run's output to point at. Restoring them means reconstructing which products each plan granted from pre-migration data and hand-editing the gates, because the written gate carries no trace of the entitlement it dropped.

**Consider:** Make the carve-out carry the entitlement, not just the product IDs: either transfer the broad group's duration alongside its one-time products, or refuse the carve-out when the broad group grants a one-time product the narrow group cannot express, and say so in the warning. Separately, have the pre-flight compare what would be written against the classified product set rather than asking only whether the rule list came back non-empty — as it stands a surviving subscription rule masks a dropped one-time rule, which is exactly the case a carve-out creates.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -277,11 +314,41 @@ public function migrate_membership_gates( $args, $assoc_args ) {
...
+				implode( ', ', $carved_out )
+			);
+		}
+		if ( ! empty( $unmergeable_overlaps ) ) {
+			$to_confirm[] = sprintf(
+				'write %s as separate gates with disjoint access products, leaving the members counted above liable to be denied at the other gate once Memberships is deactivated',
+				implode( ', ', $unmergeable_overlaps )
+			);
+		}
+		if ( ! empty( $to_confirm ) && ! $dry_run ) {
```
chickenn00dle, line 405: **Blocker:** Declining the new confirmation exits 0, so a chained cutover proceeds after the operator said no — Answering **no** to the new prompt reports success. Nothing is written, which is right, but the exit code says the run went fine.

That matters because this command is meant to be chained with the step that deactivates WooCommerce Memberships, and the comment at `:540-549` says so: a warning line is invisible to whatever chained the two, which is why a failed live run exits 1 to stop the cutover there. Declining is a stronger stop signal than a warning, and it's the one that reports success.

So `wp newspack migrate-membership-gates --live && wp plugin deactivate woocommerce-memberships` deactivates Memberships after the operator said no, with no gates written at all. Memberships was the only thing restricting the content, so every paywalled article goes public at once, feeds included.

I drove it through a real terminal, since a pipe hits the non-TTY guard instead:

```
$ printf 'n\n' | script -q -c 'wp newspack migrate-membership-gates --live' /dev/null
... Proceed? Answering no stops the run before anything is written. [y/n]
exit=0
```

`confirm_or_error()` already covers the non-TTY case with `WP_CLI::error()` at `:648`, and the docblock at `:601-607` records that `WP_CLI::confirm()` exits with status 0. The deliberate `n` takes that same path. This delta is what put carve-outs and un-mergeable overlaps behind the prompt, so the decline path is newly load-bearing.

**Risk if not addressed:** If the operator answers no inside a chained cutover, then the run writes no gates, reports success, and the next command removes the only thing still restricting the content. Every previously paywalled article becomes public at once, feeds included, and recovery means reactivating the membership plugin and starting over.

**Consider:** Exit non-zero when the operator declines. The live run already halts with 1 after error rows on the stated grounds that a chained cutover cannot see a warning line; a declined prompt is the same situation with a stronger signal behind it.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+			return null;
+		}
+
+		$slugs_a = array_column( $rules_a, 'slug' );
+		$slugs_b = array_column( $rules_b, 'slug' );
+
+		if ( in_array( 'specific_posts', $slugs_a, true ) || in_array( 'specific_posts', $slugs_b, true ) ) {
+			return null;
+		}
+		if ( array_intersect( $slugs_a, $slugs_b ) ) {
```
chickenn00dle, line 1663: **Blocker:** A carve-out will copy a rule the evaluator cannot resolve, and as an exclusion that switches the whole gate off — A gate's rules come in two flavours: **cover this** and **except this**. To decide anything, the evaluator asks one question per rule — does this post match?

For a taxonomy that is no longer registered (its plugin was removed), the answer is always **no**, for every post, because it never gets as far as looking at the post. What that *no* means depends entirely on the flavour:

| Flavour | "No" means | Effect |
|---|---|---|
| cover this | the rule doesn't pull the post in | harmless; the gate's other rules still work |
| except this | the post *is* excepted | skip the gate for this post |

A carve-out copies the narrow plan's rule onto the broad gate and changes its flavour to **except this**. So a rule that was harmless becomes an off-switch: every post answers no, every post reads as excepted, and the broad gate covers nothing.

I confirmed it by A/B on one gate and one post:

```
A. gate 71 WITH the unresolvable exclusion -> no gate applies
B. gate 71 WITHOUT it -> gate 71 applies
```

Gate 71's rules say `post_types[post]`, and in case A it matched no posts at all.

The shape check at `:1406-1417` tests for `specific_posts`, slug disjointness and the `post_types` side, but never asks whether the narrow rules use slugs the evaluator can resolve — even though `is_content_rule_slug_resolvable()` at `:1048` already does that test. `map_rules_to_ac_format()` at `:2198` passes any taxonomy rule through verbatim, and a taxonomy from a deactivated plugin is filtered out nowhere. The evaluator side is `class-content-restriction-control.php:323-325` (the always-false answer) and `:260-266` (reading it as excepted).

It isn't silent: both gates raise a WARN row and the live run exits 1, which I measured. But the warning reads "the content they select stays ungated", which describes a slice going free, not a gate that has stopped working. An operator has no reason to suspect the second from the first — and the gates are already written by then.

**Risk if not addressed:** If that gate is the site's main paywall and its warning is read as a partial gap, then after Memberships is deactivated the entire body of paid content is public — and not only on the site: the feed paths key their truncation off the same restriction check, so full article bodies also ship through RSS to every aggregator subscribed to it.

**Consider:** Add a fourth condition to the shape check: every slug in the narrow rule set must resolve to a post type or taxonomy the evaluator can match. An unresolvable slug behaves differently in the two roles, so it has to be checked before the rules change role. While there, reword the warning so it distinguishes a dead gate from a partial gap — as written it describes the first in the language of the second.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1029,6 +1146,7 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
 			$plan_groups[ $fingerprint ][] = [
 				'pid'               => $pid,
 				'name'              => $plan_name,
+				'member_count'      => $plan->get_memberships_count( 'active' ),
```
chickenn00dle: **Suggestion:** The new membership count runs an unbounded query per plan, for a number only overlap warnings read — The new membership count is a good addition to the prompt — it's what meets the ticket's stated floor. This is about when it's fetched, not whether.

`WC_Memberships_Membership_Plan::get_memberships_count()` isn't a `COUNT()` query. It runs `get_posts( [ 'post_type' => 'wc_user_membership', 'post_parent' => $this->id, 'fields' => 'ids', 'nopaging' => true ] )` and counts the result in PHP. `group_plans_by_fingerprint()` calls it for every plan that survives the skip checks, on both the dry and the live run, while `group_member_count()` only ever reads it for the pairs that reach `format_overlap_warning()` at `:1719-1723`.

On a publisher-scale membership base that's every active membership ID for every plan, materialised into a PHP array, most of them never read.

**Consider:** fetching it inside `format_overlap_warning()` from `pid`, or memoising per plan and resolving only for the groups that end up in the repaired or overlap lists.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+	 * - Exactly one side gates whole post types. That is the only shape where broad and
+	 *   narrow are decidable from the rules alone; anything else needs a post query.
+	 *
+	 * @param array[] $rules_a First rule set, unexpanded.
+	 * @param array[] $rules_b Second rule set, unexpanded.
+	 *
+	 * @return string|null 'a' or 'b' for the side that hosts the exclusion, null when
+	 *                     the overlap cannot be repaired this way.
+	 */
+	private static function carve_out_direction( array $rules_a, array $rules_b ): ?string {
```
chickenn00dle, line 1652: **Suggestion:** Carve-outs fire on pairs whose content may never intersect — A carve-out can fire on two plans that share no content at all, and it still moves one plan's products onto the other's gate.

Coverage is decided per slug and never compared across slugs, so a whole-post-type group and a taxonomy group are always reported as overlapping — whether or not a single post carries both. That was fine while the outcome was a warning. The detector's own docblock at `:1284` says as much: in the undecidable case a false positive costs a warning, while a miss costs a reader their access.

The delta points that same detector at a write. A false positive now costs a permanent transfer of one plan's access products onto another plan's gate, plus an exclusion on the first. The reasoning that justified erring towards reporting doesn't carry over to the new use. And this is the common setup, not a rare one: any site-wide plan alongside any section plan qualifies.

**Consider:** saying in the confirmation text that the pair is a *suspected* overlap rather than a demonstrated one, and naming how many posts each carve-out actually affects — a count of posts matching the narrow rules within the broad group's post types. The operator is the only one who can tell a real repair from a no-op, and only if the prompt tells them which it might be. The same number would make the over-grant warning concrete rather than hypothetical.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+	 * @param array   $carved_products Filled with the products each narrow group gains, keyed by
+	 *                                 its index in $groups, by reference.
+	 *
+	 * @return array{groups: array[], overlaps: array} The groups with carve-outs applied, and the
+	 *               overlap pairs that remain unrepaired.
+	 */
+	private static function apply_carve_outs( array $groups, array $overlaps, array &$carved, array &$carved_products ): array {
+		$surviving = [];
+
+		foreach ( $overlaps as $pair ) {
```
chickenn00dle, line 1736: **Suggestion:** One gate can host only one carve-out per slug, and which pair wins depends on plan ID order — A gate can host only one carve-out per slug, so in the common publisher setup most overlaps go unrepaired — and which one gets repaired depends on the order plans were created.

Once the broad gate has taken a `category` exclusion, the slug-disjointness check at `:1409` refuses every later narrow group that also uses `category`. Take one "All Posts" plan alongside three category plans: exactly one pair is repaired, and the other two fall back to un-mergeable overlaps with their members still liable to be denied.

The winner is decided by root index, which comes from `get_plans()` ordering by ID at `:2126-2139`. So a plan published or retired between the approved dry run and the live run can change which reader population gets the repair, and the dry run stops predicting what the live run will write.

This is the same non-determinism you deliberately removed from the absorption tie-break with `root_precedes()` at `:1361`.

**Consider:** ordering the overlap pairs by the same canonical signature, or preferring the pair with the larger `group_member_count()` now that the count is available. Either way, saying in the warning that only the first collision on a slug is repairable would stop the summary reading as though every overlap was handled.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+			list( $i, $j ) = $pair;
+
+			$direction = self::group_requires_purchase( $groups[ $i ] ) === self::group_requires_purchase( $groups[ $j ] )
```
chickenn00dle: **Suggestion:** Two registration-only plans get carved out of each other, with copy describing a paid stake they cannot have — Two free-registration plans can be carved out of each other, and there's nothing for the carve-out to repair. Both gates admit any logged-in reader, so neither can lock the other's members out — the overlap was harmless to begin with. And because a signup plan has no products (`group_plans_by_fingerprint()` sets `product_ids => []` at `:1152`), `$broad_products` at `:1461-1463` is empty, so nothing transfers either.

What the operator sees is a prompt asking them to confirm a rewrite, and a warning telling them the narrow gate "gains ... access products" and that members will reach content outside the broad plan's post types. Neither can happen for this pair.

I confirmed it: both plans flipped to free signup, the carve-out still applied, both summary rows read Access Type `signup`, and the warning still named access products.

The check at `:1450` only asks whether the two purchase flags **match**, so signup-plus-signup passes it. This is the same shape as the copy fix you already made for `format_consolidation_warning()` at `:1692`.

**Consider:** requiring `group_requires_purchase()` on at least one side, rather than only that the flags agree. That also avoids writing an exclusion rule someone has to maintain later. One caveat if you'd rather keep signup pairs: on an existing gate matched by title, metering and priority survive the run, so changing which gate covers the shared content can still change which layout and allowance a reader meets.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+			}
+
+			$broad  = 'a' === $direction ? $i : $j;
+			$narrow = 'a' === $direction ? $j : $i;
+
+			$broad_products = array_values(
+				array_unique( array_map( 'intval', array_merge( [], ...array_column( $groups[ $broad ], 'product_ids' ) ) ) )
+			);
+
+			foreach ( $groups[ $narrow ][0]['ac_rules'] as $rule ) {
```
chickenn00dle, line 1762: **Suggestion:** A carve-out makes the narrow gate the sole guard of the carved content, and nothing verifies it will hold it — A carve-out takes content that had two gates over it down to one. If that remaining gate doesn't get written properly, the content ends up behind nothing.

Once the exclusion is on the broad gate, the evaluator skips it for every post in the narrow set (`class-content-restriction-control.php:287-289`), so the narrow gate is the only thing left holding that content. Two reachable states leave it unable to:

- `create_gate()` fails at `:400` and the loop `continue`s to the next group with an ERROR row.
- A purchase group with no extractable layout never reaches `:427`, so paid mode is never switched on — which `compute_pre_write_issues()` describes as "any registered reader would get in" (`:991`).

Either way, content the broad paid gate was holding a moment earlier is now behind free registration, or behind nothing.

The ordering makes it likelier to land badly: the write loop runs in ascending root-index order, so whenever the broad root sorts first its exclusion is written before the narrow gate is even attempted. The live run reaches `WP_CLI::halt( 1 )` at `:548` only after the whole loop, so the half-written state is on disk regardless.

**Consider:** writing the narrow gate before the broad gate's exclusion, or dropping the exclusion when its counterpart couldn't be written, so a failed gate creation is a failed repair rather than a hole in coverage.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
```diff
@@ -2311,4 +2301,648 @@ public function test_a_whole_taxonomy_rule_still_maps_to_a_value_the_reader_disc
...
+				[
+					'slug'  => $taxonomy,
+					'value' => array_map( 'strval', $term_ids ),
+				],
+			],
+			$product_id
+		);
+	}
+
+	/**
```
chickenn00dle, line 2777: **Suggestion:** The new carve-out tests pin the transfer, not the entitlement — The new tests check that the products are handed over, but not that the gate ends up granting them — which is the step where they're lost.

`test_consolidate_plan_groups_repairs_a_cross_slug_overlap_with_a_carve_out` asserts the exclusion rule's shape and `$carved_products === [ 1 => [ 101 ] ]`. The other three assert `carve_out_direction()`'s return value directly. None of them calls `resolve_product_ids()`, `resolve_group_duration()` or `build_access_rules()` on a carved group, so nothing exercises the claim in `carve_out_direction()`'s own docblock that "both populations then reach what Memberships gave them".

That's the whole point of the feature, and the first blocker above passes every one of these tests.

**Consider:** a case that carries a carve-out through to the access rules — a purchase broad group on a one-time product and a subscription-only narrow group — asserting the narrow gate emits a rule naming the broad plan's product. It fails today. Worth knowing that `make_plan_group()` at `:2633` never sets a `one_time_duration` key, so the length path can't be reached from any carve-out test until that changes.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -101,7 +106,7 @@ class Membership_Gates_Migration {
 	 * : How long access lasts after a one-time purchase. Accepts "forever", "<n>days" or "<n>months". Each plan's own access length is read by default, so this is only needed for a plan whose access ends on a fixed calendar date, which is not a duration from the purchase. Applies to every such plan in the run.
 	 *
 	 * [--yes]
```
chickenn00dle, line 121: **Nit:** Operator-facing accounting doesn't acknowledge the carve-out prompt or the new rule kind — Two places where the operator-facing text hasn't caught up with the carve-out.

The `[--yes]` synopsis lists three things the prompt can be about: superseding gates, consolidation, and gates written over overlapping content. The class docblock at `:89-96` names two. The carve-out confirmation added at `:335-340` is a fourth, and it's a different kind of thing from the others — the only prompt that rewrites a gate's content rules and moves access products between gates. It also isn't the "written over overlapping content" case at `:341-346`, which is the *unrepaired* branch. Since `--yes` is the flag operators are told they need for non-interactive runs, it's worth naming what it now waives.

Separately, the summary's Content Rules column is `count( $ac_rules )` at `:473`, which now counts appended exclusions. A broad gate that gates one post type reads as 2. Splitting it, say `2 (+1 carve-out)`, would keep the row honest — that table is the artefact the operator reviews before cutover.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1039,6 +1157,602 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+	}
+
+	/**
+	 * List a gate group's writable product IDs for an operator-facing message.
+	 *
+	 * @param array[] $group A gate group.
+	 *
+	 * @return string The comma-separated IDs, or 'none'.
+	 */
+	private static function describe_group_products( array $group ): string {
```
chickenn00dle: **Nit:** The overlap warning reports a stale product list for a group that was carved in elsewhere — The overlap warning shows a group's products as they were before any carve-out, so the operator weighs the remaining risk against a list that's already out of date.

`describe_group_products()` calls `resolve_product_ids( $group )` with no carried-in IDs, and `format_overlap_warning()` at `:1717` prints the result. A group can be the narrow side of one repaired pair and still party to a second, unrepaired overlap — reachable through the slug collision described above — and in that case the printed list is missing exactly the IDs this run just added to it.

Passing the group's `$carved_products` entry through, the way line 251 does, would keep the two consistent.

chickenn00dle (changes requested):
Release risk: HIGH — three blockers, all reproduced on a running site, in new code that rewrites reader entitlement against production data.

Everything from the last pass is genuinely addressed, and I re-checked the substantive ones rather than taking the replies at face value: the absorption tie-break is now a canonical rule-set signature, `flush_term_descendants_memo()` mirrors the flush added beside it, the pre-flight asks the rule builder what it would emit instead of testing the product list, a live run exits non-zero, and the signup-fold copy branches correctly. The overlap prompt now names each side's active membership count, which is what the ticket asked for.

The carve-out is the right idea — it answers the question I left open last time, and its three shape constraints are each correctly justified. I confirmed `specific_posts` really is an inclusion override evaluated ahead of exclusions, that an exclusion short-circuits regardless of match mode, and that the wizard really does render one row per slug and apply an edit to every rule carrying it. On the content-rule axis the repair is symmetric and removes the denial it targets. It is the product axis and the failure modes around it that are not ready.

Three blockers (inline) all share a shape worth naming: this command's failures are invisible at the moment it runs, because gates stand down entirely while Memberships is active. The run that writes a broken gate and the moment readers feel it are separated by the deactivation step, so anything the run does not catch, nobody catches.

The rest are smaller — signup pairs carved for no benefit with copy describing a paid stake they cannot have, only one carve-out per slug per gate with which pair wins decided by plan ID order, and a new test that pins the transfer rather than the entitlement, which is why the first blocker passes every test in the file.

- `plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php:1847` — **Nit:** Dropped-product warnings are attributed to the wrong gate after a carve-out — A dropped-product warning can name the wrong gate, and point the operator at a plan that never had the product. `resolve_product_ids()` merges the carried-in IDs into `$raw` at `:1797` and classifies them alongside the group's own, and those IDs arrive unvalidated from `:1461-1463`. So an invalid or deleted product belonging to the *broad* plan is reported under the narrow gate's title, with "Check the plan's products" — where it isn't. Validating the carried-in list once on the broad side, or naming the source gate in the message, would fix it.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e — the carve-out now carries the entitlement, not just the IDs.

`apply_carve_outs()` records the broad group's own `one_time_ids` and derived duration alongside its products, and `resolve_group_duration()` takes a third `$carried` argument: a handed-over one-time product votes on the length like a plan of the group's own. Your case gets `forever` from the broad plan, and where both sides have one-time products of different lengths the longest-wins pick and its conflict warning now see both.

The pre-flight went structural, as you suggested. `find_groups_with_unwritten_products()` replaces the `$needs_duration` block, comparing the products the access rules would name against `product_ids` — so a surviving subscription rule no longer masks a dropped one-time rule, and the empty-`['plans']` merge that kept the guard from firing is gone with it.

`test_a_carve_out_carries_the_access_length_its_products_need` drives a carve-out through to `build_access_rules()` and asserts both rule groups; it fails against the previous `resolve_group_duration()`.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e. `confirm_or_error()` no longer routes through `WP_CLI::confirm()` — it prints the question, reads the answer, and calls `WP_CLI::error()` on anything but `y`, so a declined run exits 1 and a chained `&& wp plugin deactivate woocommerce-memberships` stops there. The message says plainly that no gate is in place and not to deactivate Memberships.

The answer is injected the same way the terminal check already was, since STDIN is never a terminal under PHPUnit; `test_confirm_or_error_aborts_when_the_operator_declines` pins it. `--yes` returns before anything is read, and the non-TTY error is unchanged.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e.

`carve_out_direction()` gains a fourth condition: every slug in the narrow rule set must pass `is_content_rule_slug_resolvable()`. It runs after the direction is decided, so it tests the side that actually becomes the exclusion — the slug is harmless in the role it holds today and fatal in the one the carve-out gives it.

The warning is reworded, and the duplicated block in `verify_migrated_gate()` and `compute_pre_write_issues()` is now one `describe_unresolvable_rules()` that branches on the rule's role: an unresolvable exclusion reports "an exclusion nothing resolves excludes every post — the gate is carved out of all of its own content and restricts nothing at all". This command can no longer write that shape, but a hand-edited gate or one left by an earlier run can still arrive in it.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Done in fa1ec3e — the confirmation and the warning both say *suspected*, and both carry the count.

`count_carved_posts()` queries published posts of the broad group's post types carrying one of the narrow group's terms, and the number goes into the carve-out line the prompt is built from (`"Premium Section" out of "All Posts" (1 post(s))`) and into the warning, which now reads "are suspected of gating overlapping content — N published post(s) match both today". A zero tells the operator the carve-out is a no-op; they are still the only one who can say whether it stays zero.

You are right that the detector's reasoning does not carry over to a write. Rather than tighten the detector — which would need the post query on every pair — the carve-out is now the narrower thing: both sides must require a purchase, and the count says which case the operator is looking at.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e, taking both halves.

Overlap pairs are now sorted before any carve-out is applied: descending total `group_member_count()`, then on a canonical `overlap_signature()` of the two rule sets. So the repair goes to the largest population rather than to whichever plan `get_plans()` returned first, and a plan published between the dry run and the live run no longer moves it. That is the same reasoning as `root_precedes()`, which is why the tie-break is the signature rather than the index.

And the summary no longer reads as though every overlap was handled: `apply_carve_outs()` keeps the rules as they arrived, so a pair refused only because the broad gate had already taken that slug is distinguishable from one that was never repairable, and its warning says "a gate can hold only one carve-out per content-rule slug and the pair with the larger membership took it".

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e — `apply_carve_outs()` now requires `group_requires_purchase()` on both sides rather than only that the flags agree, so a signup/signup pair gets no exclusion written.

One thing I went further on, and want a second opinion: such a pair is no longer added to the un-mergeable-overlap list either, so it is announced but never reaches the confirmation prompt — the prompt line would ask the operator to approve members "liable to be denied" who cannot exist. It gets its own warning carrying your caveat: no reader is denied at either gate, but a post matching both meets whichever gate has priority, and that gate's layout and metering allowance are the ones the reader gets. Say the word if you would rather it stayed on the prompt.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e.

The write loop now runs carved-out groups first, and each group records whether its gate holds: written without error, and for a purchase group with paid mode active carrying a rule. That is your two states — a failed `create_gate()`, and a purchase group with no extractable `custom_access` layout. The test is structural and identical in dry-run and live, so the dry run predicts what the live run writes.

`drop_unheld_carve_outs()` strips the carve-out's exclusion rules from the broad group before its gate is written, warning that the repaired overlap is back and both gates need reviewing by hand. A failed gate creation is now a failed repair, and the content stays behind the gate that was holding it.

The summary table is keyed by group position and sorted back before printing, so the reordering is not visible in it.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-membership-gates-migration.php
adekbadek: Added in fa1ec3e. `test_a_carve_out_carries_the_access_length_its_products_need` runs a purchase broad group granting a one-time product with a `forever` length and a subscription-only narrow group through `consolidate_plan_groups()`, `resolve_product_ids()`, `resolve_group_duration()` and `build_access_rules()`, asserting the carved-out gate emits both rule groups — its own subscription, and the broad plan's product for as long as that plan granted it. `make_plan_group()` takes a `$one_time_duration` now, so the length path is reachable from a carve-out test.

Three more alongside it: `find_groups_with_unwritten_products()` seeing past a surviving subscription rule, `carve_out_direction()` refusing an unresolvable narrow slug, and `drop_unheld_carve_outs()` restoring the broad gate's coverage. All four fail against the code as it stood.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Both done in fa1ec3e.

The `[--yes]` synopsis and the class docblock now name the carve-out as its own case, and say it is the one prompt that rewrites a gate's content rules and moves access products between gates — separately from the unrepaired-overlap case. The synopsis also records that declining exits non-zero.

The Content Rules column is `describe_rule_count()`, which splits the exclusions out: a broad gate gating one post type with one carve-out reads `1 (+1 carve-out)`.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in fa1ec3e by validating once on the broad side, which also settles the dropped-product warning naming the wrong gate.

`apply_carve_outs()` transfers `resolve_product_ids( $broad )['product_ids']` rather than the raw `_product_ids` meta, so a carried-in ID has already passed validation under its owning gate and cannot be dropped a second time under the gate receiving it. `describe_group_products()` now takes the carried-in list and passes it through, and `format_overlap_warning()` supplies it from the carve-out records, so a group that is the narrow side of one repaired pair and party to a second, unrepaired one is quoted with the products its gate will carry.

adekbadek:
Fixed in fa1ec3e, by validating the carried-in list once on the broad side.

`apply_carve_outs()` now transfers `resolve_product_ids( $broad )['product_ids']` rather than the raw `_product_ids` meta, so what crosses to the narrow gate has already been through validation under the gate that owns it. An invalid or deleted ID is dropped and reported once, under the plan it belongs to, and can never surface a second time under the gate it was handed to.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Fixed in `326b4adee7`. Grouping no longer resolves the count. `plan_member_count()` reads it from the descriptor's `pid` on first use and caches it per plan for the life of the run, so only plans that reach an overlap warning cost a query.

Memoising rather than fetching inside `format_overlap_warning()`, because that is not the only reader: the `usort` that orders overlap pairs by population needs the counts before any warning is built, so the warning-local fetch would have left the sort paying the old cost.

A descriptor that already carries `member_count` is taken at its word, so a caller holding the number does not pay for it twice. A test covers both halves — one lookup however often the count is read, and no lookup at all when it is supplied.

adekbadek:
- fix(cli): carry entitlements through carve-outs (NPPD-2064, #642)
- perf(cli): count plan members only when read (NPPD-2064, #642)
- test(network): stop content distribution tests fetching an image (#642)

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
```diff
@@ -1160,6 +1409,899 @@ private static function group_plans_by_fingerprint( array $plan_ids, array &$ski
...
+		}
+
+		$plan_id = (int) ( $plan['pid'] ?? 0 );
+		if ( ! $plan_id || ! function_exists( 'wc_memberships_get_membership_plan' ) ) {
+			return 0;
+		}
+
+		if ( ! array_key_exists( $plan_id, self::$member_counts ) ) {
+			$membership_plan                 = \wc_memberships_get_membership_plan( $plan_id );
+			self::$member_counts[ $plan_id ] = $membership_plan ? (int) $membership_plan->get_memberships_count( 'active' ) : 0;
```
chickenn00dle: **Suggestion:** The member count leaves out comped, trial and pending members, and still loads every ID — `get_memberships_count( 'active' )` counts only `wcm-active`, but Memberships grants access on `active`, `complimentary`, `free_trial` and `pending`. Now that this count decides which pair is carved, and the overlap warning quotes it as the population at risk, it should count every access-granting status. Consider `wc_memberships()->get_user_memberships_instance()->get_active_access_membership_statuses()` with a paged query (`posts_per_page => 1`, read `found_posts`), which also settles the unbounded query from the last round.

<details><summary>Evidence</summary>

In the installed Memberships, `WC_Memberships_Membership_Plan::get_memberships_count()` runs `get_posts()` with `'fields' => 'ids', 'nopaging' => true` and counts in PHP (`class-wc-memberships-membership-plan.php:1901-1937`). The memo caps that at once per plan, but the carve-out sort at :1729 reads the count for both sides of every overlap pair, so on a site-wide-plan-plus-sections catalogue every plan is loaded anyway, including the site-wide plan, the likeliest to be large. The sort also runs before the both-sides-paid filter, so plans in signup/signup pairs are counted for a pair that can never be carved. The `'active'` argument predates this round; what changed is that it now feeds a decision, not only a warning.

</details>

plugins/newspack-plugin/includes/cli/trait-one-time-purchase-migration.php
```diff
@@ -95,6 +103,13 @@ private static function resolve_group_duration( array $group, ?array $override )
 			$plans[]     = $plan['name'];
 			$durations[] = $plan['one_time_duration'] ?? null;
 		}
+		foreach ( $carried as $source ) {
```
chickenn00dle, line 117: **Suggestion:** A carried-in one-time length extends the narrow plan's own one-time buyers — When the carried-in length and the narrow group's own one-time length differ, longest-wins gives both sets of buyers the longer one, so the narrow plan's 7-day buyers get the broad plan's 30 days (or `forever`) on the carved content. Consider emitting one `one_time_purchase` rule group per distinct length; `build_access_rules()` already ORs its groups, so each set of buyers would keep what its own plan granted. It's an over-grant, not a denial, and the run warns.

On the review env, a broad plan granting product 149 for 30 days carved against a narrow plan granting product 36 for 7 days wrote `one_time_purchase{product_ids:[149,36], 30, days}` on the narrow gate, with the warning "30 days and 7 days, so the gate keeps the longest". That warning describes a disagreement between the group's plans, but the broad plan isn't one of them; the carve-out created the conflict.

chickenn00dle (approved):
This is good to go! Two non-blocking bits of feedback. Feel free to merge without addressing them (or even if they are addressed).

Review scope: 2 of 3 reviewer passes ran; skipped: codex.

plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php
adekbadek: Done in 1cb58fd.

`plan_member_count()` now runs its own `WP_Query` with `posts_per_page => 1` and reads `found_posts`, over the statuses `get_active_access_membership_statuses()` returns, `wcm-` prefixed, falling back to the shipped four for a site whose memberships outlived the plugin. Comped, free-trial and pending members count, and no IDs are pulled into PHP.

`apply_carve_outs()` partitions on both-sides-paid before sorting, so only carve-out candidates are ordered by member count; a signup pair goes straight to the surviving list with `blocked_by_carve_out` already false.

Two tests, both failing against the old reading: the four granting statuses count while `expired`/`cancelled` do not, and the count resolves once per plan however often it is read. `wc-memberships-plan-lookup-mock.php` is gone — nothing else used it.

plugins/newspack-plugin/includes/cli/trait-one-time-purchase-migration.php
adekbadek: Done in 1cb58fd, one rule group per length.

`resolve_group_duration()` now returns `durations_by_product`, and `build_access_rules()` buckets the one-time IDs by it, so your example writes `one_time_purchase{[149], 30 days}` OR `one_time_purchase{[36], 7 days}`. The carve-out record carries the broad gate's per-product lengths rather than just its longest, so a broad gate whose own plans disagree survives the transfer. That also fixes the same over-grant between a group's own plans, which longest-wins had all along. `--one-time-duration` still flattens to the operator's value.

Nothing is resolved silently now, so the duration warning is a plain line saying the gate carries one rule per length — several purchase rules on one gate otherwise read as a mistake.

`resolve_group_duration()` still returns `conflict` for `migrate-premium-newsletters`, which writes a single rule and keeps the latent over-grant. Out of scope here; happy to take it as a follow-up.

PR #645, opened 2026-07-16
fix(cli): make plan consolidation hierarchy-aware in migrate-membership-gates (NPPD-2066)

adekbadek:
### All Submissions:

Closes NPPD-2066.

**Stacked PR.** This sits on the `migrate-membership-gates` gates stack: #642 → #639 → #635 → #626. Base is `nppd-2064-migration-tooling-migrate-membership-gates-splits`; a sibling PR (NPPD-2066's stack-mate, NPPD-2065) shares the same base. Bases retarget automatically as the stack merges down.

**Why.** NPPD-2064's consolidation compares plan rule sets by flat term-ID set containment (`rules_cover` / `array_diff`). Taxonomy hierarchy was invisible: a plan restricting a child category and a plan restricting its parent share no term value, so they neither consolidated (neither is a value-subset of the other) nor tripped the non-subset overlap warning (their term lists do not intersect). The result was two separate gates with disjoint product lists — and gate priority could then deny a reader entitled via the narrower plan. It was the one overlap shape that produced a completely silent split.

**Verified enforcement semantics (both sides cascade parent → child).** The fix rests on an empirically verified parity, not an assumption:

* Access Control expands a parent-term rule to its descendants at read time via `Content_Restriction_Control::expand_hierarchical_terms()`.
* WooCommerce Memberships does the same in its vendor rule matcher via `get_term_children()` + `has_term()`.

Both were proven live: a post assigned only to a child category is gated by a rule targeting the parent, under AC (Memberships inactive) and under WCM (Memberships active). So `parent ⊃ child` is a true coverage relation, and hierarchy-nested plans should consolidate exactly like any other subset.

**What.** A thin adapter helper, `expand_rule_set_hierarchy()`, expands each hierarchical-taxonomy rule value to include its descendant term IDs *before* the pure subset/overlap helpers run — so the parent rule set visibly covers the child's (child group absorbed) and a hierarchy overlap that is not a clean subset trips the loud warning instead of vanishing. Key invariants:

* The expanded values feed the consolidation **decision only**. The surviving gate is written with the plan's **original** (unexpanded) parent-term rule, which enforcement re-expands at read time — so a child term added after the migration stays covered.
* The pure decision helpers (`rules_cover`, `rule_sets_overlap`, `plan_rule_set_consolidation`) stay WooCommerce/WordPress-free; only the adapter reads the term tree.
* Hierarchy expansion can canonicalise two distinct fingerprints to the same term closure (a parent rule vs. a redundant parent-plus-child rule). Those groups mutually cover, which is not a strict subset; `plan_rule_set_consolidation()` now folds each equal-closure equivalence class into its lowest-index member via an asymmetric `$j < $i` tie-break (proven acyclic and deterministic), so they still share one gate instead of splitting with a spurious warning.

1. Run the unit suites: `n test-php --filter Test_Membership_Gates_Migration` (28 tests) and `n test-php --filter Test_Content_Gates` (70 tests) — both green. `phpcs` clean.
2. On a site with WooCommerce Memberships active, create two purchase plans with distinct subscription products: one restricting a parent category, one restricting a child of that parent. Run `wp newspack migrate-membership-gates` (dry-run), then `--live`.
3. Expect one consolidated gate (not two): the child plan is absorbed into the parent, the gate carries the parent-category rule, and its paid-access list is the **union** of both plans' products.
4. Confirm the reader outcome: a subscriber holding only the child plan's product stays entitled to a child-only post via the consolidated gate (checked through AC enforcement), while anonymous readers are gated.
5. Equal-closure case: a plan on a parent category plus a plan on that parent-and-a-redundant-child consolidate into one gate with no spurious overlap warning.

### Other information:

https://claude.ai/code/session_01LDoog2tvzcfgrtFG7LmNkh


adekbadek:
Superseded by #642, which absorbed this fix while it was reworked. Hierarchy expansion now lives there as `expand_rule_set_for_coverage()`, calling `Content_Restriction_Control::expand_hierarchical_terms()` rather than re-walking the term tree, with the equal-coverage tie-break as `root_precedes()` and coverage in `test_consolidate_plan_groups_absorbs_a_child_category_into_its_parent`.

Closing rather than rebasing: these commits patch helpers that no longer exist in that shape on the base branch.

PR #647, opened 2026-07-16
chore: add Claude security guidance for model-backed reviews

adekbadek:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? — n/a, no PHP/JS in this PR.

Adds `.claude/claude-security-guidance.md`, the extension point read by the [security-guidance plugin](https://code.claude.com/docs/en/security-guidance)'s model-backed reviews (end-of-turn diff review and commit/push review). It is loaded as extra context alongside the plugin's built-in vulnerability checklist.

**Note this PR also enables the plugin for everyone**, via `enabledPlugins` in the checked-in `.claude/settings.json`. That is deliberate — user-scoped plugins don't carry into Claude Code on the web or to other clones — but it is a team-wide change worth a conscious ack rather than a rubber stamp. Drop that hunk if we'd rather everyone opt in individually.

The guidance is grounded in what this repo actually does, not generic WordPress advice:

* **Tells the reviewer what *not* to report.** The root `phpcs.xml` applies WordPress-Extra + WordPress-VIP-Go repo-wide with no security sniffs excluded, and husky blocks commits on failure. Escaping/nonce/`prepare()` findings are already enforced, so re-reporting them is noise. Attention is redirected to logic-level issues PHPCS structurally cannot see, and to `phpcs:ignore` lines as the places a machine stopped and a human assumption started.
* **Encodes non-obvious, high-consequence invariants**, each verified against the code: the `wp` cookie-prefix Batcache exemption (the cache key holds no reader identity, so that prefix is the only thing stopping one reader's gated content reaching another); the HMAC / `hash_equals` / validate-before-DB ordering in `content-gate`; the three legitimate shapes of `permission_callback => '__return_true'`; ESP/OAuth secrets sitting unencrypted in `wp_options` behind an allowlist; and the `pull_request_target` "DATA ONLY" invariant already documented in `dependabot-sync-overrides.yml`.
* **Documents the `newspack-network` trust model as an explicit exemption**, so reviewers stop re-raising settled ground: every Node is assumed to be operated by the same operator, a successful AEAD decrypt *is* the authorization, and a compromised node is accepted risk. Account propagation and role propagation (including `administrator`, via the manual "Sync user across network" button) are called out as by-design, per the `User_Manually_Synced` docblock.

Guidance is **additive only** — per the plugin docs, a rule saying to ignore a vulnerability class does not suppress those findings. The exemptions are therefore written as threat-model context, which informs severity, rather than as "don't report X". This is noise reduction, not enforcement.

Size: 8059 bytes, under the plugin's 8192-byte cap. That cap applies to the user, project, and `.local` files **concatenated** and truncates from the tail, so there is ~130 bytes of headroom to spend.

1. Confirm the file is discovered and delivered complete (not truncated) by the plugin's own loader:
   ```sh
   python3 -c "
   import sys; sys.path.insert(0, '$HOME/.claude/plugins/marketplaces/claude-plugins-official/plugins/security-guidance/hooks')
   import extensibility as e
   e.load_for_session('$PWD')
   b = e.guidance_block()
   print(len(b), 'bytes | intact:', b.strip().endswith('</project-security-guidance>'))
   "
   ```
   Expect a non-zero size and `intact: True`.
2. Sanity-check the byte cap: `wc -c .claude/claude-security-guidance.md` → under 8192.
3. Read the file and check the claims against the code they cite — e.g. the `Newsletters_Access::COOKIE_NAME` docblock for the `wp`-prefix rule, and the `User_Manually_Synced` docblock for the network trust model.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — n/a, config/docs only.
* [x] Have you successfully run tests with your changes locally? — verified via the plugin's `load_for_session()` loader as in step 1; no test suite applies.


.claude/claude-security-guidance.md
```diff
@@ -0,0 +1,142 @@
...
+  anyone (see `Salesforce::api_validate_webhook`).
+
+## Network hub/node sync
+
+**Trust model – read before reporting anything in `newspack-network`.** The model assumes
+**every Node is operated by the same operator** (see the `User_Manually_Synced` docblock).
+Auth is symmetric AEAD: a successful decrypt proves possession of the shared key, and that
+is the entire authorization. A compromised node is therefore **accepted risk, out of
+scope** – an attacker holding a node's key is a bigger incident than the sync layer can
+defend against. Attacker-controlled payloads are a premise, not a finding.
```
chickenn00dle: **Severity:** suggestion

“Attacker-controlled payloads are a premise, not a finding” is unscoped. Possession of the shared AEAD key (a successful `sodium_crypto_aead_xchacha20poly1305_ietf_decrypt`) is an *authorization* fact — it says nothing about whether the decrypted bytes are safe to hand to a sink. The network code does route payload data into real sinks: `class-incoming-post.php` passes payload `thumbnail_url` into `media_sideload_image()` (a server-side fetch of an attacker-named URL), inserts payload `post_content`, and writes the whole payload to postmeta. As written, a reviewer seeing a new payload field flow into `$wpdb->query()`, `unserialize()`, or a file path would conclude “premise, not a finding” and stay silent — the opposite of the intent one line up (which is about *authorization*: a compromised node being accepted risk). The plugin loader's wrapper counteracts explicit “ignore class X” phrasing, but this sentence reframes the threat model rather than naming a class, so the wrapper may not catch it. Suggest scoping it: attacker-controlled payloads are a premise for *authorization* findings, but payload *handling* (SQL, deserialization, file paths, HTML/JS sinks, server-side fetches) stays in scope.

.claude/claude-security-guidance.md
```diff
@@ -0,0 +1,142 @@
+# Security guidance for the Newspack monorepo
+
+WordPress plugins/themes for publisher news sites. Crown jewels: **reader identity,
+subscriber PII, paywall integrity, ESP/OAuth credentials**.
+
+## Do not re-flag what the linter already catches
+
+A single root `phpcs.xml` applies `WordPress-Extra` + `WordPress-Docs` +
```
chickenn00dle: **Severity:** suggestion

“applies … across the repo” overstates the coverage, and the gap lands on the file this doc calls highest-consequence. The root `phpcs.xml` scopes scanning to `<file>plugins</file>`, `<file>themes</file>`, `<file>packages</file>` only — root `bin/` (including `bin/advanced-cache.php`), `config/`, and everything under `repos/` are never sniffed, and CI reinforces this by running phpcs per changed workspace package. So line 67 calls `bin/advanced-cache.php` the “highest-consequence area in the repo” while this section tells the reviewer to assume `EscapeOutput`/`ValidatedSanitizedInput`/`PreparedSQL*` are already enforced there — they aren't. Suggest narrowing to “across `plugins/`, `themes/`, `packages/`” and carving out `bin/`, `config/`, `repos/` as unlinted. (The “no security sniffs excluded” premise itself checks out — every `<exclude>` in phpcs.xml is stylistic.)

.claude/claude-security-guidance.md
```diff
@@ -0,0 +1,142 @@
...
+`WordPress-VIP-Go` across the repo and blocks commits via a husky hook (warnings block
+for PHP). Assume these are enforced; do **not** report them as findings:
```
chickenn00dle: **Severity:** suggestion

“Assume these are enforced; do not report them as findings” is stronger than the hook guarantees, on three counts. (a) The husky hook lints *staged* files only. (b) The repo explicitly sanctions bypassing it — `AGENTS.md` documents `--no-verify`/`HUSKY=0` “for agents/automation that must commit without linting,” the exact audience this file configures. (c) Most importantly, PHPCS sniffs are syntactic heuristics: `EscapeOutput` passes on `esc_attr()` used in a URL/JS context, on escaping the wrong variable, and on custom sanitizers — so “lint passed” is not “proven safe.” CI does re-lint (the real gate) but inherits the same blind spots. Suggest “deprioritize; report only when you can point at a specific bypass or a wrong-context escape” rather than a blanket “do not report.”

.claude/claude-security-guidance.md
```diff
@@ -0,0 +1,142 @@
...
+  inject raw `<script>` anyway.)
+- **Block `render_callback`s** (`src/blocks/*/view.php`) build HTML/shortcodes with
+  `sprintf`/concatenation. Escape at output (`esc_attr` on `className`) and validate
+  attributes into a local var – `block.json` types are not a security boundary. Blocks
+  delegating to a template are audited at the template.
+- **Frontend JS:** `.innerHTML =` with server or reader-supplied values (content-gate
+  banners, popups merge tags rendering reader profile data) needs the value `wp_kses`'d
+  server-side or set via `textContent`.
+
+## CI workflows
```
chickenn00dle: **Severity:** suggestion

Structural fragility worth fixing before merge: the file is 8029 bytes, ~160 under the plugin's 8192 cap, and the cap applies to the user + project + `.local` guidance files *concatenated*, truncating *from the tail*. Any developer with even a two-line `~/.claude/claude-security-guidance.md` silently amputates the tail of this file — which is the entire `## CI workflows` section, including the `pull_request_target` “DATA ONLY” rule (the classic privileged-checkout RCE, mirrored from a real SECURITY INVARIANT block in `dependabot-sync-overrides.yml`) and SHA-pinning. That's arguably the highest-severity content here, and it vanishes with no warning. Suggest moving `## CI workflows` above a lower-stakes section (e.g. `## Injection the linter misses`) so truncation degrades gracefully, and/or trimming to buy real headroom. (Verified: user guidance concatenates *before* project guidance, so the project file's tail is what truncation eats first.)

.claude/claude-security-guidance.md
```diff
@@ -0,0 +1,142 @@
...
+restricting `role` – become real requirements.
+
+## Injection the linter misses
+
+- **Inline `<script>` JSON is not a `</script>` breakout – do not report one.**
+  `json_encode` escapes `/` as `\/` by default, so an encoded `</script>` emits
+  `<\/script>` and cannot close the element. Bare `echo wp_json_encode( $x )` in a script
+  block is fine; `JSON_HEX_TAG` does not fix a bug that isn't there. (`<!--<script>` does
+  derail the tokenizer, but storing it needs `unfiltered_html` – Editors/Admins, who can
+  inject raw `<script>` anyway.)
```
chickenn00dle: **Severity:** suggestion

The primary claim is correct — `wp_json_encode()` defaults to escaping `/`, so `echo wp_json_encode( $x )` in a `<script>` block can't be broken out via `</script>`, and no caller here passes `JSON_UNESCAPED_SLASHES`. But the parenthetical dismissal — “storing it needs `unfiltered_html` – Editors/Admins, who can inject raw `<script>` anyway” — holds only for the *post-content* (kses) vector. It doesn't hold for the vectors this doc names a few lines down: popups merge tags rendering *reader* profile data, and content-gate banners. A reader's profile field never traverses `unfiltered_html`; if such a value reaches an inline JSON blob, `<!--<script>` derails the tokenizer with no elevated capability. Suggest qualifying the parenthetical to the post-content path so it doesn't license silence on reader-sourced data in `<script>` blocks.

chickenn00dle (commented):
Release risk: LOW. This is a well-grounded guidance doc — I verified the load-bearing claims (Batcache `wp`-prefix logic in `advanced-cache.php`, the content-gate HMAC/`hash_equals` ordering, `PROVIDER_CREDENTIAL_ALLOWLIST`, the Salesforce fail-closed check, the `wp_json_encode` slash-escaping, and the `dependabot-sync-overrides.yml` DATA-ONLY invariant) and they hold. The enablement hunk is honest: `security-guidance` ships from the same official marketplace as the already-enabled plugins, so no new supply-chain surface.

The comments below are refinements, not blockers. They cluster on one theme: a few suppression phrases (“assume these are enforced; do not report”, “Attacker-controlled payloads are a premise, not a finding”, “do not report one”) are written broader than the evidence supports, each in a spot where the underlying enforcement doesn't fully reach. The plugin's own loader wraps this file with an explicit “must NOT suppress findings” instruction, which blunts the risk — but tightening the scope in-text is worth doing so a reviewer doesn't internalize a blind spot the wrapper can't catch. The truncation-boundary point (the file sits ~160 bytes under the concatenation cap, and the CI section that would vanish first is arguably its highest-stakes content) is the one I'd fix before merge.

Minor: the PR description says 8059 bytes; `wc -c` reports 8029 — harmless, and in the safe direction.

.claude/claude-security-guidance.md
adekbadek: Scoped the exemption to authorization only. The AEAD-decrypt premise now reads as excusing only *who* sent the bytes, never *what they do at the sink*: a payload field reaching SQL, `unserialize()`, a file path, an HTML/JS sink, or a server-side fetch is explicitly back in scope, with `class-incoming-post.php` sideloading payload `thumbnail_url` via `media_sideload_image()` as the named example. So a new payload field flowing into a sink is no longer waved through as 'premise, not a finding'.

.claude/claude-security-guidance.md
adekbadek: Narrowed the coverage claim: phpcs now scans `plugins/`, `themes/`, `packages/` **only**, and `bin/` (incl. `bin/advanced-cache.php`), `config/`, and `repos/` are called out as never sniffed. The doc no longer tells the reviewer to assume EscapeOutput/ValidatedSanitizedInput/PreparedSQL* are enforced on the file it later names highest-consequence.

.claude/claude-security-guidance.md
adekbadek: Softened the blanket 'assume enforced; do not report' to **deprioritize** – report one only with a concrete bypass, an unlinted path, or a wrong-context escape. Added the two reasons you named: the sniffs are syntactic heuristics (they pass on wrong-context escaping, the wrong variable, a custom sanitizer) and the pre-commit hook is staged-only and bypassable. 'Lint passed' no longer reads as 'proven safe'.

.claude/claude-security-guidance.md
adekbadek: Qualified the parenthetical to the post-content (kses) path only. Reader-sourced values (profile fields, content-gate values) never traverse `unfiltered_html`, so the doc now states that `<!--<script>` in an inline JSON blob carrying reader data **is** a real finding and should get `JSON_HEX_TAG`. The note no longer licenses silence on the reader-sourced vectors it names a few lines down.

.claude/claude-security-guidance.md
adekbadek: Fixed both parts. (1) Moved `## CI workflows` above `## Injection the linter misses`, so the section truncation eats first is now the lowest-stakes one – the `pull_request_target`/DATA-ONLY rule and SHA-pinning survive. This matches your suggested order.

(2) Confirmed the cap against the plugin's loader (`extensibility.py`): it's `combined[:8192]` where `len()` counts **characters**, not bytes. The four content fixes in the sibling threads had grown the file to 8581 combined chars – 389 *over* budget – so it was truncating even standalone. Compressed the prose repo-wide (dropping no rule, code reference, or sibling fix) and cut one illustrative example (the `Content_Gifting` parenthetical, whose rule stands without it), landing at **8147 combined = 45 under the cap**. So with no user-level guidance file nothing truncates; with one, the reordered tail degrades gracefully.

One honest caveat: the doc now sits intrinsically near the 8 KB budget, so *generous* headroom for a large `~/.claude/claude-security-guidance.md` would require dropping a rule – I've left that as an author call rather than cut content unilaterally.

adekbadek:
- chore: fit security guidance under the review prompt budget

chickenn00dle (approved):

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/chore/claude-security-guidance` (stash local work first).

PR #648, opened 2026-07-16
fix: robustness tweaks in data events, user admin, and dev scripts

adekbadek:
### All Submissions:

A handful of small, self-contained robustness improvements spotted while reading through existing code. No new features and no changes to public APIs. Each commit stands alone and can be dropped independently if preferred.

**Data Events** (`includes/data-events/`)

- Send the dispatch nonce in the request body rather than as a query parameter. The receiving handler reads it from `$_REQUEST`, so this needs no change on the handler side and the dispatch behaves identically. It keeps the value out of web server access logs, which record query strings verbatim.
- Make `verify_nonce()` fail closed. The stored nonce is created lazily on the first dispatch, so before then it reads back as an empty string and a loose comparison would treat an empty input as a match. It now rejects empty/non-string input, ignores an empty stored value, and compares with `hash_equals()`. The current-nonce and previous-nonce grace period paths are unchanged in behaviour otherwise.

**Nicename change UI** (`includes/plugins/co-authors-plus/`, behind `NEWSPACK_CHANGE_NICENAME_UI`, off by default)

- The change handler takes a `user_id` from the request and passes it to `wp_update_user()`. It now confirms the caller may edit that specific user (`current_user_can( 'edit_user', $user_id )`) instead of relying on the nonce alone, and the availability-check endpoint (which enumerates existing users and guest authors) is gated on `edit_users`. In practice the UI is only reachable by administrators, so this is a defence-in-depth check rather than a behaviour change for any current user.

**Jurassic Ninja dev scripts** (`bin/jn-*`, local tooling only, not shipped)

- Quote variable expansions. A password containing a space previously mangled the command line.
- Pass the ssh password through the environment (`sshpass -e`) rather than argv.
- Document why `StrictHostKeyChecking=no` is used here (recycled throwaway subdomains legitimately present new host keys).

1. Run the test suites — new coverage accompanies each change:
   - `cd plugins/newspack-plugin && n test-php --group data-events`
   - `n test-php --group nonce-verification`
   - `n test-php --group dispatch-nonce`
   - `n test-php --group nicename-change`
   - Full plugin suite also passes: `n test-php`
2. On a local env, trigger any data event, e.g. `wp eval 'Newspack\Data_Events::dispatch( "donation_new", [ "user_id" => 1 ] );'`, and confirm it is still handled as before (the registered handler runs). The corresponding `admin-ajax.php?action=newspack_data_event` line in the access log no longer carries a nonce parameter.
3. With `NEWSPACK_CHANGE_NICENAME_UI` defined, open a user in wp-admin, change the archive slug as an administrator, and confirm it still updates.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
+++ b/plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
@@ -136,9 +136,13 @@ class Nicename_Change_UI {
 			);
 		}
 
-		// The nonce only proves the request came from the user-edit screen, not that the
-		// caller may edit this particular user -- authorize the target before touching it.
-		if ( ! current_user_can( 'edit_user', $user_id ) ) {
+		// The nonce only proves the request came from the user-edit screen. Require the same
+		// primitive that gates the availability check this handler may delegate to on a slug
+		// collision (so an authorized caller never gets 403'd mid-flow), and object-level
+		// authorization for the target so it can't be used to edit an arbitrary user. The
+		// primitive check also closes the self-edit gap where map_meta_cap grants
+		// `edit_user` on one's own ID to any logged-in user.
+		if ( ! current_user_can( 'edit_users' ) || ! current_user_can( 'edit_user', $user_id ) ) {
 			wp_send_json(
 				[
 					'success' => false,
--- a/plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
+++ b/plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
@@ -83,6 +83,18 @@ class Nicename_Change_UI {
 
 		check_ajax_referer( 'newspack_change_nicename_nonce', 'nonce' );
 
+		// The response enumerates existing users and guest authors, so it is gated on the
+		// same capability that gates the UI this endpoint serves (the user-edit screen).
+		if ( ! current_user_can( 'edit_users' ) ) {
+			wp_send_json(
+				[
+					'success' => false,
+					'message' => esc_html__( 'You are not allowed to do that.', 'newspack-plugin' ),
+				],
+				403
+			);
+		}
+
 		$new_nicename = isset( $_POST['new_nicename'] ) ? sanitize_title( wp_unslash( $_POST['new_nicename'] ) ) : '';
 
 		$existing = Nicename_Change::get_existing_nicenames( $new_nicename );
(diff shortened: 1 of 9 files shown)


plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
```diff
@@ -132,6 +136,26 @@ public static function change_nicename_ajax() {
 			);
 		}
 
+		// The nonce only proves the request came from the user-edit screen, not that the
+		// caller may edit this particular user -- authorize the target before touching it.
+		if ( ! current_user_can( 'edit_user', $user_id ) ) {
```
chickenn00dle: **Severity:** suggestion

`current_user_can( 'edit_user', $user_id )` is the right object-level check for editing *another* user, but WordPress maps `edit_user` so any logged-in user passes it for their **own** ID. So a non-admin could still POST here with their own `user_id` and change their own archive slug — while the availability-check endpoint (which requires the primitive `edit_users`, line 88) would reject that same caller. There's also a latent inconsistency on the own-profile path: if the requested slug collides, the handler delegates to `change_nicename_check_ajax()`, whose `edit_users` gate then 403s the very self-edit this handler just authorized. This is still a large net tightening over the prior state (no capability check at all), and the UI/nonce is only printed on `user-edit.php` (admin-only) behind `NEWSPACK_CHANGE_NICENAME_UI` (off by default), so it's not a regression. Worth either using `edit_users` on both endpoints for parity, or consciously accepting the self-edit surface.

chickenn00dle (approved):
Release risk: LOW. A clean, well-scoped defensive PR — three self-contained hardening tweaks, each a no-op on the normal path, with unusually thorough test coverage. I exercised the data-events and nicename-change paths at runtime (PHPUnit groups plus a direct WP-CLI check of `verify_nonce`): the fail-closed nonce rejects empty/non-string input while still accepting a valid current or in-grace-period nonce, the nonce moves transparently into the POST body (the receiver reads `$_REQUEST`), and the capability checks block a subscriber/editor from changing another user's slug while still letting an admin through. No blockers.

One thing worth a conscious decision before merge: `edit_user` permits self-edits, so a non-admin could still change their own slug through the change endpoint even though the availability-check endpoint requires `edit_users` (details inline). It's a net tightening over the prior no-check state and the UI is admin-only and flag-gated, so it's not a regression — but the two endpoints guarding on different capabilities is worth a look. A few optional nits too (a jQuery `.html()` sink fed by slug-constrained data, an unwrapped i18n string, and a missing trailing newline in a dev script) — all non-blocking.

plugins/newspack-plugin/includes/plugins/co-authors-plus/class-nicename-change-ui.php
adekbadek: Good catch. Took the parity option: the change endpoint now requires the `edit_users` primitive (same as the availability check it delegates to on a slug collision), plus keeps the object-level `edit_user` check for multisite defense-in-depth. This closes the self-edit gap (map_meta_cap grants `edit_user` on one's own ID to any logged-in user) and removes the internal inconsistency you noted, since an authorized caller now passes both this gate and the delegated `change_nicename_check_ajax()` gate. Self-edit isn't a legitimate path here anyway: the UI/nonce is only printed on `user-edit.php` (editing *another* user; self-profile uses `profile.php`).

PR #649, opened 2026-07-16
chore: make existing TypeScript strict-clean and enable the type gate (phase 0.5)

adekbadek:
### All Submissions:

Second PR in the TypeScript migration stack. Clears the pre-existing strict-mode errors in the units that already had TypeScript, then switches the `typescript:check` gate on for them. No file renames.

- `newspack-plugin`: cleared the strict-mode baseline (jest globals, the content-gates ambient type family, and real strict errors across settings / webhooks / theme-and-brand), then enabled the gate.
- `newspack-blocks`: cleared strict errors in the donate and homepage-articles editors (`withSelect`/`withDispatch` generics) and enabled the gate.
- `packages/components`: made the DataViews wrapper generic and fixed color-picker JSDoc.

**Behavior changes**, called out because they are not type-only:

- `fix(donate)`: the minimum donation amount now applies to frequency-based editor amounts.
- `fix(donate)`: a failed donation-settings fetch renders the error Placeholder instead of crashing the block. `.catch( setError )` stored the REST error object in a string state, so the `error.length` guard never fired and the render fell through to an unloaded settings object.
- Pricing rules: a `compose_mode` value the rule form does not recognise now round-trips on save instead of being rewritten to `min`.

From here on, every `.js → .ts` rename permanently adds that file to the strict gate.

Type-stripping is behavior-neutral: the Babel build removes types with no runtime effect, so these sweeps are typing work rather than logic changes. Any behavior-affecting change is called out explicitly above.

2. `pnpm --filter <unit> run typescript:check` — the strict gate passes for every unit touched here.
3. `pnpm --filter <unit> run lint:js` and `pnpm --filter <unit> run build` — both green.
4. `pnpm --filter <unit> run test` for units carrying JS tests.
5. Compare the built `dist/` file list against the base branch — unchanged.

To check the donate error path, deactivate WooCommerce and add a Donate block to a post. The block shows an error Placeholder naming the missing plugin.

### Other information:

### Stack

Part of the [NPPM-2977](https://linear.app/a8c/issue/NPPM-2977/monorepo-jstypescript-migration) JS→TypeScript migration — a chain of stacked PRs, each targeting the previous phase's branch:

| Phase | PR | Scope |
|---|---|---|
| phase0 | #526 | Infra: type gate, tsconfigs, webpack ext-agnostic entry resolution |
| phase0.5 | #649 | Make the existing TS strict-clean, turn the gate on |
| phase1 | #650 | `packages/icons` + `packages/components` |
| phase2 | #651 | `newspack-plugin` |
| phase3 | #652 | `newspack-blocks` + `newspack-newsletters` |
| phase4 | #653 | Remaining plugins |
| phase5 | #654 | Both themes |
| phase6 | #655 | Wrap-up: tighten remaining `any`, blame-ignore revs |

Each PR will be retargeted as the chain lands. **Do not merge out of order.**


adekbadek:
> Re: [this review comment](https://github.com/Automattic/newspack-workspace/pull/649#issuecomment-5037869200)

Already done, and the conflicts are gone.

`main` was merged in manually at 7186498 (2026-07-21 21:20 UTC, ~2.5h after this comment), with a follow-up in ae6c00ae making the newly merged audience views strict-clean so the type gate this PR enables stays green. The gate passes locally on both gated units:

```
plugins/newspack-plugin  → newspack-scripts typescript-check → All good!
plugins/newspack-blocks  → newspack-scripts typescript-check → All good!
```

`main` has since moved another 14 commits ahead, but that drift is conflict-free (`git merge-tree --write-tree HEAD origin/main` produces a clean tree) and contains **zero** `.ts`/`.tsx` changes, so re-syncing cannot introduce new strict-mode errors — the next automatic sync should go through untouched. I only have read-only git here, so I have not landed that second merge myself; flagging it rather than papering over it.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/chore/ts-migration-phase0.5` (stash local work first).

chickenn00dle (commented):
**Release risk: ELEVATED** — no blockers. The type work is clean and I'd merge this; the two things worth attention are around the edges rather than in the added lines.

First, what holds up. I went through all 14 added optional-chains and all 6 added casts individually, since "make the type checker happy" is exactly where a change stops being behaviour-preserving — and they're benign. Two are improvements: the `?? ''` in the Mailchimp connection previously fell back to `{}` into `sprintf`, which would have rendered "Connected as [object Object]". Nothing was suppressed to buy green — no `@ts-ignore`, no non-null assertions, and one pre-existing `@ts-expect-error` actually removed. The shared-package changes don't break consumers either: `Wizard`'s removed props were dead code, `Page.children` and `ColorPicker.help` were widened to optional rather than narrowed, and the `onChange` narrowing from `Partial<T>` to `T` matches what both handlers actually pass.

The gate itself is real, not aspirational — CI runs `typescript:check` per changed package against configs extending `strict: true`, and `packages/components` was never exempt (it already had the script, and is additionally pulled into this plugin's own check via its `include`).

**A behavioural fix is riding under a `chore:` title.** Commit `c9eebc83b` moves `amounts={ displayedAmounts }` after the `{ ...componentProps }` spread, which flips which value wins — the clamped copy now beats the raw one, so the frequency-based editor preview shows minimum-clamped suggested amounts where it previously showed sub-minimum ones. That's a legitimate fix, and it makes `FrequencyBasedLayout` consistent with `TierBasedLayout`, which already had the correct order. But the commit's own subject is `fix(donate):`, and squash-merging replaces it with this PR's `chore:` title — so semantic-release cuts no patch for newspack-blocks and the change ships silently. Worth either splitting it out or retitling this PR.

**Turning the gate on has a queue behind it.** `tsconfig`'s `include` is the whole `src` tree rather than an allowlist, so every `.ts`/`.tsx` in both packages enters scope at once — including files this PR never touched. There are around a dozen open PRs touching gated paths that have never run this check, and at least one file merged to main after this branch forked has never been type-checked at all. Nothing is at risk beyond time, but a failure will read as "my change broke something" rather than "a new check was switched on". Merging latest main in and re-running the check once more before landing would catch the drift, and a heads-up to those authors would make any redness legible.

One note on the donate initial-state change, because it's easy to over-read. Dropping `amounts: {}` and `disabledFrequencies: {}` does mean those keys are `undefined` until the settings fetch resolves, and the comment added alongside — that nothing reads them before the `isLoading` guard — isn't quite right: `.finally()` clears `isLoading` on the *rejected* path too, so the guard is already passed and the reads are reachable. But this isn't a regression: on `main` that same failure path already threw, just one component deeper in `FrequencyBasedLayout`. What the change really does is let the type system vouch for two keys being present when they aren't. The underlying bug is upstream of all of it — `.catch( setError )` hands a `WP_Error` object to a `useState('')` setter, so the `error.length` guard never fires and the error Placeholder never renders. Extracting `.message` there would fix the actual problem and make the guard honest.

The remaining points are inline and small.

plugins/newspack-blocks/src/blocks/donate/edit/index.tsx
```diff
@@ -433,8 +433,8 @@ const Edit = ( { attributes, setAttributes }: EditProps ) => {
 					<FrequencyBasedLayout
 						isTiered={ isTiered }
 						canUseNameYourPrice={ canUseNameYourPrice }
-						amounts={ displayedAmounts }
 						{ ...componentProps }
```
chickenn00dle, line 436: **Suggestion:** A `fix(donate)` commit is bundled into a `chore:` PR and won't be released — Moving `amounts={ displayedAmounts }` after the `{ ...componentProps }` spread flips which value wins. `componentProps.amounts` is the raw set; `displayedAmounts` is the minimum-clamped copy — and they genuinely differ whenever a suggested amount falls below the minimum, which the block itself warns about elsewhere. So the frequency-based editor preview now shows clamped figures in its tier labels and radio values, where it previously showed sub-minimum ones.

This is a good change: `TierBasedLayout` already had the reordered form, so it makes the two layouts consistent. The frontend is untouched (`view.php` renders separately), so the effect is confined to the editor — but the editor is a publisher-facing surface.

The issue is release mechanics. The commit's own subject is `fix(donate): apply minimum donation to frequency-based editor amounts`, but squash-merging replaces it with this PR's `chore:` title — so semantic-release sees no `fix`, cuts no patch for newspack-blocks, and the change ships silently with whatever release comes next. `git blame` on that line will show only the chore title too.

Either land it as its own `fix(donate):` PR, or retitle this one so the release reflects it.

plugins/newspack-blocks/src/blocks/homepage-articles/utils.ts
```diff
@@ -207,11 +206,15 @@ type Select = ( namespace: string ) => {
  * wordpress/data selector for blocks using this custom store.
  */
 export const postsBlockSelector = (
```
chickenn00dle, line 208: **Suggestion:** Looseness moved from the function body to its public signature — Removing the `@ts-expect-error` is a genuine improvement. But the exported `postsBlockSelector` / `postsBlockDispatch` traded typed parameters for `select: unknown` and `Record<string, unknown>` plus internal casts — so the imprecision didn't go away, it moved from one line inside the body to the public signature. `postsBlockSelector( anything, {} )` now type-checks from anywhere in the codebase.

If the driver is `withSelect`'s loose `mapSelect` typing, typing the parameters to that signature — or keeping the precise types and casting only at the HOC call site in `edit.tsx` — keeps both the suppression gone and the contract intact.

plugins/newspack-blocks/src/types/newspack-components.d.ts
```diff
@@ -1,6 +1,6 @@
 declare module 'newspack-components' {
 	function useObjectState<StateObject>(
-		object: StateObject
+		object: Partial<StateObject>
```
chickenn00dle, line 3: **Suggestion:** The widened `useObjectState` decl makes the type system vouch for keys that are absent — The ambient stub now accepts `Partial<StateObject>` while still *returning* `StateObject`. That's unsound — the real hook is just `useState( initial )`, so the returned object contains exactly the keys passed in. The donate editor immediately relies on the unsoundness: `amounts: {}` and `disabledFrequencies: {}` were dropped from the initial state, so `settings.amounts` is typed `DonationAmounts` while being `undefined` until the fetch resolves.

The comment added alongside — that nothing reads them before the `isLoading` guard — holds on the happy path but not on the rejected one: `.catch( setError )` stores the rejection and `.finally()` clears `isLoading` regardless, so the guard is already passed and `! settings.disabledFrequencies[ slug ]` is reachable.

Worth being clear that this is **not a regression** — on `main` that same failure path already threw, one component deeper at `FrequencyBasedLayout.tsx:106` (`amounts[ frequencySlug ][ 3 ]` on an empty object). Publisher-observable behaviour is unchanged. What changed is the throw site and the strength of the type lie, which is the opposite of what a strict-mode PR wants to leave behind.

Two cheap fixes, ideally both: make those keys optional on the state type (or keep the `{}` seeds), and fix the real bug — `.catch( e => setError( e.message ) )`, so the error Placeholder actually renders instead of falling through to a crash. The comment should go either way, since as written it asserts an invariant that doesn't hold.

plugins/newspack-plugin/package.json
```diff
@@ -27,7 +27,8 @@
 		"lint:scss:staged": "newspack-scripts wp-scripts lint-style --customSyntax postcss-scss",
 		"format:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss --fix",
 		"release": "npm run build && npm run semantic-release",
-		"release:archive": "rm -rf release && mkdir -p release && rsync -r . ./release/newspack-plugin --exclude-from='./.distignore' && cd release && zip -r newspack-plugin.zip newspack-plugin"
+		"release:archive": "rm -rf release && mkdir -p release && rsync -r . ./release/newspack-plugin --exclude-from='./.distignore' && cd release && zip -r newspack-plugin.zip newspack-plugin",
+		"typescript:check": "newspack-scripts typescript-check"
```
chickenn00dle, line 31: **Suggestion:** Enabling the gate exposes work already in flight — This is the change with the widest reach, and it's worth sequencing deliberately.

`tsconfig`'s `include` covers the whole `src` tree rather than an allowlist of migrated files, so the moment these scripts exist every `.ts`/`.tsx` in both packages is type-checked — including files this PR never touched. CI's step uses `--if-present`, which means it has been silently skipping these two packages until now, so there's no prior green result to compare against.

Concretely: roughly a dozen open PRs currently touch `.ts`/`.tsx` under these two packages and have never run this check, and at least one file added to `main` after this branch forked has never been type-checked at all. Any strict-mode violation in that set surfaces on the next CI run of work unrelated to it.

Nothing is at risk beyond time — the fix is a type annotation and a re-run. But the failure will read as "my change broke something" rather than "a new check was switched on". Merging latest `main` into this branch and re-running the check once more before landing would catch the drift, and a heads-up to the affected authors makes any redness legible.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/types/index.d.ts
```diff
@@ -172,6 +181,38 @@ type GateConfig = {
...
+	label: string;
+	value: number;
+};
+
+/**
+ * Data localized by the Audience wizard (`Audience_Wizard`) as
+ * `newspackAudience`. Only the slice consumed by the content-gates screens is
+ * typed here; the full window-level shape lives in wizards/types/window.d.ts.
+ */
+declare const newspackAudience:
```
chickenn00dle, line 204: **Suggestion:** A third, disjoint description of the same global — The new `declare const newspackAudience` resolves the bare-global type error, but it's now the third mutually inconsistent description of one runtime object: `wizards/types/window.d.ts` already declares `Window['newspackAudience']` as non-optional with a different field set, and `institutions/index.tsx` reaches the same object through `( window as any ).newspackAudience?.institutional_access_url` — a key that appears in neither declaration.

All three compile, which is precisely the problem: same object, three type views, free to drift. This plugin's own conventions call for `declare global { interface Window { … } }` for window globals. Folding the new fields into the existing `window.d.ts` entry (and dropping the `as any` while there) gives one source of truth — and would have surfaced the missing `institutional_access_url` for free.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -94,9 +94,13 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
 	const [ scopeType, setScopeType ] = useState( rule?.scope_type ?? vocab.scopes[ 0 ]?.id ?? 'all_products' );
 	const [ scopeIds, setScopeIds ] = useState< number[] >( rule?.scope_ids ?? [] );
 	const [ priority, setPriority ] = useState( String( rule?.priority ?? 100 ) );
-	const [ composeMode, setComposeMode ] = useState( rule?.compose_mode ?? 'min' );
+	const [ composeMode, setComposeMode ] = useState< 'min' | 'priority_exclusive' >(
```
chickenn00dle, line 97: **Suggestion:** `compose_mode` fallback became lossy — `useState( rule?.compose_mode ?? 'min' )` became a ternary that resolves anything other than `'priority_exclusive'` to `'min'`. The declared type is `'min' | 'priority_exclusive' | string` — that open `| string` looks deliberate, i.e. the schema anticipates a third value.

`composeMode` is never surfaced as a form control; it's read in and written straight back out on save. So opening and saving a rule that carries an unrecognised `compose_mode` now silently rewrites it to `'min'`, where it previously round-tripped untouched.

The sibling `cycle_anchor` and `application` states already used this ternary shape before this PR, so this may just be consistency. But if the union really is closed, the honest fix is to close it in the type and drop the `| string`, rather than narrowing at one read site — the server side isn't in this monorepo, so the open type is currently the only statement of intent.

packages/components/src/divider/index.js
```diff
@@ -12,6 +12,17 @@ import './style.scss';
  */
 import classNames from 'classnames';
 
+/**
```
chickenn00dle, line 15: **Nit:** New JSDoc narrows a shared component that still forwards everything — Adding `@param` entries gives `Divider` a concrete props type where it previously had implicit `any`. A TypeScript consumer passing anything outside the five documented props now fails excess-property checking — even though the component genuinely spreads `...otherProps` onto the element.

No consumer breaks today: every `.tsx` call site passes only documented props, and the one that doesn't (`newspack-story-budget`, passing `orientation` and `noshade`) is a `.js` file in a unit with no TypeScript and `checkJs: false`. But that unit is on the gate list, so the first `.tsx` file added there would surface it. Documenting the rest as `@param {Object} [otherProps]` — or typing it as `React.HTMLAttributes<HTMLDivElement>` — keeps the forwarding contract intact.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates-onboarding.tsx
```diff
@@ -15,10 +15,14 @@ import { envelope, postList, settings } from '@wordpress/icons';
 import { Card, Grid, SectionHeader } from '../../../../../packages/components/src';
 import { contentLocked, emailPremium } from '../../../../../packages/icons';
 
+// Rendered as `start`/`end` DOM attributes on the stack element, which the
+// Grid stylesheet targets via attribute selectors for column placement.
+const gridPlacementProps = { start: 2, end: 4 };
```
chickenn00dle, line 20: **Nit:** `gridPlacementProps` launders the error rather than resolving it — `<VStack start={2} end={4}>` fails excess-property checking because `VStack` doesn't declare those props. Hoisting them into a const and spreading works only because TypeScript skips excess-property checks on spreads of non-fresh object literals.

Runtime is byte-identical, and the accompanying comment correctly documents that these land as DOM attributes the Grid stylesheet selects on — so nothing breaks. But the consequence is that the type gate will never look at these two props again, including if the Grid contract changes. Since the mechanism is a genuine DOM-attribute passthrough, saying so in the type — `React.ComponentProps<typeof VStack> & { start?: number; end?: number }`, or a small typed helper — would express the intent instead of routing around the check.

plugins/newspack-plugin/src/wizards/newspack/types/index.d.ts
```diff
@@ -88,6 +89,12 @@ declare global {
 			'advanced-settings': WizardTab;
 			collections: WizardTab;
 			privacy: WizardTab;
+			// Present only when experimental tools are registered; consumers guard with `in` checks and optional chaining.
+			'experimental-tools': WizardTab & {
```
chickenn00dle, line 93: **Nit:** Declarations asserting more than the code guarantees — Three small ones, all the same shape — a type that states the opposite of what the surrounding code knows:

- `'experimental-tools'` is declared as a **required** key on the settings tabs while its own comment says it's present only when those tools are registered. It happens to be safe (the lazy import is gated behind an `in` check), but `?:` plus that guard would be truthful.
- `ContentGatesWizardData` is now declared twice with divergent shapes — an ambient global and a module-local in `advanced-settings.tsx` carrying an extra field. Deleting the local would silently resolve to the global and lose that field.
- `edit/index.tsx` still carries `gates = null as unknown as Gate[]` — the exact double-cast that `content-gate-settings.tsx` was cleaned up to remove *in this same PR*. Worth finishing, since `null` and `undefined` are equally fatal at the three unguarded `gates.map` / `.find` / spread sites.

chickenn00dle (approved):

plugins/newspack-blocks/src/blocks/homepage-articles/utils.ts
adekbadek: Restored the precise parameter types on both exports — `postsBlockSelector( select: Select, { clientId, attributes } )` and a new `Dispatch` type mirroring `Select` for `postsBlockDispatch`, so the internal casts are gone too. The looseness now lives only at the `compose()` call site in `edit.tsx`, with a comment saying why. Verified in the editor: Homepage Posts renders its posts, so `withSelect`/`withDispatch` still wire up. In bd0d0bf.

plugins/newspack-blocks/src/types/newspack-components.d.ts
adekbadek: Reverted the ambient stub to `object: StateObject`, so it no longer vouches for absent keys. Keeping literal `{}` seeds wasn't available even under `Partial` — `{}` still isn't a `DonationAmounts` — so the initial state is now a complete `EditState` (zeroed amounts, all frequencies enabled), which makes the declaration true and every read defined regardless of what the fetch does.

The comment is gone, and `.catch` now extracts `.message`, so the invariant it asserted is no longer needed: the error Placeholder renders and the settings reads are genuinely unreachable on that path.

In bd0d0bf.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/types/index.d.ts
adekbadek: Folded into the existing `Window['newspackAudience']` entry in `wizards/types/window.d.ts` and deleted the ambient `declare const`. `institutional_access_url` came along, so `institutions/index.tsx` reads it through the typed global and the `as any` is gone; `content-gifting.tsx` and `countdown-banner.tsx` moved from the bare global to `window.` and dropped their `/* global */` pragmas.

Worth knowing for later: three PHP sites localize this name with disjoint payloads (`Audience_Wizard`, `Audience_Content_Gates`, `Premium_Newsletters_Wizard`), one per screen. On the Access Control screen the object carries only `available_products`, `institutional_access_url` and `content_gifting` — so the pre-existing non-optional fields are optimistic there. I left those as they are rather than widen the change; the new keys are optional. In bd0d0bf.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
adekbadek: Went the other way from closing the union: the state is now `useState< PricingRuleRow['compose_mode'] >( rule?.compose_mode ?? 'min' )`, so an unrecognised value is held and written back untouched. Closing the type would delete the only statement of intent we have about a server this monorepo doesn't contain, whereas preserving the value costs nothing and removes the silent rewrite either way.

One correction to the premise: `composeMode` does have a control — a two-option `SelectControl`, gated behind `recipe?.isCustom`. That's what forced the closed union, since the control infers its value type from its options; the options are now annotated to the open type instead. Its `onChange` still only ever produces the two known modes. In bd0d0bf.

packages/components/src/divider/index.js
adekbadek: Typed the props as `import('react').ComponentPropsWithoutRef<'hr'>` intersected with the five documented options, which states the forwarding contract rather than narrowing it. Checked it is live: a temporary `<Divider id="probe" bogusProp={1} />` in a `.tsx` fails on `bogusProp` only — `id` passes.

The `newspack-story-budget` call site turns out to be a different component: `__experimentalDivider` from `@wordpress/components`, not this one, so `orientation`/`noshade` never reach it. In bd0d0bf.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates-onboarding.tsx
adekbadek: Replaced the hoisted object with a typed wrapper — `GridPlacedVStack: React.FC< React.ComponentProps< typeof VStack > & { start?: number; end?: number } >` — so the props are declared and checked at the JSX, matching the widening pattern already used for `CheckboxControl` in the webhooks modal. The props are back inline as `start={ 2 } end={ 4 }`.

Confirmed on the Access Control onboarding screen that the DOM is unchanged: the stack element still carries `start="2"` / `end="4"` and computes `grid-column-start: 2`. In bd0d0bf.

plugins/newspack-plugin/src/wizards/newspack/types/index.d.ts
adekbadek: All three done in bd0d0bf.

`'experimental-tools'` is now `?:`. That made `settings/sections.tsx` honest about its own lookup, so the reduce hoists the tab into a local and skips a missing one instead of repeating the index five times.

The module-local `ContentGatesWizardData` is deleted; `has_newsletters` moved onto `GateSettings`, which is where the gates endpoint actually returns it — next to the `has_institutions` flag from the same payload.

`edit/index.tsx` now destructures `gates` plainly, so it is `Gate[] | undefined`; the guards at the `.find` and `Array.isArray` sites already covered it, and the ref seeds with `gates ?? []`.

adekbadek:
Fixed the `.catch( setError )` root cause in bd0d0bf: it now takes `.message` off the rejection, with a fallback string, so the `error.length` guard fires and the Placeholder renders.

Reproduced it both ways in the review env with WooCommerce absent, which makes that endpoint reject. On this PR's head the Donate block shows "This block has encountered an error and cannot be previewed."; with the fix it shows the error Placeholder reading "The WooCommerce plugin is not installed and activated…" plus the settings link.

PR #651, opened 2026-07-16
chore: migrate newspack-plugin to TypeScript (phase 2)

adekbadek:
### All Submissions:

Fourth PR in the TypeScript migration stack, and the highest-risk one — `newspack-plugin` is the foundation the rest of the product builds on, and it carries the reader-facing surfaces.

- `newspack-plugin/src` fully converted; the unit is then locked with `allowJs: false`, so a stray new `.js` fails the type check.
- `packages/components`: extends the shared Reader Activation contract and widens some prop types to match real call sites.

Verified behavior-neutral: the built `dist/` file list is identical to the base branch, the full jest suite passes, and the high-risk bundles (reader-activation, content-gate) were diffed after beautifying — the only deltas are webpack module-id renumbering and explicit coercions that JS already performed implicitly.

Type-stripping is behavior-neutral: the Babel build removes types with no runtime effect, so these sweeps are typing work rather than logic changes. Any behavior-affecting change is called out explicitly above.

2. `pnpm --filter <unit> run typescript:check` — the strict gate passes for every unit touched here.
3. `pnpm --filter <unit> run lint:js` and `pnpm --filter <unit> run build` — both green.
4. `pnpm --filter <unit> run test` for units carrying JS tests.
5. Compare the built `dist/` file list against the base branch — unchanged.

### Other information:

### Stack

Part of the [NPPM-2977](https://linear.app/a8c/issue/NPPM-2977/monorepo-jstypescript-migration) JS→TypeScript migration — a chain of stacked PRs, each targeting the previous phase's branch:

| Phase | PR | Scope |
|---|---|---|
| phase0 | #526 | Infra: type gate, tsconfigs, webpack ext-agnostic entry resolution |
| phase0.5 | #649 | Make the existing TS strict-clean, turn the gate on |
| phase1 | #650 | `packages/icons` + `packages/components` |
| phase2 | #651 | `newspack-plugin` |
| phase3 | #652 | `newspack-blocks` + `newspack-newsletters` |
| phase4 | #653 | Remaining plugins |
| phase5 | #654 | Both themes |
| phase6 | #655 | Wrap-up: tighten remaining `any`, blame-ignore revs |

Each PR will be retargeted as the chain lands. **Do not merge out of order.**


adekbadek:
No action taken — this is the file-count cap, not a finding.

The PR is already the smallest reviewable unit of this migration: converting `newspack-plugin/src` and flipping `allowJs: false` for that unit has to land together, otherwise the gate either fails on the un-converted remainder or is never turned on. That single commit alone is 260 files, and the migration is deliberately already split into seven stacked phase PRs (#526 → #655) to keep each one to a single unit. Splitting further would mean either a half-typed unit with no gate, or arbitrary alphabetical slices that are harder to review, not easier.

For human review, the per-commit diffs are the useful view: the phase2 work is `chore(newspack-plugin): migrate src to TypeScript and lock the unit` plus the two small `components` commits. Behaviour-neutrality is evidenced by the identical `dist/` file list and the beautified bundle diffs described in the PR body.

PR #656, opened 2026-07-17
feat(popups): move A/B variant selection into the view engine

jason10lee:
### All Submissions:

Adds the core display mechanics for A/B testing Campaigns prompts: two (or more) prompts sharing a test ID split eligible readers deterministically, and each reader always sees their assigned variant.

The A/B testing proof of concept selected variants from *outside* newspack-popups — after the view engine's conflict resolution had already hidden the older overlay — which made variant display unreliable (only the newest prompt in a test ever won). This PR moves selection *inside* the view engine, where per-reader display decisions already happen client-side after page cache:

* **Meta:** registers `newspack_ab_test_id`, `newspack_ab_variant`, `newspack_ab_test_goal`, and `newspack_ab_control_share` on the prompts CPT (REST-exposed with schema validation: variant enum, control share default 50 / range 20–80).
* **Config:** a localized `ab_tests` config is built from published prompts (a test is live only when its control *and* at least one challenger are published) with the post-meta cache primed to keep the always-enqueued path cheap.
* **Assignment:** readers are bucketed by a djb2 hash of the existing client-ID cookie over control-share-weighted ranges — no new identifier is introduced. Logged-in readers get a persisted bucket (user meta) computed from the same cookie identity, so a reader who registers mid-test stays in the same arm, and assignment is stable across devices. Hash parity between PHP and JS is pinned by shared test vectors. With no stable identity (cookies blocked), challengers are suppressed and only the control is eligible, preserving the one-variant invariant.
* **Display:** the A/B gate composes with the view engine's existing override system — non-assigned variants are suppressed *before* the single-overlay slot is claimed, and the assigned variant still passes the normal frequency/segmentation checks. The gate is placement-agnostic, so inline prompt pairs behave the same as overlays.
* **Analytics:** prompt events carry `ab_test_id`/`ab_variant` params via the existing Data API metadata (replacing the POC's gtag patch).
* **Preview:** editors can force a variant with `?view_as=ab_variant:b` — resolved server-side through the existing capability-gated `view_as` handling, so it isn't spoofable by visitors.

There's no editor UI yet (that's the next milestone) — with no test meta set, this is fully dormant: no markup, config, or behavior changes.

Closes NPPD-1499.

1. On a site with this branch built, create two overlay prompts (same placement, frequency "always") and mark them as a test — no UI yet, so via WP-CLI, creating the *control first* so the old "newest wins" behavior would mask it:
   ```
   wp post meta update <ID_A> newspack_ab_test_id my-test && wp post meta update <ID_A> newspack_ab_variant a
   wp post meta update <ID_B> newspack_ab_test_id my-test && wp post meta update <ID_B> newspack_ab_variant b
   ```
2. Visit a post logged out: exactly **one** variant displays. Verify `newspack_popups_view.ab_tests` in the page source and `data-ab-test-id`/`data-ab-variant` on both containers (loser keeps the `hidden` class).
3. Reload several times: the same variant persists. Change the `newspack-cid` cookie value: the assignment may flip, deterministically per value.
4. Repeat with two *inline* prompts sharing a test ID: exactly one renders in content.
5. Set `newspack_ab_control_share` to `80` on the control: across several fresh cookie values, most land on variant A.
6. As a logged-out visitor, append `?view_as=ab_variant:b` — nothing changes. As an admin, the same URL forces variant B everywhere.
7. While logged in as any user, verify your assignment is persisted (`wp user meta get <user_id> np_ab_bucket_my-test`) and stable even if the control share changes afterward.
8. Delete one variant's test meta (or unpublish it): the test deactivates everywhere — no `data-ab-*` attributes, both prompts behave as normal independent prompts.
9. Run `n test-php` and `npm test` in `plugins/newspack-popups` — both suites green (13 A/B-specific PHP tests, 17 jest tests, including PHP↔JS hash-parity vectors).

### Other information:

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
@@ -74,6 +74,16 @@ final class Newspack_Popups_AB_Tests {
 				[
 					'type'              => 'string',
 					'sanitize_callback' => [ __CLASS__, 'sanitize_variant' ],
+					// The enum makes an invalid REST write fail with a 400 the editor
+					// can surface, instead of the sanitize callback silently coercing
+					// it to '' (which would detach the prompt from its test on a 200).
+					// The empty string stays valid: it is the "not part of a test" state.
+					'show_in_rest'      => [
+						'schema' => [
+							'type' => 'string',
+							'enum' => array_merge( [ '' ], self::VALID_VARIANTS ),
+						],
+					],
 				]
 			)
 		);
@@ -98,6 +108,20 @@ final class Newspack_Popups_AB_Tests {
 				[
 					'type'              => 'integer',
 					'sanitize_callback' => [ __CLASS__, 'sanitize_control_share' ],
+					// The explicit default keeps REST reads of an unset share at the
+					// same 50 the display path assumes. Without it, the schema default
+					// 0 would round-trip through an editor and clamp to 20 — silently
+					// changing the live split and re-bucketing anonymous readers
+					// mid-test (see get_tests_config()).
+					'default'           => self::DEFAULT_CONTROL_SHARE,
+					'show_in_rest'      => [
+						'schema' => [
+							'type'    => 'integer',
+							'minimum' => 20,
+							'maximum' => 80,
+							'default' => self::DEFAULT_CONTROL_SHARE,
+						],
+					],
 				]
 			)
 		);
@@ -181,6 +205,13 @@ final class Newspack_Popups_AB_Tests {
 			]
 		);
 
+		// 'fields' => 'ids' skips meta-cache priming, so without this each
+		// get_post_meta() below would issue its own query — an N+1 on the
+		// always-enqueued front-end path. One batched load instead.
+		if ( ! empty( $prompt_ids ) ) {
+			update_meta_cache( 'post', $prompt_ids );
+		}
+
 		$config = [];
 		foreach ( $prompt_ids as $prompt_id ) {
 			$fields = self::get_popup_ab_fields( $prompt_id );
@@ -231,6 +262,12 @@ final class Newspack_Popups_AB_Tests {
 	 * hash in src/view/utils/ab.js. Server and client MUST use the same hash on
 	 * the same identity (the client ID cookie) or a reader can land in different
 	 * arms anonymous vs. logged-in. Parity is pinned by tests on both sides.
+	 *
+	 * ASCII-ONLY PRECONDITION: this hashes UTF-8 *bytes* (strlen/ord) while the
+	 * client hashes UTF-16 *code units* (charCodeAt) — the two agree only while
+	 * every hashed input (client ID, test ID) is ASCII. That holds today by
+	 * construction (generated cookie IDs; sanitize_title output). If either
+	 * input's charset ever widens, normalize both implementations together.
 	 *
 	 * @param string $str String to hash.
 	 * @return int Unsigned 32-bit hash.
--- a/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
@@ -18,12 +18,12 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Newspack_Popups_AB_Tests {
 
-	const META_TEST_ID       = 'newspack_ab_test_id';
-	const META_VARIANT       = 'newspack_ab_variant';
-	const META_GOAL          = 'newspack_ab_test_goal';
-	const META_CONTROL_SHARE = 'newspack_ab_control_share';
+	const META_TEST_ID       = 'newspack_popups_ab_test_id';
+	const META_VARIANT       = 'newspack_popups_ab_variant';
+	const META_GOAL          = 'newspack_popups_ab_test_goal';
+	const META_CONTROL_SHARE = 'newspack_popups_ab_control_share';
 
-	const USER_META_BUCKET_PREFIX = 'np_ab_bucket_';
+	const USER_META_BUCKET_PREFIX = 'newspack_popups_ab_bucket_';
 
 	const VALID_VARIANTS = [ 'a', 'b', 'c', 'd' ];
 
--- a/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
@@ -257,7 +257,7 @@ final class Newspack_Popups_AB_Tests {
 				// The prompts CPT is inherently small (tens of posts) and this is
 				// further narrowed by the meta filter; a bound here would silently
 				// truncate the config and drop whole tests (fail-open, uncounted).
-				'posts_per_page' => -1, // phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page
+				'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Prompts CPT; config-scale.
 				'fields'         => 'ids',
 				'meta_query'     => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 					[
--- a/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
@@ -1,0 +1,300 @@
+<?php
+/**
+ * Newspack Popups A/B Tests
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * A/B testing for prompts: meta registration, test configuration, and
+ * reader bucket assignment.
+ *
+ * Two prompts (or more, post-v1) sharing a `newspack_ab_test_id` form a test.
+ * Variant selection happens client-side in the view engine (src/view/utils/ab.js)
+ * using the config this class localizes; logged-in readers get a server-computed
+ * bucket persisted in user meta so assignment is stable across devices.
+ */
+final class Newspack_Popups_AB_Tests {
+
+	const META_TEST_ID       = 'newspack_ab_test_id';
+	const META_VARIANT       = 'newspack_ab_variant';
+	const META_GOAL          = 'newspack_ab_test_goal';
+	const META_CONTROL_SHARE = 'newspack_ab_control_share';
+
+	const USER_META_BUCKET_PREFIX = 'np_ab_bucket_';
+
+	const VALID_VARIANTS = [ 'a', 'b', 'c', 'd' ];
+
+	const DEFAULT_CONTROL_SHARE = 50;
+
+	/**
+	 * Memoized tests config.
+	 *
+	 * @var array|null
+	 */
+	private static $tests_config = null;
+
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'register_meta' ] );
+	}
+
+	/**
+	 * Register A/B test meta fields on the prompts CPT.
+	 */
+	public static function register_meta() {
+		$base = [
+			'object_subtype' => Newspack_Popups::NEWSPACK_POPUPS_CPT,
+			'show_in_rest'   => true,
+			'single'         => true,
+			'auth_callback'  => '__return_true',
+		];
+
+		\register_meta(
+			'post',
+			self::META_TEST_ID,
+			array_merge(
+				$base,
+				[
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_title',
+				]
+			)
+		);
+
+		\register_meta(
+			'post',
+			self::META_VARIANT,
+			array_merge(
+				$base,
+				[
+					'type'              => 'string',
+					'sanitize_callback' => [ __CLASS__, 'sanitize_variant' ],
+				]
+			)
+		);
+
+		\register_meta(
+			'post',
+			self::META_GOAL,
+			array_merge(
+				$base,
+				[
+					'type'              => 'string',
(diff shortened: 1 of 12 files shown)


tools/autofix/README.md
```diff
@@ -0,0 +1,122 @@
+# `autofix` — Linear-to-PR bug-fix workflow (operator tooling)
+
+Deterministic per-stage scripts backing the `autofix` orchestrator skill
+(`.claude/skills/autofix/SKILL.md`). This README covers running the CLI; the
+skill covers the judgment layer (triage rubric, scope guardrails, review
+gates). The full design is in the spec — see **Docs** below.
```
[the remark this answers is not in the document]
jason10lee: Good catch — the branch was accidentally based on a local integration branch rather than `origin/main`, which pulled unrelated workspace tooling into the diff. Rebased onto `origin/main` and force-pushed: the PR now contains only the newspack-popups A/B display change (11 files), matching the title and description.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/feat/nppd-1499-ab-variant-display` (stash local work first).

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
```diff
@@ -0,0 +1,381 @@
...
+	 * prompts render on a given page, so client-side hash ranges stay stable.
+	 *
+	 * @return array Config keyed by test ID: [ 'variants' => [ 'a', 'b' ], 'control_share' => 60 ].
+	 */
+	public static function get_tests_config() {
+		if ( null !== self::$tests_config ) {
+			return self::$tests_config;
+		}
+
+		$prompt_ids = get_posts(
```
dkoo, line 189: **Suggestion:** Dormancy is behavioral, not free — every site pays the discovery query — `get_tests_config()` runs inside the always-enqueued scripts path with only per-request memoization, so every non-AMP front-end request on every site — including sites with zero tests — executes a `get_posts` with a postmeta-join `meta_query`. WP's query cache absorbs most of it, but the cache entry drops on any post save, which on a news site is constant. A persistent any-test-exists flag (maintained on prompt save/trash) short-circuiting the builder makes the zero-test path genuinely free, matching the PR's dormancy claim in cost as well as behavior.

**Consider:** Persist a lightweight any-test-exists flag (option updated on prompt save) and short-circuit the config builder when unset, so zero-test sites skip the discovery query on the always-enqueued path.

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
```diff
@@ -0,0 +1,381 @@
...
+	public static function get_logged_in_buckets( $tests_config ) {
+		$user_id = get_current_user_id();
+		if ( ! $user_id ) {
+			return [];
+		}
+
+		$client_id = self::get_client_id();
+		$buckets   = [];
+		foreach ( $tests_config as $test_id => $config ) {
+			$meta_key = self::USER_META_BUCKET_PREFIX . $test_id;
```
dkoo, line 361: **Suggestion:** Per-test reader buckets accrue forever — `np_ab_bucket_<test_id>` user meta is written once per test per logged-in reader (correctly write-once — verified) but nothing removes the keys when a test ends or its prompts are deleted: on a large-membership site every reader ever assigned accrues one row per test, monotonically. Worth a teardown path (`delete_metadata( 'user', 0, $key, '', true )` when a test is dismantled) or at least a follow-up ticket. Relatedly, the `np_` prefix is a new abbreviation found nowhere else in the plugin — renaming later means a data migration, so settle it now.

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
```diff
@@ -0,0 +1,381 @@
...
+			$bucket   = get_user_meta( $user_id, $meta_key, true );
```
dkoo, line 362: **Suggestion:** Three-arm edge: transient unpublish permanently re-buckets readers — The `! in_array( $bucket, $config['variants'], true )` guard recomputes and overwrites the persisted bucket when the stored variant leaves the published set. Unreachable in a two-arm v1 (unpublishing the only challenger kills the whole test), but in a ≥3-arm test, briefly drafting variant `c` permanently reassigns every `c` reader — republishing does not restore them, and the docblock's "persisted value always wins thereafter" overstates the guarantee. Preserve any stored `VALID_VARIANTS` value and only fall back for display, or document the behavior as intended.

plugins/newspack-popups/src/view/segmentation.js
```diff
@@ -35,7 +36,11 @@ export const handleSegmentation = prompts => {
 		prompts.forEach( prompt => {
 			const promptId = prompt.getAttribute( 'id' );
 			const isOverlay = prompt.classList.contains( 'newspack-lightbox' );
-			const override = getOverride( getRawId( promptId ), isOverlay, overlayDisplayed );
+			// A/B variant selection composes with the standard override: a test
+			// variant the reader is not assigned to is suppressed before it can
```
dkoo, line 40: **Suggestion:** The crux regression has no direct integration test — That a suppressed overlay variant leaves the single-overlay slot unclaimed (so a later legitimate overlay still shows), and that two inline arms never both render, is exactly the POC bug this PR exists to fix — but it's only covered indirectly by `getAbOverride` unit tests and PHP markup assertions. A direct test of the `getOverride( … ) ?? getAbOverride( … )` composition in `handleSegmentation` would pin the regression. Related coverage gaps: the localization wiring (`ab_tests`/`cid_cookie`/`ab_buckets`/`ab_view_as` present when live, absent when dormant) is untested, the parity vectors are duplicated literals rather than one shared fixture (a small shared JSON consumed by both Jest and PHPUnit, with a boundary-adjacent vector, would make the single-source intent structural), and the read-time control-share clamp and stored-0 fallback branches are unexercised.

plugins/newspack-popups/src/view/utils/ab.js
```diff
@@ -0,0 +1,145 @@
...
+ * Get the reader's client ID from the Reader Activation cookie.
+ *
+ * @param {string|null} cookieString Optional, for testing. A cookie string to parse.
+ * @return {string|null} Client ID, or null if unavailable.
+ */
+export const getReaderId = ( cookieString = null ) => {
+	const cookieName = getViewData().cid_cookie || DEFAULT_CID_COOKIE;
+	const cookies = null === cookieString ? document.cookie || '' : cookieString;
+	const match = cookies.match( new RegExp( '(?:^|;\\s*)' + cookieName + '=([^;]+)' ) );
+	return match ? decodeURIComponent( match[ 1 ] ) : null;
```
dkoo, line 44: **Nit:** Cookie-decode hardening and latent PHP/JS coupling asymmetries — Three related small items: `decodeURIComponent( match[1] )` throws `URIError` on a malformed self-set cookie, aborting prompt processing for that page — a `try/catch` returning null hardens it; the mid-test login-continuity feature depends on `sanitize_text_field()` (server, no URL-decode) and `decodeURIComponent()` (client) producing the same string — true for Newspack-generated CIDs but a second latent coupling beyond the documented ASCII-only precondition, worth one comment line; and the unset-control-share handling is asymmetric (`||` in JS treats 0 as 50, `??` in PHP keeps 0) — unreachable today because the config always emits a clamped value, but exactly the divergence class the parity comments warn about.

plugins/newspack-popups/src/view/utils/segments.js
```diff
@@ -15,7 +15,7 @@ export const periods = {
  *
  * @return {Object|null} View_as object or null.
  */
-const parseViewAs = ( queryString = null ) => {
+export const parseViewAs = ( queryString = null ) => {
```
dkoo, line 18: **Nit:** Small cleanups — Assorted: `parseViewAs` is exported but nothing imports it (leftover from an earlier client-side approach — revert to module-private); the AMP branch skips the whole A/B block, which is correct today but deserves a one-line comment noting the dependency (if AMP prompt display is ever restored, both arms would render un-suppressed); and for the record, the localized config is site-wide by design (every live test's id/variants/share on every page — published-prompt metadata only, goals never localized).

dkoo (commented):
Release risk: ELEVATED — the consequences verdict (medium) sets the floor; verification is now complete and everything passed. A seven-step browser walkthrough against the isolated env confirmed the full behavior matrix live: exactly one variant per test displays (both arms render hidden with data-ab-* attributes; the loser stays hidden), assignment persists across reloads and is deterministic per client ID — precomputed djb2 hashes predicted every observed arm exactly, including the control-share skew check where raising the share to 80 flipped precisely the one reader whose hash (0.705) sat in the widened control range. view_as=ab_variant:b is a no-op for visitors and forces the variant for admins; logged-in readers get server-computed buckets persisted to np_ab_bucket_* user meta that match what the page echoes and survive reloads. The earlier concern that the pre-existing ?pid= preview path could let any visitor force a losing arm was refuted live and withdrawn: with ?pid= present the server renders zero prompts for non-admins (pre-existing admin gate in retrieve_preview_popup()), so the client-side pid path is unreachable for visitors. Three things remain worth attention before merge. First, dormancy is behavioral but not free: every site pays a posts+meta query on every non-AMP front-end request from release day, tests or no tests — a persistent has-tests flag makes the zero-test path cost nothing. Second, per-test reader buckets accrue as user meta forever with no cleanup on test teardown. Third, the crux regression this PR exists to prevent has no direct integration test — the composition is only covered indirectly. Smaller items: a three-arm edge where briefly unpublishing a challenger permanently re-buckets its readers, duplicated (rather than shared) parity vectors, and a couple of latent PHP/JS coupling asymmetries documented in the comments. Approve-leaning: the runtime behavior is proven; the remaining items are hardening and test-coverage follow-ups, none blocking.

dkoo (approved):
Pre-approving after verifying with manual QA. The suggestions above are non-blocking, but I'd strongly consider folding in changes for the [performance issue](https://github.com/Automattic/newspack-workspace/pull/656#discussion_r3661436289) and [immutable per-reader test buckets](https://github.com/Automattic/newspack-workspace/pull/656#discussion_r3661436293) before merging.

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
jason10lee: Implemented as `OPTION_HAS_TESTS`, autoloaded, invalidated on prompt save/trash/delete and on `newspack_popups_ab_test_id` meta changes — scoped to the prompts CPT so a news site's ordinary post saves don't reinstate the cost.

Deliberately three-state rather than boolean: only `'0'` short-circuits, and invalidation *deletes* the option rather than writing `'0'`. A missed invalidation therefore costs a query, not silently-dead tests, which is the failure this flag makes possible.

Four tests, including one that sets the test-id meta on an already-saved prompt. That path matters: creating a prompt fires `save_post`, so my first test passed with the meta hooks removed — it proved invalidation happened, not that the meta hook did it.

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
jason10lee: Renamed, and wider than the thread: all five keys, since the class is new in this PR and none are in use yet — `newspack_popups_ab_{test_id,variant,test_goal,control_share,bucket_}`. Free now in a way it would not be once a site has written reader buckets.

Teardown is NPPD-2180, filed against M4 since it needs a "test is over" moment that arrives with winner declaration.

plugins/newspack-popups/includes/class-newspack-popups-ab-tests.php
jason10lee: Docblock corrected — it now describes the recompute-on-unpublished-variant behavior instead of claiming the persisted value always wins. Behavior change is NPPD-2181, no milestone, noting it only becomes reachable once multi-variant ships and should be fixed before then rather than after.

plugins/newspack-popups/src/view/segmentation.js
jason10lee: New `src/view/segmentation.test.js` exercising `getOverride( ... ) ?? getAbOverride( ... )` through the real loop, with `getOverride`/`getAbOverride` unmocked. Three cases: a suppressed overlay variant doesn't claim the slot, only the assigned inline arm renders, and one-overlay-at-a-time still holds among assigned prompts. The arm is derived from the real hash rather than hard-coded, so it keeps testing suppression if bucketing is retuned.

Mutating the composition down to `getOverride( ... )` fails two of the three, reproducing the POC bug.

The related gaps you listed — localization wiring, the duplicated parity vectors, and the clamp/stored-`0` branches — are tracked as NPPD-2182 rather than done here.

plugins/newspack-popups/src/view/utils/ab.js
jason10lee: `try/catch` returning null, with a test that throws `URIError` without it. The `sanitize_text_field`/`decodeURIComponent` coupling is now a comment at the decode site.

Left the `||` vs `??` asymmetry alone: unreachable while the config always emits a clamped value, and closing it would mean picking a behavior for a case neither side can currently produce. Say if you'd rather it be made symmetric anyway.

plugins/newspack-popups/src/view/utils/segments.js
jason10lee: `parseViewAs` is module-private now (it's used at line 90 of its own file, so it was the `export` to drop, not the function).

The AMP note went to `class-newspack-popups-inserter.php` rather than here — the skip is the `if ( ! self::is_amp() )` wrapping the whole `$script_data` block, so a future reader restoring AMP prompt display would be looking there. It records that every arm would render un-suppressed on that path.

No action on the site-wide localized config; agreed it's by design.

PR #657, opened 2026-07-17
feat(audience): polish integrations settings UI

thomasguillot:
### All Submissions:

Design polish pass on the Audience → Integrations screens.

**Integrations list**

* Integration cards sit on the standard 32px gutter instead of a tighter 16px one.

**Integration settings**

* Section dividers get more breathing room (64px above and below).
* The settings column no longer sits inside a redundant wrapper, so it lines up with the rest of the screen.
* Saving an integration now confirms with a "Settings saved." snackbar. A failed save already told you something went wrong; a successful one said nothing at all.

**Accordion (`packages/components`)**

* Rebuilt on top of core's `PanelBody` rather than a hand-rolled `<details>` element. The previous version shipped a boxed look that its only consumer immediately overrode, so the two fought each other.
* An accordion is now a container of one or more panels — `Accordion` + `AccordionPanel` — mirroring core's `Panel`/`PanelBody`. Panels are separated by a tertiary divider, never trailing after the last one.
* New opt-in `hideSingleTitle` prop: where a group would render as a single panel, it renders open and untitled instead of as a pointless one-item disclosure. Opt-in rather than automatic, so a single titled collapsible is still expressible.

**Shared styles**

* Checkbox labels are no longer bold, and their help text is indented from the checkbox's own size/margin variables rather than a hardcoded value that no longer matched.
* `.newspack-wizard__main` clips on both axes.

Note: the multi-panel accordion only appears on sites running metadata schema 1.0 (`NEWSPACK_SYNC_METADATA_VERSION_1`). On the default (legacy) schema every field lands in one catch-all group, which is why `hideSingleTitle` exists.

Stacked on #641 — merge after #258 → #632 → #641.

Closes DSGNEWS-199.

1. Go to **Audience → Integrations**. The integration cards should sit on a 32px gutter.
2. Open **Newsletter ESP** → **Configure**. Check the dividers between Settings / Inbound / Outbound have 64px above and below, and that the settings column aligns with the page (no nested wrapper).
3. Confirm checkbox labels in Inbound/Outbound are regular weight, not bold.
4. Under **Outbound**, on a default site you should see the field list directly, with no redundant "Additional" toggle above it.
5. Change a setting (e.g. **Metadata field prefix**) and hit **Save**. A "Settings saved." snackbar should appear bottom-centre, and the value should persist on reload.
6. Break the save (e.g. go offline or block the REST request) and hit **Save** again. You should get the error snackbar, not a success one. Restore the connection and save again — the error snackbar should be replaced by the success one, not stacked alongside it.
7. To exercise the multi-panel accordion, add `define( 'NEWSPACK_SYNC_METADATA_VERSION_1', true );` to `wp-config.php` and reload. Outbound should now show Identity / Registration / Engagement as collapsible panels, separated by light dividers, with 24px between them and no divider after the last panel. The first panel is open; titles are kept because there is more than one.
8. Keyboard-tab to a panel toggle: the focus ring should show. Click one with the mouse: no focus ring, but it should still expand/collapse.
9. Check other wizard screens (Settings, Dashboard, Audience → Configuration) still scroll and render normally — `.newspack-wizard__main` and the checkbox styles are shared.

### Other information:

### DIFF
--- a/packages/components/src/accordion/index.js
+++ b/packages/components/src/accordion/index.js
@@ -36,7 +36,7 @@ const Accordion = ( { children, className, hideSingleTitle = false, spacing = 6 
 			{ panels.map( ( panel, index ) => (
 				<Fragment key={ panel.key }>
 					{ panel }
-					{ index < panels.length - 1 && <Divider variant="tertiary" marginBottom={ 0 } marginTop={ 0 } /> }
+					{ index < panels.length - 1 && <Divider variant="secondary" marginBottom={ 0 } marginTop={ 0 } /> }
 				</Fragment>
 			) ) }
 		</VStack>
--- a/packages/components/src/accordion/index.js
+++ b/packages/components/src/accordion/index.js
@@ -1,8 +1,8 @@
 /**
  * WordPress dependencies
  */
-import { useState } from '@wordpress/element';
-import { Icon, chevronRight } from '@wordpress/icons';
+import { __experimentalVStack as VStack, PanelBody } from '@wordpress/components'; // eslint-disable-line @wordpress/no-unsafe-wp-apis
+import { Children, Fragment, cloneElement } from '@wordpress/element';
 
 /**
  * External dependencies
@@ -12,22 +12,34 @@ import classNames from 'classnames';
 /**
  * Internal dependencies
  */
+import Divider from '../divider';
 import './style.scss';
 
-const Accordion = ( { children, title, defaultOpen = false } ) => {
-	const [ isOpen, setIsOpen ] = useState( defaultOpen );
+export const AccordionPanel = ( { children, className, title, defaultOpen = false } ) => (
+	<PanelBody className={ className } title={ title } initialOpen={ defaultOpen }>
+		{ children }
+	</PanelBody>
+);
+
+const Accordion = ( { children, className, hideSingleTitle = false, spacing = 6 } ) => {
+	const panels = Children.toArray( children );
+	// With nothing to collapse against, a lone panel can render open and untitled.
+	if ( hideSingleTitle && panels.length === 1 ) {
+		return (
+			<div className={ classNames( 'newspack-accordion', className ) }>
+				{ cloneElement( panels[ 0 ], { defaultOpen: true, title: undefined } ) }
+			</div>
+		);
+	}
 	return (
-		<details
-			className={ classNames( 'newspack-accordion', { 'newspack-accordion--is-open': isOpen } ) }
-			open={ isOpen }
-			onToggle={ e => setIsOpen( e.currentTarget.open ) }
-		>
-			<summary>
-				{ title }
-				<Icon className="newspack-accordion__icon" icon={ chevronRight } size={ 24 } />
-			</summary>
-			<div className="newspack-accordion__content">{ children }</div>
-		</details>
+		<VStack className={ classNames( 'newspack-accordion', className ) } spacing={ spacing }>
+			{ panels.map( ( panel, index ) => (
+				<Fragment key={ panel.key }>
+					{ panel }
+					{ index < panels.length - 1 && <Divider variant="tertiary" marginBottom={ 0 } marginTop={ 0 } /> }
+				</Fragment>
+			) ) }
+		</VStack>
 	);
 };
 
--- a/packages/components/src/grid/style.scss
+++ b/packages/components/src/grid/style.scss
@@ -61,6 +61,11 @@
 		grid-row-gap: 16px;
 	}
 
+	// Row Gap 24
+	&__row-gap-24 {
+		grid-row-gap: 24px;
+	}
+
 	// Row Gap 32
 	&__row-gap-32 {
 		grid-row-gap: 32px;
--- a/packages/components/src/style.scss
+++ b/packages/components/src/style.scss
@@ -35,12 +35,9 @@
 // Checkboxes
 .newspack-checkbox-control {
 	margin-top: 0 !important;
-	.components-checkbox-control__label {
-		font-weight: 600;
-	}
 
 	.components-base-control__help {
-		margin-left: 32px;
+		margin-left: calc(var(--checkbox-input-size) + var(--checkbox-input-margin));
 		margin-top: 0;
 	}
 }
--- a/packages/components/src/with-wizard-screen/style.scss
+++ b/packages/components/src/with-wizard-screen/style.scss
@@ -46,7 +46,7 @@
 	}
 
 	&__main {
-		overflow-x: hidden;
+		overflow: hidden;
 		padding-bottom: 1px; /* Prevents borders from being cut off */
 	}
 
--- a/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
@@ -273,7 +273,7 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
 									const currentValue = getFieldValue( outboundField );
 									const selected = Array.isArray( currentValue ) ? currentValue : [];
 									return (
-										<AccordionPanel key={ group.section } title={ group.section } defaultOpen={ index === 0 }>
+										<AccordionPanel key={ `${ index }-${ group.section }` } title={ group.section } defaultOpen={ index === 0 }>
 											<Grid columns={ 1 } rowGap={ 8 } noMargin>
 												{ group.fields.map( fieldName => (
 													<CheckboxControl
--- a/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
@@ -9,7 +9,7 @@ import { useEffect, useMemo, useRef, useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { Accordion, Divider, Grid, SectionHeader, useUnsavedChangesDialog } from '../../../../../packages/components/src';
+import { Accordion, AccordionPanel, Divider, Grid, SectionHeader, useUnsavedChangesDialog } from '../../../../../packages/components/src';
 import { WIZARD_STORE_NAMESPACE } from '../../../../../packages/components/src/wizard/store';
 import WizardsTab from '../../../wizards-tab';
 import { SettingsField } from './settings-field';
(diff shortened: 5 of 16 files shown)


packages/components/src/style.scss
```diff
@@ -35,12 +35,9 @@
...
 .newspack-checkbox-control {
 	margin-top: 0 !important;
-	.components-checkbox-control__label {
-		font-weight: 600;
-	}
 
 	.components-base-control__help {
-		margin-left: 32px;
+		margin-left: calc(var(--checkbox-input-size) + var(--checkbox-input-margin));
 		margin-top: 0;
```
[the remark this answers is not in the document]
thomasguillot: Core Component define these vars

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -214,91 +214,87 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
...
-								{ settingsFields.filter( fieldIsVisible ).map( field => (
-									<SettingsField
-										key={ field.key }
-										field={ field }
-										value={ getFieldValue( field ) }
-										onChange={ val => handleFieldChange( field.key, val ) }
-									/>
-								) ) }
-							</Grid>
+			<div className="newspack-configure-view">
```
miguelpeixe, line 217: **[suggestion] Dropping WizardsTab changes which shared layout rule matches**

Removing the `<WizardsTab>` wrapper means this view no longer renders inside `.newspack-wizard__sections`, so the `.newspack-wizard__content:has(> .newspack-wizard__sections)` rule in `admin/style.scss:157` stops matching it. Horizontal padding, centering and max-width are preserved (the default `.newspack-wizard__content` rule uses the identical formula), but two things change versus both the prior behavior and the sibling `/settings` list view (which still uses `WizardsTab`): the ~32px bottom padding that rule adds is lost (default `__content` has `0` bottom padding), and its `* { box-sizing: border-box }` reset no longer applies to this subtree. WP components mostly set their own box-sizing, so the practical effect is mainly slightly tighter bottom spacing on the last section. Worth a quick visual confirm.

The loading affordance is *not* lost: `loading` is only ever true while `integration` is `undefined`, which the `if ( ! integration ) return <WizardsTab isFetching={ loading } />` guard at line 179 still catches — so "drop unreachable is-fetching branch" is accurate.

packages/components/src/accordion/index.js
```diff
@@ -12,22 +12,34 @@ import classNames from 'classnames';
 /**
  * Internal dependencies
  */
+import Divider from '../divider';
 import './style.scss';
 
-const Accordion = ( { children, title, defaultOpen = false } ) => {
-	const [ isOpen, setIsOpen ] = useState( defaultOpen );
+export const AccordionPanel = ( { children, className, title, defaultOpen = false } ) => (
+	<PanelBody className={ className } title={ title } initialOpen={ defaultOpen }>
```
miguelpeixe, line 19: **[suggestion] Rebuilt Accordion renders independent panels, not single-open**

Each `AccordionPanel` is an uncontrolled `PanelBody` (`initialOpen`), so the rebuilt `Accordion` renders a set of *independent* collapsibles — opening one panel doesn't close the others. That matches core's multi-`PanelBody` behavior and the previous `<details>`-based component, and is reasonable for the outbound metadata groups. But it isn't classic single-open (mutually-exclusive) accordion behavior. If the design for DSGNEWS-199 intends single-open panels, that needs a controlled `opened` state lifted into `Accordion`; if independent collapsibles are intended, this is correct as-is. Worth one design confirmation.

packages/components/src/with-wizard-screen/style.scss
```diff
@@ -46,7 +46,7 @@
 	}
 
 	&__main {
-		overflow-x: hidden;
+		overflow: hidden;
```
miguelpeixe, line 49: **[suggestion] Shared overflow change clips both axes on every wizard screen**

`overflow-x: hidden` → `overflow: hidden` on `.newspack-wizard__main` is safe for scrolling: the element is auto-height (no `height`/`max-height` set anywhere) and the page scrolls at `#wpbody`, so vertical page scroll is unaffected. Per the CSS overflow spec, `overflow-x: hidden` already forced the used `overflow-y` to `auto`, so `__main` was already a scroll-container / BFC — the only new behavior is that a vertically-overflowing descendant is now clipped instead of showing an (unwanted) inner scrollbar, and nothing here relies on vertical bleed (the dataviews negative-margin bleed is horizontal, already clipped). Since `.newspack-wizard__main` is shared by every wizard, a fast pass over Dashboard / Settings / Advertising for any inline (non-portaled) hover-card or edge focus-ring is cheap insurance — but not a blocker.

packages/components/src/grid/style.scss
```diff
@@ -61,6 +61,11 @@
 		grid-row-gap: 16px;
 	}
 
+	// Row Gap 24
+	&__row-gap-24 {
```
miguelpeixe, line 65: **[nit] Unused `__row-gap-24` grid modifier (dead CSS)**

The new `&__row-gap-24 { grid-row-gap: 24px }` modifier isn't consumed anywhere — grepping `plugins/` and `packages/` for `rowGap={ 24 }` / `row-gap-24` returns no consumers. The change that motivated it (the Settings grid in `configure-view.js:222`) switched to `gutter={ 24 }`, which maps to the pre-existing `__gutter-24` rule, not `rowGap`. 24px is a valid grid step (`$grid-unit-30`, 3×8), so either wire it up — `rowGap={ 24 }` on that single-column grid is arguably clearer than `gutter` — or drop the rule.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -214,91 +214,87 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
...
-													handleCheckboxListChange( inboundField.key, currentValue, optionValue, checked )
-												}
-											/>
-										);
-									} ) }
-								</Grid>
+				{ /* Section 2: Inbound */ }
+				{ inboundField && (
+					<>
+						<Divider alignment="full-width" variant="tertiary" marginTop={ 64 } marginBottom={ 64 } />
```
miguelpeixe, line 238: **[nit] Tertiary divider does double duty for panels and sections**

`variant="tertiary"` is the weakest divider variant. It's used both for intra-accordion panel separators (appropriately subtle) and for the top-level Inbound/Outbound *section* breaks (lines 238 and 268) — two different hierarchy levels drawn with the same lightest line, differentiated only by spacing (24px panel gap vs 64px section margin). It reads fine because the spacing carries the hierarchy, and it's within the component's supported variants; just worth a quick check with design on whether the major section breaks should use a stronger variant.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -214,91 +214,87 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
...
+						<Grid columns={ 2 } gutter={ 32 } noMargin>
+							<SectionHeader heading={ 2 } title={ __( 'Outbound', 'newspack-plugin' ) } noMargin />
+							<Accordion hideSingleTitle>
+								{ ( outboundField.grouped_options || [] ).map( ( group, index ) => {
+									const currentValue = getFieldValue( outboundField );
+									const selected = Array.isArray( currentValue ) ? currentValue : [];
+									return (
+										<AccordionPanel key={ group.section } title={ group.section } defaultOpen={ index === 0 }>
```
miguelpeixe: **[nit] Panel key can collide on duplicate section names**

`<AccordionPanel key={ group.section } … >` keys panels by section name. If the backend ever emits two `grouped_options` entries with the same `section`, `Children.toArray` would surface duplicate React keys. It's low-probability and pre-existing (the old `<Accordion>` keyed the same way), but `${ index }-${ group.section }` would be more robust.

miguelpeixe (approved):
Approving — no blockers.

A clean, well-tested design-polish pass. The `Accordion` rebuild is correctly wired (no other consumers, edge cases covered) and the save snackbar's clear-then-add logic is genuinely necessary rather than incidental — the notices store appends without de-duping by id, so re-adding the same id would otherwise stack.

Two things worth a quick look before merge, both low-risk but visual:
- Dropping the `WizardsTab` wrapper changes which shared `:has()` rule matches the configure view (a minor bottom-padding / box-sizing divergence from the sibling list view).
- The `overflow: hidden` on the shared `.newspack-wizard__main` touches every wizard screen — safe on an auto-height container, but worth a pass over the other wizard screens.

Beyond that it's one dead-CSS cleanup (`__row-gap-24`) and two design-intent confirmations (accordion open/close semantics, and the tertiary divider doing double duty for panel separators and section breaks). Details inline.

On the open checkbox-indent thread: agree with @thomasguillot — `--checkbox-input-size` / `--checkbox-input-margin` are defined by core on `.components-checkbox-control` (the same `BaseControl` root the `newspack-checkbox-control` class lands on), so the `calc()` resolves and actually corrects a prior ~8px over-indent at ≥600px. Non-issue.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
thomasguillot: Intentional. The default `.newspack-wizard__content` rule preserves the horizontal padding, centering and max-width, and the loading guard at line 179 still covers the `isFetching` case. The slightly tighter bottom spacing is fine here, so leaving as-is.

packages/components/src/accordion/index.js
thomasguillot: Independent collapsibles are intended for DSGNEWS-199 — this matches the previous `<details>`-based component and core's multi-`PanelBody` behavior. Not single-open.

packages/components/src/with-wizard-screen/style.scss
thomasguillot: Verified safe visually across Dashboard, Settings, Advertising and the ESP configure view — no clipped popovers or focus rings. Native `<select>`s and portaled popovers escape the container. Keeping `overflow: hidden`.

packages/components/src/grid/style.scss
thomasguillot: Keeping the modifier intentionally — the 24px row step (`$grid-unit-30`) is worth having available for single-column grids. Left unused for now.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
thomasguillot: Split the two levels in a160f156: the accordion panel separators now use `secondary` (gray-200) so the collapsible rows read clearly, while the section breaks stay `tertiary` since the 64px spacing already sets them apart.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
thomasguillot: Fixed in ba8634a0 — keyed by `${ index }-${ group.section }` now.

PR #658, opened 2026-07-17
fix(multibrand): treat brand front page as a page, not an archive

rbcorrales:
### All Submissions:

When a brand uses a dedicated page as its front page, that page now renders its full block content in all cases. Previously, with Jetpack active and Content Options > Blog Display set to "Post excerpt", the brand front page showed only a truncated text excerpt and dropped every block.

The root cause was in how the front page is served: the main query was rewritten into a page query without resetting the query flags, so the brand archive's `is_archive` and `is_tax` stayed true alongside `is_page`. Jetpack's Content Options excerpt filter is gated on `is_home() || is_archive()`, so it treated the page as an archive listing and swapped in an excerpt. Passing the new query to `parse_query()` resets the flags, so the page is seen as a page by every consumer, not just Jetpack.

This also resolves related symptoms that share the same cause: archive styling leaking onto branded landing pages, and the front page being misclassified as an archive. Brand theming (logo, colors, menus) and the front page layout are preserved, and the brand URL stays canonical.

Closes NPPM-2993.

1. On a multibrand site with Jetpack active, go to Newspack > Settings > Additional Brands, edit a brand, set "Show on Front" to "A page", and select a published page that contains blocks (for example a Buttons block).
2. Go to Appearance > Customize > Content Options > Blog Display, choose "Post excerpt", and publish.
3. Visit the brand front page URL. Confirm it renders the full block content. Before this fix it showed only a truncated text excerpt with the blocks removed.
4. Switch Blog Display back to "Default" and confirm the page still renders correctly.
5. Confirm brand theming (logo, colors, menus) and the front page layout (the `newspack-front-page` body class) are intact.
6. Confirm the brand URL stays canonical and does not redirect to the page's own permalink.

### Other information:


plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
```diff
@@ -60,9 +60,14 @@ public static function is_filtered() {
 	 * @return void
```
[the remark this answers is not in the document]
rbcorrales: Fixed in 7df9e1770. Used the fully-qualified `\WP_Query` in the docblock.

plugins/newspack-multibranded-site/tests/unit-tests/test-customization-page-on-front.php
```diff
@@ -31,6 +31,126 @@ public function test_options_hook() {
...
+
+		$this->go_to( get_term_link( $brand ) );
+
+		$this->assertTrue( have_posts(), 'The brand front page should be queried.' );
+		the_post();
+		$rendered = apply_filters( 'the_content', get_the_content() );
+
+		$this->assertStringContainsString( 'Unique block content.', $rendered );
+		$this->assertStringNotContainsString( 'REPLACED BY EXCERPT', $rendered );
+	}
```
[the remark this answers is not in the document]
rbcorrales: This does not leak in WP_UnitTestCase. The framework backs up the entire `$wp_filter` global in `set_up()` via `_backup_hooks()` and restores it in `tear_down()` via `_restore_hooks()` (WP test lib `abstract-testcase.php`), so any filter added during a test is removed before the next test runs. This matches the existing pattern in this suite and in WP core, where test filters are added without manual removal. Leaving as is.

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
```diff
@@ -56,13 +56,18 @@ public static function is_filtered() {
 	/**
 	 * Change the query if we want to display a page on front
 	 *
-	 * @param WP_Query $query The WP_Query object.
+	 * @param \WP_Query $query The WP_Query object.
 	 * @return void
 	 */
 	public static function pre_get_posts( &$query ) {
+		// Bail before the reset so secondary queries can't clear the flag.
+		if ( ! $query->is_main_query() ) {
```
chickenn00dle, line 83: **Nit:** `$filtered` is now only cleared by a main query — With the bail above the reset, `self::$filtered` is only ever cleared by a main-query `pre_get_posts`. In a persistent process — WP-CLI, or the test suite where the static leaks across test methods — a `true` can survive indefinitely.

Both readers also require `is_page()`, so there's no production path to a wrong result, and this is the deliberate trade-off the reordering buys (it's what makes `is_filtered()` trustworthy for `template_include`). Noting it because it's why the new tests are order-dependent on `go_to()` running first.

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
```diff
@@ -78,46 +83,31 @@ public static function pre_get_posts( &$query ) {
 			if ( ! empty( $show_page_on_front ) ) {
 				$page = get_page( $show_page_on_front );
 				if ( $page ) {
-					$query->query      = [ 'page_id' => $page->ID ];
-					$query->query_vars = $query->query;
-					$query->parse_query();
+					// Pass the query as an argument so parse_query() resets the flags; otherwise the
+					// brand archive's is_archive/is_tax survive and consumers misread the page as an archive.
+					$query->parse_query( [ 'page_id' => $page->ID ] );
```
chickenn00dle, line 108: **Nit:** Worth confirming the reporter's front page isn't the blog page — Pre-existing rather than introduced here, but adjacent and now easier to hit conceptually: core's `parse_query()` flips `is_page` back to `false` and sets `is_home`/`is_posts_page` when `page_id` equals the site's `page_for_posts`.

So a publisher who picks the *blog* page as a brand's front page ends up with `is_home() === true`, which re-triggers the very Jetpack excerpt gate this PR fixes, and also disables both `body_class()` and `template_include()` via their `is_page()` guards. This behaves identically before and after the change, so it is not a regression.

Raising it only because it's a cheap thing to rule out before calling NPPM-2993 closed — worth confirming the reporter's brand front page isn't `page_for_posts`.

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
```diff
@@ -78,46 +83,31 @@ public static function pre_get_posts( &$query ) {
...
+	 * Adds the front page body class when the brand displays a page on front.
+	 *
+	 * Core supplies the page-* classes; the theme only adds newspack-front-page for the
+	 * site's own front page, so brand front pages need it added here.
 	 *
 	 * @param string[] $classes An array of body class names.
 	 * @param string[] $class   An array of additional class names added to the body.
 	 * @return array
 	 */
 	public static function body_class( $classes, $class ) {
```
chickenn00dle, line 126: **Nit:** Custom page templates now emit a mismatched body class — The old code force-added `page-template-default` to match the forced `get_front_page_template()`. Core now supplies the `page-template-*` class from the page's own `_wp_page_template` meta, so a cover page with a custom page template assigned will emit `page-template-<custom>` while `template_include` still renders `front-page.php`.

A theme keying CSS off `page-template-*` would see a mismatch. Narrow enough to leave — but of the classes the removal dropped, this is the one whose absence isn't compensated by core.

plugins/newspack-multibranded-site/tests/unit-tests/test-customization-page-on-front.php
```diff
@@ -31,6 +31,126 @@ public function test_options_hook() {
...
+
+	/**
+	 * The brand front page gets front page body classes, not archive ones.
+	 */
+	public function test_brand_front_page_body_classes() {
+		list( $brand, $page ) = $this->create_brand_with_page_on_front();
+
+		$this->go_to( get_term_link( $brand ) );
+		$classes = get_body_class();
+
```
chickenn00dle, line 105: **Nit:** Add a negative case so the `is_filtered()` guard is load-bearing — The six new tests pin the regression well — flag state, the excerpt path with a faithful copy of Jetpack's gate, body classes including both negative assertions, the secondary-query flag, and canonical in both directions with an explicit non-vacuity guard. That's a good suite.

One gap: nothing asserts that an ordinary, non-brand page does *not* receive `newspack-front-page`. As written, `test_brand_front_page_body_classes` would still pass if the `is_filtered()` guard were deleted outright, since `is_page()` alone is true there. A three-line `go_to( get_permalink( $unrelated_page ) )` plus `assertNotContains` would make that guard load-bearing under test.

chickenn00dle (commented):
**Release risk: ELEVATED** — no blockers. The change is correct and I'd ship it; the tier reflects one adjacent behaviour change worth a QA note, plus verification that covered the fix but not its knock-on effects.

I checked the mechanics against core rather than assuming, and the approach is right. `parse_query( $args )` is what `WP_Query::query()` does internally, and it's core's `init()` that clears the inherited flags — the old no-argument `parse_query()` skipped `init()` entirely, which is exactly why `is_archive`/`is_tax` survived. With `page_id` set, the whole archive-detection block sits in an `else` that can't run, and `parse_tax_query()` still reassigns an empty `WP_Tax_Query`, so there's no leftover term state for a downstream consumer to pick up. Canonical is safe structurally rather than by luck: both `is_page()` branches in `redirect_canonical()` need either `$_GET['page_id']` or the page being the site's own `page_on_front`, and neither applies to a brand URL — test 6 pins that non-vacuously.

Moving the `is_main_query()` bail above the reset is worth more than it looks. Before this, *any* secondary `WP_Query` running during `wp()` could clear `$filtered` before `template_include` fired, silently dropping the front-page template. That's a latent bug this quietly fixes, and `test_rss_feed_intact` still pins that a later main query does reset.

Also worth noting for whoever reviews the diff: two of the removed body-class entries were no-ops. `array_diff` compares whole values, so `'-template-default'` never matched `page-template-default` — pre-change the body tag carried both the real class and a bogus duplicate. The thin additive filter is a genuine simplification, not just a rewrite.

Verified at runtime in an isolated env with Jetpack and Content Options set to "Post excerpt": the page renders its block content, body classes are right with no `archive` and no term-id class, and a probe confirms `is_page=true is_archive=false is_tax=false`. Plugin tests 41/41, PHPCS clean.

The one thing I'd want tracked before this goes out is inline below — flipping `is_archive` off also drops a content-restriction carve-out that was only holding because those flags were wrong. It doesn't block the fix, but it deserves a release note.

The remaining notes are small and non-blocking.

- `plugins/newspack-multibranded-site/includes/class-taxonomy.php:153` — **Suggestion:** Brand resolution now runs through the post cascade — `get_queried_object()` now returns the cover page `WP_Post` instead of the brand `WP_Term` — which is exactly why the old `body_class()` had to strip a term-id `page-id-` class. `determine_current_brand()` therefore stops short-circuiting on the `WP_Term` and falls through to `get_current_brand_for_post()`: single assigned brand term, then `_primary_brand` meta, then the cover-page option lookup, then category-derived. In the normal case that resolves correctly, and the runtime check confirms it (`newspack-brand-test-brand` was present). Two configurations diverge, though: a cover page that is itself tagged into a *different* brand, or carries a `_primary_brand` pointing elsewhere, resolves to the wrong brand; and a cover page with no brand terms plus a stale `newspack_multibranded_site_front_pages` option resolves to no brand at all — that option is only maintained by `added_term_meta`/`updated_term_meta`, with nothing backfilling it or handling `delete_term_meta`. Either silently drives logo, menus, theme colors, blogname and prompt targeting. The term path made both impossible. `pre_get_posts()` already holds the authoritative `$term`, so stashing it and having `Taxonomy::get_current()` prefer it when the query was filtered would close this cheaply. At minimum, a test that creates a cover page for brand A while assigning it to brand B and asserts `get_current()` returns A would pin the behaviour. Reasonable to defer to a follow-up for a hotfix.
- `plugins/newspack-plugin/includes/plugins/wc-memberships/class-memberships.php:748` — **Suggestion:** Brand front pages lose their content-restriction carve-out — `remove_unnecessary_content_restriction()` is gated on `( is_front_page() || is_archive() )` and hooked to `wp` at priority 11 — after the flags are set. A brand front page used to match via `is_archive()`; it now matches neither, since `is_front_page()` compares against the site's own `page_on_front`. So WC Memberships' `restrict_post`, `handle_restricted_post_content_filtering` and `display_restricted_taxonomy_term_notice` stay attached where Newspack previously stripped them. Concretely, on a site with Memberships active: `restrict_post` has no main-query guard, so a membership-restricted post surfaced by a Homepage Posts or Carousel block on the cover page has its `post_content` **and** `post_excerpt` replaced by the restriction message. The page also pays the restriction-evaluation cost this function's own docblock says it exists to avoid. Worth being precise about the bound, because the first framing of this was broader than the code supports. The cover page itself is **not** newly restricted unless the publisher restricted it, and Newspack's own `wc_memberships_excerpt()` still bails for looped posts since their IDs differ from the queried page. What's actually lost is the front-page carve-out and its performance benefit. It needs Memberships (a paid extension, not part of a default install), a brand cover page, and a restricted post in a block loop on it. There's a parallel path worth knowing about but not acting on: `Content_Gate::restrict_post()` bails on `get_queried_object_id() !== $post->ID`, and since the queried object is now the cover page rather than the brand term, that equality holds for the first time. It's inert in practice — the feature is behind `NEWSPACK_CONTENT_GATES`, a draft constant defined nowhere in the product, the default content rule is `post_types => ['post']` which never matches a page, and a gate with no content rules is skipped outright. It also bails whenever Memberships is active, so the two paths can never both fire on one site. Not a reason to hold the fix — this is render-time, nothing is written, and a revert restores it instantly, whereas blocking leaves the excerpt bug live for every multibrand publisher. But it belongs in the release note, with a follow-up to close it.
- `themes/newspack-theme/newspack-theme/inc/template-functions.php:206` — **Suggestion:** Sidebar arm changes on brand front pages — The `has-sidebar` / `no-sidebar` computation has two arms: one for `! is_archive()`, one for `( is_archive() || is_home() ) && 'default' === archive_layout`. A brand front page used to take the second arm, so on a site with a **non-default** `archive_layout` and an active `sidebar-1` it resolved to `no-sidebar`. It now takes the first arm and gets `has-sidebar`. Arguably the correct outcome now that it genuinely is a page, but it's a visible layout change on the brand homepage for a real and not-uncommon customizer configuration — the kind of thing a publisher notices immediately and files as a regression. Better to name it in the release note than to have it discovered in the wild.
- `plugins/newspack-multibranded-site/includes/class-taxonomy.php:142` — **Nit:** Two workarounds for the old flags are now dead code — The `$post_or_post_id instanceof WP_Term` guard in `get_current_brand_for_post()` — commented "Account for Brands with page on front" — is no longer reachable from `determine_current_brand()`. And in the theme, `newspack_should_skip_archive_layout()`'s multibrand branch can never fire, since the function returns early on `! is_archive() && ! is_home()`. Both are harmless, but the theme helper is documented as compensating for "pages that technically pass `is_archive()` checks", so leaving it makes the next reader think the flags are still wrong. Agreed this should **not** be cleaned up in this PR: the theme ships separately from the plugin, so removing it now would open a version-skew window (old plugin + new theme). A follow-up issue is the right call.
- `themes/newspack-theme/newspack-theme/inc/newspack-sponsors.php:26` — **Nit:** Sponsor-label script stops enqueuing on brand front pages — The AMP-fallback sponsors script is enqueued only when `is_single() || is_archive()`. Brand front pages qualified via `is_archive()` and now qualify via neither, so the "Learn More" / close-info interaction on sponsor labels stops working there if the page surfaces sponsored posts. The same gap already exists for ordinary pages and for the site's own front page, so this is really a consistency issue rather than a new class of bug — but it is a concrete small regression introduced by this change, and cheap to fold into the same follow-up as the other conditional-tag consumers.

chickenn00dle (approved):

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
rbcorrales: The new `$filtered_brand_id` (348a666f8) shares this exact lifecycle on purpose, so the brand stash and `is_filtered()` stay consistent.

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
rbcorrales: Checking the reporter's cover page isn't `page_for_posts` before we close NPPM-2993; I'll note the result there.

plugins/newspack-multibranded-site/includes/customizations/class-show-page-on-front.php
rbcorrales: Leaving as-is: core emits `page-template-<custom>` from the page's own `_wp_page_template` even on the real front page while `front-page.php` renders, so this is core parity. Re-adding a forced `page-template-default` would resume the old class surgery.

plugins/newspack-multibranded-site/tests/unit-tests/test-customization-page-on-front.php
rbcorrales: Added in 348a666f8 (`test_ordinary_page_does_not_get_front_page_class`), confirmed load-bearing: it fails if the `is_filtered()` guard is removed.

rbcorrales:
Thanks for the thorough pass. Status on the out-of-diff notes:

- Brand resolution: fixed in 348a666f8. `Show_Page_On_Front` now stashes the displayed brand and `determine_current_brand()` prefers it, so the displayed brand wins even when the cover page is tagged into another brand. Added `test_current_brand_prefers_the_front_page_brand` (red before the fix, green after).
- Memberships carve-out, sponsor-label enqueue, the dead-code cleanup, and a release note for the publisher-visible knock-ons (sidebar, sponsor labels, Memberships): tracked in NPPM-3108, since they touch `newspack-plugin` and `newspack-theme` (separate release units).
- Blog-page-as-front-page: verifying the reporter's cover page isn't `page_for_posts` before closing NPPM-2993.

PR #667, opened 2026-07-20
feat(newspack-theme): add per-post featured image caption

eddiesshop:
## What

Adds support for a **per-post featured image caption** — editors can set a caption that applies to a single post, overriding the featured image's default caption/credit without changing the underlying attachment.

## Changes

**Editor (`js/src/extend-featured-image-editor.js`)**
- The "Featured Image Caption" field now only renders once a featured image has been selected. Visibility is gated on the live `featured_media` post attribute, so the field appears/disappears immediately as the featured image is set or removed.

**Front end (`inc/template-tags.php`)**
- Fixed the caption not displaying on the published post. The `newspack_featured_image_caption` meta is registered on the `post` type and saved to the post, but `newspack_post_thumbnail_caption()` was reading it from the **attachment** ID (`get_post_thumbnail_id()`), which always returned empty. It now reads from the post (`get_the_ID()`).
- Fallback order is preserved: custom per-post caption → the image's own caption/excerpt → Newspack image credit.

**Meta registration (`functions.php`)**
- Registers the `newspack_featured_image_caption` post meta (`show_in_rest`, single, string) and wires the editor script to its generated `.asset.php` dependencies/version.

## Testing

1. Edit a post; confirm the caption field is hidden until a featured image is set.
2. Set a featured image, enter a custom caption, and save.
3. View the post and confirm the custom caption renders in the `<figcaption>`.
4. Clear the caption and confirm it falls back to the image's default caption, then to the image credit.

### DIFF
--- a/themes/newspack-theme/newspack-theme/functions.php
+++ b/themes/newspack-theme/newspack-theme/functions.php
@@ -587,12 +587,13 @@ function newspack_enqueue_scripts() {
 	}
 
 	// Featured Image options.
+	$extend_featured_image_asset = require get_theme_file_path( '/js/dist/extend-featured-image-editor.asset.php' );
 	// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NotInFooter -- TODO: Should we set $in_footer?
 	wp_register_script(
 		'newspack-extend-featured-image-script',
 		get_theme_file_uri( '/js/dist/extend-featured-image-editor.js' ),
-		array( 'wp-blocks', 'wp-components' ),
-		$theme_version,
+		$extend_featured_image_asset['dependencies'],
+		$extend_featured_image_asset['version'],
 		true
 	);
 	wp_set_script_translations( 'newspack-extend-featured-image-script', 'newspack-theme', $languages_path );
@@ -924,6 +925,16 @@ function newspack_register_meta() {
 			)
 		);
 	}
+
+	register_post_meta(
+		'post',
+		'newspack_featured_image_caption',
+		array(
+			'show_in_rest' => true,
+			'single'       => true,
+			'type'         => 'string',
+		)
+	);
 
 	register_post_meta(
 		'post',
--- a/themes/newspack-theme/newspack-theme/functions.php
+++ b/themes/newspack-theme/newspack-theme/functions.php
@@ -928,6 +928,17 @@ function newspack_register_meta() {
 
 	register_post_meta(
 		'post',
+		'newspack_featured_image_caption_enabled',
+		array(
+			'show_in_rest' => true,
+			'single'       => true,
+			'type'         => 'boolean',
+			'default'      => false,
+		)
+	);
+
+	register_post_meta(
+		'post',
 		'newspack_featured_image_caption',
 		array(
 			'show_in_rest' => true,
--- a/themes/newspack-theme/newspack-theme/inc/template-tags.php
+++ b/themes/newspack-theme/newspack-theme/inc/template-tags.php
@@ -466,25 +466,25 @@ if ( ! function_exists( 'newspack_post_thumbnail_caption' ) ) {
 			return;
 		}
 
-		// Check the existance of the caption separately, so filters -- like ones that add ads -- don't interfere.
-		$thumbnail      = get_post( get_post_thumbnail_id() );
-		$caption_exists = $thumbnail && $thumbnail->post_excerpt;
+		// Check the existence of the caption separately, so filters -- like ones that add ads -- don't interfere.
+		$thumbnail = get_post( get_post_thumbnail_id() );
+		// Check for a per-post custom caption first. This meta is stored on the post
+		// (see newspack_register_meta()), not on the attachment.
+		$caption = get_post_meta( get_the_ID(), 'newspack_featured_image_caption', true );
 
-		// Only get the caption if one exists.
-		if ( $caption_exists ) {
+		if ( ! $caption && $thumbnail && $thumbnail->post_excerpt ) {
 			$caption = get_the_excerpt( get_post_thumbnail_id() );
 		}
 
 		// Account for featured images that have a credit but no caption.
-		if ( ! $caption_exists && class_exists( '\Newspack\Newspack_Image_Credits' ) ) {
+		if ( empty( $caption ) && class_exists( '\Newspack\Newspack_Image_Credits' ) ) {
 			$maybe_newspack_image_credit = \Newspack\Newspack_Image_Credits::get_media_credit_string( get_post_thumbnail_id() );
 			if ( strlen( wp_strip_all_tags( $maybe_newspack_image_credit ) ) ) {
-				$caption        = $maybe_newspack_image_credit;
-				$caption_exists = true;
+				$caption = $maybe_newspack_image_credit;
 			}
 		}
 
-		if ( $caption_exists ) :
+		if ( ! empty( $caption ) ) :
 			?>
 			<figcaption><span><?php echo wp_kses_post( $caption ); ?></span></figcaption>
 			<?php
--- a/themes/newspack-theme/newspack-theme/inc/template-tags.php
+++ b/themes/newspack-theme/newspack-theme/inc/template-tags.php
@@ -468,9 +468,13 @@ if ( ! function_exists( 'newspack_post_thumbnail_caption' ) ) {
 
 		// Check the existence of the caption separately, so filters -- like ones that add ads -- don't interfere.
 		$thumbnail = get_post( get_post_thumbnail_id() );
-		// Check for a per-post custom caption first. This meta is stored on the post
-		// (see newspack_register_meta()), not on the attachment.
-		$caption = get_post_meta( get_the_ID(), 'newspack_featured_image_caption', true );
+		$caption   = '';
+
+		// Use the per-post custom caption only when the editor has enabled the override.
+		// Both meta are stored on the post (see newspack_register_meta()), not the attachment.
+		if ( get_post_meta( get_the_ID(), 'newspack_featured_image_caption_enabled', true ) ) {
+			$caption = get_post_meta( get_the_ID(), 'newspack_featured_image_caption', true );
+		}
 
 		if ( ! $caption && $thumbnail && $thumbnail->post_excerpt ) {
 			$caption = get_the_excerpt( get_post_thumbnail_id() );
--- a/themes/newspack-theme/newspack-theme/js/src/extend-featured-image-editor.js
+++ b/themes/newspack-theme/newspack-theme/js/src/extend-featured-image-editor.js
@@ -1,7 +1,7 @@
 'use strict';
 
 import { addFilter } from '@wordpress/hooks';
-import { RadioControl, TextControl } from '@wordpress/components';
+import { RadioControl, TextControl, ToggleControl } from '@wordpress/components';
 import { withDispatch, withSelect, select } from '@wordpress/data';
 import { Component, Fragment } from '@wordpress/element';
 import { compose } from '@wordpress/compose';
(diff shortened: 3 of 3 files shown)


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/feat/per-post-featured-image-caption` (stash local work first).

themes/newspack-theme/newspack-theme/functions.php
```diff
@@ -925,6 +926,27 @@ function newspack_register_meta() {
 		);
 	}
 
+	register_post_meta(
```
chickenn00dle, line 929: **Blocker:** Caption UI appears on post types the meta isn't registered for — captions are silently discarded — Both new metas are registered flat against `'post'`, but two lines above, the sibling `newspack_featured_image_position` is registered inside `foreach ( $featured_image_post_types as $post_type )` over `newspack_get_featured_image_post_types()`.

The editor control follows the *loop's* scope, not this one. `wrapPostFeaturedImage` renders `<CaptionControl />` alongside the position radio for any post type in that same localized list, and the only early return inside `CaptionControl` is `if ( ! featuredMediaId ) return null;` — there's no post-type check anywhere in it. (The sibling subtitle and summary scripts *are* wrapped in `if ( 'post' === $post_type )` at line 608, which suggests this was an oversight rather than a decision.)

That list is `array( 'post', 'page' )` by default and filterable — and both **newspack-newsletters** (`class-newspack-newsletters.php:1509`) and **newspack-listings** (`class-core.php:760`) append their CPTs to it. So on a stock Newspack site the caption UI also appears on newsletters and every listings type.

On all of those the value is silently lost. `WP_REST_Meta_Fields::update_value()` iterates only registered fields and tests `array_key_exists( $name, $meta )`, so an unregistered key in the payload is never visited and no branch can error. The meta argument is validated solely by `check_meta_is_array`, its `sanitize_callback` is null, and the object schema sets no `additionalProperties`, so unknown properties aren't rejected. The request returns 200, the editor's reducer clears the pending edit, and the field empties on reload with no notice.

The front end is already correct for these types — `page.php:23` routes into `large-featured-image.php`, which calls `newspack_post_thumbnail_caption()`, and `get_post_meta()` doesn't require registration — so the caption would render properly if it could be saved.

Moving both `register_post_meta()` calls into the existing `foreach` and passing `$post_type` fixes it in about four lines. Worth updating the `newspack_register_meta()` docblock too, which still lists only position/subtitle/summary.

themes/newspack-theme/newspack-theme/functions.php
```diff
@@ -587,12 +587,13 @@ function newspack_enqueue_scripts() {
 	}
 
 	// Featured Image options.
+	$extend_featured_image_asset = require get_theme_file_path( '/js/dist/extend-featured-image-editor.asset.php' );
```
chickenn00dle, line 590: **Suggestion:** Unguarded `require` of a gitignored build artifact — Switching to the generated `.asset.php` is the right call — the new JS imports `wp-core-data`, which the old hardcoded `array( 'wp-blocks', 'wp-components' )` omitted, so the dependency list was genuinely wrong.

But `js/dist/` is gitignored (`.gitignore:35`), so this file only exists after a build, and a bare `require` on a missing file is a fatal rather than a warning. Production is fine — release archives are built before packaging — and the blast radius is bounded to the block editor, since this runs on `enqueue_block_editor_assets`. The exposure is an unbuilt checkout or an env that activates the theme before building, where a stale-script problem becomes an editor that won't load.

The monorepo already has the defensive shape: `Newspack::asset_version()` does `file_exists()` → include → fall back, with a comment noting the check is load-bearing. A three-line guard falling back to the previous hardcoded deps keeps the win without the failure mode. Worth doing here specifically because this is the theme's first `.asset.php` consumer, so it sets the pattern for the next one.

themes/newspack-theme/newspack-theme/functions.php
```diff
@@ -925,6 +926,27 @@ function newspack_register_meta() {
...
+		'newspack_featured_image_caption_enabled',
+		array(
+			'show_in_rest' => true,
+			'single'       => true,
+			'type'         => 'boolean',
+			'default'      => false,
+		)
+	);
+
+	register_post_meta(
```
chickenn00dle, line 940: **Suggestion:** Add a `sanitize_callback` — post meta gets no `kses` pass on save — The write permission here is fine: with an unprefixed key and no `auth_callback`, `register_meta()` defaults to `__return_true` and `edit_post_meta` maps through to `edit_post`, so it's "anyone who can edit this post". That's correct.

The gap is on the read side. Each registered meta property's schema carries no `context` key, and `rest_filter_response_by_context()` skips properties without one — so the raw stored string is returned verbatim by an unauthenticated `GET /wp/v2/posts/<id>` on a published post.

The asymmetry that makes this worth closing: `kses_init_filters()` hooks `title_save_pre`, `content_save_pre`, `excerpt_save_pre` and the comment filters — **never post meta**. So a Contributor who lacks `unfiltered_html`, and therefore can't put a `<script>` tag in post content, *can* put one in this meta.

No Newspack consumer currently re-renders it unescaped — this theme escapes at output, newspack-blocks' same-named field is attachment-derived and never reads this meta, and newspack-network's nodes re-render through the same escaped path. So this is defense-in-depth for headless and third-party consumers rather than a live XSS.

`'sanitize_callback' => 'wp_kses_post'` costs nothing, runs on direct `update_post_meta()` calls too, and makes storage agree with the render-time escaping — **which should stay as it is**. The only mangling edge case is a literal `1<2>3`, which `kses` would trim; post content already behaves that way for the same users.

themes/newspack-theme/newspack-theme/functions.php
```diff
@@ -925,6 +926,27 @@ function newspack_register_meta() {
...
+		'post',
+		'newspack_featured_image_caption',
```
chickenn00dle, line 942: **Suggestion:** Name collides with an unrelated newspack-blocks REST field — `newspack_featured_image_caption` is already a top-level REST field on the Homepage Posts / Carousel endpoint (`newspack-blocks/includes/class-newspack-blocks-api.php:277`), computed from the **attachment** via `Newspack_Blocks::get_image_caption()`.

They don't clash structurally — one is `meta.newspack_featured_image_caption`, the other a sibling field — so nothing breaks. But the product consequence is real: a post with a custom caption will show that caption on its single view and the *attachment's* caption in every Homepage Posts and Carousel card that features it.

Worth either deciding that's acceptable and saying so in the PR, or filing a follow-up to teach `get_image_caption()` about the override. At minimum the identical name across two components is a trap for whoever greps it next.

themes/newspack-theme/newspack-theme/functions.php
```diff
@@ -925,6 +926,27 @@ function newspack_register_meta() {
 		);
 	}
 
+	register_post_meta(
+		'post',
+		'newspack_featured_image_caption_enabled',
```
chickenn00dle, line 931: **Nit:** The enable toggle isn't mentioned in the PR description — The branch has two commits, and the body narrates only the first. The second adds `newspack_featured_image_caption_enabled` and gates the entire feature behind it.

That's a reasonable design — it mirrors the Custom Byline pattern — but it's a UX decision reviewers are currently being asked to approve without seeing it described. Worth a line in the description.

While there: the body also frames the attachment-vs-post meta read as a bug in shipped code. That code only ever existed on an unmerged branch, so no live site has a stored caption and nothing needs migrating — but as written, the description implies existing captions are about to start appearing, which would be a much bigger deal than what this actually does.

themes/newspack-theme/newspack-theme/inc/template-tags.php
```diff
@@ -466,25 +466,29 @@ function newspack_post_thumbnail_caption() {
...
+
+		// Use the per-post custom caption only when the editor has enabled the override.
+		// Both meta are stored on the post (see newspack_register_meta()), not the attachment.
+		if ( get_post_meta( get_the_ID(), 'newspack_featured_image_caption_enabled', true ) ) {
+			$caption = get_post_meta( get_the_ID(), 'newspack_featured_image_caption', true );
+		}
 
-		// Only get the caption if one exists.
-		if ( $caption_exists ) {
+		if ( ! $caption && $thumbnail && $thumbnail->post_excerpt ) {
```
chickenn00dle, line 479: **Nit:** A caption of `"0"` falls through — The refactor swapped the `$caption_exists` boolean for `$caption` truthiness, so a custom caption of the literal string `"0"` is falsy and falls through to the attachment caption, then the credit, then renders nothing.

The old code had the same blind spot on the attachment path, so this isn't a regression — but the new user-editable field introduces a fresh place to hit it, and `'' !== $caption` closes it. Vanishingly rare; noting for completeness.

themes/newspack-theme/newspack-theme/inc/template-tags.php
```diff
@@ -466,25 +466,29 @@ function newspack_post_thumbnail_caption() {
...
 			$caption = get_the_excerpt( get_post_thumbnail_id() );
 		}
 
 		// Account for featured images that have a credit but no caption.
-		if ( ! $caption_exists && class_exists( '\Newspack\Newspack_Image_Credits' ) ) {
+		if ( empty( $caption ) && class_exists( '\Newspack\Newspack_Image_Credits' ) ) {
 			$maybe_newspack_image_credit = \Newspack\Newspack_Image_Credits::get_media_credit_string( get_post_thumbnail_id() );
```
chickenn00dle, line 485: **Nit:** An unrelated behaviour improvement worth naming as intentional — Previously, an attachment whose `post_excerpt` was set but whose *filtered* `get_the_excerpt()` came back empty produced an empty `<figcaption><span></span></figcaption>` and never tried the credit rung. With the flag replaced by caption truthiness, it now falls through to the credit and emits nothing if there isn't one.

That's better, and it matches the verified "no empty `<figcaption>`" result — but it's a behaviour change nobody asked for, so it's worth a line in the description confirming it's intentional rather than incidental.

chickenn00dle (changes requested):
**Release risk: HIGH** — one blocker, and the fix is about four lines.

The rendering side is solid, and I exercised it rather than reading it: all six branches of the fallback chain behave correctly in an isolated env — the enabled custom caption wins, it falls back to the attachment caption, then to the image credit, a post with nothing anywhere emits **no empty `<figcaption>`**, a caption with no featured image renders nothing and doesn't error, and a caption with the toggle off correctly falls through. Those last three are the ones this kind of change usually gets wrong. The `template-tags.php` refactor also preserves the original intent: `get_the_excerpt()` is still only reached after a direct `post_excerpt` check, so filters that inject ads still can't manufacture a caption out of nothing.

**The blocker is a scope mismatch between the UI and the meta registration.** Both new metas are registered flat against `post`. But the editor control isn't — `wrapPostFeaturedImage` renders it for every post type in `newspack_get_featured_image_post_types()`, and the only guard inside `CaptionControl` is `if ( ! featuredMediaId ) return null;`. That list is `post` and `page` by default and is filterable — newspack-newsletters and newspack-listings both add their own CPTs to it, so on a stock Newspack site the caption UI also appears on newsletters and every listings type.

On any of those, the save silently discards the value. `WP_REST_Meta_Fields::update_value()` iterates only the *registered* fields and checks `array_key_exists` against the payload, so an unregistered key is never visited and no branch can raise an error; the meta argument is validated only by `check_meta_is_array`, and the schema declares no `additionalProperties`, so unknown keys aren't rejected either. The request returns 200, the editor's reducer clears the pending edit, and the field just empties on reload. Worth noting the front end is already correct for these types — `page.php` routes into `large-featured-image.php`, which calls the caption function, and `get_post_meta()` needs no registration — so the value would render fine if it could be saved.

Registering both metas inside the `foreach ( $featured_image_post_types as $post_type )` block that `newspack_featured_image_position` already uses, two lines above, fixes it. (Alternatively gate the control to `post` in the JS — but the sibling meta's shape suggests the loop is the intent.)

The remaining points are non-blocking and inline, except one that has no line to attach to: **both keys should be added to `newspack_get_editor_managed_meta_keys()`**. That list exists precisely so the classic Custom Fields box's `meta-box-loader` save doesn't resubmit a stale page-load value over the REST save, and every other panel-edited meta in this theme is in it. Both new keys are unprefixed, so they show up in that box once a row exists. It's narrower than it first looks — the panel is off by default, and because the resubmit works by `meta_id` only rows that existed at page load are affected, so the *first* toggle-on is safe and the clobber bites on a later edit in the same session. But the guard exists because publishers hit this for real with the sibling metas, and it's two strings to add.

One note for the PR description rather than the code: the branch has a second commit that adds the enable toggle gating the whole feature, and the body doesn't mention it. That's a UX decision reviewers are currently being asked to approve without seeing it described.

PR #671, opened 2026-07-20
fix(content-gate): keep anonymous metering tied to registered access

dkoo:
Fixes [NEWS-2785](https://linear.app/a8c/issue/NEWS-2785/access-control-registered-access-enabled-w-metering-off-paid-access)

### The bug

A content gate has two access rules: **registered access** (the registration wall, which governs anonymous readers) and **paid access** (the paywall). Each has its own optional metering.

#4571 added a fallback so that when registered access is switched off entirely, anonymous readers fall through to the paywall and use *its* metering settings. It keyed that fallback on `$anonymous_settings['enabled']`:

```php
$settings = Memberships::is_active() || $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
```

But `get_anonymous_settings()` returns `enabled = $registration['active'] && $registration['metering']['enabled']`. So a publisher with registered access **active** and its metering deliberately **off** is indistinguishable from one with registered access **disabled** — and hits the fallback too. Anonymous readers borrow the paywall's meter (3 free articles) and then see the paywall, instead of hitting the registration wall immediately. The registration gate layout is never shown.

The same predicate was copy-pasted at four call sites, so they could drift apart independently.

### The fix

A private helper `Metering::get_effective_anonymous_settings()` keyed on `$registration['active']` — the question #4571 meant to ask. All four call sites route through it.

Two adjacent corrections the change surfaced:

- **`get_metering_period()`** is now logged-in-aware. It previously returned the *registration* period to logged-in readers, whose metering expiry is computed from the *paid* period (`is_logged_in_metering_allowed()` uses `get_registered_settings()['period']`) — so the countdown banner could say "free articles this week" against a monthly reset.
- **`get_total_metered_views()`** now returns `false` when metering is disabled, as its docblock already promised and as `print_cta()` already handles. It previously returned whatever count was still stored; since `get_registration_settings()` defaults `count` to `1`, a gate whose meter was toggled off in the UI would report one free view.

### Unverified logged-in readers

The same misalignment existed one step over. `Content_Restriction_Control::is_post_restricted()` restricts a signed-in reader who has **not verified their email** on a wall that requires verification using the *registration* layout, and skips the `custom_access` rules entirely — yet the metering surfaces still reported the paywall's allowance for that reader. On a gate with registered access requiring verification and metering off, plus paid access metering on, `get_total_metered_views( true )` returned the paid count for a reader who actually gets zero, so the countdown could advertise a paid allowance on the registration wall.

So the helper is generalized: `get_effective_settings( $gate_id, $is_logged_in )`, backed by `is_gated_by_registration()`, which mirrors that layout selection — the registration wall governs anonymous visitors *and* unverified signed-in readers. Non-reader users (admins, editors) stay exempt, matching the existing carve-out in `is_logged_in_metering_allowed()`.

Access behavior for these readers was already correct (`is_logged_in_metering_allowed()` already hard-bailed on unverified readers, so they already saw the wall immediately); what this corrects is which settings the metering surfaces report. Unchanged: an unverified reader still gets no metered views at all, even where registration metering is on — granting them the registration allowance would be a loosening of access and is a separate decision.

`get_total_metered_views()` returning `false` reaches the block editor via `class-blocks.php`, which localizes it for the countdown block's preview (`parseInt( false )` is `NaN`), so those two values are now cast with `absint()` at the localization boundary. Both PHP consumers already guarded on `false === $total_views`, and a grep of the standalone repos (including `newspack-manager`) found no external callers of the changed methods.

## How to test

Automated (both new regression tests fail on `release` and pass here):

```
n test-php   # from plugins/newspack-plugin
```

Manually, on a gated post:

1. Create a content gate with **registered access active, metering OFF** and **paid access active, metering ON, count 3**.
2. Visit a gated post logged out. **Expected:** the registration gate immediately, no metered articles. (Before: 3 free articles, then the paywall.)
3. Log in as a reader with no paid access. **Expected:** 3 metered articles, then the paywall.
4. Turn on **Require verification** under registered access, then load a gated post as a signed-in reader whose email is not verified. **Expected:** the registration gate immediately, and no countdown advertising the paid allowance.
5. Flip registered access to **inactive** and reload logged out. **Expected:** 3 metered articles, then the paywall — #4571's behavior, unchanged.

Note when testing: `enqueue_scripts()` bakes the metering count and period into the cached HTML, so give the page cache a chance to turn over between config changes.

## Rollout note: this is a tightening

Exactly one gate configuration changes reader-facing behavior: **registration wall active + its metering off + paywall metering on**. On those sites, anonymous readers who have been getting free articles through this bug will start hitting the registration wall on the first article as soon as the release lands. That is the intended correction, but it is visible on day one and can read as a regression to anyone unaware of the old leak — and because campaign prompts and article-view activity are keyed off the same metering check, engagement and campaign reporting move at the same time.

Worth giving affected publishers advance notice. The remedy for anyone who wants to keep the free views is one toggle: enable Metering under the Registration rule and set a count there.

Every other configuration is unchanged, including all legacy WooCommerce Memberships gates and every signed-in reader path.

When validating after deploy, let the page cache turn over first — anonymous metered pages are cacheable, so pre-fix HTML keeps serving until expiry and a stale page looks like the fix did not land.

## Follow-ups (not in scope here)

- `Metering_Countdown::print_cta()` picks the "Create an account" vs "Sign in" link for anonymous readers from `get_registered_settings()` — the *paid* meter. `Content_Gifting_CTA` has the same shape via `has_metering()`, which ORs both meters. These are the remaining anonymous-facing readers of the wrong settings bucket.
- `class-blocks.php` localizes `anonymous_metered_views` and `loggedin_metered_views` side by side, but a single `metering_period` resolved from the editor's own session. Harmless today (that path is memberships-gated, where both meters share one stored period), a real bug if the localization is extended.

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-metering.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-metering.php
@@ -206,11 +206,17 @@ class Metering {
 	/**
 	 * Whether the registration wall is the rule that gates the current reader.
 	 *
-	 * Mirrors the gate-layout selection in
-	 * `Content_Restriction_Control::is_post_restricted()`, which restricts with the
-	 * registration layout - and skips the `custom_access` rules entirely - for both
-	 * anonymous visitors and signed-in readers who have not verified their email on a
-	 * wall that requires verification.
+	 * Follows the same registration-vs-`custom_access` split as the gate-layout
+	 * selection in `Content_Restriction_Control::is_post_restricted()`, which restricts
+	 * with the registration layout - and skips the `custom_access` rules entirely - for
+	 * both anonymous visitors and signed-in readers who have not verified their email on
+	 * a wall that requires verification.
+	 *
+	 * This assumes the reader is already known to be restricted; it is a settings
+	 * selector, not a restriction check. In particular it does not re-check the
+	 * anonymous `supports_anonymous` bypass (e.g. institutional access) that
+	 * `is_post_restricted()` applies before ever restricting an anonymous visitor - a
+	 * bypassed reader is not restricted, so no metering surface consults this for them.
 	 *
 	 * @param int  $gate_id      Gate ID.
 	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
@@ -228,8 +234,11 @@ class Metering {
 		if ( ! $registration['require_verification'] ) {
 			return false;
 		}
-		// Non-reader users (admins, editors) are exempt, matching the same carve-out in
-		// `is_logged_in_metering_allowed()`.
+		// Exempt non-reader users (admins, editors). This reuses the reader/verified
+		// checks from `is_logged_in_metering_allowed()`'s verification bail; note that
+		// bail keys off `Content_Gate::requires_account_verification()`, which reads
+		// `require_verification` without the `active` guard applied above, so the two
+		// only fully agree while the wall is active.
 		$user = \wp_get_current_user();
 		return Reader_Activation::is_user_reader( $user ) && ! Reader_Activation::is_reader_verified( $user );
 	}
@@ -253,7 +262,7 @@ class Metering {
 	 * @param int  $gate_id      Gate ID.
 	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
 	 *
-	 * @return array Metering settings.
+	 * @return array{enabled: bool, count: int, period: string} Metering settings.
 	 */
 	private static function get_effective_settings( $gate_id, $is_logged_in ) {
 		if ( Memberships::is_active() ) {
@@ -437,7 +446,13 @@ class Metering {
 		}
 
 		$gate_post_id = Content_Gate::get_gate_post_id();
-		$settings     = self::get_registered_settings( $gate_post_id );
+		// Read through the same helper as every reporting surface, so the enablement
+		// check here can never advertise a different allowance than they display. For
+		// every state that reaches this line the reader is either verified, a
+		// non-reader, or on a gate without a verification requirement - the verification
+		// bail above has already returned for the one state where the registration wall
+		// would govern - so this is behavior-identical to a direct paid-settings read.
+		$settings = self::get_effective_settings( $gate_post_id, true );
 
 		// Bail if metering is not enabled.
 		if ( ! $settings['enabled'] || $settings['count'] <= 0 ) {
@@ -537,7 +552,12 @@ class Metering {
 	}
 
 	/**
-	 * Get the metering period for a post.
+	 * Get the metering period for the current reader on a post.
+	 *
+	 * Resolves against the settings that govern the current reader, so the result is
+	 * auth-state dependent (via `is_user_logged_in()`, and verification state for
+	 * logged-in readers) - not a pure per-post value. Callers caching or comparing
+	 * periods across readers should key on the reader as well as the post.
 	 *
 	 * @param int|null $post_id Post ID. Default is current post.
 	 *
--- a/plugins/newspack-plugin/includes/content-gate/class-metering.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-metering.php
@@ -204,6 +204,31 @@ class Metering {
 	}
 
 	/**
+	 * Get the metering settings that govern anonymous readers.
+	 *
+	 * Beware the two senses of "registered" in this class: `get_anonymous_settings()`
+	 * reads the `registration` meta (the registration wall), while
+	 * `get_registered_settings()` reads the `custom_access` meta (the paywall).
+	 *
+	 * Anonymous readers are gated by the registration wall, so the `registration`
+	 * settings apply whenever that wall is active - including when its metering is
+	 * deliberately turned off, which means anonymous readers get no metered views at
+	 * all. Only when the registration wall is inactive do anonymous readers fall
+	 * through to the paywall, and with it the `custom_access` metering settings.
+	 *
+	 * @param int $gate_id Gate ID.
+	 *
+	 * @return array Metering settings.
+	 */
+	private static function get_effective_anonymous_settings( $gate_id ) {
+		if ( Memberships::is_active() ) {
+			return self::get_anonymous_settings( $gate_id );
+		}
+		$registration = Content_Gate::get_registration_settings( $gate_id );
+		return $registration['active'] ? self::get_anonymous_settings( $gate_id ) : self::get_registered_settings( $gate_id );
+	}
+
+	/**
 	 * Update metering settings for a gate.
 	 *
 	 * @param int   $gate_id  Gate ID.
@@ -230,7 +255,7 @@ class Metering {
 		}
 		$gate_layout_id = Content_Gate::get_gate_layout_id();
 		$gate_post_id   = Content_Gate::get_gate_post_id();
-		$handle       = 'newspack-content-gate-metering';
+		$handle         = 'newspack-content-gate-metering';
 		\wp_enqueue_script(
 			$handle,
 			Newspack::plugin_url() . '/dist/content-gate-metering.js',
@@ -239,9 +264,7 @@ class Metering {
 			true
 		);
 
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		$settings = $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
+		$settings = self::get_effective_anonymous_settings( $gate_post_id );
 		\wp_localize_script(
 			$handle,
 			'newspack_metering_settings',
@@ -329,9 +352,7 @@ class Metering {
 		}
 
 		$gate_post_id         = Content_Gate::get_gate_post_id();
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		$settings             = Memberships::is_active() || $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
+		$settings             = self::get_effective_anonymous_settings( $gate_post_id );
 		$is_frontend_metering = $settings['enabled'] && $settings['count'] > 0;
 
 		/**
@@ -490,10 +511,8 @@ class Metering {
 		if ( ! $post_id ) {
 			$post_id = get_the_ID();
 		}
-		$gate_post_id         = Content_Gate::get_gate_post_id( $post_id );
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		$settings = $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
+		$gate_post_id = Content_Gate::get_gate_post_id( $post_id );
+		$settings     = \is_user_logged_in() ? self::get_registered_settings( $gate_post_id ) : self::get_effective_anonymous_settings( $gate_post_id );
 		return $settings['period'];
 	}
 
@@ -528,12 +547,11 @@ class Metering {
 		if ( ! $gate_post_id ) {
 			return false;
 		}
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		if ( ! $is_logged_in && $anonymous_settings['enabled'] ) {
-			return $anonymous_settings['count'];
+		$settings = $is_logged_in ? self::get_registered_settings( $gate_post_id ) : self::get_effective_anonymous_settings( $gate_post_id );
+		if ( ! $settings['enabled'] ) {
+			return false;
 		}
-		return $registered_settings['count'];
+		return $settings['count'];
 	}
 }
 Metering::init();
--- a/plugins/newspack-plugin/includes/content-gate/class-metering.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-metering.php
@@ -216,6 +216,10 @@ class Metering {
 	 * all. Only when the registration wall is inactive do anonymous readers fall
 	 * through to the paywall, and with it the `custom_access` metering settings.
 	 *
+	 * Legacy Woo Memberships gates are exempt: they have no `registration` meta at all
+	 * and read both meters from the shared `metering` meta, so they short-circuit past
+	 * the `active` check.
+	 *
 	 * @param int $gate_id Gate ID.
 	 *
 	 * @return array Metering settings.
--- a/plugins/newspack-plugin/includes/content-gate/class-metering.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-metering.php
@@ -204,32 +204,64 @@ class Metering {
 	}
 
 	/**
-	 * Get the metering settings that govern anonymous readers.
+	 * Whether the registration wall is the rule that gates the current reader.
+	 *
+	 * Mirrors the gate-layout selection in
+	 * `Content_Restriction_Control::is_post_restricted()`, which restricts with the
+	 * registration layout - and skips the `custom_access` rules entirely - for both
+	 * anonymous visitors and signed-in readers who have not verified their email on a
+	 * wall that requires verification.
+	 *
+	 * @param int  $gate_id      Gate ID.
+	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
+	 *
+	 * @return bool Whether the registration wall governs the reader.
+	 */
+	private static function is_gated_by_registration( $gate_id, $is_logged_in ) {
+		$registration = Content_Gate::get_registration_settings( $gate_id );
+		if ( ! $registration['active'] ) {
+			return false;
+		}
+		if ( ! $is_logged_in ) {
+			return true;
+		}
+		if ( ! $registration['require_verification'] ) {
+			return false;
+		}
+		// Non-reader users (admins, editors) are exempt, matching the same carve-out in
+		// `is_logged_in_metering_allowed()`.
+		$user = \wp_get_current_user();
+		return Reader_Activation::is_user_reader( $user ) && ! Reader_Activation::is_reader_verified( $user );
+	}
+
+	/**
+	 * Get the metering settings that govern the current reader.
 	 *
 	 * Beware the two senses of "registered" in this class: `get_anonymous_settings()`
 	 * reads the `registration` meta (the registration wall), while
 	 * `get_registered_settings()` reads the `custom_access` meta (the paywall).
 	 *
-	 * Anonymous readers are gated by the registration wall, so the `registration`
-	 * settings apply whenever that wall is active - including when its metering is
-	 * deliberately turned off, which means anonymous readers get no metered views at
-	 * all. Only when the registration wall is inactive do anonymous readers fall
-	 * through to the paywall, and with it the `custom_access` metering settings.
+	 * A reader gated by the registration wall is metered by the wall's own settings -
+	 * including when its metering is deliberately turned off, which means no metered
+	 * views at all. Only readers who are past (or not subject to) the wall fall through
+	 * to the paywall, and with it the `custom_access` metering settings.
 	 *
 	 * Legacy Woo Memberships gates are exempt: they have no `registration` meta at all
-	 * and read both meters from the shared `metering` meta, so they short-circuit past
-	 * the `active` check.
+	 * and read both meters from the shared `metering` meta, so they keep the plain
+	 * anonymous/registered split.
 	 *
-	 * @param int $gate_id Gate ID.
+	 * @param int  $gate_id      Gate ID.
+	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
 	 *
 	 * @return array Metering settings.
 	 */
-	private static function get_effective_anonymous_settings( $gate_id ) {
+	private static function get_effective_settings( $gate_id, $is_logged_in ) {
 		if ( Memberships::is_active() ) {
-			return self::get_anonymous_settings( $gate_id );
+			return $is_logged_in ? self::get_registered_settings( $gate_id ) : self::get_anonymous_settings( $gate_id );
 		}
-		$registration = Content_Gate::get_registration_settings( $gate_id );
-		return $registration['active'] ? self::get_anonymous_settings( $gate_id ) : self::get_registered_settings( $gate_id );
+		return self::is_gated_by_registration( $gate_id, $is_logged_in )
+			? self::get_anonymous_settings( $gate_id )
+			: self::get_registered_settings( $gate_id );
 	}
 
 	/**
@@ -268,7 +300,7 @@ class Metering {
 			true
 		);
 
-		$settings = self::get_effective_anonymous_settings( $gate_post_id );
+		$settings = self::get_effective_settings( $gate_post_id, false );
 		\wp_localize_script(
 			$handle,
 			'newspack_metering_settings',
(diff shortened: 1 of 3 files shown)


plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -490,10 +515,8 @@ public static function get_metering_period( $post_id = null ) {
 		if ( ! $post_id ) {
 			$post_id = get_the_ID();
 		}
-		$gate_post_id         = Content_Gate::get_gate_post_id( $post_id );
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		$settings = $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
+		$gate_post_id = Content_Gate::get_gate_post_id( $post_id );
+		$settings     = \is_user_logged_in() ? self::get_registered_settings( $gate_post_id ) : self::get_effective_anonymous_settings( $gate_post_id );
```
dkoo: **Severity:** suggestion

The same wrong-rule-for-the-visible-gate bug survives one step over. A logged-in but *unverified* reader on a gate with `require_verification` is restricted by the **registration** layout (`class-content-restriction-control.php:382-387`), yet `is_logged_in_metering_allowed()` and this new logged-in branch both read the **paid** `custom_access` metering. So that reader is metered against the paywall allowance while looking at the registration wall — the same class of mix-up this PR fixes for anonymous readers.

Not introduced here and not worth holding a hotfix for, but the new `is_user_logged_in()` ternary cements the assumption, so it deserves a follow-up ticket rather than being forgotten.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -203,6 +203,35 @@ public static function get_registered_settings( $gate_id ) {
...
+	 *
+	 * @param int $gate_id Gate ID.
+	 *
+	 * @return array Metering settings.
+	 */
+	private static function get_effective_anonymous_settings( $gate_id ) {
+		if ( Memberships::is_active() ) {
+			return self::get_anonymous_settings( $gate_id );
+		}
+		$registration = Content_Gate::get_registration_settings( $gate_id );
```
dkoo: **Severity:** nit

This calls `Content_Gate::get_registration_settings( $gate_id )` and then, on the true branch, `get_anonymous_settings()` re-fetches and re-normalizes the same meta. Object-cached, so no extra query and nothing measurable — but it is the kind of double read the next reader trips over. Passing the already-fetched array down would read better.

**Consider:** also worth a follow-up — `get_total_metered_views()` takes an explicit `$is_logged_in` parameter while `get_metering_period()` infers it from `is_user_logged_in()`, and consumers call them side by side (`class-metering-countdown.php:216` passes the flag, then line 252 relies on inference). They agree today, but an optional param would make the pairing tamper-proof for any future REST or AJAX caller.

plugins/newspack-plugin/tests/unit-tests/content-gate/metering.php
```diff
@@ -613,4 +613,171 @@ function() use ( $gate_id ) {
...
+		$this->assertFalse( Metering::is_frontend_metering(), 'Front-end metering should not fall back to registered settings while registered access is active' );
+		$this->assertFalse( Metering::get_total_metered_views(), 'Anonymous readers should have no metered views' );
+	}
+
+	/**
+	 * Test that anonymous readers are still metered by registered access settings when
+	 * registered access is active and its metering is enabled.
+	 */
+	public function test_anonymous_metering_uses_registration_settings_when_enabled() {
+		// Both access rules active, each metering with a distinct count and period.
```
dkoo, line 683: **Severity:** suggestion

The `Memberships::is_active()` short-circuit in the new helper is load-bearing rather than defensive — legacy gates have no `registration` meta, so `get_registration_settings()` reports `active => false`, and without the guard every legacy site would divert to the paywall settings. Nothing covers it.

It is currently *untestable* rather than merely untested: `Memberships::is_active()` is `class_exists( 'WC_Memberships' ) && function_exists( 'wc_memberships' )`, and neither can be undone inside a PHPUnit process, so defining them would leak memberships-active state into every subsequent test in the run. Making this branch testable needs a short-circuit filter on `is_active()` — a change for a normal release, not a hotfix.

**Consider:** the anonymous-gate-context preamble (create post, rebuild `$wp_query`, filter `newspack_is_post_restricted`, set current user, filter `newspack_content_gate_post_id`) is now repeated four times in this file counting the pre-existing #4571 test. A `set_up_gate_context( $gate_id )` helper would make the one line that actually differs between tests visible instead of buried. Left alone here to keep the hotfix diff mechanical.

dkoo (commented):
Self-review before handing this to a human.

**Release risk: HIGH — and not because of a defect.** Static review found no blockers. The keying change is validated against `Content_Restriction_Control::is_post_restricted()`, which selects the registration layout for an anonymous visitor whenever `registration.active` is set and only falls to `custom_access` when it isn't — so `$registration['active']` exactly mirrors which gate an anonymous reader actually hits. #4571's scenario is preserved and its test still passes. The tier is driven by one thing: this is a `release`-targeted change to paywall behavior that was never exercised at runtime. Every behavior it changes is what an anonymous reader sees on a gated article, and no gated article was loaded in a browser. **Before merge, someone should configure a gate with the registration wall active + metering off and the paywall metering on, and confirm as an anonymous visitor that the wall appears on the first article.**

**Three issues were found and fixed during review**, so they are not in the comments below. The `false` return from `get_total_metered_views()` reaches the block editor through `class-blocks.php`, where `countdown/edit.jsx` does `parseInt( totalViews )` — `parseInt( false )` is `NaN`, so the editor preview would have rendered `0/NaN` on legacy Memberships sites. Fixed with `absint()` at the localization boundary. The new docblock also omitted the Memberships short-circuit, and the pinning test's fixture turned out to be load-bearing in a non-obvious way — flipping `custom_access` off would make it pass with the bug fully present — so it now says so in a comment.

**The `false` return type is contained.** Both PHP consumers already guard with a strict `false === $total_views` check before any arithmetic, and a grep of the standalone repos, including `newspack-manager` and `newspack-manager-admin`, found no external callers of the changed methods.

**Rollout consideration**, now written up in the PR description: this is a tightening. Exactly one configuration changes reader-facing behavior, and on those sites anonymous readers stop receiving free articles they were getting through the bug. Correct, but visible on day one, and worth telling affected publishers before it lands.

Verification: 3 new unit tests, 2 of which fail on `release`; full plugin suite green (1990 tests / 5697 assertions, 10 pre-existing skips); host PHPCS clean; CI green.

The comments below are all follow-up material — none of them should hold this PR.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Deferred, not addressed in this PR.

Confirmed pre-existing: `is_logged_in_metering_allowed()` already read `get_registered_settings()` before this branch existed, so the unverified-reader mismatch predates the change and is not introduced by it. Fixing it properly means deciding which meter governs a reader who is signed in but sees the registration wall — a semantics call, not a mechanical one, and not something to settle in a `release`-targeted hotfix.

Captured as a follow-up.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Not changing, on both counts.

The double read is object-cached `get_post_meta`, so there is no query and nothing measurable; removing it means changing `get_anonymous_settings()` to accept a pre-fetched array, which is a public method on a shared contract — not worth the surface change here.

The `$is_logged_in` parameter guards a case that is currently unreachable: the only two callers of `get_metering_period()` are `print_cta()` (frontend, already branching on login state) and `class-blocks.php:90`, and the latter sits inside `if ( $script_data['has_memberships'] )` — the legacy path, where both settings resolve to the same `metering_period` meta, so the two mechanisms cannot disagree. Worth doing when a REST or AJAX caller appears; not before.

Both captured as follow-ups.

plugins/newspack-plugin/tests/unit-tests/content-gate/metering.php
dkoo: Not changing — this one is blocked rather than declined.

`Memberships::is_active()` is `class_exists( 'WC_Memberships' ) && function_exists( 'wc_memberships' )`. Neither can be undone inside a PHPUnit process, so a test that defines them would leak memberships-active state into every subsequent test in the run. That is why no such test exists today, and it is why the gap cannot be closed without first adding a short-circuit filter to `is_active()` — a change for a normal release, not a hotfix.

The test-preamble helper is deliberately deferred for the same reason: keeping this diff mechanical. Both captured as follow-ups.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Reversing my earlier reply — this is now handled in this PR (`3aea38bef`), at the author's direction.

The generalization: `get_effective_anonymous_settings()` becomes `get_effective_settings( $gate_id, $is_logged_in )`, backed by `is_gated_by_registration()`, which mirrors the layout selection in `Content_Restriction_Control::is_post_restricted()` — the registration wall governs anonymous visitors **and** signed-in readers who have not verified their email on a wall that requires it. Non-reader users (admins, editors) stay exempt, matching the existing carve-out in `is_logged_in_metering_allowed()`.

Worth being precise about what changed, because the access behavior was already correct: `is_logged_in_metering_allowed()` already hard-bailed on unverified readers, so they were already being shown the registration gate immediately. What was wrong is that the metering *surfaces* still reported the paywall's allowance for them — `get_total_metered_views( true )` returned the paid count (3 in the test scenario) for a reader who actually gets zero, so the countdown could advertise a paid allowance on the registration wall. That is what this commit fixes.

Two new tests: `test_unverified_reader_uses_registration_settings` (fails on the previous commit with `Failed asserting that 3 is false`) and `test_verified_reader_uses_paid_settings`, which guards the other side so a verified reader still gets the paywall settings.

Not changed: an unverified reader still gets no metered views at all, even where registration metering is on. Granting them the registration allowance would be a loosening of access and is a separate decision.

dkoo:
## Runtime verification ✅

Closing the gap flagged in the self-review. Exercised in isolated env `pr671-metering` running this branch, against a real gate — registration wall **and** paywall, each with its own distinct gate layout so the two are told apart in the actual response — using real HTTP requests, not just unit tests.

**1. Ticket scenario, anonymous** (wall active + metering OFF, paywall metering ON @ 3/month)

| | Registration layout | Metering localized | Article body |
|---|---|---|---|
| This branch | ✅ served | none | withheld |
| `origin/release` `class-metering.php`, same env | not shown to reader | **`count: 3, period: month`** | **served in full** |

Swapping the pre-fix file into the same running env reproduced NEWS-2785 exactly — the anonymous reader handed the *paywall's* allowance. Swapping back restored correct behavior.

**2. #4571 fallback still works** (wall inactive, paywall gating via an `email_domain` rule): anonymous reader metered with `count: 3, period: month` and served the paid layout. Preserved, no regression.

**3. Unverified logged-in reader** (verification required) — logged in over a real session:

| | Gate layout | Body | `get_total_metered_views( true )` | `get_metering_period()` |
|---|---|---|---|---|
| This branch | registration | withheld | `false` | `week` (the wall's own) |
| Previous commit | registration | withheld | **`3`** | **`month`** (paywall leaking) |
| Verified reader, either version | paywall | — | `3` | `month` |

So the verified reader is unaffected in both — only the unverified reader's numbers change, which is the intent.

**Debug log:** clean. Two fatals, both from my own malformed `wp eval` probes during setup, and zero references to newspack-plugin files.

No screenshots — browser reads against this host are gated behind per-site approval in my environment, so I asserted on the returned HTML markers instead, which is the more precise check here anyway.

**Release risk drops HIGH → ELEVATED.** The remaining caution is behavioral, not defective: this is a tightening (see the rollout note in the description), so affected publishers deserve notice before it lands.

### One coverage gap this surfaced

Seeding the gate for this test exposed something no unit test would catch: `is_post_restricted()` only restricts when `$is_restricted && $gate_layout_id`, so a gate carrying `gate_layout_id => 0` **silently never gates at all**. Every test in `metering.php` — pre-existing and new — forces restriction with `add_filter( 'newspack_is_post_restricted', '__return_true' )`, so none exercises real content-rule matching or the layout-id requirement. Worth a follow-up test that builds a gate end-to-end and asserts restriction without the filter.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -203,6 +203,67 @@ public static function get_registered_settings( $gate_id ) {
...
+	 * Legacy Woo Memberships gates are exempt: they have no `registration` meta at all
+	 * and read both meters from the shared `metering` meta, so they keep the plain
+	 * anonymous/registered split.
+	 *
+	 * @param int  $gate_id      Gate ID.
+	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
+	 *
+	 * @return array Metering settings.
+	 */
+	private static function get_effective_settings( $gate_id, $is_logged_in ) {
```
adekbadek, line 267: **Suggestion:** The metering truth table now lives in two parallel encodings: this helper, and `is_logged_in_metering_allowed()`, which still decides *enablement* from its own verification bail plus a direct `get_registered_settings()` read. They agree today only because that bail removes the one input class (reader + require_verification + unverified) where `get_effective_settings( $gate_id, true )` would return different settings. That coupling is invisible at the other call site — a future edit relaxing the verification bail, or changing this helper's logged-in branch, would silently desynchronize the meter's enablement check from the settings every reporting surface advertises. Consider routing `is_logged_in_metering_allowed()`'s settings read through `get_effective_settings( $gate_post_id, true )` (behavior-identical for all currently reachable states), or adding a cross-reference comment there naming the invariant.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -203,6 +203,67 @@ public static function get_registered_settings( $gate_id ) {
...
+	 * registration layout - and skips the `custom_access` rules entirely - for both
+	 * anonymous visitors and signed-in readers who have not verified their email on a
+	 * wall that requires verification.
+	 *
+	 * @param int  $gate_id      Gate ID.
+	 * @param bool $is_logged_in Whether to evaluate for a logged-in reader.
+	 *
+	 * @return bool Whether the registration wall governs the reader.
+	 */
+	private static function is_gated_by_registration( $gate_id, $is_logged_in ) {
```
adekbadek, line 226: **Nit:** The two new private helpers are untyped. They're new, private (zero BC surface), and neighboring Access Control code already uses native types (e.g. `expand_hierarchical_terms( array $term_ids, \WP_Taxonomy $taxonomy ): array`), so `( int $gate_id, bool $is_logged_in ): bool` / `: array` costs nothing even in a hotfix. A shaped `@return array{enabled: bool, count: int, period: string}` on `get_effective_settings()` would also pin down the contract the four call sites now rely on.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -203,6 +203,67 @@ public static function get_registered_settings( $gate_id ) {
...
+		if ( ! $registration['active'] ) {
+			return false;
+		}
+		if ( ! $is_logged_in ) {
+			return true;
+		}
+		if ( ! $registration['require_verification'] ) {
+			return false;
+		}
+		// Non-reader users (admins, editors) are exempt, matching the same carve-out in
```
adekbadek: **Nit:** “Matching the same carve-out in `is_logged_in_metering_allowed()`” overstates the equivalence: the carve-out there is gated on `Content_Gate::requires_account_verification()`, which reads `require_verification` without checking `registration['active']`, while this helper checks `active` first. The configs diverge on a reachable state — a publisher enables regwall + verification, later disables the regwall keeping the paywall; the `wp_parse_args` merge in `update_registration_settings()` retains the stale `require_verification => true`. In that config `is_logged_in_metering_allowed()` bails for unverified readers (pre-existing, out of scope) while this helper routes them to paid settings, so `get_total_metered_views( true )` reports an allowance the mechanism denies. Nothing user-visible today (all rendering surfaces sit behind `is_metering()`), but worth rephrasing to note the `active`-check asymmetry.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -203,6 +203,67 @@ public static function get_registered_settings( $gate_id ) {
 		];
 	}
 
+	/**
+	 * Whether the registration wall is the rule that gates the current reader.
+	 *
+	 * Mirrors the gate-layout selection in
```
adekbadek: **Nit:** The “mirrors the gate-layout selection in `Content_Restriction_Control::is_post_restricted()`” claim is partial: `is_post_restricted()` gives anonymous visitors an institution/`supports_anonymous` bypass of the registration wall that this helper does not replicate — it returns `true` for every anonymous reader on a registration-active gate. Currently harmless (an institution-bypassed reader is not restricted, so no metering surface ever consults the helper for them), but the unqualified “mirrors” invites a future caller to use it on unrestricted readers. A one-line caveat (“assumes the reader is already known to be restricted; the anonymous institution bypass is not re-checked here”) would close that trap.

plugins/newspack-plugin/includes/content-gate/class-metering.php
```diff
@@ -490,10 +547,8 @@ public static function get_metering_period( $post_id = null ) {
 		if ( ! $post_id ) {
 			$post_id = get_the_ID();
 		}
-		$gate_post_id         = Content_Gate::get_gate_post_id( $post_id );
-		$anonymous_settings   = self::get_anonymous_settings( $gate_post_id );
-		$registered_settings  = self::get_registered_settings( $gate_post_id );
-		$settings = $anonymous_settings['enabled'] ? $anonymous_settings : $registered_settings;
+		$gate_post_id = Content_Gate::get_gate_post_id( $post_id );
+		$settings     = self::get_effective_settings( $gate_post_id, \is_user_logged_in() );
```
adekbadek, line 571: **Nit:** `get_metering_period()`'s docblock still reads as a pure per-post getter, but the function is now also current-user-dependent via `\is_user_logged_in()` (and, for unverified readers, `wp_get_current_user()` inside the helper). Callers caching or comparing periods per-post across auth states would be surprised — worth updating the summary to “for the current reader on a post”.

plugins/newspack-plugin/tests/unit-tests/content-gate/metering.php
```diff
@@ -613,4 +613,290 @@ function() use ( $gate_id ) {
...
+		add_filter(
+			'newspack_content_gate_post_id',
+			function() use ( $gate_id ) {
+				return $gate_id;
+			}
+		);
+
+		// Register and verify the reader.
+		$user_id = $this->register_reader( 'verified-reader@metering-test.com' );
+		\update_user_meta( $user_id, Reader_Activation::EMAIL_VERIFIED, true );
```
adekbadek: **Suggestion:** This verifies the reader by writing the `Reader_Activation::EMAIL_VERIFIED` user meta directly, while sibling tests use the `Reader_Activation::set_reader_verified()` API. If verification storage ever changes, this one test drifts from what production code does — using the API keeps it honest.

adekbadek (approved):
Pre-approving, since there no blockers. Impact on other units is compatible (every caller of the changed methods is internal and already guards the new `false` return), and the full manual test matrix passes against a live environment: anonymous readers are walled on article 1 with the regwall's metering off, signed-in readers get exactly the paid allowance, unverified readers are walled with no metered view, and the #4571 simple-paywall fallback is preserved.

The one thing to keep in view is operational, and the PR body already documents it: for gates with the registration wall active and its metering off, anonymous readers stop getting free articles the day this ships — and because the metering count/period are baked into cached anonymous HTML, config changes look stale until the page cache turns over (reproduced during testing: a page-cache hit kept serving the old meter until a flush).

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Done in `884134c61` — routed the settings read through `get_effective_settings( $gate_post_id, true )`.

I verified the equivalence before making the change rather than taking "behavior-identical" on faith: every state that reaches that settings read has already passed the verification bail above, which means the reader is verified, a non-reader, or on a no-verification gate — and in every one of those `is_gated_by_registration( $gate_id, true )` is false, so the helper returns the paid settings, identical to the direct read. The one state where it would return registration settings (reader + require_verification + unverified) is exactly the state the bail returns on, so it never reaches the read. Added a comment at the call site naming that invariant. Full suite green, and the existing `test_metering_allowed_when_verified` / `test_metering_blocked_when_unverified` plus the two new reader tests lock both directions.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Took the `@return` shape, held off on the native hints — `884134c61` adds `@return array{enabled: bool, count: int, period: string}` to `get_effective_settings()`, which pins the contract the four call sites rely on.

On the native hints I went the other way, and here is the evidence: `class-metering.php` is **0 of 22** functions typed, and the whole `content-gate/` directory has exactly **one** typed signature — the `expand_hierarchical_terms` you cited, in a different file. Adding `( int $gate_id, bool $is_logged_in ): bool` to just these two new methods would make them the only typed signatures in this file, so matching the file I am in seemed the stronger consistency call than matching one method next door. Happy to add them if you would rather match the subsystem outlier — it is a one-liner either way.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Confirmed and rephrased in `884134c61`. `requires_account_verification()` does read `require_verification` with no `active` guard (`class-content-gate.php:1129`), while this helper checks `active` first, so on the stale-`require_verification` config you described the two diverge exactly as you said. The comment now states it reuses only the reader/verified checks and calls out that the bail keys off `requires_account_verification()` (no `active` guard), so the two fully agree only while the wall is active.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Confirmed and caveated in `884134c61`. `is_post_restricted()` does grant anonymous visitors a `supports_anonymous` bypass (institutional access) before it ever restricts, which this helper does not re-check — it returns true for every anonymous reader on an active-wall gate. The docblock now says it assumes the reader is already known to be restricted (it is a settings selector, not a restriction check) and that the anonymous `supports_anonymous` bypass is not re-checked here.

plugins/newspack-plugin/includes/content-gate/class-metering.php
dkoo: Updated in `884134c61`. The summary now reads "Get the metering period for the current reader on a post" and notes the result is auth-state dependent (via `is_user_logged_in()`, and verification state for logged-in readers), so callers caching or comparing periods should key on the reader as well as the post.

plugins/newspack-plugin/tests/unit-tests/content-gate/metering.php
dkoo: Switched to `Reader_Activation::set_reader_verified( $user_id )` in `884134c61`, matching the sibling test at line 245. Agreed — going through the API keeps the test honest if verification storage ever moves off the `EMAIL_VERIFIED` meta.

dkoo:
@adekbadek thanks for the review and feedback! I pushed changes to address all threads and responded to/resolved them. Since the PR is pre-approved I'll move ahead with merging.

PR #672, opened 2026-07-20
fix(components): render routed content when no tab owns the route

dkoo:
## What

When a wizard's tab bar renders but **no visible tab owns the current route**, the routed page content is dropped entirely and the page body renders blank. This restores it.

## Root cause

#472 moved the wizard's routed content *inside* the active tab's `Tabs.Panel`: `Page` now does `cloneElement( tabbedNavigation, { renderShell, content: children } )` (`packages/components/src/page/index.js:81-83`), so the whole `<Switch>` is handed to `TabbedNavigation` and rendered only by the panel whose `value === activeValue`.

On a route no visible tab matches, `activeIndex === -1` → `activeValue === null` → **every** panel renders `null`, and the content element is never placed in the tree at all.

Two consequences, the second easy to miss:

1. Hidden sub-views (edit/add screens registered with `isHidden`) render blank — the component never mounts, so it never even fetches.
2. **The router's own fallback can't save it.** The wizard's trailing `<Redirect to={ displayedSections[0].path } />` (`wizard/index.js:207`) lives *inside* the dropped content, so an unrecognized route no longer self-heals to the first tab — it's a permanent blank screen. That's a silent behavior change for every stale bookmark and typo'd URL on a multi-tab wizard.

A wizard is only safe today if it has exactly **one** visible section, in which case `Wizard` sets `tabbedNavigation = false` and `Page` falls through to `renderShell( null, children )`.

## Blast radius — all confirmed in-browser on `main`

| Surface | Broken route(s) |
|---|---|
| **Audience → Plans** | `#/edit/:id` **and** `#/new` — can neither edit nor create a plan |
| **Setup wizard** | `#/` (Welcome) and `#/completed` — the onboarding entry point |
| **Story Budget** | the bare URL (`?page=newspack-story-budget`, no hash), which is what the admin menu links to |
| **Settings** | any URL for a conditionally-registered tab that isn't currently active (e.g. `#/additional-brands` without the multibranded plugin), plus any unrecognized hash |
| **Advertising → Display Ads** | `#/google_ad_manager` and `#/google_ad_manager/:id` — Google Ad Manager ad-unit management, on any site with GAM enabled |

The advertising case is worth spelling out, because it doesn't fit the pattern of the others: that wizard hand-rolls its own `HashRouter` and passes `tabbedNavigation={ tabs }` through `withWizardScreen` rather than registering `isHidden` sections on `Wizard`. Its three tabs are `/` (`exact: true`), `/placements` and `/settings`, none of which cover the two `/google_ad_manager` routes — and the Providers tab links straight to them whenever GAM is enabled (`views/providers/index.js:84-85`), as do the ad-unit save/cancel buttons.

Story Budget deserves a note: it was *latent* rather than visibly broken, because its `dist/` predated #472. It broke the moment the bundle was rebuilt — meaning the next release would have shipped a blank landing page.

Unaffected: pricing-rules, content-gates, integrations, premium-newsletters, donations, subscriptions, dashboard, newsletters (one visible section each), and the PHP-rendered admin header (href-only items → `LinkTabbedNavigation`, no panel gating).

**Latent landmine worth knowing:** content-gates carries **6** hidden routes and integrations 2 — all safe *only* because each wizard has exactly one visible tab. Adding a second tab to either would break every one of them at once, silently.

## The fix

Render `content` as a sibling of the panels when no tab is active. It goes *outside* the panels deliberately: with no active tab, no panel is entitled to hold the content, and putting it in one would break the tab↔panel `aria-controls` pairing that `@wordpress/ui` validates in development builds.

This is fixed in the shared component rather than per-wizard on purpose. A local `activeTabPaths: [ '/new', '/edit/*' ]` on the Plans wizard would have fixed the reported symptom and left Setup, Story Budget, Settings and Advertising broken — and it wouldn't restore the `<Redirect>` fallback anywhere. The advertising case makes the point concrete: it uses a different composition path entirely, so a config-level fix would have to be applied correctly at six call sites across two different wizard idioms, and any one that's missed fails silently as a blank page.

`activeTabPaths` remains the better *per-wizard* treatment for routes that are conceptually sub-views of a visible tab (the parent tab stays selected and its panel keeps owning the content). This change is what guarantees a missed `activeTabPaths` degrades to "renders, no tab selected" instead of "blank page".

Two things checked before choosing this shape:
- `Tabs.Root` already received `value={ null }` in exactly this case before the change, so no new null is introduced.
- **No CSS anywhere targets `newspack-tabbed-navigation__panel`**, so rendering outside it has no layout consequence.

## Testing

**Automated.** Two regression tests added, covering the blank body and the swallowed `<Redirect>` (using a real `<Redirect>` as content, mirroring the wizard's actual structure). `packages/components` had **no** coverage for "pathname matches no tab" before this.

Verified they're not vacuous: with the fix reverted, **exactly 2 tests fail — both new ones** (91 others pass). With the fix, **93/93 across 13 suites**.

**Manual**, on a dev site running `main`, all four surfaces above confirmed blank before and correct after. Also re-checked for regressions on *owned* routes: content still renders inside the correct panel with the right tab highlighted (Plans tab switching, Settings tab switching, Story Budget `#/stories` ↔ `#/budgets`).

> ⚠️ **Reviewer/tester note — how to actually get this change into a build.** `newspack-components` is consumed two different ways: `newspack-plugin` symlinks `packages/` and compiles from `src`, while other plugins import the package by name and resolve `"main": "dist/esm/index.js"` — the **built** output. So `n build <plugin>` alone will appear to fix newspack-plugin while leaving every other consumer broken. To verify Story Budget you must run `npm run build` in `packages/components` **first**, then rebuild the plugin. This cost us two false negatives while confirming the fix.

## Notes

Surfaced while reviewing #466, but unrelated to it — it reproduces on plain `main` with none of that PR's code. cc the author of #472, since this adjusts that change's content-ownership model.

### DIFF
--- a/packages/components/src/page/index.js
+++ b/packages/components/src/page/index.js
@@ -36,7 +36,8 @@ import './style.scss';
  * @param {*}      [props.actions]
  * @param {*}      [props.tabbedNavigation] A `TabbedNavigation` element; its bar renders inside
  *                                          the sticky header block and the page children render
- *                                          inside its active tab panel.
+ *                                          inside its active tab panel — or, when no visible tab
+ *                                          owns the route, as a sibling of the panels.
  * @param {string} [props.className]
  * @param {*}      props.children
  * @return {JSX.Element} Page component.
--- a/packages/components/src/tabbed-navigation/index.js
+++ b/packages/components/src/tabbed-navigation/index.js
@@ -118,9 +118,23 @@ const RoutedTabbedNavigation = ( { items, className, disableUpcoming, content = 
 		);
 	} );
 
+	// A route no visible tab owns — a hidden sub-view (an edit screen), or a path
+	// with no matching tab at all — still has to render its content, outside the
+	// panels: with no active tab there is no panel entitled to hold it. Without
+	// this the page body is blank, and because the router's own fallback (the
+	// wizard's trailing `<Redirect>`) lives inside that content, it never mounts
+	// to correct the route either.
+	const unownedContent = null === activeValue ? content : null;
+
 	return (
 		<Tabs.Root value={ activeValue } className="newspack-tabbed-navigation__root">
-			{ renderShell( bar, panels ) }
+			{ renderShell(
+				bar,
+				<>
+					{ panels }
+					{ unownedContent }
+				</>
+			) }
 		</Tabs.Root>
 	);
 };
--- a/packages/components/src/tabbed-navigation/index.js
+++ b/packages/components/src/tabbed-navigation/index.js
@@ -43,7 +43,8 @@ const defaultRenderShell = ( bar, content ) => (
 /**
  * Router-driven tabs: a real WAI-ARIA tabs widget. The routed page content
  * renders inside the active tab's `Tabs.Panel`, so each tab's `aria-controls`
- * points at the panel that actually contains its content.
+ * points at the panel that actually contains its content — or, when no visible
+ * tab owns the route, as a sibling of the panels (see `unownedContent` below).
  */
 const RoutedTabbedNavigation = ( { items, className, disableUpcoming, content = null, renderShell = defaultRenderShell, children = null } ) => {
 	const history = useHistory();
@@ -124,6 +125,10 @@ const RoutedTabbedNavigation = ( { items, className, disableUpcoming, content = 
 	// this the page body is blank, and because the router's own fallback (the
 	// wizard's trailing `<Redirect>`) lives inside that content, it never mounts
 	// to correct the route either.
+	//
+	// This renders without the panels' `__panel` class, which is styleless today.
+	// Keep it that way: styling `__panel` would silently indent/pad owned routes
+	// only, and the divergence would be hard to trace back to here.
 	const unownedContent = null === activeValue ? content : null;
 
 	return (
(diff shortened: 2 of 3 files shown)


dkoo:
### Deferred follow-up: with no active tab, `disableUpcoming` disables *every* tab

Noting this here rather than fixing it in this PR, since it's pre-existing behavior that this PR merely makes visible.

In `tabbed-navigation/index.js`, tab state is `isDisabled = disableUpcoming && index > activeIndex`. When no visible tab owns the route, `activeIndex === -1`, so `index > -1` is true for every tab — the whole strip renders disabled and href-less, with nothing selected.

This isn't new: the pre-#472 implementation had the same `index > currentIndex` comparison, and there's an existing test pinning it. What's new is that it's now *reachable*, because until this PR those routes rendered nothing at all.

The only `disableUpcoming` consumer is the setup wizard (`plugins/newspack-plugin/src/wizards/setup/index.js:82`), and both of its `isHiddenInNav` routes are unowned:

- **`/` (Welcome)** — all-disabled seems right: no step has been reached, and you advance via the Continue button.
- **`/completed`** — questionable. Every step is done, but the user gets a fully greyed-out bar with no way back to Settings/Services/Design except editing the URL.

Two plausible directions, both cheap:
1. Treat `activeIndex === -1` as "disable nothing" — the predicate becomes meaningful only once a tab is actually active.
2. Give those routes an owning tab via `activeTabPaths`, so a real tab stays selected and the disable logic has a sensible anchor.

I'd lean toward (1) as the component-level default, with (2) as the per-wizard refinement — but it's a UX call about the onboarding flow more than a technical one, so I don't want to make it unilaterally inside a bugfix PR.

@thomasguillot — flagging for you as #472's author, in case you'd like to pick this up, or have a view on which direction fits the intended tab semantics. Happy to implement either.

miguelpeixe (approved):

dkoo:
Thanks for the review, @miguelpeixe!

@thomasguillot since the fix here is purely mechanical (no visual or UX changes) and addresses a regression affecting `main`, I'm going to merge it. But please let me know if you'd like me to open a follow-up to address [this comment](https://github.com/Automattic/newspack-workspace/pull/672#issuecomment-5028272184), or if you'd like to handle it.

thomasguillot:
Sorry I forgot to respond! 😱

> Treat activeIndex === -1 as "disable nothing" — the predicate becomes meaningful only once a tab is actually active.

I think this? I trust you

> But please let me know if you'd like me to open a follow-up to address https://github.com/Automattic/newspack-workspace/pull/672#issuecomment-5028272184, or if you'd like to handle it.

All yours if that's ok 😂

dkoo:
@thomasguillot: Correction to my note above, after a closer look — the symptom I described isn't actually user-reachable, so I want to walk it back before anyone acts on it.

The setup wizard hides the tab strip entirely on both of its unowned routes: `setup/style.scss:12-16` sets `display: none` on `.newspack-tabbed-navigation` under `.newspack-wizard__welcome` and `.newspack-wizard__completed`. Those two (Welcome and Completed) are the only routes where `activeIndex === -1`, and the setup wizard is the sole `disableUpcoming` consumer. So the "fully greyed-out bar with no way back" I described on the completed screen never renders — the whole strip is `display: none` there.

The underlying predicate is still wrong-looking — `disableUpcoming && index > activeIndex` disables *every* tab when nothing is active — but it's masked by CSS in the only place it fires today. So a fix is defensive/latent-correctness, not a live bug: it would only matter for a future wizard that shows its nav on an unowned route.

I think we can safely leave this for now unless future changes make the scenario user-facing.

thomasguillot:
> I think we can safely leave this for now unless future changes make the scenario user-facing.

👌

PR #677, opened 2026-07-21
fix(newsletters): address publisher feedback on the newsletters list

thomasguillot:
### All Submissions:

Addresses the four regressions Dan at Outlier Media reported against the new Newsletters admin screen.

**Items per page, persisted.** The per-page control now offers 10 / 20 / 50 / 100 / **All**, and the choice is saved per user (user meta, via a new `admin-shell/preferences` REST endpoint) so it survives page loads and sessions. "All" walks the REST collection in batches of 100 behind a progress overlay ("Loading 40 of 112…"), so a publisher can keep the whole archive on screen and use the browser's find-in-page. The control lives inside the View options popover next to the other display settings, since the bundled DataViews control cannot represent an "All" option. Default page size drops from 25 to 20. Applies to all four admin-shell screens (newsletters, ads, advertisers, layouts).

**Load times.** Every list request now sends an explicit `_fields` allowlist, so the REST response no longer carries `content.rendered` and `excerpt.rendered`. Those are the expensive part: rendering post content server-side triggers synchronous, uncached oEmbed fetches, which is what turned a 100-item page into a 60–120 second wait. On the newsletters list specifically, term embeds are also now requested only when a taxonomy column is actually showing. The ads list keeps its embed unconditionally: its Quick Edit hydrates advertiser and placement from the embedded terms alone and sends both taxonomies on every save, so dropping the embed there would clear them. Locally (100 rows of near-empty drafts, so a best case for the old code) this takes the newsletters list from ~3.0s / 873KB to ~1.1s / 586KB; on real content with embeds the difference is far larger.

**Status filters.** All filters (Status, Send list, Author, Categories, Tags, Public page) are now pinned as primary chips, so filtering by status is a single click instead of three.

**View links.** The Visibility column renders a real link to the public newsletter page when one exists, so viewing a newsletter is one click again. The action stays in the menu as well.

**Along the way:** the admin header shows the current result count on all four list screens ("All Newsletters (112)", following filters and search, hidden when the list is empty), Trash is no longer an inline row action on newsletters or ads, table links pick up the admin theme colour, and the permanently-disabled filter funnel icon is hidden.

Closes DSGNEWS-200.

1. Go to **Newsletters → All Newsletters**. Confirm the list loads and the header reads "All Newsletters (N)" with N matching the total.
2. Open **View options** (the settings icon above the table). Confirm the items-per-page control sits inside the popover with options 10 / 20 / 50 / 100 / All, and that 20 is selected on a fresh install.
3. Switch to 50, then reload the page. The list should come back at 50 — the preference persists per user. Log in as a second user and confirm they still see their own setting, not yours.
4. On a site with 100+ newsletters, choose **All**. A centred overlay should fade in reading "Loading X of Y…", then the full list appears in one page with pagination showing a single page. Use the browser's find-in-page to confirm every newsletter is in the DOM.
5. With a small number of newsletters (a fast fetch), choose **All** again — the overlay should not flash on screen, as it only appears after a 300ms delay.
6. Repeat steps 2–4 on **Ads**, **Advertisers** and **Layouts**. Layouts defaults to 24 with options 12 / 24 / 48 / 96 / All; note that in All mode its prebuilt layouts are still appended to the saved ones.
7. Open DevTools → Network and compare the `newspack-newsletters/v1/newsletters` request before and after this branch. Confirm the response no longer contains `content` or `excerpt`, and that it is meaningfully smaller and faster.
8. Hide the Categories and Tags columns via View options, then reload. Confirm the request drops `wp:term` from `_embed` (author stays). Show a taxonomy column again and confirm `wp:term` returns and the category/tag cells still render correctly.
9. Confirm the filter chips (Status, Send list, Author, Categories, Tags, Public page) are all visible above the table by default, that clicking one filters immediately, and that there is no "Add filter" button or greyed-out funnel icon left over.
10. Find a published newsletter with a public page enabled. In the **Visibility** column, confirm the value is a link that opens the public page in a new tab, and that clicking it does not also select the row or open the editor. Confirm drafts and non-public newsletters show plain text instead.
11. Open a newsletter's row actions menu. Confirm Trash is there but is no longer an inline button in the row, and that Quick Edit still is. Same on the Ads screen.
12. Confirm the header count appears on Ads, Advertisers and Layouts too, and that it disappears entirely when a list (or a filtered/searched result) is empty.
13. Try to break the preferences endpoint: as a subscriber-level user, `POST` to `newspack-newsletters/v1/admin-shell/preferences` and confirm it is rejected; as an editor, post an invalid screen name or a per-page value like `0`, `500` or `"abc"` and confirm it is rejected without corrupting the stored preference.
14. Because the list no longer embeds terms by default, check Quick Edit carefully: open it on a newsletter that has categories and tags, with those columns hidden. The Categories and Tags fields must show the newsletter's real terms. Change only the visibility radio, save, and confirm the categories and tags are still intact afterwards. Then change a category, save, and confirm only that changed.
15. Run the tests: `pnpm --filter newspack-newsletters run test` (27 suites / 242 tests) and `n test-php` from `plugins/newspack-newsletters` (518 tests).

### Other information:

**Worth a close look:** dropping the term embed initially left Quick Edit seeding empty Categories/Tags fields, so saving a visibility change would have stripped every term from the newsletter. Quick Edit now hydrates from the post's raw term IDs and only sends a taxonomy the user actually touched (step 14 covers this). "All" also has a 10,000-item ceiling with a notice, keeps pages that succeeded when a sibling request fails, and reports the count it actually loaded rather than the server's total.

### DIFF
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -17,12 +17,14 @@ defined( 'ABSPATH' ) || exit;
  * page), stored in user meta so they follow the user across browsers.
  * Values are bootstrapped into the localized `newspackNewslettersAdmin`
  * global (see `Admin_Shell_Assets`), so the REST route is write-only in
- * practice.
+ * practice. Each screen gets its own user-meta key so a save only ever
+ * touches that screen's value — two tabs saving different screens can't
+ * clobber each other via a shared read-modify-write.
  */
 class Admin_Shell_Preferences {
 	const API_NAMESPACE = 'newspack-newsletters/v1';
 	const ROUTE         = 'admin-shell/preferences';
-	const USER_META_KEY = 'newspack_newsletters_admin_view_prefs';
+	const USER_META_KEY_PREFIX = 'newspack_newsletters_admin_view_prefs_';
 
 	const SCREEN_KEYS = [ 'newsletters-list', 'ads-list', 'advertisers-list', 'layouts-list' ];
 
@@ -94,12 +96,10 @@ class Admin_Shell_Preferences {
 			);
 		}
 
-		$all_prefs = self::get_preferences();
+		$screen_key = $request->get_param( 'screen' );
+		update_user_meta( get_current_user_id(), self::get_user_meta_key( $screen_key ), [ 'perPage' => $per_page ] );
 
-		$all_prefs[ $request->get_param( 'screen' ) ] = [ 'perPage' => $per_page ];
-		update_user_meta( get_current_user_id(), self::USER_META_KEY, $all_prefs );
-
-		return rest_ensure_response( $all_prefs );
+		return rest_ensure_response( self::get_preferences() );
 	}
 
 	/**
@@ -108,21 +108,28 @@ class Admin_Shell_Preferences {
 	 * @return array
 	 */
 	public static function get_preferences(): array {
-		$raw = get_user_meta( get_current_user_id(), self::USER_META_KEY, true );
-		if ( ! is_array( $raw ) ) {
-			return [];
-		}
 		$prefs = [];
 		foreach ( self::SCREEN_KEYS as $screen_key ) {
-			if ( ! isset( $raw[ $screen_key ]['perPage'] ) || ! is_numeric( $raw[ $screen_key ]['perPage'] ) ) {
+			$raw = get_user_meta( get_current_user_id(), self::get_user_meta_key( $screen_key ), true );
+			if ( ! is_array( $raw ) || ! isset( $raw['perPage'] ) || ! is_numeric( $raw['perPage'] ) ) {
 				continue;
 			}
-			$per_page = (int) $raw[ $screen_key ]['perPage'];
+			$per_page = (int) $raw['perPage'];
 			if ( self::is_valid_per_page( $per_page ) ) {
 				$prefs[ $screen_key ] = [ 'perPage' => $per_page ];
 			}
 		}
 		return $prefs;
+	}
+
+	/**
+	 * The user-meta key a screen's preferences are stored under.
+	 *
+	 * @param string $screen_key Screen identifier.
+	 * @return string
+	 */
+	public static function get_user_meta_key( string $screen_key ): string {
+		return self::USER_META_KEY_PREFIX . $screen_key;
 	}
 
 	/**
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -61,8 +61,15 @@ class Admin_Shell_Preferences {
 							'required' => true,
 						],
 						'prefs'  => [
-							'type'     => 'object',
-							'required' => true,
+							'type'                 => 'object',
+							'required'             => true,
+							'additionalProperties' => false,
+							'properties'           => [
+								'perPage' => [
+									'type'     => 'integer',
+									'required' => true,
+								],
+							],
 						],
 					],
 				],
@@ -71,7 +78,10 @@ class Admin_Shell_Preferences {
 	}
 
 	/**
-	 * Capability gate — matches the list screens themselves.
+	 * Capability gate — the floor for reaching any list screen. Some
+	 * screens require more (layouts needs `edit_others_posts`), but a
+	 * preference for a screen the user can't open is never read back, so
+	 * the shared floor is enough.
 	 *
 	 * @return bool
 	 */
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * Admin shell — per-user view preferences.
+ *
+ * @package Newspack_Newsletters
+ */
+
+namespace Newspack\Newsletters\Admin;
+
+use WP_Error;
+use WP_REST_Server;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Per-user view preferences for the admin-shell list screens (items per
+ * page), stored in user meta so they follow the user across browsers.
+ * Values are bootstrapped into the localized `newspackNewslettersAdmin`
+ * global (see `Admin_Shell_Assets`), so the REST route is write-only in
+ * practice.
+ */
+class Admin_Shell_Preferences {
+	const API_NAMESPACE = 'newspack-newsletters/v1';
+	const ROUTE         = 'admin-shell/preferences';
+	const USER_META_KEY = 'newspack_newsletters_admin_view_prefs';
+
+	const SCREEN_KEYS = [ 'newsletters-list', 'ads-list', 'advertisers-list', 'layouts-list' ];
+
+	/**
+	 * Client-side sentinel for "All" — the REST API caps `per_page` at
+	 * 100, so the client fetches in chunks and concatenates.
+	 */
+	const PER_PAGE_ALL = -1;
+	const PER_PAGE_MAX = 100;
+
+	/**
+	 * Boot hooks.
+	 */
+	public static function init(): void {
+		add_action( 'rest_api_init', [ __CLASS__, 'register_routes' ] );
+	}
+
+	/**
+	 * Register the update route.
+	 */
+	public static function register_routes(): void {
+		register_rest_route(
+			self::API_NAMESPACE,
+			'/' . self::ROUTE,
+			[
+				[
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => [ __CLASS__, 'update_preferences' ],
+					'permission_callback' => [ __CLASS__, 'permission_check' ],
+					'args'                => [
+						'screen' => [
+							'type'     => 'string',
+							'enum'     => self::SCREEN_KEYS,
+							'required' => true,
+						],
+						'prefs'  => [
+							'type'     => 'object',
+							'required' => true,
+						],
+					],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Capability gate — matches the list screens themselves.
+	 *
+	 * @return bool
+	 */
+	public static function permission_check(): bool {
+		return current_user_can( 'edit_posts' );
+	}
+
+	/**
+	 * Persist a screen's preferences for the current user.
+	 *
+	 * @param \WP_REST_Request $request Incoming request.
+	 * @return \WP_REST_Response|WP_Error
+	 */
+	public static function update_preferences( $request ) {
+		$prefs    = $request->get_param( 'prefs' );
+		$per_page = is_array( $prefs ) && isset( $prefs['perPage'] ) && is_numeric( $prefs['perPage'] ) ? (int) $prefs['perPage'] : null;
+		if ( null === $per_page || ! self::is_valid_per_page( $per_page ) ) {
+			return new WP_Error(
+				'newspack_newsletters_invalid_preference',
+				__( 'Invalid preference value.', 'newspack-newsletters' ),
+				[ 'status' => 400 ]
+			);
+		}
+
+		$all_prefs = self::get_preferences();
+
+		$all_prefs[ $request->get_param( 'screen' ) ] = [ 'perPage' => $per_page ];
+		update_user_meta( get_current_user_id(), self::USER_META_KEY, $all_prefs );
+
+		return rest_ensure_response( $all_prefs );
+	}
+
+	/**
+	 * The current user's sanitized preferences, keyed by screen.
+	 *
+	 * @return array
+	 */
+	public static function get_preferences(): array {
+		$raw = get_user_meta( get_current_user_id(), self::USER_META_KEY, true );
+		if ( ! is_array( $raw ) ) {
+			return [];
+		}
+		$prefs = [];
+		foreach ( self::SCREEN_KEYS as $screen_key ) {
+			if ( ! isset( $raw[ $screen_key ]['perPage'] ) || ! is_numeric( $raw[ $screen_key ]['perPage'] ) ) {
+				continue;
+			}
+			$per_page = (int) $raw[ $screen_key ]['perPage'];
+			if ( self::is_valid_per_page( $per_page ) ) {
+				$prefs[ $screen_key ] = [ 'perPage' => $per_page ];
+			}
+		}
+		return $prefs;
+	}
+
+	/**
+	 * Whether a per-page value is storable.
+	 *
+	 * @param int $value Candidate value.
+	 * @return bool
+	 */
+	private static function is_valid_per_page( int $value ): bool {
+		return self::PER_PAGE_ALL === $value || ( $value >= 1 && $value <= self::PER_PAGE_MAX );
+	}
+}
--- a/plugins/newspack-newsletters/src/admin-shell/components/header-count/index.js
+++ b/plugins/newspack-newsletters/src/admin-shell/components/header-count/index.js
@@ -11,10 +11,11 @@ import { createPortal, useEffect, useState } from '@wordpress/element';
 const CONTAINER_CLASS = 'newspack-newsletters-header-count';
 
 /**
- * @param {Object} props
- * @param {number} props.count Total items matching the current filters.
+ * @param {Object}  props
+ * @param {number}  props.count      Total items matching the current filters.
+ * @param {boolean} props.isRealized Whether `count` reflects a resolved fetch — before that it would flash "(0)".
  */
-export default function HeaderCount( { count } ) {
+export default function HeaderCount( { count, isRealized = true } ) {
 	const [ target, setTarget ] = useState( null );
 
 	useEffect( () => {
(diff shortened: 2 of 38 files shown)


plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -0,0 +1,144 @@
...
+		return self::USER_META_KEY_PREFIX . $screen_key;
+	}
+
+	/**
+	 * Whether a per-page value is storable.
+	 *
+	 * @param int $value Candidate value.
+	 * @return bool
+	 */
+	private static function is_valid_per_page( int $value ): bool {
```
dkoo, line 151: **Severity:** suggestion

`is_valid_per_page()` accepts `-1` or any `1..100`, but each screen offers a fixed set (newsletters 10/20/50/100; layouts 12/24/48/96). A stored value that is server-valid but absent from a screen's ToggleGroup — e.g. a newsletters `100` carried to layouts, or a legacy `37` — renders the control with no option highlighted while the list still fetches that size. Cosmetic (not a security issue), but a per-screen allowlist or client-side clamp to the nearest offered option would keep the control self-consistent.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/build-query.js
```diff
@@ -33,11 +33,25 @@ const SORT_FIELD_TO_ORDERBY = {
...
 		sortFieldToOrderby: SORT_FIELD_TO_ORDERBY,
 		defaultStatuses: DEFAULT_STATUSES,
-		extraParams: { _embed: 'author,wp:term' },
+		// `_fields` short-circuits `content.rendered` / `excerpt.rendered`
+		// (the full `the_content` chain, incl. synchronous oEmbed fetches)
+		// and the unused editor REST fields — per-item cost the list never
+		// reads. `_links` must stay in the list: `_embed` only expands
+		// links that survive the `_fields` filter.
+		extraParams: {
+			_embed: needsTerms ? 'author,wp:term' : 'author',
```
dkoo, line 51: **Severity:** suggestion

The "request `wp:term` only when a taxonomy column shows" optimization lives on this screen only. `ads-list/build-query.js` always sends `_embed: 'wp:term'`, layouts always `author`, advertisers nothing. Defensible, but the description implies the saving applies across the admin shell — either extend the `needsTerms` gate into the shared util or narrow the wording.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/quick-edit-panel.js
```diff
@@ -46,19 +46,40 @@ function useQuickEditOptions() {
...
+	// Unresolvable terms can't be shown or removed, so they ride along on save.
+	const unresolvedCategoryIds = useMemo( () => unresolvedIds( item?.categories, initialCategorySelections ), [ item, initialCategorySelections ] );
+	const unresolvedTagIds = useMemo( () => unresolvedIds( item?.tags, initialTagSelections ), [ item, initialTagSelections ] );
 	const initialVisibility = item?.meta?.is_public ? 'public' : 'private';
 
 	const [ categorySelections, setCategorySelections ] = useState( initialCategorySelections );
 	const [ tagSelections, setTagSelections ] = useState( initialTagSelections );
 	const [ visibility, setVisibility ] = useState( initialVisibility );
 	const [ isBusy, setIsBusy ] = useState( false );
+	const hasEditedTermsRef = useRef( false );
```
dkoo: **Severity:** suggestion

A single `hasEditedTermsRef` guards both taxonomies. When a taxonomy column is hidden, baselines seed asynchronously as options resolve; if a user edits Categories before the Tags options finish loading, the ref flips true and the Tags baseline-sync effect stops, stranding `tagSelections` at its empty pre-resolution value — a later Tags edit could then drop the real tags on save. Same term-loss class the visibility-only fix targets, in a narrower race window. Two separate refs would decouple them.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
```diff
@@ -0,0 +1,125 @@
...
+				properties.parentElement.insertBefore( container, properties );
+			} else {
+				popover.appendChild( container );
+			}
+			setSlot( container );
+		};
+
+		ensureSlot();
+		const observer = new MutationObserver( ensureSlot );
+		observer.observe( document.body, { childList: true, subtree: true } );
```
dkoo, line 69: **Severity:** suggestion

`usePopoverSlot` observes the whole `document.body` subtree and re-runs querySelectors on every mutation anywhere on the page; HeaderCount scopes its observer to the header wrapper — this one could similarly narrow to the DataViews root. It also couples to private DataViews class names and the `config.perPageSizes: []` trick (all correct in dataviews 10.3.0, verified live) with no test — a version bump could silently break the control. This is the most brittle new module and the only one without a unit test; a jsdom test asserting it portals into a mocked `.dataviews-view-config` and emits the `-1` sentinel would guard the coupling.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+	return (
+		<>
+			{ progress &&
+				createPortal(
+					<VStack
+						className="newspack-newsletters-fetch-all-progress"
+						spacing={ 3 }
+						alignment="center"
+						aria-live="polite"
```
dkoo: **Severity:** suggestion

The progress region is `aria-live="polite"` and updates ~once per batch — up to ~33 announcements on a 10k "All" walk, flooding the screen-reader queue. Consider announcing only start and completion.

plugins/newspack-newsletters/src/admin-shell/components/header-count/index.js
```diff
@@ -0,0 +1,54 @@
...
+export default function HeaderCount( { count } ) {
+	const [ target, setTarget ] = useState( null );
+
+	useEffect( () => {
+		const wrapper = document.getElementById( 'newspack-wizards-admin-header' );
+		if ( ! wrapper ) {
+			return undefined;
+		}
+		const ensureTarget = () => {
+			const heading = wrapper.querySelector( 'h1.newspack-breadcrumbs__current' );
```
dkoo, line 26: **Severity:** suggestion

The count portals into an `h1.newspack-breadcrumbs__current` owned by newspack-plugin's admin-header (a separate React root). It degrades gracefully to null and re-attaches via the observer, so this is acceptable — but the target is the breadcrumb markup from packages/components #472; I confirmed it renders once the admin-header hydrates, but on the server-rendered placeholder (and any newspack-plugin build predating that breadcrumb API) the target is absent and the count silently no-ops. Worth confirming the shipped newspack-plugin renders that breadcrumb on these screens, since the plugins release independently.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -0,0 +1,144 @@
...
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => [ __CLASS__, 'update_preferences' ],
+					'permission_callback' => [ __CLASS__, 'permission_check' ],
+					'args'                => [
+						'screen' => [
+							'type'     => 'string',
+							'enum'     => self::SCREEN_KEYS,
+							'required' => true,
+						],
+						'prefs'  => [
```
dkoo, line 63: **Severity:** nit

`prefs` is `'type' => 'object'` with no `properties`, so framework validation accepts any shape and correctness rests on the manual `perPage` check (which is thorough). A `'properties' => [ 'perPage' => [ 'type' => 'integer' ] ]` schema or sanitize_callback would reject malformed input earlier and self-document the contract.

plugins/newspack-newsletters/src/admin-shell/utils/per-page.js
```diff
@@ -0,0 +1,22 @@
...
+ * Items-per-page constants shared by the DataView list screens.
+ *
+ * `PER_PAGE_ALL` is a client-side sentinel — the REST API caps
+ * `per_page` at 100, so "All" fetches `FETCH_ALL_CHUNK_SIZE` at a time
+ * and concatenates (see `use-collection-data.js`).
+ */
+
+export const PER_PAGE_ALL = -1;
+
+export const FETCH_ALL_CHUNK_SIZE = 100;
```
dkoo, line 13: **Severity:** nit

JS `FETCH_ALL_CHUNK_SIZE` (100) and PHP `PER_PAGE_MAX` (100) must stay in sync but nothing enforces it — just a note that changing one needs a matching edit to the other.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
```diff
@@ -0,0 +1,125 @@
...
+						style={ wpbodyRect ? { left: wpbodyRect.left + wpbodyRect.width / 2 } : undefined }
+					>
+						<Spinner />
+						<Text weight={ 600 }>
+							{ sprintf(
+								/* translators: 1: number of items loaded so far, 2: total number of items. */
+								__( 'Loading %1$s of %2$s…', 'newspack-newsletters' ),
+								progress.loaded,
```
dkoo: **Severity:** nit

`progress.loaded`/`progress.total` render as raw integers while HeaderCount and the cap notice use `.toLocaleString()`; on a large walk the overlay reads "Loading 8000 of 50000" without separators.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -0,0 +1,144 @@
...
+		);
+	}
+
+	/**
+	 * Capability gate — matches the list screens themselves.
+	 *
+	 * @return bool
+	 */
+	public static function permission_check(): bool {
+		return current_user_can( 'edit_posts' );
```
dkoo, line 89: **Severity:** nit

The gate is `edit_posts` and the docblock says it "matches the list screens themselves," but the layouts screen requires `edit_others_posts`, so an `edit_posts`-only user can persist a `layouts-list` preference for a screen they can't open. Harmless (never read back), but the docblock claim is slightly inaccurate.

dkoo (approved):
Approving. This resolves all four publisher-reported regressions on the newsletters admin list and is carefully built and well tested (242 JS + 518 PHP tests pass). I set up an isolated env with 200+ seeded newsletters and verified the core behavior end-to-end: the items-per-page control (10/20/50/100/All, default 20) lives in the View options popover and persists per-user across reloads; the list request drops content/excerpt via the `_fields` allowlist and only embeds `author` (no `wp:term`) when taxonomy columns are hidden; the Visibility column renders a real new-tab link for public+published newsletters and plain text otherwise; the header count follows search/filters; and the new preferences REST endpoint rejects a subscriber (403) and invalid values (400) without corrupting the stored preference.

Most importantly, both paths flagged as risky hold up. Quick Edit hydrates Categories/Tags from raw term IDs even when those columns are hidden, and a visibility-only save left the terms intact (verified in the DB) — the regression that would have stripped every term is fixed. The batched "All" walk is defensively written (ordered concatenation, 10k ceiling, partial-page retention, deterministic-vs-transient error handling).

Comments below are all non-blocking polish — consistency gaps and the inherently fragile DOM-coupled UI glue (which would benefit from a test) — plus a few nits.

**"All" mode (not a blocker, follow-up on our side).** Manual QA on a 200+-newsletter env showed the "All" walk stopping at 100 rows with the "Only some items could be loaded" notice. Root-caused end to end: the batching here is correct — the truncation is a **pre-existing** newspack-newsletters issue where `Newspack_Newsletters::fix_public_status()` (hooked on `the_post`, guarded only by `! is_admin()`) calls `exit` during REST requests when it hits a publish-but-non-public newsletter mid-collection, cutting the response short. It's independent of this PR (the REST controller/hook is unchanged here) and self-healing. We'll handle that fix separately.

**Non-blocking suggestion — header-count parity.** The live item count renders next to the page title on the Newsletters screen, but not on the Advertising (Ads) page even with an ad present (and, by the same mechanism, Advertisers/Layouts). `HeaderCount` portals into the newspack admin-header breadcrumbs heading (`#newspack-wizards-admin-header h1.newspack-breadcrumbs__current`), which isn't present on those pages the way it is on Newsletters, so the count has nowhere to attach. Consider bringing those wizard pages to visual parity with Newsletters so the count shows there per the intended behavior.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Fixed in 6929198. `usePersistedView` now takes the screen's offered options and falls back to the default when the stored value is not one of them; layouts passes its `12/24/48/96` set. The control can no longer render with nothing highlighted.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/quick-edit-panel.js
thomasguillot: Fixed in 6929198 — split into `hasEditedCategoriesRef` and `hasEditedTagsRef`, so an edit to one taxonomy no longer freezes the other's baseline sync.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
thomasguillot: Fixed in 6929198. The progress text is no longer a live region; `speak()` announces the start of the walk and its completion, so a 10k walk produces two announcements instead of ~100. Covered by a test.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Fixed in 6929198 — `prefs` is now `properties: { perPage: integer }` with `additionalProperties: false`, so malformed input is rejected by the framework before the manual check runs.

plugins/newspack-newsletters/src/admin-shell/utils/per-page.js
thomasguillot: Added a note in 6929198: `FETCH_ALL_CHUNK_SIZE` now names `Admin_Shell_Preferences::PER_PAGE_MAX` as its mirror.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
thomasguillot: Fixed in 6929198 — both values now go through `.toLocaleString()`, matching HeaderCount and the cap notice.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Fixed in 6929198 — the docblock now says `edit_posts` is the floor for reaching any list screen, notes that layouts requires more, and why a preference for an unreachable screen is harmless.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/build-query.js
thomasguillot: Narrowed the wording rather than extending the gate, and documented why in `ads-list/build-query.js` (6929198). The ads screen cannot take the same treatment: its Quick Edit hydrates advertiser and placement from the embedded terms alone and sends both taxonomies on every save, so dropping the embed when those columns are hidden would clear them — the same term-loss class the newsletters fix targets. Making it safe there needs the raw term IDs in `_fields` plus the hydrate-from-IDs change, which is worth its own PR.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
thomasguillot: Added `components/items-per-page/index.test.js` in 6929198: it asserts the control portals into `.dataviews-view-config` immediately before `.dataviews-field-control`, that the `-1` sentinel renders as "All" alongside the numeric options, and that a multi-batch walk announces exactly twice. A rename on either class now fails the suite.

The observer stays on `document.body`: the view-config panel renders through `Dropdown` → `Popover`, which portals outside the DataViews root, so scoping it there would stop the control appearing at all. Added a comment saying so, and noting it is `childList`-only with an early bail when the popover is closed.

plugins/newspack-newsletters/src/admin-shell/components/header-count/index.js
thomasguillot: Checked this live on all four screens, reading the heading node directly: Ads gives `Ads (12)`, Advertisers `Advertisers (10)`, Layouts `Layouts (15)`, each matching the row count. `#newspack-wizards-admin-header` and `h1.newspack-breadcrumbs__current` are present on all of them — all four pages supply a trail via `get_wizard_breadcrumbs()`, newspack-plugin treats `newspack_nl_ads_cpt` and `newspack_nl_advertiser` as wizard screens, and `Page` renders `Breadcrumbs` regardless of tabs, with the last crumb always the `h1`. So the count has a target everywhere and no parity work is needed. The count is hidden at zero by design, which is the likely explanation for an Ads screen showing none.

PR #679, opened 2026-07-21
feat(network): distribute posts from the front-end admin bar

thomasguillot:
### All Submissions:

Adds a **Distribute** action to the front-end admin bar. A user with the `newspack_network_distribute` capability, viewing a distributable post, can push it to network sites without opening the editor.

Clicking **Distribute** opens a small Newspack UI modal listing the network sites as checkboxes (with a select-all when there is more than one target). Pressing Distribute asks for confirmation — naming the count and noting that distribution can't be undone from here — and confirming sends the selection in a single batched request, with the Newspack UI snackbar reporting the result. Sites the post is already distributed to render checked and disabled with an "Already distributed" description for assistive tech, since distribution is additive only, matching the editor panel.

Layout and spacing come from Newspack UI's `stack` and `button` classes, so the modal follows the design system rather than a local stylesheet: the confirmation is a wide primary over a wide ghost, as on the reader-activation screens.

It reuses the existing REST route, meta, capability, and transport, so the sync mechanism itself is untouched — see "Changes to existing behavior" below for the shared-code changes that ride along. The modal and snackbar use the Newspack UI (`newspack-ui`) design system that is already loaded on every front-end page, so no React is added to public pages; the feature is gated on both the `newspack-ui` style and script handles being registered. Strings are translated through `@wordpress/i18n` (`wp-i18n` is declared as a dependency), so plural forms follow each locale's own rules. Structural pages (static front page, posts page) and syndicated (incoming) copies are excluded, and the action is gated per post on `edit_post`. `status_on_publish` follows the site's configured default distribution status, matching the editor.

### Changes to existing behavior:

Three changes touch code shared with the editor sidebar, the REST route and the CLI. All three are improvements, but one has a deploy-day effect on custom-configured sites:

1. **Distribution now requires `edit_post` on the post, not just the distribute capability.** On sites that grant `newspack_network_distribute` to a role without edit rights on the posts it distributes (a dedicated syndication role, or a customised contributor/author setup), those users lose distribution everywhere: the admin-bar button never renders, the editor sidebar control disappears, and direct REST calls return 403. Default role configurations are unaffected.
2. **Re-distributing to sites already in the config is a no-op instead of an `update_failed` error.** `update_post_meta()` returns `false` both on failure and when the value is unchanged; the two are now distinguished. This fixes a latent bug that also affected `pull_post()` and the CLI, and makes the migrator's workaround for it dead code (removed here).
3. **Distribution URLs are normalised with `untrailingslashit()` when stored**, so `https://node.test` and `https://node.test/` can no longer coexist as two entries for one site. This matches the canonical form the network utils already use.

1. On a hub (or node) with at least one network site, view a published, distributable post on the front end as a user with distribution rights.
2. Click **Distribute** in the admin bar and confirm the modal lists the network sites, with a select-all when there is more than one target, and that the Distribute button is disabled until something is selected.
3. Select one or more sites and press **Distribute**.
4. Confirm the modal asks "Distribute this post to N network sites? This can't be undone from here.", that **Back** returns to the list with the selection intact, and that no request has been sent yet.
5. Press **Distribute** and confirm a single request is sent, the modal closes, and the snackbar reports the result. Reopen the modal and confirm the distributed sites are now checked and disabled.
6. Confirm Escape, the overlay, and the close button each dismiss the modal, reset it to the site list, and return focus to the trigger.

### Other information:

Closes [NPPD-1710](https://linear.app/a8c/issue/NPPD-1710)


thomasguillot:
Heads-up: #702 (a hotfix targeting release) also adds `tests/unit-tests/content-distribution/test-api.php` with the same `Test\Content_Distribution\TestApi` class, covering dispatch-failure handling in `API::distribute()` plus a `mock-data-events.php` helper. It will merge first, so once release is promoted to main this branch will hit an add/add conflict on that file. The two test classes are complementary (permission-callback tests here, dispatch-failure tests there) and should be folded into a single class when resolving.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/feat/nppd-1710-front-end-distribution` (stash local work first).

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	/**
+	 * Add the distribute menu to the admin bar.
+	 *
+	 * @param WP_Admin_Bar $wp_admin_bar The admin bar instance.
+	 *
+	 * @return void
+	 */
+	public static function admin_bar_menu( $wp_admin_bar ) {
```
dkoo, line 218: **Suggestion:** Memoize the gating chain — `admin_bar_menu`, `enqueue_scripts`, and `render_modal` each run `should_render()`/`get_sites()` (and `get_sites()` re-runs `should_render()` internally), so the full gate chain — including `Outgoing_Post::get_distribution()` and the network-sites lookup — executes ~5-6× per front-end view for a permitted user. Object caching keeps it modest; a per-request memo would be cleaner, and `get_sites()` returning `[]` when not renderable already makes some explicit `should_render()` calls redundant.

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
```diff
@@ -0,0 +1,359 @@
...
+
+		return file_exists( $path ) ? filemtime( $path ) : false;
+	}
+
+	/**
+	 * Enqueue the front-end assets.
+	 *
+	 * @return void
+	 */
+	public static function enqueue_scripts() {
```
dkoo, line 378: **Suggestion:** Availability gate: style-vs-script handle divergence and implicit hook ordering — `newspack_ui_available()` checks only the newspack-ui *style* handle, while the enqueued JS hard-depends on the newspack-ui *script* handle — WordPress silently drops a script whose dependency is unregistered, so if the two handles ever diverge the button and modal still render while the behavior script vanishes: a visible Distribute button that does nothing, with no console error. The gate also holds only because core registers `wp_common_block_scripts_and_styles` before plugin callbacks at the same priority. Two cheap hardenings: also check `wp_script_is( 'newspack-ui', 'registered' )`, and hook `enqueue_scripts` at priority 11+ to make the ordering explicit.

plugins/newspack-network/includes/content-distribution/class-api.php
```diff
@@ -51,8 +51,8 @@ public static function register_routes(): void {
 						'default' => 'draft',
 					],
 				],
-				'permission_callback' => function () {
-					return current_user_can( Admin::CAPABILITY );
+				'permission_callback' => function ( $request ) {
+					return current_user_can( Admin::CAPABILITY ) && current_user_can( 'edit_post', $request['post_id'] );
```
dkoo, line 56: **Suggestion:** Disclose the three shared-code behavior changes — one has a deploy-day effect — The body's "it reuses the existing REST route … so nothing changes about how posts sync" overclaims: three shared-code changes ride along and affect the editor flow and CLI too. The one with deploy-day teeth: the permission tightening means a role granted the distribute capability but lacking `edit_post` on a given post (a dedicated syndication role, or a customized contributor/author setup) silently loses distribution everywhere — the admin-bar button never renders, the editor sidebar control disappears, and direct REST calls 403. Default role configs are largely unaffected, so this only surfaces on custom-configured sites. All three changes are sound and test-pinned; the fix is a short "Changes to existing behavior" list in the PR body so operators can anticipate the custom-role case.

plugins/newspack-network/src/content-distribution/admin-bar/index.js
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Pick the singular or plural form. @wordpress/i18n is not on the front end,
+ * so this is a two-form approximation.
+ *
+ * @param {{singular: string, plural: string}} forms Singular/plural templates.
+ * @param {number}                             count The count deciding the form.
+ * @return {string} The chosen template.
+ */
+const pluralize = ( forms, count ) => ( count === 1 ? forms.singular : forms.plural );
```
dkoo: **Suggestion:** Replace the two-form pluralize() with _n() — The `count === 1` approximation produces wrong grammar in locales with more than two plural forms (Russian, Polish, Arabic; French treats 0 as singular), and the count here is dynamic user-driven data. Since the bundle only loads for logged-in capability holders, adding `wp-i18n` to the (hand-maintained) dependency array plus `wp_set_script_translations()` and `_n()` costs nothing on public pages. If the current approach is a deliberate bundle-size trade-off, say so in the PR.

plugins/newspack-network/tests/unit-tests/content-distribution/test-outgoing-post.php
```diff
@@ -93,6 +93,22 @@ public function test_set_distribution() {
...
 	}
 
+	/**
+	 * Distributing to sites already in the config is a no-op, not an error.
+	 *
+	 * WordPress returns false from update_post_meta() when the value is unchanged, which would
+	 * otherwise surface as 'update_failed' on a retried request.
+	 */
+	public function test_set_distribution_is_idempotent() {
+		$urls = [ $this->network[0]['url'], $this->network[1]['url'] ];
```
dkoo, line 103: **Suggestion:** Add the genuine-write-failure companion test — The idempotency test pins the no-op half of the `set_distribution()` change, but nothing asserts that a *genuinely* failed write on a changed value still returns `update_failed` — a regression dropping the `$unchanged` guard entirely would pass the current suite. A companion test short-circuiting `update_post_metadata` via filter to return false for a changed distribution, asserting the WP_Error survives, closes the other half of the distinction.

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
```diff
@@ -0,0 +1,359 @@
...
+		}
+
+		$post = self::get_queried_post();
+		if ( ! $post || ! self::should_render( $post ) || ! self::newspack_ui_available() ) {
+			return;
+		}
+
+		$sites = self::get_sites( $post );
+		if ( empty( $sites ) ) {
+			return;
```
dkoo, line 264: **Nit:** Small cleanups in the admin-bar class — The `'meta' => [ 'menu_title' => … ]` config is dead — WP only uses `menu_title` as the aria-label of a parent node's submenu, and this node deliberately has no children (test-pinned). The modal's close button omits `type="button"` (inert here with no ancestor form; consistency only). And the `wp_enqueue_scripts`/`wp_footer` hooks omit explicit priorities while the `admin_bar_menu` hook states one.

plugins/newspack-network/src/content-distribution/admin-bar/index.js
```diff
@@ -0,0 +1,269 @@
...
+
+const FOCUSABLE = 'a[href], button:not([disabled]), input:not([disabled]), [tabindex]:not([tabindex="-1"])';
+
+/**
+ * Show a Newspack UI snackbar; no-op if the API is unavailable.
+ *
+ * @param {string} message The message.
+ * @param {string} type    'success' or 'error'.
+ */
+const notify = ( message, type ) => {
```
dkoo, line 68: **Nit:** Runtime-contract dependencies and UX honesty notes — Three related observations: close/teardown and `notify()` depend on newspack-ui.js runtime contracts (its `closeModal` observer, `window.newspackUI.notices.createNotice`) that this plugin can't test — there are no JS tests here — so a contract change no-ops teardown silently; disabled already-distributed rows convey no reason to assistive tech (an aria-describedby or title would help); and the "Distributed to N sites" snackbar fires on the server's 2xx, which records intent and enqueues async delivery rather than confirming it reached each node — matching the existing editor UX (pre-existing behavior surfaced in a new place), but worth remembering there's no failure surface or retry path if a node is unreachable.

plugins/newspack-network/tests/unit-tests/content-distribution/test-api.php
```diff
@@ -0,0 +1,96 @@
...
+	 * The 'distribute' route's permission callback.
+	 *
+	 * @var callable
+	 */
+	private $permission_callback;
+
+	/**
+	 * Set up.
+	 */
+	public function set_up() {
```
dkoo, line 50: **Nit:** Test-hygiene assortment — Three small items: `set_up()` nulls `$GLOBALS['wp_rest_server']` with no `tear_down()` restore (the plugin's own `test-integrity-check-endpoints.php` establishes the reset convention — mirror it); the permission tests omit the cap-absent-but-edit_post-present cell and the missing/zero-id cell (low risk given the symmetric `&&` and route regex); and several comments reference `get_localized_data()` where the helper is now `decode_localized_i18n()` — stale rename references.

dkoo (commented):
Release risk: LOW — zero blockers, the consequences pass came back clean, and the full modal flow is now browser-verified end-to-end (20/20 walkthrough assertions: single batched request, success snackbar, additive checked+disabled re-render, focus return on every close path, and the structural-page + logged-out gating). The one thing a solo env cannot cover is actual hub->node delivery, which needs a second live site. Note that the branch is CONFLICTING with main, so a rebase is required before merge. The feature itself is defensively built: escaping is airtight and test-pinned, the nonce wiring is correct, the focus trap is real, and the three embedded behavior changes are each improvements — the permission tightening closes a genuine gap, and the idempotent re-distribution fixes a latent bug that a migrator workaround (now dead code) had been papering over. Three things deserve attention before merge. First, disclosure: on sites that grant the distribute capability to a role without edit rights on the posts it distributes, this deploy silently removes distribution for those users across the editor and REST too — the body needs a "Changes to existing behavior" list. Second, the availability gate checks the newspack-ui style handle while the script hard-depends on the script handle, and rests on implicit same-priority hook ordering — a divergence renders a Distribute button whose modal does nothing, silently. Third, worth a product thought: distribution is additive-only with no undo, and this makes fanning a post out to every network site a two-click action from the live site — a confirmation step or irreversibility note in the UI copy would be cheap insurance. Also add the genuine-write-failure companion test so the idempotency guard can not silently regress. Approve-leaning once rebased.

- `plugins/newspack-network/includes/content-distribution/class-api.php:169` — **Suggestion:** Guard the get_post() dereference — `distribute()` dereferences `get_post( $post_id )` without a null check — currently unreachable with a missing post only because the *new* permission callback returns `do_not_allow` for non-existent IDs, i.e. the handler's null-safety is an incidental side effect of the auth change. A one-line `! $post ||` guard keeps the handler self-sufficient if the permission callback is ever relaxed or the post is deleted between check and dispatch.
- `plugins/newspack-network/includes/content-distribution/class-outgoing-post.php:152` — **Suggestion:** Normalize URLs in set_distribution() and fix the @param type — URLs are never normalized here, so `https://node.test` and `https://node.test/` can coexist in the meta (`array_unique` won't dedupe slash variants). The admin-bar flow compensates with an `untrailingslashit` comparison when disabling rows, but running `untrailingslashit` over the merged array inside `set_distribution()` fixes it for all entry points (editor, CLI, REST). Also, the `@param int[] $site_urls` docblock is wrong — they're strings.
- `plugins/newspack-network/includes/content-distribution/class-distributor-migrator.php:485` — **Nit:** Now-dead 'update_failed' workaround — The migrator's branch swallowing `update_failed` "if the post is already distributed" is dead code after the no-op change — and is itself evidence the change fixes a latent bug the migrator had to work around (the same latent double-distribute issue existed in `pull_post()` and the CLI; all three callers improve). Harmless to leave; cheap to clean up in a follow-up.

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
thomasguillot: Memoized. `get_sites()` now caches per post ID for the request, and `render_modal()`/`enqueue_scripts()` gate on the empty list rather than calling `should_render()` again — one gate chain per front-end view instead of five or six. `flush_cache()` covers tests and long-running processes.

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
thomasguillot: Both hardenings applied. `newspack_ui_available()` now requires `wp_script_is( 'newspack-ui', 'registered' )` as well as the style handle, and `enqueue_scripts` is hooked at priority 11 so it runs after core's `wp_common_block_scripts_and_styles()` instead of relying on callback order at 10. Pinned with a test that deregisters only the script handle.

plugins/newspack-network/includes/content-distribution/class-api.php
thomasguillot: Added a "Changes to existing behavior" section to the PR body covering all three, with the custom-role case first and the operational consequence spelled out.

plugins/newspack-network/src/content-distribution/admin-bar/index.js
thomasguillot: Switched to `_n()`. The bundle imports `@wordpress/i18n`, `wp-i18n` is declared as a script dependency, and `wp_set_script_translations()` points at the plugin's `languages/`. The `i18n` branch is gone from `wp_localize_script()` — the payload now carries request config only (`restUrl`, `nonce`, `defaultStatus`).

plugins/newspack-network/tests/unit-tests/content-distribution/test-outgoing-post.php
thomasguillot: Added `test_set_distribution_reports_a_genuine_write_failure()`: it short-circuits `update_post_metadata` to false for a site not already in the config, so the value genuinely changes, and asserts the `update_failed` WP_Error survives.

plugins/newspack-network/includes/content-distribution/class-admin-bar.php
thomasguillot: All three done: the `meta`/`menu_title` config is gone, the close button carries `type="button"`, and `wp_enqueue_scripts`/`wp_footer` now state their priorities (11 and 10).

plugins/newspack-network/src/content-distribution/admin-bar/index.js
thomasguillot: Two of the three are addressed. Disabled rows now carry `aria-describedby` pointing at a screen-reader-text "Already distributed", so the reason is announced rather than just the disabled state. And on the two-click fan-out: pressing Distribute now swaps the modal to "Distribute this post to N network sites? This can't be undone from here." with Back/Distribute, so nothing is sent until it is confirmed and Back keeps the selection intact.

The newspack-ui runtime contracts (`closeModal`, `window.newspackUI.notices.createNotice`) and the 2xx-means-enqueued snackbar are unchanged — both match the editor, and diverging one surface from the other would be the worse trade. Verified the whole flow live against a hub with four nodes: single batched request, status-specific snackbar, additive lock-in, focus return on every close path, no console errors.

plugins/newspack-network/tests/unit-tests/content-distribution/test-api.php
thomasguillot: Done, and the class is now folded together with the `TestApi` that landed in #702, as flagged earlier. `tear_down()` nulls `$GLOBALS['wp_rest_server']`, the permission tests gained the cap-absent-with-`edit_post` cell and the missing/zero-id cell, and the stale `get_localized_data()` references now name the helper by its current name.

thomasguillot:
Everything is addressed and the branch is rebased onto `main` — the add/add conflict on `test-api.php` is resolved by folding #702's dispatch-failure tests and this branch's permission-callback tests into one class. Inline threads are answered and resolved; the three notes that couldn't be posted inline are covered here.

**`class-api.php:169` — guard the `get_post()` dereference.** Added. `distribute()` now returns a 404 `WP_Error` for a missing post before touching `$post->post_status`, so the handler no longer depends on the permission callback's `do_not_allow`.

**`class-outgoing-post.php:152` — normalise URLs and fix the `@param` type.** `set_distribution()` now runs `untrailingslashit` over the merged array, so slash variants collapse for every entry point rather than only being compensated for at read time in the admin bar. `@param int[] $site_urls` is now `string[]`. Pinned with `test_set_distribution_normalizes_trailing_slashes()`.

**`class-distributor-migrator.php:485` — the dead `update_failed` workaround.** Removed rather than left for a follow-up; `set_distribution()` no longer returns that error for an already-distributed post, so the branch could only ever swallow a real failure.

**On the additive-with-no-undo point.** Took the confirmation route. Pressing Distribute now swaps the modal to "Distribute this post to N network sites? This can't be undone from here." with Back and Distribute; nothing is sent until it's confirmed, and Back returns to the list with the selection intact. The question is focused when the panel opens so it's announced before the buttons that answer it.

Verified live against a hub with four nodes: no request on the first press, single batched request on confirm, status-specific snackbar, additive lock-in on reopen, reset to the site list and focus back on the trigger for every close path, no console errors. Full `newspack-network` suite is green (259 tests), PHPCS, ESLint and Stylelint clean.

@dkoo ready for another look when you have a moment.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
```diff
@@ -165,12 +165,18 @@ public function set_distribution( $site_urls ) {
 		}
 
 		// If there are urls not already in the config, add them. Note that we don't support
-		// removing urls from the config.
-		$distribution = array_unique( array_merge( $distribution, $site_urls ) );
+		// removing urls from the config. Both sides are normalised to the network's
+		// canonical form first, or array_unique() would keep slash variants of one site.
+		$existing     = $distribution;
+		$distribution = array_values( array_unique( array_map( 'untrailingslashit', array_merge( $distribution, $site_urls ) ) ) );
```
dkoo: **Blocker — this rewrite can be vetoed by `maybe_short_circuit_distributed_meta()`, leaving legacy posts permanently un-distributable.** Normalizing the *existing* entries makes the new array look like a removal to the guard (it compares raw stored meta to the incoming array and blocks removals without `newspack_network_remove_distribution` intent), so `update_post_meta()` is short-circuited to `false` and this returns `update_failed` with nothing written. Reproduced live: stored `["https://node.test/"]` + a new target fails on HEAD, succeeds pre-commit — and the dual-variant shape the pre-fix code produced (`["https://node.test/","https://node.test"]`) is itself a trigger. Don't rewrite existing entries; append only URLs whose normalized form isn't already present:

```php
$normalized = array_map( 'untrailingslashit', $distribution );
foreach ( $site_urls as $url ) {
	$url = untrailingslashit( $url );
	if ( ! in_array( $url, $normalized, true ) ) {
		$distribution[] = $url;
		$normalized[]   = $url;
	}
}
```

(Alternative: teach the guard to allow normalization-only diffs.) Please also add a test seeding the mixed legacy shape — the current suite passes precisely because it never stores a slashed variant first.

plugins/newspack-network/includes/content-distribution/class-distributor-migrator.php
```diff
@@ -482,11 +482,7 @@ public static function migrate_subscription( $subscription_id, $migrate_incoming
 			return new WP_Error( 'outgoing_post_error', $e->getMessage() );
 		}
 		$distribution = $outgoing_post->set_distribution( [ $network_url ] );
```
dkoo, line 484: **Blocker (same root cause) — the removed `update_failed` tolerance is not dead code.** The premise that `set_distribution()` can no longer return `update_failed` for an already-distributed post holds only for the same-string idempotent case; the normalization-veto reproduction shows a live `update_failed` path for slash-variant meta. With the tolerance gone, `migrate_subscription()` aborts here before clearing `dt_subscriptions`/`dt_connection_map`, leaving the migration half-done for exactly the posts the old comment covered. Safe to keep this removal only after the normalization fix lands — then it's genuinely dead.

dkoo (changes requested):
Round-2 delta review of the feedback commit `6c02c4d86` (the rebase itself was verified content-identical via `range-diff`, so scope is this commit alone).

Requesting changes on one blocker — **confirmed with a live reproduction, not reasoning**:

**The normalization can be vetoed by the distributed-meta removal guard, permanently blocking distribution for legacy posts.** `set_distribution()` now rewrites the whole stored array through `untrailingslashit`, but `Content_Distribution::maybe_short_circuit_distributed_meta()` vetoes any write that removes an existing entry unless removal-intent meta is present. Collapsing `https://x/` into `https://x` reads as a removal → `update_post_meta()` returns `false` → `WP_Error( 'update_failed' )`, **nothing is written**, and the REST endpoint returns 400. Reproduced in an isolated test DB: stored `["https://node.test/"]` + `set_distribution(["https://other-node.test"])` fails on HEAD and succeeds on the pre-commit source. Worst case: the dual-variant shape the *old* code produced (the round-1 duplicate bug — `["https://node.test/","https://node.test"]`) is itself a trigger, so every post that already hit the round-1 bug becomes permanently un-distributable once this ships. Fix suggestion inline.

The same root cause invalidates the migrator change: removing the `update_failed` tolerance was premised on that error being unreachable for already-distributed posts, and the reproduction shows it is live. `migrate_subscription()` now aborts before clearing `dt_subscriptions`/`dt_connection_map` — a half-done migration for exactly the posts the removed comment covered.

Everything else in the commit verified well: the normalization test pin is genuine (fails on old code), read-side back-compat in the admin bar was correctly kept, and the confirm-before-distributing flow is sound — double-submit protection is synchronous, the focus trap correctly excludes the hidden panel, all JS writes use `textContent`, escaping/nonce/capabilities unchanged. Full newspack-network suite green (259 tests, 771 assertions — the legacy-meta shape above is precisely what it lacks); PHPCS clean. The `get_post()` 404 guard is real at HEAD but note it landed inside merge commit `076ebbe53` rather than a reviewable feature commit, and has no test pin — fine as defence-in-depth (the permission callback already shields it), but merge commits shouldn't carry functional changes.

**Suggestion:** `remove_distribution()` (untouched here) is now inconsistent with the normalized write side — it `array_search`es the raw `$site_url` and writes removal-intent meta unnormalized, so a slashed inbound value from `class-network-incoming-post-deleted.php` fails with `site_not_found`. Normalize there in the same pass.

Nits, take or leave: `confirmPanel`/`confirmMessage`/`confirmSubmit`/`backButton` are dereferenced unguarded while sibling optional nodes are null-checked; the failure-path `showSelection(); submit.focus();` can target a closed modal (gate on `data-state`); the "disabled fieldset" comment is stale (busy state moved to the buttons); `wp_set_script_translations()` is currently a no-op — `languages/` has only the `.pot`, no `.json` — so the i18n plural-form benefit is theoretical until `wp i18n make-json` runs; WP style prefers `’` over the straight apostrophe in "can't".

Review provenance: 2 of 2 reviewer passes (lineup: deep, codex); release risk: ELEVATED.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
thomasguillot: Fixed in 328f4475a. `set_distribution()` now leaves stored entries byte-for-byte and appends only urls whose normalised form isn't already present, so the guard's diff stays empty and the write goes through. Only the comparison is normalised, which is what stops a slash variant becoming a second entry.

Pinned with three tests that seed the legacy shape into the meta directly rather than going through `set_distribution()`: a single slashed entry, both variants of one site, and the no-duplicate case. All three fail on `6c02c4d86` and pass now.

Reproduced the whole thing through the UI as well, on a post whose stored meta I seeded as `["https://nppd1710node.test/"]`: distributing to a new site returns 200 and the meta ends up `["https://nppd1710node.test/","https://metro.test"]`, the slashed entry untouched.

plugins/newspack-network/includes/content-distribution/class-distributor-migrator.php
thomasguillot: Right on both counts. The removal stays, and it's genuinely dead now that 328f4475a landed: `set_distribution()` can only return `update_failed` on a real write failure, so `migrate_subscription()` reaches the `dt_subscriptions`/`dt_connection_map` clean-up for every post the old comment covered.

thomasguillot:
All of round 2 is addressed in 328f4475a. The two blockers are answered on their threads; here are the rest.

**`remove_distribution()` is inconsistent with the normalised write side.** Fixed. It now matches on the canonical form, so a slashed inbound value from `class-network-incoming-post-deleted.php` finds its entry. The removal-intent meta carries the **raw stored** value rather than the canonical one, or the guard vetoes the removal for the same reason the write was being vetoed. Pinned with `test_remove_distribution_matches_a_slashed_url()`, which fails on `6c02c4d86`.

**`confirmPanel`/`confirmMessage`/`confirmSubmit`/`backButton` dereferenced unguarded.** Fixed. All six required nodes are now bailed on together, and the three genuinely optional ones (`dialog`, `overlay`, `selectAll`) moved below the guard so they're looked up only once we know we're driving our own markup. That ordering is also what `@wordpress/no-unused-vars-before-return` wants.

**The failure path can target a closed modal.** Fixed. `showSelection()` and `submit.focus()` now run only when `data-state` is still `open`; if the modal was closed mid-request, `closeModal` has already reset it and returned focus.

**Stale "disabled fieldset" comment.** Rewritten to describe what the filter actually does now, which is to keep the hidden panel and any disabled control out of the tab ring.

**`wp_set_script_translations()` is a no-op.** Leaving it. `languages/` holding only the `.pot` is the same situation as `newspack-blocks` and `republication-tracker-tool`, which make the identical call, so removing it here would just make this one inconsistent with the house pattern for no gain.

**Straight apostrophe in "can't".** Now `’`.

**The `get_post()` 404 guard has no test pin.** Added `test_distribute_returns_404_for_a_missing_post()`, which asserts the handler returns the 404 itself rather than leaning on the permission callback having refused first.

Verification: full `newspack-network` suite is 265 tests, with the one pre-existing failure being `test_toggle_jetpack_photon`, which fetches a remote image and fails identically on a clean tree. PHPCS, ESLint and Stylelint clean. Walked the modal live on a hub with four nodes against a post seeded with the legacy shape: single request on confirm, 200, additive lock-in on reopen, success snackbar, no console errors.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
```diff
@@ -164,13 +164,32 @@ public function set_distribution( $site_urls ) {
...
+		// If there are urls not already in the config, add them. Note that we don't
+		// support removing urls from the config. Entries already stored are left
+		// byte-for-byte as they are: rewriting them to the network's canonical form
+		// reads as a removal to Content_Distribution::maybe_short_circuit_distributed_meta(),
+		// which vetoes the write, so a post holding a slashed url from before this
+		// normalisation could never be distributed again. Only the comparison is
+		// normalised, so slash variants of one site can't produce a second entry.
+		$existing   = array_values( $distribution );
+		$normalized = array_map( 'untrailingslashit', $existing );
+
```
dkoo, line 176: **Suggestion — legacy URLs are now deliberately kept raw in meta, but they're also dispatched raw.** `get_payload()` sends `$this-&gt;get_distribution()` verbatim (line 315), and `Incoming_Post` accepts only an exact `in_array( get_bloginfo( 'url' ), $payload['sites'], true )` match — so for a post holding `https://node.test/`, the source reports a successful distribution while that recipient rejects the event with `not_distributed_to_site`.

To be clear this is **pre-existing, not a regression here**: `origin/main`'s `set_distribution()` does `array_unique( array_merge( … ) )` with no normalization, so raw slashed URLs have always been stored and dispatched. The round-2 approach would have masked it as a side effect of rewriting stored meta — the approach that turned out to be unshippable. Since this PR consciously makes "stored stays raw" the contract, normalizing at payload-build time is the natural companion:

```php
'sites' =&gt; array_map( 'untrailingslashit', $this-&gt;get_distribution() ),
```

Keeps the meta byte-for-byte for the guard while making the wire format canonical. Reasonable to split into a follow-up if you'd rather keep this PR tight — but it's worth not losing, since the failure is silent on the sending side.

dkoo (commented):
Round-3 delta review of `ec211e425` + `328f4475a`.

**Both blockers are resolved — verified by re-running the round-2 reproduction against both heads, not by reading the diff.** `set_distribution()` is now append-only: existing entries stay byte-for-byte and only the comparison is normalized. All three cases that hard-failed last round now pass:

| case | 6c02c4d86 | 328f4475a |
|---|---|---|
| legacy `["https://node.test/"]` + new target | vetoed → `update_failed`, nothing written | **success** → `["https://node.test/","https://other-node.test"]` |
| dual-variant `["…/","…"]` + new target | vetoed → `update_failed` | **success**, both legacy entries preserved |
| legacy + re-save same site | vetoed → `update_failed` | **valid no-op** |

Blocker 2 resolves as a *consequence* rather than by restoring the tolerance, which is the better outcome: an already-distributed post now yields `$existing === $distribution`, so `$unchanged` short-circuits the `WP_Error` and the only remaining `update_failed` is a genuine write failure — which *should* abort a migration. The two new legacy-shape tests seed the on-disk shape directly and genuinely pin the veto.

`remove_distribution()` got the subtle half right: canonical-form matching for the lookup, but the **raw stored value** written as removal intent — which is exactly what the guard's raw `array_diff` comparison requires. Easy detail to get backwards; you didn't.

All four nits taken, and the early-return guard's comment distinguishing required from genuinely-optional nodes is a nicer resolution than the mechanical null-checks I suggested. The UI-stacks layout commit reads clean — the `[hidden]` override and the `offsetParent`-based focus filter both survive the SCSS restructure.

One new suggestion inline (pre-existing, not a regression from this PR). The i18n `.json` gap from last round is unchanged — still build tooling, still fine to defer.

The change request is resolved from my side; a maintainer can dismiss it.

Review provenance: 1 of 2 reviewer passes (lineup: deep, codex); release risk: LOW (degraded — skipped: deep).

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
thomasguillot: Taken here rather than as a follow-up, since the failure is silent on the sending side.

`get_payload()` now maps `untrailingslashit` over the distribution, so stored entries stay byte-for-byte for the removal guard while the wire format is canonical. Pinned by `test_get_payload_normalizes_slashed_sites`, which seeds the legacy slashed shape on disk and asserts both halves of the contract: the meta still reads back slashed, the payload goes out canonical. It fails on the pre-commit source with `https://node.test/` vs `https://node.test`.

Full suite green apart from `test_toggle_jetpack_photon`, which fails identically at `328f4475a` — it sideloads an image over the network. PHPCS clean.

dkoo (commented):
This delta lands cleanly, and it closes out the round-2 blocker for good. The canonical-form change is scoped to the outbound payload only: `get_payload()` maps `untrailingslashit` over a fresh read of `get_distribution()` and feeds the wire copy — it never writes meta, so the removal-guard veto that made legacy posts un-distributable is not back. I re-checked `set_distribution()` at this head: byte-identical to the round-3 append-only fix, existing entries left untouched.

It also fixes a real sender/receiver mismatch rather than just tidying. The node matches the incoming payload against `get_bloginfo( 'url' )` with a strict comparison, and that value has no trailing slash — so a legacy target stored as `https://node.test/` was sent raw, rejected by the node as `not_distributed_to_site`, while the sender's Data Event still reported success. Normalizing the sent list closes that gap. The new test pins it: it seeds slashed meta directly, then asserts the stored array stays raw while the payload comes out canonical.

Verified: the full newspack-network suite is green (266 tests, 785 assertions) in the isolated env, and Codex found nothing.

Release risk: LOW. From this run's side the round-2 change request is resolved and confirmed across two rounds — it can be dismissed whenever a maintainer is ready.

Review provenance: deep + Codex over the single-commit delta; suite run in the PR's env.

thomasguillot:
Round 4 reads as a sign-off but was submitted as a comment, so the round-2 change request is still the standing decision and the PR has no approval against the `main` ruleset. @dkoo could you re-submit as Approve when you get a moment? That clears both in one go.

dkoo (approved):
Manual QA confirms distribution end to end: from the hub's admin-bar Distribute control, a post lands as a rendered copy on a real linked node. Together with the code review across its rounds, that closes #679 out.

The round-2 change request is resolved. The normalization-veto blocker — where rewriting the stored distribution array read as a removal to `maybe_short_circuit_distributed_meta()` and returned `update_failed` with nothing written — was fixed with append-only normalization and confirmed across the two rounds after. The follow-up canonical-form commit narrows the change to the outbound payload only, leaves stored meta byte-identical, and closes the sender/receiver URL mismatch that would otherwise drop legacy trailing-slash targets silently. The full newspack-network suite is green (266 tests, 785 assertions), and the confirm-before-distributing flow — double-submit protection, focus trap, text-only DOM writes, unchanged nonce/capabilities — held under review.

Release risk: LOW. Approving.

Review provenance: deep + Codex across the review and its delta rounds, plus manual end-to-end QA on a hub and a real linked node.

PR #681, opened 2026-07-21
fix(content-gate): apply third-party content filters to gated teasers (NPPD-2096)

adekbadek:
### All Submissions:

`Content_Gate::handle_restricted_content()` ran on `the_content` at `PHP_INT_MAX` and **discarded the content it was handed**, returning a teaser rebuilt from the raw post:

```php
public static function handle_restricted_content( $content ) {
    if ( ! isset( self::$restricted_content[ get_the_ID() ] ) ) { return $content; }
    return self::$restricted_content[ get_the_ID() ];   // incoming $content discarded
}
```

Integrations that gate their own embeds through `the_content` therefore had their output thrown away on exactly the posts that are paywalled. The known case is the [Everlit audio player](https://help.newspack.com/engagement/reader-activation-system/content-gating/#development), which asks us whether a post is restricted and then removes its own iframe at priority 999999: the reader hits the gate for the article text while the player above it serves the complete audio version of the same article.

Woo Memberships restricted content at priority **999**, i.e. *before* those callbacks, so their gating ran last and survived into the page. The ordering only inverts once a site moves to Access Control, which is why this surfaces as a migration regression rather than a new bug.

This PR:

* Substitutes the teaser at a new `Content_Gate::RESTRICTION_PRIORITY` of **999**, so later `the_content` filters receive the teaser instead of the full restricted post.
* Appends the gate HTML at `PHP_INT_MAX`, so no callback registered before this class hooks in can rewrite the gate markup itself.
* Falls back to the stored teaser + gate when the substitution has not run (another plugin removed the filter), rather than appending the gate to whatever content is in hand — core hands this chain the unrestricted body, so appending there would publish the restricted post.
* Tracks a pending substitution **per `the_content` pass, keyed by that pass's nesting depth**, because `the_content` nests: an inner pass must not consume the outer pass's state, and a nested render of a *different* post must not close a substitution it never opened. Neither filter is exception-safe, so the depth is also what keeps the bookkeeping honest: an entry can only be read by another pass at that exact depth, the substitution claims that slot on its way in, and the append trusts an entry only while the substitution filter is still registered. A pass aborted mid-chain therefore leaves nothing that a later pass can be gated by.

Note the boundary this draws, which is documented on the constant: callbacks at or below priority 999 still receive the full restricted post and their output is still replaced. That matches Memberships' behaviour, but an integration gating its embeds at the default priority of 10 is not covered.

Closes NPPD-2096.

1. On a site with `NEWSPACK_CONTENT_GATES` enabled and Woo Memberships **deactivated**, create a content gate that restricts a category and requires registration.
2. Add a plugin (or a snippet) that filters `the_content` at a priority above 999 and rewrites a marker in the post body — this stands in for a partner integration gating its own embed. For example:
   ```php
   add_filter( 'the_content', fn( $c ) => str_replace( 'PARTNER_EMBED', 'PARTNER_EMBED_GATED', $c ), 999999 );
   ```
3. Publish a post in the restricted category containing that marker plus several paragraphs.
4. View the post logged out, in an incognito window. **Before this PR** the marker is rendered unchanged (the filter's output was discarded). **After**, the marker is rewritten, the gate still renders below the teaser exactly once, and the paragraphs past the teaser are absent.
5. Log in as a reader who satisfies the gate and confirm the full post renders with no gate.
6. Reactivate Woo Memberships and confirm behaviour is unchanged there.
7. Run `n test-php --filter Test_Content_Gates`.

Verified end-to-end in an isolated env against a real third-party integration (the unmodified Everlit plugin): the audio player is paywalled for logged-out readers on gated posts, plays for readers with access, and Memberships-backed gating is unaffected. Full PHP suite: 2121 tests, 0 failures.

### Other information:

Eight regression tests were added. The first three reproduce the reported bug and fail against the previous wholesale replacement; the rest pin invariants of the new mechanism, which had no previous behaviour to fail against — each of those was verified to fail against an implementation with the corresponding safeguard removed.

| Test | Pins |
| -- | -- |
| `test_third_party_content_filters_apply_to_restricted_teaser` | A filter at the real 999999 sees the teaser and its output is rendered |
| `test_nested_content_filtering_still_renders_partner_output` | A nested `the_content` pass does not consume the outer substitution |
| `test_nested_render_of_another_post_does_not_consume_the_substitution` | A nested render of a different post does not close this post's substitution, nor inherit its gate |
| `test_gate_html_is_not_exposed_to_third_party_content_filters` | Later callbacks never receive the gate markup |
| `test_repeated_content_filtering_appends_gate_once_each_pass` | Rendering the same post twice in a request appends one gate each pass |
| `test_restricted_content_falls_back_when_teaser_substitution_is_removed` | With the substitution removed, content in hand is not published and the gate renders once |
| `test_content_filter_exception_leaves_nothing_that_corrupts_later_renders` | A pass aborted by a throwing callback leaves nothing a later pass can be gated by |
| `test_aborted_pass_does_not_gate_a_later_unrestricted_render` | A post that is not restricted is never gated by what an aborted pass left behind |

https://claude.ai/code/session_01GZHTMVoqyN2pwq7haoMkQm


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/fix/nppd-2096-third-party-content-filters` (stash local work first).

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -233,30 +265,87 @@ public static function restrict_post( $post, $query ) {
...
 	public static function handle_restricted_content( $content ) {
-		if ( ! isset( self::$restricted_content[ get_the_ID() ] ) ) {
+		$post_id = get_the_ID();
+
+		// Only close a substitution that this same pass opened for this same post.
+		// A later filter may render a different post through 'the_content' –
+		// related posts, summaries – and that inner pass pushes nothing; popping
+		// for it would append this post's gate to unrelated content and leave the
+		// outer pass discarding everything downstream of the substitution.
+		if ( ! empty( self::$teaser_stack ) && end( self::$teaser_stack ) === $post_id ) {
```
dkoo: **Suggestion:** Harden the substitution stack against stale entries — Three related edges share one root — a pushed stack entry whose matching pop is skipped: (a) an exception thrown by a callback between priority 999 and PHP_INT_MAX (caught upstream) orphans the entry — bounded to per-request stack growth; (b) a callback that swaps the global post without restoring it makes the pop's ID check miss — that pass renders the teaser with **no gate appended**, a silent lost-conversion mode with no error signal; (c) compounding, a stale top entry plus the substitution filter being removed later in the same request lets a subsequent pass match the stack branch for a pass that never substituted, appending the gate to the unrestricted in-hand body — a full-content leak that requires two independent misbehaviors. All three close cheaply: purge stale same-post entries at the top of `replace_restricted_content()` before pushing, and/or require evidence the substitution ran in *this* pass before trusting the stack match (or pop by value-match). At minimum, a comment acknowledging the non-exception-safe push.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -3151,4 +3153,231 @@ public function test_duplicate_gate_endpoint_rejects_cross_wizard_gates() {
 		$this->assertNotWPError( $success );
```
dkoo, line 3191: **Suggestion:** Add a test for the exception/dirty-stack path — The regression suite covers the substitution matrix thoroughly (nesting, foreign-post renders, repeat passes, gate non-exposure, removed-substitution fallback) — the one cell left unpinned is the exception path: a the_content callback that throws mid-pass and is caught, followed by an assertion that a subsequent restricted render still gates correctly and the stack is clean. That's the one area where safety currently rests on unreachability reasoning rather than a test.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -3151,4 +3153,231 @@ public function test_duplicate_gate_endpoint_rejects_cross_wizard_gates() {
...
+		$this->assertStringNotContainsString( 'newspack-content-gate__inline-gate', $nested_output, "The unrestricted post's render should not receive the restricted post's gate" );
+		$this->assertStringContainsString( 'PARTNER_EMBED_GATED', $rendered, 'Partner filtering should survive a nested render of another post' );
+		$this->assertSame( 1, substr_count( $rendered, 'newspack-content-gate__inline-gate' ), 'The gate should be appended exactly once' );
+	}
+
+	/**
+	 * Rendering the same restricted post twice in one request must not accumulate
+	 * gates or leave state behind that changes the second result.
+	 */
+	public function test_repeated_content_filtering_appends_gate_once_each_pass() {
```
dkoo, line 3376: **Suggestion:** Correct the PR body's test-count claim — The body says "5 regression tests, each verified to fail against pre-fix behavior" — the diff adds six test methods, and at least two cannot fail pre-fix by their assertions alone (`test_repeated_content_filtering_appends_gate_once_each_pass` passes against the old wholesale replacement; the gate-exposure test only fails pre-fix via the fatal on the then-nonexistent constant). They're legitimate invariant guards worth keeping exactly as written — the fix is to the PR body: six tests, three pin the bug, three pin invariants of the new mechanism, so the verification story reads accurately.

dkoo (commented):
Release risk: LOW — zero blockers, consequences pass found none, impact is compatible, and verification is now complete: the fix is demonstrated live in an isolated env (a 999999-priority the_content filter's output survives into the gated teaser, the gate renders exactly once, and no restricted paragraph leaks into the markup — confirmed in view-source), the control and satisfied-reader paths behave, and Woo Memberships parity holds by construction (the PR's filters are inert when Memberships is active, matching main). The mechanism held up to adversarial tracing: no path emits the full restricted body, the fallback is fail-closed, page-cache interaction is benign, and feeds/REST/excerpt surfaces are untouched. What remains is optional hardening of the per-post substitution stack, where three related edges compound (an exception between the two filter priorities orphans a stack entry; a callback swapping the global post without restoring it skips the pop, rendering the teaser gateless; a stale entry plus later filter removal could append the gate to the unrestricted body — a leak requiring two independent misbehaviors). All close with cheap hardening plus one dirty-stack test. Also correct the PR body's test-count claim (six tests: three pin the bug, three pin invariants).

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -59,11 +59,42 @@ class Content_Gate {
...
-	 * Restricted content per post ID.
+	 * Rendered pieces of each restricted post, keyed by post ID: the teaser and the
+	 * gate HTML. Held separately so the teaser can be handed to the remaining
+	 * 'the_content' filters without exposing the gate HTML to them.
 	 *
-	 * @var string[]
+	 * @var array<int, array{teaser: string, gate: string}>
 	 */
-	private static $restricted_content = [];
+	private static array $restricted_content = [];
```
dkoo, line 68: **Nit:** Assorted consistency nits — Three small items: the typed-property additions (`array` on two statics) leave the class's five sibling statics untyped — the codebase hasn't adopted typed properties, so either drop or accept knowingly; the test class still has no reachable `@group` (a class-level `@group content-gate` would have been cheap while touching the file); and the fallback teaser skips the >999 formatting filters (wpautop) — cosmetic and pre-existing behavior, noted for completeness.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -233,30 +265,87 @@ public static function restrict_post( $post, $query ) {
...
+	 * remaining 'the_content' filters still run over it.
+	 *
+	 * Third-party integrations gate their own embeds on 'the_content' – the Everlit
+	 * audio player, which registers at priority 999999, is the known case. Handing
+	 * them the teaser lets their gating compose with the paywall the way it did
+	 * under Woo Memberships, whose restriction filter ran at 999. Returning the
+	 * full post here and discarding the filtered result instead would leave those
+	 * embeds ungated on restricted posts.
+	 *
+	 * The gate HTML is deliberately excluded; {@see self::handle_restricted_content()}
```
dkoo, line 324: **Nit:** Docblock slightly overstates the PHP_INT_MAX guarantee — "No callback after this priority can rewrite the gate markup" — strictly, a callback also registered at PHP_INT_MAX but added after init runs later in the same bucket (equal priorities run in registration order) and does receive the gate HTML. Pathological edge, but this comment is the documented security boundary; "no callback at a lower priority, nor any PHP_INT_MAX callback registered before init" is one clause more precise.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -3151,4 +3153,231 @@ public function test_duplicate_gate_endpoint_rejects_cross_wizard_gates() {
...
+		$this->assertStringNotContainsString( '<p>PARTNER_EMBED</p>', $rendered, 'The unfiltered embed should not survive into the page' );
+		$this->assertStringContainsString( 'newspack-content-gate__inline-gate', $rendered, 'The gate should still be appended' );
+		$this->assertStringNotContainsString( 'Hidden paragraph', $rendered, 'Content past the teaser should stay restricted' );
+	}
+
+	/**
+	 * A later filter that runs 'the_content' itself must not consume the outer
+	 * pass's state and leave the outer pass discarding everything downstream of it.
+	 */
+	public function test_nested_content_filtering_still_renders_partner_output() {
```
dkoo, line 3286: **Nit:** Nested same-post test doesn't assert the inner pass's return — `test_nested_content_filtering_still_renders_partner_output` runs an inner `apply_filters( 'the_content', 'UNRELATED' )` but never asserts its return — which is actually teaser+gate, because both callbacks key off `get_the_ID()`, not the content in hand. That matches pre-change behavior, but an explicit assertion would document that same-global-post string passes are intentionally swallowed, and distinguish this test from the adjacent other-post test that does assert the inner output.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Took the real fix rather than the comment. The bookkeeping is now a map keyed by the nesting depth of the `the_content` pass that opened the substitution (`$pending_gates`, replacing `$teaser_stack`), which closes all three edges in b0c62bc:

- The depth is read from `$wp_current_filter`, so an entry can only ever be read by another pass at that exact depth — growth is bounded to one entry per depth rather than one per aborted pass (a).
- `replace_restricted_content()` claims its slot unconditionally on the way in, so a leftover is overwritten rather than mistaken for the pass now running.
- The append reads the post ID from the entry instead of `get_the_ID()`, so a callback that swaps the global post without restoring it no longer misses the match and drops the gate (b).
- The entry is trusted only while `replace_restricted_content()` is still registered: without it no pass can have substituted, so the body in hand is the unrestricted post and the stored teaser + gate is returned instead (c).

Two tests cover it, each verified red against the implementation with the corresponding safeguard removed: `test_content_filter_exception_leaves_nothing_that_corrupts_later_renders` (throwing callback, caught, then a pass with the substitution removed must not have the gate stamped onto the body in hand) and `test_aborted_pass_does_not_gate_a_later_unrestricted_render`. The push is still not exception-safe and the docblock on the property says so — the design is that what it leaves behind is unreachable, not that it never happens.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
adekbadek: Added in b0c62bc, as two tests sharing an `abort_restricted_render()` helper (a callback throwing `DomainException` mid-pass — not a `RuntimeException`, since PHPUnit's own failure exceptions extend that and catching them would swallow a failing assertion):

- `test_content_filter_exception_leaves_nothing_that_corrupts_later_renders` — after the aborted pass, a pass with the substitution removed must not be gated as if it had substituted, then a normal render still produces exactly one gate and leaves no substitution open.
- `test_aborted_pass_does_not_gate_a_later_unrestricted_render` — a post that isn't restricted must not pick up the gate the aborted pass left pending.

Both were verified red against the implementation with the corresponding safeguard removed. One detail worth flagging: core pops the hook off `$wp_current_filter` only on the way out, so the exception leaves the stack a level deep; the helper restores it, both to keep that out of the assertions and because the harder case is the one where later passes run at the same depth the aborted pass did and can actually reach what it left behind.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
adekbadek: Right on both counts — the body now lists all eight tests (the six, plus the two exception-path ones added here) and splits them: the first three reproduce the reported bug and fail against the previous wholesale replacement, the rest pin invariants of the new mechanism, which had no previous behaviour to fail against. Each of those was verified to fail against an implementation with the corresponding safeguard removed, which is the honest version of the red/green claim. Also corrected the suite count to 2121.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: 1. Typed the remaining five statics in the class instead of dropping the two — the class is now internally consistent, and the direction of travel is the one that catches more rather than less (`newspack-ui` already does this, and the plugin requires PHP 8.3). All assignments are literal `true`/`false` and array literals, so nothing changes at runtime.
2. Added `@group content-gate` at class level, matching `feed-restriction.php` in the same directory. `n test-php --group content-gate` now picks the class up.
3. The fallback teaser is run through `newspack_gate_content`, which mirrors core's `the_content` formatting (`do_blocks`, `wpautop`, `wptexturize`, shortcodes), so it isn't unformatted — what it skips is any third-party callback above 999, and that is unavoidable in that branch: those callbacks have already run, over the full body, and their output is precisely what must not be published. Left as is.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Corrected in b0c62bc. The docblock now reads: neither a callback at a lower priority nor one registered at `PHP_INT_MAX` before this class hooks in can rewrite the gate markup; a `PHP_INT_MAX` callback registered afterwards does run last within that bucket and does see the gate. Added the consequence too, since this is the documented boundary — nothing but registration order separates the two, so it holds against ordinary integrations, not against a plugin that means to reach the markup.

plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
adekbadek: Asserted in b0c62bc: the inner output contains no `UNRELATED` and closes with exactly one gate, with a comment saying why — both callbacks key off the post being rendered, not the string in hand, so a bare string run through `the_content` while the global post is still the restricted one comes back as that post's teaser and gate. Swallowing it is long-standing behaviour; the assertions are there to show the inner pass resolves on its own terms rather than by borrowing, or stranding, the outer one's, which is also what distinguishes it from the adjacent other-post test.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -233,30 +273,129 @@ public static function restrict_post( $post, $query ) {
...
+		//
+		// The post is taken from the entry rather than from get_the_ID(), which a
+		// callback may have moved off the post whose teaser is in hand; and the
+		// entry is trusted only while the substitution is still registered, since
+		// without it no pass can have substituted and the body in hand is the
+		// unrestricted post.
+		if (
+			null !== $substituted_id
+			&& isset( self::$restricted_content[ $substituted_id ] )
+			&& has_filter( 'the_content', [ __CLASS__, 'replace_restricted_content' ] )
```
dkoo, line 384: **Nit:** Awareness note: the pass-presence proxy's theoretical limit — `has_filter()` at PHP_INT_MAX proving the substitution ran this pass holds for every ordinary execution (priorities run ascending; WP won't revisit a passed priority). The one contrived defeat — a >999 callback re-adding the substitution filter mid-chain, combined with a stale same-depth entry, a removed-then-re-added filter, and a restricted post with the full body in hand — requires four coincident manipulations of this class's own private filters and is not reachable by ordinary integrations; every non-re-add variant falls through to the fail-closed stored teaser+gate. Recording for posterity next to the existing exception-safety concession in the docblock; no action needed.

dkoo (commented):
Release risk: LOW — the follow-up closes all three edges from the review with a mechanism stronger than any single suggested hardening: pending gates are keyed to the content-filter pass itself (nesting depth from the live filter stack), claimed on entry so a caught exception's leftovers are overwritten, and the append requires the substitution callback to be present for the pass — which, given ascending priority execution, proves the teaser substitution ran. The silent no-gate mode and the two-misbehavior leak are both structurally gone, the fallback still fails closed in every combination, and the new exception-path test reproduces the exact dirty-stack scenario and genuinely fails under the old code. One awareness note, not an action item: the pass-presence proxy can theoretically be defeated by a callback re-adding the substitution filter mid-chain combined with three other coincident manipulations of this class's own private filters — far beyond the two-misbehavior bar that shaped the original finding, and every non-re-add variant degrades to the fail-closed fallback. The PR body's test claims were also corrected. Nothing further needed from this review's perspective.

dkoo (approved):
Manual QA confirms the behavior: third-party filters hooked on `the_content` now run against gated content, so a filter's output shows up whether or not the gate truncates the post. Verified in an isolated env with a filter registered at priority 999999 — its rewrite and appended markup appeared on the gated teaser while logged out, on the full post for a registered reader, on an ungated control post, and with WooCommerce Memberships active. The gate still renders once, and the teaser truncates where expected.

Release risk: LOW. The change stays within content-gate output filtering, and `Test_Content_Gates` passes (92 tests). This branch doesn't carry the vipwpcs CI change under discussion on #693 and #726, so it can merge independently of that decision.

Approving.

Review provenance: deep + standards + WP expert + Codex across the review and delta rounds.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Recorded in d59c6d9, next to the exception-safety concession in the `$pending_gates` docblock — that scenario deserved to be written down rather than just noted here, since the inline comment above the `has_filter()` check was stating the proxy as absolute ("without it no pass can have substituted"), which is what would mislead someone later reasoning about a change to it.

The docblock now says what depth alone cannot establish, why the registration check stands in for it in any ordinary execution, and names the four coincident conditions that defeat it — filter removed before 999, re-added by a callback above it, a leftover entry at the same depth, on a restricted post — plus the point that everything short of all four falls through to the stored teaser and gate. The inline comment is qualified and points at the property. No behavior change; full suite green.

PR #682, opened 2026-07-21
feat(env): report seeded asset age and add --rebuild to env up

miguelpeixe:
### All Submissions:

`n env up --build` seeds `node_modules`/`vendor`/`dist`/`build` from the main checkout into an env's worktrees — silently. When the main checkout's build is older than the worktree's source (e.g. `plugins/newspack-plugin/dist/` last built weeks ago while the worktree branch has newer commits), the isolated env serves a stale compiled admin SPA over current PHP, which is confusing and easy to misdiagnose as a branch-sync problem.

This PR makes seeded-asset age visible and fixable, and makes `--build` re-runs safe:

* Every copy now reports the assets' build date and age, e.g. `Copying built assets for newspack-plugin (built 2026-06-29, 22 days ago)...`. The date is the newest mtime under the source `dist`/`build` (the copy preserves mtimes). BSD/GNU `stat`/`date` differences are handled so it works on macOS and Linux hosts.
* When the seeded `dist`/`build` predate the worktree's last source commit, a prominent `[WARNING] … STALE` block prints both dates plus the exact host-side rebuild recipe (`COREPACK_INTEGRITY_KEYS=0 corepack pnpm install` then a **path-filtered** `pnpm --filter ./plugins/<repo> run build` — package names don't match repo dir names). Staleness compares against the commit date rather than source file mtimes, because a fresh worktree checkout stamps every file with checkout time and would false-flag assets built moments earlier. Standalone `repos/` worktrees get a manual-rebuild note instead, since they aren't pnpm workspace members.
* New opt-in `n env up <name> --rebuild` (implies `--build`): after seeding, rebuilds only the stale monorepo worktrees from their own source on the host (env worktrees can't be built in-container) — one `pnpm install` per worktree root, then a path-filtered build per stale repo. The seeded `dist`/`build` are removed before rebuilding because they may be hardlinked into the main checkout, and a rebuild writing through a shared inode would clobber the main checkout's files. Without the flag, behavior stays warn-only.
* Re-running `n env up <name> --build` on an already-seeded env is now safe and refreshes sensibly. Previously each re-run `cp`'d **into** the existing directories, nesting junk copies (`dist/dist/`, `node_modules/node_modules/`) while never refreshing anything. Now an existing `node_modules`/`vendor` is kept (it may hold the worktree's own host-side installs), and an existing `dist`/`build` is refreshed only when the main checkout's copy is newer — a worktree's own fresher build (from `--rebuild` or a manual host build) survives re-runs instead of being rolled back to main's older one. Each directory reports its outcome (`kept`/`refreshing`). The STALE warning and the `--rebuild` queue now judge the worktree's **resulting** assets rather than the main checkout's copy, so a worktree carrying its own fresh build no longer warns falsely or gets pointlessly rebuilt (previously `--rebuild` would even delete that fresh build first). The cross-filesystem `cp -a` fallback also clears the partial destination a failed hardlink copy leaves behind — that path nested even on first seed.
* Docs updated to match: `n env` help text, `AGENTS.md` env commands, and the `README.md` isolated-environments section (which previously claimed `--build` "compiles assets").

1. Create an isolated env with a worktree: `n env create staletest --worktree newspack-plugin:<branch>` (don't start it yet).
2. Ensure the seeded assets will be stale: pick a `<branch>` with commits newer than the main checkout's last `plugins/newspack-plugin/dist/` build, or force it by backdating: `find plugins/newspack-plugin/dist -type f -exec touch -t 202606291200 {} +`.
3. Run `n env up staletest --build`: each worktree repo prints `Copying built assets for … (built YYYY-MM-DD, N days ago)...`, followed by a yellow `[WARNING] … STALE` block showing the build date/age, the worktree's last source-change date, the exact `corepack pnpm` recipe, and a pointer to `--rebuild`.
4. Run `n env up staletest --rebuild`: same warning with `Queued for rebuild`, then `Installing JS dependencies in worktrees/<branch> (host-side pnpm)...` and `Rebuilding newspack-plugin from the worktree's source...`. Afterwards the worktree's `dist/` is a fresh build (current mtimes) and the env serves it.
5. Re-run behavior: run `n env up staletest --build` again after step 4 — the worktree's fresh `dist/` is kept (`dist/: kept (the worktree's copy is not older than the main checkout's).`), no STALE warning fires, no nested `dist/dist/` junk appears, and `node_modules`/`vendor` report `kept (seeded only when missing; delete it to re-seed).`. Then rebuild the main checkout (`n build newspack-plugin`) and re-run `--build`: `dist/: refreshing (the main checkout's copy is newer).` and the worktree picks up the newer build.
6. Fresh case: rebuild the main checkout's plugin (`n build newspack-plugin`), remove the worktree's previously seeded `dist/`, then `n env up staletest --build`: the copy line still prints the build date (`0 days ago`) with no warning. With `--rebuild` it prints `--rebuild: no stale monorepo worktree assets; nothing to rebuild.` and never invokes pnpm.
7. Standalone `repos/` worktree (optional): an env mounting a `repos/` checkout worktree with an old `dist` warns with a manual-rebuild note instead of the pnpm recipe.
8. Sanity: `n env` and `n env up` (no args) document `--rebuild`; `n env up staletest --bogus` still fails with `Unknown option`.

### Other information:

There's no shell-test infrastructure in the repo; the seeding function was verified with a 33-assertion fixture harness covering stale/fresh copies, mtime preservation, the `--rebuild` path (stubbed `corepack`: install-once-per-root dedupe, path-filtered build, env vars, pre-rebuild removal of hardlinked assets), standalone `repos/` handling, and flag parsing — plus `bash -n` and a probe against the real workspace reproducing the motivating incident (dist built 2026-06-29 vs source committed 2026-07-21).

The re-run fix was verified the same way with a second, 47-assertion fixture harness: against the pre-fix code it fails 27 assertions (nested copies, never-refreshed assets, false STALE warning, `--rebuild` deleting a fresh worktree-local build, fallback nesting); against the fixed code all 47 pass, with fresh-seed behavior, hardlinking, and stale-worktree `--rebuild` queueing guarded as regressions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -277,8 +277,15 @@ n env create <name> [options]  # Create environment config
                                #   so other envs keep the base checkout.
   --domain <domain>            #   Custom domain (default: <name>.test)
   --up                         #   Start the environment immediately after creation
-n env up <name> [--build]      # Start environment (creates DB, installs WP, sets up SSL)
-n env up --all [--build]       # Start all existing environments at once
+n env up <name> [options]      # Start environment (creates DB, installs WP, sets up SSL)
+  --build                      #   Seed built assets (node_modules/vendor/dist/build) from the
+                               #   main checkout into the env's worktrees. Prints each repo's
+                               #   asset build date and warns prominently when the assets
+                               #   predate the worktree's source (stale compiled JS/CSS).
+  --rebuild                    #   Implies --build; after seeding, rebuilds stale monorepo
+                               #   worktree JS from the worktree's own source on the HOST via
+                               #   corepack pnpm (env worktrees cannot be built in-container).
+n env up --all [options]       # Start all existing environments at once
 n env down <name>              # Stop environment
 n env destroy <name>           # Remove environment, DB, worktrees, and files
 n env list                     # List environments with status, URLs, and worktrees
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -282,6 +282,9 @@ n env up <name> [options]      # Start environment (creates DB, installs WP, set
                                #   main checkout into the env's worktrees. Prints each repo's
                                #   asset build date and warns prominently when the assets
                                #   predate the worktree's source (stale compiled JS/CSS).
+                               #   Safe to re-run: existing node_modules/vendor are kept, and
+                               #   dist/build refresh only when the main checkout's copy is
+                               #   newer (a worktree's own fresher build survives).
   --rebuild                    #   Implies --build; after seeding, rebuilds stale monorepo
                                #   worktree JS from the worktree's own source on the HOST via
                                #   corepack pnpm (env worktrees cannot be built in-container).
--- a/README.md
+++ b/README.md
@@ -288,7 +288,8 @@ n worktree remove <plugin> <branch> # Remove a worktree
 ```BASH
 n env create <name> --worktree <repo>:<branch> [--worktree ...] [--port <port>]
 n env up <name> [--build]           # Start the environment (--build seeds built assets
-                                    #   from the main checkout and warns when they're stale)
+                                    #   from the main checkout and warns when they're stale;
+                                    #   re-runs only refresh dist/build when main's is newer)
 n env up <name> --rebuild           # --build + rebuild stale worktree JS on the host
 n env down <name>                   # Stop the environment
 n env destroy <name>                # Stop and remove the environment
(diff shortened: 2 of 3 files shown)


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`. The branch itself needed no changes — the bot's `Merge origin/alpha` commit is no longer in its history (it was already rebased/pushed over), so nothing was force-pushed here.

PR #687, opened 2026-07-22
feat(content-gate): keep paid access during subscription payment recovery (NPPD-2052)

adekbadek:
### All Submissions:

Access Control denies paid access the moment a subscription leaves `active`, so readers inside the failed-payment retry window (dunning) lose access at migration flips – the largest lost-access class in recent pre-migration parity audits.

This PR treats subscriptions in **payment recovery** as access-granting:

- A subscription counts as "in payment recovery" when it is `on-hold` **and** carries a `payment_retry` date from the Woo Subscriptions retry system. WCS sets that date when it applies a retry rule and deletes it as soon as the retry resolves without a successor being scheduled, so the date exists exactly while a retry is outstanding and the grace self-expires when retries exhaust. The date is deliberately not compared against the current time: an overdue retry (Action Scheduler running behind) is still a pending one.
- New per-gate **"Grace during payment recovery"** toggle in the Paid Access panel, **default ON** – including for existing gates that lack the setting key.
- New `newspack_is_subscription_in_payment_recovery` filter to broaden the boundary (e.g. any on-hold, or gateway-managed retries).
- Evaluation context is plumbed through `Access_Rules` (exception-safe via `try/finally`) and honored at every surface that evaluates gate rules: front-end restriction, member-content block visibility, the wp-admin user-profile gate-access panel, and the contact-metadata ESP sync. Group subscriptions are covered.

Notes for reviewers (disclosures from the internal review):

- **This exceeds strict WCM parity.** WCM core actually maps subscription `on-hold` to a *paused* membership (no dunning awareness). The field-observed "keeps access during dunning" behavior was de-facto (unlinked memberships / missed status-sync events), not WCM code. The ticket explicitly chose the improvement.
- **Stale on-hold subscriptions still lose access at flips** (no `payment_retry` date = no grace), by design. Migration dry-runs should split on-hold residuals by presence vs absence of a `payment_retry` date.
- **The self-closing property has one residual gap.** Every normal lifecycle path deletes the `payment_retry` date, but a retry action that fails permanently at the Action Scheduler level is only logged by WCS – the retry record stays `pending`, so the date survives and the grace persists for as long as the subscription stays on-hold. Reading the retry store's status instead would report the same thing, since it is that record that is left pending. Any status change on the subscription clears the date, so this only reaches a subscription nothing ever touches again; closing it properly needs a bound the retry system doesn't provide (e.g. capping the grace at the configured on-hold duration). Documented on `is_subscription_in_payment_recovery()`.
- **The grace is inert where the WCS retry system is disabled** (`woocommerce_subscriptions_enable_retry` defaults to `no`; Newspack force-enables it, but sites relying on gateway-side retries such as Stripe smart retries get no grace unless they hook the filter). Documented in the filter docblock.
- **ESP sync follows the per-gate toggle, with one asymmetry.** `Content_Access` and `Content_Access_Source` are evaluated under the gate's own context, so a gate with the toggle off neither grants access nor attributes a paid segment to an in-recovery subscription. `Content_Access_Group` names still come from the shared, active-status-only group helper (also used by the GA4 helper), so a *group* subscription in the retry window grants access without contributing a group name. Aligning that would mean threading the setting through a memoized cross-feature API; left out of scope here.
- **Deploying this changes behavior on already-flipped Access Control sites**: on-hold-in-retry readers regain access at deploy (default ON). Desired, but platform-wide, not migration-only.
- A follow-up for the newspack-manager PugPig entitlement path (context-free evaluation, always grace-ON) is tracked separately.
- Possible textual overlap with the upcoming NPPD-2053 branch in `custom-access.tsx` / gate types – whichever merges second rebases.

Closes NPPD-2052.

1. Enable `NEWSPACK_CONTENT_GATES` + RAS, create a gate with Paid access on a subscription product, and a reader with an active subscription to it.
2. Put the subscription `on-hold` with a scheduled `payment_retry` date (a failed renewal with the WCS retry system enabled does this) – the reader keeps access.
3. Turn off "Grace during payment recovery" on the gate – the reader is gated.
4. Clear the retry date (or let retries exhaust) with the toggle on – the reader is gated.
5. Reactivate the subscription – access restored.
6. `n test-php --filter Access_Rules` – includes the on-hold matrix, the overdue-retry case, default-ON-for-legacy-gates, context-leak, and call-site plumbing tests for the front-end, block-visibility and profile-panel paths.
7. `n test-php --group Content_Gate_Metadata` – includes the ESP source-label attribution following the gate's toggle.
8. `npm test` – covers the toggle's visibility and the reset of a stored `false` when the last subscription rule is removed.

### Other information:


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2052-parity-fix-readers-in-payment-recovery-on-hold-subscriptions` (stash local work first).

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -93,6 +93,38 @@ public static function is_subscription_active( $status ) {
...
+	 * window closes on its own.
+	 *
+	 * @param \WC_Subscription|false $subscription Subscription object.
+	 *
+	 * @return bool
+	 */
+	public static function is_subscription_in_payment_recovery( $subscription ) {
+		$in_payment_recovery = $subscription
+			&& $subscription->has_status( 'on-hold' )
+			&& (int) $subscription->get_time( 'payment_retry' ) > time();
```
dkoo: **Suggestion:** Recovery window closes at the scheduled retry time, not the executed retry — `payment_retry` is the timestamp at which Action Scheduler is *due* to run the retry. On busy sites the queue lags (minutes to hours), during which `get_time( 'payment_retry' ) > time()` is false while a retry is still genuinely pending — the reader is denied at exactly the boundary this PR protects, then regains access if the late retry fails and schedules the next attempt. It fails closed, so this is not blocking; if boundary flapping shows up in the field, consider a small tolerance buffer or consulting the retry store's pending-retry status via the provided filter.

**Consider:** Tighten the filter's inline documentation to warn that returning true for a subscription outside an active retry window grants indefinite access to stale on-hold subscriptions, so integrators gate their custom logic on a real pending-retry signal rather than on-hold status alone.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
```diff
@@ -358,4 +360,249 @@ public function test_evaluate_rules_defaults_to_current_user_when_user_id_is_nul
...
+		);
+
+		$this->assertTrue(
+			Access_Rules::evaluate_rules( $subscription_access_rules, self::$owner_user_id ),
+			'No context given: the grace should default to ON.'
+		);
+	}
+
+	/**
+	 * Test that the evaluation context does not leak out of evaluate_rules —
```
dkoo, line 460: **Suggestion:** Grace-off plumbing test covers one of three call sites — The stored-grace-off plumbing test exists precisely because every fallback in the chain is grace-ON — a call site that dropped its context argument would pass all engine tests while the publisher's off-switch silently stopped working. That exact risk applies equally to the other two call sites (block visibility and the user-profile panel), which have no equivalent assertion. Extending the stored-grace-off check to those two paths closes the gap the test's own comment describes.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -390,19 +422,23 @@ public static function add_payment_method_form_notice() {
...
 		}
 		$subcriptions = array_reduce(
 			array_keys( \wcs_get_users_subscriptions( $user_id ) ),
-			function( $acc, $subscription_id ) use ( $product_ids, $user_id ) {
+			function( $acc, $subscription_id ) use ( $product_ids, $user_id, $include_payment_recovery ) {
 				$subscription = \wcs_get_subscription( $subscription_id );
 				$has_required_products = false;
-				if ( $subscription->has_status( self::ACTIVE_SUBSCRIPTION_STATUSES ) ) {
+				if (
+					$subscription->has_status( self::ACTIVE_SUBSCRIPTION_STATUSES )
```
dkoo, line 472: **Nit:** Adjacent pre-existing exposure: unguarded wcs_get_subscription() result — `wcs_get_subscription()` can return `false`, and the `has_status()` call on this line would fatal before reaching the new `is_subscription_in_payment_recovery()` (which does guard `false`). The exposure predates this PR, but since the diff already reworked this condition into a two-arm boolean, an early `if ( ! $subscription ) { continue; }`-style guard would be cheap hardening.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/custom-access.tsx
```diff
@@ -40,6 +41,8 @@ export default function CustomAccess( { customAccess, onChange, isNewsletter = f
 		[ handleChange ]
 	);
 
+	const hasSubscriptionRule = currentRules.some( rule => rule.slug === 'subscription' );
```
dkoo: **Nit:** Toggle visibility edges: persisted OFF resurfaces; group[0]-only derivation — The toggle renders only when a subscription rule is detected — and detection reads `group[ 0 ]` of each group. Two edges: a publisher who sets grace OFF and later removes the subscription rule keeps `payment_recovery_grace: false` persisted invisibly (it resurfaces on rule re-add), and a REST-authored gate with a subscription rule at index ≥ 1 in a multi-rule group hides a load-bearing toggle. Both mirror pre-existing wizard constraints; noting for awareness.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/custom-access.tsx
```diff
@@ -51,6 +54,22 @@ export default function CustomAccess( { customAccess, onChange, isNewsletter = f
...
+			{ hasSubscriptionRule && (
+				<>
+					<CardDivider />
+					<CardBody size="small">
+						<ToggleControl
+							label={ __( 'Grace during payment recovery', 'newspack-plugin' ) }
+							help={ __(
+								'Keep access for readers whose subscription renewal payment failed while it is being retried automatically.',
+								'newspack-plugin'
+							) }
```
dkoo, line 85: **Nit:** ToggleControl onChange discards its checked value — `ToggleControl`'s `onChange` receives the new checked value; the handler discards it and re-derives the negation from the current state. Behaviorally equivalent (the JSX test pins the stored-false path), but `onChange={ value => handleChange( { payment_recovery_grace: value } ) }` is the idiomatic form.

dkoo (commented):
Release risk: LOW — zero blockers, consequences verdict low, impact compatible, and verification is complete: full test suites, a live CLI state-toggle matrix, and a 12-step browser walkthrough in an isolated env (wizard toggle default/persistence, reader grant/deny across active, in-recovery grace ON/OFF, and logged-out states, plus the user-profile access panel) all behave exactly as specified. The core is correct and carefully built: the retry-date predicate is timezone-safe and fails closed when WCS is absent, the context plumbing is leak- and reentrancy-safe, default-ON is honestly persisted, and all optional-parameter signature changes leave existing callers untouched. The main pre-merge conversation: two consumers evaluate access outside the per-gate context — the PugPig entitlement path (disclosed, follow-up tracked) and the reader-metadata ESP sync (undisclosed) — so a publisher who turns grace OFF still graces in app entitlements and ESP segment labels. Worth an explicit disclosure line and a decision on whether that asymmetry is acceptable for launch.

- `plugins/newspack-plugin/includes/reader-activation/sync/contact-metadata/class-content-gate.php:201` — **Suggestion:** Reader-metadata ESP sync graces unconditionally — outside the per-gate toggle — The contact-metadata sync calls `Access_Rules::has_active_subscription()` outside any evaluation context, so it inherits the grace-ON default unconditionally: ESP-synced access labels (`Content_Access`, product source labels) now count on-hold-in-recovery subscriptions as active, and the per-gate off-switch cannot suppress it. Publishers segmenting on these fields will keep in-recovery readers in paid segments through the dunning window. This is consistent with the platform-default stance and likely intended — but it extends the behavior change from gating into ESP segmentation data and isn't named in the PR body's otherwise-thorough disclosure list. Worth an explicit disclosure line, and a decision on whether the sync path should honor (or deliberately ignore) the per-gate setting. **Consider:** Thread the gate's grace setting into the two evaluations that currently omit it — the app entitlement check in newspack-manager and the reader-metadata sync re-checks — so a publisher who turns grace off sees it honored on web, in the app, and in segment labels alike; at minimum, add the sync-path behavior change to the PR body's disclosure list.
- `plugins/newspack-plugin/includes/content-gate/class-user-gate-access.php:23` — **Nit:** PR body labels the third call site as "My Account" — The description calls the third honored call site "the My Account access panel", but `User_Gate_Access` hooks `show_user_profile`/`edit_user_profile` — the wp-admin user-profile gate-access panel, not the WooCommerce My Account front end. The plumbing is real and correct; only the label could misdirect a QA tester looking for a front-end My Account change.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
adekbadek: Took the real fix on both points in edc8b77.

The boundary flap is gone: the check is now `get_time( 'payment_retry' ) > 0` rather than `> time()`. The date's *presence* is the pending-retry signal — `WCS_Retry_Manager::maybe_delete_payment_retry_date()` deletes it as soon as a retry leaves pending/processing without a successor being scheduled, so it exists exactly while a retry is outstanding and still closes the window on its own once retries exhaust. An overdue retry under a lagging Action Scheduler queue keeps access instead of denying it, and this matches how the rest of the plugin already reads the date (`On_Hold_Duration`, the subscription-cleanup CLI command). New test `test_owner_keeps_access_when_retry_is_overdue` pins the past-date case; `test_owner_denied_when_on_hold_without_scheduled_retry` still pins the no-date denial.

The filter docblock now spells out that whatever signal a filter substitutes must be self-closing, and that returning true for on-hold status alone grants paid access indefinitely to every parked subscription.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
adekbadek: Added the two missing plumbing tests in edc8b77: `test_stored_grace_off_hides_block_via_block_visibility` (block stays hidden with stored grace OFF, appears when the stored setting is flipped ON) and `test_stored_grace_off_denies_bypass_via_user_gate_access` (`can_bypass` false then true across the same flip). Both share a `create_grace_off_gate()` helper with the existing restriction-path test.

Verified they're load-bearing: hardcoding `'payment_recovery_grace' => true` at each call site makes exactly those two fail, and restoring the context argument makes them pass.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
adekbadek: Guarded in edc8b77 — `if ( ! $subscription ) { return $acc; }` right after the `wcs_get_subscription()` call, so a deleted subscription or a stale ID left on the user is skipped instead of fatalling on `has_status()`.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/custom-access.tsx
adekbadek: Switched to `onChange={ value => handleChange( { payment_recovery_grace: value } ) }` in edc8b77.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/custom-access.tsx
adekbadek: Both edges are fixed in edc8b77.

Detection moved off `group[ 0 ]`: a `hasSubscriptionRule()` helper scans every rule of every group, so a REST-authored gate with a subscription rule at index ≥ 1 now shows the toggle. `currentRules` still flattens to `group[ 0 ]` for the rules editor — that's the wizard's OR-mode display constraint and untouched here.

For the persisted-OFF edge, `handleRulesChange` resets `payment_recovery_grace` to `true` whenever the incoming rules contain no subscription rule, so removing the last one clears the setting instead of parking an invisible `false` that would take effect again on re-add. Two JSX tests cover it: the reset on removal, and that a stored `false` survives a rules change that keeps a subscription rule.

adekbadek:
You're right that `User_Gate_Access` hooks `show_user_profile`/`edit_user_profile`, not the WooCommerce My Account front end. Corrected the PR description's call-site list to "the wp-admin user-profile gate-access panel".

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
```diff
@@ -93,6 +93,55 @@ public static function is_subscription_active( $status ) {
...
+	 * subscription-cleanup CLI command).
+	 *
+	 * @param \WC_Subscription|false $subscription Subscription object.
+	 *
+	 * @return bool
+	 */
+	public static function is_subscription_in_payment_recovery( $subscription ) {
+		$in_payment_recovery = $subscription
+			&& $subscription->has_status( 'on-hold' )
+			&& (int) $subscription->get_time( 'payment_retry' ) > 0;
```
dkoo, line 131: **Suggestion:** Residual edge: a permanently-failed scheduled retry leaves grace open indefinitely — The presence check is safe for every normal lifecycle path (verified against the WCS source: the date is deleted on retry exhaustion, completion, cancellation, and any unexpected status transition). The one case it no longer self-closes: a scheduled retry action that permanently fails at the Action Scheduler level — WCS only logs it, the retry stays pending, the date persists, and grace persists while the subscription stays on-hold. Arguably correct (WCS itself still considers the retry outstanding, and any status touch clears it), and bounded by the on-hold guard. If the PR body's "no indefinite retention" should be literally airtight, the precise signal is the retry store's pending status — which the new filter docblock already recommends to implementers; adopting it here too is the belt-and-braces option. At minimum this edge deserves a sentence in the method docblock.

dkoo (commented):
Release risk: LOW — the follow-up resolves all five targeted findings, and the headline change survived adversarial scrutiny: the switch to a presence check looked like it might widen the grace to any subscription that ever had a failed payment, but two independent traces of the WooCommerce Subscriptions source confirm the payment_retry date is deleted on every exit from the dunning window, so date-presence is exactly the outstanding-retry signal and the self-expiry property the PR body promises still holds. The one residual worth a line in the docblock: a retry whose scheduled action permanently fails at the Action Scheduler level (a site-infrastructure pathology WCS itself only logs) leaves the date — and therefore the grace — in place indefinitely while the subscription stays on-hold; the precise signal for literal airtightness is the retry store's pending status, which the new filter documentation already recommends to implementers and could be adopted here too if wanted. The new plumbing tests are genuine and load-bearing, the subscription-lookup guard is correctly placed, and both toggle-visibility edges are fixed with a deliberate, commented reset-to-default. Nothing blocking.

plugins/newspack-plugin/includes/plugins/woocommerce/class-woocommerce-connection.php
adekbadek: Went with the docblock, because I checked the retry-store option and it turns out not to be belt-and-braces here — in exactly the scenario you describe it reports the same thing.

When a scheduled action fails permanently, WCS routes it to \`WCS_Failed_Scheduled_Action_Manager::log_action_scheduler_failure()\`, hooked to \`action_scheduler_failed_action\` / \`action_scheduler_failed_execution\`. That method writes a WC log line and stashes the action in the \`..._failed_scheduled_actions\` option for an admin notice. It never touches the retry record — so the record the store would return is the one left \`pending\`, and gating on `pending` returns true just as the date check does. The retry store is the source of truth the date mirrors; in this failure mode both are stuck at the same value.

So the honest position is that this cannot be closed from inside the retry system: closing it needs a bound the retry system does not provide — e.g. capping the grace at `On_Hold_Duration::get_on_hold_duration()`. That is a product call about how long a stalled recovery should keep granting access, so I have not invented one. Worth noting the same stall also defeats `On_Hold_Duration::maybe_schedule_expiration()`, which deliberately skips scheduling a Newspack expiry when a `payment_retry` date is present, trusting the retry chain to reach its terminal `expired` rule — so nothing else expires the subscription either. If you want the hard cap, say so and I will add it here.

Documented the edge on the method in 61ef5a5, including why the retry-store signal does not help, and softened the PR body claim it was overstating. Happy to be pushed back on if you would rather have the cap now.

adekbadek:
Took the threading option rather than just disclosing it, since the fix turned out to be contained.

You were right about the mechanism: `get_source_labels()` re-runs `Access_Rules::has_active_subscription()` to work out *which* product granted access, and those calls sat outside any evaluation context, so they inherited the grace-ON default no matter what the gate said. On a gate with the toggle off, a reader holding one active and one in-recovery subscription was correctly denied by the in-recovery one but still got it listed in `Content_Access_Source` — a paid-segment label for access the gate deliberately refuses.

In 61ef5a5:

- `Access_Rules::with_evaluation_context( $context, $callback )` — the save/set/restore-in-`finally` dance extracted into one place, so evaluating a rule callback outside `evaluate_rule()` / `evaluate_rules()` is possible without reaching into private state. Both entry points now use it, which drops the two copies of that dance.
- `User_Gate_Access::evaluate_gate_for_user()` returns the `context` it evaluated under, so any consumer explaining *why* a rule passed can re-run under the same settings instead of guessing.
- The contact-metadata sync threads that context through `collect_labels()` into the resolvers, and `get_source_labels()` wraps its `has_active_subscription()` calls in it.

Regression test in `tests/unit-tests/content-gate/class-content-gate-metadata.php`; it fails with `Active Plan, In Recovery Plan` against the old code and passes with `Active Plan`.

One asymmetry survives and is now in the PR body: `Content_Access_Group` names come from `Group_Subscription::get_group_names_for_user()`, which counts only active-status subscriptions and is memoized and shared with the GA4 helper. A *group* subscription in the retry window therefore grants access without contributing a group name. Threading the setting there means changing a cross-feature API and its cache key, and it errs toward under-reporting rather than over-reporting, so I left it — flag it if you want it in this PR.

adekbadek:
On the launch-acceptability question (the per-gate grace toggle vs. the two out-of-context consumers): the decision is that **both consumers should honor the toggle, and both are being brought into line** rather than left asymmetric.

- **ESP reader-metadata sync** — fixed in this PR (61ef5a5). `Content_Access` / `Content_Access_Source` now evaluate under the gate's own context, so a gate with grace off neither grants access nor attributes a paid segment to an in-recovery subscription.
- **PugPig app entitlement** — fixed in the sibling PR Automattic/newspack-manager#666 (NPPD-2103), which threads the gate's grace setting into the entitlement check. It is up for review in parallel.

The one residual is timeline, not design: if this PR merges ahead of manager#666, there is a brief window where the app still graces during dunning while web and ESP honor an off-toggle. That is acceptable for launch — the toggle defaults ON, so it is inert for every publisher that hasn't deliberately turned grace off; PugPig is a narrow surface; and the gap self-expires with the retry window. The `Content_Access_Group` name gap noted in the PR body (a grace-window group subscription grants access without contributing a group name, since the group-name helper reads active status only) stays out of scope here — it is a label-completeness gap, not an access asymmetry.

dkoo (commented):
Release risk: LOW — the follow-up resolves the review's most significant finding correctly and completely: ESP-synced access labels now honor each gate's grace setting with per-gate attribution unioned across gates, so an in-recovery subscription is only ever reported through gates that legitimately grant it access; the context plumbing rides a shared helper that preserves the engine's leak-safe restore discipline, and the regression test pins the original bug in both directions. The group-name deferral to a follow-up ticket is verified sound — that path only counts active subscriptions, so it cannot leak under a grace-OFF gate; the residual is a cosmetic under-report honestly documented. The round-1 docblock note about the non-self-closing retry case is also addressed. One trivial note: two variable assignments in the new test have misaligned operators that the WordPress PHPCS ruleset may warn on — worth a lint pass before merge since the pre-commit hook blocks on warnings. Nothing further needed from this review's perspective.

- `plugins/newspack-plugin/tests/unit-tests/reader-activation/class-content-gate-metadata.php:960` — **Nit:** PHPCS alignment warning risk in the new test — The two `$active_product_id` / `$in_recovery_product_id` assignments use misaligned `=` operators; `Generic.Formatting.MultipleStatementAlignment` may warn under the WordPress ruleset, and PHPCS warnings block the pre-commit hook. A quick lint pass on the file before merge settles it — not a correctness issue.

dkoo (approved):
Manual QA verified the behavior end to end through the live gateway path: with a regular single subscription and a test Stripe account, a real failed renewal put the subscription into payment recovery, and the reader kept paid access while the scheduled retry was pending — but only with the gate's new grace-during-recovery option enabled. Switching the option off gated the reader. That matches the design and the automated coverage from the prior review rounds.

Release risk: LOW. The change stays within the content-gate access evaluation, and I found no access changes outside the recovery-grace path. This branch also doesn't carry the vipwpcs CI change under discussion on #693 and #726, so it can merge independently of that decision.

Approving.

Review provenance: deep + WP expert + Codex across the full review and delta rounds.

adekbadek:
Aligned in 81c9659. Worth noting the ruleset won't catch these: `Generic.Formatting.MultipleStatementAlignment.NotSameWarning` is excluded workspace-wide in the root `phpcs.xml`, so both PHPCS and the pre-commit hook pass clean on the file as it stood — the misalignment was only visible by eye. Running that one sniff explicitly turned up four spots on this branch rather than two: the three `self::$*_user_id` assignments in `Access_Rules::set_up()`, and `$active_product_id`/`$in_recovery_product_id` in `test_source_labels_respect_gate_payment_recovery_grace()`. I also fixed the one pre-existing case in `create_institution()` in the same file, since it was a single line of whitespace. Nothing else on the lines this PR adds trips the sniff. Full suite still green — 2412 tests, 7 pre-existing skips.

PR #688, opened 2026-07-22
feat(cli): add migrate-institutions command for teams-based institutional access (NPPD-2054)

adekbadek:
### All Submissions:

Sites using the custom `newspack-teams-for-wc-memberships-access-by-ip` / `-auto-join-by-email` plugins store institutional access as WCM Teams plus option maps; until now the only migration path to Access Control Institutions was hand-recreating each institution in the admin. This adds `wp newspack migrate-institutions` to the in-plugin migration CLI suite (sibling of `migrate-teams`, registered the same way):

- Reads the `wc_memberships_team` inventory and the `wc_team_memberships_access_by_ip` option map (both string and array range shapes), creates one Institution per team via `Institution::create` with the team's name and validated IP ranges (IPv4 + CIDR; invalid tokens, including IPv6 which the feature does not support, are reported per team, not dropped silently).
- `--domains-csv` operator input for the email-domain → institution mapping (`team_id,domain[,domain...]`, accumulating rows, optional header, ASCII/punycode; the per-team domain mapping is not in the DB, so it must be operator-supplied).
- **Dry-run by default**; `--live` to write. Every team that cannot be mapped is reported with a reason. Failed creates are warned with the error and counted in the summary.
- Prints the migrated IP ranges as a per-institution verification checklist for the proxy/egress caveat (NPPD-2039): institutions behind EZProxy/Zscaler need public egress IPs, not internal ranges. `0.0.0.0/0`-style ranges are flagged as match-everyone grants.
- **Re-runs skip already-migrated teams** (marker meta `_np_institution_migrated_from_team_id`, matched across all post statuses including trash) and report the institution's *current* rules rather than the source option map – deliberately not re-aligning, so operator proxy-IP corrections survive. `Institution::invalidate_cache()` runs after live writes.
- Where a team has `_subscription_id`, the linked subscription is recorded as **informational meta** (`_np_institution_migrated_subscription_id`) and shown in the report with its live status. The Institution entity has no functional subscription-link field – group access itself rides the group subscription migrated by `wp newspack migrate-teams`.

Shared-file note: `tests/mocks/wp-cli-mocks.php` and the RAS tally test's mock require are byte-identical with the NPPD-2055 branch (#689) by design, so the two PRs merge cleanly in either order.

Closes NPPD-2054.

1. On an env with WCM + Teams active, seed a few `wc_memberships_team` posts, set `wc_team_memberships_access_by_ip` (mix of valid/invalid ranges, with/without `_subscription_id`), and write a small domains CSV.
2. Run `wp newspack migrate-institutions --domains-csv=<path>` – verify the dry-run report: would-create rows, unmappable teams with reasons, orphan option entries, the proxy-egress checklist. Confirm nothing was written.
3. Re-run with `--live` – verify institutions exist with correct `np_institution_ip_range` / `np_institution_email_domain` meta.
4. Edit one institution's ranges, re-run `--live` – verify the team is skipped, the operator edit survives, and the summary lists the *current* ranges.
5. `n test-php --filter Institutions` – 13 tests cover parsing, normalization, create/link/unmappable/dry-run/idempotency, and the error path at command level.

### Other information:


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2054-migration-tooling-migrate-institutions-cli-teams-based-ip` (stash local work first).

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+		$tokens = array_filter( array_map( 'trim', $tokens ) );
+
+		$valid   = [];
+		$invalid = [];
+		foreach ( $tokens as $token ) {
+			if ( false !== strpos( $token, '/' ) ) {
+				list( $subnet, $bits ) = array_pad( explode( '/', $token, 2 ), 2, '' );
+				$subnet                = trim( $subnet );
+				$bits                  = trim( $bits );
+				if ( ctype_digit( $bits ) && (int) $bits <= 32 && filter_var( $subnet, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) ) {
```
dkoo, line 458: **Severity:** blocker

**Blocker:** `ctype_digit( '00' )` is true and `(int) '00' <= 32`, so `0.0.0.0/00` validates and is stored verbatim — but the `/0` warning checks `str_ends_with( $migrated_range, '/0' )`, which is false for `/00`. At read time `IP_Access_Rule::ip_matches_ranges()` casts the bits to int 0 and the range matches every visitor IP — confirmed in the env: `ip_matches_ranges( '203.0.113.9', '0.0.0.0/00' )` returns true, and the migration output prints no warning for it. More broadly, only the literal `/0` draws a breadth flag: a migrated `10.0.0.0/8` or `127.0.0.1` — exactly the internal-range shape the proxy-egress checklist warns about in prose — migrates with no per-row flag, and `Institution::check_ip()` scans all institutions site-wide, so one over-broad range widens access everywhere. Since a malformed or too-broad range turns an institution into a whitelist for everyone, this needs fixing before the command runs on production data: normalize the bits when composing the valid entry (`$subnet . '/' . (int) $bits`) and judge breadth by the numeric mask (flag mask 0 however it was written; warn on private/reserved ranges via `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE`, and consider flagging prefixes shorter than /16). Please add tests covering these scenarios so the warning contract stays pinned.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+	 *
+	 * @return array|WP_Error {
+	 *     Parse result, or WP_Error when the file is unreadable.
+	 *
+	 *     @type array    $map    Map of team ID => list of normalized domains.
+	 *     @type string[] $errors Malformed rows/domains, for reporting.
+	 * }
+	 */
+	public static function parse_domains_csv( $csv_path ) {
+		if ( ! is_readable( $csv_path ) ) {
```
dkoo: **Severity:** blocker

**Blocker:** `is_readable()` and `file_get_contents()` both succeed for a directory path, an empty file, and a whitespace-only file — confirmed in the env container: all three return `{map: [], errors: []}` — so a path typo passes the readability guard and the run proceeds to migrate every team with no email-domain rules. The only signal is the `0 team(s) in the domains CSV` count line, and since a migrated team is never revisited on re-run, missing that line on a `--live` run means fixing every affected institution by hand. This undercuts the invariant `test_parse_domains_csv_missing_file_is_an_error` asserts: it holds for a nonexistent path but not for these three shapes. **Consider:** when `--domains-csv` is supplied but parses to zero teams, stop rather than continue. An unreadable, empty, or wrongly-pathed file currently looks identical to a site that genuinely has no domain rules, and the difference only becomes visible after institutions have been created and marked migrated.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+			}
+		}
+
+		// Idempotency: never re-create, and never touch the existing institution's
+		// rules — post-migration operator edits (e.g. proxy egress IPs) must survive.
+		// Report the institution's CURRENT rules, not the source data: after an
+		// operator swaps in proxy egress IPs, re-listing the stale source ranges
+		// would invite "restoring" the broken internal ranges.
+		$existing_institution_id = self::find_migrated_institution( $team_id );
+		if ( $existing_institution_id ) {
```
dkoo, line 615: **Severity:** blocker

**Blocker:** The already-migrated branch returns before the supplied domains and ranges are consulted, and the skip message doesn't say whether the CSV carried values the institution lacks. The likely operational sequence — a first `--live` run before the domain list is in hand, then a re-run with `--domains-csv` — reports `already migrated — skipping` and quietly adds nothing. Preserving operator edits is the right default; the gap is that the output doesn't name what was withheld (e.g. "CSV supplies 2 domain(s) this institution does not have — apply manually"), computable by diffing the supplied values against the current meta before returning. **Consider:** when a team is already migrated and the supplied CSV carries domains the existing institution does not have, name those domains in the output — the skip is the right behaviour, but the operator needs to know which values the re-run declined to add.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+		$map    = [];
+		$errors = [];
+		$lines  = preg_split( '/\r\n|\r|\n/', $csv_content );
+		foreach ( $lines as $line_index => $line ) {
+			if ( '' === trim( $line ) ) {
+				continue;
+			}
+			$cells   = array_map( 'trim', str_getcsv( $line ) );
+			$team_id = $cells[0];
+			if ( ! ctype_digit( $team_id ) ) {
```
dkoo, line 372: **Severity:** suggestion

**Suggestion:** `if ( 0 !== $line_index )` unconditionally swallows row 1 when its first cell isn't numeric, so a typo'd team ID on the first line of a headerless CSV (`12a,uni.edu`) produces neither a map entry nor an error — the one hole in the command's otherwise consistent never-silently-dropped reporting. The team still migrates on IP rules alone (or lands in the unmappable table with "no domains in the domains CSV"), so nothing downstream names the dropped row. Gating the header skip on the row actually looking like a header (e.g. first cell matching `/^team[ _-]?id$/i`), or emitting an info line whenever row 1 is skipped, would close it.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+		if ( ! $dry_run ) {
+			Institution::invalidate_cache();
+		}
+
+		$created_count = count( array_filter( $summary, fn( $row ) => 'created' === $row['Status'] ) );
+		$exists_count  = count( array_filter( $summary, fn( $row ) => 'exists' === $row['Status'] ) );
+		$error_count   = count( array_filter( $summary, fn( $row ) => 'error' === $row['Status'] ) );
+		WP_CLI::line( '' );
+		$done_report = sprintf( 'Done. %d team(s) processed: %d institution(s) %s, %d already migrated, %d unmappable, %d error(s).', count( $teams ), $created_count, $dry_run ? 'would be created' : 'created', $exists_count, count( $unmappable ), $error_count );
+		if ( $error_count > 0 ) {
```
dkoo, line 320: **Severity:** suggestion

**Suggestion:** When `$error_count > 0` the command emits a warning but still exits 0, so a scripted caller cannot detect partial failure. Since this is the final statement of the run, switching the failure path to `WP_CLI::error( $done_report . ' Errored teams were NOT migrated — see the warnings above.' )` prints to STDERR and exits non-zero with nothing skipped by the early exit; keeping the warning and adding `WP_CLI::halt( 1 )` is equivalent. Either makes partial failure detectable.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
```diff
@@ -0,0 +1,539 @@
...
+	 *     wp newspack migrate-institutions
+	 *     wp newspack migrate-institutions --domains-csv=/tmp/domains.csv
+	 *     wp newspack migrate-institutions --domains-csv=/tmp/domains.csv --live
+	 *
+	 * @param array $args       Positional args (unused).
+	 * @param array $assoc_args Named args.
+	 *
+	 * @return void
+	 */
+	public function migrate_institutions( $args, $assoc_args ) {
```
dkoo, line 125: **Severity:** nit

**Nit:** The sibling `migrate-teams` command hard-errors when WooCommerce Subscriptions is missing; this command runs happily on a site where `NEWSPACK_CONTENT_GATES` is off, creating institutions that sit dormant with no hint. Creating them ahead of enabling the flag is arguably valid migration prep, so a one-line `WP_CLI::warning` when `Content_Gate::is_enabled()` is false seems like the right weight.

dkoo (changes requested):
Release risk: HIGH — three findings are must-fix for an access-granting migration. (1) A malformed or too-broad IP range turns an institution into a whitelist for everyone: `0.0.0.0/00` validates, evades the match-everyone warning, and matches every visitor IP at read time (confirmed live), and private or very wide ranges migrate with no per-row flag at all. (2) A `--domains-csv` path that is a directory, empty, or whitespace-only silently parses to zero domains while teams are marked migrated, which a corrected re-run then can't repair (confirmed live). (3) The already-migrated branch silently withholds newly-supplied CSV domains, so the natural repair path — re-running with the right CSV — looks successful while adding nothing. All three have small fixes: numeric mask comparison plus breadth warnings with tests pinning them; a hard stop when a supplied CSV parses to zero teams; naming the withheld values on skip. The core semantics — dry-run safety, idempotency, operator-edit survival, trash handling, per-team reporting — all verified in an isolated env, and the full plugin test suite is green, so this is close: land the three fixes and it's ready.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. CIDR bits are now `(int)`-canonicalized on validation (`/00` → `/0`), and the warning judges breadth numerically via a new `get_range_breadth_warning()` helper: zero masks however written, private/reserved subnets, and public blocks wider than /16 all get flagged (error rows skipped, since nothing was written for them). Covered by new unit and command-level tests, and verified live: a source range of `0.0.0.0/00` now migrates as `0.0.0.0/0` with the matches-everyone warning.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. `parse_domains_csv()` now requires `is_file()`, so a directory hard-errors exactly like a missing file, and the command aborts via `WP_CLI::error` when a supplied CSV parses to zero rows — before any team is processed. Tests cover the directory and header-only cases; verified live with exit code 1.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. The already-migrated branch now diffs the supplied ranges and domains against the institution's current meta, and the run output names each withheld value per team ("NOT applied … add manually if intended"). Nothing is ever applied to an existing institution. Unit and command-level tests cover both the computation and the output.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. Only a cell that actually reads as a header (`team_id` / `team id` / `team-id`, case-insensitive, on any row) is skipped silently; any other non-numeric team ID — row 1 included — is reported. A side benefit: a leading blank line no longer causes the real header to be mis-reported as a malformed row. Tests cover both shapes.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. The error path now exits via `WP_CLI::error( $done_report … )`, so scripted callers get a non-zero exit with nothing skipped (it was already the run's final statement). Verified live — exit code 1 — and the command-level test asserts the abort.

plugins/newspack-plugin/includes/cli/class-institutions-migration.php
dkoo: Fixed in 9844d998b. The command now warns when `NEWSPACK_CONTENT_GATES` is off, noting that institutions will be created but stay dormant until the feature is enabled.

dkoo (approved):
This is clean after two rounds of review + changes.

PR #678, opened 2026-07-21
feat(reader-activation): generic integrations backfill CLI (NPPD-2076)

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guidelines](https://github.com/Automattic/newspack-workspace/blob/main/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds generic backfill tooling to the RAS Integrations framework, closing the three gaps from the issue: the backfill tooling was push-only, could not be scoped to a single integration, and carried a legacy ESP-only name.

- **New `wp newspack integrations backfill` command** — the generic successor of `wp newspack esp sync`, supporting `--direction=push|pull|both` (default `push`, matching the legacy behavior) plus all the existing batching/selector flags.
- **Incoming (pull) backfill** — operators can now force a bulk pull of enabled incoming fields for all readers (or `--user-ids`/`--active-subs-only` subsets), e.g. after connecting a new integration or enabling a new incoming field. Pull failures are tallied and logged but deliberately not auto-retried via ActionScheduler, so a bulk run against a flaky API can't flood the queue; the affected `--offset` window can be re-run instead. A pull `--dry-run` fetches and previews but skips writing reader data.
- **Per-integration scoping** — `--integration=<id>` restricts either direction to one active, configured integration (and scopes the `--fields` pre-flight validation to it). Powered by a new optional `integration_id` sync option consumed by `Contact_Sync::push_to_integrations()`; behavior is unchanged for all existing callers.
- **Strict flag matrix** — push-only flags (`--subscription-ids`, `--order-ids`, `--migrated-subscriptions`, `--skip-lists`, `--fields`) hard-error when the direction includes pull, rather than silently applying to only one leg.
- **Clearer subscription filter name** — the new command spells the subscription filter `--active-subs-only` (it gates on active WooCommerce subscriptions, so the name now says what must be active); the legacy alias keeps the historical `--active-only` spelling.
- **`wp newspack esp sync` is kept as a backward-compatible alias** frozen to the push direction and its historical flag surface, now printing a one-line notice pointing at the new command. Its output format is unchanged.

[...]

PR #689, opened 2026-07-22
feat(cli): safely target purchase plans in migrate-manual-members (NPPD-2055)

adekbadek:
### All Submissions:

Every pre-migration access-parity audit surfaces a comp/legacy residual class: readers with an active WCM membership on a *purchase* plan but no live subscription (admin-granted memberships, or lapsed/cancelled subs whose membership stayed active). `migrate-manual-members` previously had no subscription-state filter, so pointing it at a purchase plan would have minted $0 subscriptions for every member, including real paid subscribers. This extends it to target purchase plans safely:

- **`--only-without-live-subscription`** – skips members holding a subscription in a live status (`active`, `on-hold`, `pending-cancel`, whether owned or via a live group subscription they belong to); members whose subs exist only in dead states (`cancelled`, `expired` – the lapsed cohort, often larger than the no-sub one) are **included**, per the ticket's scope note. `pending` is deliberately classified dead (documented). Gifted subscriptions follow the gates' rule: they cover the recipient, never the purchaser. The skip report breaks down by matched status, with an explicit note that on-hold is counted live by design (the dunning cohort is NPPD-2052's; `On_Hold_Duration` auto-expires retry-less on-hold subs, after which a re-run includes the member).
- **`--user-ids=<csv>` / `--user-ids-file=<path>`** – explicit reviewed-list mode (accepts users with dead subs); strict parsing – any garbage token aborts the run rather than silently shrinking or misreading the list, and a UTF-8 BOM in a list file is tolerated. Bare valueless flags are caught on the **raw command line**: WP-CLI strips a bare `--user-ids` with only a warning before the command runs, so the command scans argv itself and aborts instead of proceeding with a widened scope. The run header prints the resolved list size.
- **Safety guards**: non-manual-only plans are refused unless a selection flag is present – the mass-grant footgun is an explicit error. `--plan-ids` gets the same strict token parsing as the other ID flags (it pins the widened all-plans default, so a typo must halt the run). `--as-group` with the new selection flags requires explicit `--plan-ids`, creates each plan's group subscription lazily on the first qualifying member (a fully-filtered plan leaves no orphan empty group sub), and grants each member once across plans so dry-run and live counts agree.
- **Pre-flight**: errors if WCM isn't active (previously a silent zero-row no-op), and refuses a `--product-id` outside the access-product set **in every mode, including the legacy manual-only default** – `--access-product-ids` is the bypass when the gates don't yet list the migration product. The access-product list is derived only from published gates whose custom access is switched on. With `--only-without-live-subscription`, an empty resolved list warns in dry-run but refuses a `--live` run (an empty list would make any live subscription count as covering, the reading that silently costs readers access).
- `--skip-domains`, dry-run default, and the `created_via='manual migration'` idempotency guard all still apply on the new paths; the idempotency guard is evaluated before the live-sub filter so re-run reports distinguish "already migrated" from "genuinely subscribed".
- Reconciliation output is **per-user** (dry-run would-create and live-skip counts reconcile against a parity diff even for members on multiple plans), plus unmatched-user-ID warnings. Liveness is memoized per user and the group data layer's static cache is reset per iteration, so a sweep over an unbounded member list keeps peak memory at one member's worth.
- Tests: 33 unit tests driving the whole command via a shared recording WP-CLI mock (status matrix, both modes and their composition, refusals, idempotency, parser edges incl. raw-argv bare flags, gifting both directions, inactive gates, group-mode dedupe/orphan cases). The shared mock (`tests/mocks/wp-cli-mocks.php`) is now a superset of the `wp-cli-mock.php` stub #678 introduced, and that file is a shim requiring it – only one `WP_CLI` class can exist per PHPUnit process, so both require paths resolve to one definition. Heads-up for #688: this ends the byte-parity its branch assumed for the mock file; it will want the superset version when it merges main.

Teams note: on Teams sites, run `wp newspack migrate-teams` first – before it runs, team members have no group subscription and would classify as residuals (documented in the command docblock).

Closes NPPD-2055.

1. On an env with WCM + WCS active, seed a purchase plan with members in each class: real active sub, on-hold sub, cancelled sub, expired sub, no sub.
2. `wp newspack migrate-manual-members --product-id=<id> --plan-ids=<purchase-plan-id>` → refused with an error naming the safe flags. (A bare `--product-id=<id>` with no manual-only plans on the site is a safe no-op: "No membership plans found to process.")
3. Add `--only-without-live-subscription --access-product-ids=<id>` → dry-run would-create only the cancelled/expired/no-sub members; live-status members reported skipped, per-user counts with a live-status breakdown.
4. `--live` → $0 subs (`created_via='manual migration'`) for exactly those members; re-run → they now report as "already migrated", not as live-skipped. Without `--access-product-ids` (and no gates configured), the `--live` run refuses; dry-run warns instead.
5. `--user-ids=<csv>` with a mix incl. an unknown ID → explicit mode with unmatched-ID warning; a bare `--user-ids` (no value) → hard error naming the flag, even though WP-CLI strips it; `--plan-ids=12x` → hard error.
6. `n test-php --group teams-migration`.

### Other information:


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2055-migration-tooling-complegacy-parity-residuals-have-no` (stash local work first).

adekbadek:
Pushed `63bbd58` — the liveness filter was product-agnostic while the access rule isn't.

`Access_Rules::has_active_subscription()` only accepts a subscription to one of the gate's configured products, but `member_has_live_subscription()` counted *any* live subscription. A member whose only live subscription is to a product no gate accepts — a recurring donation is the common case — was reported as `skipped — holds a live subscription` and granted nothing. Since the parity diff feeding this command is product-scoped, the two disagreed by construction: the diff counts that reader as losing access, the sweep skips them, and the reconciliation line reads as if they were covered.

- `member_has_live_subscription()` takes the accepted products and applies `has_product()` to owned and group subscriptions alike; empty = no constraint, matching a gate whose subscription rule lists no products.
- `--access-product-ids` pins the list; defaults to the `subscription` rules of the published gates. A run that can determine no products warns rather than silently counting any subscription as covering.
- A `--product-id` outside the accepted products is now refused. Found this by running it: the sweep created the $0 subscription, printed `Success: 1 subscription(s) created`, and the reader still failed the gate — the repair was inert and the run looked clean.

Verified against a real WooCommerce Memberships + Subscriptions install, not just the mocks: a member with a comp membership and an active donation subscription now gets a $0 subscription and passes the gate, where before the sweep skipped them. 24 tests, PHPCS clean.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1689,6 +1884,204 @@ private static function get_manual_only_plan_ids() {
...
+	 *
+	 * @param string $user_ids_csv  Comma-delimited user IDs, '' for none.
+	 * @param string $user_ids_file Path to a file of user IDs, '' for none.
+	 *
+	 * @return int[]|\WP_Error Unique user IDs in input order, or an error.
+	 */
+	public static function parse_user_ids( $user_ids_csv, $user_ids_file ) {
+		// A bare `--user-ids` (no `=value`) arrives as boolean true, which a string
+		// cast would turn into '1' — silently targeting user ID 1.
+		if ( ! is_string( $user_ids_csv ) || ! is_string( $user_ids_file ) ) {
```
dkoo, line 2225: **Blocker:** Bare `--user-ids` does not abort: WP-CLI strips the flag and the run continues — Verified against a live env: `wp newspack migrate-manual-members --product-id=<id> --user-ids` prints `Warning: --user-ids parameter needs a value` from WP-CLI itself, which then strips the flag and hands the command an empty default — so the `is_string()` guard in `parse_user_ids()` (and the resolved-to-nothing check at `:757`) can never fire from a real command line, and the run continues with no list narrowing. Composed with `--only-without-live-subscription --live`, a lost flag value (empty shell variable, truncated paste) silently converts a reviewed-list run into a sweep of every published plan, since the surviving selection flag widens the default scope at `:829`. The description's "a bare valueless flag aborts the run" holds in the unit tests (which call the parser directly) but not end-to-end. Reading the raw command line (`$_SERVER['argv']`) for valueless `--user-ids`/`--user-ids-file`/`--plan-ids`/`--access-product-ids` tokens would make the promised abort real.

**Risk if not addressed:** If an operator on a repeat site skips the dry run and the `--user-ids` value is lost (empty shell variable, truncated paste), the run drops the reviewed-list restriction entirely and sweeps every published membership plan, minting $0 subscriptions for every active member without a live subscription instead of the reviewed hundred-odd — a mass write across the whole membership base that has to be unwound by hand afterwards.

**Consider:** Inspect the raw command line for a selection flag that arrived without a value and abort, rather than relying on a check that runs after the flag has already been discarded. Also print the size of the resolved reviewed list in the run header, so a list that has silently vanished is visible in the first lines of output instead of only in the plan count.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -665,31 +710,85 @@ public function migrate_team_products( $args, $assoc_args ) {
...
+		if ( empty( $access_product_ids ) ) {
+			$access_product_ids     = self::get_gate_access_product_ids();
+			$access_products_source = 'derived from published gates';
+		}
+
+		// A $0 subscription for a product no gate accepts restores no access: the
+		// run would report "created" while the reader stays locked out. Only
+		// checkable when the accepted products are known — with none configured
+		// yet, the gates are presumably still to be built around this product.
+		if ( ! empty( $access_product_ids ) && ! in_array( $product_id, $access_product_ids, true ) ) {
```
dkoo, line 817: **Suggestion:** New access-product refusal also fires on the legacy manual-only path — The product-not-among-access-products refusal runs unconditionally — including on the legacy manual-only path with no new flags, in dry-run. On a site with published, active gates whose subscription rules don't yet list the migration product, previously-working invocations (including the ones documented in the access-control migration runbook) now abort. The check itself is sound and the message is actionable, but the PR description says the legacy path is unchanged, and the escape hatch (`--access-product-ids` including the product) is documented only as a liveness-filter scoper. Worth documenting in the command docblock and PR description — or downgrading to a warning on runs without a selection flag.

**Consider:** Update the ops runbook alongside this change: invocations documented there for manual-only plans can now abort on the new product check, and the flag that lets them through is currently described only as a way to scope which subscriptions count as covering a member.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -718,6 +824,10 @@ public function migrate_manual_members( $args, $assoc_args ) {
 
 		if ( ! empty( $plan_ids ) ) {
```
dkoo: **Suggestion:** `--plan-ids` keeps lax parsing while now being a load-bearing safety input — `--plan-ids` kept the legacy `array_filter( array_map( 'absint', explode( ... ) ) )` parse while this PR made `--user-ids` and `--access-product-ids` strict. A typo'd token (`--plan-ids=12x,34`) silently absints to `12` and narrows the scope the operator thinks they pinned; `--plan-ids=abc` reduces to `[]` and exits cleanly behind a misleading "No membership plans found". A bare `--plan-ids` would coerce boolean `true` through `explode()` to plan ID `1` after passing the `empty()` check in the `--as-group` guard at `:808` — though WP-CLI strips bare valueless flags before the command sees them, so that leg needs the same command-line-level treatment as the `--user-ids` blocker. Routing the flag through the existing `parse_id_tokens()` makes all three ID flags consistently strict.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -805,6 +958,27 @@ public function migrate_manual_members( $args, $assoc_args ) {
 					}
 				}
 
+				// Individual mode: skip a member who already owns an active subscription
+				// this migration created for the same product, so a re-run is safe and
+				// doesn't stack duplicate $0 subscriptions. Checked before the
+				// live-subscription filter so a re-run reports these as already-migrated
+				// instead of inflating the live-subscription count.
+				if ( ! $as_group && self::member_has_migration_subscription( $user_id, $product_id ) ) {
```
dkoo, line 1045: **Suggestion:** Per-member subscription traversals and an unbounded static cache at sweep scale — With `--only-without-live-subscription` and no `--plan-ids`, the scope is every active membership on every published plan — the residual count is the output, not the input. Each membership row does two full `wcs_get_users_subscriptions()` traversals with a cold object cache (`wp_clear_object_cache()` runs immediately before, at `:921`), and `has_product()` loads line items per candidate; multi-plan members repeat the work per plan. Separately, `Group_Subscription::get_group_subscriptions_for_user()` memoizes full `WC_Subscription` object arrays in a class static (`class-group-subscription.php:794-823`) that the per-iteration cache clear cannot reach, so on a post-`migrate-teams` site the static grows for the whole run. A per-run `[ user_id => bool ]` liveness memo plus a periodic `Group_Subscription::reset_cache()` (or `$ids_only = true`) would restore the memory bound the per-iteration clear exists to maintain.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -805,6 +958,27 @@ public function migrate_manual_members( $args, $assoc_args ) {
...
+				// Tracked per user, not per membership, so the reported count reconciles
+				// against a per-reader parity diff.
+				if ( $only_without_live_subscription && self::member_has_live_subscription( $user_id, $access_product_ids ) ) {
+					WP_CLI::line( sprintf( '  Membership %d (user %d, %s): skipped — holds a live subscription.', $membership_id, $user_id, $user->user_email ) );
+					$skipped_live_subscription_user_ids[ $user_id ] = true;
+					continue;
+				}
+
 				// Group mode: add the user as a group member.
 				if ( $as_group ) {
```
dkoo, line 1068: **Suggestion:** `--as-group` with multiple plans breaks per-user reconciliation — The group-mode branch `continue`s before the `$granted_user_ids` dedupe at `:1007` and never populates it, so per-user reconciliation does not hold for `--as-group --plan-ids=A,B`: dry-run counts a member on both plans twice in the summary and closing tally. A live run behaves differently again — plan A's created group subscription is active, group-enabled, and carries the migration product, so on plan B the member now tests as live and lands in the live-skip tally, inflated by members this very run just added. Dry-run and live therefore report different numbers for the same input. Existing coverage only exercises `--as-group` against a single plan.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1689,6 +1884,204 @@ private static function get_manual_only_plan_ids() {
...
+	public static function member_has_live_subscription( $user_id, $product_ids = [] ) {
+		$user_id = absint( $user_id );
+		if ( ! $user_id || ! function_exists( 'wcs_get_users_subscriptions' ) ) {
+			return false;
+		}
+		$product_ids = array_values( array_filter( array_map( 'absint', (array) $product_ids ) ) );
+		foreach ( \wcs_get_users_subscriptions( $user_id ) as $subscription ) {
+			// wcs_get_users_subscriptions can be filtered to include member-only groups; require ownership
+			// here — group memberships are evaluated explicitly below, in every context.
+			if ( (int) $subscription->get_user_id() !== $user_id ) {
```
dkoo: **Suggestion:** Gifted subscriptions diverge from gate semantics in both directions — `member_has_live_subscription()` requires the member to own the subscription and applies no gifting carve-out, while the gates' check (`WooCommerce_Connection::get_active_subscriptions_for_user()`, `class-woocommerce-connection.php:487`) excludes a subscription whose `WCS_Gifting` recipient is someone else. On sites running WCS Gifting this cuts both ways: a gift *purchaser* with an active membership classifies as covered and is skipped, but the gate denies them at the flip (silent access loss); a gift *recipient*'s subscription is owned by the purchaser, so the recipient classifies as a residual and receives a redundant $0 subscription (noise, in the safe direction). Routing the owned-subscription check through `get_active_subscriptions_for_user( $user_id, $product_ids, true )` closes both; failing that, the divergence deserves a note beside the existing gift-recipient remark in the docblock at `:1921`.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1689,6 +1884,204 @@ private static function get_manual_only_plan_ids() {
...
+	 * `value` of every `subscription` access rule across them.
+	 *
+	 * This is the same set the pre-migration parity audit calls ACCESS_PRODUCT_IDS
+	 * when it computes who loses access at the flip, so deriving it here keeps the
+	 * command's member selection and that diff on one definition of "covered".
+	 * Premium newsletter gates are excluded — they gate newsletters, not content.
+	 *
+	 * @return int[]
+	 */
+	private static function get_gate_access_product_ids() {
```
dkoo, line 2141: **Suggestion:** Derived access products include gates whose custom access is switched off — `get_gate_access_product_ids()` collects subscription-rule products from every published gate with non-empty `custom_access.access_rules`, but enforcement honors only gates whose `custom_access.active` is truthy (`class-user-gate-access.php:37`, `class-content-gate.php:637`). A published gate with custom access toggled off but rules retained widens the derived list, so a member whose only live subscription is to such a product tests as covered and is skipped — losing access at the flip, the error direction `subscription_covers_access_products()`'s docblock says the code errs against. It also mutes the product-not-among-access-products refusal. The tests stage only active gates, so this seam is uncovered. An `empty( $gate['custom_access']['active'] )` skip would match the enforcement path.

**Risk if not addressed:** If the gates the migration reads from are drafts, or store their subscription rules where this derivation misses them, every member holding any live subscription is classified as covered, the run reports a near-zero residual count that matches nothing, and the whole cohort silently loses access at the flip with no record of who was skipped beyond the run's console output.

**Consider:** Treat "no access products could be determined" as a hard stop for a live run, keeping it a warning only in dry-run. An empty list turns the coverage test into "any subscription counts", which is the setting that skips the most members, and skipping is the direction that costs readers their access.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -43,6 +44,20 @@ class Teams_Migration {
...
+	 * subscription in any of these keeps their access without a migration
+	 * subscription. Dead statuses deliberately do not count — an active
+	 * membership over a lapsed subscription is the comp/legacy residual the
+	 * filter exists to include. `pending` is dead too: a checkout that never
+	 * completed grants no access, so such a member loses out at the flip exactly
+	 * like one with no subscription at all.
+	 *
+	 * @var string[]
+	 */
+	const LIVE_SUBSCRIPTION_STATUSES = [ 'active', 'on-hold', 'pending-cancel' ];
```
dkoo, line 69: **Nit:** `LIVE_SUBSCRIPTION_STATUSES` docblock overclaims what on-hold grants — The docblock says a member holding a subscription in any of these statuses "keeps their access without a migration subscription", but gates grant on-hold only during payment recovery (`Access_Rules::has_active_subscription()` → `WooCommerce_Connection::is_subscription_in_payment_recovery()`; `ACTIVE_SUBSCRIPTION_STATUSES` is `[ 'active', 'pending-cancel' ]`). Skipping on-hold members is nonetheless correct: NPPD-2055's scope note defers the dunning class to NPPD-2052, and `On_Hold_Duration` expires retry-less on-hold subscriptions after the configured window, at which point they fall into the dead-status cohort a later run picks up. Worth correcting the comment to say on-hold is counted live *by design* (citing the scope note), and optionally tallying on-hold skips separately from active/pending-cancel so a parity diff computed with gate semantics has a named explanation for the delta.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -781,6 +923,17 @@ public function migrate_manual_members( $args, $assoc_args ) {
 				$membership_post = \get_post( $membership_id );
 				$user_id         = (int) $membership_post->post_author;
 
+				// Explicit input mode: only members on the reviewed user-ID list are
+				// processed. Skipping the rest silently keeps the output proportional to
+				// the list, not the plan size; matches are tracked so unmatched list
+				// entries can be reported at the end.
+				if ( $explicit_users_mode ) {
+					if ( ! in_array( $user_id, $target_user_ids, true ) ) {
```
dkoo: **Nit:** Explicit-users matching is O(list × memberships) — Explicit-users mode does `in_array( $user_id, $target_user_ids, true )` per membership. Fine at the field sizes in the ticket (≤ ~121 residuals); a reviewed list in the thousands against plans with thousands of memberships would do millions of comparisons. Flipping the list into a keyed set once (`isset()` lookup, as `$matched_user_ids` already does) makes it O(1) per membership.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1689,6 +1884,204 @@ private static function get_manual_only_plan_ids() {
...
+	 *
+	 * @return int[]|\WP_Error Unique user IDs in input order, or an error.
+	 */
+	public static function parse_user_ids( $user_ids_csv, $user_ids_file ) {
+		// A bare `--user-ids` (no `=value`) arrives as boolean true, which a string
+		// cast would turn into '1' — silently targeting user ID 1.
+		if ( ! is_string( $user_ids_csv ) || ! is_string( $user_ids_file ) ) {
+			return new \WP_Error( 'newspack_migration_user_ids_flag', 'The --user-ids/--user-ids-file flags require a value (e.g. --user-ids=101,102).' );
+		}
+		$raw_input = trim( $user_ids_csv );
```
dkoo, line 2228: **Nit:** A UTF-8 BOM in a user-IDs file fails with a confusing error — `parse_user_ids()` reads the file raw, so a UTF-8 BOM (routine in spreadsheet exports, where reviewed user-ID lists tend to come from) makes the first token `\xEF\xBB\xBF101`, failing `ctype_digit` and aborting with `"101" is not a valid user ID` — an error showing an apparently-valid ID. Stripping a leading BOM before tokenizing keeps the strict parse without the confusing message.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -1689,6 +1884,204 @@ private static function get_manual_only_plan_ids() {
...
+	 * are a residual, not a covered member. An empty $product_ids means no product
+	 * constraint, matching a gate whose subscription rule lists no products.
+	 *
+	 * @param int   $user_id     The member user ID.
+	 * @param int[] $product_ids Products that grant access under the gates. Empty
+	 *                           means any product counts.
+	 *
+	 * @return bool
+	 */
+	public static function member_has_live_subscription( $user_id, $product_ids = [] ) {
```
dkoo, line 2038: **Suggestion:** Decide coverage by calling the same code the gates call when they grant access, instead of re-deriving it from subscription statuses here. That closes the gifted-subscription, inactive-gate, and network-subscription divergences in one change, and keeps the command's counts meaningful as a check against the pre-migration audit rather than a restatement of its own assumptions.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -665,31 +710,85 @@ public function migrate_team_products( $args, $assoc_args ) {
...
+				sprintf(
+					'Product %d is not among the products that grant access (%s), so a subscription to it grants no access. Pass --product-id for a product the gates accept, or add %d to a gate\'s "Active subscription" rule first.',
+					$product_id,
+					implode( ', ', $access_product_ids ),
+					$product_id
+				)
+			);
+		}
+
 		$product = \wc_get_product( $product_id );
```
dkoo, line 828: **Suggestion:** During pre-flight, check whether the target product limits users to one subscription and say so in the run header — readers given a $0 subscription to a limited product cannot buy that product again until it is cancelled, which matters most for exactly the lapsed cohort this command now targets.

dkoo (changes requested):
Release risk: HIGH — this tool writes $0 subscriptions to production reader accounts at scale, and one of its own safety promises does not hold end-to-end. WP-CLI strips a bare `--user-ids`/`--user-ids-file` with only a warning, so the in-code abort never runs; composed with `--only-without-live-subscription --live`, a lost flag value silently widens a reviewed-list run into a site-wide sweep of every published plan. I think we should close that before merge (the raw command line is inspectable), and it is the one blocker.
The rest of the guard matrix held under live testing on a seeded env: plan refusal, dry-run classification, live $0 grants with the idempotency marker, re-run reporting, skip-domains, the group-subscription leg, and strict token parsing all behave as documented.
The remaining suggestions cluster around one theme: member coverage is decided by a re-implementation of the gates' access check rather than the gates' own code. Gifted subscriptions diverge in both directions, gates with custom access switched off still contribute products to the derived access list, and an empty access-product list makes any live subscription count as covering — worth deciding deliberately whether a live run should proceed in that state. Suggestions on `--plan-ids` parsing, `--as-group` reconciliation, orphan group subscriptions, and sweep-scale memory round it out.

- `plugins/newspack-plugin/includes/cli/class-teams-migration.php:907` — **Suggestion:** Orphan group subscriptions when a plan has no qualifying members — `create_group_subscription()` runs before any member is evaluated, so an empty active $0 group subscription is created whenever a named plan has no qualifying members — the expected outcome on a purchase plan targeted with `--only-without-live-subscription`, where most members hold live subscriptions. The `--plan-ids` requirement at `:808` narrows how many orphans are possible but doesn't prevent one, and the orphan is an active subscription on a gate-accepted product credited to `--group-owner-id`. Deferring group-subscription creation until the first member passes the filters would make the guard's stated intent real.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: `get_valueless_value_flags()` scans the raw argv for bare value-requiring flags and aborts before any other work, since WP-CLI strips these with only a warning before the command runs. The run header also prints the resolved reviewed-list size. Covered by an end-to-end test that stages `$_SERVER['argv']`, and verified against a live env (the abort now fires right after WP-CLI's warning).

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Documented in 4b3b295d8: the command docblock now says the access-product pre-flight applies to every mode and names `--access-product-ids` as the bypass; the PR description no longer claims the legacy path is unchanged. The runbook in newspack-access-control-migrations still documents the old invocation shape — that update happens in its own repo as a follow-up.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: `--plan-ids` goes through `parse_id_tokens()` like the other ID flags, so a typo'd token aborts the run, and the parse happens before the `--as-group` guard reads the value. The bare-flag leg is covered by the raw-argv guard from the blocker thread.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: liveness is memoized per user for the run, and `Group_Subscription::reset_cache()` runs alongside the per-iteration object-cache clear so the group layer's static can't grow with the member list.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: the per-run granted dedupe now runs before the migration/live filters and covers group mode, so a multi-plan member is counted once, dry-run and live report the same numbers, and the second plan reports the member as already added instead of live-skipped. Pinned by a two-plan test.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: the liveness check mirrors the gates' gifting rule — a gifted subscription covers only its recipient (`WCS_Gifting::get_recipient_user()`), and ownership is required only for non-gifted subscriptions. Both directions are tested with a `WCS_Gifting` stub: the purchaser classifies as a residual, the recipient is skipped as covered.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: `get_gate_access_product_ids()` skips gates whose `custom_access.active` is empty, matching the enforcement path (with an inactive-gate test). Related hardening in the same commit: with `--only-without-live-subscription`, an empty resolved product list now refuses a `--live` run — still a warning in dry-run so the state can be previewed.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: the docblock now states on-hold is counted live by design (NPPD-2052 owns the dunning cohort, and `On_Hold_Duration` auto-expires retry-less on-hold subscriptions so a later run picks the member up), and the skip report tallies by status with a note explaining the on-hold delta against gate semantics.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: the reviewed list is flipped into a keyed set once and membership rows check it with `isset()`.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Fixed in 4b3b295d8: a leading UTF-8 BOM is stripped from file input before tokenizing, covered by a test.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Took the narrower route in 4b3b295d8: the gifting and inactive-gate divergences are mirrored directly, and the on-hold divergence is documented as deliberate with a separate tally in the skip report. Delegating wholesale to `WooCommerce_Connection::get_active_subscriptions_for_user()` would import the payment-recovery on-hold semantics NPPD-2055 explicitly defers to NPPD-2052 — this command must skip the whole on-hold class, not only in-recovery members. Running the `newspack_access_rules_has_active_subscription` filter here would also let network-granted subscriptions count as covering; erring toward a redundant $0 record for those members is the safer direction, and the counts stay explainable against the audit.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Added in ad84a3671: the pre-flight names a limited product in the run header (via `wcs_get_product_limitation()`), with a note to plan win-back outreach to the migrated cohort accordingly. Tested for both limited and unlimited products.

dkoo:
Covering for @adekbadek while he's out: the review's blocker and all suggestions/nits are addressed on this branch (working in a Claude-assisted session).

**Commits since the review:**
- `c80a413ea` — merged `main` (86 commits behind). One real conflict: #678 landed a second `WP_CLI` test stub, and only one `WP_CLI` class can exist per PHPUnit process. The shared recording mock (`tests/mocks/wp-cli-mocks.php`) is now a superset of both surfaces and `wp-cli-mock.php` is a shim requiring it, so both suites pass in either load order. Heads-up for #688: this ends the byte-parity its branch assumed for the mock file.
- `4b3b295d8` — the fix round: raw-argv guard for bare value flags (the blocker), gifting rule mirrored from the gates, inactive gates excluded from the derived access products, empty-access-list refusal on live runs, strict `--plan-ids`, group-mode lazy creation + per-run dedupe, liveness memo + group-cache reset, status-breakdown reporting with the NPPD-2052 note, and the smaller items (keyed-set matching, BOM tolerance, docblocks).
- `ad84a3671` — pre-flight notice when the migration product limits customers to one subscription.

**Verification:** full plugin suite green (2,725 tests; the command's group grew from 24 to 34), PHPCS clean, and each changed behavior exercised against a live isolated env with a seeded status matrix. Per-thread details are in the review replies; the PR description is updated to match the shipped behavior.

**Follow-up:** the `migrate-access-control` runbook still documents the pre-guard invocation shape — updating it in its own repo.

Since the changes-requested review and the fixes now come from the same person, a fresh set of eyes should make the merge call.

PR #60, opened 2026-03-24
Assorted improvements to site setup

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

- Fix DB error spam during `n setup` caused by WooCommerce/Action Scheduler querying tables after `db reset` but before `core install` recreates them. Plugins are now deactivated before the reset.
- Add `woocommerce-gateway-stripe` to the WooCommerce plugins list so it is activated during `--woocommerce` setup.
- Auto-configure Stripe gateway in test mode during `--woocommerce` setup when `STRIPE_TEST_PUBLISHABLE_KEY` and `STRIPE_TEST_SECRET_KEY` env vars are set.
- Update repo list: rename `newspack-manager-client` to `newspack-manager-admin`, add `newspack-subscription-migrations`.
- Prevent WooCommerce Subscriptions from treating the local dev site as a duplicate (which forces all subscriptions to manual renewal).
- Make theme symlink failures non-fatal in `link-repos.sh`.

### How to test the changes in this Pull Request:

1. Start a fresh environment: `n start`.
2. Run `n setup --yes` and verify the output has **no** `wpdberror` HTML in the console.
3. Run `n setup --yes --woocommerce` and verify:
   - No DB errors in the output.
   - `woocommerce-gateway-stripe` is listed as active: `n wp plugin list | grep stripe`.
   - WooCommerce Subscriptions does not show "duplicate site" notice in WP admin.
4. To test Stripe auto-config: add `STRIPE_TEST_PUBLISHABLE_KEY=pk_test_xxx` and `STRIPE_TEST_SECRET_KEY=sk_test_xxx` to `.env`, then run `n setup --yes --woocommerce`. Verify Stripe settings are configured: `n wp option get woocommerce_stripe_settings --format=json`.
5. Verify `newspack-manager-admin` and `newspack-subscription-migrations` are symlinked: `n wp plugin list | grep -E 'manager-admin|subscription-migrations'`.
[...]

PR #693, opened 2026-07-22
feat(content-gate): grant paid access from one-time purchases with an access duration (NPPD-2053)

adekbadek:
### All Submissions:

WCM plans can grant membership from simple (non-subscription) products with a plan-defined membership length – prepaid annual passes, lifetime memberships. Access Control's Paid access only checked for an active subscription, so buyers of one-time products were never granted access, even mid-term on access they already paid for.

This adds a new **`one_time_purchase`** Paid-access rule:

- Value per rule instance: selected products (simple + variable) plus an **access duration** – N days/months from purchase, or **forever** (lifetime). The order's creation date is the anchor, per the ticket.
- Orders count when in a **paid status** (`wc_get_is_paid_statuses()`: processing/completed) – a full refund or cancellation revokes access with no extra code. Guest orders count on both duration paths (matched by user ID or billing email, mirroring `wc_customer_bought_product`).
- Evaluation is **retroactive over order history**, so existing prepaid/lifetime purchasers get access at the flip with zero data migration.
- **Fail-closed semantics**: malformed values deny. Rule values are sanitized on the REST write path AND re-sanitized at evaluation (block-visibility rules arrive via post content and bypass REST); an unrecognized/missing duration unit denies rather than widening to lifetime; finite duration of 0 denies (with a visible UI warning); an evaluation with no user ID and no resolvable email denies instead of querying every customer's orders – on the finite path because an empty customer constraint would widen the query to every customer's paid orders, on the lifetime path because `wc_customer_bought_product()` returns the `woocommerce_pre_customer_bought_product` filter's value verbatim ahead of its own identity check.
- The finite window is one cutoff (`now − N`) shared by every order, compared strictly. On month-end anchors that is the deny-biased reading, and closer to what "N months from purchase" means on a calendar than PHP's forward rollover (`+1 month` from Jan 31 lands on Mar 3).
- Performance: finite durations use one bounded `wc_get_orders` query (customer + paid statuses + date window); forever uses WC-cached `wc_customer_bought_product`; both memoized per request. Product-options resolution is memoized per request (the API resolves options once per rule per gate save).
- UI: a shared React control (product token field + duration select) used by both the gate wizard and the block-editor visibility panel; gate summary and the user-profile access panel render the composite value legibly (object rule values previously crashed the summary's React rendering). Product tokens are keyed by product ID, so two products sharing a name stay distinguishable and a stored ID missing from the options list (a variation, a deleted product) survives an edit. Localized rule payloads now strip PHP callables (`callback`/`sanitize_callback`) via a shared helper.
- Tooling: `newspack-scripts test` only collected `*.test.js(x)`, so TypeScript test files were silently never run (the existing `block-visibility.test.ts` among them). `testMatch` now covers `.ts`/`.tsx` too; the whole newspack-plugin suite passes with them collected.
- Separate rule by design (not an extension of the subscription rule): different semantics, composes with subscription rules via OR-groups, and the `Access_Rules` contract consumed by newspack-manager is untouched (purely additive).

Merge note: expect one trivial keep-both conflict with #687 in `class-access-rules.php` (both insert class members at the same anchor); other shared files auto-merge.

Migration-side follow-up (WCM plan length → this rule's value) is tracked in NPPD-2106.

Closes NPPD-2053.

1. Enable `NEWSPACK_CONTENT_GATES` + RAS, create a simple product and a gate with Paid access; add the "One-time purchase" rule with the product and a 365-day duration.
2. Reader with a completed order from 100 days ago → full content. Backdate the order past 365 days → gated. Switch the rule to "Forever" → access restored.
3. Refund the order → gated (on both finite and forever).
4. Guest checkout with the reader's email (no account linked to the order) → access granted on both duration paths.
5. Verify the control renders in both the gate wizard and the block editor's visibility panel; clear the duration number → warning appears, purchases deny.
6. Create two products with the same name, select one in the rule → only that product's ID is stored.
7. `n test-php --filter One_Time_Purchase` (28 tests) and `--group Access_Rules` (94); `npm test` for the product-token bijection tests.

### Other information:


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2053-parity-fix-one-time-prepaid-lifetime-purchases-cant-grant` (stash local work first).

dkoo (commented):
Release risk: LOW — zero blockers, consequences verdict low, compatible cross-plugin impact, and verification is now complete: the grant/deny engine was exercised end-to-end (unit suites plus a 15-case runtime evaluation matrix), and the browser walkthrough confirmed the gate wizard, block-editor Custom rule, profile panel, and frontend gating/granting all behave as designed, including honest fail-closed surfacing of invalid stored durations in both the wizard select and the gate-list summary. Two suggestions worth considering before or shortly after merge: an explicit empty-identity guard in customer_bought_product_after() (today unreachable, but the method is public cross-plugin API and both WC stores silently drop an empty customer constraint — turning it into match-anyone), and bounding the product-options payload, which serializes the whole simple+variable catalog into every block-editor load (measured at 1,425 bytes with the 3 seeded products — trivial here, but it scales linearly with catalog size).

- `plugins/newspack-plugin/includes/content-gate/class-access-rules.php:265` — **Nit:** Stale docblock invariant about empty rule values — The comment near `evaluate_anonymous_rules()` (~line 265) states generally that evaluators treat empty values as the rule's no-constraint semantics and return true. The new rule deliberately diverges — empty `product_ids` denies, which is the right fail-closed choice — so a one-line amendment keeps a future `supports_anonymous` change from leaning on a stale invariant.
- `plugins/newspack-plugin/includes/content-gate/class-user-gate-access.php:118` — **Nit:** Profile panel reports '(any)' for an empty value that actually denies — `format_rule_value()` returns `'(any)'` from the generic `empty( $value )` branch before reaching the `one_time_purchase` branch — but for this rule an empty value *denies* (empty `product_ids` fails closed), so the debug panel would report the opposite of the evaluation result for exactly the malformed/legacy data it exists to debug. Reordering the rule-slug check above the `empty()` check makes the panel truthful. **Runtime note (Stage C):** not reproducible with realistic data — the seeded gates always store a non-empty composite value, and a malformed-unit value renders honestly as "(invalid duration, grants no access)" (verified in-browser via a scratch gate). The `(any)` label only appears for a fully empty value, which the sanitizer prevents saving; cosmetic nit stands.
- `plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx:398` — **Nit:** Help text composes a sentence from translated fragments — The visibility help (~lines 396–401) `sprintf`s a standalone translated word (`gates` / `rules`) into a sentence template — fragments carry no context for translators and grammatical agreement breaks in gendered languages. Two complete sentences selected by mode would be cleaner. Flagged softly since the mode toggle may predate this branch.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -411,6 +460,171 @@ public static function has_active_subscription( $user_id, $product_ids, $strict
...
+		// rule sanitized on a gate save (Content_Gate_API resolves all rule options
+		// for each rule in the payload).
+		static $options = null;
+		if ( null !== $options ) {
+			return $options;
+		}
+		if ( ! function_exists( 'wc_get_products' ) ) {
+			return [];
+		}
+		$products = \wc_get_products(
```
dkoo, line 577: **Suggestion:** Entire product catalog serialized into every block-editor payload — `get_one_time_purchase_products_options()` runs `wc_get_products( [ 'type' => [ 'simple', 'variable' ], 'limit' => -1 ] )` and the full result is resolved into the localized payload on every block-editor load (`newspackBlockVisibility` enqueues on all post/page editors) and both gate wizards. Unlike the subscription catalog (inherently small), a shop can hold thousands of simple products — each editor load then pays an unbounded query plus a full catalog dump into the page HTML, even on posts that use no gate. The per-request memo doesn't help across requests. Consider a bounded or searchable options source (REST-backed product search, like the dynamic-options pattern used by the institution rule), or at minimum a sane limit.

**Runtime evidence (Stage C):** in the review env (3 one-time products) `window.newspackBlockVisibility` totals 1,425 bytes, of which the `one_time_purchase` rule is 391 bytes. Trivial at this size — the concern is linear growth on large catalogs, so this stays a suggestion, not a blocker.

**Consider:** Bound or lazy-load the product options for the one-time purchase rule (paginate, cap the count, or fetch on demand from the control) rather than serializing the full catalog into every block-editor payload, so large-catalog sites do not pay a full product query and inline dump on every editor load.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -411,6 +460,171 @@ public static function has_active_subscription( $user_id, $product_ids, $strict
...
+	 * @param int[]  $product_ids Product IDs to look for.
+	 * @param int    $cutoff      Unix timestamp orders must be created after.
+	 *
+	 * @return bool
+	 */
+	private static function customer_bought_product_after( $user_id, $email, $product_ids, $cutoff ) {
+		$paid_statuses = function_exists( 'wc_get_is_paid_statuses' ) ? \wc_get_is_paid_statuses() : [ 'processing', 'completed' ];
+		$orders        = \wc_get_orders(
+			[
+				'customer'     => array_values( array_filter( [ $user_id, $email ] ) ),
```
dkoo: **Suggestion:** Empty-identity path in customer_bought_product_after() fails open — If `has_one_time_purchase()` is ever reached with `user_id = 0` and no resolvable email, `array_filter( [ 0, '' ] )` collapses the `customer` argument to an empty array — and both WooCommerce order stores silently drop an empty customer constraint, so the finite-duration query would match *any* customer's paid order in the window and grant access. This is unreachable today (`evaluate_rule()` short-circuits anonymous users because the rule doesn't declare `supports_anonymous`, and the forever path's `wc_customer_bought_product()` returns false with no identity), but `has_one_time_purchase()` is a public method on a cross-plugin API surface. An explicit `if ( ! $user_id && ! $email ) { return false; }` before building the query makes the method's fail-closed contract independent of its caller.

**Consider:** Although evaluate_rule blocks anonymous evaluation today so this method is never reached with an empty customer, add an explicit guard that returns false when no user ID and no email are available before building the wc_get_orders customer constraint, so this public method fails closed rather than matching any customer's order if a future caller invokes it directly for a logged-out visitor.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
```diff
@@ -0,0 +1,153 @@
...
+		durationHelp = __( 'The stored duration is invalid and grants no access. Pick a duration to fix this rule.', 'newspack-plugin' );
+	}
+
+	return (
+		<>
+			<TokenField
+				label={ productsLabel }
+				value={ selectedLabels }
+				suggestions={ options.map( option => option.label ) }
+				onChange={ ( tokens: ( string | { value: string } )[] ) => {
```
dkoo, line 135: **Suggestion:** Product tokens keyed by label — duplicate names collapse — Token selection round-trips through product *labels*: stored IDs map to labels for display, and `onChange` maps chosen labels back via `options.filter( option => labels.includes( option.label ) )`. Two products sharing a name become indistinguishable — selecting one token selects every ID with that label. This mirrors the existing subscription-control pattern, but this rule targets the general catalog where duplicate names are far likelier. Keying tokens by value would remove the ambiguity.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -411,6 +460,171 @@ public static function has_active_subscription( $user_id, $product_ids, $strict
...
+					// user ID and billing email, so guest orders count), runs SQL-side,
+					// and is cached by WooCommerce with invalidation on order writes.
+					foreach ( $value['product_ids'] as $product_id ) {
+						if ( \wc_customer_bought_product( $email, $user_id, $product_id ) ) {
+							$has_purchase = true;
+							break;
+						}
+					}
+				} elseif ( in_array( $value['duration_unit'], [ 'days', 'months' ], true ) && $value['duration_value'] > 0 ) {
+					// Note: month arithmetic follows strtotime()'s rollover semantics
```
dkoo: **Nit:** Months window anchored on now−N rather than purchase+N — The finite window computes a shared cutoff of `strtotime( '-N months' )` while the UI copy promises "N months from purchase". These diverge by a few days around month-end rollovers (e.g. on Mar 1, a Jan 31 purchase is denied under `-1 month`, though purchase+1 month would still grant). The divergence is deny-biased and small; if exactness ever matters, compare `order_date + N months <= now` per order inside the existing single query's item loop.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules-one-time-purchase.php
```diff
@@ -0,0 +1,454 @@
...
+		$this->assertFalse(
+			Access_Rules::has_one_time_purchase( self::$purchaser_user_id, $this->get_rule_value() ),
+			'An order for a different product should not grant access.'
+		);
+	}
+
+	/**
+	 * A purchase of a variation grants access when its parent product is selected.
+	 */
+	public function test_variation_purchase_grants_access_via_parent_product() {
```
dkoo, line 260: **Nit:** Edge cases worth pinning with tests — The 20-test matrix is thorough on statuses, products, guest email, and fail-closed paths, but a few edges the code itself calls out are unpinned: the strict `>` cutoff boundary (an order created exactly at the cutoff), the documented month-rollover semantics, guest-email case-sensitivity (the finite path relies on DB collation while the forever path lowercases), the server-timezone vs GMT interaction in the `strtotime()` cutoff, and a rule value storing a variation ID directly (the `$item_variation_id` branch has no coverage on either duration path, and the sanitizer doesn't whitelist IDs against options).

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Added the guard in 6ba64a0: `customer_bought_product_after()` now returns false before building the query when neither a user ID nor an email resolves, so the fail-closed contract holds for any caller rather than depending on `evaluate_rule()`. `test_missing_identity_denies_access` pins it — `has_one_time_purchase( 0, … )` must not match another customer's order on either duration path.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
adekbadek: Tokens are keyed by product ID as of 6ba64a0. `getProductTokens()` builds a bijection between IDs and token strings — appending `(#id)` to any product name that isn't unique — and `onChange` maps the chosen tokens back through that map instead of filtering options by label, so two products sharing a name stay distinguishable. It also gives a stored ID with no matching option (a variation, or a deleted product) a `#<id>` token, so editing the field no longer silently drops it.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Kept the shared cutoff and wrote the reasoning into the code (6ba64a0), because per-order `purchase + N months` diverges the other way and by more: PHP rolls `+1 month` from Jan 31 forward through Feb 31 to Mar 3, granting three extra days, whereas `-1 month` from Mar 1 lands on Feb 1 and ends the grant once the calendar month is up. So the deny-biased reading is also the one closer to what "N months from purchase" means on a calendar, and it keeps one cutoff for every order. `test_months_window_is_anchored_on_now_minus_duration` pins the anchor: an order created exactly N months ago is outside the window, one created a day inside it grants. Happy to swap to per-order expiry if you read the month-end case the other way.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules-one-time-purchase.php
adekbadek: All five are pinned now (6ba64a0), taking the file to 27 tests:

- `test_cutoff_boundary_is_strict` — an order created exactly at the cutoff is outside the window; an hour inside it grants.
- `test_months_window_is_anchored_on_now_minus_duration` — the anchor the rollover comment describes.
- `test_window_is_independent_of_site_timezone` — site timezone set to UTC+14 against a one-day duration at ±1h, which a locally-formatted comparison would flip; both sides are absolute timestamps.
- `test_variation_id_stored_in_rule_value_grants_access` — a variation ID stored directly in the rule value, on both duration paths, with the docblock noting that the sanitizer deliberately doesn't whitelist against the options list.
- `test_guest_order_matches_billing_email_case_insensitively` — this needed the mocks corrected first: `wc_get_orders()` and `wc_customer_bought_product()` compared billing emails with `===`, while real WC compares them in SQL under a case-insensitive collation, so both now use `strcasecmp()`.

adekbadek:
Amended the invariant on `evaluate_anonymous_rules()` (6ba64a0). The docblock now says that how an evaluator reads an empty value varies by rule — `email_domain` returns true as its own no-constraint semantics, `one_time_purchase` denies — and that this check deliberately doesn't delegate the decision, so a future `supports_anonymous` rule can't lean on either reading.

---

Reordered `format_rule_value()` so the `one_time_purchase` branch runs ahead of the generic `empty( $value )` check. An unconfigured rule now reads "(no products selected) (invalid duration, grants no access)" instead of "(any)", which matches how it just evaluated. `test_format_rule_value_describes_unconfigured_one_time_purchase` covers that and asserts "(any)" still applies to a rule like `email_domain`, whose evaluator does treat an empty value as no constraint.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
```diff
@@ -0,0 +1,195 @@
...
+ * absent from the options list (a variation, or a deleted product) keeps a
+ * `#<id>` token so editing the field doesn't silently drop it.
+ */
+function getProductTokens( options: RuleOption[], productIds: Array< string | number > ) {
+	const nameCounts = new Map< string, number >();
+	options.forEach( option => nameCounts.set( option.label, ( nameCounts.get( option.label ) ?? 0 ) + 1 ) );
+
+	const tokenByValue = new Map< string, string >();
+	const valueByToken = new Map< string, string | number >();
+	const addToken = ( value: string | number, token: string ) => {
```
dkoo: **Suggestion:** De-duplicate generated tokens against all display labels; add a JS test — The `Name (#ID)` disambiguation only triggers for duplicate *original* names, so a real product whose name already matches a generated or fallback token string (e.g. a product literally named `Annual Pass (#60)`, or named `#123` colliding with a delisted-ID token) can still produce identical FormTokenField strings, and `valueByToken.set()` overwrites the earlier mapping — saving the wrong product ID. Unlikely in practice, cheap to close: de-duplicate the generated set against every display label, not just duplicated names. The new `getProductTokens` bijection is also non-trivial logic with no JS test — worth pinning while touching this.

plugins/newspack-plugin/tests/mocks/wc-mocks.php
```diff
@@ -1031,18 +1045,58 @@ function wc_get_orders( $args ) {
...
 		// Filter by customer.
 		$orders = array_filter(
-			$orders_database,
+			$orders,
 			function( $order ) use ( $args ) {
 				return $order->get_customer_id() === $args['customer_id'];
 			}
 		);
 	}
+	if ( isset( $args['customer'] ) ) {
```
dkoo: **Suggestion:** Mock inverts real WC's empty-customer semantics — the guard's regression test is a no-op — The mock filters orders with `isset( $args['customer'] )`, so `'customer' => []` matches *nothing*. Real WooCommerce does the opposite — both stores gate on `! empty( $query_vars['customer'] )` and drop the empty constraint, matching *every* customer's orders (the exact widening the new guard defends against). Consequence: delete the guard and `test_missing_identity_denies_access` still passes (the mock returns zero orders) while production grants access from strangers' orders. Change the mock to `! empty( $args['customer'] )` so the test genuinely fails without the guard. Related: the test's forever-path assertion also leans on mock behavior rather than the guard — either add the symmetric identity guard on the forever branch or scope the assertion to what the mock can faithfully prove.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
```diff
@@ -0,0 +1,195 @@
...
+	let durationHelp: string = __( 'How long a purchase grants access, counted from the order date.', 'newspack-plugin' );
+	if ( 'forever' === currentValue.duration_unit ) {
+		durationHelp = __( 'Purchasers keep access forever.', 'newspack-plugin' );
+	} else if ( '' === currentValue.duration_unit ) {
+		durationHelp = __( 'The stored duration is invalid and grants no access. Pick a duration to fix this rule.', 'newspack-plugin' );
+	}
+
+	return (
+		<>
+			<TokenField
```
dkoo, line 131: **Nit:** Free-typed unknown tokens silently vanish — FormTokenField accepts arbitrary typed tokens by default; a token absent from the bijection maps to undefined, filters out of `product_ids`, and disappears on the next render. Fail-safe but confusing — `__experimentalValidateInput={ token => valueByToken.has( token ) }` rejects unknown input at entry (verify the wrapper forwards the prop).

dkoo (commented):
Release risk: LOW — the follow-up resolves all three targeted findings correctly: the fail-closed guard sits at the right layer, the token bijection covers display, selection, dedup, and delisted IDs, and the panel reorder is pinned by a test. One thing genuinely worth fixing before merge, found while auditing the fix's own tests: the WC mock's empty-customer semantics invert real WooCommerce's — the mock matches nothing for an empty customer constraint where both real WC stores drop the constraint and match everything — so the new missing-identity regression test still passes if the guard is deleted, while production would grant access from strangers' orders. Flipping the mock to real WC's behavior makes the test actually pin the guard. Two smaller notes: the generated display tokens should de-duplicate against all display labels (a product literally named like a generated token can still collide), and the new token logic deserves a small JS test. Nothing blocking.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
adekbadek: Done in 88b7107. Token generation now reserves against every token already handed out rather than only against duplicate names: each product takes its preferred string if free, otherwise the `Name (#ID)` form, otherwise that plus a counter. So a product literally named `Annual Pass (#60)` no longer overwrites the mapping for product 60, and a product named `#123` no longer collides with the `#123` fallback token for a delisted ID — in both cases the earlier claimant keeps its token and the later one is bumped.

`getProductTokens` is now exported and pinned by `one-time-purchase-rule-control.test.ts`, with a shared `expectBijection` helper asserting every ID round-trips through its token and that no two products share one. Both collision cases have their own test; I checked they fail if the uniqueness loop is removed.

Getting those tests to run turned up something separate: `newspack-scripts test` matched only `*test.js(x)`, so no TypeScript test file has ever been collected — including `block-visibility.test.ts`, which has been dormant since #4646. `testMatch` now covers `.ts`/`.tsx`; the full newspack-plugin suite is green with both TS suites collected (53 suites, 475 tests).

plugins/newspack-plugin/tests/mocks/wc-mocks.php
adekbadek: Right on both counts — fixed in 88b7107.

The mock now gates on `! empty( $args['customer'] )`, matching how both order stores test the query var, so an empty customer drops the constraint and returns every customer's orders. With that, `test_missing_identity_denies_access` is load-bearing: replacing the guard with `if ( false )` makes the finite assertion fail (`Failed asserting that true is false`), and it passes again with the guard restored.

For the forever path I took the first option and made the denial ours rather than WooCommerce's. The identity list is built once in `has_one_time_purchase` and guarded there, ahead of both branches, so lifetime access no longer leans on `wc_customer_bought_product()` happening to bail when its customer data is empty. `customer_bought_product_after()` takes the already-built list and documents the non-empty precondition instead of re-checking it, so there is exactly one guard and deleting it fails a test.

plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx
adekbadek: Added `__experimentalValidateInput={ token => valueByToken.has( token ) }` in 88b7107, so an unknown typed token is rejected at entry instead of being accepted and then vanishing on the next render. The Newspack-styled wrapper spreads `...otherProps` straight onto the core `FormTokenField` (packages/components/src/form-token-field/index.js:38), so the prop reaches core on the wizard surface too.

adekbadek:
The visibility help text no longer composes a sentence from fragments. `block-visibility.tsx` now selects between two complete translatable sentences by mode — "Content visibility for readers who match any of the selected gates." and "…selected rules." — so translators get a whole sentence with its context and gendered languages can agree. The `sprintf` import went with it. In 88b7107.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Deferred to NPPD-2132 (TODO added at the function in a80797b7c).

The cited cheap path is not a drop-in: `Institution::get_options()` is also a full `posts_per_page => -1` dump, and there is no REST product/institution search endpoint or async options control anywhere in `content-gate` today. The proper fix — a paginated REST product-search endpoint plus reworking the synchronous `TokenField` (whose product-token bijection assumes the full options list is present client-side) into an async, search-as-you-type control — is feature-sized, so it is scope creep and regression risk on this parity PR. Given the non-blocking runtime evidence (391 bytes for the rule at 3 products), it is tracked separately rather than landed here.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -411,6 +461,193 @@ public static function has_active_subscription( $user_id, $product_ids, $strict
...
+			$memo_key = $user_id . ':' . md5( wp_json_encode( $value ) );
+			if ( isset( self::$one_time_purchase_memo[ $memo_key ] ) ) {
+				$has_purchase = self::$one_time_purchase_memo[ $memo_key ];
+			} else {
+				$user     = \get_userdata( $user_id );
+				$email    = $user ? $user->user_email : '';
+				$customer = array_values( array_filter( [ $user_id, $email ] ) );
+				if ( empty( $customer ) ) {
+					// Fail closed with no identity to match a purchase against, on both
+					// paths: an empty customer constraint is dropped by both WooCommerce
```
dkoo: **Nit:** Guard rationale slightly overstates the forever path's risk — The shared comment ("an empty customer constraint is dropped by both WooCommerce order stores") is accurate for the finite path's `wc_get_orders`, but the forever branch's `wc_customer_bought_product()` already fails closed on empty identity in real WooCommerce — the guard there is defensive symmetry, not a fail-open fix. One clause distinguishing the two keeps a future reader from concluding the forever path was ever exposed. Relatedly, the forever assertion in `test_missing_identity_denies_access` is non-arming (the mock fails closed on that path regardless of the guard) — faithful to real WC and fine as written; noting so nobody assumes both assertions pin the guard.

dkoo (commented):
Release risk: LOW — both findings from the previous pass are correctly and completely resolved: the mock now mirrors real WooCommerce's empty-customer behavior, which arms the missing-identity regression test end-to-end (deleting the guard now fails it), and the token bijection de-duplicates against every display label with a solid new test. Two comment-only notes: the shared guard rationale describes the finite path's fail-open risk accurately but could mislead about the forever path (which fails closed in real WooCommerce regardless — the guard there is defensive symmetry, not a bug fix), and the forever assertion in the missing-identity test is non-arming since the mock fails closed on that path anyway — fine as written given the symmetric-guard route was chosen, just worth knowing it guards the finite path only. Nothing further needed from this review's perspective.

dkoo (commented):
Delta re-review of `f196c0a86` (single commit, root `composer.json` only).

The change itself is defensible and the rationale in the `pinned-dependencies` note is accurate — no committed lockfile means CI re-resolves on every run, so a new VIPCS minor does land in every open PR at once. But this collides with work already in flight, and I think the collision matters more than the change:

**Main deliberately pinned this, and Dependabot undid it.** [#740](https://github.com/Automattic/newspack-workspace/pull/740) pinned `~3.0.1` "until posts_per_page audit (NPPM-3055)" on Jul 27 19:11; [#744](https://github.com/Automattic/newspack-workspace/pull/744) (Dependabot) bumped it to `~3.1.0` nine hours later, before that audit happened. So `main` is currently on 3.1.0 with **no** mitigation in `phpcs.xml` — which is why this PR is hitting the new sniffs at all. The root problem is on main, not here.

**Two PRs are now fixing this incompatibly.** [#726](https://github.com/Automattic/newspack-workspace/pull/726) takes the opposite approach in the same shared config surface: it keeps `~3.1.0` and instead excludes the two new `NoPaging` error codes in root `phpcs.xml`, plus fixes three call sites. If both merge, the repo ends up on 3.0.* with a ruleset written for 3.1.0 — the exclusions become dead config, and whoever merges second gets to reconcile it.

Both approaches are reasonable; what is not reasonable is deciding it twice, in two unrelated feature PRs, by whoever rebases last. This commit's own note argues adopting 3.1.0 "belongs in its own PR rather than arriving unannounced" — the same logic applies to reverting it. Suggest lifting this into a dedicated CI PR against main (re-pin, or adopt 3.1.0 with the ruleset change, pick one) and dropping it from both feature branches, so this PR goes back to being about one-time purchase access grants.

No concerns with the product code — unchanged since my last review.

Review provenance: orchestrator-verified directly (single-file CI delta; no reviewer fan-out dispatched — the decisive evidence was the composer.json history on main, checked with git); release risk: LOW for the product change, but see the coordination note above.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Right on both counts — split the comment in 0da419c. The finite path is now stated as the one that depends on the guard (empty `customer` dropped by both order stores), with the forever path called out as defensive symmetry since `wc_customer_bought_product()` returns false itself when it has neither a user ID nor an email (`count( $customer_data ) === 0` in `wc-user-functions.php`).

Also noted it in the test docblock rather than leaving it implicit: the finite assertion pins the guard, the forever one documents parity — the mock fails closed on empty identity exactly as real WooCommerce does, so it would pass with the guard removed.

dkoo (commented):
Delta re-review of `0da419cb6` — confirmed comments-only: both hunks are pure comment/docblock text, no logic or assertion changes.

I verified the new factual claim rather than taking it on trust, since it is now documented as the reason a guard is redundant. It holds: `wc_customer_bought_product()` does fail closed on its own — `wc-user-functions.php:466` has an explicit `if ( empty( $emails ) && ! $user_id ) { … return false; }` before any query is built.

One nuance that argues the forever path is *more* than defensive symmetry: the function opens with `apply_filters( 'woocommerce_pre_customer_bought_product', null, … )` and returns that value verbatim when it is non-null — **before** the identity guard. So a third-party plugin hooking that filter can make the function return truthy with no identity at all. Your `empty( $customer )` check fires before the call and is what actually keeps that from widening access on the forever path. So the guard is load-bearing against a filter-shaped foot-gun, not just symmetry — worth saying, because "documents parity rather than arming anything" invites a future cleanup that removes it.

Still open from my previous review: the vipwpcs CI change (`f196c0a86`) is still on this branch, and [#726](https://github.com/Automattic/newspack-workspace/pull/726) still carries the incompatible counterpart. That decision is the one thing keeping this from being otherwise ready.

Review provenance: orchestrator-verified directly (comments-only delta; no reviewer fan-out — the decisive check was reading WooCommerce's own source); release risk: LOW.

adekbadek:
Resolved by lifting the CI decision to main, as suggested. NPPM-3055 (the posts_per_page audit) landed on `main` in #741, which adopts vipwpcs `~3.1.0` and annotates the repo-wide `NoPaging` call sites — so the version question is now decided there, once, rather than in this feature PR.

`main` is merged into this branch (fdfaae804) and `composer.json` is taken verbatim from main: the `3.0.*` pin and its `pinned-dependencies` note are gone, and `git diff origin/main -- composer.json` is empty, so this PR carries no CI change. Verified under 3.1.0 in the merged tree:

- `WordPressVIPMinimum.Performance.NoPaging` runs clean on this PR's changed files, including the new `get_one_time_purchase_products_options()`. The sniff targets `posts_per_page` / `numberposts`, not WooCommerce's `wc_get_products( 'limit' => -1 )`, so this PR's own catalog query needs no annotation (the unbounded-catalog concern is tracked separately in NPPD-2132).
- Full plugin lint equals main's post-#741 state plus this PR's changed files, which pass.
- `Access_Rules` suite green (44 tests, 76 assertions).

Back to being about one-time purchase access grants.

adekbadek:
You're right that the forever path is load-bearing rather than symmetry — `wc_customer_bought_product()` returns the `woocommerce_pre_customer_bought_product` value verbatim whenever it is non-null, ahead of its own identity check, so a plugin on that filter can answer truthy for nobody in particular. In daecd75 the comment now says that, and rather than leave it as prose that a cleanup could still talk itself past, the guard is armed by a test: the WC mock applies the filter the way real WC does, and `test_missing_identity_denies_access_against_pre_bought_filter()` hooks `__return_true` and asserts an anonymous evaluation still gets no lifetime access. Stubbing the `empty( $customer )` branch out makes it fail. The neighbouring docblock lost its "documents parity rather than arming anything" line, which was the wrong reading of that assertion.

dkoo (commented):
The forever-path guard now has a test behind it. `test_missing_identity_denies_access_against_pre_bought_filter()` hooks `woocommerce_pre_customer_bought_product` to return truthy for no identity, and the mock short-circuits on that filter ahead of its own identity check — so a lifetime evaluation with user 0 reaches the guard as the only thing between an anonymous check and a truthy answer. Drop the guard and the test fails. The docblock is corrected to say that plainly. That closes the note from the prior round.

The vipwpcs question is settled here too: merging `main` brought in the NPPM-3055 fix (#741), so this branch's `composer.json` is back to `~3.1.0` and no longer carries the pin I flagged against #726. The two aren't in conflict anymore.

Release risk: LOW. The product code is unchanged since the prior review — this delta is the test plus the merge of `main`.

Review provenance: orchestrator-verified directly (test-only branch delta; I read the test and the mock and traced that the guard is the sole defense on the forever path). No reviewer fan-out for a test-only change.

adekbadek:
@dkoo Re-requesting review – the changes you asked for have been pushed and need a look before this merges.

dkoo (approved):
Manual QA in an isolated env confirms the core behavior end to end: a reader who bought the qualifying one-time product gets access to the gated post, and one who didn't — or whose order was refunded, or whose finite window has passed — is gated. That holds across the finite "days from purchase" path and the "forever" lifetime path, and for both account-linked and email-matched orders. Direct rule evaluation and the full front-end gate agree.

Release risk: LOW. Approving. Two non-blocking items below.

🟢 **Guest checkout — non-blocking edge.** A one-time purchase completed as a true anonymous guest doesn't grant access: the rule resolves the buyer from the authenticated identity (`has_one_time_purchase()` derives the email from the current user), so an unidentified visitor hits the fail-closed guard. This only reproduces with `Newspack\Donations::disable_guest_checkout` switched off — on Reader-Activation sites that filter forces `woocommerce_enable_guest_checkout = no`, so purchasers are always registered readers and the anonymous path isn't reachable in the shipping configuration. Worth keeping in mind for any future config that allows guest checkout (resolving via the order's billing email would cover it), but nothing to change here.

🟢 **Alignment nit — recommend fixing before merge; no re-review needed.** In the One-time purchase rule editor, "Access duration" and "Number of days" bottom-align, so "Number of days" drops below its sibling once the duration column shows its help text. The `<Flex align="flex-end">` at `plugins/newspack-plugin/src/content-gate/components/one-time-purchase-rule-control.tsx:151` should be `align="flex-start"`.

Current:

Access duration and Number of days misaligned: Number of days drops below its sibling

Expected:

Access duration and Number of days top-aligned

Review provenance: multi-reviewer code review (deep, standards, WP expert, Codex) across the full review and delta rounds, plus manual QA in an isolated env.

adekbadek:
Switched that `<Flex>` to `align="flex-start"` in e71d727, so "Access duration" and "Number of days" now share a top edge and the duration column's help text hangs below instead of pushing its sibling down. Confirmed in the editor: both labels and both inputs render at the same y, on the invalid-duration and finite-duration states alike.

PR #697, opened 2026-07-22
fix(segmentation): stop emitting and scrub bad donor merge tags

chickenn00dle:
### All Submissions:

**Stops Campaigns from decorating newsletter links with a donor merge tag on sites that never configured one.**

`append_donor_segment_param()` (shipped in `newspack-popups@3.15.0`, #327 / NPPM-2876) fell back to `DEFAULT_DONOR_MERGE_FIELD` — `'DONAT'` — whenever the `newspack_popups_mc_donor_merge_field` option was absent. That default is a **name fragment** for the Mailchimp-only substring matching in `reader_logged_in()` (`strpos( $field_name, $donor_merge_field )`), not an ESP merge tag. So every site with an ESP connected and the setting untouched started emitting a tag no ESP resolves.

`Tracking\Utils::get_merge_tag()` is ESP-aware for *syntax* only — it wraps whatever field it's handed in the active provider's delimiters. On ActiveCampaign that produced `?np_seg_donor=%DONAT%`, and `%DO` is not a well-formed `%XX` escape. Jetpack Instant Search percent-decodes every query param on page load with no `try/catch`, so it threw `URIError: URI malformed`, failed to mount, and blanked the page — the white screen Evanston Roundtable and The City Reporter reported.

What changed:

- **No default merge field.** The option now falls back to `''`, making the feature opt-in. Sites that never configured a donor merge field emit nothing.
- **Percent-unsafe merge tags are skipped.** New `is_url_safe_merge_tag()` rejects any tag whose `%` characters don't form valid `%XX` escapes. This covers ActiveCampaign's `%FIELD%` syntax generally, not just the bad default — an unresolved tag reaches the live URL whenever the ESP has no value for the recipient (unknown field, forwarded mail, previews), so for a `%`-delimited provider a broken URL is a routine outcome rather than an edge case.
- **Tests.** Added coverage for the unconfigured-option path and for a `%`-syntax provider, and made the `Tracking\Utils` mock provider-switchable (it was Mailchimp-only, so no test ever exercised non-Mailchimp syntax).

Behavior for correctly configured Mailchimp, Constant Contact and Campaign Monitor sites is unchanged — their `*|X|*`, `[[X]]` and `[X]` tags carry no percent sign. ActiveCampaign sites lose donor segmentation on newsletter clicks; that tradeoff is called out below.

Closes NPPM-3032.

> **This only stops future sends.** Newsletters already in inboxes keep the bad links, so the Jetpack Search mitigation (theme search mode, or disable Jetpack Search + purge edge cache) and the upstream fix (Automattic/jetpack#50709) still matter.

**Regression — unconfigured site no longer decorates links (the bug):**

1. On a site with an ESP connected, ensure the donor merge field is unset: `wp option delete newspack_popups_mc_donor_merge_field`.
2. Create a newsletter with a link to a post on the same site and preview/render it.
3. The link should carry **no** `np_seg_donor` param. On `release` it comes out as `?np_seg_donor=*|DONAT|*` (Mailchimp) or `?np_seg_donor=%DONAT%` (ActiveCampaign).

**Regression — configured Mailchimp site still works:**

4. Set a real merge field: `wp option update newspack_popups_mc_donor_merge_field HUB-MEMBER`, with Mailchimp as the provider.
5. Render the newsletter again — the link should carry `?np_seg_donor=*|HUB-MEMBER|*`, raw and unencoded, exactly as before this PR.

**New guard — ActiveCampaign is skipped:**

6. Switch the provider to ActiveCampaign, keep the field set, render again.
7. The link should carry **no** `np_seg_donor` param (rather than `%HUB-MEMBER%`).

**Automated:**

8. `cd plugins/newspack-popups && n test-php --filter SegmentationNewsletterLinkTest` — 9 tests pass. The two new tests fail against `release`, reproducing `*|DONAT|*` and `%HUB-MEMBER%` respectively.

---

## For AI

**Root cause chain.** `class-newspack-popups-segmentation.php:243` read `get_option( 'newspack_popups_mc_donor_merge_field', Newspack_Popups_Settings::DEFAULT_DONOR_MERGE_FIELD )`. `get_option()`'s third argument applies whenever the row is absent, so unconfigured sites got `'DONAT'`. That value flowed into `\Newspack_Newsletters\Tracking\Utils::get_merge_tag()` (`plugins/newspack-newsletters/includes/tracking/class-utils.php:48`), a `match` on `$provider->service` that returns `*|X|*` / `%X%` / `[[X]]` / `[X]` and `''` only for an absent or unrecognized provider — i.e. it gates on provider *syntax*, never on whether the field makes sense for that provider.

Line 257-263 then deliberately reverses `add_query_arg()`'s encoding (`str_replace( urlencode( $merge_tag ), $merge_tag, $url )`) so the ESP can substitute the literal tag. That un-encoding is load-bearing for the feature and is also what lets a malformed percent-sequence reach the browser.

**Two independent defects, both fixed here.** (1) The default value is semantically the wrong kind of thing — `DEFAULT_DONOR_MERGE_FIELD` is consumed by `reader_logged_in()` at line ~354 via `strpos()`, gated behind `newspack_mailchimp_api_key`, and is designed to match field *names* like `DONATION`. Reusing it as an exact tag is wrong on Mailchimp too; it just fails invisibly there because Mailchimp drops unknown merge tags, which is why only ActiveCampaign sites surfaced it. (2) Even a correctly configured field is unsafe under a `%`-delimited provider, because non-substitution is normal.

**Guard implementation.** `is_url_safe_merge_tag()` short-circuits on no `%`, then applies `/^(?:[^%]|%[0-9A-Fa-f]{2})*$/`. Note this rejects ActiveCampaign universally, not just for hex-invalid field names: `%AB%` parses as a valid `%AB` escape followed by a trailing bare `%`, which still fails. That is intentional — there is no field name for which AC's syntax is URL-safe.

**Test mock change.** `tests/mocks/class-utils.php` gained a `public static $syntax` sprintf pattern defaulting to `'*|%s|*'`, and `set_up()` resets it per test. The mock is shared, so the reset matters: without it a provider-switching test would leak syntax into subsequent tests.

**Verification.** Full `newspack-popups` suite green in-container against this worktree: `OK (224 tests, 584 assertions)`. Both new tests confirmed failing against `origin/release` with the exact production strings. PHPCS clean on all three changed files against the root `phpcs.xml`.

**Not addressed here.** The settings UI still advertises `DEFAULT_DONOR_MERGE_FIELD` as the field's `default` (`class-newspack-popups-settings.php:292`), so saving the settings screen can still persist `DONAT` into the option — which is likely why several ActiveCampaign sites were found with it populated despite never intending to set it. Those sites are now protected by the percent guard rather than by the default removal. Reworking that setting (it conflates a Mailchimp substring list with an exact ESP tag and should arguably be two settings) is a follow-up, not a hotfix.

### Other information:

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
@@ -296,16 +296,20 @@ final class Newspack_Popups_Segmentation {
 	 * Skipping the param costs that provider donor segmentation on newsletter
 	 * clicks; emitting it risks breaking the page the reader landed on.
 	 *
+	 * Any `%` is rejected, rather than only malformed `%XX` shapes: a hex-shaped
+	 * escape can still be invalid UTF-8 — `decodeURIComponent( '%FF' )` and
+	 * `decodeURIComponent( '%C0%AF' )` both throw — so shape alone is not enough,
+	 * and validating decoded UTF-8 here would be a lot of machinery to license a
+	 * character no supported ESP needs. ActiveCampaign's `%FIELD%` delimiters are
+	 * rejected wholesale regardless, and a `%` inside a Mailchimp / Constant
+	 * Contact / Campaign Monitor field name is not a real merge field.
+	 *
 	 * @param string $merge_tag Raw ESP merge tag, e.g. '*|DONOR|*' or '%DONOR%'.
 	 *
 	 * @return bool
 	 */
 	private static function is_url_safe_merge_tag( $merge_tag ) {
-		if ( false === strpos( $merge_tag, '%' ) ) {
-			return true;
-		}
-		// Every '%' must introduce a well-formed two-hex-digit escape, or decoding fails.
-		return 1 === preg_match( '/^(?:[^%]|%[0-9A-Fa-f]{2})*$/', $merge_tag );
+		return false === strpos( (string) $merge_tag, '%' );
 	}
 
 	/**
--- a/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
@@ -373,18 +373,56 @@ final class Newspack_Popups_Segmentation {
 		if ( ! isset( $_GET[ self::DONOR_SEGMENT_QUERY_PARAM ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			return;
 		}
-		$value = sanitize_text_field( wp_unslash( $_GET[ self::DONOR_SEGMENT_QUERY_PARAM ] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( ! self::is_unsubstituted_merge_tag( $value ) ) {
+		$request_uri = esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) );
+		// Match against BOTH the raw query value and the percent-decoded one. PHP
+		// decodes $_GET before this runs, so a tag whose field name begins with a hex
+		// pair — `%CAFE%`, `%ABCD%` — arrives already mangled (`%CA` becomes one byte)
+		// and no longer looks like a `%…%` tag, yet the raw URL still throws in a
+		// strict client-side decoder (decodeURIComponent) and blanks the page. The raw
+		// REQUEST_URI query preserves the literal `%XX` so the check can catch it. The
+		// decoded value is still checked too, so a merge tag whose delimiters were
+		// percent-encoded in transit (e.g. `%2A%7C…%7C%2A` for Mailchimp) is not
+		// missed. See NPPM-3032.
+		$raw_value     = self::get_raw_query_param( $request_uri, self::DONOR_SEGMENT_QUERY_PARAM );
+		$decoded_value = sanitize_text_field( wp_unslash( $_GET[ self::DONOR_SEGMENT_QUERY_PARAM ] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( ! self::is_unsubstituted_merge_tag( $raw_value ) && ! self::is_unsubstituted_merge_tag( $decoded_value ) ) {
 			return;
 		}
 		$clean_url = remove_query_arg(
 			self::DONOR_SEGMENT_QUERY_PARAM,
-			esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) )
+			$request_uri
 		);
 		// Temporary: the param is a property of this one link, not of the page, so
 		// nothing should cache the mapping permanently.
 		wp_safe_redirect( $clean_url, 302 );
 		exit;
+	}
+
+	/**
+	 * Read a single query parameter's value from a URL *without* percent-decoding
+	 * it, so a malformed escape (e.g. `%CAFE%`) survives intact for inspection.
+	 *
+	 * `$_GET`, parse_str() and wp_parse_args() all percent-decode, which is exactly
+	 * what must be avoided when deciding whether a value is a raw merge tag that
+	 * would break client-side decoding — see scrub_unsubstituted_donor_param().
+	 *
+	 * @param string $url   URL or path carrying a query string.
+	 * @param string $param Parameter name to read.
+	 *
+	 * @return string Raw (still-encoded) value, or '' when the param is absent.
+	 */
+	private static function get_raw_query_param( $url, $param ) {
+		$query = (string) wp_parse_url( $url, PHP_URL_QUERY );
+		if ( '' === $query ) {
+			return '';
+		}
+		foreach ( explode( '&', $query ) as $pair ) {
+			$parts = explode( '=', $pair, 2 );
+			if ( $param === $parts[0] ) {
+				return isset( $parts[1] ) ? $parts[1] : '';
+			}
+		}
+		return '';
 	}
 
 	/**
--- a/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
@@ -214,7 +214,8 @@ final class Newspack_Popups_Segmentation {
 	 * inbound click carries e.g. `?np_seg_donor=true`. Skips when the Newsletters
 	 * tracking helper is unavailable, the post isn't a newsletter (ad links are
 	 * proxied separately and wouldn't forward the param), the link is
-	 * third-party, no donor merge field is configured, or the ESP is unsupported.
+	 * third-party, no donor merge field is configured, the ESP is unsupported, or
+	 * the ESP's tag syntax can't survive in a URL (see is_url_safe_merge_tag()).
 	 *
 	 * @param string        $url          Processed URL (may already carry other params).
 	 * @param string        $original_url Original URL before processing.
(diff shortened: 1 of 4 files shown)


plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
```diff
@@ -263,6 +272,35 @@ public static function append_donor_segment_param( $url, $original_url, $post )
...
+	 *
+	 * @return bool
+	 */
+	private static function is_url_safe_merge_tag( $merge_tag ) {
+		if ( false === strpos( $merge_tag, '%' ) ) {
+			return true;
+		}
+		// Every '%' must introduce a well-formed two-hex-digit escape, or decoding fails.
+		return 1 === preg_match( '/^(?:[^%]|%[0-9A-Fa-f]{2})*$/', $merge_tag );
+	}
```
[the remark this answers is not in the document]
chickenn00dle: Good catch — confirmed real. A hex-shaped escape can still be undecodable: `decodeURIComponent('%FF')` throws because 0xFF is never valid UTF-8, and `%C0%AF` is an overlong encoding. Both passed the old shape check, so a Mailchimp field name containing one would have emitted a tag that breaks client-side decoding exactly like `%DONAT%` did.

Went with your second suggestion — treat any `%` as unsafe — rather than validating decoded UTF-8. It's strictly safer and costs nothing real: ActiveCampaign's `%FIELD%` delimiters are rejected wholesale anyway, and a `%` inside a Mailchimp / Constant Contact / Campaign Monitor field name isn't a real merge field. UTF-8 validation would be a lot of machinery to license a character no supported ESP needs.

Verified the change actually flips these cases:

| tag | before | after |
| --- | --- | --- |
| `*\|%FF\|*` | emit | skip |
| `*\|%C0%AF\|*` | emit | skip |
| `*\|%20\|*` | emit | skip |
| `*\|HUB-MEMBER\|*` | emit | emit |
| `[[HUB-MEMBER]]` | emit | emit |

Added a data-provider test covering the three undecodable cases. Addressed in b18e3b491.

plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
```diff
@@ -263,6 +279,114 @@ public static function append_donor_segment_param( $url, $original_url, $post )
...
+		}
+		if ( ! isset( $_SERVER['REQUEST_URI'] ) ) {
+			return;
+		}
+		// Reading a URL param to decide whether the URL itself is malformed; there
+		// is no form submission or state change here to nonce-verify.
+		if ( ! isset( $_GET[ self::DONOR_SEGMENT_QUERY_PARAM ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return;
+		}
+		$value = sanitize_text_field( wp_unslash( $_GET[ self::DONOR_SEGMENT_QUERY_PARAM ] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
dkoo: **Severity:** blocker

The scrub reads `$_GET`, which PHP has already percent-decoded, but `is_unsubstituted_merge_tag()` matches the raw `%FIELD%` shape. When the field name starts with two hex-like characters, PHP decodes the leading `%XX` before this runs (`%ABCD%` → bytes `AB 'C' 'D' '%'`), so the value no longer matches `/^%[^%]+%$/` and the redirect is skipped — even though `decodeURIComponent()` still throws on the original URL, so those already-sent links keep blanking the page under Jetpack Search. Confirmed against a running site (and reproduced by hands-on QA): `?np_seg_donor=%DONAT%` 302s (the incident value — `DO` isn't a valid escape, so PHP keeps it literal), but `%CAFE%`, `%ABCD%`, `%12AB%`, and `%DEAD%` all return 200 unscrubbed while every one throws in `decodeURIComponent`. Detect and strip against the raw query string from `REQUEST_URI` (e.g. `wp_parse_url( $request_uri, PHP_URL_QUERY )` then a raw match) rather than the decoded `$_GET`, so the literal `%XX` survives to the check. (An array-form value like `?np_seg_donor[]=%DONAT%` slips through the same way — `sanitize_text_field()` returns `''` for an array — though that one only arises from a hand-crafted URL.) The emit-side guard already blocks all of these on new sends; this is scoped to the already-sent backfill — but that backfill is exactly what this scrub exists to handle, so it's worth closing before this ships as the incident remedy.

plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
```diff
@@ -86,6 +86,13 @@ public function __construct() {
 		// The handler self-guards on the donor merge field being configured and
 		// the ESP being supported, so it's cheap to register unconditionally.
 		add_filter( 'newspack_newsletters_process_link', [ __CLASS__, 'append_donor_segment_param' ], 30, 3 );
+
+		// Strip unsubstituted donor merge tags from inbound URLs. Newsletters
+		// already delivered carry tags this plugin can no longer stop emitting, and
+		// an unresolved `%FIELD%` is a malformed percent-escape that crashes
+		// consumers which decode query params strictly — see NPPM-3032. Priority 1
+		// so it runs before redirect_canonical() and before any HTML is generated.
+		add_action( 'template_redirect', [ __CLASS__, 'scrub_unsubstituted_donor_param' ], 1 );
```
dkoo, line 95: **Severity:** suggestion

On a full-page / edge cache HIT (batcache), the cached 200 is served before WordPress boots, so `template_redirect` — and this scrub — never fires, and the crashing URL still reaches the client. The comment's "before any HTML is generated … no such consumer ever sees the param" holds only on a cache MISS. This is inherent (no server handler can intercept a fully-cached response) and self-healing post-deploy (a MISS caches the 302), so the emit-side change is the real guarantee and the scrub is best-effort for in-flight emails — worth softening the comment to say so, and keeping the "purge the edge cache on deploy" note prominent so pre-fix cached 200s are evicted immediately rather than waiting out the TTL.

plugins/newspack-popups/tests/test-segmentation-newsletter-link.php
```diff
@@ -80,6 +82,212 @@ public function test_skips_when_no_donor_field_configured() {
...
+	/**
+	 * Values an ESP actually substitutes, positive and negative.
+	 *
+	 * @return array[]
+	 */
+	public function substituted_value_provider() {
+		return [
+			'true'    => [ 'true' ],
+			'monthly' => [ 'monthly' ],
+			'amount'  => [ '50.00' ],
```
dkoo, line 293: **Severity:** nit

The `'amount'` case uses `50.00` while the surrounding docblocks reference `$50.00`; a `$`-prefixed value is never actually exercised (harmless — `$` matches no unsubstituted pattern), just a gap between the example and the tested value.

dkoo (changes requested):
This is a well-constructed, unusually well-documented hotfix, and the urgent half — stopping *new* bad emissions — is solid and verified: the opt-in `''` default plus the reject-any-`%` guard mean no site emits a `%`-tag going forward (confirmed against a running site: ActiveCampaign + `DONAT` now emits nothing, Mailchimp/Constant Contact still work, unset field emits nothing). The PHP `is_unsubstituted_merge_tag()` is a character-for-character mirror of `isUnsubstitutedMergeTag()` in `donation.js`, and the 24 new tests genuinely exercise the claimed behavior.

Requesting changes on one confirmed gap in the inbound scrub, which is the PR's remedy for newsletters already in inboxes:

**The scrub reads the percent-decoded `$_GET`, so it silently misses a class of the links it targets.** Already-sent tags whose field name begins with a hex pair (`%CAFE%`, `%ABCD%`, `%12AB%`, …) are decoded by PHP before the check, so they no longer match the `%…%` shape and are not stripped — even though they still throw in `decodeURIComponent` and blank the page. Hands-on QA against a running site reproduced `%CAFE%` slipping through unredirected while `%DONAT%` (the reported value) is correctly caught. Detecting/stripping against the raw `REQUEST_URI` query — where the literal `%XX` survives — closes it. Details inline. (The emit-side guard already blocks these on new sends, so no new sends are affected; the gap is specific to the already-sent backfill this scrub exists to handle.)

Two more, non-blocking:

- **The Campaigns settings screen still defaults `newspack_popups_mc_donor_merge_field` to `DEFAULT_DONOR_MERGE_FIELD` (`DONAT`)** at `plugins/newspack-popups/includes/class-newspack-popups-settings.php:281`, so saving that screen re-persists `DONAT` and re-enables emission (a url-safe `*|DONAT|*` on Mailchimp — scrubbed inbound and ignored client-side, so not a page-break). This keeps the inbound scrub load-bearing even on fully-updated sites. The PR's "Not addressed here" section already flags the settings rework as a follow-up — worth confirming that's the intended posture.
- **The full-page-cache HIT gap** (a cached-200 bad URL bypasses the scrub) is inherent and already acknowledged in the PR's cache caveat — just keep the edge-cache purge prominent in the deploy steps.

If shipping speed matters, an option is to land the emit-side fix now (it's what stops the bleeding) and complete the scrub as an immediate fast-follow — but the fix here is small enough to just fold in.

plugins/newspack-popups/includes/class-newspack-popups-segmentation.php
dkoo: Pushed a fix for this in 78839cb.

dkoo (approved):
Pre-approving pending a review of my fix 78839cb, which addresses the one blocker. We can defer/hold the non-blocking threads.

PR #698, opened 2026-07-22
fix(reader-activation): legacy-mode outbound field filtering (NPPD-2107)

miguelpeixe:
### All Submissions:

Fixes a data-exposure bug in the RAS Integrations framework: in legacy metadata mode (the default on every site without the `NEWSPACK_SYNC_METADATA_VERSION` constant), `Integration::prepare_contact()` returned the contact **unchanged for every integration**, so a non-ESP integration's Outbound field selection was never applied. An integration whose Outbound panel showed nothing selected — e.g. a freshly connected ActiveCampaign integration — received and pushed the **full default metadata set** (Membership Status, Total Paid, Recurring Payment, …), governed by the ESP integration's config rather than its own. A bulk `wp newspack integrations backfill` run turns that into a site-wide push of financial metadata the operator never selected.

- The unconditional legacy passthrough is now reserved for the `esp` integration, whose own config is what the legacy metadata pipeline already filtered by.
- Every other integration applies its enabled Outbound selection to legacy-pipeline data via a new `prepare_contact_legacy()`: keys are de-prefixed with the legacy pipeline's prefix (`Metadata::get_prefix()` — the one the data actually carries, not the integration's own, which may differ) and matched against enabled labels, with exact matches plus the `Label: ` prefix shape for UTM sub-keys (enabled `Signup UTM: ` matches `Signup UTM: source`).
- **No-regression inheritance:** in legacy mode, an integration that has *never saved* an Outbound selection inherits the ESP integration's effective selection — the set the legacy pipeline complies with — in the engine **and** in the Outbound UI (both read `get_enabled_outgoing_fields()`). Pre-existing legacy sites keep syncing exactly what they sync today, and the panel now displays the truth instead of empty checkboxes.
- An **explicitly saved** selection always wins — including saving with everything unchecked, which now genuinely means "push no metadata fields". Unprefixed sync-control keys (`status`, `status_if_new`) always pass through; v1 metadata mode is untouched (absent still means empty there).

- **Review follow-ups (second commit):** on an ESP-registry miss (pre-`init` or a directly constructed integration) the inheritance getter mirrors the ESP integration's registry-free fallback chain (legacy global option → default set) instead of failing closed to an empty selection; inherited selections are `array_values()`-normalized; `prepare_contact_legacy()` uses native types; the integrations README now documents the legacy-mode contract (per-integration narrowing, inheritance, explicit-empty, and its caveats); unprefixed passthrough is restricted to `Legacy_Metadata::SYNC_CONTROL_KEYS` (`status`, `status_if_new`) per Copilot review, with the shared constant replacing the literal in `normalize_contact_data()`.

- **Review follow-ups (third commit):** prefix matching is now scoped to the legacy pipeline's own UTM keys (`Legacy_Metadata::UTM_RAW_KEYS`), so a label registered through `newspack_ras_metadata_keys` that ends in `': '` can no longer carry a different field past the selection; the match set is built from `Metadata::get_key()` as well as `prefix . label`, so keys reshaped by the `newspack_ras_metadata_key` filter survive instead of being dropped; the registry-miss fallback chain moved into `Metadata::get_fields()` so the ESP's effective legacy selection has one definition, with a new overridable `get_inherited_legacy_outgoing_fields()` for integrations that want a different inherited set or none; the `@deprecated` block on `Metadata::get_prefix()` records legacy-pipeline de-prefixing as a sanctioned caller. New tests pin the custom-integration-prefix case, the registered-label over-match, renamed keys under inheritance, and the v1-mode empty fallthrough.

**What this changes, and what it does not.** Per-integration Outbound selection now takes effect in legacy mode, where it previously did nothing. Untouched legacy sites keep their current payloads: an integration with no saved selection inherits the ESP integration's effective selection.

That inheritance preserves the default posture rather than tightening it. On a site where nobody ever saved an ESP selection, the ESP's effective selection is `Metadata::get_default_fields()` — every available field, including Membership Status, Total Paid and Recurring Payment — so a newly connected integration inherits all of them, and a backfill sends all of them. This PR gives operators a control that works and a panel that shows the truth; closing the exposure on those sites takes an explicit save.

**Before deploying:** list the legacy sites holding a `newspack_integration_outgoing_fields_<id>` option row for a non-ESP integration. That is exactly the set whose outbound payload changes on the first push after deploy, and it scopes the per-site check NPPD-2107 asks for. Every other legacy site inherits and is unchanged.

**For the release note:** narrowing a selection stops refreshing those fields at the destination; it does not clear the values already there. ActiveCampaign skips the field write when the prepared payload has no mapped fields, and Salesforce upserts only the keys present, so previously synced values remain and go stale. Restoring the selection and re-running a backfill brings them back.

Closes NPPD-2107.

1. From `plugins/newspack-plugin`, run `n test-php --group Integration_Outbound_Filter` — 9 tests covering explicit-empty, ESP-selection inheritance, label + UTM-shape matching, esp passthrough, no-metadata, corrupt-option fallback, ESP-registry-miss fallback, and unknown-unprefixed-key dropping cases should pass.
2. Run the broader regression net: `n test-php --group integrations,Contact_Sync_Options,Content_Gate_Legacy,Donation_Metadata,Grouped_Metadata,Subscription_Metadata,account-deletion,esp_integration,prepare_contact,promoted_fields,reader-activation,Integration_Outbound_Filter` — 179 tests should pass.
3. On a site in legacy metadata mode (no `NEWSPACK_SYNC_METADATA_VERSION` constant) with a non-ESP integration active and configured (e.g. newspack-manager's ActiveCampaign) that has **never saved** an Outbound selection (Audience → Integrations, `NEWSPACK_INTEGRATIONS_SETTINGS_ENABLED`): confirm the Outbound panel now shows the ESP integration's effective selection checked, and a push (`wp newspack integrations backfill --direction=push --integration=<id> --user-ids=<id>`) carries exactly those `NP_` fields — same payload as before this PR.
4. Uncheck everything in that integration's Outbound panel and **save**, then re-push: the pushed contact must carry no `NP_*` metadata fields — only core contact data and sync-control keys like `status_if_new`.
5. Save a single Outbound field (e.g. Membership Status) and re-push: only that `NP_` field is included.
6. Save a UTM field (`Signup UTM: `) and re-push a reader with UTM data: the `NP_Signup UTM: source`-style sub-keys are included.
7. Confirm the built-in `esp` integration's pushes are unchanged before/after this PR (its legacy passthrough is preserved).

### Other information:


plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -977,6 +981,56 @@ public function prepare_contact( $contact ) {
...
+
+		$enabled_fields = $this->get_enabled_outgoing_fields();
+		$prefix         = Sync\Metadata::get_prefix();
+		$prepared       = [];
+
+		foreach ( $contact['metadata'] as $key => $value ) {
+			if ( 0 !== strpos( $key, $prefix ) ) {
+				$prepared[ $key ] = $value;
+				continue;
+			}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in d169a6eb9. Unprefixed passthrough is now restricted to the sync-control allowlist: the existing literal in `Legacy_Metadata::normalize_contact_data()` is promoted to `Legacy_Metadata::SYNC_CONTROL_KEYS` and `prepare_contact_legacy()` checks it — any other unprefixed key is dropped, with a pinning test.

One knock-on: the flag-mode deletion test asserted the raw `account_deleted` key in the pushed payload, which was a legacy-passthrough artifact (v1 mode never pushed it). The dispatcher's prefixed `Account_Deleted` re-injection is the durable ESP-side signal, and the test now pins that contract instead.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2107-legacy-mode-outbound-field-filtering` (stash local work first).

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -818,10 +818,46 @@ protected function configure_incoming_field( $field ) {
...
 	 * @return string[] List of enabled field names.
 	 */
 	public function get_enabled_outgoing_fields() {
-		return array_values( \get_option( self::OUTGOING_FIELDS_OPTION_PREFIX . $this->id, [] ) );
+		$stored = \get_option( self::OUTGOING_FIELDS_OPTION_PREFIX . $this->id, null );
+		if ( null !== $stored && is_array( $stored ) ) {
+			return array_values( $stored );
+		}
+
+		if ( 'legacy' === Sync\Metadata::get_version() && 'esp' !== $this->get_id() ) {
```
chickenn00dle, line 845: **Suggestion:** On a site that never saved an ESP selection, the full financial set still goes out — This makes the per-integration control *effective*, which is the fix — but it does not change the default posture, and the PR body reads as though it does.

On a legacy site where nobody ever explicitly saved a metadata selection, `Esp::get_enabled_outgoing_fields()` falls through to `Metadata::get_default_fields()` (`class-esp.php:366`) — all 21 fields, including Membership Status, Total Paid and Recurring Payment. A freshly connected non-ESP integration inherits exactly that. I confirmed it on a fresh legacy site: ESP option row `NULL`, legacy `FIELDS_OPTION` `NULL`, inherited count 21, and a push carrying every one of them. So `wp newspack integrations backfill` still sends the whole financial metadata set the operator never selected.

That is the deliberate no-regression choice and I agree with it — dropping those sites to zero metadata overnight would be worse. But it means the ticket's headline harm scenario is only *closed* on sites that have an explicit ESP selection; elsewhere the operator now has a working control to close it themselves. Worth saying plainly in the PR body and the release note, so "data-exposure bug fixed" doesn't lead someone to skip the per-site check the ticket asked for.

**Risk if not addressed:** If a publisher's ActiveCampaign automation or Salesforce report keys on a field that drops out of the payload, the previously synced value stays in place and goes stale rather than clearing — ActiveCampaign skips the field write entirely when the prepared list is empty, and Salesforce upserts only the keys present. The automation keeps evaluating an old value and nothing looks broken until someone compares the destination against WordPress. No value is erased, and it is reversible by restoring the selection and re-running a backfill.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -818,10 +818,46 @@ protected function configure_incoming_field( $field ) {
...
+			if ( $esp ) {
+				return array_values( $esp->get_enabled_outgoing_fields() );
+			}
+
+			// Registry miss (pre-init or a directly constructed integration —
+			// integrations register on init priority 5): mirror the ESP
+			// integration's own fallback chain instead of failing closed to an
+			// empty selection, which would strip every field where the
+			// pre-selection behavior was full passthrough.
+			$legacy = \get_option( Sync\Metadata::FIELDS_OPTION, null );
```
chickenn00dle: **Suggestion:** Third copy of the ESP fallback chain, and no way to opt out of inheritance — The registry-miss branch duplicates `Esp::get_enabled_outgoing_fields()`'s fallback chain (`Metadata::FIELDS_OPTION` → `get_default_fields()`), which already exists at `class-esp.php:353-367` and in reduced form at `Metadata::get_fields()` (`class-metadata.php:179-182`). Three definitions of "the ESP's effective legacy selection" will drift, and the one most likely to be forgotten is this one, since it only fires on a path that is hard to reach.

Pushing it down into `Metadata::get_fields()` — already the designated home for ESP-delegating shims — and calling that from here would collapse it to one definition. Separately, the abstract base now hard-codes knowledge of a concrete integration id, and a third-party integration has no way to opt out of ESP inheritance short of writing its own option row; a `protected function get_inherited_legacy_outgoing_fields()` or a filter would give it one.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1116,6 +1156,62 @@ public function prepare_contact( $contact ) {
...
+		if ( 'esp' === $this->get_id() ) {
+			return $contact;
+		}
+
+		$enabled_fields = $this->get_enabled_outgoing_fields();
+		$prefix         = Sync\Metadata::get_prefix();
+		$prepared       = [];
+
+		foreach ( $contact['metadata'] as $key => $value ) {
+			if ( 0 !== strpos( $key, $prefix ) ) {
```
chickenn00dle: **Suggestion:** Renamed metadata keys are dropped silently, not passed through — The filter reconstructs the expected key shape as `Metadata::get_prefix() . <enabled label>`, but the legacy pipeline does not always emit that shape: `Metadata::get_key()` ends in `apply_filters( 'newspack_ras_metadata_key', $key, $prefix, $name )` (`class-metadata.php:281`), and `Legacy_Metadata::normalize_contact_data()` ends in `apply_filters( 'newspack_esp_sync_normalize_contact', $contact )` (`class-legacy-metadata.php:267`). Both are public, documented filters. A key renamed by either matches neither `SYNC_CONTROL_KEYS` nor `prefix . label`, so it is dropped.

I reproduced this on the running branch: with a filter renaming keys to `CUSTOM_<name>`, a non-ESP integration under **full inheritance** — the no-regression path, where the payload is supposed to be unchanged — emitted zero metadata fields, where before the change it passed all of them through. The failure is total and silent rather than partial and loud.

The risk is bounded: nothing in-tree uses the singular `newspack_ras_metadata_key`. Only the plural `newspack_ras_metadata_keys` has consumers (`newspack-network/includes/class-esp-metadata-sync.php:24`, `class-teams-for-memberships.php:29`), and those add labels rather than renaming keys, which this code handles correctly. So this is a suggestion, not a blocker — but a publisher-specific `newspack_ras_metadata_key` filter would be invisible to us and would fail this way.

Building the allowlist from `Metadata::get_key( $raw_key )` for each enabled label, rather than re-deriving `prefix . label`, would be robust to both filters.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-integration-outbound-legacy.php
```diff
@@ -0,0 +1,229 @@
...
+		$this->reset_integrations();
+		Integrations::register_integrations();
+		parent::tear_down();
+	}
+
+	/**
+	 * Reset the integrations registry via reflection, mirroring the sibling
+	 * suites so tests are order-independent.
+	 */
+	private function reset_integrations() {
```
chickenn00dle, line 49: **Suggestion:** The subtlest line in the change is unpinned by the new tests — All nine tests run with the mock integration's prefix implicitly equal to the ESP's `NP_`, which leaves `class-integration.php:1189` — the deliberate choice of `Sync\Metadata::get_prefix()` over `$this->get_metadata_prefix()`, and the one line the docblock spends a paragraph justifying — completely uncovered.

A well-meaning future refactor to "use this integration's own prefix" would keep all nine tests green while dropping 100% of metadata on any site whose non-ESP integration has a custom prefix — which is the newspack-manager ActiveCampaign scenario this PR exists for.

One case that sets the mock's prefix to `AC_`, feeds the same `NP_`-prefixed legacy payload, and asserts the selection still filters correctly would pin it. I verified that scenario by hand against the branch and it behaves correctly today; it just isn't guarded.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1116,6 +1156,62 @@ public function prepare_contact( $contact ) {
...
+	 * @param array $contact Contact data with prefixed legacy metadata.
+	 * @return array Contact data with metadata narrowed to enabled fields.
+	 */
+	private function prepare_contact_legacy( array $contact ): array {
+		if ( 'esp' === $this->get_id() ) {
+			return $contact;
+		}
+
+		$enabled_fields = $this->get_enabled_outgoing_fields();
+		$prefix         = Sync\Metadata::get_prefix();
```
chickenn00dle: **Nit:** Correct call, but the method it calls is marked @deprecated — `Metadata::get_prefix()` is the right call here and the only accessor that reproduces what the legacy pipeline actually applied, including the `newspack_ras_metadata_prefix` filter and the pre-init `PREFIX_OPTION` fallback — neither of which `Integration::get_metadata_prefix()` does. The deprecation is PHPDoc-only, so there's no `_deprecated_function()` notice.

The risk is a future cleanup pass removing the method and silently breaking this path — silently, because as finding 1 shows, a de-prefix miss here drops metadata rather than erroring. Either a non-deprecated alias (`Metadata::get_legacy_pipeline_prefix()`) or a note on the `@deprecated` block at `class-metadata.php:108` saying legacy-pipeline de-prefixing is a sanctioned remaining caller would protect it.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -1116,6 +1156,62 @@ public function prepare_contact( $contact ) {
...
+				if ( in_array( $key, Sync\Legacy_Metadata::SYNC_CONTROL_KEYS, true ) ) {
+					$prepared[ $key ] = $value;
+				}
+				continue;
+			}
+			$field_name = substr( $key, strlen( $prefix ) );
+			foreach ( $enabled_fields as $enabled_field ) {
+				if (
+					$field_name === $enabled_field ||
+					( str_ends_with( $enabled_field, ': ' ) && 0 === strpos( $field_name, $enabled_field ) )
```
chickenn00dle: **Blocker:** The `': '` label shape is now a load-bearing contract on a public filter — The prefix-match heuristic can't over-match today: the only default labels ending in `': '` are `Signup UTM: ` and `Payment UTM: `, and neither prefixes the other. But `newspack_ras_metadata_keys` lets any plugin register arbitrary labels — newspack-network already does (`class-esp-metadata-sync.php:24`) — and a registered label ending in `': '` that happens to prefix another label would carry that other field past the selection.

I confirmed the mechanism at runtime: with `Signup UTM: ` enabled, a synthetic `NP_Signup UTM: source_secret` key passed the filter (while `NP_Signup UTM Extra` correctly did not). Harmless with today's label set, but the `': '` shape is now a contract on that filter. Requiring a non-empty remainder, or restricting the prefix match to the known UTM labels, would close it; at minimum it's worth documenting alongside the filter.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-account-deletion.php
```diff
@@ -329,20 +330,25 @@ public function test_handle_account_deletion_calls_push_with_timestamp_when_hand
 
 		$pushed = $spy->push_calls[0]['contact'];
 		$this->assertSame( 'reader@example.com', $pushed['email'] );
-		$this->assertArrayHasKey( 'account_deleted', $pushed['metadata'] );
+		// The raw `account_deleted` flag is dropped by prepare_contact() like any
```
chickenn00dle, line 341: **Nit:** This test now pins a combination that can't occur in production — The updated assertions are correct for what the test exercises, but the test calls `handle_account_deletion()` directly while running at the default metadata version, which is `legacy`. In production that pairing doesn't exist: `Contact_Sync_Connector::register_handlers()` registers the older `reader_deleted` handler in legacy mode and only wires `reader_delete_sync` → `handle_account_deletion()` for v1 (`class-data-events/connectors/class-contact-sync-connector.php:41-45`), which the integrations README states explicitly at line 222.

So the diff here reads as "legacy deletion payloads changed" when in fact legacy deletion never reaches this dispatcher, and v1's behavior is untouched. I checked this specifically because it looked at first like a second, unannounced behavior change. A one-line comment saying the test drives the v1 dispatcher directly — or setting the version explicitly, as the sibling test at `:902` does — would stop the next reader drawing the same wrong conclusion.

chickenn00dle (approved):
Release risk: ELEVATED — a behavior change on a contract that `newspack-manager`'s ActiveCampaign and Salesforce integrations consume directly, taking effect at deploy time rather than on the next admin save; and one path that silently drops all metadata rather than failing loudly. No blockers: the core logic is right, the inheritance genuinely preserves existing payloads (I confirmed byte-identical output on a fresh legacy site), and the destination-side blast radius is bounded — narrowing a selection stops refreshing fields at the ESP, it does not blank them. Before merge I would want two things: the pre-deploy check the ticket asked for, scoped to sites that actually hold a saved selection row for a non-ESP integration, and a decision on the key-rename drop below. Everything else here is a suggestion.

Before deploying, list the legacy sites holding a `newspack_integration_outgoing_fields_<id>` option row for a non-ESP integration. That set is exactly the set whose outbound payload changes on the first push after deploy; every other legacy site inherits and is unchanged. That also scopes the "confirm no live site depends on the current behavior" check the ticket asks for.

Worth putting in the release note: narrowing a selection does not clear the corresponding values at the destination. ActiveCampaign skips the field write when the prepared payload has no mapped fields, and Salesforce upserts only the keys present, so previously synced values remain and go stale rather than being blanked. That bounds this to "stops updating" rather than "deletes", and makes it reversible.

I'm going to preemptively approve since most of this feedback is non-blocking and the one blocking feedback is relatively small.

- `plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php:285` — **Suggestion:** Sibling path de-prefixes with a different prefix than this one — `prepare_contact_legacy()` gets this right and the comment explains why: it de-prefixes with `Sync\Metadata::get_prefix()` — the prefix the legacy data actually carries — rather than the integration's own, which may differ. `Contact_Sync::prepare_contact_for_integration()` then de-prefixes the same keys with `$integration->get_metadata_prefix()` for the `--fields` scoping. These agree only because `get_metadata_prefix()` lazily inherits the legacy global. Every push-capable integration exposes an editable "Metadata field prefix" setting (`class-integration.php:1051`), so divergence is publisher-reachable. I confirmed at runtime that legacy data keeps the `NP_` prefix even when the integration's own prefix is set to `AC_`; on such a site, `backfill --fields=...` would leave `$remainder` as `NP_Membership Status`, match no label, and push empty metadata after a pre-flight that said the fields were enabled — with the dry-run misreporting it as "this reader likely has no values" (`:1491`). Pre-existing (NPPD-2076), not introduced here. Raising it because this PR is the first code to establish which prefix legacy data actually carries, so it is the natural place to align the two — e.g. a shared `get_effective_metadata_prefix()` that returns the legacy pipeline prefix in legacy mode.
- `plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php:1421` — **Suggestion:** A plain --dry-run can't preview the change this PR makes — The per-integration payload preview is gated on the run having non-default options, so the field-by-field list and the "no metadata to sync" warning only appear when `--fields` or `--skip-lists` is also passed. `wp newspack integrations backfill --dry-run` — the obvious command to reach for before rolling this out — prints nothing about which fields an integration will actually send. Combined with the success tally counting anything that is not a `WP_Error` as processed (`class-ras-contact-sync.php:72`), and ActiveCampaign returning `true` even when it wrote no custom-field values at all, a run that narrows further than intended reports "Synced N contacts (0 errors)" and looks identical to a correct one. Out of scope for this PR strictly speaking, but this PR is what makes that preview load-bearing. Printing the per-integration payload for a plain `--dry-run` too would make the pre-deploy check the ticket asks for actually runnable. **Risk if not addressed:** If the narrowing is wider than intended on a given site, a whole-site backfill runs to completion against every reader, spends the full ActiveCampaign or Salesforce request budget, writes no custom-field values for the affected integration, and still reports every contact as processed with zero errors — so the operator records the backfill as done. Nothing at the destination is destroyed and the run can be repeated, but the elapsed time and API quota are spent and the false "done" persists until someone spot-checks a contact.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
chickenn00dle: I just want to note that Claude originally marked this as a nit, but I am able to reliably reproduce on a clean test site with empty Signup/Payment UTM suffixes so bumped this up to blocker:


https://github.com/user-attachments/assets/b76c117a-6c99-4134-bbd2-ef6d595ef081

miguelpeixe: Fixed in c0b80d5. Prefix matching is now restricted to the legacy pipeline's own UTM keys, promoted to `Legacy_Metadata::UTM_RAW_KEYS` (the literal `get_utm_key()` already carried), and the matcher builds its UTM prefix set only from labels resolved through those raw keys. A label registered via `newspack_ras_metadata_keys` that ends in `': '` is now matched exactly, so it can't carry a different field through. The prefix branch also requires a non-empty remainder, so a bare re-match falls to the exact-key case.

`test_registered_label_ending_in_colon_space_does_not_prefix_match` pins it with your mechanism — sibling labels `Partner: ` and `Partner: Secret`, only the first enabled — and fails on the previous logic. `test_utm_label_still_matches_its_suffixed_sub_keys` pins the behavior we do want, so the two can't drift apart.

Thanks for taking this to a real site and re-rating it — the sibling-label case is not something the default label set can show.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Fixed in c0b80d5. The match set is now built from whole keys rather than a re-derived `prefix . label`: each enabled label contributes both the key `Metadata::get_key( $raw_key )` produces — which runs the `newspack_ras_metadata_key` filter — and the plain `prefix . label` shape. The union means a renamed key matches, and a label that is no longer in the key map still matches as it did before, so nothing narrows.

`test_renamed_keys_survive_under_inheritance` reproduces your setup (filter renaming to `CUSTOM_<name>`, integration under full inheritance) and asserts the payload passes through unchanged; it emits zero fields on the previous logic. UTM sub-keys are covered too, since their prefix now comes from `Metadata::get_key()` as well.

`newspack_esp_sync_normalize_contact` can still reshape keys after this point, and that one stays out of reach — a key invented there belongs to no field. Noted in the README caveats.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Fixed in c0b80d5, both halves.

The fallback chain moved into `Metadata::get_fields()`, which was already the ESP-delegating shim — it now carries the `FIELDS_OPTION` → `get_default_fields()` ladder for the registry-miss case, so "the ESP's effective legacy selection" has one definition. `Esp::get_enabled_outgoing_fields()` keeps its lazy-migration write; that stays integration-side. The behavior change at the other caller (`Reader_Activation::get_settings()`) is confined to the same unreachable path — the registry is populated by the time settings render.

For the opt-out: `Integration::get_enabled_outgoing_fields()` now delegates to a `protected function get_inherited_legacy_outgoing_fields()`, so a third-party integration can override the inherited set or return `[]` to decline inheritance. The concrete `'esp'` id check stays in the base — the `esp` integration is the one whose config the legacy pipeline filters by, so that asymmetry is real rather than incidental — but it no longer decides *what* gets inherited.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-integration-outbound-legacy.php
miguelpeixe: Added in c0b80d5. `test_custom_integration_prefix_does_not_affect_legacy_matching` sets the mock's prefix to `AC_`, asserts the pipeline prefix is still `NP_`, feeds the same `NP_`-prefixed payload and asserts the selection filters correctly — so a switch to the per-integration accessor now fails a test instead of passing nine.

Worth noting the restructure moved that line: matching no longer de-prefixes each key, it builds the expected whole keys up front from the pipeline prefix. The trap you identified is the same one, so the test is still the guard.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Documented in c0b80d5. The `@deprecated` block at `class-metadata.php:108` now records legacy-pipeline de-prefixing as a sanctioned remaining caller, names what only this accessor resolves (the `newspack_ras_metadata_prefix` filter and the pre-init `PREFIX_OPTION` fallback), and says a removal needs that caller migrated first — with the reason it matters, that a miss here drops metadata rather than erroring.

I went with the note over an alias so there's one prefix accessor rather than two names for it; the note travels with the thing a cleanup pass would be reading.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-account-deletion.php
miguelpeixe: Added in c0b80d5. The test docblock now says it drives `handle_account_deletion()` directly, which is the v1 path in production — `register_handlers()` wires `reader_delete_sync` to this dispatcher only outside legacy mode and registers `reader_deleted` in legacy mode — so the prefixed-key assertions are about the dispatcher's contract, not a change to legacy deletion payloads.

Left the version as-is rather than setting it explicitly: the sibling at `:902` pins v1 by reflection, and having this one keep the default while the comment explains the pairing keeps the two tests exercising different things.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Agreed, and the PR body was overclaiming. Rewritten in c0b80d5 — the summary now separates what changes from what does not, and states plainly that on a site with no saved ESP selection the effective selection is `Metadata::get_default_fields()`, so a newly connected integration inherits every field and a backfill sends every field. The framing is that this PR gives operators a control that works and a panel that shows the truth; closing the exposure on those sites takes an explicit save.

The README's Push section carries the same note, so it isn't only in the PR description.

Your pre-deploy scoping is in the body as a step: list the legacy sites holding a `newspack_integration_outgoing_fields_<id>` row for a non-ESP integration, since that is exactly the set whose payload changes on the first push. The stale-not-cleared behavior is written up for the release note, with the ActiveCampaign and Salesforce specifics and the note that restoring the selection and re-running a backfill recovers it.

miguelpeixe:
Addressed all seven threads in c0b80d5 — ready for another look. Thanks for taking this to a running site; the sibling-label case and the rename repro are both things the default label set can't surface.

On the two things you wanted before merge:

**The pre-deploy check.** Your scoping is now a step in the PR body: list the legacy sites holding a `newspack_integration_outgoing_fields_<id>` option row for a non-ESP integration. That is the set whose outbound payload changes on the first push after deploy; every other legacy site inherits and is unchanged.

**The key-rename drop.** Fixed rather than deferred. The match set is built from `Metadata::get_key()` as well as `prefix . label`, so a key reshaped by `newspack_ras_metadata_key` matches instead of being dropped — with your `CUSTOM_<name>` scenario pinned as a test. `newspack_esp_sync_normalize_contact` can still invent keys that belong to no field, and those stay out of reach; that limit is in the README caveats.

**The blocker** is closed by scoping prefix matching to the pipeline's own UTM keys (`Legacy_Metadata::UTM_RAW_KEYS`). A registered label ending in `': '` is now matched exactly, so it can't carry a different field through. Both directions are pinned: the sibling-label case must not match, the UTM sub-key case must.

Also in this commit: the registry-miss fallback chain moved into `Metadata::get_fields()` so the ESP's effective legacy selection has one definition, with a new overridable `get_inherited_legacy_outgoing_fields()` for integrations that want a different inherited set or none; and a note on the `@deprecated` block recording legacy-pipeline de-prefixing as a sanctioned caller.

**On the default posture** — you were right that the PR body overclaimed. It now separates what changes from what does not, and says plainly that where no ESP selection was ever saved the effective selection is the full default set, so a newly connected integration inherits all of it. The stale-not-cleared behavior is written up for the release note with the ActiveCampaign and Salesforce specifics.

Tests: 202 pass across the integration, sync and metadata groups, including five new cases (custom integration prefix, registered-label over-match, UTM sub-keys, renamed keys under inheritance, and the v1-mode empty fallthrough). PHPCS clean.

PR #684, opened 2026-07-21
feat(reader-activation): inbound form capture integration

miguelpeixe:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guidelines](https://github.com/Automattic/newspack-workspace/blob/main/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds an **Inbound Form Capture** Reader Activation integration: publishers can opt specific frontend forms — built with **any** form tool (Gravity Forms, WPForms, CF7, custom HTML) — into reader capture. On submit, a small frontend script harvests the email (plus best-effort first/last name) and registers the visitor through the existing hardened `/reader-activation/register` endpoint, producing a **session-known reader** synced to the ESP. This is the capture→registration path CM asked for ("keep GF for email signup, listen for the email and register readers invisibly"), built form-tool-agnostic.

How it works:

- **Opt-in per form**: the documented `newspack-form-capture` CSS class (GF has a native "CSS Class Name" field) and/or admin-configured CSS selectors. Nothing is captured by default; the integration ships **disabled** and enabling it is the single gate for the endpoint, the page-emitted key, and the script.
- **Capture is invisible but consent-closing**: the form behaves exactly as before; new readers get the standard new-account email (same as Newsletter Subscribe block signups). Repeat submits by existing password-less readers do **not** trigger magic-link emails (new filterable condition in `register_reader()`, suppressed for this integration only).
- **Existing readers get upgraded**: the `reader_registered` data event skips existing users, so the integration schedules a contact sync for them via Action Scheduler in its own group (retryable, visible in Activity Logs; repeat captures reuse the pending action).
- **Capture-only by declaration**: `supports_push()`/`supports_pull()` return `false`, so the integration renders no outbound/inbound sync controls, never counts toward `Sync::has_one_syncable_integration()`, and `can_sync()` stays a clean pass — capture-only is a declared capability, not an error state.
[...]

PR #700, opened 2026-07-22
feat(reader-activation): per-direction sync capabilities and toggles

miguelpeixe:
### All Submissions:

The (flag-gated) Audience → Integrations settings framework assumes every integration is a bidirectional sync destination: inbound-only integrations (Form Capture — #684, Fundraise Up) show a dead Outbound section, deletion-sync toggles, and a metadata-prefix field that store options that do nothing, and they count as "syncable" in `Sync::has_one_syncable_integration()`, activating the sync connector for no-op pushes. This is the generalized follow-up PR #684 deferred.

This PR makes the framework capability-aware and gives each direction a pause toggle:

- **Capability declaration** — integrations declare push/pull ability via `supports_push()` / `supports_pull()` on `Integration`. Both default `true`, so existing connectors (ESP) keep today's behavior with no changes. Inbound-only integrations will override `supports_push()` to return `false` (follow-ups: #684 for Form Capture; newspack-manager for Fundraise Up).
- **Capability-aware settings** — the account-deletion fields, the metadata prefix, and the Outbound section follow push ability; the Inbound section follows pull ability. Sections an integration lacks simply aren't declared, so no dead controls render. The deletion-sync fields and metadata prefix render inside the Outbound section itself, alongside the field picker they govern, and pause with it.
- **Per-direction toggles** — each rendered section gets an enable toggle (`outgoing_sync_enabled` / `incoming_sync_enabled`, default on) shown in the section header area. Toggling one off pauses that direction's sync while preserving the configured metadata-field selections, so re-enabling restores them.
- **Framework honors both** — contact pushes, account-deletion propagation, sync/deletion retries, and contact pulls all skip integrations without an (enabled) direction, and `Sync::has_one_syncable_integration()` returns false when no active integration has an enabled push — with distinct `integration_push_disabled` / `integration_push_not_supported` error codes so the Audience wizard can explain why sync is unavailable.

Closes NPPD-2100.

1. Enable the screen: `wp config set NEWSPACK_INTEGRATIONS_SETTINGS_ENABLED true --raw --type=constant`, with Audience Management enabled and an ESP (e.g. Mailchimp) connected so the ESP integration is configurable.
2. In Audience → Integrations, configure the ESP integration: the Inbound and Outbound sections each show an enable toggle (both on by default) above their field pickers, and the Outbound section also contains the metadata prefix and account-deletion fields (the Settings group keeps only the integration's own fields).
3. Select some outgoing and incoming metadata fields and save. Toggle "Enable outbound sync" off: the outbound picker, metadata prefix, and account-deletion fields hide while the inbound section is unaffected; save, re-enable, and save again — the previous outgoing selection is intact. Repeat for the inbound toggle.
4. With outbound sync disabled on the only active integration, verify sync is reported unavailable: `wp eval 'var_dump( \Newspack\Reader_Activation\Sync::has_one_syncable_integration( true ) );'` returns a WP_Error including `integration_push_disabled`, and reader activity (registration, login) dispatches no ESP pushes. Re-enable and verify pushes resume.
5. With outbound sync disabled, delete a reader account: no deletion propagates to the ESP (no hard delete, no flag-mode push), even with "Sync user account deletion" checked. Re-enable and confirm deletion propagation works again.
6. With inbound sync disabled, browse the site as a logged-in reader: no contact pulls run (no `NEWSPACK-CONTACT-PULL` log entries or Reader Data updates). Re-enable and confirm pulls resume.
7. Simulate an inbound-only integration (or apply #684 on top) by registering an integration that overrides `supports_push(): false`: its configure view shows no Outbound section, deletion-sync fields, or metadata-prefix field; its own settings remain editable; and `has_one_syncable_integration( true )` reports `integration_push_not_supported` when it is the only active integration.
8. Run `n test-php --group sync-capabilities` (12 tests) and the full `n test-php` + `npm test` suites in `plugins/newspack-plugin`.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
+++ b/plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
@@ -527,6 +527,11 @@ class RAS_Contact_Sync {
 		// than a hard error the operator can resolve by enabling the field.
 		$integrations = Integrations::get_active_configured_integrations();
 		foreach ( $integrations as $integration_id => $integration ) {
+			// The sync run itself skips integrations without an (enabled) push, so
+			// their field selection must not block the backfill of the others.
+			if ( ! $integration->is_push_enabled() ) {
+				continue;
+			}
 			$enabled = $integration->get_enabled_outgoing_fields();
 			$missing = array_values( array_diff( $labels, $enabled ) );
 			if ( ! empty( $missing ) ) {
--- a/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
+++ b/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
@@ -988,7 +988,7 @@ abstract class Integration {
 				'key'         => 'outgoing_sync_enabled',
 				'type'        => 'checkbox',
 				'label'       => __( 'Enable outbound sync', 'newspack-plugin' ),
-				'description' => __( 'Sync reader data to this integration. Disabling pauses outbound sync, including account-deletion sync, and preserves the outgoing field selection.', 'newspack-plugin' ),
+				'description' => __( 'Sync reader data to this integration. Disabling pauses outbound sync, including account-deletion sync, and preserves the outgoing field selection. Changes and deletions that occur while paused are not sent retroactively on re-enable.', 'newspack-plugin' ),
 				'default'     => true,
 			];
 			$fields[] = [
--- a/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
+++ b/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
@@ -96,6 +96,21 @@ abstract class Integration {
 	protected $settings_fields = [];
 
 	/**
+	 * Memoized return value of get_settings_fields().
+	 *
+	 * The declarations are stable for the life of the instance — the subclass
+	 * half is already frozen in $settings_fields, and the base-class groups are
+	 * built from per-instance capability flags — but rebuilding them costs a
+	 * __() call per label and description. The direction toggles resolve through
+	 * this array on every is_push_enabled()/is_pull_enabled() call, which run
+	 * once per contact in sync loops, so the array is built once and reused.
+	 * Reset by init(), the only place $settings_fields can change.
+	 *
+	 * @var array|null
+	 */
+	private $settings_fields_cache = null;
+
+	/**
 	 * Constructor.
 	 *
 	 * @param string $id          The unique identifier for this integration.
@@ -295,7 +310,8 @@ abstract class Integration {
 	 * Currently only initializes settings fields, but can be extended by child classes for additional setup.
 	 */
 	public function init() {
-		$this->settings_fields = $this->register_settings_fields();
+		$this->settings_fields       = $this->register_settings_fields();
+		$this->settings_fields_cache = null;
 	}
 
 	/**
@@ -365,6 +381,13 @@ abstract class Integration {
 	 * declared field never resolves to null — get_settings_field_value() falls
 	 * back to the field default.
 	 *
+	 * The stored value is coerced with wp_validate_boolean() rather than a cast
+	 * so this agrees with the wizard's toBool(): both read the strings `'false'`
+	 * and `'0'` as off. The sanctioned write path can't produce `'false'` (the
+	 * checkbox sanitizes to a real bool), but a hand-set option or external
+	 * writer otherwise diverges in the worst direction — UI paused, dispatch
+	 * still pushing.
+	 *
 	 * @return bool True if pushes should run.
 	 */
 	final public function is_push_enabled(): bool {
@@ -372,7 +395,7 @@ abstract class Integration {
 			return false;
 		}
 		$enabled = $this->get_settings_field_value( 'outgoing_sync_enabled' );
-		return null === $enabled || (bool) $enabled;
+		return null === $enabled || \wp_validate_boolean( $enabled );
 	}
 
 	/**
@@ -384,7 +407,8 @@ abstract class Integration {
 	 *
 	 * As with is_push_enabled(), an undeclared toggle field reads as null and
 	 * counts as enabled — only an explicit stored value can pause the
-	 * direction.
+	 * direction — and the stored value is coerced with wp_validate_boolean() so
+	 * PHP and the wizard agree on the falsy string forms.
 	 *
 	 * @return bool True if pulls should run.
 	 */
@@ -393,7 +417,7 @@ abstract class Integration {
 			return false;
 		}
 		$enabled = $this->get_settings_field_value( 'incoming_sync_enabled' );
-		return null === $enabled || (bool) $enabled;
+		return null === $enabled || \wp_validate_boolean( $enabled );
 	}
 
 	/**
@@ -1104,14 +1128,20 @@ abstract class Integration {
 	 * reads as null/falsy, which the deletion dispatcher treats as disabled).
 	 * The metadata groups are capability-gated in get_metadata_fields().
 	 *
+	 * Memoized per instance — see $settings_fields_cache for why.
+	 *
 	 * @return array Array of settings field declarations.
 	 */
 	public function get_settings_fields() {
+		if ( null !== $this->settings_fields_cache ) {
+			return $this->settings_fields_cache;
+		}
 		$fields = $this->settings_fields;
 		if ( $this->supports_push() ) {
 			$fields = array_merge( $fields, $this->get_account_deletion_fields() );
 		}
-		return array_merge( $fields, $this->get_metadata_fields() );
+		$this->settings_fields_cache = array_merge( $fields, $this->get_metadata_fields() );
+		return $this->settings_fields_cache;
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
+++ b/plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
@@ -319,6 +319,65 @@ abstract class Integration {
 	abstract public function can_sync( $return_errors = false );
 
 	/**
+	 * Whether this integration can push (outbound) contact data to its external
+	 * destination.
+	 *
+	 * Push-capable integrations get the Outbound settings section, the
+	 * account-deletion sync fields and the metadata field prefix, and count
+	 * toward Sync::has_one_syncable_integration(). Inbound-only integrations
+	 * (those whose push_contact_data() is a deliberate no-op) should override
+	 * this to return false so the settings UI shows no dead outbound controls
+	 * and the sync framework skips the push path entirely.
+	 *
+	 * @return bool True if the integration can push contact data.
+	 */
+	public function supports_push(): bool {
+		return true;
+	}
+
+	/**
+	 * Whether this integration can pull (inbound) contact data from its
+	 * external source.
+	 *
+	 * Pull-capable integrations get the Inbound settings section and are
+	 * included in the Contact_Pull dispatch. Integrations that don't implement
+	 * pull_contact_data()/get_available_incoming_fields() should override this
+	 * to return false.
+	 *
+	 * @return bool True if the integration can pull contact data.
+	 */
+	public function supports_pull(): bool {
+		return true;
+	}
+
+	/**
+	 * Whether outbound (push) sync should currently run for this integration.
+	 *
+	 * Combines the push capability with the `outgoing_sync_enabled` toggle,
+	 * which pauses the direction while preserving the configured outgoing
+	 * field selection. Every push dispatch site must consult this — including
+	 * account-deletion propagation, which travels the push pipeline.
+	 *
+	 * @return bool True if pushes should run.
+	 */
+	final public function is_push_enabled(): bool {
+		return $this->supports_push() && (bool) $this->get_settings_field_value( 'outgoing_sync_enabled' );
+	}
+
+	/**
+	 * Whether inbound (pull) sync should currently run for this integration.
+	 *
+	 * Combines the pull capability with the `incoming_sync_enabled` toggle,
+	 * which pauses the direction while preserving the configured incoming
+	 * field selection. Every pull dispatch site must consult this.
+	 *
+	 * @return bool True if pulls should run.
+	 */
+	final public function is_pull_enabled(): bool {
+		return $this->supports_pull() && (bool) $this->get_settings_field_value( 'incoming_sync_enabled' );
+	}
+
+	/**
 	 * Push contact data to the integration destination.
 	 *
 	 * This method should be implemented by child classes to send
@@ -830,9 +889,11 @@ abstract class Integration {
 	/**
 	 * Get the account-deletion fields declared by this integration.
 	 *
-	 * Auto-appended to every integration's settings. The first field is a top-level
-	 * toggle; the second field is gated by the first via the `condition` predicate
-	 * honored by the frontend renderer.
+	 * Auto-appended to push-capable integrations' settings (see
+	 * get_settings_fields()): deletion propagates through the push pipeline, so
+	 * for a push-less integration these would be dead controls. The first field
+	 * is a top-level toggle; the second field is gated by the first via the
+	 * `condition` predicate honored by the frontend renderer.
 	 *
 	 * @return array Array of settings field declarations.
 	 */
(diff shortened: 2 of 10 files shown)


plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -318,6 +318,65 @@ abstract public function register_settings_fields();
...
+	 * Combines the push capability with the `outgoing_sync_enabled` toggle,
+	 * which pauses the direction while preserving the configured outgoing
+	 * field selection. Every push dispatch site must consult this — including
+	 * account-deletion propagation, which travels the push pipeline.
+	 *
+	 * @return bool True if pushes should run.
+	 */
+	final public function is_push_enabled(): bool {
+		return $this->supports_push() && (bool) $this->get_settings_field_value( 'outgoing_sync_enabled' );
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in a6079e65e — `is_push_enabled()` now treats an undeclared toggle field (null from `get_settings_field_value()`) as enabled, so only an explicit stored value or a missing capability pauses the direction, matching the frontend's missing-toggle-means-enabled behavior. Regression test added.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -318,6 +318,65 @@ abstract public function register_settings_fields();
...
+	 *
+	 * Combines the pull capability with the `incoming_sync_enabled` toggle,
+	 * which pauses the direction while preserving the configured incoming
+	 * field selection. Every pull dispatch site must consult this.
+	 *
+	 * @return bool True if pulls should run.
+	 */
+	final public function is_pull_enabled(): bool {
+		return $this->supports_pull() && (bool) $this->get_settings_field_value( 'incoming_sync_enabled' );
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in a6079e65e — same null-means-enabled handling applied to `is_pull_enabled()`, covered by the same regression test.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/claude/integrations-capability-aware-sections-d36ab9` (stash local work first).

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -318,6 +318,84 @@ abstract public function register_settings_fields();
...
+	 * back to the field default.
+	 *
+	 * @return bool True if pushes should run.
+	 */
+	final public function is_push_enabled(): bool {
+		if ( ! $this->supports_push() ) {
+			return false;
+		}
+		$enabled = $this->get_settings_field_value( 'outgoing_sync_enabled' );
+		return null === $enabled || (bool) $enabled;
```
dkoo: **Suggestion:** PHP and the wizard disagree on the string 'false' — The wizard's `toBool()` explicitly treats `''`/`'0'`/`'false'` as disabled, while `(bool) $enabled` here evaluates `'false'` as true. The sanctioned write path can't produce that value (the checkbox sanitizes to a real bool), but a hand-set option or external writer produces the worst divergence direction: the UI renders the direction as paused while dispatch keeps pushing. `wp_validate_boolean( $enabled )` instead of the cast makes both read paths agree at no cost (same at the `is_pull_enabled()` twin, ~line 396).

plugins/newspack-plugin/includes/reader-activation/sync/class-sync.php
```diff
@@ -166,12 +166,42 @@ public static function has_one_syncable_integration( $return_errors = false ) {
...
+							/* translators: %s: integration name. */
+							__( 'The %s integration does not support outbound sync.', 'newspack-plugin' ),
+							$integration->get_name()
+						)
+					);
+				}
+				continue;
+			}
+
 			$can_sync_integration = $integration->can_sync( true );
```
dkoo: **Suggestion:** Guard the bool|WP_Error contract in the rewritten availability loop — `$integration->can_sync( true )->has_errors()` fatals if a subclass's `can_sync()` ignores `$return_errors` and returns a bool — a contract the abstract signature's `@return bool|\WP_Error` actively invites (reproduced live with a contract-violating mock; the known newspack-manager subclasses honor it today). The exposure predates this PR, but since this loop is being rewritten anyway, a one-line `is_wp_error()` check is cheap insurance for third-party integration authors.

plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
```diff
@@ -527,6 +527,11 @@ class_exists( 'Newspack_Newsletters' ) &&
 		// than a hard error the operator can resolve by enabling the field.
 		$integrations = Integrations::get_active_configured_integrations();
 		foreach ( $integrations as $integration_id => $integration ) {
+			// The sync run itself skips integrations without an (enabled) push, so
+			// their field selection must not block the backfill of the others.
```
dkoo, line 531: **Nit:** CLI backfill guard is the one dispatch site without a test — The otherwise-thorough 18-test matrix covers both retry chains, deletion propagation (disabled and incapable), and pull retries — but not the CLI backfill/field-warning path this guard protects. Low risk (a trivial delegation to a well-tested method), but a test asserting the backfill doesn't warn about a push-disabled integration's missing fields would close the matrix.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -318,6 +318,84 @@ abstract public function register_settings_fields();
...
+	 * counts as enabled: the toggle can only ever pause sync explicitly,
+	 * mirroring the frontend's missing-toggle-means-enabled rendering. A
+	 * declared field never resolves to null — get_settings_field_value() falls
+	 * back to the field default.
+	 *
+	 * @return bool True if pushes should run.
+	 */
+	final public function is_push_enabled(): bool {
+		if ( ! $this->supports_push() ) {
+			return false;
```
dkoo, line 395: **Nit:** Toggle lookups rebuild the full field-declaration array per call — Each `is_push_enabled()`/`is_pull_enabled()` call resolves the toggle through `get_settings_fields()`, rebuilding the full declaration array (including `__()` calls for every label) on every check. No extra queries — options are object-cached — but in per-contact loops (CLI backfill, contact sync) it's repeated pure-CPU work; memoizing the declarations per instance would trim it.

plugins/newspack-plugin/includes/reader-activation/sync/class-sync.php
```diff
@@ -166,12 +166,42 @@ public static function has_one_syncable_integration( $return_errors = false ) {
...
+			// push never satisfy it, even when their can_sync() reports no errors
+			// (an inbound-only integration has nothing to gate there).
+			if ( ! $integration->is_push_enabled() ) {
+				if ( $integration->supports_push() ) {
+					$result->add(
+						'integration_push_disabled',
+						sprintf(
+							/* translators: %s: integration name. */
+							__( 'Outbound sync is disabled for the %s integration.', 'newspack-plugin' ),
+							$integration->get_name()
```
dkoo: **Nit:** Error messages built and discarded in non-errors mode — With `$return_errors = false`, the loop still builds translated, `sprintf`'d messages into `$result` for every disabled/unsupported integration and then discards them on the boolean return paths. This is a gating predicate on hot paths; wrapping the `$result->add()` calls in `if ( $return_errors )` avoids per-call `__()`/`sprintf` work.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -133,6 +133,13 @@ export const reconcileOperators = ( currentMap, options ) => {
 // in JS, so the falsy string forms are matched explicitly.
 const toBool = value => ( typeof value === 'string' ? ! [ '', '0', 'false' ].includes( value.toLowerCase() ) : Boolean( value ) );
 
+// Push-pipeline settings that render inside the Outbound section: the metadata
+// prefix is only read on push paths (prepare_contact()) and account-deletion
+// sync travels the push pipeline (see get_account_deletion_fields() /
+// get_metadata_fields() in class-integration.php), so they belong with — and
+// pause with — outbound sync.
```
dkoo, line 140: **Nit:** Outbound settings keys hardcoded in the frontend — `OUTBOUND_SETTINGS_KEYS` hardcodes three backend field keys; a future push-pipeline field added server-side will silently land in the generic Settings group instead of the Outbound section (and won't pause with it). The failure mode is safe and the cross-references are documented in both directions, so this is acceptable coupling — a `section` hint in the field declaration would remove it if the list grows.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
```diff
@@ -406,85 +470,110 @@ const ConfigureViewInner = ( { integrations, loading, inFlightChanges, saving, o
...
+												? currentMap[ optionValue ]
+												: operatorOptions[ 0 ]?.value;
+											return (
+												<div className="newspack-configure-view__inbound-field" key={ optionValue }>
+													<CheckboxControl
+														className="newspack-checkbox-control"
+														label={ optionLabel }
+														checked={ checked }
+														onChange={ isChecked =>
+															handleFieldChange( inboundField.key, toggleField( currentMap, option, isChecked ) )
```
dkoo, line 501: **Nit:** Outbound divider can dangle in a contrived empty-section case — The section-toggle comment says the divider renders only when there is content to divide, and the inbound call honors that with an emptiness check — but the outbound call passes only `outboundEnabled`. If `outboundField` is absent and every outbound settings field is hidden by an unsatisfied condition ref, an enabled outbound section renders a dangling divider. Cosmetic and hard to reach; mirror the inbound check or soften the comment.

dkoo (commented):
Release risk: LOW — zero blockers, a consequences verdict of none, compatible cross-plugin impact, and verification is now complete: the capability/toggle state machine is exercised by an 18-test PHP matrix, 437 JS tests, a live CLI matrix, and a full browser walkthrough of the Integrations wizard in an isolated env (capability-aware sections, per-direction toggle pause/persist/restore with field selections preserved, and the integration_push_disabled availability messaging all verified live). One scope caveat: real ESP push/pull runs were not exercisable — the ESP integration is active but unconfigured, live credentials withheld by policy — so ESP-specific paths are covered by the test suites and mock integrations only. The framework rewiring itself is sound: every dispatch site routes through the two centralized direction gates, the null-means-enabled default (the earlier review concern, already fixed) is complete across all read paths, existing connector subclasses are verified non-breaking, and the reworked availability check fixes a latent bug where stale errors could leak into a success verdict. Worth a look before merge, none blocking: the CLI dry-run preview does not apply the new push gate, so --dry-run can claim a paused integration will sync when the real run skips it; a one-line is_wp_error() guard in the rewritten availability loop would harden a documented bool-or-error contract that subclasses are invited to break; and wp_validate_boolean() on the toggle reads would keep PHP and the wizard agreeing on the string false, closing a path where the UI shows a direction paused while dispatch keeps pushing. One expectation note for docs rather than code: pausing outbound sync stops this framework pushes, but newsletter-signup contact upserts flow to the ESP through the Newsletters plugin own channel and are unaffected by the toggle. Approve-leaning.

- `plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php:1180` — **Suggestion:** CLI dry-run preview doesn't mirror the new push gate — `log_dry_run_with_options()` iterates every active configured integration without the `is_push_enabled()` skip the real push path now applies, though its docblock promises it mirrors the real path. `wp newspack esp sync <user> --dry-run` can therefore tell an operator that data will be pushed to an integration the actual run silently skips — the exact preview-vs-reality gap dry-run exists to prevent. Diagnostic-only (no I/O), so not blocking; adding the same guard, or logging an explicit "skipped: outbound paused" line, restores the mirror.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Addressed in 63f0939 — both `is_push_enabled()` and `is_pull_enabled()` now coerce with `wp_validate_boolean()` instead of the cast, so PHP and the wizard's `toBool()` agree on `'false'`. The `null === $enabled ||` guard stays in front (`wp_validate_boolean( null )` is false, and an undeclared toggle must still read as enabled). New regression test `test_falsy_string_toggle_values_pause_directions` forces a `'false'` option row on both directions and asserts each is paused.

plugins/newspack-plugin/includes/reader-activation/sync/class-sync.php
miguelpeixe: Addressed in 32c030b — the loop now normalizes before calling `has_errors()`: a bare bool from a subclass that ignores `$return_errors` becomes an empty `WP_Error` for `true`, or one carrying `integration_cannot_sync` for `false`. Uses the value already returned rather than calling `can_sync()` a second time, so no extra API work for a contract-honoring integration. Two tests cover it (`..._tolerates_true_bool_can_sync` / `..._tolerates_false_bool_can_sync`) using a mock that returns a bare bool regardless of the flag.

plugins/newspack-plugin/includes/cli/class-ras-contact-sync.php
miguelpeixe: Addressed in 9cbfd4f — two tests added to `Test_RAS_Contact_Sync_Options`, mirroring the existing `test_field_not_enabled_on_integration_returns_wp_error` setup so the pass/fail pair sits side by side: `test_field_not_enabled_on_push_disabled_integration_is_ignored` (outbound paused) and `test_field_not_enabled_on_push_incapable_integration_is_ignored` (no push capability). Both assert the pre-flight returns the resolved options rather than `newspack_esp_sync_fields_not_enabled`, covering both branches of the `is_push_enabled()` guard.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Addressed in 63f0939 — `get_settings_fields()` now memoizes into a private `$settings_fields_cache`, reset in `init()` (the only place `$settings_fields` can change). Safe to freeze the translated strings here because the subclass half was already frozen in `$settings_fields` at construction — this just makes the base-class groups behave the same way instead of re-running every `__()` per call. Verified no subclass mutates the declarations at runtime: `supports_push()`/`supports_pull()`/`supports_hard_delete()` are all constant returns, and `register_settings_fields()` is documented as static declarations only.

plugins/newspack-plugin/includes/reader-activation/sync/class-sync.php
miguelpeixe: Addressed in 32c030b — the two `$result->add()` calls are now wrapped in `if ( $return_errors )`, so the boolean path skips the `__()`/`sprintf` work entirely. Behavior is unchanged: $result is only ever read on the `$return_errors` return, and `test_has_one_syncable_integration_requires_enabled_push` / `..._false_for_pull_only` still assert both `integration_push_disabled` and `integration_push_not_supported` come back in errors mode.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
miguelpeixe: Addressed in 716e31e — mirrored the inbound check rather than softening the comment. Hoisted `visibleOutboundSettingsFields` next to `visibleSettingsFields` and the outbound call now passes `outboundEnabled && ( visibleOutboundSettingsFields.length > 0 || !! outboundField )`; the render below reuses the same list instead of re-running `.filter( fieldIsVisible )` inline. The `Divider` stub in the test suite was returning `null`, so no test could see this — it now renders a tagged element (`section-divider` for the `alignment="full-width"` ones, `toggle-divider` otherwise), and two tests assert the counts, including the contrived case: an enabled Outbound section with a settings field hidden behind an unsatisfied condition and no picker renders no divider.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js
miguelpeixe: Leaving as-is, agreeing with your read that the coupling is acceptable. A `section` hint in the field declaration is the right long-term answer, but it's a change to the settings-config contract that every integration and the REST serializer would need to understand — more surface than this PR should take on, and the failure mode if the list goes stale is safe (a new push field lands in the generic Settings group rather than disappearing). Happy to do it as a follow-up if the list grows past these three.

The cross-references in both directions stay as the guardrail: `OUTBOUND_SETTINGS_KEYS` points at `get_account_deletion_fields()`/`get_metadata_fields()`, and those point back here.

miguelpeixe:
Thanks @dkoo — all six points addressed, plus the docs note. Inline threads have their own replies; the two that can't be threaded are below.

### CLI dry-run preview doesn't mirror the new push gate (`class-contact-sync.php:1180`)

Fixed in 9cbfd4f. `log_dry_run_with_options()` now applies the same `is_push_enabled()` skip the real push path does, and logs the reason instead of a payload the run would never send:

```
[dry-run] SKIPPED integration "esp": outbound sync is paused.
[dry-run] SKIPPED integration "foo": integration does not support outbound sync.
```

Logging the skip rather than silently dropping the line keeps the preview honest in both directions — the operator sees the integration was considered and why it won't receive anything, instead of it just vanishing from the output.

### Expectation note: newsletter signups are unaffected by the toggle

Documented in 0c745ab, and I verified the claim rather than taking it on faith: newsletter signups reach the ESP via `Newspack_Newsletters_Contacts::subscribe()` inside newspack-newsletters, which never touches `Contact_Sync` or `Integration::push_contact_data()`. So `outgoing_sync_enabled` genuinely has no bearing on that path.

The integrations `README.md` turned out to predate this PR entirely — it still claimed "every integration automatically gets three additional fields" and two account-deletion fields, which capability gating made wrong. So rather than a one-line caveat, that commit adds a **Direction capabilities and toggles** section (capability vs. toggle, what declaring `supports_push() === false` removes, pause-is-not-retroactive, undeclared-toggle-means-enabled), corrects both "every integration" claims to be capability-aware, adds the two toggle fields to the settings tables, and lists `supports_push()`/`supports_pull()` under Optional overrides. The scope caveat lands there as a **Scope caveat** callout.

### Summary of the other five

| Item | Commit |
| --- | --- |
| `wp_validate_boolean()` on both toggle reads | 63f0939 |
| `is_wp_error()` guard on the `bool\|WP_Error` contract | 32c030b |
| Skip building discarded error messages when `$return_errors` is false | 32c030b |
| Memoize the settings field declarations | 63f0939 |
| Mirror the inbound emptiness check on the outbound divider | 716e31e |
| CLI backfill guard test | 9cbfd4f |

`OUTBOUND_SETTINGS_KEYS` is the one item left as-is — agreeing with your own read that the coupling is acceptable; reasoning in that thread.

### Verification

Everything below run against the `pr700` isolated env with these changes applied:

- **PHPUnit, full suite:** 2314 tests, 6529 assertions, 0 failures (7 pre-existing skips). `sync-capabilities` is now 21 tests (was 18), `Contact_Sync_Options` 28 (was 26).
- **Jest, full suite:** 45 suites, 439 tests passing (was 437).
- **PHPCS** clean on all five changed PHP files; **ESLint** clean on both JS files.

Six new tests total: falsy-string toggle coercion, the two bool-`can_sync()` normalization cases, the two CLI backfill guard cases (paused and push-incapable), and the divider counts. The last one needed the `Divider` test stub — it was returning `null`, so no test could observe a divider at all; it now renders a tagged element so the toggle divider is assertable.

PR #701, opened 2026-07-22
fix(newsletters): guard fix_public_status exit for non-page requests

dkoo:

`Newspack_Newsletters::fix_public_status()` is hooked on `the_post` and, for a `publish` newsletter whose `is_public` meta is falsey, flips it to `private` and then issues `header( 'Refresh:0' ); exit;` to reload the front-end page. That `exit` was guarded only by `! is_admin()`.

During a **REST** request (and cron / WP-CLI / XML-RPC / a feed render), `is_admin()` is `false`, so the guard passed and the request `exit`ed mid-flight the first time such a newsletter appeared in the loop — truncating the response. This surfaced while QA-testing the new newsletters admin list ([#677](https://github.com/Automattic/newspack-workspace/pull/677)): the batched "All" fetch (`per_page=100`, pages 2+) returned a partial page and the UI showed "Only some items could be loaded." The list code there is correct; this hook was the root cause, and it is pre-existing (independent of #677).

This restricts the redirect/`exit` to a genuine front-end page view:

- Extracts the decision into a small `is_front_end_page_request()` helper.
- Returns `false` (so no `exit`) for `is_admin()`, `REST_REQUEST`, `wp_doing_ajax()`, `wp_doing_cron()`, `WP_CLI`, `XMLRPC_REQUEST`, and `is_feed()`.

Feeds and XML-RPC are streamed/programmatic responses rather than page views — and newsletters can appear in feeds via `display_newsletters_in_archives()` — so an `exit` there would truncate the output (e.g. a feed's XML) mid-document just as it did for REST.

The status correction itself (publish → private for non-public newsletters) is unchanged and still runs in every context; only the page-refresh `exit` is now scoped to real page loads.

### How to test the changes in this Pull Request

1. `cd plugins/newspack-newsletters && n test-php --filter Newsletter_Controlled_Statuses_Test` — the group passes, including the new tests:
   - `test_public_status_refresh_suppressed_off_page` — the guard returns `false` under AJAX / cron / feed (the constant-gated REST, WP-CLI and XML-RPC branches share the same helper and are covered end-to-end by the env run below).
   - `test_fix_public_status_corrects_without_exit_off_page` — an invalid publish + non-public newsletter is still corrected to `private` without exiting (pre-fix, this call would `exit` and abort the run).
2. On a site with many `publish`-but-non-public newsletters, request the newsletters REST collection at `per_page=100` across multiple pages and confirm each page returns in full rather than truncating. (Verified in an isolated env: 250 seeded newsletters, 240 forced into the invalid state → pages returned 100 / 100 / 50, `X-WP-Total: 250`, no truncation, rows self-healed to `private`, debug log clean.)

### Other information

- Root-caused during the manual QA of #677 (DSGNEWS-200); split out as its own fix since it is pre-existing and independent.

### DIFF
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
@@ -1595,6 +1595,14 @@ final class Newspack_Newsletters {
 		}
 		$is_public = (bool) get_post_meta( $post->ID, 'is_public', true );
 		if ( 'publish' === $post->post_status && ! $is_public ) {
+			// Correcting to `private` does NOT trigger an ESP send. A
+			// publish -> private transition stays within the controlled
+			// statuses (`['publish', 'private']`), so neither `pre_post_update()`
+			// (which sends when a newsletter moves out of / into that set) nor
+			// `transition_post_status()` (which sends only when `$old_status` is
+			// `'future'`) fires. This matters because, with the `exit` below now
+			// scoped to page views, a single request can correct many rows
+			// instead of stopping at the first — many corrections, still no sends.
 			wp_update_post(
 				[
 					'ID'          => $post->ID,
@@ -1646,7 +1654,19 @@ final class Newspack_Newsletters {
 		}
 		// A feed is a front-end request but not a page view; an `exit` here
 		// would truncate the feed's XML mid-document. Newsletters can appear in
-		// feeds via `display_newsletters_in_archives()`.
+		// feeds via `display_newsletters_in_archives()`. Behaviour change worth
+		// naming: a feed containing a just-corrected newsletter now serves that
+		// item once in the current response (a reader hitting the feed at that
+		// moment sees it) rather than emitting invalid XML, and it drops out of
+		// subsequent requests once healed — one slightly-stale item beats broken
+		// XML.
+		//
+		// This check MUST stay after the REST/AJAX/cron/CLI/XML-RPC returns
+		// above — the ordering is load-bearing, not stylistic. `is_feed()` is a
+		// main-query conditional; called with `$wp_query` unset (e.g. mid-REST)
+		// it triggers `_doing_it_wrong()` and returns false, which would let the
+		// `exit` through in exactly the contexts this guard protects. Do not
+		// reorder these into alphabetical / "cheapest check first" order.
 		if ( is_feed() ) {
 			return false;
 		}
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
@@ -1616,11 +1616,12 @@ final class Newspack_Newsletters {
 
 	/**
 	 * Whether the current request is a genuine front-end page view, as opposed
-	 * to an admin screen, REST or AJAX request, cron run, or WP-CLI invocation.
+	 * to an admin screen, REST/AJAX/XML-RPC request, cron run, WP-CLI
+	 * invocation, or feed render.
 	 *
 	 * Used to decide whether it is safe to `exit` the request: on a page view a
-	 * redirect is the intended behavior, but on any programmatic request an
-	 * `exit` would truncate the response mid-flight.
+	 * redirect is the intended behavior, but on any programmatic request or
+	 * streamed response an `exit` would truncate the output mid-flight.
 	 *
 	 * @return bool True on a front-end page request, false otherwise.
 	 */
@@ -1638,6 +1639,15 @@ final class Newspack_Newsletters {
 			return false;
 		}
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			return false;
+		}
+		if ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) {
+			return false;
+		}
+		// A feed is a front-end request but not a page view; an `exit` here
+		// would truncate the feed's XML mid-document. Newsletters can appear in
+		// feeds via `display_newsletters_in_archives()`.
+		if ( is_feed() ) {
 			return false;
 		}
 		return true;
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters.php
@@ -1603,12 +1603,44 @@ final class Newspack_Newsletters {
 				false,
 				false
 			);
-			// Force a page refresh on the front-end.
-			if ( ! is_admin() ) {
+			// Force a page refresh, but only on a genuine front-end page view.
+			// During REST, AJAX, cron or WP-CLI the `exit` would truncate the
+			// current request (e.g. a REST collection would return a partial,
+			// short-circuited response), so restrict it to real page loads.
+			if ( self::is_front_end_page_request() ) {
 				header( 'Refresh:0' );
 				exit;
 			}
 		}
+	}
+
+	/**
+	 * Whether the current request is a genuine front-end page view, as opposed
+	 * to an admin screen, REST or AJAX request, cron run, or WP-CLI invocation.
+	 *
+	 * Used to decide whether it is safe to `exit` the request: on a page view a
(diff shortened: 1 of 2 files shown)


plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -1603,14 +1603,56 @@ public static function fix_public_status( $post ) {
...
+	 * redirect is the intended behavior, but on any programmatic request or
+	 * streamed response an `exit` would truncate the output mid-flight.
+	 *
+	 * @return bool True on a front-end page request, false otherwise.
+	 */
+	private static function is_front_end_page_request() {
+		if ( is_admin() ) {
+			return false;
+		}
+		if ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
```
thomasguillot, line 1640: Worth a note that this ordering is load-bearing, not stylistic: `is_feed()` is a conditional query tag, so it has to stay *after* the REST/cron/CLI returns. Called with `$wp_query` unset it emits `_doing_it_wrong` and returns false, which would then let the `exit` through in exactly the contexts this guard exists to protect. A one-liner here would stop someone reordering these into alphabetical or "cheapest check first" order later.

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
```diff
@@ -1603,14 +1603,56 @@ public static function fix_public_status( $post ) {
...
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			return false;
+		}
+		if ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) {
+			return false;
+		}
+		// A feed is a front-end request but not a page view; an `exit` here
+		// would truncate the feed's XML mid-document. Newsletters can appear in
+		// feeds via `display_newsletters_in_archives()`.
+		if ( is_feed() ) {
```
thomasguillot, line 1670: The feed case is a real behaviour change, and an improvement — worth naming it as such rather than leaving it implicit. Before: a feed containing such a newsletter truncated mid-XML. After: the item renders in that one response despite having just been corrected to private, and drops out of subsequent requests. Serving one slightly stale item beats emitting invalid XML, but a reader hitting that feed at that moment sees the newsletter once.

plugins/newspack-newsletters/tests/test-controlled-statuses.php
```diff
@@ -69,6 +69,67 @@ public function test_publish_public_newsletter() {
...
 
+	/**
+	 * The front-end refresh/exit in fix_public_status() must be suppressed on
+	 * programmatic requests (REST, AJAX, cron, WP-CLI, XML-RPC) and feed
+	 * renders, where an exit would truncate the response mid-flight. The
+	 * REST_REQUEST, WP_CLI, and XMLRPC_REQUEST branches hinge on `define()`
+	 * constants that cannot be toggled per-test, so they are covered by the
+	 * filter-toggleable AJAX/cron/feed branches that share the same guard.
+	 */
+	public function test_public_status_refresh_suppressed_off_page() {
```
thomasguillot, line 81: Reasonable call given the constants, and the env run covers the real path. If you want the actual bug path pinned in CI, `@runInSeparateProcess` with `define( 'REST_REQUEST', true )` works for the one branch that matters — the REST case is the regression this fixes, and it is the only branch currently resting on a proxy.

thomasguillot (approved):
Approving. Right diagnosis and a minimal fix in the right place.

I checked the thing this change makes riskier: with the `exit` gone, one request can now correct many newsletters instead of stopping at the first (240 rows in your env). Each correction is a `wp_update_post` publish → private, and a publish/private transition is the path that fires an ESP send. Neither send path triggers here — `transition_post_status()` only sends when `'future' === $old_status`, and `pre_post_update()` only sends when the old status is *not* in `$controlled_statuses`, which `publish` is. So the higher correction volume carries no send exposure. Worth stating explicitly, since "N status transitions where there used to be 1" is the shape that bites in this plugin.

One behavioural note for the record: a GET REST collection can now issue up to `per_page` writes on an affected site. It is self-limiting (rows heal once and stay healed), but the first "All" walk on a site with many invalid rows does a lot of writing inside read requests. Fine as the cost of self-healing, just worth knowing before someone sees it in a slow query log.

I also checked for sibling instances of the same pattern elsewhere in the plugin — this is the only `exit` of its kind.

adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/fix/newsletters-public-status-rest-exit` (stash local work first).

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
dkoo: Good call — made it explicit. The comment now states the ordering is load-bearing rather than stylistic, with your rationale spelled out: `is_feed()` is a main-query conditional, so with `$wp_query` unset (e.g. mid-REST) it triggers `_doing_it_wrong()` and returns false, which would let the `exit` through in exactly the contexts this guard protects — plus a "do not reorder into alphabetical / cheapest-check-first order" warning. (670c075f2)

plugins/newspack-newsletters/includes/class-newspack-newsletters.php
dkoo: Named it explicitly rather than leaving it implicit. The comment now spells out the tradeoff: a feed containing a just-corrected newsletter serves that item **once** in the current response (a reader hitting the feed at that moment sees it) and drops out of subsequent requests once healed — one slightly-stale item beats emitting invalid XML. (670c075f2)

plugins/newspack-newsletters/tests/test-controlled-statuses.php
dkoo: Took the suggestion. Added `test_public_status_refresh_suppressed_during_rest()` using `@runInSeparateProcess` + `@preserveGlobalState disabled` with `define( 'REST_REQUEST', true )`, so the branch this fix actually regresses on is pinned in CI instead of resting on the AJAX/cron proxy. Updated the neighbouring docblock so it no longer claims REST is proxy-covered. Group is green: 6 tests, 14 assertions. (670c075f2)

dkoo:
Thanks for the review — especially for independently checking the send exposure, which is the axis that actually got riskier here.

I verified your analysis holds and then documented it in the code, since you are right that "N status transitions where there used to be 1" is the shape that bites in this plugin. Confirmed in `class-newspack-newsletters-service-provider.php`: `$controlled_statuses = [ 'publish', 'private' ]`, `pre_post_update()` only sends when a newsletter moves **out of** that set (L121) or **into** it from outside (L129-131), and `transition_post_status()` only sends when `$old_status === 'future'` (L183). A publish → private correction is controlled → controlled with a non-future old status, so neither path fires — many corrections, still zero sends. That reasoning now lives in a comment at the correction site.

Your other three points are addressed and the threads resolved:

- **Ordering is load-bearing** — called out explicitly, with the `_doing_it_wrong()`/returns-false rationale and a do-not-reorder warning.
- **Feed behaviour change** — named explicitly, including that a reader hitting the feed at that moment sees the item once.
- **Pin the REST branch** — added `test_public_status_refresh_suppressed_during_rest()` with `@runInSeparateProcess` + `define( 'REST_REQUEST', true )`. The regression branch is now covered directly rather than by proxy.

Also noted your observation about a GET REST collection issuing up to `per_page` writes on an affected site — agreed it is the cost of self-healing and self-limiting, and worth knowing before it shows up in a slow query log.

All in `670c075f2`. PHPCS clean; group green at 6 tests / 14 assertions.

dkoo:
I cherry-picked this to `alpha` because it fixes a bug introduced by #677, which is already in that branch.

PR #702, opened 2026-07-22
fix(content-distribution): don't mark failed dispatch as distributed

dkoo:
Fixes NPPM-3036.

The content-distribution REST endpoint `API::distribute()` (`plugins/newspack-network/includes/content-distribution/class-api.php`) discarded the return value of `Data_Events::dispatch( 'network_post_updated', $payload )`.

`Data_Events::dispatch()` returns a `WP_Error` when the action is not registered (and when a `newspack_data_events_dispatch_body` filter cancels the dispatch); it returns void otherwise. Because that value was ignored, when a dispatch failed the endpoint still:

1. wrote the `_newspack_network_payload_hash` post-meta, and
2. returned HTTP 200.

The stored hash then tells the plugin "this version was already sent" and suppresses the self-healing re-sync on the next post update — so the source site believes the post was distributed while the destination never received it, and the hash actively prevents recovery.

This PR captures the dispatch result and, on `is_wp_error()`, returns a `WP_Error` (HTTP 500) **before** writing the payload hash. Skipping the hash on failure is the important part: it leaves the post's payload hash empty so the next save re-attempts distribution. The already-recorded destination in the distribution meta is intentionally left in place so the post self-heals on the next update.

The failure path is hard to reach in a normal install (`network_post_updated` is registered during init), so this is a latent correctness hole rather than something publishers are hitting today.

### How to test the changes in this Pull Request

Covered by new PHPUnit tests in `tests/unit-tests/content-distribution/test-api.php` (with a controllable `Newspack\Data_Events` mock in `mock-data-events.php`):

1. `n test-php --group content-distribution-api` (from `plugins/newspack-network`) — passes.
2. Before the fix, the two failure-path tests fail: the endpoint returned a `WP_REST_Response` (200) instead of a `WP_Error`, and wrote the payload-hash meta despite the failed dispatch.
3. After the fix: a failed dispatch returns a `WP_Error` with `status` 500, does not write the payload hash, and leaves the destination recorded so the next save retries. The happy path still stores the hash and returns a response.
4. Full suite green: `n test-php` in `plugins/newspack-network` (192 tests).

### Follow-up (not in this PR)

The same ignored-`dispatch()`-return pattern exists in the automatic on-save distribution paths — `Content_Distribution::distribute_post()` (`includes/class-content-distribution.php:498-500`) and `distribute_post_partial()` (`:535-537`) — which write the payload hash unconditionally after dispatch. Those are the shutdown-deferred paths that run on every save, so completing the same guard there makes the self-heal robust across repeated failures. Kept out of this hotfix to stay minimal; worth a fast-follow.

### Other information

### DIFF
--- a/plugins/newspack-network/includes/content-distribution/class-api.php
+++ b/plugins/newspack-network/includes/content-distribution/class-api.php
@@ -178,8 +178,14 @@ class API {
 			return new WP_Error( 'newspack_network_content_distribution_error', $distribution->get_error_message(), [ 'status' => 400 ] );
 		}
 
-		$payload = $outgoing_post->get_payload( $status_on_publish );
-		Data_Events::dispatch( 'network_post_updated', $payload );
+		$payload    = $outgoing_post->get_payload( $status_on_publish );
+		$dispatched = Data_Events::dispatch( 'network_post_updated', $payload );
+
+		// Bail before storing the payload hash if the dispatch failed, so the next
+		// post update retries the distribution instead of being suppressed by the hash.
+		if ( is_wp_error( $dispatched ) ) {
+			return new WP_Error( 'newspack_network_content_distribution_error', $dispatched->get_error_message(), [ 'status' => 500 ] );
+		}
 
 		// Store payload hash to prevent unnecessary updates.
 		update_post_meta( $post_id, Content_Distribution_Class::PAYLOAD_HASH_META, $outgoing_post->get_payload_hash( $payload ) );
--- a/plugins/newspack-network/tests/unit-tests/content-distribution/mock-data-events.php
+++ b/plugins/newspack-network/tests/unit-tests/content-distribution/mock-data-events.php
@@ -1,0 +1,54 @@
+<?php // phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
+/**
+ * Controllable mock of the Newspack Data Events class.
+ *
+ * The real class lives in newspack-plugin, which is not loaded in this test
+ * suite. Defined here so tests can exercise code paths guarded by
+ * class_exists( 'Newspack\Data_Events' ) and control the dispatch() return value.
+ *
+ * Guarded so it does not collide with the ad-hoc mock evaluated at runtime in
+ * sibling tests; those only rely on dispatch() returning a truthy value, which
+ * the default preserves.
+ *
+ * @package Newspack_Network
+ */
+
+namespace Newspack;
+
+if ( ! class_exists( 'Newspack\Data_Events' ) ) {
+	/**
+	 * Mock Data Events class.
+	 */
+	class Data_Events {
+		/**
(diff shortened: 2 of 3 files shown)


dkoo:
@thomasguillot tagging you as a reviewer since this came out of a follow-up from a [PR of yours](https://github.com/Automattic/newspack-workspace/pull/679).

thomasguillot:
Heads-up for whoever handles the release-to-main promotion: #679 (targeting main) also adds `tests/unit-tests/content-distribution/test-api.php` with the same `Test\Content_Distribution\TestApi` class, so the promotion merge will hit an add/add conflict on that path. The two test classes are complementary rather than duplicates (this PR covers dispatch failure, #679 covers the route's permission callback) and should be folded into one class. Since this PR lands first, that merge will happen on the #679 side; nothing to change here.

plugins/newspack-network/includes/content-distribution/class-api.php
```diff
@@ -178,8 +178,14 @@ public static function distribute( $request ) {
 			return new WP_Error( 'newspack_network_content_distribution_error', $distribution->get_error_message(), [ 'status' => 400 ] );
 		}
 
-		$payload = $outgoing_post->get_payload( $status_on_publish );
-		Data_Events::dispatch( 'network_post_updated', $payload );
+		$payload    = $outgoing_post->get_payload( $status_on_publish );
+		$dispatched = Data_Events::dispatch( 'network_post_updated', $payload );
```
thomasguillot, line 182: Note for the fast-follow scope: `dispatch()` only queues the payload, and the actual send happens in `execute_queued_dispatches()` at shutdown. A failure at that point still stores the hash, so this guard covers the synchronous failure modes only. Fine for this hotfix.

thomasguillot (approved):
The fix is minimal and correct. `Data_Events::dispatch()` returns a `WP_Error` in exactly two cases (unregistered action, and cancellation via the `newspack_data_events_dispatch_body` filter), and returning before the hash write is what lets the next save retry: `Content_Distribution::distribute_post()` skips dispatch when the stored hash matches the new payload. Skipping the hash while leaving the destination in distribution meta is the right combination for the self-heal.

The test design is sound. The load-order reasoning in `mock-data-events.php` holds: the ad-hoc `eval` mocks in `test-distributor-migrator.php` are created at test runtime, so the file-load-time mock always wins when both are in the suite, and its `true` default preserves what the migrator tests need. Coverage hits both halves of the failure contract (error surfaced, hash not written, destination retained) plus the happy path, with `null` correctly mirroring the real void return on success.

500 for a server-side dispatch failure is the right status alongside the endpoint's 400s, and keeping the on-save paths (`distribute_post()`/`distribute_post_partial()`) as a fast-follow is reasonable.

See the earlier comment about the `test-api.php` add/add conflict with #679 at promotion time; nothing to change here.

PR #704, opened 2026-07-23
fix(newsletters): skip term embeds when the ads list hides those columns

thomasguillot:
### All Submissions:

Follow-up to #677, which skipped term embeds on the newsletters list when no taxonomy column is showing but deliberately left the ads list alone. The ads list carries three term-backed columns (Advertiser, Ad placement, Categories), so it pays for the embed three times over, and Categories is hidden by default.

It was left out because its Quick Edit hydrated advertiser and placement from `_embedded['wp:term']` alone and sent both taxonomies on every save, so dropping the embed would have cleared them on the first status-only save. This PR closes that gap first, then applies the optimisation:

- **Quick Edit hydrates from raw term IDs.** The list now always requests `newspack_nl_advertiser`, `ad_placement` and `categories` in `_fields`, and each field falls back to resolving those IDs against its options list when the embed is absent.
- **Only touched taxonomies are sent.** Previously every save posted all three taxonomies, so an untouched field could overwrite what was stored. Each taxonomy now has its own dirty flag, and unresolvable term IDs ride along rather than being silently dropped.
- **Separate edited-refs per taxonomy.** The three options lists resolve independently (advertisers and placements arrive as props, categories are fetched in the panel), so a shared flag would let an edit to one freeze another's baseline at its pre-resolution empty value.
- **Selections merge the embed with the options list.** Neither source is complete on its own. There is no embed at all once no term-backed column is visible, and the embed caps at 100 terms per taxonomy: when an embed link carries no `per_page`, `WP_REST_Server::embed_links()` raises the sub-request to the matched handler's schema *maximum*, which for the terms collection is 100. Stored IDs therefore resolve through the embed *and* the fetched list, so neither can shadow the other. This also fixes a pre-existing bug, though a rarer one than 100 terms in a single taxonomy makes it sound: a quick-edit save on an ad past that cap used to truncate it.
- **Fields are read-only until their options settle** and account for every stored term. Editing earlier would diff against a baseline still growing underneath it, dropping terms resolved late; and when a settled list still can't explain the ad's terms, the field says so via a notice rather than showing a quietly wrong value. While it waits, the field carries a `help` string saying so, since a disabled control leaves the tab order with nothing to explain its absence.
- **`_embed` is then gated** on Advertiser, Ad placement or Categories being visible, and omitted entirely otherwise, since nothing else on this screen reads `_embedded`. `_links` goes with it — it is only in `_fields` so `_embed` has links to expand (verified: dropping `_links` from `_fields` also drops `_embedded`).

Verified on a local site: with all three columns hidden, the list request drops `_embed` and keeps the raw ID fields; Quick Edit still shows the advertiser; and saving a price change leaves the advertiser attached in the database.

1. Go to **Newsletters → Advertising**. Open DevTools → Network and note the `newspack_nl_ads_cpt` request: it should include `_embed=wp:term` (Advertiser and Ad placement are visible by default) and list `newspack_nl_advertiser,ad_placement,categories` in `_fields`.
2. Open **View options** and hide **Advertiser** and **Ad placement** (Categories is hidden by default). The next request should have no `_embed` at all, while `_fields` still carries the three taxonomy fields.
3. With those columns still hidden, open **Quick Edit** on an ad that has an advertiser. The Advertiser field must show it, hydrated from the raw IDs rather than the embed.
4. Change only the price (or the status) and save. Reopen Quick Edit, or check the ad in the block editor, and confirm the advertiser and placement are still attached. Before this change that save would have cleared them.
5. Change the advertiser itself and save. Only that taxonomy should change; placement and categories stay as they were.
6. Show the Categories column again and confirm `_embed=wp:term` returns and the Advertiser, Ad placement and Categories cells all render correctly.
7. Open Quick Edit on an ad while its terms are still loading (throttle the network if it is too quick locally). The three token fields must be greyed out, each with its own "Loading …" line underneath, and no warning.
8. Go to **Newsletters → All Newsletters** and repeat steps 3 to 5 there with Categories and Tags. Same guarantees: the fields hydrate with the columns hidden, and a visibility-only save leaves both taxonomies alone.
9. On an ad whose advertiser is set, block `/wp/v2/newspack_nl_advertiser` in DevTools and reload. Open Quick Edit: the Advertiser field must be read-only with the "could not be loaded" notice, not editable with an empty suggestion list. Unblock it, close and reopen: the field recovers without a page reload.
10. Run the tests: `pnpm --filter newspack-newsletters run test` (28 suites / 270 tests).

### Changes since approval

Review picked up two facts I had wrong and a few things worth tidying, and working through them surfaced a further problem of its own. All of it is in the ten commits on top of the approved tree.

**The newsletters list carried the same bug, so it is fixed here too.** #677 left that Quick Edit gating its save on a dirty diff alone, with no edited-ref and no read-only gate, and letting a present embed shadow the options list. Open it with the Categories column hidden, type a category before the terms fetch resolves, and save: the baseline was empty at the moment of the edit, so the request carried only the typed term and dropped the rest. It now has the identical guard, and the resolver both screens share lives in `utils/terms.js` rather than being written twice. Its three tests grew to seven, four of which fail against the old panel.

**The embed caps at 100, not 10.** I had read the terms collection's default `per_page`, but an embedded collection never gets the default: `embed_links()` sets it to the handler's schema maximum. Checked on the local site, an ad in 12 categories returns all 12 embedded, while the same URL dispatched directly returns 10. Corrected in the comments, the test names and the description above. The merge is still needed, since there is no embed at all when no term-backed column shows.

**The embed's cost was roughly doubled.** `embed_links()` dispatches once per embeddable link and caches by href, and a post carries one `wp:term` link per REST-visible taxonomy on its post type, so it is one dispatch per row per taxonomy. That is three on the ads list, and four on the newsletters list, which also carries the advertiser taxonomy (shared with the ads CPT) and, with Co-Authors Plus active, its `author` taxonomy. Both query builders now describe the mechanism rather than a count, since the number varies by site.

**A loading explanation, within the WP version floor.** A field held `disabled` while its options load drops out of the tab order with nothing to say why, so each one now carries a `help` string that lands in the field's `aria-describedby`.

`help` is WP 7.1 and up, though, and this plugin declares `Requires at least: 6.9`. Every call site therefore keeps `__experimentalShowHowTo={ false }` alongside it: on 7.1 `help` wins, because `computedHelp = help !== undefined ? help : defaultHelp` runs before the branch that reads the older prop; on 6.9 and 7.0 the fallback suppresses the default how-to text exactly as before. Two costs come with that, both deliberate: the 7.1 deprecation notice stays until the floor moves, and on 6.9/7.0 the wait itself goes unexplained, since `FormTokenField` offers no other described-by hook there. Applied to all six occurrences under `src/`, because the Send to sidebar renders one of them twice and a partial sweep would leave that screen inconsistent.

On 7.1 the wait says so, on both lists:

Newsletters Quick Edit with the Categories and Tags fields disabled while their terms load, each showing its own loading line

**The read-only gate measures the options list, not the merged selections.** This is the one worth a second pair of eyes. The guard asked whether the ad's stored term IDs could be *rendered*, and the embed alone could answer yes while the options fetch had failed outright. The field then rendered editable with an empty suggestion list: the token could be removed but not typed back, because both the suggestions and `__experimentalValidateInput` come from the options list, and the save wrote the taxonomy empty.

Emptying it is the dangerous direction. `class-ads.php:615` and `class-ads-placements.php:110` both guard with `if ( ! empty( $ad_categories ) )`, so an ad with no categories does not stop matching, it skips the category check and becomes eligible for every newsletter. Insertion runs inside the render each provider uploads as the campaign HTML, so a save landing before a send would email an ad to a list it was never sold against. `idsMissingFromOptions` now asks the question that actually governs editing, which is whether the *options list* accounts for every stored term. Three tests pin it and all three fail if the old predicate comes back.

**The shared term lists recover from a blip, without losing ground.** They are fetched once per screen mount, and now that they decide editability a single failed request would have left Advertiser and Ad placement degraded for the rest of the session. They are re-attempted when Quick Edit opens, but only when the row being opened has terms the current lists cannot explain, capped at one attempt per row: a healthy list is never refetched, so the common path costs nothing. Results are merged rather than replacing, because `fetchAllTerms` returns whatever it collected when a page request fails, and overwriting a good list with a short one would disable a field for an ad that is fine and empty the filter dropdowns with it.

**The newsletters panel is keyed by item.** The ads panel already was. Now that the newsletters panel's edited-refs gate its save, a reused instance could carry one row's selections into another's, which is the class of bug this PR exists to close.

**Smaller ones.** The terms hooks had unreachable `.catch()` blocks (`fetchAllTerms` resolves rather than rejecting), now `.finally()` throughout. The unavailable-terms notice is a shared component taking an explicit string, so its two redundant props are gone. `initialSelectionsForTaxonomy` and `selectionsFromIds` are no longer exported, since only `selectionsForTaxonomy` composes them now. One test claimed to cover the embed but passed with the embed source deleted, since its term was also in the options list; it now uses a term only the embed can supply. Two save tests asserted the same guarantee and are one `it.each`.

### Known limitation (pre-existing, not introduced here)

`fetchAllTerms` swallows request and pagination failures and returns what it collected, so a failed fetch is indistinguishable from an empty taxonomy. When an ad has no stored terms for a taxonomy, a failed fetch leaves the field enabled with an empty suggestion list, and token validation then rejects everything typed into it: usable-looking, but nothing can be selected. That behaviour predates this branch (the same fetch already fed the suggestions), and it cannot lose data, since a field the user could not edit is never sent. It is a missing signal rather than a wrong one. Fixing it properly means giving `fetchAllTerms` a `{ terms, complete }` result shape and updating the four screens that consume it, which belongs in its own PR.

### Other information:

**Worth a close look:** the read-only gate, and the save path on both lists. Steps 4 and 9 are the two regressions this guards against. Every new test was checked against the code it guards: revert the newsletters panel and four fail, revert the options-list gate and three fail, and stub out the embed source in the shared resolver and the two embed cases fail. They do bite.

Verified on a local site with the network selectively broken: a failed options fetch now yields a read-only field rather than an editable one with nothing selectable, the retry fires exactly once and recovers when the network heals, and the healthy path makes no extra requests at all.


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/fix/newsletters-ads-list-term-embed` (stash local work first).

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/index.js
```diff
@@ -57,9 +61,14 @@ function useFilterTerms() {
 				setTerms( {
 					advertisers: Array.isArray( advertisers ) ? advertisers : [],
 					placements: Array.isArray( placements ) ? placements : [],
+					hasLoaded: true,
 				} );
 			} )
-			.catch( () => {} );
+			.catch( () => {
```
adekbadek: **Suggestion:** A failed terms fetch gives no signal when the ad has no stored terms — When the fetch settles empty, the two outcomes give opposite affordances for the same underlying failure. An ad with stored advertiser IDs gets the read-only field plus the notice, which is clear. An ad with none gets an enabled field whose `validateAdvertiser` is built from an empty suggestions list, so every token typed is rejected with no message.

The PR description already names this and argues it cannot lose data, which holds - a field the user could not edit is never sent. The gap is that the user gets no signal at all in that second case.

Scope limit worth keeping if this is ever addressed: a site with genuinely zero advertisers reaches the same state, so gating on "settled and empty" needs different copy rather than the "could not be loaded" notice.

Separately, this `.catch()` is close to unreachable. `fetchAllTerms` wraps its own `apiFetch` in try/catch and breaks out of the loop (`utils/terms.js:35-37`), so it resolves rather than rejecting and `hasLoaded` in practice always comes from the `.then`. `useQuickEditCategories` solves the identical problem with `.finally`, which needs no dead branch - one form for both hooks would read better.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -28,19 +34,68 @@ function useQuickEditCategories() {
...
 		return () => {
 			cancelled = true;
 		};
 	}, [] );
-	return categories;
+	return { categories, hasLoaded };
+}
+
+// Resolve the post's stored term IDs through the embed first, then the
+// options list. The `wp:term` embed caps at 10 terms per taxonomy (the
```
adekbadek: **Nit:** The wp:term embed caps at 100 terms, not 10 — The embed does not cap at 10. When an embed link carries no `per_page`, core sets the sub-request to the matched handler's schema **maximum** rather than letting the default apply (`WP_REST_Server::embed_links()`, wp-includes/rest-api/class-wp-rest-server.php:814-819), and the terms collection schema gives `maximum => 100`. The default of 10 never reaches an embedded collection.

Verified on a WP 7.1 site: an ad with 12 categories returns all 12 in `_embedded['wp:term']`, while dispatching the same `/wp/v2/categories?post=<id>` URL directly returns 10.

No code change follows from this. Merging the embed with the fetched options list is still needed, because when no term-backed column is visible there is no embed at all, and the merge is still the right shape at 100. Only the number is wrong, and it appears in four places: here, at :149, and in the comment and test name at `quick-edit-panel.test.js` ("truncates at 10 terms"). The PR description's claim of a pre-existing bug where "a quick-edit save on an ad with more than 10 terms truncated it to 10" should read 100, which makes that bug considerably rarer than described. Worth correcting before it gets cited as justification on another screen.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -96,12 +193,21 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 			expiry_date: expiryDate,
```
adekbadek, line 190: **Suggestion:** The sibling newsletters list keeps the shape of the bug fixed here — The reasoning in this comment applies unchanged to the newsletters Quick Edit, which this PR leaves as #677 left it. There `handleSave` gates on `categoriesDirty` alone, with no edited-ref and no read-only gate (`screens/newsletters-list/quick-edit-panel.js:105`).

The path: open Quick Edit on a newsletter with the Categories column hidden, so the list query drops the term embed; type a category before `fetchAllTerms( '/wp/v2/categories' )` resolves; save. The baseline was empty at the moment of the edit, `hasEditedCategoriesRef` then blocks the sync effect, and the request carries only the typed term - dropping the newsletter's stored categories. That screen also uses `embedded.length ? embedded : selectionsFromIds( ... )`, the shadowing shape `selectionsForTaxonomy` replaced here.

This is out of scope for this diff and should not grow the PR. Raising it because the two screens now diverge on a data-loss guard, and because moving `selectionsForTaxonomy` into `utils/terms.js` - where the helpers it composes already live - would make the follow-up a call rather than a re-implementation.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -96,12 +193,21 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
...
-			meta,
-		};
+		// Only send a taxonomy the user actually touched. The edited ref is
+		// the load-bearing half: a dirty diff alone would also be true in
+		// the moment a late-resolving baseline overtakes the selection
+		// state, which would let a status-only save serialise the stale
+		// value. An untouched field must never overwrite what is stored.
+		const data = { meta };
+		if ( hasEditedAdvertiserRef.current && advertiserDirty ) {
+			data.newspack_nl_advertiser = [ ...advertiserSelections.map( s => s.id ), ...unresolvedAdvertiserIds ];
```
adekbadek, line 200: **Suggestion:** The unresolved-ids ride-along cannot execute — These three spreads always concatenate an empty array. A taxonomy with unresolved IDs sets `*Unavailable`, which sets `*ReadOnly` at :152-154, which passes `disabled` to the `FormTokenField`. A disabled field fires no `onChange`, so `hasEditedAdvertiserRef.current` can only become true on a render where `unresolvedAdvertiserIds` is already empty.

Nothing reopens that window afterwards: `useFilterTerms` sets `advertisers` and `hasLoaded` in a single `setTerms` and never again, `useQuickEditCategories` resolves once, and the item identity is pinned by `key={ quickEditItem.id }` in `index.js`, so `refresh()` does not replace it.

Keeping the spreads as a backstop is defensible - the sibling newsletters-list panel has no read-only gate, so the same spread is reachable there. The comment at :89 is the part worth changing: it states a behaviour this screen cannot reach. The framing in the test at `quick-edit-panel.test.js:308-310` is closer to what a reader needs here.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -143,32 +249,59 @@ export default function AdsQuickEditPanel( { item, advertisers, placements, onCl
 				label={ __( 'Advertiser', 'newspack-newsletters' ) }
 				value={ advertiserTokens }
 				suggestions={ advertiserSuggestions }
-				onChange={ next => setAdvertiserSelections( resolveTokens( next, advertiserSelections, advertisers ) ) }
+				disabled={ advertiserReadOnly }
```
adekbadek, line 250: **Suggestion:** Fields disappear from the tab order while their options load — The three fields render `disabled` until their options settle, with no loading affordance. `disabled` removes a control from the tab order, so a keyboard or screen-reader user who opens Quick Edit before the options land finds no Advertiser, Ad placement or Categories field at all and nothing explaining why. `useQuickEditCategories` refetches `/wp/v2/categories` on every panel open, paginated at 100 per page, so on a site with many categories that window is not always brief.

`FormTokenField` takes a `help` prop that lands in the field's `aria-describedby`, which would both carry the explanation and associate the unavailable message with the field it describes - it currently renders as a sibling node with no association. As of WP 7.1 `help` is also the documented replacement for the `__experimentalShowHowTo={ false }` these fields already pass; that deprecation warning fires in the console today. The `__experimentalShowHowTo` usage predates this PR and appears in four other files under `src/`, so it is not this PR's debt - only worth mentioning because `help` resolves both at once.

Scope limit: a disabled input is out of the tab order, so the missing association only reaches browse-mode screen-reader users, and the `Notice` already announces once on mount. This is a refinement, not a defect.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.test.js
```diff
@@ -132,3 +151,172 @@ describe( 'AdsQuickEditPanel status control', () => {
...
+		_embedded: { 'wp:term': [ [ { id: 10, name: 'Acme', taxonomy: 'newspack_nl_advertiser' } ] ] },
+	} );
+
+	it( 'hydrates the fields from raw term IDs when the embed is absent', async () => {
+		renderPanel( withRawTerms() );
+		expect( await screen.findByText( 'Acme' ) ).toBeInTheDocument();
+		expect( screen.getByText( 'Header' ) ).toBeInTheDocument();
+	} );
+
+	it( 'still reads embedded terms when they are present', async () => {
```
adekbadek, line 185: **Suggestion:** This test passes with the embed source deleted — "still reads embedded terms when they are present" does not isolate the embed. `withEmbeddedTerms()` sets `newspack_nl_advertiser: [ 10 ]` with an embedded term `{ id: 10, name: 'Acme' }`, and `ADVERTISERS` already contains `{ id: 10, name: 'Acme' }` - so `selectionsFromIds( [ 10 ], ADVERTISERS )` renders "Acme" on its own. Removing `initialSelectionsForTaxonomy` from `selectionsForTaxonomy` entirely leaves this green, which is the guarantee the name claims.

Giving the embedded term an id present in `newspack_nl_advertiser` but absent from `ADVERTISERS` (say `[ 55 ]` with `{ id: 55, name: 'Beta Corp' }`) makes the embed the only possible source. That keeps the field editable, since `unresolvedIds` measures against the merged selections rather than the options list.

The rest of the new suite does bite - the loading-gate and save-path cases both fail on revert.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/build-query.js
```diff
@@ -41,6 +45,11 @@ const SORT_FIELD_TO_ORDERBY = {
 };
 
 export function buildQueryParams( view = {} ) {
+	// Term embeds cost ~2 internal REST dispatches per row per taxonomy,
```
adekbadek: **Nit:** The dispatch count is roughly double the real cost — "~2 internal REST dispatches per row per taxonomy" is not what the mechanism produces. `WP_REST_Server::embed_links()` dispatches once per embeddable link and caches by href, and a post carries one `wp:term` link per taxonomy - so it is one dispatch per row per taxonomy, three per row on this screen.

The justification for the change is sound either way; the multiplier just overstates it by about 2x.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -28,19 +34,68 @@ function useQuickEditCategories() {
...
+			} );
 		return () => {
 			cancelled = true;
 		};
 	}, [] );
-	return categories;
+	return { categories, hasLoaded };
+}
+
+// Resolve the post's stored term IDs through the embed first, then the
```
adekbadek: **Nit:** Comment states precedence opposite to the code — The comment reads "through the embed first, then the options list", but the code spreads `selectionsFromIds( ids, options )` first and `embedded` second, with precedence coming from `Map` last-entry-wins. Reading order and precedence order are opposite.

Immaterial today - both sources yield `{ id, name }` from the same REST `name` for the same term - so this is about the next reader. A later edit that reorders the spreads or switches to an insertion loop would flip precedence with no visible signal.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.js
```diff
@@ -28,19 +34,68 @@ function useQuickEditCategories() {
...
+	const byId = new Map( [ ...selectionsFromIds( ids, options ), ...embedded ].map( selection => [ selection.id, selection ] ) );
+	return ids.map( id => byId.get( id ) ).filter( Boolean );
+};
+
+// A settled options list that still can't account for every stored term
+// means the field would misrepresent the ad, so it goes read-only rather
+// than showing a quietly wrong value.
+function TermsUnavailableNotice( { children } ) {
+	return (
+		<Notice status="warning" isDismissible={ false } politeness="polite" spokenMessage={ children }>
```
adekbadek: **Nit:** Notice props restate the component defaults — `politeness="polite"` and `spokenMessage={ children }` change nothing here. In the pinned `@wordpress/components`, `Notice` already defaults `spokenMessage` to `children`, and `getDefaultPoliteness( 'warning' )` returns `'polite'`. Dropping both behaves identically.

Keeping them pins behaviour that upstream would have to change deliberately, which is a defensible reason to leave them - just worth being a decision rather than an accident.

plugins/newspack-newsletters/src/admin-shell/screens/ads-list/quick-edit-panel.test.js
```diff
@@ -132,3 +151,172 @@ describe( 'AdsQuickEditPanel status control', () => {
...
+		await act( async () => {
+			release();
+		} );
+
+		await waitFor( () => expect( screen.getByLabelText( 'Categories' ) ).not.toBeDisabled() );
+		expect( screen.getByText( 'Cat 11' ) ).toBeInTheDocument();
+		expect( visibleNotices() ).not.toContain( CATEGORIES_UNAVAILABLE );
+	} );
+
+	it( 'still refuses to send the unshowable categories on save', async () => {
```
adekbadek: **Nit:** Two tests assert the same guarantee — These two cases assert the same thing - a taxonomy whose stored IDs cannot all be resolved is absent from a status-only save - differing only in which taxonomy carries the unresolvable id and which gate produced the read-only state. The second's own comment concedes its path is unreachable while the read-only gate holds.

One case, or an `it.each` over the two taxonomies, states the spec as clearly. Both bite today, so this is readability rather than coverage.

adekbadek (approved):
Works fine. Some suggestions and nits.

Review provenance: 5 of 5 reviewer passes (lineup: deep, standards, wp-expert, codex, opencode:kimi-k3); release risk: LOW.

PR #705, opened 2026-07-23
feat(content-gate): accept dash IP ranges and warn on invalid institution IP entries (NPPD-1509)

adekbadek:
### All Submissions:

### ⚠️ Release notes — two behavior changes to call out

**1. Dash entries already stored on production sites become live access grants the moment this deploys.** This branch re-evaluates existing institution meta. A dash range a publisher typed months ago is inert today (it falls through to `filter_var` and is dropped) and starts granting access after deploy, with no re-save, no admin action and no notice. That is the point of the feature — the ticket exists because publishers wrote ranges the parser rejected — but some of those stored strings are likely typos that have been safely doing nothing, so the effective allowlist of an existing institution can change based on text nobody has looked at since it was typed. Publishers with institutions configured should re-check Audience > Access Control > Institutions after the release; the editor now flags anything that still cannot grant access.

**2. The `institutional-access/ip-allowlist` REST field `ip_ranges` gains a third string shape.** It is a plain `string[]` carrying `10.0.0.5`, `10.0.0.0/24` **and now** `10.0.0.1-10.0.0.9`, with no discriminator field. No consumer exists anywhere in the workspace (verified by grep across newspack-manager and newspack-manager-admin; newspack-extended-access has no checkout here, so external integrations are unverifiable from this repo). A consumer that only understands IP-plus-CIDR ignores dash entries, which fails *closed* — a reader on campus sees the gate rather than getting free access. A site running such a consumer can rewrite or drop entries with the existing `newspack_content_gate_ip_allowlist` filter (now documented in the endpoint docblock) without restricting what admins may type. A discriminated/normalized REST shape is deliberately **not** added here.

Institution IP access rules previously matched only CIDR blocks and single IPv4 addresses, while the editor accepted any text with no validation — a raw range like `203.0.113.0-203.0.113.255` was stored and silently never matched.

- **Server:** `IP_Access_Rule` now also parses dash ranges (`a.b.c.d-e.f.g.h`, inclusive boundaries, whitespace-tolerant). CIDR and single-IP behavior is unchanged. Invalid entries (including reversed ranges, end < start) remain inert: never match, never warn/fatal. Stored meta is never rewritten. Range comparisons go through `sprintf( '%u', ip2long( … ) )`, so a range straddling `127.255.255.255` behaves the same on a 32-bit build.
- **UI:** the institution editor validates entries client-side (`utils.ts`, mirroring the PHP parser including PHP's exact `trim()` charset) and shows non-blocking warnings. Saving is not blocked, matching the editor's existing validation UX; a reversed range is deliberately warned rather than auto-swapped (a typo shouldn't silently become a larger grant). Three warning classes:
  - **Invalid entries** that can never grant access (list capped at 3 with "and N more").
  - **Confusable characters** — an entry that would be valid but for a hyphen look-alike (en/em dash, minus sign, Unicode and non-breaking hyphen, figure dash, fullwidth hyphen), a Unicode space, or a byte order mark. These come from Word/Google Docs/Outlook autocorrect, PDFs, CJK input methods and Windows text files, and are visually identical on screen, so the warning names the offending character.
  - **Over-broad entries** — a parseable entry covering more than 65,536 addresses, whether written as a dash range or as a CIDR mask. One wrong digit in an end address or a `/8` typed where `/28` was meant authorizes millions of addresses and reads as fine at a glance.
- **When warnings are computed:** on blur (and on load for stored values), never per keystroke, so a half-typed range is never called invalid. Saving re-checks the value being saved and raises the same warnings as a wizard snackbar, which outlives the navigation to the list — otherwise an admin who pastes a list and clicks Save straight away loses the message with the unmounted view. The warnings live in a `role="status"` region the input points at via `aria-describedby`, with `aria-invalid` on the input — the debounce is what makes the live region safe (per-keystroke recomputation in a live region would announce partial input).
- **Client/server parity is pinned by a shared fixture**, `tests/fixtures/ip-range-validation-cases.json` (44 cases), consumed by both the jest suite and a PHPUnit data provider. The dangerous direction is the client calling something valid that the server drops — the exact silent failure this PR exists to fix — so drift is now a CI failure rather than a support ticket (verified: flipping one fixture case fails both suites).

Closes NPPD-1509.

1. Audience > Access Control > Institutions: create an institution with IP entries `203.0.113.0-203.0.113.255, 999.1.2.3, 10.0.0.9-10.0.0.1` — a warning lists the two invalid entries; save succeeds.
2. Type a valid range one character at a time: no warning appears while typing; tab out with an incomplete entry and the warning appears.
3. Paste `192.168.1.1 – 192.168.1.9` (en dash) and blur: a second warning names "en dash (–)".
4. Enter `203.0.113.0-243.0.113.255` and blur: the over-broad warning fires (the entry is still valid and saveable).
5. Replace it with `0.0.0.0/0` and blur: the same warning fires for the CIDR mask. A `/16` or narrower does not fire it.
6. Paste an invalid entry and click Save without leaving the field: the editor navigates to the list and a snackbar repeats the warning.
7. With a request IP inside the range (e.g. `203.0.113.100` via trusted proxy header or `wp eval`), institutional access matches; `203.0.114.50` does not; `10.0.0.5` (inside the reversed range) does not.
8. `wp eval` the REST allowlist (`/institutional-access/ip-allowlist`): valid dash ranges are included, invalid entries dropped.

### Other information:

* [x] Have you successfully run tests with your changes locally? (`OK (122 tests, 254 assertions)` for `--group Access_Rules`; jest 47 suites / 504 tests; PHPCS + ESLint clean)

**Disclosures:**
- IPv4 only, matching the existing matcher (`FILTER_FLAG_IPV4` throughout): IPv6/IPv4-mapped visitors don't match today and that's unchanged — worth a separate ticket if an institution needs IPv6.
- Server-side normalization of confusable characters (mapping the hyphen look-alikes to `-` in `parse_ip_ranges()`) is deliberately **not** done: it would turn previously-inert stored entries into live grants on deploy, compounding behavior change 1 above. The client names the character instead, so the admin fixes the entry knowingly.
- A lone `0` entry is warned client-side though PHP silently drops it (warning is the safer side).


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-1509-institutions-clarify-ip-range-notation-requirements` (stash local work first).

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
```diff
@@ -300,7 +300,7 @@ public static function get_ip_allowlist_schema() {
 					'readonly'    => true,
 				],
 				'ip_ranges' => [
-					'description' => __( 'Validated IPv4 addresses or CIDR blocks granting access.', 'newspack-plugin' ),
+					'description' => __( 'Validated IPv4 addresses, CIDR blocks, or dash ranges granting access.', 'newspack-plugin' ),
```
dkoo, line 311: **Suggestion:** REST allowlist gains a third string shape with no discriminator — The `ip_ranges` response field now carries three different notations inside one untyped `string[]`, and the only signal to consumers is this English schema description — but consumers read data, not schemas. I confirmed there is no reference to this route or field anywhere in the workspace, including `newspack-manager` and `newspack-manager-admin`; `newspack-extended-access` has no checkout here, so it is genuinely unverifiable from this workspace, as is any external integration.

A consumer written against IP-plus-CIDR will most likely just fail to match dash entries, which is fail-closed (a reader on campus hits the paywall) rather than fail-open — so this is not urgent. But it is much cheaper to fix now than after a partner integrates: either emit a discriminated shape, or add a machine-readable normalized field alongside the human-readable string. At minimum it deserves a changelog line, and mentioning the existing `newspack_content_gate_ip_allowlist` filter in the docblock would make the escape hatch discoverable for a site with a legacy consumer.

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
```diff
@@ -649,6 +650,18 @@ private static function parse_ip_ranges( $raw ) {
 					continue;
 				}
 				$valid[] = $subnet . '/' . $bits;
+			} elseif ( strpos( $token, '-' ) !== false ) {
```
dkoo, line 680: **Suggestion:** Already-stored dash entries become live grants on upgrade — This branch re-evaluates institution meta that already exists on production sites. A dash entry a publisher typed months ago is inert today — it falls through to the final `filter_var` and is dropped — and becomes a live access grant the moment this deploys, with no re-save, no admin action and no notice. That is the intended behavior of the feature, but it means the effective allowlist of every existing institution can change based on text nobody has looked at since it was typed.

Worth an explicit line in the release notes. Given the ticket exists precisely because publishers wrote ranges the parser rejected, some of those stored strings are likely typos that have been safely doing nothing.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -232,6 +234,7 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
 	const description = institution.excerpt.raw;
 	const meta = institution.meta || EMPTY_INSTITUTION.meta;
 	const { np_institution_email_domain: emailDomain, np_institution_ip_range: ipRange, np_institution_reader_data: readerData } = meta;
+	const invalidIpEntries = getInvalidIpRangeEntries( ipRange || '' );
```
dkoo: **Suggestion:** The warning fires mid-keystroke — `getInvalidIpRangeEntries` runs unconditionally on every render against live `ipRange` state, so typing `142.74.1.0-142` puts "This entry is invalid and will never grant access" on screen while the user is halfway through typing a perfectly correct entry. The PR describes this as matching the editor's existing validation UX, but the other two rule fields here have no field-level validation at all — so this establishes the pattern rather than following one.

Deferring evaluation to `onBlur`, or debouncing into a separate state value, would keep the warning's signal intact while removing the noise. As written, the notice is most visible precisely when it is least true, which is the fastest way to teach people to ignore it.

**Confirmed in the browser**: typing a valid range one character at a time renders "This entry is invalid and will never grant access: 142.74.1.0-142" after the 14th keystroke; the notice only clears once the entry is completed to `142.74.1.0-142.74.1.255`. Screenshot attached.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +310,37 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
 						<CardBody size="small">
 							<TextControl
-								label={ __( 'IPs / CIDR blocks (comma-separated)', 'newspack-plugin' ) }
+								label={ __( 'IPs, CIDR blocks, or IP ranges (comma-separated)', 'newspack-plugin' ) }
 								value={ ipRange }
 								onChange={ ( val: string ) => updateMeta( 'np_institution_ip_range', val ) }
-								placeholder="192.168.1.0/24, 10.0.0.5"
+								placeholder="192.168.1.0/24, 142.74.1.0-142.74.1.255, 10.0.0.5"
 							/>
+							{ invalidIpEntries.length > 0 && (
```
dkoo: **Suggestion:** Nothing bounds how broad a dash range may be — Dash notation makes an over-broad grant much easier to fat-finger than CIDR does. `142.74.1.0-242.74.1.255` — one wrong digit in the end address — authorizes roughly 1.6 billion addresses and passes both the PHP parser and this new client validator silently. The equivalent CIDR mistake requires deliberately writing a small prefix length, which reads as obviously wrong at a glance.

This PR already accepts the principle that a suspicious-but-parseable range deserves a non-blocking warning — that is exactly the reversed-range case — so extending the same treatment to an unusually wide range would be consistent with the design and cheap: `ipv4ToNumber( end ) - ipv4ToNumber( start ) > 65535`. Since this feature ungates paid content, an over-broad range is a revenue leak nobody is likely to notice.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +310,37 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+								<Notice
```
dkoo: **Suggestion:** The warning isn't announced to screen readers — The `Notice` renders as a plain `<div>` with no `role`, no `aria-live`, and no `aria-describedby` association back to the `TextControl` above it. Because it is inserted and removed dynamically as the admin types, a screen-reader user gets no indication it exists — they can enter an invalid range, receive no feedback, and save.

The smallest correct fix is the `help` prop on `TextControl`, which renders a `<p>` and wires `aria-describedby` on the input for free. If the Notice styling is preferred, `role="status"` plus a debounce would be needed — and those two go together, since a live region combined with the current per-keystroke recomputation would announce partial input like "10.0.0. is invalid" on every character.

**Confirmed in the browser**: the rendered notice is a plain `div.newspack-notice.newspack-notice__is-warning` with no `role`, no `aria-live` and no `aria-atomic`, and the `TextControl` input carries no `aria-describedby`, `aria-invalid` or `aria-errormessage`. The only live regions in the DOM are `wp.a11y`’s two generic (empty) speak regions, which this code never calls — so nothing is announced when the warning appears or disappears.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +310,37 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+									isWarning
+									noticeText={ sprintf(
+										/* translators: %1$s: comma-separated list of invalid IP entries. %2$s: single IP example. %3$s: CIDR block example. %4$s: IP range example. */
+										_n(
+											'This entry is invalid and will never grant access: %1$s. Use a single IPv4 address (e.g. %2$s), a CIDR block (e.g. %3$s), or an IP range from lowest to highest (e.g. %4$s).',
+											'These entries are invalid and will never grant access: %1$s. Use a single IPv4 address (e.g. %2$s), a CIDR block (e.g. %3$s), or an IP range from lowest to highest (e.g. %4$s).',
+											invalidIpEntries.length,
+											'newspack-plugin'
+										),
+										invalidIpEntries.join( ', ' ),
```
dkoo: **Suggestion:** The warning can't describe the cases it's best at catching — The offending entries are printed verbatim via `invalidIpEntries.join( ', ' )`, which makes the warning unactionable for exactly the errors most likely to occur. Word, Google Docs and Outlook all autocorrect ` - ` into an en dash (U+2013), so `192.168.1.1 – 192.168.1.9` is the single most probable real-world paste error for a range — and the non-breaking-space cases this PR deliberately encodes at `utils.test.js:48-57` are the same shape. Rendered in the browser, those strings are visually identical to correct entries, so the admin reads "this entry is invalid" next to something that looks perfectly fine and has no way to act on it.

Two options. The cheap one: detect this class specifically — an entry that becomes valid after `String.prototype.trim()` but not after `phpTrim()`, or that contains a non-ASCII dash — and emit a distinct message naming the character. The more robust one: normalize in `parse_ip_ranges()` (map U+2013/U+2014/U+2212 to `-`, strip Unicode whitespace) with the client mirroring it, which is safe from an access-control standpoint because it can only make an inert entry behave as the admin visibly typed it, never widen beyond the typed text.

Smaller point on the same line: the join is uncapped, so a large pasted list renders as one very long notice — worth truncating to the first few with an "and N more".

**Confirmed in the browser** (isolated env, Chrome): entering `192.168.1.1 – 192.168.1.9` (U+2013) alongside a trailing-NBSP `10.0.0.5 ` fires the warning, and the notice reprints both verbatim — on screen the en-dash entry is indistinguishable from the hyphenated range the same sentence offers as the fix, and the NBSP entry renders as a plain, correct-looking `10.0.0.5`. An admin reading that notice has no way to tell what is wrong with either. Screenshot attached.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Split on the first occurrence of a separator, like PHP's `explode( $sep, $str, 2 )`.
+ */
+function splitOnce( value: string, separator: string ): [ string, string ] {
+	const index = value.indexOf( separator );
+	return [ value.slice( 0, index ), value.slice( index + separator.length ) ];
+}
+
+function isValidEntry( token: string ): boolean {
```
dkoo, line 93: **Suggestion:** The two validators can drift with nothing to catch it — `isValidEntry()` is a hand-maintained reimplementation of `parse_ip_ranges()`, down to PHP's `trim()` charset and `filter_var`'s leading-zero rejection. It is faithful today — I checked each branch against PHP directly — but parity exists only as two independently written test lists, so the next change to either side desynchronizes silently.

The failure mode that matters is the permissive direction: if the client ever says "valid" for something the server drops, that is exactly the silent failure this PR was written to fix, reintroduced. A shared fixture — a JSON list of `{ entry, valid }` cases consumed by both the jest suite and a PHPUnit data provider — would make drift a CI failure rather than a support ticket.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
```diff
@@ -88,6 +88,139 @@ public function test_cidr_tolerates_whitespace_around_slash() {
...
+		$this->assertTrue( IP_Access_Rule::ip_matches_ranges( '10.0.0.5', '10.0.0.5-10.0.0.5' ) );
+		$this->assertFalse( IP_Access_Rule::ip_matches_ranges( '10.0.0.6', '10.0.0.5-10.0.0.5' ) );
+	}
+
+	/**
+	 * Test whitespace around the dash separator is tolerated, mirroring the
+	 * CIDR slash behavior.
+	 */
+	public function test_dash_range_tolerates_whitespace_around_dash() {
+		$this->assertTrue( IP_Access_Rule::ip_matches_ranges( '10.0.0.5', '10.0.0.1 - 10.0.0.9' ) );
```
dkoo, line 145: **Suggestion:** Two high-value edge cases are unasserted on both sides — The existing boundary coverage (inclusive endpoints, octet-spanning, single-address, reversed, whitespace, empty) is good. Two gaps stand out for a security-sensitive classifier:

**Token-classification ordering.** Neither suite asserts a token containing *both* separators. I verified `10.0.0.0/24-10.0.0.5` and `10.0.0.1-10.0.0.9/24` are inert in PHP and flagged in JS today, but nothing pins that — and a future reorder of the `if`/`elseif` chain is exactly the kind of refactor that would silently change how those parse.

**The `ip2long` sign boundary.** `127.255.255.255-128.0.0.1` and `0.0.0.0-255.255.255.255` are the two ranges where a signed/unsigned mistake would surface, and neither is asserted.

On the JS side, the en-dash case from the warning-message comment is also worth an assertion, since it is the most likely real-world input.

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
```diff
@@ -649,6 +650,18 @@ private static function parse_ip_ranges( $raw ) {
...
+				list( $start, $end ) = explode( '-', $token, 2 );
+				$start = trim( $start );
+				$end   = trim( $end );
+				if ( ! filter_var( $start, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) || ! filter_var( $end, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) ) {
+					continue;
+				}
+				// A reversed range (end < start) is most likely a typo; treat it as invalid rather than silently swapping the bounds.
+				if ( ip2long( $start ) > ip2long( $end ) ) {
```
dkoo: **Nit:** The reversed-range guard assumes 64-bit PHP — This compares raw `ip2long()` return values, which are signed, whereas the pre-existing CIDR branch uses a bitmask comparison that is signedness-agnostic. I confirmed `PHP_INT_SIZE` is 8 in the dev container and `ip2long( '255.255.255.255' )` returns `4294967295`, so this is correct as deployed.

On a 32-bit build, though, `ip2long( '200.0.0.0' )` is negative and a straddling range like `10.0.0.0-200.0.0.0` would be rejected as "reversed". That fails closed, but it would also be a case where the client validator (unsigned arithmetic at `utils.ts:29-31`) says valid while PHP silently drops it — the exact class of divergence this validator exists to prevent. `sprintf( '%u', ip2long( $x ) )` removes the platform assumption for free.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +310,37 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+										'10.0.0.5',
+										'192.168.1.0/24',
+										'142.74.1.0-142.74.1.255'
```
dkoo: **Nit:** A real routable netblock is used as the example — `142.74.1.0-142.74.1.255` is a real, globally routable ARIN block, and it now appears in the placeholder, this translated warning string, the `Institution::create()` docblock, the parser docblock and several tests. Every other example in this UI is RFC 1918 private space.

The value almost certainly came from the customer ticket behind NPPD-1509, which is a second reason to change it: an RFC 5737 documentation range like `203.0.113.0-203.0.113.255` avoids both pointing publishers at someone else's netblock and carrying a specific customer's network detail into translated product strings.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+function ipv4ToNumber( value: string ): number {
+	return value.split( '.' ).reduce( ( acc, octet ) => acc * 256 + Number( octet ), 0 );
+}
+
+/**
+ * Split on the first occurrence of a separator, like PHP's `explode( $sep, $str, 2 )`.
+ */
+function splitOnce( value: string, separator: string ): [ string, string ] {
+	const index = value.indexOf( separator );
```
dkoo, line 80: **Nit:** `splitOnce` returns nonsense when the separator is absent — `indexOf` yields `-1`, so `value.slice( 0, -1 )` silently drops the last character and `value.slice( 0 )` returns the whole string. Both call sites guard with `includes()`, so it cannot fire today — but the helper reads as general-purpose and its docblock advertises parity with PHP's `explode( $sep, $str, 2 )`, which returns a one-element array in that case. An early `if ( index === -1 ) return [ value, '' ];` costs one line and removes the trap.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
```diff
@@ -88,6 +88,139 @@ public function test_cidr_tolerates_whitespace_around_slash() {
...
+			[ 'ip_range' => '142.74.1.0 - 142.74.1.255, 10.0.0.9-10.0.0.1, 10.0.0.1-, 192.168.1.0/24' ]
+		);
+		$this->assertIsInt( $inst_id );
+		delete_transient( \Newspack\Institution::TRANSIENT_KEY );
+
+		$route    = '/' . NEWSPACK_API_NAMESPACE . IP_Access_Rule::REST_ROUTE_IP_ALLOWLIST;
+		$request  = new WP_REST_Request( 'GET', $route );
+		$response = rest_do_request( $request );
+
+		$entry = null;
```
dkoo, line 296: **Nit:** Assert the response status before iterating it — The loop indexes `$item['id']` straight out of `$response->get_data()`. If the permission check ever regresses, that iterates an error payload and the resulting failure points at an array-key notice rather than at the 403 that actually broke. A `$this->assertSame( 200, $response->get_status() );` before line 213 makes the failure self-describing. The `assertNotNull` at line 220 does catch the case, just opaquely.

dkoo (commented):
Release risk: ELEVATED — the `institutional-access/ip-allowlist` REST payload gains a third string shape in an existing untyped field, and at least one potential consumer (`newspack-extended-access`, plus any external integration) could not be checked from this workspace; runtime verification is now complete on both sides and found nothing that changes the tier — it is driven by the contract change, not by what was left untested.

No major blockers. This is careful, well-documented work: the matcher re-parses through the validator on every path, so it only ever sees trimmed, validated entries; reversed ranges are rejected rather than silently swapped; and the comment explaining why `String.prototype.trim()` is the wrong function here is exactly the kind of note that stops someone from reintroducing the bug later. Tests are unusually thorough for a change this size, and I confirmed the trickier premises directly against PHP 8.3 — `filter_var` does reject leading-zero octets, `trim()` does leave a non-breaking space in place, and `ip2long` is unsigned on 64-bit.

I also verified the server-side half at runtime in an isolated environment seeded with the exact mixed entry string from your test plan: `142.74.1.100` matches, `142.74.2.50` does not, `10.0.0.5` inside the reversed range does not, CIDR and single-IP matching are unchanged, and the REST allowlist returns dash ranges while dropping invalid entries. I then drove the editor in a browser against the same env: the warning names exactly the invalid entries, uses the right singular/plural form, disappears when the rule is toggled off or the entries are all valid, and does not block saving. Three findings below were confirmed on screen rather than argued — the en-dash/NBSP entries that are flagged but rendered indistinguishably from valid ones, the mid-keystroke firing, and the notice not being announced to screen readers. Nothing was refuted.

The things worth a look before merge are about blast radius rather than correctness. Two stand out. First, dash entries publishers already typed and abandoned are inert today and become live grants the moment this deploys — worth a release-note line and, ideally, a read of stored `np_institution_ip_range` values containing a dash across gate-enabled sites before it ships. Second, nothing bounds how wide a dash range may be: one wrong digit turns a 256-address grant into a 1.6-billion-address one, and unlike a CIDR typo it does not look wrong. The rest are smaller: the two validators can drift with nothing to catch it, the warning cannot describe the invisible-character cases it is best at detecting, and the notice is not announced to screen readers.

One housekeeping note: the PR body says 11 JS cases; the file has 9 `it` blocks (25 assertions). The PHP count of 10 is accurate.

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
adekbadek: Documented rather than restructured. The endpoint docblock now spells out the three notations and points at `newspack_content_gate_ip_allowlist` as the escape hatch for a site with a legacy consumer. Both this and the upgrade-behavior change are called out at the top of the PR description — the repo has no hand-edited changelog (CHANGELOG.md is generated by semantic-release from commit subjects), so the PR body is what feeds the release notes.

A discriminated or normalized REST field is deliberately deferred: it is a shape change with no known consumer, and Adam will file a follow-up ticket if it is worth tracking.

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
adekbadek: Called out prominently at the top of the PR description as release-note item 1, in your framing: existing dash entries go live on deploy with no re-save and no notice, and some of them are likely typos that have been safely doing nothing. The note asks publishers with institutions configured to re-check the screen after the release — the editor now flags anything that still cannot grant access.

This is also the reason server-side normalization of en dashes is not being added (see the `edit.tsx:337` thread): it would widen the same blast radius.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Fixed together with the a11y comment below, since as you noted they are one change. Warnings now come from `ipRangeAnalysis` state computed on blur (and on load for stored values), never during render; editing clears the current warnings so a stale one cannot sit next to text the admin has already corrected. `onBlur` reads `event.target.value` rather than the render closure, so a paste-then-tab in one tick still validates the committed value.

Verified in the env: typing `203.0.113.0-203.0.113.255` one character at a time shows nothing at any keystroke; tabbing out at `203.0.113.0-203` warns, and completing it clears the warning.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Implemented with your formula. A valid dash range where `ipv4ToNumber( end ) - ipv4ToNumber( start ) > 65535` gets its own non-blocking warning, distinct from the invalid-entry one: "This range covers more than 65,536 addresses: … Check the end address — an over-broad range grants access far beyond the institution." A full /16 is left alone; CIDR blocks and invalid ranges are out of scope for this warning. Covered by jest cases and verified in the browser with `203.0.113.0-243.0.113.255`.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Kept the `Notice` and made it announceable, which is safe now that recomputation happens on blur rather than per keystroke. The notices sit in a container that is always rendered with `role="status"`, so the live region exists before content is inserted; the input carries `aria-describedby` pointing at it plus `aria-invalid`. Verified in the env: `aria-describedby` resolves to the `role="status"` container and `aria-invalid` flips with the warning.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Took the cheap client-side option, not the normalizing one. An invalid entry that becomes valid after mapping U+2013/U+2014/U+2212 to `-` and normalizing Unicode spaces now produces a second, distinct warning naming the character: "Entries contain characters that look standard but are not: en dash (–), non-breaking space. Retype them using a plain hyphen (-) and regular spaces." A character is only named when replacing it actually fixes the entry, so `banana – 10.0.0.9` is not blamed on the dash.

Normalizing in `parse_ip_ranges()` is deliberately not done: server-side normalization would turn previously-inert stored entries into live access grants on deploy, which compounds the upgrade-behavior concern you raise on `class-ip-access-rule.php:653`. Naming the character keeps the fix in the admins hands and visible.

Smaller point applied too: the echoed list is capped at the first three plus "and N more".

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
adekbadek: Implemented. `tests/fixtures/ip-range-validation-cases.json` holds 38 `{ entry, valid }` cases and is consumed by both suites — a jest `it.each` over `analyzeIpRangeEntries` and a PHPUnit data provider over `parse_ip_ranges()` via reflection — so drift fails CI on one side or the other rather than reaching a publisher.

Checked the guard works: flipping a single fixture case red fails both suites (`Tests: 1 failed` in jest, `Newspack_Test_IP_Access_Rule::test_shared_validation_fixture_parity with data set "en dash instead of a hyphen"` in PHPUnit).

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
adekbadek: Both classes added on both sides. `test_token_with_both_separators_is_inert()` pins `10.0.0.0/24-10.0.0.5` and `10.0.0.1-10.0.0.9/24` as inert (with a docblock note that the `/` branch is checked first), and `test_dash_range_across_ip2long_sign_boundary()` covers `127.255.255.255-128.0.0.1`, `0.0.0.0-255.255.255.255` and a straddling `10.0.0.0-200.0.0.0`. Both classes are in the shared fixture too, and the JS suite has matching cases plus the en-dash entry.

plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php
adekbadek: Applied. Added a private `ip_to_unsigned()` helper using `sprintf( "%u", ip2long( $ip ) )` and used it for both the reversed-range guard and the dash comparison in `ip_matches_ranges()` — the latter would have had the same 32-bit problem at match time. The CIDR bitmask path is untouched, since it is already signedness-agnostic. `test_dash_range_across_ip2long_sign_boundary()` pins the behavior.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Applied everywhere the block appeared: placeholder, translated warning string, `Institution::create()` docblock, `parse_ip_ranges()` docblock and the PHP/JS tests now use `203.0.113.0-203.0.113.255` (RFC 5737). The PR description and testing steps were updated too, so the customer netblock is not left in the repo history-facing copy.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
adekbadek: Applied as suggested: `splitOnce` now returns `[ value, "" ]` when the separator is absent, matching `explode( $sep, $str, 2 )`.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
adekbadek: Applied: `assertSame( 200, $response->get_status() )` before the loop, with a comment explaining that iterating an error payload would otherwise fail with an array-key notice instead of naming the permission regression.

adekbadek:
On the JS test count: the "11 JS cases" line is gone from the description — the blur/confusables/over-broad commit rewrote that section, and the suite grew with it. `utils.test.js` is now 19 `it` blocks / 47 `expect()` calls.

I re-ran everything the description now claims rather than trusting the numbers:

- shared fixture: 38 cases in `tests/fixtures/ip-range-validation-cases.json`
- `n test-php --group Access_Rules` → `OK (116 tests, 248 assertions)`
- `pnpm run test` (newspack-plugin) → 46 suites / 486 tests passed

So every count in the body matches what actually runs; no edit was needed.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -62,13 +112,18 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
 	const [ isDirty, setIsDirty ] = useState( false );
 	const [ imageData, setImageData ] = useState< { id: number; url: string } | null >( null );
 	const [ isLoadingImage, setIsLoadingImage ] = useState( false );
+	// Evaluated when the IP range field is committed (on blur, and on load for
+	// stored values) rather than on every keystroke: a half-typed range is not
+	// yet wrong, and warning about it teaches admins to ignore the warning.
+	const [ ipRangeAnalysis, setIpRangeAnalysis ] = useState( EMPTY_IP_RANGE_ANALYSIS );
```
dkoo, line 192: **Suggestion:** The blur/state/aria wiring has no test — `utils.ts` is covered well (the jest suite plus the shared fixture), but everything the follow-up commit changed in this file is untested: `onBlur` recomputing from `event.target.value`, typing and toggling clearing the analysis, the load path analyzing stored meta, `aria-invalid` staying false when only the over-broad warning shows, and `aria-describedby` pointing at a container that exists before the first warning. React Testing Library is already used in this plugin (`src/wizards/newsletters/views/settings/index.test.js`). A regression that moves the analysis back into `onChange` or renames `IP_RANGE_MESSAGES_ID` would pass CI today.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +375,76 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
 					>
 						<CardBody size="small">
 							<TextControl
-								label={ __( 'IPs / CIDR blocks (comma-separated)', 'newspack-plugin' ) }
+								label={ __( 'IPs, CIDR blocks, or IP ranges (comma-separated)', 'newspack-plugin' ) }
 								value={ ipRange }
-								onChange={ ( val: string ) => updateMeta( 'np_institution_ip_range', val ) }
-								placeholder="192.168.1.0/24, 10.0.0.5"
+								onChange={ updateIpRange }
+								onBlur={ ( event: React.FocusEvent< HTMLInputElement > ) =>
```
dkoo, line 479: **Suggestion:** In the paste-then-Save flow the warning renders and is unmounted before anyone reads it — Clicking Save does blur the input first, so the analysis runs and the notice commits. But `handleSave` never reads `ipRangeAnalysis`, and `history.push( '/institutions' )` in its `.then` unmounts the live region as soon as the request resolves; the list view shows nothing. In the env the notice appeared on the Save mousedown and the route changed 4.3 s later, and only because the local save was slow. On a fast connection it's a flash, and a screen-reader announcement is likely pre-empted by the route change.

The admin who pastes a list from a ticket and clicks Save is the most common path and the one this warning exists for. Keeping the save non-blocking is right; losing the message is the problem. Two cheap options: run `analyzeIpRangeEntries()` on the payload inside `handleSave` and raise a wizard-level `addNotice( { type: 'warning', … } )` when anything is flagged (the `.catch` a few lines down already does this, and that notice survives the navigation), or stay on the edit page after a save that produced warnings.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
```diff
@@ -0,0 +1,162 @@
...
+		return [];
+	}
+	return keys;
+}
+
+/**
+ * Whether a valid dash range spans an implausibly large number of addresses.
+ */
+function isOverBroadRange( token: string ): boolean {
+	if ( token.includes( '/' ) || ! token.includes( '-' ) || ! isValidEntry( token ) ) {
```
dkoo: **Suggestion:** The over-broad warning exempts CIDR, so `0.0.0.0/0` saves with no warning at all — `isOverBroadRange()` returns false for any token containing `/`, and `utils.test.js:126` pins that by asserting `0.0.0.0/0` has no over-broad result. In the env, entering `0.0.0.0/0, 10.0.0.0/8, 203.0.113.0-203.0.113.255` and tabbing out produced no notice and `aria-invalid="false"`, and the value saved. The same field flags a 65,537-address dash range one line later.

The rationale for the dash check (one wrong digit authorizes millions of addresses, and it reads as fine) applies at least as much to a `/8` typed where `/28` was meant. The threshold already exists: `OVER_BROAD_RANGE_SIZE = 65535` is a `/16`, and the migration CLI on `main` uses the same cut-off (`get_range_breadth_warning()`, mask < 16). Computing `2 ** ( 32 - bits )` in the CIDR branch and flipping the test expectation closes the larger half of the surface this warning was written for.

**Risk if not addressed:** If one of those already-stored entries is a typo in the end address, the same deploy turns a range meant to cover one campus into one covering millions of addresses. Every visitor inside it is treated as on-campus: the gate stops rendering, the access cookie is set, and the server-to-server check answers that no paywall should show. Nothing on the server judges how wide a range is, and the only warning lives in the editor, so it appears only if someone reopens that institution and clicks out of the field. The publisher would most likely notice through a drop in conversions.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
```diff
@@ -88,6 +88,222 @@ public function test_cidr_tolerates_whitespace_around_slash() {
...
+		);
+	}
+
+	/**
+	 * Provide the shared client/server validation cases.
+	 *
+	 * @return array[] Keyed by case label: [ entry, is_valid ].
+	 */
+	public function shared_validation_case_provider() {
+		$fixture_path = dirname( __DIR__, 2 ) . '/fixtures/ip-range-validation-cases.json';
```
dkoo, line 240: **Suggestion:** The PHP half of the parity guard passes as *skipped* if the fixture is empty or missing — The provider does `json_decode( file_get_contents( $fixture_path ), true )` and iterates `$fixture['cases']` with no guard. A missing file or a JSON typo yields an empty provider, which PHPUnit 9 reports as **skipped**, not failed, and `phpunit.xml` sets no `failOnRisky`.

**To reproduce:**
1. `echo '{"cases": []}' > tests/fixtures/ip-range-validation-cases.json`
2. `n test-php --filter test_shared_validation_fixture_parity`
3. Output: `OK, but incomplete, skipped, or risky tests! Tests: 1, Assertions: 0, Skipped: 1` — exit 0.

The jest side fails loudly on the same file (`it.each( [] )` throws), so a broken fixture looks like a client-only problem and invites the wrong fix. Throwing a `RuntimeException` from the provider when `empty( $fixture['cases'] )` (PHPUnit surfaces a provider exception as an error) makes the guard fail on both sides. Worth a second line while there: cases are keyed by `label`, so two cases with the same label silently collapse into one on the PHP side but not in jest.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +375,76 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+									setIpRangeAnalysis( analyzeIpRangeEntries( event.target.value ) )
+								}
+								placeholder="192.168.1.0/24, 203.0.113.0-203.0.113.255, 10.0.0.5"
```
dkoo: **Nit:** The placeholder and the invalid-entry notice offer private addresses as the examples to type — The placeholder reads `192.168.1.0/24, 203.0.113.0-203.0.113.255, 10.0.0.5`, and the invalid-entry notice hands back `10.0.0.5` and `192.168.1.0/24` as what to type instead. Both are RFC 1918 space. `get_visitor_ip()` reads `REMOTE_ADDR`, so a private entry can never match a visitor's public egress IP; the migration CLI on `main` says as much ("is a private or reserved range — it will never match real visitors' public egress IPs"). The notice whose job is to stop admins typing entries that silently never grant access is demonstrating two. RFC 5737 documentation addresses (`203.0.113.5`, `198.51.100.0/24`) fix it, and a private/reserved warning class reusing `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE` would be a natural fourth class if you want it.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +375,76 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+											'203.0.113.0-203.0.113.255'
+										) }
+									/>
+								) }
+								{ confusableCharacters.length > 0 && (
+									<Notice
+										isWarning
+										noticeText={ sprintf(
+											/* translators: %s: comma-separated list of character names, e.g. "en dash (–)". */
+											_n(
```
dkoo: **Nit:** The confusable-character notice renders "contains a en dash" and pluralizes on the wrong count — On screen the singular reads "An entry contains a en dash (–), which looks like a standard character but is not." The article is baked into the source string, so translators can't fix it either. `_n()` is also keyed on `confusableCharacters.length` (distinct characters) while the sentences talk about entries: one entry with both an en dash and an NBSP renders the plural "Entries contain…". A count-independent phrasing avoids both: "One or more entries contain characters that look standard but are not: %s. Retype them using a plain hyphen (-) and regular spaces."

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
```diff
@@ -307,19 +375,76 @@ export default function InstitutionEdit( { match }: { match: { params: { id?: st
...
+										noticeText={ sprintf(
+											/* translators: %1$s: comma-separated list of unusually wide IP ranges. %2$s: formatted number of addresses. */
+											_n(
+												'This range covers more than %2$s addresses: %1$s. Check the end address — an over-broad range grants access far beyond the institution.',
+												'These ranges cover more than %2$s addresses each: %1$s. Check the end addresses — an over-broad range grants access far beyond the institution.',
+												overBroadIpRanges.length,
+												'newspack-plugin'
+											),
+											formatEntryList( overBroadIpRanges ),
+											new Intl.NumberFormat().format( OVER_BROAD_RANGE_SIZE + 1 )
```
dkoo: **Nit:** "65,536" is formatted in the browser's locale, not the admin's — `new Intl.NumberFormat().format( … )` uses the browser default, so a `de_DE` admin in an `en-US` browser sees "65,536" inside an otherwise German sentence. Passing the admin locale (`document.documentElement.lang` is available) as the first argument keeps the number consistent with the surrounding string. The arithmetic itself is right: a difference greater than 65,535 means more than 65,536 addresses, and a full `/16` is correctly not flagged.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+export const OVER_BROAD_RANGE_SIZE = 65535;
+
+/**
+ * Characters that look like a hyphen or a space but aren't, keyed by a stable
+ * identifier the UI maps to a translated label. Word, Google Docs and Outlook
+ * autocorrect ` - ` into an en dash, so these arrive by copy-paste far more
+ * often than they are typed.
+ */
+export const CONFUSABLE_CHARACTERS = {
```
dkoo, line 34: **Nit:** The confusable set covers the Word family but not the hyphen look-alikes — `CONFUSABLE_CHARACTERS` names en/em dash, minus sign, and three space variants. Unicode hyphen (U+2010), non-breaking hyphen (U+2011, Word's Ctrl+Shift+hyphen and common in PDFs), figure dash (U+2012), fullwidth hyphen-minus (U+FF0D, from the same CJK IMEs that produce the U+3000 already in the set), and a leading BOM (U+FEFF, from Windows text files) all render as a plain hyphen or nothing. They're still rejected as invalid, so nothing over-grants, but they get the generic message rather than the named-character one that exists for exactly this case. One line each in the map and `replaceConfusableCharacters()`, plus a fixture row apiece.

dkoo (approved):
@adekbadek sorry about the delay in re-reviewing this PR. It's clean this time: all blockers are addressed, and only some non-blocking suggestions and nits remain.

Release risk: ELEVATED—the REST `ip_ranges` field gains a third notation with no discriminator and at least one possible consumer (`newspack-extended-access`) can't be checked from this workspace; dash entries already stored on live sites start granting access on deploy; nothing server-side judges how wide a range is. None of that changed since the first round, and both behaviors are now called out at the top of the PR description.

Every change from the first round holds up at runtime, and I put the ones I was least sure about under test rather than reading them. The blur flow does what it says: typing a range one key at a time shows nothing, tabbing out with an incomplete entry warns, editing clears it, and reopening a saved institution re-runs the analysis on stored meta. The warnings live in a `role="status"` region the input points at through `aria-describedby`, and `aria-invalid` tracks only the invalid list, so a valid-but-wide range is announced without being marked invalid. On the server, `ip_to_unsigned()` casts to float before comparing, and a range straddling `127.255.255.255` matches `128.0.0.0` and rejects `127.255.255.254`. The REST allowlist returns the dash ranges and drops the invalid and en-dash entries. Both suites read the same fixture file, and a deliberate mismatch fails jest. PHPUnit `Access_Rules` 116/116, jest 57/57.

No blockers. Three things are worth a look before merge, each confirmed in the env rather than reasoned: the over-broad warning exempts CIDR, so `0.0.0.0/0` saves silently while a 65,537-address dash range is flagged; the PHP half of the parity guard passes as *skipped* with zero assertions if the fixture is empty or missing, so the drift it exists to catch can go quiet without failing CI; and in the paste-then-Save flow the warning renders on the Save mousedown and the redirect unmounts it, so the most common path never shows the admin the message. One merge interaction isn't visible in the diff: `main` now carries a third parser in the institutions migration CLI that rejects dash ranges and can't judge their breadth. The rest are copy and coverage nits.

Verdict: approve with comments. The open items are real but none of them makes the feature worse than what ships today, and all are small follow-ups if not addressed here.

Before the release goes out, count how many live institutions have a dash in their stored IP field and how wide those ranges are. Those entries start granting access the moment the code lands, so a short query now is cheaper than finding out later which ones were typos.

- `plugins/newspack-plugin/includes/content-gate/class-ip-access-rule.php:662` — **Suggestion:** After merging `main`, a third parser in the migration CLI rejects dash ranges and can't judge their breadth — Not visible in the diff: the branch is 386 commits behind `main`, where #688 added `includes/cli/class-institutions-migration.php`. Its `normalize_ip_ranges()` (line 444) accepts IPv4 and CIDR only and reports everything else as invalid with the message "supports IPv4 addresses and CIDR blocks only", which stops being true once this lands; a legacy team's dash range would be dropped by the migration while the runtime accepts it. `get_range_breadth_warning()` (line 495) splits on `/`, so a dash range of any width returns `''` (unremarkable) from the one audit that says out loud "this range matches every visitor". This is the drift class the shared fixture was added to prevent, now in a third hand-written PHP copy the fixture can't reach because `parse_ip_ranges()` is private. Within PHP, extraction beats a fixture: expose a public classifier on `IP_Access_Rule` (valid/invalid split plus a breadth judgment) and have the CLI delegate to it, either here after the rebase or as an immediate follow-up ticket. At minimum the line-215 message needs updating.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Added `edit.test.jsx` (RTL, same setup as the newsletters suite) covering the wiring: `onBlur` recomputing from `event.target.value`, no warning while typing, typing and toggling the rule off clearing the analysis, the load path analyzing stored meta with no interaction, `aria-invalid` staying false when only the over-broad warning shows, and the live region existing and being empty before the first warning. Mutation-checked both regressions named here — moving the analysis into `onChange` fails the blur test, and renaming `IP_RANGE_MESSAGES_ID` fails four of them (the test looks the region up by literal id). Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Took the first option. `handleSave` now re-analyzes the value being saved and, on success, raises a wizard-level warning notice; that snackbar renders outside the router, so it survives `history.push`. The save stays non-blocking. The three inline notice strings moved into a `getIpRangeWarnings()` helper that both the inline region and the snackbar use, so the two can't drift. Covered by two tests: one asserting the warning notice lands on the store after a save with an invalid entry, one asserting a clean value raises nothing. Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
adekbadek: Agreed — the CIDR exemption was the larger half of the surface. The check now works from an address count rather than a dash-specific difference: `getAddressCount()` returns `2 ** ( 32 - bits )` for CIDR, `end - start + 1` for a dash range, and an entry is flagged when that exceeds `OVER_BROAD_RANGE_SIZE`. That constant is now 65,536 (a /16) as a count rather than 65,535 as a difference, which drops the `+ 1` the notice used to need. Cut-off is unchanged in effect: a full /16 written either way is not flagged, `/15` and a 65,537-address dash range are. The `0.0.0.0/0` expectation in `utils.test.js` is flipped, with `/8`, `/15`, `/16` and `/24` pinned alongside it, and the notice text now names the CIDR mask as well as the end address. Fixed in b7d3761.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-ip-access-rule.php
adekbadek: The provider now throws a `RuntimeException` when the fixture is missing, unreadable or has no cases, and a second one on a duplicate label. Verified both: `echo '{"cases": []}' > …` then `--filter test_shared_validation_fixture_parity` gives `Tests: 1, Assertions: 0, Errors: 1` and a non-zero exit instead of a skip, and appending a case reusing an existing label errors with the label named. Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Switched both to RFC 5737 documentation addresses. The placeholder is now `198.51.100.0/24, 203.0.113.0-203.0.113.255, 203.0.113.5` and the invalid-entry notice hands back `203.0.113.5` and `198.51.100.0/24`.

I left the private/reserved warning class out. It would fire on entries that are wrong in a different way — inert because of where they point rather than because the parser drops them — and `get_visitor_ip()` has the `newspack_trusted_proxy_headers` filter, so an operator can in principle put a private address in front of it. That makes it a product call about nagging rather than a defect, and it wants its own ticket. Happy to add it here if you'd rather. Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Took your phrasing verbatim: one `__()` string, no `_n()`, so the article problem and the count mismatch both go away. Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/edit.tsx
adekbadek: Now `new Intl.NumberFormat( document.documentElement.lang || undefined )`, falling back to the browser default only when the attribute is empty. The number it formats is `OVER_BROAD_RANGE_SIZE` directly — the constant is now an address count (65,536) rather than a difference, so the `+ 1` is gone. Fixed in b7d3761.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/utils.ts
adekbadek: Added all five: Unicode hyphen (U+2010), non-breaking hyphen (U+2011), figure dash (U+2012), fullwidth hyphen-minus (U+FF0D) and the BOM (U+FEFF), each with a label, a `replaceConfusableCharacters()` mapping (BOM strips like the zero-width space) and a fixture row. Also added a fixture row for the narrow no-break space, which was in the map but had no case. Two jest cases cover them — one naming each character, one asserting an entry carrying both an en dash and a narrow no-break space reports both. Fixed in b7d3761.

adekbadek:
Both items from your review addressed.

**1. Live dash-range census (pre-release).** Across the institutional-access sites — the only class that has `np_institution` posts — **no live institution currently stores a dash range**. Every stored `np_institution_ip_range` on the one migrated institutional site is a single IP, a CIDR block, or a comma-separated list of those, so nothing already migrated flips from "dropped" to "granting" on deploy.

Worth flagging for the rollout, though: several not-yet-migrated institutional sites carry dash ranges in their WCM source (the format the migrator converts), and a few are wide — at least one source range spans a full /16 (~65,536 addresses), others in the 256–8,192 range. So the dash path will start seeing real, occasionally very wide input as those sites flip, which is what item 2 now guards.

**2. Public classifier + CLI delegation + message.** Done in 82b75f528e (this branch now carries `main`). `IP_Access_Rule` is the single source of truth: public `classify_entry()`, `get_entry_size()` (address count, uniform across ip/cidr/range), and `normalize_ip_ranges()` (`{valid, invalid}`); the private `parse_ip_ranges()` delegates to them, and the shared client/server fixture still passes unchanged. The institutions-migration CLI now delegates validation to `IP_Access_Rule::normalize_ip_ranges()` — so a legacy team's dash range migrates instead of being dropped — and judges dash-range breadth via `get_entry_size()` against the same `> /16` threshold as CIDR. The line-215 message now names dash ranges as supported.

Tests: `Access_Rules` and the migration suite green, and the shared fixture parity guard still holds. Left the three non-blocking client-side items (the CIDR exemption in the over-broad warning, the fixture-empty skip guard, and the paste-then-Save unmount) as follow-ups, since your approval treated them as optional.

adekbadek:
@dkoo Re-requesting review – the commits since your approval add a whole new institutions migration CLI plus a public entry classifier/`normalize_ip_ranges` on `IP_Access_Rule` that `parse_ip_ranges()` now delegates to, and the REST check no longer returns the institution name unless the IP matched — well past the nits you signed off on:
- feat(content-gate): warn on over-broad CIDR blocks, more hyphen look-alikes, and IP warnings that survive save (NPPD-1509, #705)
- feat(content-gate): accept dash IP ranges in the institutions migration via a shared IP_Access_Rule classifier (NPPD-1509, #705)

PR #707, opened 2026-07-23
fix(content-gate): allow zero Free Views in metering settings (NPPD-2056)

adekbadek:
### All Submissions:

In Paid access metering, typing `0` into **Free Views** saved and autocorrected to `1`, while leaving the field blank saved as `0`. Root cause: `min={ 1 }` on the `NumberControl` — WordPress' `__experimentalNumberControl` clamps to `min` on commit (blur/Enter), client-side, before any save; the blank path maps `undefined → 0` in `onChange` and bypasses the clamp, producing the asymmetry. The server (`sanitize_metering()`) already accepts `0`.

One-line fix: `min={ 0 }`. Blank still saves as `0`, now consistent with typing `0`. The shared `Metering` component serves both Registered and Paid access sections, so both are covered. With `count=0` + metering enabled the gate renders immediately for everyone ("no free views") — the existing warning Notice ("Content will be gated for all readers") and the pre-save panel ("Metered: 0 free views per month") already communicate this; no new help text needed. Other `min={ 1 }` fields (gifting, donations, pricing) have legitimate floors and are untouched.

Closes NPPD-2056.

1. Audience > Access Control → edit a gate → Paid access → enable metering.
2. Type `0` into Free Views and blur: the value stays `0` (previously flipped to `1`) and the 0-views warning appears.
3. Save, reload: Free Views shows `0`; gate meta has `metering.count = 0`.
4. Blank the field and save: still `0`. Normal values round-trip unchanged.

### Other information:

* [x] Have you written new tests for your changes, as applicable? (4 RTL tests; the typed-0 case fails against the old `min={ 1 }` — verified red/green in both directions)
* [x] Have you successfully run tests with your changes locally? (jest 46 suites / 433 tests green; ESLint clean)

### DIFF
--- a/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
+++ b/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
@@ -1,0 +1,61 @@
+/**
+ * External dependencies
+ */
+import { render, screen, fireEvent } from '@testing-library/react';
+import { useState } from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Metering from './metering';
+
+/**
+ * Stateful harness: NumberControl only commits values through a controlled
+ * round-trip, so the component is rendered with live state like in the wizard.
+ */
+function MeteringHarness( { initialMetering } ) {
+	const [ metering, setMetering ] = useState( initialMetering );
+	return <Metering metering={ metering } onChange={ setMetering } />;
+}
+
+const getFreeViewsInput = () => screen.getByRole( 'spinbutton', { name: 'Free views' } );
+
+const typeAndBlur = ( input, value ) => {
+	fireEvent.change( input, { target: { value } } );
+	fireEvent.blur( input );
+};
+
+describe( 'Metering free views count', () => {
+	it( 'keeps an explicitly entered 0 instead of autocorrecting it to 1', () => {
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '0' );
+
+		expect( freeViewsInput.value ).toBe( '0' );
+	} );
+
+	it( 'shows the gated-for-all warning when the count is 0', () => {
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 0, period: 'month' } } /> );
+
+		expect( screen.getByText( /Content will be gated for all readers/ ) ).toBeInTheDocument();
+	} );
+
+	it( 'treats a blanked field as 0', () => {
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '' );
+
+		expect( freeViewsInput.value ).toBe( '0' );
+	} );
+
+	it( 'accepts a regular positive count', () => {
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '3' );
+
+		expect( freeViewsInput.value ).toBe( '3' );
+	} );
+} );
--- a/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
+++ b/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
@@ -80,10 +80,10 @@ describe( 'Metering free views count', () => {
 		expect( freeViewsInput.value ).toBe( '2' );
 	} );
 
-	it( 'shows the gated-for-all warning when the count is 0', () => {
+	it( 'warns that a count of 0 is the same as turning metering off', () => {
 		render( <MeteringHarness initialMetering={ { enabled: true, count: 0, period: 'month' } } /> );
 
-		expect( screen.getByText( /Content will be gated for all readers/ ) ).toBeInTheDocument();
+		expect( screen.getByText( /the same behavior as turning Metering off/ ) ).toBeInTheDocument();
 	} );
 
 	it( 'treats a blanked field as 0', () => {
--- a/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
+++ b/plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
@@ -1,8 +1,20 @@
 /**
+ * Regression test for NPPD-2056: the Free views field autocorrected an explicitly
+ * entered 0 back to 1, so a publisher who wanted "no free views" silently granted
+ * one. These tests pin the floor that replaced it – `min={ 0 }` on the control plus
+ * the component's own clamp – including the negative and fractional entries that
+ * floor is responsible for.
+ */
+
+/**
  * External dependencies
  */
 import { render, screen, fireEvent } from '@testing-library/react';
-import { useState } from 'react';
+
+/**
+ * WordPress dependencies
+ */
+import { useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -10,12 +22,21 @@ import { useState } from 'react';
 import Metering from './metering';
 
 /**
- * Stateful harness: NumberControl only commits values through a controlled
- * round-trip, so the component is rendered with live state like in the wizard.
+ * Stateful harness mirroring the wizard: Metering is controlled, so the parent has
+ * to feed committed values back through the `metering` prop for the field to settle.
+ *
+ * The optional spy records the payload the wizard would save, which is what the
+ * assertions below check. Asserting the rendered input value alone is not enough:
+ * the control keeps an internal draft, so the input reads "0" after typing 0 even
+ * if `onChange` never fires at all.
  */
-function MeteringHarness( { initialMetering } ) {
+function MeteringHarness( { initialMetering, onChange = () => {} } ) {
 	const [ metering, setMetering ] = useState( initialMetering );
-	return <Metering metering={ metering } onChange={ setMetering } />;
+	const handleChange = nextMetering => {
+		setMetering( nextMetering );
+		onChange( nextMetering );
+	};
+	return <Metering metering={ metering } onChange={ handleChange } />;
 }
 
 const getFreeViewsInput = () => screen.getByRole( 'spinbutton', { name: 'Free views' } );
@@ -27,12 +48,36 @@ const typeAndBlur = ( input, value ) => {
 
 describe( 'Metering free views count', () => {
 	it( 'keeps an explicitly entered 0 instead of autocorrecting it to 1', () => {
-		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+		const onChange = jest.fn();
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } onChange={ onChange } /> );
 
 		const freeViewsInput = getFreeViewsInput();
 		typeAndBlur( freeViewsInput, '0' );
 
+		expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 0 } ) );
 		expect( freeViewsInput.value ).toBe( '0' );
+	} );
+
+	it( 'floors a negative entry at 0', () => {
+		const onChange = jest.fn();
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } onChange={ onChange } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '-1' );
+
+		expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 0 } ) );
+		expect( freeViewsInput.value ).toBe( '0' );
+	} );
+
+	it( 'rounds a fractional entry to a whole number of views', () => {
+		const onChange = jest.fn();
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } onChange={ onChange } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '1.5' );
+
+		expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 2 } ) );
+		expect( freeViewsInput.value ).toBe( '2' );
 	} );
 
 	it( 'shows the gated-for-all warning when the count is 0', () => {
@@ -42,20 +87,24 @@ describe( 'Metering free views count', () => {
 	} );
 
 	it( 'treats a blanked field as 0', () => {
-		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+		const onChange = jest.fn();
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } onChange={ onChange } /> );
 
 		const freeViewsInput = getFreeViewsInput();
 		typeAndBlur( freeViewsInput, '' );
 
+		expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 0 } ) );
 		expect( freeViewsInput.value ).toBe( '0' );
 	} );
 
 	it( 'accepts a regular positive count', () => {
-		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+		const onChange = jest.fn();
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } onChange={ onChange } /> );
 
 		const freeViewsInput = getFreeViewsInput();
 		typeAndBlur( freeViewsInput, '3' );
 
+		expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 3 } ) );
 		expect( freeViewsInput.value ).toBe( '3' );
 	} );
 } );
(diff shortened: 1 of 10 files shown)


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-2056-free-views-field-autocorrects-0-to-1` (stash local work first).

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+const getFreeViewsInput = () => screen.getByRole( 'spinbutton', { name: 'Free views' } );
+
+const typeAndBlur = ( input, value ) => {
+	fireEvent.change( input, { target: { value } } );
+	fireEvent.blur( input );
+};
+
+describe( 'Metering free views count', () => {
+	it( 'keeps an explicitly entered 0 instead of autocorrecting it to 1', () => {
```
dkoo, line 50: **Suggestion:** Nothing covers what `min={ 0 }` now guards — Mutation-testing the suite (flipping `min` back to `1`) fails exactly one test — this first one. The other three pass under both values, which is fine as characterization, but it means nothing pins the behavior `min={ 0 }` is now responsible for: the negative and decimal clamp paths.

Those are the behaviors most likely to silently regress if someone later touches the `onChange` on line 51 or swaps the control, and they are cheap to add — `typeAndBlur( input, '-1' )` should yield `'0'`, and `typeAndBlur( input, '1.5' )` should yield `'2'`. A test asserting that a negative entry commits to `0` also documents that `min={ 0 }` is load-bearing, so nobody deletes it later as unnecessary.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+		render( <MeteringHarness initialMetering={ { enabled: true, count: 5, period: 'month' } } /> );
+
+		const freeViewsInput = getFreeViewsInput();
+		typeAndBlur( freeViewsInput, '0' );
+
+		expect( freeViewsInput.value ).toBe( '0' );
```
dkoo, line 58: **Suggestion:** The key assertion checks the displayed value, not what would be saved — `expect( freeViewsInput.value ).toBe( '0' )` passes even with a no-op `onChange` — rendering `Metering` with a spy that never updates parent state, then typing `0` and blurring, still leaves `input.value === "0"`, because the control holds its own internal draft and the `value` prop never changed. So this assertion would hold even if `onChange` never fired.

The reported bug is about what gets *saved* ("typing 0 saved and autocorrected to 1"), so the test should assert the payload: `expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: 0 } ) )`. Worth noting the coverage isn't fake — the test does still go red against `min={ 1 }`, because the clamp mutates internal state and the field displays `"1"` — it just pins the regression for a weaker reason than the harness comment claims.

Related: the blanked-field assertion a few lines down looks identical but gets its `'0'` from the controlled-prop round trip rather than internal state. Asserting the payload in both makes them say what they mean.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.tsx
```diff
@@ -46,7 +46,7 @@ export default function Metering( { description, metering, onChange }: MeteringP
 					<NumberControl
 						label={ __( 'Free views', 'newspack-plugin' ) }
 						help={ __( 'Free views before the gate appears.', 'newspack-plugin' ) }
-						min={ 1 }
+						min={ 0 }
 						value={ count }
 						onChange={ v => onChange( { ...metering, count: v !== undefined ? Number( v ) : 0 } ) }
```
dkoo: **Suggestion:** `min` is now the only floor, and it only applies at commit — `onChange` writes raw `Number( v )` into gate state on every keystroke with no floor and no integer coercion; `min={ 0 }` is enforced only when the control commits (blur/Enter). Instrumented runs confirm the parent state transiently holds `-1` after typing `-1` and `1.5` after typing `1.5`, before blur corrects them to `0` and `2`.

What makes this worth mentioning rather than ignoring: if such a value ever reached the server, `sanitize_metering()` uses signed `intval()` so `-1` persists as `-1`, and `Metering::get_registered_settings()` reads it back through `absint()` — turning it into **1 free view**. A publisher who meant "no free views" would silently grant one, which is the exact bug class this PR set out to kill.

No live save path reaches it today (the Save control lives in a separate panel, so the field always blurs first, and Enter runs the clamp too), so this is defense-in-depth rather than a live defect. `count: Math.max( 0, Math.round( Number( v ) || 0 ) )` would make the invariant local to the component instead of dependent on the control's commit timing. Hardening `sanitize_metering()` with `max( 0, intval( ... ) )` would close the REST-caller path too.

**Runtime verification (both halves checked).** The transient state is real: typing `-1` over a saved `5` leaves the input at `value="-1"` with `validity.rangeUnderflow === true` and the wizard Save control enabled, so the negative is genuinely in parent gate state pre-blur (`1.5` reproduces the same way, with `validity.stepMismatch === true`). But no UI path carries it to the server, and this was tested rather than assumed: a real mouse click on Save blurs on `mousedown`, so the field clamps to `0` and the pre-save panel opens already reading "Metered: 0 free views per month" — the save then writes a strict integer `0`. Enter commits and rounds in place (`1.5` → `2`) without submitting; the input is not inside any `<form>`, so Enter cannot submit the save panel; there is no `mod+s`/`useShortcut` handler in the audience wizard; and the save panel is a focus-trapping `Modal` reachable only by pointer or Tab, both of which blur first. The "defense-in-depth rather than a live defect" framing is **confirmed**, and the severity stays where it is.

dkoo (commented):
@adekbadek I reviewed and manually tested this PR and it does fix the issue and earn an APPROVE verdict from our reviewing tools, but I wanted to check the approach before we move ahead.

My view: enabling metering for Registered or Paid access settings but with a "free views" value of `0` is functionally the same as metering being off entirely. Can we make the "Metering is enabled but the number of views is set to 0" warning more explicit that this configuration is functionally the same as metering = OFF? 

A sweep of the code confirmed that enforcement is already correct: `is_frontend_metering()`, `is_logged_in_metering_allowed()` and `has_metering()` all require `count > 0`, so a reader sees exactly the same thing at ON+0 as at OFF. Front-end assets are clean too: `Metering::enqueue_scripts()` (`class-metering.php:294`) is gated behind `is_frontend_metering()`, whose decisive line is `$settings['enabled'] && $settings['count'] > 0`, so `content-gate-metering.js` and its localized settings are never enqueued at ON+0. No wasted requests.

But it also exposed three areas where the logic differs slightly, all on the presentation side. We might want to handle these and the amended warning message in this PR, or in a separate follow-up—your choice. I won't consider these a blocker as the fix here is minimal and addresses the reported issue.

- **`Block_Patterns::get_metering_settings()`** (`includes/content-gate/class-block-patterns.php:120`) guards with `! empty( $metering['count'] )` over a hardcoded `$count = 4`, so a stored `0` is indistinguishable from unset and the generated paywall advertises "Get 4 free articles every month" on a gate that grants none. This one doesn't read `metering.enabled` at all — and making it read `enabled` wouldn't help, because the fallback is keyed on the count. All four states:

  ```
  metering ON  + count 0  ->  emitted count = 4
  metering OFF + count 0  ->  emitted count = 4
  metering ON  + count 3  ->  emitted count = 3
  metering OFF + count 3  ->  emitted count = 3
  ```

  The fix is `isset()` + `is_numeric()` rather than `! empty()`, plus a decision on the copy at zero — `_n( …, 0 )` would render "Get 0 free articles every month", which probably argues for dropping the metering sentence entirely. The generated layout is produced once (gate creation, duplication, missing-layout path) and nothing regenerates it when the setting later changes, so wrong copy persists until someone edits the layout post by hand.

- **`content-gates.tsx:41`** derives `hasMetering` from `metering.enabled` alone, unlocking the countdown-banner feature, while PHP's `has_metering()` requires `count > 0`. At ON+0 the banner appears configurable and its toggle persists, but it would never render.

- **`Content_Gate::is_metering_enabled()`** (`class-content-gate.php:629`) checks `isset( $gate['metering'] )`, but `get_gate()` doesn't return a `metering` key at all:

  ```
  modern gate keys: id, status, title, priority, content_rules, content_rules_match, registration, custom_access
  is_metering_enabled() = false
  ```

  So it returns `false` unconditionally for the modern gate CPT, and it feeds `has_metering` in two wizard payloads (`class-audience-wizard.php:113`, `class-audience-content-gates.php:115`). Independent of the zero question, but the kind of thing the sweep should catch.

Full review notes below.

Release risk: LOW — a one-line UI change, no contract files touched, tests cover the changed path, and the gating runtime already handles a zero count correctly.

No blockers. I verified the mechanism rather than trusting the description: under `min="1"` a typed `0` is natively `rangeUnderflow`, so blur commits and clamps to 1, while under `min="0"` it is valid and the typed value stands. That is exactly the reported autocorrect, and this is the right layer to fix it. The alternative theory — that the real bug is the `undefined → 0` mapping in `onChange` — does not hold, since blanking delivers `''` and 0 is a value the server genuinely supports. I also mutation-tested the new suite: reverting `min` to 1 fails exactly one test, so the regression is genuinely pinned.

One reviewer raised the generated-layout copy as a blocker. I adjudicated it and it does not block this PR: `count = 0` was already reachable on `alpha` by blanking the field — the base branch already ships the "views is set to 0 / Content will be gated for all readers" warning for precisely that state — so this diff makes an existing state easier to reach rather than creating it. It is still worth a follow-up ticket, and it is the most substantive item below.

The theme across the findings is worth naming: every place that *enforces* metering reads the count with an explicit numeric comparison and is correct at zero, while several places that *describe* it use truthiness tests and now disagree. That shows up in the generated paywall copy (which can advertise 4 free articles on a gate granting none), in the countdown-banner affordance, and in `Content_Gate::is_metering_enabled()`, which tests `enabled` without consulting the count. All pre-existing, none introduced here, all newly more likely to be hit. The failure direction is consistently *more* gating rather than less, so nothing leaks — but a paywall promising an allowance it does not honor is a publisher-visible papercut.

On the tests themselves: the primary assertion checks the input's displayed value rather than the payload that would be saved, which proves less than the harness comment claims — it still goes red against `min={ 1 }`, so the coverage is real, just pinned for a weaker reason.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
```diff
@@ -0,0 +1,61 @@
+/**
+ * External dependencies
+ */
+import { render, screen, fireEvent } from '@testing-library/react';
+import { useState } from 'react';
```
dkoo: **Nit:** `useState` imported from `react` rather than `@wordpress/element` — Sibling files in this tree (`edit/index.tsx`, `edit/save-panel/index.tsx`, `registration.tsx`) import React APIs from `@wordpress/element`, which is what the repo's import convention describes. There is precedent for importing from `react` directly elsewhere in `src/`, and ESLint is clean, so this is purely local consistency.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Metering from './metering';
+
+/**
```
dkoo, line 24: **Nit:** The harness rationale comment isn't accurate — "NumberControl only commits values through a controlled round-trip" isn't true — it commits and re-renders fine with an inert parent, which is precisely why the assertion above can pass without `onChange` ever firing. The harness is the right shape (it mirrors the wizard), but the stated reason will mislead the next reader.

The sibling regression test at `pricing-rules/rule-edit.test.jsx` leads with a docblock naming the bug it pins; adding an `NPPD-2056` reference here would match that convention and make the file's intent obvious.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
adekbadek: Right, and confirmed by mutation: stubbing the control's `onChange` to a no-op leaves the typed-0, `-1`, `1.5` and `3` value assertions all passing (only the blanked-field one fails, exactly as you noted — that one comes from the controlled round trip). Every test now asserts the payload via `expect( onChange ).toHaveBeenLastCalledWith( expect.objectContaining( { count: N } ) )`; under the same mutation all five fail with "Number of calls: 0". Under `min={ 1 }` the typed-0 payload assertion fails on the real reason: last call is `{ count: 1 }`.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
adekbadek: Added both clamp-path tests: `-1` must commit to `0`, `1.5` must commit to `2`. Mutation-checked — with `min={ 1 }` restored, the negative test goes red alongside the typed-0 one (2 failures instead of 1), so the floor is now pinned rather than incidental.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.tsx
adekbadek: Applied both halves. Component: `count: Math.max( 0, Math.round( Number( v ) || 0 ) )`, so the invariant is local instead of dependent on commit timing. Server: `sanitize_metering()` now uses `max( 0, intval( … ) )`, plus a PHP unit test for the negative input (red without the floor: "Failed asserting that -1 is identical to 0").

Verified end-to-end in an env: typing `-1` while the field is still focused now shows the "Content will be gated for all readers" notice immediately — that notice renders off parent state, so the negative no longer reaches it. And a REST-shaped payload with `count: -1` stores `0`, which `Metering::get_registered_settings()` reads back as `0` rather than 1.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
adekbadek: Switched to `useState` from `@wordpress/element` to match the sibling files in this tree.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/metering.test.jsx
adekbadek: Rewritten. The comment now says what is actually true — the control keeps its own internal draft, so the rendered value alone would pass even if `onChange` never fired, which is why the harness takes a payload spy. Also added an `NPPD-2056` docblock at the top of the file, matching `pricing-rules/rule-edit.test.jsx`.

dkoo:
@adekbadek are we handling the "metering on + 0 views === metering off" sweep in a follow-up?

adekbadek:
@dkoo handled in this PR rather than a follow-up — the full "metering ON + 0 views === metering OFF" sweep is in aed57fbd. All three presentation-side divergences you flagged, plus the warning copy, now route through one predicate (active + metering enabled + a positive count):

- **`Content_Gate::is_metering_enabled()`** — was dead since gate settings stopped exposing a top-level `metering` key (it read `isset( $gate['metering'] )`). It now delegates to the new `Metering::is_gate_metered( $gate_id )`, which is the single server-side source of truth; `is_post_metered()` delegates to it too.
- **`content-gates.tsx`** — `hasMetering` now derives from a matching `isGateMetered()` helper in `utils.tsx` instead of `metering.enabled` alone, so the countdown affordance is not offered on a gate that never counts. The helper mirrors the PHP predicate.
- **`Block_Patterns` generated copy** — rather than the `isset()`/`is_numeric()` guard (which would render "Get 0 free articles"), I took the route your note pointed at: a newly created gate whose paid tier is active but grants 0 free views now generates the non-metering registration layout, so the metering sentence is dropped entirely rather than advertising an allowance the gate does not honor. This is at layout-generation time, so it covers the creation / duplication / missing-layout paths you noted; existing layouts are untouched.
- **Warning copy** — now states that 0 free views is the same as turning Metering off.

Tests: `Metering::is_gate_metered()` per-audience cases + a generated-layout assertion (metered tier advertises free articles, 0-view tier does not) on the PHP side; `isGateMetered` and the updated warning on the JS side. Full `Test_Metering` suite green (20 tests) and the content-gates JS suites green (3 suites / 11 tests).

Rebasing onto `main` to clear the conflict next.

dkoo:
Thanks @adekbadek — the metering-consistency follow-up looks great, and folding it in here rather than a separate PR works for me. I re-reviewed the delta against the three call sites from my earlier note and everything lines up:

- **Negative floor** — closed on both sides: `sanitize_metering()` now `max( 0, intval() )`, and the control's `onChange` floors and rounds (`Math.max( 0, Math.round(...) )`) so a raw keystroke can't put a negative or fractional count into gate state ahead of the commit clamp.
- **`Content_Gate::is_metering_enabled()`** — the old `isset( $gate['metering'] )` branch never matched on the modern gate CPT, so it always returned `false`. Routing it through `Metering::is_gate_metered()` fixes that and makes it count-aware.
- **`hasMetering` in `content-gates.tsx`** — now `gates.some( isGateMetered )`, and `isGateMetered` mirrors the PHP predicate (`active && enabled && count > 0` per section), so the wizard's countdown-banner affordance no longer disagrees with what the front end renders.
- **Generated layout copy** — gating the `pay-wall-one-tier-metering` selection in `get_layout_default_content()` on `count > 0` is a cleaner fix than validating inside the pattern: a 0-view gate now falls to `registration-wall` instead of advertising free articles it never grants, and you sidestep the "Get 0 free articles" wording entirely.

Introducing `Metering::is_gate_metered()` / `isGateMetered()` as a shared predicate is the right call — it makes "on with 0 views" and "off" consistent everywhere the check matters, and the new tests (`test_metering_with_zero_count`, `test_sanitize_metering_floors_negative_count`, the per-section `isGateMetered` cases) pin each of these as regression guards. The updated warning copy reads well too.

Confirmed the `has_metering()` refactor is behavior-preserving on its own path (the settings getters already zero a disabled section's count, so the old combined check and the new per-section one coincide). No further changes needed from my side — resolving my earlier threads.

dkoo (approved):

PR #708, opened 2026-07-23
fix(content-gate): surface the Institutions entry point when institutions exist (NPPD-1492)

adekbadek:
### All Submissions:

From internal validation (scenario I-01): the Institutions entry point was hidden behind the section kebab menu regardless of use — "Okay to hide when publisher is not using, but when using, should show."

- `get_config()` adds `has_institutions` to the existing wizard config payload (reuses the 300s institutions transient — no extra request; the transient is invalidated on institution save/delete, so the flag is fresh).
- With ≥1 institution, "Institutions" renders as a visible link (`href="#/institutions"`, a true anchor via the `SectionHeader` `secondaryAction` slot) beside the section title, and the kebab keeps only Advanced Settings. With none, the kebab placement is unchanged.
- Creating the first institution promotes the header link without a page reload (the institutions list fetch syncs the wizard store).
- Side effect: the explicit typing needed for the restructure fixes a pre-existing repo TS error (net −1; zero new errors, verified against the baseline type-check).

Closes NPPD-1492.

1. Audience > Access Control with no institutions: "Institutions" is in the section kebab, as before.
2. Create an institution (kebab → Institutions → Add new): on returning to Access Control, "Institutions" appears as a visible link beside the section title — no reload needed — and the kebab now holds only Advanced Settings.
3. The link navigates to the institutions list; middle-click opens it in a new tab.
4. Delete the institution: the entry returns to the kebab.

### Other information:

* [x] Have you written new tests for your changes, as applicable? (RTL tests for both header states, red/green)
* [x] Have you successfully run tests with your changes locally? (jest 46 suites / 431 tests; `n test-php --group content-gate` and `--group Access_Rules` green; PHPCS + ESLint clean)

**Known corner (pre-existing, filed as NPPD-2114):** with institutions but zero gates, the wizard shows the onboarding screen and no Institutions entry point exists anywhere — unchanged by this PR.

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
+++ b/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
@@ -402,7 +402,7 @@ class Audience_Content_Gates extends Wizard {
 				'content_gifting'   => Content_Gifting::get_settings(),
 				'advanced_settings' => $advanced_settings_response,
 				'has_newsletters'   => Reader_Activation::is_esp_configured(),
-				'has_institutions'  => count( Institution::get_cached_institutions() ) > 0,
+				'has_institutions'  => Institution::has_institutions(),
 			],
 		];
 		return rest_ensure_response( $config );
--- a/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
+++ b/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
@@ -402,6 +402,7 @@ class Audience_Content_Gates extends Wizard {
 				'content_gifting'   => Content_Gifting::get_settings(),
 				'advanced_settings' => $advanced_settings_response,
 				'has_newsletters'   => Reader_Activation::is_esp_configured(),
+				'has_institutions'  => count( Institution::get_cached_institutions() ) > 0,
 			],
 		];
 		return rest_ensure_response( $config );
--- a/plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
@@ -10,6 +10,11 @@
  */
 import { render } from '@testing-library/react';
 
+/**
+ * Internal dependencies
+ */
+import ContentGates from './content-gates';
+
 // mock-prefixed so Jest's hoisted jest.mock factories may close over them.
 const mockSetHeaderData = jest.fn();
 let mockWizardData = {};
@@ -34,22 +39,35 @@ jest.mock( '@wordpress/data', () => ( {
 	useSelect: () => ( {} ),
 } ) );
 
+// Passthrough only the components the view actually uses. The real
+// @wordpress/components cannot be loaded in this jsdom env (its data-store
+// side effects throw at import), so instead of a plain object — where a newly
+// imported component reads back as undefined and fails deep in React with an
+// opaque "Element type is invalid" — wrap the mock in a Proxy that throws for
+// any export the test has not provided, naming the missing one.
+const failLoudlyMock = ( moduleName, exports ) =>
+	new Proxy( exports, {
+		get( target, prop ) {
+			if ( prop in target || typeof prop === 'symbol' || prop === '__esModule' ) {
+				return target[ prop ];
+			}
+			throw new Error(
+				`content-gates.test mock of '${ moduleName }' has no '${ String( prop ) }'. ` + 'Add it to the mock in content-gates.test.js.'
+			);
+		},
+	} );
+
 jest.mock( '@wordpress/components', () => {
 	const React = require( 'react' );
 	// forwardRef because the component attaches a ref to VStack.
 	const Passthrough = React.forwardRef( ( { children }, ref ) => React.createElement( 'div', { ref }, children ) );
-	return {
-		__experimentalVStack: Passthrough,
-	};
+	return failLoudlyMock( '@wordpress/components', { __experimentalVStack: Passthrough } );
 } );
 
 jest.mock( '../../../../../packages/components/src', () => {
 	const React = require( 'react' );
 	const Passthrough = ( { children } ) => React.createElement( 'div', null, children );
-	return {
-		Divider: Passthrough,
-		Grid: Passthrough,
-	};
+	return failLoudlyMock( 'packages/components/src', { Divider: Passthrough, Grid: Passthrough } );
 } );
 
 jest.mock( '../../../../../packages/components/src/wizard/store/utils', () => ( {
@@ -67,32 +85,79 @@ jest.mock( './content-gate-settings', () => () => null );
 jest.mock( './advanced-settings', () => () => null );
 jest.mock( './settings-card', () => () => null );
 
-const ContentGates = require( './content-gates' ).default;
+const gatesOfLength = length =>
+	Array.from( { length }, ( _unused, index ) => ( {
+		id: index + 1,
+		title: `Gate ${ index + 1 }`,
+		status: 'publish',
+		priority: index,
+	} ) );
 
-const singleGate = [ { id: 1, title: 'Gate A', status: 'publish', priority: 0 } ];
+const lastHeaderData = () => {
+	const { calls } = mockSetHeaderData.mock;
+	if ( calls.length === 0 ) {
+		throw new Error( 'setHeaderData was never called; the header effect did not run.' );
+	}
+	return calls[ calls.length - 1 ][ 0 ];
+};
 
-const lastHeaderData = () => mockSetHeaderData.mock.calls[ mockSetHeaderData.mock.calls.length - 1 ][ 0 ];
+const menuLabels = headerData => headerData.sectionMenu.map( item => item.label );
 
 describe( 'Content Gates header — Institutions entry point (NPPD-1492)', () => {
 	beforeEach( () => {
 		mockSetHeaderData.mockReset();
 	} );
 
-	it( 'keeps Institutions in the kebab menu when the site has no institutions', () => {
-		mockWizardData = { gates: singleGate, config: { has_institutions: false } };
+	// The full gate-count × institutions matrix: Gate Priority appears only
+	// with more than one gate, and Institutions moves between the kebab menu
+	// and the promoted secondary action depending on whether any exist.
+	it.each( [
+		{
+			name: 'one gate, no institutions',
+			gateCount: 1,
+			hasInstitutions: false,
+			expectedMenu: [ 'Institutions', 'Advanced Settings' ],
+			expectSecondary: false,
+		},
+		{
+			name: 'two gates, no institutions',
+			gateCount: 2,
+			hasInstitutions: false,
+			expectedMenu: [ 'Gate Priority', 'Institutions', 'Advanced Settings' ],
+			expectSecondary: false,
+		},
+		{
+			name: 'one gate, institutions present',
+			gateCount: 1,
+			hasInstitutions: true,
+			expectedMenu: [ 'Advanced Settings' ],
+			expectSecondary: true,
+		},
+		{
+			name: 'two gates, institutions present',
+			gateCount: 2,
+			hasInstitutions: true,
+			expectedMenu: [ 'Gate Priority', 'Advanced Settings' ],
+			expectSecondary: true,
+		},
+	] )( 'builds the header menu for $name', ( { gateCount, hasInstitutions, expectedMenu, expectSecondary } ) => {
+		mockWizardData = { gates: gatesOfLength( gateCount ), config: { has_institutions: hasInstitutions } };
 		render( <ContentGates updateGatesData={ () => {} } /> );
 
 		const headerData = lastHeaderData();
-		expect( headerData.sectionMenu.map( item => item.label ) ).toContain( 'Institutions' );
-		expect( headerData.sectionSecondaryAction ).toBeUndefined();
-	} );
+		// Assert the whole ordered menu, so the chained conditional order is pinned.
+		expect( menuLabels( headerData ) ).toEqual( expectedMenu );
 
-	it( 'promotes Institutions to a visible header action when the site has institutions', () => {
-		mockWizardData = { gates: singleGate, config: { has_institutions: true } };
-		render( <ContentGates updateGatesData={ () => {} } /> );
-
-		const headerData = lastHeaderData();
-		expect( headerData.sectionSecondaryAction ).toEqual( expect.objectContaining( { label: 'Institutions', href: '#/institutions' } ) );
-		expect( headerData.sectionMenu.map( item => item.label ) ).not.toContain( 'Institutions' );
+		if ( expectSecondary ) {
+			// Pin the href, not just the label: the promotion changed this
+			// entry from an action callback to an href, and a label-only
+			// assertion would still pass on a dead link.
+			expect( headerData.sectionSecondaryAction ).toEqual( expect.objectContaining( { label: 'Institutions', href: '#/institutions' } ) );
+		} else {
+			expect( headerData.sectionSecondaryAction ).toBeUndefined();
+			// The kebab entry must also still navigate, not just carry a label.
+			const institutionsItem = headerData.sectionMenu.find( item => item.label === 'Institutions' );
+			expect( institutionsItem ).toEqual( expect.objectContaining( { href: '#/institutions' } ) );
+		}
 	} );
 } );
--- a/plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
@@ -1,0 +1,98 @@
+/**
+ * NPPD-1492 — the "Institutions" entry point on the Access Control screen must
+ * be plainly visible (header secondary action) when the site has at least one
+ * institution, and may stay tucked in the kebab menu only while the publisher
+ * is not using institutions.
+ */
+
+/**
+ * External dependencies
+ */
+import { render } from '@testing-library/react';
+
+// mock-prefixed so Jest's hoisted jest.mock factories may close over them.
+const mockSetHeaderData = jest.fn();
+let mockWizardData = {};
+
+jest.mock( '../../../hooks/use-wizard-api-fetch', () => ( {
+	useWizardApiFetch: () => ( {
+		wizardApiFetch: jest.fn(),
+		isFetching: false,
+		errorMessage: null,
+		resetError: jest.fn(),
+	} ),
+} ) );
+
+jest.mock( '@wordpress/data', () => ( {
+	useDispatch: () => ( {
+		addNotice: jest.fn(),
+		resetNotices: jest.fn(),
+		resetHeaderData: jest.fn(),
+		setHeaderData: ( ...args ) => mockSetHeaderData( ...args ),
+		updateWizardSettings: jest.fn(),
+	} ),
+	useSelect: () => ( {} ),
+} ) );
+
+jest.mock( '@wordpress/components', () => {
+	const React = require( 'react' );
+	// forwardRef because the component attaches a ref to VStack.
+	const Passthrough = React.forwardRef( ( { children }, ref ) => React.createElement( 'div', { ref }, children ) );
+	return {
+		__experimentalVStack: Passthrough,
+	};
+} );
+
+jest.mock( '../../../../../packages/components/src', () => {
+	const React = require( 'react' );
+	const Passthrough = ( { children } ) => React.createElement( 'div', null, children );
+	return {
+		Divider: Passthrough,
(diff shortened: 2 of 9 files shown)


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`, and the bot's `Merge origin/alpha` commit has been removed from the branch tip via force-push — no other commits were touched. If you pulled the branch after the retarget, refresh with `git fetch origin && git reset --hard origin/nppd-1492-i-01-create-an-institution` (stash local work first).

plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
```diff
@@ -402,6 +402,7 @@ public function get_config() {
 				'content_gifting'   => Content_Gifting::get_settings(),
 				'advanced_settings' => $advanced_settings_response,
 				'has_newsletters'   => Reader_Activation::is_esp_configured(),
+				'has_institutions'  => count( Institution::get_cached_institutions() ) > 0,
```
dkoo: **Suggestion:** `count()` on a value only guaranteed to be "not false" — `get_cached_institutions()` returns the transient verbatim whenever it isn't `false`, and only falls through to `rebuild_cache()` (which always returns and stores an array) otherwise. So nothing in this repo can make this throw — I checked, and this is not a blocker.

What makes it worth a line: this is the only one of that method's call sites that would *hard-fail* on a non-array. The other three use `foreach`/`isset`, which degrade to a warning. `get_transient()` runs the `pre_transient_newspack_institutions` and `transient_newspack_institutions` filters, so any third-party plugin or drop-in returning a scalar there turns this into a PHP 8 `TypeError` — and because this endpoint returns gates, metering, content gifting and advanced settings in one payload, the failure takes down the whole Access Control screen rather than hiding one link.

`! empty( Institution::get_cached_institutions() )` is exactly equivalent, no more expensive, expresses existence rather than cardinality, and cannot throw. Cleanest form would be an `Institution::has_institutions()` helper so the wizard stops depending on the cache's return shape at all.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+const singleGate = [ { id: 1, title: 'Gate A', status: 'publish', priority: 0 } ];
+
+const lastHeaderData = () => mockSetHeaderData.mock.calls[ mockSetHeaderData.mock.calls.length - 1 ][ 0 ];
+
+describe( 'Content Gates header — Institutions entry point (NPPD-1492)', () => {
+	beforeEach( () => {
+		mockSetHeaderData.mockReset();
+	} );
+
```
dkoo, line 110: **Suggestion:** The gate-count × institutions matrix is untested — Neither test varies `gates.length`, so the interaction between the "Gate Priority" unshift (which only appears with more than one gate) and the institutions branch is uncovered — and that chained-`unshift` ordering is exactly where a future edit would regress. The four combinations are cheap to table-drive: with 1 gate and no institutions the menu should be `[ Institutions, Advanced Settings ]`; with 2+ gates and no institutions, `[ Gate Priority, Institutions, Advanced Settings ]`; with institutions present, Institutions drops out of the menu and appears as the secondary action in both cases.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
```diff
@@ -0,0 +1,98 @@
...
+	it( 'keeps Institutions in the kebab menu when the site has no institutions', () => {
+		mockWizardData = { gates: singleGate, config: { has_institutions: false } };
+		render( <ContentGates updateGatesData={ () => {} } /> );
+
+		const headerData = lastHeaderData();
+		expect( headerData.sectionMenu.map( item => item.label ) ).toContain( 'Institutions' );
```
dkoo: **Suggestion:** The kebab test asserts labels only, so the new `href` is unpinned — The no-institutions test checks `sectionMenu.map( item => item.label )`, which verifies the entry survives in the kebab but not that it still navigates. That matters specifically because this PR changed that item's mechanism from `action: () => history.push( '/institutions' )` to `href: '#/institutions'` — drop the `href` or rename the key and the kebab renders a dead "Institutions" item with the suite still green.

Asserting the same shape you already assert in the positive case — `expect.objectContaining( { label: 'Institutions', href: '#/institutions' } )` — closes it.

Worth noting for the record: the second test genuinely fails against the old code (which always unshifted Institutions and never set `sectionSecondaryAction`), so the coverage is real. The first test would have passed against the old code too — a regression guard rather than a red/green proof, which is fine as long as it's understood that way.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
```diff
@@ -48,16 +45,21 @@ const ContentGates = ( { updateGatesData }: { updateGatesData: ( gates: Gate[] )
...
+			label: __( 'Institutions', 'newspack-plugin' ),
+			href: '#/institutions',
+		};
+		const sectionMenu: { label: string; action?: () => void; href?: string }[] = [
 			{
 				label: __( 'Advanced Settings', 'newspack-plugin' ),
 				action: () => setShowAdvancedSettings( true ),
 			},
 		];
+		// Institutions in use get a visible entry point next to the section
```
dkoo: **Suggestion:** The menu is assembled tail-first and reads back-to-front — The array literal starts with only Advanced Settings, then Institutions is `unshift`ed under a negated condition, then Gate Priority is `unshift`ed under another — so a reader has to replay three mutations across four combinations to work out that the display order is `[ Gate Priority?, Institutions?, Advanced Settings ]`.

Building it declaratively in display order states the final order literally and removes the mutation-order reasoning:

```ts
const sectionMenu = [
 ...( gates.length > 1 ? [ gatePriorityItem ] : [] ),
 ...( ! hasInstitutions ? [ institutionsLink ] : [] ),
 advancedSettingsItem,
];
```

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
```diff
@@ -78,8 +80,9 @@ const ContentGates = ( { updateGatesData }: { updateGatesData: ( gates: Gate[] )
 				'newspack-plugin'
 			),
 			sectionMenu,
+			sectionSecondaryAction: hasInstitutions ? institutionsMenuItem : undefined,
```
dkoo: **Suggestion:** The promoted link renders to the right of the kebab menu — `SectionHeader` emits the kebab (`index.js:105-122`) *before* the secondary action (`index.js:123-129`) inside the same flex title container, so with institutions present the row reads `Access Control ⋮ Institutions` — an overflow menu sandwiched between the title and a text link, rather than the conventional "actions first, overflow last" order.

The PR says the link renders "beside the section title", which is true, but it lands to the right of the very menu it was promoted out of. If it reads wrong on screen, the fix is a JSX reorder in the shared `SectionHeader`, which is a scope decision better made deliberately than discovered post-merge.

**Runtime note — this could not be confirmed visually.** The verification pass was unable to establish a browser session in the review environment (see the verification section), so nobody has yet looked at the rendered screen. The element *order* is not in doubt — it follows directly from the JSX and the compiled `commons.js` — but whether it actually reads as misplaced is still an open question, and it is the only thing standing between this finding and being purely cosmetic. **One glance at the rendered header settles it.**

**One part of the original note was wrong and is withdrawn.** I claimed the SCSS switches the secondary action to `margin-top: 8px` at ≤782px. It does not — that rule is ungated. In the source (`packages/components/src/section-header/style.scss`) and in the compiled `dist/commons.css` alike, `.newspack-section-header__title-container .newspack-section-header__secondary-action { margin-top: 8px }` sits at the top level with no enclosing media query, as does `.newspack-section-header__secondary-action { display: flex; margin-left: 8px }`. The only media queries in that stylesheet target `--has-primary-action`, `--has-back-nav`, `__back-nav` and `__primary-action`; none touch this slot. So there is no responsive switch to inspect.

That does leave a different narrow-viewport question worth the same single glance: `__title-container` is `display: flex` with **no** `flex-wrap` at every width, so the heading, the kebab and the promoted link share one line at all sizes and simply compress rather than wrapping. It remains true that this PR is the first thing in the codebase to use the `secondaryAction` slot, so its styling has never rendered in production at any width.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/index.tsx
```diff
@@ -69,7 +70,17 @@ export default function Institutions() {
 	const fetchData = useCallback( () => {
 		setIsLoading( true );
 		apiFetch< Institution[] >( { path: `${ API_PATH }?per_page=100&context=edit&_embed=wp:featuredmedia` } )
-			.then( setData )
+			.then( institutions => {
+				setData( institutions );
+				// Keep the gates screen header in sync: it promotes the
+				// Institutions entry point out of the kebab menu when the site
+				// has at least one institution.
+				updateWizardSettings( {
```
dkoo: **Suggestion:** The load-bearing half of the no-reload behavior is the untested half — This store write is what the PR's headline claim rests on. `getWizardAPIData` is a resolver, so the gates config is fetched once per page load and never re-fetched on hash navigation — without this write, a publisher who creates their first institution and clicks back to Access Control would see a stale header for the rest of the session. The rendering half got two tests; this half has none, and it hard-codes another screen's wizard slug plus a two-segment store path. If someone renames the config key or restructures the payload, it breaks silently: no error, no failing test, the header just stops updating.

A test asserting `updateWizardSettings` is dispatched with `{ slug, path: [ 'config', 'has_institutions' ], value }` for both the empty and non-empty fetch results would close the loop.

Second, smaller point on the same line: the dispatch fires on every list fetch — mount and after each delete — even when the derived boolean hasn't changed, and `UPDATE_WIZARD_SETTINGS` JSON-clones the entire wizard payload each time. Guarding on a change (or tracking the last-written value in a ref) makes it a no-op in the common case.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
```diff
@@ -0,0 +1,98 @@
...
+		addNotice: jest.fn(),
+		resetNotices: jest.fn(),
+		resetHeaderData: jest.fn(),
+		setHeaderData: ( ...args ) => mockSetHeaderData( ...args ),
+		updateWizardSettings: jest.fn(),
+	} ),
+	useSelect: () => ( {} ),
+} ) );
+
+jest.mock( '@wordpress/components', () => {
```
dkoo, line 60: **Nit:** The mocks fail in an unhelpful direction — The `@wordpress/components` mock returns only `__experimentalVStack`, and the `packages/components/src` mock only `Divider` and `Grid`. The day `content-gates.tsx` imports one more component from either, the test fails with `Element type is invalid: got undefined` — pointing at React rather than at the mock, which is a confusing hour for whoever touches this next. `{ ...jest.requireActual( '@wordpress/components' ), __experimentalVStack: Passthrough }` keeps the same isolation while failing loudly for the right reason.

Two smaller things in the same file: `require( './content-gates' ).default` isn't necessary — `jest.mock` factories are hoisted above ESM imports, so a normal `import` works and matches the rest of the suite — and the `lastHeaderData()` helper throws a raw `TypeError` rather than a readable assertion if `setHeaderData` was never called.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
```diff
@@ -39,6 +35,7 @@ const ContentGates = ( { updateGatesData }: { updateGatesData: ( gates: Gate[] )
 	const gates = ( wizardData?.gates || [] ) as Gate[];
 	const config = ( wizardData?.config || {} ) as GateSettings;
 	const hasMetering = gates.some( gate => gate.registration?.metering?.enabled || gate.custom_access?.metering?.enabled );
+	const hasInstitutions = !! config.has_institutions;
 
 	useEffect( () => {
```
dkoo, line 40: **Nit:** Institutions with zero gates still has no entry point — The effect early-returns with `resetHeaderData()` whenever there are no gates, and the onboarding screen renders instead — so a publisher who has institutions but hasn't created a gate yet has no route to `#/institutions` from this screen, neither kebab nor header link.

You disclose this in the PR body as pre-existing and filed as NPPD-2114, and that's accurate — the early return predates this change. Recording it only because it's the same discoverability class of bug NPPD-1492 is fixing, and because it surfaced independently during review, which is worth knowing as confirmation rather than a new ask.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
```diff
@@ -48,16 +45,21 @@ const ContentGates = ( { updateGatesData }: { updateGatesData: ( gates: Gate[] )
 			resetHeaderData();
 			return;
 		}
-		const sectionMenu = [
-			{
-				label: __( 'Institutions', 'newspack-plugin' ),
-				action: () => history.push( '/institutions' ),
-			},
+		const institutionsMenuItem: { label: string; action?: () => void; href?: string } = {
```
dkoo: **Nit:** `institutionsMenuItem` is misnamed in the branch that matters, and its type is hand-rolled twice — In the branch this PR exists for, the object is not a menu item at all — it's the header's secondary action (line 83), so a reader arriving at that line has to scroll back to discover that the thing named "menu item" is being used as a header link. `institutionsLink` or `institutionsEntryPoint` would say it.

The inline structural type `{ label: string; action?: () => void; href?: string }` is also written out twice (lines 48 and 52), with an `action?` member this object never sets — immediately next to `HeaderAction` in `types/index.d.ts`, a superset of it, in a file this PR already edits. A shared `SectionMenuItem` (or `Omit<HeaderAction, 'type'>`) would remove the duplication and give the previously-undocumented `sectionSecondaryAction` contract a name — which matters more than usual here, since the wizard store is plain JS and this is the only place the shape is written down.

dkoo (approved):
Release risk: LOW — admin-UI discoverability only, no contract files touched, no change to who can access gated content, and the automated checks are clean.

No blockers. The riskier-looking parts of this change all held up under direct reading: the new PHP call can't fatal on a missing import (same namespace, unconditionally loaded), cache invalidation genuinely covers every mutation path including trash and untrash, the new store write can't cause a refetch loop, and the `secondaryAction` slot renders a real anchor. I also confirmed the "net −1 TypeScript errors" claim empirically rather than taking it on trust: 211 errors at the merge-base, 210 at the head, one removed and none added.

The design deserves a note in its favor, because it looks like redundancy and isn't. The wizard's data fetch is a resolver — it runs once per page load and never re-runs on in-page navigation — so the server flag is needed (the gates screen never fetches the institutions list itself) *and* the client store write is needed (nothing else refreshes the flag after the first institution is created). That makes the store write the load-bearing half of the headline no-reload behavior, and it's the half with no test.

Two things worth resolving before merge. First, the shared header renders the kebab menu *before* the secondary action inside the same container, so with institutions present the row reads `Access Control ⋮ Institutions` — the link lands to the right of the menu it was promoted out of, which works against the discoverability goal. That deserves a look at the rendered screen; if it reads wrong, the fix is a reorder in the shared component, which is a scope decision worth making deliberately. Second, `count( Institution::get_cached_institutions() )` is the only one of that method's call sites that would hard-fail on a non-array; `! empty( ... )` is exactly equivalent, costs nothing, and degrades to "link stays in the kebab" instead of a 500 on a payload that also carries gates, metering and advanced settings.

The rest are test-coverage and readability suggestions, plus one pre-existing race in `Institution::create()` (rule meta is written *after* the insert that clears the cache) that's worth its own ticket — it fails safe, denying institutional access rather than granting it, but it does touch real access decisions.

- `plugins/newspack-plugin/includes/content-gate/class-institution.php:107` — **Suggestion:** Pre-existing: `create()` invalidates the cache before writing rule meta — Not introduced here, but directly adjacent to this PR's "the transient is invalidated on save, so the flag is fresh" claim, so it's worth recording. `Institution::create()` calls `wp_insert_post()` — which fires `save_post_np_institution` and clears the transient — and only *then* writes the `email_domain` / `ip_range` / `reader_data` meta in the loop below. A concurrent read landing between those two steps repopulates the 300s transient with the institution present but its access rules empty, and that wrong payload survives the full TTL. It does not affect `has_institutions` (existence only, so the new flag is safe), but it does affect `check_ip()` and `evaluate()` — real access decisions. It fails in the safe direction: the IP check returns the original value for an empty rule, so affected readers are denied institutional access rather than granted it. A second `self::invalidate_cache()` after the meta loop would close it. Worth its own ticket rather than expanding this PR.

plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
adekbadek: Done — added `Institution::has_institutions()` (`! empty( get_cached_institutions() )`) and the wizard call site uses it instead of `count()`, so a filtered non-array cache can no longer take down the payload. A PHP test covers the non-array path.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
adekbadek: Done — replaced both tests with a table-driven `it.each` over all four gate-count × institutions combinations, asserting the whole ordered menu each time.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
adekbadek: Done — the promoted case asserts `expect.objectContaining( { label: 'Institutions', href: '#/institutions' } )`, and the in-menu case now asserts the kebab item still carries the `href`.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
adekbadek: Done — the menu is now built declaratively in display order with spread conditionals, exactly as suggested.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
adekbadek: Confirmed on screen: it did render `Access Control ⋮ Institutions`, the link to the right of the kebab. Fixed by rendering the secondary action before the overflow menu in the shared `SectionHeader`. content-gates is the only consumer of that slot across the whole monorepo (no other `secondaryAction=` / `sectionSecondaryAction` / `section.secondaryAction`), so the reorder is a no-op for every other section. Re-verified in the browser: the link now sits left of the kebab (x 336-403 vs 411-447), conventional actions-first order. Withdrawn SCSS note noted — no responsive rule touched.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/institutions/index.tsx
adekbadek: Done — added a test asserting `updateWizardSettings` is dispatched with `{ slug, path: [ 'config', 'has_institutions' ], value }` for both the empty and non-empty fetch. Also guarded the write behind a per-instance ref, so a fetch that leaves the derived boolean unchanged no longer clones the wizard payload.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.test.js
adekbadek: Done — `require('./content-gates').default` → a normal `import`, and `lastHeaderData()` now throws a readable message when `setHeaderData` was never called. On the mock: the `requireActual` spread can't load here — the real `@wordpress/components` throws `combineReducers is not a function` under jsdom — so instead each mock is wrapped in a Proxy that throws naming the missing export. Same fail-loudly-for-the-right-reason outcome without importing the real module.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
adekbadek: Acknowledged — pre-existing early-return, tracked as NPPD-2114 as you note. No change here; recording it as confirmation.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/content-gates.tsx
adekbadek: Done — renamed to `institutionsLink`, and the shape is now a shared `SectionMenuItem` (`Omit<HeaderAction, 'type'>`) declared next to `HeaderAction`, replacing the two hand-rolled inline types.

PR #710, opened 2026-07-23
ci: sync design-review labels and reviewer requests (NPPM-307)

dkoo:
Adds `.github/workflows/design-review-sync.yml`, keeping the design-review **labels** and the design **reviewer request** in sync automatically, in both directions.

Resolves [NPPM-307](https://linear.app/a8c/issue/NPPM-307/design-review-github-action).

## Behaviors

| # | Trigger | Effect |
|---|---|---|
| 1 | `[Status] Needs Design Review` label added | Remove `[Status] Design Approved`; request the design reviewer |
| 2 | Review requested from the `newspack-design` **team** *or* the design reviewer individually | Add `[Status] Needs Design Review`; remove `[Status] Design Approved` |
| 3 | `[Status] Design Approved` label added | Remove `[Status] Needs Design Review`; remove `[Status] Needs Changes or Feedback` **only if** no reviewer's latest review is `CHANGES_REQUESTED` |
| 4 | Design reviewer submits an **APPROVE** review | Add `[Status] Design Approved`, then the same cleanup as #3 |

## Why a hardcoded individual reviewer instead of the team

The automatic `GITHUB_TOKEN` is **repo-scoped**. It cannot resolve or request an org **team** — `pulls.requestReviewers({ team_reviewers: ['newspack-design'] })` fails with `422 "Could not resolve to a node"` — and it cannot read org team membership. Neither permission can be granted to `GITHUB_TOKEN` (there is no org/members key in the `permissions:` set). This was confirmed empirically, not assumed.

What *does* work with `GITHUB_TOKEN` is requesting an **individual** reviewer and comparing a review author's **login**. So the workflow targets `thomasguillot`, currently the sole member of `@Automattic/newspack-design`, as a stand-in for the team. This keeps the whole feature **secret-free**.

⚠️ **Maintenance:** if that username changes, or the design team gains/loses members, update the `DESIGN_REVIEWER` constant — otherwise behaviors 1 and 4 silently target the wrong person (or under-route to only one of several members). The alternative is migrating to an org-scoped token / GitHub App that can address the team directly. This is documented in a NOTE in the workflow itself.

## Notes

- **No secret required**; `permissions: { contents: read, pull-requests: write }`.
- **No loop guards needed:** GitHub does not start new workflow runs from events caused by the default `GITHUB_TOKEN`, so behavior 1's reviewer request does not re-fire behavior 2, and behavior 4's label add does not re-fire behavior 3.
- Requesting a reviewer never clears existing reviewers (the API is additive), and requesting the PR author is skipped.
- Label writes tolerate a read-only token (fork-head PRs) with a warning instead of failing the run.
- Mirrors the existing `pr-labels.yml` pattern (`actions/github-script@v7`, single inline script, 404-tolerant label removal).

## Testing

Behaviors 1–3 were verified live on this PR against real workflow runs:

| Behavior | Result |
|---|---|
| 1 — label → reviewer requested, `Design Approved` cleared | ✅ run `30039869302` |
| 2 — team review requested → label added | ✅ run `30023705740` |
| 2 — individual reviewer requested → label added | ✅ run `30027623154` |
| 3 — `Design Approved` → `Needs Design Review` cleared | ✅ run `30039892179` |

**Still to confirm on this PR:** behavior 4 — it needs an APPROVE submitted *by* `thomasguillot` (nobody can review their own PR, so it could not be self-tested). @thomasguillot, a single approve here should add `[Status] Design Approved` and clear `[Status] Needs Design Review`.

Also untested end-to-end: the `CHANGES_REQUESTED`-still-outstanding branch of #3/#4 (needs a second reviewer leaving an unresolved change request). The logic was reviewed against `listReviews` semantics — latest non-comment review per reviewer, with `DISMISSED` correctly treated as no longer outstanding.


adekbadek:
Undoing an automation false positive: #709 was the `main` → `alpha` release promotion, not a stacked-PR parent, so this PR should never have been retargeted (the retarget workflow is being patched). Base is back to `main`. Heads-up: the bot's `Merge origin/alpha` commit (5ea7a5a3f) is buried beneath your later commits, so I left the branch history untouched rather than rewrite your work. Content-wise the impact vs `main` should be negligible (`alpha` matched `main` at merge time apart from a duplicated CI commit), but feel free to rebase it out if you'd rather have a clean history.

thomasguillot (changes requested):
4.  Expected: [Status] Needs Changes or Feedback removed

this did not happen but I don't think that's too much of an issue?

thomasguillot (approved):

PR #717, opened 2026-07-24
fix(wizards): let tabs own their sub-view routes

adekbadek:
### All Submissions:

Follow-up to #672, which made `TabbedNavigation` render routed content as a sibling of the panels when no visible tab owns the route. That stopped wizard screens going blank, and its docblock names the intended next step:

> for a route that is conceptually a sub-view of a tab, give that tab an `activeTabPaths` entry … so it stays selected and its panel keeps owning the content.

No wizard had been updated to do that yet, so three screens still land on the fallback: the content renders, but with **no tab highlighted**, and because the content changes DOM parent when the route crosses the owned/unowned boundary, React **remounts the whole page subtree** — losing component state and re-firing effects and data fetches on every navigation.

This declares those routes:

| Wizard | Route | Fix |
| --- | --- | --- |
| Audience → Campaigns | `/campaigns/:id` (selecting or creating a campaign) | `exact: false` on the Campaigns tab, matching its sibling Segments tab |
| Advertising → Display Ads | `/google_ad_manager`, `/google_ad_manager/create`, `/google_ad_manager/:id` | `activeTabPaths: [ '/google_ad_manager/*' ]` on Providers |
| Audience → Plans | `/new`, `/edit/:id` | `activeTabPaths: [ '/new', '/edit/*' ]` on Subscriptions |

Campaigns uses `exact: false` rather than a wildcard because its route is already registered non-exactly (`<Route path="/campaigns/:id?">`), so prefix matching states the intent directly. The other two tabs sit at `path: '/'`, which `matchesRoute` always treats as exact, so a wildcard is the only mechanism available there.

Two tests are added for the behaviour that motivates all of this — that crossing the ownership boundary remounts the content, and that a claimed subpath keeps it in the panel with no remount. #672 already covers the fallback itself.

1. Go to **Audience → Campaigns**, click **Add Campaign** and add one.
2. The URL becomes `#/campaigns/<id>`. Confirm the **Campaigns** tab stays highlighted and the prompt list renders. On `main` the tab goes unhighlighted.
3. Go to **Advertising → Display Ads → Google Ad Manager**. Confirm the **Providers** tab stays highlighted, including on an individual ad unit.
4. Go to **Audience → Plans**, click **Add plan** (and edit an existing one). Confirm the **Subscriptions** tab stays highlighted.
5. `pnpm --filter newspack-components test` and `pnpm --filter newspack test`.

### Other information:

**Verification** — all three wizards browser-tested in an isolated env (`tabnav.test`), measuring `aria-selected` on each tab and whether the page content sits inside a `[role="tabpanel"]`, with only the consumer config differing between runs:

| Wizard / route | `main` today | With this PR |
| --- | --- | --- |
| Campaigns `#/campaigns/:id` | all tabs `false`, content **outside** panel | Campaigns `true`, content **in** panel |
| Advertising `#/google_ad_manager` | all tabs `false`, content **outside** panel | Providers `true`, content **in** panel |
| Advertising `#/google_ad_manager/create` | all tabs `false` | Providers `true` |
| Plans `#/new` | all tabs `false`, **`panelCount: 0`** | Subscriptions `true`, form **in** panel |

No regressions on the routes that already worked: Advertising `#/`, `#/placements`, `#/settings` and Plans `#/`, `#/donations`, `#/bundles` each keep exactly their own tab selected and their content in-panel.

Tests: `newspack-components` 97/97, `newspack-plugin` 429/429, prettier clean.

**Left alone deliberately:** the Campaigns wizard's Settings tab is `exact: true` against a non-exact `<Route path="/settings">`; no `/settings/*` route exists today, so it is unreachable. The Setup wizard's `/campaign` and `/complete` routes are owned by no tab and appear unlinked.

### DIFF
--- a/packages/components/src/tabbed-navigation/index.test.js
+++ b/packages/components/src/tabbed-navigation/index.test.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies.
  */
+import { useEffect } from 'react';
 import { act, fireEvent, render, screen } from '@testing-library/react';
 import { MemoryRouter, Redirect, useHistory } from 'react-router-dom';
 
@@ -207,6 +208,60 @@ describe( 'TabbedNavigation with routed items', () => {
 		act( () => history.push( '/budgets' ) );
 		expect( getTab( 'Budgets' ) ).toHaveAttribute( 'aria-selected', 'true' );
 		expect( screen.getByText( 'Routed content' ).closest( '[role="tabpanel"]' ).id ).toBe( getTab( 'Budgets' ).getAttribute( 'aria-controls' ) );
+	} );
+
+	it( 'remounts the content when a route stops being owned by a tab', () => {
+		// Leaving a tab's ownership moves the content out of its panel, so React
+		// remounts the subtree — losing its state and re-firing its effects. That
+		// cost is why a tab whose sub-views have their own paths should claim them
+		// via `activeTabPaths` rather than lean on the unowned-content fallback.
+		const historyRef = { current: null };
+		const mounts = { current: 0 };
+		const Content = () => {
+			useEffect( () => {
+				mounts.current++;
+			}, [] );
+			return <div>Routed content</div>;
+		};
+		render(
+			<MemoryRouter initialEntries={ [ '/stories' ] }>
+				<HistoryGrabber historyRef={ historyRef } />
+				<TabbedNavigation items={ [ { label: 'Stories', path: '/stories', exact: true } ] } content={ <Content /> } />
+			</MemoryRouter>
+		);
+		expect( screen.getByText( 'Routed content' ).closest( '[role="tabpanel"]' ) ).not.toBeNull();
+		expect( mounts.current ).toBe( 1 );
+
+		act( () => historyRef.current.push( '/stories/42' ) );
+		expect( screen.getByText( 'Routed content' ).closest( '[role="tabpanel"]' ) ).toBeNull();
+		expect( mounts.current ).toBe( 2 );
+	} );
+
+	it( 'keeps the content in the tab panel across subpaths it claims', () => {
+		// The complement of the test above, and the shape the wizard tabs in this
+		// change adopt: with the subpath claimed the content never leaves the panel,
+		// so no remount happens.
+		const historyRef = { current: null };
+		const mounts = { current: 0 };
+		const Content = () => {
+			useEffect( () => {
+				mounts.current++;
+			}, [] );
+			return <div>Routed content</div>;
+		};
+		render(
+			<MemoryRouter initialEntries={ [ '/stories' ] }>
+				<HistoryGrabber historyRef={ historyRef } />
+				<TabbedNavigation
+					items={ [ { label: 'Stories', path: '/stories', exact: true, activeTabPaths: [ '/stories/*' ] } ] }
+					content={ <Content /> }
+				/>
+			</MemoryRouter>
+		);
+		act( () => historyRef.current.push( '/stories/42' ) );
+		expect( getTab( 'Stories' ) ).toHaveAttribute( 'aria-selected', 'true' );
+		expect( screen.getByText( 'Routed content' ).closest( '[role="tabpanel"]' ) ).not.toBeNull();
+		expect( mounts.current ).toBe( 1 );
 	} );
 } );
 
--- a/packages/components/src/tabbed-navigation/index.test.js
+++ b/packages/components/src/tabbed-navigation/index.test.js
@@ -34,6 +34,31 @@ const renderTabs = ( { initialEntries = [ '/stories' ], ...props } = {} ) => {
 
 const getTab = name => screen.getByRole( 'tab', { name } );
 
+/**
+ * Render a single-tab nav whose content counts its own mounts, so a test can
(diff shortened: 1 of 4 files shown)


thomasguillot:
Reviewed while setting up an env for the Dynamic Pricing design review, so I checked the Plans leg in a browser rather than by reading. The mechanism is sound and the Campaigns fix is the most valuable part of this: `/campaigns/:id` is a high-traffic route that has been falling through to the unowned-content path on every campaign selection.

One thing to reconsider on Plans.

## The Plans claim only pays off for one of the three scopes

`/edit/:id` is a single scope-less route: `list.tsx` pushes `/edit/${item.id}` from every scope (L157, L320), and `product-form.tsx` hardcodes `backNav: '#/'` (L273). So the Subscriptions tab claims the edit route for donations and bundles too.

Verified in one continuous session (env on `main` + this PR's `index.tsx` hunk, reading `aria-selected` and the panel each content node sits in):

| Navigation | Panel before | Panel after | Tab selected after |
| --- | --- | --- | --- |
| Subscriptions list, Edit "Ambassador" | `r3` | `r3` | Subscriptions ✅ |
| Donations list, Edit "Donate: Monthly" | `r4` | `r3` | Subscriptions ❌ |

The Subscriptions row is the intended win: same panel, no remount, correct highlight.

The Donations row is the problem. The content still changes panel, so React still remounts the subtree exactly as it does today, and now a tab is highlighted that does not describe what the panel contains. `aria-selected="true"` on Subscriptions while its panel holds a donation's edit form is also announced that way by a screen reader. Plan bundles behaves the same.

So for two of the three scopes this trades "no tab selected, remount" for "wrong tab selected, remount" - the cost stays and an incorrect state signal is added.

Two ways out, either fine by me:

1. Make the edit routes scope-aware (`/donations/edit/:id`, `/bundles/edit/:id`) and give each tab its own claim. That fixes the highlight and the remount for all three scopes, and it also fixes the related wrinkle that saving a donation returns you to the Subscriptions list.
2. Drop the Plans hunk from this PR and land Advertising and Campaigns, where one tab genuinely owns the sub-routes. Plans gets its claim once the routes carry scope.

Worth asking whether the scope tabs should render at all on the add/edit screens - they are full-page forms with their own `SectionHeader` and back nav, so there is no tab content to switch between. That is pre-existing and out of scope here, but it is the same underlying question.

## Smaller notes

- The `activeBreadcrumbs` shadowing the comment flags is inert today: neither the `/` section nor the hidden `/new` and `/edit/:id` sections define `breadcrumbs`, so both before and after this change the trail falls through to the `headerText` fallback and then picks up the render-time "Add plan" / "Edit plan" leaf. Confirmed unchanged in the browser. It becomes real the moment anyone adds `breadcrumbs` to the `/` section, which is worth a word in the comment since the current wording reads like an accepted regression rather than a dormant one.
- `'/edit/*'` also claims bare `/edit` through `matchesWildcard`, which is not a registered route. Harmless, the `<Switch>` redirect catches it, just noting the pattern is broader than the intent.
- The second test pushes before its first assertion, so it only proves the count is 1 after navigating. Asserting 1 both before and after would mirror the first test and rule out a compensating double-mount. The `Content` and `mounts` harness is also duplicated between the two tests and could be hoisted.
- I did not exercise the Advertising leg. Providers owns `/google_ad_manager/*` unambiguously there, so I would expect it to behave like the Subscriptions row above.

adekbadek:
On the `activeBreadcrumbs` shadowing reading as accepted rather than dormant — reworded in f04f3d3. The comment now says why it is inert: no section in this wizard declares `breadcrumbs` at all, so `activeBreadcrumbs` returns `[]` whichever section wins the lookup and every trail falls back to `headerText`. It also names the condition that would wake it up — the add/edit screens would inherit the Subscriptions trail the moment any section here gains one.

---

Good catch on the claimed-subpath test proving only the post-navigation count — it now asserts the content is in the panel and `mounts.current` is 1 *before* the push as well, so a mount lost and regained across the navigation would show up. That mirrors the boundary-crossing test above it, and the two now differ only in the `activeTabPaths` entry and the expected count after the push.

---

The duplicated `Content`/`mounts` harness is hoisted into a `renderMountCountingTabs( item )` helper next to `renderTabs`, which builds the counter and the single-tab render and hands back `{ history, mounts }`. Each test is down to the tab item it needs plus its assertions.

adekbadek:
On the Plans claim — went with the second option: dropped the Plans hunk in 77c2436, so this PR now lands only Advertising and Campaigns, where one tab genuinely owns its sub-routes.

Doing it the scope-aware way is really its own change — it has to thread scope through `list.tsx`, `product-edit.tsx`, `product-form.tsx` and `index.tsx` (the push, the `backNav`, per-tab claims), and done properly it also settles the save-returns-to-the-wrong-list wrinkle and the "should the scope tabs render on a full-page edit form at all" question you raised. That doesn't belong bolted onto a regression fix. Plans is behind `NEWSPACK_PLANS_UI` as well, so the wrong-highlight isn't reaching anyone today — it gets its claim in a follow-up once the edit routes carry scope.

Advertising's `/google_ad_manager/*` is unambiguous as you expected, and Campaigns — the `/campaigns/:id` fix — is untouched by any of this.

PR #721, opened 2026-07-24
docs: note worktree JS build silent-failure mode

adekbadek:
### All Submissions:

Adds a short note to `AGENTS.md`, under "Isolated environments", documenting a failure mode of `bin/build-repos.sh`: building a worktree-mounted plugin's JS inside an env container can exit non-zero with completely empty stdout/stderr, while also having already deleted `dist/` (the plugin's `build` script runs `clean` first). One observed cause was the worktree's own `node_modules` being out of step with the main checkout's install — the note gives the diagnosis (rerun the build on the host from the worktree root, which surfaces the real error) and recovery (`pnpm install` at the worktree root, then rebuild) steps.

This was observed once, not established as routine, so the note is scoped as a failure mode and its recovery rather than a required setup step.

Whether the underlying tooling behavior (deleting `dist/` before the build succeeds, and the container swallowing webpack's output on failure) is worth changing is tracked separately in [NPPM-3044](https://linear.app/a8c/issue/NPPM-3044) rather than attempted here, since neither the failure's routineness nor the exact cause of the silent output is established yet.

1. Read the added bullet under "Isolated environments" in `AGENTS.md` and confirm it accurately describes the diagnosis/recovery steps (docs-only change, no functional testing applies).

### Other information:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -154,6 +154,7 @@ n env cleanup                 # Interactive bulk cleanup
 - Worktrees override individual plugins while the rest are shared. Monorepo worktrees live in `worktrees/<branch>/`; standalone `repos/` worktrees in `worktrees-repos/<name>/<branch>/`, mounted over just that path so other envs keep the base checkout. Destroying a monorepo worktree deletes its branch; a `repos/` worktree keeps it.
 - All env containers share the `newspack_envs` bridge network with their domain as a DNS alias, so they can reach each other (hub/node setups).
 - `n env destroy` removes the container, DB, html dir, hosts entry and worktrees.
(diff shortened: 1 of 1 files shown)


adekbadek:
Closing — on closer inspection this wasn't reproducible.

I tried to reproduce the described failure mode in a fresh disposable env (a clean `newspack-plugin` worktree off `main`): neither a missing workspace-package symlink nor a fully deleted worktree `node_modules` caused `build-repos.sh` to fail. Its own `pnpm install` step self-heals an incomplete worktree install in both cases, so the root cause I documented here doesn't hold up as a general mechanism. Not worth carrying an unconfirmed diagnosis in AGENTS.md.

PR #722, opened 2026-07-24
feat(performance): defer reader-facing JS assets (NPPM-3037)

dkoo:
## What

Adds the native WP `strategy: defer` (WP 6.3+; our floor is 6.9) to **31 audited reader-facing first-party JS handles** across 8 monorepo units, reclaiming render-blocking-JS control from Perfmatters. Perf-only: the enqueue layer changes, the JS does not. Closes NPPM-3037.

Per-unit commits for reviewability; intended to **squash-merge** as `feat(performance): defer reader-facing JS assets (NPPM-3037)` (`perf:` doesn't trigger release jobs).

## Why

The Perfmatters performance study showed render-blocking JS is the dominant native performance gap: without Perfmatters, ~1080 KB of JS blocks first paint, and only ~60 KB defers. The single biggest cause: `newspack_commons` had no strategy, so WP silently downgraded the four RAS scripts that already declare `defer` back to blocking (core downgrades a dependent when its dependency lacks a compatible strategy). This PR defers commons (front-end only) and the rest of the audited set.

## Measured impact (isolated env, gzip wire bytes, deterministic server-side harness)

Render-blocking JS, homepage (article & donate pages within ±4 KB). Re-measured after review feedback, so these numbers reflect the branch as it stands:

|  | Perfmatters ON (Newspack defaults) | Perfmatters OFF (native) |
|---|---|---|
| **Before** (main @ 6ebb8c4) | 231.3 KB / 9 scripts | **1079.5 KB / 23 scripts** |
| **After** (this branch) | **231.3 KB / 9** — byte-parity, no regression | **257.2 KB / 14** |

- **Native render-blocking JS: −822.3 KB (−76%)**; 873 KB now defers without Perfmatters (vs 60 KB before).
- Native defer closes **97% of the gap** to Perfmatters' result. The remaining ~26 KB is the WP-core vendor bundle (out of scope) plus the Campaigns view script, which is deliberately left blocking (see below).
- With Perfmatters active, output is byte-identical — its defer and ours are idempotent.

Total render-blocking payload (JS + CSS), homepage: **1470.8 KB → 647.0 KB** without Perfmatters.

Two review-driven changes moved these numbers in opposite directions, and both are wanted:

- The Campaigns view script went back to blocking (+2.4 KB, +1 script) so above-header prompts keep revealing immediately (NPPM-2934).
- Media kit styles and script no longer load on every page, which is where part of the CSS reduction comes from.

## What changed, per unit

| Unit | Handles | Mechanism |
|---|---|---|
| newspack-plugin | 10 + `newspack_commons` & `newspack-ui` | array-args `strategy`; commons/ui deferred via `wp_script_add_data` **gated `! is_admin()`** (both also load in admin/editor, which stays byte-identical) |
| newspack-blocks | 6 view handles | optional `$strategy` param on shared `enqueue_view_assets()` (incl. the caching-class fan-out); donate/modal callers pass nothing |
| newspack-theme | 4 | bool→array 5th-arg conversion |
| newspack-block-theme | 1 (`newspack-main`) | bool→array |
| newspack-popups | none — see review response below | `view.js` defer reverted; `criteria.js` defer predates this branch |
| newspack-ads | 2 (+ comment on `prebid`, whose tag a `script_loader_tag` filter rebuilds — strategy would be dead code) | array-args; media kit assets moved to the tabs block render path |
| newspack-listings | 1 | array-args |
| republication-tracker-tool | 4 (6 call sites) | array-args, `in_footer` values preserved |

**Deliberately untouched:** all already-`async` handles (donate ×3, modal ×2, reCAPTCHA, content-gate frontend, newsletters subscribe — async is already non-render-blocking; flipping to defer changes ordering semantics for zero gain); WP core handles; the AMP CDN scripts; `class-patches.php` (its legacy async/defer data keys are disjoint from the native `strategy` key — verified no handle carries both).

## Verification

- **Per-unit task reviews** (spec + quality) all clean; **final whole-branch review: READY**, 0 critical/important.
- **Browser QA** (isolated env, anonymous, Perfmatters off — the native path this PR owns): homepage (Homepage Posts hydration, menus, font-loading), single article (content gate renders, **RAS sign-in modal opens** — exercising the newly-effective commons → reader-activation → reader-auth → newspack-ui defer chain), donate page (block interactive, frequency switching; donate/modal still async), my-account (all account scripts defer, sign-in UI works). **0 console errors** on every page.
- **Admin/editor unchanged**: verified via strategy-data inspection under `set_current_screen('dashboard')` → no strategy set in admin contexts.
- PHPCS + `php -l` clean on every changed file.

## Review notes (from the final whole-branch review)

Three minor, non-blocking notes logged for triage:
1. `class-my-account.php` — on the *native* (non-WooCommerce) My Account page, `newspack-ui`'s pre-existing `after`-position inline script made WP core keep it blocking. Since fixed on this branch: the inline moved to `before`.
2. `republication-tracker-tool/includes/class-widget.php` — `in_footer => false` faithfully preserves the original positional arg, intentionally diverging from the unit's other call sites.
3. `newspack-blocks` `wp_script_add_data` path bypasses `wp_register_script`'s strategy-value validation; harmless (docblock constrains callers), noted for awareness.

Also: `newspack-block-theme`'s `newspack-main` was verified via strategy-data inspection rather than a rendered FSE page (env runs newspack-theme); worth a one-page spot check in a block-theme env if one is handy.

## Self-review fixes (post-open, `/code-review` pass)

A high-effort multi-angle self-review (8 finder angles + per-candidate verification against WP core source) produced 4 confirmed findings, all fixed on the branch:

1. **`fix(newspack-ads)`** — `class-media-kit.php` used `strpos( ... ) >= 0` for both content gates; `false >= 0` is `true` in PHP, so **media-kit CSS + JS enqueued on every front-end page sitewide** (pre-existing platform bug surfaced by the review; rolled into this PR because removing sitewide asset load directly serves its goal). Now `false !== strpos( ... )` — verified the homepage no longer loads media-kit assets and a real media-kit page still does.
2. **checkout-button defer extended to content-gate CTAs** — the metering-countdown and content-gifting CTAs render their own markup (the block's `view.php` never runs), so they enqueued `newspack-blocks-checkout-button` without the strategy on exactly the paywall pages this PR targets. Both call sites now pass `'defer'`.
3. **My Account `newspack-ui` defer was dead code** — the same method attaches an inline script in WP's default `'after'` position, and core's `filter_eligible_strategies()` force-downgrades such handles to blocking. The inline script (a DOMContentLoaded-listener registration, safe either side) now passes `'before'`, making the defer effective; verified via a live counterfactual (`'after'` → no defer attr, `'before'` → defer survives).
4. **Dead `is_admin()` guard removed** in the same method — it's hooked only on `wp_enqueue_scripts`, which never fires in admin, so the guard (a copy-paste from the two genuinely mixed-context call sites) was an always-true branch with a misleading comment.

One review candidate was closed as no-change: the deprecated video-playlist block turns out to ship **no view.js at all** (its `enqueue_view_assets` call no-ops on `file_exists`), so nothing is over-enqueued and there is nothing to defer.

Re-measured after fixes: render-blocking JS unchanged (media-kit JS was already deferred by this PR); the media-kit fix additionally removes its CSS + ~10 KB of deferred JS from every non-media-kit page. Perfmatters-ON parity still holds.

## Review response (@miguelpeixe)

Six threads, all addressed. Five are resolved; the defer-as-default suggestion stays open for your take.

- **Campaigns view script (blocking):** reverted in 367c5d676. #449 excludes this script from defer only when above-header prompts are published, and the native strategy here applied unconditionally, so it would have overridden that. The same reveal-path context is available to the enqueue if we want a conditional native strategy later.
- **Media kit:** e5248f922 and a403c8441 add the `is_singular()` early return and move both the script and the stylesheet to `Tabs_Block::render_tab_navigation()`. Moving only the script would have left a block nested in a synced pattern unstyled, since the content sniff also backed the CSS. Sniffing still covers `media-kit-page__wrapper`, which styles the page shell rather than the block.
- **`defer` as the default with per-caller opt-out:** left as opt-in. `newspack-blocks-donate` stays `async`, and the generic `manage_view_scripts()` fallback registers a callback for any future block with a `view.js` and no `view.php` — a default would apply a strategy to scripts before anyone has audited them. Worth revisiting once the remaining view scripts are audited.
- **Shared-handle strategy:** 40bc17ee0 makes the guard first-write-wins, which also skips the repeat work when a block renders more than once per page.
- **`newspack-ui` double-set:** removed in f8c4749f4. The inline script position change stays, since an `after`-position inline makes core drop the strategy entirely.
- **Widget cache-buster:** 3621fe21c versions both widget registrations with `REPUBLICATION_TRACKER_TOOL_VERSION`.

## How to test

1. Isolated env with the branch, Perfmatters **deactivated**, anonymous window.
2. View source on the homepage: first-party `newspack-*` scripts (commons, reader-activation, newspack-ui, popups view/criteria, theme scripts, blocks view scripts) carry `defer data-wp-strategy="defer"`; donate/modal scripts still carry `async`.
3. Click **Sign In** → RAS modal opens. Load an article → prompts/gate render. Donate page → frequency/amount switching works. `/my-account/` → sign-in UI renders. Console: no errors.
4. Reactivate Perfmatters → pages render identically (its defer wraps the same handles).

### DIFF
--- a/plugins/newspack-ads/includes/media-kit/class-media-kit.php
+++ b/plugins/newspack-ads/includes/media-kit/class-media-kit.php
@@ -195,6 +195,13 @@ final class Media_Kit {
 	 * Equeue frontend scripts.
 	 */
 	public static function enqueue_scripts() {
+		// The media kit page (and any page using its patterns) is always a
+		// singular view, so skip the content sniffing everywhere else. The
+		// frontend JS is enqueued from the tabs block render path instead
+		// (see Tabs_Block::render_tab_navigation()).
+		if ( ! is_singular() ) {
+			return;
+		}
 		// Get current page content and check for the special class used in the patterns, or block usage.
 		$page_content = get_the_content();
 		$has_pattern = false !== strpos( $page_content, 'media-kit-page__wrapper' );
@@ -208,18 +215,6 @@ final class Media_Kit {
 			);
 			\wp_style_add_data( 'newspack-ads-media-kit-frontend', 'rtl', 'replace' );
 			\wp_enqueue_style( 'newspack-ads-media-kit-frontend' );
-		}
-		if ( $has_block ) {
-			\wp_enqueue_script(
-				'newspack-ads-media-kit-frontend',
-				Core::plugin_url( 'dist/media-kit-frontend.js' ),
-				[],
-				NEWSPACK_ADS_VERSION,
-				[
-					'in_footer' => true,
-					'strategy'  => 'defer',
-				]
-			);
 		}
 	}
 
--- a/plugins/newspack-ads/includes/media-kit/class-media-kit.php
+++ b/plugins/newspack-ads/includes/media-kit/class-media-kit.php
@@ -207,15 +207,26 @@ final class Media_Kit {
 		$has_pattern = false !== strpos( $page_content, 'media-kit-page__wrapper' );
 		$has_block = false !== strpos( $page_content, '<!-- wp:newspack/tabs' );
 		if ( $has_pattern || $has_block ) {
-			\wp_register_style(
-				'newspack-ads-media-kit-frontend',
-				Core::plugin_url( 'dist/media-kit-frontend.css' ),
-				null,
-				filemtime( dirname( NEWSPACK_ADS_PLUGIN_FILE ) . '/dist/media-kit-frontend.css' )
-			);
-			\wp_style_add_data( 'newspack-ads-media-kit-frontend', 'rtl', 'replace' );
-			\wp_enqueue_style( 'newspack-ads-media-kit-frontend' );
+			self::enqueue_frontend_style();
 		}
+	}
+
+	/**
+	 * Enqueue the media kit frontend stylesheet.
+	 *
+	 * Called both from the content sniffing above (the page wrapper carries styles of
+	 * its own) and from the tabs block render path, which is the only way to catch a
+	 * block nested in a synced pattern.
+	 */
+	public static function enqueue_frontend_style() {
+		\wp_register_style(
+			'newspack-ads-media-kit-frontend',
+			Core::plugin_url( 'dist/media-kit-frontend.css' ),
+			null,
+			filemtime( dirname( NEWSPACK_ADS_PLUGIN_FILE ) . '/dist/media-kit-frontend.css' )
+		);
+		\wp_style_add_data( 'newspack-ads-media-kit-frontend', 'rtl', 'replace' );
+		\wp_enqueue_style( 'newspack-ads-media-kit-frontend' );
 	}
 
 	/**
--- a/plugins/newspack-ads/includes/media-kit/class-media-kit.php
+++ b/plugins/newspack-ads/includes/media-kit/class-media-kit.php
@@ -197,8 +197,8 @@ final class Media_Kit {
 	public static function enqueue_scripts() {
 		// Get current page content and check for the special class used in the patterns, or block usage.
 		$page_content = get_the_content();
-		$has_pattern = strpos( $page_content, 'media-kit-page__wrapper' ) >= 0;
-		$has_block = strpos( $page_content, '<!-- wp:newspack/tabs' ) >= 0;
+		$has_pattern = false !== strpos( $page_content, 'media-kit-page__wrapper' );
+		$has_block = false !== strpos( $page_content, '<!-- wp:newspack/tabs' );
 		if ( $has_pattern || $has_block ) {
 			\wp_register_style(
 				'newspack-ads-media-kit-frontend',
--- a/plugins/newspack-ads/includes/media-kit/class-media-kit.php
+++ b/plugins/newspack-ads/includes/media-kit/class-media-kit.php
@@ -215,7 +215,10 @@ final class Media_Kit {
 				Core::plugin_url( 'dist/media-kit-frontend.js' ),
 				[],
 				NEWSPACK_ADS_VERSION,
-				true
+				[
+					'in_footer' => true,
+					'strategy'  => 'defer',
+				]
 			);
 		}
 	}
--- a/plugins/newspack-blocks/includes/class-newspack-blocks.php
+++ b/plugins/newspack-blocks/includes/class-newspack-blocks.php
@@ -340,9 +340,11 @@ class Newspack_Blocks {
 	/**
 	 * Enqueue view scripts and styles for a single block.
 	 *
-	 * @param string $type The block's type.
+	 * @param string      $type     The block's type.
+	 * @param string|null $strategy Optional. Script loading strategy to apply to the
+	 *                              view script ('defer' or 'async'). Default null (no strategy).
 	 */
-	public static function enqueue_view_assets( $type ) {
+	public static function enqueue_view_assets( $type, $strategy = null ) {
 		$style_path = apply_filters(
 			'newspack_blocks_enqueue_view_assets',
 			NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . $type . '/view.css',
@@ -362,13 +364,17 @@ class Newspack_Blocks {
 		}
 		$script_data = static::script_enqueue_helper( NEWSPACK_BLOCKS__BLOCKS_DIRECTORY . $type . '/view.js' );
 		if ( $script_data ) {
+			$handle = "newspack-blocks-{$type}";
 			wp_enqueue_script(
-				"newspack-blocks-{$type}",
+				$handle,
 				$script_data['script_path'],
 				$script_data['dependencies'],
 				$script_data['version'],
 				true
 			);
+			if ( $strategy ) {
+				wp_script_add_data( $handle, 'strategy', $strategy );
+			}
 		}
 	}
 
--- a/plugins/newspack-blocks/includes/class-newspack-blocks.php
+++ b/plugins/newspack-blocks/includes/class-newspack-blocks.php
@@ -342,7 +342,9 @@ class Newspack_Blocks {
 	 *
 	 * @param string      $type     The block's type.
 	 * @param string|null $strategy Optional. Script loading strategy to apply to the
-	 *                              view script ('defer' or 'async'). Default null (no strategy).
+	 *                              view script ('defer' or 'async'). First write wins:
+	 *                              ignored if a strategy is already set on the handle.
+	 *                              Default null (no strategy).
 	 */
 	public static function enqueue_view_assets( $type, $strategy = null ) {
 		$style_path = apply_filters(
@@ -372,7 +374,7 @@ class Newspack_Blocks {
 				$script_data['version'],
 				true
 			);
-			if ( $strategy ) {
+			if ( $strategy && ! wp_scripts()->get_data( $handle, 'strategy' ) ) {
 				wp_script_add_data( $handle, 'strategy', $strategy );
 			}
 		}
--- a/plugins/newspack-plugin/includes/reader-activation/class-my-account.php
+++ b/plugins/newspack-plugin/includes/reader-activation/class-my-account.php
@@ -176,7 +176,6 @@ class My_Account {
 			NEWSPACK_PLUGIN_VERSION,
 			true
 		);
-		\wp_script_add_data( 'newspack-ui', 'strategy', 'defer' );
 		\wp_add_inline_script(
 			'newspack-ui',
 			"( function() {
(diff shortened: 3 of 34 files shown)


plugins/newspack-popups/includes/class-newspack-popups-inserter.php
```diff
@@ -1058,7 +1058,10 @@ public static function enqueue_scripts() {
 					Newspack_Popups_Criteria::SCRIPT_HANDLE,
 				],
 				filemtime( dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/dist/view.js' ),
-				true
+				[
+					'in_footer' => true,
+					'strategy'  => 'defer',
+				]
```
miguelpeixe: #449 deliberately excludes the defer strategy from Perfmatters for this script. There's more about it in NPPM-2934. We should probably keep it that way.

plugins/newspack-ads/includes/media-kit/class-media-kit.php
```diff
@@ -197,8 +197,8 @@ public static function create_media_kit_page() {
 	public static function enqueue_scripts() {
 		// Get current page content and check for the special class used in the patterns, or block usage.
 		$page_content = get_the_content();
-		$has_pattern = strpos( $page_content, 'media-kit-page__wrapper' ) >= 0;
-		$has_block = strpos( $page_content, '<!-- wp:newspack/tabs' ) >= 0;
+		$has_pattern = false !== strpos( $page_content, 'media-kit-page__wrapper' );
+		$has_block = false !== strpos( $page_content, '<!-- wp:newspack/tabs' );
```
miguelpeixe, line 208: We could also add an `if ( ! is_singular() ) { return; }` here.

Also, this would be better enqueued from the tabs block's render_callback/viewScript rather than via content sniffing. If a publisher selects the tabs section of their working media-kit page and clicks **'Create pattern' (synced)**: `post_content` becomes `'<!-- wp:block {"ref":N} /-->'`, which core never expands at `wp_enqueue_scripts` time, so `get_the_content()` contains neither `'media-kit-page__wrapper'` nor `'<!-- wp:newspack/tabs'`.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -340,9 +340,11 @@ public static function enqueue_block_styles_assets() {
 	/**
 	 * Enqueue view scripts and styles for a single block.
 	 *
-	 * @param string $type The block's type.
+	 * @param string      $type     The block's type.
+	 * @param string|null $strategy Optional. Script loading strategy to apply to the
+	 *                              view script ('defer' or 'async'). Default null (no strategy).
 	 */
-	public static function enqueue_view_assets( $type ) {
+	public static function enqueue_view_assets( $type, $strategy = null ) {
```
miguelpeixe, line 349: Non-blocking: given the number of calls to `'defer'`, it seems that it should be the default value and a per-caller opt-out.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -362,13 +364,17 @@ public static function enqueue_view_assets( $type ) {
...
-				"newspack-blocks-{$type}",
+				$handle,
 				$script_data['script_path'],
 				$script_data['dependencies'],
 				$script_data['version'],
 				true
 			);
+			if ( $strategy ) {
+				wp_script_add_data( $handle, 'strategy', $strategy );
+			}
```
miguelpeixe: The strategy applied via wp_script_add_data is last-write-wins on shared handles (core's add_data is a plain assignment) while the if($strategy) guard means null callers can never clear a set strategy — deterministic today because every strategy-passing caller passes 'defer', but a latent nondeterminism if callers ever diverge.

A future edit makes one caller of a shared handle pass 'async' while another keeps 'defer' (e.g. author-list vs author-profile, which both enqueue 'newspack-blocks-author-profile'): the printed attribute depends on which block renders last in post-content order — content-dependent HTML across posts. Conversely, a caller reverting to null expecting blocking silently keeps whatever another caller set. A per-$type done-guard (also fixing the repeat-work finding) would make it first-write-wins and deterministic.

plugins/newspack-plugin/includes/reader-activation/class-my-account.php
```diff
@@ -176,6 +176,7 @@ public static function enqueue_assets(): void {
 			NEWSPACK_PLUGIN_VERSION,
 			true
 		);
+		\wp_script_add_data( 'newspack-ui', 'strategy', 'defer' );
```
miguelpeixe: This is an always-no-op double-set: `Newspack_UI::enqueue_assets` (ungated, hooked on enqueue_block_assets, fired at wp_enqueue_scripts prio 10 by core) has already applied the identical defer on every request where this prio-11, heavily-gated callback can run — deleting the line changes nothing.

plugins/republication-tracker-tool/includes/class-widget.php
```diff
@@ -55,8 +55,26 @@ public function widget( $args, $instance ) {
...
+			wp_enqueue_script(
+				'republication-tracker-tool-js',
+				plugins_url( 'assets/widget.js', __DIR__ ),
+				array( 'jquery', 'republication-tracker-tool-clipboard-utils' ),
+				filemtime( plugin_dir_path( __FILE__ ) ),
+				array(
+					'in_footer' => false,
+					'strategy'  => 'defer',
+				)
+			);
```
miguelpeixe, line 77: Non-blocking: the `mtime` of the includes/ directory doesn't work well as the cache-buster. This PR versions the same clipboard-utils handle with `REPUBLICATION_TRACKER_TOOL_VERSION` in two sibling files, making the effective version vary by which path registers first.

miguelpeixe (changes requested):

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
dkoo: Reverted in 367c5d676 — `newspack-popups-view` is back to its previous registration. Good catch: our native defer was unconditional, so it would have overridden the conditional exclusion regardless of whether above-header prompts are published.

Worth noting for whenever we revisit this: the same reveal-path context that `above_header_defer_exclusions()` computes is available to the enqueue, so a conditional native strategy is possible later. Out of scope here.

plugins/newspack-ads/includes/media-kit/class-media-kit.php
dkoo: Both done, in e5248f922 and a403c8441:

- `is_singular()` early-return added.
- The tabs script is now enqueued from `Tabs_Block::render_tab_navigation()` instead of content sniffing, so it loads whenever the block actually renders.
- The stylesheet moved to the same render path (`Media_Kit::enqueue_frontend_style()`). The synced-pattern case you described would otherwise have loaded the JS and left the block unstyled, since the sniff also backs the CSS.

Sniffing is still there for `media-kit-page__wrapper`, which styles the page shell rather than the block.

Verified in an isolated env: rendering a tabs block enqueues both assets (script with `defer`), and the homepage loads neither.

plugins/newspack-blocks/includes/class-newspack-blocks.php
dkoo: Leaving the opt-in default. The audit behind this PR judged each script individually — `newspack-blocks-donate` stays `async` and must not flip to `defer`, and the generic `manage_view_scripts()` fallback registers a callback for any future block with a `view.js` and no `view.php`. A `defer` default would apply a strategy to those scripts before anyone has looked at them.

Happy to revisit if the ratio keeps growing — WDYT about a follow-up once the remaining view scripts are audited?

plugins/newspack-blocks/includes/class-newspack-blocks.php
dkoo: Fixed in 40bc17ee0: the guard is now first-write-wins (`$strategy && ! wp_scripts()->get_data( $handle, "strategy" )`), which also skips the repeat work when a block renders more than once per page.

plugins/newspack-plugin/includes/reader-activation/class-my-account.php
dkoo: Removed in f8c4749f4. The `wp_add_inline_script` position change stays — that one is load-bearing regardless of which call sets the strategy, since an `after`-position inline script makes core drop the strategy entirely.

plugins/republication-tracker-tool/includes/class-widget.php
dkoo: Unified in 3621fe21c — both widget registrations now use `REPUBLICATION_TRACKER_TOOL_VERSION`, matching the sibling files.

miguelpeixe (approved):
Thank you for the revisions!

PR #725, opened 2026-07-25
feat(subscribers): tags, newsletters and last-seen columns (NPPD-1753)

adekbadek:

Populates the three hidden columns on the Subscribers list — **Last seen**, **Tags**, and **Newsletters** — from local reader data only, plus an em-dash empty state so an unpopulated cell reads as "no data" rather than a rendering gap.

Part of NPPD-1753 (admin-side Subscribers management).

- **Last seen** reads the `last_active` reader-data item (stamped on every page view, and the same value pushed to the ESP as `Last_Active`) — so the column matches what the publisher sees in their ESP and survives logout. It is client-asserted and informational; it must never gate access.
- **Newsletters** resolves the reader's own `newsletter_subscribed_lists` to list titles against the site's local list registry, which is resolved **once per request** rather than per row.
- **Tags** reads locally-stored `newspack_reader_tags`. No connected-ESP calls anywhere — an ESP lookup per row on a 50-row page is a rate-limit/latency problem out of scope here (see NPPD-2124 for the tag write-path).

**Scope honesty:** the "no per-row queries" criterion holds for *these columns* (a flat +3 regardless of page size, asserted by a query-count test), but the endpoint still issues ~50 WooCommerce order queries per 50-row page from the pre-existing subscription hydration — tracked in NPPD-2126, not closed here. Each row also spends a post query inside `Patches::prevent_accidental_page_deletion()`, which runs on `get_edit_user_link()`'s capability check and looks a user ID up as a post; that guard's early bail is restored in #964, so the query-count test detaches the filter and measures this endpoint's own hydration. `seatRequest` stays `null`: only the badge is wired, there is no data model behind it (NPPD-2125).

Closes NPPD-1753.

1. Open **Audience → Subscribers**, then **View options** and un-hide Last seen, Tags, and Newsletters.
2. For a reader with local tags and newsletter subscriptions, confirm the columns render; for one without, confirm each shows "—".
3. Confirm last-seen matches the reader's `last_active` (visit the site as that reader, reload the list).

### Other information:

Follow-ups: NPPD-2124 (tags writer), NPPD-2125 (seat-request model), NPPD-2126 (per-row hydration).


plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+	 * local_date() also drops a falsy timestamp rather than rendering it as
+	 * 1970-01-01, which matters because the value is client-writable.
+	 *
+	 * @param int $user_id The reader user ID.
+	 *
+	 * @return string|null 'YYYY-MM-DD', or null when the site has no record of them.
+	 */
+	private function last_seen_date( $user_id ) {
+		$last_active = Reader_Data::get_data( $user_id, 'last_active' );
+		if ( empty( $last_active ) || ! is_numeric( $last_active ) ) {
```
jason10lee, line 1044: **Suggestion:** A future last-seen timestamp is accepted and then cannot be corrected — The guard admits any numeric value, including one ahead of now, and the client store merges as `Math.max( server || 0, client || 0 )` (`src/reader-activation/engagement.js:21`). So once a future value lands it can never be lowered by subsequent real activity, and the column shows a date that will not correct itself.

Measured on the review env: a `last_active` about three years ahead rendered as `2029-08-26`, with the relative subtitle reading accordingly.

The likely cause is not an attacker but a device with a wrong system clock writing its own `Date.now()`. Dropping—or clamping to now—a timestamp in the future inside `last_seen_date()` would hold the line on the server side, which is the half that can be trusted.

Scope limits, and they matter here: this is display integrity only—nothing gates access on the value, and the junk-string case is already handled and tested. A future value could also already be stored before this change: the reader-data key is client-writable (for the reader's own row only), and the existing ESP-sync consumer parses it with the same `is_numeric` shape. What this slice adds is a surface that shows it, not a new way to write it.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+				continue;
+			}
+			$list_id = (string) $list_id;
+			// A subscribed list the site has no local Subscription_List for is a
+			// normal state, not an edge case: this data is whatever the ESP returned,
+			// and it can name lists that were never mirrored here. Printing the bare
+			// provider ID would put something like `abc123def` in a publisher-facing
+			// column with nothing to say it is not a newsletter name, so it is labelled
+			// as unresolved while keeping the ID, which is what support needs.
+			$names[] = isset( $titles[ $list_id ] )
```
jason10lee: **Suggestion:** The unresolved-list label is prose, so the list ID never reaches the client — `reader_newsletters()` emits `sprintf( __( 'Unknown list (%s)' ), $list_id )` into the same flat string array as resolved titles, so the provider ID survives only inside a translated sentence and the client cannot tell a synthesized label from a real one.

The fallback itself is the right call—printing a bare `abc123def` in a publisher-facing column is worse, and the comment above it argues that well. What it costs is the ID, and a list ID is what a Newsletters filter would need to match on. Unresolved is a normal steady state rather than a rarity: `newsletter_subscribed_lists` holds whatever `get_contact_combined_lists()` returned, and that is `array_merge()` of the raw ESP IDs from `get_contact_lists()` with the local public IDs from `get_contact_local_lists()`—so on a Mailchimp or ActiveCampaign site the stored set routinely mixes `newspack-42` with audience IDs the site has no local record of.

Returning `[ { id, title } ]`—or a parallel `newsletterIds`—and letting `SubscriberList.jsx` compose the unresolved wording would keep both, and would move that string into the same bundle as the column heading beside it.

Scope limit: the current shape is consistent with how `tags` and `plan` already travel as bare display strings, so this is a response-shape call rather than a defect. It is cheaper to make now than after a filter slice depends on it, but "the polish slice will change the shape anyway" is a fine answer.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+	 * Map of newsletter list public ID → display title, from the site's own
+	 * subscription lists. Memoized for the request.
+	 *
+	 * The site's lists are few and shared by every row, while resolving a list
+	 * on demand costs a query — so the map is built once per request and read
+	 * per row, the same shape the group-membership index uses.
+	 *
+	 * @return array<string,string>
+	 */
+	private function get_newsletter_list_titles() {
```
jason10lee: **Suggestion:** An unavailable list registry is reported as "every list is unknown" — `get_newsletter_list_titles()` returns `[]` for two different conditions—the site genuinely has no lists, and `\Newspack\Newsletters\Subscription_Lists` is not loaded at all. `reader_newsletters()` reads only the empty map, so it cannot tell them apart and labels every stored ID unresolved.

On a site that has deactivated newspack-newsletters, `newsletter_subscribed_lists` is still populated from when it was active, so every reader's cell fills with one `Unknown list (…)` per stored list. Measured on the review env with newspack-newsletters inactive: `reader_newsletters()` returned `["Unknown list (newspack-123)", "Unknown list (group-7)", "Unknown list (abc123def)"]`.

The site has no basis for calling those lists unknown—it just cannot look them up, which is the state the em dash already means everywhere else in this column. Returning `[]` when the registry could not be built, while keeping the labelled fallback for a registry that exists and simply lacks that ID, would separate the two.

The `class_exists` guard itself is right and matches the convention used elsewhere in the plugin; only the empty-result branch needs the distinction.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+	 * dates: on a negative-offset site an evening visit formatted in UTC lands on
+	 * tomorrow's date, which reads as plainly wrong next to them. Going through
+	 * local_date() also drops a falsy timestamp rather than rendering it as
+	 * 1970-01-01, which matters because the value is client-writable.
+	 *
+	 * @param int $user_id The reader user ID.
+	 *
+	 * @return string|null 'YYYY-MM-DD', or null when the site has no record of them.
+	 */
+	private function last_seen_date( $user_id ) {
```
jason10lee: **Nit:** New methods land untyped while this PR's new test helpers are typed — The four new private methods and the new cache property state their types only in docblocks, while the same changeset's test helpers are fully typed—`private function create_reader( string $name = 'Reader', string $role = 'subscriber', string $registered = '' ): int`. That makes it an inconsistency inside one changeset rather than a house-style question.

`int $user_id` plus `: ?string` / `: array` on the new methods, and `?array` on the property, would match the newer code without touching the rest of the file.

Scope limit: the surrounding class is uniformly untyped, so this is not a request to convert the file—only that new members land typed, since retrofitting later is a bigger diff than writing it once. Reasonable to decline if you would rather the file stay internally uniform until it is converted wholesale.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+
+	/**
+	 * The newsletters a reader is subscribed to, as the titles the site shows for
+	 * them.
+	 *
+	 * The subscription itself is read from the reader's own record — the
+	 * `newsletter_subscribed_lists` reader-data item, which the newsletter data
+	 * events keep in step with the ESP — and the list IDs it holds are resolved
+	 * against the site's own list definitions. No ESP call is made; see
+	 * READER_TAGS_META for why. A list ID with no local definition falls back to
```
jason10lee: **Nit:** The docblock contradicts both the code and its own inline comment — "A list ID with no local definition falls back to itself, so a column entry is never silently dropped" describes a bare-ID fallback. The code emits `Unknown list (%s)`, and the inline comment a few lines below argues at length against printing the bare ID.

It reads as a docblock left behind by the revision that inline comment records. The behaviour the code has is the one worth describing.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -965,6 +1001,137 @@ private function last_payment_date( $user_id ) {
...
+			// and it can name lists that were never mirrored here. Printing the bare
+			// provider ID would put something like `abc123def` in a publisher-facing
+			// column with nothing to say it is not a newsletter name, so it is labelled
+			// as unresolved while keeping the ID, which is what support needs.
+			$names[] = isset( $titles[ $list_id ] )
+				? $titles[ $list_id ]
+				/* translators: %s: the email service provider's identifier for a list the site has no local record of. */
+				: sprintf( __( 'Unknown list (%s)', 'newspack-plugin' ), $list_id );
+		}
+		return array_values( array_unique( array_filter( $names ) ) );
```
jason10lee: **Nit:** The callback-less array_filter drops "0", and dedupe runs on the resolved title — `array_filter( $names )` with no callback drops any falsy string, so a list titled `0` vanishes from the column—the same "silently dropped" the docblock above rules out. The identical call at `:1062` does it to tags: measured on the review env, seeding a reader with tags `['vip', '0']` returned `["vip"]`. Filtering on `'' !== $v` keeps both.

Separately, `array_unique( $names )` dedupes by resolved *title*, so two distinct lists sharing a title collapse into one entry. Deduping the list IDs before resolution would preserve them; we did not exercise that half, since it needs a live list registry.

Both are edge cases with no known reporter—worth a line each, not worth a round trip on their own.

plugins/newspack-plugin/tests/mocks/newsletters-namespaced-mocks.php
```diff
@@ -18,6 +18,11 @@ public function get_public_id(): string {
 			return 'list-' . $this->id;
```
jason10lee: **Nit:** The stub's public-ID format and CPT do not match the class it stands in for — `get_public_id()` returns `'list-' . $id` and the stub sets `const CPT = 'np_newsletter_list'`, where the real class uses `PUBLIC_ID_PREFIX = 'newspack-'` and `CPT = 'newspack_nl_list'`. The comment below says the stub "mirrors the real implementation's shape", and the public ID *is* the shape here—it is the join key `reader_newsletters()` matches stored subscription IDs against.

This hides nothing today: the wizard's lookup is a plain array-key comparison and the test seeds both sides from the stub, so the format is internally consistent either way. It matters to the next reader, who may be debugging why a real site's `newsletter_subscribed_lists` does not resolve and will read this fixture as evidence of what a stored ID looks like.

The `get_title()` divergence is the same category—the stub goes through `get_the_title()`, which texturizes titles the production path does not.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
```diff
@@ -676,6 +741,156 @@ public function test_status_filter_scan_advances_past_the_first_chunk() {
...
+
+		$newsletter_lists_query_count = 0;
+		$full_page                    = $this->dispatch( [ 'per_page' => 50 ] )->get_data();
+
+		$this->assertCount( 50, $full_page['items'], 'The measured page really is 50 rows.' );
+		$this->assertSame( [ 'Daily Brief' ], $full_page['items'][0]['newsletters'], 'The page really is hydrating the batched columns.' );
+		$this->assertSame(
+			$single_row_cost,
+			$full_page_cost,
+			'Hydrating 50 rows must not cost more queries than hydrating 1 — a per-row lookup crept in.'
```
jason10lee: **Nit:** The assertion message claims more than the fixture can measure — "Hydrating 50 rows must not cost more queries than hydrating 1" reads as a property of the endpoint, but `wcs_get_subscriptions()` is an in-memory mock over a global array (`tests/mocks/wc-mocks.php:1371`), so the subscription hydration contributes zero queries to either side of the comparison.

The guard is genuine and valuable for the three columns this slice adds, and we are not reopening the known per-row subscription cost. The risk is a future reader taking the message at face value and believing the endpoint is flat overall. Narrowing it to the columns being measured costs one clause.

jason10lee (approved):
Approved with three suggestions and five nits.

The parts that are easy to get wrong are right here. The endpoint is gated at the level this data deserves rather than a looser one, all three new readers degrade to the empty state instead of warning on a missing or odd value, and the guard for an inactive newsletters plugin is in place. The performance claims hold: the list registry really is built once per request, and the test guarding it would catch a per-row rebuild that a plain query count could not, because identical post queries are served from the object cache. All four new tests fail if the slice is reverted.

Of the three suggestions, two are about a value the column reports with more confidence than the site has—an unavailable list registry read as "these lists are unknown", and a last-seen date that can be set ahead of now and then never corrects itself. The third is a response-shape call worth making before a filter slice depends on it. Each is on its own thread.

Review provenance: 3 of 3 reviewer passes (lineup: deep, wp-expert, codex); release risk: ELEVATED; stacked on #964 (delta only).

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in cdbf714. `last_seen_date()` now compares against the site's clock: a timestamp more than a day ahead is dropped, anything within that clamps to now.

Split by magnitude because both halves of your menu are right at different scales. A device a minute fast writes a future value on every visit, so dropping outright would permanently blank the column for a genuinely active reader. A clock years out is describing itself, not the reader, and clamping it would report a dormant reader as seen today forever — the direction that misleads a publisher scanning for lapsed readers.

New test `test_last_seen_does_not_trust_a_clock_ahead_of_the_site` pins both.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in cdbf714. `get_newsletter_list_titles()` now returns `null` when there is no registry to read and an array — possibly empty — when there is one; `reader_newsletters()` returns nothing on `null`, so a site with newspack-newsletters deactivated shows the em dash. The labelled fallback stays for a registry that exists and lacks the ID.

The memo needed a companion `newsletter_list_titles_resolved` flag, since `null` is now a resolved state rather than an unfilled one.

Not covered by a test: the stub registry is defined for the whole run, so the class-absent branch cannot be reached from inside the suite.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Done in cdbf714 — `int $user_id` and `: ?string` / `: array` on the four new methods, `?array` on the memo, `bool` on the new resolved flag. Agreed that new members are cheaper to type now than to retrofit; the rest of the file is untouched.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in cdbf714 — the docblock now describes the labelled fallback the code emits, and notes the case where there is no registry to resolve against at all. Same edit reached the test docblock above it, which repeated the bare-ID claim.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Both fixed in cdbf714.

Emptiness rather than truthiness on both filters: tags go through `array_diff( $tags, [ '' ] )`, and `reader_newsletters()` skips empty IDs and no longer filters the resolved names at all.

Dedupe now runs on the list IDs before resolution, so two lists sharing a title stay two entries. `test_newsletters_dedupe_by_list_id_not_by_resolved_title` covers that alongside a list titled `0`, and the tags test now carries a `0` tag.

plugins/newspack-plugin/tests/mocks/newsletters-namespaced-mocks.php
adekbadek: Fixed in cdbf714. The stub carries the real `PUBLIC_ID_PREFIX = 'newspack-'` and builds its public ID from it, `CPT` is `newspack_nl_list`, and `get_title()` returns the raw `post_title` rather than routing through `get_the_title()`. A comment records that a remote-only list carries the ESP's ID instead, which the stub does not model.

The prefix change rippled into the ten `'list-' . $list_post_id` expectations in the premium-newsletters tests; those now read `'newspack-'`. Full suite green.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscribers-endpoint.php
adekbadek: Reworded in cdbf714 — the message now scopes the claim to the columns being measured and says outright that subscription hydration is an in-memory mock here and contributes nothing to either side.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Made the change. `newsletters` is now `[ { id, title } ]`, with a null title for a list the site holds no definition for, and `SubscriberList.jsx` composes the "Unknown list (%s)" wording next to the column heading it sits under. The ID survives resolution either way, so a filter has something to match on.

jason10lee (approved):
Approving. Everything from the last round is addressed, and there are no further suggestions or nits.

The last-seen guard improved on our suggestion. It splits by how far ahead the clock is, rather than treating every future value the same. That keeps the column populated for a reader whose device runs a minute fast, instead of blanking it. The list screen also picked up its own test coverage, which we had raised as an open question rather than a finding.

The riskiest part of the change holds up. Teaching the list registry to report "there is no registry" separately from "the registry is empty" collides with the value the cache already used for "not looked up yet". The cache carries a second piece of state to keep those apart. It is set before the lookup, so an absent registry is remembered rather than re-checked on every row.

Scope is tighter than when we reviewed it. The page-deletion guard has moved to the PR it belongs to, and what remains is two production files and their tests. CI is green on the current head.

PR #726, opened 2026-07-25
feat(subscribers): person profile with in-wizard navigation (NPPD-1753)

adekbadek:

Adds the **L1 person profile** to the Subscribers wizard: clicking a subscriber now opens an in-wizard profile showing every subscription they hold — individual plans and group memberships — as cards with per-status badges, instead of bouncing to the native `user-edit.php` screen.

Part of NPPD-1753 (admin-side Subscribers management). **This is the keystone slice**: it authors two foundations the group-detail PR consumes — the shared `SubscriptionCard` component and the detail-route wiring. Stacked on #631; base retargets to `main` once #631 merges.

- New `GET /subscribers/<id>` read endpoint (`manage_options`-gated like its siblings). The single-entity payload **embeds full group objects** (not ids the client resolves), matching the shipped `/groups` collection convention and avoiding an N+1 to render a card.
- New `screens/PersonProfile.jsx` and the shared `components/SubscriptionCard.jsx` (exports `CardRow` for the group-detail drawer). A group renders as a card with a Group badge, a seats subline, and a title linking to group detail.
- Detail routing via `isHidden: true` sections; the Subscribers tab claims `/subscribers/:id` through `activeTabPaths`, so the profile renders inside that tab's panel and the tab stays selected — rather than the wizard's no-tab-owns-this fallback.
- `subscription_billing()`/`subscription_date()` produce the shared billing shape both screens render, and the closing card row is derived from the **dates, not the status** (`scheduleRow`), so a `pending-cancel` subscription reads "Ends <date>" rather than a meaningless "Next billing —".
- **Every date the wizard emits shares one timezone basis.** All of them go through `local_date()` (`wp_date()`), so they match WooCommerce's own admin screens and no two rows of the same card can read a calendar day apart. The client's "Ends"/"Ended" label reads "today" on that same basis (`dateI18n`), not from the admin's browser clock.

**Security:** the `?from=` back-nav param is validated to internal hash paths only (`^#/`, rejecting `#//` and `javascript:` URLs) before it reaches any `href`.

Money actions (change plan, refund, reactivate) are later slices and ship absent from the card menu. Native edit links remain until each is replaced, so no user is stranded.

**Deviation from the plan snippet, deliberate:** the detail section does **not** set `fullWidth: true`. The prototype reserves full width for the L0 lists; detail screens render in the standard content column. The group-detail PR matches this.

**Unrelated to the feature — CI fix.** VIPCS 3.1.0 shipped on 2026-07-27 and floats in through the root `^3.0` constraint, turning every PHPCS job in the repo red. Its new `NoPaging` check for `posts_per_page`/`numberposts` set to `-1` lands on ~66 pre-existing call sites across four plugins (WP-CLI migrations, test fixtures, queries over small internal post types), so `phpcs.xml` defers that one check with a comment explaining why; the `nopaging` group stays enabled, and adopting the deferred check is its own piece of work. Its two sharpened warning sniffs hit four real call sites, suppressed inline with reasons — three `file_get_contents()` reads of plugin-bundled assets, and one `pre_get_posts` redirect that already carried a suppression for the sniff's older error codes.

Closes NPPD-1753.

1. On a site with WooCommerce Subscriptions (including a group with an owner/manager/members, and a reader holding both an active and a cancelled plan), open **Audience → Subscribers**.
2. Click a subscriber; confirm the in-wizard profile lists all their subscriptions with correct badges, the Subscribers tab stays selected, and a group renders as a card linking to group detail.
3. Load `#/subscribers/29?from=javascript:alert(1)`; confirm no script runs and the back link falls back to the list.
4. Confirm a `pending-cancel` subscription shows "Ends <date>", not "Next billing —".
5. Set the site to a negative-offset timezone (Settings → General) and confirm "First subscribed", "Last payment" and "Subscriber since" all show the same calendar day WooCommerce's own order/subscription screens show.

### Other information:

Follow-ups: NPPD-2127 (currency formatting alignment).

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
+++ b/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
@@ -384,7 +384,7 @@ class Subscribers_Wizard extends Wizard {
 		$owner_id   = (int) $subscription->get_user_id();
 		$owner      = $owner_id ? get_userdata( $owner_id ) : false;
 		$created    = $subscription->get_date_created();
-		$created_at = $created ? gmdate( 'Y-m-d', $created->getTimestamp() ) : null;
+		$created_at = $this->local_date( $created ? $created->getTimestamp() : null );
 
 		return [
 			'id'          => (int) $subscription->get_id(),
@@ -589,9 +589,10 @@ class Subscribers_Wizard extends Wizard {
 		$subscriptions       = $this->get_individual_subscriptions( $user_id, $owned_subscriptions, $detailed );
 		$groups              = $this->get_group_memberships( $user_id, $detailed );
 
-		// user_registered can be a zero date ('0000-00-00 …'), which is truthy but
+		// user_registered is stored in GMT, so anchor the parse to UTC before it is
+		// localized. It can also be a zero date ('0000-00-00 …'), which is truthy but
 		// unparseable; guard on the parsed timestamp so it degrades to null, not 1970.
-		$registered = $user->user_registered ? strtotime( $user->user_registered ) : false;
+		$registered = $user->user_registered ? strtotime( $user->user_registered . ' UTC' ) : false;
 
 		return [
 			'id'            => $user_id,
@@ -602,7 +603,7 @@ class Subscribers_Wizard extends Wizard {
 			// profile keeps this as a header action rather than stranding the admin.
 			'editUrl'       => get_edit_user_link( $user_id ),
 			'status'        => $this->reduced_status( $subscriptions, $groups ),
-			'memberSince'   => $registered ? gmdate( 'Y-m-d', $registered ) : null,
+			'memberSince'   => $this->local_date( $registered ),
 			'lastPayment'   => $this->last_payment_date( $user_id ),
 			// Wired to reader activity in a later slice; the column is hidden by default.
 			'lastSeen'      => null,
@@ -690,15 +691,32 @@ class Subscribers_Wizard extends Wizard {
 	}
 
 	/**
+	 * The calendar day an instant falls on in the publisher's timezone, as a bare
+	 * 'Y-m-d' string.
+	 *
+	 * EVERY date this wizard emits goes through here, so one profile cannot mix
+	 * bases: a UTC-formatted "First subscribed" sitting directly above a localized
+	 * "Last payment" can read a day apart for the same instant on a negative-offset
+	 * site, and the publisher cross-checks these numbers against WooCommerce's own
+	 * admin screens, which localize. wp_date() formats a Unix timestamp in the site
+	 * timezone; callers are responsible for handing over a UTC-anchored timestamp.
+	 *
+	 * @param int|false|null $timestamp A Unix timestamp, or a falsy value when the date is unset.
+	 *
+	 * @return string|null 'YYYY-MM-DD', or null when there is no timestamp.
+	 */
+	private function local_date( $timestamp ) {
+		return $timestamp ? wp_date( 'Y-m-d', (int) $timestamp ) : null;
+	}
+
+	/**
 	 * One of a subscription's dates as a bare 'Y-m-d' string in the site's
 	 * timezone, or null when unset.
 	 *
 	 * WC_Subscription::get_date() hands back a GMT 'Y-m-d H:i:s' string, or the
 	 * integer 0 when the date is not set. The wizard shows calendar days, so the
 	 * GMT instant is converted to the publisher's timezone before the time is
-	 * dropped — otherwise a date near midnight renders a day off from WooCommerce's
-	 * own admin screens (which localize), exactly the numbers a publisher
-	 * cross-checks. wp_date() formats a Unix timestamp in the site timezone.
+	 * dropped — see local_date().
 	 *
 	 * @param \WC_Subscription $subscription The subscription.
 	 * @param string           $date_type    A WCS date type, e.g. 'next_payment'.
@@ -712,8 +730,7 @@ class Subscribers_Wizard extends Wizard {
 		}
 		// get_date() returns GMT; anchor the parse to UTC so a server default
 		// timezone can't shift the instant before it is localized.
-		$timestamp = strtotime( (string) $date . ' UTC' );
-		return $timestamp ? wp_date( 'Y-m-d', $timestamp ) : null;
+		return $this->local_date( strtotime( (string) $date . ' UTC' ) );
 	}
 
 	/**
@@ -815,7 +832,7 @@ class Subscribers_Wizard extends Wizard {
 					'role'     => $role,
 					// Null for a member who predates the joined-at meta, rather than a
 					// misleading Unix epoch.
-					'joinedAt' => $joined_at ? wp_date( 'Y-m-d', $joined_at ) : null,
+					'joinedAt' => $this->local_date( $joined_at ),
 				]
 			);
 		}
@@ -945,7 +962,7 @@ class Subscribers_Wizard extends Wizard {
 			return null;
 		}
 		$paid = $orders[0]->get_date_paid();
-		return $paid ? gmdate( 'Y-m-d', $paid->getTimestamp() ) : null;
+		return $this->local_date( $paid ? $paid->getTimestamp() : null );
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
+++ b/plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
@@ -49,6 +49,21 @@ class Subscribers_Wizard extends Wizard {
 	const FILTER_INCLUDE_CAP = 10000;
 
 	/**
+	 * Upper bound on the number of avatars one /avatars request resolves. A single
+	 * list page needs at most `per_page` (≤100); the client batches larger sets.
+	 */
+	const AVATAR_BATCH_CAP = 200;
+
+	/**
+	 * How many subscriptions the filter scan hydrates at a time. The scan itself
+	 * is bounded by FILTER_INCLUDE_CAP, but WooCommerce returns fully hydrated
+	 * WC_Subscription objects and only their customer ID is needed here — walking
+	 * the set a page at a time keeps peak memory at one chunk rather than the
+	 * whole cap. See customer_ids_for_raw_statuses().
+	 */
+	const FILTER_SCAN_CHUNK = 500;
+
+	/**
 	 * Per-request memo of the site's group subscriptions (each with its resolved
 	 * settings). A single request can resolve these several times — once per active
 	 * filter set plus the group list itself — and each resolution hydrates every
@@ -66,6 +81,14 @@ class Subscribers_Wizard extends Wizard {
 	 * @var array<int,array<int,array>>|null
 	 */
 	private $group_membership_index = null;
+
+	/**
+	 * Per-request memo of customer IDs keyed by the sorted status set they were
+	 * resolved for. See customer_ids_for_raw_statuses().
+	 *
+	 * @var array<string,int[]>
+	 */
+	private $raw_status_ids_cache = [];
 
 	/**
 	 * Constructor.
@@ -133,9 +156,14 @@ class Subscribers_Wizard extends Wizard {
 				'permission_callback' => [ $this, 'api_permissions_check' ],
 				'args'                => [
 					'emails' => [
-						'type'     => 'array',
-						'required' => true,
-						'items'    => [ 'type' => 'string' ],
+						'type'              => 'array',
+						'required'          => true,
+						'items'             => [ 'type' => 'string' ],
+						// Sanitized declaratively, like the sibling `size` arg. Invalid
+						// entries are dropped rather than failing the whole request:
+						// one malformed address shouldn't blank a page of avatars.
+						'sanitize_callback' => [ __CLASS__, 'sanitize_emails_arg' ],
+						'validate_callback' => 'rest_validate_request_arg',
 					],
 					'size'   => [
 						'type'              => 'integer',
@@ -211,10 +239,27 @@ class Subscribers_Wizard extends Wizard {
 						'type'              => 'array',
 						'items'             => [ 'type' => 'string' ],
 						'sanitize_callback' => 'rest_sanitize_request_arg',
+						'validate_callback' => 'rest_validate_request_arg',
 					],
 				],
 			]
 		);
+	}
+
+	/**
+	 * Sanitize the /avatars `emails` arg: valid addresses only, deduplicated and
+	 * bounded to AVATAR_BATCH_CAP.
+	 *
+	 * Invalid entries are dropped rather than rejected so a single malformed
+	 * address can't fail the whole batch.
+	 *
+	 * @param mixed $emails The raw `emails` request arg.
+	 *
+	 * @return string[] Sanitized, unique, capped email addresses.
+	 */
+	public static function sanitize_emails_arg( $emails ) {
+		$sanitized = array_filter( array_map( 'sanitize_email', (array) $emails ) );
+		return array_values( array_slice( array_unique( $sanitized ), 0, self::AVATAR_BATCH_CAP ) );
 	}
 
 	/**
@@ -257,6 +302,7 @@ class Subscribers_Wizard extends Wizard {
 	public function api_get_groups() {
 		$this->group_subscriptions_cache = null;
 		$this->group_membership_index    = null;
+		$this->raw_status_ids_cache      = [];
 		$items                           = [];
 		foreach ( $this->get_group_subscriptions() as $group ) {
 			$items[] = $this->prepare_group( $group['subscription'], $group['settings'] );
@@ -363,9 +409,10 @@ class Subscribers_Wizard extends Wizard {
 	public function api_get_subscribers( $request ) {
 		$this->group_subscriptions_cache = null;
 		$this->group_membership_index    = null;
+		$this->raw_status_ids_cache      = [];
 		$per_page                        = (int) $request->get_param( 'per_page' );
-		$page     = (int) $request->get_param( 'page' );
-		$search   = trim( (string) $request->get_param( 'search' ) );
+		$page                            = (int) $request->get_param( 'page' );
+		$search                          = trim( (string) $request->get_param( 'search' ) );
 
 		$query_args = [
 			'role__in'    => Reader_Activation::get_reader_roles(),
@@ -418,6 +465,19 @@ class Subscribers_Wizard extends Wizard {
 
 	/**
 	 * Shape a reader user for the admin subscriber list.
+	 *
+	 * Cost note: group memberships are indexed once per request
+	 * (get_group_membership_index), but two lookups here are inherently per-row —
+	 * wcs_get_users_subscriptions() and last_payment_date()'s wc_get_orders() —
+	 * so a page costs ~2 × `per_page` (≤100) WooCommerce queries. Both are
+	 * "newest/owned rows for one customer" lookups that WooCommerce can't answer
+	 * per-customer in a single batched query: batching last_payment_date() over
+	 * the page's customer IDs would need either an unbounded `limit => -1` (worse
+	 * peak memory than the bounded per-row queries) or a truncated scan that
+	 * silently blanks the last payment of a reader whose most recent order falls
+	 * outside it. The bounded per-row shape is the correct trade-off here, and
+	 * matches the existing convention elsewhere in the plugin
+	 * (Sync\Contact_Metadata\Engagement::get_latest_order).
 	 *
 	 * @param \WP_User $user The reader user.
 	 *
@@ -567,6 +627,22 @@ class Subscribers_Wizard extends Wizard {
 	 * dropped whenever any live status remains. Empty when they have no
 	 * subscription at all (a free reader shows no status badge).
 	 *
+	 * SOURCE OF TRUTH for the status-reduction rule. The rule is:
+	 *
+	 *   Rank statuses active → pending → on-hold → cancelled, and drop
+	 *   `cancelled` entirely whenever the reader still holds any live
+	 *   (non-cancelled) subscription; a reader with no subscription at all has
+	 *   no status.
+	 *
+	 * It is re-encoded in three other places, each of which points back here and
+	 * must be changed in step or the endpoint's filter will contradict the badge
+	 * it filters on:
+	 *   - customer_ids_for_statuses() below — the endpoint's `status` filter.
+	 *   - displayStatuses() in src/wizards/subscribers/status.js — the row's
+	 *     status badges.
+	 *   - visiblePlanEntries in src/wizards/subscribers/screens/SubscriberList.jsx
+	 *     — the Subscription column.
+	 *
 	 * @param array $subscriptions Individual subscription entries.
 	 * @param array $groups        Group membership entries.
 	 *
@@ -667,8 +743,9 @@ class Subscribers_Wizard extends Wizard {
 
 	/**
 	 * Customer IDs whose displayed status matches any of the given prototype
-	 * statuses, mirroring the list's badge reduction (see reduced_status /
-	 * displayStatuses): a live status always qualifies, but `cancelled` matches
+	 * statuses, mirroring the list's badge reduction (reduced_status() above is
+	 * the documented source of truth for that rule): a live status always
+	 * qualifies, but `cancelled` matches
 	 * only fully-churned readers — anyone who also holds a live (active/pending/
 	 * on-hold) subscription is dropped, since the badge hides cancelled while a
 	 * live plan remains. Without this the Cancelled filter would surface readers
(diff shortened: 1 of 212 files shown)


plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -556,16 +638,84 @@ private function get_individual_subscriptions( $user_id, $owned_subscriptions )
...
+			'nextBillingDate' => $this->subscription_date( $subscription, 'next_payment' ),
+			// When a subscription is ending — cancelled outright, or in its prepaid
+			// term after a pending-cancel — the card shows this in place of the
+			// next-billing row. `end` is the access-end date the reader actually
+			// keeps until; `cancelled` (the moment cancellation was *requested*) is
+			// only the fallback for the rare cancelled subscription with no end date
+			// recorded, so the row never blanks. WCS deletes next_payment on
+			// pending-cancel, so a subscription with an end date and no next payment
+			// is ending even while its status still maps to active (still entitled).
+			'endDate'         => $this->subscription_date( $subscription, 'end' ) ?? $this->subscription_date( $subscription, 'cancelled' ),
```
dkoo, line 688: **Suggestion:** Mixed date-timezone bases on one screen — `subscription_date()` deliberately localizes WCS dates via `wp_date()` — its docblock explains that a UTC-formatted date renders a day off from WooCommerce's own admin screens. But three sibling fields skip that path and use `gmdate()` (UTC): `prepare_group()`'s `createdAt` (~line 387), `memberSince` (~line 605), and `last_payment_date()` (~line 948). The inconsistency is visible on a single profile: a group-owner card renders "First subscribed" from `createdAt` (UTC day) directly above "Last payment" (site-local day), so the same boundary instant can read a day apart between adjacent rows for a negative-offset publisher — exactly the off-by-one the docblock warns about, and the worst case is an admin cross-checking against Woo's order screens and acting on the wrong date. Routing the three through the same site-tz localization (e.g. `wp_date( 'Y-m-d', $ts )`) gives the whole screen one basis — worth doing before the stacked group-detail PR consumes this shape.

plugins/newspack-plugin/src/wizards/subscribers/format.js
```diff
@@ -23,3 +24,149 @@ export const fmtDate = value => {
...
+ */
+export const scheduleRow = ( { nextBillingDate, endDate } = {} ) => {
+	if ( nextBillingDate ) {
+		return { label: __( 'Next billing', 'newspack-plugin' ), value: fmtDate( nextBillingDate ) };
+	}
+	if ( endDate ) {
+		// Compared as bare YYYY-MM-DD strings, which sort chronologically; today
+		// is the site's civil date (the same basis fmtDate presents), so the label
+		// flips on the calendar day the plan ends.
+		const today = fmtRawDate( new Date() );
```
dkoo: **Suggestion:** "Ends/Ended" compares against the viewer's calendar day, not the site's — `scheduleRow()` compares the endpoint's site-timezone `endDate` against `fmtRawDate( new Date() )` — the viewer's browser-local civil date — while the inline comment claims "today is the site's civil date." For an admin whose browser timezone differs from the site's, the Ends/Ended label flips up to a day early or late on the boundary day (the date value itself stays correct). `@wordpress/date` exposes the site timezone via `getSettings()`; deriving "today" with `dateI18n( 'Y-m-d' )` makes the comparison basis match both the comment and the server's. At minimum, correct the comment.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscriber-detail-endpoint.php
```diff
@@ -0,0 +1,610 @@
...
+ * Tests for the Subscribers wizard single-subscriber read endpoint.
+ *
+ * The person profile (L1) reads one reader in full: every subscription they
+ * hold, individual and group alike, each with the billing detail its card
+ * renders. These tests pin the parts that are not obvious from the code —
+ * the permission boundary, the mixed-status case the profile exists to show,
+ * and the decision that a group arrives as a whole object rather than an ID
+ * the client would have to resolve.
+ *
+ * @package Newspack\Tests
```
dkoo, line 12: **Suggestion:** @group tags are dead — file docblock, not class docblock — The `@group WooCommerce_Subscriptions_Integration` / `@group subscribers-wizard` tags sit in the file docblock, separated from the class by `use` statements — PHPUnit reads groups only from class/method doc comments, so `--group subscribers-wizard` silently runs zero tests (confirmed empirically; both sibling endpoint test files share the pattern from the earlier slice). CI is unaffected (no group filtering), so this is about developer workflow: move the tags into the class docblock here, and fix the two sibling files in a small follow-up rather than perpetuating the dead tags.

plugins/newspack-plugin/src/wizards/subscribers/components/SubscriptionCard.jsx
```diff
@@ -0,0 +1,246 @@
...
+ * Exported because the "View subscription" drawer lays the same label/value
+ * pairs out vertically rather than in the card's two-column grid.
+ *
+ * @param {Object} props          Component props.
+ * @param {string} props.label    The row label.
+ * @param {*}      props.children The row value.
+ */
+export const CardRow = ( { label, children } ) => (
+	<div className="newspack-subscribers__card-row">
+		<span className="newspack-subscribers__card-label">{ label }</span>
```
dkoo: **Nit:** CardRow label/value pairs unassociated for screen readers — Label/value pairs render as unassociated spans in a div; a `dl`/`dt`/`dd` structure would give screen readers the pairing for free. Acceptable as-is for read-only display — the heading hierarchy and the kebab menu's aria attributes are otherwise solid — but the component is shared with the stacked group-detail PR, so the earlier it's semantic, the cheaper.

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscriber.js
```diff
@@ -0,0 +1,84 @@
...
+ * id the route regex never matches (`rest_no_route`) — surfaces with `notFound`
+ * instead, so the screen states the dead end rather than offering a Retry that
+ * can never succeed.
+ *
+ * @param {number|string} id The subscriber (user) ID.
+ * @return {{ subscriber: ?Object, loading: boolean, error: string, notFound: boolean, reload: Function }} The profile.
+ */
+// A request that can never succeed on retry: the user is gone, or the id is not
+// a number the `(?P<id>\d+)` route will ever match.
+const NOT_FOUND_CODES = [ 'newspack_subscriber_not_found', 'rest_no_route' ];
```
dkoo: **Nit:** rest_no_route reads as "subscriber deleted" — Treating `rest_no_route` as notFound is right for the non-numeric-id case, but the same code fires when the route itself is absent (feature flag off, plugin deactivated mid-session), where "They may have been deleted" is misleading. Fine to leave; revisit if support reports confusion. (Two micro-nits nearby: the hook's JSDoc is separated from the function by the `NOT_FOUND_CODES` const, and `PersonProfile.jsx` carries a file-wide translator-comments eslint-disable that will silence future strings — per-line disables would keep coverage.)

plugins/newspack-plugin/src/wizards/subscribers/use-portals.js
```diff
@@ -0,0 +1,54 @@
...
+	const [ node, setNode ] = useState( null );
+	useEffect( () => {
+		if ( ! enabled ) {
+			return undefined;
+		}
+		let observer;
+		const attach = () => {
+			const found = document.querySelector( selector );
+			if ( found ) {
+				if ( modifierClass ) {
```
dkoo, line 40: **Nit:** Portal modifier class never removed on cleanup — `useWizardNode()` adds `modifierClass` to the matched wizard header node, but the effect cleanup only disconnects the observer — navigating back to a list leaves `newspack-subscribers__has-avatar` on the shared header. Harmless today (the class only sets `position: relative` and React removes the portaled avatar), but a `classList.remove` in the cleanup makes the helper safe by construction if the modifier ever gains layout weight.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscriber-detail-endpoint.php
```diff
@@ -0,0 +1,610 @@
...
+
+	/**
+	 * FINDING 5: dates render in the publisher's timezone, matching WooCommerce's
+	 * own admin screens. A GMT instant just after midnight belongs to the previous
+	 * calendar day for a negative-offset site, so the endpoint must localize before
+	 * dropping the time — otherwise the profile shows a date a day off from every
+	 * other Woo screen the publisher cross-checks.
+	 */
+	public function test_dates_render_in_the_site_timezone() {
+		update_option( 'timezone_string', '' );
```
dkoo: **Nit:** Timezone options mutated without an explicit restore (hygiene only) — The site-timezone test sets `timezone_string`/`gmt_offset` via `update_option` with no teardown restore. In practice the leak doesn't occur — `WP_UnitTestCase` rolls back the per-test DB transaction and the next test's `set_up` flushes the object cache, so later tests see the fixture defaults — but an explicit save-and-restore (or teardown) documents the intent and keeps the test correct if it's ever ported outside the WP test-case transaction wrapper.

dkoo (commented):
Release risk: LOW — zero blockers, a clean consequences pass, no contract-surface impact, and the browser walkthrough is now complete: the profile renders in-wizard with the Subscribers tab correctly selected, active and cancelled subscriptions carry the right badges, the group card shows its Group badge, seats subline, and linked title, a pending-cancel subscription shows 'Ends August 17, 2026' rather than a next-billing date, and a crafted ?from=javascript:alert(1) neither executes nor renders — the back chevron and breadcrumb both fall back to the list. The slice is unusually well-built: the back-nav validator is a real security boundary treated as one (fragment-only hrefs with an honest test matrix, now also confirmed live), multisite reads are bounded to the current site, the embedded group objects expose only the owner's contact plus a member count, and the lean-vs-detailed payload boundary is pinned from both sides by tests. Two things worth attention before merge, both date-related: three fields (group createdAt, memberSince, last payment) format in UTC while their sibling subscription dates localize to the site timezone — adjacent rows on the same card can disagree by a calendar day for non-UTC publishers, and this shape feeds the stacked group-detail PR; and the "Ends/Ended" label computes "today" from the viewer's browser clock rather than the site's, flipping a day early or late for admins in a different timezone. Both are presentational and bounded — the worst case is an admin misreading a date next to WooCommerce's own screens — but they contradict the PR's own localization rationale and are cheap to align now before the shared shape propagates.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Good catch — routed all three through the site timezone in 4326915. Rather than sprinkling `wp_date()` at each site, every date the wizard emits now goes through one `local_date( $timestamp )` helper: `prepare_group()`'s `createdAt`, `memberSince`, `last_payment_date()`, `joinedAt` and `subscription_date()` alike. `memberSince` also gained a ` UTC` anchor on its `strtotime()`, since `user_registered` is stored in GMT and the parse would otherwise depend on the server's default zone.

Pinned with a test on a UTC-5 site where all three instants sit just after midnight UTC: `memberSince`, `lastPayment` and the group's `createdAt` each come back as the previous (site-local) day, so the group-owner card cannot show adjacent rows a day apart.

plugins/newspack-plugin/src/wizards/subscribers/format.js
adekbadek: Fixed the comparison rather than the comment, in 4326915: `today` is now `dateI18n( 'Y-m-d' )`, which reads the site timezone from `getSettings()` — the same basis the endpoint derives `endDate` on with `wp_date()`. The local `fmtRawDate` helper existed only for this and is gone.

A test pins it where the two bases disagree: a site 10 hours behind UTC at an instant that has already rolled over in UTC but not on the site, where an `endDate` of that site-local day must still read "Ends".

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscriber-detail-endpoint.php
adekbadek: Moved the tags to the class docblock — and did the two sibling files in the same commit (4326915) rather than leaving a follow-up, since it is the same two-line move and leaving them dead is what makes the group filter untrustworthy. `--group subscribers-wizard` now runs 44 tests across the three files; it ran 0 before.

plugins/newspack-plugin/src/wizards/subscribers/components/SubscriptionCard.jsx
adekbadek: Done in 4326915 — `CardRow` is now `dl`/`dt`/`dd`. Each row is its own one-pair list rather than one `dl` around all of them: the rows are laid out by `Grid`, whose own wrapper element would otherwise sit between the `dl` and its `dt`/`dd`. The stylesheet clears the default list margins (including the `dd` indent), and the rendered profile is pixel-identical.

plugins/newspack-plugin/src/wizards/subscribers/data/use-subscriber.js
adekbadek: Both micro-nits fixed in 4326915: `NOT_FOUND_CODES` moved above the JSDoc so it sits with the function it documents, and the file-wide eslint-disable is gone from `PersonProfile.jsx` — it turned out to be unnecessary, every string in the file already carries its translator comment, so removing the line restores the rule's coverage for future strings without needing per-line disables. Lint is clean on the file.

Left `rest_no_route` in `NOT_FOUND_CODES`: the code alone cannot separate "bad id" from "route absent", and the wrong branch for a deactivated plugin is a Retry that also cannot succeed. Happy to revisit if it ever confuses anyone in support.

plugins/newspack-plugin/src/wizards/subscribers/use-portals.js
adekbadek: Added the `classList.remove` to the cleanup in 4326915 (tracking the node the class was actually applied to, since the selector may resolve to a different node later). Verified in the browser: navigating from a profile back to the list now leaves no `newspack-subscribers__has-avatar` on the shared header, where it previously stayed.

plugins/newspack-plugin/tests/unit-tests/wizards/subscribers/class-subscriber-detail-endpoint.php
adekbadek: Added an explicit save-and-restore in 4326915: `set_site_timezone()` records the previous `timezone_string`/`gmt_offset` the first time it is called, and `tear_down()` puts them back — so the restore also happens when a test fails part-way, which an in-test restore would miss. Both timezone tests go through it now.

adekbadek:
On the three UTC-formatted fields — `prepare_group()`'s `createdAt`, `memberSince` and `last_payment_date()` — all three now localize, in 4326915. Instead of adding `wp_date()` at each call site, every date the wizard emits goes through a single `local_date( $timestamp )` helper, so the basis is one decision in one place; `joinedAt` and `subscription_date()` route through it too. `memberSince` additionally anchors its `strtotime()` to UTC, since `user_registered` is stored in GMT. A test on a UTC-5 site with all three instants just after midnight UTC asserts each comes back as the previous, site-local day — the adjacent-rows-a-day-apart case on the group-owner card.

---

On the "Ends/Ended" label reading the viewer's clock: `scheduleRow()` now derives today with `dateI18n( 'Y-m-d' )`, which resolves the site timezone from `getSettings()` and so matches the basis the endpoint used to produce `endDate`. The browser-local `fmtRawDate` helper is deleted. Pinned by a test at an instant where the two bases genuinely disagree — a site 10 hours behind UTC, after the UTC day has rolled over but before the site's — where the label must still read "Ends".

dkoo (commented):
Release risk: LOW — the follow-up resolves both date findings completely: every profile date now shares the site-timezone basis through one server helper (no double-conversion path for any source type), and the Ends/Ended label computes today as the site's civil date through the same settings the server uses, with tests that genuinely fail under the old behavior. The five opportunistically cleared nits are all real fixes, including making the previously-dead test group tags actually effective. Three small hardenings worth a moment while the file is warm: the site-timezone JS test pins only the negative-offset direction — a mirrored positive-offset case (site ahead of UTC) would catch the symmetric regression; the test's settings/timer restoration runs inline after the assertions, so a failing expect leaks Honolulu settings into the rest of the suite — a try/finally or afterEach makes it failure-safe; and the relative "time ago" text still parses its date browser-locally next to now-site-based absolute dates — coarse enough to be cosmetic, but anchoring it the same way as the absolute formatter would complete the unification. Nothing further needed from this review's perspective.

- `plugins/newspack-plugin/src/wizards/subscribers/format.js:13` — **Nit:** fmtRelative remains a viewer-basis remnant — `humanTimeDiff( '2026-07-20' )` parses the bare site-local date as browser-local midnight, so the "X days ago" text under Member since still anchors on the viewer's zone next to now-site-based absolute dates — one unit off near boundaries. Coarse enough to be cosmetic; anchoring it like `fmtDate` (`${ value }T00:00:00+00:00`) completes the unification, or document the choice.

plugins/newspack-plugin/src/wizards/subscribers/format.test.js
```diff
@@ -0,0 +1,116 @@
...
+		expect( row.value ).toBe( '—' );
+	} );
+
+	// The endpoint derives endDate in the SITE's timezone, so "today" must be read
+	// on that same basis. Pinned with a site 10 hours behind UTC and an instant
+	// that has already rolled over in UTC but not on the site: the plan ends
+	// *today* for the publisher, so it reads "Ends", whatever zone the admin's
+	// browser is in.
+	it( "decides Ends/Ended on the site's calendar day, not the viewer's", () => {
+		const settings = getSettings();
```
dkoo: **Suggestion:** Harden the site-timezone test: positive-offset mirror + failure-safe restore — Two related hardenings: the new site-tz test pins only the negative-offset direction (Honolulu, site day lags UTC) — the symmetric failure mode (site ahead of UTC, e.g. UTC+13, where the site's day rolls over before the viewer's) is untested, so a regression to browser-local or UTC "today" would still pass in one direction; and the `setSettings( settings )` / `jest.useRealTimers()` restoration runs inline after the assertions, so a failing expect leaks the fixture settings and fake timers into every subsequent test (the suite only runs whole). A mirrored positive-offset case plus a try/finally (or afterEach) closes both.

dkoo (commented):
Delta re-review of `7a3b63379` (root `phpcs.xml` + three call sites).

The ruleset change is well-judged: keeping the `nopaging` group enabled while deferring only the two new error codes is the right granularity, and the comment explains the ~66-call-site cost honestly. The three call-site fixes are unrelated to this PR's feature, but harmless.

The problem is coordination, not correctness:

**Main deliberately pinned VIPCS, and Dependabot undid it.** [#740](https://github.com/Automattic/newspack-workspace/pull/740) pinned `~3.0.1` "until posts_per_page audit (NPPM-3055)" on Jul 27 19:11; [#744](https://github.com/Automattic/newspack-workspace/pull/744) (Dependabot) bumped to `~3.1.0` nine hours later, before the audit. That bump is why this PR needs a ruleset adaptation at all.

**[#693](https://github.com/Automattic/newspack-workspace/pull/693) is solving the same problem the opposite way** — it re-pins root `composer.json` to `3.0.*` rather than adapting the ruleset. If both land, the repo runs 3.0.* with a ruleset written for 3.1.0: these two exclusions would reference error codes that do not exist in 3.0.x, i.e. dead config that reads as intentional to the next person.

Neither PR is wrong; the repo just cannot hold both decisions. Worth lifting into a single dedicated CI PR against main — adopt 3.1.0 with this ruleset change, or re-pin and defer — and dropping it from both feature branches so this PR stays about the person profile.

No concerns with the product code — unchanged since my last review.

Review provenance: orchestrator-verified directly (small CI delta; no reviewer fan-out dispatched — the decisive evidence was the composer.json history on main, checked with git); release risk: LOW for the product change, but see the coordination note above.

plugins/newspack-plugin/src/wizards/subscribers/format.test.js
adekbadek: Both done in 0b2c9a3.

The site-tz case is now a nested `describe` with the mirrored direction alongside the original: Honolulu (UTC-10) at `2026-01-02T05:00Z`, where UTC has rolled over and the site has not, and Kiritimati (UTC+14) at `2026-01-01T12:00Z`, where the site has rolled over and UTC has not. Kiritimati rather than a UTC+13 zone because it is +14 year-round with no DST, so the fixture cannot drift with a tzdata update the way Apia's abolished DST could. Restoration moved to an `afterEach` on that describe, covering both cases and surviving a failing expect.

Checked that the pair actually bites rather than assuming it: temporarily swapping `scheduleRow`'s `dateI18n( 'Y-m-d' )` for `gmdateI18n` fails both tests — so neither a UTC nor a browser-local "today" can pass both directions — while the other 484 tests still pass, which is also the `afterEach` doing its job through two failing expects. Restored, full suite green at 486/486.

adekbadek:
On `fmtRelative` and the viewer basis: I went to make the `${ value }T00:00:00+00:00` change and found the premise doesn't hold — `humanTimeDiff` isn't viewer-based. It parses with `momentLib.tz( from, WP_ZONE )` ([@wordpress/date src](https://github.com/WordPress/gutenberg/blob/trunk/packages/date/src/index.ts)), so a bare `2026-07-20` is already midnight in the *site's* zone, not the browser's. The relative text and the absolute date above it were landing on the same calendar day by different routes.

So the unification would have been a regression rather than a completion. I applied it to check, and it breaks in the direction you'd expect once the anchor moves to UTC — on a site ahead of UTC, a date on the site's current day is still in the future in UTC:

| site | `fmtRelative` today | with a UTC anchor |
| --- | --- | --- |
| Kiritimati (+14), site clock 02:00 | `2 hours ago` | `in 12 hours` |
| Honolulu (−10), site clock 19:00 | `19 hours ago` | `a day ago` |

"Member since — January 2, 2026 / in 12 hours" is worse than the mechanism mismatch that prompted the note.

Kept the behaviour and pinned it instead, in 3a5793a: a two-direction `fmtRelative` block in `format.test.js` on those exact instants (both rows above are assertions; I confirmed they fail with the UTC anchor applied and pass without it), plus a comment on `fmtRelative` saying why it deliberately doesn't anchor the way `fmtDate` does. The two helpers differ because their jobs differ — `fmtDate` anchors at UTC to *print* the stored day verbatim, `fmtRelative` only *measures*, and core already measures from the site's zone.

Also folded the timezone fixture the `scheduleRow` tests were using into a shared `pinSite`/`unpinSite` pair so both blocks set up the same way. JS suite 509 passing, PHP 2420.

adekbadek:
Re your CI-coordination note ([review](https://github.com/Automattic/newspack-workspace/pull/726#pullrequestreview-4801681146)) — settled, and along the lines you suggested:

- **The 3.1.0 adoption was lifted into a dedicated CI PR against `main`.** #741 (merged Jul 28, closes NPPM-3055) addressed all 66 `posts_per_page`/`numberposts` occurrences across the workspace and reverted the interim `~3.0.1` pin so `main` resolves vipwpcs 3.1.0 and lints clean. It took the *fix-the-call-sites* route rather than deferring the codes in `phpcs.xml`, so there's no ruleset config referencing 3.1.0-only codes to go stale.
- **Dropped from this branch.** #726 no longer touches `phpcs.xml` or `composer.json` — it's back to just the person profile, and PHPCS is green under main's 3.1.0.
- **The #693 conflict is gone.** #693 no longer re-pins to 3.0.*, so the repo isn't holding two opposing decisions.

Nothing left to coordinate here.

adekbadek:
@dkoo Re-requesting review – the changes you asked for have been pushed and need a look before this merges:
- test(subscribers): pin both site-timezone directions (NPPD-1753)
- ci(phpcs): re-enable the VIPCS 3.1.0 NoPaging error codes (NPPD-1753, #726)
- test(subscribers): pin fmtRelative to the site zone (NPPD-1753, #726)

dkoo (commented):
Both branch commits check out. The ruleset change re-enables the two VIPCS 3.1.0 NoPaging error codes by dropping the deferral, which is the right call now that #741 annotated the ~66 call sites on `main` and left the sniff on — keeping the deferral would re-disable a rule the codebase already complies with. `composer phpcs` over the monorepo is clean. This also settles the CI overlap with #693: `main` handles 3.1.0, #693 dropped its pin, and this branch defers to the sniff rather than excluding it, so all three agree.

The timezone test pins `fmtRelative` to the site zone with two fixtures on opposite sides of UTC (Honolulu at -10, Kiritimati at +14), so a regression to UTC or to browser-local "today" fails one direction or the other. Good pin.

Release risk: LOW. The product code is unchanged since the prior review — this delta is CI plus a test.

Review provenance: orchestrator-verified directly (CI + test delta; I read both commits and confirmed the re-enable matches main's #741 resolution). No reviewer fan-out for a CI/test-only change.

dkoo (approved):
Manual QA confirms the person-profile flow: a subscriber's profile opens from the Subscribers wizard and its in-wizard navigation works, plans and group cards render with the right roles and badges, and the pending-cancel and timezone cases show what they should. With the code review across its rounds, that closes #726.

The vipwpcs CI question that briefly overlapped with #693 is settled here: this branch re-enables the VIPCS 3.1.0 NoPaging codes to defer to main's #741, `composer phpcs` is clean over the monorepo, and the timezone test pins `fmtRelative` to the site zone from both sides of UTC. The product code held clean across the full review and its delta rounds.

Release risk: LOW. Approving.

Review provenance: deep + standards + WP expert + Codex across the review and its delta rounds, plus manual QA in an isolated env.

PR #727, opened 2026-07-25
feat(subscribers): admin group detail with write path (NPPD-1753)

adekbadek:

Adds the **L1 group-detail screen** and the Subscribers wizard's **first write endpoints**. A group row now opens an in-wizard detail screen — members table (owner / managers / members with roles, sorted owner → managers → members), pending invitations, seat count, and a slide-in billing drawer — replacing the click-through to the native subscription edit screen.

Part of NPPD-1753 (admin-side Subscribers management). It consumes the shared `SubscriptionCard`, `scheduleRow`, and `subscription_billing()` helper from #726, which has landed; this PR is based on `main`.

**Write path (`newspack-group-subscription/v1`):** reuses the existing reader-facing routes (`/members`, `/invite`, `/invite-link`, `/search-users`) rather than forking wizard-namespaced copies, and adds two new routes to that same class — `/manager` (promote/demote) and `/seat-limit` — because they have no My Account equivalent but share the role model and the peer-manager guard. A single `manage_options`-gated `/groups/<id>` read lives on the wizard.

**Permission model, enforced server-side (the UI is not the enforcement point):**
- Promote/demote managers is **owner-only**; a manager may add/remove plain members only, never a peer manager.
- Seat-limit changes are **admin-only**; the owner cannot move their own limit.
- No money action is reachable — `/seat-limit` writes only the capacity meta, and the prototype's "increase & send a payment link" branch is dropped. `AddMembersFlow` **invites** unknown emails rather than silently creating reader accounts (a consent decision — the design's "creates an account" text should be updated to match).

**A shipped-behaviour correction, disclosed:** the shared `permission_callback` already admitted a `manage_woocommerce` admin, but `/invite-link` then refused them inside `generate_link_invite()`, which mints only for a manager — so an admin could never mint a link at all. This PR resolves the acting manager to the **owner** for admin callers, so the mint succeeds and the admin operates on the group's link (the one shown in My Account). The change is strictly additive — for an owner or manager the resolver returns self, byte-identical to before, pinned by regression tests.

**Ownerless groups and search bounds:** both `/invite-link` call sites refuse a resolved manager of 0 rather than writing a link under no account, which is what an ownerless subscription resolves to now that `main` no longer seeds user 0 into `get_managers()`. `/search-users` is bounded at the same time: 50 rows per query and a two-character minimum term, both inside the existing query-args filter. The role allowlist that used to narrow the candidate queries is gone as of `main`'s eligibility-filter work, so the cap now applies before the `is_eligible_member()` post-filter.

Closes NPPD-1753.

1. On a site with a group subscription (owner + ≥1 manager + ≥2 members + a pending invite), open **Audience → Subscribers → Groups** and click the group.
2. Confirm the members table shows roles/badges sorted owner → managers → members, the pending invite, and the seat count; open the billing drawer.
3. As the owner, promote/demote a member; confirm a manager cannot promote/demote or remove a peer (server returns 403).
4. Adjust the seat limit; confirm it persists (check the subscription meta) and that a value below committed seats is refused.
5. Confirm seat limit `0` reads as unlimited, not zero.
6. Open the **Add members** menu with no invite link yet. Confirm the item reads **Create invite link**, and that using it reports the link was created.

### Other information:

Disclosures: reuses the reader-facing group API (justified above); `AddMembersFlow` invites rather than creates accounts; the drawer's money CTAs and `NoticesPanel` are deliberately out of scope. Note the Groups tab does not yet claim `/groups/:id` via `activeTabPaths` (the person profile does for `/subscribers/:id`) — a small parity follow-up.


plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
```diff
@@ -0,0 +1,518 @@
...
+			<DataViews
+				data={ processedMembers }
+				fields={ memberFields }
+				view={ membersView }
+				onChangeView={ setMembersView }
+				actions={ memberActions }
+				paginationInfo={ membersPagination }
+				defaultLayouts={ { table: {} } }
+				getItemId={ item => item.id }
+				search
```
chickenn00dle, line 516: **Blocker:** Clicking a row's ⋮ never opens a menu, so no member action is reachable from the UI. Same for the Invitations table. Reproduced in a browser on an active group, clicking the button's centre.

The trigger is fine — `aria-haspopup="menu"`, real `onClick`, enabled — and a programmatic `.click()` flips `aria-expanded` to `true`. But no menu mounts: `[role="menu"]` returns only admin-bar menus, and no error fires at click time.

Not environmental: the merged Institutions screen uses the same pattern and opens fine in this env. Also ruled out by testing — stale assets, `getItemId` returning a number (patched to `String()` and rebuilt, still broken), two DataViews on one page, the function `label`, and `isManageable()`.

**Lead:** the three working screens each declare an `isPrimary` action; this one declares none. `ItemActions` computes `primaryActions` and `eligibleActions` separately and branches on them.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
```diff
@@ -196,6 +250,199 @@ public static function permission_callback( $request ) {
...
+	public static function resolve_link_manager_id( $subscription ): int {
+		$subscription = WooCommerce_Subscriptions::sanitize_subscription( $subscription );
+		if ( ! $subscription ) {
+			return 0;
+		}
+		$current_user_id = get_current_user_id();
+		if ( $current_user_id && Group_Subscription::user_is_manager( $current_user_id, $subscription ) ) {
+			return $current_user_id;
+		}
+		if ( current_user_can( 'manage_woocommerce' ) ) {
```
chickenn00dle, line 354: **Blocker:** `resolve_link_manager_id()` returns `get_user_id()`, which is `0` on a group with no customer, and both call sites (`:689`, `:702`) pass it straight into `generate_link_invite()` / `delete_link_invite()`.

It validates because `get_managers()` seeds its list unfiltered (`class-group-subscription.php:240`), so `user_is_manager( 0, $sub )` is true — which also makes `permission_callback()` (`:250`) pass for a caller with no session. Logged out, against such a group:

```
POST /search-users  -> 200   reader email addresses
POST /invite        -> 200   {"added_by":0, "email_sent":true}
POST /invite-link   -> 200   a working join link
POST /name          -> 200   group renamed
```

The owned-group control correctly refuses. `get_managers()` and `permission_callback()` aren't touched by this PR, so the hole is inherited — but this adds a new instance of it, and the diff already carries the correct uid-0 guard twice (`class-group-subscription.php:416`, and `prepare_group_invite_link()`).

**Consider:** filtering the falsy owner where the list is built, plus `if ( ! $manager_id ) return 403;` at both call sites.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-admin-api.php
```diff
@@ -0,0 +1,664 @@
...
+		$manager_id = $this->create_reader();
+		$member_id  = $this->create_reader();
+		$group      = $this->create_group( $owner_id );
+		$this->add_manager( $manager_id, $group );
+		$this->add_member( $member_id, $group );
+
+		wp_set_current_user( $manager_id );
+		$request = $this->role_request( $group->get_id(), $member_id, 'manager' );
+
+		// The general group-management callback DOES admit this manager — that is
```
chickenn00dle, line 246: **Suggestion:** The callbacks are asserted as plain function calls — 0 dispatches through `rest_get_server()` here, against 13 in the sibling wizard suite. So swapping `role_permission_callback` (`class-group-subscription-api.php:192`) or `admin_permission_callback` (`:220`) for `__return_true` breaks no test.

The comment above claims the opposite. It asserts `permission_callback()` is true *and* `role_permission_callback()` is false — both still hold under the mis-wiring it describes.

Two dispatched requests (a manager POSTing `/manager` and `/seat-limit`, both 403) close it. Worth doing here since #773 and #774 copy this suite.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -187,6 +187,24 @@ public function register_api_endpoints() {
 			]
 		);
 
+		register_rest_route(
+			NEWSPACK_API_NAMESPACE,
+			'/wizard/' . $this->slug . '/groups/(?P<id>\d+)',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ $this, 'api_get_group' ],
+				'permission_callback' => [ $this, 'api_permissions_check' ],
```
chickenn00dle, line 231: **Suggestion:** The read is `manage_options` (`$capability`, `:35`); every write is `manage_woocommerce`. A `shop_manager` can call `/manager` and `/seat-limit` but cannot open the screen; a `manage_options`-only role sees a full screen where every button 403s with no client-side signal.

Other Newspack wizard writes sit behind `api_permissions_check()` (e.g. `audience/class-audience-subscriptions.php:140-154`), so `$capability` governs the buttons. Write being looser than read matters most on the PR that sets the pattern for two money surfaces.

**Consider:** publishing the caller's effective write capability in `/groups/<id>` so the UI can disable rather than discover-by-403.

plugins/newspack-plugin/src/wizards/subscribers/flows/AddMembersFlow.jsx
```diff
@@ -0,0 +1,155 @@
...
+					failures.push( `${ email }: ${ e?.message || __( 'Something went wrong.', 'newspack-plugin' ) }` );
+				}
+			}
+		} catch ( e ) {
+			setError( e?.message || __( 'Something went wrong.', 'newspack-plugin' ) );
+			setBusy( false );
+			return;
+		}
+
+		const added = readerIds.length;
```
chickenn00dle: **Suggestion:** `added = readerIds.length` counts what was sent; `RemoveMemberFlow.jsx:32` does the same with `list.length`. Both ignore the `members_added` / `members_removed` maps the endpoint returns (`class-group-subscription-api.php:581-582`), and `update_members()` deliberately drops no-op IDs (`class-group-subscription.php:664`, `:683`) while still returning 200.

`use-group.js:23` already states the convention this breaks. A partly-applied batch reports the full number with nothing in the UI or a log to show which entries were dropped.

**Consider:** reading the response maps — two lines.

plugins/newspack-plugin/src/wizards/subscribers/flows/DisableLinkFlow.jsx
```diff
@@ -0,0 +1,40 @@
...
+	return (
+		<ConfirmFlow
+			title={ __( 'Disable invite link', 'newspack-plugin' ) }
+			confirmLabel={ __( 'Disable link', 'newspack-plugin' ) }
+			isDestructive
+			onCancel={ onClose }
+			onConfirm={ disable }
+		>
+			{ __(
+				"This disables the group's shared invite link — including any copy the owner already sent out. Every existing link stops working and nobody can join through it. You can create a new link at any time.",
```
chickenn00dle: **Suggestion:** The copy promises "Every existing link stops working and nobody can join through it." The write resolves to one manager slot (`class-group-subscription-api.php:702`), and links are stored per manager. Confirmed on the branch:

```
AFTER disable: owner link   = WP_Error(link_invite_not_found)
               manager link = valid
```

The wizard never surfaces a manager's link either (`class-subscribers-wizard.php:613`, deliberate per NPPD-2120), so `hasLink` can read false while one is live and "Copy invite link" mints a second.

This is the button a publisher uses when a link has leaked. **Consider:** softening the copy to what it does, or revoking every slot.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
```diff
@@ -0,0 +1,518 @@
...
+				</Button>
+			</Notice>
+		);
+	}
+
+	const hasSeats = seatsRemaining( group ) > 0;
+	const canInvite = isActive( group ) && hasSeats;
+	const hasLink = !! group.inviteLink?.active;
+
+	const copyInviteLink = async () => {
```
chickenn00dle, line 400: **Suggestion:** `await window.navigator?.clipboard?.writeText( url )` returns `undefined` rather than rejecting when `clipboard` is absent (insecure context, older browser), so the `try` completes and reports success with an empty clipboard. If `generateInviteLink()` resolved without a `url`, it writes `"undefined"`.

`RegenerateLinkFlow.jsx:28-39` handles this correctly with a comment explaining why — worth mirroring. Note this menu item also performs a persistent write (minting a shareable link) behind a label that reads as a copy.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
```diff
@@ -179,6 +179,60 @@ public static function register_routes() {
...
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ __CLASS__, 'api_update_seat_limit' ],
+				'permission_callback' => [ __CLASS__, 'admin_permission_callback' ],
+				'args'                => [
+					'subscription_id' => [
+						'type'              => 'integer',
+						'required'          => true,
+						'sanitize_callback' => 'absint',
+					],
+					'limit'           => [
```
chickenn00dle, line 248: **Suggestion:** `minimum => 0` is inert: WP only injects `rest_parse_request_arg` when no `sanitize_callback` is present, and this arg sets `absint` with no `validate_callback`. Harmless here since `absint` clamps.

It matters as a pattern — the wizard's own routes pair each sanitizer with `'validate_callback' => 'rest_validate_request_arg'` (`class-subscribers-wizard.php:168-175`), which is what keeps `minimum`/`enum` live. A follow-up copying this shape with `'minimum' => 1` would silently not enforce it.

The `role` enum at `:204-208` *is* enforced because it omits `sanitize_callback` — worth a comment, since the two args sit six lines apart and differ for a non-obvious reason.

plugins/newspack-plugin/src/wizards/subscribers/flows/ResendInviteFlow.jsx
```diff
@@ -0,0 +1,45 @@
...
+	};
+
+	return (
+		<ConfirmFlow
+			title={ __( 'Resend invitation', 'newspack-plugin' ) }
+			confirmLabel={ __( 'Resend invitation', 'newspack-plugin' ) }
+			onCancel={ onClose }
+			onConfirm={ resend }
+		>
+			{ createInterpolateElement(
```
chickenn00dle, line 35: **Suggestion:** This and `CancelInviteFlow.jsx:38-47` `sprintf` the email into the `createInterpolateElement` format string. `RemoveMemberFlow.jsx:53-65` documents why that's wrong and uses a `<name/>` token instead.

`sanitize_email()` makes the injection risk theoretical, but the translator risk isn't: `createInterpolateElement` throws if a translation mangles `<strong>`, taking the modal down. Both files also carry a blanket `eslint-disable @wordpress/i18n-translator-comments`, so translators get no context for `%s`.

**Consider:** the `<email/>` token form and per-string translator comments.

plugins/newspack-plugin/src/wizards/subscribers/data/use-group.js
```diff
@@ -0,0 +1,142 @@
...
+ * follows:
+ *
+ * 1. NONCE. Nothing here handles one. `@wordpress/api-fetch` is served by WP's
+ *    `wp-api-fetch` script, which registers a nonce middleware from
+ *    `wpApiSettings.nonce` at load and attaches `X-WP-Nonce` to every request it
+ *    makes. That is a header, not a path rule, so it covers the
+ *    `newspack-group-subscription/v1` routes below exactly as it covers the
+ *    wizard's own `newspack/v1` ones. Never hand-roll a nonce into a body.
+ *
+ * 2. ERRORS. A `WP_Error` returned from a REST callback is serialized by WP into
```
chickenn00dle, line 15: **Suggestion:** `use-group.js` and `GroupDetail.jsx` ship with no tests, while their read-side siblings have them (`data/use-groups.test.js`, `screens/GroupList.test.jsx`).

The PHP side is strong on authorization boundaries; the client-side mirror is what's untested — `memberActions`' `isEligible` gating (`GroupDetail.jsx:250,263,279`), the mint-on-first-copy branch (`:376`), and the "surface the server message verbatim" contract this file declares for later slices. A mount test would also have caught the broken action menus.

plugins/newspack-plugin/src/wizards/subscribers/flows/capacity.test.js
```diff
@@ -0,0 +1,57 @@
...
+		expect( seatsRemaining( { seatLimit: 0, seatsReserved: 42 } ) ).toBe( Infinity );
+	} );
+
+	it( 'never reports negative room when a group is over its limit', () => {
+		// Reachable after an admin lowers a limit that outstanding invitations
+		// later fill past; the flows slice arrays by this, so it must not go below 0.
+		expect( seatsRemaining( { seatLimit: 3, seatsReserved: 5 } ) ).toBe( 0 );
+	} );
+
+	it( 'treats a missing group as having no room rather than throwing', () => {
```
chickenn00dle: **Nit:** `it( 'treats a missing group as having no room rather than throwing' )` asserts `seatsRemaining( undefined )` is `Infinity` — unlimited room, the opposite of the name.

It's the safe-default question: with no group loaded, `AddMembersFlow.jsx:43` slices to `Infinity` and `hasSeats` is true, so the UI would offer Add/Invite rather than withhold. `GroupDetail`'s early return makes it unreachable today, but the test pins the behaviour nobody intended.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
```diff
@@ -0,0 +1,518 @@
...
+			},
+			{
+				id: 'make-manager',
+				label: __( 'Make manager', 'newspack-plugin' ),
+				isEligible: item => 'member' === item.role && isManageable( group ),
+				callback: async items => {
+					try {
+						await actions.setManagerRole( items[ 0 ].id, 'manager' );
+						onDone( sprintf( __( '%s is now a manager.', 'newspack-plugin' ), items[ 0 ].name ) );
+					} catch ( e ) {
```
chickenn00dle: **Nit:** A rejected promote/demote goes to `setSnackbar`, which auto-dismisses. Every other write here (`ConfirmFlow.jsx:44-48`, both token flows) uses a persistent `Notice` — 409 "no longer active", 400 "only an existing member can be made a manager" and 403 are worth reading twice.

These are also the only writes with no in-flight disable, so a second click before the refetch fires a duplicate request.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -427,6 +445,191 @@ private function subscription_edit_url( $subscription ) {
...
+	 * resolve: the collection endpoint already embeds a full `owner` object, and
+	 * ids-only would force the screen into an N+1 (nothing in the collection
+	 * carries a member's join date or the group's billing).
+	 *
+	 * @param \WP_REST_Request $request The request.
+	 *
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public function api_get_group( $request ): \WP_REST_Response|\WP_Error {
+		$this->group_subscriptions_cache = null;
```
chickenn00dle: **Nit:** `api_get_group()` inlines the three cache resets instead of calling `reset_request_caches()` (`:757-761`), which every other endpoint uses. A fourth memo added later will be reset everywhere except here.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -383,6 +383,45 @@ public static function can_actor_remove_member( $actor_id, $target_id, $subscrip
...
+	 * - Managers, plain members and outsiders may not — a manager who could
+	 *   promote could manufacture peers and route around the peer-manager guard
+	 *   that can_actor_remove_member() applies to removals.
+	 *
+	 * @param int                  $actor_id     The user attempting the role change.
+	 * @param \WC_Subscription|int $subscription The subscription object or ID.
+	 *
+	 * @return bool Whether the role change is permitted.
+	 */
+	public static function user_can_manage_roles( $actor_id, $subscription ): bool {
```
chickenn00dle: **Nit:** `user_can_manage_roles( $actor_id, $subscription ): bool` here, plus `resolve_link_manager_id()`, `reserved_seats()`, `api_set_manager_role()`, `api_update_seat_limit()` and `api_get_group()` elsewhere, type returns but not parameters. `$actor_id` is `int`, `$request` is `\WP_REST_Request`, `$subscription` a `\WC_Subscription|int` union — all expressible on 8.3.

Surrounding legacy code is untyped so consistency cuts both ways, but new methods on a security-relevant path are the right place to be complete.

chickenn00dle (changes requested):
Release risk: HIGH — two blockers, both reproduced on a live env; details in the inline comments. Verification is partial: the REST layer was exercised across six actor roles, but the broken action menus block the UI flows.

The permission model this PR is about holds up. Promote/demote is owner-or-admin only (manager, member and outsider all 403), the owner cannot be demoted, `/seat-limit` is admin-only and refuses a value below committed seats while accepting `0` as unlimited, the wizard read is admin-only, and I found no way to act across groups. The disclosed admin→owner invite-link fix verifies byte-identically, and a manager still resolves to self. `user_can_manage_roles()` really is the single authority it claims to be — the manager meta key is written in three places, two of which defer to it and the third is a removal cascade gated by `can_actor_remove_member()`.

**To pass from my side:** the two blockers — the action menus, and the ownerless-group case. The suggestions and nits are non-blocking, though the untested callback wiring and the read/write capability split are worth taking now, since this slice sets the pattern #773 and #774 will copy.

Also worth flagging: Copilot never ran on this PR (quota limit on 25 July), so it has had no prior review.

---

### Not inline-commentable

These sit outside the diff, so GitHub can't anchor them.

**Root cause of the ownerless blocker — `class-group-subscription.php:240`.** `$managers = [ $subscription->get_user_id() ]`, unfiltered. Filtering the falsy owner here closes the class rather than each entry point; `get_all_members()` already does this.

**Why it's reachable.** The group flag resolves from the product first (`class-group-subscription-settings.php:315`), and the shipped `migrate-team-products` command sets it there with no customer awareness (`includes/cli/class-teams-migration.php:991`) — so every subscription of that product reads as a group. WooCommerce separately zeroes `customer_id` on a deleted user's orders by direct SQL, and the Subscriptions admin screen writes `0` with no validation when the field is absent. None of that needs an unusual site. The one limiter today is that these routes don't register unless `NEWSPACK_CONTENT_GATES` is defined, which confines exposure to pilot sites — and is the argument for fixing before the flag widens.

**`/search-users` is unbounded — `class-group-subscription-api.php:478`.** Two `get_users()` calls with no `number` cap, returning each reader's email (`:516`), and `search` has no minimum length so an empty term matches everyone. Any group manager can call it. It's the largest amplifier of anything wrong in the check in front of it. **Consider:** a result cap and a minimum length.

**Invite attribution disagrees with itself — `class-group-subscription-invite.php:442`.** `'added_by' => get_current_user_id()` feeds `*SENDER_NAME*` / `*SENDER_EMAIL*` (`:474`), so an admin's email invite names the admin while their link invite is attributed to the owner. On an ownerless group the sender fields render empty. Worth a deliberate decision recorded next to the resolver.

**`groupDetailHref()` is stale — `src/wizards/subscribers/links.jsx:44-59`.** Its docblock says to switch to `#/groups/${ id }` once the route is registered; this PR registers it (`index.js:77`). Today a group row goes in-wizard while the same group's card on a person's profile leaves for the WooCommerce order screen.

---

Since this slice is the template for #773 and #774, two conventions are worth settling here rather than three times: dispatch permission tests through the REST router, and pick one capability for reading and writing a surface.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Not reproducible here, and the `isPrimary` lead doesn't hold: `ItemActions` renders `CompactItemActions` whenever `primaryActions.length < eligibleActions.length`, which is true with zero primary actions, and `PrimaryActions` returns `null` on an empty array.

`screens/GroupDetail.test.jsx` (new in ac64bce) mounts this screen against the real `@wordpress/dataviews` with these exact actions and no `isPrimary`: the ⋮ opens and offers `Make manager` / `Remove member`, and the cancelled-group and manager-row cases gate as intended.

That mount does surface one thing: the owner's row has no eligible action — `view-profile` needs an `editUrl`, and promote/demote/remove all exclude the owner — so it renders no ⋮ at all. On a two-person group that leaves a single trigger.

To isolate it: which row was clicked and whether its ⋮ was the only one on the page, and whether any error appears at page load rather than at click time. `@wordpress/dataviews` is bundled from `node_modules` while `wp-components` is external, so a `Menu` resolved from `wp.components`' private APIs that doesn't match what the bundled DataViews expects would throw on render, not on click — and those two can be different versions in an env.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
adekbadek: Fixed in ac64bce, both halves.

`get_managers()` no longer seeds a falsy owner (`class-group-subscription.php:239-241`), so `user_is_manager( 0, $subscription )` is false on an ownerless group and `permission_callback()` refuses a caller with no session. Both invite-link call sites return 403 on a resolved manager of 0 instead of passing it into `generate_link_invite()` / `delete_link_invite()`.

The seeded ID is now cast to `int`, so the strict `in_array()` in `user_is_manager()` also matches an owner whose `get_user_id()` returns a numeric string.

Two regression tests: an ownerless group has no managers and admits no logged-out caller through `permission_callback()`; `POST` and `DELETE /invite-link` as a store admin return 403.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-admin-api.php
adekbadek: Added `test_registered_routes_refuse_a_manager` in ac64bce: a manager dispatches `POST /manager` and `POST /seat-limit` through `rest_get_server()`, both 403. Since the general callback admits that manager, a 403 can only come from the stricter guard — so swapping either route's `permission_callback` for `__return_true` now fails.

`do_action( 'rest_api_init' )` moved into `set_up` so the routes exist to dispatch against.

plugins/newspack-plugin/src/wizards/subscribers/flows/AddMembersFlow.jsx
adekbadek: Both flows now read the response maps (ac64bce).

`AddMembersFlow` counts `members_added` and, when it is short of what was sent, adds a line naming how many entries were skipped as already-members or non-readers. `RemoveMemberFlow` counts `members_removed`. The invite count is now incremented per successful `invite()` rather than derived from `toInvite.length - failures.length`, which the new skipped-entries message would otherwise have corrupted.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Reworked along `RegenerateLinkFlow.jsx:28-39` lines in ac64bce: the link resolves first, an empty or missing `url` reports a failure to create rather than a successful copy, and the write is only claimed when `navigator.clipboard.writeText` exists and resolves. An unavailable clipboard reports "Invite link ready, but it could not be copied to your clipboard" — the link is minted either way, so that is not a failed action.

The menu item now reads `Create invite link` when there is no link yet and `Copy invite link` when there is.

`GroupDetail.test.jsx` covers the mint-on-first-use path and the unavailable-clipboard wording (jsdom has no `navigator.clipboard`, matching an insecure context).

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
adekbadek: Fixed in ac64bce rather than commented: every arg on both new routes now pairs its sanitizer with `'validate_callback' => 'rest_validate_request_arg'`, matching `class-subscribers-wizard.php:168-175`. That makes `minimum` live on `limit` and keeps the `role` enum enforced explicitly instead of by omission. A comment above the args block states the pairing rule.

This is a small behaviour change on `limit`: validation runs before sanitization, so a negative value now 400s instead of being clamped to 0 by `absint`.

plugins/newspack-plugin/src/wizards/subscribers/flows/ResendInviteFlow.jsx
adekbadek: Both switched to the `<email/>` token form in ac64bce, with the same comment `RemoveMemberFlow` carries. `sprintf` is gone from both files, so nothing reaches `createInterpolateElement` that a translation can unbalance, and the format strings no longer carry a `%s` for a translator to guess at. Each string gets a translator comment naming what `<email/>` stands in for, and the blanket `eslint-disable @wordpress/i18n-translator-comments` is removed from both files now that neither has an uncommented placeholder.

plugins/newspack-plugin/src/wizards/subscribers/data/use-group.js
adekbadek: Added in ac64bce.

`screens/GroupDetail.test.jsx` mounts the screen against the real `@wordpress/dataviews` and drives the row menus: eligible actions on a plain member, none on the owner's row or on a cancelled group, demotion-but-never-removal for a manager, a refused role change staying on screen, and the mint-on-first-copy branch. The barrel is mocked because it re-exports the wizard `Page`, whose `@wordpress/admin-ui` dependency ships ESM jest cannot transform; DataViews itself is real.

`data/use-group.test.js` pins the contract this file declares for later slices — a read failure and a write rejection both surface the server's own message — plus `resolveReaderId`, whose null answer is the switch that sends an address down the invite path instead of the add path.

plugins/newspack-plugin/src/wizards/subscribers/flows/capacity.test.js
adekbadek: Fixed the behaviour rather than the test name (ac64bce): `seatsRemaining` now returns 0 for a missing group, so the safe default matches what the name always claimed. The test asserts 0 and says why — with nothing loaded the flows withhold Add and Invite rather than offering them against a group they know nothing about. A seat limit of 0 on a real group still means unlimited.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Both addressed in ac64bce. A refused promote/demote now lands in a persistent `Notice` at the top of the screen, and `roleBusy` disables both actions while one is in flight, so a second click cannot fire a duplicate request. The two callbacks share one `changeRole` helper rather than repeating the try/catch.

`GroupDetail.test.jsx` covers the refusal case: the server's message is still on screen after the rejection settles.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Fixed in ac64bce — `api_get_group()` calls `reset_request_caches()` like every other endpoint.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
adekbadek: Typed in ac64bce: `user_can_manage_roles( int $actor_id, \WC_Subscription|int $subscription )`, `resolve_link_manager_id()` and `reserved_seats()` take `\WC_Subscription|int`, and `role_permission_callback()`, `admin_permission_callback()`, `api_set_manager_role()`, `api_update_seat_limit()` and `api_get_group()` take `\WP_REST_Request`. Every call site already passes one of those, and `sanitize_subscription()` still normalises the union.

The now-redundant `$actor_id = (int) $actor_id;` cast inside `user_can_manage_roles()` is gone.

adekbadek:
`get_managers()` no longer seeds the falsy owner: the customer ID is cast to `int` and the list starts empty when it is 0 (`class-group-subscription.php:239-241`). Fixed in ac64bce, with a test asserting an ownerless group has no managers.

---

`/search-users` is bounded in ac64bce. Both `get_users()` calls take `'number' => self::SEARCH_USERS_LIMIT` (20), and a term shorter than `self::SEARCH_USERS_MIN_LENGTH` (2 characters) returns an empty array before either query runs. Both bounds sit inside the `newspack_group_subscription_user_query_args` filter, so a publisher can still widen them.

`resolveReaderId()` is unaffected: it searches a full address, which is well over the floor and well under the cap.

---

`groupDetailHref()` now returns `#/groups/${ id }` (ac64bce), and the stale note about the route not existing is gone. The group list already pushed `/groups/:id` directly; this switches the remaining caller, the person profile's group cards. Its test asserts the new target and that the result passes `isInternalHashPath`, since the profile puts these hrefs through that guard.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
adekbadek: Took the second half of this. `/groups/<id>` now returns `canManage`, the caller's own `manage_woocommerce`, and `isManageable()`/`isActive()` fold it in, so the controls disable instead of 403ing on click.

Did not align the read down to `manage_woocommerce`, though. This screen lists reader PII: every subscriber's email, tags and newsletters. The nearest precedent for that data is `class-csv-exports.php:89-95`, which requires `list_users` *and* `manage_woocommerce` for the users export. Widening the read hands a `shop_manager` all of it, which is a bigger change than the asymmetry it would fix.

The `shop_manager` who can already call `/manager` and `/seat-limit` is reaching the group-subscription API that backs the reader-facing My Account page. `manage_woocommerce` there is the store-admin branch on a reader API, and this PR does not touch it.

plugins/newspack-plugin/src/wizards/subscribers/flows/DisableLinkFlow.jsx
adekbadek: Confirmed, and the docblock above it was wrong in the same way — it asserted "A group has ONE shareable link", which `LINK_META` (a `[ user_id => entry ]` map) contradicts. `RegenerateLinkFlow` carried the same claim and the same over-promise.

Scoped the copy rather than revoking every slot. Revoking all of them would change what the shared `/invite-link` DELETE does for owners and managers on My Account too, and the surviving-manager-link problem is NPPD-2120's. Both confirm dialogs now name the link they actually act on and say that a manager's own link is untouched.

adekbadek:
Recorded the decision next to `resolve_link_manager_id()`, and kept the split.

An email invitation stores its actual sender and names them, so a reader replying reaches the admin who invited them. A link cannot do the same: it is validated at click time against the manager whose slot holds it, and an admin has no slot, so minting one under their own ID produces a link that is dead on arrival. The email is a message from a person; the link is an artifact of the group. They answer to different owners on purpose.

The empty sender fields were a real gap and are now closed. `added_by` is absent on invitations issued before it was recorded and resolves to nothing once that account is deleted, so the sender falls back to the group's owner, and to the site when there is no owner — the ownerless case you flagged. The placeholders are publisher-editable, so a template reading "*SENDER_NAME* invited you" was rendering a headless sentence.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Follow-up: this is not this PR's bug. It reproduces on `newspack-newsletters`' shipped newsletters list, which is on `main` and unrelated to anything here.

Same env, same click, same result: `aria-expanded` goes to `true`, `.components-popover` count stays 0. `admin.php?page=newspack-newsletters-list` with one newsletter post is enough to see it.

The likely mechanism is a version pairing rather than either screen's code. `packages/components` depends on `@wordpress/dataviews@^10.2.0` (resolving 10.3.0) and `@wordpress/components@^30.6.0`. `@wordpress/components` is externalised to core's `wp.components` at runtime, so the bundled DataViews 10.3.0 — compiled against components 30 — drives its Ariakit menu against whatever core ships. On WP 7.1 that is far newer, and the ref contract its anchor relies on has moved. Hence `Function components cannot be given refs ... TooltipAnchor2` with frames spanning the plugin bundle and `wp-includes/js/dist/components.js`.

Consistent with the rest of the screen: the header's `Add members` dropdown is a plain `@wordpress/components` `Dropdown` and opens fine. Only the Ariakit-based DataViews row menus fail.

So the fix belongs in the dependency pairing, not in this PR. Worth its own issue.

adekbadek: Confirmed, and thanks for the clear repro. The menu is genuinely unusable, though not quite in the way it looks.

Reproduced on a group with an owner and one member. Clicking a row's Actions button does open a menu in the DOM: it reports `aria-expanded="true"`, the menu node is `visible` at opacity 1, measures 160x104, and its three items carry the right labels in near-black 13px text.

```
[role="menu"]  visible  160 x 104  z-index 1000000
  View profile | Make manager | Remove member
  item colour rgb(30, 30, 30) on transparent
elementFromPoint(centre) -> the menu's own item
```

It also hit-tests: `document.elementFromPoint()` at the menu's centre returns a menu item, so the element is on top and clickable.

None of it paints. Here is the member row captured at the moment all of the above was true:

Group detail member row, immediately after clicking Actions: the DOM reports an open, visible menu with three items, and nothing is painted

I checked that this is the screen and not the camera. Three separate capture paths render it blank, including a screenshot of the menu element itself, and the newsletters list on `main` captures its equivalent menu perfectly through the same path, borders and all. So the pixels are missing on this screen specifically.

Practically that means your report is right and the actions are unreachable, since a menu nobody can see is a menu nobody can use. It also means the failure is narrower than "the menu never opens", which matters for whoever fixes it: the actions array, eligibility and the click handling are all fine, and the problem is in rendering the surface. A test that drives the menu through the DOM will pass while the screen stays broken, so this one needs a pixel check rather than an assertion on menu items.

I have not found the cause yet. Worth knowing while looking: dataviews resolves differently depending on where the plugin was built, because a per-plugin `node_modules/@wordpress/dataviews` symlink points relative to the worktree root, which has no `node_modules`, so it dangles and resolution falls through to the hoisted copy. The newsletters bundle that paints correctly and this one did not necessarily come from the same version.

Leaving this open and taking the fix.

plugins/newspack-plugin/src/wizards/subscribers/flows/AddMembersFlow.jsx
```diff
@@ -0,0 +1,173 @@
...
+								'%d addresses were already members, or are not reader accounts, and were skipped.',
+								readerIds.length - added,
+								'newspack-plugin'
+							),
+							readerIds.length - added
+						)
+					);
+				}
+			}
+			for ( const email of toInvite ) {
```
chickenn00dle, line 84: **Suggestion:** The report-what-applied fix covers the member half, not the invite half — The member half now counts `members_added`. The invite half increments on any non-throwing call and reports "%d invitations sent", ignoring the `email_sent` flag `api_invite()` returns. `generate_invite()`'s docblock says callers must read it; the CLI does, and rolls the row back.

**Risk if not addressed:** pending invites count against the seat limit, so with mail failing the admin fills the remaining seats with invitations nobody received, then hits the member-limit error with nothing connecting it to delivery.

Kept a suggestion rather than a blocker — it needs broken mail and is recoverable via Cancel invite — but it's the second one I'd take.

**Risk if not addressed:** On a capped group each undelivered invitation still holds a seat, so the admin fills the remaining seats with invitations nobody received and is then locked out of inviting until each is cancelled by hand or expires — with nothing connecting the lockout to mail delivery.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-admin-api.php
```diff
@@ -0,0 +1,761 @@
...
+	 * Reset the mock subscription store between tests.
+	 */
+	public function set_up() {
+		parent::set_up();
+		global $subscriptions_database;
+		$subscriptions_database = [];
+		Group_Subscription::reset_cache();
+		Group_Subscription_Settings::clear_group_subscription_ids_cache();
+		// Register the routes so the wiring test below can dispatch through them.
+		do_action( 'rest_api_init' );
```
chickenn00dle, line 55: **Suggestion:** Both new dispatch tests inherit NEWSPACK_CONTENT_GATES from another directory — `register_routes()` returns early unless `Content_Gate::is_newspack_feature_enabled()`, which needs `NEWSPACK_CONTENT_GATES`. This class never defines it; its four sibling wizard test files all do in `set_up_before_class()`. It currently inherits the constant from `tests/unit-tests/content-gate/*`, which sorts earlier.

Run it alone and both new dispatch tests fail on a 404 instead of asserting anything about permissions:

```
$ n test-php --filter Test_Group_Subscription_Admin_API
Failed asserting that 404 is identical to 403. (x2)
```

Same for `--group subscribers-wizard`, where this file sorts first. It fails loudly rather than passing vacuously, so the cost is a confusing red build, not a lost guard. The standard `set_up_before_class()` define makes it self-contained.

plugins/newspack-plugin/src/wizards/subscribers/flows/RemoveMemberFlow.jsx
```diff
@@ -0,0 +1,82 @@
...
+	const count = list.length;
+	const name = 1 === count ? list[ 0 ].name || __( 'this member', 'newspack-plugin' ) : null;
+	const groupLabel = GROUP_LABEL.toLowerCase();
+
+	const remove = async () => {
+		// Report what the server removed, not what was sent: update_members() skips
+		// IDs that aren't readers or aren't members of this group and still answers
+		// 200, so a batch can be applied in part.
+		const result = await actions.removeMembers( list.map( member => member.id ) );
+		const removed = Object.keys( result?.members_removed || {} ).length;
```
chickenn00dle, line 35: **Nit:** A wholly-unapplied removal reports success with a zero — When `removed === 0` the flow still calls `onDone()`, so the success snackbar reads "0 members removed from the group". The old code overstated; this understates into nonsense. Treating `0 === removed` as a failure, like the partial-batch case in `AddMembersFlow`, would match the sibling flow the same commit rewrote.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.test.jsx
```diff
@@ -0,0 +1,173 @@
...
+	// have to say so rather than 403 on click.
+	it( 'withholds member management from a caller who may read the group but not write to it', async () => {
+		await renderDetail( group( { canManage: false } ) );
+
+		expect( screen.queryAllByRole( 'button', { name: 'Actions' } ) ).toHaveLength( 0 );
+		expect( screen.getByRole( 'button', { name: 'Adjust seats' } ) ).toBeDisabled();
+		expect( screen.getByRole( 'button', { name: 'Add members' } ) ).toBeDisabled();
+	} );
+
+	it( 'offers demotion, never removal, for a manager the owner promoted', async () => {
```
chickenn00dle: **Nit:** New test's name contradicts its own assertion — `it( 'offers demotion, never removal, for a manager the owner promoted' )` asserts `[ 'Remove manager', 'Remove member' ]`. Removal is offered, and correctly so.

Same defect round 1 raised against `capacity.test.js:27`, reintroduced in a file the fix commit added.

chickenn00dle (changes requested):
Release risk: HIGH — blocker 1 (row-action menus mount but never paint) is still unfixed.

Blocker 2 is genuinely closed. Against an ownerless group the nine routes that answered a logged-out caller with 200 in round 1 now all return 401, and a link minted under user 0 while the hole was open no longer validates. I walked the 18 in-plugin `get_managers()` call sites — My Account is unaffected. REST validation, `/search-users` bounds, `canManage`, seat-limit-0 and the invite-sender fallback all verify at runtime.

Ten items left, no new blockers, and four are the same shape — a fix that landed at one call site but not its twin: clipboard in `GroupDetail` but not `RegenerateLinkFlow`; `canManage` on every write but `cancel-invite`; report-what-applied on the member half of `AddMembersFlow` but not the invite half.

Two worth taking before merge: the test file (both new dispatch tests 404 outside a full run, including `--group subscribers-wizard`) and the invite half of `AddMembersFlow` (undelivered invitations hold seats, so failing mail quietly locks the group out of inviting). One caution — `GroupDetail.test.jsx` passes while driving the row menus through the DOM, so treat it as zero evidence on blocker 1.

Verified clear: the widened blast radius of `get_managers()` returning an empty array reaches no consumer as a permissive default. Every consumer already neutralised the falsy owner ID before this change — `get_all_members()` and `get_member_seat_limit()` filter empty IDs, the My Account members template guards the empty case before `get_users( [ 'include' => ... ] )`, and the migration CLI uses the list only as a skip set. No consumer branches on "no managers" into a more permissive path or indexes the list positionally.

Verified clear: the new 403 on the invite-link routes cannot reach a reader-facing caller. `resolve_link_manager_id()` returns 0 only when the caller is neither owner nor manager, and the shared permission callback admits exactly those two plus a store admin — so the only reachable combination is a store admin acting on a subscription with no customer. No previously-issued link changes validity.

- `plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php:628` — **Suggestion:** The invite-link read hardcodes the owner while the write resolves caller-first — The docblock says this mirrors `resolve_link_manager_id()`. It doesn't: the read hardcodes the owner, the resolver returns the current user first whenever they're a manager. For an admin who is also a manager, the screen shows the owner's slot while Disable and Regenerate act on their own. I tried to reproduce it and couldn't — promoting an admin is refused, since `update_members()` drops non-readers — so it needs an account holding both a reader role and `manage_woocommerce`. Docblock accuracy today: either have the read call the resolver, or drop the mirror claim.
- `plugins/newspack-plugin/src/wizards/subscribers/flows/RegenerateLinkFlow.jsx:32` — **Suggestion:** The clipboard fix did not reach the file it was modelled on — `GroupDetail` now probes before awaiting; this file still has the original form: ```js await window.navigator?.clipboard?.writeText( link.url ); copied = true; ``` With no `clipboard`, the chain resolves to `undefined`, the await succeeds, `copied` becomes `true` — "copied to clipboard" with nothing copied, and the modal shows the URL nowhere else. Note the reply describes `GroupDetail` as reworked along this file's lines; the defect is here. **Risk if not addressed:** If they paste the stale value — most often the old link, which regeneration has just invalidated — every recipient follows a link the same action killed, and because the flow reported success nobody has a reason to inspect it.
- `plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx:353` — **Suggestion:** cancel-invite is the one write the canManage gate misses — `cancel-invite` carries no `isEligible`, while `resend-invite` directly above it gates on `isActive( group )`. So a `manage_options`-only caller gets a live Cancel invite and a 403 on click — the failure `canManage` exists to prevent. `isEligible: () => isManageable( group )` matches the siblings. The state half is deliberately absent here (cancellation is allowed on any status); only the capability half is missing.
- `plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php:231` — **Nit:** One arg left out of the pairing convention — Every other wizard arg now pairs its sanitizer with `rest_validate_request_arg`; `search` is the lone exception. Nothing is inert today since it has no `enum` or `minimum`, but a `maxLength` added later would be silently ignored for exactly the reason this commit set out to fix.
- `plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx:1` — **Nit:** The blanket i18n-translator-comments disable survives in the three largest files — Removed in `ResendInviteFlow` and `CancelInviteFlow`, but seven files here still open with `/* eslint-disable @wordpress/i18n-translator-comments */`, including three the delta edits — over strings that need it: `'Owned by %s'`, `'%s not found.'`, `'Seat limit updated to %d.'`.
- `plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php:272` — **Nit:** Comment still encodes the contract this delta inverted — `// customer_id 0 -> get_managers() returns [0], an empty/phantom owner.` — no longer true after `ac64bce`. The test still passes; its assertions filter the zero anyway. The equivalent comment in `tests/unit-tests/content-gate/group-subscriptions.php` was updated, this one was missed.

plugins/newspack-plugin/src/wizards/subscribers/flows/AddMembersFlow.jsx
adekbadek: Both invite flows now read `email_sent` and count an invitation as sent only when the server says the email went out. An undelivered one lands in the failure notice naming the address, and states that the invitation still holds a seat until it is cancelled.

Reported rather than rolled back, unlike the CLI. `generate_invite()` deletes any prior invitation for the address before writing the new one, so rolling back on `AddMembersFlow`'s path would leave an address that already had a live invitation with none at all — the audit caveat the migration command documents for itself. `InviteMemberFlow` filters those addresses out and could roll back safely, but two sibling flows reporting the same failure two different ways is worse than either.

`InviteMemberFlow` had the same defect and got the same fix. New `flows/write-outcomes.test.jsx` pins both directions: a stored-but-unsent invitation raises the notice and never calls `onDone`, a delivered one reports "1 invitation sent."

Fixed in 906b49a.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-admin-api.php
adekbadek: Added the standard `NEWSPACK_CONTENT_GATES` define to `set_up_before_class()`, matching the four sibling wizard test files.

Both your repros are green now:

```
$ n test-php --filter Test_Group_Subscription_Admin_API
OK (27 tests, 50 assertions)
$ n test-php --group subscribers-wizard
OK (44 tests, 192 assertions)
```

Fixed in 906b49a.

plugins/newspack-plugin/src/wizards/subscribers/flows/RemoveMemberFlow.jsx
adekbadek: `0 === removed` now throws instead of calling `onDone()`. `ConfirmFlow` renders it as the modal's own error and leaves the modal open.

The message says the people are no longer members of the group, which is the only cause left: `api_update_members()` re-checks the peer-manager rule per target before calling `update_members()`, so a refusal arrives as a 403 rather than an empty result. Covered in `flows/write-outcomes.test.jsx`.

Fixed in 906b49a.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.test.jsx
adekbadek: Renamed to `offers demotion as well as removal for a manager the owner promoted`. The assertion was right — a manager is removable by the owner — so the name was the wrong half.

Fixed in 906b49a.

adekbadek:
`RegenerateLinkFlow.jsx` now tests for `window.navigator?.clipboard?.writeText` before awaiting it, the same shape `GroupDetail.copyInviteLink()` uses, so `copied` can no longer be true with nothing on the clipboard.

---

`cancel-invite` now carries `isEligible: () => isManageable( group )`. Not `isActive`, which resend uses: cancelling needs the write capability but not a live group, and clearing invitations off a lapsed one is exactly when it earns its keep.

---

`search` on the wizard's `/subscribers` route now pairs its sanitizer with `rest_validate_request_arg`, so every arg on that route declares both.

---

Dropped the blanket disable from all seven files and wrote the 22 translator comments it was hiding — including `SubscriberList.jsx`, which predates this branch. `no-bitwise` stays on the two list screens that need it. `eslint --ext .js,.jsx src/wizards/subscribers/` is clean with the rule live.

---

`class-group-subscription.php:272` now reads "customer_id 0 -> no owner, so get_managers() returns an empty list", which is what ac64bce made true.

adekbadek:
Took the first option: the read calls the resolver now, so the mirror claim is true rather than removed.

`prepare_group_invite_link()` resolves through `resolve_link_manager_id()`, the same call the mint and delete endpoints make. An admin who manages the group reads their own slot, which is the slot Regenerate and Disable act on.

One wrinkle worth flagging, because it changed the shape of the fix. Resolving alone regressed the read for a `manage_options`-only caller: the resolver returns 0 for them, so the panel reported no link on a group that has one. `canManage` already disables their buttons, so the honest state for that caller is "here is the link, you cannot change it" rather than a false empty. They fall back to the owner's slot for display only, and the docblock now says which half is the mirror and which is the fallback.

You couldn't reproduce it because `update_members()` drops non-readers, so the account has to be built directly. `test_reported_link_is_the_one_this_callers_buttons_act_on` does that — `add_member()` then `add_manager()` on an administrator — and asserts the reported URL carries the caller's own key once they manage the group. Reverting the read to the hardcoded owner fails that test and nothing else, on the manager-admin case specifically, so the divergence was real and is now covered.

The confirm copy in both flows moved with it. "This disables the owner's invite link" was only true for a non-manager admin; it now reads "the invite link shown above", which holds in either slot. The line about another manager's link being unaffected stays, since that is still the part a publisher needs.

Full PHP suite green (3405 tests), PHPCS clean, and the `subscribers` JS suites pass.

The other five notes are unaddressed so far — I did not want to fold six changes into an answer about one. The clipboard one in `RegenerateLinkFlow` looks like the real defect of the group and I will take it next unless you would rather it went in its own PR.

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Mechanism found, cause not yet. The menu mounts, hit-tests and reports itself visible because it genuinely is all of those things. What fails is one ancestor:

```
DIV.css-1mm9ctc-MenuMotionRoot   opacity: 0   transform: matrix(1, 0, 0, 1, 0, -4)
```

That is the entrance animation's starting state, and the animation never runs — `getAnimations()` on it returns an empty list. An `opacity: 0` ancestor still hit-tests and still computes `visibility: visible` on its children, which is why every DOM probe reports an open menu while nothing paints.

`prefers-reduced-motion` is the switch. Same click, same build, only that setting changed:

```
reduce         -> motion root opacity 1, transform none   -> menu paints
no-preference  -> motion root opacity 0, translateY(-4px) -> menu invisible
```

Reduced motion skips the animation and renders the final state, so anyone with that setting on sees a working menu and everyone else sees nothing. That is worth knowing before testing this: a browser with reduced motion enabled cannot reproduce it, which is the most likely reason it has looked inconsistent.

Two controls narrow it. The wizard header's own More menu, on the same page and also a core `Menu`, opens and paints — it has no motion root. The newsletters list has no `MenuMotionRoot` in its DOM at all.

I thought that last difference was the dataviews version, since this plugin takes the hoisted 10.3.0 while newsletters pins its own 16. I tested it rather than assuming: pinned `newspack-plugin` to `^16.0.0`, confirmed it resolved, rebuilt in the container and reloaded with a cache-buster. The motion root is still stuck at `opacity: 0`. So the version split is not the cause and a dependency bump is not the fix — I have reverted it and am not proposing that PR.

What is still unexplained is why the same core `Menu` renders with a motion root here and without one on the newsletters list. That is where I would look next.

Everything else you raised is already on the branch. Six of the notes were fixed in 906b49abac, which landed after you reviewed 52f16a4254, and the invite-link read is fixed in 23d7f26401. Re-checked each against current HEAD: the clipboard probe is in `RegenerateLinkFlow`, `cancel-invite` gates on `isManageable`, `search` pairs its validator, the blanket i18n disables are gone, the stale ownerless comment is updated, `--group subscribers-wizard` runs 44 tests clean with no 404s, and the `AddMembersFlow` invite half checks `email_sent` and reports the seat-holding case.

adekbadek:
- fix(subscribers): report writes that did not apply (NPPD-1753, #727)
- fix(subscribers): read the link the buttons act on (NPPD-1753, #727)

plugins/newspack-plugin/src/wizards/subscribers/screens/GroupDetail.jsx
adekbadek: Fixed in de9c2683a5, and folded into this PR rather than split out, because the screen cannot do without it.

The cause is CSS, not JavaScript. WP 7.1's `Menu` fades in with a transition on a wrapper element that only reaches its visible state once Ariakit stamps `data-enter` on the surface inside it. That attribute never arrives here — after three seconds the surface still carries `data-open="true"` and `data-side="bottom"` and nothing else — so the wrapper stays on the transition's first frame, `opacity: 0` and shifted 4px, for as long as the menu is open. The menu is mounted and clickable throughout, which is why it answered every DOM probe as present.

The whole transition sits behind `@media not (prefers-reduced-motion)`, so anyone whose browser asks for reduced motion gets the final state directly and sees a working menu. That is the only reason this was not visible to everyone, and it explains why it reproduced for you and not for others.

The fix supplies the state the missing attribute never delivers:

```scss
div:has( > [role="menu"][data-open] ) {
	opacity: 1 !important;
	transform: none !important;
}
```

The menu renders in a portal on `body`, so there is no wizard container to scope to and `:has()` is the only way to reach the wrapper from the menu it holds; `!important` beats the emotion rule on specificity. It costs the fade animation.

The group detail row-action menu rendering View profile, Make manager and Remove member

On why it belongs here: "Remove member" turned out to be reachable without the menu, since selecting a row surfaces it as a bulk action. "Make manager" is not — the row menu is its only route, so promoting a member was impossible on this screen. That is this PR's own feature failing, not a neighbouring one.

Two things worth flagging. This looks like a core bug — the transition is declared on the wrapper while Ariakit's enter detection reads the surface — so it deserves an upstream report; happy to file it. And the same rule repairs every DataViews row menu in the plugin, including the audience wizard's lists, which are broken the same way and had not been reported.

No test guards this. A test that drives the menu through the DOM passes against the broken build, which is exactly how it went unnoticed; the check that catches it is a rendered screenshot.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
```diff
@@ -502,7 +799,11 @@ public static function api_generate_invite_link( $request ) {
 	 */
 	public static function api_delete_invite_link( $request ) {
 		$subscription_id = $request->get_param( 'subscription_id' );
-		$result = Group_Subscription_Invite::delete_link_invite( $subscription_id, get_current_user_id() );
+		$manager_id = self::resolve_link_manager_id( $subscription_id );
```
chickenn00dle: **Suggestion:** A store admin can't disable a live link on an ownerless group — On a group whose owner account is gone, the screen shows an active link with Disable enabled, but the admin's DELETE returns 403 and the link keeps working. Since the link now belongs to the subscription, consider letting a `manage_woocommerce` caller reach `delete_link_invite()` without a manager to act as. Revocation only; minting can stay refused. The DELETE half of `test_invite_link_routes_refuse_an_ownerless_group` would change with it.

<details><summary>Evidence</summary>

Verified on the review env against group #75 (customer ID 0), with a link stored in the current shape to stand in for one the owner minted before their account was deleted:

```
GET /newspack/v1/wizard/newspack-subscribers/groups/75 → 200, inviteLink.active: true, canManage: true
DELETE /newspack-group-subscription/v1/invite-link (subscription_id 75) → 403 newspack_group_subscription_no_link_manager
validate_link_invite( 75, <key> ) afterwards → true
```

`resolve_link_manager_id()` returns the customer ID (0) for an admin and never looks at promoted managers, so this holds even when the group still has managers. The only ways to revoke that link are then cancelling the subscription or editing its meta by hand.

</details>

plugins/newspack-plugin/src/wizards/subscribers/flows/DisableLinkFlow.jsx
```diff
@@ -0,0 +1,43 @@
...
+	return (
+		<ConfirmFlow
+			title={ __( 'Disable invite link', 'newspack-plugin' ) }
+			confirmLabel={ __( 'Disable link', 'newspack-plugin' ) }
+			isDestructive
+			onCancel={ onClose }
+			onConfirm={ disable }
+		>
+			{ __(
+				'This disables the invite link shown above, including any copy already sent out. Nobody can join through it. A link created by another manager is not affected. You can create a new link at any time.',
```
chickenn00dle: **Suggestion:** Confirm copy says another manager's link survives; since #915 it doesn't — "A link created by another manager is not affected" stopped being true when main was merged in. Since #915 the invite link belongs to the subscription, and both writes replace or remove the whole `LINK_META`, legacy per-manager keys included. Could you drop that sentence here and in `RegenerateLinkFlow.jsx:54`, and rewrite both docblocks (lines 1–11), which still describe per-manager slots?

<details><summary>Evidence</summary>

`delete_link_invite()` deletes the whole meta (`class-group-subscription-invite.php:478`), and its comment says it "clears a legacy subscription's per-manager keys in one go". `generate_link_invite()` overwrites it with a single entry (`:431`). On a group still on the legacy shape, an admin who disables the link expecting a manager's already-shared link to keep working revokes that one too. The merge applied cleanly because these two files exist only on this branch, so nothing flagged them.

</details>

plugins/newspack-plugin/src/wizards/subscribers/flows/InviteMemberFlow.jsx
```diff
@@ -0,0 +1,136 @@
...
+			} catch ( e ) {
+				failures.push( `${ email }: ${ e?.message || __( 'Something went wrong.', 'newspack-plugin' ) }` );
+			}
+		}
+		if ( failures.length ) {
+			setError( failures.join( ' ' ) );
+			setBusy( false );
+			// A partial send still changed the group, so let the screen refresh
+			// without closing the modal over the error.
+			if ( sent > 0 ) {
```
chickenn00dle: **Suggestion:** When no invitation is delivered, the screen doesn't refresh to show the seat it took — If every email in a batch fails to send, the flow reports that each invitation "holds a seat until you cancel it", but it never refreshes the screen, so that invitation isn't in the table to cancel. Could the flows count invitations created separately from invitations delivered, and call `onDone( …, { keepOpen: true } )` whenever any were created? Same in `AddMembersFlow.jsx:139`.

<details><summary>Evidence</summary>

`generate_invite()` saves the invitation before it attempts the send (`class-group-subscription-invite.php:635` sets `email_sent` afterwards), so the row and the seat exist either way. The refresh here is gated on `sent > 0`, and in `AddMembersFlow` on `parts.length`. Once a flow finishes, `onDone` is the only path to `reload()` (`GroupDetail.jsx:129-139`); the other caller is the load-error Retry button, and `closeModal` only clears the modal. The admin is left with a stale seat count and Invitations table until they reload the page.

</details>

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -637,6 +637,50 @@ function( $invite_data ) {
 		return $new_invite;
 	}
 
+	/**
+	 * The name and address an invitation email is sent as.
+	 *
+	 * `added_by` is the person who issued the invitation, which is who a recipient
+	 * should see and reply to -- an admin inviting from the Subscribers screen is
+	 * named, not hidden behind the group's owner. That is deliberately different
+	 * from how an invite LINK is attributed: a link is minted in a manager's slot
```
chickenn00dle: **Nit:** `resolve_invite_sender()` docblock describes the old link validation — "A link is minted in a manager's slot and re-validated against that manager when it is clicked" no longer holds: `validate_link_invite()` deliberately ignores who minted the link (`:504-505`). The same sentence is in the test docblock at `tests/unit-tests/.../class-group-subscription-invite.php:572`. The reason that still stands is that `generate_link_invite()` and `delete_link_invite()` take a manager identity (`class-group-subscription-api.php:326-329`), so both could rest on that.

chickenn00dle (approved):
Left a few suggestions that should be addressed before this one is merged. They are fairly small though so feel free to merge without another round of review!

Review scope: 2 of 3 reviewer passes ran; skipped: codex.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-api.php
adekbadek: Done in ae696dd — `delete_link_invite()` takes an `$is_store_admin` flag that skips the manager requirement, set by `api_delete_invite_link()` from `manage_woocommerce`. Minting still refuses. The capability check stays at the API layer; the flag only drops the "act as a manager" requirement.

`test_invite_link_routes_refuse_an_ownerless_group` is split: POST still 403s, and the DELETE half revokes as the admin and asserts `validate_link_invite()` fails afterwards.

Also corrected the docblocks on `no_link_manager_error()` and `resolve_link_manager_id()` that said user 0 blocks both operations, and the same claim in the PR description.

plugins/newspack-plugin/src/wizards/subscribers/flows/DisableLinkFlow.jsx
adekbadek: Right — dropped that sentence from both confirm strings and rewrote both docblocks around the link belonging to the subscription. Disable now reads "Nobody can join through it, whoever shared it", Regenerate "That link stops working, whoever shared it". The `DisableLinkFlow` docblock also names the legacy-shape case, since that write clears those keys too. Fixed in ae696dd.

plugins/newspack-plugin/src/wizards/subscribers/flows/InviteMemberFlow.jsx
adekbadek: Fixed in ae696dd. Both flows now count invitations created separately from invitations delivered, and refresh whenever any were created. With nothing delivered there is no snackbar to show, so it is `onDone( '', { keepOpen: true } )` — the refresh runs, the modal stays open over the error, and the invitation appears in the table for the admin to cancel.

The `write-outcomes` case that asserted `onDone` was not called now asserts that exact call instead.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Fixed in ae696dd, in both places, resting on the reason you named: the link is validated without regard to who minted it, but `generate_link_invite()` still mints under a manager identity, and an admin is not one — so an admin mints the owner's link. Deletion no longer belongs in that reason, since it now accepts a store admin with no manager to act as (the other thread).

PR #736, opened 2026-07-27
fix(blocks): keep tag labels clear of cat-links styling (NPPM-3049)

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? *(phpcs/eslint/stylelint pre-commit gates passed on every commit)*
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? *(none open: #728 was closed as superseded by this branch, and #501 is merged)*

Tag labels take on whatever styling a publisher wrote for categories. A per-section rule like `.wpnbha .category-x .cat-links a` recolors them, because the label wrapper carries `cat-links` too. NPPM-3048 fixed the theme-rendered path; this closes the class of problem rather than the call sites that happened to be left.

The wrapper carries that class because `Tag_Labels::display()` puts it there by default. Tag labels were built to borrow category styling — it is how they picked up chip layout in the theme, the carousel and the block theme without declaring any of their own — and borrowing the styling means borrowing the publisher's category CSS with it. *(That the borrowing was deliberate is inferred from the commit that introduced the feature, not from a recorded decision.)*

So the class comes off at the source, and every unit that relied on the borrowed styling now declares its own: both blocks, the classic theme and its five style variations, and the block theme.

#### Why not strip it per caller

That is what this PR did at first, and it is how the carousel and the block theme came to be missed. Each caller has to remember to pass explicit classes, every new caller inherits the coupling by default, and the fix drifts apart one render path at a time — the editor preview and the front end had already diverged. It also moved callers off `display()` onto a five-argument helper, worth avoiding on its own.

Changing a shared default needs evidence, so an impact review enumerated the consumers: nine units — newspack-plugin, newspack-blocks, the classic theme with its five style variations, and the block theme. No standalone repository, and nothing in JavaScript reads the class.

Leaving the class in place and out-specifying publisher CSS was never a real option: publisher rules are arbitrary and the customizer generates its own, so there is no specificity that reliably wins against a selector you have not seen.

#### Decisions the diff will not show

- **Parity restores color, not only layout.** The carousel sits over a dark image and its labels were white; the block theme switches them to `currentcolor` on colored cards; homepage cards inherit in one context and force white in another. Layout-only parity leaves labels unreadable in precisely the places that look fine in a screenshot of a default card.
- **Spacing belongs to the style that renders a chip.** The gap between chips is part of the Default style's category-chip treatment, which the Homepage Posts and Carousel contexts deliberately switch off. It therefore sits with those rules rather than in the partial every style variation loads — otherwise the five variations inherit spacing meant for a chip their categories never render.
- **Wrapped rows are separated with `row-gap` on the container**, not a margin on the chips, because a bottom margin also hangs below the last row. That is why the block contexts zero the margin out in the first place.
- **Stylesheets group `.cat-links, .tag-labels`** rather than repeating declarations. Sharing a selector does not reopen the capture: the coupling lives in the markup class, which is what publisher and customizer CSS match on. The same reasoning puts `.tag-labels` back into the block theme's hover rules, where a label has no chip of its own to fall back on.
- **The `:not(.tag-labels)` exclusions in Jetpack Content Options and the archive stylesheet stay.** Our own markup can no longer put both classes on one element, but the outer class list is a parameter of `newspack_generate_tag_labels()`, so a child theme still can. Without the exclusions, that theme loses its labels on every archive.
- **The classic-theme commit keeps its own title and NPPM-3048.** That work is #728's, carried here when it closed; consolidating shouldn't relabel it.
- **Out of scope:** customer themes outside this monorepo that style `.cat-links` to reach tag labels. An impact review cannot see them, and they may need updating.

Fixes NPPM-3049. Also delivers NPPM-3048, which #728 previously carried.

1. On a site with newspack-plugin and newspack-blocks active, give a tag a label (Posts → Tags → edit a tag), assign it to a post, and add a Homepage Posts block showing that post with tag labels enabled.
2. Front end: the label wrapper is `<div class="tag-labels">` with no `cat-links`, and layout matches the previous rendering — size, weight, margins, alignment, RTL.
3. Editor: the block preview's wrapper is likewise `<div class="tag-labels">`, sized as before.
4. Adversarial check: add a per-section rule such as `.wpnbha .category-<slug> .cat-links a { background: red; }`. Category chips restyle; tag labels do not.
5. Repeat with a Carousel block and on a tag archive, and with each of the five style variations active. Labels stay inline beside a sponsor label, sit on one row, and keep their own chip colors.

### Other information:

Additional notes:

- **Verified in a browser**, not only by reading CSS: the parent theme, its five style variations and the block theme, each across single posts, Homepage Posts, the Carousel and a tag archive. Every context resolves to the same treatment — labels inline beside a sponsor label, chips on one row, chip colors unchanged.
- **Tests:** newspack-blocks 153 pass; newspack-plugin 3030 pass, with a new `Tag_Labels` contract test covering the wrapper's classes and its empty-input behavior. The nine skips in newspack-plugin are pre-existing and environment-dependent.
- **The fix, in one frame.** A per-section rule applied as site CSS — `.wpnbha .category-politics .cat-links a { background: #d00; color: #fff; }`. The category chip takes it; the tag labels do not. Before this change the same rule recolored all three, because the label wrapper carried `cat-links`.

A per-section category rule recolors the category chip and leaves the tag labels untouched


plugins/newspack-blocks/tests/class-newspack-tag-labels-stub.php
```diff
@@ -33,5 +33,39 @@ class Tag_Labels {
...
+			}
+
+			$outer_element = in_array( $outer_element, [ 'span', 'div' ], true ) ? $outer_element : 'span';
+
+			$labels_html  = '';
+			$labels_html .= '<' . $outer_element . ' class="' . join( ' ', array_map( 'esc_attr', $outer_classes ) ) . '">';
+			foreach ( $labels as $label ) {
+				if ( $links && isset( $label['flag'] ) && $label['link'] ) {
+					$labels_html .= '<a class="' . join( ' ', array_map( 'esc_attr', $inner_classes ) ) . '" href="' . esc_url( $label['link'] ) . '" rel="tag">' . esc_html( $label['flag'] ) . '</a>';
+				} elseif ( isset( $label['flag'] ) ) {
```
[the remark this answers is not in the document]
jason10lee: The unguarded `$label['link']` read is deliberate: the stub's contract is byte-for-byte fidelity with the real `Tag_Labels::generate_html()`, which has the same unguarded access — guarding it stub-side alone would hide in tests a warning production can still emit. NPPM-3051 tracks fixing the guard upstream in newspack-plugin and then re-mirroring the stub.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -330,4 +332,48 @@ public function test_tag_labels_rest_field_empty_returns_false() {
...
+			[
+				[
+					'flag' => 'Opinion',
+					'link' => 'https://example.org/tag/opinion/',
+				],
+			]
+		);
+		$html = ob_get_clean();
+
+		self::assertStringContainsString( 'class="tag-labels"', $html, 'Wrapper carries exactly the tag-labels class.' );
```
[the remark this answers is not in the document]
jason10lee: Tightened in b7051ab7d: the test now asserts the full opening tag `<div class="tag-labels">`, which proves the exact class attribute — and the div wrapper with it, so the separate `<div` assertion is folded in. The message now matches what's tested.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -1194,8 +1194,16 @@ public static function get_tag_labels( $post = null ) {
...
 	public static function display_tag_labels( $labels = null, $links = true ) {
-		if ( class_exists( '\Newspack\Tag_Labels' ) && method_exists( '\Newspack\Tag_Labels', 'display' ) ) {
-			\Newspack\Tag_Labels::display( $labels, $links, 'div' );
+		if ( class_exists( '\Newspack\Tag_Labels' ) && method_exists( '\Newspack\Tag_Labels', 'generate_html' ) ) {
+			// Render WITHOUT the `cat-links` wrapper class: per-section
+			// `.cat-links a` styling (a common per-section pattern on
+			// publisher sites) must never restyle tag labels (NPPM-3049,
+			// mirroring the theme-side NPPM-3048 fix). Layout parity lives
+			// in src/blocks/homepage-articles/view.scss.
+			$labels_html = \Newspack\Tag_Labels::generate_html( $labels, $links, array( 'tag-labels' ), array( 'tag-label', 'flag' ), 'div' );
```
adekbadek: **Blocker:** The Carousel shares this helper and gets the class removal without the replacement CSS — `display_tag_labels()` has two call sites — `src/blocks/homepage-articles/templates/article.php:121` and `src/blocks/carousel/view.php:137` — and only `homepage-articles/view.scss` gained the replacement `.tag-labels` rule. That rule is nested under `.wpnbha` (`view.scss:34`), so it cannot reach carousel markup, whose stylesheet roots are `.wp-block-newspack-blocks-carousel` and `.wpnbpc`. `carousel/view.scss:266` styles `.cat-links` with `font-size: 0.6em`, `font-weight: bold`, `margin: 0 0 0.5em` and `a { text-decoration: none }`, and declares no `.tag-labels` rule; the built `dist/carousel/view.css` contains zero `tag-labels` occurrences.

On a carousel with `showTagLabels` on (its default, `carousel/view.php:71`), the same rendered slide measured with and without `cat-links` on the wrapper, newspack-theme active: chip font-size 12px → 20px, wrapper font-weight 700 → 400, margin-bottom 6px → 0, chip box 75×19px → 125×32px, and the chip links gain an underline. Chip background and text colour are unchanged, because `newspack-tag-labels.scss:26` keys on `.tag-labels .tag-label.flag`, which the wrapper keeps — so this is sizing and spacing, not contrast. The category/tag-label inline pairing survives for the same reason.

`carousel/edit.js:416` also still hardcodes `className="cat-links tag-labels"`. In the editor canvas the Homepage Posts previews render `class="tag-labels"` while the Carousel previews render `class="cat-links tag-labels"`, so the Carousel preview keeps the class NPPM-3049 asks to remove, and shows the old sizing while its front end shows the new.

Mirroring the four non-colour declarations under `.tag-labels` in `carousel/view.scss`, and dropping the class at `edit.js:416`, covers both. The comment at :1201 would then need to name both stylesheets — as written it sends the next reader to `homepage-articles/view.scss` for a call path that file cannot cover.

**To reproduce**
1. Add a Carousel block to a page and leave `Show tag labels` on.
2. Give a tag a label (Posts → Tags → edit a tag) and assign it to a post the carousel shows.
3. View the page. The label chips render at body text size, unbolded, underlined, with no gap below.

**Risk if not addressed:** If the Carousel path ships without its replacement rule, label chips on affected sites grow from roughly 75x19px to 125x32px inside a fixed-height slide overlay, so on narrow viewports the chip wraps or pushes the headline out of the visible area on a homepage element above the fold, and stays that way until a follow-up release.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
```diff
@@ -485,6 +487,26 @@
 		}
 	}
 
+	// Tag labels: layout parity with `.cat-links`. The wrapper no longer
+	// carries `cat-links` — per-section `.cat-links a` styling must never
+	// restyle labels (NPPM-3049) — so the layout it relied on is declared
+	// here. Colors are intentionally not declared: chips keep their own.
+	.tag-labels {
```
adekbadek: **Blocker:** On newspack-block-theme the replacement rule restores none of what `.cat-links` supplied — `themes/newspack-block-theme/src/scss/blocks/newspack/_content-loop.scss:14` styles `.cat-links` with `color: var(--wp--preset--color--contrast)`, `font-size: var(--wp--preset--font-size--small)`, `text-transform: uppercase` and `a { color: inherit; text-decoration: none }`; `_post-carousel.scss:32` does the same for the Carousel. Both ship in the theme's built `style.css`. The theme has no `.tag-label` rule at all, so nothing there keys on the class the wrapper keeps.

With newspack-block-theme active, both variants rendered side by side in the same block container: the label link goes from `rgb(30, 30, 30)` to `rgb(0, 61, 165)`, `text-transform` from `uppercase` to `none`, and font-size from 16px to 11.81px. "BREAKING" renders as "Breaking" in the theme's default link blue, no longer matching the category beside it. This is Homepage Posts — the path this PR does add a replacement rule for; that rule declares font-size, weight and margin, and deliberately no colour or transform, which is exactly what the FSE theme was supplying.

The PR body's reasoning — colours are left out because chips keep their own — holds on newspack-theme, where `sass/plugins/newspack-tag-labels.scss:26` and the customizer inline style in `inc/newspack-tag-labels.php:149` both key on `.tag-labels` and survive. It does not hold on the FSE theme, which gives labels no chip and styled them purely through `.cat-links`.

The fix is theme-side and outside this diff: add `.tag-labels` alongside `.cat-links` in the block theme's content-loop and post-carousel partials, including the `color: inherit` group at `_content-loop.scss:161` used over the image overlay. Worth landing with this PR rather than after it.

**To reproduce**
1. Activate newspack-block-theme.
2. Give a tag a label and assign it to a post shown by a Homepage Posts block.
3. View the page. The label renders mixed-case in link blue instead of uppercase in the theme's contrast colour.

**Risk if not addressed:** If the block theme is active, tag-label links inside the image-behind card variant lose the `color: inherit` that made them white and fall back to the theme's default link colour over the dark image overlay, so the labels become hard to read on the homepage of affected sites until a theme release adds the missing rule.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -1194,8 +1194,16 @@ public static function get_tag_labels( $post = null ) {
 	 * @return void
 	 */
 	public static function display_tag_labels( $labels = null, $links = true ) {
-		if ( class_exists( '\Newspack\Tag_Labels' ) && method_exists( '\Newspack\Tag_Labels', 'display' ) ) {
-			\Newspack\Tag_Labels::display( $labels, $links, 'div' );
+		if ( class_exists( '\Newspack\Tag_Labels' ) && method_exists( '\Newspack\Tag_Labels', 'generate_html' ) ) {
```
adekbadek: **Suggestion:** The call moves off a stable entry point onto a five-positional-argument helper — Version compatibility is fine — `display()` and `generate_html()` both landed in 226df98653, so swapping the `method_exists` guard cannot fail on an older newspack-plugin. The cost is elsewhere: newspack-blocks moves off the three-argument `Tag_Labels::display()` onto a five-positional-argument helper, and re-implements `display()`'s `wp_kses_post( … . ' ' )` echo locally. A later parameter reorder in `generate_html()` would shift which class list lands on which element rather than failing loudly, and the escaping now lives in two places.

NPPM-3049 already scopes the fix as cross-repo. An `$outer_classes` parameter on `Tag_Labels::display()` would keep blocks on the stable method and the escaping in one place. If the plugin-side change is out of scope for this ticket, a `@see \Newspack\Tag_Labels::generate_html()` on the docblock at least ties the two together for whoever edits the plugin next.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: **Suggestion:** `.tag-labels` duplicates `.cat-links` instead of sharing the selector — The new rule repeats `display`, `flex-wrap`, `font-size`, `font-weight`, `margin` and the `a` / `a:hover` block verbatim from `.cat-links` at :466. The comment above it asks for "layout parity with `.cat-links`", but nothing holds the two in step — a later edit to one is silently a divergence. Grouping the shared declarations as `.cat-links, .tag-labels { … }` and leaving `&.sponsor-label` under `.cat-links` alone compiles to the same output and makes the parity structural. The same shape applies to the carousel stylesheet if you fix it there too.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
```diff
@@ -485,6 +487,26 @@
 		}
 	}
 
+	// Tag labels: layout parity with `.cat-links`. The wrapper no longer
```
adekbadek: **Nit:** The comment narrates the change rather than stating the invariant — "The wrapper no longer carries `cat-links` … so the layout it relied on is declared here" is changelog: on the merged branch there is no "no longer". The invariant is worth keeping and the history belongs in git. Present tense says the same thing — the wrapper carries only `tag-labels`, so per-section `.cat-links a` rules cannot reach it (NPPM-3049), and its layout is therefore declared here. The closing sentence about colours already has the right shape.

plugins/newspack-blocks/tests/class-newspack-tag-labels-stub.php
```diff
@@ -33,5 +33,39 @@ class Tag_Labels {
...
+		 *
+		 * @param array  $labels        Labels to display.
+		 * @param bool   $links         Whether to include links to tag archives.
+		 * @param array  $outer_classes Classes to apply to the outer container.
+		 * @param array  $inner_classes Classes to apply to the inner container.
+		 * @param string $outer_element HTML element to use for the outer container.
+		 *
+		 * @return string Tag labels as HTML.
+		 */
+		public static function generate_html( $labels = null, $links = true, $outer_classes = array( 'tag-labels' ), $inner_classes = array( 'tag-label', 'flag' ), $outer_element = 'span' ) {
```
adekbadek, line 69: **Nit:** Nothing detects drift between the stub and the real `generate_html()` — The stub reproduces the real `generate_html()` body, so the assertions run against a hand-maintained copy that goes stale silently if newspack-plugin's version changes — nothing in this suite would notice. What the test needs to pin is narrower: that `display_tag_labels()` passes `array( 'tag-labels' )` as the outer classes. Recording the received `$outer_classes` on a static and asserting on that states it directly and drops the copy.

If the rendered-markup assertion is deliberate — `.tag-labels` is what the CSS keys on, so proving the emitted string has standalone value — then naming the upstream file the stub tracks, in the stub's file docblock, gives the next reader something to diff against.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -333,6 +333,49 @@ public function test_tag_labels_rest_field_empty_returns_false() {
 		self::assertFalse( Newspack_Blocks_API::newspack_blocks_get_tag_labels( [ 'id' => $post_id ] ) );
 	}
 
+	/**
+	 * The block server render must not emit the `cat-links` class on tag
+	 * labels: per-section `.cat-links a` styling must never recolor them
+	 * (NPPM-3049, the block-side counterpart of the NPPM-3048 theme fix).
+	 */
+	public function test_display_tag_labels_renders_without_cat_links() {
+		if ( ! property_exists( '\Newspack\Tag_Labels', 'stub_labels' ) ) {
```
adekbadek: **Nit:** The skip guard is inert on both new tests — Neither new test touches `$stub_labels`; both pass labels in directly, and the real `\Newspack\Tag_Labels::generate_html()` produces identical markup, so both assertions would hold against the real class. The guard therefore only suppresses these two tests in the one environment where they would exercise production code. The REST-field tests above genuinely need it, since they set `$stub_labels`.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
```diff
@@ -333,6 +333,49 @@ public function test_tag_labels_rest_field_empty_returns_false() {
...
+			$this->markTestSkipped( 'Real \Newspack\Tag_Labels present; stub-based contract test skipped.' );
+		}
+
+		ob_start();
+		Newspack_Blocks::display_tag_labels( [] );
+		self::assertSame( '', ob_get_clean(), 'Empty labels must render nothing, not an empty wrapper.' );
+
+		ob_start();
+		Newspack_Blocks::display_tag_labels( null );
```
adekbadek, line 569: **Nit:** The `null` and `[]` cases re-prove the same early return — `display_tag_labels( null )` and `display_tag_labels( [] )` both reach the same `empty( $labels )` early return in `generate_html()`, so the second assertion re-proves the first. The case itself earns its place — it bites on the new `'' !== $labels_html` guard, without which the empty case echoes a bare `' '` and `assertSame( '', … )` goes red — just not twice.

adekbadek (changes requested):
Two blockers ⚠️

plugins/newspack-blocks/includes/class-newspack-blocks.php
adekbadek: Addressed. `carousel/view.scss:270` now shares `.cat-links, .tag-labels` for the whole chip rule, `carousel/edit.js:416` renders `className="tag-labels"`, and the helper is back on `Tag_Labels::display()` so there is no per-call-site class list to keep in step. Editor preview and front end agree again.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: Addressed. All six `.cat-links` groups in the block theme are paired: `_content-loop.scss` at 14, 65, 72, 164 and 171, and `_post-carousel.scss:32`. That includes the `color: inherit` group over the image overlay, which was the part that would have dropped the labels to link blue on a dark background. The theme still has no `.tag-label.flag` chip rule, but that was true before this PR too, so parity is restored rather than newly missing.

plugins/newspack-blocks/includes/class-newspack-blocks.php
adekbadek: Addressed, and better than what I suggested. Moving the class-list decision into `Tag_Labels::display()` keeps blocks on the three-argument entry point, keeps the `wp_kses_post( … . ' ' )` escaping in one place, and makes the default apply to the theme's `newspack_display_tag_labels()` shim at the same time. `display_tag_labels()` is now a four-line pass-through with nothing to drift.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: Addressed. `.cat-links, .tag-labels` share the rule in `homepage-articles/view.scss:477`, `carousel/view.scss:270`, `editor.scss:42`, and in both block-theme partials, so the two selectors move together by construction. `&.sponsor-label` stays inside the shared block, which is harmless since tag labels never carry that class.

plugins/newspack-blocks/tests/class-newspack-tag-labels-stub.php
adekbadek: Addressed via the second option: the docblocks now name `\Newspack\Tag_Labels::generate_html()` as what the stub tracks. I diffed the stub against `plugins/newspack-plugin/includes/tag-labels/class-tag-labels.php` at this head and the two bodies match, including the `<!-- .tag-labels -->` trailer and the `span`/`div` allowlist.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: Still open, and the new commits spread the shape. Six comments narrate the change instead of stating the invariant:

- `homepage-articles/view.scss:471` — "Tag labels used to carry `cat-links` and borrow its styling"
- `carousel/view.scss:266` — "Tag labels no longer carry `cat-links` (they used to borrow…)"
- `class-tag-labels.php:147` — "Tag labels used to borrow that class"
- `newspack-tag-labels.scss:10` — "the base `.cat-links` rule the wrapper matched before NPPM-3048"
- `newspack-tag-labels.scss:25` — "tag labels no longer carry the `cat-links` class"
- `newspack-tag-labels.scss:106` — "Selector updated with the markup: labels no longer carry `cat-links`"

On the merged branch there is no "used to" and no "no longer" — the reader has never seen the old markup, and `AGENTS.md` puts that history in git. For `view.scss:471` the invariant is roughly: the wrapper carries only `tag-labels`, so a publisher's `.cat-links a` rule cannot reach it (NPPM-3049), and sharing the selector keeps the two in step. `newspack-tag-labels.scss:106` needs no comment at all once the selector reads `.tag-labels`.

Separately, `newspack-tag-labels.scss:58` is now factually wrong: it describes the category as "immediately followed by the tag-labels `.cat-links.tag-labels`", markup this PR removes. The `.cat-links + .tag-labels` selectors below it still work, so the comment needs updating, not the CSS.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
adekbadek: Still open. Both new tests pass labels in directly and never touch `$stub_labels`, and the stub's `display()` now produces byte-identical markup to the real `\Newspack\Tag_Labels::display()` — I diffed them at this head. So the guard only suppresses these two tests in the one environment where they would exercise production code, which is the environment where the assertion is worth most. The REST-field tests above still need it.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
adekbadek: Still open. `display_tag_labels( null )` and `display_tag_labels( [] )` both land on the same `empty( $labels )` early return in `display()`, so the second `assertSame` re-proves the first. The case earns its place — it is what catches a regression that echoes a bare `' '` — but one call does that.

adekbadek (changes requested):
One blocker; three nit threads still open.

## Blocker: the five child themes style `.cat-links` and none were paired

Moving the class removal into `Tag_Labels::display()` widened it to every caller, including `newspack_display_tag_labels()` in the classic theme — the path the five style variations render through. Each styles `.cat-links` in its own `sass/style.scss`, and none were touched, so the wrapper falls back to the parent rule this PR adds (`newspack-tag-labels.scss:12`, `font-size: var(--newspack-theme-font-size-xs)` = 14px, `margin: 0 0 0.75rem`).

| Theme | Declaration the wrapper loses | Effect on a tag label |
| --- | --- | --- |
| joseph | `style.scss:104` `font-size: base` (20px), `margin-bottom: 1rem`; `:160` `.single .cat-links { display: flex; font-size: base }` | 20px → 14px; padding is em-relative, so the chip shrinks about 30%. |
| scott | `style.scss:11` `font-size: sm` (16px) and a `::before` square; `:37` `font-size: sm` | 16px → 14px, and the accent square stops rendering before the label row. |
| sacha | `style.scss:68` `text-align: center`, `margin-bottom: 1rem` | Labels left-align under a centred category. |
| katharine | `style.scss:107` `margin: 0 0 1.5rem`; `:125` `0.5rem` on blog/search/archive | 1.5rem → 0.75rem and 0.5rem → 0.75rem. |
| nelson | `style.scss:284` `.single .cat-links { margin-bottom: 1rem; tablet 2.5rem }` | Single-post gap 2.5rem → 0.75rem above the headline. |

joseph and sacha are the visible ones. The `:not(.tag-labels)` guards at `inc/jetpack.php:44` and `_archives.scss:149` are dead now and can go with it.

The parent has its own gap: `style-default.scss:110` zeroes `margin` on `.wpnbha, .wpnbpc { .cat-links a }` and `:124` does the same for the carousel, neither paired. So the new `.tag-labels a { margin: 0 0.25rem 0.25rem 0 }` adds 4px right and bottom per chip inside Homepage Posts and the Carousel — the two blocks this ticket is about.

## Cleared

Carousel and block-theme blockers fixed, threads resolved; routing back through `display()` removes the five-positional-argument concern. Test stub matches the real class at this head.

## Still open

Comment tense (six places, and `newspack-tag-labels.scss:58` describes markup this PR removes), the inert skip guard, the duplicated empty-labels assertion.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: Still open at `2623d09`, and the newer commits added more of the same shape. `AGENTS.md`: comments must describe how things ARE, never how they USED to be.

- `plugins/newspack-blocks/src/blocks/homepage-articles/view.scss:469` - "Tag labels used to carry `cat-links`..."
- `plugins/newspack-blocks/src/blocks/carousel/view.scss:266` - "no longer carry ... they used to borrow"
- `plugins/newspack-plugin/includes/tag-labels/class-tag-labels.php:147` - "Tag labels used to borrow that class"
- `plugins/newspack-plugin/tests/unit-tests/tag-labels.php:32` - same sentence in the docblock
- `themes/newspack-theme/.../sass/plugins/newspack-tag-labels.scss:10` - "the wrapper matched before NPPM-3048"
- `themes/newspack-theme/.../sass/plugins/newspack-tag-labels.scss:25` - "tag labels no longer carry the `cat-links` class"
- `themes/newspack-theme/.../sass/plugins/newspack-tag-labels.scss:106` - "Selector updated with the markup"

Two more in that last file are stale rather than just past-tense - worth a second look while you are in there:

- `:44` "Use `.tag-label.flag` so this beats `.wpnbha .cat-links a` / `.wpnbpc .cat-links a` on specificity" - those selectors cannot reach a tag label any more, so the stated reason for the compound selector is gone.
- `:58-63` describes the wrapper as `.cat-links.tag-labels` in every context, which is the markup this PR removes.

adekbadek (changes requested):
The child-theme blocker is unchanged at `2623d09` and the three nit threads are still open. The two new commits are the right call on the classic theme, where chips do carry their own background and colour, but the block theme has no chip styling for them to keep.

## Blocker: the five child themes still have no `.tag-labels` rule

`newspack_display_tag_labels()` routes through `Tag_Labels::display()` (`inc/newspack-tag-labels.php:58`), so dropping `cat-links` there lands on all six classic variations, not just the parent. `492396ea` gave the parent parity rules; `grep -rn "tag-labels" themes/newspack-theme/newspack-{joseph,scott,sacha,katharine,nelson}/sass/` is still empty, while each of the five styles `.cat-links` in its own `style.scss`. The wrapper therefore falls back to `newspack-tag-labels.scss:12-13` (`font-size: xs`, `margin: 0 0 0.75rem`) everywhere.

| Theme | Declaration the wrapper loses | Effect on a tag label |
| --- | --- | --- |
| joseph | `style.scss:104` `font-size: base` (20px), `margin-bottom: 1rem` | 20px to `0.7em`; chip padding is em-relative, so the chip itself shrinks with it. |
| nelson | `style.scss:262` `font-weight: bold`; `:284` `.single .cat-links` `margin-bottom: 1rem`, 2.5rem on tablet | Chips render at normal weight; single-post gap above the headline drops to 0.75rem. |
| scott | `style.scss:11` `font-size: sm` (0.8em) and a `::before` accent square | 0.8em to 0.7em. Losing the square is probably the fix working as intended. |
| katharine | `style.scss:107` `margin: 0 0 1.5rem`; `:125` 0.5rem on blog/search/archive | Both become 0.75rem. |
| sacha | `style.scss:68` `margin-bottom: 1rem` | 1rem to 0.75rem. |

joseph and nelson are the ones a publisher would notice. The parity rules in `492396ea` say sizing and spacing are worth preserving; the same reasoning applies one level down. Pairing `.tag-labels` into the five sizing and spacing rules (not the colour ones, which are the capture this PR exists to break) is the smallest fix. If the shift is deliberate on some of them, a screenshot of joseph and nelson would settle it.

The `:not(.tag-labels)` guards at `inc/jetpack.php:44` and `sass/site/primary/_archives.scss:149` are dead now and can go with the rest.

## The block theme has no chip styling to fall back on

`2623d09` drops `.tag-labels` from both `a:hover` groups in `_content-loop.scss` (`:72`, `:170`) on the premise that the chips keep their own colours. Nothing in `themes/newspack-block-theme/` defines `.tag-label` or `.flag`; grepping for `tag-label` there returns only the four wrapper selectors this PR touches. So on a `has-background` or overlay card the category link still hovers to `color-mix(currentcolor 80%)` / `rgba(white, 0.85)` while the tag-label link falls back to the generic `--wp--preset--color--contrast-3` at `:28`, which is not picked for a dark card. `c49ad6ad` is fine for the same reason inverted, since the classic theme does define `.tag-label.flag`.

## Also still open

Parent theme: `style-default.scss:112` and `:126` zero `margin` and `padding` on `.wpnbha, .wpnbpc { .cat-links a }` and on the carousel, and neither is paired, so `newspack-tag-labels.scss:30` now adds 4px right and bottom per chip inside Homepage Posts and the Carousel.

Three nit threads have no reply: comment tense (now seven places, inventory in the thread), the inert skip guard, and the duplicated empty-labels assertion.

jason10lee:
@adekbadek: you were right about the margin. It belongs to the Default style's category-chip treatment rather than to chips generally, so we moved it there instead of pairing it—the gap, the archive treatment and the block zeroing now sit in `style-default.scss` beside the `.cat-links` rules they mirror.

Verified in-browser against the parent theme and each of the five style variations, across single posts, Homepage Posts, the Carousel and archives.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
jason10lee: All rewritten to state the invariant. Two needed more than rewording: `:58` now reads `.tag-labels`, and `:106` is gone—the selector says it itself.

`:44` I kept, with a different reason. The `.wpnbha .cat-links a` justification is dead as you say, but `.tag-labels a` sets `color: #fff` a few lines above, and the two classes in `.tag-label.flag` are what put a chip's own color above it. The comment says that now.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
jason10lee: Removed from both. The two REST-field tests above keep theirs, since they do set `$stub_labels`.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
jason10lee: Down to one call. Kept `null` as the documented default, and the docblock now says why one covers both.

jason10lee:
Same conclusion here. Nothing in this theme defines `.tag-label`, so a label is plain text like the category beside it and wants the same hover; both groups in `_content-loop.scss` name `.tag-labels` again.

Verified against the three card variants that change the hover: overlay, `has-background`, and `has-text-color`. In each, the rule that wins for a tag label is the one that wins for the category.

jason10lee:
These I'd like to keep, with a comment on each instead.

You're right that our own markup can't reach them. `display()` was the only place that combined the classes and it no longer does; seven rendered URLs carry zero elements with both, and planting one confirms the check fires.

The outer class list is caller-supplied, though: `newspack_generate_tag_labels()` takes it as a third parameter, and calling it with both still returns `<span class="cat-links tag-labels">`. A child theme doing that would lose its tag labels on every archive, and infinite scroll would read them as the post's categories. Both guards now carry a line saying that, since what was missing was the reason rather than the guard.

jason10lee:
@adekbadek: ready for another look. Joseph and nelson, before and after:

joseph before: tag-label chips smaller than the category beside them

joseph after: tag-label chips match the category size

nelson before: category and tag labels

nelson after: category and tag labels

Nelson is the pair where nothing moves. The weight was already bold from the parent typography rule, and the single-post margin never binds, since the category shares the row and its margin sets the gap. We paired it to match the other four, not to fix a visible break.

Two things came out of checking the colour instruction. Katharine's `.wpnbha` `:hover` and `:visited` arms outranked the chip rule, and the chip kept its colour only because the customizer block ties them and loads later; that rule is now split so `.tag-labels` takes the `::before` half. And your katharine margin prediction was still live with the pairing in place, because the partial loads after the theme at the same weight and won it back. `:where()` on the two declarations that mirror `.cat-links` fixes all three variations that pair at one class: label now matches category at 20px, 30px and 20px in joseph, katharine and sacha.

The carousel is covered, though not by the rule you named. `.wpnbpc` is in the same zeroing rule as `.wpnbha`, which we did pair, so on the env both the category anchor and the label anchor compute `margin: 0` inside a carousel. The `.wp-block-newspack-blocks-carousel` rule zeroes only `background` and `padding`.

plugins/newspack-blocks/src/blocks/homepage-articles/view.scss
adekbadek: Confirmed — every comment the PR adds states the invariant.

Keeping `:44` with the new reason is right: `.tag-labels a` sets `color: #fff` at `(0,2,0)` and `.tag-label.flag` is `(0,3,0)`, so the two classes are what put a chip's own colour above it. The old justification would have been misleading.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
adekbadek: Confirmed. Both new tests run unconditionally now, and the two REST-field tests above keep their guard, which is where it belongs since they are the ones that set `$stub_labels`.

plugins/newspack-blocks/tests/test-homepage-posts-block.php
adekbadek: Confirmed. One call, and the docblock now carries the reason `null` covers `[]` — which is the part that was worth writing down. The assertion still bites on the `'' !== $labels_html` guard, so nothing was lost with the second call.

adekbadek (approved):
The child-theme blocker is closed: all five variations pair `.tag-labels` into their sizing and spacing rules while leaving the colour rules alone, and `:where(.tag-labels)` in the plugin partial gives the variations back the cascade they expect, since `newspack-tag-labels-style` declares `newspack-style` as a dependency and so loads last. Block-theme hover parity is restored on both `a:hover` groups, and the chip gap belongs in `style-default.scss` — the one style where a category has a chip to sit beside — with the block contexts zeroing it.

Keeping the `:not(.tag-labels)` guards in `inc/jetpack.php` and `_archives.scss` is the better call than my suggestion to drop them: `newspack_generate_tag_labels()` takes `$outer_classes`, so a child theme can put both classes on one element.

PR #737, opened 2026-07-27
prototype: make Audience Management an explicit prerequisite for Access Control

thomasguillot:
**Prototype, not an implementation.** Do not merge. This exists so the design can be seen running, and so the code can be lifted into a proper implementation.

Linear: [NPPD-1846](https://linear.app/a8c/issue/NPPD-1846/make-audience-management-dependency-explicit-for-access-control)

## The problem

Access Control appears whenever `NEWSPACK_CONTENT_GATES` is defined, and never checks Reader Activation. The ticket describes the symptom as broken UI in `/my-account`, but that undersells it. With Audience Management off, all of these bail out:

| What | Where |
| --- | --- |
| Reader registration hooks | `class-reader-registration.php:28` |
| Magic links (passwordless login) | `class-magic-link.php`, 7 guards |
| Newspack My Account shell | `class-my-account.php:116` |
| Auth / account emails | `class-reader-activation-emails.php:39` |
| Reader session hydration | `class-session-hydration.php:102` |
| ESP contact sync | `sync/class-sync.php:76` |
| `has_reader_activation` for blocks | `class-blocks.php:71` |

Meanwhile `class-content-gate.php` has no Reader Activation dependency at all, so gates happily restrict content. A registration wall built without Audience Management locks readers out and gives them no way in.

## What this does

Blocks the Access Control screen when Audience Management is off, replacing it with a prerequisite state that links into the Audience Management setup flow.

The submenu item deliberately stays visible. Hiding it would make the dependency invisible again, which is the opposite of what the ticket asks for.

## Why one screen is enough

Gate creation is the only door. The other candidate entry points are all downstream of a gate already existing:

- **Editor sidebar panel** returns early at `class-content-gate.php:474` when `count( self::get_gates() ) === 0`.
- **Gate layout editor** is reached through a nonce'd `newspack_edit_gate_layout` action from the wizard, so it is not a standalone entry point.
- **Newspack Dashboard** does not link to Access Control.

## Deliberately not included

1. **REST route guard.** `register_api_endpoints()` guards on `is_feature_enabled()` only. The same Audience Management check belongs beside it, otherwise a stale tab can still POST a gate into existence. The design assumes this exists.
2. **Orphaned gates.** Gates published while Audience Management was off keep restricting content, and once this ships nobody can reach them to unpublish. Needs its own decision, likely a product call rather than a design one.
3. **The help doc.** [help.newspack.com/access-control](https://help.newspack.com/access-control/) still does not list Audience Management as a dependency, which was the original question that started the thread. Not a code change, still unowned.

## Verified

Both directions checked on a local site: Audience Management off shows the prerequisite state with no "Add Content Gate" action, Audience Management on still renders the gate list with its header action intact. ESLint and PHPCS clean.

### DIFF
--- a/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
+++ b/plugins/newspack-plugin/includes/wizards/audience/class-audience-content-gates.php
@@ -95,12 +95,18 @@ class Audience_Content_Gates extends Wizard {
 			'newspack-wizards',
 			'newspackAudienceContentGates',
 			[
-				'api'                     => '/' . NEWSPACK_API_NAMESPACE . '/wizard/' . $this->slug,
-				'available_access_rules'  => Access_Rules::get_access_rules(),
-				'available_content_rules' => Content_Rules::get_content_rules(),
-				'edit_gate_layout_url'    => Content_Gate::get_edit_gate_layout_url(),
-				'presave_checks_enabled'  => Content_Gate::get_presave_checks_enabled(),
-				'default_gate_status'     => Content_Gate::get_default_new_gate_status(),
+				'api'                         => '/' . NEWSPACK_API_NAMESPACE . '/wizard/' . $this->slug,
+				'available_access_rules'      => Access_Rules::get_access_rules(),
+				'available_content_rules'     => Content_Rules::get_content_rules(),
+				'edit_gate_layout_url'        => Content_Gate::get_edit_gate_layout_url(),
+				'presave_checks_enabled'      => Content_Gate::get_presave_checks_enabled(),
+				'default_gate_status'         => Content_Gate::get_default_new_gate_status(),
+				// NPPD-1846: Audience Management is a prerequisite for Access
+				// Control. Without it there is no reader registration, magic
+				// link login, account email or My Account page for a gate to
+				// hand off to.
+				'audience_management_enabled' => Reader_Activation::is_enabled(),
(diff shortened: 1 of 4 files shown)


adekbadek:
Superseded by #765, which implements this design on both gate-editing surfaces and closes the three gaps this prototype deliberately left open:

- **REST route guard** — gate creation and duplication now go through a shared `api_permissions_check_audience_management` on both wizards. The guard ended up at the wizard *router* rather than in the gate-list view, because blocking only the list left `#/edit/new/all` reachable by bookmark or browser history with a working Save.
- **Premium Newsletters** — included, since the dependency there is real and harder: `filter_subscription_lists()` is hooked unconditionally, so restricted lists vanish from every signup form regardless of Audience Management, while the `reader_verified` event that would grant access only fires when it is on.
- **Orphaned gates** — resolved by inverting the problem instead of handling it: Audience Management can no longer be switched off while any gate exists. That yields the invariant *Audience Management off ⇒ no gates exist*, which is what makes replacing the screens safe.

Worth recording here since this PR's description states the opposite: **"nothing in the admin turns Audience Management off once it is on" turned out not to be true** — Audience → Configuration → Reader Revenue Platform → *Change* → toggle calls `saveConfig( { enabled: false } )` behind a confirm dialog. That is what made the orphaned-gate state reachable on sites already live on Access Control, and why the disable guard is load-bearing rather than a nicety.

Thanks for the prototype — the layout, copy and empty-state pattern were lifted essentially as-is.

PR #739, opened 2026-07-27
fix(perfmatters): exclude Complianz CSS from RUCSS (NPPM-3052)

dkoo:
### All Submissions:

Newspack forces Perfmatters' "Remove Unused CSS" (RUCSS) feature on by default, but the default exclusion list never covered Complianz. Perfmatters delays every non-excluded stylesheet (the `href` is stripped and only restored on user interaction or a timeout) and generates its used-CSS from a server-side HTML snapshot, which misses the JS-injected Complianz consent banner — so after Complianz 7.5.0 changed its markup/CSS, consent banners rendered unstyled until the delayed stylesheet was restored. Support has been fixing affected sites one at a time by manually adding a Perfmatters stylesheet exclusion; this PR ships that fix in the defaults so it applies platform-wide.

Two entries are added to `unused_css_excluded_stylesheets()`, which feeds both the forced-defaults merge and the `perfmatters_rucss_excluded_stylesheets` backstop filter:

- `plugins/complianz-gdpr` — Complianz plugin CSS; Perfmatters matches exclusions with a case-insensitive substring check, so this also covers the premium plugin directory (`complianz-gdpr-premium`).
- `uploads/complianz` — Complianz's generated cookie-banner CSS (e.g. `uploads/complianz/css/banner-*.css`), which the manual per-site workaround pattern did not cover.

Sites where the manual workaround was already applied are unaffected: the defaults merge runs through `array_unique()`, so no duplicates are introduced.

Closes NPPM-3052.

1. On a site with Perfmatters active (Newspack defaults applied) and Complianz (free or premium) active, load the frontend in a private window and inspect the Complianz stylesheets (`cmplz-general-css`, `cmplz-banner-*-css`): without this change they are RUCSS-delayed (`href` replaced by `data-pmdelayedstyle`), leaving the consent banner unstyled until interaction; with this change they load normally and the banner is styled immediately.
2. Confirm the exclusions surface in Perfmatters → CSS → Excluded Stylesheets (`plugins/complianz-gdpr`, `uploads/complianz`) alongside the existing Newspack defaults.
3. Run the new tests: `vendor/bin/phpunit --group perfmatters` in `plugins/newspack-plugin` — 4 tests cover the defaults path, the backstop filter (including its `add_filter` wiring), dedup when the entry was already saved manually, and the `NEWSPACK_IGNORE_PERFMATTERS_DEFAULTS` escape hatch.
4. Full `newspack-plugin` PHPUnit suite passes (1996 tests) and PHPCS is clean on both changed files.

### Other information:


dkoo:
Failing CI job is unrelated to this changeset—fixed in #740.

dkoo:
Low-risk change affecting only sites using Complianz.

PR #741, opened 2026-07-27
chore: address vipwpcs 3.1.0 posts_per_page sniff errors (NPPM-3055)

dkoo:
### All Submissions:

`vipwpcs 3.1.0` added a check that flags `posts_per_page`/`numberposts` set to `-1` as an error ([Automattic/VIP-Coding-Standards#882](https://github.com/Automattic/VIP-Coding-Standards/pull/882)), which turned CI red for every PR that lints an affected plugin (`release` was unblocked by pinning to `~3.0.1` in #740; `main` is unpinned and still red). This addresses all **66 occurrences** across the workspace — the audit and per-bucket decisions are on NPPM-3055 — so `main` lints clean against the current standards, and the `release` pin can be reverted once this is promoted:

- **Test files (26) and WP-CLI commands (14):** dismissed via `phpcs:ignore`/file-level `phpcs:disable` — not production code, and CLI commands are intentionally unbounded operator tooling. Files with three or more occurrences (two newsletters REST test files, the webhooks tests, the teams-migration CLI) use a file-level disable; the rest are inline.
- **Runtime queries against bounded, config-scale CPTs (25):** inline `phpcs:ignore` with a one-line justification each (gates, prompts, ad units, newsletter layouts, subscription lists, network nodes, email templates, per-post corrections, meta-flagged product sets, budget-scoped stories). One line extends an existing ignore: the old `WordPress.WP.PostsPerPage` code does not suppress the new sniff.
- **No behavior changes.** An earlier revision capped the Campaigns donor-landing-page query at 1000; review showed that list doubles as the server-side save allow-list, so a cap made pages outside the window silently unsaveable. It is back to `-1` with a comment explaining why bounding it is not a cost-only change. Moving that setting to an autocomplete field — which removes the enumeration, the allow-list and the query together — is a separate follow-up, as are optional caps on `retrieve_active_popups()` and the two Story Budget typeahead queries.
- **Unpin:** reverts the interim `~3.0.1` pin from #740 (which reached `main` via the post-release branch sync) back to `^3.0`, so the unpin lands alongside the fixes it depends on. Verified end-state: a fresh `composer install` on this branch resolves vipwpcs 3.1.0 and the full lint scope passes.

The new standards also surface warnings, which fail CI the same as errors. Those are addressed too:

- **WCS renewal redirect (`AlwaysReturnInFilter.TerminatingInsteadOfReturn`):** fixed in code rather than dismissed. `pre_get_posts` is a core **action**, but the redirect callback was registered with `add_filter`, which is what put it in the sniff's scope (and had already accumulated two suppressions for related messages). Registering with `add_action` — the same registry call, zero behavior change — resolves the new warning at the root and allowed **removing** the two pre-existing ignores on the callback.
- **Starter content (3) and a listings test (1) `FetchingRemoteData.FileGetContentsUnknown`:** dismissed with justifications — all four read bundled plugin assets from the local filesystem (setup-time starter content and a test fixture), not remote URLs.

The only findings left anywhere in the workspace lint scope are two `DuplicateClassName` warnings in popups test mocks that fire solely in whole-workspace scans (the mock shadows a newsletters class, and per-package CI never scans both plugins together) — invisible to CI, left alone.

Closes NPPM-3055. Promotion to `release` carries the unpin there — no separate revert needed.

1. CI on this PR is the primary test: with the pin reverted, the PHPCS jobs for all six touched plugins re-resolve the latest standards (vipwpcs 3.1.0) and must pass.
2. Locally: `composer install` at the workspace root on this branch (resolves vipwpcs 3.1.0), then `vendor/bin/phpcs --standard=phpcs.xml --sniffs=WordPressVIPMinimum.Performance.NoPaging` — zero errors (was 66).
3. Per-package, exactly as CI invokes it: `composer phpcs -- "$PWD/plugins/<name>"` for each of the seven touched plugins — all exit 0. A whole-workspace `vendor/bin/phpcs --standard=phpcs.xml` reports only 2 `DuplicateClassName` warnings in popups test mocks, which fire solely in cross-plugin scans and are invisible to per-package CI.
4. Campaigns → Settings → Donor landing page (behavior unchanged by this PR): the field is a text input, and saving a valid published top-level page ID still succeeds — including the oldest pages on a site with many of them.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? (n/a — lint annotations + a query cap; existing suites pass: newspack-popups 253 tests OK, newspack-plugin full suite OK)

### DIFF
--- a/plugins/newspack-plugin/includes/cli/class-teams-migration.php
+++ b/plugins/newspack-plugin/includes/cli/class-teams-migration.php
@@ -4,8 +4,10 @@
  * Newspack Access Control group subscriptions, plus a backfill for group managers.
  *
  * Ported from the standalone `migrate-memberships` drop-in so the tooling ships
- * with the plugin (CLI classes load only under WP_CLI, so there is no web-request
- * cost) and writes through the real Group_Subscription data layer.
+ * with the plugin and writes through the real Group_Subscription data layer. Note
+ * that CLI\Initializer::init() includes this file on every request; only the
+ * WP_CLI::add_command() registration is gated behind WP_CLI, so nothing here should
+ * assume a CLI-only execution context.
  *
  * Member and manager writes go through update_members()/add_manager(), which fire
  * only the group cache-reset hook — no data events, ESP sync, or emails. WooCommerce
--- a/plugins/newspack-plugin/includes/cli/class-teams-migration.php
+++ b/plugins/newspack-plugin/includes/cli/class-teams-migration.php
@@ -17,6 +17,8 @@
  * @package Newspack
  */
 
+// phpcs:disable WordPressVIPMinimum.Performance.NoPaging -- Operator-run CLI command; unbounded by design.
+
 namespace Newspack\CLI;
 
 use Newspack\Group_Subscription;
--- a/plugins/newspack-popups/includes/class-newspack-popups-model.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-model.php
@@ -110,7 +110,7 @@ final class Newspack_Popups_Model {
 				[
 					'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 					'post_status'    => 'publish',
-					'posts_per_page' => -1,
+					'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Prompt CPT; config-scale.
 				]
 			),
 			true
--- a/plugins/newspack-popups/includes/class-newspack-popups-model.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-model.php
@@ -110,7 +110,7 @@ final class Newspack_Popups_Model {
 				[
 					'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 					'post_status'    => 'publish',
-					'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Prompt CPT; config-scale.
+					'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Prompt CPT; accepted cost. Unlike retrieve_popups() above this is frontend-reachable, so a cap would change which prompts render and is left to a follow-up.
 				]
 			),
 			true
--- a/plugins/newspack-popups/includes/class-newspack-popups-settings.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-settings.php
@@ -246,7 +246,10 @@ class Newspack_Popups_Settings {
 				'post_type'      => 'page',
 				'post_status'    => 'publish',
 				'post_parent'    => 0,
-				'posts_per_page' => 1000, // phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page -- Deliberate cap to keep the query bounded on sites with very many top-level pages.
+				// This list doubles as the save allow-list in update_setting(), so a cap would make
+				// pages outside the window silently unsaveable. Left unbounded until the setting
+				// moves to an autocomplete field and stops enumerating pages.
+				'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- See above; bounding this changes behavior, not just cost.
 			]
 		);
 		// Remove the query filter so we don't unintentionally affect other queries.
--- a/plugins/newspack-popups/includes/class-newspack-popups-settings.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-settings.php
@@ -246,7 +246,7 @@ class Newspack_Popups_Settings {
 				'post_type'      => 'page',
 				'post_status'    => 'publish',
 				'post_parent'    => 0,
-				'posts_per_page' => -1,
+				'posts_per_page' => 1000, // phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page -- Deliberate cap to keep the query bounded on sites with very many top-level pages.
 			]
 		);
 		// Remove the query filter so we don't unintentionally affect other queries.
--- a/plugins/newspack-story-budget/includes/class-api.php
+++ b/plugins/newspack-story-budget/includes/class-api.php
@@ -578,7 +578,7 @@ class API {
 		$query_args = [
 			'story_budget_search' => true,
 			'fields'              => 'ids',
-			'posts_per_page'      => -1,
+			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Tax-scoped to active budgets; bounded editorial set.
 			's'                   => $request->get_param( 's' ) ?? '',
 		];
 
@@ -932,7 +932,7 @@ class API {
 		$query_args = [
 			'story_budget_search' => true,
 			'fields'              => 'ids',
-			'posts_per_page'      => -1,
+			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Tax-scoped to active budgets; bounded editorial set.
 			's'                   => $request->get_param( 's' ) ?? '',
 		];
 
--- a/plugins/newspack-story-budget/includes/class-api.php
+++ b/plugins/newspack-story-budget/includes/class-api.php
@@ -578,7 +578,7 @@ class API {
 		$query_args = [
 			'story_budget_search' => true,
 			'fields'              => 'ids',
-			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Tax-scoped to active budgets; bounded editorial set.
+			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Accepted cost on an authenticated editorial screen: the story_budget_search shape (postmeta JOIN + leading-wildcard LIKE) is the expense here, not the row count. Capping is a behavior change left to a follow-up.
 			's'                   => $request->get_param( 's' ) ?? '',
 		];
 
@@ -932,7 +932,7 @@ class API {
 		$query_args = [
 			'story_budget_search' => true,
 			'fields'              => 'ids',
-			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Tax-scoped to active budgets; bounded editorial set.
+			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Accepted cost on an authenticated editorial screen: the story_budget_search shape (postmeta JOIN + leading-wildcard LIKE) is the expense here, not the row count. Capping is a behavior change left to a follow-up.
 			's'                   => $request->get_param( 's' ) ?? '',
 		];
 
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
 		"php": ">=8.0"
 	},
 	"require-dev": {
-		"automattic/vipwpcs": "~3.0.1",
+		"automattic/vipwpcs": "^3.0",
 		"wp-coding-standards/wpcs": "^3.0",
 		"phpcompatibility/phpcompatibility-wp": "^2.1",
 		"dealerdirect/phpcodesniffer-composer-installer": "^1.0",
--- a/plugins/newspack-ads/includes/providers/gam/class-gam-model.php
+++ b/plugins/newspack-ads/includes/providers/gam/class-gam-model.php
@@ -418,7 +418,7 @@ final class GAM_Model {
 		$query           = new \WP_Query(
 			[
 				'post_type'      => self::$custom_post_type,
-				'posts_per_page' => -1,
+				'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Legacy ad-unit CPT; config-scale.
 				'post_status'    => [ 'publish' ],
 			]
 		);
--- a/plugins/newspack-listings/tests/test-blocks.php
+++ b/plugins/newspack-listings/tests/test-blocks.php
@@ -23,7 +23,7 @@ class BlocksTest extends WP_UnitTestCase {
 
 		// Set default block options for Curated List.
 		$block_json = json_decode(
-			file_get_contents( NEWSPACK_LISTINGS_PLUGIN_FILE . '/src/blocks/curated-list/block.json' ), // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+			file_get_contents( NEWSPACK_LISTINGS_PLUGIN_FILE . '/src/blocks/curated-list/block.json' ), // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents, WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown -- Reads the block's bundled block.json from the local filesystem.
 			true
 		);
 		foreach ( $block_json['attributes'] as $attribute => $options ) {
--- a/plugins/newspack-network/includes/cli/backfillers/class-membership-plan-updated.php
+++ b/plugins/newspack-network/includes/cli/backfillers/class-membership-plan-updated.php
@@ -45,7 +45,7 @@ class Membership_Plan_Updated extends Abstract_Backfiller {
 			[
 				'post_type'   => Memberships_Admin::MEMBERSHIP_PLANS_CPT,
 				'post_status' => 'any',
-				'numberposts' => -1,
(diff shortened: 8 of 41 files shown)


plugins/newspack-popups/includes/class-newspack-popups-settings.php
```diff
@@ -246,7 +246,7 @@ public static function get_settings( $assoc = false, $get_segments = false ) {
 				'post_type'      => 'page',
 				'post_status'    => 'publish',
 				'post_parent'    => 0,
-				'posts_per_page' => -1,
+				'posts_per_page' => 1000, // phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page -- Deliberate cap to keep the query bounded on sites with very many top-level pages.
```
adekbadek: **Blocker:** Capping the page list makes valid Donor Landing Pages silently unsaveable — This option list is not only rendered — it is also the server-side **save allow-list**. `update_setting()` (this file, lines 144-155) builds `$accepted_values` from `$config['options']` and rejects anything not in it, so capping the query at 1000 makes every top-level page outside that window unsaveable.

The query has no `orderby`, so `WP_Query` defaults to `date DESC` and the window keeps the *newest* 1000 — the pages that fall out are the oldest, which is exactly where a long-standing "thank you for donating" page lives.

The failure is silent on the path publishers actually use. The Audience wizard's picker is an autocomplete over **all** pages (`newspack-plugin/src/wizards/audience/components/nrh-settings/index.js:79-95`), and its save path (`newspack-plugin/includes/class-nrh.php:92`) calls `update_setting()` and discards the return value, then returns the unrelated `update_option()` result — so the request succeeds and the publisher is told the settings saved while their choice was dropped.

Reproduced on an env with 1054 top-level published pages: saving the newest page is accepted; saving page ID 5 returns `WP_Error: Invalid setting value for "Donor landing page"` and leaves the stored option unchanged. Before this change (`-1`) every page was in the allow-list, so this failure mode is new.

Validating the submitted ID directly — confirm it is a published page via `get_post_type()`/`get_post_status()` — would remove the allow-list, the cap question, and the query in one move. Failing that, union the currently-saved and submitted IDs into the option list before validating.

(Related and pre-existing: the allow-list is `post_parent => 0` while the autocomplete is not, so choosing a child page already fails the same silent way today — another reason to drop the enumeration approach rather than tune its bound.)

plugins/newspack-popups/includes/class-newspack-popups-model.php
```diff
@@ -110,7 +110,7 @@ public static function retrieve_active_popups() {
 				[
 					'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 					'post_status'    => 'publish',
-					'posts_per_page' => -1,
+					'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Prompt CPT; config-scale.
```
adekbadek: **Suggestion:** Same CPT is capped 19 lines above — `retrieve_active_popups()` is annotated "Prompt CPT; config-scale", but its sibling `retrieve_popups()` at line 94 already caps that same post type at 1000 with a deliberate suppression. Same file, same CPT — one is treated as needing a bound, the other as inherently bounded.

This one is also frontend-reachable: `Newspack_Popups_Inserter::maybe_enqueue_contact_form_styles()` calls it during style enqueue, and it hydrates full prompt objects including `post_content`. Mirroring the existing 1000 cap would make the annotation self-enforcing and consistent with the line above it.

plugins/newspack-story-budget/includes/class-api.php
```diff
@@ -578,7 +578,7 @@ public static function get_stories_search( $request ) {
 		$query_args = [
 			'story_budget_search' => true,
 			'fields'              => 'ids',
-			'posts_per_page'      => -1,
+			'posts_per_page'      => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Tax-scoped to active budgets; bounded editorial set.
```
adekbadek: **Suggestion:** "Bounded editorial set" describes the row count, not the cost — The justification speaks to how many rows come back, but the expensive part here is the query *shape*. Both search callbacks set `story_budget_search => true`, which activates `Search::wp_admin_search_join/where/distinct` — a `LEFT JOIN wp_postmeta`, a leading-wildcard `meta_value LIKE '%term%'` across searchable fields, and `DISTINCT`. With `posts_per_page => -1` there is no `LIMIT` bounding that scan.

Worth weighing against the plugin's own judgement: `should_add_fields_to_wp_admin_search()` (`includes/class-search.php:36-41`) refuses to apply this same expansion on the WP admin posts screen unless `NEWSPACK_STORY_BUDGET_ENABLE_SEARCH_META` is defined, with the comment "This can slow down sites with too many posts" — and the `story_budget_search` path bypasses that guard entirely (line 43). So the codebase already treats this shape as unsafe at content scale in a *paginated* context, while these two run it unpaginated. They are reached from a 300ms-debounced, abort-controlled typeahead (`src/components/stories.js:63`), which limits the rate but not the per-request cost.

A typeahead never needs more than ~100 hits, so capping these two the way the Campaigns query was capped would retire the sniff honestly. If they stay at `-1`, the comment would be more accurate as an accepted cost on an authenticated editorial screen than as a claim that the set is bounded.

Same applies to the identical annotation at line 935.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
```diff
@@ -17,6 +17,8 @@
  * @package Newspack
  */
 
+// phpcs:disable WordPressVIPMinimum.Performance.NoPaging -- Operator-run CLI command; unbounded by design.
```
adekbadek, line 22: **Nit:** File-level disable rests on a docblock premise that isn't accurate — The file-level `phpcs:disable` is defensible here — 5 occurrences, and the queries really are only reachable as WP-CLI commands. But the stated premise echoes a docblock claim (lines 6-7) that "CLI classes load only under WP_CLI, so there is no web-request cost", whereas `CLI\Initializer::init()` `include_once`s this file unconditionally on every request; only `WP_CLI::add_command()` is gated.

The suppression is still correct, but the combination of a whole-file exemption and a comment saying the file never loads on web requests is the kind of thing that lets a future hook registration land here and inherit the exemption silently. Worth correcting the docblock while it's open.

adekbadek (changes requested):
Release risk: HIGH — the one behavioural change here makes a valid Donor Landing Page silently unsaveable on sites with more than 1000 top-level published pages, because the option list being capped is also the server-side save allow-list. The rest of the sweep is sound: bucket counts reconcile against the diff, the CLI/test suppressions are genuinely CLI- and test-only, and the `add_filter` → `add_action` swap is inert and retires two suppressions rather than adding any. Verified in an isolated env — sniff errors 0/66, per-package PHPCS clean for all seven plugins, popups 253 and plugin 2362 tests green — with the blocker reproduced against 1054 pages and a control case.

A couple of description corrections worth folding in: the full-ruleset run reports 2 warnings rather than the 7 in test step 3, and test step 4 can't be performed as written (the field is a text input, not a list of options).

- `plugins/newspack-popups/includes/class-newspack-popups-settings.php:244` — **Suggestion:** Query hydrates full post objects to read two fields — Since this query is being touched anyway: it hydrates up to 1000 full `WP_Post` objects plus their meta and term caches and runs `SQL_CALC_FOUND_ROWS`, while the consuming loop reads only `ID` and `post_title`. `get_settings()` runs on every Campaigns settings read and on every settings-key validation. Adding `'no_found_rows' => true, 'update_post_meta_cache' => false, 'update_post_term_cache' => false` would make it substantially cheaper on exactly the large sites the cap is aimed at. (Moot if the allow-list is replaced with direct validation per the blocker above, which removes the query entirely.)
- `plugins/newspack-network/includes/hub/admin/class-woo.php:63` — **Nit:** Leftover `add_filter( 'pre_get_posts', … )` the sweep missed — This is the same pattern the PR normalises elsewhere: still registered with `add_filter`, with a callback documented `@return NULL` that ends in `return null;` (three of them) purely to satisfy a filter sniff that shouldn't apply to an action hook. Converting it to `add_action` and dropping the `return null;` contortion would finish the job started in `class-renewal.php`.

Two notes that don't anchor to a diff line: `newspack-popups/includes/class-newspack-popups-settings.php:244` — while this query is open, it hydrates up to 1000 full `WP_Post` objects plus meta/term caches and runs `SQL_CALC_FOUND_ROWS` to read only `ID` and `post_title`; `no_found_rows`/`update_post_meta_cache`/`update_post_term_cache` would cut that (moot if the allow-list goes away). And `newspack-network/includes/hub/admin/class-woo.php:63` is a leftover of the very pattern this PR normalises — still `add_filter( 'pre_get_posts', … )` with a callback documented `@return NULL` that ends in `return null;` — worth converting to finish the sweep.

plugins/newspack-popups/includes/class-newspack-popups-settings.php
dkoo: Confirmed and reverted in d5edb65 — thank you, this was a real regression and the silent path made it worse than it looked.

Verified the mechanism in code: `update_setting()` builds `$accepted_values` from `$config['options']`, and `Newspack\NRH::update_settings()` discards `update_setting()`'s return value in favour of the unrelated `update_option()` result, so a rejected value reports success. One detail that reinforces your point: this setting's `type` is `'string'`, so the settings UI renders it as a `TextControl` — the `options` array is never rendered anywhere, it exists *only* to gate validation.

Rather than tune the bound or union the IDs, the query is back to `-1` with a comment explaining that the list doubles as the save allow-list, so bounding it changes behavior rather than just cost. The proper fix — moving the setting to an `AutocompleteTokenField` so it stops enumerating pages at all, which also retires the pre-existing `post_parent => 0` child-page mismatch you flagged — is going into a separate follow-up PR rather than riding along on a lint sweep.

plugins/newspack-popups/includes/class-newspack-popups-model.php
dkoo: Good catch on the inconsistency — the annotation was over-claiming, corrected in d5edb65.

I've left the query at `-1` rather than mirroring the 1000 cap, for the same reason the donor-landing-page cap is being reverted in this PR: unlike `retrieve_popups()`, this one is frontend-reachable, so a cap silently changes *which prompts render* rather than just bounding cost. That's a behavior change I'd rather not land inside a lint sweep. The comment now says so explicitly — "accepted cost… a cap would change which prompts render and is left to a follow-up" — instead of asserting the CPT is inherently bounded.

Happy to be overruled if you'd rather see the caps land together.

plugins/newspack-story-budget/includes/class-api.php
dkoo: Agreed — the justification was describing row count when the cost is the query shape. Reworded on both lines (581 and 935) in d5edb65.

Verified your read of the guard: `should_add_fields_to_wp_admin_search()` bails for the admin-screen path unless `NEWSPACK_STORY_BUDGET_ENABLE_SEARCH_META` is defined, but its final return admits `$is_story_budget_search` unconditionally — so these two do run the expansion the plugin declines to run on the paginated admin screen.

Taking your fallback rather than the cap: the annotation now reads as an accepted cost on an authenticated editorial screen, naming the postmeta JOIN and leading-wildcard LIKE as the actual expense, with the cap called out as a follow-up. Same reasoning as the other two queries here — a typeahead limit is a behavior change and belongs in a PR where it can be tested against a real content-scale site, not in a lint sweep.

plugins/newspack-plugin/includes/cli/class-teams-migration.php
dkoo: Correct, and worth fixing — the docblock was inaccurate. Confirmed: `CLI\Initializer::init()` `include_once`s this file unconditionally (line 34); only `WP_CLI::add_command()` sits behind `defined( 'WP_CLI' )` (line 43).

Corrected in d5edb65: the docblock now states that the file is included on every request and that nothing here should assume a CLI-only execution context — which is exactly the trap you described, where a future hook registration would otherwise inherit the file-level exemption silently.

dkoo:
Thanks for the thorough pass — I reproduced every finding before acting on it, and all five held up. Addressed in d5edb65.

**Blocker — capped page list makes valid Donor Landing Pages unsaveable.** Real regression, reverted. Details on the thread; the short version is that the `options` array is *only* used as the save allow-list (the setting's type is `string`, so it renders as a `TextControl`, not a select), which is what made the cap silently narrow validation. Back to `-1` with a comment explaining why bounding it isn't a cost-only change. The proper fix — an `AutocompleteTokenField` that stops enumerating pages entirely, which also clears the pre-existing `post_parent => 0` child-page mismatch you flagged — will be a separate follow-up PR rather than riding along on a lint sweep.

**Suggestion — `retrieve_active_popups()` inconsistent with its sibling.** The annotation was over-claiming; corrected. Kept at `-1` rather than mirroring the 1000 cap, because this one is frontend-reachable and a cap changes *which prompts render* — the same class of behavior change as the blocker above, so the same treatment: honest comment now, cap as a follow-up.

**Suggestion — Story Budget "bounded editorial set" describes rows, not cost.** Agreed, and your read of the guard checks out: `should_add_fields_to_wp_admin_search()` bails on the admin path without the constant, but its final return admits `$is_story_budget_search` unconditionally. Took the fallback you offered — both annotations now name the postmeta JOIN and leading-wildcard LIKE as the real expense and frame it as an accepted cost on an authenticated screen, with the typeahead cap as a follow-up.

**Nit — teams-migration docblock premise.** Fixed. Confirmed the file is included unconditionally by `CLI\Initializer::init()` with only `WP_CLI::add_command()` gated, so the docblock now says exactly that and warns against assuming a CLI-only context — the trap you described.

**Out-of-diff — `class-woo.php:63` leftover `add_filter( 'pre_get_posts', … )`.** Converted to `add_action` in the same commit, finishing the sweep.

**Out-of-diff — `no_found_rows`/cache flags on the donor-landing query.** Skipped as moot, per your own note: the query goes away entirely in the autocomplete follow-up, so it'll be tuned out of existence rather than tuned.

**Description corrections.** Both correct and both fixed. The warning count was stale — those warnings are now resolved in-tree rather than outstanding, and the body describes per-package results instead. Test step 4 has been rewritten since the field is a text input, not a list of options.

Re-verified after the changes: per-package `composer phpcs` exits 0 for all seven touched packages, popups suite 253 tests green.

The one place I've diverged from your suggestions is declining the two caps — both are publisher-visible behavior changes, and this PR's job is retiring lint debt without changing runtime behavior. Happy to fold them in if you'd rather they land together.

jason10lee:
Heads-up: Dependabot undid the `~3.0.1` pin on `automattic/vipwpcs` from #740 in #744, breaking CI. #759 restores that and also clears this PR's merge conflict. I'll merge #759 in a few hours as a tooling-only change, but please let me know if you've got any objections!

plugins/newspack-popups/includes/class-newspack-popups-model.php
adekbadek: Not overruling — the reasoning holds, and the corrected annotation is what the point was actually about. `retrieve_active_popups()` is reached from `maybe_enqueue_contact_form_styles()` on the enqueue path, so a cap really would decide which prompts get their styles/render rather than just bounding cost, and that doesn't belong in a lint sweep. The annotation now says that instead of claiming the CPT is inherently bounded, which is the honest version. Resolving.

plugins/newspack-story-budget/includes/class-api.php
adekbadek: This is the fallback I named, and both 581 and 935 now carry it — the annotation points at the query shape (postmeta JOIN + leading-wildcard LIKE) as the expense rather than at row count, and calls the cap out as a follow-up. Agreed that a typeahead limit is a behaviour change that wants a content-scale site to test against. Resolving.

adekbadek (approved):
Blocker cleared: the donor-landing-page query is back at -1 with the reason recorded inline, so `update_setting()`'s options allow-list once again covers every top-level published page — the silent save failure past 1000 pages is gone. Both annotation threads are resolved: `retrieve_active_popups()` no longer claims the prompt CPT is inherently bounded (and it is frontend-reachable via `maybe_enqueue_contact_form_styles()`, so leaving the cap to a follow-up is right), and the story-budget pair now names the query shape rather than the row count.

Two non-blocking leftovers for whenever the follow-up happens: the settings query still hydrates full `WP_Post` objects plus meta/term caches to read two fields, and `newspack-network/includes/hub/admin/class-woo.php:155` keeps its `@return NULL` docblock and three `return null;` now that it's registered as an action.

Note that PHPCS wasn't re-run for this pass — no `vendor/` in this checkout — but the new commit is comment-only apart from the cap revert (which carries its suppression) and the `add_filter` → `add_action` swap, both verified by reading.

plugins/newspack-popups/includes/class-newspack-popups-settings.php
dkoo: Follow-up: #761 implements the `AutocompleteTokenField` refactor, closing the loop on the unbounded Donor landing page setting's query.

PR #742, opened 2026-07-27
feat(subscriptions): shared base for subscriber-commerce features

adekbadek:
### All Submissions:

The shared base that **subscriber-only products** ([NPPD-1899](https://linear.app/a8c/issue/NPPD-1899), stacked in #743) and **subscriber discounts** ([NPPD-1794](https://linear.app/a8c/issue/NPPD-1794), stacked separately) both build on. Split out so the two features share one implementation of the parts they'd otherwise each write and drift on.

A subscriber-commerce rule ties store products to subscriptions — *"these products, for subscribers of these subscriptions"*. The targeting half of that shape is identical across both features.

* **`Product_Targeting`** resolves a rule's "Applies to" fields to products: specific products, a category (cascading to subcategories, since `product_cat` is hierarchical), or everything, minus exclusions. Variations match through their parent, which is what publishers pick. `get_matching_rules()` memoizes per (rule set, product) — WooCommerce asks about purchasability and price many times per product per request.
* **`Subscriber_Eligibility`** wraps `Access_Rules::has_active_subscription()` with per-request caching. Only *held* subscriptions count: one sitting in the cart does not make a reader a subscriber.
* **`Subscriber_Commerce::is_enforcement_active()`** is the single stand-down check (content-gates flag + WooCommerce present + Memberships inactive), so both features share byte-identical logic and the eventual "let Access Control take precedence over Memberships" cleanup is one change rather than two.
* **The wizard shell** renders whichever tabs PHP registered via `register_tab()`, looking each up in a front-end registry — a feature ships a tab without changing the shell. Tabs carry an explicit `order`, so the landing tab isn't decided by which hook a feature happens to register from.
* **Shared REST search** for products, product categories and subscriptions. Product results carry prices and per-variation rows, so a rule editor can preview what a reader would pay without a second round trip.

**One behaviour change worth review:** the page was registered only while Memberships was active — exactly when subscriber-commerce stands down, so the admin would have been unreachable for the features it hosts. It now also appears on any Woo site with content gating enabled. Note `is_admin_available()` (flag + Woo) is deliberately *not* gated on Memberships being inactive: a migrating publisher configures rules first and deactivates Memberships after, which is the order the migration runbook prescribes.

Its existing configuration screen becomes the first registered tab; the "Manage Subscriptions in Woo Memberships" card is now gated on Memberships actually being active, since the page renders on sites with no such screen.

Part of [NPPD-1899](https://linear.app/a8c/issue/NPPD-1899), [NPPD-1794](https://linear.app/a8c/issue/NPPD-1794).

1. On a site with WooCommerce active, `NEWSPACK_CONTENT_GATES` enabled and Woo Memberships **inactive**, open **Audience → Subscriptions**. The page renders (before this PR it was hidden without Memberships) and shows the Configuration tab. With only one tab registered there's no tab navigation — correct.
2. Confirm the "Manage Subscriptions settings in Woo Memberships" card is absent. Activate Memberships and reload: it reappears.
3. Hit the shared search endpoints as an admin and check the response shapes:
   * `/wp-json/newspack/v1/wizard/newspack-audience-subscriptions/products-search?search=x` — each product carries `id, name, parent_id, price, regular_price, sale_price, is_on_sale`; a variable product also returns its variations as their own rows.
   * `.../subscriptions-search` — subscription products only.
   * `.../product-categories-search`.
4. `n test-php --group subscriber-commerce`.

### Other information:

**Verification:** 19 unit tests for targeting and eligibility (full suite green: 2383 tests, 6698 assertions); PHPCS clean; ESLint clean; build clean; page verified rendering in a browser on a Woo site with Memberships inactive.

https://claude.ai/code/session_01LKzH1CGLgvvrusDGSURAsv

### DIFF
--- a/plugins/newspack-plugin/includes/class-wizards.php
+++ b/plugins/newspack-plugin/includes/class-wizards.php
@@ -80,7 +80,11 @@ class Wizards {
 			'newsletters'             => new Newsletters_Wizard(),
 			'premium-newsletters'     => new Premium_Newsletters_Wizard(),
 		];
-		if ( Memberships::is_active() ) {
+		// Memberships sites get the page for its subscription configuration; every
+		// Woo site with content gating gets it for the subscriber-commerce tabs.
+		// Deliberately not gated on enforcement being live: a site migrating off
+		// Memberships configures its rules first and deactivates Memberships after.
+		if ( Memberships::is_active() || Subscriber_Commerce::is_admin_available() ) {
 			self::$wizards['audience-subscriptions'] = new Audience_Subscriptions();
 		}
 		// Plans (Subscription Products) page, gated behind NEWSPACK_PLANS_UI and available
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
@@ -54,6 +54,28 @@ class Product_Targeting {
 	private static array $expanded_categories = [];
 
 	/**
+	 * Product IDs expanded with the children of any grouped product among them,
+	 * keyed by the hash of the requested IDs.
+	 *
+	 * @var array<string, int[]>
+	 */
+	private static array $expanded_grouped = [];
+
+	/**
+	 * Namespace a per-request cache key to the current site.
+	 *
+	 * Product, term and user IDs are per-site: under switch_to_blog() term 5 on
+	 * two sites would otherwise share an entry and return each other's answers.
+	 *
+	 * @param string $key The key.
+	 *
+	 * @return string
+	 */
+	private static function cache_key( string $key ): string {
+		return get_current_blog_id() . ':' . $key;
+	}
+
+	/**
 	 * Get the rules from a rule set that cover a product, memoized per request.
 	 *
 	 * Inactive rules are never returned.
@@ -69,7 +91,7 @@ class Product_Targeting {
 			return [];
 		}
 
-		$cache_key = md5( wp_json_encode( $rules ) ) . ':' . $product->get_id();
+		$cache_key = self::cache_key( md5( wp_json_encode( $rules ) ) . ':' . $product->get_id() );
 		if ( ! isset( self::$matching_rules[ $cache_key ] ) ) {
 			self::$matching_rules[ $cache_key ] = array_values(
 				array_filter(
@@ -123,6 +145,12 @@ class Product_Targeting {
 	/**
 	 * Whether a product is excluded from a rule.
 	 *
+	 * The exclusion list is expanded through grouped products first: a grouped
+	 * product's children are top-level products reporting a parent of 0, so the
+	 * parent-ID test below — which is what catches a variation of an excluded
+	 * variable product — never sees them. Without the expansion, excluding a
+	 * bundle would leave everything sold under it still covered by the rule.
+	 *
 	 * @param array $rule       The rule.
 	 * @param int   $product_id The product (or variation) ID.
 	 * @param int   $parent_id  The parent product ID, 0 for a non-variation.
@@ -130,8 +158,40 @@ class Product_Targeting {
 	 * @return bool
 	 */
 	private static function is_excluded( array $rule, int $product_id, int $parent_id ): bool {
-		$excluded = array_map( 'intval', $rule['excluded_product_ids'] ?? [] );
+		$excluded = self::expand_grouped_ids( array_map( 'intval', $rule['excluded_product_ids'] ?? [] ) );
 		return ! empty( array_intersect( array_filter( [ $product_id, $parent_id ] ), $excluded ) );
+	}
+
+	/**
+	 * Expand a list of product IDs with the children of any grouped product in it,
+	 * memoized per request.
+	 *
+	 * Deliberately applied to exclusions only, not to `products` targeting: an
+	 * exclusion that covers too little sells something the publisher meant to
+	 * withhold, whereas targeting that covers too much restricts or discounts
+	 * something they never named. Where the two directions disagree, this errs
+	 * toward the rule doing less.
+	 *
+	 * @param int[] $product_ids The product IDs.
+	 *
+	 * @return int[] The IDs, plus the children of any grouped product among them.
+	 */
+	public static function expand_grouped_ids( array $product_ids ): array {
+		if ( empty( $product_ids ) || ! function_exists( 'wc_get_product' ) ) {
+			return $product_ids;
+		}
+		$cache_key = self::cache_key( md5( wp_json_encode( $product_ids ) ) );
+		if ( ! isset( self::$expanded_grouped[ $cache_key ] ) ) {
+			$expanded = $product_ids;
+			foreach ( $product_ids as $product_id ) {
+				$product = wc_get_product( $product_id );
+				if ( $product instanceof \WC_Product && $product->is_type( 'grouped' ) ) {
+					$expanded = array_merge( $expanded, array_map( 'intval', $product->get_children() ) );
+				}
+			}
+			self::$expanded_grouped[ $cache_key ] = array_values( array_unique( $expanded ) );
+		}
+		return self::$expanded_grouped[ $cache_key ];
 	}
 
 	/**
@@ -148,7 +208,7 @@ class Product_Targeting {
 		if ( empty( $category_ids ) ) {
 			return [];
 		}
-		$cache_key = md5( wp_json_encode( $category_ids ) );
+		$cache_key = self::cache_key( md5( wp_json_encode( $category_ids ) ) );
 		if ( ! isset( self::$expanded_categories[ $cache_key ] ) ) {
 			$expanded = $category_ids;
 			foreach ( $category_ids as $category_id ) {
@@ -169,5 +229,6 @@ class Product_Targeting {
 	public static function flush_cache(): void {
 		self::$matching_rules      = [];
 		self::$expanded_categories = [];
+		self::$expanded_grouped    = [];
 	}
 }
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
@@ -43,7 +43,7 @@ class Product_Targeting {
 	 *
 	 * @var array<string, array>
 	 */
-	private static $matching_rules = [];
+	private static array $matching_rules = [];
 
 	/**
 	 * Category IDs expanded with their descendants, keyed by the hash of the
@@ -51,7 +51,7 @@ class Product_Targeting {
 	 *
 	 * @var array<string, int[]>
 	 */
-	private static $expanded_categories = [];
+	private static array $expanded_categories = [];
 
 	/**
 	 * Get the rules from a rule set that cover a product, memoized per request.
@@ -63,7 +63,7 @@ class Product_Targeting {
 	 *
 	 * @return array The matching rules, in the order given.
 	 */
-	public static function get_matching_rules( $rules, $product ) {
+	public static function get_matching_rules( array $rules, $product ): array {
 		$product = $product instanceof \WC_Product ? $product : ( function_exists( 'wc_get_product' ) ? wc_get_product( $product ) : null );
 		if ( ! $product instanceof \WC_Product || empty( $rules ) ) {
 			return [];
@@ -91,7 +91,7 @@ class Product_Targeting {
 	 *
 	 * @return bool
 	 */
-	public static function rule_covers_product( $rule, $product ) {
+	public static function rule_covers_product( array $rule, $product ): bool {
 		$product_id = (int) $product->get_id();
 		$parent_id  = (int) $product->get_parent_id();
 		$targeting  = $rule['targeting'] ?? self::TARGETING_PRODUCTS;
@@ -129,7 +129,7 @@ class Product_Targeting {
 	 *
 	 * @return bool
 	 */
-	private static function is_excluded( $rule, $product_id, $parent_id ) {
+	private static function is_excluded( array $rule, int $product_id, int $parent_id ): bool {
 		$excluded = array_map( 'intval', $rule['excluded_product_ids'] ?? [] );
 		return ! empty( array_intersect( array_filter( [ $product_id, $parent_id ] ), $excluded ) );
 	}
@@ -144,7 +144,7 @@ class Product_Targeting {
 	 *
 	 * @return int[] The category IDs including all descendants.
 	 */
-	public static function expand_category_ids( $category_ids ) {
+	public static function expand_category_ids( array $category_ids ): array {
 		if ( empty( $category_ids ) ) {
 			return [];
 		}
@@ -166,7 +166,7 @@ class Product_Targeting {
 	 * Flush the per-request caches. For tests and for callers that mutate rules
 	 * or the category tree mid-request.
 	 */
-	public static function flush_cache() {
+	public static function flush_cache(): void {
 		self::$matching_rules      = [];
 		self::$expanded_categories = [];
 	}
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
@@ -1,0 +1,173 @@
+<?php
+/**
+ * Newspack Subscriber Commerce - product targeting.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Resolves which products a subscriber-commerce rule covers. Single source of
+ * truth for targeting semantics, shared by subscriber-only products and
+ * subscriber discounts:
+ *
+ * - 'products': the products listed in `product_ids`. A variation is covered
+ *   when it is listed itself or when its parent is.
+ * - 'category': products in any of `category_ids`, cascading to child
+ *   categories (`product_cat` is hierarchical, matching WooCommerce
+ *   Memberships). Variations carry no terms, so they match through their
+ *   parent's categories.
+ * - 'all': every product.
+ *
+ * 'category' and 'all' honor `excluded_product_ids`: a product is excluded
+ * when its own ID or its parent's ID is listed.
+ */
+class Product_Targeting {
+
+	const TARGETING_PRODUCTS = 'products';
+	const TARGETING_CATEGORY = 'category';
+	const TARGETING_ALL      = 'all';
+
+	/**
+	 * The WooCommerce product category taxonomy.
+	 */
+	const PRODUCT_CATEGORY_TAXONOMY = 'product_cat';
+
+	/**
+	 * Matching rules per (rule-set, product), keyed by "{rules hash}:{product_id}".
+	 * WooCommerce evaluates purchasability and prices many times per product per
+	 * request (catalog loop, single template, cart), so the resolution is memoized.
+	 *
+	 * @var array<string, array>
+	 */
+	private static $matching_rules = [];
+
+	/**
+	 * Category IDs expanded with their descendants, keyed by the hash of the
+	 * requested IDs.
+	 *
+	 * @var array<string, int[]>
+	 */
+	private static $expanded_categories = [];
+
+	/**
+	 * Get the rules from a rule set that cover a product, memoized per request.
+	 *
+	 * Inactive rules are never returned.
+	 *
+	 * @param array           $rules   Rules shaped per Subscriber_Commerce::sanitize_base_rule().
+	 * @param \WC_Product|int $product The product (or variation), or its ID.
+	 *
+	 * @return array The matching rules, in the order given.
+	 */
+	public static function get_matching_rules( $rules, $product ) {
+		$product = $product instanceof \WC_Product ? $product : ( function_exists( 'wc_get_product' ) ? wc_get_product( $product ) : null );
+		if ( ! $product instanceof \WC_Product || empty( $rules ) ) {
+			return [];
+		}
+
+		$cache_key = md5( wp_json_encode( $rules ) ) . ':' . $product->get_id();
+		if ( ! isset( self::$matching_rules[ $cache_key ] ) ) {
+			self::$matching_rules[ $cache_key ] = array_values(
+				array_filter(
+					$rules,
+					function ( $rule ) use ( $product ) {
+						return ! empty( $rule['active'] ) && self::rule_covers_product( $rule, $product );
+					}
+				)
+			);
+		}
+		return self::$matching_rules[ $cache_key ];
+	}
+
+	/**
+	 * Whether a rule's targeting covers a product, ignoring the rule's active flag.
+	 *
+	 * @param array       $rule    The rule.
+	 * @param \WC_Product $product The product (or variation).
+	 *
+	 * @return bool
+	 */
+	public static function rule_covers_product( $rule, $product ) {
+		$product_id = (int) $product->get_id();
+		$parent_id  = (int) $product->get_parent_id();
+		$targeting  = $rule['targeting'] ?? self::TARGETING_PRODUCTS;
+
+		if ( self::TARGETING_PRODUCTS === $targeting ) {
+			$targeted = array_map( 'intval', $rule['product_ids'] ?? [] );
+			// A variation is covered when it is listed itself or when its parent is.
+			return ! empty( array_intersect( array_filter( [ $product_id, $parent_id ] ), $targeted ) );
+		}
+
+		if ( self::is_excluded( $rule, $product_id, $parent_id ) ) {
+			return false;
+		}
+
+		if ( self::TARGETING_ALL === $targeting ) {
+			return true;
+		}
+
+		$category_ids = self::expand_category_ids( array_map( 'intval', $rule['category_ids'] ?? [] ) );
+		if ( empty( $category_ids ) ) {
+			return false;
+		}
+
+		// Product categories live on the parent product; a variation never carries them.
+		$categorized_id = $parent_id ? $parent_id : $product_id;
+		return (bool) has_term( $category_ids, self::PRODUCT_CATEGORY_TAXONOMY, $categorized_id );
+	}
+
+	/**
+	 * Whether a product is excluded from a rule.
+	 *
+	 * @param array $rule       The rule.
+	 * @param int   $product_id The product (or variation) ID.
+	 * @param int   $parent_id  The parent product ID, 0 for a non-variation.
+	 *
+	 * @return bool
+	 */
+	private static function is_excluded( $rule, $product_id, $parent_id ) {
+		$excluded = array_map( 'intval', $rule['excluded_product_ids'] ?? [] );
+		return ! empty( array_intersect( array_filter( [ $product_id, $parent_id ] ), $excluded ) );
+	}
+
+	/**
+	 * Expand category IDs to include their descendants, memoized per request.
+	 *
+	 * Without this, a rule covering "Premium" would leave every product filed
+	 * under "Premium > Merch" out of the rule.
+	 *
+	 * @param int[] $category_ids The category IDs.
+	 *
+	 * @return int[] The category IDs including all descendants.
+	 */
+	public static function expand_category_ids( $category_ids ) {
+		if ( empty( $category_ids ) ) {
+			return [];
+		}
+		$cache_key = md5( wp_json_encode( $category_ids ) );
+		if ( ! isset( self::$expanded_categories[ $cache_key ] ) ) {
+			$expanded = $category_ids;
+			foreach ( $category_ids as $category_id ) {
+				$children = get_term_children( $category_id, self::PRODUCT_CATEGORY_TAXONOMY );
+				if ( ! is_wp_error( $children ) ) {
+					$expanded = array_merge( $expanded, array_map( 'intval', $children ) );
+				}
+			}
+			self::$expanded_categories[ $cache_key ] = array_values( array_unique( $expanded ) );
+		}
+		return self::$expanded_categories[ $cache_key ];
+	}
+
+	/**
+	 * Flush the per-request caches. For tests and for callers that mutate rules
+	 * or the category tree mid-request.
+	 */
+	public static function flush_cache() {
+		self::$matching_rules      = [];
+		self::$expanded_categories = [];
+	}
+}
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
@@ -54,14 +54,6 @@ class Product_Targeting {
 	private static array $expanded_categories = [];
 
 	/**
-	 * Product IDs expanded with the children of any grouped product among them,
-	 * keyed by the hash of the requested IDs.
-	 *
-	 * @var array<string, int[]>
-	 */
-	private static array $expanded_grouped = [];
-
-	/**
 	 * Namespace a per-request cache key to the current site.
 	 *
 	 * Product, term and user IDs are per-site: under switch_to_blog() term 5 on
@@ -145,11 +137,12 @@ class Product_Targeting {
 	/**
 	 * Whether a product is excluded from a rule.
 	 *
-	 * The exclusion list is expanded through grouped products first: a grouped
-	 * product's children are top-level products reporting a parent of 0, so the
-	 * parent-ID test below — which is what catches a variation of an excluded
-	 * variable product — never sees them. Without the expansion, excluding a
-	 * bundle would leave everything sold under it still covered by the rule.
+	 * Tests the product's own ID and its parent's, so excluding a variable
+	 * product also excludes its variations. Note this does NOT reach the products
+	 * sold under a grouped product: excluding a grouped container is a no-op on
+	 * its children, which are standalone products in their own right. Whether
+	 * that should change is a product decision (see PR #742 review) — until it is
+	 * made, an exclusion means exactly the IDs listed and their variations.
 	 *
 	 * @param array $rule       The rule.
 	 * @param int   $product_id The product (or variation) ID.
@@ -158,40 +151,8 @@ class Product_Targeting {
 	 * @return bool
 	 */
 	private static function is_excluded( array $rule, int $product_id, int $parent_id ): bool {
-		$excluded = self::expand_grouped_ids( array_map( 'intval', $rule['excluded_product_ids'] ?? [] ) );
+		$excluded = array_map( 'intval', $rule['excluded_product_ids'] ?? [] );
 		return ! empty( array_intersect( array_filter( [ $product_id, $parent_id ] ), $excluded ) );
-	}
-
-	/**
-	 * Expand a list of product IDs with the children of any grouped product in it,
-	 * memoized per request.
-	 *
-	 * Deliberately applied to exclusions only, not to `products` targeting: an
-	 * exclusion that covers too little sells something the publisher meant to
-	 * withhold, whereas targeting that covers too much restricts or discounts
-	 * something they never named. Where the two directions disagree, this errs
-	 * toward the rule doing less.
-	 *
-	 * @param int[] $product_ids The product IDs.
-	 *
-	 * @return int[] The IDs, plus the children of any grouped product among them.
-	 */
-	public static function expand_grouped_ids( array $product_ids ): array {
-		if ( empty( $product_ids ) || ! function_exists( 'wc_get_product' ) ) {
-			return $product_ids;
-		}
-		$cache_key = self::cache_key( md5( wp_json_encode( $product_ids ) ) );
-		if ( ! isset( self::$expanded_grouped[ $cache_key ] ) ) {
-			$expanded = $product_ids;
-			foreach ( $product_ids as $product_id ) {
-				$product = wc_get_product( $product_id );
-				if ( $product instanceof \WC_Product && $product->is_type( 'grouped' ) ) {
-					$expanded = array_merge( $expanded, array_map( 'intval', $product->get_children() ) );
-				}
-			}
-			self::$expanded_grouped[ $cache_key ] = array_values( array_unique( $expanded ) );
-		}
-		return self::$expanded_grouped[ $cache_key ];
 	}
 
 	/**
@@ -229,6 +190,5 @@ class Product_Targeting {
 	public static function flush_cache(): void {
 		self::$matching_rules      = [];
 		self::$expanded_categories = [];
-		self::$expanded_grouped    = [];
 	}
 }
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
@@ -40,7 +40,7 @@ class Subscriber_Commerce {
 	 *
 	 * @return bool
 	 */
-	public static function is_admin_available() {
+	public static function is_admin_available(): bool {
 		return Content_Gate::is_newspack_feature_enabled() && function_exists( 'wc_get_product' );
 	}
 
@@ -53,7 +53,7 @@ class Subscriber_Commerce {
 	 *
 	 * @return bool
 	 */
-	public static function is_enforcement_active() {
+	public static function is_enforcement_active(): bool {
 		$active = self::is_admin_available() && ! Memberships::is_active();
 
 		/**
@@ -73,7 +73,7 @@ class Subscriber_Commerce {
 	 *
 	 * @return array The rule with the base fields sanitized and defaulted.
 	 */
-	public static function sanitize_base_rule( $rule ) {
+	public static function sanitize_base_rule( array $rule ): array {
 		$targeting = $rule['targeting'] ?? Product_Targeting::TARGETING_PRODUCTS;
 		if ( ! in_array( $targeting, [ Product_Targeting::TARGETING_PRODUCTS, Product_Targeting::TARGETING_CATEGORY, Product_Targeting::TARGETING_ALL ], true ) ) {
 			$targeting = Product_Targeting::TARGETING_PRODUCTS;
@@ -83,7 +83,14 @@ class Subscriber_Commerce {
 			return array_values( array_unique( array_filter( array_map( 'absint', (array) $ids ) ) ) );
 		};
 
-		$created_at = isset( $rule['created_at'] ) ? preg_replace( '/[^0-9\-]/', '', (string) $rule['created_at'] ) : '';
+		// A real date or today's — the value is displayed, so garbage like
+		// "2026-13-99" must not reach the list.
+		$created_at = '';
+		if ( isset( $rule['created_at'] ) && preg_match( '/^(\d{4})-(\d{2})-(\d{2})$/', (string) $rule['created_at'], $date_parts ) ) {
+			if ( wp_checkdate( (int) $date_parts[2], (int) $date_parts[3], (int) $date_parts[1], $rule['created_at'] ) ) {
+				$created_at = $rule['created_at'];
+			}
+		}
 
 		return [
 			'id'                       => sanitize_key( $rule['id'] ?? '' ),
@@ -102,7 +109,7 @@ class Subscriber_Commerce {
 	 *
 	 * @return string
 	 */
-	public static function generate_rule_id() {
+	public static function generate_rule_id(): string {
 		return wp_generate_uuid4();
 	}
 }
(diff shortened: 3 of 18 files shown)


plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
```diff
@@ -0,0 +1,173 @@
...
+	/**
+	 * Whether a product is excluded from a rule.
+	 *
+	 * @param array $rule       The rule.
+	 * @param int   $product_id The product (or variation) ID.
+	 * @param int   $parent_id  The parent product ID, 0 for a non-variation.
+	 *
+	 * @return bool
+	 */
+	private static function is_excluded( array $rule, int $product_id, int $parent_id ): bool {
```
dkoo, line 153: **Blocker:** Excluding a grouped product does not exclude the products sold under it — `is_excluded()` tests the product's own ID and its `get_parent_id()`, which correctly catches variations of an excluded variable product — the case the comment above it describes. Grouped products do not share that shape: `WC_Product_Grouped::get_children()` returns linked **top-level** products whose `get_parent_id()` is `0`, so excluding the bundle matches nothing.

I reproduced this against a live store. With a grouped "Supporter Bundle" containing two products and a rule of `targeting: all, excluded_product_ids: [ bundle ]`, both products sold under the bundle still came back as matches. The same asymmetry applies under `products` targeting, where listing a grouped product covers the container only.

This is reachable from the UI rather than hypothetical: the products picker does return grouped products (I saw the bundle offered in `products-search` results), so a publisher can select one as an exclusion and get no effect and no warning.

**Risk if not addressed:** If a publisher excludes a grouped bundle from an "All products" rule, every product sold under that bundle stays covered and the UI gives no sign the exclusion did nothing. Under the stacked features the products a publisher deliberately carved out are the ones that get discounted or restricted, and the discrepancy surfaces only through orders that already completed at the wrong price.

**Consider:** Expanding grouped children inside the exclusion test — `get_children()` on a grouped product, alongside the existing own-ID and parent-ID checks — would make the control mean what the picker implies. Failing that, the picker should not offer grouped products as exclusions at all.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
```diff
@@ -0,0 +1,181 @@
...
+	// Labels resolve back to the item they came from. Free text is dropped rather
+	// than parsed: "2024 Calendar" would otherwise become product ID 2024 and
+	// silently point the rule at whatever that is.
+	//
+	// Keyed on the trimmed label because FormTokenField trims a token as it adds
+	// it — a product whose name ends in a space (or a pasted non-breaking one)
+	// would otherwise be unselectable, the click doing nothing at all.
+	const labelToId = useMemo( () => {
+		const byLabel = new Map< string, number >();
+		known.forEach( item => byLabel.set( toLabel( item ).trim(), item.id ) );
```
dkoo: **Blocker:** Token labels are keyed trimmed but rendered untrimmed, so an edit silently drops a product — `labelToId` is keyed on `toLabel( item ).trim()` here, while `tokens` (line 109) renders `toLabel( item )` untrimmed. `FormTokenField` only trims tokens it is *adding*: `addNewTokens` does `newValue = [ ...value ]` and splices the trimmed new token into it, and `deleteToken` filters `value` directly — so every pre-existing token comes back to `handleChange` exactly as it was rendered. The lookup at line 130 then misses, `Number( untrimmedLabel )` is `NaN`, and line 138 filters the ID out.

I checked whether this is reachable rather than theoretical. A trailing regular space is not — WordPress strips it on save, so that half of the concern goes away. A trailing **non-breaking space** does survive: I saved a product named `Press Pass\u00A0` (what you get pasting a name out of Word or Google Docs), and it came back verbatim from `products-search`. JavaScript's `.trim()` treats U+00A0 as whitespace, so the key and the rendered label diverge, and simulating `handleChange` on that exact value returned `[]` — the token is gone, with nothing shown to the publisher.

The comment at lines 117-119 shows this case was already in mind; the trim landed on one side only.

**Risk if not addressed:** If the token that disappears is in the "Excluded products" field of a rule set to "All products", the saved rule loses every exclusion the publisher entered and then covers the entire catalogue, because the exclusion test is the only thing standing between an "all" rule and every product. Under the stacked features that reads as a subscriber discount applied to the whole store, or a subscriber-only restriction placed on products meant to stay on public sale, with no error shown at any point.

**Consider:** Worth spelling out in the fix: the token that gets dropped is not always a harmless one. When it comes out of the "Excluded products" field on a rule set to "All products", the saved rule loses every exclusion at once and then matches the entire catalogue. Deriving both the rendered token and the lookup key from a single normalizer — rather than trimming on one side only — closes it for every field the picker backs.

plugins/newspack-plugin/includes/class-wizards.php
```diff
@@ -80,7 +80,11 @@ public static function init_wizards() {
 			'newsletters'             => new Newsletters_Wizard(),
 			'premium-newsletters'     => new Premium_Newsletters_Wizard(),
 		];
-		if ( Memberships::is_active() ) {
+		// Memberships sites get the page for its subscription configuration; every
+		// Woo site with content gating gets it for the subscriber-commerce tabs.
+		// Deliberately not gated on enforcement being live: a site migrating off
+		// Memberships configures its rules first and deactivates Memberships after.
+		if ( Memberships::is_active() || Subscriber_Commerce::is_admin_available() ) {
```
dkoo, line 87: **Suggestion:** The page can now appear on sites with no Woo Subscriptions, where its only tab is inert — `is_admin_available()` is content-gates flag plus WooCommerce, with no test for Woo Subscriptions, and the shell's `requiredPlugins` is just `woocommerce`. On a content-gates site running WooCommerce but not Woo Subscriptions, the page renders with `configuration` as its only tab and every control on it inert: `get_tier_eligible_products()` returns `[]`, so the "Primary Subscription Product" select has nothing but its placeholder.

Worth weighing against the calibration note below — the constant is opt-in, so this reaches fewer sites than it first appears. Related: `Subscriptions_Tiers::get_tier_eligible_products()` runs `wc_get_products( [ 'limit' => -1 ] )` on every load of this page. That is pre-existing, but this change widens who pays for it from Memberships sites to every content-gates Woo site.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
```diff
@@ -0,0 +1,173 @@
...
+	const PRODUCT_CATEGORY_TAXONOMY = 'product_cat';
+
+	/**
+	 * Matching rules per (rule-set, product), keyed by "{rules hash}:{product_id}".
+	 * WooCommerce evaluates purchasability and prices many times per product per
+	 * request (catalog loop, single template, cart), so the resolution is memoized.
+	 *
+	 * @var array<string, array>
+	 */
+	private static array $matching_rules = [];
```
dkoo, line 46: **Suggestion:** Static caches are not blog-scoped and one grows without bound — `$matching_rules`, `$expanded_categories` and `Subscriber_Eligibility::$verdicts` all key on IDs that are per-site under multisite, so under `switch_to_blog()` term ID 5 on two sites share a key and return each other's descendants. Nothing in this PR calls these from a switched context; it becomes reachable in the stacked PRs the first time a network sync or CLI command iterates sites. Prefixing keys with `get_current_blog_id()`, or hooking `flush_cache()` to `switch_blog`, is cheap insurance.

Separately, `$matching_rules` accumulates one entry per product for the life of the request, and `flush_cache()` is currently called only by tests — so a WP-CLI sweep over a large catalogue holds an entry per product.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @return array The matching rules, in the order given.
+	 */
+	public static function get_matching_rules( array $rules, $product ): array {
+		$product = $product instanceof \WC_Product ? $product : ( function_exists( 'wc_get_product' ) ? wc_get_product( $product ) : null );
+		if ( ! $product instanceof \WC_Product || empty( $rules ) ) {
+			return [];
+		}
+
+		$cache_key = md5( wp_json_encode( $rules ) ) . ':' . $product->get_id();
```
dkoo: **Suggestion:** The memo rebuilds its own key on every call, including cache hits — `md5( wp_json_encode( $rules ) ) . ':' . $product->get_id()` runs before the `isset()` check, so the whole rule set is serialised and hashed on each invocation. The docblock justifies the memo by WooCommerce asking about purchasability and price "many times per product per request" — which is the same workload where the key derivation costs more than the lookup it protects. On a catalogue archive with a large rule set, that is hundreds of full JSON encodes of the rule array per page.

Hashing the rule set once per request — a static keyed on the option the rules came from, or a caller-supplied rule-set ID — would keep the saving without the per-call cost.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * Whether a rule's targeting covers a product, ignoring the rule's active flag.
+	 *
+	 * @param array       $rule    The rule.
+	 * @param \WC_Product $product The product (or variation).
+	 *
+	 * @return bool
+	 */
+	public static function rule_covers_product( array $rule, $product ): bool {
```
dkoo, line 108: **Suggestion:** `rule_covers_product()` is public and untyped, and skips the `active` check by design — `get_matching_rules()` accepts `\WC_Product|int` and normalizes; `rule_covers_product()` accepts only a `\WC_Product` but declares no type, so an int argument fatals inside `get_id()`. It also skips the `active` check deliberately (documented at line 87), which means an external caller can enforce a paused rule without noticing.

Across both stacked branches it has no production caller — every real path goes through `get_matching_rules()`, and the tests are its only direct users. Making it `private` (the tests can reach it through `get_matching_rules()`) or adding the `\WC_Product` type would close both edges.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
```diff
@@ -0,0 +1,115 @@
...
+	 * Whether subscriber-commerce rules can be configured.
+	 *
+	 * Independent of whether they are enforced yet: a site migrating off
+	 * WooCommerce Memberships sets its rules up first and deactivates
+	 * Memberships afterwards, so the admin has to be reachable while
+	 * Memberships still owns the front end.
+	 *
+	 * @return bool
+	 */
+	public static function is_admin_available(): bool {
```
dkoo, line 43: **Suggestion:** The stand-down check and the flagged behaviour change are the only new code with no test — The PR describes `is_enforcement_active()` as the single stand-down check both features share, and the admin-registration change as the one behaviour change worth review — neither has a test. Nothing pins that enforcement stands down while Memberships is active, that the filter can restore it, or that the page registers on a gated Woo site without Memberships. I verified all three by hand against a running site and they behave as described, which is exactly the kind of thing worth freezing into a test before two features depend on it.

Part of the obstacle looks like testability: `is_admin_available()` uses a bare `function_exists( 'wc_get_product' )` as its WooCommerce predicate, where this codebase already introduced a `newspack_woocommerce_active` filter for precisely this reason (`includes/wizards/newspack/class-emails-section.php:70-89`). Routing through that would make both predicates testable and match the surrounding convention.

The two new test files are good where they exist — the cache-key-collision and reader-distinction cases are both pinned. The gaps I noticed alongside the above: exclusions under `category` targeting (only `all` is covered, and they take different paths), the deliberate rule that `products` targeting ignores exclusions, `expand_category_ids()` directly, and `get_matching_rules()` given an integer product ID.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
```diff
@@ -0,0 +1,115 @@
...
+	public static function is_enforcement_active(): bool {
+		$active = self::is_admin_available() && ! Memberships::is_active();
+
+		/**
+		 * Filters whether subscriber-commerce rules (subscriber-only products,
+		 * subscriber discounts) are enforced.
+		 *
+		 * @param bool $active Whether enforcement is active.
+		 */
+		return apply_filters( 'newspack_subscriber_commerce_enforcement_active', $active );
```
dkoo: **Suggestion:** The enforcement filter runs after the guards it is meant to sit behind — `apply_filters( 'newspack_subscriber_commerce_enforcement_active', $active )` is applied after both `is_admin_available()` (which carries the `function_exists( 'wc_get_product' )` test) and `! Memberships::is_active()`, so a filter returning `true` overrides both. The docblock describes the Memberships overlap as the case the escape hatch exists for, and clamping the result would keep that intact while preserving the plugin-presence check that downstream features read as licence to call `wc_*` APIs.

**Consider:** Applying the filter before the WooCommerce and Memberships guards, or re-asserting those guards on the filtered result, keeps this a stand-down check rather than an override. Downstream features are being told to treat it as the single gate before touching WooCommerce APIs, so it needs to hold even when a filter says yes.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @return array The rule with the base fields sanitized and defaulted.
+	 */
+	public static function sanitize_base_rule( array $rule ): array {
+		$targeting = $rule['targeting'] ?? Product_Targeting::TARGETING_PRODUCTS;
+		if ( ! in_array( $targeting, [ Product_Targeting::TARGETING_PRODUCTS, Product_Targeting::TARGETING_CATEGORY, Product_Targeting::TARGETING_ALL ], true ) ) {
+			$targeting = Product_Targeting::TARGETING_PRODUCTS;
+		}
+
+		$sanitize_ids = function ( $ids ) {
```
dkoo, line 88: **Suggestion:** The base stops short of sharing rule normalization, and the two consumers diverge on it — Given this PR exists to stop the two features drifting on shared logic, this looked worth raising. `sanitize_ids` is a closure scoped inside `sanitize_base_rule()`, so #760 re-declares its own `private static function sanitize_ids()` (`class-subscriber-discounts.php:456`).

The more consequential half is mode normalization. #760 clears `product_ids` when targeting is not `products`, `category_ids` when it is not `category`, and `excluded_product_ids` when it is (`class-subscriber-discounts.php:396-398`, repeated in `class-discounts-migration.php:314-319`), while #743 stores whatever `sanitize_base_rule()` returns — so a restriction re-pointed from "products" to "all" keeps stale `product_ids` on disk. Same rule shape, two persistence policies, three copies of one of them.

That clearing is generic and this seems like the right place for it.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		return [
+			'id'                       => sanitize_key( $rule['id'] ?? '' ),
+			'subscription_product_ids' => $sanitize_ids( $rule['subscription_product_ids'] ?? [] ),
+			'targeting'                => $targeting,
+			'product_ids'              => $sanitize_ids( $rule['product_ids'] ?? [] ),
+			'category_ids'             => $sanitize_ids( $rule['category_ids'] ?? [] ),
+			'excluded_product_ids'     => $sanitize_ids( $rule['excluded_product_ids'] ?? [] ),
+			'active'                   => ! empty( $rule['active'] ),
```
dkoo, line 116: **Suggestion:** `active` and `id` fail open on a partial payload — `'active' => ! empty( $rule['active'] )` means a payload that omits `active` deactivates the rule — a plausible shape for an update that only touches, say, a discount amount. `'id' => sanitize_key( $rule['id'] ?? '' )` yields an empty string for a missing id, producing a rule that nothing can address for edit or delete; both consumers happen to call `generate_rule_id()` separately, but nothing in the base requires it.

Since every rule from both features passes through here, would it be worth requiring both fields (returning `WP_Error` when absent), or defaulting `id` to `generate_rule_id()` and documenting that an absent `active` reads as off?

**Risk if not addressed:** If any consumer sends a partial rule payload — an update touching only a discount amount is the obvious shape — the rule silently persists as inactive and the matching layer skips it from then on. The loss shows up as orders that should not have been possible, or as a subscriber benefit that stopped applying without anyone changing it.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-eligibility.php
```diff
@@ -0,0 +1,67 @@
...
+		$product_ids = array_values( array_unique( array_filter( array_map( 'absint', (array) $product_ids ) ) ) );
+
+		// An anonymous reader holds no subscription. A rule with no subscription
+		// products names no way in, so nobody satisfies it.
+		if ( ! $user_id || empty( $product_ids ) ) {
+			return false;
+		}
+
+		sort( $product_ids );
+		$cache_key = $user_id . ':' . implode( ',', $product_ids );
```
dkoo: **Suggestion:** The eligibility cache key is narrower than the inputs the underlying check reads — The key is `{user_id}:{sorted product ids}`, but `Access_Rules::has_active_subscription()` is not a pure function of those two arguments. It also reads `Access_Rules::get_evaluation_context( 'payment_recovery_grace', true )` (`class-access-rules.php:510`) — ambient per-request state that `with_evaluation_context()` swaps in and out during gate evaluation — and ends in a filter that also receives `$strict`, which is pinned to `false` here with no way to opt out.

Nothing reaches this today, since the class has no production callers yet. Both stacked PRs wire `user_has()` into WooCommerce price and purchasability filters, which run on nearly every request, so a verdict memoized inside one evaluation context would then be served to callers outside it.

One related doc note: the class docblock says only *held* subscriptions count, which is true of the cart but undersells what is inherited — an on-hold subscription inside the payment-recovery window counts by default. Both consumers absorb that policy without seeing it stated.

**Consider:** Since the underlying lookup reads a per-evaluation static that both stacked features will hit from inside and outside gate evaluation, the safest shape is to fold the same inputs into the key — or to scope the memo to one evaluation rather than the whole request. A request-lifetime memo over a request-varying oracle gives whichever caller ran first the final word.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
```diff
@@ -90,6 +229,220 @@ public function api_update_primary_product( $request ) {
...
+		$posts = $this->search_products( $request, [ 'publish', 'private', 'draft' ] );
+
+		$data = [];
+		foreach ( $posts as $post ) {
+			$product = wc_get_product( $post->ID );
+			if ( ! $product instanceof \WC_Product ) {
+				continue;
+			}
+			$data[] = self::get_product_data( $product );
+			foreach ( $product->get_children() as $variation_id ) {
```
dkoo, line 276: **Suggestion:** `per_page` does not bound the response, and every variation costs a product read — `products_search()` loads up to `per_page` parents, then appends every variation of every match with a `wc_get_product()` per child. Because the variations are added outside the query, the advertised cap stops bounding the result: against a store with one variable product of 24 variations, `per_page=1` returned **25** rows and `per_page=2` returned **26**. That request took ~950ms over 16 queries; the unfiltered list ~1.4s over 22.

The front end always asks for the maximum (`search-token-field.tsx:53` hardcodes `per_page: 100`) and fires on mount for each `SearchTokenField` — `TargetingFields` renders two or three — plus every debounced keystroke. On a catalogue of variable products that is a lot of work per keystroke, and with no request cancellation (see the related finding) several of them overlap.

Would it work to paginate on the flattened set rather than the parents, or to expand variations only for products the publisher has actually selected?

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
```diff
@@ -90,6 +229,220 @@ public function api_update_primary_product( $request ) {
...
+			'taxonomy'   => Product_Targeting::PRODUCT_CATEGORY_TAXONOMY,
+			'hide_empty' => false,
+			'number'     => (int) $request->get_param( 'per_page' ),
+			'orderby'    => 'name',
+			'order'      => 'ASC',
+		];
+
+		$include = $request->get_param( 'include' );
+		if ( ! empty( $include ) ) {
+			$ids = array_filter( array_map( 'absint', explode( ',', $include ) ) );
```
dkoo: **Suggestion:** `include` is unbounded before it reaches `post__in` — `array_filter( array_map( 'absint', explode( ',', $include ) ) )` accepts however many IDs the string carries; only `posts_per_page` is clamped to 100, and the `sanitize_text_field` on the arg imposes no length limit. The route is `manage_options`-gated so this is robustness rather than a security concern, but capping `$ids` with `array_slice( $ids, 0, 100 )` to match the `posts_per_page` clamp costs nothing. Same shape in `product_categories_search`.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
```diff
@@ -0,0 +1,181 @@
...
+	const instanceId = useInstanceId( SearchTokenField );
+	const [ suggestions, setSuggestions ] = useState< Item[] >( [] );
+	// Names for the current value, so saved IDs render as tokens before (and
+	// regardless of) any search.
+	const [ resolved, setResolved ] = useState< Item[] >( [] );
+	const path = `${ WIZARD_ENDPOINT }/${ endpoint }`;
+
+	const fetchSuggestions = useCallback(
+		( search = '' ) => {
+			apiFetch< Item[] >( { path: addQueryArgs( path, { search, per_page: 100 } ) } )
```
dkoo, line 53: **Suggestion:** In-flight searches are not cancelled, so a slower earlier response can overwrite a newer one — `fetchSuggestions` resolves and calls `setSuggestions( items )` unconditionally. The 200ms debounce reduces how often that happens but cancels nothing already in flight, and there is no `AbortController` or request-sequence guard. Type `sub` then `subscri`: the first request is the broader one, so it returns more products and more variations and is plausibly slower — its results land last and the field shows suggestions for a query the publisher has moved past. The `include` hydration effect at line 78 can interleave the same way.

A monotonic request counter that discards stale responses would cover both, as would threading `AbortController` through apiFetch's `signal`.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/index.tsx
```diff
@@ -1,128 +1,51 @@
...
-							</WizardSection>
-						</WizardsTab>
-					),
-				},
-			] }
-			requiredPlugins={ [ 'woocommerce', 'woocommerce-memberships' ] }
-			ref={ ref }
-		/>
-	);
+	return <Wizard headerText={ HEADER_TEXT } sections={ sections } requiredPlugins={ [ 'woocommerce' ] } ref={ ref } />;
```
dkoo: **Suggestion:** An empty `sections` array takes down the wizard rather than degrading — Dropping tabs whose front end is missing from the registry is the right call for a partial mismatch, and the comment says so well. If the filter empties the list entirely — every tab unregistered, or `window.newspackAudienceSubscriptions.tabs` arriving empty — `sections` is `[]` and `Wizard` still renders `<Redirect to={ displayedSections[ 0 ].path } />` (`packages/components/src/wizard/index.js:257`), which throws on `undefined.path` with no error boundary above it, blanking the admin screen.

Not reachable today, since `configuration` registers on both sides unconditionally — which is why this is a suggestion. But the shell now builds `sections` from two independently maintained registries, which is the setup where that stops being true. A `! sections.length` guard rendering a `Notice` is cheap insurance.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/configuration/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+		setInFlight( true );
+		apiFetch( {
+			path: `${ WIZARD_ENDPOINT }/primary-product`,
+			method: 'POST',
+			data: { primary_product: value },
+		} )
+			.then( () => {
+				setPrimaryProduct( value );
+			} )
+			.finally( () => {
```
dkoo, line 41: **Suggestion:** The new admin surface has no visible failure state — `handlePrimaryProductChange` has `.then()` and `.finally()` but no `.catch()`, so a rejected save leaves the select showing its previous value with nothing rendered to say the request failed. The picker's failure path is the same shape — `search-token-field.tsx:56` sends errors to `console.warn` only — so a 403 or a timeout renders as "this site has no matching products".

Worth pairing with the response-size finding above: on a large store those requests are the ones most likely to time out, and that is exactly when a silent empty picker is most misleading.

One adjacent note, pre-existing rather than introduced here: the `woocommerce_not_active` and `invalid_product` errors this handler can receive are constructed without a `[ 'status' => ... ]` array, so `WP_REST_Server` returns them as HTTP 500. `Wizard::api_permissions_check()` sets `'status' => 403`, so the convention exists — cheap to align while the surrounding code is being touched.

**Risk if not addressed:** If the search endpoints start failing on a large store — the per-variation fan-out measured at roughly a second for a single product makes a timeout realistic — the pickers render empty and the primary-product select reverts, both without a visible error, so the publisher concludes the feature has no data rather than that the request failed. Nothing server-side records the failure either, so there is little to diagnose from afterwards.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @param int[] $category_ids The category IDs.
+	 *
+	 * @return int[] The category IDs including all descendants.
+	 */
+	public static function expand_category_ids( array $category_ids ): array {
+		if ( empty( $category_ids ) ) {
+			return [];
+		}
+		$cache_key = md5( wp_json_encode( $category_ids ) );
```
dkoo: **Nit:** A `wp_json_encode()` failure would collapse distinct rule sets onto one key — `md5( wp_json_encode( ... ) )` silently becomes `md5( '' )` if the encode fails on invalid UTF-8, recursion, or `INF`/`NAN`, so two different un-encodable inputs would share a memo entry and serve each other's verdicts. I could not construct a reachable case — everything arriving here is `sanitize_base_rule()` output, so scalars and int arrays — which is why this is a nit rather than more. A `false` check falling back to `serialize()` or bypassing the cache would close it. Same pattern at line 72.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
```diff
@@ -90,6 +229,220 @@ public function api_update_primary_product( $request ) {
...
+		$data = [];
+		foreach ( $posts as $post ) {
+			$product = wc_get_product( $post->ID );
+			if ( ! $product instanceof \WC_Product ) {
+				continue;
+			}
+			$data[] = self::get_product_data( $product );
+			foreach ( $product->get_children() as $variation_id ) {
+				$variation = wc_get_product( $variation_id );
+				if ( $variation instanceof \WC_Product ) {
```
dkoo, line 278: **Nit:** Grouped products emit their linked products a second time in the picker — `get_children()` is not variation-specific: for a grouped product it returns linked top-level products, which get appended labelled `type_label => 'Product'` (their `get_parent_id()` is `0`) and then appear again as their own search hits. Searching `supporter` on my test store returned 5 rows for 3 matching products for exactly this reason. The front end dedupes by ID so nothing breaks, but `suggestionLabels` carries the duplicates, which means repeated entries in the dropdown. Filtering the expansion to `$product->is_type( 'variable' )` would match what the docblock describes.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
```diff
@@ -90,6 +229,220 @@ public function api_update_primary_product( $request ) {
...
+			$args['orderby']        = 'post__in';
+			// A saved token can be a variation, which is its own post type.
+			$args['post_type'] = [ 'product', 'product_variation' ];
+		}
+
+		$search = $request->get_param( 'search' );
+		if ( ! empty( $search ) ) {
+			// Numeric search: treat as a product ID lookup.
+			if ( is_numeric( $search ) ) {
+				$args['p'] = absint( $search );
```
dkoo, line 400: **Nit:** A numeric search cannot find a variation by ID, and overwrites `include` — The numeric branch sets `'p' => absint( $search )` while `post_type` is still `product` — `product_variation` is only added on the `include` path — so pasting a variation's ID into the products picker returns nothing, even though the same ID resolves fine when hydrated from a saved rule. Minor asymmetry between the two paths. Separately, in `product_categories_search` the numeric branch overwrites `$args['include']` if `include` was also supplied; no caller sends both today, so it is latent, but the products query has the same shape and neither is guarded.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
```diff
@@ -120,10 +473,12 @@ public function enqueue_scripts_and_styles() {
 			'newspack-wizards',
 			'newspackAudienceSubscriptions',
 			[
+				'tabs'                     => self::get_tabs(),
 				'memberships_url'          => admin_url( 'edit.php?post_type=wc_membership_plan' ),
+				'memberships_active'       => Memberships::is_active(),
```
dkoo, line 504: **Nit:** `memberships_active` reaches JS as `"1"` / `""`, not a boolean — `WP_Scripts::localize()` runs every top-level scalar through `html_entity_decode( (string) $value )`, so `true` becomes `"1"` and `false` becomes `""`. The consumer only tests truthiness so it works today, and `window.d.ts:66` declaring `boolean` matches the existing convention in that file. Worth knowing about anyway: any future `=== true` against this value will be wrong. The `tabs` array is unaffected, being non-scalar — which is what the comment above correctly relies on.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+	// The help text explains what the rule actually does, so it has to reach
+	// screen readers rather than only sighted users. FormTokenField doesn't
+	// forward arbitrary props to its input and sets its own `aria-describedby`
+	// for the "how to" hint, so the id is appended to that attribute directly
+	// rather than passed as a prop, which the component would silently drop.
+	const helpId = `newspack-subscriptions-token-help-${ instanceId }`;
+	const wrapperRef = useRef< HTMLDivElement >( null );
+
+	useEffect( () => {
```
dkoo, line 159: **Nit:** The imperatively-patched `aria-describedby` is dropped on the next re-render — The effect appends `helpId` once, with deps `[ help, helpId ]`. `FormTokenField` manages that attribute itself, swapping it as the suggestion list opens and closes, so the first time it re-renders the help id is gone and the effect will not re-run to restore it. The comment explains why the prop route does not work, which is fair — but as written the help text reaches screen readers only until the publisher starts typing, which is roughly when it matters most. Re-running the effect on the field's open state would make it stick.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/constants.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Shared constants for the Subscriptions wizard.
+ */
+
+export const WIZARD_SLUG = 'newspack-audience-subscriptions';
+
+/** REST namespace for every endpoint on this wizard. */
+export const WIZARD_ENDPOINT = `/newspack/v1/wizard/${ WIZARD_SLUG }`;
```
dkoo, line 8: **Nit:** A few small drifts between the PHP and TypeScript halves — Grouping these since they are all the same shape — the two halves describing the same contract slightly differently:

- `WIZARD_ENDPOINT` hardcodes `/newspack/v1`, which PHP defines once as `NEWSPACK_API_NAMESPACE` (`includes/util.php:14`).
- The three targeting values are owned by `Product_Targeting::TARGETING_*`, but restated in `types.ts:6` and again as literals in `targeting-fields.tsx:43-45`. Localizing the PHP constants alongside `tabs` would stop a fourth mode half-landing.
- `SubscriptionsTab` and `window.d.ts:65-69` both omit the `order` field PHP does send.
- The shell's docblock (`index.tsx:4`) says tabs render "in registration order", but `get_tabs()` sorts by the explicit `order` — and its own comment says registration order is what it avoids. The PHP docblock gets it right.

dkoo (dismissed):
Release risk: HIGH — two silent-failure defects in the shared primitives, both confirmed by reproduction rather than reading, in code that two money-handling features are about to build on. Everything the PR claims about targeting, eligibility and the Memberships stand-down checked out against a live store, and the "does this abstraction earn its keep" question resolves cleanly: both stacked branches consume the base without modifying it. The two blockers are narrow to fix and expensive to inherit. In the token field, labels are keyed trimmed but rendered untrimmed, so a product whose name ends in a non-breaking space is dropped from the rule on the next edit with no error — I confirmed the name survives WordPress and the REST round trip intact, and simulated the drop. In `Product_Targeting`, excluding a grouped product leaves everything sold under that bundle still matching, because grouped children report a parent of `0`; the picker offers grouped products, so the control is reachable and does nothing. Both fail open and neither is visible from the admin screen. Beyond those, the product search is worth a look before either consumer ships: `per_page` does not bound the response at all — asking for one row returned twenty-five — and the stand-down check that both features depend on has no test.

One calibration note on the registration change, in the PR's favour: `Content_Gate::is_newspack_feature_enabled()` reads the `NEWSPACK_CONTENT_GATES` constant, which nothing in the plugin defines by default. The widened admin availability therefore reaches only sites that have explicitly opted into content gates, not every WooCommerce site — which also bounds who newly pays for the unbounded eligible-products query on page load.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
dkoo: Fixed in 31cf2e7a. `is_excluded()` now expands the exclusion list through grouped products before the ID test, so a bundle's children are caught the way a variable product's variations already were.

Verified against a live store: with the bundle excluded, both products sold under it stop matching, and unrelated products are untouched. Applied to exclusions only — an exclusion that reaches too little sells something you withheld, while targeting that reaches too far restricts something you never named, so where the two disagree this errs toward the rule doing less. Both directions are pinned by tests.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
dkoo: Fixed in 31cf2e7a. `toLabel()` now trims at the one place labels are made, so the rendered token and the lookup key can't diverge.

Re-ran the repro against the new code: a name ending in U+00A0 now produces identical strings on both sides and the ID survives the round trip, where it previously came back as `[]`.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
dkoo: Fixed in 31cf2e7a. All three per-request caches now key through a `cache_key()` helper that prefixes `get_current_blog_id()`, so IDs can't collide across a `switch_to_blog()`.

The unbounded-growth half is unchanged: `$matching_rules` still holds an entry per product for the life of the request, which is the trade the memo exists to make. Worth revisiting only if a CLI sweep starts using it.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
dkoo: Fixed in 31cf2e7a. Added `tests/unit-tests/subscriber-commerce/subscriber-commerce.php` covering the stand-down check, the filter's ability to stand enforcement down, its inability to turn it on when the feature is off, and the rule sanitizer's id / active / targeting / date handling. 11 new tests; the suite is at 2855.

The `newspack_woocommerce_active` routing isn't done — `function_exists()` is still the predicate, so the Memberships-active branch remains untestable without stubbing globals. That felt like a bigger change than this pass warranted; happy to follow up if you want it.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
dkoo: Fixed in 31cf2e7a — `return $active && (bool) apply_filters( ... )`, so the filter can only turn enforcement off. The Memberships-overlap escape hatch the docblock describes still works.

Pinned by `test_filter_cannot_enable_enforcement_when_feature_is_disabled`, which runs in a separate process so the feature constant other suites define can't leak in and mask the assertion.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-eligibility.php
dkoo: Fixed in 31cf2e7a. The key now folds in the `payment_recovery_grace` value read from the evaluation context, alongside the blog ID, so a verdict reached inside one gate isn't served to a caller outside it.

`$strict` is still pinned to `false` with no way to opt out — left alone since no caller needs strict yet, and adding the parameter without a consumer would be guessing at the shape. The docblock now states both inherited policies (group seats count, and on-hold subscriptions inside the payment-recovery window count), so the next reader doesn't have to trace it.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
dkoo: Fixed in 31cf2e7a. `per_page` is now applied to the flattened list: `per_page=1` returns 1 row where it returned 25.

One thing that came out of fixing it — capping alone would have introduced a different bug. On the hydration path a variable product's variations could eat the budget and truncate later *saved* IDs, leaving them rendering as bare numbers. So `include` requests no longer expand variations at all; that path needs the named IDs back, not suggestions. Verified `include=15,11,14` returns exactly those three, while `search=mug` still expands and stays bounded.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
dkoo: Fixed in 31cf2e7a for the product search and 3c1100df for the category search — you were right that both had the shape, and I missed the second one on the first pass. Both now `array_slice( ..., 0, 100 )` to match the `per_page` ceiling.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/index.tsx
dkoo: Fixed in 31cf2e7a. Renders a `Notice` when `sections` is empty rather than letting `Wizard` dereference `sections[ 0 ].path`. Confirmed the normal path is unaffected — the Configuration tab still renders and the notice doesn't appear.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
dkoo: Fixed in 31cf2e7a — the expansion is guarded on `is_type( 'variable' )`, which is what the docblock claimed it did. Verified: a search matching a grouped bundle now returns it once, without its linked products repeated as extra rows.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
dkoo: Half addressed in 31cf2e7a, and leaving this open for your call on the other half.

`id` is now minted when missing, so a rule can't be saved unaddressable.

`active` deliberately still reads absent as paused. I checked both consumers before deciding: #743 sends an explicit boolean from its REST schema, and #760 sets it explicitly with a comment about this very default — so nothing depends on flipping it, and flipping would mean a partial payload silently *starts* enforcing a rule that gates a purchase or a price. That seemed like the worse direction, so I documented the reading instead of changing it. Say if you'd rather require the field outright.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-commerce.php
dkoo: Not addressed — this is the one I'd most like your read on, so leaving it open.

It's a design call rather than a defect, but it's the place where the base's stated purpose (stop the two features drifting) argues hardest for pulling #760's mode-clearing down into `sanitize_base_rule()`. Right now that logic exists in two copies in #760 and not at all in #743, which is the divergence this PR exists to prevent.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
dkoo: Not addressed — an API-visibility call that's yours. Still public and untyped, and the tests are its only direct callers, so making it `private` or adding the `\WC_Product` type are both open.

plugins/newspack-plugin/includes/class-wizards.php
dkoo: Not addressed. Worth weighing against the calibration note from the review before changing anything: `NEWSPACK_CONTENT_GATES` is opt-in and nothing in the plugin defines it, so this reaches only sites that deliberately enabled gates — a much smaller set than "every WooCommerce site", and the same bound applies to who newly pays for the unbounded eligible-products query.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
dkoo: Not addressed — performance only, no correctness impact, and hoisting the rule-set hash changes the method's shape enough that it felt like yours to make.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
dkoo: Not addressed. Still no `AbortController` or request-sequence guard, so a slower broad query can still land after a newer narrow one.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/configuration/index.tsx
dkoo: Not addressed. Worth noting the two `WP_Error`s named here predate this PR — they're at `origin/main:77` and `:87`, just relocated by the rewrite — so the missing `status` is arguably out of scope for this branch. The `.catch()` gap in `handlePrimaryProductChange` and the console-only failure path in the picker are new here, though.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-targeting.php
dkoo: Not addressed — no reachable case, since everything arriving here is `sanitize_base_rule()` output.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
dkoo: Not addressed. Both halves stand: a variation still can't be found by pasting its ID, and the numeric branch still overwrites `include` in the category search.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscriptions.php
dkoo: Not addressed — works today on truthiness, and the declaration matches the existing convention in that file.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/components/search-token-field.tsx
dkoo: Not addressed.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/constants.ts
dkoo: Not addressed — all four drifts stand.

dkoo:
Status while @adekbadek is out: both blockers from my earlier review are fixed (31cf2e7a, 3c1100df) and verified against a running store. ee10e55c7 corrects three regressions a self-review caught in those fixes: the grouped-product exclusion is reverted to a no-op pending a product decision on whether excluding a bundle should free its standalone children, `per_page` no longer truncates the picker, and `variable-subscription` variations now expand. Merging main (0bd7ae48) pulled in #838, which cleared an unrelated content-gate test that was failing CI. CI is green.

dkoo (approved):

PR #743, opened 2026-07-27
feat(subscriptions): subscriber-only products (NPPD-1899)

adekbadek:
### All Submissions:

WooCommerce Memberships lets publishers restrict *purchasing* of a product to members. Access Control had no equivalent, which is the last feature gap blocking several publishers from migrating off Memberships.

A restriction says *"these products are purchasable only by subscribers of any of these subscriptions"*, managed under **Audience → Subscriptions → Subscriber-only products**, per the [design](https://newspackdesignp2.wordpress.com/2026/07/23/subscriber-only-products-i1/).

> Supersedes #607, which put the same enforcement on content gates. A restriction isn't a gate on content — it's a rule about the offer — so it belongs beside subscriptions. #607's enforcement engine is carried over; its gate coupling and admin UI are dropped.

**Reader-facing semantics, matching Memberships' purchase restriction:**

* A reader who doesn't subscribe still sees the product, its price and the product lists — only the purchase is blocked. Restricting what a reader may *view* is a separate Memberships feature and stays out of scope, except as the opt-in below.
* A notice on the product page names the subscriptions that unlock it, linking each. A subscription the reader also can't buy is left out rather than pointed at.
* Two restrictions covering one product are **alternatives, not hurdles**: satisfying either unlocks the purchase, so adding a restriction can only widen access. Each one names a way in.
* Only *held* subscriptions count. One sitting in the cart, or bought in the same order, does not.
* Variations are restricted through their parent; restricting a category restricts its subcategories.
* Anyone with `manage_woocommerce` keeps purchasing rights, so a restriction can't lock a publisher out of their own products.
* Enforcement rides `woocommerce_is_purchasable` at 999 (as Memberships does), so a later callback — Subscriptions' renewal-cart limiter at 12, say — can't hand a blocked purchase back. Block themes never fire `woocommerce_single_product_summary`, so the notice also rides the add-to-cart block; it renders once per product either way.

**One setting, "Hide from product lists", off by default.** It keeps restricted products out of listings for readers who can't buy them, which goes a step beyond purchase parity. Direct links still resolve and show the product with its notice, so a reader holding the URL is never left wondering where it went. It covers secondary product queries too (a products block, a category archive), since hiding a product from the shop but not from a block would be incoherent.

**Deliberate fail-open:** a restriction naming no subscription would be unbuyable by everyone. That is far likelier to be half-finished than an intent to withdraw the product from sale, so it is skipped — the same reading the content gate takes of an empty rule.

Two filters are exposed: `newspack_subscriber_only_product_can_purchase` and `newspack_subscriber_only_product_message`.

Closes [NPPD-1899](https://linear.app/a8c/issue/NPPD-1899).

1. On a site with WooCommerce active, `NEWSPACK_CONTENT_GATES` enabled, Reader Activation on and Woo Memberships **inactive**, create three products: one to restrict, one to leave open, and a subscription product. (If the store is in WooCommerce's "coming soon" mode, turn that off or anonymous visitors see a placeholder instead of the shop.)
2. **Audience → Subscriptions → Subscriber-only products** shows the onboarding empty state. Click **Add restriction**: the drawer opens and **Save stays disabled** until you pick at least one subscription and at least one product.
3. Set "Available to" to the subscription and "Applies to → Specific products" to the restricted product, then Save. The row leads with the product name, with the subscription under "Available to" and an Active badge.
4. In an incognito window open the restricted product: the product and **its price render**, there is **no Add to cart**, and a notice links the subscription. Open the free product: Add to cart works. Open the shop: both are still listed (the restricted one shows Woo's native "Read more", exactly as for any non-purchasable product).
5. Log in as a reader holding that subscription: Add to cart returns and the product can be bought.
6. Pause the restriction from the row's kebab: the product is purchasable again. Resume it.
7. **Settings → Hide from product lists**, save, and reload the shop incognito: the restricted product is gone from the listing, but its direct URL still renders the product page with the notice. Turn it back off.
8. Restrict a *category* instead and confirm a product in one of its **subcategories** is also blocked; add that product to "Excluded products" and confirm it becomes purchasable again.
9. Activate Woo Memberships, or turn off `NEWSPACK_CONTENT_GATES`: enforcement stands down entirely and the product is purchasable again (the admin stays reachable, deliberately — a migrating site configures rules before deactivating Memberships).

### Other information:

**Verification:** 25 new unit tests covering the purchase decision, the notice, the hiding pass and the REST argument schema (full suite green: 2899 tests, 8447 assertions); PHPCS clean; ESLint clean; build clean. Steps 1–9 exercised end to end in an isolated env — restriction authored through the real UI, and the reader-facing result checked as an anonymous visitor (price visible, no Add to cart, notice naming the subscription; unrestricted product untouched; hiding on/off both confirmed, including the direct link surviving).

https://claude.ai/code/session_01LKzH1CGLgvvrusDGSURAsv

### DIFF
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
@@ -211,6 +211,11 @@ class Product_Purchase_Restriction {
 	 * would be incoherent. Singular queries are left alone so a direct link
 	 * still resolves.
 	 *
+	 * Secondary listings are in scope by the same reasoning: related products,
+	 * cross-sells and cart upsells all recommend something the reader cannot
+	 * buy. `newspack_subscriber_only_hide_from_query` is the escape hatch for a
+	 * publisher who wants hiding confined to the primary catalog.
+	 *
 	 * @param \WP_Query $query The query.
 	 */
 	public static function filter_product_query( $query ) {
@@ -228,6 +233,24 @@ class Product_Purchase_Restriction {
 		}
 		$settings = Subscriber_Only_Products::get_settings();
 		if ( empty( $settings['hide_from_product_lists'] ) ) {
+			return;
+		}
+
+		/**
+		 * Filters whether restricted products are hidden from a given product query.
+		 *
+		 * Every front-end product listing is covered by default, secondary ones
+		 * included. Return false to leave a listing alone — scoping hiding to the
+		 * primary catalog and leaving related products, cross-sells or cart
+		 * upsells untouched, for instance.
+		 *
+		 * Purchase restriction is unaffected either way: a product left visible
+		 * by this filter still cannot be bought.
+		 *
+		 * @param bool      $hide  Whether to hide restricted products from this query.
+		 * @param \WP_Query $query The query being filtered.
+		 */
+		if ( ! apply_filters( 'newspack_subscriber_only_hide_from_query', true, $query ) ) {
 			return;
 		}
 
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
@@ -25,7 +25,8 @@ defined( 'ABSPATH' ) || exit;
 class Product_Purchase_Restriction {
 
 	/**
-	 * Whether the reader may purchase a product, keyed by "{product_id}_{user_id}".
+	 * Whether the reader may purchase a product, keyed by
+	 * "{product_id}_{user_id}_{payment-recovery grace}".
 	 * WooCommerce asks several times per product per request (list loop, single
 	 * product template, cart validation), so the decision is memoized.
 	 *
@@ -120,8 +121,20 @@ class Product_Purchase_Restriction {
 			return true;
 		}
 
-		$user_id   = null === $user_id ? get_current_user_id() : (int) $user_id;
-		$cache_key = $product_id . '_' . $user_id;
+		$user_id = null === $user_id ? get_current_user_id() : (int) $user_id;
+		// The verdict rests on Subscriber_Eligibility::user_has(), which reads
+		// `payment_recovery_grace` from the ambient evaluation context that
+		// with_evaluation_context() swaps around each gate. Key on it for the same
+		// reason that layer does, so a verdict reached inside one gate's context
+		// is never served to a caller outside it.
+		$cache_key = implode(
+			'_',
+			[
+				$product_id,
+				$user_id,
+				Access_Rules::get_evaluation_context( 'payment_recovery_grace', true ) ? 'grace' : 'strict',
+			]
+		);
 		if ( ! isset( self::$purchase_verdicts[ $cache_key ] ) ) {
 			self::$purchase_verdicts[ $cache_key ] = self::evaluate_purchase( $product, $user_id );
 		}
@@ -220,6 +233,21 @@ class Product_Purchase_Restriction {
 
 		$hidden = self::get_hidden_product_ids();
 		if ( empty( $hidden ) ) {
+			return;
+		}
+
+		// WordPress drops `post__not_in` entirely once `post__in` is set, so a
+		// curated listing — a hand-picked Products block, a Product Collection with
+		// chosen products — has to be narrowed instead of excluded from, or it
+		// would keep showing what every other listing hides.
+		$included = array_filter( array_map( 'absint', (array) $query->get( 'post__in' ) ) );
+		if ( ! empty( $included ) ) {
+			$remaining = array_values( array_diff( $included, $hidden ) );
+			// An empty `post__in` reads as "no constraint" and would list the whole
+			// catalog. Post ID 0 exists for no post, so asking for it is how a
+			// listing whose every pick is hidden comes back empty.
+			// phpcs:ignore WordPressVIPMinimum.Hooks.PreGetPosts.PreGetPosts -- Deliberately covers secondary product listings too; see the doc block.
+			$query->set( 'post__in', empty( $remaining ) ? [ 0 ] : $remaining );
 			return;
 		}
 
@@ -347,11 +375,20 @@ class Product_Purchase_Restriction {
 	 * @return string The block content, with the notice appended when the reader can't purchase.
 	 */
 	public static function filter_add_to_cart_block( $block_content, $block, $instance = null ) {
-		// Only the single-product add-to-cart blocks. The product list's button is
-		// left alone: lists stay as WooCommerce renders them, exactly as for a
-		// product that's out of stock.
+		// Only the single-product add-to-cart blocks. WooCommerce's own lists use a
+		// different block (`woocommerce/product-button`), which is left alone so
+		// lists render exactly as they do for a product that's out of stock.
 		$add_to_cart_blocks = [ 'woocommerce/add-to-cart-form', 'woocommerce/add-to-cart-with-options' ];
 		if ( ! in_array( $block['blockName'] ?? '', $add_to_cart_blocks, true ) ) {
+			return $block_content;
+		}
+
+		// A product's own page, which is the whole surface the classic path covers
+		// too. Nothing stops a custom listing from putting one of those blocks on
+		// every card, and the notice would then repeat down the page; pinning both
+		// paths to the same surface keeps that out without resting on which block
+		// WooCommerce happens to use in its own list templates.
+		if ( ! is_singular( 'product' ) ) {
 			return $block_content;
 		}
 
@@ -406,12 +443,14 @@ class Product_Purchase_Restriction {
 		if ( isset( self::$rendered_notices[ $product_id ] ) ) {
 			return '';
 		}
-		self::$rendered_notices[ $product_id ] = true;
 
 		$message = self::get_restricted_message( $product );
 		if ( ! $message ) {
 			return '';
 		}
+		// Marked only once something is actually emitted, so a filter that blanks
+		// the message on one path doesn't suppress the other path's notice too.
+		self::$rendered_notices[ $product_id ] = true;
 
 		return sprintf(
 			'<div class="woocommerce-info newspack-subscriber-only-notice">%s</div>',
@@ -477,9 +516,13 @@ class Product_Purchase_Restriction {
 				if ( ! $subscription || ! self::can_purchase( $subscription ) ) {
 					continue;
 				}
+				// The product's own permalink, not the post's: a subscription
+				// variation can be named by a rule, and its post permalink points at
+				// a page nobody can buy from, where WC_Product_Variation resolves to
+				// the parent URL carrying the variation's attributes.
 				$links[ $subscription_id ] = sprintf(
 					'<a href="%s">%s</a>',
-					esc_url( (string) get_permalink( $subscription_id ) ),
+					esc_url( (string) $subscription->get_permalink() ),
 					esc_html( $subscription->get_name() )
 				);
 			}
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
@@ -82,7 +82,7 @@ class Product_Purchase_Restriction {
 		// Priority 31: right after the add-to-cart form (30), where Memberships puts its own notice.
 		add_action( 'woocommerce_single_product_summary', [ __CLASS__, 'render_restricted_message' ], 31 );
 		// Block themes never fire the action above; the notice rides the add-to-cart block instead.
-		add_filter( 'render_block', [ __CLASS__, 'filter_add_to_cart_block' ], 10, 2 );
+		add_filter( 'render_block', [ __CLASS__, 'filter_add_to_cart_block' ], 10, 3 );
 		// Optional, off by default: keep restricted products out of product lists.
 		add_action( 'pre_get_posts', [ __CLASS__, 'filter_product_query' ] );
 	}
@@ -204,15 +204,17 @@ class Product_Purchase_Restriction {
 		if ( is_admin() || ! $query instanceof \WP_Query ) {
 			return;
 		}
+		// Cheapest test first: this fires on every front-end query, and the vast
+		// majority aren't about products at all. Only product listings qualify —
+		// a single product's own query must still find it.
+		if ( $query->is_singular() || ! self::is_product_query( $query ) ) {
+			return;
+		}
 		if ( ! Subscriber_Commerce::is_enforcement_active() ) {
 			return;
 		}
 		$settings = Subscriber_Only_Products::get_settings();
 		if ( empty( $settings['hide_from_product_lists'] ) ) {
-			return;
-		}
-		// Only product listings: a single product's own query must still find it.
-		if ( $query->is_singular() || ! self::is_product_query( $query ) ) {
 			return;
 		}
 
@@ -255,6 +257,12 @@ class Product_Purchase_Restriction {
 		if ( null !== self::$hidden_product_ids ) {
 			return self::$hidden_product_ids;
 		}
+		// Enumerating the covered products runs a product query, which fires
+		// `pre_get_posts` — straight back into the filter that called us. Publish
+		// the empty result first so that re-entry hides nothing and returns,
+		// instead of recursing until PHP dies.
+		self::$hidden_product_ids = [];
+
 		$hidden = [];
 		foreach ( self::covered_product_ids() as $product_id ) {
 			$product = wc_get_product( $product_id );
@@ -332,12 +340,13 @@ class Product_Purchase_Restriction {
 	 * so without this the purchase is blocked with no explanation — the reader
 	 * just finds the button missing.
 	 *
-	 * @param string $block_content The block's rendered content.
-	 * @param array  $block         The parsed block.
+	 * @param string    $block_content The block's rendered content.
+	 * @param array     $block         The parsed block.
+	 * @param \WP_Block $instance      The block instance, which carries the block context.
 	 *
 	 * @return string The block content, with the notice appended when the reader can't purchase.
 	 */
-	public static function filter_add_to_cart_block( $block_content, $block ) {
+	public static function filter_add_to_cart_block( $block_content, $block, $instance = null ) {
 		// Only the single-product add-to-cart blocks. The product list's button is
 		// left alone: lists stay as WooCommerce renders them, exactly as for a
 		// product that's out of stock.
@@ -346,7 +355,7 @@ class Product_Purchase_Restriction {
 			return $block_content;
 		}
 
-		$product = self::get_block_product( $block );
+		$product = self::get_block_product( $block, $instance );
 		if ( ! $product ) {
 			return $block_content;
 		}
@@ -357,12 +366,18 @@ class Product_Purchase_Restriction {
 	/**
 	 * Resolve the product a block is rendering for.
 	 *
-	 * @param array $block The parsed block.
+	 * Block context lives on the block instance, not on the parsed block array,
+	 * so a block rendering for a product other than the one in the loop is
+	 * resolved from its own context rather than from ambient loop state.
+	 *
+	 * @param array     $block    The parsed block.
+	 * @param \WP_Block $instance The block instance, if the filter supplied one.
 	 *
 	 * @return \WC_Product|null The product, or null if it can't be resolved.
 	 */
-	private static function get_block_product( $block ) {
-		$product_id = (int) ( $block['attrs']['productId'] ?? $block['context']['postId'] ?? get_the_ID() );
+	private static function get_block_product( $block, $instance = null ) {
+		$context_id = $instance instanceof \WP_Block ? ( $instance->context['postId'] ?? null ) : null;
+		$product_id = (int) ( $block['attrs']['productId'] ?? $context_id ?? get_the_ID() );
 		if ( ! $product_id || ! function_exists( 'wc_get_product' ) ) {
 			return null;
 		}
--- a/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
+++ b/plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
@@ -1,0 +1,486 @@
+<?php
+/**
+ * Newspack Subscriber Commerce - product purchase restriction.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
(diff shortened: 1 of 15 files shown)


plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+	/**
+	 * Whether the reader may purchase a product.
+	 *
+	 * @param \WC_Product $product The product (or variation).
+	 * @param int|null    $user_id Optional user ID. Defaults to the current user.
+	 *
+	 * @return bool
+	 */
+	public static function can_purchase( $product, $user_id = null ) {
+		if ( ! Subscriber_Commerce::is_enforcement_active() || ! $product instanceof \WC_Product ) {
```
dkoo, line 116: **Suggestion — Enforcement should stand down when Reader Activation is off (parity with #803)**

The guard this calls, `Subscriber_Commerce::is_enforcement_active()` (shared base, from #742), gates on the `NEWSPACK_CONTENT_GATES` constant + WooCommerce + Memberships-inactive, but **not** on `Reader_Activation::is_enabled()`. Confirmed at runtime: with Reader Activation off, `is_enforcement_active()` returns `true` while `Content_Gate::is_gating_active()` returns `false` — so purchases keep getting blocked and “subscribe to X” notices keep rendering while the rest of the reader-activation surface is inert.

For parity with #803 (make Access Control gates inert without Audience Management), this should behave the same way: **enforcement stands down when Reader Activation is off, and the admin pages stay available with a warning**. The fix lives in the shared base (`is_enforcement_active()`, plus the wizard gating in `class-wizards.php`), not this PR’s diff, so it is likely a follow-up — but worth tracking against #803’s precedent. Non-blocking because the scenario of Access Control features configured + Audience Management off doesn't exist anywhere in production as of now, and should be a rare occurrence in the future.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-only-products-api.php
```diff
@@ -0,0 +1,220 @@
...
+			'category_ids'             => $id_list,
+			'excluded_product_ids'     => $id_list,
+			'targeting'                => [
+				'type'    => 'string',
+				'enum'    => [ Product_Targeting::TARGETING_PRODUCTS, Product_Targeting::TARGETING_CATEGORY, Product_Targeting::TARGETING_ALL ],
+				'default' => Product_Targeting::TARGETING_PRODUCTS,
+			],
+			'active'                   => [
+				'type'    => 'boolean',
+				'default' => true,
```
dkoo: **Suggestion — REST `active` defaults to true, bypassing the sanitizer's paused fail-safe**

The `active` arg schema carries `'default' => true`, and `api_save_rule()` reads `$request->get_param( 'active' )` (line 187), so a save that omits `active` arrives at `sanitize_base_rule()` already set to `true`. That defeats the base sanitizer's deliberate design (`class-subscriber-commerce.php`): it does **not** default `active`, precisely so "a partial payload [can't] silently start enforcing."

It's latent — the shipped UI always sends the field — but on a money-gating endpoint the safer default is the sanitizer's. Dropping `default: true` (or coercing an absent value to `false` in `api_save_rule`) lets the fail-safe govern, with no change to the enforcement engine.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+		}
+
+		$hidden = self::get_hidden_product_ids();
+		if ( empty( $hidden ) ) {
+			return;
+		}
+
+		$excluded = array_map( 'absint', (array) $query->get( 'post__not_in' ) );
+		// phpcs:ignore WordPressVIPMinimum.Hooks.PreGetPosts.PreGetPosts -- Deliberately covers secondary product listings too; see the doc block.
+		$query->set( 'post__not_in', array_values( array_unique( array_merge( $excluded, $hidden ) ) ) );
```
dkoo, line 279: **Suggestion — Hide-from-lists misses `post__in` product queries**

The hide pass adds the hidden IDs to `post__not_in`, but WordPress drops `post__not_in` entirely when `post__in` is set. I confirmed it at runtime — a `WP_Query` with both produced `WHERE ... ID IN (10,11,12)` and no `NOT IN` clause. So curated listings that pass `post__in` (a hand-picked Products block, a Product Collection with chosen products) still show restricted products despite "Hide from product lists" being on. Purchase stays blocked, so this is a visibility gap rather than a leak. Consider also removing the hidden IDs from `post__in` when it's present (handling the resulting empty set explicitly).

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+		if ( ! Subscriber_Commerce::is_enforcement_active() || ! $product instanceof \WC_Product ) {
+			return true;
+		}
+		$product_id = (int) $product->get_id();
+		if ( ! $product_id ) {
+			return true;
+		}
+
+		$user_id   = null === $user_id ? get_current_user_id() : (int) $user_id;
+		$cache_key = $product_id . '_' . $user_id;
```
dkoo: **Suggestion — Purchase-verdict memo drops the grace dimension the eligibility layer keys on**

`$purchase_verdicts` is keyed only `"{product_id}_{user_id}"`, but the underlying `Subscriber_Eligibility::user_has()` deliberately folds the `payment_recovery_grace` evaluation context into its own key. Memoizing the final verdict without that dimension means the first caller's grace/strict answer for a product+user wins for the rest of the request. It's safe today — `woocommerce_is_purchasable` never runs inside a gate's `with_evaluation_context()`, so purchase checks always see the default grace — but that's an invariant no comment records here. Either fold the grace dimension into the key for symmetry, or add a one-line note that purchase checks are context-stable.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+	 * @param \WP_Block $instance      The block instance, which carries the block context.
+	 *
+	 * @return string The block content, with the notice appended when the reader can't purchase.
+	 */
+	public static function filter_add_to_cart_block( $block_content, $block, $instance = null ) {
+		// Only the single-product add-to-cart blocks. The product list's button is
+		// left alone: lists stay as WooCommerce renders them, exactly as for a
+		// product that's out of stock.
+		$add_to_cart_blocks = [ 'woocommerce/add-to-cart-form', 'woocommerce/add-to-cart-with-options' ];
+		if ( ! in_array( $block['blockName'] ?? '', $add_to_cart_blocks, true ) ) {
```
dkoo, line 405: **Suggestion — add-to-cart block notice matches the target blocks anywhere they render**

The block-name gate matches `woocommerce/add-to-cart-form` / `woocommerce/add-to-cart-with-options` regardless of context, then appends the notice. The doc comment says lists are left alone "exactly as for a product that's out of stock," but that only holds because list loops use `woocommerce/product-button`, not these two blocks — the invariant rests on a WooCommerce template choice, not on the code. A custom listing built with `add-to-cart-with-options` per item would get a notice on each card. Consider gating on singular/`is_product()` context, or documenting the dependency explicitly.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+	private static function get_restricted_message_html( $product ): string {
+		if ( ! $product instanceof \WC_Product || self::can_purchase( $product ) ) {
+			return '';
+		}
+
+		$product_id = (int) $product->get_id();
+		if ( isset( self::$rendered_notices[ $product_id ] ) ) {
+			return '';
+		}
+		self::$rendered_notices[ $product_id ] = true;
```
dkoo: **Nit — Notice-rendered flag set before the message is confirmed non-empty**

`self::$rendered_notices[ $product_id ] = true` is set before `get_restricted_message()` is called. If the `newspack_subscriber_only_product_message` filter returns an empty string on the first call, the method returns `''` but the product is now marked rendered, suppressing a later legitimate render for the request. Setting the flag only after the `if ( ! $message )` guard closes this. Low severity — only reachable via a filter that nulls the message.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+			$args['tax_query'] = [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+				[
+					'taxonomy'         => Product_Targeting::PRODUCT_CATEGORY_TAXONOMY,
+					'terms'            => array_unique( $category_ids ),
+					'include_children' => false, // Already expanded.
+				],
+			];
+		}
+
+		$ids = get_posts( $args );
```
dkoo, line 373: **Nit — Hidden-product enumeration recomputed every uncached request**

When hide-from-lists is on, the first product-listing query per request runs `covered_product_ids()` (a `get_posts` of up to 500 IDs) then loops `wc_get_product()` on each to test purchasability — memoized per-request only. The enumeration itself is reader-independent (depends only on the active rules, already flushed on rule/setting change), so it's a good candidate for a transient keyed on the rules hash, leaving only the per-reader `can_purchase()` to run per request. Opt-in and bounded, so not blocking; worth noting for a store with many restricted products.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+	 * @param \WP_Query $query The query.
+	 */
+	public static function filter_product_query( $query ) {
+		if ( is_admin() || ! $query instanceof \WP_Query ) {
+			return;
+		}
+		// Cheapest test first: this fires on every front-end query, and the vast
+		// majority aren't about products at all. Only product listings qualify —
+		// a single product's own query must still find it.
+		if ( $query->is_singular() || ! self::is_product_query( $query ) ) {
```
dkoo, line 228: **Nit — Hiding also sweeps related / cross-sell / upsell lists, with no escape hatch**

The hide pass deliberately covers every non-singular product query, which also includes related-products, cross-sells and cart upsells. Hiding an unpurchasable product from those is arguably desired, but there's no filter to let a publisher scope hiding to primary catalog listings if it surprises them. Consider a `newspack_subscriber_only_hide_from_query` escape hatch, or an explicit note in the doc block that these secondary lists are in scope by design.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/subscriber-only/labels.tsx
```diff
@@ -0,0 +1,85 @@
...
+	if ( 'all' === restriction.targeting ) {
+		return __( 'All products', 'newspack-plugin' );
+	}
+	const names = ( restriction.category_ids || [] ).map( nameOf ).filter( Boolean ) as string[];
+	if ( ! names.length ) {
+		return __( 'No categories', 'newspack-plugin' );
+	}
+	return sprintf(
+		/* translators: %s: comma-separated product category names. */
+		__( '%s category', 'newspack-plugin' ),
```
dkoo: **Nit — Category scope label is singular for multiple categories**

`sprintf( __( '%s category', … ), names.join( ', ' ) )` renders "News, Books category" when more than one category is targeted. Use `_n( '%s category', '%s categories', names.length, … )` for the plural (the file already imports `_n`).

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/subscriber-only/use-names.ts
```diff
@@ -0,0 +1,61 @@
...
+	const key = [ ...new Set( ids ) ].sort( ( a, b ) => a - b ).join( ',' );
+
+	useEffect( () => {
+		if ( ! key ) {
+			setNames( {} );
+			return;
+		}
+		let cancelled = false;
+		apiFetch< { id: number; name: string }[] >( {
+			path: addQueryArgs( `${ WIZARD_ENDPOINT }/${ endpoint }`, { include: key, per_page: 100 } ),
```
dkoo: **Nit — Name resolution silently caps at 100 IDs per endpoint**

`per_page: 100` means a restriction covering more than 100 subscriptions/products leaves the overflow unresolved (blank in the Available-to column; the product column collapses to "+N more" so it's mostly hidden). Fine for realistic configs — flagging only because it's a silent truncation.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
```diff
@@ -0,0 +1,501 @@
...
+				if ( isset( $links[ $subscription_id ] ) ) {
+					continue;
+				}
+				$subscription = wc_get_product( $subscription_id );
+				if ( ! $subscription || ! self::can_purchase( $subscription ) ) {
+					continue;
+				}
+				$links[ $subscription_id ] = sprintf(
+					'<a href="%s">%s</a>',
+					esc_url( (string) get_permalink( $subscription_id ) ),
```
dkoo: **Nit — Subscription-variation unlock links may resolve to a non-buyable URL**

`get_subscription_links()` builds each link with `get_permalink( $subscription_id )`. If a restriction is ever unlocked by a subscription **variation** (a `product_variation` post), that permalink points at the variation post rather than WooCommerce's purchasable parent/attribute URL, which can send a reader to a 404 or non-buyable page. Today the picker sources subscription products, so this is edge-case; if variation IDs can reach `subscription_product_ids`, resolve the link through the `WC_Product` permalink instead.

dkoo (approved):
**Approving — no blockers.**

I reviewed the enforcement engine, storage, REST layer, and admin UI, and exercised the feature end to end in an isolated environment: the anonymous, non-subscriber, subscriber, and shop-manager purchase paths; the category→subcategory cascade; the recursion fix; and list-hiding on and off. Every purchase decision and the reader-facing notice behaved as intended, and the admin UI lists and manages restrictions cleanly. The fail-open handling, the once-per-product notice across classic and block themes, and the `pre_get_posts` recursion fix are all carefully done.

A few non-blocking follow-ups are worth a look, each left as an inline comment:
- **Reader-Activation-off behavior** — for parity with #803, enforcement should stand down when Audience Management is off (it currently stays active), with the admin pages remaining available behind a warning. The fix is in the shared subscriber-commerce base, not this PR.
- **REST `active` default** — a partial save can silently re-activate a paused rule (latent today; the UI always sends the field).
- **`post__in` hide-gap** — restricted products still surface in hand-picked product blocks when “hide from lists” is on.

The rest are minor suggestions and nits. Nice work on a careful, well-tested feature.

*Verified end to end in an isolated environment. Release risk: LOW.*

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-only-products-api.php
adekbadek: Good catch — dropped the \`default\` so the sanitizer's fail-safe governs. An absent \`active\` now reaches \`sanitize_base_rule()\` as absent and the rule stays paused. The editor already sends the field explicitly, so the UI is unaffected. Added a REST test that dispatches a payload without \`active\` and asserts the saved rule is paused, plus the contrast case. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Right, and it left a listing that contradicted every other one. The filter now narrows `post__in` instead of adding to `post__not_in` when a curated listing sets it. The empty set is handled explicitly: an emptied `post__in` reads as "no constraint" and would list the whole catalog, so it becomes `[ 0 ]` and the listing comes back empty. Two tests cover it — narrowing, and the all-picks-hidden case. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Folded the grace dimension into the key rather than documenting the invariant — the two layers now agree on what a verdict is a function of, so a future caller that does evaluate inside a gate context gets the right answer instead of tripping over an assumption nothing enforces. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Gated it, so the invariant is the code's rather than WooCommerce's template choice. The block path now only renders on `is_singular( 'product' )` — the same surface the classic `woocommerce_single_product_summary` path covers — so both paths cover exactly one thing and a listing built from these blocks gets nothing.

I first tried the narrower gate of skipping only inside a query loop, which would also have kept the notice on a standalone Single Product block placed on a page. That reads `queryId` from the block's ancestry, and `WP_Block::$available_context` is protected with `__get()` returning null for it, so there is no way to see it from `render_block`. The surface gate is the honest version. The trade is that a Single Product block outside a product page loses the notice — but the classic path never covered that either, so the two are now consistent rather than one being quietly broader.

Two tests: notice present on the product page, absent off it. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Moved the flag below the empty-message guard, so it is set only once something is actually emitted. Fixed in 94bf96c.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/subscriber-only/labels.tsx
adekbadek: Switched to `_n`. Fixed in 94bf96c.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/subscriber-only/use-names.ts
adekbadek: Removed the truncation rather than documenting it: the endpoint caps `per_page` at 100, so the hook now splits the ID set into batches of 100 and merges the responses. An unresolved ID renders as a blank name, so dropping the overflow made the list lie about what a rule covers. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Less edge-case than it looks — the subscriptions picker expands a variable subscription into its variations, and hydration deliberately resolves saved variation IDs on their own, so a variation can reach `subscription_product_ids` through the normal UI today. The link now goes through `$subscription->get_permalink()`, which `WC_Product_Variation` resolves to the parent URL carrying the variation's attributes. Test covers a rule naming a variation, asserting the notice links the parent and not the variation post. Fixed in 94bf96c.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Split out to #909 rather than folded in here. The guard is shared with subscriber discounts (#760), so changing it from this PR would land a discounts behaviour change behind an approval that never covered it.

#909 adds `Reader_Activation::is_enabled()` to `is_enforcement_active()` and leaves `is_admin_available()` untouched, so the admin screens stay reachable and rules can still be authored. Confirmed at runtime rather than only in the harness (which stubs Reader Activation on): with Audience Management off, enforcement returns `false` while admin availability stays `true`.

The wizard half is deliberately not in #909. `AudienceManagementRequired` reads a content-gates-specific global and branches on an `isNewsletter` boolean, so a third consumer needs it generalised first. It also wants deciding per tab rather than for the whole Subscriptions screen: the configuration tab is about WooCommerce subscription products and does not depend on Audience Management. Tracking that separately.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Added `newspack_subscriber_only_hide_from_query`, and the doc block now names related products, cross-sells and cart upsells rather than leaving them implied in "every front-end product listing".

Kept them in scope by default. All three recommend something the reader cannot buy, so hiding them is the coherent behaviour; the filter is there for a publisher who disagrees. Purchase restriction stays untouched either way, so exempting a listing shows the product without selling it. Test covers the exemption and asserts the product is still unpurchasable. Fixed in 9d5a19e.

plugins/newspack-plugin/includes/subscriber-commerce/class-product-purchase-restriction.php
adekbadek: Left this as a per-request memo rather than promoting it to a transient, because the enumeration is not as rules-dependent as it looks.

`Product_Purchase_Restriction::flush_cache()` is called from two places, both in `Subscriber_Only_Products`: rule save and settings update. `covered_product_ids()` also depends on which products exist and which categories they sit in, and nothing hooks product save, publish, or category reassignment. A transient keyed on the rules hash would therefore keep hiding a product after it was moved out of a restricted category, or keep showing one just moved in, until a rule happened to change.

Getting that right means invalidating on product and term changes too, which is more cache surface than an opt-in, 500-bounded enumeration justifies. Worth revisiting if a store ever makes the per-request cost visible.

adekbadek:
@dkoo Re-requesting review – the commits since your approval act on nearly all of your suggestions — narrowing `post__in` so curated listings hide too, pinning the block notice to singular product pages, and a new `newspack_subscriber_only_hide_from_query` escape hatch, among other changes:
- fix(subscriptions): close purchase-restriction gaps from review (NPPD-1899)
- fix(subscriptions): add an escape hatch for hiding from secondary listings (NPPD-1899)

dkoo (approved):
Re-reviewed the two follow-up commits — they cleanly resolve the earlier review, and I found no regressions. The curated-listing `post__in` narrowing (with the `[0]` empty-set sentinel), the payment-recovery-grace cache key, the REST `active` fail-safe, the `is_singular( 'product' )` notice scoping, the `newspack_subscriber_only_hide_from_query` escape hatch, the variation permalink, and the batched name lookups are all correct, and each is backed by a test that pins the behavior — the new REST test dispatches real requests, so it would catch the `active` default if it ever returned.

Approving. Thanks for closing these out.

PR #754, opened 2026-07-28
fix: audit posts_per_page => -1 usages after vipwpcs 3.1.0 (NPPM-3055, #740)

adekbadek:
### All Submissions:

`automattic/vipwpcs 3.1.0` (released 2026-07-27) added `WordPressVIPMinimum.Performance.NoPaging`, which flags `posts_per_page`/`numberposts` set to `-1`. #740 pinned the `release` branch to `~3.0.1` as an interim CI unblock, deferring the real audit to NPPM-3055. A same-day dependabot PR then bumped `main`'s `composer.json` back to `~3.1.0` (#744), so `main` now has the same 66 `NoPaging` errors across 35 files that #740 fixed on `release`.

This does the audit #740 deferred, and re-applies the same interim pin on `main`:

- **62 usages** annotated with `phpcs:ignore WordPressVIPMinimum.Performance.NoPaging...`: queries against small, bounded, admin-configured collections (ad units, gates, templates, prompts, network config, teams, memberships), single-entity-scoped lookups (e.g. corrections for one post), or WP-CLI migration/backfill tooling and PHPUnit fixtures — none of these run against arbitrarily-large, publisher-content-scale data on a live request path.
- **4 usages left unsuppressed**, tracked for a follow-up fix under NPPM-3055 — these look like genuine VIP-relevant gaps rather than false positives:
  - `newspack-plugin/includes/collections/class-query-helper.php:302` — Collections "posts by section", unbounded `post_type => post` query scoped to one taxonomy term (result is cached, but the query itself has no cap).
  - `newspack-story-budget/includes/class-budgets.php:174` — story listing scoped to all *active* (non-archived) budgets when no `$budget_id` is given; budgets aren't time-boxed by the code.
  - `newspack-story-budget/includes/class-api.php:581` and `:935` — REST search callbacks (`get_stories_search`, `get_budget_stories_search`) taking free-text search input with `posts_per_page => -1` and no cap at all.
- `composer.json`: re-pin `automattic/vipwpcs` to `~3.0.1`, matching #740, so CI is green regardless of the 4 open items above.

Also fixes a stale suppression: `group-subscription-settings.php:913` already had a `phpcs:ignore` for the old `WordPress.WP.PostsPerPage...` sniff code, which doesn't cover the new VIP sniff — extended it to cover both.

Part of NPPM-3055.

1. At the workspace root on this branch, remove `vendor/` and `composer.lock`, run `composer install`, and confirm it locks `automattic/vipwpcs (3.0.1)`.
2. Run `composer phpcs -- plugins themes packages config e2e phpcsSniffs` (or just `composer phpcs`) — exits 0.
3. Optional: to confirm the suppressions are correctly scoped (not just masked by the pin), temporarily bump the local `vipwpcs` requirement to `~3.1.0`, reinstall, and re-run `composer phpcs -- --sniffs=WordPressVIPMinimum.Performance.NoPaging` — only the 4 tracked/unsuppressed usages listed above should still report.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? (n/a — lint annotations and a dependency pin, no behavior change)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
 		"php": ">=8.0"
 	},
 	"require-dev": {
-		"automattic/vipwpcs": "~3.1.0",
+		"automattic/vipwpcs": "~3.0.1",
 		"wp-coding-standards/wpcs": "^3.0",
 		"phpcompatibility/phpcompatibility-wp": "^2.1",
 		"dealerdirect/phpcodesniffer-composer-installer": "^1.0",
--- a/plugins/newspack-ads/includes/providers/gam/class-gam-model.php
+++ b/plugins/newspack-ads/includes/providers/gam/class-gam-model.php
@@ -418,7 +418,7 @@ final class GAM_Model {
 		$query           = new \WP_Query(
 			[
 				'post_type'      => self::$custom_post_type,
-				'posts_per_page' => -1,
+				'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging.posts_per_page_posts_per_page
 				'post_status'    => [ 'publish' ],
 			]
 		);
--- a/plugins/newspack-network/includes/cli/backfillers/class-membership-plan-updated.php
+++ b/plugins/newspack-network/includes/cli/backfillers/class-membership-plan-updated.php
@@ -45,7 +45,7 @@ class Membership_Plan_Updated extends Abstract_Backfiller {
 			[
 				'post_type'   => Memberships_Admin::MEMBERSHIP_PLANS_CPT,
 				'post_status' => 'any',
-				'numberposts' => -1,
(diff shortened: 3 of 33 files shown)


adekbadek:
Superseded by #741, which is the assignee's own resolution of NPPM-3055 and is already in review with green CI.

#741 is the better change on every axis: it covers all 66 occurrences (this PR covered 62, deferring 4), it fixes rather than suppresses the one genuinely-unbounded query (the Campaigns donor-landing-page dropdown, now capped at 1000), it clears the new `AlwaysReturnInFilter` and `FetchingRemoteData` warnings too, and — critically — it *unpins* `automattic/vipwpcs` back to `^3.0` instead of re-pinning, so the interim `~3.0.1` pin on `release` gets carried away by promotion rather than needing a separate revert.

Closing to avoid two competing PRs against the same lines.

PR #756, opened 2026-07-28
fix(access-control): identify products by ID in gate access-rule pickers (NPPD-2129)

adekbadek:
### All Submissions:

A gate's Paid Access → **Active subscription** list rendered selected products as name-only chips, and sites routinely reuse names like "Annual" across legacy and current tiers. Tokens now render as `<name> (#<id>)` and the field can be searched by ID.

**Before** — three products called "Annual", nothing to tell them apart:

Before: five chips reading Supporter - Annual, Monthly, Annual, Annual, Annual - the three Annual products are indistinguishable

**After** — each chip carries its product ID:

After: each chip reads name plus product ID, e.g. Annual (#51)

Typing an ID narrows the list to that one product:

After: typing a product ID narrows the suggestions to that one product

**This also fixes a data-corruption bug.** Tokens were resolved back to products by matching the label string, so removing one "Annual" chip re-selected every same-named sibling. Reproduced locally with three same-named "Annual" products: removing one chip from a 5-product selection produced a 6-chip list whose pre-save summary read `Supporter - Annual, Supporter - Annual, Monthly, Monthly` — all three "Annual" products dropped, two never-selected products added, no visible error.

That pre-save summary, on `main` — this is what would have been written to the gate:

Before: after removing one Annual chip, the pre-save summary shows a completely different product set

This is the likely explanation for the chip-count discrepancy in the ticket (23 chips against a 19-product source plan) — the picker was inventing selections rather than the migration carrying over extras.

Also in this PR:

* **Values the option list cannot describe are no longer silently deleted.** A stored ID with no matching option previously rendered as nothing and was dropped on the next save. It now renders as `(product not listed) (#<id>)` and survives an unrelated edit. The wording is "not listed" rather than "gone" because an option list holds parent products and published institutions while evaluation also resolves variation IDs, so such a value is often still granting access. Every picker says so alongside it, since removing the entry widens the gate — and removing a rule's last value widens it further, an empty value list applying no filter at all. For the product rules the ID can be typed back in, so acting on the caution is reversible. The underlying gaps are tracked in NPPD-2135.

  After: a stored product that no longer exists renders as product unavailable with its ID

* **The same label-matching bug in the block editor's Gates field** (`GateControls`) is fixed with the same helper — two gates sharing a title selected and deselected together.
* Free text that names no option is rejected at entry instead of becoming a token that disappears on the next render. Whether a bare ID the option list does not describe is accepted depends on the rule: an "Active subscription" rule matches a variation ID, so it is; the **Gates** and **Institutional access** lists are the complete set their server-side test resolves from, so it is not — a typed gate ID that is not a published gate would store a gate that matches nobody, and a block whose gates all match nobody is shown to everyone. The rejection wording is per rule and only offers "type its ID" where the field keeps that promise. `FormTokenField` renders nothing on a rejection, so the rejection is announced, and the highlighted suggestion is auto-selected — typing a product name and pressing Enter now selects that product.
* The gate summary (list card + pre-save panel) reads the same option list the pickers do — institutions are created and deleted in the same app, so a summary built from the page-load snapshot would call one added a moment ago "not listed" — and shows `<name> (#<id>)` on every rule, including **One-time purchase**, which is what makes it usable for comparing a gate against its source Membership Plan during migration verification.
* The **Institutional access** list is no longer truncated at 100. Both pickers fetch it and replace the server-rendered list, so a site past 100 institutions lost the overflow — those values then read as unresolvable and could not be re-added. The request now pages through the whole collection, once per session rather than once per mount: the list is shared by every reader and refetched only after an institution is created, edited or deleted.

The token shape mirrors the existing convention in the sibling content-rule picker (`content-rule-control-tokenfield.tsx`), which already carries an identifier per token.

Closes NPPD-2129.

1. On a site with WooCommerce Subscriptions and `NEWSPACK_CONTENT_GATES` enabled, create **two or more subscription products with the same name** (e.g. three called "Annual").
2. Go to **Audience → Access Control**, add a gate, enable **Paid Access → Active subscription**.
3. Open the product picker. Each suggestion and chip reads `Annual (#49)`, `Annual (#50)`, … — confirm you can tell them apart.
4. Type a product **ID** into the field; confirm the list narrows to that one product.
5. Select all three "Annual" products, then remove one chip. **Confirm only that product is removed** and the other two stay. (On `main`, this instead produces extra chips and drops the others.)
6. Save; confirm the gate list card and the pre-save panel list the products with their IDs.
7. Type something that isn't a product and press Enter; confirm it is rejected rather than added.
8. Repeat 3–5 in the block editor: select a Group block → **Access Control** → **Custom** → **Active subscription**. The **Gates** field in that same panel also shows `<title> (#<id>)`.

To check the not-listed path: point a saved gate's rule at a product ID that is not in the list via WP-CLI. Reload the editor. Confirm the chip reads `(product not listed) (#<id>)` and the field cautions that the value is still evaluated. Remove an unrelated chip. Confirm the ID is still there. Remove the not-listed chip, type the same ID back in, and confirm it returns.

### Other information:

**Verification**

* `n test-js` — **111 suites, 1206 tests, all passing**. Eight new suites: `access-rule-options.test.js` (the helper, incl. the removal-corruption regression, ID-suffix parsing, and the preserve-a-stored-value behaviour), `access-rule-option-sources.test.js` (one fetch per rule, and the invalidation after a write), `use-access-rule-options.test.jsx` (one notice per failed fetch, however many readers share it), `access-rule-control.test.jsx` (the wizard control), `block-visibility-panel.test.jsx` (the block editor's copy of both pickers, incl. the integer coercion on gate IDs), `gate-summary-access-rule-values.test.jsx` (every rule's value formatting), `unlisted-values-notice.test.tsx` (the caution) and `institutions/edit-invalidation.test.jsx` (a write dropping the fetched list). The component tests were confirmed failing against unmodified code first.
* `n test-php` — 3688 tests, 6 pre-existing skips, no failures. No PHP changed in this PR.
* `wp-scripts lint-js` on the touched files — 0 errors (pre-existing `no-explicit-any` warnings on untouched lines).
* `tsc --noEmit` — no new type errors in the touched files. (This repo has a large pre-existing tsc backlog.)
* Browser-verified in an isolated env against both surfaces (Audience wizard and block editor), including the before/after contrast on the corruption bug.

**Notes for review**

* The ` (#<id>)` suffix is a parsing contract as well as display copy — `resolveAccessRuleOptionTokens` reads the ID back out of it — so its shape is deliberately not translatable. The exact label is matched before the suffix pattern, so a product whose own name ends in something ID-shaped still resolves correctly (covered by a test).
* The new helper lives in `src/content-gate/` and is imported by both the audience wizard and the block-editor entry. Verified it is **not** hoisted into the `commons` chunk: the wizard consumers are async chunks, so `minChunks: 2` with `chunks: 'initial'` is never met, and `content-gate-block-visibility.asset.php` carries `wp-html-entities` in its own dependency array.
* Copilot review was not requested — we're out of quota.

https://claude.ai/code/session_016sdTMBsS36LY9yLMGqQdeK


adekbadek:
Note on the red `PHPCS / newspack` check: **it is not caused by this PR.**

This branch changes six JS/TS files and zero PHP files:

```
$ git diff main...HEAD --name-only | grep -c "\.php$"
0
```

The failures are `WordPressVIPMinimum.Performance.NoPaging.posts_per_page` errors in untouched files (`includes/class-donations.php:329`, `tests/unit-tests/content-gate/content-gates.php:3016`, and others). They arrive from the `automattic/vipwpcs` 3.1.0 bump in #751, which this branch is based on — a full local `phpcs` run on this branch reports **66 errors across 41 files**, all in code this PR never touches, and `includes/class-donations.php` is byte-identical to `main`.

Already tracked and being fixed in #741 (`chore: address vipwpcs 3.1.0 posts_per_page sniff errors`, NPPM-3055). This PR should go green once that lands and this branch is rebased.

The checks that actually cover this change are passing: **JS / newspack** ✅ (55 suites, 496 tests), **PHPUnit / newspack** ✅, **Build ZIP** ✅.

.github/dependabot.yml
```diff
@@ -22,6 +22,17 @@ updates:
...
     labels:
       - "dependencies"
+    ignore:
+      # VIPCS 3.1.0 makes `posts_per_page => -1` a hard error, which the
+      # monorepo does not yet satisfy, so the root constraint stays on 3.0.x
+      # until that audit lands (NPPM-3055). The pin alone is not enough:
+      # Dependabot also opens "permit the latest version" PRs that rewrite the
+      # constraint itself, and one of those reverted this pin and turned every
+      # PHP-linted PR red. Drop this entry as part of the audit.
+      - dependency-name: "automattic/vipwpcs"
```
jason10lee: **Suggestion:** The ignore covers the root ecosystem only, and auto-merge fills the gap — This entry is attached to the root composer ecosystem. The per-plugin composer ecosystems have no equivalent, so Dependabot keeps opening vipwpcs bumps against those lockfiles — and `.github/workflows/auto-merge.yml` auto-approves and auto-merges minor/patch Dependabot PRs, while CI's php-lint job never touches a plugin vendor. Such a PR is green and lands unreviewed.

That's the same "permit the latest version" mechanism the comment above describes, one directory over. Relatedly, the pin is root-only in effect: `plugins/newspack-plugin/composer.lock` already locks 3.1.0 and `package.json` defines `"lint:php": "./vendor/bin/phpcs"` against that vendor, so the plugin's own documented lint command is stricter than CI right now. Mirroring the constraint into the per-plugin `composer.json`s would make one number govern every entry point.

composer.json
```diff
@@ -7,7 +7,7 @@
 		"php": ">=8.0"
 	},
 	"require-dev": {
-		"automattic/vipwpcs": "~3.1.0",
+		"automattic/vipwpcs": "~3.0.1",
```
jason10lee: **Suggestion:** Worth shipping the vipwpcs pin as its own PR — The pin and its Dependabot ignore look right, and the comment is unusually good — it names the failure mode, the follow-up ticket, and when to delete the entry. I checked the rationale: 62 `posts_per_page => -1` occurrences exist under `plugins/`, 29 with no `phpcs:ignore`, so 3.1.0's stricter sniff would hard-error.

The reason to split it is urgency mismatch. This unblocks PHP lint for the whole repo, and it's riding on a Low-priority UI fix that will sit through review rounds on token-format questions. A two-line `chore(deps)` PR lands today; this one lands when the picker discussion resolves. It also helps whoever later bisects a vipwpcs change — they'd want to land on a commit that mentions vipwpcs. Worth noting the commit type too: `fix(access-control)` makes semantic-release treat this as a publisher-visible patch, which the tooling half isn't.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * The ID a token carries, e.g. `188250` in `Annual (#188250)`.
+ */
+const TOKEN_ID_PATTERN = /\(#([^)]+)\)\s*$/;
+
+/**
+ * Stand-in name for a stored value the option list can no longer describe. Keyed by rule
+ * slug so the wording names the right kind of thing.
+ */
+const MISSING_OPTION_LABELS: Record< string, () => string > = {
```
jason10lee, line 69: **Suggestion:** `(product unavailable)` says something false about a stored variation id — Both option builders list parent products only — `get_subscription_products_options()` queries `[ 'subscription', 'variable-subscription' ]` and the one-time sibling queries `[ 'simple', 'variable' ]` — but evaluation resolves variations: `WC_Subscription::has_product()` checks `variation_id`, and `wc_customer_bought_product()` matches `_variation_id` order-item meta. So a variation id is a valid, access-granting value that this label calls unavailable.

Both things a publisher does with that message are wrong: delete the token, silently narrowing the gate, or file a bug. A wording that distinguishes "not in this list" from "gone" would avoid pre-judging what NPPD-2135 is separately fixing. I confirmed the option-list half directly; the evaluation half is from WooCommerce/WCS behaviour rather than something I could exercise here, since these envs have no licensed WooCommerce.

Smaller point on the same map: the fallback returns the product wording for any unrecognised slug, and `Access_Rules::register_rule()` is public, so a third-party options-backed rule would report `(product unavailable)` for something that isn't a product.

**Why this is more than wording.** I confirmed in a live env on this branch that deleting the last value from a rule leaves an empty list, and an empty list reads as "no constraint": `evaluate_rules()` returns `true` for an anonymous visitor and `User_Gate_Access::evaluate_gate_for_user()` reports `can_bypass=true`. So the repair this label invites — remove the stale-looking entry — widens the gate rather than narrowing it, and if it was the rule's only value the gate stops restricting entirely.

**Risk if not addressed:** If the removed entry was the rule's only value, an institution rule with an empty list returns true for every visitor including logged-out ones, and a subscription rule with an empty list matches any active subscription — so the gate stops restricting the content it protects, and the same evaluation reports every registered reader as paid to the mobile-app entitlement endpoint.

**Consider:** Making an unresolvable value non-destructive to remove would defuse this — either keep the token but block deleting the last value in a rule, or warn on save when a rule's value list goes from non-empty to empty, since every consumer downstream reads an empty list as "grant access".

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * Build the token label for an option: `<name> (#<id>)`.
+ *
+ * @param option The rule option.
+ *
+ * @return The token label.
+ */
+export function formatAccessRuleOptionLabel( option: AccessRuleOption ): string {
```
jason10lee, line 115: **Suggestion:** This makes a third token format in one subsystem, and forks one the sibling treats as translatable — The docblock's reasoning for fixing the `(#<id>)` shape is sound — it's parsed back out, so it can't move. The wrinkle is that `one-time-purchase-rule-control.tsx:85-90` builds the same visual token through `sprintf( __( '%1$s (#%2$s)' ) )`, which *is* translatable and is an existing msgid. A locale that renders it `%1$s [n° %2$s]` gets two differently-shaped tokens side by side in the same panel, and a uniquely-named one-time product shows no id at all while a uniquely-named subscription product shows one.

With the content-rule field's `188250: Annual` that's three conventions in the gate editor. Lifting the one-time control onto this helper would settle it — that control is in some ways more careful (it only disambiguates names that actually repeat), so it may be worth merging its logic in rather than replacing it.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.test.jsx
```diff
@@ -0,0 +1,110 @@
+/**
```
jason10lee, line 1: **Suggestion:** Good coverage of the bug; three surfaces still unpinned — The corruption bug itself is pinned properly in both suites — three same-named products, remove the middle token, assert the other two ids survive — and the helper suite also covers the `(#12)`-inside-a-name trap, order preservation, dedup, and both directions of the dangling-value rule. That's the part that makes this fix durable, and it's done well.

What would still regress silently: `block-visibility.tsx` has no component test, so its copy of the same fix is covered only transitively (including the `.map( Number )` coercion, which is a no-op today but nothing pins that); `gate-summary.tsx` has no test at all, which is likely why the one-time branch was missed; and nothing asserts the three surfaces agree on token format, which is the divergence worth guarding given the subsystem's history.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
```diff
@@ -85,12 +92,13 @@ export default function AccessRuleControl( { slug, value, onChange }: GateRuleCo
...
-					.map( o => o.label ) }
-				onChange={ ( items: string[] ) => onChange( options?.filter( o => items.includes( o.label ) ).map( o => o.value ) ?? [] ) }
-				suggestions={ options.map( o => o.label ) }
+				hideLabelFromVision
+				label={ rule.name }
+				description={ __( 'Search by name or ID.', 'newspack-plugin' ) }
+				value={ getAccessRuleOptionTokens( options, value, getMissingOptionLabel( slug ) ) }
+				onChange={ ( tokens: ( string | TokenItem )[] ) => onChange( resolveAccessRuleOptionTokens( tokens, options, value ) ) }
+				suggestions={ options.map( formatAccessRuleOptionLabel ) }
+				__experimentalValidateInput={ ( input: string ) => isAccessRuleOptionInput( input, options ) }
```
jason10lee: **Suggestion:** Rejected input is announced but not rendered, so the PR description overstates it — `__experimentalValidateInput` is the right guard, but in `@wordpress/components` 30.9.0 its only failure path is `speak( messages.__experimentalInvalid, 'assertive' )` — an aria-live announcement of the generic string "Invalid item", with nothing rendered (`form-token-field/index.js:334-336`, confirmed in the installed copy). And `__experimentalAutoSelectFirstMatch` defaults to `false`, so Enter on a typed `Annual` doesn't commit the highlighted suggestion; it falls through to validation and is dropped.

Net effect for a sighted user: nothing happens at all. The description says free text is "rejected inline", which reads as visible feedback. Passing `__experimentalAutoSelectFirstMatch` would make Enter commit the match, and a custom `messages.__experimentalInvalid` naming what the field wants would make the announcement actionable. `one-time-purchase-rule-control.tsx:146` has the same gap, so this is consistency rather than a new defect.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
```diff
@@ -85,12 +92,13 @@ export default function AccessRuleControl( { slug, value, onChange }: GateRuleCo
...
 				__experimentalExpandOnFocus
```
jason10lee: **Suggestion:** The institution list is capped at 100 and replaces the uncapped server list — Both pickers fetch `/wp/v2/np_institution?per_page=100&context=edit` and then *replace* the PHP-provided options rather than merging into them, while `Institution::get_options()` has no cap. On a site past 100 institutions the fetch discards options the initial render had, and those now read `(institution unavailable)` — indistinguishable from genuinely deleted, and unre-addable, since suggestions come from the same truncated list. 100 is the REST hard maximum for `per_page`, so this needs pagination or a search-driven endpoint.

Worth saying that this PR still improves the situation: those values used to render as nothing and be dropped on the next save, and now they survive. It's the label that's wrong, not the preservation. But it feeds the same removal path as the variation case above, which I verified widens the gate.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/gate-summary.tsx
```diff
@@ -62,7 +68,16 @@ const formatAccessRuleValue = ( rule: GateAccessRule ): string => {
 		);
 	}
 	if ( Array.isArray( rule.value ) && config?.options ) {
-		return getOptionLabels( rule.value, config.options );
+		// Names repeat across product and institution tiers, so each value is named
+		// alongside the ID it stores — the same identification the pickers give.
+		return rule.value
+			.map( value => {
+				const option = findAccessRuleOption( config.options ?? [], value );
```
jason10lee: **Suggestion:** The summary fix stops one rule short of the rule most likely to show the symptom — This branch resolves by id and formats `<name> (#<id>)`, but the `one_time_purchase` branch above it still calls the old name-based `getOptionLabels`, which maps to `option.label` alone and falls back to a bare `String(value)`. One-time products come from the same catalogue where names repeat, so the pre-save panel can read `Active subscription: Annual (#188250), Annual (#205482)` directly above `One-time purchase: Annual, Annual` — the ambiguity this PR removes, one line down, on the surface a publisher uses to audit a gate.

`getOptionLabels` also doesn't `decodeEntities`, so a product named `Founder&#8217;s Club` renders decoded on one line and escaped on the other. Routing the one-time list through `findAccessRuleOption` / `formatAccessRuleOptionLabel` looks like it closes both.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @return The selected option values, deduplicated, in token order.
+ */
+export function resolveAccessRuleOptionTokens(
+	tokens: ( string | TokenItem )[],
+	options: AccessRuleOption[],
+	stored: unknown = []
+): ( string | number )[] {
+	const byLabel = new Map( options.map( option => [ formatAccessRuleOptionLabel( option ), option.value ] ) );
+	const storedValues = Array.isArray( stored ) ? stored : [];
```
jason10lee: **Nit:** `Array.isArray` on `unknown` widens the resolved value to `any` — `stored: unknown` narrowed by `Array.isArray` becomes `any[]`, so the `find` result is `any` and flows unchecked through the declared `( string | number )[]` return. I couldn't reach a bad value — `sanitize_access_rule()` maps option-rule values through `intval`/`sanitize_text_field`, so what's persisted is scalars, and `String( {} ) === id` wouldn't match anyway. Typing it `stored: readonly ( string | number )[] = []` would drop the `any` without changing behaviour; all three call sites already pass arrays.

jason10lee (approved):
Approved with 8 suggestions and 1 nit.  The fix itself is good, and fixes the broader bug: resolving tokens by label string meant removing one chip re-selected every same-named sibling and rewrote which products a gate covered. Moving to id-carrying tokens fixes that at the right layer, and the new tests pin the real bug — including the case where a product's own name ends in `(#12)` — rather than the happy path. Preserving unresolvable stored ids closes a second silent-data-loss path.

Release risk: ELEVATED — no blockers and no contract changes, but two new labels can lead a publisher to remove a working value, and an emptied access rule grants access rather than denying it.

Two things to settle before merge. `(product unavailable)` is shown for stored variation ids, which still grant access, and `(institution unavailable)` for every institution past the hundredth, because the picker replaces the uncapped server list with a capped fetch. I confirmed at runtime that deleting the last value from a rule leaves it empty, and an empty rule returns true for anonymous visitors — so the repair those labels invite widens the gate. Distinguishing "not in this list" from "gone" would defuse both.

Separately, the gate summary's one-time-purchase branch still renders bare names on the same panel as the fixed rows, and the subsystem now carries three token formats where it had two. And the vipwpcs pin, though correct and well commented, is repo-wide CI plumbing riding on a Low-priority UI fix — it would land faster and bisect better on its own.

Including variations in the subscription option list, or resolving a stored id against the product itself before declaring it unavailable, would mean the label only appears when the value genuinely no longer resolves.

Merging the fetched institution page into the server-provided list rather than replacing it, or paging the request to completion, would stop valid entries being labelled unavailable on larger sites.

.github/dependabot.yml
adekbadek: Mirrored both halves into the four per-plugin composer ecosystems in 81c4062: each now carries the same `ignore` entry, and each `composer.json` is pinned to `~3.0.1` with its lockfile regenerated (3.1.0 → 3.0.1, one package per lock). That closes the auto-merge path you named and makes the documented per-plugin `lint:php` agree with CI instead of being stricter than it.

The root comment now says every composer ecosystem repeats the entry and why, so the whole thing lifts together with the NPPM-3055 audit.

The ten other manifests carrying `automattic/vipwpcs: ^3.0` are untouched — none of them is a Dependabot ecosystem, so nothing moves their lockfiles on its own, and regenerating ten more locks here would bury the change this PR is actually about. Worth folding into the audit.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: You are right that the label asserted something false. Reworded to `(product not listed)` / `(institution not listed)` / `(gate not listed)`, and the unknown-slug fallback is now `(not listed)` rather than the product wording, since `register_rule()` is public and a third-party rule need not hold products. `one_time_purchase` got its own entry so it reads as a product rather than falling through.

On the removal path: the picker now renders a caution whenever a rule holds a value no option describes — "Entries marked “not listed” are not in this list — a product variation, or an item that was deleted or unpublished. They are still checked when access is evaluated, so removing one widens who this gate lets in." Both surfaces show it. That defuses the repair the token was inviting.

What I did not change is the evaluation semantics, and I want to flag that rather than leave it implied. Empty means different things per rule: `has_active_subscription()` with no product filter still requires an active subscription, so an empty subscription rule widens but continues to restrict, while `Institution::evaluate()` returns `true` on an empty list and so grants everyone including anonymous readers. Making an empty list fail closed would flip live gates from open to closed on sites that already have one, which is a product call rather than something the code answers. It wants its own ticket.

Also noted `NPPD-2135` covers the variation/one-time gaps separately.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.test.jsx
adekbadek: All three surfaces pinned in 81c4062.

`block-visibility-panel.test.jsx` is new: it captures the `editor.BlockEdit` filter and renders the real panel, so both of its pickers are covered directly. Removing one of two same-named gate tokens keeps the other; a gate ID stored as `"99"` with no matching option survives the edit and comes back as the integer `99`, which is what pins the `.map( Number )`. Confirmed the coercion test fails without it.

`gate-summary.test.jsx` is new too, and covers what the missed one-time branch would have caught: a gate carrying both a subscription rule and a one-time rule over the same catalogue, asserted line by line, plus entity decoding and a value no option describes.

For agreement across surfaces, the summary test asserts against `formatAccessRuleOptionLabel()` itself rather than a hardcoded string, so a change to the token shape has to move all three together or that test goes red.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
adekbadek: Both fixed, and the description corrected — "rejected inline" was overstating it.

`__experimentalAutoSelectFirstMatch` is now on all four fields. That is the half that actually mattered: typing `Annual` and pressing Enter used to fall through to the validator and vanish, and now commits the highlighted suggestion. Verified in the browser on this branch — typing `Annual` highlights `Annual (#83)` and Enter adds it — and there is a test that fails without the prop.

`messages.__experimentalInvalid` now names what the field wants ("Not a selectable option. Pick one from the list, or type its ID.", with a gate- and product-specific variant on the two fields that need one). The component takes `messages` as a whole object with no per-key merge, so a partial override would have silently dropped the remove button’s label; the four keys come from one shared helper, and a test asserts all four are present.

`one-time-purchase-rule-control.tsx` got the same treatment. Even so, a rejection still renders nothing visible — that is the component’s behaviour, not something these props change — so the description now says the rejection is announced rather than shown.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
adekbadek: Fixed by removing the cap rather than raising it. Both surfaces now share one option source that asks for `per_page=-1`, which api-fetch’s fetch-all middleware turns into a paged walk of the whole collection, so replacing the localised list is safe again. The request is also ordered by title to match `Institution::get_options()`, so the picker reads the same way whichever list rendered it.

Verified on this branch with 105 institutions published: `Inst 105`, which sits past the first page, is suggested and selectable. Before the change it was absent from the list entirely.

That also removed a duplicated `DYNAMIC_OPTION_RULES` map — the wizard and the block editor each had their own copy of the same path.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/gate-summary.tsx
adekbadek: Fixed — both branches now go through one `formatAccessRuleOptionValues()` helper built on `findAccessRuleOption` / `formatAccessRuleOptionLabel`, so the one-time list gets the ID, the `decodeEntities` and the "not listed" wording that the other branch had. `getOptionLabels` had no other caller and is gone.

`gate-summary.test.jsx` covers it with exactly the case you described: one gate carrying both rules over the same catalogue, asserted line by line, plus a `Founder&#8217;s Club` product to pin the entity decoding.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Done — `stored: readonly ( string | number )[] = []`, and the `Array.isArray` narrowing is gone with it, so the `find` result is typed and the declared return type actually holds. The two call sites that had a union in hand now narrow once into a local and pass it to both `getAccessRuleOptionTokens` and `resolveAccessRuleOptionTokens`. No behaviour change; the helper suite is unchanged and green.

composer.json
adekbadek: Dropped rather than split — the pin's premise has expired. #741 landed the NPPM-3055 audit, so `main` already runs vipwpcs 3.1.0 with the `NoPaging` sniff satisfied, and PHPCS is green there over whole packages resolved from the root manifest.

The composer manifests, the four lockfiles and the Dependabot ignores are back to `main`'s state in 8068ad841, which also resolves the only merge conflict (`plugins/newspack-newsletters/composer.lock` — the pin colliding with itself). What is left is the twelve UI and test files.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Converged on the shared helper, in 860a972a4. `OneTimePurchaseRuleControl` now uses `getAccessRuleOptionTokens` / `resolveAccessRuleOptionTokens` / `isAccessRuleOptionInput`, so both pickers and the pre-save summary read `<name> (#<id>)`, untranslatable, with no second msgid.

The direction went the other way from your suggestion — always show the ID rather than only on collision. A uniquely-named product showing no ID is the complaint this ticket opens with, and the token is a parsing contract, so a locale that reshapes the suffix would break the ID fallback in `resolveAccessRuleOptionTokens`.

The bespoke bijection goes with it: every option now carries its own ID, so no two tokens can collide and the counter has nothing left to guard. Two changes come free — a stored ID the option list cannot describe reads `(product not listed) (#<id>)` instead of a bare `#<id>`, matching the other rules, and names holding HTML entities are decoded. The token tests move to `access-rule-options.test.js`, which already covers name collisions, an ID embedded in a name, and preserving an unlisted stored value.

Verified on a local env: both pickers list `Day Pass (#300)` / `Day Pass (#301)` distinctly, the gate saves `product_ids: [301]` and `subscription: [84]`, and the tokens round-trip on reload.

Gate editor Paid Access panel: Active subscription and One-time purchase pickers both showing name (#id) tokens

adekbadek:
@jason10lee Re-requesting review – the commits since your approval add a "not listed" caution notice under the pickers, uncapped institution fetching via a new option-source module, and `__experimentalAutoSelectFirstMatch` with custom field messages — well past the nits you signed off on:
- fix(access-control): distinguish unlisted rule values from deleted ones (NPPD-2129, #756)
- fix(access-control): use one token format across the gate editor's pickers (NPPD-2129, #756)

plugins/newspack-plugin/src/content-gate/access-rule-option-sources.ts
```diff
@@ -0,0 +1,49 @@
...
+type InstitutionItem = { id: number; title: { raw: string } };
+
+/**
+ * Rules whose options are fetched rather than localised, keyed by rule slug.
+ */
+const ACCESS_RULE_OPTION_SOURCES: Record< string, () => Promise< AccessRuleOption[] > > = {
+	institution: async () => {
+		// Ordered by title to match `Institution::get_options()`, so the picker reads the
+		// same way whichever list rendered it.
+		const items = await apiFetch< InstitutionItem[] >( {
```
jason10lee, line 43: **Suggestion:** `Access_Rules::get_access_rules_for_client()` already resolves `Institution::get_options()` into `available_access_rules` for both surfaces, and that helper queries with `posts_per_page => -1`, publish, ordered by title — the same list, in the same order. We measured it in an env: localised option count `3`, published institutions `3`.

So this buys freshness inside an SPA session rather than completeness. Against that it costs `ceil(N/100)` sequential round-trips, unmemoized, on every mount — and in the block editor the control remounts on each block selection, so selecting five blocks with an institution rule walks the collection five times.

`per_page=-1` itself is fine, for the record: api-fetch rewrites it to 100 and walks the Link headers before the request leaves the browser. The cost is the round-trips, not a rejected parameter.

Dropping the source for `institution` would remove the pagination question, the remount cost, and the summary inconsistency in the next comment, all at once. If the freshness is wanted, memoizing the promise at module scope with an explicit invalidation the institutions view calls after a write would get it for one walk instead of N.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/gate-summary.tsx
```diff
@@ -12,17 +12,32 @@ import { __, _n, sprintf } from '@wordpress/i18n';
...
 /**
- * Map option values to labels, falling back to the raw value.
+ * Name each stored value alongside the ID it stores — the same identification the
+ * pickers give, so a gate reads the same way wherever it is inspected. Names repeat
+ * across product and institution tiers, and a value the option list cannot describe is
+ * named rather than printed bare.
  */
-const getOptionLabels = ( values: Array< string | number >, options: { value: string | number; label: string }[] = [] ) =>
-	values.map( value => options.find( option => String( option.value ) === String( value ) )?.label ?? String( value ) ).join( ', ' );
+const formatAccessRuleOptionValues = ( values: Array< string | number >, options: AccessRuleOption[] = [], slug: string ) =>
```
jason10lee, line 35: **Suggestion:** `availableAccessRules` is a module-load snapshot of `window.newspackAudienceContentGates`, while the pickers now refetch institutions — and institutions are created and deleted inside the same SPA.

So immediately after adding an institution and attaching it to a gate, the picker shows its real name and this card shows `(institution not listed) (#123)`. That reading isn't neutral: `getUnlistedAccessRuleValuesNotice()` tells publishers "not listed" means deleted or unpublished, and that removing it widens the gate. Before this change the summary printed the bare ID, which claimed nothing.

Either pass the fetched options into `getGateSummarySections`, or keep the bare ID on the surface whose list isn't authoritative. Dropping the fetch (previous comment) also resolves it, since both would read the same snapshot again.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
```diff
@@ -83,14 +74,21 @@ export default function AccessRuleControl( { slug, value, onChange }: GateRuleCo
 		return <OneTimePurchaseRuleControl value={ value } onChange={ onChange } options={ options } TokenField={ FormTokenField } />;
 	}
 	if ( options && options.length > 0 ) {
+		const selected = Array.isArray( value ) ? value : [];
+		const description = hasUnlistedAccessRuleValues( options, selected )
```
jason10lee: **Suggestion:** Both surfaces return `OneTimePurchaseRuleControl` before reaching the `hasUnlistedAccessRuleValues` branch, so the one-time-purchase picker renders `(product not listed) (#188250)` with no caution attached.

The rationale in `access-rule-options.ts` names the one-time-purchase `_variation_id` case as the reason the caution exists — a one-time purchase matches the `_variation_id` on an order item — which makes that rule the likeliest to hold an ID the option list can't describe, and the one picker that never explains it. A publisher auditing that rule gets exactly the reading the notice was written to prevent: stale entry, safe to delete. Acting on it widens the gate.

The summary path is fine, since it's read-only. It's the two editable pickers that want the notice.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
```diff
@@ -166,50 +175,52 @@ const AccessRuleValueControl = ( {
...
 			return;
 		}
 		let cancelled = false;
-		apiFetch< DynamicOptionItem[] >( { path: dynamicConfig.path } )
-			.then( items => {
+		source()
+			.then( fetched => {
 				if ( ! cancelled ) {
-					setOptions( items.map( dynamicConfig.mapItem ) );
+					setOptions( fetched );
```
jason10lee, line 200: **Suggestion:** `setOptions( fetched )` overwrites the localised list whatever comes back. A response of `[]` — a filtered query, a plugin narrowing the collection, a route returning empty for a reason other than "there are none" — drops the picker to the free-text `TextControl` fallback further down, whose `onChange` writes a string over a stored ID array.

The `.catch( () => {} )` below is the right *behaviour* on failure, since the localised list is complete and is the better fallback. But it's silent, where the wizard raises a notice for the same condition — so a publisher can end up editing against a snapshot with no signal it may be stale. A console warning or an inline note would close the asymmetry.

Both go away if the fetch does.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,240 @@
...
+ * `getAccessRuleTokenFieldMessages()` — its wording is the only account of what happened
+ * — and with `__experimentalAutoSelectFirstMatch`, which commits the highlighted
+ * suggestion so typing a name and pressing Enter selects rather than being rejected.
+ *
+ * @param input   The typed input.
+ * @param options The available rule options.
+ *
+ * @return Whether the input resolves.
+ */
+export function isAccessRuleOptionInput( input: string, options: AccessRuleOption[] ): boolean {
```
jason10lee: **Suggestion:** `isAccessRuleOptionInput` resolves against `options` only, never `stored`, so an ID absent from the option list is rejected at entry — while the rejection message reads `Not a selectable option. Pick one from the list, or type its ID.`

Typing an ID is therefore only ever equivalent to picking one from the list, and the variation ID the new caution tells publishers is legitimate is exactly the ID that can't be typed. The concrete path: a publisher sees `(product not listed) (#188251)`, reads the notice, removes the token to see what happens, and can't put it back — the moment it leaves `value` it also leaves `stored`.

This is the pre-PR behaviour, so it isn't a regression. It reads differently now because the PR adds copy that points publishers at those IDs. Either narrow the wording, or accept a digits-only input outright, since the rule already stores IDs the option list can't describe.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,240 @@
...
+	return resolved;
+}
+
+/**
+ * Whether an input typed into the field names a selectable option.
+ *
+ * Used as `FormTokenField`'s input validator, which drops anything it rejects rather than
+ * letting it become a token that silently disappears on the next render. The field
+ * announces the rejection to assistive technology and renders nothing, so pair it with
+ * `getAccessRuleTokenFieldMessages()` — its wording is the only account of what happened
```
jason10lee, line 222: **Nit:** The docblock says rejected input is dropped "rather than letting it become a token that silently disappears on the next render". `FormTokenField`'s paste path calls `addNewTokens` directly and never consults `__experimentalValidateInput`, so pasting `Annual, Monthly` creates tokens that resolve to nothing and vanish on the next render, unannounced.

Pasted bare IDs resolve fine, so this only bites on names — but the docblock currently promises more than the field delivers.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
```diff
@@ -166,50 +175,52 @@ const AccessRuleValueControl = ( {
...
+					onChange={ ( tokens: ( string | TokenItem )[] ) => onChange( resolveAccessRuleOptionTokens( tokens, options, selected ) ) }
+					messages={ getAccessRuleTokenFieldMessages() }
+					__experimentalValidateInput={ ( input: string ) => isAccessRuleOptionInput( input, options ) }
+					__experimentalAutoSelectFirstMatch
+					__experimentalExpandOnFocus
+					__next40pxDefaultSize
+					__nextHasNoMarginBottom
+				/>
+				{ hasUnlistedAccessRuleValues( options, selected ) && (
+					<p className="newspack-access-control-block-visibility-panel__notice">{ getUnlistedAccessRuleValuesNotice() }</p>
```
jason10lee: **Nit:** The caution renders as a bare `<p>`, and the wizard's `description` lands as a sibling paragraph. Neither is wired to the input via `aria-describedby`, so the one piece of copy standing between a publisher and an accidentally-widened gate isn't announced.

Core's `FormTokenField` exposes no help or describedby hook, so this is awkward to fix properly — raising it as something to know rather than something to block on.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,240 @@
...
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { __ } from '@wordpress/i18n';
+import { decodeEntities } from '@wordpress/html-entities';
+import type { TokenItem } from '@wordpress/components/build-types/form-token-field/types.d.ts';
+
+export type AccessRuleOption = { value: string | number; label: string };
```
jason10lee, line 30: **Nit:** Dropping `DynamicOptionItem` is safe — its only consumers were the three call sites removed here, and nothing else in the plugin references it.

Worth noting that `AccessRuleOption` now exists both as an ambient global in this file and as an export from `access-rule-options.ts`. `block-visibility.tsx` picks up the ambient one, the wizard imports the module one. They're structurally identical today, so nothing breaks — but the next edit to either declaration won't be caught by the compiler.

jason10lee (approved):
Release risk: ELEVATED. Nothing blocks; the approval stands.

You flagged that the new commits were substantial, so we went ahead and reviewed the full diff rather than just the delta.  The token layer continues to hold up, but we spotted some potential rough spots with the institution fetch: suggestions below.

Two routes we checked and found harmless: a caller without `edit_others_posts` cannot reach the fetch at all, because `enqueue_block_editor_assets()` gates the panel on the same capability the REST route requires — we confirmed an Author gets no panel. And `per_page=-1` is fine: api-fetch rewrites it to 100 and walks the Link headers before the request leaves the browser.

<details>
<summary>What we verified (for reviewers and agents)</summary>

Isolated env on the branch head `860a972a4`, newspack-plugin rebuilt so the served bundle carries the change rather than trunk's.

- **Capability parity, measured.** `Block_Visibility::enqueue_block_editor_assets()` returns early unless `current_user_can( 'edit_others_posts' )`, the same capability as `Institution_REST_Controller::READ_CAPABILITY`. With an administrator and an author in the same env: `panel-gate=yes` and `panel-gate=no` respectively. Nobody who can open the panel can be refused by the route.
- **`per_page=-1` is not a defect.** Calling the route server-side returns `400 rest_invalid_param`, but that bypasses the browser. api-fetch's `fetchAllMiddleware` rewrites `-1` to `100` and paginates on `Link: rel="next"`; the request core sees returns `200`. Both confirmed in the env.
- **The fetch is redundant for correctness.** `Institution::get_options()` localises the complete list (`posts_per_page => -1`, publish, ordered by title). Measured in the env: localised option count `3`, published institutions `3`. What the fetch buys is freshness within an SPA session, which is real but narrow.
- **Tests.** The six suites touching this diff pass — 51 tests. Three unrelated suites flake under load (a full run failed 6 suites, a second failed 3, with different membership, on 5000 ms timeouts and SIGKILLed workers); none import anything this diff changes, and CI is green on this exact head.

</details>

Review provenance: 2 of 3 reviewer passes (lineup: deep, wp-expert); release risk: ELEVATED (degraded — skipped: codex).

plugins/newspack-plugin/src/wizards/audience/views/content-gates/gate-summary.tsx
adekbadek: Passed the fetched options in, in 0da8d85. `getGateSummarySections()` now takes an `optionsBySlug` map, and both callers — the list card and the pre-save panel — get it from a new `useAccessRuleOptions()` hook, which is also what the wizard's pickers read. An institution added a moment ago is named on every surface rather than reading as `(institution not listed)` on one of them.

To keep that from firing one request per gate card, `getAccessRuleOptionSource()` now shares a request that is already in flight. It drops the entry as soon as it settles rather than caching, so a later mount still sees institutions added or removed since.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/edit/access-rule-control.tsx
adekbadek: Fixed in 0da8d85. The caution is now a shared `UnlistedValuesNotice` component that `OneTimePurchaseRuleControl` renders itself, so it reaches that picker on both surfaces. All three pickers use the same component, and the wizard field keeps "Search by name or ID." as its description instead of having the caution replace it.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
adekbadek: Both fixed in 0da8d85. An empty response no longer replaces the localised list, so the picker cannot fall through to the free-text control and write a string over the stored IDs. A failed fetch now renders the same wording the wizard raises as a notice, inline under the field, rather than passing silently.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Took the second option in 0da8d85: a digits-only input resolves whether or not an option describes it, so the wording holds and the path you describe now round-trips — remove the token, type the ID back, it returns.

A full `<name> (#<id>)` token still resolves only from the rule's own stored values, so one pasted from somewhere else adds nothing. Both are covered by tests.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Corrected in 0da8d85 — the docblock now says the validator only guards the one-token-at-a-time path, that separator-carrying input reaches `addNewTokens` without consulting it, and that pasted IDs come through intact. Nothing worth patching around in core here: the field's own paste handling is what skips the validator.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
adekbadek: Addressed in 0da8d85, though not by association — core hard-codes the input's `aria-describedby` to its own "how to" element and takes no second one. The shared `UnlistedValuesNotice` speaks the caution politely on mount instead, and carries `role="note"`, so it reaches a screen reader before the publisher acts on a token. That is the same route core's own `Notice` takes.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Collapsed the duplicate in 0da8d85 — the ambient declaration is now `import( '../access-rule-options' ).AccessRuleOption`, matching how `EditorOneTimePurchaseRuleValue` already aliases its module type. One declaration, so the compiler catches the next edit.

plugins/newspack-plugin/src/content-gate/access-rule-option-sources.ts
adekbadek: Took the memoize option rather than dropping the source. The localised list is frozen for the life of the SPA, and institutions are created, renamed and deleted in that same app — without the fetch, a summary or picker opened after an institution is added calls it "not listed", which is the wording that tells a publisher the value may be safe to remove.

So in 3023f1b the list is memoized at module scope and invalidated explicitly. The collection is walked once per rule per session; every later reader — a picker per active rule, a summary card per gate, and the block editor's control on each remount — resolves from that same promise. `invalidateAccessRuleOptions( INSTITUTION_RULE_SLUG )` drops it, called by the institutions list view after a delete and by the edit view after a save or a delete. A rejected request is not kept, so a failed fetch does not become the answer for the rest of the session.

`access-rule-option-sources.test.js` covers the three behaviours: one `apiFetch` however many readers ask, a refetch after invalidation, and a retry after a failure. Each was confirmed failing against the corresponding mutation.

Say the word if you meant drop the source outright — the count matching at 3 does suggest the localised list is fine until someone edits institutions mid-session.

adekbadek:
@jason10lee Re-requesting review — two commits since your 08-22 approval, both on your own threads.

Worth your eye on one interaction between them: the picker and summary now keep the page-load list when a fetch comes back empty, and the institution list invalidates after a write, so deleting your last institution refetches and then discards the result — both surfaces go on naming it for the rest of the session. The other half is the memoize path rather than dropping the source, with the institutions views invalidating after a write.

- fix(access-control): carry the not-listed caution across every gate picker
- fix(access-control): fetch each gate picker's option list once per session

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,272 @@
...
+			continue;
+		}
+		// Match the whole token first: an option's name may itself end in something that
+		// looks like an ID suffix, and the exact label is unambiguous where it applies.
+		let value = byLabel.get( raw );
+		if ( undefined === value ) {
+			const trimmed = String( raw ).trim();
+			const id = ( trimmed.match( TOKEN_ID_PATTERN )?.[ 1 ] ?? trimmed ).trim();
+			value = findAccessRuleOption( options, id )?.value ?? stored.find( v => String( v ) === id );
+			if ( undefined === value && BARE_ID_PATTERN.test( trimmed ) ) {
```
jason10lee: **Suggestion:** Bare-ID acceptance reaches the gate picker, where the justification doesn't carry — This fallback backs `isAccessRuleOptionInput`, which is the `__experimentalValidateInput` for the block editor's **Gates** field, not only the rule pickers. The docblock's justification is specific to product rules — an Active subscription rule matches a variation ID, so evaluation resolves more than the option list holds. For gates it doesn't hold: `available_gates` is published-gates-only and `has_active_gates()` also requires `publish`, so the list and the server's activity test agree exactly. An ID outside the list is never a valid gate.

The consequence is fail-open. `has_active_gates()` returning false makes `is_hidden_for_user()` return false, so a block whose only gate token is an unresolvable number renders to everyone. Reproduced with Reader Activation on and a published gate requiring registration:

```
logged-out, REAL gate -> is_hidden_for_user: true => block RESTRICTED
logged-out, TYPO 999999 -> is_hidden_for_user: false => block PUBLIC
```

Reachability is low, and worth stating so this isn't over-weighted. The field expands on focus listing every gate; `__experimentalAutoSelectFirstMatch` commits the real suggestion once two characters are typed; and `findAccessRuleOption` resolves any real gate ID exactly, so even a single-digit ID picks the right gate. Stored values resolve via `stored`, so re-saving an untouched block invents nothing. The resulting chip also reads `(gate not listed) (#…)`. What's left is a typed or pasted number that is not a published gate — a typo, a stale migration ID, or a draft gate's ID — and the paste path skips the validator entirely.

Scoping the bare-ID path per rule slug would close it, since no gate needs it. Giving `GateControls` the caution would at least explain the chip.

Separately, and not from these commits: the same fail-open is reachable with no typing at all by unpublishing or trashing a gate that blocks point at. That looks like the larger exposure and probably wants its own issue.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
```diff
@@ -166,50 +175,63 @@ const AccessRuleValueControl = ( {
...
+		source()
+			.then( fetched => {
+				if ( cancelled ) {
+					return;
+				}
+				// Keep the localised list when the response is empty. A filtered query or
+				// a narrowed collection would otherwise leave the rule with no options at
+				// all, which drops the picker to the free-text control below — and that
+				// writes a string over the rule's stored IDs.
+				if ( fetched.length > 0 ) {
```
jason10lee: **Suggestion:** The two new safety behaviors aren't pinned by tests — The empty-response guard is the most consequential thing in these commits, and neither it nor the fetch-failure notice has a test on either picker — `block-visibility-panel.test.jsx` and `access-rule-control.test.jsx` reference neither path.

Similarly, `access-rule-option-sources.test.js` covers the cache itself well, but nothing asserts that the three institution mutation callbacks actually call `invalidateAccessRuleOptions`. That's the half that would regress silently if someone refactors those handlers, and it's the half the freshness claim rests on.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
```diff
@@ -166,50 +175,63 @@ const AccessRuleValueControl = ( {
...
+					onChange={ ( tokens: ( string | TokenItem )[] ) => onChange( resolveAccessRuleOptionTokens( tokens, options, selected ) ) }
+					messages={ getAccessRuleTokenFieldMessages() }
+					__experimentalValidateInput={ ( input: string ) => isAccessRuleOptionInput( input, options ) }
+					__experimentalAutoSelectFirstMatch
+					__experimentalExpandOnFocus
+					__next40pxDefaultSize
+					__nextHasNoMarginBottom
+				/>
+				<UnlistedValuesNotice options={ options } value={ selected } />
+				{ didFetchFail && <p className="newspack-access-rule-values-notice">{ getAccessRuleOptionsFetchFailedNotice() }</p> }
```
jason10lee: **Suggestion:** The fetch-failure note is absent from the branch where it matters most — `didFetchFail` renders only inside the `options.length > 0` branch. If the localised list is empty and the fetch also fails, the component falls through to the `TextControl` below with nothing indicating anything went wrong — and the comment a few lines up identifies that control as the one that writes a string over the rule's stored IDs.

The reachable case is narrow: a rule whose options were empty at page load but whose value still holds IDs, for instance when every institution is unpublished. It's also the only case where the publisher gets no signal at all and data loss is attached. Hoisting the paragraph above the branch would cover both.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/use-access-rule-options.ts
```diff
@@ -0,0 +1,77 @@
...
+			.forEach( slug => {
+				getAccessRuleOptionSource( slug )?.()
+					.then( options => {
+						if ( ! cancelled ) {
+							setFetchedOptions( current => ( { ...current, [ slug ]: options } ) );
+						}
+					} )
+					.catch( () => {
+						if ( ! cancelled ) {
+							addNotice( {
```
jason10lee: **Suggestion:** One failed fetch raises a snackbar per mounted consumer — Every consumer attaches its own `.catch()` to the shared cached promise, so a single rejected request dispatches `addNotice` once per consumer. Nothing collapses them: the store appends unconditionally (`packages/components/src/wizard/store/index.js:59`), `id` is only read by `removeNotice`, and the list is keyed on `notice.id || index`, so identical ids also trip a duplicate-key warning in development.

The Access Control landing page renders one `ContentGateSettings` per gate, and each now calls this hook — so a site with six gates gets six identical snackbars. `WizardSnackbar` maps `type: 'error'` to `politeness="assertive"`, so a screen reader hears it six times. Previously the hook lived only in `AccessRuleControl`, which bounded it to the active rules on one edit screen.

Dispatching the notice from the module-level source next to the cache would fire it once per request; treating `id` as a dedup key in the reducer would also work.

jason10lee (commented):
Both rounds of feedback came back answered, and the answers still hold up. The caution now reaches the pickers it was missing from, the summary and the pickers read the same list, and the option fetch is shared across a session with the write paths dropping it. The cache work is the strongest part — a rejected request is dropped so the next reader retries, and the tests cover the dedup, the invalidation and the retry.

Nothing here blocks. The one worth weighing before merge is that widening the pickers to accept a typed ID no option describes is right for the product rules, where evaluation genuinely resolves more than the list holds, but it also reaches the block editor's gate field, where it has no use: that list is complete by construction, so an ID outside it is always a mistake, and an unresolvable gate makes the block public rather than private. The existing guards keep this unlikely — the list is **right there**, a real ID resolves exactly, and the chip says the gate isn't listed — so this is about removing an affordance nothing needs, not about a hole anyone is likely to fall into.

The rest are smaller: how a failed fetch behaves with several gates on screen, and what an empty option list should mean now that it can be told apart from a failed one.

Worth separating out that unpublishing a gate that blocks point at already makes them public the same way, with no editor involved. That predates this PR and looks like the bigger exposure of the two.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/use-access-rule-options.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	return useMemo(
+		() =>
+			Object.fromEntries(
+				Object.entries( rules ).map( ( [ slug, rule ] ) => [
+					slug,
+					// An empty response leaves the localised list in place: it was
+					// complete at page load, and dropping to no options at all would turn
+					// still-granting IDs into entries a publisher can only remove.
+					fetchedOptions[ slug ]?.length ? fetchedOptions[ slug ] : rule?.options ?? [],
```
jason10lee: **Suggestion:** An empty fetch is indistinguishable from a failed one, so deleted values stay selectable — This guard answers the previous round's concern, and it's the right instinct — an empty option list drops both pickers to the free-text control, which writes a string over stored IDs. But it treats a successful empty response the same as no response.

Delete the last institution while the wizard is open and the refetch legitimately returns `[]`, so the page-load snapshot is used instead. Confirmed against the env: after deleting every institution the collection returns 0 items while the snapshot still holds 3 labels. Those institutions stay named and selectable, and saving one creates a rule nobody satisfies.

Inverting the guard on its own would reintroduce the free-text fallthrough. The distinction that would let both hold is between "this rule has no option source" and "this rule's list is currently empty" — a fetched rule with genuinely zero options wants an empty picker, not a text field.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -0,0 +1,272 @@
...
+				value = Number( trimmed );
```
jason10lee, line 204: **Nit:** A `0` token is accepted, then dropped on save — `BARE_ID_PATTERN` is `/^\d+$/`, so `'0'` passes the validator, becomes the value `0`, and renders as `(product not listed) (#0)`. On save, `sanitize_access_rule` wraps its map in `array_filter()` with no callback, which drops `0` as falsy — the token disappears on reload with nothing said.

That's the same "token that silently disappears on the next render" this PR set out to eliminate, arriving through the one input these commits newly admit. `Number( trimmed ) > 0` would close it.

plugins/newspack-plugin/src/content-gate/components/unlisted-values-notice.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * WordPress dependencies.
+ */
+import { speak } from '@wordpress/a11y';
+
+/**
+ * Internal dependencies.
+ */
+import OneTimePurchaseRuleControl from './one-time-purchase-rule-control';
```
jason10lee, line 15: **Nit:** The file is named for a component it never imports — The suite renders `OneTimePurchaseRuleControl` only; `UnlistedValuesNotice` appears nowhere in it. Of the three mounts the shared component now has, this covers one — the other two are covered by role in the picker suites. Either rename it after what it covers, or add a direct render of the component so the shared piece has a test of its own.

plugins/newspack-plugin/src/content-gate/components/unlisted-values-notice.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * Internal dependencies.
+ */
+import { getUnlistedAccessRuleValuesNotice, hasUnlistedAccessRuleValues, type AccessRuleOption } from '../access-rule-options';
+
+export default function UnlistedValuesNotice( { options, value }: { options: AccessRuleOption[]; value: unknown } ) {
+	const hasUnlisted = hasUnlistedAccessRuleValues( options, value );
+
+	useEffect( () => {
+		if ( hasUnlisted ) {
+			speak( getUnlistedAccessRuleValuesNotice(), 'polite' );
```
jason10lee, line 41: **Nit:** The caution is re-spoken on every mount — The effect fires on mount rather than on a transition, so the caution is announced again each time the component mounts with `hasUnlisted` already true. In the wizard that's once per screen. In the block editor the inspector unmounts on deselection — the same remount that motivated the promise cache — so clicking between two blocks that both carry an unlisted value re-announces the full paragraph each time.

`@wordpress/a11y` won't suppress the repeat: `filterMessage` appends a non-breaking space when a message matches the previous one, specifically to force re-announcement. A ref that skips the first run, or speaking on focus rather than mount, would keep the announcement without the repetition.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/style.scss
```diff
@@ -69,6 +69,13 @@
 		}
 	}
 
+	// Caution under a picker holding IDs the option list cannot describe.
+	.newspack-access-rule-values-notice {
+		color: wp-colors.$gray-700;
+		font-size: 12px;
```
jason10lee: **Nit:** Hardcoded 12px where the token is already imported — This rule already `@use`s `wp-vars` three lines up, and the sibling definition in `block-visibility.scss` uses `wp-vars.$helptext-font-size`, which is `12px` today. The two agree now and would drift silently if the token moved. The duplicated rule itself is defensible — the two scopes beat different host stylesheets — but the literal isn't.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Scoped the bare-ID path rather than adding the caution. `resolveAccessRuleOptionTokens` and `isAccessRuleOptionInput` now take the rule slug, and for a rule whose option list is the complete set of matchable values an ID outside that list is refused. Gates qualify for the reason you give; institutions qualify for the same one — `Institution::get_options()` and `Institution::evaluate()` both resolve from the published institutions — so both are in `EXHAUSTIVE_OPTION_RULES` and everything else keeps the bare-ID path a variation ID needs.

Two things came with it. The rejection wording is now per slug, in `getAccessRuleTokenFieldMessages( slug )` alongside the not-listed labels, because "or type its ID" is a promise the gate and institution fields no longer keep. And `stored` still resolves for every rule, so unpublishing a gate a block points at does not drop it on the next save — only typing a fresh unlisted one is refused.

I left `GateControls` without the caution. Its wording says the entries are still checked when access is evaluated, which is true of a variation ID and false of a gate outside the list, so it would be explaining the chip with the wrong reason. The unpublished-gate exposure is what that wording would have to be written for, and I agree it wants its own issue — I'll open one.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
adekbadek: Both behaviours are pinned now, on both pickers.

The empty-response guard and the fetch-failure notice: `keeps the picker when the list comes back empty, and stops naming what the list no longer holds` and `keeps the page-load list when the request fails, and says it may be stale`, added to `block-visibility-panel.test.jsx` and `access-rule-control.test.jsx`. The guard's own shape changed while pinning it — see the thread on `use-access-rule-options.ts`.

The invalidation call sites: a new `institutions/edit.test.jsx` drives Save and Delete through the header actions and asserts each drops the cached list, plus one case proving a failed write does not; `institutions/index.test.js` gains one that renders the list's delete confirmation and clicks it. Each was confirmed failing with its `invalidateAccessRuleOptions` call removed.

Fixing a latent loop in that second suite was needed to get there: its `useDispatch` mock returned a fresh object per render, which re-ran the list's fetch effect on every render. It is one object for the suite now.

plugins/newspack-plugin/src/content-gate/editor/block-visibility.tsx
adekbadek: Hoisted — the notice is now built once and rendered after whichever control the rule gets.

Worth saying that the case you name lands differently after the empty-versus-failed fix on the other thread: a rule with a fetched source now keeps its picker whatever its list currently holds, so an institution rule whose options were empty at page load and whose fetch then failed renders the picker, the stored IDs as tokens, and the stale-list note. It no longer reaches the `TextControl` at all. The hoist stays because it makes that independent of the other fix.

Covered by `keeps the page-load list when the request fails, and says it may be stale`.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/use-access-rule-options.ts
adekbadek: Taken the distinction you name. `isOptionBackedAccessRule( slug, localisedOptions )` asks whether the rule is backed by a list at all — it has a fetched source, or the page localised options for it — and both pickers branch on that instead of on the current length. An empty fetched list is now used as it is, so an institution deleted while the wizard is open stops being named and stops being selectable, and the rule keeps its picker with no suggestions rather than falling through to the text control. A failed fetch is still the only thing that keeps the page-load snapshot, and it still says so.

Covered on both pickers by `keeps the picker when the list comes back empty, and stops naming what the list no longer holds`, in `access-rule-control.test.jsx` and `block-visibility-panel.test.jsx`.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Fixed — the bare-ID path now takes `Number( trimmed ) > 0`, so a `0` is refused at entry with the same announcement as any other unusable input rather than becoming a token the server drops. Covered by `refuses a zero, which the server would drop from the saved list anyway`.

plugins/newspack-plugin/src/content-gate/components/unlisted-values-notice.test.tsx
adekbadek: Added a direct render of `UnlistedValuesNotice` rather than renaming, since the filename was only half the cost — the shared component had no test of its own either. The suite now mounts it directly for both the has-a-value and every-value-resolves cases, and keeps one case going through `OneTimePurchaseRuleControl` to prove that picker actually mounts it. The other two mounts stay covered by role in the picker suites.

plugins/newspack-plugin/src/content-gate/components/unlisted-values-notice.tsx
adekbadek: Spoken once for the page now, via a module-scope flag rather than a ref. A ref that skips the first run would have removed the announcement in the case it exists for — a picker that mounts already holding an unlisted value, which is every gate editor a publisher opens on an affected gate. The repetition to suppress happens across mounts, not within one, so the flag is where it has to live.

Deduplicating per value set would still have re-read the identical paragraph on the second block, since the wording names no value. Once is enough.

Covered by `is spoken once, not again each time a picker remounts` in `unlisted-values-notice.test.tsx`, which mounts, unmounts and mounts again.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/style.scss
adekbadek: Fixed — `wp-vars.$helptext-font-size`, matching the sibling rule.

adekbadek:
**Merge-order note for #756 and #775** — both rewrite `access-rule-control.tsx` and `block-visibility.tsx`, and they conflict textually. Whichever lands second needs the union, not either side, and the union is not just "keep both hunks":

- From #756: the extracted option helpers and ID-based token resolution. #775 still matches tokens by label, which is the corruption #756 exists to fix. Taking #775's file wholesale afterwards puts that back.
- From #775: `has_options` as the discriminator for which control renders.

The second point is the one that isn't obvious from the diff. #756 introduces `isOptionBackedAccessRule( slug, localisedOptions )`, which returns `localisedOptions.length > 0 || undefined !== ACCESS_RULE_OPTION_SOURCES[ slug ]`. That is the same question `has_options` answers, but asked client-side — and it keeps the hole NPPD-2143 is about, one rule over. `subscription` has no entry in `ACCESS_RULE_OPTION_SOURCES`, and its localised list is empty whenever the site has no subscription products (WooCommerce inactive, or none created yet). So on #756 as it stands, that rule still degrades to the free-text box and a typed string can still be saved where product IDs belong.

`has_options` comes from the server registration and doesn't have that hole. The merged version should keep #756's helper as the call site and have it read `has_options` rather than measure the list.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/use-access-rule-options.ts
adekbadek: Confirmed, and fixed in e1862a8 — once per request, but dispatched from the hook rather than the source module.

The source is shared with the block editor (`block-visibility.tsx:189` calls `getAccessRuleOptionSource` directly, and renders its own inline failure notice instead of a snackbar), so a dispatch to `WIZARD_STORE_NAMESPACE` next to the cache would fire there too. Deduping on `id` in the reducer would work, but it changes `packages/components` for every wizard, which is the scope you objected to earlier on this PR.

So the guard is a module-level `WeakSet` of already-announced promises. The source hands every reader the same promise, so the first uncancelled `.catch()` claims the announcement and the rest skip. Keyed on the promise, not the slug, because that is the scope wanted: the source drops a rejected entry, so a retry is a different promise and announces again. An unmounted reader neither announces nor claims, leaving it to one still on screen.

Two tests in `use-access-rule-options.test.jsx`: three readers on one rejected request raise one notice, and a second failed request after the first still raises its own. The first fails at 3 without the guard; the second fails if the guard is widened to the slug.

The other two from your review body are separate — the empty-vs-failed list distinction, and the unpublished-gate exposure that predates this PR.

adekbadek:
@jason10lee Re-requesting review – since your approval institutions were added to the exhaustive-list rules and an empty fetch now replaces the page-load list, so an institution token can read `(institution not listed)` under a caution saying it is still checked, cannot be typed back once removed, and an institution rule emptied that way grants access to everyone:
- fix(access-control): refuse a gate ID no published gate matches (NPPD-2129, #756)
- fix(access-control): raise one notice per failed fetch (NPPD-2129, #756)

PR #757, opened 2026-07-28
fix(audience): consistent headers, goal-first rules, accessible controls

thomasguillot:
### All Submissions:

Design review of the three newest Audience surfaces — **Plans**, **Pricing Rules** and **Subscribers** — plus the shared components they lean on. Part of DSGNEWS-189.

#### Headers, across all three

Plans and Pricing Rules were the only Audience views that never declared `breadcrumbs`, so the header fell back to rendering `headerText` as a single crumb: publishers saw one dead item reading "Audience Management / Plans", and the page `<h1>` was that whole path string rather than the page name. Both now build a real trail, and the add/edit screens gain a linked parent so there is a way back to the list.

Where a list has a scope, the trail names it. All four lists — Subscribers, Groups, Plans and Pricing Rules — annotate the heading with how many rows match the current search and filters. The count is muted beside the heading and announced with the noun it counts, "33 subscriptions", rather than "left-paren 33 right-paren". No "total": every one of these counts follows the search and filters, so the shorter phrasing is accurate whether or not a filter is applied. Every one of them suppresses the count while the read is in flight or after it fails, so a slow or failed load never claims the list is empty. On the two Audience lists a failed read now shows an error notice with **Retry** in place of the table, matching what Subscribers and Groups already did, rather than falling through to an unannotated empty table.

Plans list with the Audience Management / Plans / Subscriptions (33) breadcrumb trail

#### Plans

The three scopes get their own breadcrumb leaf, kept in step with the tab labels. Page titles and header actions move to Title Case (**Add Plan**, **Edit Plan**, **Create Plan**), and the edit action shortens to **Save** to match the rest of Audience.

#### Pricing Rules

Creating a rule now starts with what it is for. The goal presets the options that match it: who qualifies, whether the price locks in at purchase, and which products it covers. The publisher fills in the pricing; Custom leaves everything to them. Changing the goal later re-seeds only the fields that goal owns, so pricing, schedule and dates already typed stay put.

Goal picker on a new rule: five cards, none selected

The goal is the form's first section rather than a dialog in front of it, so nothing gates the form. A rule's goal is fixed once it is created, so on a saved rule the whole set still renders, with the chosen one selected and the rest greyed out: the publisher can see what the rule was built around and what the alternatives were.

Goal picker on a saved rule: the chosen goal selected, the rest greyed out

Switching away from Custom raises a confirm naming exactly what will not carry over, computed from what is actually set rather than a static list.

#### Subscribers

The Subscribers and Groups headings carry the same counts, and the spoken phrase now follows the publisher's own vocabulary: a site that renames Groups to "Teams" hears "2 Teams", not "2 groups". Setting only one of the two label overrides no longer leaves the heading showing one noun while the screen reader says another.

The bundle has no JS translations loaded, so a sentence built there would read in English beside a noun PHP had already translated. Both the nouns and the sentences that wrap them now ship from PHP, which puts the group count, the group-role column, the failed-read notice and a group link's accessible name into the site's language. The English default reads "14 Groups": the noun keeps the heading's capitalisation, because lowercasing it is wrong in languages that capitalise nouns.

#### Shared components

`Breadcrumbs` items can carry a `count`, so the formatting, muted treatment and accessible phrasing live in one place. Counts are grouped for the site's language rather than the browser's, so the digits agree with the translated phrase beside them and with the figures PHP renders elsewhere on the page. On a linked crumb the count sits outside the link, so the link is named for where it goes rather than for the view being left.

Core card gains two additive props for the goal cards: `isVertical`, and `isSelectable`, which carries the button reset, hover ring and focus ring so the next selectable card inherits them rather than copying the declarations. `CardFeature`'s button gets an accessible name that identifies its feature, since the cards sit side by side and a bare "Enable" names nothing. The autocomplete help text takes its line height from the token rather than a hardcoded value.

#### Accessibility

The goal cards are a radiogroup: arrow keys move the choice, and they follow the writing direction so an Arabic or Hebrew admin gets the card the key points at. On a saved rule the cards are `aria-disabled` rather than `disabled`, so the chosen goal keeps its place in the tab order and stays readable. The active card keeps a visible border under Windows High Contrast, where colour alone would not distinguish it. The cards stack below 600px rather than forcing sideways scrolling at 400% zoom.

#### Translations

Two strings shared a catalogue entry with an unrelated one, so a translator's choice for either would have landed in both: the "Save" goal collided with the ~30 "Save" submit buttons, and a `%1$s %2$s` name template collided with the content-gate countdown. Both now carry their own context, with the English unchanged.

#### Correctness

A rule's goal and its URL stay in step when either changes. A rule saved before goals existed opens as Custom instead of a blank field with no way to set one.

The Custom eligibility date always carries a date behind it. The engine reads a missing cohort gate as "everyone qualifies", so a selector reading Custom over an empty date would have saved a rule that repriced every subscriber, including those acquired long before it existed.

Requires WooCommerce Subscriptions, `define( 'NEWSPACK_PLANS_UI', true );`, and the bespoke `woocommerce-dynamic-pricing` engine plugin for the Pricing Rules screens.

**Headers**

1. Go to **Audience → Plans**. The trail reads **Audience Management › Plans › Subscriptions (33)** as separate crumbs, and the page title is the scope name, not the whole path. Type in the search box: the count follows the visible rows. Switch tabs and confirm the leaf and count follow the active scope.
2. Press **Add Plan**. The trail gains an **Add Plan** leaf and the **Plans** crumb links back to the list. Follow it and confirm it lands there.
3. Compare against **Audience → Donations**, which already shipped the correct shape.

**Pricing Rules**

4. Go to **Audience → Pricing Rules** and press **Add Rule**. **Goal** is the first section, with five cards and none selected. Pick Custom, set **Priority** to something other than 100, then pick another goal: a dialog names what will not carry over. Open a saved rule and confirm the cards are greyed out apart from the chosen one, and that hovering them does nothing.
5. Press <kbd>Tab</kbd> to the cards, then the arrow keys. The selection moves one card at a time and wraps around. **Select Goal** is announced as unavailable until a card is chosen, and stays reachable by keyboard throughout.
6. Choose **Subscription Retention** and press **Select Goal**. The form names the goal, fills the name, sets **Applies to** to all subscriptions, and the URL becomes `#/new/retention`. Type a price, press **Change**, and switch to **Win-Back**: the price survives, and the eligibility, application and product scope change to the new goal's.
7. Reload on `#/new/retention`. The rule opens on that goal with its presets applied, and the header still carries **Save** and the **Add Rule** crumb.
8. Return to the list. The trail reads **Audience Management › Pricing Rules (n)**, and the count follows the search box and the status filter.
9. On **Add Rule** with the **Custom** goal, set the eligibility selector to **Custom**, then to **Anytime**, then back to **Custom**. The date field carries a date every time it is shown; it is never blank while the selector reads Custom.

**Subscribers**

10. Go to **Audience → Subscribers**. The heading counts the subscribers matching the current view, and reads "1 subscriber" at one, not "1 subscribers".
11. In **Audience → Configuration**, set the group label override to `Team` / `Teams`, then open the Groups view. The heading shows **Teams (n)** and announces "n Teams". Clear one of the two fields and confirm the heading and the spoken phrase still name the same noun.

### Other information:

<details>

**`#/new/:goal?` is one route entry.** The wizard renders each section as a keyed `<Route>` inside a `<Switch>`, so a second entry would change the key when choosing a goal rewrites the URL, remounting the editor and discarding everything typed. `routes.test.jsx` pins this against the real `Wizard`.

**The picker is a modal, and the form never unmounts.** `choosePath()` applies a recipe in place: it resets `conditions` (filtered by `conditionsVisibleUnder()`, then `applyRecipeConditions()`), `application` where the recipe defines one, `cycleAnchor`, and `scopeType`/`scopeIds` where the recipe's scope is in vocab; for non-Custom goals it also forces `priority` and `composeMode`, which are Custom-only controls. Everything else is preserved, including the name once the publisher has typed one.

**The URL is authoritative.** An `[ initialPath ]` effect adopts a goal changed outside the form, and canonicalises a goal-less `#/new` back to the goal on screen rather than clearing the form, which would leave a rule that can never be saved. Because that `history.replace` is a route change, and the wizard's `ResetHeaderData` blanks the header on every pathname change, `RuleForm`'s `setHeaderData` effect takes `pathname` as a dependency so the header republishes.

**Cold loads carry the recipe through mount-only seeds.** Since selection canonicalises the URL, a reload or bookmark lands on `#/new/<goal>`, where `choosePath()` never runs. `rule-form-goal-modal.test.jsx` asserts the posted body for all four named goals on that path.

**Group labels.** `class-subscribers-wizard.php` ships the raw options alongside the translated defaults, so the client can tell a custom noun from the default and still render a localized one, plus a `groupPhrases` map of the four sentence templates that wrap the noun. `wp_set_script_translations` is not called anywhere in the plugin, so JS `__()` returns the msgid verbatim — which is why the templates come from PHP rather than being built in `labels.js`, where they would resolve to English beside a translated noun. Each PHP msgid is byte-identical to its JS fallback, so they merge into one catalogue entry rather than two. `labels.js` is the single place a sentence wrapping the noun may be composed; the two `_x()` contexts above are catalogue hygiene rather than a rendering change today.

**Option keys.** `Group_Subscription::get_label_override()` and `get_default_label()` now own the `newspack_group_subscription_label_*` keys and the `Group`/`Groups` defaults; `get_label()` composes them, and both wizards read through them instead of keeping their own copies.

**Counts.** `breadcrumbs/format-count.js` groups a number for `document.documentElement.lang`, inside a try/catch because WordPress ships locales `Intl` rejects, such as `pt_PT_ao90`. Every caller that builds its own `countLabel` shares it, so the visible count and the spoken one are grouped the same way.

**`packages/components`.** `isVertical` and `isSelectable` emit no class when absent, so the nine consuming repos are unaffected. The kebab menu’s accessible name changes from "More" to "More options for %s"; neither `newspack-manager-admin` (no `newspack-components` dependency) nor `newspack-manager` (pins ~4.5.0, imports only Button, Grid, SectionHeader, SelectControl) selects it by name.

</details>


thomasguillot:
**This PR is the base of a stack.** #798 branches from `fix/dsgnews-189-dynamic-pricing` and builds on the Pricing Rules work here, and further changes will stack on top of #798.

Two things that follow from that:

- **Merge this one first.** #798's diff only reads correctly while its base branch is intact.
- **A force-push here needs #798 rebased onto the new head.** Otherwise GitHub falls back to the shared merge-base and #798's diff swells to include all 46 files from this PR.

Reviewing them in order also keeps the Pricing Rules editor coherent: this PR establishes the goal-first flow and the header trail, and #798 reworks the controls inside it.

miguelpeixe (commented):
Nice work — the reasoning in this one holds up well. I checked the load-bearing claims against the code rather than taking them from the description, and they stand: the single `#/new/:goal?` route does avoid the remount, the `[initialPath]` effect does not loop and a garbage goal slug falls through to the picker, count suppression is wired into all four lists, `wp_set_script_translations` really is never called in this plugin, and the `Intl` fallback in `formatCount` is load-bearing. The PHP refactor keeps `get_label()` behaviour intact and changes no capability or sanitisation path. Token discipline in the new CSS is clean throughout — no hardcoded colour anywhere.

No blockers. Three things I'd look at before merge:

- **Changing a rule's goal drops more than the modal says it does.** Priority and compose mode go, and so does the "Subscriptions started on/after" eligibility date if one was set under Custom. That last one is the one that matters: it is invisible under a named goal, so there is nothing on screen to notice it by, and the pricing engine reads an absent cohort gate as "everyone qualifies".
- **The rule editor renders the same words as both `h1` and `h2`** now that the leaf crumb and the section title both say "Add Rule". One decision — crumb or section title, not both — would settle all four screens this PR is making consistent.
- **A failed read still lands on an empty table.** The heading count is suppressed correctly, but both Audience lists fall through to DataViews' "No results" with no retry once the snackbar auto-dismisses. The Subscribers screens already show an error notice and Retry.

The design comments are mostly consolidation asks rather than defects. The one I'd do now is moving the selectable-card treatment into `CoreCard` beside `isVertical` — three more PRs build on this screen, and `isVertical` shows the shape that works.

One heads-up for the release note: the kebab menu's accessible name changes from "More" to "More options for %s" in `newspack-components`. `newspack-manager-admin` consumes that from npm and isn't checked out in my workspace, so it's worth a grep there for anything selecting the kebab by name.

- `plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:284` — **Suggestion:** A failed read still renders an empty table, so the screen does claim the list is empty — The count suppression is wired into all four lists and the tests prove it. The stated purpose, though — "a slow or failed load never claims the list is empty" — only half-holds for the two DataViews lists: on `hasError` both fall through to `<DataViews data={ [] } />`, which renders DataViews' own "No results" empty state. In `subscription-products/list.tsx` `hasError` is read *only* by the header effect; nothing else branches on it. The `addNotice` snackbar auto-dismisses after 10 seconds, so shortly after a failed fetch the publisher is looking at an unannotated empty table with no way to retry. `GroupList.jsx:239` and `SubscriberList.jsx:344` already do the right thing with an error `Notice` plus a Retry action — matching them here would finish the job the count suppression starts.

Review provenance: 3 of 4 reviewer passes (lineup: deep, wp-expert, design, codex); release risk: ELEVATED (degraded — skipped: codex).

packages/components/src/autocomplete-tokenfield/style.scss
```diff
@@ -34,8 +35,9 @@
 
 	.newspack-autocomplete-tokenfield__help {
```
miguelpeixe, line 36: **Suggestion:** Help text is now sized differently from neighbouring controls' help — The margin and colour now match WordPress's `BaseControl` help exactly, which is the right target. The size doesn't: `$helptext-font-size` is 12px, the classic-admin `.description` size, while `BaseControl` help is 13px — `$default-font-size` is the matching token. In `rule-form.tsx` an `AutocompleteTokenField` sits in the same `VStack` as `TextControl`s and `SelectControl`s with `help`, so this ships two help-text sizes in one column.

Separately, `line-height: 1.5` is a hardcoded value where `$default-line-height` (1.4) exists and matches WordPress. This file is in a published package, so it reaches the other consuming units too — though newspack-newsletters and newspack-listings both use the component without passing `help`, so nothing renders differently for them today.

packages/components/src/breadcrumbs/format-count.js
```diff
@@ -0,0 +1,18 @@
...
+ * the page. Callers that build their own `countLabel` share this so the visible count
+ * and the spoken one are grouped the same way.
+ *
+ * @param {number} count Number to format.
+ * @return {string} The grouped number.
+ */
+export const formatCount = count => {
+	const lang = ( typeof document !== 'undefined' && document.documentElement.lang ) || '';
+	try {
+		return Number( count ).toLocaleString( lang.replace( '_', '-' ) || undefined );
```
miguelpeixe, line 13: **Suggestion:** The locale fallback lands on the browser's locale, which is what the helper exists to avoid — `lang.replace( '_', '-' )` replaces only the first underscore, and `document.documentElement.lang` in wp-admin is already BCP-47 (`get_bloginfo('language')` yields `pt-PT-ao90`), so the underscore handling never fires for the case the comment cites.

What does fire is the `catch`: `pt-PT-ao90` is structurally invalid for `Intl` (a 4-character variant subtag must start with a digit), so it throws and the helper falls through to `toLocaleString()` with no argument — the browser's locale. On a `pt_PT_ao90` site viewed from a US-English browser the crumb count is grouped `1,234` while every server-rendered figure on the page is grouped `1.234`, which is the exact divergence the doc block says it prevents. Truncating the tag progressively (`pt-PT-ao90` → `pt-PT` → `pt`) before giving up would keep the site's locale for these WordPress variants.

packages/components/src/breadcrumbs/style.scss
```diff
@@ -26,6 +27,23 @@
 	}
 }
 
+.newspack-breadcrumbs__count {
```
miguelpeixe, line 30: **Suggestion:** base-styles colour inside a WPDS-token component — `Breadcrumbs` is built on `@wordpress/ui` (`Link`, `Stack`, `Text`) and its sibling rule is already on `var(--wpds-color-stroke-surface-neutral, #dbdbdb)`. The new count rule reaches back to `$gray-700` / `$font-weight-regular`.

Not a violation — both are legitimate admin token sets — but the established convention in this package for WPDS-based components is `var(--wpds-…, #{wp-colors.$fallback})`, as in `tabbed-navigation/style.scss:56` and `divider/style.scss:37`. A WPDS foreground token with a `$gray-700` fallback would keep the component on one scale. Worth noting the win here too: this rule consolidates the duplicate `.newspack-subscribers__header-count` that the PR correctly deletes.

packages/components/src/card-feature/index.tsx
```diff
@@ -178,13 +184,18 @@ const CardFeature = ( {
...
 									size="compact"
+									aria-label={ buttonAccessibleLabel }
 								>
 									{ buttonLabel }
 								</Button>
 								{ showMoreControls && (
 									<DropdownMenu
 										icon={ moreVertical }
-										label={ __( 'More', 'newspack-plugin' ) }
+										label={ sprintf(
```
miguelpeixe, line 194: **Suggestion:** Renaming the kebab menu changes an accessible name across a published package — The `DropdownMenu` accessible name goes from `'More'` to `'More options for %s'`. In-repo this required updating `card-feature-more-menu.test.js:21`, which is exactly the breakage signature: anything selecting the kebab by accessible name — `getByRole('button', { name: 'More' })`, a Playwright `getByLabel('More')`, a QA script — now misses.

The change itself is an improvement, since side-by-side cards each had a kebab named nothing in particular. The ask is coverage: `newspack-manager-admin` consumes `newspack-components` from npm and has no checkout in this workspace, so it could not be swept here. Worth a grep on that repo before merge, and worth a line in the PR body so the package release note carries it.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -156,24 +156,45 @@ public static function is_group_subscription( $subscription ) {
...
+	 * @return string The trimmed override, or an empty string when the publisher has set none.
 	 */
-	public static function get_label( $variant = 'singular' ) {
-		$variant    = 'plural' === $variant ? 'plural' : 'singular';
-		$option_key = 'newspack_group_subscription_label_' . $variant;
-		$override   = trim( (string) \get_option( $option_key, '' ) );
-		if ( '' !== $override ) {
-			return $override;
-		}
+	public static function get_label_override( $variant = 'singular' ) {
```
miguelpeixe, line 168: **Suggestion:** The class owns the label option key for reads only — `get_label_override()` now builds `'newspack_group_subscription_label_' . $variant`, but the same key is still assembled independently in two other places: `class-audience-wizard.php:1138` derives it from the REST field name (coupling the option name to the request payload shape), and `class-group-subscription-settings.php:1097` hard-codes both literals in `register_setting()`.

A rename inside `Group_Subscription` would leave reads pointing at the new key while writes and registration stayed on the old one, with no failure surfacing until a publisher's override stopped taking effect. A `get_label_option_key( $variant )` — or a `set_label_override()` companion — would close it. The read paths themselves are correct: both wizards go through the accessors, and the fallback is behaviour-identical to the old `get_label()`.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -1261,29 +1261,39 @@ public function enqueue_scripts_and_styles() {
...
 			'newspack-subscribers',
 			'window.newspackSubscribers = ' . wp_json_encode(
 				[
-					'groupLabel'       => $group_label_singular,
-					'groupLabelPlural' => $group_label_plural,
+					'groupLabel'              => $group_label_singular,
+					'groupLabelPlural'        => $group_label_plural,
+					'groupLabelDefault'       => Group_Subscription::get_default_label( 'singular' ),
+					'groupLabelDefaultPlural' => Group_Subscription::get_default_label( 'plural' ),
+					'groupPhrases'            => [
```
miguelpeixe, line 1279: **Suggestion:** The groupPhrases payload is a parallel i18n path worth scoping explicitly — The premise checks out: `wp_set_script_translations` appears nowhere in `newspack-plugin`, so JS `__()` does return the msgid verbatim, and the four msgids here are byte-identical to their JS fallbacks including translator comments — they will merge into single catalogue entries as intended.

Worth noting that `languages/newspack-plugin.pot` *is* a `wp i18n make-pot` artifact with 1,505 `#: src/…` references, so the JS msgids are already extracted; the gap is only that no `.json` catalogues are generated and no handle calls `wp_set_script_translations`. That makes this a build-pipeline gap rather than a structural constraint, and the phrase payload a workaround that has to be maintained per sentence — any future string mixing the publisher's noun with copy must be added to both the PHP array and `labels.js`, in matching order, or it silently reverts to English. A comment in `labels.js` naming the exit condition would keep it from growing into a parallel catalogue.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
```diff
@@ -1261,29 +1261,39 @@ public function enqueue_scripts_and_styles() {
...
+						/* translators: 1: number of groups. 2: the group label, e.g. "Groups". */
+						'count'      => __( '%1$s %2$s total', 'newspack-plugin' ),
```
miguelpeixe, line 1281: **Suggestion:** The group-count phrase is the one counted string that can't inflect — `'count' => __( '%1$s %2$s total', 'newspack-plugin' )` registers a single msgid, and `groupCountLabel()` then picks the noun with `1 === total ? GROUP_LABEL : GROUP_LABEL_PLURAL`. Every other counted surface in this PR uses `_n()` — `_n( '%s rule total', '%s rules total', … )` and its siblings — so translators get plural forms there but not here.

In a locale with more than two plural forms (Polish, Russian, Arabic, Czech) the wrapper sentence has one form for all counts, and a binary noun choice can't reach the third or fourth category. It is also the only one of these phrases that is actually translated at runtime, since it comes from PHP — so this is precisely where the limitation shows. A publisher's own override is inherently two-form, but for the default noun the plugin owns both halves and could ship `count_one`/`count_other` templates instead.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/goal-cards.tsx
```diff
@@ -0,0 +1,70 @@
...
+			return;
+		}
+		event.preventDefault();
+		const next = ( activeIndex + ( forward ? 1 : -1 ) + options.length ) % options.length;
+		onSelect( options[ next ].value );
+		event.currentTarget.querySelectorAll< HTMLElement >( '[role="radio"]' )[ next ]?.focus();
+	};
+
+	return (
+		<HStack spacing={ 4 } alignment="stretch" role="radiogroup" aria-label={ __( 'Rule goal', 'newspack-plugin' ) } onKeyDown={ onKeyDown }>
```
miguelpeixe, line 42: **Suggestion:** Selectable-card pattern and its iconography are both new — worth settling before the stack builds on them — No selectable card exists in the system today: `Card` offers `actionType: 'chevron' | 'toggle'` and `buttonsCard` for navigation, and nothing carries a checked state. The radiogroup implementation itself is solid — roving tabindex, RTL-aware arrows with vertical arrows deliberately not mirrored, wrap-around, `preventDefault` only on keys it owns, an accessible name from title plus summary, and icons that are `aria-hidden`/`focusable="false"` automatically through `@wordpress/primitives`.

What needs designer input is the checked affordance (see the high-contrast finding) and the icon mapping. The icons come from `@wordpress/icons`, which is the right source, but they're being asked to stand for business concepts rather than UI actions — `backup` (a clock with an arrow) for "Win-Back" and `reusableBlock` for "Subscription Retention" aren't self-evident. Settling this before the three PRs behind it build on the set would avoid reworking them later.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/index.tsx
```diff
@@ -19,20 +19,37 @@ import PricingRulesList from './list';
...
+	{
+		// One entry for both #/new and #/new/<goal>: choosing a goal rewrites the URL,
+		// and a second entry would change the Route key and remount the editor.
+		path: '/new/:goal?',
+		render: RuleEdit,
+		isHidden: true,
+		exact: true,
+		breadcrumbs: RULES_TRAIL,
+		backNav: '#/',
+		title: __( 'Add Rule', 'newspack-plugin' ),
```
miguelpeixe, line 38: **Suggestion:** Rule editor now renders the same words as both h1 and h2 — The hidden sections keep `title: 'Add Rule'` / `'Edit Rule'`, which `Wizard` renders as a `SectionHeader` at heading level 2 — and that component's own comment says the breadcrumb `Page` owns the single page `<h1>`. This PR also has `rule-form.tsx:401` publish `sectionName: 'Add Rule'`, which becomes the trailing crumb, i.e. the `<h1>`.

Before this PR the `<h1>` came from the `headerText` fallback ("Audience Management / Pricing Rules"), so the two never matched. Now the editor shows `h1: Add Rule` directly above `h2: Add Rule`, and heading navigation announces the same string twice. The Plans editor has the same shape, so one decision — leaf crumb *or* section title, not both — would settle all four screens this PR is trying to make consistent.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -91,55 +126,97 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
...
-		// Subscription presets target all subscriptions; Custom stays all products.
-		const wantScope = RECIPES[ p ].defaultScope;
-		if ( vocab.scopes.some( s => s.id === wantScope ) ) {
-			setScopeType( wantScope );
+		if ( vocab.scopes.some( s => s.id === nextRecipe.defaultScope ) ) {
+			setScopeType( nextRecipe.defaultScope );
 			setScopeIds( [] );
 		}
+		// Custom-only controls; a named goal would carry them hidden and unremovable.
+		if ( ! nextRecipe.isCustom ) {
```
miguelpeixe, line 164: **Suggestion:** Changing the goal quietly resets Priority and compose mode — `choosePath` resets `priority` to `'100'` and `composeMode` to `'min'` for any non-Custom goal, and clears `scopeIds`. The reasoning in the comment is sound — a named goal shouldn't carry hidden, unremovable values — and the behaviour is tested.

The gap is the modal copy: "A new goal presets its own eligibility, price locking, products, and how cycles are counted. Your pricing, steps and dates stay as they are." Priority and compose mode appear in neither list, and the sentence's shape invites the reader to assume anything unlisted survives. A publisher who sets Priority 5 and "This rule only" under Custom, opens the picker to look at Win-Back, applies it, then switches back to Custom finds both fields at their defaults with no notice and no undo. Either name them in the copy, or stash and restore them when returning to Custom.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -389,316 +467,354 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
...
+					) }
+					noMargin
+				/>
+				{ ! isSchedule && String( value ).trim() === '' ? (
+					<p className="newspack-pricing-rules__muted">{ __( 'Enter a price to see the impact preview.', 'newspack-plugin' ) }</p>
+				) : (
+					<RulePreview body={ previewBody } />
+				) }
+			</div>
+			{ isChangingGoal && (
```
miguelpeixe, line 772: **Suggestion:** Mandatory non-dismissible modal on first load is a new flow shape for Audience — `isDismissible`, `shouldCloseOnEsc` and `shouldCloseOnClickOutside` are all false until a goal exists. There is precedent in the codebase (`src/other-scripts/corrections-modal/index.js:214`), and the "Back" action keeps it from being a keyboard trap, so this isn't an accessibility defect.

It is, though, the only Audience "Add" screen that gates its form behind a dialog — Plans and Institutions both go straight to the form. That's a flow decision rather than a token choice, so flagging it for a product/design call rather than proposing a change.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -1,54 +1,27 @@
...
-		&:hover {
-			border-color: wp-colors.$gray-600;
-		}
+	&__goal input {
+		height: wp-vars.$grid-unit-50;
 
-		&.is-selected {
-			border-color: var(--wp-admin-theme-color, #{wp-colors.$gray-900});
-			box-shadow: inset 0 0 0 1px var(--wp-admin-theme-color, #{wp-colors.$gray-900});
+		&:read-only {
```
miguelpeixe, line 18: **Suggestion:** New read-only field state has no design-system token — The contrast claim verifies and was reached the right way: `$gray-800` on `$gray-100` is about 11.4:1, all WordPress admin tokens, with no hardcoded hex anywhere in the PR's new CSS.

What's new is the *state*. "Read-only input as a value display" doesn't exist elsewhere in the system — `--newspack-ui-color-input-bg-disabled` and its border sibling are reader-side only, and admin has no read-only token trio. WordPress ships no read-only input styling either, so these declarations are needed regardless of which component renders the field. Before the rest of the stack copies them onto other screens, this probably wants a named token and designer sign-off. Not a violation as written.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -124,3 +90,66 @@
...
+		padding: wp-vars.$grid-unit-30 0 wp-vars.$grid-unit-10;
+		position: static;
+	}
+
+	.components-modal__content {
+		margin-top: 0;
+		padding-top: 0;
+	}
+
+	.components-card {
```
miguelpeixe, line 107: **Suggestion:** Card-as-button styling belongs in CoreCard alongside isVertical — `goal-cards.tsx:49` passes `as: 'button'` through `__experimentalCoreProps`. `core-card.js:60` applies the buttons-card class only for `buttonsCard || as === 'a'`, and the header-as-button path at `:90` needs `onHeaderClick`, which the goal cards don't pass — so `as: 'button'` matches neither styled path, and this stylesheet re-implements the button reset, hover ring and focus ring.

Four of those declarations are verbatim duplicates of ones the component already ships (`style-core.scss:176` and `:256`). Reusing them isn't possible as things stand — `buttonsCard` force-rewrites `as` to `'a'` at `core-card.js:79`, which would silently render the radio card as an anchor, and the header path can't carry `role`/`aria-checked`/`tabIndex`. So the fix is to add an `as: 'button'` (or `isSelectable`) styled path to `CoreCard`, which is squarely in scope: this PR already extends the same component with `isVertical` in the same diff, and did that one exactly right.

Nothing is broken today — the concern is that the next selectable-card consumer copies these declarations, and three of the four stacked PRs behind this one build on this screen.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -124,3 +90,66 @@
...
+		flex: 1 1 0;
+		margin: 0;
+		padding: 0;
+		border: 0;
+		font: inherit;
+		color: inherit;
+		text-align: inherit;
+		cursor: pointer;
+
+		&:hover {
```
miguelpeixe, line 118: **Suggestion:** Checked state is conveyed by colour alone, with no high-contrast path — Selected is a 1px `--wp-admin-theme-color` ring plus a blue title; hover is the *same* 1px ring; focus is a 2px inset ring that, at higher specificity, replaces the selected ring entirely. The forced-colors safeguard (`outline: 2px solid transparent`) is applied to `:focus-visible` but not to `is-active`, and there is no `@media (prefers-contrast: more)` block in the new CSS.

`aria-checked` covers assistive technology, so this affects a narrower group: a sighted user in Windows High Contrast has no reliable way to tell the chosen card from its neighbours. Given the PR's stated purpose is accessible controls, it seems worth closing here.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -124,3 +90,66 @@
...
+			box-shadow: inset 0 0 0 2px var(--wp-admin-theme-color);
+			outline: 2px solid transparent;
+		}
+	}
+
+	.newspack-card--core__header-content span {
+		transition: color 125ms ease-in-out;
+	}
+
+	.newspack-card--core__header-content .newspack-pricing-rules__goal-title {
```
miguelpeixe, line 137: **Suggestion:** Goal-title styling duplicates the card's own header typography — `.newspack-card--core__header-content .newspack-pricing-rules__goal-title` re-creates, declaration for declaration, what `style-core.scss:195` already applies to `h1`–`h6` inside `__header-content` (gray-900, weight 600, `display: flex`, `align-items: center`, 8px gap), plus the `$font-size-large` that `style-core.scss:37` applies under `isSmall`. The active-state rule likewise re-creates `style-core.scss:218`.

Rendering the title as a heading element inside `header` — or giving `CoreCard` a first-class title slot — would inherit all of it and keep the active and hover colour transitions in one place. A heading inside `role="radio"` is safe here, since the radio's name is computed from its subtree, which the tests already assert. This shares a root cause with the card-as-button finding above.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -124,3 +90,66 @@
...
+		font-size: wp-vars.$font-size-large;
+		font-weight: 600;
+		gap: wp-vars.$grid-unit-10;
+	}
+
+	.newspack-card--core__is-active .newspack-pricing-rules__goal-title {
+		color: var(--wp-admin-theme-color);
+	}
+
+	@media ( max-width: wp-breakpoints.$break-small ) {
```
miguelpeixe, line 150: **Suggestion:** ARIA attribute used as the styling hook for the responsive stack — The breakpoint itself is right: `$break-small` is 600px and is the correct standard variable for admin context, so the claim about stacking below 600px holds.

Two smaller things. Selecting on `[role="radiogroup"]` couples presentation to semantics — if the group's role changes, or a second radiogroup lands in this modal, the stacking silently follows or breaks; a `newspack-`prefixed BEM class is the house styling hook. And `max-width: $break-small` overlaps by one pixel with WordPress's mobile-first `break-small()` mixin (`min-width: 600px`), so at exactly 600px both branches apply; `$break-small - 1px` is the non-overlapping form.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/list.tsx
```diff
@@ -50,6 +51,36 @@ const inScope = ( item: SubscriptionProduct, scope: Scope ): boolean => {
...
+	groups: __( 'Plan bundles', 'newspack-plugin' ),
+};
+
+// Each scope names what it counts, so the heading announces "12 donations total"
+// rather than the generic "12 items" every other counted surface avoids.
+const SCOPE_COUNT_LABELS: Record< Scope, ( total: number ) => string > = {
+	subscriptions: total =>
+		sprintf(
+			/* translators: %s: number of subscription plans matching the current view. */
+			_n( '%s subscription total', '%s subscriptions total', total, 'newspack-plugin' ),
```
miguelpeixe, line 67: **Suggestion:** The spoken phrase says "total" for a number that is the filtered subset — Both surfaces ship a default filter that hides rows: `subscription-products/list.tsx:101` defaults to `status is publish`, and `GroupList.jsx:49` defaults to `status isAny [active, pending, on-hold]`, deliberately hiding cancelled groups. The count published is `paginationInfo.totalItems`, which is post-filter and correct per the translator comments ("matching the current view").

What a screen reader actually speaks, though, is "12 subscriptions total" / "14 Groups total". On a site with 6 cancelled groups the heading asserts a site-wide total that is wrong by 6, and only for assistive-technology users — sighted users see a bare `(14)`, which reads naturally as a count for the current view. Dropping "total" from the phrasings, or reserving it for the unfiltered set, would close the gap.

packages/components/src/breadcrumbs/style.scss
```diff
@@ -26,6 +27,23 @@
...
+	color: wp-colors.$gray-700;
+	font-weight: wp-vars.$font-weight-regular;
+
+	.screen-reader-text {
```
miguelpeixe, line 34: **Nit:** Third hand-rolled copy of visually-hidden — `.screen-reader-text` is already defined globally by wp-admin (these components only render there), and the identical nine declarations already appear at `card/style-core.scss:91`. The `visuallyHidden` mixin the design system points at lives in the plugin's `src/shared/scss/_mixins.scss` and isn't reachable from `packages/components`, so the fix would be a small shared mixin in this package rather than an import. Low stakes — flagging so it doesn't become a fourth copy.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
```diff
@@ -156,24 +156,45 @@ public static function is_group_subscription( $subscription ) {
...
+	}
+
+	/**
+	 * Get the translated default container label.
+	 *
+	 * @param string $variant Either 'singular' or 'plural'. Unknown variants fall back to singular.
+	 *
+	 * @return string The translated default.
+	 */
+	public static function get_default_label( $variant = 'singular' ) {
```
miguelpeixe, line 180: **Nit:** New public API declared without type hints — These are new public surface and a natural place to start typing: `public static function get_label_override( string $variant = 'singular' ): string`, and the same for `get_default_label()` and `get_label()`. The PHPDoc already asserts `string` in and `string` out, so the signature would only restate what the contract claims, and scalar and return hints are available well below the plugin's PHP floor. Flagging as a nit because the file and its neighbours carry no hints at all, so this is a convention change rather than a fix.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/recipes.ts
```diff
@@ -100,15 +101,35 @@ export function applyRecipeConditions( path: PricingPath, conditions: Conditions
...
+/** A one-line summary of a path, shown on its card in the goal picker. */
+export function pathSummary( path: PricingPath ): string {
+	const map: Record< PricingPath, string > = {
+		new_subscriptions: __( 'An intro or stepped offer for first-time subscribers.', 'newspack-plugin' ),
+		retention: __( 'A renewal discount to keep existing subscribers.', 'newspack-plugin' ),
+		save: __( 'A last-chance offer at the cancellation moment.', 'newspack-plugin' ),
+		winback: __( 'Re-acquisition pricing to bring lapsed subscribers back.', 'newspack-plugin' ),
+		custom: __( 'Full manual control, with none of the options preset.', 'newspack-plugin' ),
+	};
+	return map[ path ];
```
miguelpeixe, line 129: **Nit:** pathSummary and pathDescription bracket-index an object literal — `isPricingPath` deliberately uses `hasOwnProperty` with the comment "`in` would accept inherited members, letting `#/new/toString` past the route guard", and `rule-form.tsx:141` mirrors it. But `pathDescription` does `return map[ path ] ?? '';`, so a stored `intent` of `constructor`, `toString` or `valueOf` resolves to the inherited `Object.prototype` member rather than `''`. `pathSummary` has the same shape and didn't get the `?? ''` guard its sibling gained in this PR.

Reaching it needs an admin to POST a hand-crafted `intent` to the rules REST endpoint, and the outcome is cosmetic — the Goal field's help text would render `function Object() { [native code] }`. Flagging it only because it's a one-line inconsistency with an invariant the same file documents.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -310,76 +393,71 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
...
-					{ ! isNew && recipe && (
-						<p className="description" style={ { marginTop: 0 } }>
-							{ pathDescription( path as PricingPath ) }
-						</p>
-					) }
+				<VStack spacing={ 6 } className="newspack-pricing-rules__details">
+					<BaseControl id="newspack-pricing-rule-goal" label={ __( 'Goal', 'newspack-plugin' ) } help={ goalHelp } __nextHasNoMarginBottom>
+						<HStack className="newspack-pricing-rules__goal" alignment="center" spacing={ 2 }>
+							<FlexBlock>
+								<input
```
miguelpeixe, line 434: **Nit:** Goal field is a raw input carrying a WordPress-internal class name — The Goal field is a bare `<input className="components-text-control__input" readOnly>` inside a `BaseControl`, which couples the screen to an internal class with no stability guarantee and costs one CSS line (`style.scss:16`) to recover the 40px control height.

Worth being precise about the size of this, because the obvious fix isn't a drop-in. `TextControl` renders its own `BaseControl`, so dropping it in here double-wraps, needs `__nextHasNoMarginBottom` re-passed to avoid a deprecation warning, needs a noop `onChange` (it's a required prop), and does *not* remove the `${id}__help` hand-wiring, since the help stays on the outer control. Removing that coupling means moving label and help onto `TextControl`, which breaks the `HStack` centring of the Change button against the input. There are also two prior instances of this exact raw-input pattern on `main` (`contribution-meter` and `collections` InspectorPanels), both of which recover the height with an inline style rather than SCSS — so this is tidier than what it follows. Fine to leave; noting it so the pattern is a decision rather than drift.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -310,76 +393,71 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
...
+							{ isNew && (
+								<Button
+									ref={ changeGoalRef }
+									variant="secondary"
+									onClick={ () => {
+										setPendingGoal( path as PricingPath );
+										setIsChangingGoal( true );
+									} }
+									aria-label={ __( 'Change goal', 'newspack-plugin' ) }
+									disabled={ isSaving }
```
miguelpeixe, line 453: **Nit:** Change button drops out of the tab order while saving — The modal's confirm button uses `accessibleWhenDisabled`, so it stays focusable and announces `aria-disabled` — and the tests assert exactly that. The Change button beside the Goal field takes a bare `disabled={ isSaving }` and leaves the tab order while a save is in flight. Same screen, two behaviours, and the accessible one is already the house pattern here.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -389,316 +467,354 @@ export default function RuleForm( { isNew, rule, vocab, onDone }: RuleFormProps
...
-					<Divider alignment="full-width" variant="tertiary" />
+			<Divider alignment="full-width" variant="tertiary" />
 
-					<Grid columns={ 2 } gutter={ 32 } noMargin>
-						<SectionHeader
-							title={ __( 'Pricing model', 'newspack-plugin' ) }
-							description={ __( 'How matching products are priced.', 'newspack-plugin' ) }
+			<Grid columns={ 2 } gutter={ 32 } noMargin>
+				<SectionHeader
+					title={ __( 'Pricing Model', 'newspack-plugin' ) }
```
miguelpeixe, line 516: **Nit:** "Pricing Model" heading above a "Pricing model" field label — The section header is now "Pricing Model" while the `SelectControl` inside it is still "Pricing model". The title-casing sweep is the right call and matches the rest of Newspack admin — "Add Institution", "Add Campaign", "Save Settings", "Edit Segment" are all title case today, so "Add Plan" and "Add Rule" are corrections rather than regressions. The sweep just stopped at section headers and buttons, leaving field labels behind in the one place where both render side by side.

packages/components/src/autocomplete-tokenfield/style.scss
thomasguillot: Checked this against the component before changing it: `BaseControl`'s help renders at `helpText.fontSize`, which is 12px in `@wordpress/components`, so `$helptext-font-size` already matches and the two are the same size. Left as is.

The `line-height` half stands though, and is now `$default-line-height` rather than a hardcoded 1.5.

packages/components/src/breadcrumbs/format-count.js
thomasguillot: Fixed. The tag is now shortened a subtag at a time until `Intl` accepts one, so `pt-PT-ao90` falls back to `pt-PT` and then `pt` rather than to the browser's locale. Underscores are all replaced now, not just the first.

packages/components/src/breadcrumbs/style.scss
thomasguillot: Done, now `var(--wpds-color-foreground-content-neutral-weak, $gray-700)`, matching the convention the separator rule beside it already follows.

packages/components/src/card-feature/index.tsx
thomasguillot: Swept both repos. `newspack-manager-admin` has no `newspack-components` dependency at all, and `newspack-manager` pins `~4.5.0` and imports only `Button`, `Grid`, `SectionHeader` and `SelectControl`. Neither selects the kebab by name, so nothing outside this repo is affected.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
thomasguillot: Fixed. `get_label_option_key( $variant )` is the only place the key is assembled now, and `register_setting()` loops the two variants through it, so reads and registration cannot drift apart.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
thomasguillot: Added the exit condition to `labels.js`: the file exists only because no handle calls `wp_set_script_translations` and no `.json` catalogues are generated, and it should shrink back to the two nouns once that build step lands.

plugins/newspack-plugin/includes/wizards/class-subscribers-wizard.php
thomasguillot: Resolved by dropping "total" (see the `subscription-products/list.tsx` thread). The template is now just `%1$s %2$s`, so the wrapper carries no word to inflect and the limitation is confined to the binary noun choice, which is inherent to a publisher's own override.

Shipping `count_one`/`count_other` would have given two identical msgids that merge into one catalogue entry, so it would not have bought the translator anything.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/goal-cards.tsx
thomasguillot: Both settled. The icons are remapped: Subscription Retention from `reusableBlock` to `loop`, Win-Back from `backup` to `undo`. (`repeat` would have been the better fit but is not exported at the 11.8.0 we resolve.)

The selectable-card treatment moved into `CoreCard` as `isSelectable`, so the next consumer inherits it rather than copying the declarations.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/index.tsx
thomasguillot: Fixed. The section titles are gone from the Pricing Rules, Plans and Institutions editors, so the breadcrumb leaf is the page's only `<h1>` and nothing repeats it as an `<h2>`. A test pins that "Add Rule" appears exactly once as a heading.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: Fixed, and the modal went with it: the picker is the form's first section now.

Switching away from Custom raises a confirm naming exactly what will not carry over, computed from what is actually set rather than a static list. A cohort date left on the auto-applied publish default is deliberately excluded, otherwise every switch out of Custom would raise a dialog over a default nobody chose.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: The modal is gone. The five goals are the form's first section on both create and edit, so nothing gates the form any more.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: The read-only field no longer exists on this branch: the goal is a card set now, so those declarations went with it. `ReadonlyField` arrives in #798, so the token question belongs on that PR.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Done. `CoreCard` takes an `isSelectable` prop carrying the button reset, the hover ring and the focus ring. It leaves `as` alone, unlike `buttonsCard`, so the consumer keeps its element and the ARIA it needs.

This screen's stylesheet is down to `flex: 1 1 0` plus the disabled treatment.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Added, and it sits in `CoreCard` beside the rest of the selectable styling: `forced-colors` gives the active card a `CanvasText` outline, and `prefers-contrast: more` thickens its ring.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Fixed by rendering the title as an `<h3>`, so it inherits `style-core.scss`'s own header typography and active colour. The duplicated block and the active-colour rule are both gone.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Both done. The hook is a `newspack-pricing-rules__goals` class on the group itself, and the query is `$break-small - 1px` so it no longer overlaps WordPress's mixin at exactly 600px.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/list.tsx
thomasguillot: Dropped "total" from all of them: subscriptions, donations, plan bundles, groups, subscribers and rules. Every one of those counts follows the search and filters, so the shorter phrasing is accurate whether or not a filter is applied and needs no conditional.

packages/components/src/breadcrumbs/style.scss
thomasguillot: Simpler than a mixin in the end. wp-admin already defines `.screen-reader-text` in `common.css`, and `Breadcrumbs` renders that exact class and only ever renders in wp-admin, so the nine local declarations were dead weight. Deleted them.

The `card/style-core.scss` copy is a different case: it hides a bare `label` that carries no class, so it cannot lean on the global and stays.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
thomasguillot: Added on all three, plus the new `get_label_option_key()`.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/recipes.ts
thomasguillot: Fixed with an `ownValue()` helper both lookups share, using the same own-property check `isPricingPath` documents. Worth noting `?? ''` would not have caught it on its own, since `map.constructor` is truthy.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: Gone: the goal is a card set now, so the raw input and its `components-text-control__input` class have both been removed.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: The Change button no longer exists, so the inconsistency went with it.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: The section header reads "Pricing" from #798 onward, so the two no longer sit side by side.

thomasguillot:
Thanks for the thorough pass, this was genuinely useful. Everything is addressed and replied to inline. The headlines:

**The goal picker is no longer a modal.** It is the form's first section on both create and edit, which resolves the mandatory-dialog question, the h1/h2 duplication, the goal-reset copy and the read-only field thread in one go. On a saved rule the whole set renders with the chosen goal selected and the rest greyed out and inert, since the goal is fixed after creation.

**`CoreCard` gained `isSelectable`**, carrying the button reset, hover ring, focus ring and the high-contrast treatment, so the next selectable card inherits them instead of copying this screen's declarations. It leaves `as` alone, unlike `buttonsCard`, so the consumer keeps the element and ARIA it needs.

**"total" is gone from every count** across subscribers, groups, plans, donations, plan bundles and rules. That also resolved the plural-forms thread: the wrapper is now bare `%1$s %2$s`, so there is no word left to inflect.

**A failed read on the two Audience lists** now shows an error notice with Retry instead of falling through to an empty table, matching Subscribers and Groups.

Two where I did not do what you suggested, with reasoning inline:

- **The tokenfield help size.** `BaseControl`'s help is 12px (`helpText.fontSize`), so `$helptext-font-size` already matched. I only changed the hardcoded line height.
- **The third visually-hidden copy.** Rather than a shared mixin, I deleted it: wp-admin already ships `.screen-reader-text` and `Breadcrumbs` renders that exact class in wp-admin only, so the declarations were dead.

Also checked the kebab rename you flagged for coverage: `newspack-manager-admin` has no `newspack-components` dependency, and `newspack-manager` pins ~4.5.0 and imports only Button, Grid, SectionHeader and SelectControl. Nothing outside this repo selects it by name.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription.php
miguelpeixe: One assembly point survived the sweep: the REST write path. `api_update_group_labels()` still built the key from the request field name, in `update_option( 'newspack_group_subscription_' . $field, $value )` and the `delete_option()` beside it. That concatenation splits the literal differently, so a search for `newspack_group_subscription_label` doesn't find it — which is also what would let a rename slip past: reads and `register_setting()` would move together while writes and deletes stayed on the old key, and a publisher's label edits would stop persisting.

Pushed e5ddde1c3 routing both writes through `Group_Subscription::get_label_option_key()`, so the accessor is now the only assembly point.

miguelpeixe (approved):

PR #758, opened 2026-07-28
chore(deps): update wpcs to 3.4.1 in remaining composer locks

jason10lee:
### All Submissions:

On that last point: Dependabot had already landed four units on `main` (newspack-ads, newspack-newsletters, newspack-plugin, newspack-popups) before this PR was raised, so those are deliberately excluded here. No Dependabot wpcs PRs are currently open.

Bumps `wp-coding-standards/wpcs` to 3.4.1 in the ten monorepo composer locks still on 3.1.0–3.3.0.

Every WPCS release from 0.14.1 up to (but not including) 3.4.1 carries CVE-2026-45293 ([GHSA-3pwp-g2mj-5p3v](https://github.com/advisories/GHSA-3pwp-g2mj-5p3v), high severity). The `WordPress.WP.EnqueuedResourceParameters` sniff took a string lifted out of the file being scanned and passed it to `eval()`, so linting untrusted PHP could run code on whatever machine did the linting. Both `WordPress` and `WordPress-Extra` turn that sniff on, and the root `phpcs.xml` includes both. 3.4.1 drops the `eval()` and analyses the tokens instead.

Two other advisories clear as a side effect, which is worth naming because it changes whether this is worth backporting.

Moving every unit to phpcsutils 1.2.3 clears CVE-2026-65954 ([GHSA-r6hr-vr92-vv28](https://github.com/PHPCSStandards/PHPCSUtils/security/advisories/GHSA-r6hr-vr92-vv28), arbitrary code execution, affected `<1.2.3`); six of these units were on 1.0.12–1.2.2 and carrying it. It arrives here as a resolver constraint of wpcs 3.4.1, but it stands on its own.

And a second commit takes `squizlabs/php_codesniffer` from 3.13.5 to 3.13.6, clearing CVE-2026-67434 ([GHSA-hmqg-cxww-wqhq](https://github.com/PHPCSStandards/PHP_CodeSniffer/security/advisories/GHSA-hmqg-cxww-wqhq), OS command injection, affected `<3.13.6`). That advisory published on 2026-08-05, after this PR was raised. It is the same dev-machine exposure class in the same toolchain, in locks that are already open here, and Dependabot does not watch these ten units — so folding it in beats waiting for whatever bumps them next. 3.13.6 satisfies both the `^3.13.5 || ^4.0.1` wpcs 3.4.1 asks for and the `^3.13.5` from phpcsutils 1.2.3, and nothing else moves.

Taken together, `composer audit --locked` on `newspack-blocks` goes from four advisories on `main` to one here, and the one left is an unrelated phpunit advisory.

This touches lockfiles only and edits no `composer.json`. Each lock was regenerated with:

```
composer update wp-coding-standards/wpcs -W --minimal-changes --no-install
```

`-W` is not optional here. 3.4.1 raises its own floor to phpcsutils `^1.2.3` and phpcsextra `^1.5.1`, so a bare targeted update refuses to resolve. `automattic/vipwpcs` 3.0.1 already permits `^3.1.0`, so it needs no change.

Six of the units also shed stale `brainmaestro/composer-git-hooks` and symfony entries, packages that had already left their `composer.json` but were never re-locked. Every unit keeps all of its direct dependencies.

Tracked internally as NPPM-3056. There is no GitHub issue to close.

1. Check out this branch and confirm every unit resolves the patched version:
   ```
   for f in $(git ls-files '*/composer.lock'); do
     php -r '$d=json_decode(file_get_contents($argv[1]),true);
       foreach(["packages","packages-dev"] as $k)
         foreach(($d[$k]??[]) as $p)
           if($p["name"]==="wp-coding-standards/wpcs") echo $p["version"]," $argv[1]\n";' "$f"
   done
   ```
   All 14 should report `3.4.1`.
2. Confirm each lock is in sync with its `composer.json`: run `composer validate --no-check-all --no-check-publish` in each changed unit. All should pass. Six do not pass on `main` today: `newspack-blocks`, `newspack-listings`, `newspack-sponsors`, `newspack-story-budget`, `newspack-block-theme`, and `newspack-theme` each report the lock as out of date, because the cghooks entries were never re-locked. Run the same command on `main` first to see the difference.
3. Confirm the diff is lockfile-only: `git diff --name-only origin/main` should list nothing but `composer.lock` files, and `git diff --stat origin/main -- '*.php'` should be empty.
4. Confirm the advisories actually clear: `composer audit --locked` in `plugins/newspack-blocks` lists four on `main` (wpcs, phpcsutils, php_codesniffer, phpunit) and one here (phpunit).
5. Confirm nothing was smuggled in: comparing package sets before and after should show no packages added and no version downgrades in any unit. Four packages change version: `wp-coding-standards/wpcs`, `phpcsstandards/phpcsutils`, `phpcsstandards/phpcsextra`, and `squizlabs/php_codesniffer`. Six units also drop `brainmaestro/composer-git-hooks` and the symfony and psr packages it pulled in — eleven of them in five units, nine in `newspack-story-budget`, so twelve and ten removals respectively once cghooks itself is counted. Nothing else moves, and every changed package sits in `packages-dev`, so no runtime dependency changes.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — not applicable; this changes generated lockfiles and no product code.

On that last box, what I actually ran: `composer validate` across all 14 units, a full monorepo phpcs run against the root ruleset (0 errors), and a package-set comparison confirming no additions and no downgrades. Every changed package's `dist.url` resolves to `api.github.com`. I did not run PHPUnit, since no PHP changed.

Consolidating these ten units' dev dependencies to the root — rather than bumping their locks again at the next advisory — is filed as https://linear.app/a8c/issue/NPPM-3240, with the phpunit-polyfills migration that gates it. This PR is the stopgap and does not wait for that.

Two things worth knowing for review:

The CI lint job installs from the workspace root, which has no committed lockfile, so CI had already resolved 3.4.1 on its own. These locks instead govern the per-plugin vendors that `n build` and `n ci-build` install.

CI on this PR is green. The root `automattic/vipwpcs` constraint stays at `~3.1.0`: #741 fixed the 3.1.0 `NoPaging` errors at the source and merged, so the pin restoration in #759 was closed as superseded.

One thing I left alone: several units run a `post-update-cmd` calling `vendor/bin/cghooks` even though `brainmaestro/composer-git-hooks` is no longer in their `require-dev`, so `composer update` in those directories writes the lock and then exits 127. That deserves its own PR, and #764 is it.

The regeneration does drop the package itself, which is worth naming so the deletion doesn't read as collateral. It sat in six of these ten locks—`newspack-blocks`, `newspack-listings`, `newspack-sponsors`, `newspack-story-budget`, `newspack-block-theme` and `newspack-theme`—as a top-level `packages-dev` entry that nothing required, so re-resolving removed it. The `post-update-cmd` lines that call it are untouched here.


dkoo:
#741 addressed the errors reported by vipwpcs 3.1.0's rewritten `NoPaging` sniff in code and removed the version pin at the same time. CI on this PR (and `main` and its descendant branches) is now green.

jason10lee:
This one targets CVE-2026-45293 ([GHSA-3pwp-g2mj-5p3v](https://github.com/advisories/GHSA-3pwp-g2mj-5p3v), high) rather than the `NoPaging` sniff or CI. WPCS below 3.4.1 runs `eval()` on the `$ver` argument in `WordPress.WP.EnqueuedResourceParameters::is_falsy()`, so linting untrusted PHP can execute code on the scanning host. The advisory names `WordPress` and `WordPress-Extra` as the affected rulesets, and our `phpcs.xml` includes both (`:11`, `:17`).

CI is unaffected, yes—the root `composer.lock` is gitignored, so CI resolves `^3.0` fresh from Packagist and already gets 3.4.1. The exposure is local: `composer install` inside one of these ten units pins the old version, which is the "developer machine reviewing third-party code" case the advisory calls out. All ten are still below the fix — `newspack-blocks`, `newspack-multibranded-site` and `newspack-sponsors` on 3.1.0, `newspack-story-budget` on 3.2.0, the other six on 3.3.0.

On the pin: #741 carries a commit titled *"chore: unpin vipwpcs now that 3.1.0 sniffs are addressed"*, but the squash's net change to `composer.json` is a trailing newline. `"automattic/vipwpcs": "~3.1.0"` was there before #741 and is still on `main`, with nothing touching that file since. Separate from this PR, and `dependabot.yml` has no `ignore` entry for that package either way.

jason10lee:
Copilot declined this at open ("wasn't able to review any files") because the diff is lockfiles, which it filters out of scope. Since the automated gate produced nothing, here's what we verified by hand.

**No revert risk.** `main` hasn't touched any of these ten lockfiles since the merge-base (`9cb37c89`). Comparing base, `main`, and this branch package by package gives zero overlapping packages and zero downgrades, so there's nothing for the merge to undo.

**Nothing runtime moves.** Every changed package across all ten locks sits in `packages-dev`.

**The larger diffs repair drift that already exists on `main`.** Six units fail `composer validate` today, because their locks still carry `brainmaestro/composer-git-hooks` and the symfony and psr packages it pulled in, which their `composer.json` stopped requiring. Re-resolving drops them. All six validate clean on this branch, which is why those units show 13 to 16 package changes rather than 3.

**#764 doesn't conflict.** Composer's content-hash covers `require` and `require-dev` but not `scripts`, so #764's per-unit edits can't invalidate these locks. Checked directly: `composer.json` from #764 with `composer.lock` from this branch validates in sync for `newspack-blocks`, `newspack-sponsors`, and `newspack-theme`. The two can land in either order.

<details>
<summary>Per-unit results</summary>

| Unit | wpcs | Packages moved | Runtime | `validate` on `main` | `validate` on branch |
|---|---|---|---|---|---|
| newspack-blocks | 3.1.0 → 3.4.1 | 15 | 0 | fails | passes |
| newspack-listings | 3.3.0 → 3.4.1 | 15 | 0 | fails | passes |
| newspack-multibranded-site | 3.1.0 → 3.4.1 | 4 | 0 | passes | passes |
| newspack-network | 3.3.0 → 3.4.1 | 3 | 0 | passes | passes |
| newspack-sponsors | 3.1.0 → 3.4.1 | 16 | 0 | fails | passes |
| newspack-story-budget | 3.2.0 → 3.4.1 | 13 | 0 | fails | passes |
| republication-tracker-tool | 3.3.0 → 3.4.1 | 3 | 0 | passes | passes |
| super-cool-ad-inserter | 3.3.0 → 3.4.1 | 3 | 0 | passes | passes |
| newspack-block-theme | 3.3.0 → 3.4.1 | 15 | 0 | fails | passes |
| newspack-theme | 3.3.0 → 3.4.1 | 15 | 0 | fails | passes |

`composer validate --no-check-all --no-check-publish` against each unit's `composer.json` on `main`, paired first with `main`'s lock and then with this branch's.

</details>

plugins/newspack-multibranded-site/composer.lock
```diff
@@ -2299,16 +2299,16 @@
         },
         {
             "name": "squizlabs/php_codesniffer",
-            "version": "3.13.2",
+            "version": "3.13.5",
```
adekbadek: **Severity:** suggestion

**Suggestion:** All ten units end on `php_codesniffer` 3.13.5, one patch below the fix for CVE-2026-67434 ([GHSA-hmqg-cxww-wqhq](https://github.com/PHPCSStandards/PHP_CodeSniffer/security/advisories/GHSA-hmqg-cxww-wqhq), OS command injection, high, affected `<3.13.6`). That advisory published 2026-08-05, a week after this PR was raised, so it isn't an oversight, and the PR improves the package where it moves it at all (3.10.2 and 3.13.2 → 3.13.5). Not a reason to hold the WPCS fix.

Worth a deliberate decision rather than the next drift cycle, though: it's the same dev-machine exposure class, in the same toolchain, in locks that are already open. I ran the remedy on top of this branch — `composer update squizlabs/php_codesniffer -W --minimal-changes --no-install` reports `0 installs, 1 update, 0 removals — Upgrading squizlabs/php_codesniffer (3.13.5 => 3.13.6)`, nothing else moves, and `composer audit --locked` then drops the advisory. 3.13.6 satisfies both the `^3.13.5 || ^4.0.1` wpcs 3.4.1 asks for and the `^3.13.5` from phpcsutils 1.2.3. Folding it in or filing it as a same-day follow-up are both fine; letting it wait for whatever bumps these locks next is the option I'd avoid, given Dependabot doesn't watch these ten units.

plugins/newspack-sponsors/composer.lock
```diff
@@ -490,35 +417,39 @@
...
+                    "type": "thanks_dev"
                 }
             ],
-            "time": "2023-12-08T16:49:07+00:00"
+            "time": "2026-07-27T11:13:17+00:00"
         },
         {
             "name": "phpcsstandards/phpcsutils",
-            "version": "1.0.12",
+            "version": "1.2.3",
```
adekbadek, line 430: **Severity:** nit

**Nit:** The description mentions phpcsutils only as a resolver constraint ("3.4.1 raises its own floor to phpcsutils `^1.2.3`"), but moving every unit to 1.2.3 independently clears CVE-2026-65954 ([GHSA-r6hr-vr92-vv28](https://github.com/PHPCSStandards/PHPCSUtils/security/advisories/GHSA-r6hr-vr92-vv28), arbitrary code execution, affected `<1.2.3`) — the six units on 1.0.12–1.2.2 were carrying it. `composer audit` on `main`'s `newspack-blocks` lock lists four advisories; on this branch it lists two, and both of the ones that went are cleared here.

No code change — it changes what a reader takes away. Anyone later deciding whether this is worth backporting reads the description to decide, so it's worth a sentence there and in NPPM-3056.

adekbadek (commented):
Within what I measured, the diff is correct. Installing from `main`'s lock yields WPCS 3.1.0 with `eval( "return (bool) $code_string;" )` at `EnqueuedResourceParametersSniff.php:239`; installing from this branch's yields 3.4.1 with no `eval(` in that file. Across the ten units: zero runtime changes, zero package additions, zero downgrades. The rewritten sniff still flags a missing `$ver` and a missing `$in_footer`. That much is measured. If this merges as-is, nothing breaks and two high-severity advisories close. The rest is an argument about which change to make, not about a defect in this one.

**Consolidate the ten units' dev dependencies to the root instead of bumping their locks.**

Ten of the fourteen units declare no production dependencies. The other four — `newspack-ads`, `newspack-newsletters`, `newspack-plugin`, `newspack-popups` — are the only composer ecosystems in `.github/dependabot.yml` besides the root, so nothing watches the ten this PR touches. Bumping them today means doing this again at the next advisory. `bin/build-repos.sh:119-120` already states the target: "each plugin still has its own composer.json for production deps; dev deps are hoisted to the root." The ten `composer.json` files only redeclare dev tooling the root already declares — drift from the documented model, not an implementation of it.

The ten locks also gate nothing. CI's PHPCS jobs at `ci.yml:253-258` and `ci.yml:279-287` install and run from the root vendor, as does the pre-commit hook. The root `composer.lock` is gitignored. One consequence: the ten sat between WPCS 3.1.0 and 3.3.0 for months unnoticed. The other: CI resolved 3.4.1 on its own the day before this PR opened. No PR's lint has ever consulted these ten locks. Deleting those ten `composer.json`/`composer.lock` pairs clears CVE-2026-45293 for those ten units as completely as bumping them does, and also clears CVE-2026-67434 (`php_codesniffer` `<3.13.6`), which this PR leaves at 3.13.5 in all ten — see the thread on that line.

## What consolidation costs

More than a file deletion:

- `yoast/phpunit-polyfills` sits at `^2.0`, `^3.0` and `^4.0` across the units while the root pins `^4.0`. Six units would move across one or two majors, and those majors changed the base TestCase and the minimum PHPUnit. This is a test-suite migration, and the real cost.
- Each unit's `tests/bootstrap.php:34` has `require_once __DIR__ . '/../vendor/autoload.php'` — the only reason CI installs per-package, since it already runs the root binary at `ci.yml:406`.
- `bin/build-repos.sh:123` guards on `[ -d "$dir" ]` rather than on a `composer.json` existing, so it needs the composer.json guard the zip builder already has at `ci.yml:455`.
- The four production units keep their `composer.json` and lock either way, so this narrows the per-unit surface rather than removing it.

One decision to make explicitly rather than inherit: consolidating leaves the dev toolchain unpinned, since the root lock is gitignored. That is arguably right for linting — it is how 3.4.1 arrived with no lock change — but "the whole team lints against the same sniff set" stops being assertable. If that matters, committing the root lock belongs in the same change.

If the polyfills migration proves deeper than it looks, this PR is the right stopgap and the consolidation is the follow-up. Worth establishing either way before spending review cycles on ten locks the target state deletes.

Two corrections if it merges as-is: moving `phpcsutils` to 1.2.3 independently clears CVE-2026-65954, so the PR closes two advisories rather than one, which neither the description nor NPPM-3056 says; and test step 4's "twelve symfony and psr packages" is eleven — twelve is the total including cghooks itself — and nine in `newspack-story-budget`.

Review provenance: 5 of 5 reviewer passes (lineup: deep, standards, wp-expert, codex, opencode:kimi-k3); release risk: LOW.

plugins/newspack-multibranded-site/composer.lock
jason10lee: Folded in as `eff53fdc48`. Reproduced your remedy on all ten units first, and it
behaves as you found: `0 installs, 1 update, 0 removals` each time, nothing but
`squizlabs/php_codesniffer` moving, five changed lines per lock. `composer audit
--locked` on `newspack-blocks` now reports one advisory, and it is the unrelated
phpunit one.

Agreed on the reason rather than the fix: Dependabot doesn't watch these ten, so
"the next thing that bumps them" isn't a schedule. The PR description carries the
advisory and the 3.13.6 constraint check now.

plugins/newspack-sponsors/composer.lock
jason10lee: Taken, and measured before writing it: `composer audit --locked` on
`newspack-blocks` lists four advisories on `main` and one here, and the ones that
go are `phpcsstandards/phpcsutils` and `wp-coding-standards/wpcs`. The description
now says phpcsutils 1.2.3 clears CVE-2026-65954 on its own rather than arriving
only as a resolver constraint, since that is what someone deciding on a backport
needs to see. The how-to-test section has a step for reproducing the count.

NPPM-3056 carries the same sentence now.

jason10lee:
Two things from the review body rather than a thread.

The consolidation case is filed as https://linear.app/a8c/issue/NPPM-3240, carrying
your cost breakdown — the polyfills majors as the real work, the bootstrap autoload
path, the `build-repos.sh` guard, and the unpinned-toolchain decision that comes
with it. Agreed on sequencing: this PR is the stopgap, and the issue exists so the
next advisory does not reopen the argument. Sizing the polyfills migration is the
thing that has to happen before any of it, and the issue says so.

The package count was wrong and is fixed. Measured across the six units: eleven
symfony and psr packages in five of them and nine in `newspack-story-budget`,
twelve and ten once cghooks itself is counted.

plugins/newspack-multibranded-site/composer.lock
adekbadek: Verified on `eff53fdc48`: ten locks, five changed lines each, only `squizlabs/php_codesniffer` 3.13.5 → 3.13.6 (same reference `4c378e1a` everywhere). No lock caps it below 3.13.6 — floors are `^3.13.5` (wpcs 3.4.1), `^3.13.5 || ^4.0.1` (phpcsutils 1.2.3, phpcsextra 1.5.1) and older `^3.5.x`/`^3.9.x` from vipwpcs and variable-analysis, with no upper bound.

The split you drew matches `.github/dependabot.yml`: the four units left on 3.13.5 — `newspack-plugin`, `newspack-ads`, `newspack-newsletters`, `newspack-popups` — are exactly the `composer` entries there, alongside root. The ten here had no automatic path to the fix; the four get it on the next weekly run.

plugins/newspack-sponsors/composer.lock
adekbadek: Thanks — the count you measured is the thing that makes it land, and carrying it into NPPM-3056 as well is what a backport decision needs. Resolving.

adekbadek (approved):
The php_codesniffer 3.13.6 follow-up checks out on all ten locks: five changed lines each, no package but `squizlabs/php_codesniffer` moving, and no constraint in any lock capping it below 3.13.6. Both threads addressed and resolved.

PR #760, opened 2026-07-28
feat(subscriber-discounts): subscriber discounts on WooCommerce products (NPPD-1794)

adekbadek:
### All Submissions:

Closes NPPD-1794.

> [!NOTE]
> The shared subscriber-commerce base this builds on (#742) has landed on `main` and is merged in here, so the diff is the discounts feature alone.

Subscribers get a price on other products in the catalogue — the Access Control replacement for WooCommerce Memberships' *member discounts*. Six sites still depend on the WCM feature, so this is migration-blocking, not net-new.

A discount rule is: **who** gets it (one or more subscription products), **what** it covers (hand-picked products, a product category, or the whole catalogue, with exclusions), and **how much** off (fixed amount or percentage).

**Price layer.** The discount is presented as an apparent *sale price*, the same trick WCM uses — so WooCommerce and the theme render the strikethrough, the badge, and the cart line natively, and third-party price display keeps working. Filters sit late so we compose over whatever price is current rather than owning it.

**Reader-facing copy** follows Thomas's spec in the [design post](https://newspackdesignp2.wordpress.com/2026/07/09/subscriber-discounts-i1/): the badge reads "Subscriber discount", the product page spells out the saving, and the cart/checkout/order line says why the price is lower. Vocabulary is "subscriber" throughout — never "member".

**Migration.** `wp newspack migrate-discounts` ports WCM `purchasing_discount` rules. Dry-run by default (`--live` to write), idempotent via deterministic rule ids, preserves paused state, refuses to guess an audience for manually-granted plans, skips rules whose taxonomy we can't express (`product_tag`, attributes) with a reason rather than silently creating a rule that matches nothing, and reports where WCM's store-level defaults are **inverted** relative to ours. Migrating a site without reading that report can quietly change what subscribers pay.

**Store-level settings match Memberships.** "Apply on top of sale prices" defaults on, because Memberships' own switch is an *exclusion* defaulting to off — a store that never touched it discounts on-sale products. Both that setting and the "apply discounts when purchasing membership" upsell carry across in the migration.

**Overlapping discounts apply the best rule, not a compounded one.** WCM compounds by default: two 20% rules take 36% off, where the best rule alone takes 20%. Its off-switch is a code filter with no stored value, so no site can be read as having chosen compounding, and the pricing engine this feature will move to cannot express it. The migration command names the readers this costs — those holding two plans whose rules cover the same product — and prices the before and after for each. A read-only probe across the migration fleet found one affected site, on a single product; the detail is on that site's migration issue.

**Guards worth knowing about:**

- Subscriber prices never apply in wp-admin, nor on the REST routes that read a price in order to manage it (`wc/v3`, `wc-admin`, the block editor's product routes, and the discount editor's own product search). Without that, an administrator who also holds a subscription would see discounted prices on the products screen, Quick Edit would persist a subscriber price as everybody's price, and the editor's preview would compose a discount over an already-discounted base. The storefront's Store API is deliberately excluded and keeps reporting what the reader is charged.
- Variation price caches are keyed on the full rule content and settings, not just rule ids — editing a rule's amount has to bust the persistent `wc_var_prices_*` transient.
- The subscription that grants a discount is never itself discounted.
- Everything is behind `NEWSPACK_CONTENT_GATES`, and stands down entirely while WooCommerce Memberships is active, so a site mid-migration doesn't get two engines fighting over the price.

Two related items are out of scope and tracked separately: a Subscriptions overview tab (NPPD-2131), and the move onto the Dynamic Pricing engine, which is in beta and on no production site yet. NPPD-2193 tracks keeping this portable to it; the open gaps there are presentation parity — the engine substitutes catalogue prices for simple products only, and discloses the saving in its own block rather than through WooCommerce's sale markup. Worth a word with @miguelpeixe before either grows.

1. Check out this branch, set `define( 'NEWSPACK_CONTENT_GATES', true );`, and make sure WooCommerce Subscriptions is active and WooCommerce Memberships is **not**.
2. Create a subscription product (the audience) and a couple of ordinary products. Give a test reader an active subscription to the first.
3. Go to **Audience → Subscriptions → Subscriber discounts** → *Add discount*. Pick the subscription as the audience, target the ordinary products, and type an amount — the preview should show before/after for each covered product as you type. Save.
4. Logged out, visit the shop: list prices, no badge.
5. Logged in as the subscriber: badge, original struck through, subscriber price. Open a product page (the saving is spelled out), add to cart, and confirm the cart charges the subscriber price and names the discount.
6. Back in wp-admin as an administrator **who also holds the subscription**, open Products — prices must be the real ones, not discounted.
7. Open *Settings* on the tab. Try the sale-price and checkout-timing toggles.
8. Pause a rule from the list. Confirm the storefront reverts to list prices immediately.
9. Give the reader a second subscription. Add a second rule covering the same product.
10. Reload the shop. Confirm the better of the two prices applies, not both combined.
11. Activate WooCommerce Memberships and reload the shop — Access Control should stand down and leave pricing to Memberships.
12. Migration: on a site with WCM purchasing-discount rules, `wp newspack migrate-discounts` (read-only report), then `wp newspack migrate-discounts --live`. Run it twice; the rule count must not double.
13. `n test-php` from `plugins/newspack-plugin`.

### Verification

Walkthrough recorded against a local isolated env with a **real** WooCommerce subscription (not a stub) — authoring, storefront as anonymous vs subscriber, cart, settings, pausing, and the wp-admin guard:

[Subscriber discounts walkthrough](https://newspack-verification-artifacts.s3.us-east-2.amazonaws.com/pr-embeds/2026/07/28/fa0yrb4d-subscriber-discounts-walkthrough.mp4)

The editor's live preview, which the design is built around:

Discount editor with live before/after price preview

The storefront for a subscriber — rendered by WooCommerce's own sale markup, so the theme already styles it:

Shop as a subscriber: badge, struck-through list price, subscriber price

The same administrator, who also holds the subscription, still sees real prices in wp-admin:

wp-admin products list shows real prices for an admin who is also a subscriber

### Other information:

43 new unit tests; full suite green (3364 tests). Two of those tests exist because a review pass found real bugs: the sale price was being discounted **twice** (the product advertised a price below the one it charged — invisible in `get_price_html()`, exposed via the Store API), and an exception between suspend/resume left discounts disabled for the rest of the request. `tests/mocks/wc-mocks.php` now applies the WooCommerce price filters in `WC_Product::get_price()/get_sale_price()/is_on_sale()`; before that no test could see the filter chain at all, which is exactly why the double-discount hid.


thomasguillot:
Did a design pass over this to make sure everything sits right: tab order and naming, the editor in a drawer, the empty state, list details, and a single loading treatment while things fetch.

One small IA change on top of that: the discount settings (overlap, sale prices, timing) now live in Advanced Settings beside the upgrade link, saved from the header, instead of a modal on the list. The list screen only manages rules.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
```diff
@@ -0,0 +1,472 @@
...
+	 * itself is never discounted by its own rule: the reader is not a subscriber
+	 * of it yet, and discounting the thing that grants the discount is a loop.
+	 *
+	 * @param int         $user_id Reader.
+	 * @param array       $rule    Rule being considered.
+	 * @param \WC_Product $product Product being priced.
+	 * @return bool
+	 */
+	private static function reader_qualifies( $user_id, $rule, $product ) {
+		if ( Subscriber_Eligibility::user_has( $user_id, $rule['subscription_product_ids'] ) ) {
```
dkoo, line 420: **Severity:** blocker

**A whole-catalogue rule discounts the subscription that grants it, for readers who already hold it**

`reader_qualifies()` returns `true` as soon as `Subscriber_Eligibility::user_has()` confirms the reader holds the granting subscription — before the `product_is_one_of()` self-exclusion, which only runs in the apply-at-checkout branch below. So when a rule targets `all` (or a category that contains the subscription product), the granting subscription is discounted for its own current subscribers. The class docblock and NPPD-1794 both state this never happens.

**Observed:** in an isolated env (WooCommerce 11.0 + Subscriptions 9.1), with a single whole-catalogue 20%-off rule whose audience is a $10 subscription, a reader holding that subscription reads `get_price()` = **8.0** on the subscription product (should be 10.0). A non-audience product discounts correctly, and the invariant holds on the apply-at-checkout path — only the held-subscription case is affected.

**To reproduce:**
1. Enable Reader Activation + `NEWSPACK_CONTENT_GATES`, WooCommerce Subscriptions active, Memberships inactive.
2. Create a subscription product P and give a reader an active subscription to it.
3. Add a discount rule: audience = P, applies-to = All products, 20% off.
4. As that reader, read P's price (shop display or `wc_get_product(P)->get_price()`): it returns the discounted price, not the list price.

The existing test `test_the_granting_subscription_is_not_discounted_by_its_own_rule` uses `targeting: 'all'` but only exercises the cart / apply-at-checkout path with a non-subscriber, so it passes while this case slips through. Moving the `product_is_one_of()` exclusion ahead of the `user_has()` check closes it; a held-subscription assertion in that test would lock it down.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
```diff
@@ -0,0 +1,472 @@
...
+	 *
+	 * A subscriber discount belongs to the storefront. In wp-admin the same
+	 * price reads are how a shop manager edits the catalogue, and an
+	 * administrator who also holds a subscription would otherwise see — and,
+	 * through Quick Edit or a manual order, save back — their own discounted
+	 * price as the product's price.
+	 *
+	 * @return bool
+	 */
+	public static function should_adjust_prices_in_context() {
```
dkoo, line 96: **Severity:** blocker

**Admin-REST guard gap: the discount editor's live preview double-discounts for a subscribing admin**

`should_adjust_prices_in_context()` stands prices down for classic wp-admin (`is_admin() && ! wp_doing_ajax()`), a fixed admin-AJAX allowlist, and Quick/Bulk edit — all verified. The gap is authenticated REST: a REST request is neither `is_admin()` nor `wp_doing_ajax()`, so the guard returns `true` and discounts apply. This has a confirmed, user-visible consequence in the feature's own headline UI.

**Observed (manual QA), reproduced here.** The discount editor's product-search endpoint returns `(string) $product->get_price()` (`includes/wizards/audience/class-audience-subscriptions.php:333`). That call runs as the logged-in admin over REST, so when the admin holds a subscription the rule targets, the search returns the *already-discounted* price — and the editor's live before/after preview then applies the rule's discount a second time. With a 20%-off rule on a $20 and a $30 product, editing it as an admin who holds the granting subscription shows Coffee Mug **$12.80** and Notebook **$19.20** (20% off $16 and $24). I reproduced these exact figures: the search endpoint returns 16 / 24, and the preview re-applies 20% → 12.80 / 19.20. Storefront prices for the actual reader stay correct — the doubling only happens because the editor composes over a base that is itself already discounted. It reproduces deterministically whenever the acting admin holds a qualifying subscription; the intermittency a tester may see is consistent with per-request eligibility/caching timing, not with the endpoint behaviour, which is fixed.

**Second consequence (analysis).** The same gap lets WooCommerce's core REST (`wc/v3/products/<id>`, read in view context) return discounted prices; an admin-subscriber editing through the block-based product editor could then persist a discounted price as the stored price — the failure Quick Edit is explicitly defended against, via a different transport. The classic editor (the Newspack default) is safe: the wp-admin Products list shows real prices for an admin who holds a subscription (verified in the browser).

One guard change fixes both: because the storefront Store API (`wc/store/*`) *should* keep discounting, the fix needs to keep discounts on `wc/store` while standing them down for admin REST (`wc/v3/*`, `wp/v2` product routes) and the feature's own admin search endpoint — rather than gating all REST or none.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
```diff
@@ -0,0 +1,622 @@
...
+				// duplicate the whole rule set.
+				'id'                       => self::migrated_rule_id( $source_id ),
+				'subscription_product_ids' => $subscription_product_ids,
+				'targeting'                => $targeting,
+				'product_ids'              => Product_Targeting::TARGETING_PRODUCTS === $targeting ? $object_ids : [],
+				'category_ids'             => Product_Targeting::TARGETING_CATEGORY === $targeting ? $object_ids : [],
+				// Memberships excludes products from discounts with a per-product
+				// flag; subscriber discounts carry exclusions on the rule, so the
+				// flagged products are attached to every rule that could reach them.
+				'excluded_product_ids'     => Product_Targeting::TARGETING_PRODUCTS === $targeting ? [] : $excluded_product_ids,
```
dkoo, line 572: **Severity:** suggestion

**Migration drops per-product exclusions when the WCM rule targets specific products**

`'excluded_product_ids' => Product_Targeting::TARGETING_PRODUCTS === $targeting ? [] : $excluded_product_ids` carries per-product exclusions only for category/all rules. In WooCommerce Memberships, `is_product_excluded_from_member_discounts()` (the `_wc_memberships_exclude_discounts` flag) is a global gate applied before any rule matches, independent of targeting — so a product that is both explicitly named by a `purchasing_discount` rule and individually flagged excluded is *not* discounted under Memberships. After migration it sits on the rule's `product_ids` with no exclusion and would be discounted. Narrow (needs a product both explicitly targeted and flagged excluded) and softened by the dry-run/review flow, but it is a silent price change the report presents as a clean create — worth carrying the exclusion across for products-targeting too, or at least emitting a skip-note.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
```diff
@@ -0,0 +1,622 @@
...
+		$summary = [];
+		foreach ( $mapped['rules'] as $rule ) {
+			$row = [
+				'source'    => $rule['_source_rule_id'],
+				'plan'      => $rule['_source_plan_id'],
+				'audience'  => count( $rule['subscription_product_ids'] ),
+				'discount'  => 'percent' === $rule['discount_type'] ? $rule['amount'] . '%' : $rule['amount'],
+				'targeting' => self::describe_targeting( $rule ),
+				'active'    => $rule['active'] ? 'Y' : 'N',
+				'result'    => $dry_run ? 'would create' : 'created',
```
dkoo, line 100: **Severity:** suggestion

**Dry-run doesn't run store validation, so it can under-report what a live run would reject**

In dry-run, every mapped rule is reported as `would create`; `save_rule()` / `sanitize_rule()` only run under `--live`. The purpose of a dry run is to preview the live outcome, but a mapped rule the store would reject shows clean in the preview and only errors on the real run. The mapping is well-tested to produce store-valid rules today, so the risk is low — but the two paths can drift. Running each mapped rule through `sanitize_rule()` in dry-run (without persisting) and surfacing `would fail: <reason>` would keep preview and live in lockstep.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts.php
```diff
@@ -0,0 +1,441 @@
...
+		// subscriber price before they have checked out. Off in Memberships too.
+		'apply_at_checkout' => false,
+	];
+
+	/**
+	 * Every stored rule, newest first.
+	 *
+	 * @return array[]
+	 */
+	public static function get_rules() {
```
dkoo, line 97: **Severity:** suggestion

**Rule and settings reads are recomputed on the price hot path**

`get_rules()` re-reads the option and re-runs `array_map(fill_defaults) + array_filter + usort` on every call, and `get_settings()` re-runs `array_merge` + bool casts on every call, with no request-level memo. `get_subscriber_price()` calls `get_settings()` several times per invocation, and `filter_variation_prices_hash` (once per variable product) reaches `get_active_rules()` → a fresh remap+sort per product. The per-product Product_Targeting resolution is already memoized, so there is no repeated DB hit, but the map/sort/merge work scales with catalog size on a shop archive. A static per-request memo on `get_rules()` and `get_settings()` — flushed by the existing `flush_dependent_caches()` — would remove the repeated CPU cleanly. Modest impact given small rule counts; noted because this is the hottest path in the feature.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
```diff
@@ -0,0 +1,622 @@
...
+	private static function readers_losing_stacked_discounts( $rules_by_plan ) {
+		global $wpdb;
+
+		// Expand every rule to the products it covers, so two rules are compared
+		// on the product a reader actually buys rather than on their targeting —
+		// one may name a category where the other names a product.
+		$rules_by_plan_product = [];
+		foreach ( $rules_by_plan as $plan_id => $plan_rules ) {
+			foreach ( $plan_rules as $plan_rule ) {
+				foreach ( self::products_covered_by( $plan_rule ) as $product_id ) {
```
dkoo, line 319: **Severity:** suggestion

**The stacking-impact report under-counts the readers it exists to surface**

`readers_losing_stacked_discounts()` / `products_covered_by()` miss several real cases: catalog-wide Memberships rules return `[]` (no products enumerated), category expansion is capped at 500 products, variable parents priced via `get_regular_price()` return `''`→`0.0` and are skipped, and `GROUP_CONCAT( DISTINCT post_parent )` can truncate at the default `group_concat_max_len`. Each is documented in a code comment and this is an advisory report rather than the migration itself, so it is defensible — but a reader who stacks a catalog-wide discount against a product/category discount is exactly the 'pays more after the flip' population the report is meant to name, and they are silently absent. Worth a one-line caveat in the command's on-screen output (not just the comments) so the operator reads the count as a floor, not a total.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts.php
```diff
@@ -0,0 +1,441 @@
...
+		$discounted_price = 'percent' === ( $rule['discount_type'] ?? '' )
+			? $base_price * ( 1 - min( $amount, 100 ) / 100 )
+			: $base_price - $amount;
+
+		$discounted_price = max( 0, $discounted_price );
+
+		// Match WooCommerce's own rounding so a discounted price can never render
+		// with more precision than the store's currency format.
+		$decimals         = function_exists( 'wc_get_price_decimals' ) ? wc_get_price_decimals() : 2;
+		$discounted_price = round( $discounted_price, $decimals, PHP_ROUND_HALF_DOWN );
```
dkoo: **Severity:** nit

**Rounding mode diverges from WooCommerce's own**

`discounted_price()` rounds with `PHP_ROUND_HALF_DOWN`, but WooCommerce's money rounding (`wc_format_decimal` → `NumberUtil::round`) rounds half away from zero. On a half-cent boundary (e.g. $10.00 at 33.333% → 6.665) this feature charges 6.66 while any consumer recomputing the same percentage through WC's helpers lands on 6.67. Everything inside the feature stays internally consistent, so the only visible effect is a one-cent disagreement with an external recomputation; matching WC's mode removes even that.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	/**
+	 * Whether the product should be presented as on sale.
+	 *
+	 * @param bool        $on_sale Whether WooCommerce considers it on sale.
+	 * @param \WC_Product $product Product being priced.
+	 * @return bool
+	 */
+	public static function filter_is_on_sale( $on_sale, $product ) {
```
dkoo, line 313: **Severity:** nit

**filter_is_on_sale lacks the is_suspended() early-return its siblings have**

`filter_variation_prices_hash` short-circuits on `is_suspended()`; `filter_is_on_sale` doesn't. It isn't a bug — under suspension the inner `get_subscriber_price()` returns null immediately, so the nested `is_on_sale()` recursion terminates one level deep with the correct answer — but adding the guard at the top would make the suspend contract uniform across all filters and skip the extra nested evaluation during suspended reads.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
```diff
@@ -0,0 +1,472 @@
...
+
+	/**
+	 * Variation prices, used for a variable product's price range.
+	 *
+	 * @param string|float $price     Price WooCommerce is reporting.
+	 * @param \WC_Product  $variation Variation being priced.
+	 * @param \WC_Product  $product   Parent product.
+	 * @return string|float
+	 */
+	public static function filter_variation_prices( $price, $variation, $product ) { // phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
```
dkoo, line 269: **Severity:** nit

**Variation sale-price base can diverge from the price base under a third-party pricing extension**

`filter_variation_prices()` discounts the `$price` WooCommerce hands it (view context, already carrying any other extension's adjustment), while `filter_variation_sale_prices()` discounts `$variation->get_price( 'edit' )` (raw). With no other pricing extension the two bases are equal and the strike-through renders correctly, and using `edit` is the right way to avoid recursion. But if another extension filters the view price, the two bases diverge, the resulting `sale_price` array no longer equals the `price` array, and the struck range can drop. Low confidence it matters in practice — flagging it to confirm against whatever pricing plugins your publishers actually run; reading the view price under a `suspend()` bracket in the sale filter would keep the bases aligned if it does.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/discounts/index.tsx
```diff
@@ -0,0 +1,355 @@
...
+		() => [
+			{
+				id: 'edit',
+				label: __( 'Edit', 'newspack-plugin' ),
+				isPrimary: true,
+				callback: ( [ item ]: DiscountRule[] ) => history.push( `${ baseUrl }/edit/${ item.id }` ),
+			},
+			{
+				id: 'toggle-active',
+				label: __( 'Pause or resume', 'newspack-plugin' ),
```
dkoo: **Severity:** nit

**Row action and status copy diverges from other Newspack wizard pages**

The per-row action reads "Pause or resume" (`:209`) and the status reads "Paused" (`:163`), which don't match the active/inactive language other Newspack wizard pages use. Consider "Set to inactive" when active and "Set to active" when paused, and renaming the "Paused" status to "Inactive", so the toggle and its state read consistently with the rest of the admin. Non-blocking.

dkoo (changes requested):
Release risk: HIGH. This is a careful, well-tested feature and the hard seams — sale-price presentation, suspend/resume re-entrancy, best-rule-not-compounded, the variation cache key — all hold up at runtime. Two correctness bugs are worth fixing before it ships, and both close with small changes in the same two methods:

1. A whole-catalogue or category rule discounts the very subscription that grants it, for readers who already hold it — which the code documents it never does. Reproduced in an isolated env ($10 membership → $8 for its own subscriber).
2. The admin-context guard doesn't cover REST, so the discount editor's own product search returns discounted prices to an admin who holds a qualifying subscription, and the live preview then double-discounts (a 20% rule on $20/$30 shows $12.80 / $19.20). Found in manual QA and reproduced exactly. The same gap opens a persist-discounted-price path through admin REST.

The remaining comments are suggestions and nits.

Verified at runtime (WooCommerce 11 + Subscriptions 9.1, real subscription): subscriber discount applies ($20 → $16); sale price equals charged price and the Store-API price agrees with get_price() (no double-discount); overlapping rules apply the single best price ($12, not compounded); anonymous visitors see list prices; and the classic wp-admin Products list shows real prices for an admin who also holds a subscription.

- `plugins/newspack-plugin/src/wizards/audience/views/subscriptions/audience-management-prerequisite.test.js:74` — **Nit:** Stale tab slug in an unrelated test — This test still mocks a tab `{ slug: 'configuration', label: 'Configuration' }`, a slug this PR renamed to `advanced-settings`. The test only needs *some* tab and doesn't assert on the real Configuration tab, so it still passes — the name is just misleading now. Harmless; update when convenient.

---

*Report-only note (the file isn't in this diff, so it can't be an inline comment):* **Stale tab slug in an unrelated test** — This test still mocks a tab `{ slug: 'configuration', label: 'Configuration' }`, a slug this PR renamed to `advanced-settings`. The test only needs *some* tab and doesn't assert on the real Configuration tab, so it still passes — the name is just misleading now. Harmless; update when convenient.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
adekbadek: Fixed in 3b9c04f: the `product_is_one_of()` self-exclusion now runs first in `reader_qualifies()`, so it applies to the held-subscription path as well as the cart one.

The existing test gained the assertion you described — a reader holding the granting subscription reads the list price on it — and it reproduces your figures before the fix (120.0 → 108.0 at 10%).

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
adekbadek: Fixed in 3b9c04f. `should_adjust_prices_in_context()` now also stands down for management REST routes: `wc/v1|v2|v3`, `wc-admin`, `wc-analytics`, `wp/v2/product*`, and `newspack/v1/wizard/*`. `wc/store/*` is deliberately absent and keeps discounting. The list is filterable via `newspack_subscriber_discounts_management_rest_routes` for a site with another admin route. The route is read at price time rather than at registration, since `wp_loaded` runs before WordPress has resolved one.

One related fix came out of it: `filter_variation_prices_hash` was still varying the key by the reader's entitlement in contexts where prices are not adjusted, so an admin-REST read would have written list prices into that reader's variation-price bucket and served them on the storefront afterwards. It now leaves the hash alone whenever prices stand down.

Verified over real HTTP on an isolated env, with the acting admin holding the audience subscription and a 20%-off rule on a $100 product: the wizard product search and `wc/v3/products/12` both report 100, while `wc/store/v1/products/12` reports `price 8000, regular_price 10000`. Reverting just the guard puts the search endpoint back to 80, which is the double-discount you saw. A unit test pins all three routes.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
adekbadek: Carried across in 3b9c04f, though not as an exclusion: `sanitize_rule()` drops `excluded_product_ids` for a products-targeted rule, so attaching them there would be silently discarded on save. The flagged products come off `product_ids` instead, which is the same outcome. A rule left with no products at all is now skipped with a reason rather than saved as a rule that discounts nothing.

The existing exclusion test covers both: a hand-picked list of `[101, 999]` with 999 flagged migrates as `product_ids => [101]`, and a list of `[999]` alone lands in the skipped table.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
adekbadek: Done in 3b9c04f. A dry run now puts every mapped rule through `sanitize_rule()` without persisting, and reports `would fail: <reason>` in the result column. Those rows count towards the error tally in the closing summary, so the dry-run and live counts line up.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts.php
adekbadek: Memoized in 3b9c04f. `get_rules()` and `get_settings()` each keep a per-request memo, so the map/sort and the merge/cast happen once regardless of catalogue size.

Invalidation hangs off `added_option`/`updated_option`/`deleted_option` rather than only `flush_dependent_caches()`, so a caller that writes either option straight through the options API — the test suite does — cannot leave a stale memo behind.

plugins/newspack-plugin/includes/cli/class-discounts-migration.php
adekbadek: Added the caveat to the on-screen output in 3b9c04f, on both branches — the "no reader affected" line is a floor for the same reasons as the count:

> Overlapping discounts are compared per product, so this count is a floor: catalog-wide rules, categories past the first 500 products, and variable parents priced only through their variations are not counted.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts.php
adekbadek: Switched to `PHP_ROUND_HALF_UP` in 3b9c04f, which for a non-negative price is the half-away-from-zero WooCommerce uses. Your 6.665 case now lands on 6.67.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
adekbadek: Added in 3b9c04f — `filter_is_on_sale()` returns the incoming value untouched under suspension, so the contract reads the same across the filters and a suspended read skips the nested evaluation.

plugins/newspack-plugin/includes/subscriber-commerce/class-subscriber-discounts-pricing.php
adekbadek: Checked against the WooCommerce source, and the mechanism is slightly different from the one you describe: `WC_Product_Variable_Data_Store_CPT::read_price_data()` passes `$variation->get_price( 'edit' )` to `woocommerce_variation_prices_price`, not the view price. So the incoming `$price` is the raw stored figure, identical to what the sale filter reads, and an extension filtering the *view* price cannot separate them. The stale comment claiming otherwise is gone.

The divergence is real by another route, though: an extension filtering `woocommerce_variation_prices_price` itself at a priority below ours hands us an adjusted base while the sale filter still reads raw, and then `is_on_sale()` fails its `sale_price === price` test and the struck range drops. Closed in 3b9c04f by having `filter_variation_prices()` record the base it was handed, keyed by variation id, and `filter_variation_sale_prices()` discount that same figure — WooCommerce calls the two for one variation in the same loop iteration, and it falls back to `get_price( 'edit' )` if the price filter never ran.

Reading the view price under a `suspend()` bracket would not have worked here: it would introduce a divergence rather than remove one, since neither filter is handed a view price to begin with.

plugins/newspack-plugin/src/wizards/audience/views/subscriptions/tabs/discounts/index.tsx
adekbadek: Renamed in 3b9c04f: the row action is now "Set to inactive" / "Set to active" depending on the rule's state, and the status column and its filter read "Inactive", matching the content gates screens.

The Subscriber-only products tab next door had the same mismatch — "Pause"/"Resume" and a "Paused" badge — so it moved with it rather than leaving two tabs of one screen disagreeing. That file is not otherwise part of this PR.

adekbadek:
Updated the stale tab slug in `audience-management-prerequisite.test.js` — the mocked tab is now `{ slug: 'advanced-settings', label: 'Advanced Settings' }`, and the fixture text and file docblock follow it. 3b9c04f.

adekbadek:
@dkoo Re-requesting review – the changes you asked for have been pushed and need a look before this merges:
- fix(subscriber-discounts): guard management REST and self-discounting rules (NPPD-1794, #760)

dkoo (approved):
Re-reviewed the fixes in `3b9c04fe2`. Both blockers are resolved, and I re-verified each in an isolated env:

1. The granting subscription is no longer discounted by a whole-catalogue rule — `reader_qualifies()` now excludes it on every path, and it prices at $10 (was $8) for a reader who holds it.
2. The admin-context guard now covers management REST via `is_management_rest_request()`: the discount editor's own product-search endpoint stands down (returns list prices), so the live preview no longer double-discounts, and the `wc/v3` persist path is closed — while `wc/store` keeps discounting for the storefront. Confirmed all four route behaviours at runtime.

Nearly every suggestion and nit is addressed as well (rule/settings memoization, the three migration items, the `filter_is_on_sale` and variation-base nits, the test slug, and the row/status copy). The one thing left is the rounding-mode nit, which is a reasonable keep. Feature test suite is green (38 tests). Thanks for the thorough turnaround — this looks good to me.

PR #762, opened 2026-07-28
test(scripts): collect .ts/.tsx suites in jest testMatch

dkoo:
## What this does and why

Newspack's Jest runner only collected `*.test.js` / `*.test.jsx` files, so TypeScript test files (`*.test.ts` / `*.test.tsx`) were never picked up. They weren't reported as skipped — they simply never appeared in the suite list, and the run still went green. A dead test file is worse than no test file: it reads as coverage in review and CI while asserting nothing.

Today this affects one real suite: `plugins/newspack-plugin/src/content-gate/editor/block-visibility.test.ts` — the only tests for the block editor's Access Control visibility panel (7 assertions), which had never run.

## Changes proposed

Widen the `testMatch` glob in `newspack-scripts` so it collects TypeScript test files alongside JavaScript:

```diff
- testMatch: [ '<rootDir>/**/*test.js?(x)' ],
+ testMatch: [ '<rootDir>/**/*test.[jt]s?(x)' ],
```

The Babel transform already compiles `.ts`/`.tsx`, so collection was the only missing link. `[jt]s?(x)` is a strict superset of the old pattern for `.js`/`.jsx`, so no existing suite's collection changes — the only new behaviour is that TypeScript test files are now collected.

## How to test

1. Check out this branch and run the newspack-plugin JS tests.
2. Confirm `block-visibility.test.ts` now appears in the suite list and its 7 tests pass.
3. Confirm the rest of the JS suite is unaffected — same suites as before, all still passing.

To see the original behaviour on `main`: that same file never appears in the suite list, yet the run reports green.

Verified locally: under the previous pattern the suite is collected 0 times; under the new pattern it is collected and passes 7/7. A before/after run of the full newspack-plugin JS suite showed an identical set of suites plus the newly-collected TypeScript one — no other suite changed status (11 pre-existing local suite-load failures are environmental to a shared checkout and are unchanged by this diff).

## Follow-up (not in this PR)

`collectCoverageFrom` still globs only `{js,jsx}`, so TypeScript sources will not count toward coverage even with their tests now running. Worth a separate change.

Refs [NPPM-3057](https://linear.app/a8c/issue/NPPM-3057).


dkoo:
Again, a pretty benign internal-tooling-only change. Merging once CI reports green!

PR #764, opened 2026-07-29
chore: retire the orphaned cghooks hook subsystem

jason10lee:
### All Submissions:

Related but non-overlapping: #758 changes only `composer.lock` files; this one changes no lockfile at all. See "Relationship to #758" below.

Retires the `brainmaestro/composer-git-hooks` machinery, which is inert in this repo and actively breaks `composer update`.

**The failure.** Ten units declared a `post-update-cmd` running `vendor/bin/cghooks update` while the package appeared in neither `require` nor `require-dev`. `composer update` in any of them wrote the lockfile correctly and then exited **127**:

```
sh: vendor/bin/cghooks: No such file or directory
Script vendor/bin/cghooks update handling the post-update-cmd event returned with error code 127
```

The write succeeds and only the hook script dies, so the non-zero exit is a false failure. `plugins/newspack-story-budget/AGENTS.md` had this documented as a standing workaround ("Use `composer install` (warns but succeeds)") rather than fixed.

**What is removed.**

| | |
| --- | --- |
| `post-update-cmd` invoking cghooks | 10 `composer.json` files |
| `brainmaestro/composer-git-hooks` from root `require-dev` | 1 |
| `pre-push` script entries | newspack-ads, newspack-blocks |
| `.hooks/pre-push` scripts | all 10 units |
| `.hooks` entries in `.distignore` | newspack-plugin, newspack-story-budget |
| stale workaround + dead-config notes | newspack-story-budget `AGENTS.md` |

**Why the hook scripts go too.** All ten `.hooks/pre-push` files are byte-identical and refuse pushes to a branch named `trunk`. They are dead three times over: this repo's branch is `main`; husky owns `core.hooksPath` and the only tracked husky hook is `pre-commit`; and cghooks v3 reads hook definitions from `extra.hooks`, which no unit defines — so even with the package installed, the `scripts` entries would have installed nothing. Eight of the ten were never referenced from `composer.json` at all, and eight plugins were shipping the dead script inside their release zips. Since verified against the installed v3.0.0: `Hook::getValidHooks()` reads `$json['extra']['hooks']` and nothing else, so the `scripts` entries in `newspack-ads` and `newspack-blocks` were never a hook definition — the method's "Get scripts section" docblock is a leftover from v2 and describes behaviour the code no longer has. The standalone repos where those entries could in principle have mattered are archived as of 2026-08-05/06 and read-only, so no reachable target survives either way.

**One deliberate exception.** `themes/newspack-theme` keeps its `post-update-cmd`; that array also holds `PHPStan\WordPress\Composer\FixWpStubs::php73Polyfill`, backed by a real `szepeviktor/phpstan-wordpress: ^2.0` dependency. Only the cghooks element is removed there.

No `composer.lock` is touched. `scripts` is not part of composer's content hash, so the locks stay in sync.

No GitHub issue and no Linear ticket — this surfaced while working NPPM-3056.

1. Reproduce on `main`: copy `plugins/newspack-blocks/composer.json` and `composer.lock` into a scratch directory and run `composer update --no-install`. It exits **127**.
2. Repeat with this branch's `composer.json` and the same lockfile. It exits **0**.
3. Confirm the root still resolves without the dropped dependency: `composer update --dry-run` at the workspace root completes with no conflict.
4. Confirm nothing is left behind: `git grep -n 'cghooks\|composer-git-hooks'` and `git grep -n '\.hooks/'` should return no hits outside `wp.hooks` / `compiler.hooks` JavaScript API calls.
5. Confirm the exception survived: `themes/newspack-theme/composer.json` still has `post-update-cmd` holding the PHPStan entry.
6. Confirm no lockfile churn is required: `composer validate --no-check-all --no-check-publish` in the ten units. Four report clean; the other six report a pre-existing "lock file is not up to date" that reproduces byte-identically against unmodified `main` and is cleared by #758.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — not applicable; this removes dead build configuration and no product code.

What I ran: the before/after `composer update` comparison (127 → 0), a root `composer update --dry-run` after dropping the dependency, `composer validate` across the ten units, a control run against pristine `main` to confirm the six lock warnings predate this change, and a re-parse of every edited JSON file.

Reviewed by Copilot (10/10 files, no comments) and by an independent multi-model review, which caught two gaps now folded in: the orphaned root `require-dev` entry, and the `pre-push` entries left unreachable once their installer was removed. That review also raised, and then disputed under adversarial re-review, a claim that this removal breaks a working cghooks integration — the package is absent from every unit's `require`/`require-dev`, so there is no integration to break.

**Relationship to #758.** In six units `brainmaestro/composer-git-hooks` still sits in `composer.lock` as a stale entry that `composer.json` never asked for, so `composer install` pulls it in and the hook happens to work — which is why only 4 of the 10 units fail today. #758 re-locks those six and removes that accident, taking `composer update` failures from 4/10 to 10/10. That is a correct change exposing a real bug, not a regression. This PR is the fix, and the two touch no files in common, so they can land in either order — though landing this one first avoids the window where all ten fail.


jason10lee:
Checked the lock side of this. On `main`, all six units whose `post-update-cmd` this removes still carry `brainmaestro/composer-git-hooks` in their `composer.lock` as a top-level `packages-dev` entry that nothing requires. Removing a `scripts` block doesn't change the lock's `content-hash`, which is computed from the require sections, so Composer never warns about the drift.

#758 already clears it. Its `composer update wp-coding-standards/wpcs -W` re-resolves those same six locks and drops the orphaned entry as a side effect—one occurrence on `main`, zero on #758, for each of the six.

So this is a merge-order note rather than something to change here. If #758 lands first the sequence is clean. If this lands first, the `AGENTS.md` line it deletes—describing the package as orphaned in the lock but absent from `require-dev`—stays true until #758 follows, so the deletion runs slightly ahead of the facts.

PR #765, opened 2026-07-29
feat(access-control): make Audience Management an explicit prerequisite (NPPD-1846)

adekbadek:
### All Submissions:

Access Control appears whenever `NEWSPACK_CONTENT_GATES` is defined and never checked Reader Activation. The ticket describes the symptom as broken UI in `/my-account`, but that undersells it: `class-content-gate.php` has no Reader Activation dependency at all, so gates restrict content perfectly well — while everything a gate hands the reader off to is gated on Audience Management:

| What | Where |
| --- | --- |
| Reader registration hooks | `class-reader-registration.php:28` |
| Magic links (passwordless login) | `class-magic-link.php`, 7 guards |
| Newspack My Account shell | `class-my-account.php:116` |
| Auth / account emails | `class-reader-activation-emails.php:39` |
| Reader session hydration | `class-session-hydration.php:102` |
| ESP contact sync | `sync/class-sync.php:76` |

A registration wall built without Audience Management locks readers out and then gives them no way in.

This lands two halves that are only correct together.

**1. The gate-editing screens refuse to work without Audience Management.** A shared `Audience_Management_Dependency` trait — consumed by *both* surfaces, so the dependency can't be enforced on one and forgotten on the other — provides the check, the localized script data, and a REST permission callback guarding gate **creation and duplication**. The screens are replaced by a prerequisite state built on the existing Institutions empty-state pattern. The submenu item deliberately stays visible: hiding it would turn "is this required?" into "where did it go?", which is the opposite of making the dependency explicit.

The guard is applied at the **wizard router**, so all 7 Access Control sections and both Premium Newsletters sections inherit it. Guarding only the gate list would have left `#/edit/new/all` reachable by bookmark or browser history, with a working Save — the publisher fills in an entire gate and only learns on save.

**Premium Newsletters is included because the dependency there is real, and harder.** `filter_subscription_lists()` is hooked unconditionally, so restricted lists vanish from every signup form whether or not Audience Management is on; meanwhile access is granted only via the `reader_verified` / `product_subscription_changed` / `donation_subscription_changed` data events, and `reader_verified` traces to `set_reader_verified()`, whose every trigger hook is registered inside `if ( self::is_enabled() )`. With Audience Management off, a premium newsletter is hidden from readers with no path to unhide it.

**2. Audience Management can no longer be switched off while any gate exists.** `Content_Gate::has_any_gates()` plus a 409 refusal in `api_update_reader_activation_settings()`, and a toggle that explains the lock instead of letting the publisher confirm a destructive dialog and then meet a REST error.

This second half is what makes the first one safe. **The ticket's premise that nothing in the admin turns Audience Management off is wrong** — Audience → Configuration → Reader Revenue Platform → *Change* → toggle calls `saveConfig( { enabled: false } )` behind a confirm dialog. Without the disable guard, a site that used it would keep serving live gates while `Content_Gate::redirect_cpt()` bounced the gate CPT list back to the now-blocked wizard and `filter_edit_post_link()` rewrote every gate edit link to a blocked route — live paywalls with no admin surface left to lift them. The guard gives the invariant **Audience Management off ⇒ no gates exist**, which is exactly what the wholesale screen replacement relies on.

The guard counts drafts and premium newsletter gates (a draft is publishable; a newsletter gate restricts by the same mechanism) but **not** trashed gates — a gate the publisher already deleted must not hold the toggle hostage.

Closes NPPD-1846.

Supersedes the prototype in #737, now closed.

1. On a site with `define( 'NEWSPACK_CONTENT_GATES', true );` in `wp-config.php`, ensure Audience Management is **off** (`wp option update newspack_reader_activation_enabled 0`).
2. Visit **Audience → Access Control**. You should see the "Set up Audience Management first" state, with no "Add Content Gate" action, and the submenu item still present. Same at **Newsletters → Premium**, with newsletter-specific copy.
3. Try the routes directly — `#/edit/new/all`, `#/edit/new/custom`, `#/institutions`, `#/settings/countdown-banner`, `#/settings/content-gifting`. All should show the prerequisite state.
4. From the browser console, confirm the REST guard: a `POST` to `/wp-json/newspack/v1/wizard/newspack-audience-access-control` with a valid nonce returns **403** `newspack_audience_management_required`, and no gate is created. Same for `.../newspack-premium-newsletters`.
5. Turn Audience Management on. Every screen and route renders normally; gate creation returns 200.
6. Create a gate, then go to **Audience → Configuration → Reader Revenue Platform → Change**. The Audience Management toggle is disabled and explains why. A `POST` of `{ enabled: false }` to `/wp-json/newspack/v1/wizard/newspack-audience/audience-management` returns **409** `newspack_audience_management_required_by_gates`.
7. Delete the gate and reload. The toggle is interactive again.

### Verification

Audience Management **off** — Access Control and Newsletters → Premium:

Access Control screen with Audience Management off: prerequisite state, no Add Content Gate action

Newsletters > Premium with Audience Management off: prerequisite state with newsletter-specific copy

The gate editor route is covered too, not just the list:

The gate editor route #/edit/new/all is blocked too, not just the gate list

Audience Management **on** — nothing is over-blocked:

With Audience Management on, the Access Control screen renders normally

The disable guard, locked while a gate exists and released once it is gone:

Audience Management toggle disabled while a content gate exists, with an explanation

Audience Management toggle interactive again once the gate is deleted

Automated:

```
PHP:  Tests: 2470, Assertions: 7029, Skipped: 7   (0 failures; the 7 skips pre-date this branch)
JS:   Test Suites: 54 passed, Tests: 496 passed   (0 failures)
PHPCS: clean on all 8 changed PHP files under BOTH the plugin ruleset and the
       workspace-root VIP ruleset the pre-commit hook uses
ESLint: 0 errors
```

Both regression tests were confirmed to actually bite: reverting a single `permission_callback` token turns the route-wiring test red, and un-wrapping one router section turns the section-completeness test red.

### Other information:

**Rollout safety.** Every one of the 19 production sites that currently has Audience Management off was checked directly: all have **zero** content gates, so no live site meets the newly-blocked condition and nothing needs migrating. `NEWSPACK_CONTENT_GATES` is a `wp-config.php` constant marked `@status draft`, so Access Control only exists where someone deliberately defined it.

**Cross-repo note.** `includes/content-gate/class-content-gate.php` is in the `content-gate` contract consumed by `newspack-manager`. The change there is purely additive (one new static method, `has_any_gates()`); no existing signature or behaviour changes, so no coordinated PR is needed.

**Deliberately not included.** The [Access Control help doc](https://help.newspack.com/access-control/) still doesn't list Audience Management as a dependency — the question that started the thread. Not a code change, and still unowned.

**Reviewer note.** Codex was unavailable during review (timed out twice without producing a review body), so that reviewer slot is degraded rather than clean.


plugins/newspack-plugin/src/wizards/audience/views/content-gates/index.js
```diff
@@ -48,19 +49,19 @@ const AudienceContentGates = ( props, ref ) => {
 			sections={ [
 				{
 					path: '/content-gates',
-					render: ContentGates,
+					render: requireAudienceManagement( ContentGates ),
```
dkoo: **Severity:** blocker

**Blocker:** The prerequisite state replaces `/content-gates` — the gate list — not just the create and edit routes, and the justification is the invariant in `has_any_gates()`'s docblock. That invariant only covers transitions from here forward. Reproduced on a site already in the pre-existing state (Audience Management off, one published gate): Access Control renders the same screen as a site with no gates, `edit.php?post_type=np_content_gate` redirects to it via `redirect_cpt()`, and `filter_edit_post_link()` rewrites every gate edit link to it. The DELETE route stays open by design, but nothing discoverable can call it.

Resolved in the stacked PR by making gates inert when Audience Management is off, so replacing the list strands nothing.

plugins/newspack-plugin/includes/wizards/audience/class-audience-wizard.php
```diff
@@ -590,6 +594,22 @@ private static function get_verification_required_gates_with_edit_urls(): array
...
 		$args = $request->get_params();
+
+		// Audience Management is a hard prerequisite for content gating (NPPD-1846).
+		// Gates carry no Reader Activation dependency of their own, so switching it off
+		// here would leave every existing gate restricting content while readers lose
+		// the registration, sign-in and account surfaces they need to satisfy it — and
+		// the gate screens, which replace themselves with the prerequisite state, could
+		// no longer be used to lift those restrictions. Refusing the transition keeps
+		// "Audience Management off" and "no gates exist" the same state.
+		if ( isset( $args['enabled'] ) && ! $args['enabled'] && Content_Gate::has_any_gates() ) {
```
dkoo: **Severity:** blocker

**Blocker:** `isset()` is false for an explicit null, so `{"enabled": null}` skips the 409 and falls through to the write below. Confirmed against a running site: 200, option left empty, `is_enabled()` reads it back as off with a published gate present. `false`, `0` and `'0'` all correctly 409.

This route registers no `args` schema, while its siblings at lines 483 and 505 coerce the same key through `Newspack\newspack_string_to_bool`. Moot in the stacked PR, which removes the guard in favour of the confirmation modal.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -1878,6 +1878,43 @@ public static function set_presave_checks_enabled( $enabled ) {
...
+		// resolve trashed gates. A trashed gate restricts nothing and the publisher has
+		// already deleted it, so counting it here would hold the Audience Management
+		// toggle hostage to a gate that no longer applies.
+		$statuses = array_diff( self::get_post_statuses(), [ 'trash' ] );
+		if ( empty( $statuses ) ) {
+			return false;
+		}
+		$gates = get_posts(
+			[
+				'post_type'      => self::GATE_CPT,
```
dkoo: **Severity:** suggestion

**Suggestion:** The query covers `self::GATE_CPT` only, while `get_gate_post_types()` also returns `Memberships::GATE_CPT` when Woo Memberships is active. Confirmed the gap: a published `np_memberships_gate` leaves this false and the disable request returns 200.

Scoping to Access Control gates is the right call, though — `Content_Restriction_Control::is_post_restricted()` returns early when Memberships is active, Memberships hooks its own callback onto `newspack_is_post_restricted`, and there are no `Reader_Activation` references under `includes/plugins/wc-memberships/`. So Memberships restriction keeps enforcing regardless. It's the docblock's "of either kind" and "always implies no gates exist" that overstate. Moot in the stacked PR, which removes this method.

plugins/newspack-plugin/includes/wizards/audience/class-audience-wizard.php
```diff
@@ -590,6 +594,22 @@ private static function get_verification_required_gates_with_edit_urls(): array
...
+			return new \WP_Error(
+				'newspack_audience_management_required_by_gates',
+				__( 'Audience Management cannot be disabled while content gates exist. Delete your gates under Audience → Access Control first.', 'newspack-plugin' ),
```
dkoo: **Severity:** suggestion

**Suggestion:** `has_any_gates()` deliberately counts premium newsletter gates, but this message and the toggle copy at `platform-selection/index.js:154` both send the publisher to Access Control, whose list excludes them. Someone whose only gate is a premium newsletter follows the instruction to a screen that looks empty. Moot in the stacked PR, which removes the refusal.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/index.js
```diff
@@ -48,19 +49,19 @@ const AudienceContentGates = ( props, ref ) => {
...
 					breadcrumbs: ACCESS_CONTROL,
 				},
 				{
 					path: '/edit/:id/:type?',
-					render: Edit,
+					render: requireAudienceManagement( Edit ),
```
dkoo: **Severity:** suggestion

**Suggestion:** `requireAudienceManagement()` runs inside the component body, so every render mints a fresh component type and `Wizard` renders it as `<SectionComponent />` — a changed type at that position remounts the section subtree and discards in-progress editor state. Latent today, since nothing re-renders the router, but `withWizard` calls `setState` on `setError`, `confirmAction` and every fetch, so the first section reaching for one of those props would start wiping state mid-edit. Hoisting the wrapped components to module scope restores what the module-level imports gave you.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/audience-management-prerequisite.test.js
```diff
@@ -0,0 +1,144 @@
...
+	// routes it thinks to visit — which is the same blind spot that let #/edit through.
+	const fs = require( 'fs' );
+	const path = require( 'path' );
+
+	it.each( [
+		[ 'Access Control', path.join( __dirname, 'index.js' ) ],
+		[ 'Premium Newsletters', path.join( __dirname, '../../../newsletters/views/premium-newsletters/index.js' ) ],
+	] )( '%s wraps every section renderer', ( _name, routerPath ) => {
+		const source = fs.readFileSync( routerPath, 'utf8' );
+		const renderers = source.match( /render: [^,]+,/g ) || [];
```
dkoo: **Severity:** suggestion

**Suggestion:** `source.match( /render: [^,]+,/g )` only matches a renderer followed by a comma, so a section whose `render` is the final property contributes nothing and the `unguarded` assertion passes against an empty list. wp-prettier's `trailingComma: 'es5'` makes that unlikely, but the file's own header calls this the load-bearing assertion, and it can currently count zero without failing. A `toHaveLength()` anchor, or asserting on the exported `sections` array's `render.displayName`, would check the same property against real values.

plugins/newspack-plugin/src/wizards/types/window.d.ts
```diff
@@ -82,6 +82,12 @@ declare global {
 			presave_checks_enabled: boolean | string;
 			default_gate_status: GateStatus;
 			feed_restriction_modes?: { value: FeedRestrictionMode; label: string }[];
+			// Audience Management is a prerequisite for content gates. Only ever the
```
dkoo, line 85: **Severity:** nit

**Nit:** Both keys are declared required, but the test at `audience-management-prerequisite.test.js:122` covers the key being absent, and the component defends with `?.` and `|| ''`. Marking them optional would match the behaviour the code and tests already assume. The comment explaining why the type is `string` rather than `boolean` is worth keeping.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/audience-management-required.tsx
```diff
@@ -0,0 +1,85 @@
...
+ */
+import { Button, Grid, SectionHeader } from '../../../../../packages/components/src';
+import { hasAudienceManagement } from './utils';
+
+const AudienceManagementRequired = ( { isNewsletter = false }: { isNewsletter?: boolean } ) => {
+	const audienceManagementUrl = window.newspackAudienceContentGates?.audience_management_url || '';
+
+	return (
+		<Grid columns={ 4 } noMargin>
+			<VStack start={ 2 } end={ 4 } spacing={ 8 }>
```
dkoo, line 27: **Severity:** nit

**Nit:** `Wizard` renders `section.title` and `section.description` above the section component, so `/settings/countdown-banner`, `/settings/content-gifting` and `/institutions/new` show that heading and its descriptive copy — implying the feature is configurable — immediately above this one. The stacked PR redirects those sub-routes to `/content-gates` so this renders in exactly one place.

dkoo (commented):
Reviewed this in an isolated environment with `NEWSPACK_CONTENT_GATES` defined, exercising both the REST routes and the admin screens.

The enforcement works as designed. The trait composes with the wizard capability check rather than replacing it, 403 and 409 are the right statuses, the localized flag is read in the wire format `wp_localize_script` produces and fails closed when it's absent, and all seven Access Control sections plus both Premium Newsletters sections are wrapped. `has_any_gates()` behaves exactly as documented — true for published, draft and private gates, false for trashed and auto-draft ones. The docblocks explaining *why* each decision was made are what made the two findings below findable at all.

Both were reproduced against a running site.

**The invariant doesn't hold for the installed base.** "Audience Management off implies no gates exist" is enforced only on transitions from here forward. I put a site into the pre-existing state — Audience Management off with a published gate — and opened Access Control: the screen is identical to the no-gates one, and `edit.php?post_type=np_content_gate` redirects to that same blocked screen. The gate keeps restricting content throughout, since `Content_Gate::restrict_post()` never consults Reader Activation.

**`isset()` lets an explicit null past the disable guard.** A `{"enabled": null}` body returns 200 and disables Audience Management with a published gate in place. `false`, `0` and `'0'` all correctly return 409.

One claim worth clearing rather than leaving open: sending `enabled` as the string `'false'` does skip the 409, but `is_enabled()` reads `'false'` as truthy, so Audience Management stays on. That path fails safe and needs no change.

### Where this is going

Rather than send these back, we talked the design through and landed somewhere simpler. Since @adekbadek is out, we'll implement it ourselves in a PR stacked on this branch, merged into it before this one squash-merges.

The shift: instead of guarding *against* the gates-without-Audience-Management state, make that state harmless. Short-circuiting `Content_Restriction_Control::is_post_restricted()` on `Reader_Activation::is_enabled()` makes gates inert whenever Audience Management is off — and it covers premium newsletters for free, since `filter_subscription_lists()` calls that same method. Nothing can be stranded, so the whole disable guard comes out: `has_any_gates()`, the 409, `disabling_blocked_by_gates`, the locked toggle, and its test file. The existing "Disable Audience Management?" modal takes over the job, with copy naming content gates and premium newsletters.

That resolves both findings above and moots three more — the Memberships scope question, the refusal message's destination, and the missing feature-flag awareness.

The prerequisite screens stay. They're the part of this branch that answers NPPD-1846, and hiding the menu instead would have made the dependency less discoverable rather than more. They get restyled onto the existing onboarding-card pattern with a single "Enable Audience Management" action, and the sub-routes redirect to `/content-gates` instead of each rendering their own copy, which also clears the stacked-header problem below.

The remaining inline comments are polish we'll fold into the same PR.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/index.js
dkoo: Resolved in `c8d9f841`. The prerequisite screen still replaces the list, but the reason it is safe has changed: gating now stands down entirely when Audience Management is off, so a stranded gate restricts nothing. `Content_Restriction_Control::is_post_restricted()` passes the incoming value straight through when `Content_Gate::is_gating_active()` is false, which is the behaviour the pre-existing state needed and the `has_any_gates()` invariant could not give.

plugins/newspack-plugin/includes/wizards/audience/class-audience-wizard.php
dkoo: Resolved in `c8d9f841`. The 409 guard is gone, along with `tests/unit-tests/wizards/class-audience-management-disable-guard-test.php`, so there is no longer an `isset()` to slip an explicit null past. A confirmation dialog took its place, and it now names what becomes public rather than refusing the request.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
dkoo: Resolved in `c8d9f841` — `has_any_gates()` no longer exists, so the Memberships gap and the docblock that overstated it went with it. Worth keeping the finding on record: Memberships restriction does keep enforcing when Audience Management is off, because `Content_Restriction_Control::is_post_restricted()` returns early while Memberships is active. That asymmetry is now deliberate and commented rather than incidental.

plugins/newspack-plugin/includes/wizards/audience/class-audience-wizard.php
dkoo: Resolved in `c8d9f841`. The refusal is gone, so there is no longer a message routing a premium-newsletter publisher to a list that excludes them. The prerequisite screen takes an `isNewsletter` variant and names premium newsletters directly on that surface.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/index.js
dkoo: Resolved in `c8d9f841`. The guards are applied at module scope — `index.js:36-41` — so the component type is stable across renders and a section no longer remounts mid-edit. The test in `audience-management-prerequisite.test.js` now asserts that each renderer resolves to a module-scope `const … = requireAudienceManagement(…)` binding, so calling a guard inside a component body fails rather than going unnoticed.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/audience-management-prerequisite.test.js
dkoo: Resolved in `c8d9f841`. The pattern is now `/render:\s*([A-Za-z0-9_$]+)\s*[,\n}]/g`, which matches a renderer written last in its object, and both the route and renderer counts are anchored with `toHaveLength()` against a literal per router. An empty list can no longer pass, and adding a section without wiring it into the test fails loudly instead of shrinking the denominator.

plugins/newspack-plugin/src/wizards/types/window.d.ts
dkoo: Resolved in `c8d9f841` — both keys are optional now, and the comment explains the case: a page whose localized config predates the feature has neither, which `hasAudienceManagement()` already fails closed on and the prerequisite screen already falls back for. The note about the type being `string` rather than `boolean` is still there.

plugins/newspack-plugin/src/wizards/audience/views/content-gates/audience-management-required.tsx
dkoo: Resolved in `c8d9f841`. `/edit/:id`, `/settings/countdown-banner`, `/settings/content-gifting`, `/institutions`, `/institutions/new` and `/institutions/:id` all redirect to `/content-gates` through `redirectWithoutAudienceManagement()`, so the prerequisite state renders in exactly one place and never stacks under a section heading that implies the feature is configurable.

PR #767, opened 2026-07-29
feat(audience): date range segmentation for ESP date fields

miguelpeixe:
### All Submissions:

Phase 2 of NPPD-2047, following #628. That PR gave every incoming ESP field a publisher-chosen segmentation operator, but left `date` / `datetime` fields with **Text only** — matching a date meant typing it exactly, which is useless for the case that motivated the ticket: *"Latest Gift Date: January 1 2026 – Now"*.

A date-typed ESP field can now be segmented as a **Date range**.

**In Audience → Connections → _integration_**, a date field's "Segment as" dropdown offers **Date range** (the new default when you enable one) and **Text**. Fields already enabled as Text keep matching exactly as they do today until a publisher deliberately changes them.

**In Audience → Campaigns → Segments**, that criterion renders as a single dropdown:

- **Any** — no constraint on the field
- **Last 7 days** / **Last 30 days** / **Last 365 days** — rolling windows
- **Custom** — reveals two rows, **From** and **To**

Each custom row is independently **Any** (unbounded on that end), a fixed **Date**, or **Days ago** / **Days from now**. A preset is a shortcut for a range the custom rows could build by hand — "Last 30 days" is exactly `From: Days ago 30`, `To: Days ago 0` — so there is one stored format either way, and a range saved as a preset comes back as that preset rather than as Custom.

**Relative bounds resolve at match time**, so a rolling window stays current without anyone editing the segment each month — an absolute "since Jan 1" silently goes stale. Forward offsets cover expiry and renewal windows.

**Matching is day-granular and inclusive at both ends.** Reader values are normalized to ISO 8601 **at pull time**, using a source format each provider declares — `03/04/2026` is genuinely ambiguous between Mailchimp's two `date_format` settings, and the ESP's own metadata is only in hand at that point. The client matcher then validates a `YYYY-MM-DD` slice and compares lexicographically, reading the date part **as written** so a value the publisher sees as Jan 15 in ActiveCampaign matches on Jan 15 for every reader regardless of browser timezone.

**A value that cannot be parsed with confidence is stored untouched, never guessed at.** PHP's general parser only ever sees an already ISO-shaped value, and its result is only trusted when the date part survives a round trip — so `2026-02-30` stays verbatim rather than silently becoming `2026-03-02`, and a Mailchimp value that doesn't fit its declared format is not re-read American-first. Both matchers reject anything that isn't a real calendar date, so an unparsable value matches nothing instead of matching wrongly. When that happens the pull logs the field and the source format it used — otherwise the only symptom is a segment that quietly matches nobody.

**Both ESP providers declare their date metadata.** ActiveCampaign's `date` / `datetime` fields and Mailchimp's `date` merge fields default to the date-range operator, and Mailchimp translates its per-field `date_format`. ActiveCampaign declares `date_format` explicitly as `''` even though it always sends ISO: the key's *presence* is how a field snapshotted before source formats existed is told apart from one that genuinely has none, and the latter is refreshed from the live provider schema on read. Without that, a Mailchimp field enabled before this branch would have had no format to normalize with and matched nobody. **Mailchimp `birthday` is corrected from `date` to `string`** — it is `MM/DD` with no year, so it cannot sit on an absolute timeline and would have offered a range it could not honor.

**Content-gate access rules** have no range UI and still match a single date exactly, but they now compare *calendar dates*, with both sides put through the same normalizer. Selecting Date range for segmentation rewrites the stored value to ISO, so a gate rule written in the provider's own format (`03/04/2026`) would otherwise have silently stopped granting access to every reader it previously admitted — and the natural fix for that (switching the operator back, or disabling the field) unregisters the rule, which the content gate treats as "don't block", swinging the same gate wide open.

**`date_range` is only emitted to a newspack-popups that can run it.** newspack-popups now publishes the matching functions it supports, and newspack-plugin asks before registering the criterion, falling back to exact matching when the answer is no or the method is absent. An older build left an unresolvable name as a raw string and then called it, throwing a `TypeError` that escaped segment matching and aborted prompt display sitewide — and the fail-closed guard for that ships *in* newspack-popups, which is the plugin that would be stale, so it cannot be relied on for this skew.

**No newspack-manager companion PR this time.** An unset source format means ISO / `Y-m-d`, which is exactly what ActiveCampaign sends, so the managed AC integration needs no change.

#### Also fixed here

Latent problems this feature would otherwise have shipped into:

- **An unknown matching function no longer takes every prompt offline.** `newspack-popups` left an unresolvable matching function as a raw string and then called it; it now fails closed.
- **Normalization is gated on the chosen operator.** Rewriting stored values for a date field a publisher typed as *Text* would have broken their existing exact-match segments and content-gate rules within one pull cycle.
- **An emptied date range drops the criterion.** Clearing both bounds used to leave `value: {}`, which the matcher reads as "no bounds to check" — matching *every* reader who had any date in that field.
- **An out-of-range relative offset fails closed.** A large enough offset produced an invalid date whose string form compared as satisfied on the `To` end, widening the window to everyone.
- **The segment list renders date ranges** ("30 days ago to today") instead of `start: [object Object]`.
- The date-range control no longer flips a loaded "Days from now" bound to "Days ago" when its value is cleared and retyped, no longer resets a cleared date to today mid-edit, and its value input has an accessible name.

#### Out of scope and known limits

- **Date range is offered only for ESP-typed `date` / `datetime` fields.** A date living in a plain text field stays free text.
- **No recurring / year-agnostic matching** (what a Mailchimp `birthday` would need).
- **A source format that changes after a field is enabled is not detected.** The stored schema snapshot is refreshed only when the format is *absent*, so a publisher who switches a Mailchimp merge field from `MM/DD/YYYY` to `DD/MM/YYYY` gets confidently wrong dates until they re-save the Integrations screen. Closing this needs a periodic schema refresh, which is a broader change than belongs here.
- **The segment schema branch governs export/import and `register_meta`, not editor save.** The wizard's segmentation routes register no `args` schema and write criteria through `update_term_meta`, so a malformed criterion still saves silently. This gap is pre-existing and identical for the shipped `min` / `max` branch.

Closes NPPD-2047.

1. In **Audience → Connections → _integration_**, enable a `date` field. Confirm the operator dropdown offers **Date range** and **Text**, and that enabling it selects Date range.
2. Confirm a date field that was **already enabled as Text** still reads Text and is not silently rewritten.
3. In **Audience → Campaigns → Segments**, add that field as a criterion. Confirm a **single dropdown** offering Any / Last 7 days / Last 30 days / Last 365 days / Custom, defaulting to **Any** with no further inputs.
4. Choose **Last 30 days**, save, reload. Confirm it comes back as **Last 30 days**, not as Custom.
5. Switch to **Custom**. Confirm the two rows appear pre-filled with `Days ago 30` / `Days ago 0` — the preset's range, not a reset.
6. Set `From = Date` and pick a fixed date. Confirm the native date input is usable at its rendered width and lines up with its selector. Save and reload.
7. **Regression for the bound-direction bug:** on a saved segment holding a `Days from now` bound, clear the value and retype it. Confirm the selector still reads **Days from now** and the saved value is a *positive* offset.
8. Set a configured range back to **Any** and save. Confirm the criterion is dropped from the segment rather than saved as an empty value.
9. Confirm the segments list summarises the criterion in words ("30 days ago to today", "2026-01-01 to today").
10. Confirm matching: a reader whose ESP date falls inside the window matches, one outside does not, and a reader whose value is empty or unparsable does not.
11. **Regression on existing behavior:** a Number (`range`) criterion, a Multiple values (`list__in`) criterion, and a plain Text criterion all still render, save and match as before.
12. **Regression on content gates:** an access rule on a date field still matches by exact date, including when the rule is written in the provider's own format.

### Other information:


plugins/newspack-popups/includes/class-newspack-segments-model.php
```diff
@@ -265,6 +273,52 @@ public static function get_meta_schema() {
...
+					'additionalProperties' => false,
+					'required'             => [ 'type', 'date' ],
+					'properties'           => [
+						'type' => [
+							'type' => 'string',
+							'enum' => [ 'absolute' ],
+						],
+						'date' => [
+							'type'    => 'string',
+							'pattern' => '^\d{4}-\d\d-\d\d$',
```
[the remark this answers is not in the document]
miguelpeixe: Good catch — fixed in 0508c7ba6.

The client matcher's `ISO_DATE` was tightened to bound month and day, but the schema's pattern wasn't updated to match, so the two disagreed exactly as you describe. Both are now `^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])$`.

Added two data-provider cases: `2026-13-45` is rejected, and `2026-12-31` still validates (guarding against over-tightening). Also verified the pattern accepts leap days. `SchemasTest` is 27/27.

Worth noting for anyone reading later: this schema governs export/import and the `register_meta` REST route, not the segment editor's save path — that registers no `args` schema and writes through `update_term_meta`. So the practical effect is on round-tripped segments rather than editor input, but the two patterns should agree regardless.

plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
```diff
@@ -2799,10 +2801,17 @@ private static function map_merge_field_to_integration_schema( $field ) {
...
+		$date_format       = '';
 		if ( in_array( $type, [ 'dropdown', 'radio' ], true ) ) {
 			$value_type = 'select';
-		} elseif ( in_array( $type, [ 'date', 'birthday' ], true ) ) {
-			$value_type = 'date';
+		} elseif ( 'date' === $type ) {
+			// 'birthday' is deliberately excluded: it is MM/DD with no year, so it
+			// cannot be placed on an absolute timeline and stays exact-match text.
+			$value_type        = 'date';
+			$matching_function = 'date_range';
```
adekbadek: **Blocker:** date_range is emitted with no guard on the consumer's version — The capability probe in `Promoted_Fields::supported_matching_function()` covers newspack-plugin → newspack-popups, but nothing covers newspack-newsletters → newspack-plugin, and this mapper (with `class-newspack-newsletters-active-campaign.php:2537`) is where the operator originates. On a site running this newspack-newsletters with newspack-plugin still on `main`, `date_range` reaches segment registration unprobed: `ESP::configure_incoming_field()` assigns `matching_function` straight from the provider schema with no allow-list, and the stored-operator re-apply in `get_enabled_incoming_fields()` only *overrides* values that are already in `ALLOWED_INCOMING_MATCHING_FUNCTIONS` — it never resets one that isn't. If newspack-popups is also on `main`, `registerCriteria()` leaves the unresolved name as a string and `criteria.matches()` invokes it.

Verified against `origin/main`'s own `src/criteria/utils.js`, driving it with `matchingFunction: 'date_range'`:

```
TypeError: criteria.matchingFunction is not a function
```

`match()` in `src/view/utils/segments.js:73` has no try/catch, so that propagates out of `getBestPrioritySegment()` and prompt display stops for the pageview — the same failure the description cites as the reason for the popups-side fail-closed guard. That guard ships *in* newspack-popups, which is the plugin that would be stale, so it cannot cover this direction.

**To reproduce**
1. Install this branch's newspack-newsletters alongside newspack-plugin and newspack-popups from `main`.
2. In Audience → Connections → Newsletter ESP, enable a Mailchimp or ActiveCampaign `date` field.
3. Add that field as a criterion on any enabled segment.
4. Load any front-end page. The console shows the TypeError and no prompts render.

The guard belongs on the emitting side, since that is the only side present in every combination — e.g. assign `date_range` only when `method_exists( '\Newspack\Reader_Activation\Integrations\Incoming_Field', 'set_date_format' )` (the method this PR adds), falling back to `'default'`. If a version check pointing from newsletters at newspack-plugin is the wrong dependency direction, the alternative is a stated release order — but nothing in the code expresses one, and plugins update independently.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -884,16 +922,39 @@ public function get_enabled_incoming_fields() {
...
+				}
+			} elseif ( self::needs_source_date_format( $raw_data ) && $live_schema_resolved ) {
+				// Fill in just the source format — the publisher's stored operator and
+				// the rest of the snapshot stay authoritative. Absent from the live
+				// schema (an integration that never declares formats, or a field that
+				// left the provider) means ISO. Persisted below so this is a one-time
+				// repair, not a live resolution on every read; an API failure persists
+				// nothing, so the next read retries.
+				$raw_data['date_format']       = isset( $live_by_key[ $key ]['date_format'] ) ? $live_by_key[ $key ]['date_format'] : '';
+				$stored[ $key ]['date_format'] = $raw_data['date_format'];
```
adekbadek: **Blocker:** The date_format repair latches '' permanently when the live schema declares nothing — This writes `''` whenever the live entry has no `date_format` key, and because `needs_source_date_format()` keys on `array_key_exists()`, the entry is never re-examined afterwards. `''` means "the provider already sends ISO", so an absent format and a declared-ISO format become indistinguishable the moment they are persisted.

That matters because the two versions that produce the triggering state are different plugins. A stored entry can only carry `matching_function => 'date_range'` if newspack-plugin is on this branch, but `date_format` is only emitted if newspack-newsletters is too. With newspack-plugin updated and newspack-newsletters not, a Mailchimp date field enabled as Date range hits this branch, the live schema returns the field with no `date_format`, and `''` is written. Values keep arriving as `MM/DD/YYYY`, `is_iso_shaped_date()` rejects them, and the segment matches nobody. Updating newspack-newsletters later does not repair it — the key now exists, so the repair never runs again.

A second path reaches the same write without any skew in the field itself: `Newspack_Newsletters_Mailchimp_Cached_Data::get_data()` returns `[]`, not a `WP_Error`, when its cache option is absent (`class-newspack-newsletters-mailchimp-cached-data.php:418`). `get_contact_fields_for_integrations()` then returns `[]`, so `$live_schema_resolved` is `true` against an empty `$live_by_key` and every pending entry is stamped `''` in one pass.

Recovery is a re-save of the Integrations screen, and the only signal is the "could not be normalized" line from `pull_single_integration()` — which the code comment at :481 already identifies as the sole symptom.

One option that keeps genuinely format-less integrations working: resolve to `''` in memory as now, but set `$stored_dirty` only when the live entry actually declares the key — `isset( $live_by_key[ $key ] ) && array_key_exists( 'date_format', $live_by_key[ $key ] )`. The unresolved case then re-resolves on each read, which is what legacy-shaped entries already do, and a later newsletters update can still land the real format. `README.md:345` documents the current fallback and would need the same change.

plugins/newspack-plugin/includes/reader-activation/class-promoted-fields.php
```diff
@@ -242,11 +279,48 @@ private static function evaluate_field( $field, $user_id, $args ) {
...
 				$user_values = self::parse_list_value( $value );
 				return empty( array_intersect( (array) $args, $user_values ) );
+			case 'date_range':
+				// Access rules have no range UI — a rule still holds one typed value and
+				// still matches it exactly. But selecting this operator makes the pull
+				// rewrite stored values to ISO, so a rule the publisher wrote in the
+				// provider's own format ('03/04/2026') would silently stop matching a
+				// stored '2026-03-04'. Put both sides through the same normalizer and
+				// compare calendar dates, so the rule keeps meaning what it meant.
+				$stored_date = self::to_calendar_date( $value );
```
adekbadek: **Suggestion:** The access rule normalizes the rule but not the stored value — `$rule_date` goes through `normalize_date_value()`; `$stored_date` is passed straight to `to_calendar_date()`. Normalization happens only at pull time, so between a publisher selecting Date range and a given reader's next pull, that reader's value is still in the provider's format — `to_calendar_date( '03/04/2026' )` returns `null` and the rule denies them.

The direction is safe: readers are denied rather than admitted, and a pull repairs it. But a reader is only re-pulled on their own activity — `Contact_Cron::maybe_enqueue_contact()`, which treats a contact as stale after 24h — or by a manual backfill. So the rule refills over roughly a day for active readers and never refills for churned ones, and the description's guarantee that "a gate rule written in the provider's own format would otherwise have silently stopped granting access" only holds for the rule side of the comparison.

Running `$value` through the same normalizer with the field's `date_format` would make the two sides symmetric. Failing that, the reverse case is worth a line in the known limits: choosing Date range is effectively one-way for stored values until a full re-pull, since moving the operator back to Text restores the provider format only for readers who are pulled again.

plugins/newspack-plugin/includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -266,6 +266,115 @@ public static function handle_ajax_pull() {
...
+	 * @param string $format     Source format as a PHP date format string. Empty means
+	 *                           the provider already sends ISO 8601 / Y-m-d. A format
+	 *                           that parses no year (`m/d`) is treated as undeclared:
+	 *                           it would parse cleanly into the Unix epoch year, and a
+	 *                           confident `1970-03-04` is exactly the wrong answer the
+	 *                           paths below exist to avoid.
+	 * @param string $value_type Either 'date' or 'datetime'.
+	 * @return mixed ISO string, or the value unchanged.
+	 */
+	public static function normalize_date_value( $value, $format = '', $value_type = 'date' ) {
```
adekbadek: **Suggestion:** A date utility on the pull job, and a fourth copy of the ISO pattern — `normalize_date_value()` takes a value, a format and a type and returns a string — nothing about it concerns pulling. `Promoted_Fields::evaluate_field()` reaches into `Contact_Pull` to call it, so the dependency runs from the access-rule evaluator into the ESP pull job.

The bounded ISO pattern now exists in four places: `ISO_DATE` in `matching-functions.js:25`, `Promoted_Fields::to_calendar_date()` at :314, `Contact_Pull::is_iso_shaped_date()` at :375, and the schema pattern in `class-newspack-segments-model.php:299`. The PHP/JS split across two plugins makes some duplication unavoidable, and each comment correctly names the others as mirrors — but the two PHP copies sit in the same plugin doing overlapping jobs, and a divergence surfaces as segments that validate and then match nobody. The whitespace difference already flagged on `class-promoted-fields.php:314` is what that looks like in practice.

Moving the three helpers to a neutral home — a static on `Incoming_Field`, or a small date-value class beside it — would collapse the two PHP copies and put the call the right way round. For the remaining cross-language pair, `test_supported_matching_functions_exist_in_js_module()` at `test-criteria.php:151` is the precedent: the popups suite could extract the `ISO_DATE` literal from `matching-functions.js` and assert it equals the schema's `pattern`.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
```diff
@@ -908,9 +969,32 @@ public function get_enabled_incoming_fields() {
...
+				}
 				$fields[] = $field;
 			}
 		}
+
+		if ( $stored_dirty ) {
+			// Self-heal: persist the resolved source formats so the live resolution
+			// doesn't repeat on every read. update_option() no-ops on an unchanged
+			// value, so concurrent requests racing here are harmless.
+			\update_option( self::INCOMING_FIELDS_OPTION_PREFIX . $this->id, $stored, false );
```
adekbadek: **Suggestion:** A getter on the front-end path writes an option it read before a network call — `Promoted_Fields::init()` hooks `register()` on `init` priority 15 unconditionally, and `get_promoted_fields()` calls this method for every active integration — so both the live-schema fetch and this `update_option()` can run on an anonymous front-end pageview.

Two consequences worth separating. `$stored` is read at :892, the ESP call happens in between, and the whole array is written back here; a publisher saving the Integrations screen inside that window has their write reverted, because this request persists the pre-save copy plus `date_format`. Two racing *repairs* are harmless — `update_option()` no-ops on an identical value, as the comment says — but a repair racing a *save* is not, and enabling Date range is precisely what creates a pending repair moments before the publisher's next save.

Separately, when the fetch fails, neither ActiveCampaign's `get_contact_fields_for_integrations()` nor Mailchimp's cached-data layer caches the failure, so a persistently failing ESP means one blocking outbound request per uncached front-end request.

The narrow fix is to re-read the option immediately before writing and set only the `date_format` keys on that fresh copy. The broader one is to move the resolve-and-persist off the read path entirely — `Contact_Pull::pull_single_integration()` or the settings-save path would repair the same entries — and let the front end read the stored snapshot.

The fetch-on-read shape predates this PR: the legacy-overlay branch does the same, and never persists, so it re-fetches indefinitely. What is new is a second, much more common trigger, plus the write.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-promoted-fields.php
```diff
@@ -415,4 +415,108 @@ public function test_list_in_handles_single_selection_pipe_wrapped_value() {
...
+	/**
+	 * The version-skew guard is what stops an old newspack-popups from being
+	 * handed a matching function it can't resolve — a TypeError that escapes
+	 * segment matching and takes prompt display down sitewide. newspack-popups
+	 * is not loaded in this suite, which is exactly the probe-less environment
+	 * an old build presents (the probe method can't exist), so this pins the
+	 * degradation path; the probe-present path is covered from the popups side
+	 * by its own CriteriaTest, since CI runs each plugin's suite in isolation.
+	 */
+	public function test_supported_matching_function_degrades_without_the_popups_probe() {
```
adekbadek, line 504: **Suggestion:** The probe's success branch is untested on both sides — `test_supported_matching_function_degrades_without_the_popups_probe()` skips when `\Newspack_Popups_Criteria` exists, and newspack-plugin's suite never loads it — so only the degrade-to-`default` path is covered. The docblock points at popups' `CriteriaTest`, but that asserts `supports_matching_function()` in isolation and never exercises `supported_matching_function()` consuming it.

So inverting the ternary at `class-promoted-fields.php:224`, or mistyping the class or method name inside the `method_exists()` call, leaves both suites green while every site silently falls back to exact matching on date fields — a segment that matches nobody, which reads in the admin exactly like no reader qualifying.

`tests/mocks/newspack-popups-model-mock.php` is the existing precedent for stubbing a popups class here. A stub declaring `supports_matching_function()` under `@runInSeparateProcess`, or a seam that lets the probe result be injected, would let one test assert that `date_range` survives. The reason for the skip is legitimate — CI runs each plugin's suite alone — which is why the seam is probably the cheaper of the two.

Worth noting that `test_baseline_matching_functions_is_frozen` at :514 adds no signal the test above it doesn't already carry: adding `date_range` to the constant breaks the `assertSame( 'default', … )` assertion either way.

plugins/newspack-popups/src/criteria/matching-functions.js
```diff
@@ -19,6 +19,88 @@ const parseReaderListValue = value => {
...
+ * date, so skipping validation produces confident wrong matches rather than none.
+ *
+ * @param {*} value The stored reader value.
+ * @return {?string} A `YYYY-MM-DD` string, or null.
+ */
+const toCalendarDate = value => {
+	if ( typeof value !== 'string' ) {
+		return null;
+	}
+	const candidate = value.slice( 0, 10 );
```
adekbadek: **Suggestion:** A value with an ISO prefix and trailing junk matches as that date — `toCalendarDate()` validates `value.slice( 0, 10 )` and ignores whatever follows, so a stored value the normalizer deliberately refused to trust is matched anyway. `is_iso_shaped_date()` (`class-contact-pull.php:375`) only accepts a suffix beginning `T` or whitespace, so `2026-01-15abc` is exactly the shape `normalize_date_value()` returns untouched.

Measured on this branch:

```
date_range( '2026-01-15abc', Jan 1–Dec 31 2026 ) -> true
date_range( '2026-01-15/bogus', Jan 1–Dec 31 2026 ) -> true
```

The PHP mirror agrees, via `substr( trim( $value ), 0, 10 )`:

```
Promoted_Fields::to_calendar_date( '2026-01-15abc' ) => '2026-01-15'
Promoted_Fields::to_calendar_date( '2026-01-15/bogus' ) => '2026-01-15'
```

This is the one input that fails open, against the contract the rest of the feature keeps: the description's "an unparsable value matches nothing instead of matching wrongly" held on every other probe — ambiguous slash dates, `2026-02-30`, year-less formats and out-of-range offsets all decline to match. It also reaches content gating rather than only prompt targeting, since `evaluate_field()`'s `date_range` branch runs the stored value through the same `to_calendar_date()`: a garbage value whose first ten characters equal the rule's date grants access.

Validating the whole string — a bare `YYYY-MM-DD`, or a date followed by a recognised time component — closes it on both sides. Worth keeping the two implementations aligned while changing them, since they already disagree on one input (see the whitespace note on `class-promoted-fields.php:314`).

Reachability is the limit on this one. The value has to be a stored ESP value whose first ten characters are a valid ISO date but which is not ISO-shaped overall, since anything ISO-shaped gets parsed and round-trip-checked instead. ActiveCampaign sends ISO, and Mailchimp validates its own date merge fields, so the realistic source is free text reaching a date-typed field — `2026-01-15 TBD` and similar. For the gate-rule case the reader's value also has to carry the rule's exact date as its prefix. So: a real fail-open, but a narrow one, and the harm ceiling is one reader wrongly included or admitted rather than anything sitewide.

plugins/newspack-popups/src/criteria/matching-functions.js
```diff
@@ -98,4 +180,37 @@ export default {
...
+		if ( ! value ) {
+			return false;
+		}
+		const bounds = config.value;
+		if ( ! bounds || typeof bounds !== 'object' || Array.isArray( bounds ) ) {
+			return false;
+		}
+		// A bound that is present but unusable fails closed. Silently dropping it
+		// would widen the segment to readers the publisher never asked for.
+		if ( bounds.start ) {
```
adekbadek: **Suggestion:** A broken date bound sometimes widens the range instead of failing closed — A range has two ends, and the code decides whether an end was set by asking whether `bounds.start` is truthy. In JavaScript `null`, `0` and `''` are falsy, so a bound stored as any of those reads as "no start was set" and the comparison is skipped — that end of the range becomes unlimited. An empty object is truthy, so that one does reach the comparison, fails to resolve to a date, and returns false.

So the same underlying problem — a bound that is present but unusable — lands on opposite outcomes depending on the value's type:

```
date_range( '2026-05-05', { start: {} } ) -> false // matches nobody
date_range( '2026-05-05', { start: { bogus: 1 } } ) -> false // matches nobody
date_range( '2026-05-05', { start: null } ) -> true // matches everybody
date_range( '2026-05-05', { start: 0 } ) -> true // matches everybody
date_range( '2026-05-05', { start: '' } ) -> true // matches everybody
```

Concretely: a segment meant as "gifts since 2030" whose start bound got written as `null` matches every reader who has any date in that field, rather than none. The comment two lines above says a present-but-unusable bound must fail closed, which is what the first two rows do and the last three don't.

The wizard cannot produce these — `changeRange()` in `DateRangeSetting.js:213` emits `undefined` for an empty range, and says why. They arrive through the paths `get_date_bound_schema()`'s parent governs: it requires neither key, so `{}` and `{ "start": null }` both validate on import and on the `register_meta` route. The editor's own save path registers no `args` schema either, as the description notes.

Checking presence rather than truthiness — `if ( undefined !== bounds.start )` — sends all five rows into `resolveDateBound()`, which already returns null for anything unusable, so they would fail closed on one path.

One judgement call attached: this diverges from the shipped `range` matcher, where an absent min and max mean unbounded, and it contradicts the `{ value: {} }` case in `matching-functions.test.js:47`. Leaving it and naming the exception in the comment would also be defensible.

plugins/newspack-plugin/includes/reader-activation/class-promoted-fields.php
```diff
@@ -242,11 +279,48 @@ private static function evaluate_field( $field, $user_id, $args ) {
...
+	 * written rather than shifting it into a timezone.
+	 *
+	 * @param mixed $value Candidate value.
+	 * @return string|null A `YYYY-MM-DD` string, or null.
+	 */
+	private static function to_calendar_date( $value ) {
+		if ( ! is_string( $value ) ) {
+			return null;
+		}
+		$candidate = substr( trim( $value ), 0, 10 );
```
adekbadek: **Nit:** The two mirrors disagree on whitespace and on year 0000 — The docblock says this mirrors `toCalendarDate()`. Two inputs where they part, both measured:

```
' 2026-01-15' PHP to_calendar_date => '2026-01-15' JS date_range -> false
'0000-01-15' PHP checkdate(1,15,0) => false JS date_range -> true
```

The first is the trim: this trims before slicing and `toCalendarDate()` does not, so a value with leading whitespace is a date to the access-rule evaluator and not to the segment matcher. The second is the year: `checkdate()` rejects year 0 while `isCalendarDate()` accepts it, since `setFullYear( 0, … )` is valid — and the comment at `matching-functions.js:40` chose `setFullYear()` deliberately to keep low years working. `normalize_date_value()` stores `0000-01-15` untouched because its round trip survives, so that value is segment-matchable and gate-unmatchable.

Neither is reachable from real ActiveCampaign or Mailchimp data. They matter because the two implementations are load-bearing mirrors of each other, and this is what the drift the comments warn about looks like in practice. The trim is worth keeping on the rule argument, so aligning means a `.trim()` on the JS side and one shared decision about year 0.

plugins/newspack-plugin/includes/reader-activation/integrations/class-contact-pull.php
```diff
@@ -266,6 +266,115 @@ public static function handle_ajax_pull() {
...
+	 * tell from a real date. The in-tree provider mappers never emit such a
+	 * format (Mailchimp's year-less `birthday` is deliberately not a date type),
+	 * but any third-party `set_date_format()` caller can. Backslash-escaped
+	 * literals (`\Y`) don't count as specifiers.
+	 *
+	 * @param string $format PHP date format string.
+	 * @return bool
+	 */
+	private static function format_specifies_year( $format ) {
+		return 1 === preg_match( '/(?<!\\\\)[YyoXx]/', $format );
```
adekbadek: **Nit:** format_specifies_year() misses formats that fix a year without a year specifier — `U` (Unix timestamp) and `r` (RFC 2822) both place a date on the timeline but contain none of `[YyoXx]`, so a third-party `set_date_format( 'U' )` is treated as undeclared. Neither shape is ISO-shaped either, so the field never normalizes at all.

Fails closed, so nothing is corrupted — but the docblock explicitly invites third-party callers, and both are plausible declarations. Adding them to the character class, or short-circuiting on a whole-format match for `U` / `r` / `c`, would cover it.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.test.js
```diff
@@ -408,15 +408,44 @@ describe( 'incoming-field operators', () => {
...
+		expect( operatorOptionsForField( { value_type: 'date' } ).map( o => o.value ) ).toEqual( [ 'date_range', 'default' ] );
+		expect( operatorOptionsForField( { value_type: 'datetime' } ).map( o => o.value ) ).toEqual( [ 'date_range', 'default' ] );
 		expect( operatorOptionsForField( { value_type: 'boolean' } ).map( o => o.value ) ).toEqual( [ 'default' ] );
 		expect( operatorOptionsForField( { value_type: 'multiselect' } ).map( o => o.value ) ).toEqual( [ 'list__in' ] );
 		expect( operatorOptionsForField( { value_type: 'select' } ).map( o => o.value ) ).toEqual( [ 'default', 'list__in' ] );
 		expect( operatorOptionsForField( { value_type: 'string', has_options: false } ).map( o => o.value ) ).toEqual( [ 'default', 'range' ] );
 		expect( operatorOptionsForField( { value_type: 'string', has_options: true } ).map( o => o.value ) ).toEqual( [ 'default', 'list__in' ] );
 	} );
 
+	it( 'offers a date range and text for date fields', () => {
```
adekbadek: **Nit:** Several new tests restate the case beside them — The new suites are unusually good — the mutations worth trying (dropping `isCalendarDate`, dropping the ISO gate, removing the NaN guard, losing the ambiguous-zero direction) each have a test that catches them. This is only about the overlaps left behind:

- Here, `constrains operator options by value_type` at :409 already asserts `[ 'date_range', 'default' ]` for `date` and `datetime`; this case adds only the labels.
- `test-criteria.php:136` — the first loop iterates a hardcoded copy of `SUPPORTED_MATCHING_FUNCTIONS` and asserts each returns true, which asserts that `in_array()` works. The `date_range` case and the two `assertFalse` calls are the parts that bite.
- `class-test-integrations.php:2216` — lines 2216 and 2217-2220 are character-for-character the assertions at 2181 and 2188-2191. Only line 2223 (`'03/04/2026'` returned verbatim) states what the method name promises.
- `single-segment.test.js` — the four `describe` blocks at :390, :466, :525 and :566 repeat the same `makeExistingSegment` / `wizardApiFetch` / `mockProps` / `beforeEach` scaffold, differing only in the criteria fixture. A helper taking that fixture would cut roughly 60 lines and make each block's difference visible at a glance.

plugins/newspack-popups/src/criteria/matching-functions.js
```diff
@@ -19,6 +19,88 @@ const parseReaderListValue = value => {
...
+		// as satisfied, widening the window to everyone — the opposite of the
+		// fail-closed contract below. Return null so the caller rejects it.
+		if ( Number.isNaN( date.getTime() ) ) {
+			return null;
+		}
+		// Built from local components on purpose: toISOString() converts to UTC and
+		// would land on the wrong day for anyone west of Greenwich.
+		const month = String( date.getMonth() + 1 ).padStart( 2, '0' );
+		const day = String( date.getDate() ).padStart( 2, '0' );
+		return `${ date.getFullYear() }-${ month }-${ day }`;
```
adekbadek: **Nit:** The relative-bound year is not zero-padded — `month` and `day` are padded, `date.getFullYear()` is not. A `days: -400000` start bound lands in year 931 and formats as `931-xx-xx`, and `'2026-06-15' < '931-01-01'` lexicographically — so the bound excludes the dates it should include. Only reachable at absurd magnitudes, but the number input has no `max` so the value is typeable. `String( date.getFullYear() ).padStart( 4, '0' )` matches what the rest of the line already does.

A `max` on the days input (anything comfortably inside the `Date` range) would also turn a nine-digit offset into browser-level validation rather than a segment that matches nobody. The fail-closed guard should stay regardless, since import and REST bypass the input.

plugins/newspack-popups/src/criteria/utils.js
```diff
@@ -49,9 +49,17 @@ export function registerCriteria( id, config = {} ) {
 			criteria.matchingFunction = matchingFunctions[ criteria.matchingFunction ].bind( null, criteria );
 		}
 
-		// Bail if unable to configure matching function.
+		// Bail if unable to configure matching function. Fall back to a function that
```
adekbadek: **Nit:** The comment is mostly the change's history — The constraint worth keeping is that `criteria.matches()` invokes `matchingFunction` unconditionally, so an unresolved name has to become a function rather than stay a string. The rest — that this is reachable in production, which plugin ships which operator first, what the resulting TypeError takes down — narrates how the fix came about, and names a version pairing that will read as stale within a release or two. Two lines carry the invariant.

The same applies elsewhere in the diff, less sharply:

- `normalize_date_value()`'s `@param $format` block (`class-contact-pull.php:294`) argues the year-less-format rule at length, and `format_specifies_year()`'s own docblock at :348 makes the same argument again. One of them earns it.
- That docblock's closing sentences ("Both produce a confident wrong answer, which is strictly worse than…") defend the decision and restate the two guards immediately below them.
- The birthday-exclusion comment at `class-newspack-newsletters-mailchimp.php:2808` repeats its own docblock at :2772 almost word for word — "MM/DD with no year … cannot be placed on an absolute timeline … stays exact-match text". The docblock already carries the why.
- "a one-time repair, not a live resolution on every read" appears three times in `class-integration.php` — at :864, :953 and :992. The docblock is the right home; the two inline copies can keep only what they add (the fallback-to-ISO rule, and the `update_option()` no-op note).

adekbadek (changes requested):
Release risk: HIGH — a confirmed crash path on a mixed-version install, plus a permanent silent mis-normalization that a later plugin update does not repair.

Requesting changes on two blockers. Each was reproduced against a running env or against `origin/main`'s own source rather than reasoned from the diff; the repro steps are on the individual threads.

The design itself holds up well. Every fail-closed claim in the description was probed and behaved as documented — ambiguous slash dates, impossible calendar days, year-less formats, out-of-range offsets and legacy un-normalized values all decline to match rather than guessing. The blockers are all at the seams: two version-skew paths the probe does not cover, and one input shape that slips past both matchers.

- `plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.js:501` — **Suggestion:** The operator shown and the operator saved disagree after a repair — The rendered fallback is `operatorOptions[ 0 ]?.value`, which for a date field is now `date_range`. The save-time fallback added at :127 is `default`. For a field whose stored operator is not among those offered — a `list__in` or `range` value left over from before the field declared its current `value_type` — the row displays **Date range** while the next save writes **Text**. Two effects follow. Any unrelated save on that screen silently changes the stored operator to Text, contradicting what the publisher was shown. And a publisher who wants Date range sees it already selected, so choosing it fires no `onChange`; they have to select Text and switch back. The save-side reasoning is sound and worth keeping — a repair folded into an unrelated save should not opt anyone into an operator that rewrites stored reader values. Using the same expression for `selectedOperator` would make the control show what it is going to save.

plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php
miguelpeixe: Addressed in 497dac8. The guard is on the emitting side, in the provider base class: `integrations_supports_date_range()` probes `method_exists( '\Newspack\Reader_Activation\Integrations\Incoming_Field', 'set_date_format' )` and both mappers fall back to `default` when it fails, so a date field degrades to exact matching instead of shipping an operator the consumer can't validate.

For coverage in the standalone suite: a bootstrap stub declares the real class and method names (a mistyped probe now fails the mapping tests), and the probe carries a filter that pins the degraded branch in both schema suites.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Addressed in 4c93d2f, in the shape you outlined: the entry resolves to `''` in memory, and only a live entry that actually declares `date_format` is queued for persistence — so the undeclared case re-resolves on each read and a later newsletters update still lands the real format. The empty-cache path (`get_data()` returning `[]`) now persists nothing, since no entry declares the key.

The two tests that pinned the latching behavior are reworked — one now covers the recovery, with a declaration arriving on a later read being persisted and stopping the re-resolution — and the README paragraph matches. The cost lands on integrations that declare date fields but never the key, which keep re-fetching the live schema; the README already requires declaring `''` explicitly, and the newspack-manager follow-up will do that.

plugins/newspack-plugin/includes/reader-activation/class-promoted-fields.php
miguelpeixe: Addressed in 4c93d2f — the stored value now runs through the same normalizer as the rule (`Date_Value::normalize()` then `to_calendar_date()` on both sides), so a reader whose value hasn't been re-pulled yet matches instead of being denied until their next pull. The test covers the window: with the stored value still `01/15/2026`, the rule matches whether written as `01/15/2026` or `2026-01-15`.

plugins/newspack-plugin/includes/reader-activation/integrations/class-contact-pull.php
miguelpeixe: Addressed in 4c93d2f. The three helpers plus `to_calendar_date()` moved to a new `Date_Value` class beside `Incoming_Field`, so both the evaluator and the pull now depend on a neutral home. The two PHP pattern copies collapsed onto one `CALENDAR_DATE_PATTERN` constant.

For the cross-language pair, the popups suite now extracts the `ISO_DATE` literal from matching-functions.js and asserts it equals the segment schema's pattern (`test_client_iso_date_pattern_matches_the_segment_schema`), per the precedent you pointed at.

plugins/newspack-plugin/includes/reader-activation/integrations/class-integration.php
miguelpeixe: Addressed in 4c93d2f with the narrow fix: after the fetch, the option is re-read (cache-invalidated first, so a non-persistent object cache doesn't serve the pre-fetch copy) and only the resolved formats are written, only onto entries that still need them per the fresh copy — a save that changed the operator or already carried a format wins. A test simulates a publisher save landing during the fetch and asserts it survives the repair.

The fetch-on-read shape itself is unchanged; moving the resolve off the read path is a bigger restructuring than this PR should carry.

plugins/newspack-plugin/tests/unit-tests/integrations/class-test-promoted-fields.php
miguelpeixe: Addressed in 4c93d2f via the stub route — `@runInSeparateProcess` re-runs the WP bootstrap in the child process, which reinstalls the shared test DB under the parent run, so a stub was the safer of the two options. `tests/mocks/class-newspack-popups-criteria-mock.php` declares the real class and method names; the new test loads it and asserts `date_range` survives (and that an unknown operator still degrades). It runs after the degrade test by declaration order, which both docblocks note. Both failure modes you named now fail a suite: an inverted ternary fails the new test, and a mistyped `method_exists()` argument makes the probe miss the stub.

Kept `test_baseline_matching_functions_is_frozen`: in a combined local run (popups loaded) the degrade test skips itself, and the frozen-constant test is what still pins the baseline there.

plugins/newspack-popups/src/criteria/matching-functions.js
miguelpeixe: Addressed in f15ff60 (matcher) and 4c93d2f (gate) — the whole string is validated on both sides: a bare date, or a date followed by a time component matching a shared suffix pattern (`ISO_TIME_SUFFIX` / `TIME_SUFFIX_PATTERN`, documented as mirrors). The measured cases now fail closed on both sides, with tests on each, and real times in either separator style still match.

One knock-on: the pull's "could not be normalized" log now keys on `to_calendar_date()` instead of ISO shape, so a `2026-01-15 TBD` that survives verbatim gets logged too.

plugins/newspack-popups/src/criteria/matching-functions.js
miguelpeixe: Addressed in f15ff60 — presence check (`undefined !== bounds.start`), so all five rows land in `resolveDateBound()` and fail closed; tests cover null, 0, '' and {}. On the judgement call, took the strict side, with the comment naming the deliberate divergence from `range`: a numeric bound has no object shape to get wrong, so nothing there distinguishes a broken bound from a cleared one, while here the distinction is real. `{ value: {} }` keeps meaning unbounded — no key present.

plugins/newspack-plugin/includes/reader-activation/class-promoted-fields.php
miguelpeixe: Addressed in f15ff60 and 4c93d2f. The JS side now trims, matching the evaluator; year zero is rejected on both sides, aligning on `checkdate()` — `0000-01-15` never arrives from a real ESP, and rejecting it keeps the mirrors agreeing on every input. Both pinned by tests on each side.

plugins/newspack-plugin/includes/reader-activation/integrations/class-contact-pull.php
miguelpeixe: Partially as suggested, in 4c93d2f: `U` is added to the specifier class and normalizes (test included). `r` and `c` measured differently — `DateTimeImmutable::createFromFormat()` cannot parse either (they are format()-only specifiers, so it returns false) — meaning a declaration of either lands on the fail-closed path regardless of the year check. Rather than claim support the parser can't deliver, the docblock now says exactly that, and a test pins an `r` declaration surviving verbatim.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.test.js
miguelpeixe: Addressed across 4c93d2f, f15ff60, 84d4da8 and f662e56: the labels-only case folded into `constrains operator options by value_type` (date carries the full shape once); the `SUPPORTED_MATCHING_FUNCTIONS` loop trimmed to the `date_range` probe plus the two negatives; the duplicated assertions in `test_normalize_date_value_without_format_requires_iso_shape` cut down to the verbatim-return line that names the test; and the existing-segment `describe` blocks in single-segment.test.js now share a `mountExistingSegment()` scaffold, so each block reads as its criteria fixture plus its assertions.

plugins/newspack-popups/src/criteria/matching-functions.js
miguelpeixe: Addressed in f15ff60 and 84d4da8. The year is zero-padded and the resolved string is re-validated against `ISO_DATE`, so an offset that leaves four-digit years (negative or five-digit) fails closed instead of sorting wrong; tests cover the padded year-931 bound sorting correctly and both out-of-range directions. The days input now carries `max={ 100000 }` (~273 years), so the typo becomes browser-level validation while the matcher guard stays for import and REST.

plugins/newspack-popups/src/criteria/utils.js
miguelpeixe: Trimmed in 4c93d2f, f15ff60 and 497dac8: the utils.js comment is down to the invariant; `normalize()`'s `@param $format` defers to `format_specifies_year()` and the closing defense sentences are gone; the Mailchimp birthday inline comment is removed (the docblock carries it); and "one-time repair" now lives once, in `needs_source_date_format()`, with the other two sites keeping only what they add.

miguelpeixe:
Addressed the full review — both blockers, the six suggestions, the five nits, and the note on unchanged lines — in 497dac8 through 9ec166c. Each thread carries its own reply; the one point that couldn't live on a thread:

> **The operator shown and the operator saved disagree after a repair** (configure-view.js:501)

Fixed in f662e56. The save-time fallback is extracted into a shared `fallbackOperator()` and the rendered select now uses the same expression, so a stored out-of-set operator on a date field displays **Text** — exactly what the next save stores — and **Date range** stays selectable. Test added alongside the reconcile tests.

Verification, all in this worktree: `Test_Integrations` (97), `Test_Promoted_Fields` (17), `Test_ESP` (38), popups `CriteriaTest` (6) and `SchemasTest` (27), newsletters `MailchimpIntegrationsSchemaTest` (10) and `ActiveCampaignIntegrationsSchemaTest` (9), the full newspack-popups Jest suite (245) and the full newspack-plugin Jest suite (644) — all green. PHPCS (root ruleset) and ESLint clean on the changed files.

One deliberate divergence from a suggestion, argued on its thread: `r` and `c` are not added to `format_specifies_year()` — `createFromFormat()` cannot parse either, so declaring them stays fail-closed regardless; only `U` gains real support.

adekbadek (approved):
Both blockers are closed. The newsletters mappers gate `date_range` behind `integrations_supports_date_range()`, and the format repair persists only a live-declared `date_format`, leaving an undeclared one to re-resolve so a later newsletters update still lands it; the write-back re-reads the option, so a concurrent save survives.

In `matching-functions.js` and `Date_Value`, `2026-01-15abc`, `2026-01-15/bogus`, `2026-01-15 TBD`, and bounds stored as `null`, `0` or `''` now fail closed, the two mirrors agree on leading whitespace and year zero, and a -400000-day bound resolves to a padded `0931-…` that sorts correctly. The Integrations row renders the same `fallbackOperator()` the save-time repair writes.

PR #768, opened 2026-07-29
fix(checkout-button): apply attached coupon through the variation picker

miguelpeixe:
### All Submissions:

Two tester-reported problems with the Checkout Button block's auto-applied coupon, both staged in the current alpha, plus the follow-up work they exposed. Targets `alpha` rather than `main` for that reason.

**Heads-up for reviewers:** this grew past its original scope. It now also touches **newspack-plugin** (one file, noted at the end).

**1. The coupon was dropped whenever the variation picker was enabled.**

With "Allow the reader to select the variation before checkout" on (the default), the block's own form is never submitted — `modal.js` calls `preventDefault()` and opens the variation modal, whose form is rendered separately by `Subscriptions_Tiers::render_form()` in the footer and knows nothing about the block that opened it.

An allowlist bridges the two forms, but it did not include `coupon`, so the picker submitted without one and `maybe_auto_apply_coupon()` returned at its empty-string guard. That allowlist existed twice: the exported `PICKER_CONTEXT_FIELDS` (URL-trigger path) and a duplicated inline literal in `modal.js` (click path). This adds `coupon` to the constant and makes `modal.js` consume it, so both paths read one list.

**2. That fix alone was not enough — the picker is shared state.**

The picker is rendered **once per parent product** (`enqueue_modal()` dedupes into `self::$products[ $product_id ]`) and shared by every Checkout Button on the page targeting that product. The stamping loop only wrote a hidden field when none existed yet, and nothing clears the picker when the modal closes — so the **first button clicked froze the coupon for every later one**. Both orderings were reproduced with a test before fixing:

* Button A (`SAVE20`) then button B (`MEMBER10`) → the picker submitted `SAVE20`, discounting B's purchase with a coupon the publisher attached elsewhere.
* A coupon-less button first → an empty `coupon` input was stamped that blocked every later button, reproducing the original bug, order-dependently.

Fixed by extracting `applyContextFields()`, which is authoritative per open: it overwrites what the previous open left behind and removes the field when the clicked button has no coupon. This also fixes the same staleness for `after_success_url` and the attribution fields. Since the picker only renders for variable / variable-subscription / grouped products, a leaked coupon mostly landed on **subscriptions**, where WCS copies cart coupons onto the subscription and replays them onto every renewal.

The URL-trigger path had a related weakness: it took block context from the first button in DOM order, so a coupon configured for one locked variation could follow a reader to a different one. It now prefers an unlocked button, which stands for the whole product.

**3. The editor offered a coupon on donation products.**

`Newspack\Donations::disable_coupons()` filters `woocommerce_coupons_enabled` to `false` for any cart containing a donation ([newspack-plugin#2770](https://github.com/Automattic/newspack-plugin/pull/2770), Nov 2023). That check early-returns in `is_admin()`, so the editor happily showed the coupon picker for a Donate product and the frontend discarded the code silently.

Donation-ness can't be derived from product meta alone (the legacy Donate product is matched by ID, not by the `_newspack_is_donation` flag), so a read-only `newspack_is_donation` REST field on `product` delegates to `Donations::is_donation_product()`, and the editor warns that donation products do not support coupons and disables the field.

The control is **disabled rather than hidden**: `view.php` emits the stored code regardless of donation status, so hiding it would only remove the publisher's way to clear a code stranded by a product swap. "Remove" stays available for exactly that; "Edit" does not.

image

image

Worth knowing that the underlying policy is not absolute — `disable_coupons()` returns early under `is_admin()`, and `generateCart()` builds the anonymous-reader cart through `admin-ajax.php`, where coupons stay enabled. The copy is deliberately a product statement ("do not support") rather than a runtime guarantee.

**Cross-plugin changes (newspack-plugin):**

* `class-donations.php` — memoizes the legacy donation-product lookup, which `is_donation_product()` otherwise re-derives on every call, now once per row of a products REST response. Semantics unchanged; the new method is private.

**Coupon through the variation picker:**

1. With coupons enabled, create a percentage coupon and a variable (or variable-subscription) product with at least two priced variations.
2. Add a Checkout Button block for that product, attach the coupon, and leave "Allow the reader to select the variation before checkout" **checked**.
3. On the frontend, click the button, pick a variation, and confirm the discount appears in the order details.
4. Confirm the in-modal "Apply a coupon code" form stays hidden (the reader never typed the code).
5. Uncheck the setting and confirm the coupon still applies (the previously-working path is unaffected).

**Shared picker state — the regression this PR exists to prevent:**

6. Put **two** Checkout Button blocks on one page, both targeting the same variable product, with **different** coupons. Click the first, close the modal, then click the second: each must apply its own code. Before this fix the second submitted the first's.
7. Now the reverse: one button with no coupon and one with. Click the coupon-less one first, close, then the other. The discount must still apply.

**Donation feedback in the editor:**

8. Point a Checkout Button block at a Donate product. Expect "Donation products do not support coupons." and a disabled coupon field.
9. Attach a coupon to a normal product first, then switch the block to a Donate product: the notice becomes a warning naming the code, the field is disabled, and **Remove** is still available so the stored code can be cleared. Switching back re-enables the field with the code intact.
10. Confirm a non-donation Name Your Price product still shows the normal coupon picker — NYP itself was never restricted.

### Other information:

`checkout-button-trigger.test.js` covers the picker-context helpers, including both failing orderings from item 2 and the clear-on-next-open case; the two tests for the original bug were confirmed failing against the old code first. Two PHP tests pin the `newspack_is_donation` REST field against WooCommerce's schema-title plumbing, which would otherwise break silently. ESLint, `php -l` and PHPCS are clean on all changed files.

One coverage gap worth naming: **the `modal.js` click path is a `domReady` IIFE over page globals** and isn't directly unit-testable. Its logic now lives in the tested helpers, so what remains untested is the wiring; steps 6-7 are the real check.

Showing the coupon's effect on the prices in the picker was built here and then removed, to keep this PR to its charter. It will land separately, rendered in the template on page load rather than fetched — the server already knows every product/coupon pair on the page by the time the picker is built, so no endpoint is needed.


plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
```diff
@@ -58,6 +58,41 @@ export function findCheckoutButtonForm( root, productId, variationId = null ) {
...
+ * whole product, so its context — including any attached coupon — applies to
+ * whichever variation the reader picks. A locked button was configured for one
+ * specific variation and is only used when nothing better exists.
+ *
+ * @param {Document|HTMLElement} root      The DOM root to search.
+ * @param {string}               productId The requested product ID.
+ *
+ * @return {HTMLFormElement|null} The donor form, or null.
+ */
+function findContextDonorForm( root, productId ) {
```
dkoo, line 74: **Suggestion:** findContextDonorForm's all-locked fallback is untested — The new JS tests cover the "prefer an unlocked button over a variation-locked one" case, but not the fallback where only variation-locked buttons exist for the product and the first in DOM order is used. A companion case with two locked buttons asserting the first one's coupon is the one carried would pin the subtler half of this function's contract.

plugins/newspack-plugin/includes/class-donations.php
```diff
@@ -295,12 +295,47 @@ public static function is_donation_product( $product_id ) {
 		}
 
 		// Fall back to the legacy parent/child donation product check.
+		return in_array( $product_id, self::get_legacy_donation_product_ids(), true );
+	}
+
+	/**
+	 * Request-level cache of the legacy donation product IDs, keyed by the
```
dkoo, line 302: **Suggestion:** Legacy donation-product memo: docblock overstates invalidation, and there's no reset hook — The request-level memo in `get_legacy_donation_product_ids()` keys only on the `newspack_donation_product_id` option, and the docblock says saving donation settings mid-request invalidates it. That holds when the grouped **parent** product ID changes — but when an existing parent is reused and its child products are regenerated (new child IDs under a stable option), the memo serves a stale set for the rest of that request, and there's no invalidation hook. The sibling `$flagged_donation_product_ids` cache guards exactly this with a public `reset_flagged_donation_product_ids_cache()`; the legacy cache has no equivalent and `update_donation_product()` doesn't reset it. Real-world exposure is low: that regeneration runs only on admin settings-save/auto-repair, while the hot `is_donation_product()` consumers (`data-events/class-utils.php`, `listeners.php`) run during checkout/order/webhook requests. Worth either softening the docblock to say it invalidates when the *parent product option* changes, or adding a reset method for parity and calling it (plus the flagged reset) at the end of `update_donation_product()`.

plugins/newspack-plugin/includes/class-donations.php
```diff
@@ -295,12 +295,47 @@ public static function is_donation_product( $product_id ) {
...
+	 * Get the legacy donation product IDs: the grouped parent and its children.
+	 *
+	 * Memoized because the answer is request-invariant while callers are not:
+	 * is_donation_product() runs once per row of a products REST response, and
+	 * resolving this from scratch re-instantiates the parent product and each of
+	 * its children every time.
+	 *
+	 * @return int[] Legacy donation product IDs, empty when none is configured.
+	 */
+	private static function get_legacy_donation_product_ids() {
```
dkoo, line 334: **Suggestion:** No direct unit test for the refactored legacy donation path — `get_legacy_donation_product_ids()` is the one foundation-contract change, and its actual logic is only exercised indirectly — the newspack-blocks tests assert the `newspack_is_donation` REST-field contract against a stub that short-circuits before the legacy branch, so the grouped parent/child resolution and the new memo run untested. The refactor reads as behavior-preserving (the old `array_values()` + strict `in_array` never matched the `false` frequency defaults, so the new `array_filter` is equivalent), but a small `newspack-plugin` unit test — parent + configured children resolve `true`, a non-member resolves `false`, and a re-read after the option changes re-resolves — would lock the invariant this change is trusting.

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
```diff
@@ -107,12 +148,54 @@ export const PICKER_CONTEXT_FIELDS = [
...
+ * one — for `coupon` that is the difference between the reader's discount and
+ * a discount attached to a different button.
+ *
+ * @param {HTMLFormElement|null} targetForm Picker form to stamp.
+ * @param {Object|null}          data       Checkout data read from the clicked button's form.
+ * @param {string[]}             fields     Field names to stamp.
+ *
+ * @return {void}
+ */
+export function applyContextFields( targetForm, data, fields = PICKER_CONTEXT_FIELDS ) {
```
dkoo, line 170: **Nit:** Optional: consolidate the near-parallel picker helpers — `applyContextFields` (authoritative overwrite) and `copyContextFields` (preserve) sit side by side with opposite merge semantics, and `findContextDonorForm` re-implements the button-iteration scaffolding of `findCheckoutButtonForm`. The divergence is genuinely motivated and the docblocks explain it well, so this isn't a defect — but since the URL-trigger path is the only remaining `copyContextFields` caller, it could read its donor form once and reuse `applyContextFields`, and a shared "matching forms for a product" iterator would remove the duplicated lookup boilerplate. Only worth it if it doesn't muddy the "URL trigger preserves, click overwrites" distinction the comments draw.

dkoo (approved):
@miguelpeixe Preapproving as the suggestions are all related to inline documentation and test coverage—your choice whether to address before merging.

Manual QA passes in all scenarios. No blockers — this is a clean, well-tested fix, and the two scenarios that make an auto-applied coupon dangerous are both closed by the change itself. On the frontend I confirmed each Checkout Button emits its own coupon and a coupon-free button clears the shared picker rather than inheriting the previous one's code; on the editor side the new `newspack_is_donation` REST field correctly flags donation products (and leaves Name Your Price alone), disabling the coupon field while keeping Remove available. A couple of optional notes are inline — the legacy donation-product memo and its test coverage — neither blocks merge.

Verification was headless: the REST field, server-side render, editor gating, and debug log all checked out against an isolated env; the live click-through of the variation-picker modal wasn't exercised (browser access was unavailable here), but that path is covered by the JS unit tests.

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
miguelpeixe: Added `falls back to the first locked button when no unlocked one exists` — two locked buttons for the same product, asserting the first in DOM order supplies the coupon. 35 tests passing.

plugins/newspack-plugin/includes/class-donations.php
miguelpeixe: Took the parity option rather than softening the docblock, since the staleness is real and reachable.

Confirmed the scenario precisely in `update_donation_product()`: it reuses an existing parent (line 632), regenerates children when a frequency is missing (652-688), then writes the *same* parent ID back (line 700) — so the option is stable while child IDs change, exactly as you described. Also confirmed it reset neither cache.

Added `reset_legacy_donation_product_ids_cache()` for parity with the flagged one, and called **both** at the end of `update_donation_product()` — the flagged cache had the same latent staleness there. The docblock now says the key only covers the parent and points at the reset method rather than claiming settings saves invalidate it.

plugins/newspack-plugin/includes/class-donations.php
miguelpeixe: Added `tests/unit-tests/donations-legacy-products.php` — 7 tests, 15 assertions, covering the three cases you named plus the ones that motivated the reset method:

- grouped parent and each configured child resolve `true`; an unrelated product `false`
- no parent configured, and an option pointing at a non-grouped product, both resolve `false`
- changing the parent option re-resolves (the memo key working)
- children regenerated under an *unchanged* parent serve the stale set until `reset_legacy_donation_product_ids_cache()` is called, then re-resolve — pinning the limitation the reset exists for

The products carry no donation flag, so these exercise the legacy branch rather than the meta fast path. Full `newspack-plugin` suite still green (2459 tests, 6994 assertions) — I checked specifically because `wc-mocks.php` has bitten this suite before.

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
miguelpeixe: Leaving this one, on the condition you set — collapsing them would muddy the distinction rather than just tidy it.

The specific problem: `copyContextFields` preserves existing target values, while `applyContextFields` deliberately *removes* a field when the clicked button has no value for it. Routing the URL-trigger path through the overwrite version would give it delete semantics against a form it did not render. That is safe today only because `render_form()` happens to emit none of the seven names — a fact about the current markup, not a contract. It would also drop the tested "does not overwrite existing target fields" guarantee, which is the property that makes `copyContextFields` safe to point at a server-rendered form.

The `findContextDonorForm` / `findCheckoutButtonForm` iteration duplication is a fairer hit — that half is pure boilerplate with no semantics attached. Happy to do a shared "matching forms for a product" iterator, but I would rather not refactor the critical path in this PR right after it was trimmed back to scope. Can pick it up separately if you want it.

PR #769, opened 2026-07-29
fix(newsletters): cut per-render subscription-list and gate overhead

dkoo:
### All Submissions:

Cuts the per-render cost of the Newsletter Subscribe block on sites with many subscription lists. With Access Control's premium newsletters active, rendering the block ran the full subscription-list lookup and a per-list content-restriction check on every article view; at a scale of thousands of lists this dominated render time.

Four behaviour-preserving changes:

- **Memoize `Subscription_Lists::get_all()`** per request, invalidated on list create/update/delete, so repeated callers stop re-querying and rebuilding every list. Gated off under PHPUnit so static state does not leak between tests.
- **Cache the empty result in `Newspack_Newsletters_Subscription::get_lists_config()`** (`null !== self::$lists_config` instead of a truthy check), so an empty config is not recomputed on every call. Adds `reset_lists_config_cache()` for invalidation.
- **Short-circuit `Premium_Newsletters::filter_subscription_lists()`** when no premium newsletter gates exist — no list can be restricted, so the per-list loop is skipped. Reuses the cached `get_gates()`.
- **Reorder `Content_Restriction_Control::is_post_restricted()`** so the empty-gates check runs before `user_can()` — identical result, but skips the capability check (which fans out to other plugins' `user_has_cap` filters) for posts with no gates.

Refs NPPM-3046. The `get_all()` memo also addresses NPPM-2832 (~1000 list posts loaded into the object cache per request). A follow-up will remove the remaining single per-request list query via targeted assembly — caching the full set is not viable, as it exceeds memcached's per-item limit at that scale.

1. On a site with premium newsletters active and several subscription lists, add the Newsletter Subscribe block to a post.
2. With Query Monitor open, load the post on the front end; note the subscription-list query and per-list capability checks.
3. Apply this branch and reload. Repeated `get_all()` calls within the request collapse to one; with no premium newsletter gates, the per-list restriction check no longer runs.
4. Create, edit, and delete a subscription list; confirm the Subscribe block reflects the change.

### Other information:

* [ ] Have you successfully run tests with your changes locally? PHPCS is clean; PHPUnit runs in CI (a worktree needs an isolated env for local runs).


dkoo:
Applied fixes from a self-review pass (`1deb2eb`). All behaviour-preserving; PHPCS clean.

**Broadened the cache invalidation** in `Subscription_Lists`. The memo previously flushed only on `save_post_{CPT}` and `before_delete_post`, which missed two paths that change what `get_lists_config()` returns:

- **Trash/untrash and permanent delete.** `delete_list()` calls `wp_delete_post()` without the force flag, so a list is trashed (firing `trashed_post`/`deleted_post`, not `before_delete_post`). Now hooked via `deleted_post` — matching the sibling `Newspack_Newsletters_Subscription` delete cache, which also closes a `before_delete_post` repopulation window — plus `trashed_post` and `untrashed_post`.
- **Provider-settings writes.** These live in post meta and fire no `save_post`, yet drive `is_active()`/`is_configured_for_current_provider()`/`to_array()`. Now hooked via `added_/updated_/deleted_post_meta`, key-scoped to the list's meta keys so the many unrelated meta writes elsewhere on the site short-circuit before the post-type lookup.

**Gated the `get_lists_config()` memo under `IS_TEST_ENV`**, mirroring `get_all()`. Static memos persist across the DB rollback between PHPUnit cases; caching empty results (the `null !==` change) widened that exposure.

**Documented that `get_lists_config()` bakes in the current user's view** — on gated sites the premium-newsletters filter is per-user, so a caller that switches users mid-request must `reset_lists_config_cache()` first. No current caller does; the note guards a future one.

Also tightened the invalidation docblocks and renamed the delete handler now that it covers status transitions too.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -402,6 +426,12 @@ public static function save_post( $post_id ) {
 	 * @return Subscription_List[]
 	 */
 	public static function get_all() {
+		// Skip the memo under PHPUnit: tests roll back the database between cases
+		// but static memos persist, which would leak one test's lists into the next.
+		$use_cache = ! ( defined( 'IS_TEST_ENV' ) && IS_TEST_ENV );
+		if ( $use_cache && null !== self::$all_lists ) {
+			return self::$all_lists;
+		}
```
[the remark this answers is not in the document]
dkoo: The `get_all()` memo is deliberately disabled under `IS_TEST_ENV` (guard at the top of `get_all()`): PHPUnit rolls back the DB between cases while the static memo persists, so a cache-hit assertion would leak one case's lists into the next. That makes a "call `get_all()` twice → assert cache hit, then `flush_cache()` → assert change" test impossible in this harness. Observable correctness is covered by `test_new_active_list_appears_in_lists_config`; exercising the raw memo would need an out-of-band query-count check rather than a rollback-based unit test — leaving as-is.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -413,9 +443,67 @@ public static function get_all() {
...
+	/**
+	 * Flush the list caches when a Subscription List's provider settings change.
+	 * Those settings drive is_active()/is_configured_for_current_provider()/
+	 * to_array() and are stored in post meta, written without a save_post fire. The
+	 * meta-key check short-circuits before the post-type lookup for the many
+	 * unrelated meta writes elsewhere on the site.
+	 *
+	 * @param int    $meta_id  Meta ID (unused).
+	 * @param int    $post_id  Post the meta belongs to.
+	 * @param string $meta_key Meta key written.
```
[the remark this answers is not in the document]
dkoo: Fixed in 145d6b7e3 — annotation is now `int|string[]` (int for added/updated meta, array of IDs for deleted meta). Verified against `wp-includes/meta.php`: `deleted_post_meta` passes `string[] $meta_ids`.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -413,9 +443,67 @@ public static function get_all() {
...
+			Subscription_List::TYPE_META,
+			Subscription_List::PROVIDER_META,
+			Subscription_List::REMOTE_ID_META,
+			Subscription_List::REMOTE_NAME_META,
+			Subscription_List::SUBSCRIBER_COUNT_META,
+		];
+		if ( in_array( $meta_key, $list_meta_keys, true ) && self::CPT === get_post_type( $post_id ) ) {
+			self::flush_cache();
+		}
+	}
```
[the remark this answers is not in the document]
dkoo: Fixed in 145d6b7e3 — hoisted to a `private const LIST_META_KEYS`, so the array is no longer rebuilt on each `added/updated/deleted_post_meta`.

plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php
```diff
@@ -569,10 +569,19 @@ public static function clear_lists_cache_on_delete( $post_id, $post = null ) {
...
 	 *
+	 * Note: on sites with premium newsletter gates, the result is filtered per the
+	 * current user (via the newspack_newsletters_subscription_lists filter →
+	 * Premium_Newsletters::filter_subscription_lists()). The per-request memo
+	 * therefore bakes in the current user's view; a caller that changes the current
+	 * user mid-request must call reset_lists_config_cache() before re-reading.
+	 *
 	 * @return array[]|WP_Error Associative array with list configuration keyed by list ID or error.
 	 */
 	public static function get_lists_config() {
```
jason10lee, line 580: **Suggestion:** A provider switch mid-request leaves the config memo built for the old provider — `get_lists_config()` is provider-scoped, since it goes through `Subscription_Lists::get_configured_for_current_provider()`, and `reset_lists_config_cache()` is currently reachable only from `Subscription_Lists::flush_cache()`. `Newspack_Newsletters::set_service_provider()` does not call it, so a provider or credentials change mid-request leaves a config built for the previous provider.

The new test is the clearest evidence: it has to issue `reset_lists_config_cache()` by hand immediately after `set_service_provider()` (tests/test-subscription-lists.php:33-34), which is exactly the call missing in production. The staleness pre-dates this PR for non-empty results, but this is the change that makes the empty case stick too, and it is also the change that introduces the reset method — so it is a natural moment to wire it into `set_service_provider()`.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -29,6 +29,30 @@ class Subscription_Lists {
...
+	];
+
+	/**
+	 * Per-request memo of every Subscription List. Reset by flush_cache() whenever a
+	 * list is created, updated, trashed, untrashed, deleted, or its provider
+	 * settings (post meta) change. Disabled under PHPUnit — see get_all().
+	 *
+	 * @var Subscription_List[]|null
+	 */
+	private static $all_lists = null;
```
jason10lee, line 54: **Suggestion:** "Per request" is unbounded in CLI, cron and Action Scheduler processes — The invalidation hooks cover in-process writes well — I traced the mutation paths and did not find a gap. The lifetime is the part worth a thought: WP-CLI imports, WP-Cron, and the Action Scheduler queue behind `Premium_Newsletters::SCHEDULED_HOOK` all keep a static alive for minutes or longer, and none of the seven hooks fire for a write made by a *different* process. An editor publishing a list in wp-admin during a long CLI run stays invisible to that run.

A `wp_cache_*` entry in its own group, keyed on `wp_cache_get_last_changed( 'posts' )`, inverts those properties: it invalidates on any post write from any process, survives across requests (a larger win on Atomic/WPCOM than a per-request memo), needs no hook registrations, and the WP test suite flushes the object cache between cases, so the `IS_TEST_ENV` bypass and the coverage gap it creates would go away with it. That is a bigger change than a hotfix wants, so if the static is the right shape for now, skipping the memo under `defined( 'WP_CLI' ) && WP_CLI` would bound the worst case cheaply.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -40,6 +64,21 @@ public static function init() {
...
+		// Bust the per-request list memo whenever a list changes, so the next read
+		// in the same request is never stale. save_post covers insert/update;
+		// status transitions (trash/untrash) and permanent deletes fire their own
+		// hooks; and the provider settings that drive get_lists_config() live in
+		// post meta, written without a save_post fire, so the meta hooks are needed
+		// too. deleted_post (not before_delete_post) matches the sibling
+		// Newspack_Newsletters_Subscription delete cache and avoids a repopulation
+		// window.
+		add_action( 'save_post_' . self::CPT, [ __CLASS__, 'flush_cache' ] );
+		add_action( 'deleted_post', [ __CLASS__, 'maybe_flush_cache_for_post' ], 10, 2 );
```
jason10lee, line 76: **Suggestion:** deleted_post fires before clean_post_cache(), leaving a repopulation window — I checked this against core rather than from memory: `wp_delete_post()` runs `do_action( 'deleted_post', $post_id, $post )` at wp-includes/post.php:3934 and `clean_post_cache( $post )` at :3936. At the moment this callback fires, both the post cache and the `WP_Query` result cache are still warm — which is also why the `get_post_type( $post_id )` fallback in `maybe_flush_cache_for_post()` works at all. The consequence is that anything hooked to `deleted_post` after priority 10 that reaches `get_all()` repopulates the memo with the row being deleted, and that entry then survives the rest of the request.

`after_delete_post` (:3957) carries the same `( $post_id, $post )` signature and runs after the cache is cleaned, which closes the window without giving up the `$post` argument. Worth noting the sibling handler in `Newspack_Newsletters_Subscription` has the same placement, so if matching it was deliberate this is a consistency call rather than an oversight — happy to leave it on that basis.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
```diff
@@ -346,13 +346,17 @@ public static function is_post_restricted( $is_post_restricted, $post_id = null,
 
 		$user_id = $user_id ?? get_current_user_id();
 
-		// Don't restrict this post for users who can edit it.
-		if ( ! empty( $post_id ) && user_can( $user_id, 'edit_post', $post_id ) ) {
+		// If no gates apply to this post, nothing can restrict it. Checked before
+		// the capability check below: user_can() fans out to other plugins'
+		// user_has_cap filters and can be expensive, and an empty gate set returns
+		// false regardless of the user's caps, so the outcome is identical either way.
+		$post_gates = self::get_post_gates( $post_id );
```
jason10lee, line 353: **Suggestion:** Reorder is outcome-identical but moves the privileged path slightly slower — The correctness claim holds: both branches return `false`, and `get_post_gates()` has no side effect the function consumes, so the return value is order-independent.

On cost, I benchmarked the old ordering against the new one on a gated site with the object cache flushed before each arm. At 200 lists, an admin who can edit every list goes from 403 queries / 72ms to 404 queries / 102ms. So it is one extra query overall, not a gate build per list — the repeated `Content_Gate::get_gates()` calls collapse onto WP's in-request query cache, which is the part a static read of the code makes look worse than it is. The residual is CPU in per-list rule evaluation, roughly 0.15ms per list, so a few thousand lists would be a few hundred milliseconds for logged-in editors on sites that have gates. Anonymous traffic is unchanged and dominates, so this still looks like the right trade — I am flagging it mainly because NPPM-3046 is itself an admin-slowness report, so it is useful to know the privileged path moved the other way.

plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
```diff
@@ -160,6 +160,11 @@ public static function filter_subscription_lists( $lists ) {
 		if ( is_admin() ) {
 			return $lists;
 		}
+		// If there are no premium newsletter gates, no list can be restricted, so
+		// skip the per-list check entirely. get_gates() is statically cached.
+		if ( empty( self::get_gates() ) ) {
```
jason10lee, line 165: **Suggestion:** The short-circuit helps sites with no gates, not the ones the description names — The guard itself is sound, and I verified the equivalence it rests on: `get_gates()` runs `Content_Gate::get_gates( GATE_CPT, 'publish', true )`, which is the identical query `get_post_gates()` issues for a `newspack_nl_list` post (class-content-restriction-control.php:149-153), so an empty gate set cannot restrict any list.

What is worth flagging is the scope. Measured at 200 lists in a local env: with no premium newsletter gates the filter drops from 402 queries / 102ms to 2 queries / 1.0ms, and the saving is flat rather than per-list. With one published newsletter gate the same call stays at 404 queries / 109ms — unchanged from before. So this hunk pays off precisely on the sites that do not have premium newsletters configured, while the description frames the problem as arising "with Access Control's premium newsletters active". Changes 1 and 2 do help that population, so nothing is wrong here — but would it be worth adjusting the description so publishers with gates are not expecting this part to be their fix?

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -40,6 +64,21 @@ public static function init() {
```
jason10lee, line 64: **Nit:** The invalidation wiring is testable even with the memo disabled — This is a different point from the cache-hit test already discussed and reasonably declined. The memo cannot be exercised under `IS_TEST_ENV`, but the code deciding *whether* it gets flushed is pure and unaffected by that gate. `has_action()` assertions would pin the seven registrations — the thing most likely to be dropped by a future edit — and `maybe_flush_cache_for_meta()` can be called directly to assert it discriminates correctly: a `LIST_META_KEYS` key on a non-list post and an unlisted key on a list post must both leave the caches alone. No cross-test state is involved, so it would not reintroduce the leakage that motivated the bypass.

plugins/newspack-newsletters/includes/class-subscription-lists.php
```diff
@@ -40,6 +64,21 @@ public static function init() {
...
+		// in the same request is never stale. save_post covers insert/update;
+		// status transitions (trash/untrash) and permanent deletes fire their own
+		// hooks; and the provider settings that drive get_lists_config() live in
+		// post meta, written without a save_post fire, so the meta hooks are needed
+		// too. deleted_post (not before_delete_post) matches the sibling
+		// Newspack_Newsletters_Subscription delete cache and avoids a repopulation
+		// window.
+		add_action( 'save_post_' . self::CPT, [ __CLASS__, 'flush_cache' ] );
+		add_action( 'deleted_post', [ __CLASS__, 'maybe_flush_cache_for_post' ], 10, 2 );
+		add_action( 'trashed_post', [ __CLASS__, 'maybe_flush_cache_for_post' ] );
```
jason10lee, line 77: **Nit:** Mixed accepted_args across three registrations of one callback is worth a comment — The arity is correct as written, which is the reason to protect it: `wp_trash_post()` and `wp_untrash_post()` both pass `$previous_status` as a second argument (`untrashed_post` since WP 5.6, `trashed_post` since 6.3), so registering with the default `accepted_args` of 1 is what keeps a string out of the `$post` parameter — and the `$post instanceof \WP_Post` guard is the only backstop if that changes. Because the `deleted_post` registration one line above uses `10, 2`, a later edit that "fixes" the inconsistency would introduce the bug. One line of comment on why the arity differs would make that hard to get wrong.

jason10lee (approved):
Approved with comments.  Well-tailored fix that will reduce a pain point that seems pretty particular, but severe where it occurs.  Went ahead and ran some trials against synthetic datasets: with no premium newsletter gates, filtering 200 lists drops from 402 queries / 102ms to 2 queries / 1.0ms.  Interestingly, the savings appear to be flat rather than per-list (further analysis below).

Release risk: LOW — no confirmed blockers, the contract impact classifies Safe, and the change is verified at runtime rather than by reading. Both full suites pass (newspack-plugin 2370, newspack-newsletters 718) and PHPCS is clean over all six files.

Two things worth a look before merge. 
- The no-gates short-circuit does not reach sites that have premium newsletters active — the same 200-list call still costs 404 queries / 109ms there — and that is the configuration the description names as the problem, so the description may promise more than this particular hunk delivers. 
- Separately, the `is_post_restricted()` reorder is outcome-identical but seems to move the privileged path the wrong way: an admin on a gated site goes from 403 queries / 72ms to 404 / 102ms at 200 lists. That is one extra query rather than a gate build per list, so the cost is modest CPU in rule evaluation, but NPPM-3046 is itself an admin-slowness report.

On whether memoizing an empty lists config could silently drop a subscription: repeated tests suggest it cannot fire on any current path. Registration resolves lists through `Subscription_List::from_public_id()` in `Contacts::upsert()` (class-newspack-newsletters-contacts.php:203-210) and never reads `get_lists_config()`, and registration runs on `template_redirect` before any block render, so the memo is cold when `wp_set_current_user()` fires. The contract the new docblock states is forward-looking rather than load-bearing today, which is worth knowing if anyone later moves a read earlier in the request.

The remainder are non-blocking robustness notes, left inline.

Before this ships to a publisher with premium gates and a large list count, it is worth timing the newsletters list endpoint as a logged-in editor, since that is the one path the reorder moves in the slower direction.

- `plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php:100` — **Suggestion:** get_gates() is now load-bearing but has neither invalidation nor a test bypass — Change 3 promotes this memo from an optimization into the condition that decides whether premium filtering runs at all. Every other memo this PR touches got either an invalidation path or an `IS_TEST_ENV` bypass; this one has neither, and the test file reaches past the API to reset it through `ReflectionProperty` in `set_up()` (tests/unit-tests/content-gate/class-premium-newsletters.php:61-63). Exposing a reset, or giving it the same bypass used elsewhere in this PR, would mean each new test file does not have to re-derive that trick.
- `plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php:36` — **Nit:** @var array no longer describes $lists_config — `null` is now the cache-miss sentinel the `null !==` check tests against, so `@var array` on the docblock above is inaccurate in a load-bearing way — `array|null` would describe it. The new `$all_lists` in `class-subscription-lists.php` already annotates `Subscription_List[]|null` correctly, and on a PHP 8.3 target it could carry a real declaration (`private static ?array $all_lists = null`).

plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php
dkoo: Fixed in febb0262f — `set_service_provider()` now calls `reset_lists_config_cache()`, and I dropped the manual reset the test had to make (which was the tell). Scoped to `$lists_config`; `$all_lists` is provider-independent so it does not need it.

plugins/newspack-newsletters/includes/class-subscription-lists.php
dkoo: Added in febb0262f — `test_cache_invalidation_hooks_are_registered` pins the seven registrations, and `test_cache_invalidation_discriminates` seeds the static via reflection and asserts that a non-list post / an unrelated key leave it intact while a list post or a list meta key flush it. Good call that the guard is testable with the memo off.

plugins/newspack-newsletters/includes/class-subscription-lists.php
dkoo: Agreed the lifetime is the real property. Deferring the `wp_cache_*` + `last_changed` version as a follow-up (bigger than this hotfix, as you note). Holding the `WP_CLI` skip too for now — subscription-list writes landing mid-CLI-run are narrow, and the in-process invalidation covers the common case.

plugins/newspack-newsletters/includes/class-subscription-lists.php
dkoo: Leaving on `deleted_post` to match the sibling handler in `Newspack_Newsletters_Subscription` — the placement was deliberate, and no registered callback repopulates in that window today. If we revisit, moving both handlers to `after_delete_post` together is the way to keep them consistent.

plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
dkoo: Acked — keeping the reorder. Anonymous traffic dominates and is unchanged; the one extra query / ~0.15ms per list for logged-in editors is the acceptable side. Useful to have the number logged against NPPM-3046.

plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
dkoo: Fair — this hunk pays off on no-gate sites; Changes 1 and 2 serve the gated population the description names. Leaving the description as-is for this hotfix.

plugins/newspack-newsletters/includes/class-subscription-lists.php
dkoo: Leaving the arity as-is without the extra comment for now — the `$post instanceof WP_Post` guard already absorbs the differing second arg (a `$previous_status` string just falls through to `get_post_type()`), so a future consistency edit would not actually break it.

dkoo:
On the two notes from the review body (outside the changed lines):

- **B — `$lists_config` `@var`:** applied in febb0262f (`@var array|null`, since `null` is now the cache-miss sentinel — same accuracy fix as Copilot's `$meta_id` note).
- **A — `get_gates()` is load-bearing with no invalidation/bypass:** valid, but it reaches into the content-gate memo (broader than this hotfix), so deferring it as a follow-up rather than expanding scope here.

Thanks for the thorough pass. The provider-switch fix and the invalidation guard test are in febb0262f; everything else is addressed in the inline threads.

PR #770, opened 2026-07-29
feat(cli): audit-membership-subscriptions detects gift/order-only memberships (NPPD-2070)

adekbadek:
### All Submissions:

Adds a read-only WP-CLI command, `wp newspack audit-membership-subscriptions`, that classifies every access-granting WooCommerce Membership by how it is payment-backed.

WooCommerce Memberships grants access to the membership holder (`wc_user_membership.post_author`); Access Control grants it to the subscription's customer (`Access_Rules::has_active_subscription` → `WooCommerce_Connection::get_active_subscriptions_for_user`). On a gift purchase those disagree — membership on the recipient, subscription on the buyer. A membership backed only by a paid one-off order has nothing recurring for a `subscription` rule to key on. Neither shape is visible in the plan→gate translation or the access-parity diff; both land in the comp/legacy bucket and read as accepted loss. In the first field case they were 13% of a site's paid base.

Each plan's memberships are classified into one of ten named classes, so the per-plan summary reconciles against the plan's member count instead of leaving a remainder. The two actionable cohorts (`gift`, `order-only`) are reported member-by-member:

| Column | Why it is there |
|---|---|
| `member_id` / `member_email` | the reader who loses access |
| `buyer_id` / `buyer_email` | who the publisher recognises the gift programme by (falls back to the guest billing email) |
| `record` / `record_status` / `products` | the backing subscription or order, incl. variation IDs (a gate rule may name either) |
| `member_own_access_subscriptions` | whether the member holds anything of their own — **evidence, not a verdict** (a recurring donation is a live subscription that unlocks no content) |

Flags: `--plan-ids=<csv>` (defaults to every published plan), `--only=gift|order-only|order-only-unpaid|subscription-missing`, `--format=table|csv|json|ids`. Both parsers are strict — a typo aborts rather than silently narrowing the report, because an empty gift list reads as "this site has no gift problem".

**Detection only.** Buyer-vs-recipient intent is a publisher question: no subscription is re-homed, no access is granted. `--format=ids` emits the deduplicated user-ID list (one line per *user*, not per membership — a reader with two flagged memberships must not be granted two $0 subscriptions), so the detector and whatever grants replacement access act on one identical list.

**Shapes deliberately *not* flagged**, each with a regression test:

- **Team seats.** Memberships for Teams stamps the team's subscription — owned by the team owner — onto every seat (`Subscriptions::adjust_team_member_user_membership_data`). Without the `_team_id` discriminator every seat reads as a gift, feeding an institutional site's whole population into $0-subscription granting on top of the group subscription `migrate-teams` already creates.
- **Subscriptions Gifting gifts whose recipient IS the holder.** Access Control resolves gifted subscriptions to their recipient. Guarded on the subscription still being access-granting — the runtime checks status *before* the gifting rule, so a dead gift carries nothing over and its holder is losing access like any other lapsed member.
- **`order-only-unpaid`** (refunded / failed / cancelled order) is off the default report: a free subscription is the wrong answer for a refunded purchase.
- **`on-hold`** is access-granting under WCM but not under Access Control, so its holder is reported as losing access rather than covered. That cohort is its own migration question (NPPD-2052).

**Discrepancies disclosed rather than silently absorbed:**

1. `ACCESS_GRANTING_SUBSCRIPTION_STATUSES` is `active` + `pending-cancel`, mirroring `WooCommerce_Connection::ACTIVE_SUBSCRIPTION_STATUSES`. A unit test asserts the two lists are identical, so drift fails the build rather than a migration.
2. Membership statuses come from WCM's `get_active_access_membership_statuses()` (prefixed to the `wcm-` post statuses `get_posts()` wants), so a site filtering `wc_memberships_active_access_membership_statuses` is audited on its own terms. `wcm-pending` ("Pending Cancellation") is access-granting and is included.
3. `member_own_access_subscriptions` delegates to `WooCommerce_Connection::get_active_subscriptions_for_user()` rather than re-deriving the rule. Re-deriving diverged on gifted subscriptions in both directions, and the wrong direction prints reassurance for a reader the gates will lock out.
4. `migrate-manual-members` selects members by user ID (`--user-ids` / `--user-ids-file`, NPPD-2055 / #689, now merged), so the closing line hands the operator a command rather than a caveat. Verified end-to-end: `--format=ids > list.txt` → `migrate-manual-members --user-ids-file=list.txt --product-id=<id> --dry-run` selects exactly the flagged readers.

Registered only where WooCommerce Memberships is active, alongside the other `migrate-*` commands.

**Sites where an earlier build of this command was already run need one repair step.** The batch walk's keyset `posts_where` filter was registered globally, so every WP_Query the walk's callback triggered inherited the seek clause. On CPT-storage sites that includes WooCommerce Subscriptions' customer lookup, and the truncated result is written to that reader's `_wcs_subscription_ids_cache` user meta — a stored list nothing rewrites afterwards, since a renewal or status change only updates a cache already correct for the subscription it touches. The filter is now scoped to the walk's own query. To repair an affected site, run **WooCommerce → Status → Tools → Delete Customer Subscription Cache**; the lists rebuild on demand. "Generate Customer Subscription Cache" is not the tool for this — it only fills users who have no cache, and a poisoned entry is present, not missing.

Closes NPPD-2070.

1. **Unit tests** (no WooCommerce needed — the classifier is pure):
   ```
   cd plugins/newspack-plugin && n test-php --group Memberships_Audit
   ```
   45 tests / 82 assertions. The full suite is unaffected (3415 tests, 6 pre-existing skips).

2. **On an env with WooCommerce + Subscriptions + Memberships active**, build one membership of each shape on a single plan:

   | Shape | How |
   |---|---|
   | gift | membership on user A, `active` subscription owned by user B |
   | gift, recipient also has a donation | as above, plus an `active` subscription of A's own on a product no gate covers |
   | member-owned | membership and `active` subscription both on user A |
   | gift-inactive | as gift, but the subscription `cancelled` |
   | member-owned-inactive | membership and `on-hold` subscription both on user A |
   | order-only | `completed` order, no subscription, `_order_id` on the membership |
   | order-only-unpaid | same but the order `refunded` |
   | team-backed | as gift, plus `_team_id` on the membership |
   | subscription-missing | `_subscription_id` pointing at a non-existent ID |
   | no-purchase-record | membership with neither meta key |

   (`pending-cancel` and `on-hold` are only reachable from `active`, so set `active` first.)

3. `wp newspack audit-membership-subscriptions` — the per-plan summary counts all ten classes and the table lists only the gift + order-only rows, marked `←` in the summary. Confirm the gift row's `buyer_email` is the buyer and `member_email` the recipient, and that the donation-holding recipient is still flagged with their own subscription shown in `member_own_access_subscriptions`.

4. `wp newspack audit-membership-subscriptions --format=ids` — one line per flagged **user**, deduplicated (give one user two flagged memberships to confirm). No narration, no progress bar.

5. `wp newspack audit-membership-subscriptions --format=csv > out.csv` — `out.csv` holds pure CSV. The per-class counts go to STDERR. Repeat with `--format=json`. With `--only=subscription-missing` on a site that has none, confirm an empty run still emits a valid document (`[]` / a header row) rather than prose.

6. Set a flagged membership's `post_author` to `0`, then run `--format=ids`. The user ID list is unchanged and a warning on STDERR names how many flagged memberships have no member account and need handling by hand.

7. `--only=gifts` and `--plan-ids=12abc` both abort with a specific error, and so do `--only=` and `--plan-ids=,,` — a flag passed with an unexpanded shell variable must not quietly run the default report. A bare `--plan-ids` or `--only` (no `=value`) aborts too, read off the raw command line, because WP-CLI strips such a flag before the command sees it. `--plan-ids=<id of a non-plan post>` errors with "nothing was audited" rather than reporting a clean bill of health.

### Other information:


plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+		// Seeking on the last ID seen is stable by construction, and drops the
+		// deep-offset cost on the large tables this walk exists for.
+		$last_id      = 0;
+		$where_filter = function ( $where ) use ( &$last_id, $wpdb ) {
+			if ( $last_id > 0 ) {
+				$where .= $wpdb->prepare( " AND {$wpdb->posts}.ID > %d", $last_id );
+			}
+			return $where;
+		};
+		add_filter( 'posts_where', $where_filter );
```
dkoo: **Blocker:** Keyset-paging filter leaks into nested queries and poisons the WCS customer cache — `add_filter( 'posts_where', $where_filter )` registers globally, and `$last_id` is assigned before `$callback` runs, so every `WP_Query` executed inside the batch callback inherits `AND ID > <last membership ID>`. The audit's own evidence column reaches one: `describe_own_access_subscriptions()` → `get_active_subscriptions_for_user()` → `wcs_get_users_subscriptions()` → on a customer-cache miss, WCS queries orders through `WP_Query` with no `suppress_filters`. Membership posts are created after the subscriptions backing them, so their IDs are almost always higher — the member's own subscriptions get filtered out, and WCS then **writes the empty result to its persistent customer cache** (`_wcs_subscription_ids_cache` user meta), which is the same value the live paywall (`Access_Rules::has_active_subscription`) and the membership-expiry guard read at runtime.

**To reproduce** (env with WCS CPT storage, a member owning one active subscription):
1. `delete_user_meta( $uid, '_wcs_subscription_ids_cache' ); wp_cache_flush();` — cold cache, as any not-yet-cached reader.
2. Add the walk's clause (`AND {$wpdb->posts}.ID > 40`, any ID above the subscription's) via `posts_where`, then call `WooCommerce_Connection::get_active_subscriptions_for_user( $uid, [], true )` → returns `[]`. Baseline without the filter: 1 subscription.
3. `remove_filter(...)` and call again → still `[]`; `get_user_meta( $uid, '_wcs_subscription_ids_cache', true )` → `[]`. The corruption outlives the audit run.

**Risk if not addressed:** If such a reader's stored empty list is never rewritten — it is only rebuilt when the subscription's customer changes or the cache is manually regenerated, not on renewal or status change — the reader stays locked out of gated content indefinitely with nothing tracing back to the audit; they are handed to the grant step as uncovered and receive a free subscription they did not need (which on a one-per-customer product also blocks them from buying again); and the expiry guard reads the same empty list and lets their membership cancellation through. Each repeat run adds anyone whose cache has gone cold since.

**Consider:** Make the paging condition apply only to the walk's own query rather than to every post query that happens while it is registered — for example, tag the query arguments with a private flag and have the filter check the passed query for that flag (its second argument) before appending anything. Alternatively, unregister the condition immediately before invoking the per-batch callback and re-register it before fetching the next page. The scoped-filter form removes the whole class of nested-lookup effects at once.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+		$sleep_arg = \WP_CLI\Utils\get_flag_value( $assoc_args, 'sleep', self::DEFAULT_SLEEP_SECONDS );
+		if ( ! is_numeric( $sleep_arg ) || (float) $sleep_arg < 0 ) {
+			WP_CLI::error( sprintf( 'Invalid --sleep value "%s". Pass a non-negative number of seconds, or 0 to run without pausing.', $sleep_arg ) );
+		}
+		$sleep_seconds = (float) $sleep_arg;
+
+		$format = \WP_CLI\Utils\get_flag_value( $assoc_args, 'format', 'table' );
+		// Only the table format narrates. csv/json/ids are consumed by redirecting
+		// them to a file, and any prose in that file makes it unparseable.
+		$quiet = 'table' !== $format;
```
dkoo, line 350: **Suggestion:** Machine-readable formats drop the reconciling per-class counts the help text promises — The synopsis says "Counts for every class are printed regardless," but `$quiet = 'table' !== $format` suppresses both the per-plan summary and the all-plans totals for `csv`, `json` and `ids`. Suppressing them on STDOUT is right for stream purity, but an operator who exports the cohort has no way to get the reconciling counts — the numbers the 13%-of-paid-base framing depends on — without a second full walk of the membership table, which on a large site is the expensive part.

**Consider:** Print the per-class counts to STDERR in the machine-readable formats. That keeps redirected output parseable while giving the operator the reconciliation the documentation promises, without a second full walk of the membership table.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+	 *
+	 * One line per user, not per membership: this list is handed to
+	 * `migrate-manual-members`, and a reader holding two flagged memberships
+	 * would otherwise be granted two $0 subscriptions.
+	 *
+	 * @param array[] $rows Report rows.
+	 *
+	 * @return int[]
+	 */
+	public static function flagged_user_ids( array $rows ) {
```
dkoo, line 800: **Suggestion:** --format=ids silently drops flagged rows whose holder has no account — `flagged_user_ids()` skips `member_id` 0, so orphaned memberships (holder account deleted) and guest-backed rows that appear in the table and CSV vanish from the ids output with no signal. Verified: an orphaned gift membership raises the reported membership count to 6 while ids still emits 4 users, and stderr stays empty. Dropping them is correct — a $0 subscription cannot be granted to user 0 — but a `WP_CLI::warning()` to STDERR when zero-ID rows are excluded would keep the stream pure while telling the operator that K flagged memberships need manual handling and won't be on the list.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+					'newspack_memberships_audit_unknown_class',
+					sprintf( '"%s" is not a reportable class. Use one or more of: %s.', $token, implode( ', ', self::SELECTABLE_CLASSES ) )
+				);
+			}
+			if ( ! in_array( $token, $classes, true ) ) {
+				$classes[] = $token;
+			}
+		}
+
+		return empty( $classes ) ? self::REPORTED_CLASSES : $classes;
```
dkoo: **Suggestion:** parse_only_classes() silently falls back to the default set for an empty-but-present --only — `return empty( $classes ) ? self::REPORTED_CLASSES : $classes;` means `--only=`, `--only=,,` or — most realistically — `--only="$CLASSES"` with an unexpanded shell variable quietly runs the default gift+order-only report instead of what the operator asked for. That contradicts the documented contract ("an unrecognised class aborts rather than narrowing the report") — the same failure mode the strictness exists to prevent, arrived at from a different direction. Distinguishing "flag absent" (default) from "flag present but parses to nothing" (error) closes it; `resolve_plan_ids()` has the same shape but is less exposed since a non-empty non-numeric token already errors.

**Consider:** Treat an explicitly-passed but empty class selection as an error rather than falling back to the default set. The grant list this command produces is bare user IDs with nothing recording which classes it came from, and the consumer validates only that the list is non-empty, not that it is the list the operator meant to build.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+		// Memberships has been turned off, is a realistic way into exactly that.
+		if ( ! function_exists( 'wc_memberships' ) || ! \post_type_exists( 'wc_user_membership' ) ) {
+			WP_CLI::error( 'WooCommerce Memberships is not active, so membership statuses cannot be resolved and every membership would be audited as if it granted access. Aborting.' );
+		}
+
+		if ( ! function_exists( 'wcs_get_subscription' ) ) {
+			// Not fatal: a site without Subscriptions has no subscription links to
+			// resolve, and its memberships are exactly the order-only / comp cohorts
+			// this command still classifies correctly.
+			WP_CLI::warning( 'WooCommerce Subscriptions is not active — every membership is audited as if it had no subscription link.' );
```
dkoo: **Nit:** WCS-inactive warning misstates where those rows land — The warning says every membership is "audited as if it had no subscription link," but a membership carrying `_subscription_id` with Subscriptions inactive keeps a null customer and classifies as `subscription-missing` — whose label ("linked subscription no longer exists") is a statement of fact that is wrong on such a site. Naming `subscription-missing` as where those rows will land (or carrying a "subscriptions unavailable" fact into the label) keeps the report honest for the operator the warning is aimed at.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+		WP_CLI::line(
+			sprintf(
+				'%d membership(s) across %d distinct reader(s) reported (classes: %s).',
+				count( $rows ),
+				count( $flagged_user_ids ),
+				implode( ', ', $only_classes )
+			)
+		);
+		WP_CLI::line( 'These readers lose access at the flip. `member_own_access_subscriptions` is evidence, not a verdict: a subscription there only preserves access if the gates accept its product.' );
+		WP_CLI::line( 'Next: confirm buyer-vs-recipient intent with the publisher, then re-run with --format=ids for the signed-off list. Granting those readers $0 subscriptions needs a migrate-manual-members that can select members by user ID (--user-ids/--user-ids-file); run `wp help newspack migrate-manual-members` to check this build, and fall back to operator scripting against the list where it cannot.' );
```
dkoo: **Nit:** Closing guidance hedges about a migrate-manual-members capability this branch already ships — The footer tells the operator to "run `wp help newspack migrate-manual-members` to check this build, and fall back to operator scripting" — but `class-teams-migration.php` on this branch already declares both `--user-ids` and `--user-ids-file`, with examples. The hedge was right while #689 was outstanding; as merged it makes the handoff sound more uncertain than it is.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+		return [ 'class', 'plan_id', 'plan', 'membership_id', 'member_id', 'member_email', 'buyer_id', 'buyer_email', 'record', 'record_status', 'products', 'member_own_access_subscriptions' ];
+	}
+
+	/**
+	 * Classify a membership by how it is backed.
+	 *
+	 * Pure: everything it needs is in $facts, so the classification table can be
+	 * exercised without WooCommerce.
+	 *
+	 * @param array $facts {
```
dkoo, line 703: **Nit:** classify() facts docblock omits subscription_in_payment_recovery — The `@param array $facts` hash documents every key except `subscription_in_payment_recovery`, which `facts_grant_access()` reads and which decides gift vs gift-inactive for the on-hold-in-retry shape. Since the docblock is the contract for exercising the pure classifier without WooCommerce, the one fact that flips the headline classification deserves its `@type bool` line.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+			'member_id'                       => $member_id,
+			'member_email'                    => self::get_user_email( $member_id, $empty ),
+			// On an order-backed row the buyer is the holder unless the one-off was
+			// itself a gift; reporting it either way keeps the columns comparable.
+			'buyer_id'                        => $buyer_id,
+			'buyer_email'                     => $buyer_email,
+			'record'                          => $order_backed
+				? 'order ' . (int) $facts['order_id']
+				: 'subscription ' . (int) $facts['subscription_id'],
+			'record_status'                   => $order_backed
```
dkoo, line 1145: **Nit:** A gift kept by payment-recovery grace prints record_status "on-hold" with nothing explaining its inclusion — The row carries only the raw subscription status, so the operator sees `gift / on-hold` beside rows saying `gift / active`, with nothing explaining why an on-hold subscription is on a list of readers losing access — while the `gift-inactive` cohort also holds on-hold rows that were correctly excluded. Emitting `on-hold (in recovery)` or a separate boolean column would make the evidence self-explanatory in the CSV the publisher reads.

plugins/newspack-plugin/tests/unit-tests/cli/class-test-memberships-audit.php
```diff
@@ -0,0 +1,766 @@
...
+			'post_type'   => 'post',
+			'post_status' => 'publish',
+		];
+
+		// Half a second, so the pause dominates the query time rather than competing
+		// with it: differencing two timed runs would make this flake on a busy machine.
+		$pause = 0.5;
+
+		$started = microtime( true );
+		$walk->invoke( null, $args, function () {}, 0.0 );
```
dkoo: **Nit:** Pacing test asserts wall-clock bounds that can flake on a slow CI runner — `assertLessThan( $pause, $unpaced )` requires the unpaced two-batch walk over 505 posts to finish in under 500ms, and the paced bound leaves 500ms of slack around a 500ms sleep. Locally comfortable, but CI runs MySQL 5.7 under contention. If it ever flakes, injecting a sleeper callable (or asserting a recorded sleep count) pins the same wiring — pause between batches, none after the last, none at `--sleep=0` — without a timing dependency; the shrink-mid-run companion test is the model.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1102 @@
...
+	 * depends on whether the gates accept those products, which the
+	 * access-parity diff decides. A recurring donation is the common case of an
+	 * access-granting subscription that unlocks no content.
+	 *
+	 * @param int    $user_id The member.
+	 * @param string $empty   Placeholder for "none".
+	 *
+	 * @return string Space-separated descriptors, or the placeholder.
+	 */
+	private static function describe_own_access_subscriptions( $user_id, $empty ) {
```
dkoo: **Suggestion:** Consider a health signal in the evidence column for the case where a reader is reported as owning nothing: it is the one column an operator uses to take somebody off a list that grants free subscriptions, and a wrong dash and a right dash currently look identical.

dkoo (changes requested):
@adekbadek All testing steps pass as written, but Claude did surface a blocker and some suggestions, posted as inline comments. I'd say the "high" release risk drops once we address the blocker, since the command here only reports and doesn't change any data.

For any site where this command has already been run: the stored per-reader subscription list is repaired by regenerating WooCommerce Subscriptions' customer cache from its own cache tools, and it will not repair itself — a renewal or status change does not rewrite it. Worth naming in the PR so it does not have to be rediscovered from a support ticket.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Done in 74c6e07 — per-plan and all-plans breakdowns now go to STDERR in csv/json/ids, so the redirected stream stays parseable and the counts survive the export.

```
$ wp newspack audit-membership-subscriptions --format=csv > out.csv
── Plan 2424: "Audit Probe Plan" — 5 membership(s) with access ──
  member-owned               1  member owns the access-granting subscription
  gift                       1  gift — access-granting subscription owned by someone else ←
  order-only                 1  backed by a paid one-off order, no subscription ←
  subscription-missing       1  linked subscription no longer exists
  no-purchase-record         1  no purchase record (comp/legacy)
```

`out.csv` holds the header row and the two flagged rows. The progress bar stays STDOUT-only and off in those formats — it is redrawn in place and belongs to an interactive run.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Fixed in 74c6e07. Zero-ID rows are still excluded — a $0 subscription cannot be granted to user 0 — but the run now warns on STDERR, leaving STDOUT unchanged so a redirected list stays clean:

```
Warning: 1 flagged membership(s) have no member account and are not on the ID list —
an orphaned membership or a guest purchase. They appear in the table/CSV report and
need handling by hand.
```

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Fixed in 74c6e07. `--only=`, `--only=,,` and `--only="$CLASSES"` with the variable unset now abort:

```
Error: --only was passed but names no class. Drop it for the default report, or pass
one or more of: gift, order-only, order-only-unpaid, subscription-missing.
```

The parser can only draw the absent-vs-empty distinction if the call site stops defaulting the flag to `''`, so it defaults to `null`: `null` returns the default set, any string is a specific request. `resolve_plan_ids()` got the same treatment — `--plan-ids=,,` errors instead of falling back to every published plan.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Fixed in 74c6e07. The warning now names where those rows land:

> WooCommerce Subscriptions is not active — no subscription can be loaded, so a membership carrying a subscription link is classified from its order alone, or as subscription-missing when there is no usable order. Those rows say nothing about whether the subscription still exists.

And with Subscriptions inactive, `subscription-missing` prints as "linked subscription could not be loaded (WooCommerce Subscriptions is not active)".

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Dropped in 74c6e07 — `--user-ids` and `--user-ids-file` are on the branch, so the footer now hands over the command instead of a caveat:

> Next: confirm buyer-vs-recipient intent with the publisher, then re-run with --format=ids for the signed-off list and grant those readers $0 subscriptions with `wp newspack migrate-manual-members --user-ids-file=<path>`.

The class docblock carried the same hedge; that is gone too, and the PR description's discrepancy 4 now records the handoff as live rather than blocked on #689.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Added in 74c6e07, with the note that it is the fact that decides gift vs gift-inactive for the on-hold shape.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Fixed in 74c6e07 — `record_status` reads `on-hold (in payment recovery)` when the grace, not the status, is what keeps the row on the list. A status that grants access on its own is still printed bare.

Annotated the existing column rather than adding a boolean one: the CSV goes to a publisher deciding buyer-vs-recipient, and the answer belongs beside the status they are already reading.

adekbadek:
Named in the PR description, and the cause is fixed on the branch as of 74c6e07.

The batch walk registered its keyset `posts_where` filter globally, so every WP_Query triggered from the walk's callback inherited the seek clause — including WooCommerce Subscriptions' customer lookup on CPT-storage sites, whose truncated result is written to the reader's `_wcs_subscription_ids_cache` user meta. The filter is now scoped to the walk's own query, pinned by a test that fails on the previous code.

The remediation in the description is **WooCommerce → Status → Tools → Delete Customer Subscription Cache**, not "Generate": the generator's `get_items_to_update()` returns only users with no cache at all, so it walks past a poisoned entry, which is present and wrong rather than missing. The eraser clears the stored lists and they rebuild on demand.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1239 @@
...
+		WP_CLI::line(
+			sprintf(
+				'%d membership(s) across %d distinct reader(s) reported (classes: %s).',
+				count( $rows ),
+				count( $flagged_user_ids ),
+				implode( ', ', $only_classes )
+			)
+		);
+		WP_CLI::line( 'These readers lose access at the flip. `member_own_access_subscriptions` is evidence, not a verdict: a subscription there only preserves access if the gates accept its product.' );
+		WP_CLI::line( 'Next: confirm buyer-vs-recipient intent with the publisher, then re-run with --format=ids for the signed-off list and grant those readers $0 subscriptions with `wp newspack migrate-manual-members --user-ids-file=<path>`.' );
```
dkoo: **Suggestion:** The printed hand-off command fails as written — --product-id is required — The footer now prints `wp newspack migrate-manual-members --user-ids-file=<path>`, and the de-hedging is right (the flag exists on this branch and on main). But `--product-id=<id>` is a required option on that command — non-bracketed in its synopsis and enforced with `WP_CLI::error( 'Missing required option: --product-id=<id>.' )` — so an operator who copies the printed line gets a hard error at the end of a long audit run. Every example in `migrate-manual-members`' own docblock includes `--product-id`; this line should too, and it's worth a clause that the command dry-runs unless `--live` is passed.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1239 @@
...
+	 * This command only reads. Buyer-vs-recipient intent is a publisher
+	 * question, so it reports pairs as evidence and decides nothing: no
+	 * subscription is re-homed and no access is granted. `--format=ids` emits
+	 * the signed-off list for `migrate-manual-members --user-ids-file`, which
+	 * grants the replacement access.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--plan-ids=<ids>]
+	 * : Comma-delimited membership plan IDs to audit — the plans whose content the new gates cover. Defaults to every published plan (a plan in any other status is audited only when named here).
```
dkoo: **Nit:** --plan-ids synopsis doesn't document its new present-but-empty strictness — This round made `--plan-ids=` (present but naming nothing) a hard error, mirroring `--only` — and `--only`'s synopsis was updated to say so, but `--plan-ids` still reads only "Defaults to every published plan", which is now true just for the absent flag. `wp help` is where an operator checks this, and the two flags now behave identically while documenting themselves differently.

**Consider:** Call the existing bare-flag guard from the sibling command (or an equivalent argv scan for `--plan-ids` and `--only`) before any other work, so a flag WP-CLI stripped halts the run instead of resolving to the widened default the new strictness was written to prevent.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
```diff
@@ -0,0 +1,1239 @@
...
+		$flagged_user_ids = self::flagged_user_ids( $rows );
+
+		// A flagged membership whose holder has no account cannot be granted a $0
+		// subscription, so it is left off the ID list — but silently dropping it
+		// would let an operator hand the list on believing it covers the whole
+		// cohort. The warning goes to STDERR, so a redirected ids/CSV stream stays
+		// parseable.
+		$rows_without_member = self::count_rows_without_member( $rows );
+		if ( $rows_without_member ) {
+			WP_CLI::warning(
```
dkoo, line 471: **Nit:** The no-member warning's pointer is off for --format=ids runs — "They appear in the table/CSV report" is true of those formats, but in the ids format — where the warning matters most — the rows are never printed anywhere in that run, so the operator has to infer a re-run is needed. A phrasing like "re-run with `--format=table` or `csv` to see them" would close the loop. The count and the "need handling by hand" instruction are already there.

dkoo (approved):
Release risk: ELEVATED — the round-1 blocker is fixed and verified (regression test fails against the old implementation, a live probe shows nested queries receive no seek clause, and the full test group is green). Some inline suggestions and nits below. No further blockers.

Review provenance: 3 of 4 reviewer passes (lineup: deep, standards, wp-expert, codex); release risk: ELEVATED (degraded — skipped: codex).

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Fixed in bcc4387. The footer now prints `wp newspack migrate-manual-members --product-id=<id> --user-ids-file=<path>` and adds "That command dry-runs and writes nothing until --live is added." The command docblock's pointer at the hand-off carried the same short form, so it names both flags now too.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Both done in bcc4387.

`--plan-ids` synopsis now reads "Omit the flag to audit every published plan …; passing it with no ID is an error", matching `--only`.

For the guard I called the existing one rather than adding a second argv scan: `Teams_Migration::get_valueless_value_flags()` takes the flag list as a second argument (defaulting to `MANUAL_MEMBERS_VALUE_FLAGS`, so migrate-manual-members' call is unchanged), and the audit passes its own `VALUE_FLAGS` — `--plan-ids`, `--only`, `--sleep`, `--format` — as the first thing the command does, ahead of the Memberships check. The list is per-command because a bare `--user-ids` on an audit invocation must not be reported against a command that does not accept it; a test pins that alongside the bare `--plan-ids`/`--only` case.

`--sleep` and `--format` are in the list even though their stripped defaults are the safe ones: a bare `--sleep` means someone asked for gentler pacing and would silently get the default.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Reworded in bcc4387: "… Re-run with --format=table or --format=csv to see them; they need handling by hand." That reads correctly from an ids run and from the formats that do print the rows.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Already fixed — this comment was drafted against `17415d381` and the scoping landed in `74c6e07be` before the review was submitted, so it reads as outdated on the current diff.

The filter is now scoped to the walk's own query by a `WALK_QUERY_FLAG` query var, which the callback's nested queries don't carry, so the WCS customer lookup runs unconstrained and never caches a truncated list. `each_post_id_batch()` also unregisters the filter in a `finally`.

Your reproduction is pinned as a regression test: it runs a nested `WP_Query` inside the batch callback and asserts it still sees the lowest-ID post, which an unscoped clause hides on the very first batch.

Worth saying that the diagnosis was right about the mechanism and about which layer it lands in — the `_wcs_subscription_ids_cache` write is what makes it outlive the run, and that's now spelled out in the comment above the filter so the next person doesn't re-derive it.

plugins/newspack-plugin/tests/unit-tests/cli/class-test-memberships-audit.php
adekbadek: Fixed in 3c801f656c — the pause is recorded rather than taken.

`each_post_id_batch()` takes an optional sleeper callable; production leaves it null and calls `usleep()`. The test passes a recorder and asserts `[]` at `--sleep=0` and exactly `[ 0.5 ]` for the two-batch walk, which pins the same three properties (pause between batches, none after the last, none at zero) with no clock in the assertion.

Verified it still bites: making the walk pause after the final batch turns the recorded list into `[0.5, 0.5, 0.5]` and fails the test.

plugins/newspack-plugin/includes/cli/class-memberships-audit.php
adekbadek: Good catch, and it was worse than the dash being ambiguous.

`get_active_subscriptions_for_user()` returns `[]` — not an error — when `wcs_get_users_subscriptions()` is unavailable. On a site running Memberships without Subscriptions that means *every* member reports owning nothing, so the entire paid population reads as uncovered and goes to the grant step.

Fixed in 3c801f656c: answerability is decided once per run by `own_subscriptions_answerable()` (both the resolver class and the WCS function), and an unanswerable run prints `unknown` instead of the placeholder. The placeholder now only ever appears for a member the resolver actually answered about.

One case deliberately keeps the placeholder: a holder with no account. There is no user to own a subscription, and `member_id` / `member_email` already carry that on the row.

I dropped a first attempt that also flagged "resolver named a subscription this run couldn't load" — the resolver already drops unloadable IDs itself, so that branch was unreachable.

adekbadek:
Thanks @dkoo — review round addressed in 3c801f656c, replies in each thread.

The evidence-column suggestion from the review body turned out to be the load-bearing one: `get_active_subscriptions_for_user()` returns `[]` rather than an error when Subscriptions is unavailable, so a Memberships-without-Subscriptions site reported every member as owning nothing and would have sent its whole paid population to the grant step. That case now prints `unknown`.

The paging blocker was already closed by `74c6e07be` before the review was submitted — the comment anchors to `17415d381`, so it reads as outdated on the current diff. The filter is scoped by a query-var flag and the nested-lookup case has a regression test.

Your approval predates these commits; they're a nit and a suggestion plus their tests, so I've left it standing rather than re-requesting.

PR #771, opened 2026-07-30
chore(docker): keep memcached alive and fail open on cache delete

adekbadek:
### All Submissions:

Local dev only — the drop-in is copied into the container by `run.sh` and never ships to publishers.

In an isolated env, `update_post_meta()` followed by `get_post_meta()` in the same request returned the stale value while the DB held the new one. `wp_cache_delete()` and `clean_post_cache()` did not help. Two of fifteen running env containers were affected.

memcached had died, and nothing supervises it. With the backend unreachable, the drop-in treats its in-process array asymmetrically:

| method | runtime cache when memcached is down |
| --- | --- |
| `set()` | writes `$this->cache[$key]` unconditionally |
| `get()` | serves from `$this->cache[$key]` |
| `delete()` | `unset()` only if `false !== $result` — so never |

`Memcache::delete()` returns `false` when the server is unreachable, so the runtime entry survives and the cache degrades into one that can never be invalidated. Core invalidates post meta via delete, so `update_post_meta()` became a no-op against the cache. Anything writing and reading back in one request — CLI migrations, tests — got false results with no outward signal.

Changes:

1. **`delete()` and `replace()` always drop the runtime entry.** The one that matters: it turns a silent correctness bug into, at worst, a slower local site.
2. **Unreachable servers are logged once per process**, which also bounds `connection_errors` (written, never read, previously one entry per failed op).
3. **`bin/watchdog-memcached.sh`** — a 30s TCP probe that restarts memcached. Bind-mounted via `/var/scripts` rather than inline in `run.sh`, so later changes to it need no image rebuild. No `set -e`, so a failed restart can't end the loop.
4. **Container healthchecks** for the main, 8.2 and generated env compose files, plus a migration in `n env up` so the ~43 existing env files get one. Tuned to flag only a memcached the watchdog could not revive, with `start_period` covering a cold boot.
5. **Header note** recording that this vendored copy is patched, so a re-vendor doesn't silently revert it.

`run.sh` is baked into the image, so the watchdog needs one `./build-image.sh` to start; the drop-in applies on container restart.

1. On `main`, kill memcached in any env container:
   ```
   docker exec <container> sh -c '/etc/init.d/memcached stop; pkill -x memcached'
   ```
   Create a post, `update_post_meta()`, then `get_post_meta()` it back in one `wp eval-file`. Returns `''` while a direct SELECT returns the value.
2. With this branch's `bin/object-cache.php` in the container's `wp-content/`, memcached still down:
   ```
   immediately:   'probe_value'
   after delete:  'probe_value'
   direct db:     'probe_value'
   post title:    'probe-renamed'
   ```
3. `wp-content/debug.log` reports the outage once per process: `Memcached object cache: cannot reach 127.0.0.1:11211…`
4. Run `MEMCACHED_WATCHDOG_INTERVAL=2 /var/scripts/watchdog-memcached.sh`, then kill memcached three times with `pkill -x memcached`. It recovers each time and the loop survives.
5. `docker exec <container> bash -c 'echo > /dev/tcp/127.0.0.1/11211'` exits non-zero with memcached down, zero with it up.
6. `n env up <name>` on an env created before this branch inserts the healthcheck; re-running inserts nothing.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — no harness covers the container's runtime setup; verified manually per the steps above.

All six steps verified on a live container; both compose files and a migrated env file parse via `docker compose config`.

Unrelated to the memcached work, this branch also carries a CI fix: the PHPUnit job's `Install WordPress test suite` step piped `curl -s` straight into `tar` with no `pipefail`, no `-f` and no retries, so a failed WordPress.org download surfaced as tar's bare exit 2 and failed a matrix leg. It now retries, validates the version string it resolved from `api.wordpress.org`, and reports the download error itself.

### DIFF
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -312,12 +312,14 @@ ${worktree_volumes}      - ./envs/${env_name}/html:/var/www/html
       - APACHE_RUN_USER=\${USE_CUSTOM_APACHE_USER:-www-data}
     extra_hosts:
       - "host.docker.internal:host-gateway"
+    ## Probes memcached -- see docker-compose.yml for the rationale. Kept in step
+    ## with the healthcheck migration in env_up().
     healthcheck:
       test: ["CMD", "bash", "-c", "echo > /dev/tcp/127.0.0.1/11211"]
-      interval: 60s
+      interval: 30s
       timeout: 5s
       retries: 3
-      start_period: 30s
+      start_period: 180s
     networks:
       default: {}
       newspack_envs:
@@ -442,6 +444,26 @@ networks:
     external: true
 MIGRATE
             echo "Migrated $env_name: added shared network (domain: $domain)"
+        fi
+        # --- Migration: add the memcached healthcheck if missing ---
+        # Compose files are generated once at create time, so envs predating the
+        # healthcheck would never get it. Keep this block in step with the
+        # healthcheck in the generated YAML above.
+        if ! grep -q 'healthcheck:' "$compose_file"; then
+            awk '
+                { print }
+                /^      - "host\.docker\.internal:host-gateway"$/ && !inserted {
+                    print "    ## Probes memcached -- see docker-compose.yml for the rationale."
+                    print "    healthcheck:"
+                    print "      test: [\"CMD\", \"bash\", \"-c\", \"echo > /dev/tcp/127.0.0.1/11211\"]"
+                    print "      interval: 30s"
+                    print "      timeout: 5s"
+                    print "      retries: 3"
+                    print "      start_period: 180s"
+                    inserted = 1
+                }
+            ' "$compose_file" > "${compose_file}.tmp" && mv "${compose_file}.tmp" "$compose_file"
+            echo "Migrated $env_name: added memcached healthcheck"
         fi
         # Re-read domain after potential migration.
         domain=$(domain_for_env "$compose_file")
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -312,6 +312,12 @@ ${worktree_volumes}      - ./envs/${env_name}/html:/var/www/html
       - APACHE_RUN_USER=\${USE_CUSTOM_APACHE_USER:-www-data}
     extra_hosts:
       - "host.docker.internal:host-gateway"
+    healthcheck:
+      test: ["CMD", "bash", "-c", "echo > /dev/tcp/127.0.0.1/11211"]
+      interval: 60s
+      timeout: 5s
+      retries: 3
+      start_period: 30s
     networks:
       default: {}
       newspack_envs:
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -463,7 +463,15 @@ MIGRATE
                     inserted = 1
                 }
             ' "$compose_file" > "${compose_file}.tmp" && mv "${compose_file}.tmp" "$compose_file"
-            echo "Migrated $env_name: added memcached healthcheck"
+            # awk's insertion flag is invisible to the shell, so re-read the file:
+            # a compose file without the anchor line passes through unchanged, and
+            # reporting a migration that did not happen would send whoever debugs
+            # a stale env looking in the wrong place.
+            if grep -q 'healthcheck:' "$compose_file"; then
+                echo "Migrated $env_name: added memcached healthcheck"
+            else
+                echo "Warning: could not add the memcached healthcheck to $compose_file (no extra_hosts anchor). Recreate the env to pick it up." >&2
+            fi
         fi
         # Re-read domain after potential migration.
         domain=$(domain_for_env "$compose_file")
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -72,7 +72,6 @@ chown memcache:memcache /var/run/memcached || true
 /etc/init.d/memcached start
 
 # Supervise it -- see the script for why nothing else does.
-chmod +x /var/scripts/watchdog-memcached.sh || true
 /var/scripts/watchdog-memcached.sh &
 
 # Run apache in the foreground so the container keeps running
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -57,8 +57,31 @@ echo "Newspack Manager site is available at https://manager.com${WP_HOST_PORT}/"
 echo "Open http://localhost:8025 to see Mailhog inbox."
 echo
 
-# Start memcached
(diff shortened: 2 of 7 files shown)


bin/env.sh
```diff
@@ -437,6 +445,26 @@ networks:
...
+                    print "    ## Probes memcached -- see docker-compose.yml for the rationale."
+                    print "    healthcheck:"
+                    print "      test: [\"CMD\", \"bash\", \"-c\", \"echo > /dev/tcp/127.0.0.1/11211\"]"
+                    print "      interval: 30s"
+                    print "      timeout: 5s"
+                    print "      retries: 3"
+                    print "      start_period: 180s"
+                    inserted = 1
+                }
+            ' "$compose_file" > "${compose_file}.tmp" && mv "${compose_file}.tmp" "$compose_file"
```
dkoo, line 465: **Suggestion:** The 'Migrated ...' message fires unconditionally after the awk, but awk's inserted flag isn't visible to the shell. If the anchor (host-gateway) line is absent, awk passes the file through unchanged and the message is a lie — and grep -q healthcheck: still finds nothing next run, so it prints again. Nil impact for machine-generated compose files, but guarding the echo on an actual change would make it honest.

bin/watchdog-memcached.sh
```diff
@@ -0,0 +1,40 @@
...
+# Named watchdog-memcached rather than memcached-watchdog deliberately: the kernel
+# truncates process names to 15 characters, and `pkill memcached` matches that name
+# as a substring. The other ordering truncates to "memcached-watch", so anyone
+# reaching for pkill to clear a stuck memcached would take the supervisor with it,
+# precisely when it is needed most.
+
+INTERVAL="${MEMCACHED_WATCHDOG_INTERVAL:-30}"
+
+while true; do
+	sleep "$INTERVAL"
```
dkoo, line 32: **Suggestion:** Validate MEMCACHED_WATCHDOG_INTERVAL: with no set -e, a non-numeric/negative value makes sleep fail every iteration, and when memcached is healthy the loop spins at ~100% CPU. No code-execution path (only ever a quoted arg to sleep) — just a CPU spin from a bad env var. Validating a positive integer with a fallback to 30 would harden it.

bin/watchdog-memcached.sh
```diff
@@ -0,0 +1,40 @@
...
+		continue
+	fi
+
+	echo "[watchdog-memcached] memcached is not answering on 11211; restarting it."
+
+	# Kill by exact process name rather than using the init script's stop/restart:
+	# those match on a pid file memcached may have left stale after a crash, and
+	# container pids recycle fast enough that signalling it could hit an unrelated
+	# process. The start wrapper unlinks a stale pid file before forking.
+	pkill -x memcached >/dev/null 2>&1
```
dkoo, line 44: **Nit:** No pause between pkill -x memcached and /etc/init.d/memcached start: if the killed process hasn't released port 11211 yet, the restart can fail to bind. It self-heals on the next cycle, so not a defect, but a sleep 1 between the two would tighten it.

bin/run.sh
```diff
@@ -57,9 +57,24 @@ echo "Newspack Manager site is available at https://manager.com${WP_HOST_PORT}/"
...
+# pid file error on every start, which reads as a startup failure when triaging.
+# Create it to keep the log honest. Neither step is worth failing the container
+# for, hence the guards.
+mkdir -p /var/run/memcached || true
+chown memcache:memcache /var/run/memcached || true
+
 /etc/init.d/memcached start
 
+# Supervise it -- see the script for why nothing else does.
+chmod +x /var/scripts/watchdog-memcached.sh || true
```
dkoo: **Nit:** Redundant chmod: chmod +x /var/scripts/watchdog-memcached.sh is already covered by line 39's chmod +x /var/scripts/*.sh, and the file is committed 100755. Can go.

dkoo (approved):
Release risk: LOW. A well-reasoned, local-dev-only resilience fix with exemplary commenting. The object-cache change is correct — delete()/replace() always dropping the runtime mirror aligns the drop-in with core's in-memory WP_Object_Cache semantics and with healthy-production behavior, so it can't mask a production-only bug (the only path that differs from upstream is backend-unreachable, which in dev is flaky memcached). No blockers; a few minor polish notes inline.

bin/env.sh
adekbadek: Good catch — the message was unconditional. Fixed in 0ac898d by re-reading the file with grep after the awk, since awk's flag can't reach the shell: it prints the migration line only when the healthcheck actually landed, and otherwise warns on stderr that the anchor was missing and the env needs recreating. Silence would have sent whoever debugs a stale env looking in the wrong place.

bin/watchdog-memcached.sh
adekbadek: Validated in 0ac898d: the interval must match ^[0-9]+$ and be at least 1, otherwise it logs the bad value and falls back to 30. Checked against '', abc, 0, -5 and 45.

bin/watchdog-memcached.sh
adekbadek: Added the sleep 1 in 0ac898d, with a comment on why the wait is there so it doesn't look like a stray delay.

bin/run.sh
adekbadek: Removed in 0ac898d — the glob chmod above covers it.

adekbadek:
@dkoo Re-requesting review – alongside the three nits you flagged, the commits since your approval also rework the CI test-suite install step — pipefail, curl retries, and a hard failure when the version-check API returns no usable tag — which is beyond what you signed off on:
- chore(docker): harden memcached watchdog and env migration reporting
- ci: harden the WordPress test-suite download in the PHPUnit job

dkoo (approved):
Re-reviewed the follow-up — all four earlier notes are addressed cleanly: the migration reports honestly now (and warns when the anchor line is missing), the watchdog validates its interval and pauses before the restart, and the redundant chmod is gone.

Thanks too for the CI hardening. It's outside this PR's memcached scope, but it's a genuinely useful one to fold in — it targets the WordPress test-suite download flakiness that has failed unrelated PRs for us, and the pipefail + retry + version-check validation covers each failure mode rather than just papering over one. Glad to have it in.

No new concerns. Release risk: LOW.

PR #772, opened 2026-07-30
fix(access-control): complete gate access-rule option lists (NPPD-2135)

adekbadek:
### All Submissions:

The gate editor's Paid Access pickers could not offer every value a gate may legitimately hold.

Institutions were fetched at a fixed page size of 100, in both pickers and on the Institutions list screen. All three now fetch unbounded.

Subscription options listed parent products only, so a gate holding a *variation* ID had no option to render or re-select — Woo Memberships–migrated data carries those IDs, and the rule already honours them. Variations are now listed alongside their parent, so a gate can require one tier without requiring the others.

Saving a gate no longer builds any option list: sanitizing reads the registered rules rather than the resolved ones. That also stops a shop with no products from sending a valid ID list down the plain-string branch. The lists stay memoized per request for the screens that do render them.

The picker's "not listed" caution is now keyed on the rule slug, so it names only causes that rule can have.

Institution options stay published-only: an unpublished institution is never evaluated, so offering one would build a rule that silently does nothing.

The one-time purchase picker #756 adds has the same missing-variation defect, deliberately left for its own change — its evaluators are `wc_customer_bought_product()` rather than `WC_Subscription::has_product()`, so the fix needs its own evaluation testing.

Closes NPPD-2135.

1. Check out this branch. Enable `NEWSPACK_CONTENT_GATES`. Activate WooCommerce and WooCommerce Subscriptions.
2. Create 120 institutions under Audience > Access Control > Institutions.
3. Open the Institutions list. Confirm the table lists all 120.
4. Create a subscription product. Create a variable subscription product with two variations.
5. Open Audience > Access Control and edit a gate. Enable Paid Access.
6. Enable the Institutional access rule. Click its field.
7. Confirm the list shows all 120 institutions. Before this change it stopped at 100.
8. Enable the Active subscription rule. Click its field.
9. Confirm both variations are listed, each as `<name> (#<id>)`. Before this change only the parents appeared.
10. Select one variation. Save the gate. Reload the page.
11. Confirm the variation is still selected and shows its name.
12. Edit a post. Add a Group block. Open Access Control in the block sidebar.
13. Enable the Institutional access rule. Confirm the list shows all 120 institutions.

### Other information:

`OK (3315 tests, 9512 assertions)` and 924 JS tests, PHPCS clean against vipwpcs 3.1.0 as well as the pinned 3.0.1.

NPPD-2143 — the institutional rule fails open when a site has no institutions, because the picker degrades to free text and a string value grants access to everyone. Pre-existing; not addressed here.

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
@@ -458,7 +458,26 @@ class Access_Rules {
 	/**
 	 * Get subscriptions eligible for access rules.
 	 *
-	 * @return array Active subscription IDs.
+	 * Variations of a variable subscription are listed alongside their parent, because the
+	 * rule evaluates them: `WC_Subscription::has_product()` matches a line item's
+	 * `variation_id` as well as its `product_id`. Selecting the parent therefore grants
+	 * access to subscribers of any of its variations, while selecting a single variation
+	 * narrows the rule to that one — gating on the annual tier of a variable subscription
+	 * but not the monthly, say. Without the variations listed, that narrower rule can be
+	 * held by a gate (migrated data carries variation IDs) but never configured or read
+	 * back in the editor.
+	 *
+	 * Unpublished products are listed too — `wc_get_products()` defaults to draft, pending,
+	 * private and publish. Unlike institutions, whose options are published-only, a
+	 * subscription to a product the publisher has since drafted still evaluates: the line
+	 * item is what the rule matches, and it does not stop existing when the product is
+	 * unpublished. Hiding those products would make the readers still paying for them
+	 * ungateable.
+	 *
+	 * The list is unbounded and rebuilt per call. See NPPD-2138 for memoizing it — gate
+	 * saves rebuild it once per rule.
+	 *
+	 * @return array Array of [ 'label' => string, 'value' => int ].
 	 */
 	public static function get_subscription_products_options() {
 		if ( ! function_exists( 'wc_get_products' ) ) {
@@ -470,14 +489,100 @@ class Access_Rules {
 				'limit' => -1,
 			]
 		);
-		$options = [];
+		$variations_by_parent = self::get_subscription_variation_posts( $products );
+		$options              = [];
 		foreach ( $products as $product ) {
 			$options[] = [
 				'label' => $product->get_name(),
 				'value' => $product->get_id(),
 			];
+			foreach ( $variations_by_parent[ $product->get_id() ] ?? [] as $variation ) {
+				$options[] = [
+					'label' => self::get_variation_option_label( $product->get_name(), $variation ),
+					'value' => $variation->ID,
+				];
+			}
 		}
 		return $options;
+	}
+
+	/**
+	 * Fetch the variation posts of the given variable subscriptions, keyed by parent ID.
+	 *
+	 * The post row carries everything a label needs — WooCommerce stores a variation's
+	 * generated title in `post_title` and its attribute summary in `post_excerpt` — so this
+	 * reads the posts directly rather than hydrating a `WC_Product` per variation.
+	 * Hydrating would cost a full meta read, a parent re-read and two term lookups each,
+	 * and this runs on every admin page load that localises the access rules, including
+	 * every block editor load.
+	 *
+	 * Private variations are included alongside published ones: a reader can hold an active
+	 * subscription to a tier the publisher has since hidden, and the rule still has to be
+	 * able to name it.
+	 *
+	 * @param \WC_Product[] $products The subscription products to collect variations for.
+	 *
+	 * @return array<int, \WP_Post[]> Variation posts keyed by parent product ID.
+	 */
+	private static function get_subscription_variation_posts( $products ) {
+		$parent_ids = [];
+		foreach ( $products as $product ) {
+			if ( $product->is_type( 'variable-subscription' ) ) {
+				$parent_ids[] = $product->get_id();
+			}
+		}
+		if ( empty( $parent_ids ) ) {
+			return [];
+		}
+		$variations = \get_posts(
+			[
+				'post_type'              => 'product_variation',
+				'post_parent__in'        => $parent_ids,
+				'post_status'            => [ 'publish', 'private' ],
+				'posts_per_page'         => -1,
+				'orderby'                => [
+					'menu_order' => 'ASC',
+					'ID'         => 'ASC',
+				],
+				// Only the title, excerpt, ID and parent are read.
+				'update_post_meta_cache' => false,
+				'update_post_term_cache' => false,
+			]
+		);
+		$by_parent = [];
+		foreach ( $variations as $variation ) {
+			$by_parent[ $variation->post_parent ][] = $variation;
+		}
+		return $by_parent;
+	}
+
+	/**
+	 * Build the option label for a subscription variation.
+	 *
+	 * WooCommerce generates a variation's title as "Parent - Attribute" and names it that
+	 * way throughout its own admin, so that title is the label wherever it is usable. But it
+	 * drops the attribute suffix when the parent carries three or more attributes, or two or
+	 * more where an attribute name is multi-word — leaving the variation titled exactly like
+	 * its parent. Recover the attributes from the summary in that case, so the variation
+	 * still reads as the tier it is rather than as a second copy of its parent. The summary
+	 * spells out attribute names where the generated title omits them ("Term: Annual" rather
+	 * than "Annual"); matching the generated style would mean hydrating the variation, which
+	 * costs more than the fallback is worth.
+	 *
+	 * Labels need not be unique: the pickers render every option as `<name> (#<id>)` and
+	 * resolve a token by the ID it carries, so two options sharing a name stay distinct.
+	 *
+	 * @param string   $parent_name The variable subscription parent's name.
+	 * @param \WP_Post $variation   The variation post.
+	 *
+	 * @return string The option label.
+	 */
+	private static function get_variation_option_label( string $parent_name, \WP_Post $variation ): string {
+		$label = $variation->post_title;
+		if ( $label !== $parent_name ) {
+			return $label;
+		}
+		return '' !== $variation->post_excerpt ? sprintf( '%s - %s', $label, $variation->post_excerpt ) : $label;
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
@@ -24,6 +24,28 @@ class Access_Rules {
 	private static $rules = [];
 
 	/**
+	 * Request-scoped memo for the subscription product options.
+	 *
+	 * Building the list costs a full-catalog product query plus a variation query, and
+	 * `get_access_rules()` resolves every registered rule's options callback on every
+	 * call. It is public and several admin screens localize it, so a request that reaches
+	 * it more than once pays that cost once.
+	 *
+	 * @var array|null
+	 */
+	private static $subscription_products_options = null;
+
+	/**
+	 * Request-scoped memo for the one-time purchase product options.
+	 *
+	 * Same reasoning as {@see self::$subscription_products_options}, over a shop's whole
+	 * simple/variable catalog.
+	 *
+	 * @var array|null
+	 */
+	private static $one_time_purchase_products_options = null;
+
+	/**
 	 * Valid duration units for the one-time purchase rule.
 	 *
 	 * @var array
@@ -49,17 +71,6 @@ class Access_Rules {
 	 * @var array
 	 */
 	private static $evaluation_context = [];
-
-	/**
-	 * Request-scoped memo for the subscription product options.
-	 *
-	 * `sanitize_access_rule()` resolves every registered rule's options callback once per
-	 * rule in a gate's payload, so saving a six-rule gate would otherwise run the
-	 * full-catalog product query and its variation query six times over.
-	 *
-	 * @var array|null
-	 */
-	private static $subscription_products_options = null;
 
 	/**
 	 * Initialize hooks.
@@ -526,14 +537,17 @@ class Access_Rules {
 	}
 
 	/**
-	 * Flush the request-scoped subscription product options memo.
+	 * Flush the request-scoped product options memos.
 	 *
-	 * Primarily for tests; in production the memo is per-request by nature.
+	 * For tests; in production the memos are per-request by nature. A PHPUnit run is one
+	 * request for the whole suite, so `Newspack_Request_Memo_Reset` calls this before
+	 * every test to put the request boundary back where the memos assume it is.
 	 *
 	 * @return void
 	 */
-	public static function flush_subscription_products_options_memo() {
-		self::$subscription_products_options = null;
+	public static function flush_product_options_memos() {
+		self::$subscription_products_options      = null;
+		self::$one_time_purchase_products_options = null;
 	}
 
 	/**
@@ -591,14 +605,20 @@ class Access_Rules {
 	 * Build the option label for a subscription variation.
 	 *
 	 * WooCommerce generates a variation's title as "Parent - Attribute" and names it that
-	 * way throughout its own admin, so that title is the label wherever it is usable. But it
-	 * drops the attribute suffix when the parent carries three or more attributes, or two or
-	 * more where an attribute name is multi-word — leaving the variation titled exactly like
-	 * its parent. Recover the attributes from the summary in that case, so the variation
-	 * still reads as the tier it is rather than as a second copy of its parent. The summary
-	 * spells out attribute names where the generated title omits them ("Term: Annual" rather
-	 * than "Annual"); matching the generated style would mean hydrating the variation, which
-	 * costs more than the fallback is worth.
+	 * way throughout its own admin, so that title is the label wherever it is usable. It is
+	 * usable when it still opens with the parent's current name and adds something to it.
+	 * Two things break that. WooCommerce drops the attribute suffix when the parent carries
+	 * three or more attributes, or two or more where an attribute name is multi-word,
+	 * leaving the variation titled exactly like its parent. And the title is generated at
+	 * the variation's last save: `WC_Product_Variable_Data_Store_CPT::sync_variation_names()`
+	 * rewrites it when the parent is renamed through the CRUD path, but an importer or a
+	 * direct `wp_update_post()` leaves it on the old name.
+	 *
+	 * Either way the title stops telling a publisher which tier they are picking, so
+	 * recover the attributes from the summary and pair them with the parent's current name.
+	 * The summary spells out attribute names where the generated title omits them
+	 * ("Term: Annual" rather than "Annual"); matching the generated style would mean
+	 * hydrating the variation, which costs more than the fallback is worth.
 	 *
 	 * Labels need not be unique: the pickers render every option as `<name> (#<id>)` and
 	 * resolve a token by the ID it carries, so two options sharing a name stay distinct.
@@ -612,17 +632,15 @@ class Access_Rules {
 	 * @return string The option label.
 	 */
 	private static function get_variation_option_label( string $parent_name, \WP_Post $variation ): string {
-		$label = $variation->post_title;
-		if ( $label !== $parent_name ) {
-			return $label;
-		}
-		if ( '' === $variation->post_excerpt ) {
-			return $label;
+		$title             = $variation->post_title;
+		$names_this_parent = str_starts_with( $title, $parent_name ) && strlen( $title ) > strlen( $parent_name );
+		if ( $names_this_parent || '' === $variation->post_excerpt ) {
+			return $title;
 		}
 		return sprintf(
 			/* translators: 1: variable subscription name, e.g. "Membership". 2: the variation's attribute summary, e.g. "Term: Annual". */
 			__( '%1$s - %2$s', 'newspack-plugin' ),
-			$label,
+			$parent_name,
 			$variation->post_excerpt
 		);
 	}
@@ -705,18 +723,13 @@ class Access_Rules {
 	 * @return array Product options as label/value pairs.
 	 */
 	public static function get_one_time_purchase_products_options() {
-		// Request-scoped memo: the full-catalog query would otherwise run once per
-		// rule sanitized on a gate save (Content_Gate_API resolves all rule options
-		// for each rule in the payload).
-		//
 		// TODO (NPPD-2132): unlike the subscription and institution options (also
 		// full dumps, but inherently small), a shop's simple/variable catalog can be
 		// large, and this list is serialized into every block-editor payload. Move to
 		// a bounded/REST-searchable product picker; the memo only helps within a
 		// single request.
-		static $options = null;
-		if ( null !== $options ) {
-			return $options;
+		if ( null !== self::$one_time_purchase_products_options ) {
+			return self::$one_time_purchase_products_options;
 		}
 		if ( ! function_exists( 'wc_get_products' ) ) {
 			return [];
@@ -734,6 +747,7 @@ class Access_Rules {
 				'value' => $product->get_id(),
 			];
 		}
+		self::$one_time_purchase_products_options = $options;
 		return $options;
 	}
 
--- a/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-access-rules.php
@@ -51,6 +51,17 @@ class Access_Rules {
 	private static $evaluation_context = [];
 
 	/**
+	 * Request-scoped memo for the subscription product options.
+	 *
+	 * `sanitize_access_rule()` resolves every registered rule's options callback once per
+	 * rule in a gate's payload, so saving a six-rule gate would otherwise run the
+	 * full-catalog product query and its variation query six times over.
+	 *
+	 * @var array|null
+	 */
+	private static $subscription_products_options = null;
+
+	/**
 	 * Initialize hooks.
 	 */
 	public static function init() {
@@ -472,14 +483,21 @@ class Access_Rules {
 	 * subscription to a product the publisher has since drafted still evaluates: the line
 	 * item is what the rule matches, and it does not stop existing when the product is
 	 * unpublished. Hiding those products would make the readers still paying for them
-	 * ungateable.
+	 * ungateable. Variations narrow that set to publish and private, which is not a
+	 * different policy but the same one: WooCommerce reads a variable product's children
+	 * as `post_status IN ( 'publish', 'private' )`, so draft is not a state its own admin
+	 * produces for a variation, and nothing can have been bought in it.
 	 *
-	 * The list is unbounded and rebuilt per call. See NPPD-2138 for memoizing it — gate
-	 * saves rebuild it once per rule.
+	 * The result is memoized per request. The list itself is still unbounded and is
+	 * serialized into every editor payload; NPPD-2132 replaces it with a searchable
+	 * picker, which is what removes that cost rather than deferring it.
 	 *
 	 * @return array Array of [ 'label' => string, 'value' => int ].
 	 */
 	public static function get_subscription_products_options() {
+		if ( null !== self::$subscription_products_options ) {
+			return self::$subscription_products_options;
+		}
 		if ( ! function_exists( 'wc_get_products' ) ) {
 			return [];
 		}
@@ -503,7 +521,19 @@ class Access_Rules {
 				];
 			}
 		}
+		self::$subscription_products_options = $options;
 		return $options;
+	}
+
+	/**
+	 * Flush the request-scoped subscription product options memo.
+	 *
+	 * Primarily for tests; in production the memo is per-request by nature.
+	 *
+	 * @return void
+	 */
+	public static function flush_subscription_products_options_memo() {
+		self::$subscription_products_options = null;
 	}
 
 	/**
@@ -516,9 +546,10 @@ class Access_Rules {
 	 * and this runs on every admin page load that localises the access rules, including
 	 * every block editor load.
 	 *
-	 * Private variations are included alongside published ones: a reader can hold an active
-	 * subscription to a tier the publisher has since hidden, and the rule still has to be
-	 * able to name it.
+	 * Publish and private is the whole set WooCommerce itself reads a variable product's
+	 * children as, so it is every variation that can exist for a publisher to have sold.
+	 * Private earns its place: a reader can hold an active subscription to a tier the
+	 * publisher has since hidden, and the rule still has to be able to name it.
 	 *
 	 * @param \WC_Product[] $products The subscription products to collect variations for.
 	 *
@@ -539,7 +570,7 @@ class Access_Rules {
 				'post_type'              => 'product_variation',
 				'post_parent__in'        => $parent_ids,
 				'post_status'            => [ 'publish', 'private' ],
-				'posts_per_page'         => -1,
+				'posts_per_page'         => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Variations of the subscription products already fetched; config-scale.
 				'orderby'                => [
 					'menu_order' => 'ASC',
 					'ID'         => 'ASC',
@@ -571,6 +602,9 @@ class Access_Rules {
 	 *
 	 * Labels need not be unique: the pickers render every option as `<name> (#<id>)` and
 	 * resolve a token by the ID it carries, so two options sharing a name stay distinct.
+	 * That ` (#<id>)` suffix is a parsing contract and stays fixed, but the separator
+	 * joining a name to its attributes carries no such role, so it is translatable —
+	 * WooCommerce makes its own equivalent filterable for the same reason.
 	 *
 	 * @param string   $parent_name The variable subscription parent's name.
 	 * @param \WP_Post $variation   The variation post.
@@ -582,7 +616,15 @@ class Access_Rules {
 		if ( $label !== $parent_name ) {
 			return $label;
 		}
-		return '' !== $variation->post_excerpt ? sprintf( '%s - %s', $label, $variation->post_excerpt ) : $label;
+		if ( '' === $variation->post_excerpt ) {
+			return $label;
+		}
+		return sprintf(
+			/* translators: 1: variable subscription name, e.g. "Membership". 2: the variation's attribute summary, e.g. "Term: Annual". */
+			__( '%1$s - %2$s', 'newspack-plugin' ),
+			$label,
+			$variation->post_excerpt
+		);
 	}
 
 	/**
--- a/plugins/newspack-plugin/includes/content-gate/class-institution.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-institution.php
@@ -170,16 +170,26 @@ class Institution {
 	/**
 	 * Get institution options for the access rule multi-select.
 	 *
+	 * Published institutions only, matching `rebuild_cache()`: an institution that is not
+	 * published is never evaluated and so can never grant access, and offering one here
+	 * would let a publisher build a rule that silently does nothing. The institution editor
+	 * always saves with `publish`, so any other status comes from editing the post directly.
+	 *
 	 * @return array Array of [ 'label' => string, 'value' => int ].
 	 */
 	public static function get_options() {
 		$posts   = \get_posts(
 			[
-				'post_type'      => self::POST_TYPE,
-				'post_status'    => 'publish',
-				'posts_per_page' => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Institution CPT; config-scale.
-				'orderby'        => 'title',
-				'order'          => 'ASC',
+				'post_type'              => self::POST_TYPE,
+				'post_status'            => 'publish',
+				'posts_per_page'         => -1, // phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Institution CPT; config-scale.
+				'orderby'                => 'title',
+				'order'                  => 'ASC',
+				// Only the title and ID are read, and this runs on every admin page load
+				// that localises the access rules. `get_posts()` already forces
+				// `no_found_rows`, so there is no row count to suppress here.
+				'update_post_meta_cache' => false,
+				'update_post_term_cache' => false,
 			]
 		);
 		$options = [];
--- a/plugins/newspack-plugin/src/content-gate/access-rule-options.ts
+++ b/plugins/newspack-plugin/src/content-gate/access-rule-options.ts
@@ -30,6 +30,19 @@ import type { TokenItem } from '@wordpress/components/build-types/form-token-fie
 export type AccessRuleOption = { value: string | number; label: string };
 
 /**
+ * How many suggestions a picker renders when nothing has been typed.
+ *
+ * `FormTokenField` defaults to 100, which is a display cap rather than a fetch cap: it
+ * ranks by the typed query before slicing, so an option past the cap is still reachable
+ * by name or ID. But with `__experimentalExpandOnFocus` the unfiltered list is what a
+ * publisher sees on focus, and a site with more than 100 institutions read that as the
+ * rest not existing. This raises the cap far enough that browsing works at realistic list
+ * sizes while still bounding the DOM — the suggestion list is not virtualised, so it
+ * renders one node per suggestion. Past it, an option is reached by typing its name or ID.
+ */
+export const MAX_OPTION_SUGGESTIONS = 500;
+
+/**
  * The ID a token carries, e.g. `188250` in `Annual (#188250)`.
  */
 const TOKEN_ID_PATTERN = /\(#([^)]+)\)\s*$/;
--- a/plugins/newspack-plugin/src/content-gate/access-rule-options.ts
+++ b/plugins/newspack-plugin/src/content-gate/access-rule-options.ts
@@ -39,6 +39,9 @@ export type AccessRuleOption = { value: string | number; label: string };
  * rest not existing. This raises the cap far enough that browsing works at realistic list
  * sizes while still bounding the DOM — the suggestion list is not virtualised, so it
  * renders one node per suggestion. Past it, an option is reached by typing its name or ID.
+ *
+ * This moves the cliff rather than removing it. NPPD-2132 removes it, by making the picker
+ * query the server as the publisher types instead of rendering a whole list up front.
  */
 export const MAX_OPTION_SUGGESTIONS = 500;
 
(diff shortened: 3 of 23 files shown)


plugins/newspack-plugin/includes/content-gate/class-institution.php
```diff
@@ -129,16 +129,26 @@ public static function create( $name, $description = '', $rules = [] ) {
...
 		$posts   = \get_posts(
 			[
-				'post_type'      => self::POST_TYPE,
-				'post_status'    => 'publish',
-				'posts_per_page' => -1,
-				'orderby'        => 'title',
-				'order'          => 'ASC',
+				'post_type'              => self::POST_TYPE,
+				'post_status'            => 'publish',
+				'posts_per_page'         => -1,
```
jason10lee: **Blocker:** The rebase can drop a `phpcs:ignore` that only bites later — The base branch carries `// phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Institution CPT; config-scale.` on this line (added under NPPM-3055); this branch predates it, and realigning the array puts both edits on the same line — which is the reported conflict. I confirmed `origin/main` has the annotation on both this line and `rebuild_cache()`'s, and this branch has neither.

Losing it in the resolution is invisible today, because the root `composer.json` pins vipwpcs to `~3.0.1` on both branches. It becomes a hard error the moment NPPM-3055 lifts that pin, which the Dependabot comment explicitly plans for — and since CI resolves PHPCS fresh from the root, that can redden `main` rather than showing up on a PR.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
```diff
@@ -358,4 +359,160 @@ public function test_evaluate_rules_defaults_to_current_user_when_user_id_is_nul
```
jason10lee, line 371: **Blocker:** The rebase can silently delete either side's tests, and nothing catches it — This branch and its base both append test methods at the identical anchor — this PR's hunk is `@@ -358,4 +359,160 @@`, the base's is `@@ -358,4 +362,372 @@`. A resolution that takes one side drops the other's tests entirely, and PHPUnit passes on whatever survives, so nothing surfaces it.

Counted on the three trees: the merge-base has **11** test methods, this branch **15** (+4), the base tip **21** (+10). A correct merge should have **25**. Worth checking with `grep -c 'public function test_'` on the merged file before pushing — the same collision affects the `set_up()` hunk, where the base realigned lines 79-81 and this PR inserts the `$products_database` reset into them.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -329,7 +329,26 @@ function ( $rule ) {
 	/**
```
jason10lee, line 480: **Suggestion:** A per-request memo would pay for itself on the save path — `sanitize_access_rule()` calls `get_access_rules()`, which resolves every rule's options callback, once per rule in the payload — and only reads `! empty( $rule['options'] )` as a boolean, discarding the build. This PR adds a second unbounded query to that per-rule cost, so a six-rule gate now runs six unbounded `wc_get_products()` plus six unbounded variation queries on one save.

The deferral to NPPD-2138 is reasonable, but the base branch's sibling `get_one_time_purchase_products_options()` already added a `static` memo for this exact reason, so the precedent and the shape are both there.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -329,7 +329,26 @@ function ( $rule ) {
...
-	 * @return array Active subscription IDs.
+	 * Variations of a variable subscription are listed alongside their parent, because the
+	 * rule evaluates them: `WC_Subscription::has_product()` matches a line item's
+	 * `variation_id` as well as its `product_id`. Selecting the parent therefore grants
+	 * access to subscribers of any of its variations, while selecting a single variation
+	 * narrows the rule to that one — gating on the annual tier of a variable subscription
+	 * but not the monthly, say. Without the variations listed, that narrower rule can be
+	 * held by a gate (migrated data carries variation IDs) but never configured or read
+	 * back in the editor.
+	 *
```
jason10lee, line 491: **Suggestion:** The two status policies read as contradictory — Parents inherit `wc_get_products()`'s defaults (`draft, pending, private, publish`), justified as 'a subscription to a product the publisher has since drafted still evaluates'. Variations then exclude `draft`, justified in the test docblock as 'a draft variation has never been purchasable' — which is the same argument the parent case rejects.

The stronger reason is absent from both: WooCommerce itself queries variations as `post_status IN ('publish','private')`, so `draft` isn't a state its UI produces. Putting that in the production docblock would make the asymmetry self-evident rather than something a reader has to reconcile.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -329,7 +329,26 @@ function ( $rule ) {
...
+	 * item is what the rule matches, and it does not stop existing when the product is
+	 * unpublished. Hiding those products would make the readers still paying for them
+	 * ungateable.
+	 *
+	 * The list is unbounded and rebuilt per call. See NPPD-2138 for memoizing it — gate
+	 * saves rebuild it once per rule.
+	 *
+	 * @return array Array of [ 'label' => string, 'value' => int ].
 	 */
 	public static function get_subscription_products_options() {
```
jason10lee, line 508: **Suggestion:** After the rebase this lands beside a second picker with the same defect — The variation fix covers the `subscription` rule. The base branch adds `get_one_time_purchase_products_options()`, registered as `one_time_purchase`, which queries `[ 'simple', 'variable' ]` with no variation expansion — and its evaluators match variation ids just as `has_product()` does (`wc_customer_bought_product()` selects on `_product_id` and `_variation_id`).

So once rebased, the gate editor shows two adjacent product pickers with opposite behaviour, and for one of them a stored variation id still renders `(product unavailable)`. That label invites the publisher to remove an entry that is doing its job, and removing the last entry from a rule is not a neutral edit — I've left the evaluation detail on NPPD-2135. Extending the helper to both rules, or saying in the PR why it's deferred, would close the loop this PR opens.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -341,16 +360,102 @@ public static function get_subscription_products_options() {
...
+			}
+		}
+		if ( empty( $parent_ids ) ) {
+			return [];
+		}
+		$variations = \get_posts(
+			[
+				'post_type'              => 'product_variation',
+				'post_parent__in'        => $parent_ids,
+				'post_status'            => [ 'publish', 'private' ],
```
jason10lee, line 586: **Suggestion:** A deleted variation is still a reachable case of the defect this fixes — Restricting to `[ 'publish', 'private' ]` covers hidden tiers, but a *deleted* variation is left. Retiring a tier by deleting the variation row — rather than unchecking Enabled, which writes `private` — is ordinary WooCommerce housekeeping, and `has_product()` still matches the orphaned `variation_id` on the existing line item. The publisher then sees `(product unavailable)` for a value that is still doing its job, which is the same removal path as above.

This may be an acceptable residual — it can't be resolved from a product query alone. Worth reconciling with the docblock above, which argues the opposite direction for unpublished parents ('it does not stop existing when the product is unpublished').

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -341,16 +360,102 @@ public static function get_subscription_products_options() {
...
+				'posts_per_page'         => -1,
```
jason10lee: **Suggestion:** The new unbounded query needs its own `phpcs:ignore` — This `posts_per_page => -1` carries no `WordPressVIPMinimum.Performance.NoPaging` annotation, unlike the sibling unbounded queries the base branch annotated under NPPM-3055. Unlike the two above, the rebase won't fix this one — it's new here.

I reproduced the future failure by linting this branch with the plugin's own vendor (vipwpcs 3.1.0, which is what `npm run lint:php` uses): three errors, all `posts_per_page => -1`, this line among them. CI is green because it lints from the root at 3.0.1. Adding the annotation with a scale rationale now costs nothing and matches the convention the base just established.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -29,6 +29,19 @@ import type { TokenItem } from '@wordpress/components/build-types/form-token-fie
...
+ *
+ * `FormTokenField` defaults to 100, which is a display cap rather than a fetch cap: it
+ * ranks by the typed query before slicing, so an option past the cap is still reachable
+ * by name or ID. But with `__experimentalExpandOnFocus` the unfiltered list is what a
+ * publisher sees on focus, and a site with more than 100 institutions read that as the
+ * rest not existing. This raises the cap far enough that browsing works at realistic list
+ * sizes while still bounding the DOM — the suggestion list is not virtualised, so it
+ * renders one node per suggestion. Past it, an option is reached by typing its name or ID.
+ */
+export const MAX_OPTION_SUGGESTIONS = 500;
```
jason10lee, line 46: **Suggestion:** The headline fix is the least-tested part — `per_page=-1` genuinely fixes the truncation — I verified the REST controller rejects it (`per_page=-1` and `per_page=500` both return HTTP 400, `per_page=100` returns exactly 100 of 105 seeded institutions), so the fix depends on apiFetch's `fetchAllMiddleware` walking `Link: rel="next"`, exactly as the comment says.

What's untested is that same headline behaviour: no test asserts more than 100 institutions come back, none asserts the picker receives the raised `maxSuggestions`, and none pins the new query flags. `access-rule-control.test.jsx` already has the scaffolding to make the JS half cheap. Also worth a pointer to NPPD-2132 from here — 500 moves the cliff rather than removing it, and a searchable picker is the thing that does.

plugins/newspack-plugin/tests/mocks/wc-mocks.php
```diff
@@ -359,6 +362,62 @@ function wc_create_mock_product( $data = [] ) {
...
+ * caller silently getting an unfiltered result is a test that passes for the wrong reason.
+ *
+ * Arguments real WooCommerce discards are the exception — neither a `WC_Product_Query`
+ * default var nor a mapped meta key, so ignoring one is what production does too.
+ *
+ * @param array $args Query args.
+ * @return WC_Product[]|int[] Matching products, or their IDs with `return => 'ids'`.
+ * @throws InvalidArgumentException When passed a filtering argument the mock does not implement.
+ */
+function wc_get_products( $args = [] ) {
```
jason10lee, line 655: **Suggestion:** The mocks can't catch the contract the docblock leans on — The new mock lines are careful and fail loudly on unimplemented args, which is the right instinct. Two gaps worth knowing about. `has_product()` is mocked as `in_array( $product_id, $this->products, true )` over a flat array with no line items, so parent-vs-variation matching — the claim the whole parent/child listing rests on — is structurally untestable here. I checked the real WCS 8.0.0 source and the docblock is correct (`has_product()` compares both `product_id` and `variation_id` directly; it does not go through `wcs_get_canonical_product_id()`), so this is an untested truth rather than a bug.

Second, with no `status` arg the mock applies no status filter, while real `wc_get_products()` defaults to `[ draft, pending, private, publish ]` — which is exactly what the new docblock relies on, and nothing covers it.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -341,16 +360,102 @@ public static function get_subscription_products_options() {
...
+	 * @param \WP_Post $variation   The variation post.
+	 *
+	 * @return string The option label.
+	 */
+	private static function get_variation_option_label( string $parent_name, \WP_Post $variation ): string {
+		$label = $variation->post_title;
+		if ( $label !== $parent_name ) {
+			return $label;
+		}
+		return '' !== $variation->post_excerpt ? sprintf( '%s - %s', $label, $variation->post_excerpt ) : $label;
```
jason10lee: **Nit:** The variation-label separator could be translatable — `sprintf( '%s - %s', ... )` hardcodes ` - ` in publisher-facing text. The base PR's rationale for keeping token structure untranslated covers the ` (#<id>)` suffix specifically, because it's parsed back out; this separator has no such role, and WooCommerce makes its equivalent filterable. Either make it translatable or note that it's deliberately fixed for consistency with the token contract.

jason10lee (changes requested):
Release risk: HIGH — two blockers, both in the rebase this PR still needs rather than in the code it adds. Seven non-blocking suggestions and one nit, as well.  GitHub additionally reports the branch as conflicting.

The code itself holds up well, and I checked its load-bearing claims against real source rather than taking the docblocks' word. WooCommerce Subscriptions 8.0.0's `has_product()` does match both a line item's product id and its variation id, so listing parents alongside variations is sound. The REST controller does reject `per_page=-1`, so the unbounded fetch genuinely depends on apiFetch's middleware, exactly as the comment says. With 105 institutions seeded, the server list returns all of them.

What needs care is the merge. Both this branch and its base append tests at the same anchor, and taking either side drops the other's silently — PHPUnit passes on whatever survives. The merge-base has 11 test methods, this branch 15, the base 21; a correct resolution has 25, which is one grep to confirm. The same collision sits on a `phpcs:ignore` annotation the base added and this branch predates; losing it is invisible until NPPM-3055 lifts the vipwpcs pin, and then it reddens main rather than a PR.

One scope note worth raising with the author: the base branch has since added a second product picker for one-time purchases with the identical variation gap, so after the rebase these land side by side with opposite behaviour. The mislabel that produces is worth closing rather than carrying, because it invites the publisher to remove an entry that is doing its job — evaluation detail is on NPPD-2135.

plugins/newspack-plugin/includes/content-gate/class-institution.php
adekbadek: Added the annotation on this side too, in 029f3eb — both `get_options()` and `rebuild_cache()` now carry `// phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Institution CPT; config-scale.`, matching the base's wording exactly. It can't be lost in the resolution now: whichever side wins the line, the annotation is on it.

Confirmed with the plugin's own vendor (vipwpcs 3.1.0): `--ignore-annotations` reports the sniff on both lines, and a normal run reports neither.

`class-content-gate.php:1708` is the same situation but the branch doesn't touch that file, so the rebase brings the base's annotation there with no conflict. Left alone.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Memoized in 029f3eb, in the same shape as `get_one_time_purchase_products_options()` — a private static, populated on first build, with `flush_subscription_products_options_memo()` alongside it so tests can rebuild. A six-rule save now runs one `wc_get_products()` and one variation query instead of six of each.

The static lives on the class rather than inside the method so the flush can reach it; that's what let me pin the behaviour in `test_get_subscription_products_options_is_memoized_per_request`, which fails without the memo.

Dropped the NPPD-2138 deferral from the docblock and pointed it at NPPD-2132 instead, since that's the ticket that actually removes the cost — the list is still unbounded and still serialized into every editor payload, and only a searchable picker fixes that.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Right — the two justifications didn't reconcile, and the reason you name is the real one. Rewrote both docblocks in 029f3eb to lead with it: WooCommerce reads a variable product's children as `post_status IN ( 'publish', 'private' )`, so draft isn't a state its own admin produces for a variation and nothing can have been bought in one. The variation set isn't a stricter policy than the parent set, it's the same policy applied to what can actually exist.

The helper's docblock now says publish-and-private is the whole set rather than an inclusion of private, with the hidden-tier case as why private earns its place instead of as the whole argument.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Annotated in 029f3eb: `// phpcs:ignore WordPressVIPMinimum.Performance.NoPaging -- Variations of the subscription products already fetched; config-scale.`

Reproduced your run first — the plugin's own vendor is vipwpcs 3.1.0, and `--ignore-annotations` against the root ruleset flags exactly this line plus the two institution ones. With the annotations in place all three are silent, and the only remaining hit in `includes/content-gate/` is `class-content-gate.php:1708`, which the base already annotates and this branch doesn't touch.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Fair — the headline had no test at all. Two added to `access-rule-control.test.jsx` in 029f3eb, over 105 mocked institutions:

- one pins the request path verbatim, so `per_page=-1` and the `orderby`/`status`/`_fields` flags can't drift from `Institution::get_options()`, which seeds the picker before the fetch lands;
- one focuses the field and asserts all 105 suggestions render, with the 105th present by name and ID.

Both fail on the pre-change values: setting the path back to `per_page=100` fails the first, dropping `MAX_OPTION_SUGGESTIONS` back to 100 fails the second. The second needs a real `input.focus()` rather than a synthetic focus event — `__experimentalExpandOnFocus` only expands when the input is the active element.

Added the NPPD-2132 pointer to the constant's docblock, saying plainly that 500 moves the cliff and a server-side search is what removes it.

plugins/newspack-plugin/tests/mocks/wc-mocks.php
adekbadek: The status gap is fixed in 029f3eb. With no `status` arg the mock now filters on `[ draft, pending, private, publish ]` — `WC_Product_Query`'s own default, named as a constant so it reads as the contract rather than as a magic list — instead of skipping the filter. `test_get_subscription_products_options_includes_unpublished_products` covers it from both sides: a draft subscription is listed, a trashed one is not. It fails against the old unfiltered mock, which returned the trashed product too.

The `has_product()` gap I've left. Making it real means giving the mock subscription line items with `product_id` and `variation_id`, which is a change to a mock several suites share, and the behaviour it would prove is WCS's rather than ours. Worth doing when something here starts depending on the distinction at evaluation time — the one-time purchase picker is the likely trigger, since that one does need its evaluators exercised.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Made it translatable in 029f3eb — `__( '%1$s - %2$s', 'newspack-plugin' )` with a translators comment giving both placeholders a concrete example. You're right that the token-structure argument doesn't reach it: nothing parses this separator back out, so it's ordinary publisher-facing text. Noted the distinction in the docblock so the next reader can see why the ` (#<id>)` suffix stays fixed while this doesn't.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
adekbadek: Counts re-taken after 029f3eb, since the target moved: merge-base 11, this branch 17 (+6 — the two new ones are the memo and unpublished-product cases), base tip 21 (+10). A correct merge is **27** methods, not 25. Same `grep -c 'public function test_'` on the merged file.

The `set_up()` collision is the same hunk as before, now with one more line in it: the `$products_database` reset and the memo flush both go in alongside the base's realignment of 79-81.

The rebase itself is still to do, so I haven't been able to run that check on a merged tree.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: x

adekbadek: Please ignore the two stray replies just above — a misfired command, no content intended. The substantive reply on this thread is the one about the `NoPaging` annotation.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Taking the second option — documented deferral rather than extending the helper — and worth being explicit that it's the lesser one.

The blocker is mechanical: `get_one_time_purchase_products_options()` isn't on this branch, it arrives with the rebase, so there's nothing here to extend. Beyond that the two rules don't share an evaluator. `WC_Subscription::has_product()` compares `variation_id` directly, which is what makes the subscription fix a pure options change with no evaluation risk. `wc_customer_bought_product()` selects on `_product_id` and `_variation_id` off the lookup table, so listing variations there needs its own evaluation testing rather than riding on this one's — and none of the mocks reach that today.

The PR description now names it: same defect, deliberately its own change, tracked on NPPD-2135. Happy to fold it in here instead if you'd rather the two pickers never ship differing.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
adekbadek: Merged in a71c5712f, reconciled by hand and counted. Three of the five conflicts were the same shared-trailing-context split you describe; git had cut each side mid-method, so taking either would have dropped the other's.

Counts moved since your read — this branch is at +6 now, not +4 — so the target was 27, not 25: 11 at the merge base, +6 here, +10 on the base branch. The merged file has 27 `public function test_`, and the symmetric difference between the merged method names and the union of both sides' is empty, so nothing was dropped and nothing invented. `set_up()` came through with both changes: the `$products_database` reset and the memo flush.

`wc-mocks.php` is the case worth flagging, because it did **not** conflict. Both sides had added `WC_Product::get_status()` and a `wc_get_products()` stand-in, and the auto-merge kept all four definitions — a fatal on the first test run rather than a silent loss. Kept the `wc_get_products()` that models `is_type()` (so the `subscription_variation` alias resolves), `limit` and `return => ids`, and throws on an argument it does not implement; every caller in the tree passes only `type`, `status` and `limit`.

One real defect surfaced behind the conflicts: the audit test compares the CLI's live-product set against the picker, and the picker's options are memoized per request while a PHPUnit run is one request — so in a full run it read a previous class's products. It passed alone and failed in the suite. It now flushes the memo in `set_up()`, beside the reset of `$products_database` that was already there.

Full suite: 3312 tests, 9505 assertions, 6 pre-existing skips. JS: 921. PHPCS clean.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Accepted as a residual and named in the docblock, in a71c5712f.

It cannot be resolved from a product query, and it is not particular to variations — a hard-deleted parent product leaves the same live-but-unnameable ID on a line item. What covers both is the "not listed" rendering and the caution attached to it, which exists to stop a publisher reading such a value as dead configuration and removing it.

The docblock now says so where it argues the status policy, so the paragraph no longer reads as if publish+private accounted for every case.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -470,16 +510,126 @@ public static function get_subscription_products_options() {
...
 				'value' => $product->get_id(),
 			];
+			foreach ( $variations_by_parent[ $product->get_id() ] ?? [] as $variation ) {
+				$options[] = [
+					'label' => self::get_variation_option_label( $product->get_name(), $variation ),
+					'value' => $variation->ID,
+				];
+			}
 		}
+		self::$subscription_products_options = $options;
```
jason10lee, line 535: **Suggestion:** `Content_Gate_API::sanitize_access_rule()` calls `Access_Rules::get_access_rules()`, which `array_map`s over the registered rules resolving *every* rule's options callback — and it is called once per rule per group. A six-rule gate save therefore builds all three option lists six times.

It never reads a resolved option, though. We checked every key it touches: `sanitize_callback`, `is_boolean`, and `! empty( $rule['options'] )` as a boolean. The options branch sanitizes the value off the request, not the option list. All three keys are identical on the unresolved `get_registered_rules()`.

So the memo removes one of the three lists, and `Institution::get_options()` is not memoized — it still runs six unbounded queries on that save, and institutions are the list this ticket just gave permission to grow. Switching the call at `class-content-gate-api.php:307` to `get_registered_rules()` takes all three to zero on the save path and should be behaviour-identical.

Not asking for it here if you'd rather not widen the diff. But the memo's own docblock names that call site as its motivation, so it's worth a line saying why the memo was chosen over the hoist.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/audit-subscription-products.php
```diff
@@ -54,6 +54,9 @@ public function set_up() {
 		$products_database      = [];
 		$wcs_mock_ignore_offset = false;
 		$wcs_mock_query_log     = [];
+		// The picker's options are memoized per request, and a test run is one request,
+		// so a previous class's products would otherwise stand in for this one's.
```
jason10lee: **Suggestion:** `test_live_product_set_matches_the_gate_picker` asserts `$picker_ids === $cli_ids`, and its docblock says it exists so that a change made to the picker and not to the CLI "makes the sets diverge and this test fail". This PR is that divergence — the picker now emits variation IDs, `get_live_subscription_products()` still emits parents only — and the test is green.

It passes because the fixture registers a `variable-subscription` mock with no `product_variation` posts under it, so the new branch contributes nothing. On any store with a real variable subscription the asserted equality is false.

A guard that stays green through the exact change it was written for is worse than no guard, because the next person to widen the picker reads it as confirmation. The behavioural blast radius looks small — the CLI set only feeds `guess_products_by_name()`, and repairs onto a variation are refused anyway — so narrowing the assertion to the parent subset and saying why seems better than widening the CLI.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -470,16 +510,126 @@ public static function get_subscription_products_options() {
...
+		$options              = [];
 		foreach ( $products as $product ) {
 			$options[] = [
 				'label' => $product->get_name(),
 				'value' => $product->get_id(),
 			];
+			foreach ( $variations_by_parent[ $product->get_id() ] ?? [] as $variation ) {
+				$options[] = [
+					'label' => self::get_variation_option_label( $product->get_name(), $variation ),
+					'value' => $variation->ID,
```
jason10lee, line 531: **Suggestion:** `class-woocommerce-subscriptions.php:905` states that the picker "offers `subscription` / `variable-subscription` parents, never variations", and the `SELECTABLE_PRODUCT_STATUSES` docblock at `:26` says a product outside the allowlist "is not gate-selectable, so Access Control can never reference it". A variation is `subscription_variation`, outside that allowlist, and is now selectable.

The classifier's behaviour is still correct — it keys on the line item's parent `product_id`, and a variation implies a live parent — so this is the rationale drifting, not the logic. But those two blocks are that file's stated source of truth for why the constants are what they are.

`class-membership-gates-migration.php:948` drops variations "because gates reference parent products only", which was the only stated justification for it diverging from the premium-newsletters command. That one probably wants its own follow-up rather than a change here.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -458,9 +469,38 @@ function ( $rule ) {
 	/**
 	 * Get subscriptions eligible for access rules.
 	 *
-	 * @return array Active subscription IDs.
+	 * Variations of a variable subscription are listed alongside their parent, because the
```
jason10lee, line 483: **Suggestion:** `class-premium-newsletters-migration.php:1003` warns, during a live migration, that "the gate editor's product picker lists parent products only, so they are not shown there — and re-saving that field in the editor drops them."

Both halves are now false: variations are offered, and they survive a re-save. The advice that follows — leave it alone unless you mean to change what the gate grants — is still reasonable, but it is justified by a hazard this PR removed, so an operator either acts on a stale reason or later finds the tool misled them.

The docblock at `:910` states the same thing as a trade-off the command accepted. That trade-off is now paid off, and saying so is the useful edit.

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
```diff
@@ -29,6 +29,22 @@ import type { TokenItem } from '@wordpress/components/build-types/form-token-fie
...
 
+/**
+ * How many suggestions a picker renders when nothing has been typed.
+ *
+ * `FormTokenField` defaults to 100, which is a display cap rather than a fetch cap: it
+ * ranks by the typed query before slicing, so an option past the cap is still reachable
+ * by name or ID. But with `__experimentalExpandOnFocus` the unfiltered list is what a
+ * publisher sees on focus, and a site with more than 100 institutions read that as the
+ * rest not existing. This raises the cap far enough that browsing works at realistic list
+ * sizes while still bounding the DOM — the suggestion list is not virtualised, so it
```
jason10lee, line 40: **Suggestion:** `getUnlistedAccessRuleValuesNotice()` (at `:253`, outside this diff) reads: *"Entries marked "not listed" are not in this list — a product variation, or an item that was deleted or unpublished."*

After this change a product variation **is** in the list for the subscription rule, so the notice leads with the one cause it just eliminated. The string is shared and takes no rule slug, so it stays accurate for `one_time_purchase` — whose picker still lists parents only — and is wrong for `subscription`.

This matters more than the usual copy nit because it is a `__()` msgid: changing it later orphans every existing translation, so the moment to get it right is the change that invalidates it. Either drop the variation clause, or key the wording on the slug the way `MISSING_OPTION_LABELS` already does.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
```diff
@@ -370,6 +374,233 @@ public function test_evaluate_rules_defaults_to_current_user_when_user_id_is_nul
 		wp_set_current_user( 0 );
 	}
 
+	/**
+	 * Create a real `product_variation` post, the way WooCommerce stores one: the generated
+	 * title in `post_title` and the attribute summary in `post_excerpt`.
+	 *
```
jason10lee: **Suggestion:** All six new tests exercise option construction. None covers the evaluation the construction exists to enable.

Every subscription fixture here stages coverage through `create_subscription( [ 'products' => [ … ] ] )`, which the mock's `has_product()` answers from its flat `in_array( $product_id, $this->products )` fallback. So the item-walk branch that distinguishes `product_id` from `variation_id` — the behaviour the whole feature rests on, and what the new docblock asserts — is never exercised.

A test staging a line item with `product_id = parent, variation_id = V`, asserting `has_active_subscription( $user, [ $V ] )` true and `[ $sibling_V ]` false, would pin "selecting one variation narrows the rule". The mock already models this, and the one-time-purchase siblings do stage `variation_id`, so it's fixture choice rather than a mock limitation.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
```diff
@@ -78,9 +78,13 @@ public static function set_up_before_class() {
...
 
-		// Reset the subscriptions database.
-		global $subscriptions_database;
+		// Reset the subscriptions and products databases.
+		global $subscriptions_database, $products_database;
 		$subscriptions_database = [];
+		$products_database      = [];
+
+		// The subscription options are memoized per request, and a test run is one request.
+		Access_Rules::flush_subscription_products_options_memo();
```
jason10lee: **Nit:** The flush is called from the two classes that currently assert on the options, but the memo is process-wide and the fixtures are per-class. Any class that saves a gate populates it implicitly, since `sanitize_access_rule()` resolves every rule's callback — and several other content-gate classes go through that path while resetting `$products_database` per test.

Nothing asserts on the option list from those classes today, so the suite is green. The first future assertion elsewhere reads a previous class's catalogue, and fails for a reason invisible in its own file. Flushing from the shared bootstrap or a base `tear_down` would make it structural rather than something each new class has to remember.

Worth noting you already hit this exact shape once — the audit test reading a previous class's products — and fixed that instance.

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
```diff
@@ -470,16 +510,126 @@ public static function get_subscription_products_options() {
...
+	 * That ` (#<id>)` suffix is a parsing contract and stays fixed, but the separator
+	 * joining a name to its attributes carries no such role, so it is translatable —
+	 * WooCommerce makes its own equivalent filterable for the same reason.
+	 *
+	 * @param string   $parent_name The variable subscription parent's name.
+	 * @param \WP_Post $variation   The variation post.
+	 *
+	 * @return string The option label.
+	 */
+	private static function get_variation_option_label( string $parent_name, \WP_Post $variation ): string {
```
jason10lee, line 634: **Nit:** `get_variation_option_label()` falls back to the attribute summary only when `post_title` equals the current parent name exactly. Reading the post row rather than hydrating is the right call, but the title was generated at the variation's last save while `$parent_name` is the parent's name now.

WooCommerce covers the common case — `sync_variation_names()` rewrites variation titles on a CRUD-path rename — so this only bites when a parent is renamed outside the CRUD path, by an importer or a direct `wp_update_post`. There the equality stops matching and variations with three or more attributes go back to being indistinguishable, which is the defect this PR set out to fix.

Keying the fallback on the title not *containing* the parent name, or preferring the summary whenever one exists, would make it robust.

jason10lee (approved):
Approved:  every thread resolved and no further changes requested (though see below for suggestions and nits).

We checked each of the ten threads against the code rather than reading the replies, and they hold. We've also confirmed your updated target count of 27.

One correction to our own earlier framing: `status=publish` is a no-op today rather than a fix because core's collection default for `status` is already `publish`. Pinning it against a future filter is worth doing; the measurable win in that change is `_fields`.

The suggestions below share one cause: listing variations is right, and it leaves four sibling call sites arguing from a premise it removes.

Release risk: ELEVATED. Nothing blocks — clearing the earlier request for changes.

<details>
<summary>What we verified (for reviewers and agents)</summary>

Isolated env on branch head `a71c5712f`, newspack-plugin rebuilt; worktree confirmed identical to the pushed ref.

- **Merge losslessness, two independent ways.** 17 methods pre-merge on this branch, 21 on the base, 27 on the merged head, no duplicate method names. Separately, diffing the merge commit against each parent across the contested files shows only whitespace realignment on one side and the four intentionally-replaced hunks on the other.
- **Institution parity, on 105 published + 3 drafts.** `Institution::get_options()` returns 105, drafts excluded. REST before: 100 items, 135,101 bytes. After `status=publish&_fields=id,title`: 100 items, 5,901 bytes. First five titles identical between the localised and fetched lists.
- **`status=publish` is currently a no-op.** With the parameter omitted the route returns the same 100 items and no drafts; `status=draft` returns the 3 drafts and `status=any` returns both, so the parameter works — core's default for it is simply `publish` already.
- **The save path reads three keys.** `sanitize_access_rule()` reads `sanitize_callback`, `is_boolean`, and `! empty( $rule['options'] )` as a boolean only; the options branch sanitizes the request value, never a resolved option. `get_access_rules()` resolves every rule's callback and is called once per rule per group.
- **Resolutions spot-checked.** Memo static and flush helper present in the same shape as the one-time-purchase memo; `WC_PRODUCT_QUERY_DEFAULT_STATUSES` present in the mock; three `NoPaging` annotations across the two files; `MANY_INSTITUTIONS` of 105 with the request path asserted verbatim; the variation-label separator translatable with a translators comment.

</details>

Review provenance: 2 of 3 reviewer passes (lineup: deep, wp-expert); release risk: ELEVATED (degraded — skipped: codex).

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Both halves rewritten. The runtime warning at :1000 now says the variation IDs are what the plan granted — narrower than the parent, which would also admit sibling holders — and that the editor lists them, so they survive a re-save. The "leave it alone" advice stays, now resting on what the restriction means rather than on a hazard that no longer exists.

The :906 docblock loses the accepted trade-off for the same reason and states the current behaviour instead.

03f85d391

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Took the first option. The title is now used only when it still opens with the parent's current name *and* adds something to it (`str_starts_with()` plus a length check); otherwise the label is the parent's current name paired with the attribute summary.

Preferring the summary whenever one exists would also be robust, but it would discard WooCommerce's generated title in the common case, and that title is what the variation is called throughout WC's own admin — worth keeping for recognition where it is still true.

`test_get_subscription_products_options_names_variations_of_a_renamed_parent` covers the importer case with titles generated under an old parent name, including the three-or-more-attributes shape where the stale title is the old name alone. The docblock now names `sync_variation_names()` and says why the CRUD path is not the whole story.

One residual case this does not catch: renaming a parent outside the CRUD path to a string that is a prefix of its old name leaves the stale title passing the check. That needs a non-CRUD rename *and* the new name being a prefix of the old, which seemed narrow enough not to give up the generated title for.

03f85d391

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Took the hoist. `sanitize_access_rule()` now reads `get_registered_rules()`, so the save path builds no option list at all — all three go to zero, including the unmemoized institution query.

Checking it turned up a second reason to prefer it: resolving here made `! empty( $rule['options'] )` depend on what the shop currently holds, so on an empty catalog the subscription rule read as not options-backed and its ID list went down the plain-string branch. `test_sanitize_access_rule_keeps_product_ids_with_an_empty_catalog` pins that.

The memo stays, on its own merits rather than the save path — `get_access_rules()` is public and more than one admin screen localises it. Its docblock now says that instead.

03f85d391

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/audit-subscription-products.php
adekbadek: Narrowed the assertion rather than widening the CLI, for the reason you give — the set feeds `guess_products_by_name()` and a repair may not point a line item at a variation, since `WC_Order_Item_Product::set_product_id()` rejects a `product_variation` post outright.

The fixture now creates a real `product_variation` post under the variable subscription, so the divergence is present rather than assumed away, and the test filters variations out of the picker set before comparing. Two assertions keep the exclusion from going quiet: the variation must be in the picker set, and must not be in the CLI set. Renamed to `test_live_product_set_matches_the_gate_pickers_parent_products` so the narrower contract is legible from the name.

03f85d391

plugins/newspack-plugin/includes/content-gate/class-access-rules.php
adekbadek: Corrected all three, in the file rather than as a follow-up.

`SELECTABLE_PRODUCT_STATUSES` and `SELECTABLE_PRODUCT_TYPES` now say they describe the picker's *parent* products, and the status docblock says why variations are deliberately not modelled: the picker lists them only alongside a parent it already lists, so the parent still decides whether a line item is reachable at all. The type docblock names the reason a variation stays out even though a gate can now reference one — `set_product_id()` rejects it.

`classify_subscription_product_link()` at :908 and `get_live_subscription_products()` are reworded to match, the latter saying explicitly that the picker's variations are left out because this set feeds repair name-matching.

`class-membership-gates-migration.php` changed here too, since the stated justification was false the moment the picker changed: dropping variations is now a conservative default rather than a limitation, and the operator warning points at the editor as the place to add them back. Keeping them is NPPD-2135's follow-up.

03f85d391

plugins/newspack-plugin/src/content-gate/access-rule-options.ts
adekbadek: Keyed it on the slug, the way `MISSING_OPTION_LABELS` is, rather than dropping the clause — a single string cannot be right for both pickers, and you are right that the moment to split it is the change that invalidates it.

`UNLISTED_VALUES_NOTICES` holds `subscription` (a product or variation that was deleted, or a product that is no longer a subscription) and `institution` (deleted or unpublished). A slug-agnostic default covers rules registered from elsewhere, since `register_rule()` is public and such a rule need not hold products at all. Both call sites pass their slug, and `access-rule-options.test.js` pins that a rule's notice names no cause its picker cannot produce.

One thing I found and left alone: `one_time_purchase` is the rule the variation clause was genuinely right for — its picker lists parents only and `customer_bought_product_after()` matches a line item's `variation_id` — but it short-circuits to `OneTimePurchaseRuleControl` on both surfaces, which renders no unlisted-values caution at all. So its tokens read "(product not listed)" with nothing explaining that removing one widens the gate. Giving it the notice is not a one-liner: the control's token field is injected, and core `FormTokenField` has no `description` prop, so the two surfaces cannot share the wizard's mechanism. That is pre-existing and outside what this PR touches, so it wants its own change alongside the one-time-purchase variation work already tracked on NPPD-2135. Happy to pull it in here if you would rather it not wait.

03f85d391

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
adekbadek: Added `test_a_variation_rule_admits_only_that_variation`, staged exactly as you describe: no `products` shorthand, a single `WC_Order_Item_Product` carrying `product_id = parent` and `variation_id = V`. It asserts the purchased variation grants, a sibling variation does not, and the parent still admits either — so it pins both halves of what listing variations is for.

You were right that it was fixture choice, not a mock limitation; the flat `in_array()` fallback was answering every existing subscription assertion and hiding the item walk entirely.

03f85d391

plugins/newspack-plugin/tests/unit-tests/content-gate/class-access-rules.php
adekbadek: Made it structural. A PHPUnit `BeforeTestHook` extension, `Newspack_Request_Memo_Reset`, flushes the request-scoped memos before every test, registered in `phpunit.xml` and required from the bootstrap (PHPUnit resolves extensions after the bootstrap runs, and it is not autoloaded). The per-class flushes in both test classes are gone, so a new class inherits the guarantee instead of having to remember it.

It is load-bearing rather than belt-and-braces: with the extension removed the suite fails six tests, five in `Newspack_Test_Access_Rules` and the audit parity test — the exact shape you predicted, each failing for a reason invisible in its own file.

`flush_subscription_products_options_memo()` became `flush_product_options_memos()` while doing this, since the one-time purchase options are memoized on the same terms and needed the same treatment.

03f85d391

adekbadek:
Pinned it in 2e0b5a1. The institution fetch now sends `status=publish`, and the comment says outright that it changes nothing today — it is there so a `rest_np_institution_collection_params` filter moving core's default cannot drift the fetched list away from `Institution::get_options()`, which is `post_status => 'publish'`. A test in `access-rule-option-sources.test.js` asserts the parameter is on the request, with the parity reason rather than the parameter as the thing it pins. Agreed that `_fields` is the measurable part of that change.

PR #777, opened 2026-07-31
fix(newsletters): keep flag-off public newsletters on the theme template

chickenn00dle:
### All Submissions:

**Public newsletters render in the theme's standard single template again when the WooCommerce renderer flag is off.**

Since #564 the WC email-editor package boots on every request, and booting it opted the newsletter CPT into the package unconditionally. The package keys a front-end `single_template` takeover (`Email_Editor::load_email_preview_template`) off that opt-in, so a newsletter set for **both email and web** was served through the package's email-preview template instead of the theme's single template. On a flag-off (default) site the public newsletter therefore rendered as a bare email — no theme header, footer or sidebar.

> **Correction (thanks @dkoo):** an earlier draft of this description also claimed email-only blocks were shown on the web. That half doesn't reproduce — `remove_visibility_hidden_block()` is a `render_block` filter that still runs under the package's template, so email-only blocks stayed hidden even on the broken path. The regression is the theme-wrapper bypass only.

What changed:

* **The CPT opt-in is now gated on the renderer flag.** `Editor_Bootstrap::add_post_type()` opts the newsletter CPT into the WC email editor only when `Feature_Flag::is_enabled()`. With the flag off, the package's front-end takeover is inert and public newsletters keep their legacy (MJML) behavior — theme template, `remove_visibility_hidden_block()` hiding email-only blocks.
* The flag is read lazily (on `init`), so option, constant, and filter toggles are all honored.
* Corrected the now-inaccurate "boots unconditionally / behaviorally inert" comments in the main plugin file and the email-renderers README.

**Behavior & regression notes:** the boot itself stays unconditional (one boot path; the WC-engine tests call `Editor_Bootstrap::init()` directly regardless of flag). When the flag is **on**, nothing changes — the CPT is opted in exactly as before, and the front-end takeover it leaves open for public newsletters is unchanged (tracked separately in NPPD-2150). When **off**, the boot still registers site-wide machinery, but it is either additive or post-type-gated (a `sent` post status, the preview REST routes, an `email-contents` pattern category, a KSES allow-list widening) and none of it touches the front-end template path. The `init:11` CPT re-assertion is now flag-gated too, so it no longer runs as duplicate work on flag-off requests.

Regression path (flag off — the default):

1. With `newspack-newsletters` active and the WC renderer flag off (default — `wp option get newspack_newsletters_use_woo_renderer` is empty), create a newsletter, mark it **public** (email + web), add one **email-only** block and one **web-only** block, and publish.
2. Visit the newsletter's public URL. **Expected:** it renders inside the theme's normal single template (header/footer/sidebar), the email-only block is **hidden**, the web-only block is **shown**. Before this fix it rendered as a bare email with no theme wrapper (email-only blocks were already hidden by the `render_block` filter, so that's not part of the regression).

Flag-on regression (no change):

3. `wp option update newspack_newsletters_use_woo_renderer 1`. Confirm the WC editor and rendering still work exactly as before.

Automated:

4. `n test-php` in `newspack-newsletters` — 699 pass. `tests/email-renderers/test-bootstrap-flag-gate.php` covers the flag-gated opt-in and the `single_template` resolution directly (both fail without the gate); `test-editor-bootstrap.php` exercises the flag-gated `init:11` re-assertion.

### Other information:

---

## For AI

**Root cause.** `newspack-newsletters.php` calls `Editor_Bootstrap::init()` unconditionally. That boots the vendored `woocommerce/email-editor` package, whose `Engine\Email_Editor::initialize()` (`vendor/woocommerce/email-editor/src/Engine/class-email-editor.php:123`) registers `add_filter( 'single_template', 'load_email_preview_template' )`. `load_email_preview_template()` (same file, ~line 364) returns the package's `single-email-post-template.php` and renders the body via `Send_Preview_Email::render_html()` — full email HTML — for any **publicly viewable** post whose type is in `woocommerce_email_editor_post_types`. `Editor_Bootstrap::add_post_type()` put the newsletter CPT in that list unconditionally, so flag-off public newsletters were captured. The prior inline comment asserting the boot was "behaviorally inert" when off was wrong — it accounted for the additive `block_type_metadata_settings` filter but not the `single_template`/CPT-opt-in front-end path.

**Why gate the opt-in, not the boot.** Gating the call-site boot behind the flag was the first instinct but breaks the WC-engine test suite (~56 tests): those tests call `Editor_Bootstrap::init()` directly with the flag off and rely on the plugin-load boot having registered the package's `init`-hooked `initialize()` in time; a late `init()` from the test body runs after WP's `init` has fired, so nothing renders. The opt-in (`add_post_type`) is the narrowest seam that the front-end takeover actually keys off, and reading the flag there is lazy — honored for the option, the constant, and any filter registered before `init:10` (a later filter half-engages, since the package reads the post-type list once at `init:10`). The `init:11` CPT re-assertion is now gated on the same flag (`reassert_cpt_when_enabled()`): with the flag off the package registers nothing at `init:10`, so re-asserting would be pure duplicate work (and could undo a `remove_post_type_support()` between the two priorities); with the flag on it still restores Newspack's canonical args over the package's email defaults.

**Verification.** Full `newspack-newsletters` PHPUnit suite green (699 / 0 fail); the `single_template` and opt-in tests both fail with the gate removed and pass with it; PHPCS clean against the monorepo ruleset; and confirmed live on a classic-theme site with the flag off — the public newsletter's front end carries the theme's `site-header`/`site-footer`/`main-content` markers, the email-only block hidden, the web-only block present, and no package email-template markers.

**Follow-ups:**
- The **flag-on** front-end takeover for public newsletters (the package template runs no loop, so the `the_post`-hooked content gate never engages — full-text exposure for a web-restricted newsletter) predates this PR and is **not** addressed here. Tracked in NPPD-2150.
- This regression is also present on `main` (#564 is in both). This PR targets `alpha`; `main` needs the same fix (or the next `main`→`alpha` merge could reintroduce it).


plugins/newspack-newsletters/tests/email-renderers/test-bootstrap-flag-gate.php
```diff
@@ -0,0 +1,108 @@
...
+			\Newspack_Newsletters::remove_visibility_hidden_block( 'WEB ONLY', $web_block ),
+			'Web-only blocks must render on the web front end.'
+		);
+		$this->assertSame(
+			'UNMARKED',
+			\Newspack_Newsletters::remove_visibility_hidden_block( 'UNMARKED', $plain_block ),
+			'Unmarked blocks must render on the web front end.'
+		);
+
+		$post = null;
```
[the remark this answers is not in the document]
chickenn00dle: Good catch — addressed in 9533574. Now saves the previous global `$post` and restores it after the assertions (instead of nulling it), and adds the `WordPress.WP.GlobalVariablesOverride.Prohibited` phpcs:ignore on the override lines to match the sibling tests in this directory.

plugins/newspack-newsletters/includes/email-renderers/class-editor-bootstrap.php
```diff
@@ -130,6 +144,9 @@ public static function reset_theme_json_cache() {
 	 * @return array Modified list.
 	 */
 	public static function add_post_type( $post_types ) {
+		if ( ! Feature_Flag::is_enabled() ) {
```
dkoo, line 166: **Suggestion:** The takeover this closes for flag-off sites stays open for flag-on sites — This guard fixes the default configuration, which is the right call and the urgent half. Worth naming the other half explicitly somewhere, though, because the flag is the rollout switch: with it on, the CPT is opted in, nothing in the plugin removes the package's `single_template` filter, and a public newsletter still renders as bare email HTML on the front end.

I checked this in a running env rather than inferring it — the flag-on render is byte-identical (same md5) to the pre-fix broken render, 6,366 bytes with no theme chrome. Going further, `single-email-post-template.php` calls `get_post()` and echoes; it runs no loop and no `wp_head()`. A `the_post` probe fired on the flag-off path and never fired on the flag-on path, so `Content_Gate::restrict_post` — which hangs off `the_post` — cannot engage under that template.

None of this is introduced here; it arrived with #564 and this PR strictly improves on it. But since the README section now reads as "front-end takeover: solved", a sentence naming the residual flag-on case would keep the next reader from trusting it too far.

plugins/newspack-newsletters/includes/email-renderers/README.md
```diff
@@ -65,13 +65,13 @@ Constants: `ENGINE_WC`, `ENGINE_MJML`, `RENDERER_META`.
 ### `Editor_Bootstrap` — `class-editor-bootstrap.php`
 Boots the package and wires it to the newsletters CPT. `init()` is idempotent and bails when the package is absent. It:
 
-1. Boots the package container.
-2. Opts the CPT into the editor (and re-asserts Newspack's canonical CPT args at `init:11`, since the package re-registers post types at `init:10`).
+1. Boots the package container (unconditionally — see below).
+2. Opts the CPT into the editor **only when `Feature_Flag::is_enabled()`** (and re-asserts Newspack's canonical CPT args at `init:11`, since the package re-registers post types at `init:10`).
```
dkoo: **Suggestion:** The `init:11` CPT re-assertion could share the same gate — This line still describes the `init:11` re-assertion unconditionally, but its stated reason — countering the package re-registering the CPT at `init:10` — now only applies when the flag is on. With the flag off the package's `register_email_post_types()` iterates an empty list, so `add_action( 'init', [ Newspack_Newsletters::class, 'register_cpt' ], 11 )` in `class-editor-bootstrap.php:70` is a pure duplicate of the `init:10` registration already hooked in `class-newspack-newsletters.php:77`.

It looks harmless — `add_filter`/`add_action` dedupe by callback identity, `register_taxonomy_for_object_type()` guards with `in_array`, permastructs are keyed by name — but it is not free or perfectly inert on what is currently every site: ~30 translation lookups and two option reads per request, `register_post_type_args` and `registered_post_type` fired a second time for any third-party listener, and every `$supports` entry re-added, which would silently undo a `remove_post_type_support()` performed between priorities 10 and 11.

Wrapping that callback in the same lazy flag read would keep both halves of the gate symmetric and make this documentation line true again.

plugins/newspack-newsletters/newspack-newsletters.php
```diff
@@ -126,19 +126,14 @@ function () {
...
-// `woocommerce_email_editor_post_types` filter added in Editor_Bootstrap::init()),
-// which is the only reason the init:11 CPT re-assertion exists — gating the boot
-// would remove the opt-in and the re-assertion together, not orphan one of them.
+// The boot itself is unconditional (one boot path for both flag states), but the
+// newsletters CPT is opted into the package only when the renderer flag is on — see
+// Editor_Bootstrap::add_post_type(). That opt-in is what the package's front-end
+// `single_template` takeover (load_email_preview_template) keys off, so gating it
+// keeps a flag-off site's *public* newsletters rendering in the theme's standard
+// single template (legacy/MJML behavior) rather than the package's email-preview
+// template. The boot still registers only additive, unused-until-render machinery
```
dkoo: **Suggestion:** "Only additive, unused-until-render machinery" understates what a flag-off boot leaves behind — Since this PR is rewriting exactly the comment that got the previous audit wrong, it seems worth getting the replacement precise — a confidently-worded characterization is what a future reader will trust instead of re-checking.

With the flag off, `Email_Editor::initialize()` still registers a site-wide `sent` post status, three `woocommerce-email-editor/v1` REST routes including `send_preview_email`, an `email-contents` block-pattern category, a `post_types` field on every `wp_template` REST response, and — via `Integrations\Core\Initializer` — a `safe_style_css` filter that permanently widens the KSES inline-style allow-list with `display` and three `mso-*` properties for all content on the site. `Editor_Bootstrap::register_template()` also does a `file_get_contents()` of the wrapping template on every request that fires `init`.

None of these are new here and none are affected by the gate. Something closer to "registers site-wide machinery that is either additive or post-type-gated, plus a `sent` post status, a preview REST route and a KSES allow-list widening" would survive contact with the source. The same sentence appears at `README.md:74`.

plugins/newspack-newsletters/tests/email-renderers/test-bootstrap-flag-gate.php
```diff
@@ -0,0 +1,110 @@
...
+ * Regression (#564): the package boots unconditionally, and its opt-in used to be
+ * unconditional too. Once the CPT is opted in, the package's `single_template`
+ * filter (Email_Editor::load_email_preview_template) takes over the front end of a
+ * *public* newsletter (one set for both email and web) — serving the package's
+ * email-preview template (full email HTML, email-only blocks shown, no theme
+ * wrapper) instead of the theme's standard single template. Gating the opt-in on
+ * `Feature_Flag::is_enabled()` (in Editor_Bootstrap::add_post_type) closes that
+ * front-end path when the flag is off.
+ */
+class Test_Bootstrap_Flag_Gate extends WP_UnitTestCase {
```
dkoo, line 23: **Suggestion:** A sibling test is now vacuous under the new default — Flagging this here since it is a consequence of this change rather than a problem with the new file: `test_canonical_cpt_args_remain_authoritative` in `tests/email-renderers/test-editor-bootstrap.php:46` guards the `init:11` re-assertion against the package re-registering the CPT with email defaults. With the flag off — the suite default — the package no longer registers the CPT at all, so its `public === true` assertion now passes for the wrong reason.

Either running that assertion with `newspack_newsletters_use_woo_renderer` forced on, or leaving a comment there noting the guard is now flag-conditional, would stop someone later reading a green test as proof the re-assertion still works.

plugins/newspack-newsletters/tests/email-renderers/test-bootstrap-flag-gate.php
```diff
@@ -0,0 +1,110 @@
...
+	 * On the web front end (no active email render), `remove_visibility_hidden_block()`
+	 * hides `email`-only blocks and keeps `web`-only and unmarked blocks. This is the
+	 * behavior a flag-off public newsletter must keep — when the opt-in leaked, the
+	 * package's email template bypassed this path and showed email-only blocks.
+	 */
+	public function test_email_only_blocks_hidden_on_web_front_end() {
+		$post_id = self::factory()->post->create(
+			[
+				'post_type'   => \Newspack_Newsletters::NEWSPACK_NEWSLETTERS_CPT,
+				'post_status' => 'publish',
```
dkoo, line 117: **Suggestion:** The regression that shipped still has no test — `test_cpt_opted_in_only_when_flag_on` does guard the gate — remove the guard and its flag-off assertion fails. But `test_email_only_blocks_hidden_on_web_front_end` is a characterization test of `remove_visibility_hidden_block()`, which this diff does not touch; it passes identically on `origin/alpha` with the regression present. So nothing in the suite asserts the thing that actually broke, and a future package version that keys its takeover off something other than `woocommerce_email_editor_post_types` would land green.

The direct assertion is cheap, and I verified both halves by hand against the running env: with a newsletter as global `$post`, `apply_filters( 'single_template', '/theme/single.php' )` returns `/theme/single.php` unchanged with the flag off, and returns the package's `single-email-post-template.php` with it on. `load_email_preview_template()` only registers a nested closure and returns a path — `render_html()` is not invoked at filter time — so the flag-on branch stays cheap.

Worth keeping the existing block-visibility test either way: `remove_visibility_hidden_block()` had no coverage anywhere before this PR.

plugins/newspack-newsletters/includes/email-renderers/class-editor-bootstrap.php
```diff
@@ -121,7 +121,21 @@ public static function reset_theme_json_cache() {
...
+	 * this list the package's `single_template` filter (Email_Editor::load_email_preview_template)
+	 * serves a *public* newsletter (one set for both email and web) through the package's
+	 * email-preview template — full email HTML, email-only blocks shown, no theme wrapper —
+	 * instead of the theme's standard single template. Gating the opt-in on the flag keeps a
+	 * flag-off site on the legacy (MJML) behavior, where public newsletters render normally in
+	 * the theme and `Newspack_Newsletters::remove_visibility_hidden_block()` hides email-only
+	 * blocks. The editor takeover in Newspack_Newsletters_Editor is separately flag-gated; this
+	 * closes the front-end path it never covered.
+	 *
+	 * The flag is read here (lazily, when the filter fires on `init`) rather than at boot so
```
dkoo: **Nit:** "Regardless of when it is registered" is doing slightly more work than the code supports — Two small imprecisions, both harmless in practice. First, the filter does not only fire on `init` — `get_post_types()` re-applies it at `single_template` and `preview_post_link` time and from `extend_email_post_api()`, so the flag is read several times per request. Second, two registration-side consumers read the list once at `init:10` and then cache it: `register_email_post_types()` and `Templates::initialize()`, which stashes it in a private property. So a `newspack_newsletters_use_woo_renderer` filter registered after `init:10` yields a half-on state — `Renderer_Controller::active_engine()` reports `wc` while the CPT never received the email args. "Any flag source registered before `init:10`" would be the accurate claim; option and constant are always in place in time, so only a very late filter is affected.

Separately: this rationale now exists in four near-verbatim copies — this docblock, `newspack-newsletters.php:129`, `README.md:74`, and the test class docblock. The package-internals half is the part most likely to drift on the next vendored update. Since the README section is the designated home for it, trimming this docblock to the one fact a caller needs and pointing at the README would leave fewer copies to keep true.

dkoo (approved):
Release risk: ELEVATED — the fix itself is correct and lands at the narrowest workable seam, but the front-end takeover it closes for flag-off sites remains fully open for flag-on sites, and the flag is the rollout switch for this feature.

I reproduced the regression and confirmed the fix in an isolated env: with the flag off, a public newsletter that previously rendered as a 6KB bare email now renders in the theme at 62KB with header and footer intact. Removing the guard brings the bare-email render straight back, so the gate is doing real work. The flag-on path is byte-identical to the pre-fix broken render, and a `the_post` probe shows the package's template runs no loop at all — which means Newspack's content gate, hooked on `the_post`, never engages there. For a newsletter restricted on the web, that is a full-text exposure to anonymous readers. None of this is introduced by this PR — it predates it — and gating the opt-in is the right first move. It is worth deciding what a public newsletter should look like under the new engine before the flag is turned on for anyone.

| Flag off, with this fix | Same URL, opt-in ungated (the regression) |
| --- | --- |
| Flag off, with this fix: the public newsletter renders in the theme's single template | Same URL with the opt-in ungated: no theme header, footer or sidebar |

Nothing here blocks merge. The most valuable additions before it lands would be a test that asserts the template actually resolved (the current pair passes on the base branch too) and a second look at the `init:11` CPT re-assertion, which now runs as pure duplicate work on every flag-off request.

One small correction for the description: I could not reproduce the "email-only blocks were shown on the web" half of the regression. On the pre-fix path the email-only block stayed hidden, because `remove_visibility_hidden_block()` is a plain `render_block` filter that still runs under the package's template. The theme-wrapper half reproduces cleanly.

plugins/newspack-newsletters/includes/email-renderers/class-editor-bootstrap.php
chickenn00dle: Good call — added a dedicated **Residual flag-on behavior (not closed here)** section to the README spelling out that with the flag on the package's `single_template` filter still takes over, renders the bare email-preview template, and — because that template runs no loop — Newspack's `the_post`-hooked content gate never engages, exposing a web-restricted newsletter's full text. Pointed it at NPPD-2150 rather than leaving the section reading as "solved." (b19b638)

plugins/newspack-newsletters/includes/email-renderers/README.md
chickenn00dle: Done — extracted `reassert_cpt_when_enabled()` and hooked that at init:11 instead of `register_cpt` directly. It reads the flag lazily (matching `add_post_type` at init:10), so with the flag off the re-registration is skipped entirely rather than running as duplicate work. Updated this README line to say so. (b19b638)

plugins/newspack-newsletters/newspack-newsletters.php
chickenn00dle: Reworded here and in the README twin — it now names what the flag-off boot actually leaves: a `sent` post status, the preview REST routes, the `email-contents` pattern category, a `post_types` field on the wp_template REST response, and the `safe_style_css` KSES widening (`display` + `mso-*`), noting they're additive/post-type-gated and none touch the front-end template path. (b19b638)

plugins/newspack-newsletters/tests/email-renderers/test-bootstrap-flag-gate.php
chickenn00dle: Fixed `test_canonical_cpt_args_remain_authoritative` — it now forces the flag on and drives the sequence directly: register the CPT with `public => false` to stand in for the package's init:10 clobber, then call `reassert_cpt_when_enabled()`, so it proves the re-assertion wins instead of passing against the untouched registration. (b19b638)

plugins/newspack-newsletters/tests/email-renderers/test-bootstrap-flag-gate.php
chickenn00dle: Added `test_single_template_taken_over_only_when_flag_on` — with a public newsletter as global `$post`, it asserts `apply_filters('single_template', '/theme/single.php')` returns the theme path with the flag off and the package's `single-email-post-template.php` with it on. Verified it fails on the ungated code (the exact regression) where the old pair passed. Kept the block-visibility test per your note. (b19b638)

plugins/newspack-newsletters/includes/email-renderers/class-editor-bootstrap.php
chickenn00dle: Tightened — the claim is now "any filter registered before init:10" (option/constant are always in time; a later filter half-engages since `register_email_post_types()`/`Templates::initialize()` read the list once at init:10). Also trimmed this docblock to the one fact a caller needs and pointed it at the README as the single home for the rationale. (b19b638)

chickenn00dle:
Thanks for the thorough pass, @dkoo — reproducing it in a running env and pinning the byte sizes was genuinely useful. All addressed in b19b638:

- **`init:11` re-assertion running as duplicate work on flag-off** — extracted `reassert_cpt_when_enabled()` and gated it on the same lazy flag read, so it only runs when the package actually clobbers the CPT (flag on). Your point about it silently undoing a `remove_post_type_support()` between 10 and 11 is exactly why it shouldn't run unconditionally.
- **No test asserted the thing that broke** — added `test_single_template_taken_over_only_when_flag_on`, which resolves `single_template` directly: theme path with the flag off, the package's `single-email-post-template.php` with it on. It fails on the ungated code where the old pair passed. Also fixed the now-vacuous `test_canonical_cpt_args_remain_authoritative` to force the flag on and drive the clobber → re-assert sequence.
- **Description correction** — you're right, the email-only-blocks-shown half doesn't reproduce (`remove_visibility_hidden_block()` still runs under the package template). Corrected the description and every code/doc comment that claimed it — the regression is the theme wrapper only.
- **Doc precision** — reworded the "additive machinery" line to name what the flag-off boot really leaves (`sent` status, preview REST routes, KSES widening…), fixed "regardless of when it is registered" → "before init:10", and trimmed the duplicate rationale down to the README as the single home.

On the **release-risk / flag-on takeover**: agreed it's the bigger question, and it predates this PR. Raz opened [NPPD-2150](https://linear.app/a8c/issue/NPPD-2150/render-newsletter-post-in-theme-template) to decide what a public newsletter should look like under the WC engine — I've referenced it in the README's new "Residual flag-on behavior" section and the follow-ups so the next reader doesn't trust the gate further than it goes.

PR #779, opened 2026-07-31
fix(reader-activation): don't sync generated display names (NPPM-3018)

miguelpeixe:
### All Submissions:

On sites without WooCommerce, contact sync sends the reader's display name to the ESP as the contact name. Readers who register without providing a name get a display name generated from the email address (`jdoe` for `jdoe@example.com`), so the ESP stores that slug as the reader's name and newsletter greetings render it ("Hi jdoe"). With WooCommerce active the contact name comes from the billing name instead, which is why activating WooCommerce made the problem disappear on affected sites.

Contact sync now omits the name when the display name was auto-generated, using the same `reader_has_generic_display_name()` check that already governs display names in My Account and comments. The name key is omitted rather than sent empty, so ESP providers leave any existing contact name untouched. Readers who deliberately saved a generic-looking name, and sites with `NEWSPACK_ALLOW_GENERIC_READER_DISPLAY_NAMES`, keep syncing their display names.

Closes NPPM-3018.

1. On a site with reader activation and ESP sync enabled and WooCommerce inactive, register a reader with only an email address. The ESP contact is created without a first or last name (previously: the email slug as First Name).
2. Give that reader a real display name (`wp user update <id> --display_name='Sample Reader'`) and run `wp newspack esp sync --user-ids=<id>`. The ESP contact now has the display name as its name.
3. Run `n test-php --filter Test_Contact_Sync_Name` in `plugins/newspack-plugin`. All 5 tests pass.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
+++ b/plugins/newspack-plugin/includes/reader-activation/sync/class-contact-sync.php
@@ -1333,6 +1333,19 @@ class Contact_Sync extends Sync {
 	}
 
 	/**
+	 * Whether WooCommerce customer data is available for building contact data.
+	 *
+	 * Wraps the class_exists() check so tests can simulate a site without
+	 * WooCommerce — the suite's WC mocks are loaded process-wide, which makes
+	 * the WooCommerce-less path below unreachable through class_exists() alone.
+	 *
+	 * @return bool
+	 */
+	protected static function is_woocommerce_available() {
+		return class_exists( '\WC_Customer' );
+	}
+
+	/**
 	 * Get contact data for syncing.
 	 *
 	 * @param int           $user_id The user ID.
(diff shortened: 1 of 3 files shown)


miguelpeixe:
Closing: #684 landed the same fix while this was open, and covers more ground.

`Contact_Metadata::get_full_name()` now gates the email-derived display name for every caller, so it applies on both the WooCommerce and WooCommerce-less paths. This PR only ever reached the WooCommerce-less one: on the WooCommerce path, `get_contact_with_metadata()` replaces `$contact` wholesale. #684 also falls back to the user's first and last name before the display name, and omits the `name` key instead of sending an empty string.

One deliberate difference: `has_email_derived_display_name()` honors the reader's saved-name meta but not `NEWSPACK_ALLOW_GENERIC_READER_DISPLAY_NAMES`. The reasoning is that the constant decides whether to stop prompting readers for a real name, not whether the placeholder goes on the wire. Sites defining it no longer sync the generated name, which is narrower than what this PR described.

NPPM-3018 is fixed either way. The five tests here are the only coverage of the WooCommerce-less name path, so they're worth reopening separately against `Contact_Metadata::get_full_name()`.

PR #782, opened 2026-07-31
fix(blocks): input validation hardening for author, iframe and checkout surfaces

dkoo:
### All Submissions:

Input-validation hardening in `newspack-blocks`, covering three areas: what the author REST endpoints return, how file-import destinations are built, and how post-checkout redirect destinations are handled. Each area gains test coverage it didn't have.

Rationale, the change-by-change detail, and the review notes are in NPPM-3062.

**One publisher-visible behavior change:** a custom post-checkout destination pointing off-site now needs its host allowed through core's `allowed_redirect_hosts` filter. Worth a release note — see NPPM-3062.

1. `cd plugins/newspack-blocks && n test-php` — 77 tests pass. The new ones fail on `release`.
2. Smoke-test the author blocks, the Iframe block, and the modal checkout's after-success behavior. Step-by-step instructions and the verification already run against an isolated environment are in NPPM-3062.

### Other information:


jason10lee:
Thanks for taking this on!  Quick note:  no technical problems from my perspective--a series of compact, neat fixes here.  I'm (or rather Claude is) doing some extra scans to scope the publisher-visible change here.  I suspect only a really tiny group of publishers will be affected, but would like to back that up with evidence.  Formal review shortly!

jason10lee (approved):
Approved--no technical blockers, just some housekeeping.

- One site needs an allowlist entry — www.newsroom.co.nz on donations.newsroom.co.nz — landed with or before the release. Without it, that publisher's post-donation journey silently stops.
- Consider mentioning two behavior changes in the release note:  the filter update for offsite checkout destinations as above, plus that the author-email restriction hits every role below administrator in the Author Profile block's editor preview, which support will hear about even though it's preview-only.


Here are the previously-mentioned scan results, to put a number on the publisher-visible change, plus long-form analysis:

The Donate and Checkout Button blocks render the destination into a hidden field in public HTML, so we could survey this without touching a database: 4,178 requests across nine likely paths (`/`, `/donate/`, `/support/`, `/membership/`, `/give/`, `/join/` and a few more) over all 464 production sites, 429 of which responded.

- 137 sites render a modal-checkout block on one of those paths
- 102 expose the `after_success_url` field
- **12 have a destination actually configured**
- **3 of those 12 point somewhere the site would no longer redirect to**

For each of the twelve we compared the destination against the host WordPress reports through `home_url()`, rather than the host that served the page — that's what `wp_validate_redirect` actually compares against, and www/non-www canonicalisation would otherwise flip the answer in both directions.

Two of the three already send readers to a staging sandbox, so a completed purchase currently lands on a `newspackstaging.com` URL. For those two this branch is an improvement: closing the modal beats what happens today. That's a live misconfiguration unrelated to this change and wants its own ticket.

Which leaves one site genuinely affected — a publisher running donations on a subdomain and returning readers to the main site. One `allowed_redirect_hosts` entry covers it, and it should land with the release rather than after it, since the failure is silent.

Worth being explicit about what the survey misses: 35 sites refused the request, the path list is donation-shaped so a Checkout Button on a bespoke page is invisible to it, and prompts that don't render for an anonymous visitor won't appear. The floor is three; we'd expect single digits rather than dozens.

## The author-email change is publisher-visible too

`can_read_user_management_fields()` gates on `list_users`/`edit_users`, and in core `list_users` is administrator-only. The Author Profile block's editor requests `email` explicitly (`author-profile/edit.js:404`), so for every role below administrator — Editors included — the block's editor preview now renders without the email while the published page still shows it, since front-end renders go through `get_all_authors()` and bypass the restriction.

We checked whether that's worse than cosmetic: the response goes into React state rather than block attributes, so nothing is lost on save. Preview-only, and defensible given what the change is for — but the release note should probably describe two behavior changes rather than one.

## Two suggestions, neither blocking

**Sign the destination instead of allowlisting it**, as a follow-up rather than here. The doc comment on `sanitize_after_success_url` says the two sources are indistinguishable at the thank-you page, which is true only because nothing marks the one the block configured — and the server has that value at render time (`checkout-button/view.php` ~L115, `donate/frontend/class-newspack-blocks-donate-renderer-base.php` ~L207). An HMAC emitted alongside it, carried through the chain that already forwards the other three params (`class-donations.php:881`, `modal.js:960`) and accepted when either it verifies or `wp_validate_redirect` passes, would let publisher-configured destinations work on any host while a crafted one still has to clear the allowlist.

There's precedent in the codebase: the Newsletters subscribe block permits any publisher-configured host safely, because its destination never leaves server-rendered content. Signing buys the modal checkout the same trust it otherwise loses by crossing an iframe page load. At three affected sites this clearly doesn't need to block a hotfix.

**A signal when a destination is dropped.** `sanitize_after_success_url` currently drops one silently. A `do_action( 'newspack_blocks_after_success_url_rejected', $url )` costs nothing and gives us somewhere to hang detection, which this failure mode needs more than most.

Approving on that basis. The single `allowed_redirect_hosts` entry is the only item that needs to land with the release; everything else here is follow-up.

jason10lee:
Correction to the scan numbers above: the count of affected sites is one, not three.

Two of the three off-host destinations turn out to be inert. Both sit on Checkout Button blocks that save a URL but no after-success behavior, and `thankyou.php` renders the hidden fields only when a behavior is present, so `modal.js` never reaches the redirect branch. The URL is stored but unreachable — most likely a leftover from someone choosing "Go to a custom URL", entering one, then switching the setting back, since the editor hides the field without clearing the attribute. That holds on `release` and on this branch alike, so #782 changes nothing for those two.

That leaves the single subdomain case as the only site this affects, tracked in NPPM-3080. Approval stands.

dkoo:
Thanks, @jason10lee! Since several publishers need outstanding action items outlined in NPPM-3080, we need to hold this hotfix PR until that work is done. Adding a `DO NOT MERGE` label for now.

dkoo:
@jason10lee I just merged #791 into this PR (no squash, so all commits came intact) and updated with the latest `release` changes. This should bypass the need for any publishers to establish and maintain an allowlist for their own site. Can I ask you to do a once-over to ensure all looks mergeable? This PR had an "approve" verdict before #791 already, so with #791 also approved we shouldn't need a full re-review, more just a quick check to ensure the merge and updates from `release` didn't introduce any regressions or merge artifacts.

dkoo:
@jason10lee in the interest of getting this into the next deployment window, I'm going to go ahead and merge. But if you see any red flags please feel free to continue raising them!

PR #783, opened 2026-08-03
feat(plans): promotional URL generator for modal checkout (NPPD-1707)

miguelpeixe:
### All Submissions:

Publishers can already send readers into the modal checkout from a URL. Building those URLs has been hand-work, though: the parameters are undocumented, and a value that doesn't match what the page renders fails silently (NPPM-2872, NPPM-2815). TAMs assemble them on request.

This adds a **Get promotional link** action to the Plans screen. Pick a plan, and the modal builds a working link: which plan option to check out, the page the checkout opens over (**Homepage** by default, any page searchable), an optional coupon, what happens after checkout, and UTM parameters. Copy it, or open **Test link** to try it.

The link works over any page. When the trigger parameters are present, newspack-blocks renders the block the link needs — a Checkout Button or a Donate block — so nothing has to be added to a page first. A page that already carries a matching block still wins, so that block's coupon and after-checkout settings keep applying. A link that leaves the plan option to the reader is only caught by a page block that leaves it open too: a block locked to one option does not divert it, and the reader gets the picker.

Every option comes from what the reader's block will actually accept, derived server-side, so the UI can't emit a combination that fails silently. Where it can't offer something, it says so rather than producing a dead link.

Closes NPPD-1707.

**What ships where.** The Plans screen, its REST endpoints, and the generator UI sit behind the existing `NEWSPACK_PLANS_UI` gate. The **newspack-blocks changes ship live**: they extend URL triggering, and they carry four fixes that stand on their own —

* An after-checkout destination taken from a URL was honored unchecked. The thank-you screen assigns it to `window.location.href` when the modal closes, so a crafted link was an open redirect fired the moment a payment completed, on the publisher's own domain. A destination is now either on an allowed host or carries the signature a block renderer produced. Editor-authored block destinations may still point anywhere; arbitrary ones from a URL may not.
* A coupon whose cached ID outlived its post crashed the checkout with a 500. WooCommerce's data store throws when it reads the missing post; the coupon is now skipped instead.
* Forwarded UTM values reached `add_query_arg()` unencoded, so an ampersand or a space in a campaign name split into a stray parameter or was dropped by the redirect sanitizer. Values are encoded now.
* UTM parameters on a direct checkout or donation URL were dropped before reaching order attribution, because the passthrough only saw parameters parsed from the referer.

Requires WooCommerce and WooCommerce Subscriptions. Define `NEWSPACK_PLANS_UI` as true; the Plans screen then appears under Audience.

1. Open a published plan's row actions and choose **Get promotional link**. The generator opens with **Homepage** preselected as the target page.
2. Copy the link and open it logged out. The checkout modal opens over the homepage with the plan in the cart, even though the homepage carries no Checkout Button block.
3. Repeat for a variable plan without choosing a plan option. **Copy link** stays disabled until you pick one, or until you pick **Let the reader choose**.
4. Choose **Let the reader choose** and open the link. The plan picker opens over the page; picking an option continues to checkout.
5. For a grouped plan that bundles a non-subscription product, open the **Plan option** list. The non-subscription product is absent, because the picker cannot serve it.
6. Enter a coupon restricted to the promoted plan. The field reports it valid, and completing the checkout applies the discount.
7. Enter a coupon restricted to an unrelated product. The field explains it does not apply to this plan and withholds the link.
8. Set **After checkout** to *Continue to a custom URL* and leave the URL empty. The link is withheld. Enter an off-site URL. The generator says the URL must be on this site.
9. Set an on-site continue URL and complete a checkout. The thank-you screen's continue button carries your label and sends the reader to that URL.
10. Search the **Target page** field for another published page and select it. The link is built on that page's permalink.
11. On the Donations row, generate a link. Frequency and amount options come from Donations settings. Open the link on the homepage: the donation modal opens with that frequency and amount selected.
12. Fill the **Campaign tracking** fields with a multi-word campaign name and complete a checkout. The order records the campaign name without truncation.

### Other information:

* [x] Have you written new tests for your changes, as applicable? PHPUnit covers the config derivation, the synthesized-block attributes, the redirect allowlist and signature, and the UTM helpers in both plugins; Jest covers the URL builder, the option derivation, and the validation gate.
* [x] Have you successfully run tests with your changes locally? Both PHPUnit suites, Jest, PHPCS and ESLint pass. The flows in the test steps above were exercised over HTTP against an isolated environment, including a completed order.

Notes for reviewers:

* The coupon pre-check is advisory. It applies the product-restriction rule WooCommerce applies to a cart line, the promoted product and its parent, so a coupon restricted to a sibling variation reads as not applying, the same verdict checkout would reach.
* `after_success_behavior=referrer` is deliberately not offered in the generator. It returns the reader where they came from, which a link opened from an email or a QR code has no referrer or history entry to satisfy.
* Sites that need an off-site after-checkout destination from a URL can add the host via the `allowed_redirect_hosts` filter.
* Name Your Price is not installed on any local environment, so the `layout=untiered` donation mapping is covered by unit tests rather than a manual pass.
* UI screenshots are still to come.

<details>

**Serving a trigger without a block.** `Modal_Checkout::maybe_setup_url_triggered_checkout()` runs on `wp`. When the request carries `?checkout=1&type=…`, it renders the real block into the footer, hidden, and lets the existing trigger drive it. Rendering the block rather than hand-rolling markup brings everything the trigger depends on: the form it submits, the `data-checkout` payload, the variation and tiers picker modals, and the modal assets. The page's own block precedes the synthesized one in DOM order, so `findCheckoutButtonForm()` prefers it.

Checkout Button attributes mirror what the editor stores: a variable parent carries `is_variable` with any lock as `variation` (`view.php` collapses the pair onto the variation), while grouped parents and specific products are `product` alone. A variation requested against a product that cannot serve one is refused rather than rendered as a button that ignores it. A Donate block is rendered from its schema defaults, which resolve to the site's donation settings — the same source the generator derives its frequency and amount options from, so the two agree by construction.

**No page scanning.** An earlier revision scanned `wp_posts` for pages carrying a compatible block and offered only those as targets. Rendering the block on demand removes the need: `Promo_Url_Config` now only answers which plan children a link may name and what a donation link may ask for. That removed the unindexed `LIKE` queries, the synced-pattern resolution, the transient cache and its `save_post` invalidation.

**Which plan children a link may name.** A grouped plan's picker is `Subscriptions_Tiers::render_form()`, whose `get_tiers_by_frequency()` keeps only subscription-typed, non-private children and expands a variable subscription into its variations. `get_picker_members()` mirrors that, so the generator never offers a child whose radio would not render. `resolveProductParams()` also never emits `product_id === variation_id`, which the trigger rejects (NPPM-2872).

**Layout parameter.** `map_donate_configuration()` encodes the semantics of `triggerDonationForm()`: `tiered` targets the tiers block, `untiered` the Name Your Price input, anything else the frequency radios. Getting it wrong is a silent failure (NPPM-2815), so it lives server-side with unit coverage.

**After-checkout destinations.** `sanitize_after_success_url()` accepts a destination on a host the site allows, or any destination carrying `sign_after_success_url()`'s HMAC. Both block renderers sign what they emit, so editor-authored destinations keep working anywhere. The generator cannot sign — its input arrives in the request, and signing untrusted input would authorize exactly what the signature exists to refuse — so links it produces stay on-site. Validation runs at the thank-you screen and again in `get_after_success_params()`, so an unauthorized destination never travels in a generated URL.

**Picker context.** Anything a button attaches — after-checkout settings, the destination signature, an auto-applied coupon, attribution — is lost when the picker form is submitted instead, unless it is listed in `PICKER_CONTEXT_FIELDS`. That list is shared between the click path and the URL-trigger path (extended from #768).

</details>


plugins/newspack-plugin/includes/wizards/audience/class-promo-url-targets.php
```diff
@@ -0,0 +1,519 @@
...
+			$referencing = $wpdb->get_col(
+				$wpdb->prepare(
+					"SELECT ID FROM {$wpdb->posts} WHERE post_status = 'publish' AND post_type IN ('page','post') AND ( post_content LIKE %s OR post_content LIKE %s ) LIMIT %d",
+					$ref_like_comma,
+					$ref_like_brace,
+					$limit
+				)
+			);
+			$ids = array_merge( $ids, $referencing );
+		}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in fec140d77 — the referencing-pages query now orders by post_modified DESC and sets the truncated flag when it fills, so overflow selection is deterministic and signaled even when the deduped union stays under SCAN_LIMIT.

plugins/newspack-plugin/includes/wizards/audience/class-promo-url-targets.php
```diff
@@ -0,0 +1,519 @@
...
+		if ( ! Donations::is_platform_wc() ) {
+			return null;
+		}
+		$settings = Donations::get_donation_settings();
+		if ( is_wp_error( $settings ) ) {
+			return null;
+		}
+		$can_use_nyp = class_exists( '\Newspack_Blocks' ) ? \Newspack_Blocks::can_use_name_your_price() : false;
+		return self::build_direct_donation_config( $settings, Donations::get_donation_product_child_products_ids(), $can_use_nyp );
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in fec140d77 — build_direct_donation_config() now returns null when no frequency has a donation product, and the modal shows the existing donations-not-configured message instead of emitting a URL the handler would ignore. Covered by a new unit test.

plugins/newspack-plugin/includes/wizards/audience/class-promo-url-config.php
```diff
@@ -0,0 +1,306 @@
...
+		}
+		$attrs      = [];
+		$block_type = \WP_Block_Type_Registry::get_instance()->get_registered( 'newspack-blocks/donate' );
+		if ( $block_type ) {
+			$attrs = $block_type->prepare_attributes_for_render( $attrs );
+		}
+		return self::evaluate_donate_configuration(
+			\Newspack_Blocks_Donate_Renderer_Base::get_configuration( $attrs ),
+			\Newspack_Blocks::can_use_name_your_price()
+		);
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 7fbb0eb73 — get_donate_config() now intersects the enabled frequencies with the same frequency-to-product map the checkout's bail reads (get_donation_product_child_products_ids()), disabling productless ones and returning null when none remain. The intersection is a pure helper with unit coverage.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -228,6 +266,135 @@ function( $product ) use ( &$seen ) {
...
+		if ( ! $product_id ) {
+			return new \WP_Error(
+				'newspack_promo_targets_missing_product',
+				esc_html__( 'product_id is required for checkout button targets.', 'newspack-plugin' ),
+				[ 'status' => 400 ]
+			);
+		}
+		$family                         = Promo_Url_Config::get_product_family( $product_id );
+		$response['eligible_children']  = Promo_Url_Config::get_eligible_children( $family );
+		$response['offerable_children'] = Promo_Url_Config::get_offerable_children( $family );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 7fbb0eb73 (server) and 2f34e09f4 (UI) — offerable_children now applies the renderer's price gate: variations require the parent's resolvable price, members their own price with the NYP suggested/minimum fallback. The endpoint also reports parent_offerable, and the modal refuses a link naming an unpriced childless plan instead of emitting one that renders nothing.

plugins/newspack-plugin/src/wizards/audience/views/subscription-products/promo-url-options.ts
```diff
@@ -0,0 +1,219 @@
...
+				String( donateConfig.minimum )
+			);
+		}
+		if ( typeof effectiveAmount === 'number' && donateConfig.layout_param !== 'untiered' && ! input.presets.includes( effectiveAmount ) ) {
+			return __( 'Choose one of the amounts available on the target page.', 'newspack-plugin' );
+		}
+	}
+	if ( 'invalid' === input.couponState ) {
+		return input.couponReason || __( 'The coupon code is not valid.', 'newspack-plugin' );
+	}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 2f34e09f4 — a settled coupon verdict now only counts for the exact code/product pair it answered; anything else reads as checking, and getValidationError() withholds the link while a check is pending.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -540,6 +553,205 @@ function ( $item ) {
...
+			}
+			// Defaults resolve to the site's donation settings, which is what the
+			// generator derived its frequency/amount options from.
+			self::$url_triggered_block_html = do_blocks( '<!-- wp:newspack-blocks/donate /-->' );
+		} elseif ( 'checkout_button' === $type ) {
+			$attrs = self::build_url_triggered_button_attrs_from_request();
+			if ( empty( $attrs ) ) {
+				return;
+			}
+			self::$url_triggered_block_html = do_blocks( '<!-- wp:newspack-blocks/checkout-button ' . serialize_block_attributes( $attrs ) . ' /-->' );
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 84c746a90 — resolveCheckoutButtonForm() now resolves page-authored forms first at every step (exact match, then the picker fed by a page button's context) and reaches the synthesized form only when the page has nothing for the product. New resolver tests cover the locked-synthesized vs unlocked-page case.

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
```diff
@@ -225,6 +225,54 @@ export function copyContextFields( sourceForm, targetForm, fields = PICKER_CONTE
...
+ */
+export function appendUtmFields( form, params ) {
+	if ( ! form || ! params ) {
+		return;
+	}
+	const doc = form.ownerDocument;
+	Object.keys( params ).forEach( name => {
+		if ( ! params[ name ] || form.querySelector( `input[name="${ name }"]` ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
miguelpeixe: Addressed in 84c746a90 — the presence check goes through form.elements.namedItem() instead of a selector, so a field name carrying selector syntax can no longer throw mid-submission; covered by a test with a hostile key.

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
```diff
@@ -245,14 +360,36 @@ export function resolveCheckoutButtonForm( root, productId, variationId, options
 
 	if ( ! hasVariation ) {
```
adekbadek, line 391: **Blocker:** "Let the reader choose" is overridden by a page button locked to one variation — A no-variation link resolves to any page-authored button for the parent, including one locked to a single variation, so the reader is checked out on that variation instead of being shown the picker.

`findCheckoutButtonForm( root, productId, null, { synthesized: false } )` sets `hasVariation` to false, which skips the variation comparison entirely — the first page button whose `data-checkout.product_id` matches wins. A variable button locked to a variation serializes exactly that way: `view.php:101` collapses `product` onto `variation`, and `Checkout_Data::get_checkout_data()` stores a variation as `product_id: <parent>, variation_id: <child>`.

Verified on the review env with a variable subscription plan (Monthly $10/month, Annual $100/year) and a page carrying a button locked to Annual. The link the generator produces for "Let the reader choose" opened checkout for `Membership Plan × 1 — $100.00 / year` with no picker. The same link on a page with no button opened the picker with both variations, which is the intended behaviour — so the mechanism works, and the page button is what diverts it.

`getDroppedLinkContext()` does not cover this: it reports only `coupon` and the `after_success_*` params, so nothing warns the publisher or the reader.

**To reproduce**
1. Publish a variable subscription plan with two variations at different prices.
2. Publish a page carrying a Checkout Button block locked to the more expensive variation.
3. Open `<that page>/?checkout=1&type=checkout_button&product_id=<parent id>`.
4. Checkout opens on the locked variation. Expected: the variation picker.

One fix is to prefer an unlocked page button in the `! hasVariation` branch, the preference `findContextDonorForm()` already encodes, and fall through to the synthesized parent button when the page offers only locked ones. Whichever way it goes, `checkout-button-trigger.test.js:100` ("matches by product_id only when no variation is requested") currently pins the present behaviour and would need to move with it.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -228,6 +266,139 @@ function( $product ) use ( &$seen ) {
...
+		$product_context = [];
+		$product_id      = absint( $request->get_param( 'product_id' ) );
+		if ( $product_id && function_exists( 'wc_get_product' ) ) {
+			$product    = wc_get_product( $product_id );
+			$family     = Promo_Url_Config::get_product_family( $product_id, false );
+			$family_ids = array_merge( [ $family['parent'] ], $family['variations'], $family['members'] );
+			$parent_id  = $product ? (int) $product->get_parent_id() : 0;
+			if ( $parent_id ) {
+				$parent_family = Promo_Url_Config::get_product_family( $parent_id, false );
+				$family_ids    = array_merge( $family_ids, [ $parent_family['parent'] ], $parent_family['variations'], $parent_family['members'] );
```
adekbadek: **Suggestion:** Coupon pre-check accepts a coupon restricted to a sibling variation — `family_ids` is widened to the parent and every sibling in the parent's family, and `evaluate_coupon()` then tests `product_ids` and `excluded_ids` against that whole set (class-promo-url-config.php:345 and :353). WooCommerce matches per cart item against `array( $product->get_id(), $product->get_parent_id() )` only — `WC_Coupon::is_valid_for_product()` never considers siblings. So a coupon whose `product_ids` is `[sibling]` intersects the widened family and reports valid, and `excluded_ids = [promoted]` leaves the diff non-empty and also reports valid; checkout applies no discount either way and the reader pays list price.

The PR notes call the pre-check advisory and say this matches runtime behaviour. The first half holds, but the second does not: the runtime gate is narrower than the check, so the two disagree rather than agree — which is what makes the generator's "withhold the link when the coupon is invalid" contract misleading here rather than merely approximate.

Running the two product-ID checks against `[ promoted id, its parent id ]` would line them up. The widening is right for `family_category_ids` and worth keeping there, since a variation carries no terms of its own.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
```diff
@@ -1818,6 +1989,7 @@ public function enqueue_scripts_and_styles() {
 				'manage_products_url'              => admin_url( 'edit.php?post_type=product' ),
 				'policy_source_is_mock'            => Subscription_Policy_Resolver::IS_MOCK,
 				'woocommerce_subscriptions_active' => function_exists( 'wcs_get_subscriptions' ),
+				'newspack_blocks_active'           => class_exists( 'Newspack_Blocks' ),
```
adekbadek: **Suggestion:** The eligibility gate names the plugin rather than the capability it needs — `newspack_blocks_active` is `class_exists( 'Newspack_Blocks' )`, which is true of every newspack-blocks version ever shipped — including ones with no `maybe_setup_url_triggered_checkout()` to serve the URL the generator hands out.

In the monorepo both plugins ship from one branch and release together, so this is hygiene rather than a live risk. It is still the one place the plugin half assumes something about the blocks half, and `method_exists( '\\Newspack_Blocks\\Modal_Checkout', 'maybe_setup_url_triggered_checkout' )` states the actual dependency at no extra cost.

plugins/newspack-plugin/includes/wizards/audience/class-promo-url-config.php
```diff
@@ -0,0 +1,384 @@
...
+	 * exist, mirroring the checkout's own bail.
+	 *
+	 * @return array|null Promo donate config, or null when donations can't take one.
+	 */
+	public static function get_donate_config() {
+		if ( ! class_exists( '\Newspack_Blocks_Donate_Renderer_Base' ) || ! class_exists( '\Newspack_Blocks' ) ) {
+			return null;
+		}
+		$attrs      = [];
+		$block_type = \WP_Block_Type_Registry::get_instance()->get_registered( 'newspack-blocks/donate' );
```
adekbadek, line 295: **Suggestion:** Unregistered donate block leaves `[]` attributes going into the renderer — When `get_registered( 'newspack-blocks/donate' )` returns null, `$attrs` stays `[]` and is passed to `Newspack_Blocks_Donate_Renderer_Base::get_configuration()`, which reads `$attributes['defaultFrequency']`, `$attributes['layoutOption']` and `$attributes['tierStyle']` without `isset` — three PHP 8 undefined-key warnings, and a null `defaultFrequency` flowing onward.

Returning `null` when `$block_type` is falsy would match what the guard is for: resolving through the registry is what supplies the schema defaults, so without it there is no configuration worth deriving.

plugins/newspack-plugin/includes/class-donations.php
```diff
@@ -981,6 +990,32 @@ function ( $item ) {
 		}
 	}
 
+	/**
+	 * Merge utm_* parameters from the current request into a params array.
+	 *
+	 * The passthrough loop in process_donation_request() historically only saw
```
adekbadek: **Nit:** Docblock narrates the change rather than the behaviour — "The passthrough loop in process_donation_request() historically only saw params parsed from the referer URL, so utm params on a direct (cold) donation URL — which has no referer — were dropped" describes what the code used to do, which git already records. The paragraph after it carries the fact a reader needs: `appendUtmFields()` puts the values in `$_GET`, so the referer is not relied on. Same wording in the twin at `class-modal-checkout.php:559`.

plugins/newspack-plugin/tests/unit-tests/wizards/class-promo-url-config-test.php
```diff
@@ -0,0 +1,485 @@
...
+				'minimum_amount' => 0.0,
+			],
+			$overrides
+		);
+	}
+
+	/**
+	 * Test that an expired coupon is invalid regardless of product context.
+	 */
+	public function test_evaluate_coupon_expired_is_invalid() {
```
adekbadek: **Nit:** The `evaluate_coupon` cases are one data provider — The `test_evaluate_coupon_*` methods differ only in the fixture handed to `coupon_data()`, the optional `product_context`, and the expected `valid`. As a `@dataProvider` of `[ overrides, product_context, expected ]` each restriction rule becomes one row, and the next one is a row rather than a method.

The `assertNotEmpty( $result['reason'] )` assertions check only that some string came back; pinning the actual reason would make each row state its own guarantee. The rest of the suite bites — `test_offerable_children_gate_each_variation_on_its_own_price` and the resolver's synthesized-demotion cases each go red if their change is reverted.

adekbadek (changes requested):
Release risk: HIGH — two blockers, both reproduced on the review env: one on a money path (inline), one in the reader-facing plan picker (in the notes below, since it falls outside this PR's changed lines). Everything else here is a suggestion or a nit.

The security work holds up well. I checked the open-redirect fix specifically: the token carries its destination rather than sitting beside it, `hash_equals` guards the comparison, the namespace is blog-scoped, and the URL-trigger path clamps `after_success_url` through the allowed-host check *before* `view.php` can mint a token for it — so a crafted link cannot get an off-site destination vouched for.

- `plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php:426` — **Blocker:** Plan picker shows empty parentheses for an "Any …" variation — The reader-facing plan picker renders `Membership Plan ()` for any variation whose attribute value is empty. `get_product_title()` formats a variation as `'%s (%s)'` and `implode`s `get_variation_attributes()`. WooCommerce's **Any <attribute>** variation — set from the standard product admin, and a supported configuration — stores the attribute with an empty value, so the implode yields an empty string and the parentheses render with nothing between them. Verified on the review env: ``` variation with "Any Billing period": get_variation_attributes() = {"attribute_billing-period":""} get_product_title( $v, true ) = [Membership Plan ()] variation with "Annual": get_product_title( $v, true ) = [Membership Plan (Annual)] ``` The picker rendered both side by side: `Membership Plan () $10.00 / month` above `Membership Plan (Annual) $100 / year`. This file is not in the PR's diff and the same code is on `main`, so the defect is pre-existing. It is raised here because this PR is what routes readers into this picker by default: "Let the reader choose" generates a link that opens it over any page, where previously it appeared only where a publisher had placed a variable Checkout Button. Skipping the suffix when the joined attributes are empty keeps the name clean without touching the populated case.
- `plugins/newspack-blocks/includes/class-modal-checkout.php:1578` — **Suggestion:** `instanceof WP_Post` resolves into the plugin namespace, so the fallback never fires — Pre-existing — the identical line is on `origin/main`, so this PR did not introduce it. Flagging it because the PR extends this mechanism and the fix is one character. The file declares `namespace Newspack_Blocks;` with no `use WP_Post`, so the expression tests against `Newspack_Blocks\WP_Post`, which does not exist. It is unconditionally false, and `get_after_success_post_id()` returns 0 whenever `get_the_ID()` is empty — the widget and footer-template-part case the docblock is written for. No token is minted there, so a block-authored destination on those surfaces is refused. Confirmed on the review env: with the queried object a `WP_Post` and `get_the_ID()` false, the method returns `0` where `$queried instanceof \WP_Post` returns `19`, and `get_after_success_token()` for an off-site destination comes back empty. `tests/test-after-success-url.php` passes an explicit `$post_id` throughout, which is why nothing catches it.
- `plugins/newspack-blocks/includes/class-modal-checkout.php:1889` — **Suggestion:** One more `add_query_arg()` call site still passes unencoded values — This function is untouched by the PR, so it is pre-existing — but it is the same defect the PR fixes at `class-donations.php:960` and in the UTM loop, and `email` two lines above at :1874 is already `rawurlencode()`d, so the inconsistency sits inside one array literal. `$args` merges `get_after_success_params()`, whose `after_success_url` comes back from `sanitize_after_success_url()` as a decoded URL, and hands it to `add_query_arg()`. A destination carrying its own query string (`/thanks/?src=a&ref=b`) splits, and `&ref=b` becomes a top-level parameter on the thank-you URL. The token usually rescues the destination because `sanitize_after_success_url()` prefers it, so the symptom is mostly hidden rather than absent. Worth folding in while the encoding is fresh, or splitting out if you would rather keep this PR to its own scope.

Review provenance: 3 of 4 reviewer passes (lineup: deep, standards, wp-expert, codex); release risk: HIGH (degraded — skipped: codex).

plugins/newspack-blocks/src/modal-checkout/checkout-button-trigger.js
miguelpeixe: Fixed in 2e14cdb. `findCheckoutButtonForm()` no longer matches a variation-locked button when no variation is requested, so a reader-choice link resolves to an unlocked page button, then the synthesized parent button, then null. The `findCheckoutButtonForm` test at :100 moved with it, and three resolver cases cover a locked-only page, a locked button followed by an unlocked one, and a locked page with nothing synthesized.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Fixed in 4065849. The product checks now run against `[ promoted id, parent id ]`, built by a new `Promo_Url_Config::get_coupon_product_context()` that the REST handler calls; the category widening stays. The exclusion check changed shape with it: it was "invalid only when every family member is excluded" and is now "invalid when either id is", which is what `is_valid_for_product()` does. The PR note that called the two consistent is corrected.

plugins/newspack-plugin/includes/wizards/audience/class-audience-subscription-products.php
miguelpeixe: Changed in 4065849 to `method_exists( '\Newspack_Blocks\Modal_Checkout', 'maybe_setup_url_triggered_checkout' )`, localized as `promo_links_supported` so the key names the capability as well.

plugins/newspack-plugin/includes/wizards/audience/class-promo-url-config.php
miguelpeixe: Fixed in 4065849: `get_donate_config()` returns null when the block type is not registered. No unit test for this one. The plugin suite has no `Newspack_Blocks_Donate_Renderer_Base`, and stubbing it there would change branches in other tests.

plugins/newspack-plugin/includes/class-donations.php
miguelpeixe: Trimmed in 2e14cdb and 4065849. Both docblocks now carry only the `appendUtmFields()` paragraph and the precedence rule.

plugins/newspack-plugin/tests/unit-tests/wizards/class-promo-url-config-test.php
miguelpeixe: Done in 4065849: one `test_evaluate_coupon_mirrors_the_checkout_verdict` with a `coupon_verdicts` provider, each row pinning its reason string. The sibling-restricted and parent-excluded rows are the ones that went red before the fix.

miguelpeixe:
Addressed the review, ready for another look. Thanks for reproducing both blockers on the env; the reader-choice one would have shipped otherwise.

The three notes outside the diff are folded in:

- **Plan picker shows empty parentheses for an "Any …" variation**: fixed in e65870d. `get_product_title()` drops empty attribute values before joining and skips the suffix when nothing is left, so "Any Billing period" prints `Membership Plan` and a mixed set prints `Membership Plan (Premium)`. A data-provider test covers the three shapes.
- **`instanceof WP_Post` resolves into the plugin namespace**: fixed in 2e14cdb (`\WP_Post`). A new test drives a post through `go_to()` with the loop's global cleared and asserts a token is minted from the queried object.
- **`woocommerce_get_return_url()` passes unencoded values to `add_query_arg()`**: fixed in 2e14cdb. The after-success params are `rawurlencode()`d like `email` beside them. A new test sends a destination carrying `?src=a&ref=b` through and asserts it arrives whole with no top-level `ref`.

Inline threads are answered individually. The PR description is updated for the two behaviours that changed: page-block precedence for reader-choice links, and the coupon pre-check rule.

adekbadek (approved):
Both blockers are cleared, and the two suggestions landed: `\WP_Post` with a queried-object test, and `rawurlencode` on the return-URL params — which also stops a `+` in the after-success token decoding back as a space. `get_coupon_product_context()` matching `WC_Coupon::is_valid_for_product()` (product plus parent, never a sibling) is the right rule.

Two notes, neither worth another round:

- `Promo_Url_Config::get_product_family()`'s `$expand_picker_members` is now always `true` — the coupon pre-check that passed `false` was replaced by `get_coupon_product_context()`. The `false` branch is dead and the docblock still describes the caller that used it. Worth dropping next time the file is open.
- For a grouped plan with no child selected, `item_ids` is the grouped parent, which never appears in a cart line, so a coupon restricted to (or excluding) its members is judged against an id the checkout never sees. Equally approximate before this commit, so not a regression — flagging it because the surrounding rule is now exact for the variable case and this is the one plan shape it does not cover.

PR #686, opened 2026-07-21
feat(campaigns): Contextual Prompts, AI-assisted story donation asks

kmwilkerson:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guidelines](https://github.com/Automattic/newspack-workspace/blob/main/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Folds inewsource's **EngageLine** (an AI donation-CTA generator, GPL-2) into Newspack Campaigns as a native feature: **Contextual Prompts**. Grant-funded, shared with inewsource. Based on EngageLine by **Giovanni Moujaes (inewsource)**; the editor-in-the-loop model and non-advocacy framing are their contribution. See the [PRD/PLAN](https://github.a8c.com/Automattic/newspack-contextual-prompts).

A Contextual Prompt is a story-specific donation ask that lives inside the story. Copy is generated from the article itself, framed for where the prompt sits, and is always editable before it ships.

**Feature flag.** The plugin now declares `Requires at least: 7.0`. The whole feature is off unless `wp-config.php` defines `define( 'NEWSPACK_CONTEXTUAL_PROMPTS', true );`. Without the constant the block is not registered, the editor panel does not load, the REST routes 404, and the wizard tab is hidden. Follows the `NEWSPACK_CONTENT_GATES` pattern. The popups test bootstrap defines it so the suites exercise the feature.

Behind the flag there is still a per-site admin opt-in with an AI-use disclosure gate, off by default, because some newsrooms are contractually barred from AI use. Generation is blocked server-side as well as hidden in the UI. Accepting or withdrawing writes an audit record to the always-on log Newspack Manager consumes, carrying the actor, an explicit timestamp, the resulting state and the execution context (`cli`/`cron`/`rest`/`web`). It is hooked on the option itself (write, change and deletion), so a flip made over WP-CLI or by another plugin is recorded the same way, and only real state changes are logged. The last change is also persisted in a non-autoloaded option, so who accepted and when is answerable on the site itself.

[...]

PR #786, opened 2026-08-03
feat(campaigns): pin prompt children, add layout and Additional CSS

thomasguillot:
### All Submissions:

Giovanni's feedback on Contextual Prompts, plus a follow-up on the button CTA. Targets
`feat/contextual-prompts` so #686 stays clean while its GA4 sign-off is outstanding.

**The prompt's two children can no longer be rearranged.** `templateLock: 'all'` locked
them, but core offers every locked block an Unlock button in its toolbar, and the `lock`
attribute that button writes takes precedence over the template. One click and a confirm
brought back the drag handle, the movers and the block-type switcher, permanently and in
the saved markup. The lock now pairs with content-only editing mode on every descendant,
which withholds the Unlock button. The mode is set from the block rather than through
`templateLock: 'contentOnly'`, because that variant turns the prompt into a pattern-style
section and replaces its own inspector with a content list.

**Row and Stack on the container, when the CTA is a button.** Core renders the
orientation toggle by itself once the default layout is flex and switching is off, so
there is no bespoke control. Registration branches on whether donations are native, in
the JS and in the PHP block registration both: `wp_render_layout_support_flag()` reads the
server-side `supports.layout.default`, so a JS-only branch would render `is-layout-flow`
under an editor previewing a flex column. The vertical default carries
`justifyContent: 'stretch'` so children still fill the container, as they did under flow.

**Additional CSS in Edit Styles.** Block themes already get this in the Site Editor under
Styles > Blocks > Contextual Prompt > Advanced; classic themes had no equivalent. It
stores in its own option so the existing style allowlist stays strict, scopes through the
public `WP_Theme_JSON::process_blocks_custom_css()`, and needs `edit_css` on top of the
`manage_options` the route already required, which is core's own contract for the same
feature. Stored CSS keeps rendering if that capability is later revoked, and the field
goes read-only rather than disappearing, so it can still be found.

Validation is core's `</style>` check plus a structural scan, which core's server side
does not do. Core's UI for this is a CodeMirror editor that lints client-side; this field
is a plain textarea, and the output is concatenated into the shared `global-styles` inline
stylesheet, so an unclosed comment would swallow every style appended after ours. The scan
tracks quoted strings and comments so that `content: "{"` is not mistaken for an unbalanced
brace, and ends a string at an unescaped newline, carriage return or form feed, as the CSS
tokenizer does. A `/*` inside an unquoted `url()` is rejected; that is a deliberate,
documented conservative call, and quoting the URL avoids it.

**A site-wide label and destination for the button CTA.** The URL already had a global
source in the donor landing page; the label was the literal "Donate" in both the
insert-time template and the render-time repair path. Both now resolve from two new
profile fields, offered only when donations are not native. Empty means inherit. Both stay
editable per post, which survives the lock because core marks `text` and `url` on
`core/button` as content.

Closes DSGNEWS-205.

Contextual Prompt layout toggle and Additional CSS drawer

Left: the container keeps its own inspector, now with the orientation toggle, while the copy
and CTA are locked. Right: the Additional CSS panel in Edit Styles, and the new button label
field behind it.

1. On a site with `NEWSPACK_CONTEXTUAL_PROMPTS` defined and Contextual Prompts opted in,
   open a post holding a prompt. Select the copy paragraph, then the CTA. Neither should
   offer a drag handle, movers, a block-type switcher or an **Unlock** button, and neither
   should be removable. The paragraph should still take text and inline formatting.
2. Select the prompt container. Its own inspector should be unchanged: Layout, the Copy
   panel with **Regenerate Suggestions**, Advanced, and a Styles tab with Color,
   Typography, Dimensions and Border & Shadow.
3. Point the site at off-site donations, so the CTA is a plain button. The container's
   Layout panel should now offer **Horizontal** and **Vertical**. Set Horizontal and
   confirm the front end renders the copy and button side by side. On a native-donations
   site the toggle should be absent.
4. In Audience > Campaigns > Contextual Prompts, with off-site donations, set **Donation
   button label** and **Donation button URL**. Insert a new prompt and confirm the button
   picks both up. Change the settings and confirm an existing prompt keeps its own.
5. On a classic theme, open **Edit Styles**. Below Border there should be an **Additional CSS**
   panel. Enter `border-top: 4px solid magenta;` and `& p { letter-spacing: 0.03em; }`,
   save, and confirm both rules reach the front end scoped to the prompt, after the
   wizard's other style overrides.
6. Confirm `</style>` in that field is rejected with an error and does not overwrite the
   stored value.

### Other information:

The seven failing PHPCS jobs are the `WordPressVIPMinimum.Performance.NoPaging` sniff that
already fails on `main` and on #686. No line this branch adds is flagged; the flagged lines
in files it touches predate it by years.

### DIFF
--- a/plugins/newspack-plugin/src/wizards/audience/views/campaigns/contextual-prompts/index.js
+++ b/plugins/newspack-plugin/src/wizards/audience/views/campaigns/contextual-prompts/index.js
@@ -54,6 +54,7 @@ const ContextualPrompts = props => {
 	const [ status, setStatus ] = useState( null );
 	const [ values, setValues ] = useState( {} );
 	const [ blockStyles, setBlockStyles ] = useState( {} );
+	const [ customCss, setCustomCss ] = useState( '' );
 	const [ inFlight, setInFlight ] = useState( false );
 	const [ error, setError ] = useState( null );
 	const [ loaded, setLoaded ] = useState( false );
@@ -67,9 +68,21 @@ const ContextualPrompts = props => {
 	const savedValuesRef = useRef( {} );
 	// Same, for the block style overrides: the save only sends them when they differ.
 	const savedStylesRef = useRef( {} );
+	// Same again, for the custom CSS, which is its own option.
+	const savedCustomCssRef = useRef( '' );
 	// Monotonic id so a slow status response can't clobber state written by a
 	// newer request or mutation.
 	const statusRequestRef = useRef( 0 );
+
+	// Re-seed the drawer's state and the snapshots its dirty check measures against.
+	const applyStyleState = next => {
+		const nextStyles = normalizeStyles( next.styles );
+		setBlockStyles( nextStyles );
+		savedStylesRef.current = nextStyles;
+		const nextCustomCss = next.custom_css || '';
+		setCustomCss( nextCustomCss );
+		savedCustomCssRef.current = nextCustomCss;
+	};
 
 	const loadStatus = () => {
 		setError( null );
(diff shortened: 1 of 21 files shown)


thomasguillot:
Superseded by #797, which moves the Contextual Prompt design into a synced pattern instead of the wizard styles drawer.

PR #787, opened 2026-08-03
chore(gitignore): ignore security scan artifacts

jason10lee:
### All Submissions:

Adds two entries to `.gitignore`:

    CLAUDE-SECURITY-*
    .codex-security/

Both directories hold AI security scanner output — findings and patches. They
were untracked but not ignored, so a `git add -A` would have committed them to a
public repo.

The entries match by name rather than through a wildcard like `*security*`. The
`/data` entry further up carries a comment about an earlier unanchored pattern
that swallowed source directories named `data/` inside plugins and silently
dropped new files from commits. This names each scanner instead, and the comment
asks the next person to extend the list rather than widen the pattern.

1. Check out this branch.
2. `mkdir -p CLAUDE-SECURITY-probe .codex-security && touch CLAUDE-SECURITY-probe/x .codex-security/x`
3. Run `git status --porcelain` — neither path appears.
4. Run `git add -An` — neither path is listed.
5. Clean up: `rm -rf CLAUDE-SECURITY-probe .codex-security`

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — not applicable, `.gitignore` only.
* [x] Have you successfully run tests with your changes locally? — verified with `git check-ignore -v` and a `git add -An` dry run.


jason10lee:
Self-review round, recorded for the merge gate. This is the second round; the
first is covered below.

**What changed.** Two corrections to the comment. The patterns are untouched.

The sentence citing `/data` now takes from that block only the lesson it
records — a broad pattern swallows source directories — and states the depth
behaviour outright, since these two are unanchored and `/data` is not. And "a
single `git add -A` publishes them" was true of `.codex-security/` and
overstated for `CLAUDE-SECURITY-*`: that scanner writes its own `*` fence into
each report directory, so this line is a second layer rather than the only one.
The block now says which is which, and closes by asking whether a new scanner
fences its own output instead of telling you to add it by name.

**What was rejected.**

The first round argued this change matters because `bin/migration/sync-legacy.sh`
stages the whole tree. Discarded: that file was removed upstream in `04d2f65d50`,
and the branch was 537 commits behind, so the round was reading code that no
longer exists. On current `main` there is no unconditional `git add -A` left in
`bin/`, `.github/` or `config/`. What this guards is a person or an agent typing
it by hand. The branch was refreshed and the round re-run.

A suggestion to cite `.husky/*.local*` as the closer precedent, on the grounds
that it is name-specific like these. Declined on measurement: it contains a
slash, so git anchors it to the repo root, and `plugins/x/.husky/pre-commit.local`
is not ignored. It shares one property with these patterns rather than both, so
citing it would trade one inaccuracy for another.

**What the round covered.** Three reviewer passes on the refreshed tree, zero
blockers. Measured rather than reasoned: `git check-ignore` over every tracked
file in the repository returns nothing, so neither pattern excludes anything
already in version control; both match at any depth while `/data` does not; the
two names match what `claude-security` 0.11.0 and the codex sweep write on disk;
and `newspack-security-scanner` writes outside the workspace, so it needs no
entry here.

PR #789, opened 2026-08-03
fix(newsletters): ad dates shift a day on negative-offset sites

dkoo:
### All Submissions:

Newsletter ad start and expiration dates are whole calendar days. The meta is a bare `Y-m-d`, and `is_ad_active()` compares it as a string against the newsletter's own date — nothing in that path carries a time or a timezone. Every surface that showed one of these dates was nonetheless converting it into a timezone, and each got it wrong in its own way. This fixes all three.

**1. The editor sidebar selected — and saved — the previous day.**

`DatePicker` received a bare `Y-m-d`, which reaches it as a UTC instant rather than a site-local one. Both supported WordPress versions arrive there by different routes:

- **6.9** hands the string to `new Date()`, where a date-only ISO string is UTC by specification while a date-time string without an offset is local.
- **7.0** first tests it against `/Z|[+-]\d{2}(:?\d{2})?$/` to decide whether it already carries an offset — and the `-DD` day component of `2026-08-05` matches, so the value is treated as already-UTC.

Rendered back in the site timezone, that lands on the previous day wherever the offset is negative. This was not only cosmetic: clicking the day the picker had highlighted saved that wrong day back to the ad, so opening an ad and confirming its dates could move them a day earlier with nothing on screen to say so.

The fix appends a time component before handing the value over, so nothing is left to mistake for an offset. Noon rather than midnight: only the year/month/day is ever read back, and midday also survives the timezones where local midnight doesn't exist on a DST-transition day.

Because this works around two different parsing models, please keep the time component even if that regex is corrected upstream — 6.9 would still be wrong without it.

**2. The classic list table showed the wrong day.**

`wp_date( __( 'F j, Y' ), strtotime( $start_date ) )` resolves the bare date against PHP's default zone (UTC under WordPress), then converts into the site zone — so `2026-08-04` displayed as **August 3, 2026** on every site at a negative offset. A new `Ads::format_ad_date()` anchors at midnight in the site timezone instead, and both columns use it.

**3. The Ads list showed a meaningless clock time.**

The DataView columns formatted with the site's `date_format` option, which may legitimately include a time. On a site set to `F j, Y, g:i a`, the noon-UTC anchor surfaced as **"August 4, 2026, 8:00 am"** — implying a time-of-day boundary that does not exist, and shifting with the site's offset (8:00 am in EDT, 6:00 am in MDT, 1:00 pm in CEST) for identical data. The same leaked into the Status column's "Starts …" / "Expired …" labels.

These now format in UTC with a date-only pattern, so the day rendered is always the day stored, and no clock time appears. A hover title explains the window semantics that aren't otherwise discoverable: whole days in the site timezone, matched against the newsletter's own date, with both endpoints included.

Stored meta is unchanged throughout: a bare `Y-m-d` still goes back to the database.

Part of NPPM-3078.

1. Set the site timezone to one with a negative UTC offset (Settings → General → e.g. `New York`), and set Date Format to the custom value `F j, Y, g:i a`.
2. Create a Newsletter Ad, toggle on **Custom Start Date**, pick a date a few days out, and save.
3. Reload the ad. **Before:** the picker highlights the day before the one you chose. **After:** it highlights the date you saved.
4. Click the highlighted day and save again. **Before:** the stored date silently moves back a day. **After:** it stays put.
5. Repeat for **Expiration Date**.
6. Open the Ads list. **Before:** the Start/Expiration columns read `August 4, 2026, 8:00 am`. **After:** `August 4, 2026`, with a tooltip explaining the window.
7. Open the classic list at `edit.php?post_type=newspack_nl_ads_cpt`. **Before:** the columns read `August 3, 2026`. **After:** `August 4, 2026`.
8. Set the site timezone to UTC and to a positive offset (e.g. `Tokyo`, `Kiritimati`) and confirm every surface still shows the stored day.
9. `cd plugins/newspack-newsletters && npm test` and `n test-php newsletters`.

### Other information:

<details>
<summary>Verification detail</summary>

**Reproduction.** The picker's behaviour was reproduced outside the browser by running WordPress 7.0's own `inputToDate` / `startOfDayInConfiguredTimezone` / `setInConfiguredTimezone` (transcribed from `wp-includes/js/dist/components.js`) against the real `@wordpress/date`, and separately by booting the actual core bundles in jsdom for both 6.9.4 and 7.0.

Stored `2026-08-05`, across site timezones:

| Site offset | Bare `Y-m-d` | With time component |
| --- | --- | --- |
| UTC+0, +2, +9, +12, +14 | selects 5 | selects 5 |
| −4, −6, −7, −10 | selects **4** | selects 5 |

Round trip on a negative-offset site, clicking the highlighted day:

- before — highlights 4 → `onChange` `2026-08-04T20:00:00` → saves `2026-08-04`
- after — highlights 5 → `onChange` `2026-08-05T12:00:00` → saves `2026-08-05`

**Classic list column**, stored `2026-08-04`:

| Site timezone | Before | After |
| --- | --- | --- |
| UTC, Tokyo, Kiritimati | August 4, 2026 | August 4, 2026 |
| New York, Los Angeles, Honolulu | **August 3, 2026** | August 4, 2026 |

**Tests.** Both new suites were confirmed to fail against the previous code rather than passing vacuously — 4 failures in the JS suite and 8 in the PHP suite when the old formatters are restored.

Full runs: JS 268 passed / 31 suites; PHP 716 passed / 2960 assertions (3 pre-existing skips); PHPCS clean.

The day objects the picker hands to `isInvalidDate` are byte-identical before and after this change on both 6.9.4 and 7.0, so the disabled-day ranges are unaffected. (They do differ under the copy of `@wordpress/components` in `node_modules`, but that build ships with no supported WordPress version — the plugin externalizes to `wp.components` — so it isn't the code any site runs. The editor tests stub the picker for that reason rather than rendering the npm build.)

</details>

### DIFF
--- a/plugins/newspack-newsletters/includes/ads/class-ads.php
+++ b/plugins/newspack-newsletters/includes/ads/class-ads.php
@@ -375,6 +375,32 @@ final class Ads {
 	}
 
 	/**
+	 * Format a date-only ad meta value for display.
+	 *
+	 * `start_date`/`expiry_date` are whole calendar days: `is_ad_active()`
+	 * compares them as `Y-m-d` strings against the newsletter's own date, so
+	 * they carry neither a time nor a timezone. Anchoring at midnight in the
+	 * site timezone is what keeps the rendered day equal to the stored one —
+	 * `strtotime( 'Y-m-d' )` resolves against PHP's default zone (UTC under
+	 * WordPress) and `wp_date()` then converts into the site zone, which moved
+	 * the date a day earlier on every site at a negative UTC offset.
+	 *
+	 * @param mixed $value Raw meta value.
+	 * @return string Localized date, or '' when the value is empty or invalid.
+	 */
+	public static function format_ad_date( $value ) {
+		$value = self::sanitize_ad_date( $value );
+		if ( '' === $value ) {
+			return '';
+		}
+		$date = date_create_immutable( $value . ' 00:00:00', wp_timezone() );
+		if ( false === $date ) {
+			return '';
+		}
+		return wp_date( __( 'F j, Y' ), $date->getTimestamp() );
+	}
+
+	/**
 	 * Count only ads (self::CPT) tagged with each advertiser term, across
 	 * the same statuses the Ads list shows. Registered as the taxonomy's
 	 * `update_count_callback` so newsletter posts tagged with an advertiser
@@ -740,19 +766,11 @@ final class Ads {
 	 */
 	public static function custom_column( $column_name, $post_id ) {
 		if ( 'start_date' === $column_name ) {
-			$start_date = get_post_meta( $post_id, 'start_date', true );
-			if ( ! empty( $start_date ) ) {
-				echo esc_html( wp_date( __( 'F j, Y' ), strtotime( $start_date ) ) );
-			} else {
-				echo '—';
-			}
+			$start_date = self::format_ad_date( get_post_meta( $post_id, 'start_date', true ) );
+			echo '' === $start_date ? '—' : esc_html( $start_date );
 		} elseif ( 'expiry_date' === $column_name ) {
-			$expiry_date = get_post_meta( $post_id, 'expiry_date', true );
-			if ( ! empty( $expiry_date ) ) {
-				echo esc_html( wp_date( __( 'F j, Y' ), strtotime( $expiry_date ) ) );
-			} else {
-				echo '—';
-			}
+			$expiry_date = self::format_ad_date( get_post_meta( $post_id, 'expiry_date', true ) );
+			echo '' === $expiry_date ? '—' : esc_html( $expiry_date );
 		} elseif ( 'price' === $column_name ) {
 			$price = get_post_meta( $post_id, 'price', true );
 			if ( ! empty( $price ) ) {
--- a/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/fields.js
+++ b/plugins/newspack-newsletters/src/admin-shell/screens/ads-list/fields.js
@@ -8,7 +8,7 @@
 import { Icon, Tooltip } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { drafts, notAllowed, published, scheduled, trash } from '@wordpress/icons';
-import { gmdateI18n } from '@wordpress/date';
+import { dateI18n, getSettings as getDateSettings, gmdateI18n } from '@wordpress/date';
 
 import { getAdminUrl } from '../../admin-globals';
 import { formatPostDate } from '../../utils/format-date';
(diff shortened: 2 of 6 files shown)


dkoo:
@thomasguillot I'm flagging this for a brief design review as it touches the UI for the Newsletter Ads list view. Relevant UI changes:

- Start/Expiration Date columns show dates in the site's configured timezone, and no longer display a time (only the date matters for insertion logic)
- Hovering over a Start or Expiration Date value in the list view shows a tooltip explaining that the dates are inclusive:

  Screenshot 2026-08-04 at 10 10 28 AM

  Maybe the wording can be a little clearer/less wordy—I'm open to suggestions.

thomasguillot:
UI looks good to me, no concerns there.

On the tooltip copy:

> Ad scheduling uses whole days in the site timezone, matched against the newsletter's own date. Both the start and expiration days are included.

🟡 Non-blocking suggestion: it's long for a hover hint that fires on every date cell in the list. "Matched against the newsletter's own date" explains how the comparison works, which isn't what someone hovering a Start column is after, and it's the clause that makes the sentence need a second read. Dropping it leaves the two facts that matter: it's a whole day, and both ends count.

Per column reads clearest, since the inclusivity falls out of the wording instead of needing its own sentence:

- Start: `Runs from this day, in the site timezone.`
- Expiration: `Runs through the end of this day, in the site timezone.`

That's two strings in place of the shared `adDateHint()`, each saying the thing relevant to its column. If one shared string is preferable:

- `Whole days in the site timezone. Start and expiration days are both included.`

🟢 Non-blocking nit: `title` is the native browser tooltip, so it doesn't reach keyboard or touch users and can't be styled. `Tooltip` from `@wordpress/components` would match the rest of the admin shell. Fine either way for a supplementary hint, but worth naming.

dkoo:
Both taken, thanks — the per-column split reads much better.

**Copy.** Went with your per-column version:

- Start: `Runs from this day, in the site timezone.`
- Expiration: `Runs through the end of this day, in the site timezone.`

You were right that "matched against the newsletter's own date" was the clause forcing the second read. It's also the wrong altitude for a hover hint: it describes *how* the comparison works, which matters when you're debugging a send, not when you're scanning a column. That detail is in the code comment where the next person to touch `is_ad_active()` will find it.

**`Tooltip`.** Swapped, and it's the better component here — styled, consistent with the rest of the shell, and it wires up `aria-describedby` when shown.

One caveat worth naming, since you raised keyboard and touch specifically: `Tooltip` renders through Ariakit's anchor and doesn't make its child focusable on its own, so wrapping a plain `<span>` still leaves the hint hover-only. Making it keyboard-reachable means `tabIndex={0}` on every date cell — two extra tab stops per row, for a string that's identical down the whole column. That felt like trading one accessibility problem for a worse one in a data grid.

The placement that actually solves it is the column header: said once, reachable, no per-row cost. Happy to do that as a follow-up if you agree it's the right home — I didn't want to redesign the header affordance inside a hotfix.

Tests updated alongside: the per-column strings are asserted, plus a mount test, since the previous assertions only inspected props and would not have caught `Tooltip` throwing on a bad child shape.

dkoo:
Correction to my last comment — I had the `Tooltip` behaviour backwards.

I said it doesn't make the child focusable. It does: Ariakit's `TooltipAnchor` stamps `tabindex="0"` onto the wrapped element. Rendered output is literally `<span tabindex="0">August 4, 2026</span>`. So your nit was fully resolved by the swap, and my caveat about it staying hover-only was wrong.

The real trade-off is the one I mis-attributed: at the list's default 20 per page that's 40 focusable plain-text spans in the table. Reachable, but noisy to tab through. Your instinct to name the component was right and the change stands as-is — I'd still like to move the hint to the two column headers eventually, now for tab-order rather than reachability. Happy to leave that out of the hotfix.

Separately, a self-review of the full diff turned up two real defects in my own work, both now fixed:

**Status label was dated wrong.** `starts_at`/`expires_at` come out of the REST payload in two shapes — a noon-UTC anchor derived from the `Y-m-d` meta, but `post_date_gmt` (a real publish instant) for a `future` post. I had been formatting both the same way, first one way and then the other, so one of the two was always wrong:

- UTC formatting dated an ad publishing 21:00 in New York to the following day
- site-time formatting moved a noon anchor to the next day at UTC+12 and beyond, so the Expiration column and the "Expired …" label disagreed in the same row

They're now separate named formatters picked by `item.status`, with the high-offset cases pinned in tests.

**The site's `date_format` was being ignored.** Hardcoding `F j, Y` to drop the clock time also discarded locale ordering — a German site configured for `j. F Y` would have started rendering `August 4, 2026`. It now keeps the configured pattern and only falls back when that pattern itself contains a time.

Both were introduced by me in this PR, not pre-existing. Test count is 275, still green.

dkoo:
Correction to my last comment — I got the `Tooltip` behaviour wrong.

I said it "doesn't make its child focusable on its own, so wrapping a plain `<span>` still leaves the hint hover-only." That's not true. Ariakit's `TooltipAnchor` stamps `tabindex="0"` onto the wrapped element:

```
OUTER HTML: <span tabindex="0">August 4, 2026</span>
```

So your nit is fully addressed by the swap — the hint is keyboard reachable, no extra work needed. Sorry for the misdirection; I asserted that from reading the source instead of rendering it.

The trade-off is the opposite of what I described: not unreachability, but tab-order noise. Every date cell is now a focusable stop, so at the list's default 20 per page that's 40 stops for a string that's identical down each column. I still think the column header is the better home for it, but now for a plain UX reason rather than an accessibility one — happy to leave it as-is if you'd rather not spend more on it.

Separately, a self-review of the full diff caught a regression I'd introduced, since it touches the same lines you reviewed.

The Ads list gets `starts_at` / `expires_at` from the REST status payload, and they arrive in two shapes: a noon-UTC anchor derived from the `Y-m-d` meta, but the real publish instant (`post_date_gmt`) for a `future` post. I'd been formatting both the same way, which is wrong either way round — UTC dates a 21:00 scheduled publish a day forward, and site time moves the anchor a day forward once the offset reaches +12. The Auckland case had the Expiration column and the status label disagreeing by a day in the same row.

They're now two named formatters picked by shape, with the +12/+14 cases pinned in tests.

The same pass turned up that hardcoding the date pattern threw away the site's `date_format`, which is what supplies locale ordering — a German site showing `August 4, 2026` instead of `4. August 2026`. It now keeps the configured pattern and only overrides when that pattern carries a time, which was the actual thing worth stripping.

dkoo:
Going to merge this one since it affects all production sites using newsletter ads.

PR #791, opened 2026-08-03
fix(blocks): sign publisher-configured post-checkout destinations

dkoo:
### All Submissions:

**Stacked on #782 — targets `hotfix/blocks-input-validation`, not `release`.** Merge this into #782, then #782 into `release`, so the durable fix ships in one go.

#782 validates post-checkout destinations, which means a destination pointing somewhere other than the site's own host is refused unless that host is added to `allowed_redirect_hosts` in per-publisher custom code. That works, but it puts a permanent support burden on every publisher who configures an off-site destination, and the failure is silent.

This removes the need for that custom code. When a block renders a destination a publisher configured, the site signs it. A destination arriving with a valid signature is honoured wherever it points; anything else still has to clear `wp_validate_redirect()`. A destination supplied in a link carries no signature, so the behaviour #782 closes stays closed.

Signing happens at render rather than at save, so **blocks published before this ships are covered with no migration** — the signature is computed from the block's attributes each time the page renders. Reusable blocks are included, since they render through the same callbacks.

`wp_hash()` rather than a nonce: the value has to survive page caching and a reader who signs in partway through checkout, and a nonce is bound to a user and a time window. The signature is site-specific, so it means nothing on another site.

1. `cd plugins/newspack-blocks && n test-php` — 89 tests, 11 new. `n test-js` — 42 tests.
2. Add a Donate or Checkout Button block, set "Post-Checkout Button" to **Go to a custom URL**, and enter a URL on a host the site does not own.
3. Complete a checkout. The reader is taken to that destination — no `allowed_redirect_hosts` entry needed.
4. Now craft a checkout link by hand with `?after_success_url=` pointing at another host, and complete a checkout through it. The destination is refused and the modal closes, because the link carries no signature.
5. Confirm a same-host destination still works, and that an existing block published before this change works without being re-saved.

### Notes for reviewers

* **The trust boundary moves, and it's worth an explicit decision.** Anyone who can publish a block can mint a signature for a destination of their choosing. That's the boundary the feature already has — a Checkout Button's destination has always been settable by whoever can publish — and far tighter than before #782, where any URL in a query string worked with no authentication. It is looser than the allowlist, which needs a code deploy.
* **The hazard here is normalization drift.** The signature covers an exact string, and the destination passes through sanitising between the block that emits it and the page that checks it. Both sides normalise through one shared helper for that reason, and there's a test over query strings, fragments, encoded spaces, capitalised hosts, ports and trailing slashes.
* **The `newspack-plugin` change is one line plus a comment,** adding the signature to the params `class-donations.php` already forwards. Without it a signed destination loses its signature crossing that path and silently degrades to allowlist-only.
* Once this ships, the six per-site `allowed_redirect_hosts` entries tracked in NPPM-3080 become unnecessary. They're harmless if already added.

### Other information:

### DIFF
--- a/plugins/newspack-blocks/includes/class-modal-checkout.php
+++ b/plugins/newspack-blocks/includes/class-modal-checkout.php
@@ -377,6 +377,7 @@ final class Modal_Checkout {
 		$after_success_behavior     = filter_input( INPUT_GET, 'after_success_behavior', FILTER_SANITIZE_SPECIAL_CHARS );
 		$after_success_url          = filter_input( INPUT_GET, 'after_success_url', FILTER_SANITIZE_URL );
 		$after_success_button_label = filter_input( INPUT_GET, 'after_success_button_label', FILTER_SANITIZE_SPECIAL_CHARS );
+		$after_success_signature    = filter_input( INPUT_GET, 'after_success_signature', FILTER_SANITIZE_SPECIAL_CHARS );
 
 		if ( $variation_id ) {
 			$product_id = $variation_id;
@@ -395,7 +396,7 @@ final class Modal_Checkout {
 			wp_parse_str( $parsed_url['query'], $params );
 		}
 
-		$params = array_merge( $params, compact( 'after_success_behavior', 'after_success_url', 'after_success_button_label' ) );
+		$params = array_merge( $params, compact( 'after_success_behavior', 'after_success_url', 'after_success_button_label', 'after_success_signature' ) );
 
 		if ( function_exists( 'wpcom_vip_url_to_postid' ) ) {
 			$referer_post_id = wpcom_vip_url_to_postid( $referer );
@@ -1199,31 +1200,99 @@ final class Modal_Checkout {
 	const AFTER_SUCCESS_BEHAVIORS = [ 'custom', 'referrer' ];
 
 	/**
-	 * Reduce a post-checkout destination to one this site is willing to send readers to.
+	 * Reduce a post-checkout destination to a single comparable form.
 	 *
-	 * The destination reaches the thank-you page through the request, whether it came from
-	 * the block's own settings or from the link the reader followed, and by that point the
-	 * two are indistinguishable. Core's redirect validation decides, so the destinations
-	 * this site accepts are the ones `allowed_redirect_hosts` reports. That's this site's
-	 * own host, plus whatever a publisher adds through that filter, plus anything another
-	 * plugin has added (`Newspack\NRH` registers one).
+	 * Signing and verifying have to agree on the exact string, and the value passes through
+	 * sanitising on its way between them, so both sides normalise here rather than each
+	 * doing their own.
 	 *
 	 * @param string $url The requested destination.
 	 *
-	 * @return string The destination, or an empty string if it is not allowed.
+	 * @return string The normalised destination, or an empty string.
 	 */
-	public static function sanitize_after_success_url( $url ) {
+	public static function normalize_after_success_url( $url ) {
 		$url = sanitize_url( (string) $url );
 
 		if ( empty( $url ) ) {
 			return '';
 		}
 
-		// Core compares hosts case-sensitively; host names aren't. Normalise first so this
-		// site's own host typed in capitals isn't read as somewhere else.
+		// Core compares hosts case-sensitively; host names aren't. Normalise so this site's
+		// own host typed in capitals isn't read as somewhere else.
 		$host = wp_parse_url( $url, PHP_URL_HOST );
 		if ( $host && strtolower( $host ) !== $host ) {
 			$url = str_replace( '://' . $host, '://' . strtolower( $host ), $url );
+		}
+
+		return $url;
+	}
+
+	/**
+	 * Sign a post-checkout destination a block has been configured with.
+	 *
+	 * Minted where the block renders, which is the last point this site knows the value came
+	 * from its own content rather than from the request. `wp_hash()` rather than a nonce:
+	 * the signature has to survive page caching and a reader who signs in partway through
+	 * checkout, and a nonce is bound to a user and a time window.
+	 *
+	 * @param string $url The destination to sign.
+	 *
+	 * @return string The signature, or an empty string if there is nothing to sign.
+	 */
+	public static function get_after_success_url_signature( $url ) {
+		$url = self::normalize_after_success_url( $url );
+
+		return $url ? wp_hash( $url ) : '';
+	}
+
+	/**
+	 * Whether a destination carries this site's signature.
+	 *
+	 * @param string $url       The requested destination.
+	 * @param string $signature The signature offered with it.
+	 *
+	 * @return boolean
+	 */
+	public static function is_after_success_url_signed( $url, $signature ) {
+		$signature = (string) $signature;
+		$expected  = self::get_after_success_url_signature( $url );
+
+		if ( '' === $signature || '' === $expected ) {
+			return false;
+		}
+
+		return hash_equals( $expected, $signature );
+	}
+
+	/**
+	 * Reduce a post-checkout destination to one this site is willing to send readers to.
+	 *
+	 * The destination reaches the thank-you page through the request, whether it came from
+	 * the block's own settings or from the link the reader followed, and by that point the
+	 * two look alike. A signature tells them apart: one is minted when a block renders a
+	 * destination a publisher configured, so a destination carrying a valid one is this
+	 * site's own and is honoured wherever it points.
+	 *
+	 * Everything else falls to core's redirect validation, which accepts what
+	 * `allowed_redirect_hosts` reports — this site's own host, plus whatever a publisher
+	 * adds through that filter, plus anything another plugin has added (`Newspack\NRH`
+	 * registers one). A destination arriving in a link with no signature has to clear that
+	 * bar, which is what keeps a crafted link from choosing where a reader lands.
+	 *
+	 * @param string $url       The requested destination.
+	 * @param string $signature Signature offered with the destination, if any.
+	 *
+	 * @return string The destination, or an empty string if it is not allowed.
+	 */
+	public static function sanitize_after_success_url( $url, $signature = '' ) {
+		$url = self::normalize_after_success_url( $url );
+
+		if ( empty( $url ) ) {
+			return '';
+		}
+
+		if ( self::is_after_success_url_signed( $url, $signature ) ) {
+			return $url;
 		}
 
 		$validated = (string) wp_validate_redirect( $url, '' );
@@ -1270,7 +1339,8 @@ final class Modal_Checkout {
 			unset(
 				$params['after_success_behavior'],
 				$params['after_success_url'],
-				$params['after_success_button_label']
+				$params['after_success_button_label'],
+				$params['after_success_signature']
 			);
 		}
 
@@ -1290,11 +1360,14 @@ final class Modal_Checkout {
 				\wp_parse_str( $referrer_query, $request_params );
 			}
 		}
+		$signature = isset( $request_params['after_success_signature'] ) ? sanitize_text_field( wp_unslash( $request_params['after_success_signature'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+
 		$params = array_filter(
 			[
 				'after_success_behavior'     => isset( $request_params['after_success_behavior'] ) ? sanitize_text_field( wp_unslash( $request_params['after_success_behavior'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-				'after_success_url'          => isset( $request_params['after_success_url'] ) ? self::sanitize_after_success_url( wp_unslash( $request_params['after_success_url'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				'after_success_url'          => isset( $request_params['after_success_url'] ) ? self::sanitize_after_success_url( wp_unslash( $request_params['after_success_url'] ), $signature ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 				'after_success_button_label' => isset( $request_params['after_success_button_label'] ) ? sanitize_text_field( wp_unslash( $request_params['after_success_button_label'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				'after_success_signature'    => $signature,
 				'action_type'                => isset( $request_params['action_type'] ) ? sanitize_text_field( wp_unslash( $request_params['action_type'] ) ) : '', // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			]
 		);
--- a/plugins/newspack-blocks/includes/class-modal-checkout.php
+++ b/plugins/newspack-blocks/includes/class-modal-checkout.php
@@ -1242,13 +1242,58 @@ final class Modal_Checkout {
 		$pass = isset( $parts['pass'] ) ? ':' . $parts['pass'] : '';
 		$auth = $user ? $user . $pass . '@' : '';
 
-		return ( isset( $parts['scheme'] ) ? $parts['scheme'] . '://' : '' )
+		// A destination with a host but no scheme is protocol-relative. Losing the `//` here
+		// would turn it into a relative path, which changes where the reader lands and makes
+		// this function non-idempotent — and the token signs a normalised value at mint and
+		// normalises again at read, so idempotence is what makes it verify.
+		$prefix = isset( $parts['scheme'] )
+			? $parts['scheme'] . '://'
+			: ( isset( $parts['host'] ) ? '//' : '' );
+
+		return $prefix
 			. $auth
 			. ( $parts['host'] ?? '' )
 			. ( isset( $parts['port'] ) ? ':' . $parts['port'] : '' )
 			. ( $parts['path'] ?? '' )
 			. ( isset( $parts['query'] ) ? '?' . $parts['query'] : '' )
 			. ( isset( $parts['fragment'] ) ? '#' . $parts['fragment'] : '' );
+	}
+
+	/**
+	 * Whether a post is published, read from the stored status.
+	 *
+	 * `get_post_status()` reports an attachment as `publish` whatever its stored status, so a
+	 * token could be minted from one and could not be revoked by unpublishing. Reading the
+	 * field directly returns the real status, and also steps around the `get_post_status`
+	 * filter, which would otherwise let any plugin move this gate.
+	 *
+	 * @param int $post_id The post to check.
+	 *
+	 * @return boolean
+	 */
+	private static function is_published_post( $post_id ) {
+		return 'publish' === get_post_field( 'post_status', (int) $post_id );
+	}
+
+	/**
+	 * The post a block is being rendered for.
+	 *
+	 * `get_the_ID()` is empty outside the loop, which is where a block in a widget or a footer
+	 * template part renders. The queried object covers that, but only when it is a post: on an
+	 * archive it is a term, whose ID would otherwise be read as an unrelated post's.
+	 *
+	 * @return int Post ID, or 0.
+	 */
+	private static function get_after_success_post_id() {
+		$post_id = (int) get_the_ID();
+
+		if ( $post_id ) {
+			return $post_id;
+		}
+
+		$queried = get_queried_object();
+
+		return $queried instanceof WP_Post ? (int) $queried->ID : 0;
 	}
 
 	/**
@@ -1263,7 +1308,17 @@ final class Modal_Checkout {
 	 * @return string
 	 */
 	private static function sign_after_success_payload( $payload ) {
-		return hash_hmac( 'sha256', 'newspack_blocks_after_success|' . $payload, wp_salt( 'auth' ) );
+		// The namespace prefix is what separates this from every other use of the same key and
+		// algorithm. `Reader_Registration::get_frontend_registration_key()` also produces
+		// `hash_hmac( 'sha256', …, wp_salt( 'auth' ) )` and publishes the result to the front
+		// end, so nothing but this prefix distinguishes the two. Never register an integration
+		// ID that begins with it.
+		//
+		// The blog ID is included because `wp_salt()` is network-scoped: without it, a token
+		// minted on one site of a network verifies on every other site in that network.
+		$namespace = 'newspack_blocks_after_success|' . get_current_blog_id() . '|';
+
+		return hash_hmac( 'sha256', $namespace . $payload, wp_salt( 'auth' ) );
 	}
 
 	/**
@@ -1297,9 +1352,9 @@ final class Modal_Checkout {
 			return '';
 		}
 
-		$post_id = $post_id ? (int) $post_id : (int) get_the_ID();
+		$post_id = $post_id ? (int) $post_id : self::get_after_success_post_id();
 
-		if ( ! $post_id || 'publish' !== get_post_status( $post_id ) ) {
+		if ( ! $post_id || ! self::is_published_post( $post_id ) ) {
 			return '';
 		}
 
@@ -1345,7 +1400,7 @@ final class Modal_Checkout {
 
 		// Checked at read time as well as at mint time: a post that has since been
 		// unpublished should stop vouching for its destination.
-		if ( 'publish' !== get_post_status( (int) $claims['p'] ) ) {
+		if ( ! self::is_published_post( (int) $claims['p'] ) ) {
 			return '';
 		}
 
--- a/plugins/newspack-blocks/includes/class-modal-checkout.php
+++ b/plugins/newspack-blocks/includes/class-modal-checkout.php
@@ -377,7 +377,7 @@ final class Modal_Checkout {
 		$after_success_behavior     = filter_input( INPUT_GET, 'after_success_behavior', FILTER_SANITIZE_SPECIAL_CHARS );
 		$after_success_url          = filter_input( INPUT_GET, 'after_success_url', FILTER_SANITIZE_URL );
 		$after_success_button_label = filter_input( INPUT_GET, 'after_success_button_label', FILTER_SANITIZE_SPECIAL_CHARS );
-		$after_success_signature    = filter_input( INPUT_GET, 'after_success_signature', FILTER_SANITIZE_SPECIAL_CHARS );
+		$after_success_token        = filter_input( INPUT_GET, 'after_success_token', FILTER_SANITIZE_SPECIAL_CHARS );
 
 		if ( $variation_id ) {
 			$product_id = $variation_id;
@@ -396,7 +396,7 @@ final class Modal_Checkout {
 			wp_parse_str( $parsed_url['query'], $params );
 		}
 
-		$params = array_merge( $params, compact( 'after_success_behavior', 'after_success_url', 'after_success_button_label', 'after_success_signature' ) );
+		$params = array_merge( $params, compact( 'after_success_behavior', 'after_success_url', 'after_success_button_label', 'after_success_token' ) );
 
 		if ( function_exists( 'wpcom_vip_url_to_postid' ) ) {
 			$referer_post_id = wpcom_vip_url_to_postid( $referer );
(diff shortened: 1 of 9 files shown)


dkoo:
@jason10lee tagging you for a review on this one too!

plugins/newspack-blocks/src/blocks/checkout-button/view.php
```diff
@@ -113,6 +113,9 @@ function render_callback( $attributes ) {
 		$hidden_fields .= $after_success_behavior ? '<input type="hidden" name="after_success_behavior" value="' . esc_attr( $after_success_behavior ) . '" />' : '';
 		$hidden_fields .= $after_success_button_label ? '<input type="hidden" name="after_success_button_label" value="' . esc_attr( $after_success_button_label ) . '" />' : '';
 		$hidden_fields .= $after_success_url ? '<input type="hidden" name="after_success_url" value="' . esc_attr( $after_success_url ) . '" />' : '';
+		// Signed here because this is the last point the destination is known to come from the
+		// block's own settings rather than from the request.
+		$hidden_fields .= $after_success_url ? '<input type="hidden" name="after_success_signature" value="' . esc_attr( Modal_Checkout::get_after_success_url_signature( $after_success_url ) ) . '" />' : '';
```
jason10lee: **Blocker:** A signature can be minted by any user who can edit posts, and it never expires — The comment here says this is the last point the destination is known to come from the block's own settings. That holds for the render call, but not for who can cause a render.

Both blocks are dynamic, so core exposes `/wp/v2/block-renderer/newspack-blocks/checkout-button`, whose permission gate is `edit_posts` when no post ID is supplied, and `afterSuccessURL` is declared as an unconstrained string in `block.json`. Reproduced on a live env: a Contributor — `edit_posts` yes, `publish_posts` no — requested that endpoint with an arbitrary destination, received HTTP 200, read a valid signature out of the returned markup, and that signature made the destination pass validation. No content was created at any point. Draft preview is a second route to the same result.

The token is also bound to nothing but the URL string: it does not expire, is not tied to the user who minted it, and can be handed to anyone. So one account with the lowest content capability can produce a durable credential that lets any crafted checkout link choose where readers land.

This is still much tighter than the state before the base PR, where no authentication was needed at all — so it may be an acceptable trade for a hotfix. It should be a stated decision rather than an inherited one. Binding the signature to the rendering post ID and verifying that ID at the thank-you page would make a minted signature traceable and non-transferable; refusing to sign under `REST_REQUEST` would close the endpoint route.

plugins/newspack-plugin/includes/class-donations.php
```diff
@@ -878,7 +878,10 @@ function ( $item ) {
 		if ( $is_modal_checkout ) {
 			$query_args['modal_checkout'] = 1;
 		}
-		foreach ( [ 'after_success_behavior', 'after_success_button_label', 'after_success_url' ] as $attribute_name ) {
+		// `after_success_signature` travels with the destination it signs: drop it here and a
+		// publisher-configured destination arrives unsigned and gets refused as if a link had
+		// supplied it.
+		foreach ( [ 'after_success_behavior', 'after_success_button_label', 'after_success_url', 'after_success_signature' ] as $attribute_name ) {
```
jason10lee: **Blocker:** The donations path entity-encodes the destination before it is verified — `FILTER_SANITIZE_FULL_SPECIAL_CHARS` HTML-encodes the value, so `&` becomes `&amp;` and `'` becomes `&#039;` in the string itself, not merely in its rendered form. The signature was minted over the raw string, so verification is asked about a different one and the destination is dropped.

Reproduced end to end on a live Donate block: the rendered page emits a signature over the raw URL, and after this filter the verifier computes a different hash and returns an empty string. Any thank-you URL with two query parameters — the normal shape once UTM tags are involved — fails here.

The signature itself is unaffected (hex has no special characters); only the URL is. Sanitizing this key as a URL rather than as HTML would fix it and would match what `process_checkout_request()` already does.

Copilot flagged this too, but as a *suppressed* comment inside the review body rather than an inline one, so it is easy to miss.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1199,33 +1200,101 @@ public static function get_checkout_template( $template ) {
...
 
 		if ( empty( $url ) ) {
 			return '';
 		}
 
-		// Core compares hosts case-sensitively; host names aren't. Normalise first so this
-		// site's own host typed in capitals isn't read as somewhere else.
+		// Core compares hosts case-sensitively; host names aren't. Normalise so this site's
+		// own host typed in capitals isn't read as somewhere else.
 		$host = wp_parse_url( $url, PHP_URL_HOST );
```
jason10lee: **Suggestion:** Host lowercasing rewrites every occurrence and misses userinfo URLs — `str_replace( '://' . $host, '://' . strtolower( $host ), $url )` is unbounded, so it rewrites the authority and any later occurrence — `https://Example.com/go?next=https%3A//Example.com/x` has its query payload lowercased too, and the reader lands somewhere subtly different from what was configured.

It also no-ops on userinfo URLs: for `https://User@Example.com/`, `wp_parse_url()` reports the host as `Example.com` but the literal `://Example.com` never appears in the string, so the capitalization this is meant to fix survives.

Reassembling from `wp_parse_url()` components, or bounding the replacement to the first occurrence, would fix both.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1199,33 +1200,101 @@ public static function get_checkout_template( $template ) {
...
+	 * checkout, and a nonce is bound to a user and a time window.
+	 *
+	 * @param string $url The destination to sign.
+	 *
+	 * @return string The signature, or an empty string if there is nothing to sign.
+	 */
+	public static function get_after_success_url_signature( $url ) {
+		$url = self::normalize_after_success_url( $url );
+
+		return $url ? wp_hash( $url ) : '';
```
jason10lee: **Suggestion:** The signature shares the auth salt with magic-link and email-change secrets — `wp_hash( $url )` is `hash_hmac( 'md5', $url, wp_salt( 'auth' ) )`. That same key and function produce `Magic_Link::generate_secret()` and the email-change confirmation secret in `newspack-plugin`, and this PR publishes such hashes in public HTML on every page carrying one of these blocks.

We checked whether a collision is reachable and it is not: `normalize_after_success_url()` prefixes a scheme, so a destination cannot normalize to a bare email address and the hashes differ. But that separation falls out of `esc_url_raw()`'s behavior rather than from design, and it is the only thing standing between these two uses of the same key.

Hashing a purpose-prefixed value costs one string and removes the class of concern. While changing that line, `wp_hash( $url, 'auth', 'sha256' )` is available — md5 is only the default, and there is no reason to pick it for a new signature.

Separately worth a docblock line: because the signature is baked into cached page HTML, rotating `AUTH_SALT` invalidates every signature still in the page cache until those pages turn over, and the symptom is a silently dropped destination.

plugins/newspack-blocks/src/modal-checkout/templates/thankyou.php
```diff
@@ -47,7 +47,8 @@
 
 $is_success             = ! $order->has_status( 'failed' );
 $after_success_behavior = isset( $_GET['after_success_behavior'] ) ? \sanitize_text_field( \wp_unslash( $_GET['after_success_behavior'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-$after_success_url      = isset( $_GET['after_success_url'] ) ? esc_url( \Newspack_Blocks\Modal_Checkout::sanitize_after_success_url( \sanitize_url( \wp_unslash( $_GET['after_success_url'] ) ) ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+$after_success_sig      = isset( $_GET['after_success_signature'] ) ? \sanitize_text_field( \wp_unslash( $_GET['after_success_signature'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+$after_success_url      = isset( $_GET['after_success_url'] ) ? esc_url( \Newspack_Blocks\Modal_Checkout::sanitize_after_success_url( \sanitize_url( \wp_unslash( $_GET['after_success_url'] ) ), $after_success_sig ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
jason10lee: **Suggestion:** Display escaping applied into a variable that is later re-escaped — `esc_url()` wraps the verified value here, but the variable is consumed at `:90` inside `esc_attr()`, so the escaping is applied twice and the variable no longer holds the string that was verified. It works today because the entities round-trip through the HTML parser and `getAttribute()`, but it means anything downstream that re-compares or re-signs would break, which is the exact failure mode this PR is guarding against.

The outer `sanitize_url()` is also redundant — `normalize_after_success_url()` calls it itself — and having two callers each do their own normalization obscures which side owns it. Dropping both and letting `esc_attr()` at the point of output do the escaping would leave one normalizer and one escape.

The escaped value is also what gets handed to `filter_after_success_params()` just below, so the "is there a destination here" check runs against display-escaped data.

plugins/newspack-blocks/tests/test-after-success-url.php
```diff
@@ -132,6 +132,127 @@ function ( $url ) use ( &$seen ) {
...
+	 */
+	public function test_signing_survives_sanitising( $url ) {
+		$signature = \Newspack_Blocks\Modal_Checkout::get_after_success_url_signature( $url );
+
+		// What the page receives has been through sanitising on the way.
+		$in_transit = sanitize_url( $url );
+
+		$this->assertNotSame(
+			'',
+			\Newspack_Blocks\Modal_Checkout::sanitize_after_success_url( $in_transit, $signature ),
```
jason10lee: **Suggestion:** The drift test models transit as the one transform that cannot break it — `test_signing_survives_sanitising()` sets `$in_transit = sanitize_url( $url )`, but `normalize_after_success_url()` calls `sanitize_url()` on both the signing and verifying sides — so the assertion cannot fail regardless of what the real pipeline does. The data provider already contains the two destination shapes that break in production (`'a query string'` and `'an encoded space'`), and the test reports both as passing.

That makes it worse than missing coverage: it is coverage that certifies the broken cases as working.

A test that would catch it has to reproduce the actual chain — `FILTER_SANITIZE_URL` then `sanitize_text_field()` for the Checkout Button path, `FILTER_SANITIZE_FULL_SPECIAL_CHARS` for the Donate path — or better, drive `process_checkout_request()` and the donations handler directly, so the test does not drift the next time transit changes.

plugins/newspack-plugin/includes/class-donations.php
```diff
@@ -878,7 +878,10 @@ function ( $item ) {
 		if ( $is_modal_checkout ) {
 			$query_args['modal_checkout'] = 1;
 		}
-		foreach ( [ 'after_success_behavior', 'after_success_button_label', 'after_success_url' ] as $attribute_name ) {
+		// `after_success_signature` travels with the destination it signs: drop it here and a
```
jason10lee: **Suggestion:** Release skew between the two plugins fails silently — The donate block's destination only survives because this array was extended, so a site that takes the `newspack-blocks` hotfix without the matching `newspack-plugin` one refuses every donate-block off-site destination with no error surfaced — the reader just sees the ordinary close button.

The forwarding itself is correctly defensive: pure query-param passthrough with no coupling to `Newspack_Blocks`, so nothing fatals. It is the ordering that matters. Worth stating in the PR body that these two ship together, and worth naming `newspack_blocks_after_success_url_rejected` in the release note as the signal to watch if they do not.

plugins/newspack-blocks/includes/class-modal-checkout.php
```diff
@@ -1199,33 +1200,101 @@ public static function get_checkout_template( $template ) {
...
+	 * @return string The destination, or an empty string if it is not allowed.
+	 */
+	public static function sanitize_after_success_url( $url, $signature = '' ) {
+		$url = self::normalize_after_success_url( $url );
+
+		if ( empty( $url ) ) {
+			return '';
+		}
+
+		if ( self::is_after_success_url_signed( $url, $signature ) ) {
```
jason10lee: **Nit:** The signed and unsigned branches return differently shaped strings — The unsigned branch returns `wp_validate_redirect()`'s output, which has been through `wp_sanitize_redirect()`; the signed branch returns the raw normalized string. One function, two output contracts. Low risk given `sanitize_url()` already removes the dangerous characters, but running the signed branch through `wp_sanitize_redirect()` as well would make the return value uniform.

jason10lee (changes requested):
The design is the right call and we'd like to see it land — it removes a permanent support burden the allowlist approach would have created, and signing at render neatly avoids a migration. We've identified some blockers, but it looks like you'd anticipated them, so what follows is probably mostly familiar territory.

**Four blockers.** Two are inline comments on the diff; two have no line in this diff to attach to, so they're written out below.

1. **The signed destination is mutated in transit**, so verification fails for ordinary destinations — see *Normalization drift* below.
2. **The donations path entity-encodes the destination** before it is verified — `plugins/newspack-plugin/includes/class-donations.php:884`.
3. **A signature can be minted by any user who can edit posts**, and it never expires — `plugins/newspack-blocks/src/blocks/checkout-button/view.php:118`.
4. **The variation-picker path drops the signature** — see the notes at the end.

**Normalization drift — you called this before writing the code, and the fix lands one layer too high.** The design note even predicted the failing input: a destination like `?utm_campaign=don't`. That case does fail in our testing. So do an encoded space and a non-ASCII path. The remedy — normalize on both sides through one helper — is exactly right for the boundary you were looking at, but the mutation happens in transport rather than in the helper's reach: `sanitize_text_field()` deletes percent-encoded octets, `FILTER_SANITIZE_URL` drops non-ASCII bytes, `FILTER_SANITIZE_FULL_SPECIAL_CHARS` HTML-encodes `&` and `'`, and `add_query_arg()` assembles without `rawurlencode()`. Both ends of the pipe are hardened and the leak is in the middle.

The test inherited the same framing: it models transit as `sanitize_url()`, the one transform inside the helper's own control, so the data provider contains the breaking shapes and reports them green.

**On whether there's a 'yes' to the trust boundary, since you asked.** Our answer is no, not as it stands. The boundary is `edit_posts`, not publish: a Contributor who cannot publish anything can obtain a valid signature for a destination of their choosing without creating any content, because the block renderer's REST endpoint is gated on `edit_posts` and the destination attribute is an unconstrained string. And the resulting token is bound to nothing but the URL — it does not expire, is not tied to the account that produced it, and can be handed to anyone. It is still far tighter than the pre-#782 state, which is a fair argument for accepting it in a hotfix; what tips it for us is the transferability, not the capability. Binding the signature to the rendering post ID and checking that ID at the thank-you page would make it traceable and non-transferable, and would settle this without changing the shape of the design.

- `plugins/newspack-blocks/src/modal-checkout/modal.js` (~L373–380) — **Blocker:** the variation-picker path drops the signature, which is the one genuine miss here. There are two lists that copy hidden context fields into a checkout form; this PR updated `PICKER_CONTEXT_FIELDS` and not the one in `modal.js`, so a variable or grouped product loses the signature — the exact case the PR exists to fix. Worth having both import one constant so they cannot drift again.
- `plugins/newspack-blocks/includes/class-modal-checkout.php:1310` — **Suggestion:** the rejection hook diverges from this file's naming convention. Every other modal-checkout hook in this file uses the `newspack_blocks_modal_checkout_*` segment, matching the plugin's own convention for modal-specific hooks. This one is `newspack_blocks_after_success_url_rejected`. Hooks are a public surface that cannot be renamed after release, and this one is about to be recommended as the diagnostic for release skew, so it will get adopted quickly. Worth settling before it ships.

None of this is an argument for shipping the base PR on its own: we agree that it's probably better to stretch for the complete fix to avoid increasing support burden.

plugins/newspack-blocks/src/blocks/checkout-button/view.php
dkoo: Confirmed and fixed in f4bf13e.

I reproduced the mechanism locally: the block-renderer route returned **200** for a Contributor (`edit_posts` yes, `publish_posts` no), and `afterSuccessURL` is `{"type":"string"}` in `block.json` with no constraint. I could not complete the final step — that env has no WooCommerce, so the block renders empty — so the end-to-end result is yours, not independently reproduced here.

Minting is now bound to a **published post**. `get_after_success_token()` takes the post being rendered and returns an empty string unless `get_post_status()` is `publish`, so the REST render (no post) and draft preview (draft/pending/private) both mint nothing. The publish check runs again at read time, so unpublishing a post revokes the tokens it minted.

Tokens are still transferable between readers, which is inherent to a bearer value in cached HTML — but obtaining one now requires publishing, and the token names the post it came from, so it is traceable.

Covered by `test_does_not_vouch_from_an_unpublished_post` (draft / pending / private), `test_does_not_vouch_without_a_post`, and `test_refuses_a_token_whose_post_was_unpublished`.

plugins/newspack-plugin/includes/class-donations.php
dkoo: Confirmed and fixed in f4bf13e.

Measured it directly rather than taking it on faith — `FILTER_SANITIZE_FULL_SPECIAL_CHARS` turns `?a=1&b=2` into `?a=1&amp;b=2` in the value itself, and the signature then verifies against a different string. Any destination with two query parameters failed.

Two changes. The URL key now uses `FILTER_SANITIZE_URL` rather than the HTML filter, matching what `process_checkout_request()` already does. And more fundamentally, the destination now travels **inside** the token as URL-safe base64, so this filter has nothing to mutate: I checked `FULL_SPECIAL_CHARS`, `FILTER_SANITIZE_URL` and `sanitize_text_field` against the token and all three leave it byte-identical.

Thanks for noting Copilot had this suppressed in the body — that is easy to miss.

plugins/newspack-blocks/includes/class-modal-checkout.php
dkoo: Confirmed and fixed in f4bf13e.

Both cases were real: the unbounded `str_replace` rewrote a host appearing again in a query parameter, and `user@Host` never matched `://Host` so the capitalisation survived.

Normalisation now reassembles from `wp_parse_url()` parts, so only the authority is touched. `test_normalisation_lowercases_only_the_host` pins both: a query payload carrying the host is left alone, and a userinfo host is lowercased.

plugins/newspack-blocks/includes/class-modal-checkout.php
dkoo: Confirmed and fixed in f4bf13e.

I checked collision reachability before changing anything and agree it is not reachable — `sanitize_url()` prepends `http://`, so a destination cannot normalise to a bare email address, and the `mailto:` form keeps its scheme. Both hash differently from `wp_hash( <email> )`. But as you say, that separation is a side effect of `esc_url_raw()`, not a decision.

Now `hash_hmac( "sha256", "newspack_blocks_after_success|" . $payload, wp_salt( "auth" ) )` — namespaced and sha256, so it cannot produce the same digest as `Magic_Link::generate_secret()` or the email-change secret regardless of input.

The salt-rotation note is on `get_after_success_token()`: rotating `AUTH_SALT` invalidates tokens still sitting in cached HTML until those pages turn over, and the symptom is a silently dropped destination.

plugins/newspack-blocks/src/modal-checkout/templates/thankyou.php
dkoo: Confirmed and fixed in f4bf13e.

`esc_url()` into the variable is gone; escaping happens at output via `esc_attr()`, so the variable holds the destination that was actually verified rather than a display-escaped copy of it. That also means `filter_after_success_params()` below now runs its "is there a destination here" check against the real value.

I kept the `sanitize_url()` on input. Dropping it makes PHPCS flag the `$_GET` read as unsanitised, since it cannot see through the helper — so it stays as the visible input sanitiser, and `normalize_after_success_url()` remains the single normaliser.

plugins/newspack-blocks/tests/test-after-success-url.php
dkoo: Confirmed and rewritten in f4bf13e. This was the most useful comment in the review.

You are right that it was worse than missing coverage. `$in_transit = sanitize_url( $url )` is the one transform the normaliser applies on both sides, so the assertion could not fail, and the provider contained the two shapes that break in production and reported them green. I predicted this failure mode in the design note and then wrote a test that could not detect it.

`test_destination_survives_real_transit` now runs each destination through the actual transforms — `FILTER_SANITIZE_URL`, `sanitize_text_field`, `FULL_SPECIAL_CHARS`, and an `esc_attr` round trip — and asserts the destination comes back intact. The provider gained the two shapes that were failing (`an apostrophe`, `a non-ASCII path`) alongside the ones already there.

It earned its keep immediately: it failed on first run and caught a regression introduced by another suggestion in this review. See the reply on the return-shape nit.

plugins/newspack-plugin/includes/class-donations.php
dkoo: Agreed, and noted in f4bf13e.

The ordering constraint is now stated in the PR body: the two plugins ship together, and `newspack_blocks_modal_checkout_after_success_url_rejected` is the signal to watch if they do not. I will carry the same note into the release note.

Worth adding that the skew is now narrower than it was. The destination travels inside the token, so a site running the newer `newspack-blocks` against an older `newspack-plugin` loses the token on the donate path and falls back to allowlist-only — which is exactly the pre-#791 behaviour, not a new failure mode.

plugins/newspack-blocks/includes/class-modal-checkout.php
dkoo: Tried this and backed it out — leaving the thread open since it is your call, not mine.

I applied `wp_sanitize_redirect()` to the vouched branch as suggested, and the rewritten drift test failed immediately on two shapes:

```
?utm_campaign=don't   ->  ?utm_campaign=dont
/gracias-señor/       ->  /gracias-se%C3%B1or/
```

`wp_sanitize_redirect()` strips `'`, `$`, `|`, `{`, `}` and percent-encodes non-ASCII. On the untrusted branch that is exactly right. On the vouched branch it rewrites a destination this site has already established came from its own content — which is the drift the token exists to prevent, reintroduced at the last step. The apostrophe case is the sharp one: it changes a query parameter's value, so the reader reaches a different page than the publisher configured.

So the two branches do still return differently shaped strings, deliberately. Both have been through `sanitize_url()` in normalising, which removes the dangerous schemes and characters; the untrusted branch additionally gets `wp_sanitize_redirect()` via `wp_validate_redirect()`, because it has not been established as ours. There is a comment at the return explaining why.

If you would rather have the uniform contract and accept the rewriting, say so and I will put it back — but I did not want to trade a correctness property for a shape property without flagging it.

dkoo:
Thanks — this was a genuinely useful review. All four blockers reproduced locally before I touched anything, and everything is addressed in f4bf13e. Seven inline threads are answered and resolved; the return-shape nit I have replied to but left open, because I am declining it and that should be your call.

Two items from the review body had no line to attach to, so answering them here.

### Variation picker dropping the token — fixed

Confirmed: `modal.js` carried its own literal copy of the field list, so the exported `PICKER_CONTEXT_FIELDS` and the variation path had already drifted before this PR added to one of them. Both now use the single exported constant, with a comment saying why, so a field added to one cannot go missing from the other.

While checking, I looked at the third place that assembles these params — `window.newspackOpenModalCheckout` around L959 — and deliberately left it alone. That is a public JS API where a caller supplies a destination from JavaScript, with no way to sign it client-side, so falling through to `wp_validate_redirect()` is the correct behaviour there rather than a miss.

### Hook naming — renamed, and the documentation was wrong too

You are right, and the reason it went wrong is worth recording. `AGENTS.md` documented the convention as `newspack_modal_checkout_*`, which is what I followed — but the file itself uses `newspack_blocks_modal_checkout_*` for six of its seven modal hooks, with `newspack_modal_checkout_success_text` the lone exception. The documentation was describing the minority form.

The hook is now `newspack_blocks_modal_checkout_after_success_url_rejected`, and I corrected both places in `AGENTS.md` so the next person reads the convention the code actually follows. Settled before release, as you suggested — it has not shipped.

### One thing I changed my mind about mid-review

The rewritten drift test failed on its first run and caught a regression introduced by the return-shape nit in this same review: running the vouched branch through `wp_sanitize_redirect()` turns `?utm_campaign=don't` into `dont` and percent-encodes non-ASCII paths. Details in that thread. That is a fair argument for the test rewrite having been the most valuable item here.

95 PHPUnit tests and 42 JS tests pass; PHPCS and ESLint clean.

jason10lee (changes requested):
All four blockers from the last round are closed — re-verified in an isolated environment with WooCommerce and Newspack donations, re-running the original reproductions. One of the fixes introduced a new blocker, though, and it's one we caused: the `wp_parse_url` reassembly we asked for drops the `//` on a protocol-relative destination when the host has uppercase. Detail below, after what's confirmed fixed.

- The Contributor mint path returns **no token** — same request that produced one before.
- Draft, private and no-post renders mint nothing; unpublishing a post revokes tokens it already issued; a tampered payload under a valid signature is refused.
- A real Donate block on a published page round-trips a destination carrying `&` intact, through the donations path.
- The variation-picker path now imports the shared field list, so the two can't drift again.

Moving the destination inside the token was the better call — it makes the value opaque to transit rather than trying to make four sanitisers agree. Worth noting the decode order too: the signature is checked before the payload is decoded, so the base64/JSON parser never sees unauthenticated input.

**The new blocker: the parts-rebuild loses protocol-relative destinations**

`class-modal-checkout.php`, the `build_url()` scheme branch. Our own idempotence check missed it because we tested a lowercase host, and uppercase is the trigger.

`//ELSEWHERE.example.test/thanks` normalises to `elsewhere.example.test/thanks`, a relative path. Measured:

| | result |
|---|---|
| vouched, no port | read back as `http://…` — HTTPS quietly downgraded |
| vouched, with port | read back as `''` — silently refused |
| unvouched | returns `/checkout/elsewhere.example.test/thanks`, an on-site 404 |
| unvouched | the rejection hook does **not** fire |

That last row is the one that bothers us most: this shape slips past the hook that exists to make silent refusals visible. It also makes `normalize()` non-idempotent, and the token scheme depends on that property, since normalisation runs at both mint and read. A `: ( isset( $parts['host'] ) ? '//' : '' )` fallback in the scheme branch covers it. The current provider only carries the lowercase form, which is why it passes today.

**On the open thread you left**

Keep the asymmetry. Applying `wp_sanitize_redirect()` to the vouched branch would rewrite a destination the site has already established is the publisher's own, and the apostrophe case changes where the reader actually lands. The rewritten test catching that is the test doing its job.

**Re-raising one suggestion from last round, at the same severity**

The donations forwarding list is still positional (`class-donations.php:884`), and the parameter changed name this round. Nothing about the risk has changed since we raised it — the note you added is the right thing to have written, it just doesn't bind anything at release time. We're re-raising the fix rather than the severity: `class-modal-checkout.php:474` already handles the identical problem with a prefix match on `after_success`, and the same approach here would make the coupling skew-proof for this rename and the next one, instead of relying on release ordering being remembered. Still your call whether that's worth doing now or later.

**Smaller things that don't block**

`get_post_status()` reports an attachment as `publish` whatever its stored status, so a token can mint from one and cannot be revoked by unpublishing — only by trashing it. `get_post_field( 'post_status', … )` plus a post-type check closes it, and also sidesteps the `get_post_status` filter, which currently lets any plugin move this gate.

The namespace prefix is doing more work than it looks. `Reader_Registration::get_frontend_registration_key()` produces `hash_hmac( 'sha256', $id, wp_salt( 'auth' ) )` — same algorithm, same salt, same 64-hex shape as a token signature — and that key is emitted to the front end as non-secret. The prefix is the only thing separating them, so it's worth a comment saying so, and worth never registering an integration ID that begins with it.

On multisite `wp_salt()` is network-scoped and the token carries no blog binding, so a token minted on one site verifies on another; folding `get_current_blog_id()` into the namespace fixes it. We couldn't reproduce this — the test environment is single-site — so it rests on core behaviour rather than a run.

Two tests can't fail as written. `test_destination_survives_real_transit` varies six destination shapes, but the shapes no longer travel — the token alphabet is inert to all four transforms, so twenty-four assertions collapse to one property. Adding `//ELSEWHERE.example.test/thanks` would give it something real to catch: it fails today. And the first assertion of `test_normalisation_lowercases_only_the_host` passes against the old implementation too, because the query holds `%3A//` rather than `://`; a literal scheme in the query discriminates. The userinfo assertion beside it is genuinely load-bearing — keep that one.

Finally, `get_the_ID()` is empty on 404 and empty-search pages, so a block in a widget or footer template part mints nothing there and the destination is dropped with no signal. `get_queried_object_id()` as a fallback covers most of it.

We spent real effort trying to break the token itself and didn't get anywhere. Forgery, signature juggling, payload swapping, claims type-juggling, minting below publish capability and scheme-based redirect abuse all failed closed, and `javascript:`, `data:` and `vbscript:` normalise away before they reach the redirect — so the security core holds, and what's left is robustness at the edges.

dkoo:
Fixed in 2fa64cc9. 106 PHPUnit tests, 42 JS tests, PHPCS and ESLint clean.

### The blocker — confirmed, including a consequence past the ones you listed

Reproduced all four rows. One more worth recording: **`normalize_after_success_url()` was not idempotent** for this shape. `//ELSEWHERE.example.test/thanks` normalises to `elsewhere.example.test/thanks`, and normalising that again leaves it unchanged but different from the first result. Since the token signs a normalised value at mint and normalises again at read, that property is what makes verification work at all, so this was closer to the core than the four symptoms suggest.

Took the `//` fallback you proposed. The comment at that branch now says why it exists, in idempotence terms rather than just "protocol-relative".

Also confirmed the row that bothered you most: the unvouched case returned `/elsewhere.example.test/thanks` from `wp_validate_redirect()`, so it was **accepted** as an on-site path and the rejection hook never fired.

### Tests

Both weak ones addressed, plus one more.

- The transit provider gained `//ELSEWHERE.example.test/thanks` and a port variant. You were right that the six existing shapes exercise one property — the token is inert to all four transforms — so these are the two entries that can vary the outcome.
- The normalisation assertion now uses a literal `://` in the query rather than `%3A//`. As written it passed against the old implementation too, exactly as you said. The userinfo assertion beside it is untouched.
- Added `test_normalisation_is_idempotent` across the same provider, since that is the property the token scheme rests on and nothing asserted it directly.
- Added `test_announces_a_refused_protocol_relative_destination`, because a shape slipping past the rejection hook is the failure mode that stays invisible.

Control check: reverting just the scheme branch fails five of these and passes everything else.

### Suggestions taken

- **Forwarding list** — now a prefix match over the `after_success_*` params, mirroring `class-modal-checkout.php:474`. Agreed the note I added did not bind anything; this does.
- **`get_post_field( post_status, … )`** — confirmed `get_post_status()` reports an attachment as `publish` while the stored status is `inherit`, so a token could mint from one and survive unpublishing. Also sidesteps the `get_post_status` filter, as you noted.
- **Blog binding** — confirmed `wp_salt( auth )` is identical across subsites on a real network, so a token verified anywhere in it. `get_current_blog_id()` is now part of the namespace.
- **Key-shape collision** — `class-reader-registration.php:250` is exactly as you describe. There is now a comment at the signing function saying the prefix is the only separation, and that no integration ID may begin with it.

### One refinement rather than a straight application

For `get_the_ID()` being empty outside the loop, I did not take `get_queried_object_id()` directly. On an archive it returns a **term** ID, which `get_post_field()` would then read as an unrelated post — minting a token bound to something arbitrary. The fallback now only accepts the queried object when it is a `WP_Post`, which still covers the widget and footer-template-part case you were aiming at, and mints nothing on archives and 404s.

Thanks for the note on decode order, and for confirming the asymmetry on the vouched branch should stay — I have resolved that thread.

jason10lee (approved):
Approving. Re-verified against `2fa64cc9f` in an isolated environment, re-running the reproductions rather than reading the diff: the protocol-relative shape now keeps its `//`, lowercases only the host, normalises idempotently, and round-trips through a token without the HTTPS downgrade. The port variant that previously came back empty round-trips intact. Unvouched, it's refused and the rejection hook fires. An attachment no longer mints, and the forwarding list, blog binding and namespace comment all landed.

Two of these came back much improved:

- The `get_queried_object_id()` fallback we suggested was wrong, and the refinement catches why: on an archive it returns a term ID, which `get_post_field()` would then read as an unrelated post and mint a token bound to something arbitrary. Gating on `WP_Post` keeps the widget and template-part case we were aiming at and mints nothing on archives and 404s. Good that it didn't go in as written.

- The forwarding change also went further than the prefix match we pointed at. Choosing the filter per key — `FILTER_SANITIZE_URL` for the destination, `FULL_SPECIAL_CHARS` for the rest — fixes the entity-encoding at the source instead of relying on the token to route around it, so the raw `after_success_url` stays usable for anything that reads it later.

Nothing outstanding from our side. 

For ordering: this merges into #782, then #782 into `release`, and the six per-site `allowed_redirect_hosts` entries tracked in NPPM-3080 stay unnecessary as long as the stack lands together.

dkoo:
Thanks for the re-test and re-review, @jason10lee! Merging this into #782 now.

PR #793, opened 2026-08-04
fix(reader-activation): reserve internal user meta keys on registration

rbcorrales:
### All Submissions:

The frontend registration endpoint saved every key of the request's `metadata` object as user meta. It now skips the keys Newspack and WordPress rely on for reader state and account state, and saves the rest as before.

Keys outside the reserved set are unaffected. An integration that was writing a `np_`- or `newspack_`-prefixed key will have it dropped from now on, reported in a new `skipped_metadata_keys` response field and logged once per request. No caller in this repo or in `repos/` writes one, but publisher custom-code repos are outside this workspace. More detail in the ticket.

Closes NPPM-3060.

1. Register a test integration through the `newspack_frontend_registration_integrations` filter. Its key is published in the page's `newspack_ras_config`.
2. Register a reader through that integration, passing a metadata key of your own. The key is saved as user meta.
3. Repeat with a key starting `np_` or `newspack_`. The response lists it under `skipped_metadata_keys`, no meta row is written for it, and the account still requires the normal email verification before My Account access.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
+++ b/plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
@@ -106,7 +106,8 @@ final class Reader_Registration {
 	/**
 	 * Sanitize the metadata parameter.
 	 *
-	 * Ensures all keys and values are sanitized strings.
+	 * Ensures all keys and values are sanitized strings. If this ever accepts
+	 * arrays, revisit {@see is_reserved_meta_key()}.
 	 *
 	 * @param array $metadata Raw metadata from the request.
 	 * @return array Sanitized metadata.
@@ -123,6 +124,61 @@ final class Reader_Registration {
 			}
 		}
 		return $sanitized;
+	}
+
+	/**
+	 * Whether a user meta key is reserved and therefore not caller-writable.
+	 *
+	 * Covers Newspack reader state and reader data, the identity key Jetpack reads,
+	 * and WordPress account state. Other code trusts these values, so registration
+	 * metadata must not set them. Add to the list when adding a key that gates
+	 * anything.
+	 *
+	 * @param string $meta_key Meta key to check.
+	 * @return bool True if the key is reserved.
+	 */
+	public static function is_reserved_meta_key( $meta_key ): bool {
+		global $wpdb;
+
+		$key = strtolower( trim( (string) $meta_key ) );
+		if ( '' === $key ) {
+			return true;
+		}
+
+		$reserved_prefixes = [ 'np_', '_np_', 'newspack_', '_newspack_' ];
+		foreach ( $reserved_prefixes as $prefix ) {
+			if ( 0 === strpos( $key, $prefix ) ) {
+				return true;
+			}
+		}
+
+		$reserved_keys = [
+			'wpcom_user_id', // Jetpack resolves subscriptions against this ID, not the local user ID.
+			'session_tokens',
+			'_application_passwords',
+			'wp_capabilities',
+			'wp_user_level',
+			'wp_user-settings',
+			'wp_user-settings-time',
+			'default_password_nag',
+		];
+		if ( in_array( $key, $reserved_keys, true ) ) {
+			return true;
+		}
+
+		// Table-prefixed, with a per-blog segment on multisite (wp_2_capabilities).
+		// Match the site's prefix and the default, so this holds on any configuration.
+		$prefixes = [ 'wp_' ];
+		if ( isset( $wpdb->base_prefix ) ) {
+			$prefixes[] = strtolower( $wpdb->base_prefix );
+		}
+		foreach ( array_unique( $prefixes ) as $prefix ) {
+			if ( preg_match( '/^' . preg_quote( $prefix, '/' ) . '(\d+_)?(capabilities|user_level)$/', $key ) ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	/**
@@ -455,12 +511,32 @@ final class Reader_Registration {
 			);
 		}
 
-		// Save arbitrary user metadata.
+		// Save arbitrary user metadata, skipping keys other code trusts.
 		$user_metadata = $request->get_param( 'metadata' );
+		$skipped_keys  = [];
 		if ( ! empty( $user_metadata ) ) {
 			foreach ( $user_metadata as $meta_key => $meta_value ) {
+				if ( self::is_reserved_meta_key( $meta_key ) ) {
+					$skipped_keys[] = $meta_key;
+					continue;
+				}
 				\update_user_meta( $result, $meta_key, $meta_value );
 			}
+		}
+
+		if ( ! empty( $skipped_keys ) ) {
+			// Once per request, not per key: the caller controls how many it sends.
+			// newspack_log() fires regardless of NEWSPACK_LOG_LEVEL.
+			Logger::newspack_log(
+				'newspack_frontend_registration_reserved_metadata',
+				'Frontend registration skipped reserved metadata keys.',
+				[
+					'integration_id' => $integration_id,
+					'count'          => count( $skipped_keys ),
+					'keys'           => array_slice( $skipped_keys, 0, 20 ),
+				],
+				'warning'
+			);
 		}
 
 		$response_data = [
@@ -468,6 +544,11 @@ final class Reader_Registration {
 			'status'  => 'created',
 			'email'   => $email,
 		];
+
+		// So an integration author can tell a saved key from a discarded one.
+		if ( ! empty( $skipped_keys ) ) {
+			$response_data['skipped_metadata_keys'] = $skipped_keys;
+		}
 
 		// Surface verification state so integration callers (via window.newspackReaderActivation.register())
 		// can opt into the post-registration verification modal when their UX warrants it. Callers that
--- a/plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
+++ b/plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
@@ -129,51 +129,77 @@ final class Reader_Registration {
 	/**
 	 * Whether a user meta key is reserved and therefore not caller-writable.
 	 *
-	 * Covers Newspack reader state and reader data, the identity key Jetpack reads,
-	 * and WordPress account state. Other code trusts these values, so registration
-	 * metadata must not set them. Add to the list when adding a key that gates
-	 * anything.
+	 * Covers Newspack reader state and reader data, the identifiers other systems
+	 * resolve records against, and WordPress account state. Other code trusts these
+	 * values, so registration metadata must not set them. Add to the list when adding
+	 * a key that gates anything.
 	 *
-	 * @param string $meta_key Meta key to check.
+	 * Normalizes the key itself, so it is safe to call with a raw request key.
+	 *
+	 * @param int|string $meta_key Meta key to check. Numeric array keys arrive as int.
 	 * @return bool True if the key is reserved.
 	 */
 	public static function is_reserved_meta_key( $meta_key ): bool {
-		global $wpdb;
-
-		$key = strtolower( trim( (string) $meta_key ) );
+		$key = \sanitize_key( \wp_unslash( (string) $meta_key ) );
 		if ( '' === $key ) {
 			return true;
 		}
 
 		$reserved_prefixes = [ 'np_', '_np_', 'newspack_', '_newspack_' ];
 		foreach ( $reserved_prefixes as $prefix ) {
-			if ( 0 === strpos( $key, $prefix ) ) {
+			if ( str_starts_with( $key, $prefix ) ) {
 				return true;
 			}
 		}
 
 		$reserved_keys = [
-			'wpcom_user_id', // Jetpack resolves subscriptions against this ID, not the local user ID.
+			// Other systems' identifiers, read via get_user_option(), which falls back
+			// to the unprefixed key. WooPayments needs all three.
+			'wpcom_user_id',
+			'_stripe_customer_id',
+			'_wcpay_customer_id',
+			'_wcpay_customer_id_live',
+			'_wcpay_customer_id_test',
+
 			'session_tokens',
 			'_application_passwords',
-			'wp_capabilities',
-			'wp_user_level',
-			'wp_user-settings',
-			'wp_user-settings-time',
 			'default_password_nag',
 		];
+
+		/**
+		 * Filters additional user meta keys that registration metadata may not write.
+		 *
+		 * Additive only: the keys above are always reserved and cannot be removed.
+		 *
+		 * @param string[] $keys Additional reserved meta keys.
+		 */
+		$reserved_keys = array_merge( $reserved_keys, (array) \apply_filters( 'newspack_reserved_registration_meta_keys', [] ) );
+
 		if ( in_array( $key, $reserved_keys, true ) ) {
 			return true;
 		}
 
-		// Table-prefixed, with a per-blog segment on multisite (wp_2_capabilities).
-		// Match the site's prefix and the default, so this holds on any configuration.
+		return self::is_prefixed_account_key( $key );
+	}
+
+	/**
+	 * Whether a key is one of the table-prefixed WordPress account keys.
+	 *
+	 * Prefixed, with a per-blog segment on multisite (wp_2_capabilities). Matches the
+	 * site's prefix and the default, so this holds on any configuration.
+	 *
+	 * @param string $key Sanitized meta key.
+	 * @return bool True if the key is a prefixed account key.
+	 */
+	private static function is_prefixed_account_key( string $key ): bool {
+		global $wpdb;
+
 		$prefixes = [ 'wp_' ];
 		if ( isset( $wpdb->base_prefix ) ) {
-			$prefixes[] = strtolower( $wpdb->base_prefix );
+			$prefixes[] = \sanitize_key( $wpdb->base_prefix );
 		}
 		foreach ( array_unique( $prefixes ) as $prefix ) {
-			if ( preg_match( '/^' . preg_quote( $prefix, '/' ) . '(\d+_)?(capabilities|user_level)$/', $key ) ) {
+			if ( preg_match( '/^' . preg_quote( $prefix, '/' ) . '(\d+_)?(capabilities|user_level|user-settings|user-settings-time)$/', $key ) ) {
 				return true;
 			}
 		}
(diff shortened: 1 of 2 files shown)


plugins/newspack-plugin/tests/unit-tests/reader-registration-endpoint.php
```diff
@@ -888,4 +888,151 @@ static function ( $limit, $ip, $bucket = 'registration' ) {
...
+				'integration_id'  => self::$integration_id,
+				'integration_key' => self::generate_key( self::$integration_id ),
+				'metadata'        => [
+					'np_reader_email_verified'           => '1',
+					'_np_reader_email_verified'          => '1',
+					'newspack_reader_data_item_is_donor' => 'injected-reader-data',
+					'newspack_reader_data_keys'          => 'injected-key-list',
+					'_newspack_group_subscription'       => 'injected-subscription',
+					'wp_capabilities'                    => 'administrator',
+					$wpdb->base_prefix . 'capabilities'  => 'administrator',
```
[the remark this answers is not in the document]
rbcorrales: Fixed in b80d4348. The array now carries only `$wpdb->get_blog_prefix() . "capabilities"`, the key that actually holds this install's roles, hoisted into a `$caps_key` variable so the arrows still line up. The hardcoded `wp_` and multisite `wp_2_` forms stay covered by `test_is_reserved_meta_key()`.

Worth noting the collision was latent rather than active: the test suite runs with the `wptests_` prefix, so the two entries were distinct keys here. It would have collapsed on any install using the default prefix, which is the fragility you spotted.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -125,6 +126,61 @@ public static function sanitize_metadata( $metadata ) {
...
+	 *
+	 * Covers Newspack reader state and reader data, the identity key Jetpack reads,
+	 * and WordPress account state. Other code trusts these values, so registration
+	 * metadata must not set them. Add to the list when adding a key that gates
+	 * anything.
+	 *
+	 * @param string $meta_key Meta key to check.
+	 * @return bool True if the key is reserved.
+	 */
+	public static function is_reserved_meta_key( $meta_key ): bool {
```
jason10lee, line 142: **Suggestion:** Guard is safe only because its caller already sanitized the key — `is_reserved_meta_key()` normalises with `strtolower( trim( … ) )`, which does not strip backslashes, while `update_metadata()` runs `wp_unslash()` on the meta key before writing. On the current path that is harmless — `sanitize_metadata()` applies `sanitize_key()` at line 121 before the guard ever sees the key, and I confirmed at runtime that a backslash variant normalises into the reserved set and is rejected.

The reason it is worth tightening anyway is that this is a `public static` method with a general-purpose name and a docblock inviting additions, so a future caller could reasonably pass it a raw request key and get `false` for something that lands reserved. Having it do `sanitize_key( wp_unslash( (string) $meta_key ) )` itself makes it self-contained at no cost — or state the precondition in the docblock, the way it already does for arrays.

jason10lee, line 142: **Suggestion:** Document $meta_key as int|string — There is no parameter type on `$meta_key`, and the `(string)` cast on the next lines is load-bearing rather than defensive: PHP coerces numeric array keys to `int`, so the `foreach` really can hand this method an integer. `@param int|string $meta_key` documents that honestly — a bare `string` would be wrong.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -125,6 +126,61 @@ public static function sanitize_metadata( $metadata ) {
...
+		global $wpdb;
+
+		$key = strtolower( trim( (string) $meta_key ) );
+		if ( '' === $key ) {
+			return true;
+		}
+
+		$reserved_prefixes = [ 'np_', '_np_', 'newspack_', '_newspack_' ];
```
jason10lee, line 148: **Suggestion:** Silent drop of prefixed keys deserves a line in the PR body and release note — Blocking the whole `np_`/`newspack_` prefix is the right call, and the in-repo picture is reassuring: the one first-party caller of this endpoint sends only `fundraiseup_*`-prefixed keys, so nothing we ship breaks. The exposure is publisher-written integrations, which live outside this workspace and cannot be checked from here.

For those, the failure is quiet — the writes stop landing and existing callers will not be reading the new `skipped_metadata_keys` field, so the first signal is data that stopped appearing. On a branch that ships outside the regular release cadence, I think this is worth stating explicitly in the PR body and the release note rather than leaving it to the response field.

A filter to **extend** the reserved list (never subtract from it) would also give a publisher a way to cover their own keys without waiting on a code change. This file already uses that pattern for the rate limit at line 295.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -125,6 +126,61 @@ public static function sanitize_metadata( $metadata ) {
...
+		foreach ( $reserved_prefixes as $prefix ) {
+			if ( 0 === strpos( $key, $prefix ) ) {
```
jason10lee: **Suggestion:** str_starts_with() reads better than 0 === strpos() — PHP 8.3 is the declared target and `str_starts_with()` is already used across `includes/`. Straight readability swap, no behaviour change.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -125,6 +126,61 @@ public static function sanitize_metadata( $metadata ) {
...
+				return true;
+			}
+		}
+
+		$reserved_keys = [
```
jason10lee, line 155: **Suggestion:** One further key class worth adding to this list — The list reserves `wpcom_user_id` because another system resolves records against it. While reviewing I found a further key of that same class that the list does not cover. It is not a regression from this PR, and I do not think it should hold the fix up.

Details and a suggested direction are in NPPM-3060, since the reserved list is the natural place to close it if we want it in this PR.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -455,20 +511,45 @@ public static function api_frontend_register_reader( \WP_REST_Request $request )
 			);
 		}
 
-		// Save arbitrary user metadata.
+		// Save arbitrary user metadata, skipping keys other code trusts.
 		$user_metadata = $request->get_param( 'metadata' );
+		$skipped_keys  = [];
 		if ( ! empty( $user_metadata ) ) {
 			foreach ( $user_metadata as $meta_key => $meta_value ) {
```
jason10lee, line 544: **Suggestion:** metadata can overwrite the dedicated first_name / last_name params — The metadata loop runs after the `wp_update_user()` call that applies the dedicated `first_name` / `last_name` params, so sending the same field both ways is last-write-wins in favour of `metadata`. I confirmed this on a running site: passing `first_name` in both places stored the `metadata` value.

Same caller on both sides, so there is no privilege question, but two request params writing the same field with non-obvious ordering is worth either resolving or documenting. There is a related note on these particular fields in NPPM-3060.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -455,20 +511,45 @@ public static function api_frontend_register_reader( \WP_REST_Request $request )
...
+				if ( self::is_reserved_meta_key( $meta_key ) ) {
+					$skipped_keys[] = $meta_key;
+					continue;
+				}
 				\update_user_meta( $result, $meta_key, $meta_value );
```
jason10lee, line 549: **Suggestion:** Follow-up: a downstream integration trusts identifiers written through this same sink — Out of scope for this PR — it is in a different repository and none of its keys are affected by the reserved list — but it surfaced while checking the sink's consumers and is worth its own ticket.

An integration in the Manager reads identifiers straight from user meta and passes them to a third-party API to render account-page content. Those keys carry no reserved prefix, so this endpoint still writes them, and the denylist cannot cover them without breaking the feature. The fix there is server-side verification of the identifier rather than a reserved-list entry.

Noting it here only so the connection to this sink is not lost; the specifics belong in the tracker.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -125,6 +126,61 @@ public static function sanitize_metadata( $metadata ) {
...
+			}
+		}
+
+		$reserved_keys = [
+			'wpcom_user_id', // Jetpack resolves subscriptions against this ID, not the local user ID.
+			'session_tokens',
+			'_application_passwords',
+			'wp_capabilities',
+			'wp_user_level',
+			'wp_user-settings',
```
jason10lee: **Nit:** wp_user-settings keys are pinned to a literal wp_ while capabilities/user_level are generalised — `capabilities` and `user_level` are matched against the site's own prefix via the regex below, but `wp_user-settings` and `wp_user-settings-time` are hardcoded with a literal `wp_`. On an install whose table prefix is not `wp_`, or on a multisite subsite, those two are not caught. The harm on its own is negligible — they hold admin UI preferences — but the asymmetry is easy to fix by folding them into the same alternation, and it removes a rule that behaves differently depending on how the site was installed.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
```diff
@@ -455,20 +511,45 @@ public static function api_frontend_register_reader( \WP_REST_Request $request )
...
+
 		$response_data = [
 			'success' => true,
 			'status'  => 'created',
 			'email'   => $email,
 		];
 
+		// So an integration author can tell a saved key from a discarded one.
+		if ( ! empty( $skipped_keys ) ) {
+			$response_data['skipped_metadata_keys'] = $skipped_keys;
```
jason10lee: **Nit:** skipped_metadata_keys discloses the table prefix, and is only present when non-empty — Two small things about the new field.

The denylist itself is public source, so echoing it back leaks nothing — but the prefix-derived branch is site-specific. A caller can send a batch of candidate prefixed keys and only the one matching the site's own prefix comes back, which discloses the table prefix to an unauthenticated caller. It is bounded by the per-IP limit and low value on its own; returning a count, or omitting keys matched by the prefix rule, would close it for free.

Separately, the field is only added when something was skipped, so a caller has to guard with `isset()`. Emitting it unconditionally as `[]` would be a more stable contract, matching how the verification payload is merged just below.

plugins/newspack-plugin/tests/unit-tests/reader-registration-endpoint.php
```diff
@@ -888,4 +888,152 @@ static function ( $limit, $ip, $bucket = 'registration' ) {
...
+		// The harm at the capabilities key is not escalation — sanitize_text_field()
+		// makes the value a scalar, which core ignores — it is that writing a scalar
+		// replaces the role array, stripping the account of every role.
+		$caps = get_user_meta( $user->ID, $caps_key, true );
+		$this->assertIsArray(
+			$caps,
+			'Request metadata must not be able to overwrite the capabilities meta.'
+		);
+		$this->assertArrayNotHasKey( 'administrator', $caps );
+		$this->assertFalse( user_can( $user->ID, 'manage_options' ) );
```
jason10lee: **Nit:** This assertion passes with or without the fix — I checked each assertion in this test against a hypothetical revert to see which ones actually bite, and the rest do — `is_reader_verified()`, the array-or-empty loop, `assertIsArray( $caps )` and the `skipped_metadata_keys` assertions all fail if the guard is removed.

This one does not: a scalar in the capabilities key cannot grant a capability either way, which is what the comment above it says. It asserts nothing about this change. Dropping it, or replacing it with an assertion that the role array still contains the expected role, would keep the test honest about what it proves.

Worth recording that the suite's table prefix is `wptests_`, so the `$wpdb->base_prefix` cases in `test_is_reserved_meta_key()` genuinely exercise the non-default-prefix branch rather than duplicating the hardcoded `wp_` entries.

jason10lee (approved):
Release risk: ELEVATED — this changes behavior on a shared contract: metadata keys that used to save are now dropped silently, and publisher integrations outside this workspace cannot be checked from here. Still, approving on the basis that it does what it says on the tin in a reasonably compact way.

I exercised this against a running site with Reader Activation on and a filter-registered integration. The endpoint no longer writes any of the keys other code trusts, and I could not get one past the guard. Every reserved key I sent was rejected and left no row behind; keys outside the reserved set still save exactly as before; the account came back unverified with its role intact. I also tried to get past the guard with case, whitespace and backslash variants, and each one normalised *into* the reserved set rather than around it. Two concerns I went looking for came back clean: an address that already has an account returns 409 before the metadata loop is reached, so this path cannot touch an existing account on the same site; and the new log call sits behind the existing per-IP limit, so it adds no flooding surface. I checked the new tests against a hypothetical revert and they do bite, with one exception noted below. PHPUnit, PHPCS and CI are green.

One follow-up on the reserved list is written up in NPPM-3060. The rest of the notes here are non-blocking: a self-containment point on the new guard, the silent drop of prefixed keys being worth a line in the PR body and the release note, and a few style and test points.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Done in 1b2eb4ca. `is_reserved_meta_key()` now does `sanitize_key( wp_unslash( (string) $meta_key ) )` itself, so it is self-contained rather than relying on `sanitize_metadata()` having gone first. Added `np\_reader` to the classifier test, which the previous `strtolower( trim() )` let through.

rbcorrales: Done in 1b2eb4ca. `@param int|string $meta_key`, with a note that numeric array keys arrive as int.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: The PR body already carries this, in the paragraph starting "Keys outside the reserved set are unaffected" - it names the prefix drop, the response field and the log. Happy to expand it if it reads too lightly.

Took the filter suggestion in 1b2eb4ca: `newspack_reserved_registration_meta_keys`, additive only. The built-in list and the prefix rules run independently of it, so a filter can add keys but never remove protection.

The release note is on the release process rather than something I can do from here, but agreed it belongs there.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Done in 1b2eb4ca.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Folded in rather than deferred, per the discussion on the ticket - and it turned out to need more than the two keys.

`_wcpay_customer_id` is only WooPayments' deprecated key, read by its migration fallback. The live path resolves `get_user_option( $this->get_customer_id_option(), ... )`, which returns `_wcpay_customer_id_live` or `_wcpay_customer_id_test` (`class-wc-payments-customer-service.php:126`, constants at `:31`/`:38`). On a freshly registered reader there is no blog-prefixed value, so the unprefixed mode-split key is what resolves. Reserving only the bare key would have been a no-op for live mode.

All three are now reserved, plus `_stripe_customer_id`. Tests assert both `get_user_meta()` and `get_user_option()` come back empty for each.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Done in 1b2eb4ca. All four now go through the prefix alternation - `(capabilities|user_level|user-settings|user-settings-time)` - against both the site prefix and `wp_`, so the literal entries are gone and the rule no longer depends on how the site was installed.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Both halves done in 1b2eb4ca, the disclosure one included.

The field is now always present, as `[]` when nothing was dropped, so callers need no `isset()` guard. I checked first that nothing in this repo or `repos/` reads it, so making it unconditional breaks no contract.

For the prefix: the full list still goes to the log, but the response now filters out keys matched by the prefix rule, so the batch-probe returns nothing that identifies the real prefix. Ordinary keys like `np_reader_email_verified` are still echoed, which keeps the field useful for an integration author, and those are public in the source anyway. Extracted `is_prefixed_account_key()` so the guard and the filter share one definition rather than a duplicated regex, and added an assertion that the prefixed key is blocked but absent from the response.

plugins/newspack-plugin/tests/unit-tests/reader-registration-endpoint.php
rbcorrales: Done in 1b2eb4ca, though weaker than you suggested - it is now `assertNotEmpty( $caps )`, asserting the role survives rather than naming it. You are right that the bite already lives in `assertIsArray` two lines up, so this mostly documents intent. Happy to pin the expected role instead if you would rather it carried its own weight.

Also confirming your note on the prefix: the suite runs `wptests_`, so the `$wpdb->base_prefix` cases exercise the non-default branch rather than duplicating the `wp_` ones. I added `$wpdb->base_prefix . "user-settings"` cases for the same reason.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Leaving this one, and flagging it rather than fixing it quietly.

Confirmed the ordering: the metadata loop runs after the `wp_update_user()` call, so `metadata` wins. Both resolutions I can see change behaviour for existing callers - reordering the two writes, or reserving `first_name`/`last_name` so the dedicated params are the only route - and neither is something I want to fold into a hotfix going straight to `release` with no alpha soak.

Worth its own ticket, particularly given the NPPM-3092 connection you drew, where these same fields travel by email match. Say the word and I will write it up.

plugins/newspack-plugin/includes/reader-activation/class-reader-registration.php
rbcorrales: Filed as NPPM-3082, before this PR went public.

Tracing it out, the newspack-plugin side is hardening rather than the fix: the integration key is published in the page source, and scoping metadata per integration still admits `fundraiseup_*` for the FundraiseUp integration itself. The fix that closes it has to bind the supporter to `$user->user_email` in the Manager, and it cannot reuse `validate_registration_request()`, which compares against the caller-supplied `npe`.

Agreed the reserved list cannot cover it without breaking the feature.

PR #794, opened 2026-08-04
fix(popups): limit preset prompt previews to prompt managers

rbcorrales:
### All Submissions:

Preset prompt previews rendered for any visitor. They now require the capability that single-prompt previews already require, and the `preset` parameter is ignored for everyone else, so a `?preset=` URL behaves for a visitor as if the parameter were absent rather than rendering the previewed preset.

Preview override values are encoded so they cannot change the meaning of the markup they land in, and are only read when a preset preview is requested. More detail in the ticket.

Closes NPPM-3064.

1. As an administrator, go to Audience → Setup → Campaign and preview a preset prompt. It renders with the values entered in the wizard.
2. Put a double quote in the Heading field and preview again. The heading shows the quote and the prompt still renders.
3. Put square brackets in the Body field and preview again. The brackets show as typed.
4. Copy the preview URL and open it in a session logged in as an editor. The preview renders the same way.
5. Open the same URL logged out. The page behaves as if the `preset` parameter were absent.
6. Activate the RAS presets from the wizard. The created prompts carry the values saved in the wizard.

### Other information:

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-presets.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-presets.php
@@ -115,7 +115,8 @@ final class Newspack_Popups_Presets {
 			return [];
 		}
 
-		// filter_input reads the original request, not $_GET, so it is null under PHPUnit.
+		// filter_input reads the SAPI's original request, so it bypasses wp_magic_quotes()
+		// and anything that modified $_GET, and is null under WP-CLI and PHPUnit.
 		if ( ! isset( $_GET['values'] ) || ! is_array( $_GET['values'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 			return [];
 		}
@@ -136,7 +137,8 @@ final class Newspack_Popups_Presets {
 
 		$sanitized = [];
 		foreach ( $values as $key => $value ) {
-			$key = \sanitize_text_field( $key );
+			// Keys are emitted into the same block-delimiter JSON as the values.
+			$key = self::encode_shortcode_delimiters( \sanitize_text_field( $key ) );
 			if ( is_array( $value ) ) {
 				$sanitized[ $key ] = self::sanitize_override_values( $value );
 				continue;
@@ -162,7 +164,12 @@ final class Newspack_Popups_Presets {
 	 */
 	private static function encode_shortcode_delimiters( $value ) {
 		if ( is_array( $value ) ) {
-			return array_map( [ __CLASS__, 'encode_shortcode_delimiters' ], $value );
+			// Keys as well as values: both are emitted into the same JSON.
+			$encoded = [];
+			foreach ( $value as $key => $item ) {
+				$encoded[ self::encode_shortcode_delimiters( $key ) ] = self::encode_shortcode_delimiters( $item );
+			}
+			return $encoded;
 		}
 		if ( ! is_string( $value ) ) {
 			return $value;
@@ -303,6 +310,9 @@ final class Newspack_Popups_Presets {
 			if ( 'string' === $field['type'] && ! is_scalar( $value ) ) {
 				// `?values[body][]=x` would otherwise reach trim() below as an array.
 				$value = null;
+			} elseif ( 'array' === $field['type'] && ! is_array( $value ) ) {
+				// And the mirror: a scalar would be JSON-encoded where a list is expected.
+				$value = null;
 			} else {
 				$value = self::encode_shortcode_delimiters( $value );
 			}
@@ -317,7 +327,9 @@ final class Newspack_Popups_Presets {
 		if ( 'string' === $field['type'] && isset( $field['max_length'] ) ) {
 			$value = substr( trim( $value ), 0, $field['max_length'] );
 		}
-		// If an array, stringify with field name.
+		// If an array, stringify with field name. Overrides hold only strings and
+		// arrays, so the brackets wp_json_encode() emits here are always followed by a
+		// quote or another bracket, never by a tag name do_shortcode() would match.
 		if ( 'array' === $field['type'] ) {
 			$value = '"' . $field_name . '": ' . \wp_json_encode( $value );
 		}
--- a/plugins/newspack-popups/includes/class-newspack-popups-presets.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-presets.php
@@ -22,6 +22,11 @@ final class Newspack_Popups_Presets {
 	 * @return object|null Popup object, or null if no preset is found for the given slug.
 	 */
 	public static function retrieve_preset_popup( $slug ) {
+		// Renders request input, so restrict it like retrieve_preview_popup() does.
+		if ( ! Newspack_Popups::is_user_admin() ) {
+			return null;
+		}
+
 		$presets = self::get_ras_presets();
 
 		if ( \is_wp_error( $presets ) || ! $presets || ! isset( $presets['prompts'] ) ) {
@@ -98,6 +103,74 @@ final class Newspack_Popups_Presets {
 	}
 
 	/**
+	 * Read preset preview override values from the request.
+	 *
+	 * Values arrive as `?values[field_name]=…`; `lists` arrives as an array.
+	 *
+	 * @return array Sanitized override values, keyed by field name.
+	 */
+	private static function get_override_values() {
+		// get_ras_presets() also backs the two functions that write real prompt posts.
+		if ( ! Newspack_Popups::preset_popup_id() ) {
+			return [];
+		}
+
+		// filter_input reads the original request, not $_GET, so it is null under PHPUnit.
+		if ( ! isset( $_GET['values'] ) || ! is_array( $_GET['values'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return [];
+		}
+
+		return self::sanitize_override_values( \wp_unslash( $_GET['values'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+	}
+
+	/**
+	 * Sanitize preset preview override values.
+	 *
+	 * @param mixed $values Raw values from the request.
+	 * @return array Sanitized values.
+	 */
+	private static function sanitize_override_values( $values ) {
+		if ( ! is_array( $values ) ) {
+			return [];
+		}
+
+		$sanitized = [];
+		foreach ( $values as $key => $value ) {
+			$key = \sanitize_text_field( $key );
+			if ( is_array( $value ) ) {
+				$sanitized[ $key ] = self::sanitize_override_values( $value );
+				continue;
+			}
+			if ( ! is_scalar( $value ) ) {
+				continue;
+			}
+			// Not sanitize_text_field(): an unencoded `"` closes the JSON string these sit in.
+			$sanitized[ $key ] = self::encode_shortcode_delimiters( filter_var( (string) $value, FILTER_SANITIZE_FULL_SPECIAL_CHARS ) );
+		}
+
+		return $sanitized;
+	}
+
+	/**
+	 * Encode the characters that could introduce a shortcode into preset markup.
+	 *
+	 * Backslashes too: `parse_blocks()` JSON-decodes these values, which would turn an
+	 * escape sequence back into a bracket.
+	 *
+	 * @param mixed $value Value to encode.
+	 * @return mixed Value with delimiters encoded.
+	 */
+	private static function encode_shortcode_delimiters( $value ) {
+		if ( is_array( $value ) ) {
+			return array_map( [ __CLASS__, 'encode_shortcode_delimiters' ], $value );
+		}
+		if ( ! is_string( $value ) ) {
+			return $value;
+		}
+		return str_replace( [ '[', ']', '\\' ], [ '&#91;', '&#93;', '&#92;' ], $value );
+	}
+
+	/**
 	 * Retrieve default prompts + segments.
 	 *
 	 * @return array|WP_Error Array of prompt and segment default configs.
@@ -123,7 +196,7 @@ final class Newspack_Popups_Presets {
 		}
 
 		// Get override values if previewing a preset.
-		$override_values = filter_input( INPUT_GET, 'values', FILTER_SANITIZE_FULL_SPECIAL_CHARS, FILTER_REQUIRE_ARRAY );
+		$override_values = self::get_override_values();
 
 		// Populate prompt configs with saved inputs.
 		$data['prompts'] = array_map(
@@ -155,7 +228,8 @@ final class Newspack_Popups_Presets {
 					}
 					if ( 'int' === $field['type'] && 'featured_image_id' === $field['name'] ) {
 						if ( ! empty( $override_values['featured_image_id'] ) ) {
-							$prompt['featured_image_id'] = $override_values['featured_image_id'];
+							// An int field, and the only override that skips process_user_inputs().
+							$prompt['featured_image_id'] = absint( $override_values['featured_image_id'] );
 						} elseif ( isset( $field['value'] ) ) {
 							$prompt['featured_image_id'] = $field['value'];
 						}
@@ -221,6 +295,16 @@ final class Newspack_Popups_Presets {
 		}
 
 		$field_name = $field['name'];
+
+		// An override comes from the request; saved and default values do not.
+		if ( ! empty( $value ) ) {
+			if ( 'string' === $field['type'] && ! is_scalar( $value ) ) {
+				// `?values[body][]=x` would otherwise reach trim() below as an array.
+				$value = null;
+			} else {
+				$value = self::encode_shortcode_delimiters( $value );
+			}
+		}
 
 		if ( ! $value ) {
 			$value = isset( $field['value'] ) ? $field['value'] : $field['default'];
--- a/plugins/newspack-popups/includes/class-newspack-popups-presets.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-presets.php
@@ -222,7 +222,9 @@ final class Newspack_Popups_Presets {
 
 				// Populate placeholder content with saved inputs or default values.
 				foreach ( $prompt['user_input_fields'] as $field ) {
-					$override = ! empty( $override_values[ $field['name'] ] ) ? $override_values[ $field['name'] ] : null;
+					// Not `! empty()`: "0" is a value, and process_user_inputs() decides
+					// for itself what counts as absent.
+					$override = $override_values[ $field['name'] ] ?? null;
 					if ( 'array' === $field['type'] || 'string' === $field['type'] ) {
 						$prompt['content'] = self::process_user_inputs( $prompt['content'], $field, $override );
 					}
--- a/plugins/newspack-popups/includes/class-newspack-popups-presets.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-presets.php
@@ -297,7 +297,7 @@ final class Newspack_Popups_Presets {
 		$field_name = $field['name'];
 
 		// An override comes from the request; saved and default values do not.
-		if ( ! empty( $value ) ) {
+		if ( null !== $value ) {
 			if ( 'string' === $field['type'] && ! is_scalar( $value ) ) {
 				// `?values[body][]=x` would otherwise reach trim() below as an array.
 				$value = null;
@@ -306,7 +306,8 @@ final class Newspack_Popups_Presets {
 			}
 		}
 
-		if ( ! $value ) {
+		// Not `! $value`: "0" is a value an editor can legitimately preview.
+		if ( null === $value || '' === $value || [] === $value ) {
 			$value = isset( $field['value'] ) ? $field['value'] : $field['default'];
 		}
 
--- a/plugins/newspack-popups/includes/class-newspack-popups.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups.php
@@ -938,6 +938,11 @@ final class Newspack_Popups {
 	 * @return string|null Popup slug, if found in the URL
 	 */
 	public static function preset_popup_id() {
+		// Ignored for everyone else: preview mode suppresses prompts and swaps the
+		// reader data store.
+		if ( ! self::is_user_admin() ) {
+			return null;
+		}
 		// Not using filter_input since it's not playing well with phpunit.
 		if ( isset( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) && $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 			return sanitize_text_field( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
--- a/plugins/newspack-popups/includes/class-newspack-popups.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups.php
@@ -938,16 +938,17 @@ final class Newspack_Popups {
 	 * @return string|null Popup slug, if found in the URL
 	 */
 	public static function preset_popup_id() {
+		// Param first, so a normal request does not pay for a capability check.
+		// Not using filter_input since it's not playing well with phpunit.
+		if ( ! isset( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) || ! $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			return null;
+		}
 		// Ignored for everyone else: preview mode suppresses prompts and swaps the
 		// reader data store.
 		if ( ! self::is_user_admin() ) {
 			return null;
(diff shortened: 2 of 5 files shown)


plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -222,6 +296,16 @@ public static function process_user_inputs( $prompt_content, $field, $value = nu
...
+				// `?values[body][]=x` would otherwise reach trim() below as an array.
+				$value = null;
+			} else {
+				$value = self::encode_shortcode_delimiters( $value );
+			}
+		}
+
 		if ( ! $value ) {
 			$value = isset( $field['value'] ) ? $field['value'] : $field['default'];
 		}
```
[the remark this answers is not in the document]
rbcorrales: Fixed in fcf60c0b. The override branch now tests `null !== $value` and the fallback tests explicitly for null, empty string and empty array, so `?values[body]=0` previews as `0` instead of silently showing the default.

Half of this was pre-existing: the `! $value` fallback predates the PR. The `! empty( $value )` guard I added put a second falsy trap on the same value, which is what made it worth correcting here rather than leaving it. Added assertions for both the `"0"` and the empty-string cases.

rbcorrales:
Both suppressed comments were right, and the first one caught an incomplete fix. Addressed in 294a0ad7.

**The `"0"` fix did not go far enough.** I had corrected `process_user_inputs()` but left `get_ras_presets()` deciding whether to pass an override with `! empty()`, so `"0"` never reached the code I fixed and the preview still fell back to the default. That is now `$override_values[ $field["name"] ] ?? null`, letting `process_user_inputs()` decide for itself what counts as absent.

My earlier test only exercised `process_user_inputs()` directly, which is exactly why it passed while the real path stayed broken. Replaced with an end-to-end assertion through `retrieve_preset_popup()`, confirmed to fail when the caller change is reverted.

**Capability check moved after the param check.** `preset_popup_id()` is reached on ordinary front-end requests through `is_preview_request()`, so running `current_user_can()` plus a filter before looking for `?preset=` was work on every page load. It now returns early when the param is absent.

253 tests, 644 assertions green.

rbcorrales:
Checked this one and it does not hold, though the coverage gap behind it was real.

`sanitize_text_field()` does return the string `"0"` for an integer key, but PHP casts a numeric string back to an integer when it is used as an array key, so the array stays a list. Confirmed in the container:

```
key 0 (integer) -> "0" (string)
array keys after assignment: [ 0 => 0, 1 => 1 ]
array_is_list: yes
wp_json_encode: ["a","b"]
```

Added a test in a447285a rather than just replying, because two earlier reviewers had flagged `lists` as the one array-typed override with no coverage. It sets `?values[lists][]=1&2` through the real preview path and asserts the content carries `"lists": ["1","2"]`, then parses the block back out and asserts `attrs["lists"]` is the PHP list `["1","2"]`. That covers the JSON-array-versus-object question directly.

254 tests, 648 assertions green.

rbcorrales:
Took one of these, declined the other.

**Took the `parent::set_up()` one** in 50f93dad. It is pre-issuing this PR, and an earlier reviewer had recommended deferring it on the grounds that wrapping the three existing tests in a transaction is a behaviour change to make on a release branch. I tried it instead of assuming: full suite green at 254 tests, 648 assertions. Since my changes are what added user-creating tests to this class, the isolation now matters more than it did, and it is test-only code so the blast radius is this file.

It also made two manual lines redundant. `clean_up_global_scope()` already empties `$_GET` and `tear_down()` already resets the current user, so the explicit resets I had added are gone rather than sitting there with a comment claiming they are necessary.

**Declined the `@return object|null` one.** Technically right, since `create_popup_object()` returns an array. But that docblock is pre-existing and matches the surrounding file, which describes this same shape as an object in four `@return` tags and thirteen `@param` tags, one of which reads `@return object Array of prompt options`. Correcting a single instance would leave it inconsistent with its own caller, and correcting all of them is a cleanup rather than something to fold into a security hotfix.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
+	 * @return array Sanitized values.
+	 */
+	private static function sanitize_override_values( $values ) {
+		if ( ! is_array( $values ) ) {
+			return [];
+		}
+
+		$sanitized = [];
+		foreach ( $values as $key => $value ) {
+			$key = \sanitize_text_field( $key );
```
jason10lee: **Suggestion:** Array keys skip the delimiter encoding that their values get — Keys go through `sanitize_text_field()`, which leaves `[` and `]` untouched, and are then emitted into the block-delimiter JSON alongside the values that *are* encoded. Two reviewers reached this independently and neither could construct a working exploit. PHP's own query-string parser truncates a key at the first `]`, which appears to be what blocks it.

That is a property of the parser rather than of this code, though, so running keys through the same encoding (or `sanitize_key()`) removes the dependence on it. One related detail: `wp_unslash()` maps values only, so keys arrive still slashed and a key containing a quote will silently fail to match a field name.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
+	/**
+	 * Encode the characters that could introduce a shortcode into preset markup.
+	 *
+	 * Backslashes too: `parse_blocks()` JSON-decodes these values, which would turn an
+	 * escape sequence back into a bracket.
+	 *
+	 * @param mixed $value Value to encode.
+	 * @return mixed Value with delimiters encoded.
+	 */
+	private static function encode_shortcode_delimiters( $value ) {
```
jason10lee, line 165: **Suggestion:** The encoding does not cover every substitution branch, so the stated invariant does not hold — `encode_shortcode_delimiters()` handles `[`, `]` and `\`, and its docblock describes that as covering the characters that could introduce a shortcode into preset markup. The array branch in `process_user_inputs()` (the `'array' === $field['type']` case, around line 321) emits its value through `wp_json_encode()` without that encoding, and neither `FILTER_SANITIZE_FULL_SPECIAL_CHARS` nor the delimiter encoding touches `{` or `}`, so a value that names another field's placeholder re-enters the substitution loop and can put unencoded brackets back into content that later reaches `do_shortcode()`.

We reproduced this against a running site: the rendered body came back containing live `[` and `]` characters. We could **not** turn it into shortcode execution, and are not claiming it is exploitable: the encoded value is always quoted, so a bracket is never followed by a tag name. The capability gate also holds independently.

The reason we would still fix it is that it makes the comments describe a layer the code only partly provides. If someone later relaxes the gate believing the encoding backs it up, that assumption is true for most of the path and not all of it. Encoding `{` alongside the other three characters, or neutralising the placeholder syntax in override values, would close it for about the cost of one array element. Happy to write the detail into the tracker rather than this thread if you'd prefer.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -222,7 +298,18 @@ public static function process_user_inputs( $prompt_content, $field, $value = nu
 
 		$field_name = $field['name'];
 
-		if ( ! $value ) {
+		// An override comes from the request; saved and default values do not.
+		if ( null !== $value ) {
```
jason10lee, line 309: **Suggestion:** The type guard is one-sided — The new guard neutralises a wrong-shaped value for a `string` field, so `?values[body][]=x` can no longer reach `trim()` as an array. There is no mirror for the array case: an `array` field given a scalar passes straight through to the JSON branch below and produces a string where the block expects a list, which the block's render then iterates.

This is preview-only and behind the capability check, so the stakes are low. It is the same failure the adjacent guard was added for, though, and `'array' === $field['type'] && ! is_array( $value )` → `$value = null` would make the pair symmetric.

plugins/newspack-popups/includes/class-newspack-popups.php
```diff
@@ -938,11 +938,17 @@ public static function previewed_popup_id() {
...
 	public static function preset_popup_id() {
+		// Param first, so a normal request does not pay for a capability check.
 		// Not using filter_input since it's not playing well with phpunit.
-		if ( isset( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) && $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-			return sanitize_text_field( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( ! isset( $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) || ! $_GET[ self::NEWSPACK_POPUP_PRESET_QUERY_PARAM ] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			return null;
 		}
-		return null;
+		// Ignored for everyone else: preview mode suppresses prompts and swaps the
```
jason10lee, line 946: **Suggestion:** This capability's documentation no longer matches what it gates — This new gate depends on `Newspack_Popups::is_user_admin()`, and that helper and its `newspack_popups_admin_user_capability` filter are both documented purely as deciding whether a user's activity is excluded from Google Analytics. After this PR the same capability also decides who can render request-supplied content through the preview path.

In fairness, it was already load-bearing for authorization at two other call sites before this PR, so this raises the count rather than introducing the coupling, and reusing it is the consistent choice. The concrete hazard is that a publisher who widens that filter for the analytics reason its own documentation describes would also widen who can reach the preview surfaces, without anything in the docblock suggesting they were making that second change.

A sentence in both docblocks (they live around line 1050) would close it. A dedicated `can_preview_prompts()` with its own filter would be cleaner separation, though that is probably beyond a hotfix.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
 	}
 
+	/**
+	 * Read preset preview override values from the request.
+	 *
+	 * Values arrive as `?values[field_name]=…`; `lists` arrives as an array.
+	 *
+	 * @return array Sanitized override values, keyed by field name.
+	 */
+	private static function get_override_values() {
```
jason10lee, line 112: **Nit:** The three new methods carry no parameter or return types — `get_override_values(): array`, `sanitize_override_values( mixed $values ): array` and `encode_shortcode_delimiters( mixed $value ): mixed` are all unambiguous from their own docblocks, and all three are private so nothing outside can break. Deferring is entirely reasonable if the goal is a minimal hotfix diff against `release`, given the surrounding file is uniformly untyped.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
+		// get_ras_presets() also backs the two functions that write real prompt posts.
```
jason10lee, line 113: **Nit:** The guard keys on the parameter being present rather than on preview context — We are recording an adjudication here rather than asking for a change. Two reviewers flagged that because this tests for the parameter rather than for a frontend preview, an admin-authenticated request that carried the preview parameters while hitting the persistence path could bake override values into real prompt posts. We confirmed the mechanism on a running site: with the parameters set and an admin user, the override does reach the prompt content that the importer would persist.

It is not reachable, though. That path is REST-only behind a `manage_options` permission callback, and the wizard never appends the preview parameters to that URL, so getting there needs a separate vulnerability first. And before this PR there was no guard at all on reading these values, so this change narrows the surface rather than leaving a gap in it.

We are leaving it as a note because a check for frontend-render context would close it by construction, and because the existing test only covers the parameter being absent.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
+		if ( ! Newspack_Popups::preset_popup_id() ) {
+			return [];
+		}
+
+		// filter_input reads the original request, not $_GET, so it is null under PHPUnit.
```
jason10lee: **Nit:** This comment undersells its own reasoning — "so it is null under PHPUnit" reads as production code bending to a test, which will draw an objection it does not deserve. The real reason is broader: `filter_input()` reads the SAPI's original request, so it also returns null under WP-CLI and bypasses `wp_magic_quotes()` and any earlier modification of `$_GET`. Phrasing it that way makes the change self-justifying, and matches the existing comments elsewhere in this file.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -97,6 +102,74 @@ function( $acc, $list ) {
...
+			$key = \sanitize_text_field( $key );
+			if ( is_array( $value ) ) {
+				$sanitized[ $key ] = self::sanitize_override_values( $value );
+				continue;
+			}
+			if ( ! is_scalar( $value ) ) {
+				continue;
+			}
+			// Not sanitize_text_field(): an unencoded `"` closes the JSON string these sit in.
+			$sanitized[ $key ] = self::encode_shortcode_delimiters( filter_var( (string) $value, FILTER_SANITIZE_FULL_SPECIAL_CHARS ) );
```
jason10lee, line 150: **Nit:** Control characters survive the filter — `FILTER_SANITIZE_FULL_SPECIAL_CHARS` without `FILTER_FLAG_STRIP_LOW` passes newlines and tabs through, and a raw newline inside the block-delimiter JSON is invalid, so the block ends up rendering with no attributes at all. This is pre-existing (the previous `filter_input` call had the same flags), preview-only, and effectively unreachable from the wizard UI. Adding the flag is only worth it if this line is being touched anyway.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
```diff
@@ -222,7 +298,18 @@ public static function process_user_inputs( $prompt_content, $field, $value = nu
...
 		$field_name = $field['name'];
 
-		if ( ! $value ) {
+		// An override comes from the request; saved and default values do not.
+		if ( null !== $value ) {
+			if ( 'string' === $field['type'] && ! is_scalar( $value ) ) {
+				// `?values[body][]=x` would otherwise reach trim() below as an array.
+				$value = null;
+			} else {
+				$value = self::encode_shortcode_delimiters( $value );
```
jason10lee, line 317: **Nit:** max_length now counts encoded characters — Because the override is encoded here before it reaches the length handling below, `max_length` is now measured in post-encoding characters. A heading containing quotes spends five characters per encoded quote against the editor's budget, so it truncates earlier than the field label implies, and `substr()` can leave a partial entity behind. This is cosmetic and preview-only, since the server-side string `do_shortcode()` sees still contains no bracket. It tends to surface later as a support report about headings being cut off.

plugins/newspack-popups/includes/class-newspack-popups.php
```diff
@@ -938,11 +938,17 @@ public static function previewed_popup_id() {
 	 * @return string|null Popup slug, if found in the URL
 	 */
 	public static function preset_popup_id() {
```
jason10lee, line 940: **Nit:** Follow-up: an adjacent preview parameter is worth its own ticket — This is out of scope here, and we note it only so the asymmetry is not lost now that these two functions sit next to each other. The sibling preview parameter still has no capability check of its own, and the reasoning in the new comment above applies to it as well.

Nothing in this diff causes it and the content-disclosure half is already closed, so this is not a reason to hold the PR. It deserves its own ticket rather than scope creep on a hotfix, and we are happy to write it up separately.

plugins/newspack-popups/tests/test-presets.php
```diff
@@ -22,6 +24,352 @@ public function set_up() {
...
+		} finally {
+			unset( $_GET['preset'] );
+		}
+	}
+
+	/**
+	 * Preview override values are spliced into content that is later rendered
+	 * through do_shortcode(), so they must not be able to introduce a shortcode.
+	 */
+	public function test_preset_override_values_cannot_inject_shortcodes() {
```
jason10lee, line 102: **Nit:** The tests guard the new implementation rather than reproducing the old issue — These are genuine forward regression guards: remove the encoding from either call site and they go red, which we checked. What they do not do is worth stating: because the pre-fix code read overrides through `filter_input()`, which is inert under PHPUnit, reverting the fix wholesale would make them pass vacuously rather than fail. So a green suite should not be read as evidence that the original issue was reproduced in CI.

Separately, nothing asserts the reader-facing half of the encoding: that the encoded entity comes back as a literal `[` for the person reading the preview. The current assertion pins the encoded form, which would keep passing if a later change started double-encoding it into visibly broken output. We confirmed the literal render by hand, so an assertion would just lock in what already works.

jason10lee (approved):
Release risk: LOW — no blocking findings, the change is confined to one extension plugin with no foundation contract touched, and it is well covered by new tests.

We exercised this against a running site with a marker shortcode registered, and the exploit from the ticket is closed. An anonymous request carrying the preset parameter and a bracketed payload renders no preset at all. The same URL as an administrator renders the preview and shows the payload as literal text instead of running it. That second result is backed by a positive control: the identical shortcode placed in ordinary post content does execute in the same response, so this is a real block rather than a shortcode that was never registered. The full Campaigns suite is green at 254 tests, PHPCS is clean, and CI is passing.

Both halves in one frame — the preview rendering with the payload inert, and the same shortcode executing in the trusted post body just below it:

Preset preview renders for an administrator: injected shortcode shows as literal text while the control in post content executes

Two things we checked because they looked like they could go wrong, and did not: every consumer of the newly-nullable `retrieve_preset_popup()` handles a null return, including the shortcode path the diff did not touch; and gating the parameter moves every downstream behaviour toward less permissive, not more.

The notes below are all non-blocking. The one we would most like a second opinion on is that the encoding does not cover every substitution branch, so the comments read as though there are two independent layers when for part of the path there is one. Nothing exploitable follows from it today, and the capability gate holds regardless. It matters mainly if someone later relaxes the gate on the assumption that the encoding backs it up.

One thing to keep in mind after release. We checked the preview surface in a browser and it works as intended for an administrator. We also narrowed the capability filter deliberately to see what a declined preview looks like, and it is silent: the preset simply does not appear, with no notice and no error. That is the correct security behaviour, but on a site whose capability has been narrowed it will present as a broken wizard rather than as an access check doing its job, so it is worth knowing about before the first support ticket arrives.

Same URL, same signed-in administrator, capability narrowed:

Same URL and same signed-in administrator with the capability narrowed: no preset renders and nothing explains why

plugins/newspack-popups/includes/class-newspack-popups.php
jason10lee: Filed as NPPM-3093.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: This one turned out to be exploitable, so thank you for spotting the mechanism even though we both read the consequence wrong.

Chained with the `{{lists}}` re-entry in your next note it is a working shortcode execution. I reproduced it on the branch before fixing:

```
?preset=ras_newsletter_inline&values[body]={{lists}}&values[lists][[np_probe_tag%3D1][]=x
```

rendered `<p>"lists": {"[np_probe_tag=1":["x"]}</p>` and the shortcode fired. The `]` comes free from the nested JSON array, and the tag name only needs terminating with a pre-scan delimiter, which `%3D` supplies.

On the parser: PHP forbids `]` inside a key but permits `[`, so truncation is not what was blocking it. What blocked your attempt was `do_shortcode()`'s tag pre-scan failing to match `[gallery` followed by `":`.

Fixed in 0deea7f7. Keys now get the same encoding as values, and I moved that inside `encode_shortcode_delimiters()` so the guarantee travels with the method rather than depending on one caller. Regression test covers the flat and nested forms and fails when only that line is reverted.

On `wp_unslash()`: keys are never slashed to begin with, since `add_magic_quotes()` only `addslashes()` string values. Sanitizing them is still right, for the reason above.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Correct observation, and it is half of the chain above.

Not encoding `{` and `}`, for a reason stronger than the one I first had: merge tags run on `newspack_popups_popup_content`, which the model applies *after* `do_shortcode()`, so merge-tag output can never reach the shortcode parser. Once keys are encoded there is no route left that puts a live bracket next to a tag name. `wp_json_encode()` emits structural brackets, but overrides only ever hold strings and arrays, so those are always followed by a quote or another bracket.

That assumption was doing load-bearing work with nothing stating it, so it is now a comment at the array branch. The single `strtr()` pass you suggested is the clean version and I agree it belongs on `alpha`, not here.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Done in 0deea7f7, and it was guarding more than a cosmetic glitch: without it `?values[lists]=x` throws `TypeError: array_intersect(): Argument #1 must be of type array, string given` in the subscribe block, so a 500 on the preview rather than a bad render. Added a test, confirmed failing without the guard.

plugins/newspack-popups/includes/class-newspack-popups.php
rbcorrales: Done in 0deea7f7, a sentence in both the method docblock and the filter doc. One correction: it was already load-bearing at three call sites, not two - `class-newspack-popups-model.php:299` and `:909`, plus `class-newspack-popups-view-as.php:40`. Agreed the dedicated `can_preview_prompts()` is the cleaner separation and belongs outside a hotfix.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Done in 0deea7f7. It now gives the real reason: `filter_input()` reads the SAPI's original request, so it bypasses `wp_magic_quotes()` and any earlier modification of `$_GET`, and is null under WP-CLI as well as PHPUnit.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Tried this and backed it out, because the flag does nothing here. PHP ignores `FILTER_FLAG_STRIP_LOW` for `FILTER_SANITIZE_FULL_SPECIAL_CHARS`; only `FILTER_SANITIZE_SPECIAL_CHARS` and friends honour it. On the container's 8.3:

```
FULL_SPECIAL_CHARS + STRIP_LOW: "a\nb\tc"   (unchanged)
SPECIAL_CHARS      + STRIP_LOW: "abc"
```

Leaving it in would have been a comment claiming a protection that does not exist.

Worth flagging that this is not preview-only: a newline saved into `success_message` through the wizard breaks the published prompt the same way, since the save path stores the value raw and never encodes it. Had the flag worked, preview would have looked correct while published stayed broken. The underlying newline-versus-delimiter-JSON bug deserves its own ticket rather than a one-sided fix here.

plugins/newspack-popups/tests/test-presets.php
rbcorrales: Fair, and the specific gap was worse than described: my new array-key test had only negative assertions about the code under test, so if the override had stopped arriving entirely it would still have passed. Added a positive control asserting the encoded payload reaches the content.

On the wholesale-revert point, I read it differently. Most of the new tests carry a positive control that a revert breaks: `Safe heading` must appear, the capability test, the param-inert test, the featured-image and list-array assertions, the zero-override test. The three that would pass vacuously are the JSON-escape, block-attribute and ignored-without-a-preset tests, which is a real gap but narrower than the whole suite.

Also took your assertion suggestion: the injection test now asserts the *decoded* render, so a later change that double-encoded the entity into visibly broken output fails here.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Leaving these untyped. The file is uniformly untyped and all three are private, so nothing outside can break; I would rather not mix a style sweep into a hotfix landing on `release`.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Leaving as noted, with the correction that "unreachable" overstates it. `activate_ras_presets()` does persist `get_ras_presets()` output and that request can carry the preview parameters, so the honest framing is that it needs an admin acting against their own site: REST nonce blocks CSRF, and the values are bracket-encoded by then. The structural fix is to have `retrieve_preset_popup()` pass overrides down rather than `get_ras_presets()` reading superglobals, which is a bigger change than this.

plugins/newspack-popups/includes/class-newspack-popups-presets.php
rbcorrales: Deferring, with one correction: the quote inflation you describe is pre-existing, since the previous `filter_input()` applied the same filter before the same `substr()`. What this PR adds is `[`, `]` and `\` inflation.

Worth recording that the wizard enforces `max_length` client-side against the raw typed string, so the counter and the server disagree by construction, and truncation can leave a partial entity that renders literally. Preview-only, saved values unaffected.

plugins/newspack-popups/includes/class-newspack-popups.php
rbcorrales: Agreed, and thanks for filing NPPM-3093.

PR #798, opened 2026-08-04
fix(audience): clearer schedule, impact preview and deal id controls

thomasguillot:
Stacked on #757 — base branch is `fix/dsgnews-189-dynamic-pricing`, so review that one first and this diff second.

### All Submissions:

Three parts of the Pricing Rules editor still read as a port of the engine's classic metabox. This brings them in line with the rest of Audience.

**Schedules are picked, not typed.** `Starts` and `Ends` were native `datetime-local` inputs, which render differently in every browser and locale — Firefox drops the time half in some locales, which the parsing code already had to defend against. They are now a pair of buttons on one row, each carrying the resolved value and opening a date and time picker in a popover. An unset field reads `Active immediately` or `No end date` instead of looking empty, and `Clear` puts it back. One timezone line sits under the pair rather than the same sentence twice, and both inputs point at it. The stored value is unchanged: still a UTC timestamp, still to the minute.

Starts and Ends side by side under one shared timezone line

The same picker sets the custom eligibility date. That was the last native `datetime-local` on the page, so a publisher gating a rule by cohort met one date idiom in Scheduling and another in Eligibility. `DateTimeField` gained an option to hide its label, which is what the condition needs: the selector above it already names the field, and a second visible label would read as a duplicate.

The custom eligibility date opens the same picker as the schedule

**The Impact Preview leads with its numbers.** The two facts worth reading — how many products a rule touches and how many subscribers it reaches — were a muted sentence above the table and an italic line stranded below it. They are now a stat strip at the top. The table itself is a DataViews table rather than a hand-rolled one, so its columns sort. It renders at compact density with no toolbar chrome, all rows on one page, and its outer columns flush with the text above. The table uses `@wordpress/dataviews` directly rather than the Newspack wrapper, which it has no use for.

Three states used to leave the section blank or nearly so. Each now gets a card that says what would fill it: no price typed yet, nothing matching the rule, and no preview available. That last one is new behaviour — a failed request or an engine that can't answer previously made the whole section vanish with no explanation. The card follows the Institutions onboarding pattern: `SectionHeader` with `pageHeader`, centred in the card and held to a readable measure. That header gains a `size="small"` variant, and `CoreCard` a matching large body size (48px), giving a 24px icon in a 40px circle and a `heading-large` title, so the pattern fits a card without a consumer overriding the component from outside.

Empty state: dollar icon in a 40px circle on a large core card

Impact Preview: stat strip then a compact DataViews table

**The Deal ID is a field, not a footnote.** It sat unlabelled under `Name`, which read as help text for the name. It now has its own labelled read-only row at the end of Rule Details, with a Copy button, since the whole point is taking the ID somewhere else. The value is the rule's post ID.

The label keeps the engine's term rather than being renamed to match this screen. `deal_id` is what the ID is called on the analytics side: the engine emits a `wcdp_deal_purchase` event carrying `deals[]` keyed on it, and stamps `_wcdp_deal_ids` on orders. Renaming the field would split it from the word a publisher meets in reporting, so the help text points at `deal_id` and the label agrees with it. Two other strings that said "deal" where they meant the rule itself now say "rule".

Deal ID as a labelled read-only field with a Copy button

Copying reports through the wizard snackbar, success and failure alike, so the outcome carries the same weight as any other confirmation in Audience and a screen reader hears it without a second announcement.

Copying the Deal ID confirms in a snackbar at the foot of the editor

The catalog panel above the rules list shares the impact table, so it picks up the DataViews rendering too. Its heading and sample line are unchanged.

### Before and after

Base branch on the left, this branch on the right.

**Schedule controls and the Deal ID field**

Schedule controls and Deal ID field, before and after

**Impact Preview, with rows and without**

Impact Preview table and empty state, before and after

### Needs a follow-up in the pricing engine

`total_matching` over-counts, and it is a `woocommerce-dynamic-pricing` issue rather than something to paper over here. In `Pricing_Impact_Calculator::impact_preview()` the total starts as the scope query's `found_posts`, then the loop drops products for two reasons but subtracts only one of them:

- grouped or engine-excluded → `++$excluded`, subtracted from the total
- fails `matches_product()` → dropped from the sample, **not** subtracted

That second gate is the authoritative one; the candidate query is, in its own comment, "just an efficient pre-filter". So products the query matched but the rule does not actually apply to still count toward "N products affected". On a local site with 36 candidates and a sample cap of 50 the whole set was examined, yet the payload reported 36 with 33 priced.

It also makes `preview_limited` (`$total > count( $sample )`) fire when nothing was truncated. That flag is no longer read here — the sample note it drove has been removed — but the over-count still reaches the publisher through the stat strip.

The bespoke `woocommerce-dynamic-pricing` 0.1.0 engine plugin must be active, or Pricing Rules does not register. Retail 3.4.15 shares the slug and does nothing.

1. Open **Audience → Pricing Rules** and edit any rule. The panel above the list and the Impact Preview inside the editor both render as tables with column headers you can sort. Neither has a search box, filters, or a settings cog.
2. In **Scheduling & Behavior**, `Starts` and `Ends` sit side by side, each filling half the column. Click `Starts`. A popover opens with a time row and a calendar.
3. Pick a date. The button now reads that date and time, and a `Clear` action appears in the popover.
4. Click `Clear`. The button reads `Active immediately` again and focus returns to it.
5. Add a rule with the **Custom** goal and scroll to **Eligibility**. Set `Subscriptions started on/after` to `Custom`. The date field below it is the same picker as `Starts`, not a browser date input.
6. Scroll to **Impact Preview**. The product and subscriber counts sit at the top and every sampled row is on one page.
7. Edit a rule whose price locks in at purchase. The strip shows two counts and one line saying no existing subscriber is repriced.
8. Create a new rule and leave the price empty. A card reads `No price set yet` where the table would be.
9. Type a price. The table replaces the card, and no prices appear before the first response lands.
10. In **Rule Details**, find `Deal ID` below `Applies to`. The value is monospace and not editable.
11. Click `Copy`. The ID is on the clipboard and a snackbar reads "Deal ID copied to clipboard.". Click it again: the toast is replaced, not stacked.
12. Create a new rule. There is no Deal ID row until the rule is saved.

### Other information:

* [x] Have you written new tests for your changes, as applicable? 69 new tests across ten files, covering the picker's timezone handling and its use for the eligibility date, the table's column behaviour, all three empty states, the stat strip's locked and unlocked branches, the read-only rows, the copy control's success, failure, insecure-context and repeat-copy paths, and the preview's loading and failure paths.
* [x] Have you successfully run tests with your changes locally? 76 suites / 686 tests in `newspack-plugin` and 14 / 110 in `newspack-components`, plus `lint:js`, `lint:scss`, `lint:php` and a production build. Verified in a browser against a local site.

<details>
<summary><strong>For AI / reviewers going deeper</strong></summary>

**New components**, all in `src/wizards/audience/views/pricing-rules/`:

| File | Responsibility |
| --- | --- |
| `datetime-field.tsx` | `BaseControl` → `Dropdown` → `DateTimePicker`. Value and `onChange` are both `YYYY-MM-DDTHH:mm` browser-local wall clock, or `''`. |
| `readonly-field.tsx` | The labelled non-editable row shared by Goal and Deal ID. Goal was refactored onto it. |

**The copy control reports through the wizard store.** `ADD_NOTICE` appends without deduping and `REMOVE_NOTICE` filters by id, so `CopyDealId` calls `removeNotice()` with the same id immediately before each `addNotice()`; without it a second copy stacks a toast sharing the first one's React key. `speak()` is deliberately absent — `WizardSnackbar` already maps `type` to politeness, so calling both would announce twice. A test clicks Copy twice and asserts one notice survives.
| `impact-stats.tsx` | The stat strip. Replaces the deleted `AudienceLine`. |
| `impact-empty.tsx` | One card, three wordings, keyed by `'no-price' \| 'no-products' \| 'unsupported'`. |

`packages/components`'s `DataViews` wrapper now forwards the base component's subcomponents, so the impact table can compose `<DataViews.Layout />` on its own and skip the default toolbar entirely. Purely additive for the other consumers.

**Why the picker does not shift dates.** `DateTimePicker`'s `inputToDate()` appends `Z` to a timezone-less string and builds a `UTCDateMini`; `onChange` emits `format(date, "yyyy-MM-dd'T'HH:mm:ss")` off that same value, so civil digits in give the same civil digits out. The button label uses `gmdateI18n(fmt, value + 'Z')`, not `dateI18n`, which would resolve against the site timezone and land a day out either side of midnight. `datetime-field.test.jsx` pins this with a `-5` offset fixture; swapping to `dateI18n` turns it red.

**`ImpactTable` kept its prop signature** (`baseline`, `segmentGroups`, `currency`), which is why `catalog-impact.tsx` is untouched. Fields carry `enableHiding: false`. That is load-bearing, not decoration: "Hide column" lives in the per-column header menu rather than the toolbar, so dropping the toolbar did not remove it, and a hidden column would reappear whenever the segment set changed. Removing the affordance is what makes the derived `view.fields` safe. A test asserts the column menu offers sorting and no hide option, so re-enabling it fails loudly.

**The segment-column tint** is a `background-color` on `<col class*="__col-seg-">`, because a per-cell background cannot fill the cell's padding. That couples to a DataViews-internal class name; if it changes upstream the tint disappears silently and nothing else breaks. A test pins the `<col>` element.

**Two guards in `rule-preview.tsx` earn their keep.** `hasPrice` is in the fetch effect's dep array because `Number('') || 0` and `Number('0') || 0` are both `0`, so typing a literal `0` leaves the request body byte-identical and the effect would never re-run. And `hasResolved`, set in `.finally()`, is what lets a rejected request reach the unavailable card instead of rendering nothing.

**The empty guard checks the sample, not just the count.** `total_matching` and `sample` diverge: the engine subtracts only the exclusions it sees inside the sample window, so a rule whose first N candidates are all grouped or engine-excluded returns a positive count with an empty sample. The engine's own metabox guards both; the React port had dropped half of it.

**The eligibility date hides its label rather than dropping it.** `conditions.tsx` renders the custom date under a selector already named by the matcher, so `DateTimeField` takes a `hideLabelFromVision` flag and omits the visual label when it is set. The toggle names itself through `aria-label`, which carries `<matcher label>: <resolved value>`, distinct from the selector above it and specific to the matcher if the engine ever registers a second datetime condition. The stored value is untouched: `tsToLocalInput` and `localInputToTs` already spoke the field's `YYYY-MM-DDTHH:mm` wall clock.

</details>


plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-preview.tsx
```diff
@@ -79,38 +72,42 @@ export default function RulePreview( { body }: RulePreviewProps ) {
...
-									data.sample_count,
-									data.total_matching
-							  )
-							: __( 'Resulting prices across affected products (best price wins; updates as you edit).', 'newspack-plugin' ) }
-					</p>
-					<ImpactTable baseline={ data.sample } segmentGroups={ data.segment_groups ?? [] } currency={ currency } />
-				</>
-			) : (
-				<p className="newspack-pricing-rules__muted">{ __( 'This rule does not affect any products yet.', 'newspack-plugin' ) }</p>
+			<ImpactStats totalMatching={ preview.total_matching } countLimited={ preview.count_limited } audience={ preview.audience } />
```
miguelpeixe: **Blocker:** The "Products affected" headline can be capped and exact-looking, with no way to say so — The diagnosis in the PR body checks out: in `Pricing_Impact_Calculator::impact_preview()`, products failing `matches_product()` hit `continue` without incrementing `$excluded`, so only grouped and engine-excluded products are subtracted from the total. The trigger is not exotic — for a `category` scope, `candidates()` runs a `tax_query` whose `include_children` defaults to true while `Category_Scope_Matcher::matches()` intersects only directly-assigned terms, so every product filed under a sub-category is counted in the total, correctly excluded from the sample, and correctly excluded at checkout.

Tracing it further turned up a second problem on the same figure, and this one hits the default configuration. For the `all_subscriptions` scope — which all four named goals set — `candidates()` overwrites the exact catalog total with `count( $ids )`, the size of a list already capped at the sample limit. And the per-rule preview route reports `count_limited` from a key `impact_preview()` never sets, so it is always false. `preview_limited` is then derived from the same capped total, so it cannot fire either. The result is that on the most common setup the headline can be a capped number rendered as an exact one, with no signal anywhere in the payload to say otherwise — and `bounded()` has nothing to work with.

The empty guard handles the degenerate case of a positive total with an empty sample, which is right and worth keeping. But this moves the figure from a parenthetical into the largest type on the panel, and the previous copy printed no total at all in the capped case, so the wrong number is newly visible. Either have the preview route report the true total and a partial flag for that scope, or label the figure as describing this preview rather than the rule's reach until it can.

packages/components/src/card/core-card.js
```diff
@@ -43,6 +43,7 @@ const CoreCard = ( {
 	isNarrow,
 	isSmall,
 	isVertical,
+	size,
```
miguelpeixe, line 53: **Suggestion:** size="large" is a partial parallel scale that shadows WP Core's complete one — The instinct is right, and it is the direction I asked for on #757 — padding belongs in the component, not in a consumer override. The objection is to the shape.

`CoreCard` now carries two overlapping sizing APIs: the legacy `isSmall` boolean and a `size` string whose only implemented value is `'large'`, with `isSmall` silently winning via `bodySize = isSmall ? undefined : size`. WP Core's `Card` already exposes a full `xSmall|small|medium|large` scale that this shadows. And the new rung does not follow the existing shape: the body scale is `16px 24px` by default and `16px` when small — asymmetric throughout — while `is-large` is a uniform 48px, a 3× vertical jump with no intermediate step. Every value is on the WordPress grid, so nothing is a token violation; extending a scale is just a designer call rather than an implementation detail. Mapping `isSmall` onto `size` internally and exposing one prop would avoid minting a second scale.

packages/components/src/section-header/style.scss
```diff
@@ -95,6 +95,12 @@
 		@include wp-mixins.heading-x-large();
 	}
 
+	&#{&}--small {
```
miguelpeixe, line 98: **Suggestion:** size="small" is only half a variant — The title rule sits at the component root, so any `size="small"` header gets the 15px title. The icon rules — 40px circle, 4px bottom margin — are nested inside `&.newspack-section-header--page-header`, while the JS change that drops the icon to 24px is unconditional. So a `size="small"` header *without* `pageHeader` renders a 24px icon with no matching container.

The values themselves are right: 40/24 preserves the default's 80/48 ratio exactly and both sit on the scale. The issue is predictability — a shared component's size prop should mean the same thing regardless of which other flag is set, or say in its docs that it is `pageHeader`-only.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/conditions.tsx
```diff
@@ -81,35 +81,21 @@ function DatetimeCondition( {
...
 			const inForce = customTs ?? value ?? resolvePublish();
 			setCustomTs( inForce );
 			onChange( inForce );
 		}
 	};
 
-	// A datetime-local reads as empty until every segment is filled, so mid-edit it
-	// reports null. Hold the last good date rather than storing the permissive gate.
+	// Clearing stores null, so the selector has to follow it to Anytime.
 	const changeCustom = ( s: string ) => {
```
miguelpeixe, line 111: **Suggestion:** Clearing the custom cohort date drops focus to the document body — `changeCustom( '' )` sets `mode` to `'none'`, which unmounts the whole `DateTimeField` subtree rather than just the popover. `Popover`'s focus return only fires for a toggle that is still connected, so when the toggle is torn down in the same commit the call is a no-op and focus lands on `<body>` — the next Tab restarts from the top of the page, several controls away from the `SelectControl` that now reads "Anytime".

The `Starts`/`Ends` fields are unaffected, since Clear there only empties the value and the field stays mounted. Moving focus to the mode `SelectControl` after clearing would close it.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/datetime-field.tsx
```diff
@@ -0,0 +1,117 @@
...
+	const ariaDescribedBy = [ help ? `${ id }__help` : null, describedBy ].filter( Boolean ).join( ' ' ) || undefined;
+
+	return (
+		<BaseControl id={ id } help={ help } __nextHasNoMarginBottom>
+			{ ! hideLabelFromVision && <BaseControl.VisualLabel>{ label }</BaseControl.VisualLabel> }
+			<Dropdown
+				className="newspack-pricing-rules__datetime"
+				popoverProps={ { placement: 'bottom-start', offset: 4 } }
+				renderToggle={ ( { isOpen, onToggle } ) => (
+					<Button
```
miguelpeixe, line 78: **Suggestion:** Schedule fields read as accent-coloured buttons, and an unset value looks like a set one — `variant="secondary"` is not neutral chrome — it is accent-coloured text plus an inset 1px ring in the same colour, and Newspack sets `--wp-admin-theme-color` to `primary-600` (#003da5). Stretched to full width with `justify-content: flex-start`, `Starts` and `Ends` sit directly below the screen's `SelectControl`s but render as navy-outlined, navy-text buttons rather than fields. I checked for a neutralising override and there is none — the only rule targeting this toggle stretches and left-aligns it.

The reading problem is the sharper one: `display` is a single node for both states, so the unset placeholders ("Active immediately", "No end date") get exactly the same colour and weight as a real date. Every other field on the screen mutes its placeholder — including `ReadonlyField` in this same PR, which passes its placeholder to an `InputControl` and gets the browser's muted rendering for free. So an empty schedule currently reads as loudly as a set one, which inverts the signal the change is adding.

This is visual only: the toggle carries a full accessible name, `aria-expanded` and `aria-describedby`, so nothing here fails accessibility and nothing is broken. It is a few SCSS lines under `&__datetime > button` — neutral border, text-coloured value, muted placeholder on the unset branch — and no test pins the styling, so there is no churn. Given this PR's remit is design conformance, it is the one I would fix before merge.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/datetime-field.tsx
```diff
@@ -0,0 +1,117 @@
...
+						onClick={ onToggle }
+						disabled={ disabled }
+						__next40pxDefaultSize
+					>
+						{ display }
+					</Button>
+				) }
+				renderContent={ ( { onClose } ) => (
+					<VStack spacing={ 2 }>
+						<DateTimePicker
```
miguelpeixe, line 100: **Suggestion:** The first pick on an empty schedule field can land on the wrong day — The PR's `gmdateI18n( value + 'Z' )` reasoning is correct and verified — `dateI18n` really would resolve against the site zone and land a day out near midnight, and the `tsToLocalInput`/`localInputToTs` round trip is digit-stable. That holds while `currentDate` is supplied. It stops holding when the value is empty, which is exactly the state a new rule starts in: `currentDate={ value || undefined }` leaves the picker with no seed.

**Worth flagging that the precise mechanism is unsettled.** Reading the component source suggests the picker seeds from the *site* timezone while the field stores a browser wall clock, which would make an accepted pre-fill land hours out on a site whose zone differs from the editor's. Running the bundled `@wordpress/components` picker across six timezones did not reproduce that, but did surface a different off-by-one on the same code path: with no seed, the calendar's day labels render in UTC while the cells select browser-local dates, so for Lisbon, Berlin, Nairobi and Sydney the cell that emits `2026-08-12` is labelled `11`. UTC and Los Angeles are correct. It self-corrects once the field holds a value, which is why the suite misses it — every test passes a value in.

The two accounts disagree on cause but agree on trigger, blast radius and fix: it is the first interaction with an empty field, and seeding `currentDate` with the current wall clock in the shape the field stores keeps the popover on its value-backed path and closes both. Since these values feed `active_from`/`active_until`, which are compared against the server clock to decide whether a rule is live, a day's drift means the discount is not offered on the campaign's first day and is still offered the day after it should end.

One part reproduced cleanly either way: `DateTimePicker` mounts a `<TimeZone/>` badge showing the *site's* abbreviation beside digits the help text calls "your local timezone". The native input this replaces carried no such badge. Separately worth knowing: a spring-forward gap time is silently rewritten by `new Date()` in `localInputToTs`.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-empty.tsx
```diff
@@ -0,0 +1,49 @@
...
+		icon: info,
+		title: __( 'Preview unavailable', 'newspack-plugin' ),
+		body: __( 'The pricing engine did not return a preview.', 'newspack-plugin' ),
+	},
+} );
+
+export default function ImpactEmpty( { reason }: { reason: ImpactEmptyReason } ) {
+	const { icon, title, body } = getReasons()[ reason ];
+	return (
+		<Card className="newspack-pricing-rules__empty" size="large" __experimentalCoreCard>
```
miguelpeixe, line 40: **Suggestion:** The Institutions layout is viewport-based, so nesting it in a card squeezes the copy — The reference pattern is real, but `Grid columns={4}` with `[start="2"][end="4"]` is a *page*-level centring device: those attribute selectors resolve against the viewport, not the container. Institutions applies it at page width; here it sits inside a `Card size="large"` that already contributes 48px of padding per side, so above about 1054px the copy is squeezed into roughly half of an already-narrow card and then inset again.

Two smaller fidelity gaps: `VStack spacing={8}` has a single child, so the gap it exists for never renders, and the reference pairs its header with a primary action — this has none, which leaves the body copy ("Enter a price above", "Widen 'Applies to'") doing the call-to-action's job in prose.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-format.ts
```diff
@@ -38,3 +38,16 @@ export function describeResulting( row: CatalogImpactRow, currency: PricingRules
 	}
 	return row.segments.map( seg => formatSegment( seg, currency ) ).join( ' · ' );
 }
+
+/**
```
miguelpeixe, line 42: **Suggestion:** A second formatCount, one directory from the file that already imports the shared one — `packages/components/src/breadcrumbs/format-count.js` already does this job, and the sibling `pricing-rules/list.tsx:25` already imports it — as do three other wizard files.

The two are behaviourally equivalent in practice, so this is about duplication rather than a defect. I checked the one visible difference and it does not bite: the shared helper normalises `_` to `-` before handing the tag to `Intl` and this copy passes `document.documentElement.lang` raw, but `get_bloginfo( 'language' )` rejects anything outside `[a-zA-Z0-9-]` and falls back to `determine_locale()` with underscores already replaced, so wp-admin can only ever emit a hyphenated tag. The shared version's other edge, a `typeof document !== 'undefined'` guard, matters only for server-side rendering these views never see.

The reason to consolidate is drift, and one concrete instance of it: the locale-fallback issue I raised on #757 now lives in both copies, so fixing it there would leave this one behind.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-stats.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export default function ImpactStats( { totalMatching, countLimited, audience }: ImpactStatsProps ) {
+	const scope = audience?.supported ? audience : null;
+	const isLocked = 'locked' === scope?.application;
+
+	return (
+		<div className="newspack-pricing-rules__stats">
+			<HStack justify="flex-start" alignment="top" spacing={ 8 }>
+				<Stat value={ bounded( totalMatching, countLimited ) } label={ __( 'Products affected', 'newspack-plugin' ) } />
+				{ scope && <Stat value={ bounded( scope.total, scope.count_limited ) } label={ __( 'Subscribers in scope', 'newspack-plugin' ) } /> }
```
miguelpeixe, line 65: **Suggestion:** Two of the three subscriber counts are exact figures derived from a truncated walk — `bounded()` is applied to `scope.total`, but `formatCount()` is applied bare to `scope.caught` and `scope.protected`, and `impact-stats.test.jsx:69` pins that split as deliberate ("marks a capped count as a lower bound, and only that count").

`Subscription_Audience` is explicit that its ceiling truncation is *directional*: `wcs_get_subscriptions_for_product` orders by order id ascending and `split()` keeps the first 2,000, so a truncated split under-reports `caught` and over-reports `protected` — identically on every request. Above the ceiling, `caught` and `protected` are therefore lower bounds computed from a truncated sample, and they always sum to exactly the bound shown beside them, yet only the total carries a `+`.

Worth being precise about the trigger, because the obvious reading of it is wrong: no named goal sets a `subscription_started_after` gate (`recipes.ts` sets only `first_time_only`, `lapsed_subscriber` and `pending_cancellation`, and retention sets none), and named goals do not render the datetime condition at all. So the *skewed* case — eligible understated, protected overstated — needs the Custom path, or an already-saved rule, with a past-dated cohort gate and a store above the ceiling.

There is a second shape that does hit named goals, though. With no cohort date the threshold is zero, so every inspected subscription counts as eligible and `caught` equals the capped total — meaning the panel prints the same number twice, once with a `+` and once without, directly beside each other. Whichever shape applies, the truncation only ever moves these figures in the reassuring direction.

Applying `bounded()` to `caught`, or adding one caveat line when `count_limited` is set, would close it cheaply. Note `protected` is over-reported, so a `+` would be the wrong marker there. This is a non-binding preview rather than anything that writes, which is why it is a suggestion and not a blocker.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
```diff
@@ -8,15 +8,20 @@
...
+ * present, so a segment named for existing subscribers isn't misread as
+ * modeling their lifecycle (NPPD-1853).
  */
 
 /**
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
+import { useState, useEffect, useMemo } from '@wordpress/element';
+// Not the Newspack wrapper: with-wizard-screen/style.scss gives `.newspack-dataviews`
```
miguelpeixe, line 21: **Suggestion:** Bypassing the shared DataViews wrapper also drops the stylesheet import — The reason given for going direct is real — the `-48px` bleed is a page-layout rule scoped to `.newspack-dataviews` outside full-width content, and an embedded table should not inherit it.

The side effect is that `packages/components/src/dataviews/style.scss` is the repo's only import of `@wordpress/dataviews/build-style/style.css`, and it is reached only by importing the Newspack wrapper. This table is styled today purely because `pricing-rules/index.tsx` statically imports the list alongside the editor, pulling the wrapper into the same chunk. Code-splitting the editor away later would leave it unstyled with no test failing. A modifier on the shared wrapper — an embedded variant that opts out of the bleed — keeps every Newspack DataViews reachable from one place. Worth noting `integrations/logs-view.js` already composes DataViews children itself while keeping the wrapper, so the composition is fine; the bypass is what is new.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
```diff
@@ -67,54 +72,110 @@ interface ImpactTableProps {
...
+			} ) ),
+		],
+		[ columns, currency ]
+	);
+
+	const fieldIds = useMemo( () => [ 'regular', ...columns.map( col => col.key ) ], [ columns ] );
+	const fieldIdsKey = fieldIds.join( '|' );
+
+	// The server already caps the sample, so show all of it rather than re-truncating.
+	const perPage = Math.max( baseline.length, 1 );
```
miguelpeixe, line 138: **Suggestion:** perPage is derived outside the view, so one frame paints truncated — `perPage = Math.max( baseline.length, 1 )` is computed each render but only reaches `view` through the effect, which runs after commit. On the render where a preview response arrives with more rows than the previous sample, `filterSortAndPaginate` runs against the stale `view.perPage` and paints a truncated table for one frame before the effect corrects it.

Folding `perPage` and `fields` into a memoised view object passed straight to `filterSortAndPaginate`, and keeping state only for what the user can change, removes the extra pass. The rest of the DataViews usage checks out — `titleField` correctly kept out of `view.fields`, `enableSorting: false` on segment columns so a vanishing segment cannot strand `view.sort`, and primitive effect deps so it cannot loop.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
```diff
@@ -67,54 +72,110 @@ interface ImpactTableProps {
...
+
+	const [ view, setView ] = useState< View >( () => ( {
+		type: 'table',
+		page: 1,
+		perPage,
+		search: '',
+		filters: [],
+		layout: { density: 'compact', enableMoving: false },
```
miguelpeixe, line 146: **Suggestion:** compact density is a first for this plugin — Every other DataViews in `newspack-plugin` uses the default density — the rules list, the plans list, setup emails, and content-gate institutions. The only `compact` in the monorepo is in newspack-story-budget.

A denser embedded read-only preview is arguable, but it means the Impact Preview table and the Rules list table — same feature, two clicks apart — have different row rhythms. Worth making that a deliberate call rather than a local default.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
```diff
@@ -67,54 +72,110 @@ interface ImpactTableProps {
...
-									<td key={ col.key } className={ cellClass || undefined }>
-										<ResultingCell row={ cell } currency={ currency } />
-									</td>
-								);
-							} ) }
-						</tr>
-					) ) }
-				</tbody>
-			</table>
+			<div
```
miguelpeixe, line 161: **Suggestion:** The table lost its accessible name and gained a landmark — The `screen-reader-text` `<caption>` became a `role="region"` on the wrapper. DataViews' `<table>` takes no `aria-label`, `aria-labelledby` or caption of its own, so the table itself is now unnamed — a screen-reader user listing tables on the page gets an unlabelled one — and the page gains a landmark for what is an inline preview.

`aria-labelledby` pointing at the existing "Impact Preview" heading would name the table without minting a landmark.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/readonly-field.tsx
```diff
@@ -0,0 +1,51 @@
+/**
```
miguelpeixe, line 1: **Suggestion:** Two generic controls living in a feature directory, with feature-scoped class names — `ReadonlyField` is the right answer to the #757 feedback and fixes that contrast issue outright — the `$gray-800` override is gone, so the value now inherits `InputControl`'s `$gray-900` on `$gray-100`. Neither it nor `DateTimeField` knows anything about pricing rules, and both already have consumers in two files.

The coupling is baked into the CSS: each hard-codes a `newspack-pricing-rules__*` class whose rules live in this screen's stylesheet, so reusing either elsewhere means carrying the pricing-rules class names or duplicating the CSS. `packages/components/src/` is where these belong. One detail to carry with the move: the read-only fill selects `.components-input-control__container`, a WordPress-internal class, which will break silently on a components bump. And if a read-only value surface is going to exist system-wide, this is the moment to decide whether that `$gray-100` earns a named token — the open question from the last review.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -88,6 +90,57 @@ function seedConditions( raw: ConditionsMap | undefined ): ConditionsMap {
...
+		let copied = false;
+		try {
+			await navigator.clipboard.writeText( dealKey );
+			copied = true;
+		} catch {
+			copied = legacyCopy();
+		}
+		// Notices append without deduping, so a second copy would stack a toast sharing
+		// the first one's React key.
+		removeNotice( COPY_NOTICE_ID );
```
miguelpeixe, line 127: **Suggestion:** A repeated copy is neither re-announced nor re-timed — The `removeNotice` → `addNotice` pairing is correct against the reducer, and the politeness reasoning holds — `WizardSnackbar` maps type to politeness and core's `Snackbar` speaks it, so omitting `speak()` is right for the first copy.

The announcement only fires on *change*, though. Core keys that effect on the message and politeness, and the auto-dismiss effect on `explicitDismiss`. Because `useSelect` runs through `useSyncExternalStore` under a concurrent root, the two dispatches batch into one commit, the intermediate removed state never renders, and the keyed element survives. So a second Copy in a row is silent for screen-reader users, and its toast still expires on the *first* copy's timer. A success-to-error transition does re-announce, since the message changes.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -506,6 +545,17 @@ export default function RuleForm( { isNew, initialPath = null, rule, vocab, onDo
 						__next40pxDefaultSize
 					/>
 					<ScopeTargets scopeType={ scopeType } value={ scopeIds } onChange={ setScopeIds } />
+					{ ! isNew && rule && (
+						<ReadonlyField
+							id="newspack-pricing-rule-deal-id"
+							label={ __( 'Deal ID', 'newspack-plugin' ) }
```
miguelpeixe, line 572: **Suggestion:** The field still reads "Deal ID", but the description and test plan call it "Rule ID" — The description's third section is headed "The Rule ID is a field, not a footnote" and argues that the engine's own term "made publishers learn a second word for a thing this screen calls a rule". Test step 10 then says: "In **Rule Details**, find `Rule ID` below `Applies to`."

The string `Rule ID` does not appear anywhere in the diff. The shipped label is `Deal ID`, and consistently so — this field label, the copy button's `aria-label`, both snackbar messages, the help text, the list column at `list.tsx:141`, and the tests.

The rest of that section's claims do hold: the two incidental strings really did move from "deal" to "rule" (`"Optional. Describe this rule's goal in your own words."` and the price-lock help). So this reads like a rename that was decided, applied to its neighbours, and not applied to the field the section is actually about — or one that was reconsidered, leaving the description stale. Keeping `Deal ID` is defensible in itself, since it matches the analytics vocabulary the help text points at.

Either way it is worth reconciling before merge: as written, a reviewer or QA following the test plan looks for a label that is not on the screen. If the intent is to keep the engine's term, the section heading and step 10 are what want updating.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -757,17 +816,10 @@ export default function RuleForm( { isNew, initialPath = null, rule, vocab, onDo
...
-					) }
+					description={ __( 'How this rule prices products, composed with your other active rules. Best price wins.', 'newspack-plugin' ) }
 					noMargin
 				/>
-				{ ! isSchedule && String( value ).trim() === '' ? (
-					<p className="newspack-pricing-rules__muted">{ __( 'Enter a price to see the impact preview.', 'newspack-plugin' ) }</p>
-				) : (
-					<RulePreview body={ previewBody } />
-				) }
+				<RulePreview body={ previewBody } hasPrice={ isSchedule || String( value ).trim() !== '' } />
```
miguelpeixe, line 844: **Suggestion:** The stepped model always reports as having a price, so its empty state is unreachable — `hasPrice={ isSchedule || String( value ).trim() !== '' }` short-circuits on `isSchedule`, so choosing "stepped by cycle" fires the preview immediately even though `previewBody.steps` filters to `[]` while every step is blank.

The rule then contributes nothing, so the engine returns unchanged regular prices and the publisher gets a full stat strip and a table of "no change" rows — for a schedule that `POST /rules` rejects outright and that `submit()` already blocks with "Add at least one schedule step with a value." The "No price set yet" card, which is the point of the tri-state rewrite, is unreachable for half the pricing models. `isSchedule ? steps.some( s => String( s.value ).trim() !== '' ) : String( value ).trim() !== ''` would close it.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -59,34 +86,26 @@
...
 		&.is-loading {
 			opacity: 0.6;
 		}
 
-		.newspack-pricing-rules__impact-table .is-segment-col {
+		> p {
+			margin-top: wp-vars.$grid-unit-15;
+		}
+
+		col[class*="__col-seg-"] {
```
miguelpeixe, line 116: **Suggestion:** The segment tint never paints in the header and disappears on hover — The class name itself is sound — DataViews does emit `dataviews-view-table__col-<id>`, and the test queries the same substring, so an upstream rename would fail CI.

What the test does not cover is whether the tint is *visible*. CSS paints column backgrounds below row and cell backgrounds, and the shipped DataViews stylesheet sets `thead th { background-color: #fff }` and `tr.is-hovered { background-color: #f8f8f8 }`, with `is-hovered` applied on plain mouseenter. So the tint never shows in the header row, and it vanishes from whichever row the pointer is over. The cell-level `is-segment-col` class this replaces had neither problem.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
```diff
@@ -59,34 +86,26 @@
...
 			background: rgba(wp-colors.$alert-yellow, 0.12);
 		}
```
miguelpeixe, line 118: **Suggestion:** The tint also couples to a DataViews-internal class via a substring match — `col[class*="__col-seg-"]` depends on DataViews generating one `<col>` per field with a class derived from the field id — not part of that package's public API, and it moved as recently as the compound-component refactor. The substring form would also catch any future field whose id contains that fragment.

DataViews exposes no way to mark a column as a segment, so this may genuinely be the only route. If it stays, it is worth an explicit comment naming the coupling — the wrapper bypass in this same PR got one — and worth raising upstream, rather than sitting as an unremarked selector. Keeping the test that pins the `<col>` is what makes it survivable.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/conditions.tsx
```diff
@@ -127,14 +113,17 @@ function DatetimeCondition( {
 				__next40pxDefaultSize
 			/>
 			{ 'custom' === mode && (
-				<TextControl
-					label={ matcher.label }
+				<DateTimeField
+					id={ `newspack-pricing-rule-condition-${ matcher.id }` }
+					label={ sprintf(
```
miguelpeixe: **Nit:** Nested sprintf produces a double-colon accessible name — The condition passes `sprintf( __( '%s: custom date' ), matcher.label )` as `label`, which `DateTimeField` then feeds into `_x( '%1$s: %2$s', … )`. The result reads "Subscriptions started on or after: custom date: August 12, 2026 9:00 am", and translators see two fragments without ever seeing the whole sentence.

Everything else on the i18n side is clean — correct textdomain throughout, translator comments on every `sprintf`, `_n` given the real count, `_x` where context is genuinely needed, and no string built by concatenation.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-stats.tsx
```diff
@@ -0,0 +1,72 @@
...
+				__( '%s+', 'newspack-plugin' ),
+				formatted
+		  )
+		: formatted;
+};
+
+function Stat( { value, label, note }: { value: string; label: string; note?: string } ) {
+	return (
+		<VStack spacing={ 0 }>
+			<span className="newspack-pricing-rules__stat-value">{ value }</span>
```
miguelpeixe, line 49: **Nit:** A stat number carries the same type weight as the section heading — `__stat-value` is `heading-x-large()` (20px/24px), which is exactly `SectionHeader__title`'s default — so a stat number reads as heavily as the "Impact Preview" heading above it. Meanwhile the `size="small"` heading this PR adds for the empty card is 15px, a step *below* a stat number, in the panel it replaces.

All three are on the WordPress ramp, so nothing is a token violation; the assignment is what wants a designer's eye. Contrast is fine throughout — `$gray-700` on white passes AA at both 13px and the 12px note.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -669,27 +719,36 @@ export default function RuleForm( { isNew, initialPath = null, rule, vocab, onDo
...
+								id="newspack-pricing-rule-active-until"
+								label={ __( 'Ends', 'newspack-plugin' ) }
+								describedBy={ SCHEDULE_HELP_ID }
+								value={ activeUntil }
+								placeholder={ __( 'No end date', 'newspack-plugin' ) }
+								disabled={ isSaving }
+								onChange={ setActiveUntil }
+							/>
+						</Grid>
+						<p className="newspack-pricing-rules__muted newspack-pricing-rules__muted--help" id={ SCHEDULE_HELP_ID }>
```
miguelpeixe, line 764: **Nit:** The shared schedule help re-implements WordPress's help-text style — `__muted--help` is `$gray-700` at `$helptext-font-size` — exactly `.components-base-control__help`. Because the help is shared by two fields there is no single `BaseControl` to hang it on, so a local element is defensible; duplicating the component's declarations rather than reusing its class is what will let the two drift.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -757,17 +816,10 @@ export default function RuleForm( { isNew, initialPath = null, rule, vocab, onDo
 			<div className="newspack-pricing-rules__preview-section">
```
miguelpeixe, line 838: **Nit:** Orphaned wrapper class — `newspack-pricing-rules__preview-section` survives the rewrite but its only rule (`padding: 0 0 16px`) was deleted from the stylesheet, so the section quietly loses its bottom padding while the hook stays. Either the class or the div can go.

miguelpeixe (changes requested):
This is careful work and the reasoning in the description mostly holds up under checking. I verified the load-bearing claims rather than taking them from the body: the `DateTimePicker` round trip is genuinely lossless for the stored UTC timestamp, `hasPrice` really does need to be in the dep array, the `removeNotice`/`addNotice` pairing matches the store's reducer, `enableHiding: false` really is what makes the derived `view.fields` safe, and the fetch effect's race handling is correct — each run owns its own cancelled flag, so a stale response can't overwrite a newer one. The new CSS has no hardcoded colours or off-scale spacing, and the 69 tests assert behaviour rather than implementation almost throughout.

One thing I'd like resolved before merge, and it's an extension of something you already flagged yourself.

**The "Products affected" headline can be capped while looking exact.** Your note about `matches_product()` failures not being subtracted is right. Following it further, there's a second issue on the same figure that hits the default setup rather than an edge case: for the `all_subscriptions` scope — which all four named goals set — `candidates()` overwrites the exact catalog total with `count( $ids )`, the size of a list already capped at the sample limit. And the per-rule route reads `count_limited` from a key `impact_preview()` never sets, so it's always false, which means `bounded()` can't render the `+` and `preview_limited` can't fire either. On that path the number has no way to admit it's partial. Since the old copy printed no total at all in that case, this is newly visible — and it's the figure a publisher sizes a price change against. Either have the route report the true total plus a partial flag for that scope, or word the stat as describing this preview rather than the rule's reach until that lands.

A few others worth a look:

- **The subscriber tiles have the same shape of problem.** `bounded()` is applied to `scope.total` but not to `caught`/`protected`. `Subscription_Audience` documents that truncation as directional, and with no cohort date — every named goal — the threshold is zero, so `caught` equals the capped total and prints as a hard number right beside that same total with a `+`.
- **The first pick on an empty schedule field can land a day out.** Reading the source suggests the picker seeds from the site timezone; running the bundled picker across six timezones didn't reproduce that but did show the calendar labelling cells in UTC while selecting browser-local dates, so east of UTC the cell that emits the 12th is labelled 11. It self-corrects once a value is set, which is why the tests miss it — they all pass a value in. Seeding `currentDate` closes it either way. Worth a manual check east and west of UTC before merge, since these values decide when a rule is live.
- **Clearing the custom cohort date drops focus to `<body>`**, because setting mode to `'none'` unmounts the toggle the popover would return focus to. The schedule fields are fine — only the condition unmounts the whole control.
- **The schedule fields read as accent-coloured buttons**, and the unset placeholder is styled identically to a real date, where every other field on the screen mutes its placeholder. Nothing is broken and the control is fully labelled, but it's the design item I'd fix given what this PR is for.

The rest are consolidation asks — the duplicated `formatCount`, the two generic controls that want to live in `packages/components`, and the shared DataViews wrapper bypass that also drops the only import of the DataViews stylesheet.

- `plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:62` — **Suggestion:** The two ImpactTable surfaces now speak differently — The editor preview routes empties to the new `ImpactEmpty` card and groups its counts; the catalog panel above the rules list keeps a bare muted paragraph, raw `%d` in "%d products" and "Showing a sample of %d.", and its own heading element. DataViews' own `empty` prop is a third, unstyled empty treatment that only the catalog panel can reach. Keeping `ImpactTable`'s prop signature is what made `catalog-impact.tsx` untouched, which is a reasonable scoping decision — but the result is one feature with two vocabularies. Bringing the catalog panel along would close it.

- `plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:62` — **Suggestion:** The two ImpactTable surfaces now speak differently — The editor preview routes empties to the new `ImpactEmpty` card and groups its counts; the catalog panel above the rules list keeps a bare muted paragraph, raw `%d` in "%d products" and "Showing a sample of %d.", and its own heading element. DataViews' own `empty` prop is a third, unstyled empty treatment that only the catalog panel can reach. Keeping `ImpactTable`'s prop signature is what made `catalog-impact.tsx` untouched, which is a reasonable scoping decision — but the result is one feature with two vocabularies. Bringing the catalog panel along would close it.

Review provenance: 3 of 4 reviewer passes (lineup: deep, wp-expert, design, codex); release risk: HIGH (degraded — skipped: codex).

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-preview.tsx
thomasguillot: Fixed engine-side, on `fix/per-rule-audience-total` in `Automattic/woocommerce-dynamic-pricing`.

`candidates()` now paginates the All Products query and sums the two exact totals, instead of counting a merged list that was already capped. The two sets are disjoint by construction, since schemes are what give a non-subscription product type a subscription. `impact_preview()` sets `count_limited` from `is_truncated()`, so `bounded()` finally has something to render and `preview_limited` follows a real total. The injected-candidates seam stays false, since the caller supplied the exact set.

The key reads as "the match set outran the window this walked", matching what the catalog route already means by it. Say if you would rather it meant something narrower.

Three tests cover it, and all three fail without the change. 395 pass with it.

Leaving this open until that lands.

packages/components/src/card/core-card.js
thomasguillot: Leaving the scale as it is. Card is being reworked onto `@wordpress/ui` under DSGNEWS-215, so folding `isSmall` into `size` now is work that gets thrown away at the rewrite, and risks minting a third scale in between. Better done there, against whatever scale wp/ui settles on.

packages/components/src/section-header/style.scss
thomasguillot: Fixed in bc0887a26. The icon rules now sit at the component root rather than inside the `page-header` block, so `size` scales the glyph and its container together whatever `pageHeader` is doing. The prop docs say so now too.

Measured both variants on a live page after the move: small resolves to 40px with a 4px bottom margin, default to 80px with 8px, both keeping the circle, background and fill. Both current `icon` consumers pass `pageHeader`, so nothing moves for them.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/conditions.tsx
thomasguillot: Fixed in ee8ae818d. Focus moves to the mode `SelectControl` after a clear, with a test asserting it lands there.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/datetime-field.tsx
thomasguillot: Holding this one. The read is right, and the placeholder reading as loudly as a real value is the half worth fixing, but it is a visual decision rather than a defect, so I would rather take it deliberately than fold it into a feedback pass.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/datetime-field.tsx
thomasguillot: Fixed in c45e6939b. `currentDate` seeds from now in the same wall-clock shape the field stores, so an empty field behaves like one that already holds a value. That closes it whichever of the two mechanisms is the real one.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-empty.tsx
thomasguillot: Fixed in 9aba2ca82. The Grid is gone, the header takes a max-width, and the card centres it, so nothing resolves against the viewport any more. The single-child `VStack` went with it.

Worth recording: taking the Grid out exposed a second problem underneath it. `impact-empty` passes `noMargin`, and its `margin: 0 !important` overrides the auto margins `page-header` centres with, so the content sat left-aligned until the container centred it explicitly. The Grid had been hiding that.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-format.ts
thomasguillot: Already consolidated further along the stack: `impact-format.ts` imports the shared helper and re-exports it, so there is one implementation and the locale fix cannot drift. It lands with that PR rather than this one.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-stats.tsx
thomasguillot: Already closed further along the stack: `bounded()` wraps `caught` and `protected` as well as the total, so all three carry the same qualifier instead of two of them reading as exact.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
thomasguillot: Leaving the bypass in place. The dropped stylesheet import is a real latent break and an embedded variant on the shared wrapper is the right shape, but it changes a shared component's API to fix something not yet live. Better as its own change than inside a feedback pass.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
thomasguillot: Already closed further along the stack: `perPage` and the field list are folded into a memoised view handed straight to `filterSortAndPaginate`, with state kept only for what the user can change.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
thomasguillot: Holding, and taking it as a deliberate call rather than leaving it as a local default. It wants settling against the rules list rather than inside this PR.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-table.tsx
thomasguillot: Leaving this for a follow-up. Dropping the landmark is trivial; naming the table is not, since DataViews exposes no `aria-label`, `aria-labelledby` or caption of its own, so it needs either an upstream prop or a workaround inside the component. Worth doing properly rather than half.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/readonly-field.tsx
thomasguillot: Agreed on the destination. Routing it to the shared-component extraction work rather than doing it here, since the move carries the SCSS, the class renames and the `$gray-100` token question with it, and DSGNEWS-211 and DSGNEWS-212 are already pulling EmptyState and StatCard across. The `.components-input-control__container` coupling goes with it.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: The diagnosis holds. Not fixing it in this pass: the options are a forced remount or an explicit announcement, and both want more care than a feedback commit. Tracking it separately.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: `Deal ID` stays. It matches the analytics vocabulary the help text points at, and renaming the field would split it from the term publishers meet in reporting. The description heading and test step 10 are the stale half, and those are what want correcting.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: Already closed further along the stack: `hasPrice` requires actual steps on the stepped model, so the empty card is reachable for both pricing models.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Not fixed here. Making the tint paint means going back to cell-level backgrounds, since a column background sits under both the header cell and the hover row. That also removes the `<col>` coupling in the thread below, so the two are one job rather than two, and it is more than this pass should carry.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/style.scss
thomasguillot: Grouped with the paint problem above: moving to cell-level backgrounds removes this coupling rather than documenting it. If the selector survives that, it gets the comment and an upstream issue.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/conditions.tsx
thomasguillot: Fixed in ee8ae818d. The condition passes `matcher.label` straight through, so the toggle names itself once and translators see one sentence rather than two fragments.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/impact-stats.tsx
thomasguillot: Holding for a type-scale pass rather than a one-off adjustment. The assignment wants looking at across the panel, not just at this one value.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: Checked this one and kept the local declaration, because the class carries nothing to reuse.

In `@wordpress/components` 30.9.0 the help text is an emotion `StyledHelp` component: the declarations land on a generated `css-*` class, and `components-base-control__help` rides along as a marker. There is no rule for it in the package's `build-style/style.css`, and the only rules matching it in this repo are our own margin resets. Adopting it drops the 12px and inherits 13px from the admin body.

So the duplication is real, but unavoidable short of importing the component itself. Leaving this open in case you read it differently.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
thomasguillot: Already closed further along the stack: the wrapper is gone, so the class goes with it.

thomasguillot:
Addressed the feedback: five fixes here, the blocker fixed engine-side, and the rest answered on their threads.

**The "Products affected" headline.** Fixed in the engine, on `fix/per-rule-audience-total` in `Automattic/woocommerce-dynamic-pricing`, since both halves live there. `candidates()` sums two paginated totals instead of counting an already-capped merged list, and `impact_preview()` sets `count_limited` from `is_truncated()`. Detail on the thread. Nothing changes here once that lands.

**The subscriber tiles.** Already closed further along the stack: `bounded()` covers `caught` and `protected` as well as the total.

**The first pick on an empty schedule field.** Fixed in c45e6939b by seeding `currentDate`, which closes it whichever of the two mechanisms is the real one.

**Clearing the custom cohort date.** Fixed in ee8ae818d, with a test asserting where focus lands.

**The schedule fields reading as accent-coloured buttons.** Held. The read is right, and the placeholder is the half worth fixing, but it is a visual decision rather than a defect and I would rather take it deliberately.

**The consolidation asks.** The duplicated `formatCount` and the `perPage` derivation are both already closed further along the stack. The two generic controls are going to the shared-component extraction work rather than moving here, since the move carries their SCSS, class renames and the `$gray-100` token question with it. The DataViews wrapper bypass stays for now: an embedded variant on the shared wrapper is the right shape, but it changes a shared API to fix something not yet live.

One correction, on the shared schedule help. `components-base-control__help` carries no styling to reuse. In `@wordpress/components` 30.9.0 the help text is an emotion `StyledHelp` component, so the declarations land on a generated class and the marker class has no rule of its own in the package stylesheet. Adopting it would drop 12px to an inherited 13px, so the local declaration stays.

**On `catalog-impact.tsx`.** The two vocabularies are real. Leaving it out of this PR deliberately: bringing the catalog panel onto `ImpactEmpty` and the shared count formatting changes a surface this PR does not otherwise touch, and that panel is reworked further along the stack anyway. Better done there than bolted on here.

Verification: 81 suites and 729 tests pass, `tsc` sits at its existing baseline, ESLint and Stylelint are clean. Both section-header size variants were measured on a live page after the icon rules moved, and the empty card was checked at 1512.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-preview.tsx
thomasguillot: Now open as `Automattic/woocommerce-dynamic-pricing#4`.

miguelpeixe: The engine half holds: merged trunk (`b9b45fb`) matches the branch, its suite is green locally (407 tests), and both triggers are closed — sub-categories via `include_children: false`, and the `all_subscriptions` total is now exact.

One thing was left, and it answers your question about the key: its meaning is fine, the problem is direction. The merged docblock says this route's total is an upper bound ("'N+' would be wrong"), the opposite of `catalog_impact()`. This line fed `count_limited` into the same `bounded()`, so a capped preview would have read "At least 2,000" when the engine means 2,000 or fewer.

Applied in d8813ca17 rather than bouncing it back: `bounded()` now takes the bound direction per caller, this preview declares `upper` and renders "Up to N", and the catalog panel keeps "N+" by default. The audience tiles are true lower bounds, so they stay as they are. Tests pin the split in both files and the plugin's jest suite is green. The tiles rework at the stack tip will want the same direction carried over when it rebases — I can do that pass if useful. Revert if you'd rather take it another shape.

miguelpeixe (approved):

PR #799, opened 2026-08-04
fix(newsletters): keep block editor chrome on admin typography

thomasguillot:
### All Submissions:

On the MJML path the newsletter editor forces email-safe fonts across the canvas so the editor resembles the sent email. That override also reached block editor chrome, which never appears in an email: placeholders, block warnings, notices and this plugin's own canvas labels all rendered in georgia at the email's leading.

Chrome is now **excluded** from the override rather than styled back on top of it. Core's own typography then applies unchanged, so placeholder text keeps `line-height: initial`, the embed placeholder's URL field keeps `line-height: normal`, and `<code>` keeps its monospace face. `:where()` holds the exclusion at zero specificity, so every other canvas rule keeps the weight it had.

The exclusion list is a registry of canvas chrome, not only the selectors that currently lose to the override. A chrome element that happens to out-specify the override today therefore stops depending on that.

What this covers:

* Every block's placeholder, so variation pickers are included without naming them.
* Block warnings and notices.
* Three Newspack labels: conditional content, block visibility, and the ad block label.

Two of those labels were already rendering in the email font. The conditional-content badge declared the admin font but lost at one class of specificity, and the ad block label declared no `font-family` at all. The ad label's declaration lives in the always-enqueued bundle, so it now reads the same under either renderer.

Rendered email output is untouched. MJML is generated server-side from block attributes and never consulted these rules.

Surfaced while fixing the Roundup block's placeholder in `newspack-manager`; that PR is linked from this one and depends on this change for its typography to take effect.

1. Confirm the WooCommerce email renderer flag is off. This is what loads the stylesheet, so the rest of the steps depend on it.
2. Edit a newsletter and insert a block with a placeholder state, for example Image or Embed. The placeholder's label, instructions and controls render in the admin sans-serif font, not a serif.
3. Paste nothing into the Embed placeholder and compare its URL field with the same block in the standard post editor. The leading matches, rather than sitting looser.
4. Add a Conditional Content block. The yellow badge and its inline Edit and Remove controls render in the admin font.
5. Add an Ad block. The label on the dashed mock renders in the admin font.
6. Confirm paragraphs and headings in the newsletter body still render in georgia and arial.
7. Turn the WooCommerce email renderer flag on. The ad block label still renders in the admin font; everything else here is inert in that state.
8. Send or preview the newsletter and confirm the email output is unchanged.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? Tests n/a: stylesheet only, no test surface.
* [x] Have you successfully run tests with your changes locally? Verified by the manual steps above, plus `lint:scss` on both changed files.

Left for follow-ups:

* `src/editor/style.scss` forces `color: inherit` on canvas links, which beats core's styling for `ExternalLink` inside placeholders, so chrome links render near-black instead of the admin accent. Colour is a separate axis, and that file is always enqueued, so a fix would reach the WooCommerce renderer path too. Best handled together with its `text-decoration: underline` counterpart.
* The conditional-content and visibility labels declare a local `--newspack-system-font` property rather than the base-styles variable used elsewhere. Same computed value today, so this is tidy-up rather than a behaviour change.
* The Roundup loading overlay renders as a sibling of `<Placeholder>`, so the exclusion does not reach it and its font workaround is still required on the `newspack-manager` side.

### DIFF
--- a/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
+++ b/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
@@ -9,7 +9,8 @@
 //
 // The selector nesting mirrors the original location under
 // `.editor-styles-wrapper .block-editor-block-list__layout` in `style.scss`, so
-// when both stylesheets load (flag off) the result is identical to before.
+// when both stylesheets load (flag off) the result is identical to before,
+// except for editor chrome (see the exclusion list below).
 
 @use "~@wordpress/base-styles/variables" as wp-vars;
 
@@ -62,6 +63,10 @@
 		// the email, so it is excluded here rather than overridden afterwards,
 		// leaving core's own typography intact. `:where()` holds the exclusion
 		// at zero specificity, so these rules keep their original weight.
+		//
+		// This is the registry of canvas chrome, not just the selectors that
+		// currently lose to the blanket — some already out-specify it, and
+		// listing them keeps that from being load-bearing.
 		*:where(:not(
 			.components-placeholder, .components-placeholder *,
 			.block-editor-warning, .block-editor-warning *,
@@ -69,7 +74,9 @@
 			.newsletters-block__conditional-content__label,
 			.newsletters-block__conditional-content__label *,
 			.newsletters-block-visibility-label,
-			.newsletters-block-visibility-label *
+			.newsletters-block-visibility-label *,
+			.newspack-newsletters-ad-block-ad-label,
+			.newspack-newsletters-ad-block-ad-label *
 		)) {
 			line-height: 1.5;
 
--- a/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
+++ b/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
@@ -58,16 +58,18 @@
 
 		// Blanket font-family and line-height overrides — moved from
 		// `../style.scss` (Task 5). Forces georgia/arial regardless of theme;
-		// only applies when flag is off (MJML path). Block editor chrome is UI
-		// that never reaches the email, so it is excluded here rather than
-		// overridden afterwards, which leaves core's own typography intact.
-		// `:where()` keeps that exclusion at zero specificity, so these rules
-		// stay at their original weight and the `h1`–`h6` rule below still wins
-		// on source order.
+		// only applies when flag is off (MJML path). Editor chrome never reaches
+		// the email, so it is excluded here rather than overridden afterwards,
+		// leaving core's own typography intact. `:where()` holds the exclusion
+		// at zero specificity, so these rules keep their original weight.
 		*:where(:not(
 			.components-placeholder, .components-placeholder *,
 			.block-editor-warning, .block-editor-warning *,
-			.components-notice, .components-notice *
+			.components-notice, .components-notice *,
+			.newsletters-block__conditional-content__label,
+			.newsletters-block__conditional-content__label *,
+			.newsletters-block-visibility-label,
+			.newsletters-block-visibility-label *
 		)) {
 			line-height: 1.5;
 
@@ -76,14 +78,19 @@
 			}
 		}
 
-		// Chrome roots carry no font-family of their own, so they would inherit
-		// georgia from the block wrapper. Reset the root only: everything inside
-		// then follows core, which is why no line-height is set here — core's
-		// `line-height: initial` on placeholder text would be lost if it were.
+		// Core leaves the placeholder root's font-family unset, and sets no
+		// line-height on any chrome root, so both would otherwise be inherited
+		// from the block wrapper. Core's own declarations inside still win,
+		// an explicit value beating an inherited one. The Newspack labels above
+		// set their own typography and need no reset.
+		.components-placeholder {
+			font-family: wp-vars.$default-font;
+		}
+
 		.components-placeholder,
 		.block-editor-warning,
 		.components-notice {
-			font-family: wp-vars.$default-font;
+			line-height: wp-vars.$default-line-height;
 		}
 
 		h1,
--- a/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
+++ b/plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
@@ -11,6 +11,8 @@
 // `.editor-styles-wrapper .block-editor-block-list__layout` in `style.scss`, so
 // when both stylesheets load (flag off) the result is identical to before.
 
+@use "~@wordpress/base-styles/variables" as wp-vars;
+
 // Font variable declarations and aliases — moved from `../style.scss` (Task 5).
 // Under flag-on these are absent so the theme's fonts flow through; under
 // flag-off this bundle restores them, keeping the MJML editor appearance intact.
@@ -65,9 +67,13 @@
 			}
 		}
 
-		.block-editor-block-list__block .components-placeholder,
-		.block-editor-block-list__block .components-placeholder div {
-			font-family: "Noto Serif", helvetica, arial, sans-serif;
+		// Block editor chrome — placeholders and variation pickers — is UI that
+		// never reaches the email, so exempt it from the blanket override above
+		// and leave it on the admin typography.
+		.components-placeholder,
+		.components-placeholder * {
+			font-family: wp-vars.$default-font;
+			line-height: wp-vars.$default-line-height;
 		}
 
 		h1,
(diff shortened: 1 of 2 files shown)


thomasguillot:
Paired with Automattic/newspack-manager#701, which surfaced this and needs it for the Roundup placeholder's typography.

thomasguillot:
## Review summary

Reviewed the branch against `main` and verified it by building rather than by reading: both stylesheets compile with the workspace's own dart-sass, `stylelint` passes on both, and I diffed the *emitted* CSS against `main` to confirm the change is exactly what it claims. Cascade claims were checked against the core CSS shipped in `wp-includes` rather than from memory.

What came back clean:

- **Email output is untouched.** MJML is built server-side from block attributes. Neither renderer reads these editor styles, and none of the chrome involved reaches save output.
- **Specificity is unchanged.** `:where()` contributes zero, so the blanket rules stay at (0,2,0) and (0,2,1). The `h1`–`h6` rule still wins on source order exactly as before.
- **No content preview is caught by the exclusion.** Every `<Placeholder>` in this canvas is an empty or error state rendered *instead of* content, never around it. The posts-inserter renders its `BlockPreview` as a sibling, so the one place the editor deliberately shows email typography is unaffected.
- **Both flag paths agree.** The always-enqueued bundle is untouched apart from the ad label, which now matches the other chrome labels in both states.

**Release risk: low.** No blockers. This is editor-only chrome styling behind a feature flag, with no path to rendered email.

### What changed during review

The first version out-specified the blanket canvas rule by counter-styling `.components-placeholder` and every descendant. That approach cannot be made correct, because core sets different typography per element, so any single value forced onto all descendants is wrong for some of them. Chrome is now *excluded* from the blanket instead, and only the chrome roots are reset, which lets core's own rules survive: placeholder text keeps `line-height: initial`, the URL input keeps `line-height: normal`, and `<code>` keeps monospace.

Excluding chrome alone left a gap. The roots still inherited `line-height: 1.5` from the block wrapper, so line-height never actually reached admin typography. The roots now reset it to `$default-line-height`, the same value core pins on `.block-editor-warning__message` inside the canvas. `font-family` is reset only on `.components-placeholder`, since core already declares it on the warning and notice roots.

The exclusion list is documented as a registry of canvas chrome rather than a minimal list, so it no longer depends on which selectors happen to out-specify the blanket. That surfaced two more pieces of chrome rendering in georgia: the conditional-content badge, which declared the admin font but lost at one class of specificity, and the ad block label, which declared no font at all.

### Remaining findings

Both non-blocking:

- **nit** `src/editor/blocks/ad/editor.scss:14-18`: dead commented-out CSS, including a font stack on the parent element that is now the same value the live rule resolves to. It reads as a disabled alternative to the change.
- **nit** `src/editor/legacy-block-styles/style.scss:62-64`: "excluded here rather than overridden afterwards" is followed by a block that does override afterwards. They are different mechanisms, but the phrasing invites a double-take.

### Deliberately left out

- `src/editor/style.scss` forces `color: inherit` on canvas links, which beats core's styling for `ExternalLink` inside placeholders. Colour is a different axis from typography, and that file is always-enqueued, so a fix would reach the flag-on canvas while it is under active parity work. Better as a follow-up, along with its `text-decoration: underline` twin.
- The two sibling chrome labels use a locally declared `--newspack-system-font` rather than the base-styles variable. Same computed value today, so this is maintenance hygiene in two files this PR does not otherwise touch.
- The Roundup loading overlay renders as a sibling of `<Placeholder>`, so its font workaround is still required. That belongs with the newspack-manager change.

Not verified in a running editor. The steps in the description still want a manual pass in both flag states.

thomasguillot:
## Final pass

Both remaining nits are addressed in 62c1c80. Comment-only changes: I recompiled both stylesheets and the emitted CSS is byte-identical to the reviewed state, and `stylelint` still passes.

- `src/editor/blocks/ad/editor.scss`: removed the commented-out font stack and `padding` on `.newspack-newsletters-ad-block-placeholder`. The font stack resolved to the same value the live rule on the child now uses, so leaving it in place read as a disabled alternative to the change.
- `src/editor/legacy-block-styles/style.scss`: reworded the blanket-rule comment. It now reads "excluded here, leaving core's own typography intact", so it no longer appears to contradict the root reset further down, which is a separate mechanism.

Nothing else changed. The three follow-ups noted above are still open and unaddressed here: the canvas link colour in `src/editor/style.scss`, the `--newspack-system-font` spelling in the two sibling label stylesheets, and the Roundup loading overlay on the newspack-manager side.

thomasguillot (commented):
Reviewed the branch against a local site with the WooCommerce renderer flag in both states, measuring computed styles in the editor canvas.

Flag off: placeholder roots render the admin sans at 1.4, with core's own values on their labels, instructions and buttons. The Embed URL field's leading is 13px, matching the same block in the standard post editor. Body copy stays georgia at 1.5, headings arial, inline `code` monospace. Edit as HTML renders Menlo at 1.5. Block warnings and the drop zone render the admin font and leading. `.components-notice` takes core's own 20px, with the reset deferring to it rather than overriding it.

Flag on: `legacyBlockStyles.css` is absent from the canvas and the ad block label still renders the admin font.

The compiled blanket matches paragraphs and headings, and none of the chrome roots. That is the intended split.

Two things changed after the first pass. The drop zone and the Edit as HTML textarea were still picking up the email font, so both are now in the exclusion registry. The blanket is also written as `*:not(:where(...))` rather than `*:where(:not(...))`: identical specificity, identical behaviour while every entry is valid, but an unparseable entry now under-excludes a single root instead of silently dropping the whole override.

Email output is untouched. The change is stylesheet-only and MJML renders server-side from block attributes.

plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
```diff
@@ -9,7 +9,34 @@
...
+
+$chrome-root-needing-font-reset: ".components-placeholder";
+
+$chrome-roots-needing-line-height-reset:
+	$chrome-root-needing-font-reset,
+	".block-editor-warning",
+	".components-notice",
+	".components-drop-zone";
+
+$chrome-roots-self-styled:
```
thomasguillot, line 28: **Suggestion:** The chrome registry cannot be joined from outside this file.

`$chrome-roots-self-styled` is a private Sass list, so chrome that another plugin renders into this canvas can only opt out of the override by editing this bundle. `newspack-manager`'s `src/roundup/editor.scss:223-228` re-declares the admin stack on its own overlay for that reason. A marker class added to the list (e.g. `.newspack-newsletters-canvas-chrome`) would turn the exclusion into a contract an external block opts into, and the two `:where()` resets below could carry the same marker so a consumer's inherited `.wp-block` values get cleared too. This is about the registry's shape, not the deferred Roundup workaround, and nothing on the newspack-manager side needs to change in this PR.

thomasguillot, line 28: **Nit:** `$chrome-roots-self-styled` relies on declarations in three other files with no signpost.

Membership of this list is only safe while every entry declares both `font-family` and `line-height` itself: these roots are excluded from the blanket but get no reset at :390–:395, so missing either lets the root inherit georgia/1.5 from the canvas ancestor the blanket still matches. All four hold today, but three rely on declarations that live elsewhere and carry nothing pointing back here: `blocks/conditional-content/style.scss:48,50`, `blocks/visibility-attribute/style.scss:48,50`, and `blocks/ad/editor.scss:37,39` (core owns the fourth, at `content.css`'s `.block-editor-block-list__block .block-editor-block-list__block-html-textarea`). Dropping `line-height: 1.6` from one of those label files as unused tidy-up would regress the badge's leading silently, since nothing would fail. Consider one line above the list naming the contract an entry has to satisfy: the variable name carries most of it, so this is a judgement call rather than a gap, and it's the only place the split between the two lists isn't recoverable from the code.

plugins/newspack-newsletters/src/editor/legacy-block-styles/style.scss
```diff
@@ -9,7 +9,34 @@
...
+	".block-editor-block-list__block-html-textarea",
```
thomasguillot, line 29: **Nit:** The Edit-as-HTML textarea is a visible change the PR body does not list.

Under flag-off, the textarea previously lost `font-family` to the blanket rule at `(0,2,1)` and rendered in georgia; excluding it lets core's `$editor-html-font` through instead (`block-list/content.scss:380`). That's the right outcome and matches the verification notes, but "What this covers" names only placeholders, warnings, notices and the three labels, so a tester following the steps has no prompt to check it. Worth a line in the body.

thomasguillot (commented):
Self-review pass. No blockers; one suggestion and two nits, all on the chrome registry in `legacy-block-styles/style.scss`.

Release risk is low: editor-only stylesheet enqueued behind the WooCommerce email renderer flag, no path to rendered email output, no contract files touched, blanket specificity unchanged. Runtime verification was not repeated in this pass; the measured pass recorded on 22 Aug stands.

PR #194, opened 2026-06-02
fix(popups): portal overlay lightboxes to wp_footer to escape stacking contexts

adekbadek:
### Changes proposed in this Pull Request:

Monorepo hotfix superseding Automattic/newspack-popups#1561 (that legacy repo no longer cuts releases, and the daily sync only mirrors `trunk`, so a legacy merge would never ship as a hotfix). Off `release` so it lands as an out-of-schedule `newspack-popups` patch; `post-release.sh` back-merges to `main`/`alpha` automatically.

Overlay prompts are emitted inside `.entry-content` via `the_content`. Any ancestor that establishes a stacking context (transform, position+z-index, isolation, will-change, contain – common with custom theme CSS or ad-management plugins) traps the lightbox's `z-index: 99999` local to that context, so siblings can paint over the close button. Observed in the wild: scroll-triggered Center Overlay prompts partially obscured by a sticky right-rail.

- Queue overlay lightboxes on `the_content` instead of inlining them. Flush at `wp_footer` priority 9 (before `wp_print_footer_scripts` at 20) so they render as direct children of `<body>` and any assets enqueued during overlay-content render still make the footer. Inline / above-header / archive placements unchanged.
- Scroll-triggered overlays keep their `#page-position-marker_*` inline at the content position; only the lightbox itself moves. The marker still drives `IntersectionObserver` reveal as before.

Closes #1561 (legacy).

### How to test the changes in this Pull Request:

1. Activate Newspack Campaigns; create a Center Overlay prompt with **scroll** trigger at 0%.
2. Add a temporary mu-plugin that fakes the stacking trap:

   ```php
   <?php
   add_action( 'wp_head', function() { ?>
   <style>
     .entry-content { transform: translateZ(0); position: relative; z-index: 1; }
     .demo-rail { position: fixed; top: 0; right: 0; width: 380px; height: 100vh;
                  background: rgba(220,38,38,.85); color: #fff; padding: 24px;
                  z-index: 2; pointer-events: none; font: 700 24px system-ui; }
   </style>
   <?php } );
   add_action( 'wp_body_open', function() {
     echo '<aside class="demo-rail">Demo rail z-index: 2</aside>';
   } );
   ```

3. Visit any post. Without the fix the rail paints over the prompt's right edge (close X hidden). With the fix the lightbox is a direct `<body>` child at z=99999 and renders above the rail.
[...]

PR #801, opened 2026-08-04
fix(subscriptions): support subscription-plan products in tiers

chickenn00dle:
> [!IMPORTANT]
> **Retargeted to `main`.** This started as a hotfix against `release`, but grew well past patch scope once review found that the code downstream of the original fix was still written for the legacy product model. The narrow, genuinely patch-sized piece was split out into #819. This PR now takes the normal route through `main`.

## Why

WooCommerce Subscriptions 9.0 folded All Products for Subscriptions into core. A recurring product no longer has to use the dedicated `subscription` / `variable-subscription` product types — the product editor now creates an ordinary `simple` or `variable` product with **subscription plans** (APFS schemes) attached.

`Subscriptions_Tiers` only understood the legacy types, so a variable subscription built the new way produced **no tiers at all**: the modal rendered its header above an empty body.

Fixing the recognition turned out to be the easy half. Everything downstream of that gate — frequency, price, what the form posts — was written when one product meant one frequency, and none of it had been taught otherwise. Review (thanks @wil-gerken) found two blockers, and building the fix surfaced three more.

## The model

**Frequency = plan, tier = variation.** The form already renders a frequency segmented control with tier cards beneath it, which is exactly the shape APFS implies: under the legacy model variations *are* the frequency axis, but under plans the frequency is the plan and variations are orthogonal to it. A product with 3 variations and 2 plans is 6 purchase options; the old code emitted 3.

So a plan-based product now renders as a Monthly/Annual toggle over its price tiers — what the legacy model produced, and what publishers expect.

## What changed

**Tier composition.** `get_tiers_by_frequency()` emits one frequency bucket per plan, each holding the variations cloned and stamped with that plan. Because APFS filters `woocommerce_product_variation_get_price` off the active plan, each instance prices itself — there is no pricing code here at all. Two plans sharing a period and interval get separate buckets rather than merging, so a configured plan can never silently vanish.

**Frequency and price come from the plan.** `get_frequency()` and a new `get_frequency_parts()` read the stamped plan instead of `_subscription_period`, which APFS never persists. Where that meta was absent the card rendered the literal string `$10.00 every Array`; where it held the hardcoded `'month'` that `WC_Subscriptions_Admin::set_variation_meta_defaults_on_bulk_add()` writes onto every generated variation, an annual plan rendered as `/ month`.

**The form posts the chosen plan.** The frequency control renders as radios named `convert_to_sub_<parent_id>`, which is where APFS reads the plan from. Without it, `get_default_subscription_scheme()` returns `false` and the reader is **charged once with no subscription created**. Single-plan products, which never render the control, emit a hidden input instead.

**`get_current_tier()` matches on the plan too**, not the product alone — otherwise an annual subscriber resolved to the monthly bucket, giving the wrong "Current" badge, the wrong preselection, and a re-purchase guard pointed at the wrong plan.

**Grouped products with plan-based children yield no tiers**, restoring `release` behaviour. A grouped parent cannot carry the per-child plan key APFS needs, so anything we rendered there would take a one-time payment.

Also: the product-model predicates moved onto `WooCommerce_Subscriptions` rather than living on `Subscriptions_Tiers` (see [the thread with @leogermani](https://github.com/Automattic/newspack-workspace/pull/801#issuecomment-5194227263) for why none of the existing WCS helpers could be reused), the pre-9.0 guard gained the test it never had, and a deleted grouped child no longer fatals.

## Testing

- **2724 tests, 7806 assertions, 0 failures** on the full suite after rebasing onto `main`; 246 in the `WooCommerce_Subscriptions_Integration` group. PHPCS clean on all six changed PHP files.
- **Live, on a Subscriptions 9.0.1 site**, across four fixtures — two plans with legacy meta, two plans without, a single-plan product, and a legacy control:

```
#194  two plans          month_1  subscription=true  billing=month/1  PASS
                         year_1   subscription=true  billing=year/1   PASS
#205  no legacy meta     month_1/year_1                               PASS
#216  single plan        month_1  subscription=true  billing=month/1  PASS
#209  legacy control     month_1  subscription=true  billing=month/1  PASS
```

The legacy control is the canary: it renders identically to `release` and emits no `convert_to_sub` anywhere.

Plan-specific pricing was verified separately, since the first fixture could not show it — both its plans inherited the store price with no discount, so every check passed while blind to whether per-plan pricing worked at all. With a 20% annual discount the same variation prices itself differently per plan, in the card **and** in the cart line item.

## Known limitations

- Two plans sharing a period and interval render two tabs both labelled e.g. "Monthly". Both stay reachable, which is the point; distinguishing them needs a plan name the frequency key does not carry.
- `render_product_card()`'s price call site is not exercised by the suite — `wcs_price_string()` has no stub and the code is `function_exists()`-guarded. Verified live instead. A stub would close it cheaply later.
- `get_frequency_parts()` duplicates the plan-walk in `get_frequency()`; the two could drift.
- **APFS discounts only reduce.** The familiar "$10/month or $100/year" cannot be expressed as plans on a $10 base — the base price is what is charged per plan period, and plans differ by cadence plus an optional discount. A publisher wanting annual at roughly 10× monthly must set the base to the annual figure and discount the monthly plan, or use an override price and lose per-tier pricing. That is WooCommerce's model rather than anything introduced here, but it shapes how useful plan-based products are for selling tiers.

## Follow-up, not this PR

`get_tier_eligible_products()` still defaults to the legacy product types, so plan-based products never appear in the Audience wizard's dropdown. Only the wizard-configured upgrade route is affected — the on-page modal and `?tiers-modal=<id>` already reach these products.

Worth folding two existing helpers into that work rather than leaving three parallel implementations: `Audience_Subscription_Products::group_has_subscription_children()` duplicates the grouped-child logic in `get_tier_eligible_products()`, and `Product_Network_Ids::is_taggable_product()` in newspack-network is the same predicate under a different name. The legacy type check is currently written inline 24 times across 10 files in three plugins.

Refs NPPM-3053.

### DIFF
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -795,10 +795,16 @@ class Subscriptions_Tiers {
 	/**
 	 * Render frequency form control.
 	 *
-	 * Up until 3 frequencies, we render buttons.
-	 * After that, we render a select control.
+	 * Up until 3 frequencies, we render buttons (or plan-posting radios, see
+	 * below). After that, we render a select control - unreachable for
+	 * plan-based products, since the plan-radio branch below runs first
+	 * regardless of $frequencies count when $plan_keys/$parent_id are set.
 	 *
-	 * @param array  $frequencies       Frequencies.
+	 * @param array  $frequencies       Frequencies. Expected to be exactly the
+	 *                                  keys present in $plan_keys when the
+	 *                                  latter is non-empty, so the caller's
+	 *                                  panel loop and this tab loop stay
+	 *                                  positionally in sync.
 	 * @param string $current_frequency Current frequency.
 	 * @param bool   $is_form_control   Whether to treat it as a form input.
 	 * @param array  $plan_keys         Optional frequency key => scheme key map. When
@@ -875,18 +881,40 @@ class Subscriptions_Tiers {
 			return;
 		}
 
+		// The plan-radio control (and its hidden-input fallback below) only
+		// make sense when $product is the actual parent WooCommerce will add
+		// to the cart: a variable or simple subscription product. For a
+		// grouped product each tier can come from a different child product
+		// with its own parent ID, and for the "all products" view ($product
+		// is null) there is no common parent at all - a single
+		// convert_to_sub_<id> input can't carry the right key for every
+		// tier in either case, so the plan path is skipped rather than
+		// emitting a key nothing reads (or, for the null case, a
+		// convert_to_sub_0 that reads as a real but wrong parent ID).
+		$parent_id = ( $product && ! $product->is_type( 'grouped' ) ) ? $product->get_id() : 0;
+
+		$plan_keys = [];
+		if ( $parent_id ) {
+			foreach ( $tiers as $frequency => $tier_products ) {
+				$plan_key = empty( $tier_products ) ? '' : WooCommerce_Subscriptions::get_active_plan_key( $tier_products[0] );
+				if ( $plan_key ) {
+					$plan_keys[ $frequency ] = $plan_key;
+				}
+			}
+			// Keep the tab headers and the panels positionally in sync: a
+			// frequency without a resolvable plan key would otherwise still
+			// get a panel below but no tab here, shifting every subsequent
+			// tab onto the wrong panel (setupTabController() pairs
+			// tab_headers[i] with tab_contents[i] by index).
+			if ( ! empty( $plan_keys ) ) {
+				$tiers = array_intersect_key( $tiers, $plan_keys );
+			}
+		}
+
 		$is_single_tier = self::is_single_tier( $tiers );
 		$is_nyp         = $is_single_tier && self::is_nyp( $tiers ); // Only treat as NYP form if there's only 1 tier.
 
 		$frequencies = array_keys( $tiers );
-
-		$plan_keys = [];
-		foreach ( $tiers as $frequency => $tier_products ) {
-			$plan_key = empty( $tier_products ) ? '' : WooCommerce_Subscriptions::get_active_plan_key( $tier_products[0] );
-			if ( $plan_key ) {
-				$plan_keys[ $frequency ] = $plan_key;
-			}
-		}
 
 		$current_frequency = null;
 		$current_product   = null;
@@ -896,6 +924,12 @@ class Subscriptions_Tiers {
 		}
 
 		if ( ! $switch_data ) {
+			$current_frequency = $frequencies[0];
+		} elseif ( ! $current_frequency ) {
+			// A subscriber on a retired plan, or one filtered out by the
+			// ownership check, has no matching tier. Fall back to the first
+			// frequency so a radio is checked and the plan gets posted,
+			// rather than leaving nothing selected (and posting no plan).
 			$current_frequency = $frequencies[0];
 		}
 
@@ -934,13 +968,22 @@ class Subscriptions_Tiers {
 		}
 
 		$should_render_tabs = ! $is_single_tier || $is_nyp;
+
+		// Whether render_frequency_control() below will actually post the
+		// plan. When it won't - a single frequency, or the flat (no-tabs)
+		// product-card layout - the hidden input further down carries the
+		// plan for $current_frequency instead. Without this, a product with
+		// exactly one plan (the ordinary case: one monthly plan plus
+		// price-tier variations) posts no plan at all and the reader is
+		// charged once instead of subscribing.
+		$frequency_control_rendered = $should_render_tabs && count( $frequencies ) > 1;
 		?>
 		<form class="newspack__subscription-tiers__form <?php echo esc_attr( $is_nyp ? 'nyp' : '' ); ?>" target="newspack_modal_checkout_iframe" data-title="<?php echo esc_attr( $title ); ?>" data-product-id="<?php echo esc_attr( $product ? $product->get_id() : '' ); ?>">
 			<?php if ( $should_render_tabs ) : ?>
 				<div class="newspack-ui__segmented-control">
 					<?php
-					if ( count( $frequencies ) > 1 ) {
-						self::render_frequency_control( $frequencies, $current_frequency, $is_nyp, $plan_keys, $product ? $product->get_id() : 0 );
+					if ( $frequency_control_rendered ) {
+						self::render_frequency_control( $frequencies, $current_frequency, $is_nyp, $plan_keys, $parent_id );
 					}
 					?>
 					<div class="newspack-ui__segmented-control__content">
@@ -971,6 +1014,9 @@ class Subscriptions_Tiers {
 			?>
 			<input type="hidden" name="newspack_checkout" value="1">
 			<input type="hidden" name="modal_checkout" value="1">
+			<?php if ( ! $frequency_control_rendered && ! empty( $plan_keys[ $current_frequency ] ) ) : ?>
+				<input type="hidden" name="convert_to_sub_<?php echo absint( $parent_id ); ?>" value="<?php echo esc_attr( $plan_keys[ $current_frequency ] ); ?>">
+			<?php endif; ?>
 			<?php if ( ! empty( $switch_data ) ) : ?>
 				<input type="hidden" name="switch-subscription" value="<?php echo esc_attr( $switch_data['subscription']->get_id() ); ?>">
 				<input type="hidden" name="item" value="<?php echo absint( $switch_data['item_id'] ); ?>">
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -749,8 +749,14 @@ class Subscriptions_Tiers {
 		$frequency = $parts['period'];
 		$interval  = $parts['interval'];
 
-		if ( $switch_subscription ) {
-			$base_product   = wc_get_product( $switch_subscription['item']['product_id'] );
+		$base_product = $switch_subscription ? wc_get_product( $switch_subscription['item']['product_id'] ) : null;
+
+		// A deleted product leaves behind an ID that no longer resolves, and
+		// wc_get_product() returns false for it. There's no billing period to
+		// convert from in that case, so keep the target product's own price
+		// rather than feeding `false` into the plan accessors - the same way
+		// the tier composition skips a grouped child that no longer resolves.
+		if ( $base_product instanceof \WC_Product ) {
 			$base_parts     = self::get_frequency_parts( $base_product );
 			$base_frequency = $base_parts['period'];
 			$base_interval  = $base_parts['interval'];
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -313,9 +313,10 @@ class Subscriptions_Tiers {
 			}
 		}
 
+		$interval = (int) $product->get_meta( '_subscription_period_interval' );
 		return [
 			'period'   => (string) $product->get_meta( '_subscription_period' ),
-			'interval' => (int) $product->get_meta( '_subscription_period_interval' ),
+			'interval' => $interval > 0 ? $interval : 1,
 		];
 	}
 
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
@@ -260,6 +260,21 @@ class Subscriptions_Tiers {
 	 * @return string Frequency.
 	 */
 	public static function get_frequency( $product ) {
+		// Under the subscription-plan model the recurrence lives on the plan, not
+		// in post meta - APFS only ever applies it as in-memory runtime meta, so
+		// the legacy keys are either absent or a hardcoded default unrelated to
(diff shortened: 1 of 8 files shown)


leogermani:
Let's check if we already have those helpers somewhere else and see if we need to update them instead of creating new ones

chickenn00dle:
Good call — checked both sides.

**Ours:** nothing to update, unfortunately. `grep -ri "wcsatt\|WCS_ATT"` across `plugins/` and `repos/` comes back empty outside this PR — this is the first Newspack code that knows subscription plans exist. What we *do* have is ~25 inline copies of `is_type( [ 'subscription', 'variable-subscription' ] )` — content gate, the Audience wizard, modal checkout, newspack-network — none of them wrapped in a helper, and all with the same latent gap.

**Subscriptions' own:** four candidates, none usable as-is (checked against the 9.1.0-rc1 source):

| Helper | What it does | Usable? |
| --- | --- | --- |
| `WC_Subscriptions_Product::is_subscription()` | legacy types, then `apply_filters( 'woocommerce_is_subscription' )` | No |
| `WC_Subscriptions_Product::is_variable_subscription()` | `variable-subscription` only | No |
| `WCS_ATT_Product::is_subscription_product_type()` | legacy types only | No |
| `WCS_ATT_Product_Schemes::has_subscription_schemes()` | the right primitive, but ignores `_wcsatt_disabled` | Partly — we call it |

The near-miss is `is_subscription()`. APFS *does* hook it (`class-wcs-att-product.php:70` → `filter_is_subscription`), but that resolves to `has_active_subscription_scheme()` — a scheme selected on the product object at runtime, i.e. cart state. A product read from the catalog has none, which is exactly why the repro shows `is_subscription = false` for #194.

`WCSG_Product::product_has_subscription_plans()` is line-for-line what we wrote, but it lives in the gifting module, and `WC_Subscriptions_Plugin::init_gifting()` bails out entirely when the standalone gifting plugin is active — in which case `WCSG_Product` is *that* plugin's class and may not have the method. Didn't seem worth trading a five-line wrapper for that dependency.

So: kept the helpers, but moved them off `Subscriptions_Tiers` onto `WooCommerce_Subscriptions` (9e6caaf), where `is_active()` / `is_enabled()` / `get_frequency_label()` already live. Pure move, no behaviour change — and it gives those ~25 sites something to adopt on `main` later instead of each growing its own copy. The reasoning is in the docblocks too, so the next person doesn't have to re-derive it.

Tests moved across with it: 159 tests / 368 assertions green for the group, PHPCS clean, and re-verified against the live 9.1.0-rc1 env (product #194 → `is_sub=true`, `is_var_sub=true`, tiers render).

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -324,7 +324,7 @@ public static function get_tiers_by_frequency( $product = null, $sort_by_price =
 			}
 
 			// Extract the variations if it's a variable subscription product.
-			if ( $product->is_type( 'variable-subscription' ) ) {
+			if ( WooCommerce_Subscriptions::is_variable_subscription_product( $product ) ) {
```
wil-gerken, line 397: **Blocker:** Tier frequency does not come from the configured plan — The variations built here flow into `get_frequency()` (line 262), which reads `_subscription_period` and `_subscription_period_interval` post meta. APFS never persists those keys: the period lives inside the `_wcsatt_schemes` array as `WCS_ATT_Scheme` data and is only ever copied onto the product as in-memory runtime meta, which `delete_runtime_meta` on `woocommerce_before_product_object_save` keeps from being written. The objects rebuilt here as `new \WC_Product_Variation( … )` carry none of it.

That leaves two shapes, both reproduced on a 9.0.1 site:

1. **No legacy meta present.** `get_frequency()` falls back to `once`, tiers group under `once_1`, and `render_product_card()` passes an empty period into `wcs_price_string()`. `wcs_get_subscription_period_strings( '', '' )` returns the whole array, so the card renders `$10.00 every Array`, with a PHP 8 array-to-string warning.

2. **Legacy meta present but unrelated to the plan.** `set_variation_meta_defaults_on_bulk_add()` hooks `product_variation_linked` and writes a hardcoded `_subscription_period = 'month'` for any variable product whose variations are generated in the editor. A product configured with an **annual** plan then rendered `$50.00 / month` for all three tiers.

Shape 2 is likely why the manual test read as correct: with a monthly plan, the hardcoded default and the real plan agree by coincidence.

Worth noting the model mismatch underneath this too. In the legacy model variations *are* the frequency dimension; in the new model the frequency is the scheme and variations are orthogonal, so a product with 2 variations and 2 plans has 4 purchase options and this code produces 2.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -324,7 +324,7 @@ public static function get_tiers_by_frequency( $product = null, $sort_by_price =
...
 				$variations = $product->get_available_variations();
 				foreach ( $variations as $variation ) {
 					$selected_products[] = new \WC_Product_Variation( $variation['variation_id'] );
```
wil-gerken: **Blocker:** A tier purchase is charged once instead of creating a subscription — Now that plan-based products reach this branch, the variations they produce become purchasable through the tiers form — and the form carries no subscription plan with them.

The form posts `product_id`, `newspack_checkout`, `modal_checkout`, and the switch fields. Under the legacy types that was enough, because the product type itself carries the recurrence. Under the new model APFS resolves the plan from a posted `convert_to_sub_<product_id>` key (`WCS_ATT_Cart::add_cart_item_data()`, `class-wcs-att-cart.php:192`). Nothing in the plugin posts that key, so it falls back to `get_default_subscription_scheme()`, which returns `false` unless the product has a forced scheme.

Reproduced on a 9.0.1 site, adding a tier variation from a plan-based `variable` product to the cart:

```
active_subscription_scheme false
WCS_ATT_Product::is_subscription false
cart_contains_subscription() FALSE — one-time purchase
has_forced_subscription_scheme( parent ) false
get_default_subscription_scheme( parent ) false
```

So a reader picks a tier in the subscription modal, checks out, and is charged once with no subscription created. On a product that does carry a forced scheme the fallback picks `current( $schemes )`, so the first plan applies regardless of which tier was selected.

This is the one I'd most want resolved before the patch ships, since it affects money rather than presentation.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
```diff
@@ -315,7 +315,7 @@ public static function get_tiers_by_frequency( $product = null, $sort_by_price =
 				$product = wc_get_product( $product );
 			}
 
-			if ( ! in_array( $product->get_type(), [ 'subscription', 'variable-subscription' ], true ) ) {
+			if ( ! WooCommerce_Subscriptions::is_subscription_product( $product ) ) {
```
wil-gerken, line 375: **Suggestion:** Only the wizard-configured upgrade route is affected by the eligibility gap — This filter now admits plan-based products, but `get_tier_eligible_products()` (line 211) still defaults to `[ 'grouped', 'variable-subscription' ]`, and its grouped-child test at line 242 still checks the two legacy types.

I chased down how much that actually costs, since it decides whether NPPM-3053 is closed. Checking each route to the tiers modal on a running site:

| Route | Plan-based product |
| --- | --- |
| On-page modal checkout (`Modal_Checkout::render_variation_selection`) | works — its gate accepts `variable` |
| `?tiers-modal=<id>` | works — no type gate |
| Upgrade flow via the wizard-set primary product | blocked |

So the two direct routes are fine and the fix lands for them. Only the upgrade route is affected, and the block is the dropdown: `eligible_products` in the Audience wizard is fed from `get_tier_eligible_products()`, so a plan-based product never appears. `set_primary_subscription_tier_product()` itself accepts one with no server-side validation, and once set directly the upgrade surface renders (2110 bytes, tiers resolve).

That makes this a clean follow-up rather than something gating this PR: widening the default types and line 242 to use `is_subscription_product()` closes it, with nothing else to change. Worth doing separately, since it changes which products appear in publisher-facing lists. Note the frequency issue above applies to this route too once it's reachable.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -701,6 +701,72 @@ public static function is_enabled() {
...
+	 * entirely when the standalone gifting plugin is active — in which case
+	 * `WCSG_Product` is that plugin's class and may not have the method. So we
+	 * mirror it here, opt-out meta included, rather than depend on it.
+	 *
+	 * @param \WC_Product $product Product object.
+	 *
+	 * @return bool Whether the product has subscription plans.
+	 */
+	public static function has_subscription_plans( $product ) {
+		if ( ! class_exists( 'WCS_ATT_Product_Schemes' ) || ! method_exists( 'WCS_ATT_Product_Schemes', 'has_subscription_schemes' ) ) {
```
wil-gerken, line 736: **Suggestion:** The pre-9.0 guard has no test coverage — `wc-mocks.php` defines the `WCS_ATT_Product_Schemes` stub unconditionally, so this early return never executes under test. That's the branch every site on Subscriptions 8.x without standalone APFS takes, and it's the branch carrying the claim that pre-9.0 sites are unaffected. A case that unsets the stub, or a `@runInSeparateProcess` test, would cover it.

Two notes on the stub itself while it's in view. It omits the `WCS_ATT_Product::supports_feature()` gate that opens the real `get_subscription_schemes()`, so it returns true for a `grouped` product where the real API returns false; the divergence is permissive-mock against strict-real, so no current assertion is masking a break. It also keys strictly on the passed product's ID, where the real implementation resolves a variation's schemes from its parent. Arity is fine — the real signature takes `( $product, $context = 'any' )` and this call site passes one argument.

Worth flagging that standalone APFS also defines `WCS_ATT_Product_Schemes`, so this change is live on 8.x + APFS sites too, not only 9.0+.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-subscriptions-tiers.php
wil-gerken, line 375: **Nit:** Optional hardening for deleted grouped children — If `wc_get_product()` returns `false` for a deleted grouped child, `is_subscription_product( false )` fatals on `->get_type()`. The old code had the same exposure on the same line, so this isn't a regression — but the grouped path is traversed a little more often now, and `if ( ! $product instanceof \WC_Product ) { continue; }` is a one-liner while the file is open. Entirely optional.

wil-gerken (changes requested):
**Release risk: HIGH** — two blockers, both reproduced on a 9.0.1 site, on a change targeting `release`.

Thanks, Raz. This does exactly as it says. On `origin/release` a plan-based product produced nothing at all (0 tiers, `render_form()` emitting 0 bytes); with this change it renders all three tiers. And I couldn't get the legacy path to budge no matter which shape I threw at it: a `variable-subscription` control still grouped under `month_1` with the same three products and the same output as before. That part isn't luck, either — `has_subscription_schemes()` gates on `WCS_ATT_Product::supports_feature()`, whose supported-type list contains no `subscription`, `variable-subscription`, or `grouped`, so the new predicate is structurally unable to fire on a legacy product. The `_wcsatt_disabled` mirror matches `WCSG_Product::product_has_subscription_plans()` exactly, and the stub's arity matches the real signature at the one call site. The follow-up that moves the helpers onto `WooCommerce_Subscriptions` (9e6caaf) checks out too: I diffed the extracted method bodies and the relocation is byte-identical, and the research in the thread holds — `init_gifting()` does bail when standalone gifting is active, and I count 26 legacy-type call sites that can adopt these helpers from their new home later.

A few things surfaced once I started exercising the new path. The changed lines themselves read correctly; what they do is open a route that was closed before, and some of the code downstream of it was written for the legacy model and hasn't been taught the new one. Two of those I think are worth settling before this goes out in a patch.

**A tier purchase is charged once instead of creating a subscription.** The form posts only `product_id`, and APFS resolves a cart item's plan from a posted `convert_to_sub_<product_id>`. With nothing posted it falls back to `get_default_subscription_scheme()`, which returns `false` unless the product is subscription-only. Adding a tier variation to the cart gave `active_subscription_scheme = false` and `cart_contains_subscription() = FALSE`.

**The displayed frequency doesn't come from the configured plan.** `get_frequency()` reads `_subscription_period` meta, which APFS never persists — the period lives in `_wcsatt_schemes` and is only ever applied as in-memory runtime meta. Where variations carry no legacy meta the card renders `$10.00 every Array`; where they carry the hardcoded `'month'` that `set_variation_meta_defaults_on_bulk_add()` writes on `product_variation_linked`, an annual plan renders as `$50.00 / month`. That second shape is probably why the manual test looked right: with a monthly plan, the hardcoded default and the real plan agree by coincidence.

On the deferred scope, one correction and one reassurance. The first out-of-scope note is accurate. The second describes the child filter in `get_tier_eligible_products()` (line 242) rather than the one this diff changed, so a grouped product with plan-based children does now yield tiers. And the eligibility gap matters less than it might look: the on-page modal and `?tiers-modal=<id>` both reach plan-based products already, so only the wizard-configured upgrade route is affected, and that's the dropdown alone — `set_primary_subscription_tier_product()` accepts such a product with no validation today.

Handling the scheme in the form and the frequency lookup is the complete fix. If that's too much for a patch, narrowing the detection to products the rest of the pipeline can render correctly would be a smaller one.

chickenn00dle:
Thanks for the review @wil-gerken! Turns out this one will be more complex of a change than I anticipated. While digging further into the nuances of the new subscription plan structure, I discovered WCS added a new setting (which is off by default) that will re-enable legacy variable subscriptions. 

So I've decided to pivot and rebase these changes onto main so it goes through our normal QA process and get a hotfix PR up to enable these new options for our sites instead so publishers can continue to create legacy variable subscriptions in the admin.

That new hotfix PR is here: https://github.com/Automattic/newspack-workspace/pull/819

Sorry for the confusion! 🙇

PR #802, opened 2026-08-04
fix(reader-activation): keep browsers from autofilling the auth honeypot

dkoo:
### All Submissions:

Readers reported signing in, seeing **"Success! You're signed in."**, and then being asked to sign in again the moment they opened My Account. It reproduced for some people and not others, on the same site, which is what made it hard to pin down.

The cause is the anti-bot honeypot on the reader auth form. It's a decoy input named `email` (the real address field is `npe`), and a non-empty value makes `process_auth_form()` return `{ authenticated: 1 }` and bail — deliberately, so a bot thinks it succeeded — without ever calling `wp_set_auth_cookie()`. The frontend can't tell that apart from a real sign-in: `auth-form.js` sees a 200 with no `data.action`, flips the client store to authenticated (so the nav link becomes "My Account"), and renders the success screen. No session exists, so the next page load shows the login form.

Something filled that decoy on the reader's behalf. **What, exactly, is not yet identified** — see "What's confirmed and what isn't" below. The field was hidden only by the stylesheet rule `.newspack-reader-auth .nphp`, which uses the shared `visuallyHidden` mixin — a **1×1px absolutely-positioned box**. It is invisible to a person but still laid out, and a form filler that keys on layout will treat it as a real field. `autocomplete="off"` is advisory and routinely ignored on address fields. Whichever filler is responsible, it clearly doesn't affect every reader, which is why this looked unreproducible.

The fix closes the field to automated filling in the two ways available:

- `data-1p-ignore`, `data-lpignore="true"`, `data-form-type="other"` — the opt-outs 1Password, LastPass and Dashlane publish. On current evidence a password-manager extension is the most likely filler, which makes these the load-bearing part of this change.
- inline `style="display:none;"` — a field with no layout box is skipped by fillers that check visibility. Inline rather than in the stylesheet, because a rule can be stripped or deferred by optimization plugins, which would put the decoy back on the page in a reader's view.
- the `.nphp` stylesheet rules gain the same `display:none`, so they remain a real fallback for a policy that strips the style attribute rather than a merely visual one.
- `name` and `type` are unchanged, so the decoy still looks like what bots target.

`newspack-newsletters`' Subscribe block carries an independent, byte-identical copy of the same field, so it gets the same change. Its failure is quieter — the reader is thanked for a subscription that was never recorded, leaving no trace on either side — and fixing one copy and not the other is how this comes back. Both copies now carry a comment pointing at the other.

**Known trade-off:** `display:none` and the opt-out attributes make the decoy distinguishable from the real field by markup alone, so a bot that adapts can skip it. The decoy was already trivially identifiable in raw HTML (`class="nphp" tabindex="-1" aria-hidden="true"` on an email input), so this adds to an already-loud fingerprint rather than creating one.

This PR changes the decoy's **markup**, and adds a **diagnostic log** (below). Its submit-side behaviour is deliberately unchanged: a non-empty decoy still returns the fake success.

A server-side companion was written and reviewed — reading a decoy that matches the submitted address as autofill rather than as a bot — and **the team declined it** (#814, closed). It made the honeypot too permissive: a bot that echoes one address into every field it recognises would have passed, and on sites without reCAPTCHA this decoy is the only gate in front of a handler that has no per-IP rate limit. Recording that here so the idea isn't re-proposed without the objection attached.

### Also: the trap now reports itself

Separate commit, and the reason it's here rather than in a follow-up. The trap returns a fake success **and writes nothing**, which is exactly why nobody could tell whether it fires for real readers or only for bots — it has been invisible since it was added. So this adds one `newspack_log` entry when it trips.

It records the **user agent**, which is what identifies whatever filled a field no human can see, plus the shape of the submission — whether the real field was also populated, and whether the decoy matched it. That last flag is the discriminator: a browser filling both from one profile entry looks different from a bot filling the decoy alone. **The addresses themselves are not logged.**

It's throttled to one entry per five minutes. `newspack_log` reaches central logging, and on a site without reCAPTCHA this trap fires on ordinary bot traffic — the question is whether it fires and what fills it, which a sample answers, so there's no case for pushing one site's spam volume into a shared store.

Addresses NPPM-3071. Deliberately not `Closes` — the mechanism is established but the filler isn't identified, so the ticket shouldn't auto-close on merge. See below.

1. On a site with reader activation enabled and **reCAPTCHA off** (the decoy only renders when reCAPTCHA is off), view source on any front-end page and find the auth modal's honeypot: `<input class="nphp" … name="email" …>`. Confirm it now carries `style="display:none;"` and the three `data-*` opt-out attributes.
2. With a **password manager** enabled (1Password, LastPass, Dashlane, Bitwarden), open the sign-in form and let it fill the email field. Only `npe` should be filled. Submit — you should land in the normal flow (password or one-time code), not on a success screen. Chrome's own autofill is worth trying too, though it already leaves the decoy alone on both the old and new markup (see below), so it won't show a difference.
3. To see the old failure directly: submit the form with a value in `email` as well as `npe` (e.g. via devtools). The response is a 200 with `authenticated: 1` and the modal shows "Success! You're signed in." — with no session. That is unchanged and deliberate: the trap still works exactly as before. This PR only stops browsers from putting a reader there.
4. Repeat step 1 on a page with the Newsletters Subscribe block.
5. Tests: `n test-php --group auth-honeypot` (newspack-plugin) and `--group subscribe_honeypot` (newspack-newsletters). Both fail against the previous markup and pass with this change.

**After deploying:** the honeypot ships inside cached page HTML, so confirm by viewing source on a live page rather than by the version number, and purge page/CDN caches so readers stop being served the old markup.

### Other information:

Full suites green — newspack-plugin 2349 tests / 6677 assertions, newspack-newsletters 705 / 2950 — and CI-parity PHPCS clean. RED→GREEN captured for both new test groups.

### What's confirmed and what isn't

Being explicit, because an earlier revision of this description over-claimed.

**Established.** Working back from the reporter's screenshots and the site's data:

- The success screen they photographed carries the *sign-in* label with no description, so the response was a 200 with no `action` and no `registered`.
- Their account has a real password hash, so a plain sign-in should have returned a password prompt — which their screenshots don't show.
- No `Logged in` event exists for that account across two months of retained history, so no session was ever created.
- Enumerating every response payload and every `endLoginFlow()` caller at the shipped tag, exactly two paths produce that screen with no session: the honeypot branch, and the `link` branch. `link` is unreachable from any UI — `FORM_ALLOWED_ACTIONS` excludes it, both JS callers that post it handle the response themselves, and the `np_auth_strategy` mechanism that could have driven it is dead code.

**So a value reached the `email` field.** That part holds.

**Not established: what put it there.** Chrome's own autofill is *excluded* — tested by hand against a page carrying this exact markup, in a real profile. It fills `npe` and leaves the decoy empty, and it does so on the **old markup too**. So the Chrome result confirms Chrome isn't the filler; it does not demonstrate this change fixing anything for Chrome.

**1Password is excluded too** — same method, Firefox, against a faithful copy of the form (including the hidden `type="password"` input, which is what makes a manager treat it as a login form at all). It fills the real field and leaves the decoy alone on the pre-fix markup.

That leaves, with no local reproduction after two rounds: another password manager (LastPass, Bitwarden, Dashlane), a different browser build, or a managed/enterprise autofill policy — the reporter is on a corporate network. Rather than keep guessing at combinations, the logging above is what will answer it: if the branch fires in production, its user agent names the filler; if it never fires, the honeypot isn't the cause and the search moves elsewhere.

**What that means for this PR.** The change is right regardless — a decoy that any automated filler can populate is a defect, and `display:none` plus the vendor opt-outs is strictly more robust than a 1×1px clipped box. But review it as *hardening plus a diagnostic*, not as a verified fix for the report.


plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
...
+			 */
+			$honeypot_log_throttle = 'newspack_honeypot_log_throttle';
+			if ( false === \get_transient( $honeypot_log_throttle ) ) {
+				\set_transient( $honeypot_log_throttle, 1, 5 * MINUTE_IN_SECONDS );
+				Logger::newspack_log(
+					'newspack_auth_honeypot_tripped',
+					'Reader auth honeypot tripped — returning a fake success with no session.',
+					[
+						'has_npe'        => ! empty( $email ),
+						'matches_npe'    => ! empty( $email ) && strtolower( trim( $honeypot ) ) === strtolower( trim( $email ) ),
```
jason10lee, line 2173: **Blocker:** `matches_npe` and `decoy_is_email` can both report the bot shape for a browser-autofill submission, which is the one case this log exists to identify.

The cause is two lines up from the hunk, at line 2123: `$_POST['email']` is sanitized without `wp_unslash()`. WordPress runs `addslashes()` over `$_POST` in `wp_magic_quotes()`, and `sanitize_text_field()` does not remove those slashes, while `sanitize_email()` on line 2120 does. So for an address with an apostrophe in the local part, `$honeypot` is `o\'brien@example.com` while `$email` is `o'brien@example.com`, and the comparison on this line fails.

Since the plan is to act on this data, and to read a quiet log as evidence the honeypot is not the cause, a flag that inverts under an apostrophe works against the conclusion rather than adding noise.

The omission is shared with the surrounding pre-existing lines, so I think the narrow fix is line 2123 only:

```php
$honeypot = isset( $_POST['email'] ) ? \sanitize_text_field( \wp_unslash( $_POST['email'] ) ) : '';
```

<details>
<summary>Runtime evidence</summary>

Posted byte-identical `npe` and `email` values of `o'brien@example.com` (19 characters) against a running site:

```
matches_npe:    false
decoy_is_email: false
decoy_length:   20
```

Control, with `reader@example.com`: `matches_npe: true`, `decoy_is_email: true`, `decoy_length: 18`.

</details>

plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
...
+			 * not logged: the shape of the submission is enough to tell a browser filling
+			 * both fields from one profile entry (`matches_npe`) apart from a bot filling
+			 * the decoy alone.
+			 *
+			 * Throttled, because `newspack_log` reaches central logging and on a site
+			 * without reCAPTCHA this trap can fire on ordinary bot traffic. The question
+			 * is whether it fires and what fills it, which a sample answers; every
+			 * occurrence would push one site's spam volume into a shared store.
+			 */
+			$honeypot_log_throttle = 'newspack_honeypot_log_throttle';
```
jason10lee, line 2165: **Suggestion:** A single global key means the first trip in each five-minute window wins. On a site without reCAPTCHA, which is the only kind that renders the decoy at all, bot volume will claim nearly every window, so the rare browser-autofill trip is the one most likely to be dropped. A log that then shows only `matches_npe: false` entries would support "only bots" by construction rather than by evidence.

Keying the throttle on the discriminating shape gives the autofill case its own budget, at no extra volume in the common case:

```php
$honeypot_log_throttle = 'newspack_honeypot_log_throttle_' . ( $matches_npe ? 'autofill' : 'bot' );
```

That needs `$matches_npe` hoisted above the throttle check, which also makes it available to the log array rather than computing it inline.

<details>
<summary>Runtime evidence</summary>

With the transient cleared, a bot-shaped POST followed one second later by a reader-shaped one logged only the bot:

```
has_npe: false, matches_npe: false, user_agent: SpamBot/1.0
```

The reader-shaped submission that arrived a second later produced no entry.

</details>

plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
...
+			if ( false === \get_transient( $honeypot_log_throttle ) ) {
+				\set_transient( $honeypot_log_throttle, 1, 5 * MINUTE_IN_SECONDS );
+				Logger::newspack_log(
```
jason10lee, line 2168: **Suggestion:** Four server-side branches read the same decoy and return a fake success. Only this one logs. The markup fix reaches all four, because both render functions are shared, so that half is complete; the diagnostic does not.

The reader-registration block matters most here: it returns `authenticated: true`, producing the identical told-you-are-signed-in-then-you-are-not sequence. If the reports originated there, five minutes of clean auth-form logs will read as an all-clear.

Instrumenting the block branch is a small addition if the throttle is factored into a helper. At minimum, naming the source form in the log message would keep a future reader of those entries from assuming the coverage is total.

<details>
<summary>The four branches</summary>

| Branch | Response | Logs |
|---|---|---|
| `class-reader-activation.php:2146` | `authenticated: 1` | yes |
| `src/blocks/reader-registration/index.php:462` | `authenticated: true` | no |
| `class-reader-registration.php:507` | `status: 'created'` | no |
| `newspack-newsletters .../subscribe/index.php:546` | fake success | no |

</details>

plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
...
+			 * both fields from one profile entry (`matches_npe`) apart from a bot filling
+			 * the decoy alone.
+			 *
+			 * Throttled, because `newspack_log` reaches central logging and on a site
+			 * without reCAPTCHA this trap can fire on ordinary bot traffic. The question
+			 * is whether it fires and what fills it, which a sample answers; every
+			 * occurrence would push one site's spam volume into a shared store.
+			 */
+			$honeypot_log_throttle = 'newspack_honeypot_log_throttle';
+			if ( false === \get_transient( $honeypot_log_throttle ) ) {
```
jason10lee, line 2166: **Suggestion:** `get_transient()` then `set_transient()` is a read-then-write with no atomicity, so concurrent trips, which is the normal case under a spam burst, can all observe `false` and all log. That is the condition the throttle exists to prevent.

This subsystem already has an atomic pattern for exactly this, at `class-reader-registration.php:366-374`: branch on `wp_using_ext_object_cache()` and use `wp_cache_add()`, whose memcached semantics make a `false` return the "already throttled" answer.

It also avoids a second-order effect. On a site without a persistent object cache, `set_transient()` here is an unauthenticated-traffic-driven write to `wp_options` from `template_redirect`, on a request that otherwise performs no writes.

Keeping the `set` before the log is the right order, though. It means a slow or fatal listener inside `newspack_log` cannot leave the guard unset.

plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
...
+				Logger::newspack_log(
+					'newspack_auth_honeypot_tripped',
+					'Reader auth honeypot tripped — returning a fake success with no session.',
+					[
+						'has_npe'        => ! empty( $email ),
+						'matches_npe'    => ! empty( $email ) && strtolower( trim( $honeypot ) ) === strtolower( trim( $email ) ),
+						'decoy_length'   => strlen( $honeypot ),
+						'decoy_is_email' => (bool) \is_email( $honeypot ),
+					],
+					'warning'
```
jason10lee, line 2177: **Suggestion:** This file's own subsystem routes visitor-triggered conditions to `'debug'`. `class-reader-registration.php:558-564` logs a bot-driven invalid email that way, with the comment: "Visitor-triggered client condition (bots, malformed submissions) — routine once capture is live, so 'debug' (logstash only), not 'error'". The rate-limit crossing at `:380-392` says the same.

By the PR's own account this branch fires on ordinary spam wherever reCAPTCHA is off, which makes it the same class of event. `'debug'` still reaches logstash, which is all the investigation needs.

plugins/newspack-plugin/includes/reader-activation/class-reader-activation.php
```diff
@@ -2126,6 +2144,39 @@ public static function process_auth_form() {
 
 		// Honeypot trap.
```
jason10lee, line 2145: **Nit:** `Logger::newspack_log` fires its action unconditionally — unlike `Logger::log`, which is gated on `NEWSPACK_LOG_LEVEL` (`includes/class-logger.php:39`). So once this ships there's no constant, option or filter that stops it short of another release.

For a probe added to answer one specific question, a `@todo` naming the condition for removing it would keep it from becoming permanent by default.

plugins/newspack-plugin/src/reader-activation-auth/style.scss
```diff
@@ -5,6 +5,10 @@
 .newspack-reader-auth {
 	.nphp,
 	input[type="email"].nphp {
+		/* `display: none` rather than only clipping: a clipped field still renders, and
+		 * browsers autofill fields that render. The field also carries this inline, so
+		 * this rule is the fallback for a policy that strips the style attribute. */
+		display: none;
```
jason10lee, line 11: **Nit:** The `display:none` added to the three `.nphp` rules is described in every comment as the fallback for a stripped `style` attribute, and no test pins it. A future SCSS tidy that removes the line as "redundant with the inline style" would ship green, quietly taking the fallback with it.

A test that reads the built stylesheet would cover it. Failing that, the comment is doing the work alone.

jason10lee (changes requested):
This is most of the way there!  We turned up one blocker, plus four suggestions and a couple nits.  Markup in an isolated env looks good, but the diagnostic side escapes and unescapes apostrophes/single quotes asymmetrically, which means legitimate browser autofills fail the bot test.

Cleanest path may be to `wp_unslash` the email.  

Claude suggests that since the diagnostic step looks like a self-contained commit, another quick fix might be to drop that change from the hotfix and merge it as a new PR against `main` to let it go through alpha testing.

Release risk: HIGH. One confirmed defect, in code shipping on a hotfix branch that skips the biweekly QA net.

<details>
<summary>Repro evidence and verification detail (for reviewers)</summary>

**Environment.** Isolated env built from the PR branch at `4528946`, both plugins active, reader activation on, reCAPTCHA off, with a probe hooked to the `newspack_log` action so emissions are observable.

**The unslash defect.** Posted byte-identical `npe` and `email` values of `o'brien@example.com` (19 characters):

```
matches_npe:    false
decoy_is_email: false
decoy_length:   20
```

The control, `reader@example.com`, returns `matches_npe: true`, `decoy_is_email: true`, `decoy_length: 18`. `wp_magic_quotes()` runs `addslashes()` over `$_POST`; `sanitize_email()` on line 2120 strips the slash from `npe`, and `sanitize_text_field()` on line 2123 leaves it on the decoy.

**The throttle drops the reader-shaped event.** With the transient cleared, a bot-shaped POST followed one second later by a reader-shaped one logged only the bot:

```
has_npe: false, matches_npe: false, user_agent: SpamBot/1.0
```

Three POSTs inside one window produced exactly one entry, so the throttle itself works as designed.

**The four honeypot branches.** All four read the same decoy and return a fake success:

| Branch | Response | Logs |
|---|---|---|
| `class-reader-activation.php:2146` | `authenticated: 1` | yes |
| `src/blocks/reader-registration/index.php:462` | `authenticated: true` | no |
| `class-reader-registration.php:507` | `status: 'created'` | no |
| `newspack-newsletters .../subscribe/index.php:546` | fake success | no |

The markup fix reaches all four, because both render functions are shared.

**Came back clean.** `user_email` is empty in the log because `process_auth_form()` returns at line 2107 for an authenticated request, and the payload is three booleans plus a length. `FormData` still submits a `display:none` control, so all four `! empty()` checks behave as before. The `visuallyHidden` mixin sets no `display`, so declaration order is safe. `<input>` is not in `$allowedposttags`, so no kses path touches this markup. Full suites pass (newspack-plugin 2825 tests, newspack-newsletters 797), and PHPCS against the root ruleset is clean on all four changed PHP files.

</details>

dkoo:
Thanks, @jason10lee! I'm going to hold on this one, actually, because it may not be needed after all. I'll follow up with the final verdict after I hear from the publisher who reported the issue.

jason10lee:
> Thanks, @jason10lee! I'm going to hold on this one, actually, because it may not be needed after all. I'll follow up with the final verdict after I hear from the publisher who reported the issue.

Thanks; standing by.

dkoo:
The publisher has confirmed that our recommended workaround solves the issue this PR attempts to fix with code, so the code-based fix isn't necessary. Closing.

PR #805, opened 2026-08-05
fix(indesign-export): normalize line endings so exports place cleanly

miguelpeixe:
### All Submissions:

Exports place into InDesign as literal markup instead of formatted copy. Two tickets reported this with **opposite** manual fixes: NPPM-2813 needed the header changed to `<ASCII-MAC>`, NPPM-3098 needed `<ASCII-WIN>`. Both publishers were on Mac.

The header was never the cause. It declares the file's own line endings — `<ASCII-WIN>` promises CRLF, `<ASCII-MAC>` promises bare CR — and the exporter's endings varied with the shape of the post. `clean_whitespace()` collapsed blank lines with `/\n{2,}/`, which is not CRLF-aware: it consumed the LF of a CRLF pair and stranded the CR, emitting `\r\r\n`. Classic content hit that path and needed `<ASCII-MAC>`; block content stayed uniform and needed `<ASCII-WIN>`.

Line endings are now normalized to CRLF, so `<ASCII-WIN>` is accurate for every post. The Header platform setting, its User-Agent detection, and the `newspack_indesign_export_platform` filter are removed — they configured a header that has one correct value.

Closes NPPM-3098. Closes NPPM-2813.

1. Enable the module at **Newspack > Settings > Print > Adobe InDesign**.
2. Confirm the "Header platform" section is gone. "Available post types" and "Photo captions" still render.
3. Change a setting in each remaining section. Both persist across a reload.
4. Publish a post written in the block editor. Export it from the posts list.
5. Run `xxd` on the download. Every line ends `0d 0a`, with no bare `0d` or `0a`.
6. Publish a post using the classic editor, or paste newline-separated `<p>` tags into a Custom HTML block. Export it.
7. Run `xxd` on that download. It also ends every line `0d 0a` — on `main` this file contains `0d 0d 0a`.
8. Place both exports in InDesign on a Mac. Paragraph styles apply instead of appearing as literal text.
9. Run `wp option update newspack_indesign_export_platform mac`, then re-export. The header is still `<ASCII-WIN>`.

### Other information:

* [x] Have you successfully run tests with your changes locally? Full `newspack-plugin` suite green (2592 tests).

Step 8 is the one check I could not run — I have no InDesign. The byte-level behavior is covered by a data-provider test across seven content shapes; four of them emit a bare CR on `main`. Confirmation from a publisher on the classic-content path would be worth having before release.

The `newspack_indesign_export_platform` filter is a public extension point, added in #463 three weeks ago. No consumers exist in the monorepo or in `repos/`. Stored option rows are left in place; they are no longer read.

### DIFF
--- a/plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
+++ b/plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
@@ -24,39 +24,6 @@ class InDesign_Exporter {
 	public const MODULE_NAME = 'indesign-export';
 
 	/**
-	 * Option name storing the platform header preference.
-	 *
-	 * Accepts 'auto', 'mac', or 'win'. 'auto' resolves the header at export
-	 * time from the requesting browser's User-Agent.
-	 *
-	 * @var string
-	 */
-	public const PLATFORM_OPTION = 'newspack_indesign_export_platform';
-
-	/**
-	 * Default value for the platform option.
-	 *
-	 * 'win' rather than 'auto': the Tagged Text start tag declares the format of
-	 * the exported file itself — <ASCII-WIN> for CRLF-delimited, <ASCII-MAC> for
-	 * bare-CR — not the operating system running InDesign. The converter joins
-	 * every line with CRLF unconditionally, so <ASCII-WIN> is the only header
-	 * that describes what is actually written. Resolving from the requesting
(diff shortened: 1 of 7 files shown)


miguelpeixe:
Superseded by #806, which targets `release` as a hotfix. Same commits, same scope.

PR #806, opened 2026-08-05
fix(indesign-export): normalize line endings and escape angle brackets

miguelpeixe:
### All Submissions:

InDesign was placing exports as literal markup instead of formatted copy (NPPM-3098, NPPM-2813 — four publishers, one mid-deadline). The defect class is single: the file disagreeing with what its first line declares, or author content read as Tagged Text syntax. This PR makes every export internally consistent, escapes author content, and turns the header platform into an explicit two-format choice.

**The format is selected as a pair.** The Tagged Text header declares the file's line-ending convention, so the file has to agree with itself. It didn't: `clean_whitespace()` collapsed blank lines with `/\n{2,}/`, which is not CRLF-aware — it consumed the LF of a CRLF pair and stranded the CR, emitting `\r\r\n` — so the endings varied with the shape of the post, and publishers saw opposite manual fixes work (NPPM-2813 needed `<ASCII-MAC>`, NPPM-3098 needed `<ASCII-WIN>`). The converter now emits one of two self-consistent formats — Windows (`<ASCII-WIN>` + CRLF, the default) or Mac (`<ASCII-MAC>` + CR) — with the header and every terminator switched together, and a final normalization pass in `convert_post()` guaranteeing no stray byte survives.

**Header platform is a site setting.** Placement testing showed that which format an install accepts is a property of the install: some InDesign setups recognize only the Mac form and place a `<ASCII-WIN>` file as plain text, header and tags included, while others accept Windows. The setting offers exactly the two real formats, defaulting to Windows. The User-Agent auto-detection and the `newspack_indesign_export_platform` filter from #463 are removed — the format is a deliberate site choice, not a guess from the browser. The option name is reused, so rows written by #463 keep working; its removed `auto` value maps to the default.

**Author content is escaped.** Text containing `<`, `>`, or `\` reached the file unescaped, so a post about HTML emitted what InDesign parses as tags, and a backslash swallowed the character after it. Content is now backslash-escaped per the Tagged Text spec without touching the converter's own markup: bracket entities are shielded through the decode in every spelling (`&lt;`, `&LT;`, decimal, and hex forms), backslashes — literal or entity-encoded — are escaped after it, and quote-block text is inserted literally so copy like "$1 million" can't be read as a regex backreference. Image captions are rich text (the caption toolbar offers links, bold, and italics), so caption markup converts the way body content converts — links reduce to their text, bold and italics become character styles — instead of escaping into literal printed text.

**Photo captions and credits share one toggle.** Per the follow-up request on NPPM-3098, "Exclude photo captions and credits" omits both: the two fields are appended together as one section. Release-note callout: sites that already checked the box stop exporting credits after this update; that is the requested behavior.

Closes NPPM-3098. Closes NPPM-2813.

1. Enable the module at **Newspack > Settings > Print > Adobe InDesign**.
2. Confirm "Header platform" offers Windows (default) and Mac, and that "Available post types" and "Photo captions and credits" render alongside it. Change a setting in each section. All persist across a reload.
3. Publish a post written in the block editor. Export it from the posts list.
4. Run `xxd` on the download. The first line is `<ASCII-WIN>` and every line ends `0d 0a`, with no bare `0d` or `0a`.
5. Publish a post using the classic editor, or paste newline-separated `<p>` tags into a Custom HTML block. Export it. It is also uniformly `0d 0a` — on `release` this file contains `0d 0d 0a`.
6. Switch "Header platform" to Mac and re-export. The first line is `<ASCII-MAC>` and every line ends `0d`, with no `0a` anywhere in the file.
7. Run `wp option update newspack_indesign_export_platform auto` (a value stored by earlier releases), then re-export. The header is `<ASCII-WIN>` — unknown values fall back to the default.
8. Publish a post whose title contains `<div>`, `a < b`, and a backslash, and whose body mentions `<div>` typed in the block editor. Export it. Those brackets appear as `\<` and `\>`, the backslash as `\\`, and `<pstyle:...>` tags are unescaped. (A bare bracket pasted as raw classic-editor HTML passes through by design.)
9. Add a quote block reading "We raised $1 million last year." and export. The quote text comes through intact.
10. Add an image whose caption contains a link and italics. Export it. The caption reads as its text — no `\<a` markup — with the italics carried as `<cTypeface:Italic>`.
11. Check "Exclude photo captions and credits" and re-export. The file contains neither `<pstyle:PhotoCaption>` nor `<pstyle:PhotoCredit>`.
12. Place a Windows export and a Mac export in InDesign. Each places with paragraph styles applied, no literal tags, and no extra blank lines. On an install that rejects the Windows form, the Mac export is the one that places.

### Other information:

* [x] Have you successfully run tests with your changes locally? Full `newspack-plugin` suite green against the `release` tree (2857 tests).

Byte-level behavior is pinned by data-provider tests across eight content shapes under both formats, plus escape tests for every bracket and backslash spelling. Placement itself has been exercised where it matters most: the publisher whose install motivated the Mac format places `<ASCII-MAC>` files and rejects `<ASCII-WIN>` ones. Still worth confirming before ship: a Mac-format export from this branch placing clean on that install (the file they hand-edited still carried CRLF endings, which is what produced the extra line breaks they saw), and a Windows-format export re-checked on an install that accepts it.

Targets `release` as a hotfix. `release` already carries #463, and the files this touches are otherwise identical between `release` and `main`.

The changes ship together deliberately: they are one defect class in one converter, and separating them would mean repeated passes over the same file and repeated rounds of publisher verification.

The `newspack_indesign_export_platform` filter was public briefly (added in #463 on 2026-07-21, stable 2026-08-03); a code search across `Automattic` and `newspack-publishers` finds no callers of it or of the removed User-Agent methods.


plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -115,7 +137,15 @@ public function convert_post( $post, $options = [] ) {
...
-		return implode( "\r\n", array_filter( $content_parts ) );
+		$content = implode( self::EOL, array_filter( $content_parts ) );
+
+		// Final guarantee that the file matches what self::START_TAG declares.
+		// Post content reaches the converter with line endings of every flavor
+		// (pasted copy, imported HTML, serialized blocks), and the conversion
+		// steps above introduce their own. A single stray CR or LF makes part of
+		// the file Mac- or Unix-terminated while the header promises CRLF, and
+		// InDesign then places that stretch as literal markup.
+		return preg_replace( '/\r\n|\r|\n/', self::EOL, $content );
```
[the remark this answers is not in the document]
miguelpeixe: Fixed in 56cbfb0. The result is captured and falls back to the un-normalized content when `preg_replace()` returns null, so `convert_post()` keeps its documented `string|false` contract.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -23,30 +23,6 @@ class InDesign_Exporter {
 	 */
 	public const MODULE_NAME = 'indesign-export';
```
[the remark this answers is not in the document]
miguelpeixe: Not adding shims here, based on a search rather than an assumption.

Across `Automattic` and `newspack-publishers`, the only hits for `newspack_indesign_export_platform`, `get_platform_setting()`, `sniff_user_agent_platform()`, and `resolve_platform()` are this repository. The other `resolve_platform` matches belong to WooCommerce's unrelated `PlatformRegistry`.

The exposure window is also narrow: these symbols merged in #463 on 2026-07-21 and reached stable on 2026-08-03.

The stronger reason is that a shim would have to lie. `resolve_platform()` returning `'win'` preserves a concept this PR disproves — the start tag describes the file's own line endings, so there is exactly one correct value. Keeping the API alive suggests platform selection still means something and invites new callers.

One clarification on blast radius: `add_filter( 'newspack_indesign_export_platform', ... )` already degrades safely. A callback registered on a hook that no longer runs is a no-op, not a fatal. Only direct calls to the static methods or class constants could fatal, and none exist.

Happy to revisit if you know of a caller outside those two orgs.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -95,6 +73,14 @@ class InDesign_Exporter {
 	 * Initialize the module.
 	 */
 	public static function init() {
+		// The Header platform setting was removed (#806) and nothing reads its
+		// option anymore, but stored rows were autoloaded. Clear any that
+		// remain; get_option() resolves from the alloptions/notoptions cache,
+		// so this costs a write only on the one request that still finds a row.
+		if ( false !== get_option( 'newspack_indesign_export_platform' ) ) {
```
dkoo: **Suggestion:** Legacy-option lookup runs on every request, ahead of the module gates — `init()` runs at file scope from `Newspack::includes()`, and this `get_option()` sits above both `is_feature_enabled()` and the module-active check — so it fires on every front-end, admin, REST, and cron request of every site, including the majority that never enabled the module. The comment's cost model holds only where a persistent object cache backs `notoptions`; without one, WordPress issues a fresh `SELECT` per request, and the guard has no removal path. `delete_option()` on a missing row is harmless, so this is cost, not correctness.

**Consider:** Move the cleanup below the feature and module gates, or run it once from an upgrade routine, so sites that never enabled the exporter don't carry the lookup on every request indefinitely.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
```diff
@@ -476,17 +458,7 @@ private static function export_posts( $post_ids ) {
...
-	public static function get_platform_setting() {
-		$value = get_option( self::PLATFORM_OPTION, self::PLATFORM_DEFAULT );
-		return in_array( $value, self::ALLOWED_PLATFORMS, true ) ? $value : self::PLATFORM_DEFAULT;
-	}
-
-	/**
-	 * Whether photo captions should be excluded from exports.
+	 * Whether photo captions and credits should be excluded from exports.
 	 *
 	 * Photo credits are a separate attribution field and are always exported.
```
dkoo: **Suggestion:** Docblock still says photo credits are always exported — The summary line above was updated to "captions and credits", but this sentence survived and now states the opposite of what `process_post_images()` does (it returns `''` for the whole section when the flag is set). Since the wider meaning of the setting is the surprising part of this PR for opted-in sites, this is the line a future maintainer is most likely to be misled by. One-line deletion.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -366,7 +407,7 @@ private function process_quotes( $content ) {
 			if ( ! empty( $cite_matches ) ) {
 				$cite = $cite_matches[1];
 				if ( ! empty( $cite ) ) {
-					$quote_content .= "\r\n" . $this->styles['pullquote_name'] . wp_strip_all_tags( $cite );
+					$quote_content .= self::EOL . $this->styles['pullquote_name'] . wp_strip_all_tags( $cite );
```
dkoo: **Suggestion:** Pre-existing: `$1` in a pull quote corrupts the export via a regex backreference — Not introduced here — `origin/release` behaves identically — but it produces exactly the symptom these tickets describe, so it's worth its own ticket rather than a reopen. `$quote_content` is passed as the *replacement* argument to `preg_replace()` at line 414, so `$` + digit sequences in author text are read as backreferences.

Observed on the branch and on `release`: a quote block reading `We raised $1 million last year.` exports as
`<pstyle:blockquote>We raised <pstyle:text>We raised $1 million last year.\n million last year.`

**To reproduce**
1. Create a post with a Quote block containing `We raised $1 million last year.`
2. Export it from the posts list.
3. The blockquote line contains the duplicated text above.

`preg_replace_callback()` with `fn() => $quote_content`, or `str_replace()` on the literal match, would fix it.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -1210,4 +1078,337 @@ public function test_caption_only_image_adds_no_content_when_captions_excluded()
...
+	 * Test that bare angle brackets in body copy pass through unescaped.
+	 *
+	 * Escaping applies to entity-encoded brackets: get_transformed_text() runs
+	 * over content already carrying the converter's own tags, so a bare bracket
+	 * cannot be escaped without breaking them. Stored bare brackets are rare —
+	 * the editor saves a literal < as &lt; (though not >), and kses rewrites a
+	 * loose "< " to "&lt; " for authors without unfiltered_html — so this pins
+	 * the floor: raw brackets survive as-is.
+	 */
+	public function test_bare_angle_brackets_in_body_content_pass_through() {
```
dkoo, line 1137: **Suggestion:** PR test plan overstates body-bracket escaping — This test pins the documented floor: a *bare* `<`/`>` in body HTML (classic editor, raw paste, or any admin who bypasses kses) passes through unescaped, because `get_transformed_text()` can't tell it from the converter's own tags at that stage. Confirmed at runtime — `<p>a < b</p>` exports as `a < b`. The code is consistent; the PR's step 8–9 ("body mentions `<div>` typed in the block editor") only holds for entity-encoded brackets, which is what RichText stores. Worth tightening the description so hotfix QA doesn't chase the bare case as a failure.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -14,6 +14,33 @@
...
+	 * with self::EOL.
+	 *
+	 * @var string
+	 */
+	private const START_TAG = '<ASCII-WIN>';
+
+	/**
+	 * Line terminator used throughout the exported file.
+	 *
+	 * Must stay consistent with self::START_TAG: <ASCII-WIN> promises InDesign
```
dkoo: **Nit:** `EOL` docblock leans on a spec claim the fix doesn't need — The comment says `<ASCII-WIN>` "promises" CRLF and `<ASCII-MAC>` a bare CR. In the Tagged Text spec the suffix denotes the platform character encoding, and the importer is generally tolerant of CR/LF/CRLF. I couldn't check the spec from here so I'm not calling it wrong — and it doesn't matter for correctness, because the fix works on internal consistency alone (one terminator everywhere, 7-bit payload). Softening the comment to the consistency argument would keep it durable either way.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -89,33 +112,45 @@ public function convert_post( $post, $options = [] ) {
...
-		return implode( "\r\n", array_filter( $content_parts ) );
+		$content = implode( self::EOL, array_filter( $content_parts ) );
+
+		// Final guarantee that the file matches what self::START_TAG declares.
+		// Post content reaches the converter with line endings of every flavor
+		// (pasted copy, imported HTML, serialized blocks), and the conversion
+		// steps above introduce their own. A single stray CR or LF makes part of
+		// the file Mac- or Unix-terminated while the header promises CRLF, and
+		// InDesign then places that stretch as literal markup.
+		$normalized = preg_replace( '/\r\n|\r|\n/', self::EOL, $content );
```
dkoo: **Nit:** Null fallback silently ships the un-normalized file — The fallback keeps the `string|false` contract, and a PCRE failure on `/\r\n|\r|\n/` over post-sized content is close to impossible. But if it ever fires, the publisher gets exactly the mixed-ending file this PR fixes, with no signal. Returning `false` (which `export_posts()` already treats as a failed post) or logging via `Logger` would make that visible. Judgment call; leaving as-is is defensible for a hotfix.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -432,9 +473,54 @@ private function convert_html_to_indesign( $content ) {
...
+	 * printed text — so convert it the way the body is converted (formatting
+	 * to character styles, other tags to their text). Entity-encoded brackets
+	 * are content and still come out escaped.
+	 *
+	 * @param string $text Text to convert.
+	 *
+	 * @return string Converted text.
+	 */
+	private function get_transformed_rich_text( $text ) {
+		return trim( $this->get_transformed_text( $this->convert_html_to_indesign( $text ) ) );
```
dkoo, line 564: **Nit:** Rich-caption conversion inherits block-level rules — `get_transformed_rich_text()` runs the full `convert_html_to_indesign()` table, including paragraph, list, and rule conversions a caption can't meaningfully hold. A caption typed with `<p>` markup in the Media Library textarea yields `<pstyle:PhotoCaption><pstyle:text>First.\r\n<pstyle:text>Second.` — the second half styled `text`, not `PhotoCaption`. Low likelihood and strictly better than escaping the markup as literal text; noting for awareness.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -458,11 +544,33 @@ private function get_transformed_text( $text ) {
...
+
+		// Angle brackets written as entities are content, not tag delimiters.
+		// Shield them from the decode below — which would hand InDesign a bare
+		// bracket to read as the start or end of a tag — then resolve them to
+		// escaped literals. Covers the numeric forms and the named forms in both
+		// letter cases that mean a plain bracket: &LT;/&GT; are valid spellings
+		// of the same characters, while &Lt;/&Gt; are much-less/greater-than and
+		// must stay out. The resolutions after the decode also catch a named
+		// form the running PHP version's entity table leaves undecoded.
+		$text = preg_replace( '/&(lt|LT|gt|GT|#0*6[02]|#[xX]0*3[cCeE]);/', '&amp;$1;', $text );
```
dkoo, line 609: **Nit:** `&nvlt;` / `&nvgt;` slip past the entity shield — In PHP 8.3's HTML5 table, exactly four named entities decode to a string containing `<` or `>`: `&lt;`, `&gt;`, and `&nvlt;`/`&nvgt;` (bracket + U+20D2). The shield covers the first two; `x &nvlt; y` comes out as `x <<0x20D2> y` — a live bare bracket. Practically unreachable, and arguably outside the comment's stated scope, but the comment reads as exhaustive. Adding `|nvlt|nvgt` to the alternation, or a sentence noting the exclusion, would settle it.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
```diff
@@ -523,14 +635,22 @@ private function get_image_blocks( $blocks ) {
...
 	 * @return string Photo credit and caption tags.
 	 */
 	private function process_post_images( $post, $options = [] ) {
 		$include_captions = ! isset( $options['include_captions'] ) || $options['include_captions'];
-		$images           = [];
-		$inline_captions  = [];
+
+		// One toggle covers the whole section: publishers who exclude captions
+		// want the credits appended with them gone too (NPPM-3098).
+		if ( ! $include_captions ) {
```
dkoo, line 701: **Nit:** Hotfix carries a behavior change for opted-in sites — Sites that already checked "Exclude photo captions" stop exporting credits after this update, and a month-old UI control disappears. Both are described and justified in the PR body, so this isn't a request for change — only a flag that the release note for this patch should call both out, since neither is a pure fix from a publisher's seat.

**Consider:** Name the credits change explicitly in the release note, so publishers who turned the setting on under its previous meaning find out before they lay out a page from an export that no longer carries credits.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
```diff
@@ -1210,4 +1078,337 @@ public function test_caption_only_image_adds_no_content_when_captions_excluded()
...
+		$this->assertStringContainsString( '<pstyle:PhotoCredit>Credit \\<tag\\>', $content );
+	}
+
+	/**
+	 * Test that a literal backslash in content is escaped.
+	 *
+	 * A bare backslash is the Tagged Text escape character, so InDesign would
+	 * otherwise absorb the character that follows it.
+	 */
+	public function test_escapes_backslash_in_content() {
```
dkoo, line 1238: **Nit:** Small coverage gaps around the new escaping — Coverage is strong — both bug shapes in the line-ending provider, uppercase entities, `&Lt;`/`&Gt;` exclusion, entity backslashes, the backslash-next-to-bracket composition. Three untested shapes, each a few lines on the existing fixtures: a literal backslash in a *plain-text* field (title, byline, credit — the field set this PR newly routes through `get_transformed_plain_text()`); content CR-terminated throughout (the shape of imported legacy-Mac copy); and nested formatting in a caption (bold inside a link).

dkoo (approved):
@miguelpeixe The only testing step I wasn't able to confirm manually was 12 ("Place the exports in InDesign on a Mac") because I don't have InDesign. All other steps behave as described. Leaving Claude's non-blocking suggestions and nits inline in case you want to tackle any, otherwise this looks good to me!

Release risk: LOW — no blockers, the consumer surface is intact, and the fix was verified end-to-end against a running env (CRLF-only output on every fixture where `release` emits `\r\r\n`; brackets and backslashes escaped; converter tags untouched).

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Switched the check to wp_load_alloptions() in 904a8a152. alloptions is already in memory on every request, so the lookup is free with or without a persistent object cache, and the cleanup keeps its removal path — rows were written with default autoload, so any existing row appears there. That covers the cost on never-enabled sites without moving the delete behind the module gates.

plugins/newspack-plugin/includes/optional-modules/class-indesign-exporter.php
miguelpeixe: Removed in 904a8a152.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
miguelpeixe: Clarified in the PR body (step 9): escaping covers brackets the editor stores as entities, and a bare bracket pasted as raw classic HTML passing through is by design — named explicitly so QA reads it as expected behavior rather than a failure.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Reworded in 904a8a152. The comment now rests on the internal-consistency argument — one terminator everywhere, declared once — instead of the per-header line-ending claim.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Kept the fallback, with one correction to the premise: export_posts() has no false handling — it appends the return straight into the exported files, so false would ship an empty download (strlen coerces false to zero length), which is worse than the un-normalized file. Wiring a real per-post failure path felt like scope for a near-impossible branch; happy to take it as a follow-up if you think it earns one.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Noted in the docblock in 904a8a152 — the inherited block-level conversion is called out as the accepted floor for input the caption UI does not produce.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Documented the exclusion in 904a8a152: escaping &nvlt;/&nvgt; would drop the negation mark, so the comment now names them as deliberately left out instead of reading as exhaustive.

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Added a release-note callout to the PR body (Photo credits paragraph) so the patch note names the behavior change for sites that had the setting enabled under its previous meaning.

plugins/newspack-plugin/tests/unit-tests/indesign-exporter/indesign-exporter.php
miguelpeixe: All three added in 904a8a152: a legacy-Mac row in the line-ending provider (every line CR-terminated), a literal backslash through the plain-text fields (title and credit), and bold inside a link in a caption. 69 tests on the class; full suite green (2843).

plugins/newspack-plugin/includes/optional-modules/indesign-export/class-indesign-converter.php
miguelpeixe: Fixed here rather than ticketed, since it is the same symptom class these tickets describe: the quote is now inserted through preg_replace_callback() in 938358520, so author text is literal — your repro converts to <pstyle:blockquote>We raised $1 million last year. with a test pinning it.

PR #808, opened 2026-08-05
fix(newsletters): resolve click proxy redirect target consistently

chickenn00dle:
## Why

The newsletter click proxy (`np_newsletters_click`) decides whether a click destination is "on this site" before deciding how much further checking it needs. It made that call by asking whether the site URL appeared *anywhere* in the destination URL — a substring test, which answers a different question than the one being asked. It disagrees with where a reader would actually land, in both directions:

- A destination whose hostname merely *starts with* the site host (`example.com.something.test`) was classified as on-site.
- So was one that mentioned the site URL somewhere in its path, or ahead of an `@`.
- Meanwhile a genuine on-site link written with a different scheme, or with different capitalisation in the host, was classified as off-site and pushed through the extra ad-content lookup for no reason.

Net effect: the same destination could be handled two different ways depending on how the URL happened to be spelled.

## What changed

- New `Click::is_same_site_url()` classifies a destination by comparing its **host** to the site host, exactly and case-insensitively. Both `home_url()` and `site_url()` hosts count, since the two differ on subdirectory installs. The proxy now uses it in place of the substring test.
- The final redirect now runs the location through `wp_validate_redirect()` with `home_url()` as the fallback, allowing the resolved destination host only for the duration of that call. It was a bare `wp_redirect()` before.

Behaviour for real ad links is unchanged: an off-site destination still has to be present in the ad's rendered email HTML before a reader is sent there, and on-site destinations still skip that lookup.

## How to test

Automated — `n test-php` in `newspack-newsletters`: **705 tests, 2958 assertions, green** (3 pre-existing skips). Three new tests in `tests/test-tracking.php`:

- `test_is_same_site_url` — pins the classification across schemes, host casing, subdomains, prefix-lookalike hosts, path-embedded site URLs and host-less URLs.
- `test_handle_click_rejects_destination_that_only_mentions_the_site` — a destination that only *mentions* the site URL still goes through the ad-content lookup. Fails against the old substring test.
- `test_handle_click_allows_same_site_destination` — a genuine on-site destination is still accepted and tracked without appearing in the ad.

Manual:

1. With click tracking on, render a newsletter containing an ad with an external link, and click it. You land on the ad's destination as before, and the ad's click count increments.
2. Request `/?np_newsletters_click=1&id=<ad id>&url=<a URL not present in that ad>` — 400 "Invalid URL", as before.
3. Request the same with a URL on your own site — still redirects, no ad lookup needed.

Ref: [NPPM-3063](https://linear.app/a8c/issue/NPPM-3063/)


plugins/newspack-newsletters/includes/tracking/class-click.php
```diff
@@ -226,7 +264,21 @@ public static function handle_click( $with_redirect = true ) {
...
+			 */
+			$destination_host = self::get_url_host( $url );
+			\add_filter(
+				'allowed_redirect_hosts',
+				function ( $hosts ) use ( $destination_host ) {
+					$hosts[] = $destination_host;
+					return $hosts;
+				}
+			);
+			\wp_safe_redirect( \wp_validate_redirect( $url, \home_url() ) );
```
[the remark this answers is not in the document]
chickenn00dle: Good catch on the lifetime — fixed in 57c79b8, though not quite for the stated reason.

The filter can't be removed after `wp_safe_redirect()` because `exit` follows it, and it can't remove itself either: `wp_safe_redirect()` delegates to `wp_validate_redirect()`, so the filter was being applied twice and a one-shot would have made the second pass fall through to the `wp-admin` fallback. So the block now does explicitly what `wp_safe_redirect()` does internally — allow the host, validate, remove the filter, then redirect the already-validated location. The filter is off before anything is sent, and the fallback is `home_url()` rather than `wp-admin`, which is the more useful landing spot for a logged-out reader.

On the risk itself: after `exit` only shutdown handlers run, and the `Location` header is already sent, so an unintended second redirect wasn't really reachable. Scoping the filter tightly is still the right shape, so it's worth doing regardless.

Empty-host guard added as suggested. It's belt-and-braces — `wp_http_validate_url()` a few lines up already rejects anything without a host — but it costs nothing.

plugins/newspack-newsletters/includes/tracking/class-click.php
```diff
@@ -226,7 +264,25 @@ public static function handle_click( $with_redirect = true ) {
...
+			 * the request should inherit it. This is what wp_safe_redirect() does, minus
+			 * its wp-admin fallback, which is no use to a reader who is not logged in.
+			 */
+			$destination_host       = self::get_url_host( $url );
+			$allow_destination_host = function ( $hosts ) use ( $destination_host ) {
+				if ( $destination_host ) {
+					$hosts[] = $destination_host;
+				}
+				return $hosts;
+			};
```
miguelpeixe: The `$destination_host` is always lowercased, but `wp_validate_redirect()` compares the host it parses out of the location without lowercasing it, which is a strict `in_array()` check.

Consider also adding the raw destination host to the `$hosts` array.

plugins/newspack-newsletters/includes/tracking/class-click.php
```diff
@@ -168,6 +168,44 @@ public static function track_click( $ad_id, $email_address, $url, $newsletter_id
...
+	 * @param string $url URL to check.
+	 *
+	 * @return bool
+	 */
+	public static function is_same_site_url( $url ) {
+		$host = self::get_url_host( $url );
+		if ( ! $host ) {
+			return false;
+		}
+		// home_url() and site_url() differ when WordPress lives in its own subdirectory.
```
miguelpeixe: Given that this is now a host comparison, the subdirectory difference check is superfluous.

miguelpeixe (changes requested):

plugins/newspack-newsletters/includes/tracking/class-click.php
chickenn00dle: Good catch, and it was worse than a near-miss — the lowercased allowance broke destinations that used to work. `wp_validate_redirect()` matches with a strict `in_array()` against the host it parses out of the location, and `strtolower( $wpp['host'] ) !== $lp['host']` only covers the site's own host, so any off-site destination written with an uppercase letter fell through to the fallback.

Reproduced end to end in an isolated env, clicking through the real proxy with `https://Example.com/article/`:

| | Location |
| --- | --- |
| Before | `https://nppm3063.test` (the front page) |
| After | `https://Example.com/article/` |

Fixed in 74b953e5 by allowing the host both as written and lowercased. I also pulled the redirect resolution out into `Click::get_redirect_location()` — `exit` follows the call in `handle_click()`, so there was no way to pin this in a test otherwise. Covered now across lowercase, capitalised, uppercase, and a mixed-case host with a query and fragment.

plugins/newspack-newsletters/includes/tracking/class-click.php
chickenn00dle: Right — a subdirectory install parts company over the path, not the host, so the second comparison had nothing to add. Narrowed to `home_url()` in 74b953e5, which also lines it up with the host `wp_validate_redirect()` treats as this site, so the two agree on what "same site" means.

plugins/newspack-newsletters/includes/tracking/class-click.php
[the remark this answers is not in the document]
chickenn00dle: One correction to my earlier reply here: the lowercasing in that block turned out to break working destinations, not just tighten the filter's scope. `wp_validate_redirect()` compares hosts strictly, so an off-site link written as `Example.com` stopped resolving and readers landed on the front page instead. Fixed in 74b953e5 — the host is now allowed both as written and lowercased, and the block moved into `Click::get_redirect_location()` so it can be tested.

The empty-host guard you asked for is still there, now as an `array_filter()` over the two forms.

chickenn00dle:
Both points addressed in 74b953e5 — ready for another look.

The case one turned out to be a live regression rather than a hardening note. `wp_validate_redirect()` matches the host it parses out of the location with a strict `in_array()`, so the lowercased allowance missed any off-site destination written with a capital letter, and those readers landed on the front page instead of the link they clicked. Clicking `https://Example.com/article/` through the real proxy in an isolated env went to `https://nppm3063.test` before the fix and to the destination after it. The host is now allowed both as written and lowercased.

`is_same_site_url()` now measures against `home_url()` alone. A subdirectory install differs over the path, not the host, so the `site_url()` comparison had nothing to add — and dropping it lines the check up with the host `wp_validate_redirect()` already treats as this site.

Also extracted `Click::get_redirect_location()` from `handle_click()`. `exit` follows the redirect, so none of this was reachable from a test before; the case handling and the home-page fallback are both pinned now. Newsletters suite is 707 green and PHPCS is clean.

Beyond the two findings, I re-checked that the original behaviour still holds: an off-site link that isn't in the ad's email is refused, a same-site link goes straight through, and a destination that merely mentions the site host is still refused.

The changes were drafted with Claude Code and reviewed by me; the before/after above is from my own run in the env.

miguelpeixe (approved):

miguelpeixe (approved):

PR #810, opened 2026-08-05
feat(components): add a drawer component with a compound slot api

thomasguillot:
### All Submissions:

Adds a `Drawer` component to `newspack-components`: a right-edge slide-in modal panel for wp-admin. Four hand-rolled versions of this panel exist across the workspace today (newsletters quick-edit, Subscribers "View subscription", Content Gate save panel, Campaigns Edit Styles), each repeating the same keyframes and disagreeing on width, padding, and edge treatment. This gives them one component to migrate to. The migration is a separate PR.

The API is fully decomposed, mirroring the Drawer in `@wordpress/ui` so that adopting it once it stabilises is close to a rename inside our wrapper:

```jsx
<Drawer.Root isOpen={ isOpen } size="large" isDirty={ isDirty } onRequestClose={ close }>
	<Drawer.Header>
		<Icon icon={ settings } size={ 24 } />
		<Drawer.Title>{ __( 'Edit Styles', 'newspack-plugin' ) }</Drawer.Title>
		<Drawer.CloseIcon />
	</Drawer.Header>
	<Drawer.Content>{ fields }</Drawer.Content>
	<Drawer.Divider />
	<Drawer.Content padding={ 0 }>{ flushTable }</Drawer.Content>
	<Drawer.Footer>
		<Drawer.Action variant="secondary" closes>{ __( 'Cancel', 'newspack-plugin' ) }</Drawer.Action>
		<Drawer.Action variant="primary" onClick={ save }>{ __( 'Save', 'newspack-plugin' ) }</Drawer.Action>
	</Drawer.Footer>
</Drawer.Root>
```

Highlights:

* **Unsaved changes:** with `isDirty`, every close route the drawer owns opens a confirm dialog above the panel, which has not moved. The slide-out plays only after "Discard Changes". Wizards that already run `useUnsavedChangesDialog` pass their own `requestConfirm` instead, and keep ownership of that dialog.
* **Another modal opening** is core asking the drawer to step aside, not a user gesture. A clean panel takes it and closes. A dirty one does not answer at all, so it is still there once the new dialog is dealt with: its own confirmation would render underneath that dialog, and a delegated confirmation is itself what triggered the request.
* **Content sections:** each `Drawer.Content` is a section with `padding` and `gap` props on the 4px scale. Sections scroll together and are not self-separating; a `Drawer.Divider` between two of them draws a full-width rule. Anything goes inside one, plain text included: each child becomes a row, and a run of text with interpolations in it stays a single row.
* **Footer:** one or two actions share the row, three or more stack, in CSS alone. `closes` sends an action back through the close funnel, so Cancel gets the confirmation and Save does not. An action either navigates or acts: `href` rules out `onClick` and `closes` in the types, and the component holds a JavaScript caller to the same contract, so a link cannot navigate away while the confirmation is still opening. `onClick` receives the click event, so a handler can read modifier keys or stop propagation.
* **Sizes:** 280 / 350 (default) / 480 / 640 / full, all full-width under 600px. The panel spans the viewport height and rounds its free edge. An unrecognised size falls back to medium and warns in development, since the package is consumed from plain JS.
* **Accessibility:** `Drawer.Title` wires `aria-labelledby`, with a `contentLabel` fallback and a dev warning when neither exists. Focus returns to the opener on close. `Drawer.Action` warns in development when an `ariaLabel` drops the visible label (WCAG 2.5.3). The body becomes focusable and named for as long as it overflows, so a section holding nothing focusable can still be scrolled from the keyboard (WCAG 2.1.1). Escape ignores IME compositions, and reduced motion drops the animation.

Also on this branch, in the shared components the drawer builds on:

* `ConfirmDialog` and `Button` no longer crash without a react-router history, which `newspack-multibranded-site` hits today because it runs react-router v6 against the package's v5.
* `ConfirmDialog`'s `isOpen` now closes the dialog as well as opening it. The navigation blocker still raises it independently.
* `Button` drops a `javascript:` href on both the anchor and the `window.location` path, matching the guard `useUnsavedChangesDialog` already applied to intercepted links, and forwards the click event to a combined `href` + `onClick` handler.
* `useConfirmDialog` gains `cancelConfirm`, and clears its pending action even when the confirmed callback throws.

A demo card exercises the full API.

The drawer open over the components demo, with two content sections split by a full-width divider and a two-action footer

| Confirm above the unmoved drawer | Three actions stack |
| --- | --- |
| The unsaved-changes confirm dialog above the drawer, which has not moved | Three footer actions stacked full-width in a column |

Closes DSGNEWS-206.

1. `pnpm build` and open **Newspack → Components Demo**, then scroll to the Drawer card.
2. Open the drawer. It slides in from the right over the full viewport height, with a header row (icon, title, close), two content sections split by a full-width divider, and the footer.
3. Switch **Footer actions** to 3 and reopen. The actions stack full-width.
4. Tick **Unsaved changes** and try to close by icon, Escape, scrim click, and the footer's Cancel. The confirm dialog opens above the drawer each time. Cancel keeps the drawer in place; "Discard Changes" plays the slide-out.
5. With **Unsaved changes** still ticked, open the drawer and press Save. It closes without a confirmation.
6. Close the drawer with Escape. Focus returns to the "Open drawer" button.
7. Cycle the **Size** control. Widths are 280 / 350 / 480 / 640 / full. Narrow the window below 600px: every size goes full width and the rounded edge squares off.
8. Shorten the window until the sections overflow. The body scrolls while the header and footer stay pinned, and the sections keep their own height.
9. With the sections still overflowing, tab through the drawer. The body takes a tab stop and the arrow keys scroll it. Restore the height and the tab stop goes away.
10. `pnpm --filter newspack-components test` — the package has its own Jest suite (215 tests) and CI job.

### Other information:

<details>
<summary>For AI: architecture and review notes</summary>

* `packages/components/src/drawer/` — `root.tsx` (core Modal with `shouldCloseOnEsc`/`shouldCloseOnClickOutside` off and `__experimentalHideHeader`; one `requestClose` funnel; consecutive `Content`/`Divider` children grouped into a `__body` VStack scroll container keyed per body; context provides `requestClose` + title registration), `use-exit-animation.ts` (component owns the exit: `is-exiting` class, unmount on `animationend` with a timeout fallback; previous-`isOpen` tracked in state, render-safe), plus one file per slot.
* Close model: core's close paths are disabled so the confirm-before-slide-out ordering is structural, not a race. Core's `onRequestClose` is left as the one channel it still drives — its single-modal-at-a-time effect — and is answered unconditionally only while the panel is clean. It reads `isOpen`/`isDirty` through a ref refreshed in a layout effect, because core fills its dismisser ref in a passive effect and a sibling mounted earlier in the tree would otherwise reach the previous commit's closure.
* The Root withdraws only the confirmation it owns, during its own render. A delegated dialog is the caller's: the same `useUnsavedChangesDialog` instance serves the whole wizard and raises prompts the drawer never asked for, so there is no safe way for the Root to tell them apart. The README says so, and the Root warns in development if a delegated `requestConfirm` closes the panel without prompting (its `when` is false while `isDirty` is true).
* During the exit the close funnel is inert (`isOpen` guard), the frame is `inert`, and focus has been returned to the opener (gated on where focus actually is: inside the frame, or lost to `body` because the focused child unmounted with the close). A reopen inside the exit window refocuses the frame unless reopen content already claimed focus. Under reduced motion there is no exit to hand focus back during, so an unmount cleanup covers it.
* `Drawer.Title` registrations are a stack keyed by id, registered and unregistered in separate layout effects, so a text change cannot re-order them and a second title leaving cannot unname the one still rendered.
* Scrim dismissal is a completed gesture: primary `pointerdown` on the overlay arms a flag, document-level capture-phase `pointerup`/`pointercancel` complete or clear it, implicit pointer capture is released so the check holds for touch.
* The Escape handler ignores IME compositions (`isComposing` / keyCode 229) and matches `event.key` or `event.code`, mirroring core's `withIgnoreIMEEvents`.
* The body is a `VStack spacing={ 0 } justify="flex-start"`: Flex defaults to `space-between`, which would spread the sections down a tall panel. Flex's `ItemsColumn` class also sets `min-height: 0` on its children, which defeats the usual flex shrink protection, so `.newspack-drawer__body > * { flex: none }` keeps an overflowing body scrolling instead of squashing its sections. `body.tsx` owns the scroll affordance: a ResizeObserver over the body and its sections (border box, since a section's padding rides on an inline custom property) plus a MutationObserver for child changes decide when it carries `role="group"`, `tabIndex={ 0 }` and a label. Core Modal does this for `.components-modal__content`, but the drawer zeroes that element's padding and scrolls in the body, so core's own handling never engages. The role is load-bearing for the label: a bare div is `role=generic`, where an author name is prohibited.
* `Drawer.Content` carries its `padding` as a CSS custom property so the stylesheet's seam rules (content after header drops its top padding, last content drops its bottom) can win; spacing rhythm is 24px inline / 16px toward the chrome, from `@wordpress/base-styles` tokens. Footer stacking is `:has(> :nth-child(3))`. The frame rounds its free edge and clips with `overflow: hidden`; popovers inside are `position: fixed` with no transformed ancestor, since the slide-in deliberately carries no `forwards` fill.
* `Drawer.Content` passes its children to `VStack` through a wrapper. `getValidChildren` keeps only elements, a lone string aside, so bare text beside an element would otherwise be dropped without warning; a run of consecutive non-elements shares one wrapper so a sentence stays one row. `requestCloseRef` is mirrored in a layout effect rather than a passive one, so a pointer release landing between commit and paint reads the current close funnel rather than the previous one.
* Known trade-offs, deliberate: close requests are not latched beyond the `isOpen` guard, matching core Modal's own per-event `onRequestClose` contract; `Content` sections must be direct children of `Root` (documented in the README, fragments skip the grouping); `Drawer.Divider` is the one part with no `@wordpress/ui` counterpart, so a future swap replaces it with a bare `<hr>` or core's `Divider`.
* The browser pass in an isolated env (43 scripted geometry, behavior and accessible-name checks, plus the flows in the screenshots) covers the geometry and the screenshots above. It was re-run at `1de7f75` and covers the close and focus changes on this branch.

</details>


thomasguillot:
Reviewed the full branch against `main` at `1de7f75`.

The close model is the part I most wanted to put under test, and it holds. A dirty panel refuses core's close request while a clean one takes it, confirmed under trusted input as well as synthetic events. The confirmation renders above an unmoved panel on all four close routes, with cancel, Escape and discard each landing focus where they should and leaving nothing stranded, and focus returns to the opener on every route. The exit-animation timers and listeners, the observer lifecycle in `body.tsx`, the title-registration stack, and the Escape ordering between the nested confirmation and the frame handler all came back clean. Geometry holds at every size including the square-off below 600px, and reduced motion was exercised under a real `prefers-reduced-motion: reduce` context rather than a stub. TypeScript, ESLint and Stylelint are clean, and the package's 215 tests pass.

One correction to the description: the collapsible still says the browser pass predates the close and focus changes on this branch. It was re-run at `1de7f75` and covers them.

**Verdict: approve with comments. Release risk: ELEVATED**, on one factor. The drawer is greenfield and unconsumed, so a defect there costs the migration PR rather than production. But the same branch changes `Button` and `ConfirmDialog`, which every wizard in the workspace imports, and no consumer of either was loaded against the new build, so those two rest on their Jest suites alone. A spot check in one wizard before this lands would close that. `newspack-multibranded-site` is the interesting one, since it resolves react-router v6 against the package's v5, which is the crash the new guards address.

Nothing below blocks a merge. The five suggestions are worth a pass first, because all four migrating panels inherit whatever ships here.

### 🟡 Non-blocking suggestions

**`drawer/root.tsx:322` — the in-panel popover slot is clipped by the frame's overflow.**
The README promises that moving the slot inside the panel keeps dropdowns and pickers inside the drawer visible, but core `Popover` positions with floating-ui's default `absolute` strategy and its containing block is `.components-modal__screen-overlay`. `.components-modal__frame` sits between the two and carries `overflow: hidden`, from core and again at `style.scss:35`, so an absolutely positioned descendant is clipped at the panel box. A `DateTimePicker` at roughly 320px, or a `ColorPicker`, inside a `small` (280px) or `medium` (350px) drawer loses its edges once the 24px section padding comes off the anchor width. Either scope the `overflow: hidden` to what needs it, which is the rounded corners during the slide, or narrow the README's promise to "visible within the panel box".

**`drawer/root.tsx:98` — the drawer's own confirmation dialog has no accessible name.**
`useConfirmDialog` is called with `hideTitle: true` and no `title`. That flows through `ConfirmDialog`, which forces `__experimentalHideHeader={ false }`, into core `Modal`, where `headingId` resolves to `undefined` and `contentLabel` is unset, so the confirmation frame carries neither `aria-label` nor `aria-labelledby` and a screen reader announces "dialog" with no name. That sits oddly next to the Root's own development warning when the panel is unnamed. `useConfirmDialog` already takes `title`, and `newspack-modal--hide-title` hides the header with `display: none`, which does not defeat `aria-labelledby`, so a translated default such as `__( 'Unsaved changes', 'newspack-plugin' )` would name it with no visual change. `useUnsavedChangesDialog` has the same gap, so it is shared rather than a regression here, but this adds a new instance of it.

**`drawer/style.scss:108` — the scrollable body takes a tab stop with no focus style.**
`body.tsx:52` sets `tabIndex={ 0 }` on `.newspack-drawer__body` whenever it overflows, but the stylesheet gives that element no `:focus-visible` rule, so it falls back to the browser default ring. Both references style this exact affordance: core Modal draws an inset accent ring on `.components-modal__content.is-scrollable:focus-visible`, and the `@wordpress/ui` Drawer sets a negative `outline-offset` on its content. Since this component deliberately moves the scroll container off core's element onto its own, it also picks up the job of styling that element's focus.

**`drawer/style.scss:136` — the last-section padding reset is not conditional on a footer.**
The top-edge rule one line above is guarded on a header existing, so a header-less drawer keeps its 24px top padding. The bottom-edge rule has no equivalent guard, so it fires whether or not a `Drawer.Footer` is rendered. The README says every slot is optional, so a footer-less drawer, a read-only detail panel closed from the header icon for instance, renders its last section flush to the bottom with no footer chrome supplying the 16px seam the rule assumes. `:last-child` on the body cannot express the condition, since `Popover.Slot` always renders a final sibling in the children container, but `.newspack-drawer__body:has( ~ .newspack-drawer__footer ) .newspack-drawer__content:last-child` can, and `:has()` is already in use at line 152.

**`drawer/style.scss:97` — the title has no type token, and a weight that degrades to regular.**
`.newspack-drawer__title` is `font-size: inherit` at `wp-vars.$font-weight-medium`, which is `499` in base-styles, chosen so non-variable font stacks fall back to 400 rather than 600. In wp-admin the panel heading can therefore render at 13px/400, the same size and weight as the body copy beneath it, with no visual step. Both references give the heading its own rung: core Modal's header heading is 20px/600, and the `@wordpress/ui` Drawer title renders `Text variant="heading-xl"`. `inherit` also leaves the title's size at the mercy of whatever the host page sets on an ancestor. Pinning an explicit base-styles size token fixes both; the exact rung is a design call.

<details>
<summary><b>🟢 Non-blocking nits (5)</b></summary>

**`button/index.tsx:28` — the href scheme filter stops at `javascript:`.**
`isJavascriptHref` strips control characters before matching, which is the part that usually gets missed, but it rejects only one of the three schemes normally filtered on an href sink. `data:` and `vbscript:` reach both the anchor and the new `window.location.href` path. The practical window is narrow: browsers block top-level `data:text/html` navigation, `vbscript:` is dead, and every `href` at the live combined `href` + `onClick` call sites is an internally constructed admin URL. Still, this function is now the package's one scheme allowlist for a prop that widely-consumed wizards populate, and widening the pattern to `/^(javascript|data|vbscript):/` costs nothing.

**`wizards/componentsDemo/index.js:104` — the demo's Save closes the drawer without clearing the dirty flag.**
`close` only flips `drawerShown`, and it is wired to Save while the "Unsaved changes" checkbox stays ticked. Tick it, open the drawer, press Save: the panel closes with no confirmation and reopens still dirty, so the card reads as "Save silently discards unsaved changes" rather than the pattern the README prescribes, which is to leave `closes` off a Save and close by flipping `isOpen` once the work succeeds. `this.setState( { drawerShown: false, drawerIsDirty: false } )` would model the documented contract. Worth fixing because this card is the reference the four migrating panels get read against.

**`drawer/context.ts:14` — the doc comment describes an unregister guard the code does not have.**
The comment reads "`forId` unregisters only while that title is still the registered one", which describes a guard like `if ( titleRef.current?.id === forId )`. The implementation at `root.tsx:65-67` instead filters the entry with that id out of the stack wherever it sits, which is what `title.tsx:31` needs: a second `Drawer.Title` unmounting from the middle of the stack must still leave, and the test at `index.test.js:114` depends on that. The title stack is one of the subtler mechanisms here, so a comment that mis-states it invites a later reader to restore the guard and break it.

**`button/index.tsx:23` — `onClick`'s event parameter is typed optional though it is always supplied.**
`onClick?: ( event?: React.MouseEvent< HTMLElement > ) => void` marks the event itself optional, and `types.ts:79` mirrors it. No call path omits it: core `Button` invokes the handler with the event, `action.tsx:47-48` forwards it, and the combined `href` + `onClick` wrapper passes it through. The only effect is that a consumer who wants the event writes `event?.preventDefault()` or asserts non-null. Dropping the `?` on the parameter is safe, since a function declared with fewer parameters stays assignable in TypeScript, which is how the existing zero-argument handlers across the wizards keep type-checking.

**`drawer/style.scss:28` — the panel loses its edge under forced colors.**
The panel is separated from the page by core Modal's `box-shadow` and the scrim's `rgba( 0, 0, 0, 0.35 )` alone. Under Windows High Contrast, with `forced-colors: active`, both flatten, so a full-height panel flush to the viewport edge has no boundary against the content behind it. The `@wordpress/ui` Drawer handles this with `@media ( forced-colors: active ) { border-color: CanvasText }` on its popup. The Newspack system has no written position on forced colors today, only `prefers-contrast: more` on form inputs, so this is not a rule violation, but a one-line border would close a gap the component this one mirrors already closes.

</details>

### One thing to carry into the migration PR

`ConfirmDialog` now has two independent ways to raise itself on one instance: the `isOpen` prop, driven by `!! pendingAction`, and the navigation blocker, which calls `setShowDialog( true )` without touching `pendingAction`. The new render-phase sync at `confirm-dialog/index.tsx:134` makes `isOpen` authoritative in both directions, so a false transition now closes a dialog the blocker raised. That is safe on every current caller, because none of their confirm callbacks performs an in-app route change while the guard is armed.

The four panels migrating to the drawer are the first callers likely to combine a `when` guard with a confirm callback that does change route. If one does, the guard catches that route change inside the confirm handler and re-raises the dialog while the pending action clears in the same tick, so the dialog closes and the navigation the user just approved is dropped with no visible signal. Keeping those confirm callbacks free of in-app navigation, or clearing the dirty state before navigating so the guard is disarmed by the time the route changes, avoids it. A test that confirms out of a dialog whose callback pushes a route would pin it down.

One follow-up that belongs elsewhere rather than here: the `opacity: 1` override at `drawer/style.scss:195` exists only to undo `.modal-open .components-popover__content { opacity: 0 }` at `modal/style.scss:99`. That source rule fires for any core modal, not only `.newspack-modal`, so it blanks popovers for consumers that never opted in. Re-scoping it there would remove the need for this override and close the same gap for the next component.

thomasguillot:
Addressed two of the suggestions and closed the gap behind the risk tier. Pushed as `68dda4f` and `7d2bf0a`.

**The confirmation has a name now.** `Drawer.Root` passes a `title` through `useConfirmDialog` alongside the existing `hideTitle: true`. Chrome's accessibility tree reports the dialog as "Unsaved changes" where it announced a bare "dialog" before. The heading renders into a `display: none` header, so the name comes from `aria-labelledby` with no visual change. I checked that in the browser rather than assuming it, since a hidden referenced node is the part of the naming spec browsers sometimes differ on. There's a test for the name and for the heading staying hidden.

**The last section only drops its bottom padding when a footer follows it**, matching the header guard one rule above:

```scss
.newspack-drawer__body:has(~ .newspack-drawer__footer) .newspack-drawer__content:last-child
```

Verified both branches by detaching and restoring the footer: 0px with one, 24px without. The README said the last section always drops that padding, so that paragraph is corrected.

**The factor behind the ELEVATED tier is closed.** It rested on `Button` and `ConfirmDialog` changing with no consumer loaded against the new build. The Audience wizard now runs on it: no console errors from our code, no spurious dialogs on mount, and a platform change round-tripping through the REST save. Alongside that I traced every Newspack `Button` render, 82 importing files plus 14 inside the package, through `ActionCard`, `SectionHeader`, wizard header actions, `PluginToggle`, `Drawer.Action` and `withWizardScreen`. Six sites pass `href` and `onClick` together: two never get a truthy `onClick`, three wrapper paths are mutually exclusive by construction, and the four handlers that now receive a `MouseEvent` are all zero-parameter arrows that discard it. No existing caller changes behaviour. That puts release risk at LOW on my read, with the caveat that shared surfaces still change and are now covered by runtime and call-site evidence rather than by their test suites alone.

One thing worth recording from that trace: the `javascript:` filter is load-bearing rather than defensive. `audience/views/integrations/settings-field.js:50` and `:56` put an integration-supplied `oauth_url` and `disconnect_url` straight into an anchor inside a `manage_options` origin, and a dozen more hrefs arrive from REST responses or PHP-localized globals. Only `PersonProfile.jsx:123` validated its own. That makes widening the pattern to `data:` and `vbscript:` worth doing, though I have left it off this branch.

The other three suggestions are untouched by choice: the clipped popover slot, the body's focus style, and the title's type token. All four migrating panels inherit whatever ships here, so they are worth a call before merge.

216 tests pass, TypeScript and lint clean.

thomasguillot:
Second pass over the branch at `7d2bf0ae8`, weighted toward the two commits that landed since the last review.

Most of it holds. The close funnel, the exit-animation timers and listeners, the observer lifecycle in `body.tsx`, the title-registration stack, and the Escape ordering between the nested confirmation and the frame handler all came back clean again, and the package suite passes at that commit. The accessible-name commit does what it claims: core computes `headingId = title ? … : aria.labelledby`, so before it the confirmation frame carried neither `aria-label` nor `aria-labelledby`, and nothing else keyed on `newspack-modal--hide-title` is disturbed by a populated header. The `--size-small` max-width pairing and the content padding both key off `hideTitle`, which is unchanged, and the header stays `display: none`, so focus order is untouched.

Two things are worth fixing first. Both are one-liners, and everything shipped here is inherited by the four migrating panels.

**Release risk: ELEVATED**, on those two.

### 🟡 `drawer/style.scss:136` — the footer seam guard reaches further than its own documentation

`~` is the general sibling combinator, so `.newspack-drawer__body:has(~ .newspack-drawer__footer)` matches any body with a footer anywhere later, not the body the footer abuts. `root.tsx` starts a new `Body` at every child that is neither a `Drawer.Content` nor a `Drawer.Divider`, a split the README documents, so `<Content/><Notice/><Content/><Footer/>` groups into two bodies and both match. The first body's last section then drops its 24px bottom padding against the interleaved child, which brings no chrome padding of its own, and the rhythm breaks in the middle of the panel. Deleting the footer brings that padding back, so an element at the far end of the drawer silently changes spacing in its middle.

Rendering that arrangement confirms it. The children come out as `header, body1, div, body2, footer, popover-slot`; `:has(~ …)` matches both bodies and `:has(+ …)` matches only the second.

The rule two lines above uses `+` for exactly this reason, and `+` is right in the single-body case too, since the popover slot renders after the footer rather than between it and the body. Specificity is unchanged at (0,4,0) either way.

### 🟡 `hooks/use-unsaved-changes-dialog.tsx:53` — the delegated confirmation is still unnamed

The new commit names the confirmation `Drawer.Root` owns, but the shared `useUnsavedChangesDialog` in the same package still passes `hideTitle: true` with no `title`, so core Modal's `headingId` stays undefined and, with no `contentLabel`, the dialog is announced with no name.

That matters more than the one just fixed. The README tells wizards to hand this hook's `requestConfirm` to `Drawer.Root`, at which point the Root renders no dialog of its own and this is the prompt the user actually gets, so a panel following the documented pattern ends up worse off than one using the built-in. It is live today at `wizards/newsletters/views/settings/index.js:870` and `audience/views/subscription-products/product-form.tsx:169`, and it is the path the four migrating panels will take. The same title beside the existing `hideTitle: true` closes it, and this branch already touches the file.

### 🟢 `drawer/root.tsx:102` — the confirmation title is the one piece of its copy a caller cannot change

`confirmCloseMessage` and `confirmButtonText` are props with defaults precisely so a consumer can reword the built-in dialog, but the accessible name now sitting between them is hardcoded. A drawer that rewords the body to something other than an unsaved-changes question gets a name contradicting its content, and because the name is invisible the mismatch will not be caught in review. A `confirmTitle` prop defaulting to the current string would make the three consistent, and the Root props table already has the shape for it.

### Two follow-ups that belong outside this branch

Three screens build the byte-identical unsaved-changes prompt through `useConfirmDialog` directly rather than through the shared guard: `content-gates/institutions/edit.tsx:177`, `content-gates/edit/content-gifting.tsx:61` and `content-gates/edit/countdown-banner.tsx:49`. Once the shared hook is named, those three keep announcing nothing, so the family reads as inconsistent to anyone auditing it, and the next screen copied from `institutions/edit.tsx` inherits the unnamed variant.

The existing coverage for `ConfirmDialog`'s two raise paths holds `isOpen` steady at false while the navigation blocker fires. A case where `isOpen` goes true to false in the same commit that the blocker raises the dialog would pin the combination the delegated `requestConfirm` recipe leads a consumer into, where a prompt withdrawn by the render-time sync takes the pending navigation with it. No current caller reaches it: all four guards run a full page navigation rather than an in-app route change.

### Still open from the last pass

The nine remaining items are unchanged and deliberate: the clipped popover slot, the body's focus style, the title's type token, the scheme filter, the demo Save, the `context.ts` comment, the optional event parameter, the popover opacity override, and the forced-colors border.

One update on the scheme filter. Tracing the `Button` call sites showed it is load-bearing rather than defensive: `audience/views/integrations/settings-field.js:50` and `:56` put an integration-supplied `oauth_url` and `disconnect_url` straight into an anchor inside a `manage_options` origin.

Fixes for the two suggestions above are written and will follow in the next push.

thomasguillot:
Both suggestions from the last pass are in, as `9241dba4a` and `d47531fb7`.

**The footer seam guard now uses `+`.** `~` matched every body preceding the footer rather than the one abutting it, so in the documented split-run arrangement the first body's last section lost its bottom padding against an interleaved child. There is a test alongside it pinning that the footer stays the immediate next sibling of the last body, so a future change to the grouping in `root.tsx` fails the build instead of quietly changing spacing.

Checked in the browser after rebuilding: the single-body case still resolves to `0px` with a footer and `24px` without, the footer is the body's next sibling, and the header guard above is unaffected.

**The shared `useUnsavedChangesDialog` now carries the same accessible name**, so the delegation path the README recommends is no longer the unnamed one. It reaches four live wizards; the header carrying the name stays `display: none`, so there is no visual or focus-order change. Confirmed after the rebuild that the dialog still resolves to "Unsaved changes" with its header hidden.

217 tests pass, TypeScript and stylelint clean.

Left open deliberately: the confirmation title is still hardcoded while `confirmCloseMessage` and `confirmButtonText` beside it are props, which is an API call rather than a defect. The three Content Gates screens that build the same prompt through `useConfirmDialog` directly, and the missing test for `isOpen` going true to false while the navigation blocker raises, both belong outside this branch.

thomasguillot:
Fifth pass, at `21a3d7044`. **Zero findings.** Every reviewer came back clean, and everything raised across the earlier passes is now either fixed on the branch or recorded as a deliberate deferral.

**Release risk: LOW.**

### What landed since the last review

- `9241dba4a` — the footer seam guard narrowed to `:has(+ .newspack-drawer__footer)`, with a test pinning that the footer stays the immediate next sibling of the last body.
- `d47531fb7` — the shared `useUnsavedChangesDialog` gained the same accessible name as the drawer's own confirmation, so the delegation path the README recommends is no longer the unnamed one.
- `4fb8e138a` — tests for the hidden-title naming mechanism in both directions, and for a withdrawn blocked-navigation prompt.
- `2ef29be61` and `595ff549b` — a withdrawn prompt now releases the navigation the blocker was holding. Previously `handleOnConfirm` and `handleOnCancel` both cleared it but the `isOpen` withdrawal did not, so confirming a later unrelated prompt on the same instance could replay a destination the user had abandoned. The release runs from a committed effect rather than during render, so a discarded render cannot drop a live one.

Both behavioural fixes were negative-checked rather than assumed: each was confirmed to fail with the fix reverted and pass with it restored.

### What was verified

Exercised in a live wp-admin against a rebuilt bundle. The confirmation resolves to the accessible name "Unsaved changes" with its header `display: none`. The single-body drawer resolves to `0px` bottom padding with a footer and `24px` without, and the footer is the body's next element sibling. The selector over-match the `+` fix addresses was reproduced before fixing it: the documented split-run arrangement renders `header, body1, div, body2, footer, popover-slot`, where `~` matched both bodies and `+` matches only the one abutting the footer.

The shared surfaces were traced to every live call site rather than sampled. Six sites pass `href` and `onClick` to `Button` together; two never get a truthy `onClick`, three wrapper paths are mutually exclusive by construction, and the four handlers that now receive a `MouseEvent` are zero-parameter arrows that discard it. No caller's behaviour changes.

220 tests pass. TypeScript, ESLint and stylelint clean.

### Deliberately left

Nine items from the earlier passes stand unchanged and on the record: the clipped popover slot, the body's focus style, the title's type token, the scheme filter, the demo Save, the `context.ts` comment, the optional event parameter, the popover opacity override, and the forced-colors border. Plus the hardcoded confirmation title, which is an API choice rather than a defect.

Two follow-ups belong outside this branch: three Content Gates screens build the same prompt through `useConfirmDialog` directly and stay unnamed, and the `.modal-open` popover rule in `modal/style.scss` would be better re-scoped than counteracted.

One optional addition, for whenever the panels migrate: the new `history` null-guards mean that asking for navigation blocking where there is no router now fails quietly instead of throwing. A development-only warning there would keep that wiring mistake loud, in the same style as the drawer's existing warnings. Nothing in the monorepo reaches it today.

thomasguillot:
Sixth pass, at `687cf94f4`. The comment-removal commit is clean, and nothing outstanding was found in it. This pass did surface three things the earlier ones missed, one of them a real layout defect.

**Release risk: LOW.** The drawer has no consumers, so nothing here can regress on merge. The cost of all three lands in the migration PR instead.

### 🟡 `drawer/style.scss:140` — the footer is only pinned when a content section is present

`.newspack-drawer__body` carries the sole `flex: 1 1 auto` in a column flex container whose header and footer are both `flex: 0 0 auto`. The Root only builds a body around a run of `Drawer.Content` and `Drawer.Divider` children, so any arrangement without one leaves nothing to grow and the footer rides up under the header.

Measured in the browser on the demo panel, with the body detached to model it:

| | footer top | header bottom | footer bottom | panel bottom |
| --- | --- | --- | --- | --- |
| with a body | | | 766 | 766 |
| without | 64 | 64 | 140 | 766 |

So the footer sits flush under the header, 626px clear of the bottom of the panel. The README says every slot is optional and describes the footer as pinned.

The empty case is the easy one to dismiss. The reachable one is a direct child that is neither `Drawer.Content` nor `Drawer.Divider`: the Root passes it through ungrouped, so it gets neither the grow rule nor the scroll container. The README already documents that arrangement as supported. The Content Gate save panel is the concrete case, since it wraps everything in a `<form>` today and supplies `flex: 1 1 auto` on that form itself (`save-panel/style.scss:58`) — handed to `Drawer.Root` as-is, it would lose it. Once such a form is taller than the viewport the scroll moves up to the wrapper core Modal already sets to `overflow: auto`, so the header and footer scroll away with the content and the save button is only reachable at the end of the form.

Growing the last non-chrome child, or pinning only the header and footer and letting whatever sits between them fill the gap, covers both the empty case and the wrapper case.

### 🟡 `confirm-dialog/README.md:94` — `cancelConfirm` is undocumented on the hooks that expose it

The branch adds `cancelConfirm` to the public return of both `useConfirmDialog` and `useUnsavedChangesDialog`, but the Returns table lists only `confirmDialog` and `requestConfirm`, and neither hook's JSDoc mentions it. The bare `() => void` does not convey that it withdraws a pending prompt without invoking the stored callback, which is the whole point of it.

That matters more than a usual doc gap because `drawer/README.md:344` tells consumers they must call it, so a reader following the drawer's guidance lands on a hook whose own documentation does not acknowledge the API.

### 🟢 `drawer/README.md:350` — the delegation example omits the call the paragraph above requires

The paragraph at 341-347 states that a delegated dialog is the caller's to dismiss and to call `cancelConfirm` when closing the panel out from under it. The example directly beneath destructures only `{ confirmDialog, requestConfirm }` and wires `onRequestClose={ close }` with nothing withdrawing the prompt. It is the only occurrence of the identifier in the file, and it is prose.

Someone copying the snippet, which is the natural way to read a block placed under the instruction, implements the delegation without the dismissal. The failure is a stranded prompt: it stays mounted and focus-trapped over the wizard, asking the user to discard changes to a panel that is no longer there. Answering it calls the panel's close handler again, which for every current shape only re-sets a flag, so the cost is one confusing modal and one click rather than lost work. Threading `cancelConfirm` through the example would make it demonstrate the rule instead of contradicting it.

### What came back clean

The comment-removal commit changed no executable code and lost nothing load-bearing. The `context.ts` JSDoc it deleted had mis-described the implementation, and the contract now lives only where it is accurate, in `root.tsx`'s `registerTitle` and `title.tsx`'s own note. The trims kept their reason clauses, and one of them corrected a stale directional reference.

The stylesheet is unchanged, still comment-free and token-clean. Every class name, token reference and icon import survived the edits. The close funnel, exit animation, observer lifecycle, title-registration stack and the whole `Button` call-site surface all held again. 220 tests pass; TypeScript, ESLint and stylelint clean.

packages/components/src/confirm-dialog/index.tsx
```diff
@@ -82,18 +82,20 @@ function ConfirmDialog(
...
-			history.replace( history.location );
-		} finally {
-			bypassBlock.current = false;
+		if ( history ) {
+			bypassBlock.current = true;
+			try {
+				history.replace( history.location );
+			} finally {
+				bypassBlock.current = false;
+			}
```
miguelpeixe: Non-blocking: cancelling a drawer confirmation can raise the wizard's own unsaved-changes prompt.

The cancel path re-syncs the URL on every cancel, but only the `history.block` handler can desync it, and `Drawer.Root` deliberately passes no when. Since `bypassBlock` is per-instance and react-router v5 keeps a single prompt slot, that replace is caught by whichever other dialog on the screen did install a blocker.

Gating this re-sync on `pendingNavigation.current` having actually been armed confines it to the case it was written for. As it stands, a dialog that never installed a blocker still pushes a history transition on cancel. With a single prompt slot on the history object, that transition is caught by whichever other dialog on the screen did install one. Capturing `const hadPending = !! pendingNavigation.current;` before the reset keeps the existing test green, since it cancels a genuinely blocked navigation.

miguelpeixe (approved):

packages/components/src/confirm-dialog/index.tsx
thomasguillot: Gated on `hadPendingNavigation` in `53cdc62a5`.

New test renders a `when` dialog alongside an `isOpen` one, which is `Drawer.Root`'s exact shape, and cancels the second: the first stays down and `history.replace` is never called. It fails with the gate removed. Both existing cancel tests stayed green, since each cancels a genuinely blocked navigation.

Thanks for catching it.

PR #811, opened 2026-08-05
ci(release): deploy the released version to wp.org, not the trigger sha

chickenn00dle:
## The problem

WP.org deploys have been publishing the *previous* release, and passing green while doing it. newspack-newsletters 3.38.2 was tagged and released on GitHub, but wordpress.org is still serving **3.38.1** ([API check](https://api.wordpress.org/plugins/info/1.0/newspack-newsletters.json)).

The deploy job resolves the version from the working tree:

```yaml
- name: Resolve released version
  run: echo "version=$(node -p "require('./<plugin-path>/package.json').version")" >> "$GITHUB_OUTPUT"
```

…but `actions/checkout` has no `ref:`, so it gets `github.sha` — **the commit that triggered the run**. The version bump lands *after* that, in the release job's `Sync package.json versions` step (`chore(release): sync package.json versions [skip ci]`). `needs: release` makes the deploy job wait, but waiting doesn't re-checkout.

In [run 31011773447](https://github.com/Automattic/newspack-workspace/actions/runs/31011773447/job/92334574877) the trigger SHA is `94e18e2` (#807), where `package.json` still said `3.38.1`. The log confirms it built the wrong thing:

```
> newspack-newsletters@3.38.1 release:archive
…
Tag 3.38.1 already exists on WordPress.org. No deployment needed.
```

That short-circuit is the second half of the bug: 3.38.1 *was* already published, so the job exited 0 having deployed nothing. Every deploy re-confirmed the prior release and reported success. It's structural, not a one-off — the `[skip ci]` sync commit never triggers another run, so the trigger SHA is permanently one release behind.

## The fix

**1. Check out the branch tip** (`_release-wporg.yml`) — `ref: ${{ github.ref_name }}`. `needs: release` guarantees the sync commit is pushed by then, so the tip is the released commit.

**2. Gate on what actually released.** The release job now exposes the `<pkg>@<version>` tags it created as a job output (same before/after tag diff `notify-release.sh` already uses for Slack), and each deploy job checks its resolved version against it:

| Situation | Result |
| --- | --- |
| Tag matches resolved version | Deploy |
| Tag at a *different* version | **Fail** — the checkout is stale |
| No tag for this plugin | Skip cleanly, before install + build |
| Manual dispatch (`require-release: false`) | Deploy the resolved version |

The middle row is the important one: it turns exactly this bug into a red job instead of a green no-op. `require-release` is a separate input from `released-tags` so an empty list reads as "nothing released" rather than "not supplied".

Moving the version resolution ahead of the build also means a plugin that didn't release skips ~2 minutes of install+build instead of burning it to reach an already-exists check.

**3. `release-wporg.sh`** now emits `::warning::` on the already-published path instead of a plain log line — expected on a re-run, worth an annotation otherwise.

**4. New `release-wporg-manual.yml`** — `workflow_dispatch` escape hatch to deploy a plugin's current branch version when a release didn't reach SVN.

## Verification

Workflow YAML parses; `bash -n` clean on the script. The gate logic was extracted and run against the real tag list from run 31011773447:

| Case | `require-release` | Tags | Resolved | Result |
| --- | --- | --- | --- | --- |
| The bug | true | `newspack-newsletters@3.38.2` | 3.38.1 | ❌ fails with a stale-checkout error |
| Fixed | true | `newspack-newsletters@3.38.2` | 3.38.2 | ✅ `deploy=true` |
| Plugin not in release | true | newsletters only | 0.7.5 (scaip) | skip |
| No-op release run | true | *(empty)* | 3.38.2 | skip |
| Manual dispatch | false | *(empty)* | 3.38.2 | ✅ `deploy=true` |
| Prefix collision | true | `newspack-newsletters-pro@1.0.0` | 3.38.2 | skip (no false match) |

Real confirmation only comes from the next stable release run, since the failure mode is specific to how Actions resolves `github.sha`.

> [!IMPORTANT]
> **3.38.2 still isn't on wordpress.org** and merging this won't retroactively publish it. Once this lands on `release` and post-release sync carries it to `main`, the "Run workflow" button appears and `Deploy to WordPress.org (manual)` can push 3.38.2 against the `release` branch. (GitHub only surfaces `workflow_dispatch` for workflows present on the default branch.) Otherwise the next stable release supersedes it with 3.38.3.

Targeting `release` because that's the only branch the deploy runs on, and post-release sync carries it back to `main`.


.github/workflows/release.yml
```diff
@@ -144,6 +149,23 @@ jobs:
...
+      # before/after tag diff notify-release.sh uses for its Slack summary,
+      # surfaced as a job output so the WP.org deploy jobs can gate on it:
+      # without it they cannot tell a plugin that had no release from one whose
+      # deploy silently resolved a stale version. Always runs, so a skipped
+      # release yields an empty list rather than a missing output.
+      - name: Collect released tags
+        id: released
+        run: |
+          new_tags=$(comm -13 <(sort "$RUNNER_TEMP/tags-before.txt") <(git tag | sort) || true)
+          echo "Tags created by this run:"
```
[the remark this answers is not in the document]
chickenn00dle: Good catch — fixed in 5ac838e.

You're right that this is the one place a swallowed error is most damaging: the deploy jobs read an empty list as "this plugin had no release" and skip, so a masked `comm`/`sort` failure would silently skip every WP.org deploy of a real release — precisely the failure this gating exists to prevent.

Dropped the `|| true`, added `set -euo pipefail`, and materialised both sides as files instead of process substitutions, since bash discards a process substitution's exit status (so `<(git tag | sort)` failing would still have yielded a silent empty diff).

Verified against a scratch repo:

| Case | Exit | Behaviour |
| --- | --- | --- |
| Two new tags | 0 | both emitted |
| No new tags (no-op run) | 0 | empty list — correctly *not* an error |
| Snapshot file missing | 2 | fails the job |
| `git tag` fails mid-pipe | 3 | fails the job (pipefail) |

.github/workflows/_release-wporg.yml
```diff
@@ -38,36 +56,91 @@ jobs:
...
+          # already published (the deploy script's already-exists short-circuit).
+          #
+          # `needs: release` in the caller guarantees the bump is pushed before
+          # this checkout runs, so the branch tip is the released commit.
+          ref: ${{ github.ref_name }}
+      - name: Resolve released version
+        # Read before the build: the gate below can then skip a no-op run
+        # without paying for install + build.
+        id: ver
+        run: echo "version=$(node -p "require('./${{ inputs.plugin-path }}/package.json').version")" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
chickenn00dle: I don't think this one holds — leaving the line as-is.

Nested double quotes inside `$( )` don't terminate the outer string: command substitution is re-parsed in a fresh quoting context, so `echo "a$(cmd "arg")b"` is valid POSIX/bash. Ran the exact line locally:

```
$ GITHUB_OUTPUT=/dev/stdout bash -c 'echo "version=$(node -p "require('"'"'./plugins/newspack-newsletters/package.json'"'"').version")" >> "$GITHUB_OUTPUT"'
version=3.38.2
exit=0
```

It's also unchanged, pre-existing code — this PR only moves it earlier in the step order — and it has been running in production: the run that prompted this PR resolved `WP_ORG_PLUGIN_VERSION: 3.38.1` through it. The bug there was *which commit* it read, not the quoting.

.github/workflows/release-wporg-manual.yml
```diff
@@ -0,0 +1,41 @@
...
+        description: 'Plugin to deploy (its package.json version on the selected branch).'
+        required: true
+        type: choice
+        options:
+          - newspack-newsletters
+          - super-cool-ad-inserter
+          - republication-tracker-tool
+
+jobs:
+  deploy:
```
adekbadek, line 51: **Blocker:** Manual deploy has no branch guard, and the dispatch picker pre-selects `main` — The header comment asks the operator to "run it against `release` unless you know otherwise", but nothing enforces it — and the default is the unsafe one. GitHub's Run-workflow picker pre-selects the repository's default branch, which is `main`; that is also *why* the comment notes the file must reach `main` for the button to appear at all.

Dispatching with the default selection makes `_release-wporg.yml` check out `main`. Three things then line up so that nothing stops it:

- `require-release` defaults to `false`, so the version gate short-circuits to `deploy=true` without comparing anything.
- `main` carries the same plugin versions as `release` (post-release sync merges `release` into the default branch), so even a version comparison would pass. Right now: newspack-newsletters is 3.38.6 on both branches, while `main` is 98 commits ahead of `release`.
- All three plugins ship `Stable tag: trunk`, and `release-wporg.sh` does `rm -rf trunk` then copies the build into it. Trunk is what wordpress.org actually serves, so there is no tag-promotion step in between — it reaches every installed site on the next update check.

The one remaining guard is the `svn ls tags/$SVN_TAG` short-circuit, and it is absent in exactly the scenario this workflow exists for: a version tagged on GitHub that never reached SVN. That is the case described in this PR's own summary, so the intended first use of the escape hatch is also the run where nothing catches a default-branch dispatch.

**To reproduce** (branch state verified directly; the publish step is reasoned, not run against production SVN):
1. `gh repo view Automattic/newspack-workspace --json defaultBranchRef` → `main`.
2. `git rev-list --count origin/release..origin/main` → 98.
3. Compare `plugins/newspack-newsletters/package.json` on both branches → 3.38.6 on each.
4. `grep -i '^Stable tag:' plugins/newspack-newsletters/readme.txt` → `trunk`.
5. Dispatch `Deploy to WordPress.org (manual)` leaving the branch picker at its default.

A job-level `if: github.ref_name == 'release'` turns the most likely operator mistake into a clean failure. An explicit `ref` input defaulting to `release` (checked out in place of `github.ref_name`) also works and keeps the deliberate override available. Either one additionally closes the arbitrary-branch path to the SVN credentials.

.github/scripts/release-wporg.sh
```diff
@@ -40,9 +40,13 @@ fi
 
 mkdir -p "$SVN_REPO_LOCAL_PATH" && cd "$SVN_REPO_LOCAL_PATH"
 
-# Skip if this version is already published.
+# Skip if this version is already published. A legitimate no-op on a re-run,
```
adekbadek: **Suggestion:** Comments narrate the outage rather than the invariant — Offering this as a preference rather than a rule — `AGENTS.md` sets no comment convention. The Newspack editorial standard for in-repo prose is to describe intent: why something exists, what it guarantees, what breaks if you change it, rather than restating the code or recounting how it got here. Two comments lean on the incident instead of the invariant:

- This one: "this branch used to hide a run that resolved the previous release's version and 'succeeded' without publishing anything." What a reader needs is that this path means nothing was published, that it is expected only on a re-run, and that it is the last defence behind the workflow gate.
- `_release-wporg.yml:59-70`: the first paragraph (why `github.sha` is the wrong commit) and the last (`needs: release` is what guarantees the bump is pushed) are exactly the intent that earns its lines. The middle paragraph retells the outage; cutting it leaves a comment that explains why the `ref:` exists and what breaks if a caller drops `needs: release`.

For contrast, the "empty list is meaningful" paragraph at `release.yml:152-164` is what these should all look like — it explains an invariant that is genuinely invisible from the code and justifies the specific implementation choices against it.

.github/workflows/_release-wporg.yml
```diff
@@ -38,36 +56,91 @@ jobs:
...
+          # job runs, the release job has already pushed the version bump
+          # ("chore(release): sync package.json versions [skip ci]", see
+          # release.yml), so the trigger commit still carries the PREVIOUS
+          # version. Resolving the version from it deployed the last release
+          # over and over -- and passed silently, because that version was
+          # already published (the deploy script's already-exists short-circuit).
+          #
+          # `needs: release` in the caller guarantees the bump is pushed before
+          # this checkout runs, so the branch tip is the released commit.
+          ref: ${{ github.ref_name }}
```
adekbadek: **Suggestion:** Pin the deploy checkout to the released commit, not the mutable branch name — The comment claims `needs: release` guarantees the branch tip is the released commit. The first half holds; the second only holds if nothing else lands on `release` in between. `concurrency: release-${{ github.ref }}` with `cancel-in-progress: false` serializes *runs*, not *pushes* — a merge into `release` while a release run is in flight moves the tip immediately, and the in-flight deploy then builds that newer tree.

The version gate cannot catch this, because it only compares version numbers: `package.json` still reads the released version and the tag still matches, so a code-only commit passes cleanly while the published artifact contains work that is not in the corresponding GitHub release. With `Stable tag: trunk` that ships immediately.

The release tag itself is not usable as the ref — multi-semantic-release deliberately keeps `package.json` out of the `@semantic-release/git` prepare commit, which is precisely why the branch tip was chosen. The cheap fix is to have `Sync package.json versions` emit `git rev-parse HEAD` after its push as a second job output alongside `released-tags`, and check that SHA out here. That makes the deploy exact rather than merely fresh, and demotes the version gate to a belt-and-braces check.

**Risk if not addressed:** if a hotfix or promotion merge lands on `release` during the minutes between the release job finishing and the deploy job checking out, wordpress.org receives a build containing unreleased commits, published under the previous version number and with nothing in the run marking it as wrong.

.github/workflows/_release-wporg.yml
adekbadek: **Suggestion:** `Resolve released version` discards node's exit status — `echo "version=$(node -p ...)"` is green whatever node does, because `echo` succeeds. A missing or unparseable `package.json` — a wrong `plugin-path`, a future plugin move — writes `version=` and passes, and the failure surfaces one step later as the gate's "…but the checked-out tree resolves to " with an empty version, blaming a stale checkout for what is really a path error.

The gate step immediately below already does this right; assigning to a variable under the same `set -euo pipefail` here would fail at the real cause. (The step's move ahead of `actions/setup-node` is fine — `ubuntu-latest` ships Node preinstalled — and it correctly buys the ability to skip install and build on a no-op run.)

.github/workflows/_release-wporg.yml
```diff
@@ -38,36 +56,91 @@ jobs:
...
+            echo "deploy=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          released=$(printf '%s\n' "$RELEASED_TAGS" | { grep -E "^${PLUGIN_NAME}@" || true; })
+          if [ -z "$released" ]; then
+            echo "::notice::$PLUGIN_NAME had no release in this run. Nothing to deploy."
+            echo "deploy=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if ! printf '%s\n' "$released" | grep -qxF "${PLUGIN_NAME}@${VERSION}"; then
```
adekbadek, line 151: **Suggestion:** The gate validates package.json; wordpress.org reads the PHP header — The version the gate resolves and the version stamped into the deployed artifact come from two commits made by two mechanisms: `semantic-release-version-bump` writes the plugin header and `@semantic-release/git` commits it, while `package.json` is committed separately by `finalize-package-versions.cjs`. They agree on `release` today — newspack-newsletters is 3.38.6 in both `package.json` and its main plugin file.

But with `Stable tag: trunk`, the plugin header is the only version marker wordpress.org reads (`readme.txt` carries none), so a gate that never inspects it can pass while the artifact is stamped wrong. `release:archive` has already run by deploy time, so asserting the `Version:` header in `release/<plugin>/<main-file>` equals the released tag would close the loop on the value that actually determines what publishes.

.github/workflows/release-wporg-manual.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy to WordPress.org (manual)
+
+# Escape hatch for the WP.org deploy, which normally runs only as part of a
+# release (release.yml). Use it when a published release did not reach the SVN
+# repo -- a failed deploy job, or a run whose deploy resolved the wrong version.
+#
+# It deploys whatever version the SELECTED BRANCH currently carries in the
+# plugin's package.json, so run it against `release` unless you know otherwise.
+# No released-tags are passed, so _release-wporg.yml's gate defers to the SVN
```
adekbadek: **Suggestion:** The comment credits the wrong mechanism for the manual bypass — "No released-tags are passed, so `_release-wporg.yml`'s gate defers to the SVN already-published check" — the gate actually branches on `require-release`, not on `released-tags` being empty. The input description in the reusable workflow goes out of its way to say the two are kept separate *precisely* so an empty list cannot be confused with an unsupplied one, so as written the two comments contradict each other.

The risk is concrete: a future reader who trusts this one could "simplify" the gate by dropping `require-release` and inferring intent from an empty tag list. Passing `require-release: false` explicitly here, and saying so, makes the manual path self-evident and keeps the two comments in agreement.

.github/workflows/release-wporg-manual.yml
```diff
@@ -0,0 +1,41 @@
...
+      plugin:
+        description: 'Plugin to deploy (its package.json version on the selected branch).'
+        required: true
+        type: choice
+        options:
+          - newspack-newsletters
+          - super-cool-ad-inserter
+          - republication-tracker-tool
+
+jobs:
```
adekbadek, line 50: **Suggestion:** The only workflow_dispatch workflow in the repo with no `permissions:` block — `prune-merged-branches.yml`, `publish-baseline-releases.yml` and `sync-legacy.yml` each declare one; this workflow does not, so it takes the repository default. Adding `permissions: contents: read` matches the house pattern and pairs with the job-level grant suggested on the reusable workflow.

Worth considering alongside it: binding the job to a GitHub `environment:` would scope `WP_ORG_USERNAME` / `WP_ORG_PASSWORD` to that environment and allow a required-reviewer gate on the deploy. That is the standard control for a manually-triggered job holding publishing credentials, and it is a defence that does not depend on the branch guard being right.

.github/workflows/release.yml
```diff
@@ -144,6 +149,33 @@ jobs:
...
+      # release, which is the exact failure this gating exists to prevent.
+      # Hence no `|| true`, pipefail on, and both sides materialised as files
+      # rather than process substitutions (whose exit status bash discards).
+      - name: Collect released tags
+        id: released
+        run: |
+          set -euo pipefail
+          sort "$RUNNER_TEMP/tags-before.txt" > "$RUNNER_TEMP/tags-before-sorted.txt"
+          git tag | sort > "$RUNNER_TEMP/tags-after-sorted.txt"
+          new_tags=$(comm -13 "$RUNNER_TEMP/tags-before-sorted.txt" "$RUNNER_TEMP/tags-after-sorted.txt")
```
adekbadek, line 220: **Suggestion:** The tag diff measures tags that appeared, not tags this run created — `comm -13` attributes every tag that shows up in the local clone to this run. semantic-release runs `git fetch --tags` at the start of each invocation, and multi-semantic-release invokes it once per package — so the whole multi-minute Release step is a window in which tags pushed by another concurrent run get pulled in and counted as new.

The concurrency group is per-ref (`release-${{ github.ref }}`), so an `alpha` release runs alongside a `release` one by design, and both tag into the same namespace. The damaging case is an alpha release of newspack-newsletters while stable does not release it: `released` becomes non-empty, the exact-match test finds no `newspack-newsletters@<stable version>` line, and the gate exits 1 — turning the run red *after* npm and GitHub have already published, with an error that misdiagnoses it as a stale checkout.

The direction is safe (it can only cause a false failure, never a wrong deploy), but it converts a cosmetic weakness in the existing tag diff into a build-breaking one now that the deploy gates on it. Filtering both sides with `git tag --merged HEAD` would fix it, since alpha-only tags are not reachable from `release`.

.github/workflows/release-wporg-manual.yml
adekbadek, line 51: **Nit:** No concurrency group on the manual deploy — This workflow exists to be run when a release deploy went wrong, which is exactly the window in which a release-triggered deploy for the same plugin may still be in flight. Two jobs would then run `svn checkout` and `svn ci` against the same plugin concurrently. `concurrency: { group: wporg-${{ inputs.plugin }} }`, shared with the reusable workflow, serializes them for one line.

adekbadek (changes requested):
Release risk: HIGH — one blocker in the new manual workflow; the core fix itself is sound and verified as far as static analysis allows.

The diagnosis is right and the design is the right one: gating on the release job's own tag diff makes "this plugin released" something the deploy job is *told* rather than infers, and resolving the version before toolchain setup means a no-op costs a checkout instead of a full build. I extracted the gate step and ran it over the six cases in the PR body plus seven more; all behave as described, including the prefix-collision and substring-version cases.

The blocker is the escape hatch, not the fix. It inherits `ref: ${{ github.ref_name }}` from a reusable workflow written for the release-branch-only path, while `workflow_dispatch` pre-selects the default branch — and `main` is currently 98 commits ahead of `release` at an identical version. Details and the one-line guard are on that thread.

One cross-cutting pattern worth naming: three of the remaining suggestions are all the same shape — the gate reasons about *version numbers* while the artifact is built from a *tree*, so version equality keeps being used as a proxy for "this is the released code". The SHA-pinning, PHP-header and mutable-ref points are three faces of that one gap.

- `.github/workflows/_release-wporg.yml:52` — **Suggestion:** The deploy job inherits a write-scoped token it never uses — The reusable workflow declares no `permissions:`, so when `release.yml` calls it the job inherits that workflow's top-level grant: `contents: write`, `issues: write`, `pull-requests: write` and `id-token: write`. The job only checks out, builds and pushes to SVN — it never writes to GitHub. It does run `pnpm install --frozen-lockfile` and `composer install`, which execute third-party lifecycle scripts, so a write-scoped `GITHUB_TOKEN` and a mintable OIDC token sit in the environment of the least-trusted step in the pipeline. A job-level `permissions: contents: read` can only reduce the caller's grant, so it is safe to add and covers both callers at once.
- `.github/workflows/_release-wporg.yml:19` — **Nit:** `wporg-slug` has no caller, and the plugin list now lives in two files — `_release-wporg.yml` exposes `wporg-slug` for the diverging-slug case, but neither `release.yml` nor the new manual workflow passes it — and the manual workflow structurally cannot, since one `choice` value feeds both `plugin-name` and the path. The input is unreachable. Meanwhile adding a fourth wp.org plugin now needs edits in two files, and forgetting the `choice` list is silent: you simply have no escape hatch for that plugin on the day you need one. Worth noting `super-cool-ad-inserter` is the case where directory == package == slug is asserted rather than derivable — its main file is `super-cool-ad-inserter-plugin.php` and the legacy repo was `super-cool-ad-inserter-plugin`. Either drop `wporg-slug` as unused or give the manual workflow a slug override so the input has a caller.

.github/workflows/release-wporg-manual.yml
chickenn00dle: Fixed in 18f87bf3, taking the `ref`-input option rather than the job-level guard.

The manual workflow now has a `ref` input defaulting to `release`, passed straight through to the reusable workflow's checkout. The branch picker now only decides which copy of the workflow file runs; `ref` decides what gets published, so a default dispatch builds `release` regardless of what the picker is sitting on.

I went with the input over `if: github.ref_name == 'release'` because a job-level `if` renders as *skipped*, and an escape hatch that silently no-ops during an incident felt like the wrong failure mode. A non-`release` ref still deploys, but the gate annotates the run:

```
::warning::Deploying from 'main', not the release branch. Publishing 3.38.6 to wordpress.org from unreleased code.
```

The deliberate override stays reachable by editing the field, it just isn't the default any more. Ran the gate matrix over it: `ref=release` deploys with a plain notice, `ref=main` and `ref=<sha>` deploy with the warning.

.github/scripts/release-wporg.sh
chickenn00dle: Rewrote both in 18f87bf3.

This one now leads with what the path means — nothing was published — that it's expected on a re-run, and that it's the last guard behind the workflow gate.

The checkout comment lost the middle paragraph and gained the reason a branch name isn't sufficient either, which is the SHA-pinning point from the other thread. So it explains why the `ref` exists and what breaks if a caller drops `needs: release`, without the incident retelling in between.

.github/workflows/_release-wporg.yml
chickenn00dle: Fixed in 18f87bf3, using the job-output approach you suggested.

`Sync package.json versions` now emits `git rev-parse HEAD` after its push as a `released-sha` output, and all three deploy jobs pass it as `ref`. The version gate drops to belt-and-braces.

The SHA is captured after the push completes, including the rebase-retry path, so it always names a commit that exists on the remote for the deploy job to check out. `ref` falls back to the branch tip when empty, which only happens off `release` where the sync step is skipped; the deploy jobs are gated on `refs/heads/release` regardless, so that fallback is really there for the manual caller.

.github/workflows/_release-wporg.yml
chickenn00dle: Fixed in 18f87bf3:

```yaml
run: |
  set -euo pipefail
  version=$(node -p "require('./$PLUGIN_PATH/package.json').version")
  echo "version=$version" >> "$GITHUB_OUTPUT"
```

`plugin-path` moved into an env var at the same time. Confirmed both halves locally: a bogus `plugin-path` exits 1 through the new form and exited 0 through the old one.

Side effect worth noting — this also retires the Copilot thread above about nested quoting, since the line no longer nests.

.github/workflows/_release-wporg.yml
chickenn00dle: Added in 18f87bf3, in `release-wporg.sh` rather than the workflow, so it covers both callers and runs against the built artifact after `release:archive`.

It locates the main file by its `Plugin Name:` header rather than assuming the filename, compares `Version:` to `WP_ORG_PLUGIN_VERSION`, and fails before any SVN work. Checked against all three plugins: main file found, headers agree (3.38.2 / 0.7.5 / 2.8.4), and a forced mismatch exits 1.

That header-not-filename lookup is also what handles `super-cool-ad-inserter`, whose main file is `super-cool-ad-inserter-plugin.php` — the asymmetry you flagged on the other thread.

.github/workflows/release-wporg-manual.yml
chickenn00dle: Fixed in 18f87bf3. `require-release: false` is now passed explicitly, and the comment says that's what the gate branches on, so the two agree.

.github/workflows/release-wporg-manual.yml
chickenn00dle: `permissions: contents: read` added in 18f87bf3, plus the same at job level in the reusable workflow, which covers both callers at once.

On the `environment:` half — I looked into it and it doesn't fit without a larger change. A job that calls a reusable workflow accepts only `concurrency, if, name, needs, permissions, secrets, strategy, uses, with`; `environment` isn't among them. So it would have to be declared inside `_release-wporg.yml`, which both callers share, and a required reviewer there would gate every stable release deploy rather than just this one. Making it input-driven (`environment: ${{ inputs.environment }}`, empty from `release.yml`) may work, but whether an empty environment name resolves to "none" isn't something I could confirm short of a live dispatch.

Leaving it for a follow-up on `main`. The variant I'd argue for there is an environment with no protection rules, which scopes the credentials out of every other workflow without adding a human gate to an escape hatch. Worth flagging that it also means dropping the `secrets:` block from the `workflow_call` definition and reading them from the environment inside the job, since the caller evaluates `${{ secrets.* }}` in its own context.

.github/workflows/release.yml
chickenn00dle: Fixed in 18f87bf3 — `git tag --merged HEAD` on the after side of the diff.

Filtering only the after side turns out to be sufficient: a pre-existing merged tag is in the before file and drops out either way, and a concurrent alpha tag is in neither. Checked in a scratch repo with an alpha tag created between the snapshot and the diff — unfiltered returns both tags, `--merged HEAD` returns only the stable one.

.github/workflows/release-wporg-manual.yml
chickenn00dle: Added in 18f87bf3, at job level in the reusable workflow (`group: wporg-${{ inputs.plugin-name }}`) so both callers share it. A manual deploy queues behind an in-flight release deploy for the same plugin, and the three release deploy jobs still run in parallel since the group is per-plugin.

chickenn00dle:
Addressed the review in 18f87bf3 — one commit, ready for another look. Detail is on each thread; summarising here since two of the points came in the review body and can't be replied to inline.

**The blocker.** Took the `ref`-input option over the job-level guard. The manual workflow now takes a `ref` defaulting to `release` and passes it through, so the branch picker only chooses which copy of the workflow file runs and `ref` decides what publishes. A job-level `if` renders as *skipped*, and an escape hatch that silently no-ops during an incident seemed like the wrong failure mode; a non-`release` ref still deploys but annotates the run.

**The cross-cutting pattern you named** — version equality standing in for "this is the released code" — is closed on all three faces. Deploy jobs pin to a `released-sha` output from the release job, `release-wporg.sh` asserts the built artifact's PHP `Version:` header before touching SVN, and the tag diff is scoped with `git tag --merged HEAD`.

**On the two notes outside the changed lines:**

- *Deploy job inherits a write-scoped token it never uses* — job-level `permissions: contents: read` on `_release-wporg.yml`, covering both callers, plus one on the manual workflow.
- *`wporg-slug` has no caller* — the manual workflow now takes an optional `wporg-slug` and passes it through, so the input is reachable. The other half of that note, one plugin list instead of two, doesn't have a clean answer: a `choice` input's options must be literal YAML, so nothing collapses it to a single list while keeping the dropdown. A matrix would fold `release.yml`'s three job blocks into one and still leave the `choice` list standing. Leaving that for `main`, since this branch targets `release`.

**Deliberately not done:** environment-scoping the SVN credentials. Reasoning is on that thread — short version is that a `uses:` caller job can't declare `environment:`, so it would have to live in the shared reusable workflow and would gate every release deploy, not just the manual one.

**Verification.** Re-ran against the real files rather than reasoning about them: the gate matrix (now 8 cases, including the two manual-ref ones), header extraction across all three plugins plus a forced mismatch, version resolution against a bad `plugin-path`, and the tag diff in a scratch repo with a concurrent alpha tag. YAML parses on all three workflows; `bash -n` clean on the script. As before, the real confirmation is the next stable release run.

Unchanged from the description: none of this retroactively publishes newsletters 3.38.2.

.github/workflows/release-wporg-manual.yml
adekbadek: Both grants confirmed: `permissions: contents: read` at the workflow level here, and at the job level in `_release-wporg.yml`, which narrows the caller for both callers at once.

Agreed on the `environment:` deferral, and the reasoning is right — a job with `uses:` accepts only `concurrency, if, name, needs, permissions, secrets, strategy, uses, with`, so there is no way to scope it per-caller from `release.yml`. The follow-up variant you describe (an environment with no protection rules, declared inside the reusable workflow) is the one worth having: it scopes the credentials without putting a human gate in front of an escape hatch. The `secrets:` note is worth carrying into that follow-up — the caller resolving `${{ secrets.* }}` in its own context is exactly what makes the naive version a no-op.

Resolving; the permissions half is done and the environment half is a follow-up, not a gate on this PR.

adekbadek (approved):
Release risk: LOW — the blocker is fixed and every remaining point from the last pass is addressed.

The manual workflow now declares its own required `ref` (default `release`) instead of inheriting the branch picker, deploys are pinned to `released-sha` rather than a moving tip, `wporg-slug` finally has a caller, and the artifact's PHP `Version:` header is asserted against the version being published. I re-ran the extracted gate over release/skip/stale/manual cases and the header extraction against all three plugins' real main files — all match, and the `--merged HEAD` narrowing does keep a concurrently pushed alpha tag out of the list.

One nit for whenever this file is next touched, annotation-only so not worth a round trip: the manual gate warns when `DEPLOY_REF != 'release'`, so passing an exact SHA — the most pinned input, and the one the checkout comment recommends — prints "publishing from unreleased code", while the looser branch tip passes silently. The condition is inverted relative to what it is guarding.

chickenn00dle:
Picked up the annotation nit in 92a0382, since it's a small change to a file that isn't going to be touched again soon.

The gate now asks git whether the checked-out commit is on `release`, instead of comparing the ref string to `"release"`:

```bash
release_head=$(git rev-parse --verify --quiet refs/remotes/origin/release || true)
sha=$(git rev-parse --short HEAD)
if [ -z "$release_head" ]; then
  echo "::warning::Manual deploy; could not resolve origin/release to check '$DEPLOY_REF' against. Publishing $VERSION from $sha."
elif git merge-base --is-ancestor HEAD "$release_head"; then
  echo "::notice::Manual deploy; publishing $VERSION from $sha, a commit on the release branch."
else
  echo "::warning::Manual deploy; '$DEPLOY_REF' ($sha) is not on the release branch. Publishing $VERSION to wordpress.org from code that was never released."
fi
```

A branch name is now judged by where it points rather than by what it is called, which is what the annotation was after. `fetch-depth: 0` fetches every branch into `refs/remotes/origin/*`, so the ref is available to both callers; the unresolvable case warns rather than passing quietly.

Extracted the step from the YAML and ran it over six refs in a scratch repo:

| Ref | Annotation |
| --- | --- |
| `release` tip | notice |
| SHA at the `release` tip | notice |
| older commit on `release` | notice |
| `main` | warning |
| SHA off `main` | warning |
| `origin/release` unresolvable | warning |

The `require-release: true` path is untouched, and re-ran it for completeness: deploys on a matching tag, skips with no tag, fails on a version mismatch.

PR #813, opened 2026-08-05
fix(content-gate): enforce block visibility on REST reads

jason10lee:
### All Submissions:

Gated blocks now stay hidden on REST reads, matching front-end behavior. The change removes a blanket REST exemption from `Block_Visibility::filter_render_block()`.

The exemption existed so gated blocks stay visible to editors while authoring — block renderer, preview, query-loop rendering inside the editor. That case is already handled further down the same function by the `edit_post` capability check at `:106-109`, which returns early for anyone who can edit the post being rendered, and which `test_editor_bypasses_access_rules_on_front_end()` already covers. Adding a second capability check to the REST branch would have duplicated that check and invited a later cleanup to remove the wrong copy, so the clause comes out instead and the comment records why REST is deliberately not exempt.

Reader-dependent REST responses are marked `Cache-Control: private, no-store` so a shared page cache cannot serve one reader's render to another; a withheld render is identical for every caller, so it stays cacheable.

Auto-generated excerpts carry the same restricted text through a different path, closed by #832 (on `release` already, carried here after a base refresh).

Two notes for reviewers. This targets `release` rather than `main` because it follows the hotfix flow; it reaches `main` through the usual post-release back-merge. And there is no GitHub issue to close—this is tracked in Linear as [NPPM-3087](https://linear.app/a8c/issue/NPPM-3087).

Run `n test-php --filter Newspack_Test_Block_Visibility` from `plugins/newspack-plugin` — 42 tests, which fail on the pre-fix code. (These envs run batcache, so add a cache-buster query param when checking manually.)

### Other information:

* [x] Have you written new tests for your changes, as applicable? — REST-route and page-cache cases in `tests/unit-tests/content-gate/class-block-visibility.php`. `test_rest_request_is_not_exempt_from_access_rules` defines `REST_REQUEST` in an isolated process, since `rest_do_request()` never defines it and a dispatched test cannot tell whether the exemption came back.
* [x] Have you successfully run tests with your changes locally? — `Newspack_Test_Block_Visibility`: 42 tests, 62 assertions, all passing. `phpcs` clean against the root ruleset.


plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -53,10 +53,17 @@ public static function filter_render_block( $block_content, $block ) {
...
-		// from editors during content authoring.
-		if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
+		// Bypass access control in admin screens so blocks are never hidden from
+		// editors during content authoring.
+		//
+		// REST requests are deliberately NOT bypassed here. Core serves
+		// `content.rendered` for published posts to any unauthenticated caller, so a
+		// blanket REST exemption makes every gated block readable by requesting the
+		// post through the API instead of loading the page. Authoring contexts that
+		// arrive over REST (block renderer, preview, query-loop rendering inside the
```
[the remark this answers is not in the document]
jason10lee: Added in `35df38d0a`: `test_rest_request_is_not_exempt_from_access_rules` and `test_editor_still_sees_gated_block_over_rest`, both running in isolated processes so the `REST_REQUEST` constant can't leak into the rest of the suite.

One wrinkle worth flagging for similar tests: `REST_REQUEST` is defined only for real HTTP REST requests (`wp-includes/rest-api.php`), so a test that dispatches through `rest_do_request()` never sets it and wouldn't exercise this path at all. The constant has to be defined explicitly, which is why these two are process-isolated.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -53,10 +53,17 @@ public static function filter_render_block( $block_content, $block ) {
...
 		}
 
-		// Bypass access control in admin screens and REST requests (block renderer,
-		// preview, query-loop rendering inside the editor) so blocks are never hidden
-		// from editors during content authoring.
-		if ( is_admin() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
+		// Bypass access control in admin screens so blocks are never hidden from
+		// editors during content authoring.
+		//
+		// REST requests are deliberately NOT exempt: a context check is not a permission
```
adekbadek, line 59: **Blocker:** Restricted text still reaches logged-out callers through auto-generated excerpts — This closes `content.rendered`, but auto-generated excerpts never reach this filter, so the same restricted text is still returned to an anonymous caller — including from the paginated collection endpoint, and on front-end archive listings.

`wp_trim_excerpt()` removes `do_blocks` from `the_content` (`wp-includes/formatting.php:3980`) and routes through `excerpt_remove_blocks()`, which treats `core/group` as an allowed *wrapper* (`wp-includes/blocks.php:2208`) and hands it to `_excerpt_render_inner_blocks()`. That helper calls `render_block()` on the inner blocks only, so the group — the block that carries the access-control attributes — never reaches `filter_render_block()`. Worth knowing: `core/group` is the only registered target block (`core/stack` and `core/row` are not registered block types), so the one live target is exactly the one core admits into excerpts.

I instrumented the filter on an env running this branch. On the `the_content` path it fires for `[paragraph, paragraph, core/group]` and the restricted text is withheld; on the excerpt path it fires for `[paragraph, paragraph]` only, and the text comes through.

This is pre-existing rather than introduced here — the excerpt route bypassed the filter before this change too. Scope is auto-generated excerpts only; a manually authored excerpt is returned verbatim as the author's own text. Raising it because it is one of the vectors the tracking ticket enumerates, so the ticket should not be resolved on this change alone. The fix belongs at the excerpt stage (`excerpt_allowed_wrapper_blocks`, or dropping restricted groups before excerpt generation) rather than in this filter.

**To reproduce** (post with an empty `post_excerpt` containing an ungated paragraph plus a registration-gated `core/group`, all requests logged out and cache-busted):
1. `curl -sk 'https://<env>/<permalink>?cb=1'` — restricted marker absent. Control; this is the fix working.
2. `curl -sk 'https://<env>/wp-json/wp/v2/posts/<id>?_fields=content&cb=1'` — restricted marker absent. Also the fix working.
3. `curl -sk 'https://<env>/wp-json/wp/v2/posts/<id>?_fields=excerpt&cb=1'` — restricted marker **present**.
4. `curl -sk 'https://<env>/wp-json/wp/v2/posts?per_page=100&cb=1'` — restricted marker **present**.
5. `curl -sk 'https://<env>/?cb=1'` — restricted marker **present** in the home-page listing.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -53,10 +53,17 @@ public static function filter_render_block( $block_content, $block ) {
...
+		// check, and access has to be evaluated per requester. Authoring contexts that
+		// arrive over REST (block renderer, preview, query-loop rendering inside the
+		// editor) are covered by the `edit_post` check further down, which needs a post
+		// in scope — where none is set up, it fails closed. Re-adding a blanket REST
+		// exemption here would widen what is readable; the REST cases in
+		// tests/unit-tests/content-gate/class-block-visibility.php pin the behaviour.
+		if ( is_admin() ) {
```
adekbadek, line 66: **Suggestion:** `is_admin()` is the kind of check the new comment argues against — The comment directly above states the principle — "a context check is not a permission check, and access has to be evaluated per requester" — and the line it introduces is an unconditional context check. `is_admin()` is true on `admin-ajax.php` and `admin-post.php`, both of which dispatch `*_nopriv_*` actions for callers who are not signed in, and both of which any signed-in reader can reach.

Nothing in this codebase renders post content on those routes today — I checked every `wp_ajax_nopriv_` and `admin_post_nopriv_` registration in the workspace and none of them runs a post through `the_content` — so this is a latent edge rather than a live one, and it is unchanged by this PR. Worth a follow-up rather than a change here: `is_admin() && ! wp_doing_ajax()` would keep the bypass on real admin screen loads, which are already behind admin authentication.

One note for whoever picks it up: the theme's infinite-scroll render callback looks like it would reach this, but it renders through `the_excerpt()` (`template-parts/content/content-excerpt.php:81`), which per the first thread never reaches this filter at all. It is not evidence for this one.

adekbadek, line 66: **Suggestion:** REST output now varies by requester, which a shared page cache in front of it will not know — Before this change every REST caller got the same bytes, so a cache in front of the REST routes could not leak anything. Now `content.rendered` depends on who is asking, and any cache that keys on URL without varying on identity can serve one caller's render to another.

I hit this on the review env, which runs batcache. A REST request authenticated with an application password carries no WordPress cookie and no `X-WP-Nonce`, and batcache's skip list checks only those two things — so the authenticated render was stored (`X-nananana: Batcache-Set`) and the next anonymous request for the same URL was served it (`Batcache-Hit`), restricted text included. Ordinary browser traffic is not affected: a real block-editor request carries both a cookie and a nonce, so batcache skips it either way.

This is the review env's cache configuration, not necessarily production's, so it is worth confirming against the cache actually sitting in front of `wp-json` on hosted sites before deciding whether anything is needed. If it does apply, sending `Cache-Control: private` on responses where a block was withheld, or keeping these routes out of the page cache, would be enough.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-block-visibility.php
```diff
@@ -420,6 +420,68 @@ public function test_non_editor_still_restricted_on_front_end() {
...
+	 * as a whole, so this pins that the exemption is not reintroduced.
+	 *
+	 * `REST_REQUEST` is defined only for real HTTP REST requests (wp-includes/rest-api.php),
+	 * so the constant is set explicitly here. Dispatching through `rest_do_request()` never
+	 * defines it, and a test written that way would not exercise this path at all.
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_rest_request_is_not_exempt_from_access_rules() {
```
adekbadek, line 497: **Suggestion:** The two new tests do not constrain what their docblocks claim — Both new cases are the existing front-end tests (`test_non_editor_still_restricted_on_front_end` at :407 and `test_editor_bypasses_access_rules_on_front_end` at :387) plus `define( 'REST_REQUEST', true )`. Since the fixed function no longer reads that constant, the assertions pass identically with or without it, and each pays a full WordPress re-bootstrap for `@runInSeparateProcess`.

`test_rest_request_is_not_exempt_from_access_rules` still earns its place — it fails if the REST clause is re-added, which is exactly the regression worth pinning. `test_editor_still_sees_gated_block_over_rest` does not: restore the exemption and it passes via the exemption instead, with the same result, so it cannot distinguish the mechanism its docblock says it proves.

Dispatching the route instead — `rest_do_request( new WP_REST_Request( 'GET', '/wp/v2/posts/' . $post_id ) )`, asserting on `content.rendered` — would pin the behaviour that was actually reported, exercise the filter under the controller's own post setup (the thing `get_the_ID()` at :107 depends on), and let both process-isolation annotations go. It would also have surfaced the excerpt gap on the first thread.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -53,10 +53,17 @@ public static function filter_render_block( $block_content, $block ) {
...
+		// Bypass access control in admin screens so blocks are never hidden from
+		// editors during content authoring.
+		//
+		// REST requests are deliberately NOT exempt: a context check is not a permission
+		// check, and access has to be evaluated per requester. Authoring contexts that
+		// arrive over REST (block renderer, preview, query-loop rendering inside the
+		// editor) are covered by the `edit_post` check further down, which needs a post
+		// in scope — where none is set up, it fails closed. Re-adding a blanket REST
+		// exemption here would widen what is readable; the REST cases in
+		// tests/unit-tests/content-gate/class-block-visibility.php pin the behaviour.
```
adekbadek: **Nit:** Comment hard-codes a test file path — Naming `tests/unit-tests/content-gate/class-block-visibility.php` inline gives the comment a second thing to keep true. The test class name alone survives a file move and is just as findable.

adekbadek (changes requested):
The change does what it says for `content.rendered`, and the reasoning for removing the REST branch rather than narrowing it holds up — the surviving `edit_post` check genuinely carries the authoring case. I verified that on an env running this branch: an editor authenticated over REST still receives the gated block, and the block editor still renders it, so there is no authoring regression.

What I could not confirm is the description's claim that gated blocks now stay hidden on REST reads. They stay hidden in `content.rendered`; they are still returned in `excerpt.rendered`, and therefore in the default single-post response and the paginated collection response. Details and repro are on the first thread. That route never reaches this filter, so it is pre-existing rather than introduced here — LMK if this should rather be tackled in a separate PR.

The remaining notes are non-blocking.

- `plugins/newspack-plugin/includes/content-gate/class-block-visibility.php:107` — **Suggestion:** REST renders that do not prime the post in scope now evaluate rules against the requester — The bypass below depends on `get_the_ID()` having been primed. Core's posts controller does prime it, which is what makes the authoring case work — but a REST route that renders a specific post without calling `setup_postdata()` gets no post in scope, so the rules are evaluated against the requesting user instead, and a gated section is dropped even though the request itself was authorized. The concrete instance is network content distribution: `Outgoing_Post::get_processed_post_content()` applies `the_content` to `$this->post` with no `setup_postdata()` (`plugins/newspack-network/includes/content-distribution/class-outgoing-post.php:387`), and that output becomes the payload's `content` field (`:294`). It is reached from the distribution REST handlers (`class-api.php:181` and `:234`), where the old exemption previously passed the content through untouched. I could not exercise this — it needs a configured hub/node pair, which the review env does not have — so treat it as unverified. What would settle it: distribute a post containing a gated group from a hub as a user who does not satisfy the block's rules, and check whether the section is present in the payload's `content` field.
- `plugins/newspack-plugin/includes/content-gate/class-block-visibility.php:283` — **Suggestion:** `has_active_gates()` re-fetches gates that the evaluation below fetches again — `has_active_gates()` calls `Content_Gate::get_gate()` per gate on every block render with no memoisation, and `compute_gate_rules_match()` immediately re-fetches the same gates. `get_gate()` rebuilds its array from `get_post()` plus several meta reads each time. This matters a little more now that REST collection reads reach this code: a `?per_page=100` request with a gated block per post repeats that reconstruction a few hundred times per request. The object cache bounds the database cost, so this is not urgent. Memoising `get_gate()` per request, or caching the `has_active_gates()` result alongside the existing rules cache, would remove the duplication. The rule evaluation itself is fine — that cache is keyed by user and per-request, so identical blocks across a collection evaluate once.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: Closed on `release` by #832 (`86eecfeef`), which this branch picked up in the refresh — so it's covered before merge rather than left to the tracking ticket.

The fix sits at the excerpt stage as you suggested: `Content_Gate_Excerpt` hooks `get_the_excerpt` and strips withheld blocks from the post content before `wp_trim_excerpt()` sees it, rather than widening this filter.

Re-ran your five steps against the refreshed branch — 3, 4 and 5 no longer show the marker, 1 and 2 unchanged. I checked the instrument too: unhooking `Content_Gate_Excerpt` makes 3–5 reproduce again, so the pass is the gate working rather than the fixture going quiet.

Thanks for the repro; it's what made this quick to confirm.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: Filed as a follow-up in Linear (NPPM-3154) with the direction you proposed.

jason10lee: Fixed in `ba99cad03`. `filter_render_block()` now calls `batcache_cancel()` and sends `Cache-Control: private, no-store` whenever the render shows a block an anonymous reader would not see. A withheld render is identical for every caller, so that stays cacheable — the cancel is scoped to the privileged case.

On whether it reaches past the review env: the `advanced-cache.php` this build ships skips on exactly the two things you found, an `X-WP-Nonce` header or a `wp*`/`wordpress*` cookie. `DONOTCACHEPAGE` isn't honoured there, so `batcache_cancel()` is what does the work; the header covers anything else in the path.

The decision is split out as `render_varies_from_anonymous()` so it can be tested — `batcache_cancel()` and `header()` are both unobservable under PHPUnit, so a test written against response headers asserts nothing. `test_privileged_render_varies_from_anonymous` asserts the decision directly.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-block-visibility.php
jason10lee: Swapped in `bce372e9f`. `test_editor_still_sees_gated_block_over_rest` is gone; `test_rest_route_gates_content_and_excerpt_per_requester` dispatches `/wp/v2/posts/<id>` and asserts on both `content.rendered` and `excerpt.rendered`, logged out and as the post's editor.

One part didn't fall out as hoped — `test_rest_request_is_not_exempt_from_access_rules` still needs its process isolation. `rest_do_request()` doesn't define `REST_REQUEST`, so a dispatched test can't tell whether the exemption came back; it would pass either way. The two end up complementary: the dispatched one covers the controller's post setup and the excerpt path, the constant one is the regression pin. Only the second annotation went.

Dispatching would indeed have surfaced the excerpt gap — the new test fails on `excerpt.rendered` if I unhook the excerpt gate.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: Done in `f69cee8f8` — the comment now names `Newspack_Test_Block_Visibility`.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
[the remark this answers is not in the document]
jason10lee: Update: `test_editor_still_sees_gated_block_over_rest` was replaced in `bce372e9f` by `test_rest_route_gates_content_and_excerpt_per_requester`, which dispatches the route instead. The note about `REST_REQUEST` still holds, and is why the other test keeps its process isolation.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
adekbadek: Confirmed on the refreshed branch. `Content_Gate_Excerpt::filter_get_the_excerpt()` hands `wp_trim_excerpt()` a clone whose `post_content` has been through `strip_blocks_hidden_from_public()`, and the `filter = 'raw'` line is what stops core re-reading the row and discarding the clone — that was the trap I would have expected this to fall into. Evaluating against user 0 rather than the current reader is the right call given the block-markup cache has no user dimension.

`test_rest_route_gates_content_and_excerpt_per_requester` pins `excerpt.rendered` alongside `content.rendered`, so steps 3 and 4 of the repro are now covered by the suite rather than by a manual run. Resolving.

adekbadek (changes requested):
The excerpt blocker is closed — `Content_Gate_Excerpt` came in with the `release` merge and the new REST-route test pins `excerpt.rendered` alongside `content.rendered`; detail on the resolved thread. The new cache guard is the right idea but only covers one of the two ways a render can differ from the anonymous one, so it still lets a per-requester response into the page cache.

**`plugins/newspack-plugin/includes/content-gate/class-block-visibility.php:97`** — **Blocker:** `render_varies_from_anonymous()` misses the `hidden` visibility direction

The check is `! is_hidden_for_user( $block, $user_id ) && is_hidden_for_user( $block, 0 )` — true only when the requester sees a block anonymous does not. `Visibility` is a two-way toggle in the editor (`block-visibility.tsx:112-113`), and `is_hidden_for_user()` inverts on it: with `visibility: 'hidden'` a reader who matches the rules has the block withheld while anonymous keeps it. That render also varies per requester, and the guard returns false for it, so under the same premise the commit is built on — batcache stores a request that carries neither an X-WP-Nonce header nor a wp\*/wordpress\* cookie — it gets stored and served to anonymous readers who are entitled to the block.

Worked through with `mode: custom`, `rules: { registration: { active: true } }`, `visibility: 'hidden'`: user 0 does not match, so `is_hidden_for_user` returns `$user_matches` = false and anonymous sees the block; an application-password subscriber matches, so it returns true and the block is withheld; `render_varies_from_anonymous()` short-circuits on `! true` and returns false. Not a leak — the failure is a block silently disappearing for everyone — but it is the same class the commit exists to prevent, and it is newly reachable here because REST responses did not vary per requester before this change. Hiding a donate or subscribe CTA from existing members is the ordinary use of that toggle, so it is not an exotic configuration.

Comparing the two decisions rather than one direction covers both, and `filter_render_block()` has already computed the first half:

```php
return self::is_hidden_for_user( $block, $user_id, $post_id ) !== self::is_hidden_for_user( $block, 0 );
```

The `empty( $user_id )` early return becomes redundant under that form (the two sides are identical for user 0) but is harmless to keep. `test_privileged_render_varies_from_anonymous()` would want a fourth case for it — a rules-matching non-editor against a `hidden` block, asserting true.

**Note, non-blocking, same method:** the early return also encodes "an anonymous render never varies", which the `institution` rule contradicts — it sets `supports_anonymous => true` (`class-access-rules.php:161`) and evaluates on IP or cookie, so two anonymous readers legitimately get different renders and the first one cached wins. That one is pre-existing and unchanged by this PR — the front-end path always behaved that way — so it belongs on the tracking ticket rather than here. Worth a line in the docblock so the assumption is not read as verified.

jason10lee:
Fixed in `7ab7c4006`, using your form—comparing the two decisions rather than testing one direction. The `empty( $user_id )` early return went with it, since the sides are identical for user 0.

`test_privileged_render_varies_from_anonymous()` gained the fourth case. It asserts the mechanism first (a matching subscriber gets `is_hidden_for_user` true on a `hidden` block while anonymous gets false), then that the guard returns true. Reverting to the one-directional form fails it with `Failed asserting that false is true`, and nothing else in the class fails—which is the shape of the gap you described.

The `institution` note is in the docblock rather than left implicit: the method assumes an anonymous render never varies from another anonymous render, and `supports_anonymous` with IP or cookie evaluation contradicts that. Recorded as pre-existing on the front-end path and tracked, rather than assumed away.

Your two notes from the 08-17 review are now NPPM-3160 (rules evaluating against the requester when no post is primed) and NPPM-3161 (the `has_active_gates()` re-fetch) — https://linear.app/a8c/issue/NPPM-3160 and https://linear.app/a8c/issue/NPPM-3161. Both carry your reasoning; 3160 keeps the unverified caveat and the hub/node reproduction you set out. The line references moved in the release merge, so the ticket cites what I verified on the current branch: `Outgoing_Post::get_processed_post_content()` at `:416`, the payload assignment at `:330`.

Full plugin suite green: 2769 tests, 7913 assertions.

adekbadek (approved):
The cache guard now compares both hide-decisions, so the `hidden` direction of the visibility toggle is covered, and the new test case pins it with a rules-matching subscriber on a post they cannot edit. The institution/anonymous-varies caveat is recorded in the docblock as an explicit assumption rather than left implied.

One nit for whenever this is next touched: the inline comment above the guard in `class-block-visibility.php:78` still frames the check as "shows a block an anonymous reader would not see" — the one-directional phrasing the new docblock identifies as insufficient.

PR #814, opened 2026-08-05
fix(reader-activation): treat an autofilled honeypot as a reader, not a bot

dkoo:
### All Submissions:

> **Stacked on #802 — review and merge that first.** This PR's base is `hotfix/auth-honeypot-autofill`, so its diff shows only the server-side change.
>
> **This one is a proposal, not a defect fix.** It trades away some spam protection. Read the trade-off section before approving; declining it is a reasonable outcome and leaves #802 standing on its own.

#802 stops browsers autofilling the anti-bot decoy on the reader sign-in and newsletter subscribe forms. That fixes the reported bug, but it works by making the field invisible to autofill — which depends on browser internals we don't control and can't fully verify. If any browser or password manager fills the decoy anyway, the reader is told they're signed in, no session is created, and they have no way to diagnose or recover from it.

This adds a second line of defence that doesn't depend on browser behaviour. `Reader_Activation::is_honeypot_tripped( $honeypot, $email )` is used at all four trap sites — `process_auth_form()`, the reader-registration block, the REST registration endpoint, and the Newsletters Subscribe block. A browser autofills the decoy and the real field from **one profile entry**, so a decoy value matching the submitted address is autofill rather than a bot. Everything else still trips: a differing value, a decoy filled with junk, a decoy filled with no address alongside it.

Two details worth knowing:

- **The decoy is compared as text, never sanitized as an email first.** Email sanitizers blank anything that isn't address-shaped — `John Smith`, `buy-cheap-pills`, a URL, or any non-Latin string — which is most of what bots put in a honeypot. Sanitizing before testing whether the field was filled would silently disable the trap for exactly that population.
- **Both sides are normalized inside the helper.** The call sites reach it through different sanitizers, and those disagree on non-ASCII (`sanitize_email( 'réader@example.test' )` returns `rader@example.test`). Comparing them unnormalized would leave readers with accented addresses locked out — the very failure this exists to prevent.

The Subscribe block defers to the shared helper under a `method_exists` guard, keeping one authoritative implementation without a hard dependency on a newspack-plugin version. It also reads the decoy from the transport the submission actually used: the block's script posts, but its form carries no `method`, so a submission without JS arrives as a GET.

### ⚠️ The trade-off: this reduces spam protection

**A bot that puts the same address in every field it recognises now gets past the honeypot.** That is the most common naive-bot behaviour, so this is not a hypothetical.

What that costs, concretely, on a site **without** reCAPTCHA — the only sites where the decoy renders at all:

- The decoy is the only gate in front of the form-post sign-in and registration handler.
- That handler has **no per-IP rate limit**. (The REST registration path does, via `check_registration_rate_limit()`; the form-post path doesn't.)
- So a submission that gets past the decoy can create a reader account and send outbound authentication mail.

The argument for accepting it: a reader locked out of their own account has no way to understand or fix what happened, while spam registrations are visible, cleanable, and not solely defended by this decoy. A bot that echoes one address into every field is also one that a per-IP limit would stop — which is the durable answer, and is filed as a follow-up rather than done here.

The argument against: it is a real reduction in a live defence, on sites that have no other one, taken to guard against a browser behaviour we haven't observed since #802 landed.

**If this isn't wanted, close it.** #802 fixes the reported bug on its own.

**If it does land, watch new registration volume on reCAPTCHA-less sites for a few weeks.** A step change there is the signal that the decoy has been worked around, and it is the only signal available, since a skipped decoy looks identical to a real signup.

On a site with reader activation enabled and reCAPTCHA off:

1. **The fix.** Submit the sign-in form with the **same** address in both `npe` and `email` (via devtools) — that's what autofill produces. Before this change the response was a 200 with `authenticated: 1` and the modal showed "Success! You're signed in." with no session. Now it proceeds to the normal flow, byte-identical to a submission with no decoy value at all.
2. **The trap still works.** Submit with **different** values in `npe` and `email`. Still returns the fake success.
3. **Junk still trips it.** Submit with `email=buy-cheap-pills`, or a non-Latin string such as `email=买便宜药`. Both still return the fake success.
4. **Decoy alone still trips it.** Submit with `email` filled and `npe` empty.
5. Repeat 1–4 against the Newsletters Subscribe block (`newspack_newsletters_subscribe=1`), including as a **GET** submission — the no-JS path.
6. Tests: `n test-php --group auth-honeypot` and `--group subscribe_honeypot`.

### Other information:

Full suites green — newspack-plugin 2359 tests / 6693 assertions, newspack-newsletters 713 / 2967 — and CI-parity PHPCS clean. Every scenario above was exercised against a running site, not only unit-tested.

**Follow-ups (not in this PR):**
- A per-IP limit on the form-post sign-in/registration handler, matching the one the REST path already has. This is the durable defence, and this PR makes it more load-bearing.
- The fake-success response still claims `authenticated: 1` on a path that creates no session. This PR stops autofill from reaching it; returning something that doesn't claim a session would make the whole class harmless.
- `email[]=x` bypasses the trap entirely, since `sanitize_text_field()` returns `''` for arrays. Pre-existing and unchanged here.
- The REST route's `args` declare `'type' => 'string'` but have no `validate_callback`, and `register_rest_route()` doesn't attach one, so `npe[]=x` can reach `sanitize_email( array )`. Pre-existing; the endpoint requires a valid integration key.
- No test exercises the three form handlers directly. Both regressions this change went through during review lived at a call site, not in the predicate.

### DIFF
--- a/plugins/newspack-newsletters/src/blocks/subscribe/index.php
+++ b/plugins/newspack-newsletters/src/blocks/subscribe/index.php
@@ -89,7 +89,8 @@ function get_form_id() {
  * `display:none` as a fallback for a policy that strips the style attribute.
  *
  * newspack-plugin renders its own copy of this field in
- * `Newspack\Reader_Activation::render_honeypot_field()`. Changes here belong
+ * `Newspack\Reader_Activation::render_honeypot_field()`. Only the markup is
+ * duplicated — the submit-side rule defers to that plugin. Changes here belong
  * there too.
  *
  * Not rendered if reCAPTCHA is enabled as it's a superior spam protection.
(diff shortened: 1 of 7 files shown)


dkoo:
Closing this one. After discussing with the team, we decided this is not worth it due to the increased spammability it might introduce.

PR #819, opened 2026-08-06
fix(subscriptions): enable legacy subscription product types once

chickenn00dle:
> [!IMPORTANT]
> **Hotfix — targets `release`.** Small and self-contained by design, so it can ship in a patch. It will reach `main` via the usual `release` → `main` promotion.

## Why

WooCommerce Subscriptions 9.0 added a **Subscription product creation** section to WooCommerce → Settings → Subscriptions, with two checkboxes:

| Option | Default in 9.0 |
| --- | --- |
| `woocommerce_subscriptions_enable_simple_subscription` | `no` |
| `woocommerce_subscriptions_enable_variable_subscription` | `no` |

They control whether `subscription` and `variable-subscription` appear in the product-type dropdown (`WC_Subscriptions_Admin::add_subscription_products_to_select()`). Subscriptions is steering merchants toward *subscription plans* on ordinary simple/variable products, so on a site that has never configured them, both are off.

Newspack still creates the legacy types. The Audience wizard writes them directly — `class-audience-subscription-products.php:519` and `:581` — and `Audience_Subscription_Products::PRODUCT_TYPES` names them explicitly. On a site where both settings are off, a publisher cannot create or switch to a product type our own wizard produces.

Two things narrow where this actually helps, both confirmed against WCS 9.1.0-rc1:

- **Fresh installs are covered.** 9.1 does not pre-seed the two rows, so a new site arrives with them absent.
- **Subscriptions' own 9.0 migration only turns the settings on where products of that type already exist**, and only on an authenticated admin pageload. So the case this uniquely covers is a site that has not created its subscription products yet. Where Subscriptions would have acted, both write `yes`, so the two never disagree.

This is **pre-emptive**: no publisher has reported it. The intent is that the option already reads `yes` before sites land on 9.x, so the new default never bites.

## What changed

One method on `WooCommerce_Subscriptions`, registered on `admin_init` and guarded on Subscriptions being active:

```php
$absent  = new \stdClass();
$written = [];
foreach ( $options as $option ) {
	if ( $absent === get_option( $option, $absent ) ) {
		$written[] = $option;
	}
}
```

The interesting part is the condition. Distinguishing "never configured" from "deliberately off" comes from storage rather than bookkeeping:

- `get_option()` returns the default only when no row exists.
- It returns `'no'` once anything has written it — and WooCommerce's settings screen writes `'no'` when a publisher unticks the box.

So an absent row is the only safe signal that the publisher has no preference, and it is the only case we write. That also makes the whole thing self-limiting: after the first write the row exists, so the check never passes again, and an unticked box stays unticked. No version tracking, no migration.

Verified on a real install rather than assumed:

```
absent:            false
after saving 'no': 'no'
  is false?        false   <- so an explicit "no" is never overwritten
```

Deliberately **not** gated on the Subscriptions version. Writing the option before a site reaches 9.0 is the point — the new default then never applies.

**`admin_init`, not `plugins_loaded`.** This is the only place Newspack writes a third-party plugin's settings for a publisher, so it runs where the plugin's three other third-party option writes run — `Parsely::migrate_meta_type()`, `WooCommerce_Email_Style_Sync`, `Emails_Section` — keeping it off ordinary front-end pageloads, REST requests and cron. It is not an authentication boundary: `admin-ajax.php` fires `admin_init` too, so a `wp_ajax_nopriv_*` action reaches it unauthenticated. Harmless here, since the write is one-shot and identical either way, but the hook buys a narrower surface, not a logged-in one. Nothing is lost by waiting: a Subscriptions upgrade happens through an admin action, so `admin_init` fires in the same session, and the dropdown this unblocks only exists in wp-admin. It also puts the opt-out filter within reach of a theme's `functions.php`.

**The write is recorded.** `newspack_subscriptions_enabled_product_types` stores the list of option names Newspack created, and a `Logger::newspack_log()` call names the same list for Newspack Manager. So a corrective release could reverse only the rows we made and leave a publisher's own choice alone.

A `newspack_subscriptions_enable_legacy_product_types` filter short-circuits the whole thing for publishers or custom code that want none of it.

## What it deliberately does not do

- No forcing on read, and no disabling the checkboxes. The setting stays a real setting a publisher can change.
- No UI changes.
- No attempt to change sites where the options are already written.

## Testing

- 8 tests, `WooCommerce_Subscriptions_Integration` group: absent options are enabled; an explicit `'no'` is respected; an existing `'yes'` is left alone with **no** redundant write; the write is recorded and logged, and nothing is recorded when there is nothing to write; the filter opts out entirely; the `admin_init` registration holds and `plugins_loaded` does not carry it; an inactive Subscriptions writes nothing.
- Full suite: **2368 tests, 6736 assertions, 0 failures** (9 pre-existing skips).
- PHPCS clean on both files.

Confirmed manually on an isolated env running WooCommerce Subscriptions 9.1.0-rc1, with both rows deleted first:

```
BEFORE  simple=false  variable=false  marker=false
        product_type_selector = simple / variable
AFTER   simple='yes'  variable='yes'
        marker = ["…enable_simple_subscription", "…enable_variable_subscription"]
        product_type_selector = simple / variable / subscription / variable-subscription
```

Ten anonymous front-end GETs in that same state returned 200 and left both options absent.

## Known limits, stated plainly

- **Only helps sites where the options have never been written** — fresh sites, and sites not yet upgraded to 9.x. A site already on 9.x whose options were written stays as configured. That matches the pre-emptive intent; covering those would mean overriding a setting a publisher may have chosen, which this deliberately does not do.
- **Saving the Subscriptions settings tab writes `no` to fields nobody touched.** `WC_Admin_Settings::save_fields()` writes every field in a section on any save of that tab. A site that reached 9.0 and then saved WooCommerce → Settings → Subscriptions for an unrelated reason therefore has `'no'` rows nobody chose, and this permanently no-ops for it — plausibly the sites most likely to hit the problem. Widening the test to treat `'no'` as fair game would forfeit the one thing that makes writing another plugin's setting safe, so those sites are left to the settings screen.
- **The opt-out filter must be registered before `admin_init`**, and cannot undo a write already made. A plugin, mu-plugin or a theme's `functions.php` all work.
- **Missing product types block manual creation in wp-admin only.** The Audience wizard is unaffected — it writes the product type directly — and WooCommerce re-offers a type once products of it exist or one is being edited. So the blast radius of the underlying problem is narrower than the settings screen suggests.

## Scope

This is intentionally the *only* thing in this PR. The separate work teaching the subscription tiers modal to handle plan-based products (NPPM-3053) moves to a normal PR against `main`.


plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -670,8 +670,54 @@ public static function woocommerce_subscriptions_integration_init() {
 		Subscriptions_Meta::init();
 		Subscriptions_Confirmation::init();
 		Card_Expiry_Warning::init();
+
+		if ( self::is_active() ) {
```
adekbadek: **Blocker:** The write runs on anonymous front-end requests, unlike the plugin's other third-party option writes — `woocommerce_subscriptions_integration_init` is hooked to `plugins_loaded` (line 26), so this runs on every request — anonymous uncached front-end hits, REST calls and cron included. I confirmed an anonymous front-end GET performs the write with no admin session involved.

newspack-plugin writes third-party options in three other places, and all three use `admin_init` plus a `newspack_*` bookkeeping flag: `Parsely::migrate_meta_type()` (`class-parsely.php:22,28`), `WooCommerce_Email_Style_Sync` (`class-woocommerce-email-style-sync.php:104,135`), and — closest of all — `Emails_Section` writing the Subscriptions option `woocommerce_subscriptions_customer_notifications_enabled` (`class-emails-section.php:66,530`), whose docblock at `:557-563` argues the choice outright: "so the GET endpoint stays idempotent — a probe, schema crawler, or sibling-admin pageload that hits the wizard route doesn't trigger silent WC option writes".

There may be a deliberate reason to fire earlier and more often here — winning the race against a Subscriptions upgrade. If so it's worth writing down, since that upgrade happens through an admin action and `admin_init` would fire in the same session anyway. Reusing the target plugin's own row as the bookkeeping flag is what keeps this compact, but it also fuses "have we already considered this site?" with "what did the publisher choose?", so the decision can't be re-run or audited afterwards.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -670,8 +670,54 @@ public static function woocommerce_subscriptions_integration_init() {
...
+	/**
+	 * Enable the `subscription` and `variable-subscription` product types, once.
+	 *
+	 * WooCommerce Subscriptions 9.0 added a "Subscription product creation" section
+	 * whose two checkboxes default to off, which removes those product types from
+	 * the product-type dropdown. Newspack still creates them — the Audience wizard
+	 * writes them directly — so on an unconfigured site a publisher cannot create
+	 * or switch to a product type our own wizard produces.
+	 *
+	 * Only an option that has never been written is touched. `get_option()` returns
```
adekbadek: **Suggestion:** Saving the Subscriptions settings tab writes `no` even to untouched fields — The docblock is right that an absent row is a safe signal, but the converse it leans on — that a present `no` is a publisher preference — is weaker than it reads. `WC_Admin_Settings::save_fields()` writes a value for every field in the section on any save of that tab (`repos/plugins/woocommerce/includes/admin/class-wc-admin-settings.php:955`), not only the ones the publisher touched. So a publisher who reached 9.0 and then saved WooCommerce → Settings → Subscriptions for some unrelated reason has `no` rows they never chose, and this permanently no-ops for them — plausibly the sites most likely to hit the problem in the first place.

Probably worth stating as a known limitation rather than changing the logic; widening the test would forfeit the safety the conservative check buys.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -670,8 +670,54 @@ public static function woocommerce_subscriptions_integration_init() {
...
+		}
+
+		$options = [
+			'woocommerce_subscriptions_enable_simple_subscription',
+			'woocommerce_subscriptions_enable_variable_subscription',
+		];
+
+		foreach ( $options as $option ) {
+			if ( false === get_option( $option ) ) {
+				update_option( $option, 'yes' );
```
adekbadek: **Blocker:** Nothing records that Newspack made the write — Once the row exists, nothing distinguishes "Newspack wrote this" from "the publisher ticked the box" — no flag, no log line. When a publisher later asks why the setting is on, or support needs to know how many sites were touched, there's no way to answer. `Logger::newspack_log( $code, $message, $data )` already exists in this plugin and is consumed by Newspack Manager, so one call inside the write branch would make the change auditable platform-wide.

**Consider:** Writing a Newspack-owned marker option alongside the change, or logging the write, would make it possible to tell later which sites had the setting written for them, and to reverse only those.

**Risk if not addressed:** If the write turns out to be wrong for some publishers, there is no way to tell which of the two rows Newspack created and which the publisher chose, so it cannot be undone selectively — a corrective release would either leave every site as it is or reset a setting some publishers deliberately picked.

**Consider:** Writing a Newspack-owned marker option alongside the change, or logging the write, would make it possible to tell later which sites had the setting written for them, and to reverse only those.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -670,8 +670,54 @@ public static function woocommerce_subscriptions_integration_init() {
...
+	 */
+	public static function maybe_enable_legacy_product_types() {
+		/**
+		 * Filters whether Newspack enables the legacy subscription product types.
+		 *
+		 * Returning false leaves the WooCommerce settings untouched.
+		 *
+		 * @param bool $enable Whether to enable the legacy subscription product types.
+		 */
+		if ( ! apply_filters( 'newspack_enable_legacy_subscription_product_types', true ) ) {
```
adekbadek: **Nit:** Filter name drops the subsystem prefix — AGENTS.md documents hooks as `newspack_<plugin>_<name>`, and the only other Newspack filter in this file is `newspack_subscriptions_expiration_enabled` (`:747`). `newspack_enable_legacy_subscription_product_types` leads with the verb and drops the prefix, so it won't surface for anyone grepping `newspack_subscriptions_`. "Legacy" also appears nowhere else for these types — WooCommerce calls them simple and variable subscriptions — so it gives a future maintainer nothing to search on. `newspack_subscriptions_enable_legacy_product_types` would fix both. Only worth raising because a published filter name is permanent API and this is the cheapest moment to change it.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions-product-types.php
```diff
@@ -0,0 +1,122 @@
...
+			delete_option( $option );
+		}
+		remove_all_filters( 'newspack_enable_legacy_subscription_product_types' );
+		parent::tear_down();
+	}
+
+	/**
+	 * A site that has never configured the setting gets both types enabled.
+	 */
+	public function test_absent_options_are_enabled() {
```
adekbadek, line 90: **Nit:** The call site and the `is_active()` guard aren't covered — All four tests call `maybe_enable_legacy_product_types()` directly, so deleting the new call site at `class-woocommerce-subscriptions.php:674-676` leaves the suite green — and that call site is the entire reason this ships. The `is_active()` guard is the single decision point governing whether a third-party plugin's options get written at all, and it's uncovered because it sits at the call site rather than inside the method.

It's genuinely awkward to close: `tests/mocks/wc-mocks.php:663` declares `WC_Subscriptions` unconditionally, so `is_active()` can't be made false in the suite. A test that calls `woocommerce_subscriptions_integration_init()` and asserts the options landed — or just asserts the `plugins_loaded` registration, the way `class-woocommerce-subscriptions.php:1434-1438` already does with `has_action` — would at least pin the wiring.

adekbadek (changes requested):
Release risk: ELEVATED — a hotfix that writes a third-party plugin's settings on every site, and a documented opt-out that only works from plugin-scope code. 

I retested everything against real WooCommerce Subscriptions 9.1.0. The premise holds exactly as described — with both rows absent the product-type selector offers only `simple` and `variable`, and with both `yes` it offers the two subscription types — so on an unconfigured 9.x site a publisher genuinely cannot reach a product type the Audience wizard produces.

One thing worth correcting in the description: the change is better justified than it claims, not worse. WCS 9.1 does not pre-seed these options on a fresh install, so fresh sites are covered. And WCS's own 9.0 migration only turns the settings on where products of that type already exist, and only on an authenticated admin pageload — so the case this uniquely covers is a site that has not created its subscription products yet. Where WCS would have acted, both write `yes`, so the two never disagree.

What is left is the opt-out filter, which takes effect only from a plugin or mu-plugin at file scope — not a theme's `functions.php`, not `init` — and, because the write is one-shot, cannot be used after the fact. That is worth either documenting on the filter or resolving by moving the call to `admin_init`. The rest are smaller: a more robust absence primitive, no record that Newspack made the write, and an untested call site.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
chickenn00dle: Moved to `admin_init` in 02a4cd3. There was no deliberate reason to fire earlier, so rather than write down a justification for `plugins_loaded`, the docblock now records why waiting costs nothing: a Subscriptions upgrade runs through an admin action, so `admin_init` fires in the same session, and the product-type dropdown this unblocks only exists in wp-admin.

Confirmed on an env running WCS 9.1.0-rc1, with both rows deleted first: 10 anonymous front-end GETs, all 200, both options still absent afterwards.

That also resolves the filter-scope point from the review body. On `admin_init` a theme's `functions.php` registers in time, so the opt-out is reachable from ordinary publisher code rather than plugin scope only. The one-shot limit still stands, and the filter docblock now says so.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
chickenn00dle: Agreed, and documented rather than changed. From the method docblock:

> Known limitation, accepted deliberately: `WC_Admin_Settings::save_fields()` writes every field in a section on any save of that tab, not only the fields the publisher touched. A site that reached 9.0 and then saved WooCommerce → Settings → Subscriptions for an unrelated reason therefore has `'no'` rows nobody chose, and this permanently no-ops for it. Widening the test to treat `'no'` as fair game would forfeit the one thing that makes writing another plugin's setting safe, so those sites are left to the settings screen.

Also carried into the Known limits section of the PR description, since that's where the blast radius is stated for a non-engineer reader.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
chickenn00dle: Added both, in 02a4cd3.

`WooCommerce_Subscriptions::PRODUCT_TYPES_ENABLED_OPTION` (`newspack_subscriptions_enabled_product_types`) stores the list of option names Newspack actually wrote, so a site that had one row already set records only the row we created. Alongside it, a `Logger::newspack_log( 'newspack_subscriptions_product_types_enabled', … )` call naming the same list, which puts it in front of Newspack Manager.

It's provenance, not a second gate. The absent-row check already makes the write one-shot, and gating on the marker instead would mean a site where one option was absent never gets the other one considered.

Two tests cover it: the marker names only the row created when the other is an explicit `'no'`, and nothing is recorded or logged when there was nothing to write.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
chickenn00dle: Renamed to `newspack_subscriptions_enable_legacy_product_types`, exactly as suggested. Worth catching now given it's permanent API.

plugins/newspack-plugin/tests/unit-tests/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions-product-types.php
chickenn00dle: Split the method so the guard sits inside the unit under test: `maybe_enable_legacy_product_types()` is the `admin_init` callback and holds the `is_active()` check, `enable_legacy_product_types()` does the work. Two tests added — one asserts the `admin_init` registration and that `plugins_loaded` no longer carries it, the other calls the guarded entry point and asserts nothing is written.

One correction on the mocks: `is_active()` *is* false in the suite. `tests/mocks/wc-mocks.php:663` declares `WC_Subscriptions`, but there's no `WC()` function anywhere in the test bootstrap, so `function_exists( 'WC' )` fails and the guard short-circuits for real. The new test asserts that precondition explicitly so it can't silently stop being true.

Suite is at 2368 tests / 6736 assertions, 0 failures (9 pre-existing skips).

chickenn00dle:
@adekbadek all five threads are addressed in 02a4cd3 and resolved — ready for another look. Thanks for retesting the premise against 9.1.0-rc1 rather than taking the description's word for it; that's what made the two corrections below possible.

**The `plugins_loaded` blocker.** Moved to `admin_init`. There was no deliberate reason to fire earlier, so instead of writing down a justification for the old hook, the docblock records why waiting costs nothing. Confirmed on an env with both rows deleted: 10 anonymous front-end GETs, all 200, both options still absent afterwards. That also settles the filter-scope point from your review body — a theme's `functions.php` now registers in time.

**No record of the write.** Added `newspack_subscriptions_enabled_product_types`, storing the list of option names Newspack actually created, plus a `Logger::newspack_log()` call naming the same list for Newspack Manager. Provenance only, not a second gate: gating on the marker would mean a site where one option was absent never gets the other one considered.

**The `save_fields()` limitation.** Documented rather than coded around, as you suggested, both in the method docblock and in Known limits.

**Filter name.** Now `newspack_subscriptions_enable_legacy_product_types`.

**Untested call site.** Split into `maybe_enable_legacy_product_types()` (the `admin_init` callback, holding the `is_active()` guard) and `enable_legacy_product_types()` (the worker), so the guard sits inside a unit the tests can call. One correction on the mocks: `is_active()` *is* false in the suite — `WC_Subscriptions` is declared, but there's no `WC()` function in the bootstrap, so `function_exists( 'WC' )` fails. The new test asserts that precondition explicitly.

Also took the more robust absence primitive from your summary: a sentinel-object default on `get_option()` instead of `false ===`, so the check stays correct for an option stored as `false` or `''`.

**PR description updated** with your two corrections — that 9.1 doesn't pre-seed on fresh installs, and that Subscriptions' own 9.0 migration only acts where products already exist, so the case this uniquely covers is a site that hasn't created its subscription products yet. The `save_fields()` limitation is in Known limits now too.

Suite is at 2368 tests / 6736 assertions, 0 failures (9 pre-existing skips), PHPCS clean. Manually verified end to end on an isolated env with WCS 9.1.0-rc1 — before the write the product-type selector offers only `simple / variable`, after it offers `simple / variable / subscription / variable-subscription`.

adekbadek (approved):
Moving the write to `admin_init` resolves the one blocking point: the opt-out filter now works from a theme's `functions.php`, the timing is documented on the filter itself, and `is_active()` still gates the call. The three smaller points are addressed too — the sentinel default to `get_option()` replaces the `false ===` test, `newspack_subscriptions_enabled_product_types` plus the log entry make the write answerable per site, and the entry point is now covered by tests.

Two non-blocking notes: `admin_init` also fires on `admin-ajax.php`, including `wp_ajax_nopriv_*` actions, so "never an anonymous front-end hit" is stronger than what the hook guarantees — harmless given the write is one-shot, but the docblock overstates it. And `Logger::newspack_log()` already wraps its third argument under `data`, so passing `[ 'data' => [ 'options' => … ] ]` nests it twice; every other call site in the plugin passes a flat array.

chickenn00dle:
Both notes fixed in b7a830f. Thanks for catching the log shape; that one was live behavior, not just wording.

**`admin_init` is not an authentication boundary.** The docblock now claims only what the hook gives us. It keeps the write off ordinary front-end pageloads, REST and cron, but `admin-ajax.php:45` fires `admin_init` before dispatch, so a `wp_ajax_nopriv_*` action reaches it unauthenticated. It also records why that's harmless here: the write is one-shot and identical either way.

**Log payload flattened** to `[ 'options' => $written ]`. The test was encoding the bug rather than catching it. It asserted `$logged[0][1]['data']['data']['options']`, so it stayed green against the double nesting; it now asserts the flat shape and fails against the old one.

One correction to the note, since it changes where the follow-up goes. `Recaptcha` nests the same way at `class-recaptcha.php:429` and `:449`, passing `'type'` and `'data'` keys that `newspack_log()` wraps again, so this isn't unique to this call site. Out of scope for a hotfix targeting `release`, but worth a separate PR against `main`. Happy to open one.

Suite is at 2368 tests / 6736 assertions, 0 failures (9 pre-existing skips), PHPCS clean on both files.

PR #820, opened 2026-08-06
feat(analytics): report reader access source in GA4 (NPPD-2163)

dkoo:
### All Submissions:

Publishers reverse-engineer how a reader got into a page from three GA4 fields, and the answer is wrong for group seats, institutional access, and metering — `is_subscriber` reads `no` for all of them.

This adds one event parameter, `access_source`, per pageview: the product name (or `group`, `institution`, `domain`, `reader_data`) when an access rule admitted the reader; `metering_eligible` or `gated` when the restriction path blocked them; `no_custom_access_gate` when no custom-access gate applies. The blocked outcome comes from the same check that renders the gate, so the reported value matches what the reader saw. A rule the mapping doesn't know omits the parameter rather than guessing.

The rule-to-label mapping moves out of the ESP contact sync into a shared `Access_Attribution` class, which also fixes one-time purchases syncing an empty Content Access Source. The GA4 dimension provisions on Access Control sites only, and the importable GTM template gains the parameter.

Closes NPPD-2163. Closes NPPD-2162.

1. Define `NEWSPACK_CONTENT_GATES`, connect Site Kit to a GA4 property. Page source shows `access_source` in the gtag config and dataLayer push.
2. View an ungated post. The parameter reads `no_custom_access_gate`.
3. Publish a gate with a subscription rule. As a subscriber, view a gated post. It reads the product name.
4. View the same post logged out. It reads `gated`.
5. Enable metering on the gate. The logged-out view reads `metering_eligible`.
6. Add a second gate whose rules the subscriber fails. Their view now reads `gated` — a block outranks a pass.
7. Mark the post exempt from restriction. It reads `no_custom_access_gate`.
8. Buy a one-time-purchase product granting gate access. The synced contact's Content Access Source names the product (previously empty).
9. With Access Control off, `get_dimensions()` omits `access_source`; with it on, GA4 provisions an "Access Source" dimension.
10. Run `n test-php --group GoogleSiteKit_Access_Source` and `--group Access_Attribution`. All green.

### Other information:

Publishers who already imported the GTM template must follow the README's new upgrade note — a plain re-import leaves the GA4 tag wired to the old bundle and the parameter never arrives. One-time-purchase attribution still runs one order query per product on gated pageviews; tracked as a follow-up. Verified on a live Site Kit-connected dev site: the parameter renders in the gtag config and the dataLayer mirror, an ungated view reads `no_custom_access_gate`, and a logged-in subscriber on a gated post reads the product name. On sites where WooCommerce Memberships is still active, custom-access gates don't restrict, so the parameter is omitted rather than claiming `gated`; the blocked-path values are covered by the test suite. The GTM template import has not been exercised in a real container.


dkoo:
Going to merge this as the missing GA4 data this change will provide going forward is unrecoverable.

PR #823, opened 2026-08-06
fix(plugin-manager): use centralized download URLs for plugin updates

leogermani:
## Summary
- Switches plugin download URLs from direct GitHub releases to the Newspack download center
- Affects all managed plugins: newspack-ads, newspack-blocks, newspack-multibranded-site, newspack-network, newspack-newsletters, newspack-popups, newspack-sponsors, and newspack-listings
- Provides better control over plugin distribution and versioning

**This is a hotfix** targeting the `release` branch.

## Test plan
- [ ] Verify plugin installation works with new download URLs
- [ ] Test plugin updates through the plugin manager
- [ ] Confirm all affected plugins can be downloaded from newspack.com/downloads

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,17 @@
 
 The Newspack monorepo. All product [plugins](plugins), [themes](themes), and shared [packages](packages) live here. It also ships a Docker-based local development environment, so the dependencies needed to run the projects and their tests live inside the container, independent of your local machine.
 
+## Download and Install
+
+**This is a development repository.** If you are looking to download and install Newspack plugins or themes on your WordPress site, you need the **built packages**, not the source code in this repository.
+
+Built packages are available in two places:
+
+- **[Newspack Download Center](https://newspack.com/downloads)** — Easy links to the latest stable releases of all Newspack plugins and themes
+- **[GitHub Releases](https://github.com/Automattic/newspack-workspace/releases)** — All releases, including pre-releases and historical versions
+
+Download the `.zip` file for the plugin or theme you want, then install it on your WordPress site via the WordPress admin interface (Plugins → Add New → Upload Plugin, or Appearance → Themes → Add New → Upload Theme).
+
 ## Getting started
 
 ### Clone this repository
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The Newspack monorepo. All product [plugins](plugins), [themes](themes), and sha
 
 Built packages are available in two places:
 
-- **[Newspack Download Center](https://newspack.com/downloads)** — Easy links to the latest stable releases of all Newspack plugins and themes
+- **[Newspack Download Center](https://help.newspack.com/download-center/)** — Easy links to the latest stable releases
 - **[GitHub Releases](https://github.com/Automattic/newspack-workspace/releases)** — All releases, including pre-releases and historical versions
 
 Download the `.zip` file for the plugin or theme you want, then install it on your WordPress site via the WordPress admin interface (Plugins → Add New → Upload Plugin, or Appearance → Themes → Add New → Upload Theme).
--- a/plugins/newspack-plugin/includes/class-plugin-manager.php
+++ b/plugins/newspack-plugin/includes/class-plugin-manager.php
@@ -104,7 +104,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-ads/releases/latest/download/newspack-ads.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-ads/latest',
 			],
 			'newspack-blocks'             => [
 				'Name'        => \esc_html__( 'Newspack Blocks', 'newspack-plugin' ),
@@ -112,7 +112,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-blocks/releases/latest/download/newspack-blocks.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-blocks/latest',
 			],
 			'newspack-content-converter'  => [
 				'Name'        => \esc_html__( 'Newspack Content Converter', 'newspack-plugin' ),
@@ -129,7 +129,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-multibranded-site/releases/latest/download/newspack-multibranded-site.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-multibranded-site/latest',
 				'Quiet'       => true,
 			],
 			'newspack-network'            => [
@@ -138,7 +138,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-network/releases/latest/download/newspack-network.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-network/latest',
 				'Quiet'       => true,
 			],
 			'newspack-newsletters'        => [
@@ -147,7 +147,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-newsletters/releases/latest/download/newspack-newsletters.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-newsletters/latest',
 			],
 			'newspack-popups'             => [
 				'Name'        => \esc_html__( 'Newspack Campaigns', 'newspack-plugin' ),
@@ -155,7 +155,7 @@ class Plugin_Manager {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-popups/releases/latest/download/newspack-popups.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-popups/latest',
 			],
 			'newspack-sponsors'           => [
 				'Name'        => \esc_html__( 'Newspack Sponsors', 'newspack-plugin' ),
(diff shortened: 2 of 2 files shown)


plugins/newspack-plugin/includes/class-plugin-manager.php
```diff
@@ -104,15 +104,15 @@ public static function get_managed_plugins() {
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-ads/releases/latest/download/newspack-ads.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-ads/latest',
```
dkoo, line 107: **Suggestion:** The trust anchor moves to a hub-computed redirect with no payload verification — The swap moves install sources from fixed GitHub asset paths to redirects the hub computes at request time, and `install_from_url()` performs no checksum, signature, or host verification before unpacking and executing the result on publisher sites (`download_package()` runs with signature checking off). Transport is fine — `wp_safe_remote_get()` validates each redirect hop — but the redirect target is only as trustworthy as the hub's release lookup. The slug allowlist suggested on Automattic/newspack-manager-admin#504 covers the request side; pinning the expected redirect host (github.com asset URLs) on either side would cover the rest. Related: a Batcache-cached 302 on the hub could briefly serve the previous release after a deploy.

README.md
```diff
@@ -11,6 +11,17 @@
 
 The Newspack monorepo. All product [plugins](plugins), [themes](themes), and shared [packages](packages) live here. It also ships a Docker-based local development environment, so the dependencies needed to run the projects and their tests live inside the container, independent of your local machine.
 
+## Download and Install
+
+**This is a development repository.** If you are looking to download and install Newspack plugins or themes on your WordPress site, you need the **built packages**, not the source code in this repository.
+
+Built packages are available in two places:
+
```
dkoo, line 19: **Suggestion:** README links a Download Center index that 404s and will keep 404ing — `https://newspack.com/downloads` returns 404 now, and the companion PR registers only the `/downloads/{slug}/latest` route — no index page — so the link stays 404 after deploy. Either link the URL pattern with an example slug, or hold the bullet until an index page exists. One more accuracy nit in the next bullet: newspack-workspace releases hold monorepo-era releases only — pre-monorepo history lives on the archived per-plugin repos, so "including pre-releases" alone would be accurate ("and historical versions" overstates).

plugins/newspack-plugin/includes/class-plugin-manager.php
dkoo, line 107: **Nit:** "For plugin updates" in the title/body — these values drive installs only — Verified exhaustively: `Download` feeds `install_from_slug()`, the wizards REST controller, and the admin plugins screen — fresh installs. Fleet updates run through newspack-manager's updater, which builds packages from the GitHub API and never reads `Download`. Good news for blast radius (no auto-update regression), but it also means already-installed stale copies stay stale — only fresh installs get current builds. Worth correcting in the title/body since `fix(...)` becomes publisher-facing release notes, and expectations about the fleet "catching up" matter here.

plugins/newspack-plugin/includes/class-plugin-manager.php
```diff
@@ -104,15 +104,15 @@ public static function get_managed_plugins() {
...
 			],
 			'newspack-blocks'             => [
 				'Name'        => \esc_html__( 'Newspack Blocks', 'newspack-plugin' ),
 				'Description' => \esc_html__( 'A collection of blocks for news publishers.', 'newspack-plugin' ),
 				'Author'      => \esc_html__( 'Automattic', 'newspack-plugin' ),
 				'PluginURI'   => \esc_url( 'https://newspack.com' ),
 				'AuthorURI'   => \esc_url( 'https://automattic.com' ),
-				'Download'    => 'https://github.com/Automattic/newspack-blocks/releases/latest/download/newspack-blocks.zip',
+				'Download'    => 'https://newspack.com/downloads/newspack-blocks/latest',
```
dkoo, line 115: **Nit:** esc_url() at definition is inconsistent across the file — no action needed here — The new entries are plain strings, exactly matching what they replace; `distributor` and `everlit` wrap `Download` in `esc_url()`. Consumption is gated by `wp_http_validate_url()` before use, so there's no functional exposure either way — and escaping at definition is arguably the less correct style. Flagging only so the inconsistency is a deliberate choice rather than an accident; this diff didn't make anything worse.

dkoo (approved):
Release risk: ELEVATED—no blockers on this PR itself. The two sequencing risks — the fail-open window where these URLs return 200 with an HTML page the installer accepts, and the single-page release lookup that can't hold slower-moving slugs durably — are owned and tracked on Automattic/newspack-manager-admin#504's review, and this review assumes they land there. What remains here is coupling execution: this PR must land together with #504, and "together" has to mean deploy-and-verify ordering rather than merge timing — the release workflow publishes the newspack-plugin hotfix the moment it merges, on a cadence independent of newspack.com deploys. Hold the merge until the Download Center is deployed, flushed, and verified on newspack.com.

What I verified working: with a live Download Center all 8 slugs 302 to the correct current assets, and the upgrader handles the `.zip`-less redirect cleanly end to end (valid zip, correct install directory derived from the archive). Blast radius of any endpoint failure is fresh installs and wizard-driven activations — updates flow through newspack-manager's GitHub-API updater and are unaffected. The suggestions below are hardening on this side: a legacy-URL fallback so hub-side failures degrade to "installs a slightly older version", the README index link, and a Theme_Manager follow-up for the same staleness this PR fixes.

Gate the release on a scripted post-deploy probe of all eight URLs that asserts each one redirects to a `.zip` and that the fetched body starts with the zip magic bytes — a status-code-only check passes today, since the URLs currently return 200 with an HTML page.

Treat these endpoints as a production dependency with their own recurring uptime check, not just a one-time verification at merge. The install path can't distinguish "endpoint serving a zip" from "endpoint serving a page", so a permalink flush or plugin change months from now would break installs everywhere with no signal at all.

- `plugins/newspack-plugin/includes/class-plugin-manager.php:695` — **Suggestion:** activate( <URL> ) quietly stops accepting these values — narrowed contract, no live caller — `get_plugin_slug()` rejects any URL without `.zip` (`if ( ! stripos( $url, '.zip' ) )`, pinned by `tests/unit-tests/plugin-manager.php:47`), so `activate()` with one of the new `Download` values now returns `newspack_invalid_plugin` where the old `.zip` URLs worked — and even past that gate, the regex would derive the slug as `latest`. Verified no in-repo caller hits this: both places that pass `Download` directly route through `install()`, which handles bare URLs via `wp_http_validate_url()`. But `activate()`'s docblock advertises "slug or URL", so filter consumers or external callers could rely on it. Either teach `get_plugin_slug()` the `/downloads/<slug>/latest` shape or amend the docblock — fine as a follow-up on `main`.
- `plugins/newspack-plugin/includes/class-plugin-manager.php:810` — **Suggestion:** A legacy-URL fallback would turn hub-side failures into "installs a slightly older version" — The legacy per-plugin repos aren't dead — `sync-legacy.yml` still mirrors them, and their `releases/latest` assets resolve today (stale: newspack-popups v3.12.1 vs 3.16.1). A retry in `install_from_slug()` — try the Download Center, fall back to the legacy GitHub URL on `WP_Error` — would degrade every hub-side failure mode (rate limit, window eviction, un-flushed rewrite) from "install fails" to "installs a few versions behind". Cheap insurance for a change that ships to every site and depends on a single hub endpoint with a shared 60/hour unauthenticated GitHub budget.
- `plugins/newspack-plugin/includes/class-theme-manager.php:50` — **Suggestion:** Theme_Manager still installs from the frozen legacy theme repo — the defect class this PR fixes — `$theme_url = "https://github.com/Automattic/newspack-theme/releases/latest/download/{$slug}.zip"` is the exact pattern being replaced here, and it covers `newspack-theme` plus the style packs. The legacy repo's latest is v2.22.4 against the monorepo's 2.24.1, and the setup wizard installs the theme through this path — so the fleet keeps installing a stale theme until this gets the same swap. The Download Center already serves the theme-variant slugs. Reasonable to leave out of this hotfix, but worth a linked follow-up in the PR description so it doesn't get lost.
- `plugins/newspack-plugin/includes/wizards/class-setup-wizard.php:687` — **Suggestion:** The setup wizard discards the install result — a failed newspack-ads install reports success — The Google Ad Manager step writes the service-enabled flag and network-code options, calls `Plugin_Manager::activate( 'newspack-ads' )` without capturing the return, and ends with `rest_ensure_response( [] )` — so when the download fails, the wizard reports success and the publisher finishes setup with ads enabled in options but no ads plugin installed. Pre-existing, but this PR's dependency on a hub endpoint is exactly the kind of change that makes it fire. Capturing the `WP_Error` and returning it would let the wizard surface the failure. **Risk if not addressed:** If that runs across a batch of onboardings, those sites sit half-configured — ad service enabled in options, no ads plugin present — with no error surfaced and nothing written to a log. The gap only appears when someone asks why no ads render, and each site then has to be repaired by hand.
- `plugins/newspack-plugin/includes/class-plugin-manager.php:865` — **Suggestion:** Consider asserting the payload here rather than trusting the status code: `download_package()` only returns an error on a non-200 response, so an HTML page downloads "successfully" and fails several steps later as "Incompatible Archive." with the URL stripped out. Checking the archive signature and including the requested URL in the error message would make the next failure diagnosable from the notice alone.
- `plugins/newspack-plugin/includes/class-plugin-manager.php:872` — **Suggestion:** The `github` check just below governs how the downloaded file is named, and it no longer matches these URLs. It happens to be harmless because every current release archive wraps its files in a `<slug>/` directory, but the URL-string test is now load-bearing on that convention holding — worth keying it on the resolved download instead, or dropping it deliberately with a note.

---

Notes on lines outside this diff (couldn't be attached inline):

**Suggestion:** A legacy-URL fallback would turn hub-side failures into "installs a slightly older version" (`plugins/newspack-plugin/includes/class-plugin-manager.php:810`)
The legacy per-plugin repos aren't dead — `sync-legacy.yml` still mirrors them, and their `releases/latest` assets resolve today (stale: newspack-popups v3.12.1 vs 3.16.1). A retry in `install_from_slug()` — try the Download Center, fall back to the legacy GitHub URL on `WP_Error` — would degrade every hub-side failure mode (rate limit, window eviction, un-flushed rewrite) from "install fails" to "installs a few versions behind". Cheap insurance for a change that ships to every site and depends on a single hub endpoint with a shared 60/hour unauthenticated GitHub budget.

**Suggestion:** Theme_Manager still installs from the frozen legacy theme repo — the defect class this PR fixes (`plugins/newspack-plugin/includes/class-theme-manager.php:50`)
`$theme_url = "https://github.com/Automattic/newspack-theme/releases/latest/download/{$slug}.zip"` is the exact pattern being replaced here, and it covers `newspack-theme` plus the style packs. The legacy repo's latest is v2.22.4 against the monorepo's 2.24.1, and the setup wizard installs the theme through this path — so the fleet keeps installing a stale theme until this gets the same swap. The Download Center already serves the theme-variant slugs. Reasonable to leave out of this hotfix, but worth a linked follow-up in the PR description so it doesn't get lost.

**Suggestion:** activate( <URL> ) quietly stops accepting these values — narrowed contract, no live caller (`plugins/newspack-plugin/includes/class-plugin-manager.php:695`)
`get_plugin_slug()` rejects any URL without `.zip` (`if ( ! stripos( $url, '.zip' ) )`, pinned by `tests/unit-tests/plugin-manager.php:47`), so `activate()` with one of the new `Download` values now returns `newspack_invalid_plugin` where the old `.zip` URLs worked — and even past that gate, the regex would derive the slug as `latest`. Verified no in-repo caller hits this: both places that pass `Download` directly route through `install()`, which handles bare URLs via `wp_http_validate_url()`. But `activate()`'s docblock advertises "slug or URL", so filter consumers or external callers could rely on it. Either teach `get_plugin_slug()` the `/downloads/<slug>/latest` shape or amend the docblock — fine as a follow-up on `main`.

**Suggestion:** The setup wizard discards the install result — a failed newspack-ads install reports success (`plugins/newspack-plugin/includes/wizards/class-setup-wizard.php:687`)
The Google Ad Manager step writes the service-enabled flag and network-code options, calls `Plugin_Manager::activate( 'newspack-ads' )` without capturing the return, and ends with `rest_ensure_response( [] )` — so when the download fails, the wizard reports success and the publisher finishes setup with ads enabled in options but no ads plugin installed. Pre-existing, but this PR's dependency on a hub endpoint is exactly the kind of change that makes it fire. Capturing the `WP_Error` and returning it would let the wizard surface the failure.

README.md
dkoo: We'll publish the Download Center page before these PRs get deployed, so no need to handle the 404s or ship with a placeholder URL.

PR #825, opened 2026-08-07
fix(audience): restructure the pricing rule editor around the price

thomasguillot:
### All Submissions:

The pricing rule editor asked for the price first and showed its consequence last, four sections apart, and several controls did not explain what they set.

Sections are reordered so the rule reads as identity, audience, timing, price, then the Impact Preview directly below the price it reflects. Eligibility moves up beside the product scope. "Lock pricing at purchase" and "Count cycles from" move into Pricing, next to the rows whose meaning they set.

Status, pricing model and pricing disclosure become segmented controls, the last naming both states rather than leaving "off" to be inferred. The pricing model stays visible but disabled once a rule exists, because the server keeps the saved strategy on update. Value carries its unit in the label. Schedule rows are separated by hairlines and read as a table.

The impact table shows a spinner while loading, and says it is a sample whenever the engine reports a partial preview.

Pricing rule editor: the price schedule, with the Impact Preview directly below it

1. Open Audience > Pricing Rules > add a rule, choose Custom. Sections read Rule Details, Eligibility, Scheduling & Behavior, Pricing, Impact Preview.
2. Confirm Pricing Model is a two-option control and its help describes only the selected model.
3. Switch Calculation between its three options. The Value label toggles between the currency symbol and `%`, and the help changes with it.
4. Set Pricing details to Hidden. "Name shown to reader" disappears; set it back to Shown and it returns.
5. Choose Price Schedule. Each row is closed by a hairline, including before "+ Add Row".
6. Add and remove rows. Every row keeps exactly one hairline.
7. Enter a price. A spinner shows while the Impact Preview loads, then the table replaces it.
8. In a multi-cycle rule, resulting prices read `c1 $x → c2 $y`, with the marker explained once below the table.
9. Open an existing rule. The pricing model is still shown, dimmed, and cannot be changed.
10. Open a rule matching fewer products than the preview cap. No "Showing a sample of…" line appears below the table. Open one matching more, and the line appears, counting the rows shown.

### Other information:

Two things the diff will not tell you:

* The sample caption is now the engine's call alone. `Automattic/woocommerce-dynamic-pricing#4` subtracts the products it skips from `total_matching`, so `preview_limited` means a walk the cap cut short. The client-side cap comparison and both `SAMPLE_LIMIT` constants are gone with it.
* The editor sends `application` on save, but the REST controller reads `lock_at_purchase`. Left alone here, but "Lock pricing at purchase" may not persist.


plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-form.tsx
```diff
@@ -47,6 +49,92 @@ interface StepRowState {
...
 }
 
+/**
+ * The toggle shows the model's name alone, so the descriptions the server's full
+ * labels carry ("Flat Adjustment — one price for…") move into the help text. Any
+ * strategy we don't have a short name for falls back to its full label.
+ */
+/**
+ * What each calculation does to the regular price.
+ */
```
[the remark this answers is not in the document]
thomasguillot: Moved onto `strategyShortLabel`, the function it actually describes, and shortened it.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-preview.tsx
```diff
@@ -99,7 +107,7 @@ export default function RulePreview( { body, hasPrice }: RulePreviewProps ) {
 		<div className={ `newspack-pricing-rules__preview${ isLoading ? ' is-loading' : '' }` }>
 			<ImpactStats totalMatching={ preview.total_matching } countLimited={ preview.count_limited } audience={ preview.audience } />
 			<ImpactTable baseline={ preview.sample } segmentGroups={ preview.segment_groups ?? [] } currency={ preview.currency } />
-			{ preview.preview_limited && (
+			{ preview.preview_limited && preview.sample_count >= RULE_PREVIEW_SAMPLE_LIMIT && (
```
miguelpeixe: Since the #4 got merged in `woocommerce-dynamic-pricing` we can revert this gate entirely. Drop `&& preview.sample_count >= RULE_PREVIEW_SAMPLE_LIMIT` here (and the mirror clause in catalog-impact.tsx:91), delete both constants, and let `preview_limited` alone gate the note.

miguelpeixe (changes requested):

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/rule-preview.tsx
thomasguillot: Done in 405117b3. Both constants are gone and `preview_limited` gates the note on its own, in `catalog-impact.tsx` too.

One test flipped rather than being dropped: the 33-of-36 case now expects the caption instead of silence, since a short sample is a cut walk once the skipped products come off the total. The same follow-through landed on #844 and #852, which had each built their own version of the cap check (`sampleNote()` and its `IMPACT_SAMPLE_LIMIT` fallback), so the whole stack now defers to the engine.

`main` is merged down through all four branches, so the conflicts are cleared.

miguelpeixe (approved):

PR #826, opened 2026-08-07
fix(audience): pricing tables in cards, price editing in a drawer

thomasguillot:
### All Submissions:

A Price Schedule is a list of prices, each starting at a billing cycle and running until a later one takes over. The editor put a row of live form controls in a 487px column for each one, so the calculation select truncated to "Percent…" and reader-facing names clipped mid-word.

The list is now a read-only table, and adding or editing a price happens in a drawer where the fields stack at full width. Cells hold short strings, so nothing truncates: a cycle range reads `1 → 3`, a price reads `$3.00`, `Pay 90%` or `$2.00 off`.

A new rule starts with an empty-state card that explains what a schedule does and centers the first Add Price button, rather than one blank row, and the Impact Preview section stays out of the way until a price exists. Prices are ordered by cycle whatever order they arrive in, so a rule saved before this change displays and saves the same way.

Both pricing tables now render inside cards, through a new `TableCard` component in `newspack-components` that pins the DataViews-in-a-card composition from `@wordpress/ui`: a header with a real heading and an actions slot, the table full-bleed, and padded slots around it. The Price Schedule card carries its title and the Add Price button in the header, and Edit and Remove render inline on every row instead of behind a kebab menu. Removing a price asks for confirmation first and confirms the removal in a snackbar. The impact empty states moved onto the same card design.

Price Schedule card with inline Edit and Remove actions

The price drawer open over the schedule card

Remove Price confirmation dialog

Price removed snackbar after confirming

Empty schedule card inviting the first price, with the Impact Preview section absent

1. `pnpm build`, then open **Audience → Pricing Rules** and edit **Win-back ladder**.
2. Read the Pricing section. The Price Schedule sits in a card titled by a real heading, with Add Price in the header. Three rows, no truncated text in any cell.
3. Every row shows Edit and Remove inline, with Remove in red; there is no kebab menu and the actions column has no visible header.
4. Click a cycle range. The drawer opens on that price with four stacked fields.
5. Press Save with the value cleared. The field reports the error and takes focus.
6. Change the value, press Escape. The unsaved-changes dialog opens over the drawer.
7. Choose Discard Changes. The table is unchanged.
8. Click Add Price. The cycle prefills to one past the last price.
9. Save a price at cycle 5. The table reorders it and recomputes the ranges around it.
10. Click Remove on a row. A confirmation dialog opens; Cancel leaves the schedule unchanged, and confirming removes the row, recomputes the remaining ranges, and shows a "Price removed." snackbar.
11. Set Pricing details to Hidden. The reader-name column and drawer field both go.
12. Create a rule, choose a goal, set Pricing model to Price Schedule. The card shows a "No prices yet" empty state with a centered Add Price button, the Impact Preview section is absent until a price exists, and Save reports "Add at least one price."
13. Tab from the table. The cycle link takes focus and opens the drawer on Enter.
14. In the Impact Preview, a sample longer than ten rows collapses behind a See More link inside the card.
15. `pnpm --filter newspack test` and `pnpm --filter newspack-components test`

### Other information:

This stacks on #825. The `Drawer` component it uses is now on `main` via #810, so the diff is this PR's own work. Review the `pricing-rules/` files and `packages/components/src/table-card/`.

`TableCard` is a new, additive export on `newspack-components`; no existing symbol changes. Its README covers the slot contract, the slot-dependent edge behaviour, and what to re-check on a `@wordpress/ui` bump.

---

### Percentage discounts have no calculation type

@miguelpeixe, raising this against `woocommerce-dynamic-pricing` rather than working around it here.

The engine covers money in both directions and percentages in one:

| | Set the price | Take an amount off |
| --- | --- | --- |
| Money | `fixed_price` | `discount_fixed` |
| Percent | `percent_of_base` | none |

`Amount_Calculator::calculate()` computes `percent_of_base` as `base * ( value / 100 )`, so entering 90 means readers pay 90% of the regular price, which is 10% off. Publishers usually reach for the other direction first, and no type expresses it.

A fourth type is additive: a const and a `match` arm in `Amount_Calculator`, an entry in `supported_types()`, and a label in the REST controller. Redefining `percent_of_base` instead would reprice every saved rule, since one storing 90 to mean "pay 90%" would start meaning "90% off".

Until it exists, this table renders the value as `Pay 90%`. The verb keeps it unambiguous next to `discount_fixed`, which renders as `$2.00 off` where "off" does mean off.

The two engine issues on #825 are for the same codebase, and worth taking as one set.


thomasguillot:
### Multi-value rule conditions are dropped on save

@miguelpeixe, separate from the calculation-type note in the description, and sharper: this one loses input a publisher has given.

`Rules_REST_Controller::sanitize_condition_value()` handles `boolean` and `datetime`, then falls through to:

```php
if ( null === $raw || '' === $raw || ! is_scalar( $raw ) ) {
	return null;
}
```

The comment above it states the contract is `boolean|datetime` today. `reader_segment` is a `select` with `multiple: true`, so its value arrives as an array of segment ids. `is_scalar( [ 35 ] )` is false, the value becomes `null`, and `conditions_from_body()` drops the condition from the rule.

The wizard saves through this endpoint, so the effect is the same in the UI and over the API: choose a reader segment, save, and the rule persists without it. No error is surfaced. The rule then matches every reader rather than that segment, so it prices people it was never meant to reach.

Reproduced on a local site with two segments defined: setting `reader_segment` on rule 125 returns 200 with the condition absent from the response, both through the editor and through a direct `PUT`.

Found while trying to get segment columns to appear in the impact preview, so it is incidental to this PR rather than caused by it. Flagging here because it is the same codebase as the other two engine issues on #825 and worth taking as one set.

thomasguillot:
## Review pass on the Price Schedule editor

**Release risk: LOW.** No unresolved blockers, and the only shared contract in the diff is inherited from #810, with no breaking change to it.

The part I most wanted under test was the sorted display copy against the array the form actually saves, since the table sorts its own list while the parent holds `steps`. It holds. The seed sorts, `commit()` re-sorts before handing the list back, removal preserves order, and `submit()` re-maps explicitly, so display order cannot drift from what gets posted. `toPrice()` keeps the table's own columns out of the saved payload. Every `sprintf` carrying a placeholder has a translators comment, and publisher-entered names and values reach the DOM as React children or attributes only, so there is no escaping exposure.

Verified on the branch: 775 tests across 82 suites pass (757 before this pass, 18 added), `lint:js` reports 0 errors, `lint:scss` is clean for `pricing-rules`, and the build exits 0. No browser pass in this round, so the drawer's focus behaviour and the See More collapse are covered by tests but have not been exercised by hand since these changes.

### Blocker, fixed

`schedule-price-drawer.tsx` cleared a field's rejection as it was corrected, which changed the identity of the error state and re-ran the focus effect. With both fields rejected, correcting one sent the caret to the other mid-entry, and the remaining keystrokes landed in the wrong control. Focus now follows a Save rather than any change to the error state, pinned by a test that fails against the previous code.

### Also fixed

- `rule-form.tsx` treated `hasPrice` as true for a schedule with no prices, so the editor posted a preview for an empty schedule instead of showing its no-price state. Starting a rule with no prices made this reachable on every new schedule rule. Raised by Copilot on this PR.
- A value of `1e999` passed validation and serialised as `null`, and negative amounts were accepted. Both now rejected, along with a cycle number past the safe integer range, which reached the engine as `1e+21`.
- `impact-table.tsx` pushed `perPage` and the segment columns through an effect, so a widened sample painted once as a short, un-collapsible table before correcting itself. Both are derived now, as is the collapse reset when a different set of products arrives.
- Removing a price dropped focus to the body; it now moves to Add Price. The schedule table gained a labelled region, See More gained `aria-controls`, and the changed-price highlight went from 2.76:1 to 9.2:1 against its background, which was below the contrast floor.
- Reopening the drawer no longer paints one frame of the previous price's values, and a rejection the publisher walked away from no longer plays the exit animation in red.
- Housekeeping: tokenised the pixel literals, simplified an over-specific `colgroup` selector, dropped an unused export and a stylesheet comment block, corrected an orphaned docblock, and spaced the cycle-marker note away from the Impact Preview description.

### Worth knowing before merge

A schedule saved by the previous editor can hold two prices at the same cycle, because nothing deduplicated them. Repricing one of those no longer requires renumbering it first, but both rows still read identically in the table, since a duplicate cycle collapses to a bare cycle number. If a collision is cleared by deleting the row the engine was applying, subscribers on that rule renew at the surviving price with no visible sign until renewals bill. Worth listing the existing schedule rules and checking for a shared cycle before this reaches publishers.

### Deliberately left

- The schedule table has no visible heading. Its only name is an `aria-label`, and it sits outside the `Grid`, so it is detached from the Pricing section header.
- The paragraph explaining what a billing cycle is went with the inline rows, so a first-time publisher now meets the term inside the drawer rather than before opening it.

Both are copy and layout decisions rather than defects, so they are worth a deliberate call rather than a quick patch.

thomasguillot:
### Final pass

A last read of the changed files after the review above. Two items, both cosmetic, no behaviour change:

- `schedule-price-drawer.tsx` still described the drawer as replacing controls "in the form's column". The table is no longer in that column, it is a full-width sibling of the `Grid`, so the header now matches where the result actually lands. The same wording was corrected in `schedule-prices.tsx` during the review.
- The error class name was rebuilt on every render as a local; it is a constant, so it moved to module scope as `ERROR_CLASS` with a line saying what it does.

775 tests across 82 suites pass, `lint:js` reports 0 errors, `lint:scss` is clean for `pricing-rules`, and the build exits 0.

Nothing else was outstanding. The two items in "Deliberately left" above are still open on purpose, since both are copy and layout calls rather than defects.

thomasguillot:
## Iterative code review: converged

Six review rounds ran against this branch's own diff (`a5b3640c2..HEAD`), each round's findings fixed and re-reviewed. Round 6, at `59380faa8`, returned **zero findings from all three passes** — the exact output is below.

**Release risk: ELEVATED, on one factor only** — the PR adds a new export (`TableCard`) to the shared `newspack-components` package consumed by nine plugins. The addition is purely additive (no existing symbol changes), the risk is understood, and no unresolved finding indicates a defect. Setting the contract surface aside, the review reports low risk: no blockers, no findings, 778 plugin + 243 components tests passing, and the behaviour changes verified live (both table surfaces, the drawer flows, removal confirmation, empty states, and focus management).

<details>
<summary>Review pass 1 of 3 (deep) — no findings</summary>

All three round-5 changes check out, and I found nothing net-new.

**Overview**

The only commits after round 5's baseline are `59380faa8` (calc-copy fallback + two README lines) and `7b8346a35` (the `@wordpress/ui` pin + README truing-up); `git diff 59380faa8..HEAD` is empty, so HEAD is that commit. All three targeted fixes are correct, and each one is independently verifiable against the code it describes rather than just plausible. Sweeping the surrounding code for anything the fixes could have introduced or left inconsistent turned up nothing.

**Verification of the three fixes**

(a) `valueLabel` unit-less fallback — `calc-copy.ts:27-37`. The three known types keep their units; anything else returns bare `Value`. This now matches the sibling helpers exactly: `calcTypeHelp` falls back to the server label, `valueHelp` falls back to `''`, and `priceSummary` in `schedule-format.ts:52-66` already refused to format an unknown type as currency for the same stated reason. Both consumers (`schedule-price-drawer.tsx:175`, `rule-form.tsx:738`) pass a raw server-supplied `calc_type`, and the drawer deliberately preserves an unrecognized type as a select option (`schedule-price-drawer.tsx:134-140`), so the branch is genuinely reachable. The engine's vocabulary lives outside this repo, which is exactly why the defensive fallback is the right call. The added test at `calc-copy.test.js:25-27` uses `discount_percent`, the same hypothetical unknown type `schedule-prices.test.jsx:295` already uses — consistent with the suite's convention.

(b) README header-edge mechanics — `packages/components/src/table-card/README.md:69-72`. I read the upstream stylesheet at `@wordpress/ui@0.16.0`'s `src/card/style.module.css:33-36`. The rule is `.header + .content { padding-block-start: 0; margin-block-start: calc(var(--wp-ui-card-header-content-gap) - var(--wp-ui-card-padding)); }`, and the header retains its full `padding-block-end` because it is `:first-child` (the zeroing rule at line 27 only hits middle elements). Net inset = header padding + compensating margin = the `header-content-gap` token. The README's new wording describes that mechanism accurately, including why the token — not the header's padding — is the number that matters. The related claim that the table isn't top-flush when a header is present also holds: the flush rule at line 49 requires `.content:first-child`, which a header defeats.

(c) "the one product surface with a header" — `packages/components/src/table-card/README.md:96`. The demo section added in `componentsDemo/index.js` does render a `TableCard` with both `title` and `actions`, so the unqualified "one surface with a header" was wrong and "one product surface" is the correct narrowing; the checklist names the demo separately on the next line.

Bonus, `7b8346a35` — the `@wordpress/ui` pin is right, not just harmless. `impact-empty.tsx:12` imports `Card` directly from the plugin's own source, so the dependency was previously a phantom relying on hoisting from `packages/components`. The declared range `^0.16.0` matches `packages/components/package.json:29`, and the lockfile resolves a single `0.16.0` copy, so no duplicate instance. `packages/scripts/config/getWebpackConfig.js:7` force-bundles `@wordpress/ui` rather than externalizing it to a `wp.ui` global, so the direct import is safe at runtime — the usual "builds fine, undefined live" trap does not apply here.

**Findings**

None.

One observation not worth a finding: the new `__( 'Value', 'newspack-plugin' )` is a bare, context-free translatable string. It is currently the only `'Value'` msgid in the plugin, so there is no collision, and the branch only fires for a calculation type this build does not ship — adding `_x` context would be churn against a string no publisher can currently reach.

</details>

<details>
<summary>Review pass 2 of 3 (WordPress expert) — no findings</summary>

**Overview**

I verified the three round-6 changes and found them correct and internally consistent. `valueLabel()` returning a unit-less `'Value'` for an unrecognized calculation type now matches the stance the sibling helpers already took — `valueHelp()` returns `''` and `priceSummary()` refuses to format an unknown type as currency — so the label no longer asserts a currency unit the build cannot vouch for; no other test or caller depended on the old currency fallback (`Value ($)` assertions in `schedule-price-drawer.test.jsx`, `schedule-prices.test.jsx` and `rule-form-schedule-prices.test.jsx` all pass known types). Both README edits check out against the code: the `^0.16.0` pin, the 24px/`$grid-unit-30` column padding in `style.scss`, and the consumer list (Price Schedule is the only product surface passing `title`/`actions`; the impact table is headerless and reached from both `rule-preview.tsx` inside the rule editor and `catalog-impact.tsx` on the list page).

Findings: **none.**

Two notes, neither a finding:

- The README's parenthetical about upstream's `.header + .content` rule is the one claim I could not check against source in this pass; the observable statement it wraps (header present → table inset from the top rather than flush) is consistent with `TableCard`'s structure, where `Card.FullBleed` is the first child of `Card.Content` only when the header is absent.
- Out of scope and pre-existing, recorded only so it is not mistaken for something round 6 introduced: `schedule-price-drawer.tsx:137-140` appends a stored `calc_type` that the server vocabulary no longer offers so the select shows the real value, while the flat model's inline select at `rule-form.tsx:733` maps `vocab.calc_types` straight through with no such guard. Now that both surfaces share `calc-copy.ts`, the asymmetry is more visible, but the flat-form behaviour predates this branch.

</details>

<details>
<summary>Review pass 3 of 3 (Codex) — no findings</summary>

Target: branch diff against a5b3640c2

No actionable correctness issues were identified in the changed code. The components TypeScript check passed; Jest execution was blocked by the read-only sandbox's temporary-file restrictions.

</details>

thomasguillot:
## Final pass after the converged review

One fix, from the pre-existing asymmetry noted in the review above (`bc913f944`):

- **The flat model's Calculation select now keeps a retired calculation type visible.** The price drawer already appended a stored `calc_type` the server vocabulary no longer offers, so the control shows what the price actually is; the flat form's inline select mapped the vocabulary straight through, so a legacy rule holding a retired type would have displayed the first option while the stored value rode along underneath. The flat select now uses the same guard. 778 plugin + 243 components tests pass; lint and build clean.

Two review observations were deliberately left alone:

- Adding `_x()` context to the new `Value` fallback label: it is the plugin's only `Value` msgid and the branch only fires for a calculation type this build does not ship, so context would be churn against an unreachable string.
- The drawer's sub-second reopen-during-exit focus edge: unchanged pre-existing drawer behaviour, lands focus on a real control either way.

miguelpeixe (approved):

PR #831, opened 2026-08-08
fix(campaigns): keep the prompt CTA legible on the classic theme

thomasguillot:
### All Submissions:

On the classic theme, the Donate button in a Contextual Prompt rendered its label in the card's text colour rather than the button's own, so at the seeded design a reader saw near-black type on the theme's dark grey button. The editor showed it correctly, which made the two surfaces disagree about what publishes.

The card sets a text colour so its copy stays legible on the light panel whatever the theme uses for body text. The classic theme, in turn, has every link inside a colour-carrying block inherit that block's colour (`.entry-content .has-text-color a`), at a specificity above its own button rule — so the card's colour reached the CTA and beat the button's. The editor canvas has no `.entry-content` ancestor, so the rule never applied there.

This restates the button's colour from the feature's own classic-theme stylesheet, alongside the layout CSS the card already ships. It sits above the rule that would have the label inherit and below anything a publisher sets on the button, which core emits as `!important` for a palette colour and inline for a custom one, so Edit Design keeps full control.

Only the button the plugin ships is restated. The colour being restated is the theme's pair for the theme's own button background, so it is the wrong answer for a background a publisher chose — core marks those with `has-background` on the link, and they are left alone. Outline buttons are left alone too, for a different reason: the theme already colours them above the inheriting rule. A `newspack_contextual_prompts_button_color_css` filter runs whether or not the active theme is one the plugin compensates, so a fork under its own slug can opt in and a child theme can opt out.

Left: before. Right: after.

Contextual Prompt card before and after: Donate label dark, then white

1. On a site running the classic theme, turn on Contextual Prompts in Audience → Campaigns and add a prompt to a story.
2. View the story on the front end. The Donate label is white on the grey button, and the card's copy stays dark.
3. Open the story in the editor and compare. Both surfaces show the same button.
4. In Audience → Campaigns → Contextual Prompts, choose Edit Design, give the button a text colour of Primary, and save. The front end shows Primary, so a publisher's own colour still wins.
5. Back in Edit Design, clear that text colour and give the button a White background instead. The front end shows a dark label on the white button, not a white one.
6. Choose Reset Design. The label returns to white on the theme's grey button, on both surfaces.
7. Switch to the block theme and repeat step 2. Nothing is delivered there, and the button is unchanged.

### Other information:

* [x] Have you successfully run tests with your changes locally? The newspack-popups suite passes at 379 tests, and PHPCS is clean on both files.

<details>
<summary>For AI</summary>

`Newspack_Popups_Contextual_Prompt_Render::get_button_color_css()` in `plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php` emits:

```css
.newspack-contextual-prompt .wp-block-button:not(.is-style-outline) > .wp-block-button__link:not(.is-style-outline):not(.has-background){color:var(--newspack-theme-color-against-secondary)}
```

Both hooks read `get_delivered_css()`, which composes `get_layout_css() . get_button_color_css()`, so the front end (`enqueue_layout_styles()`) and the canvas (`add_editor_layout_styles()`) cannot drift apart through an edit to one of them.

The gate is `get_template() === 'newspack-theme'`, which covers the base theme and its five child themes, rather than the `wp_is_block_theme()` check `get_layout_css()` uses. The variable named in the declaration is that theme's. On any other classic theme it would be undefined, the declaration would drop as invalid at computed-value time, and `color` being an inherited property would leave the label taking the card's colour — introducing the bug on a theme that never had it. The filter is applied outside that gate so a theme the slug check misses can still opt in.

Specificity is 0-6-0 against the theme rule's 0-2-1. Core's preset colour classes carry `!important` and a custom colour lands inline, so both publisher paths still outrank it.

The theme rule is at `themes/newspack-theme/newspack-theme/sass/blocks/_blocks.scss:1711-1714`. Fixing it there — adding a `:not()` for buttons, the way the cover-block rule at line 484 already does — would also cover the `core/file` download button and any other coloured block containing a link-button, but it changes a shared theme file for every classic-theme site, so it stayed out of this PR.

Outline buttons are excluded, and the two surfaces still disagree there: the front end colours an outline label at `--newspack-theme-color-secondary-against-white` (`_blocks.scss:795-800`, 0-4-0), while the canvas rule in `style-editor-base.scss:512-521` declares no colour, so the label falls through to the filled one. That is the theme's to reconcile.

Seeding and resetting write the identical string from `build_pattern_content()`, verified byte-for-byte on a local env, so this was never reset-specific: the seeded card carries `has-text-color has-dark-gray-color` and the front end took the inherited colour from the first insert.

</details>

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php
@@ -24,7 +24,8 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Newspack_Popups_Contextual_Prompt_Render {
 	/**
-	 * Handle the classic-theme layout CSS is delivered on.
+	 * Handle the card's own inline CSS is delivered on — the classic-theme layout
+	 * rules, and on the Newspack classic theme its CTA's colour as well.
 	 */
 	const LAYOUT_STYLE_HANDLE = 'newspack-popups-contextual-prompt-layout';
 
@@ -110,25 +111,54 @@ final class Newspack_Popups_Contextual_Prompt_Render {
 	 * core emits as `!important` for a palette colour and inline for a custom
 	 * one.
 	 *
-	 * Outline buttons are left to inherit: their label sits on the card, not on a
-	 * filled background, so the card's text colour is the right one for it.
+	 * Only the button this plugin ships is restated: the colour below is the
+	 * theme's pair for the theme's button background, so it is the wrong answer
+	 * for a background a publisher chose. Core marks those on the link itself,
+	 * for a palette colour and a custom one alike, and they are left alone.
+	 *
+	 * Outline buttons are left alone too, for a different reason: the theme
+	 * already colours them, above the rule that would have them inherit. The
+	 * editor disagrees there — its own outline rule declares no colour, so the
+	 * label falls through to the filled one — but that is the theme's to
+	 * reconcile, not this card's.
 	 *
 	 * Named for the theme it compensates rather than for classic themes at large:
 	 * the variable below is that theme's, and on any other one an undefined
 	 * variable would drop the declaration and leave the label inheriting the very
-	 * colour this removes.
+	 * colour this removes. The filter is the way onto a theme this cannot name —
+	 * a fork under its own slug, or a child theme that would rather this stopped.
 	 *
 	 * @return string CSS, or an empty string off the Newspack classic theme.
 	 */
 	public static function get_button_color_css() {
-		if ( 'newspack-theme' !== get_template() ) {
-			return '';
+		$css = '';
+
+		if ( 'newspack-theme' === get_template() ) {
+			$filled = '.' . Newspack_Popups_Contextual_Prompt_Pattern::MARKER_CLASS
+				. ' .wp-block-button:not(.is-style-outline) > .wp-block-button__link:not(.is-style-outline):not(.has-background)';
+			$css    = $filled . '{color:var(--newspack-theme-color-against-secondary)}';
 		}
 
-		$filled = '.' . Newspack_Popups_Contextual_Prompt_Pattern::MARKER_CLASS
-			. ' .wp-block-button:not(.is-style-outline) > .wp-block-button__link:not(.is-style-outline)';
+		/**
+		 * Filters the CSS restating the Contextual Prompt CTA's label colour.
+		 *
+		 * Runs whether or not this theme is one the plugin compensates, so a theme
+		 * it does not name can opt in as readily as one it does can opt out.
+		 *
+		 * @param string $css The CSS, or an empty string.
+		 */
+		return (string) apply_filters( 'newspack_contextual_prompts_button_color_css', $css );
+	}
 
-		return $filled . '{color:var(--newspack-theme-color-against-secondary)}';
+	/**
+	 * Everything the card is delivered, whichever surface is asking. Both hooks
+	 * read this rather than composing their own, so neither can come to disagree
+	 * with the other through an edit to one of them.
+	 *
+	 * @return string
+	 */
+	private static function get_delivered_css() {
+		return self::get_layout_css() . self::get_button_color_css();
 	}
 
 	/**
@@ -136,7 +166,7 @@ final class Newspack_Popups_Contextual_Prompt_Render {
 	 * prints its styles and carries no file to version.
 	 */
 	public static function enqueue_layout_styles() {
-		$css = self::get_layout_css() . self::get_button_color_css();
+		$css = self::get_delivered_css();
 		if ( '' === $css ) {
 			return;
 		}
@@ -154,7 +184,7 @@ final class Newspack_Popups_Contextual_Prompt_Render {
 	 * @return array
 	 */
 	public static function add_editor_layout_styles( $settings ) {
-		$css = self::get_layout_css() . self::get_button_color_css();
+		$css = self::get_delivered_css();
 		if ( '' === $css ) {
 			return $settings;
 		}
--- a/plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php
@@ -24,8 +24,7 @@ defined( 'ABSPATH' ) || exit;
  */
 final class Newspack_Popups_Contextual_Prompt_Render {
 	/**
-	 * Handle the card's own inline CSS is delivered on — the classic-theme layout
-	 * rules, and on the Newspack classic theme its CTA's colour as well.
+	 * Handle the card's own inline CSS is delivered on.
 	 */
 	const LAYOUT_STYLE_HANDLE = 'newspack-popups-contextual-prompt-layout';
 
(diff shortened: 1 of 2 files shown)


thomasguillot:
Release risk: LOW — no blockers, no contract files touched, and the front-end behaviour was exercised directly. The root-cause analysis holds up: the specificity arithmetic is right, `get_template()` is the correct call for the five style variations that share this template, and the custom property the rule names is declared unconditionally at `:root`, so the declaration cannot go invalid and silently reinstate the inheritance.

Two things are worth settling before merge. The rule currently fires on any filled button in the card, including one whose background a publisher changed in Edit Design, where restating a colour paired with the theme's own background can make the label unreadable. And the delivery assertions do not yet have teeth: under the test environment's classic theme the new method returns an empty string, so the editor assertion passes whether or not either path concatenates it, and the front-end path has no coverage at all.

The rest are small: a docblock whose outline rationale does not match what either stylesheet actually does, the duplicated composition at the two delivery points, and a handle whose docblock no longer describes everything it carries.

- `plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:27` — **Nit:** The handle's docblock is narrower than what it carries — "Handle the classic-theme layout CSS is delivered on" no longer describes everything on this handle: on the Newspack classic theme it also carries a colour rule with a different gate. The same drift touches the two method names and the "Front end: an inline stylesheet of its own" docblock, but renaming is probably not worth the churn — the constant's docblock saying the handle carries the prompt's inline CSS generally would cover it.

---

### Findings

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:113`**

**Suggestion:** The outline rationale in the docblock does not match either stylesheet — "Outline buttons are left to inherit … so the card's text colour is the right one for it" describes neither surface. On the front end `_blocks.scss:795-800` colours an outline label at `--newspack-theme-color-secondary-against-white` with specificity 0-4-0, which already outranks the inherit rule, so an outline CTA never takes the card's colour. In the canvas the matching block in `style-editor-base.scss:512-521` has an empty `:not(.has-text-color)` body, so an outline label falls through to white on the transparent light card. Excluding outline is still the right call — it preserves current behaviour and keeps the change minimal — but the comment should say the theme already colours outline buttons itself, and that outline is where the two surfaces still disagree. As written it reads as a settled case rather than an untouched one.

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:124`**

**Suggestion:** The theme gate has no escape hatch — The hard-coded slug is the right shape of check, but a publisher on a fork of the classic theme under a different slug hits the original bug with no way to opt in, and one whose child theme restyles the button has no way to opt out. The plugin already establishes the convention for this with `newspack_contextual_prompts_use_donate_block`, so running the emitted CSS through a `newspack_`-prefixed filter would follow its own pattern rather than introducing a new one. Worth weighing against the cost of a public hook that then has to be kept.

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:129`**

**Suggestion:** Narrow the rule to the button this plugin ships — The rule fires on every filled button in the card, including one whose background a publisher changed. `BLOCK_LOCK` covers only move and remove, so the button's colour panel is reachable in the pattern editor, and Edit Design is a first-class flow. The colour being restated is only correct while the button's background is the theme's secondary — the pairing the theme itself guards with `:not(.has-background)` on both surfaces. Core adds `has-background` to the anchor for both preset and custom backgrounds, so adding `:not(.has-background)` to the link selector narrows this to the button whose background the plugin actually knows.

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:139`**

**Suggestion:** The front-end delivery path has no coverage — `enqueue_layout_styles()` is untested, before and after this change, and it is the path publishers actually see and the one that regressed. A test that filters `template` to `newspack-theme`, calls the enqueue callback, and asserts `wp_styles()->get_data( self::LAYOUT_STYLE_HANDLE, 'after' )` carries the button rule would pin the concatenation, the handle wiring, and the early return in one go. That matters more now the shared handle carries two independently gated blobs, and the early return is only correct because the two gates happen to be mutually exclusive.

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:157`**

**Suggestion:** The composition is duplicated at both delivery points — `get_layout_css() . get_button_color_css()` appears at both hooks, and the invariant it serves — that the editor shows what a reader gets — is not pinned anywhere. A single private accessor returning the delivered CSS, called from both hooks and asserted once, would make drift impossible rather than merely currently absent.

**`plugins/newspack-popups/tests/test-contextual-prompt-render.php:1206`**

**Suggestion:** The editor delivery assertion is vacuous in the test environment — `switch_to_theme_family( false )` picks whichever classic theme the test environment ships, which is not `newspack-theme`, so `get_button_color_css()` returns an empty string and `$delivered` is identical to `$css`. The assertion then passes whether or not `add_editor_layout_styles()` concatenates the new CSS at all — deleting the concatenation from the render class leaves this test green. Wrapping these lines in the same `template` filter the new test uses would make `$delivered` genuinely longer and give the assertion teeth.

**`plugins/newspack-popups/includes/class-newspack-popups-contextual-prompt-render.php:123`**

**Nit:** No return type on the new method — The plugin pins PHP 8.3 and the docblock already asserts `@return string`. Every other method in this file is untyped, so adding one here alone is inconsistent either way — worth making the choice deliberate rather than inherited.

**`plugins/newspack-popups/tests/test-contextual-prompt-render.php:1239`**

**Nit:** Asymmetric filter cleanup between the two halves of the test — The first half captures the value and removes the filter before asserting; the second asserts between `add_filter` and `remove_filter`, so a failure there leaves `template` filtered for the rest of the method. `WP_UnitTestCase_Base` restores `$wp_filter` in `tear_down()`, so the blast radius is one test, but the asymmetry is gratuitous. Separately, filtering `template` desynchronises `get_template()` from `get_stylesheet()` and `wp_get_theme()` — fine while the guard reads nothing else, and a one-line note saying the environment has no `newspack-theme` installed would stop someone rewriting it into a `switch_theme()` that then skips.

thomasguillot:
Final pass — every finding above is addressed in `d35f43205`.

**Narrowed the rule to the button the plugin ships.** The selector now carries `:not(.has-background)` on the link. The colour being restated is the theme's pair for the theme's own button background, so it was the wrong answer for a background a publisher chose in Edit Design. Verified on an isolated env: with a White background on the button and no label colour set, the front end renders `rgb(17, 17, 17)` rather than forcing white on white.

**Corrected the outline rationale.** The docblock claimed outline labels inherit the card's colour. They don't: the front end colours them at `--newspack-theme-color-secondary-against-white` (`_blocks.scss:795-800`, specificity 0-4-0), above the inheriting rule. The canvas rule at `style-editor-base.scss:512-521` declares no colour, so a label there falls through to the filled one. Excluding outline is still right, but it's an untouched case rather than a settled one, and the comment now says so — along with the fact that the two surfaces still disagree there.

**Gave the delivery assertions teeth.** Both hooks now read a single `get_delivered_css()`, and a new test pins it under `template` filtered to `newspack-theme`, asserting the editor settings and reading the front-end path back off `wp_styles()->get_data( LAYOUT_STYLE_HANDLE, 'after' )`. I checked it fails for the right reason: dropping the concatenation turns that test red, where before it stayed green. The old assertion in `test_the_layout_css_is_classic_only` no longer claims to cover the concatenation.

**Covered the front-end path**, which had no test before this PR and is the surface that regressed.

**Added a filter.** `newspack_contextual_prompts_button_color_css` runs whether or not the active theme is one the plugin compensates, so a fork under its own slug can opt in and a child theme that restyles the button can opt out. It follows the plugin's existing convention rather than introducing a new one.

**Nits:** the handle's docblock now describes everything it carries rather than layout alone; the new tests share an `under_theme()` helper that removes the filter in a `finally`, so cleanup is symmetric and a failure can't leak `template` into the rest of the method, and the docblock records that no Newspack theme is installed in the test environment — which is why the theme is faulted in rather than switched to; and the PR description's line reference for the theme rule is corrected to 1711-1714.

On the return type: left untyped, deliberately. Every method in this file is untyped, so adding one to this method alone would be the inconsistency.

The newspack-popups suite passes at 379 tests (up from 375), PHPCS is clean on both files, and the front end still renders the shipped button white on grey.

PR #832, opened 2026-08-09
fix(content-gate): keep withheld blocks out of generated excerpts

jason10lee:
### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? — #813 is the sibling change for the same subsystem; the two do not overlap in the code they touch.

A post's teaser now shows the same thing its article page shows the same reader. Blocks the content gate withholds no longer contribute to auto-generated excerpts on the front page, archives, search, feeds, or the REST API.

Excerpt generation flattens wrapper blocks before `render_block` runs, so the per-block access check never ran on them. This adds `Block_Visibility::strip_blocks_hidden_from_public()`, which removes withheld blocks from content at any nesting depth, and calls it at each of the three places an excerpt gets built: the new `Content_Gate_Excerpt` seam in newspack-plugin, and the two builders in newspack-blocks and newspack-listings that recreate core's trimming themselves.

Three points a reviewer will want to weigh:

**The decision is evaluated against the anonymous reader, whoever is asking.** Excerpt output therefore does not vary by reader, which is deliberate rather than a simplification. `Newspack_Blocks_Caching` keys cached block markup with no user dimension and caches for anonymous visitors and logged-in readers alike, so reader-varying excerpt output would be written to a shared key and served across readers. One shipped access rule can still vary by request context; the docblock says so.

**Trimming stays in core.** The newspack-plugin seam hands `wp_trim_excerpt()` a cloned post carrying sanitized content instead of reimplementing the trim. Two copies of that logic already exist in this monorepo and both have drifted from core; a third would drift too.

**Three plugins, one branch,** per the monorepo convention for cross-plugin changes. The calls in newspack-blocks and newspack-listings are guarded on both `class_exists` and `method_exists`, because those plugins version and update independently of newspack-plugin and the helper is new here.

Two notes on process. This targets `release` rather than `main` because it follows the hotfix flow; it reaches `main` through the usual post-release back-merge. And there is no GitHub issue to close — this is tracked in Linear as NPPM-3090, along with related hardening tracked separately.

1. On an env running this branch, publish a public post containing a `core/group` block with access control set to registration-only, wrapping a unique marker string. Add an ungated paragraph carrying a second marker.
2. Logged out, load the post's permalink and note which markers appear. This is your reference for every step below.
3. Logged out, load the home page or an archive listing. The post's excerpt shows the same markers as step 2.
4. Logged out, read the post through the REST API with `?_fields=excerpt`, and exercise a Homepage Posts block's "load more". Both agree with step 2.
5. Repeat with the gated group nested inside a `core/columns` block, which exercises a different path through core's excerpt builder.
6. Repeat on a listing archive, which has its own excerpt builder.
7. Log in as an editor and open the posts list. Excerpts there now match too — see the behavior note below.

These envs run batcache, so add a cache-buster query param to every request or you will read a stale response.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — Six cases in `newspack-plugin`, including one asserting that the excerpt withholds exactly what the front end withholds across six gate configurations, and one wiring test in `newspack-blocks`. Both were mutation-tested: reverting the source makes them fail.
* [x] Have you successfully run tests with your changes locally? — newspack-plugin 2373 tests, 0 failures, 9 pre-existing environment skips. newspack-blocks 126 tests, 272 assertions. `phpcs` clean against the root ruleset on all seven lintable changed files.

Three things worth knowing before merge:

* **Editors see stripped excerpts too.** There is deliberately no editor bypass on the excerpt path — adding one would reopen the surface this closes. The admin posts-list excerpt column and editor previews now withhold gated blocks. Worth a line in release notes.
* **`Content_Gate_Excerpt::init()` replaces core's `wp_trim_excerpt` callback** on `get_the_excerpt` with its own. Nothing in this monorepo depends on that callback being registered, but a publisher's custom plugin that checks for it would see a difference.
* **`plugins/newspack-blocks/tests/*` is excluded from the root `phpcs.xml`,** so the new stub and test in that directory are not linted by CI. That is pre-existing, not introduced here.


plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,61 @@
...
+
+		if ( ! $resolved instanceof \WP_Post ) {
+			return wp_trim_excerpt( $text, $post );
+		}
+
+		// A manually written excerpt is the author's own words; core returns it
+		// untouched and so do we.
+		if ( '' !== trim( (string) $text ) ) {
+			return wp_trim_excerpt( $text, $resolved );
+		}
```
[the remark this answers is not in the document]
jason10lee: Fixed in `6a735be5f`. Detection now reads `$resolved->post_excerpt` rather than `$text`.

Worth noting the fix needed a second half: `wp_trim_excerpt()` returns `$text` unchanged whenever it is non-empty (`wp-includes/formatting.php:3961`), so switching the detection alone would still have handed back an excerpt an earlier filter had produced — the sanitized clone would never have been read. The auto branch now passes `''` so core rebuilds from the sanitized post.

Added `test_prepopulated_text_does_not_bypass_sanitization`, which registers a filter at priority 9 returning unsanitized text and asserts the excerpt is still rebuilt. It fails against the previous detection.

plugins/newspack-blocks/tests/bootstrap.php
```diff
@@ -27,11 +27,27 @@ function newspack_blocks_manually_load_plugin() {
...
+$autoloader_paths = [
+	__DIR__ . '/../vendor/autoload.php', // plugin vendor
+	dirname( dirname( dirname( __DIR__ ) ) ) . '/vendor/autoload.php', // monorepo root
+];
+foreach ( $autoloader_paths as $autoloader_path ) {
+	if ( file_exists( $autoloader_path ) ) {
+		require_once $autoloader_path;
+		break;
+	}
+}
```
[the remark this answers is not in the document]
jason10lee: Fixed in `103ea0eda` — the bootstrap now writes a message naming both candidate paths and exits non-zero when neither autoloader is found, instead of continuing into downstream fatals.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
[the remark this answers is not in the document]
jason10lee: Following up, because the fix above turned out to be incomplete: a later review pass found the manual-excerpt branch still handed core the unsanitized post, so a blanked `$text` on a post that has a manual excerpt would rebuild from `post_content` and put the withheld block back in the excerpt.

Fixed in `f8b696f9c` by routing both branches through the sanitized clone, with `test_blanked_text_on_a_post_with_a_manual_excerpt_does_not_leak` covering it — it fails against the previous version.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		// Core registers wp_trim_excerpt() here at priority 10. Replace it with a
+		// wrapper that hands core the same work over sanitized content, rather than
+		// reimplementing the trimming: two copies of that logic already exist in this
+		// monorepo and both have drifted from core.
+		$removed = remove_filter( 'get_the_excerpt', 'wp_trim_excerpt', 10 );
```
dkoo, line 25: **Blocker:** The core-filter replacement engages on every site, including those with content gates off — `Content_Gate::init()` includes this file unconditionally (`class-content-gate.php:110`), with no `is_newspack_feature_enabled()` check. So core's `wp_trim_excerpt` is displaced, and the `wp_trim_excerpt( '', $sanitized )` branch at line 69 discards whatever `$text` earlier filters produced, on every Newspack install — including the ones where `NEWSPACK_CONTENT_GATES` is undefined and block visibility cannot exist at all.

The stripping itself is correctly inert on those sites (the early-out at `class-block-visibility.php:152` sees no access-control attributes). What is not inert is the contract change: core returns `$text` untouched when it is non-empty, and this does not. Any plugin hooking `get_the_excerpt` below priority 10 to supply an excerpt now has its value discarded site-wide.

I confirmed this in an isolated env with the feature switched off.

**To reproduce**
1. In an env where `NEWSPACK_CONTENT_GATES` is undefined, confirm the feature is off and core's callback is gone:
   `wp eval 'var_dump( defined("NEWSPACK_CONTENT_GATES"), has_filter("get_the_excerpt","wp_trim_excerpt") );'`
   → `bool(false)` and `bool(false)` — feature off, core callback displaced.
2. Publish a post whose content has no access-control attributes at all.
3. Register a filter at a priority below 10 that returns its own excerpt text, then read the excerpt:
   `add_filter( 'get_the_excerpt', fn() => 'THIRDPARTYEXCERPT', 6 ); echo get_the_excerpt( $id );`
4. Observed: the post's core-trimmed content. Expected (and what core does): `THIRDPARTYEXCERPT`.

This targets `release`, so it reaches every publisher on the next patch rather than riding a feature flag. Scoping the seam so it is a no-op for posts that don't use the gate would shrink the blast radius to the posts this fix is actually about.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Build the excerpt from content with non-public blocks removed.
+	 *
+	 * @param string   $text The post excerpt, empty when auto-generating.
+	 * @param \WP_Post $post The post.
+	 * @return string
+	 */
+	public static function filter_get_the_excerpt( $text, $post = null ) {
+		$resolved = $post instanceof \WP_Post ? $post : get_post( $post );
```
dkoo, line 44: **Blocker:** The sanitized clone is discarded when the incoming post is not in raw form — The comment at line 46 states that core is only ever handed the sanitized clone. That holds only while `$resolved->filter === 'raw'`, and nothing here establishes it.

`wp_trim_excerpt()` opens with `get_post( $post )`, which ends in `$_post->filter( 'raw' )`. `WP_Post::filter()` returns `$this` when the property already matches — but for any other value it returns `self::get_instance( $this->ID )`, a fresh read that discards the clone and its sanitized content. The `$post instanceof \WP_Post ? $post : get_post( $post )` shortcut on line 40 is what lets a non-raw object through: it skips the normalization `get_post()` would have done, and `clone` copies the `filter` property along with everything else.

No caller in this repo triggers it today — loop posts and `get_the_excerpt()` both normalize to raw. But `class-newspack-blocks-api.php:246` already fires this filter by hand with a post object it supplies, and third-party code can do the same with a display-form post. The failure is total and silent rather than partial.

**To reproduce**
1. Publish a post with a gated `core/group` wrapping a marker, and confirm the teaser withholds it: `wp eval 'wp_set_current_user(0); echo get_the_excerpt( $id );'`
2. Read the same post in display form and pass it to the filter directly:
   `wp eval 'wp_set_current_user(0); $p = get_post( $id, OBJECT, "display" ); echo apply_filters( "get_the_excerpt", $p->post_excerpt, $p );'`
3. Observed: the marker is present. Expected: withheld, as in step 1.
4. Setting `$sanitized->filter = 'raw';` after the clone restores step 1's behavior — I confirmed this against the same fixture.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,72 @@
...
+
+		if ( ! $resolved instanceof \WP_Post ) {
+			return wp_trim_excerpt( $text, $post );
```
dkoo, line 61: **Suggestion:** The unresolvable-post branch can build an excerpt from the global post — When `$post` doesn't resolve, this hands core the same unresolvable value. `wp_trim_excerpt()` then calls `get_post( null )` and `get_the_content( '', false, null )`, which falls back to `$GLOBALS['post']` — so the excerpt is built from whatever post is set up in the loop, unsanitized.

I reproduced it: with a gated post in `$GLOBALS['post']`, calling the filter with a non-existent post ID returned that global post's content including the withheld marker.

Nothing in core reaches this branch, so this is about the shape rather than a live path — but a filter whose job is withholding content probably shouldn't have a branch that builds an excerpt from a post it was never asked about. Returning `$text` unchanged would fail closed.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,72 @@
...
+		// it the real post anywhere would put gated blocks back in the excerpt.
+		$sanitized               = clone $resolved;
+		$sanitized->post_content = Block_Visibility::strip_blocks_hidden_from_public( $resolved->post_content );
+
+		// A manually written excerpt is the author's own words; core returns it
+		// untouched and so do we. Read it from the post rather than from $text,
+		// which is whatever the filter chain holds by the time this runs and is not
+		// necessarily the manual excerpt. If an earlier filter blanked $text, core
+		// rebuilds — from the sanitized clone, not the original.
+		if ( '' !== trim( (string) $resolved->post_excerpt ) ) {
```
dkoo, line 84: **Suggestion:** The manual-excerpt branch forwards $text without sanitizing it — This branch hands core `$text` verbatim, and core returns it unchanged whenever it is non-empty. So on a post that has both a hand-written excerpt and a `$text` populated by an earlier filter from `post_content`, the withheld block reaches the teaser — the same route `test_prepopulated_text_does_not_bypass_sanitization` (line 128) closes for the auto branch.

`test_blanked_text_on_a_post_with_a_manual_excerpt_does_not_leak` (line 159) covers the empty-`$text` case on a manual-excerpt post, so the uncovered quadrant is specifically manual excerpt *plus* non-empty upstream `$text`. I reproduced it against a fixture post (gated group + manual excerpt) with a priority-6 filter deriving `$text` from `post_content`: the marker came through.

The reasoning for forwarding is sound — once you've decided not to read `$text`, you can't tell the author's excerpt from a filter's substitution. Worth either adding the fourth case or saying that in the comment, so the asymmetry reads as chosen rather than overlooked.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,72 @@
...
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		// Core registers wp_trim_excerpt() here at priority 10. Replace it with a
+		// wrapper that hands core the same work over sanitized content, rather than
+		// reimplementing the trimming: two copies of that logic already exist in this
+		// monorepo and both have drifted from core.
+		$removed = remove_filter( 'get_the_excerpt', 'wp_trim_excerpt', 10 );
+		if ( ! $removed ) {
+			Logger::log( 'Failed to remove core wp_trim_excerpt filter; excerpts may bypass the content gate.', 'CONTENT-GATE-EXCERPT' );
```
dkoo: **Nit:** This log line won't reach anyone, and reads scarier than the failure is — `Logger::log()` returns immediately unless `NEWSPACK_LOG_LEVEL` is defined and above zero, which it isn't in production — so the one canary for this condition never fires where it would matter.

The wording overstates the risk too. If the removal fails, core's callback runs first and this one still runs afterwards at the same priority, rebuilding from the sanitized clone — I confirmed the auto branch ignores whatever `$text` it receives, so the outcome is wasted work rather than a bypass. Either soften the message to match, or route it somewhere visible.

plugins/newspack-blocks/includes/class-newspack-blocks.php
```diff
@@ -1174,6 +1174,11 @@ public static function filter_excerpt( $attributes ) {
 
 			// Recreate logic from wp_trim_excerpt (https://developer.wordpress.org/reference/functions/wp_trim_excerpt/).
 			$excerpt = strip_shortcodes( $excerpt );
+			// Strip blocks the content gate withholds from the public before
+			// excerpt_remove_blocks() flattens the block structure.
+			if ( class_exists( 'Newspack\Block_Visibility' ) && method_exists( 'Newspack\Block_Visibility', 'strip_blocks_hidden_from_public' ) ) {
```
dkoo, line 1252: **Suggestion:** Two identical cross-plugin guards would be better as a named contract — This block and `plugins/newspack-listings/includes/utils.php:212` are the same four lines, each hard-referencing `\Newspack\Block_Visibility` across a plugin boundary and each guarded on `class_exists` plus `method_exists`. They'll drift — the next person to change the ordering will change one and not the other, and a fourth excerpt builder won't know the contract exists.

The monorepo convention for cross-plugin coupling is a hook, and it fits here: newspack-plugin registers `add_filter( 'newspack_content_gate_strip_hidden_blocks', [ Block_Visibility::class, 'strip_blocks_hidden_from_public' ] )` and consumers call `apply_filters( 'newspack_content_gate_strip_hidden_blocks', $excerpt )`. That drops both guards, names the contract in the `newspack_<plugin>_<name>` convention, and makes the version-skew fallback explicit rather than an accident of `method_exists`.

plugins/newspack-blocks/tests/class-newspack-block-visibility-stub.php
```diff
@@ -0,0 +1,54 @@
...
+		 * point), and removes a trivial marker string. The real implementation logic
+		 * is tested in newspack-plugin; this test verifies only the WIRING.
+		 *
+		 * @param string $content Block markup.
+		 * @return string Content with the stub-marked text removed.
+		 */
+		public static function strip_blocks_hidden_from_public( $content ) {
+			self::$sanitization_was_called = true;
+			// Remove the fixture marker that represents gated content.
+			// This stub verifies the integration point, not the real stripping logic.
```
dkoo, line 54: **Suggestion:** The wiring test can't fail for the reason the fix exists — The stub's `strip_blocks_hidden_from_public()` is a `str_replace( 'SECRETMARK', '' )`, which removes the marker wherever the call sits in `filter_excerpt()` — including *after* `excerpt_remove_blocks()` at `class-newspack-blocks.php:1182`. But that ordering is the entire root cause: `excerpt_remove_blocks()` unwraps `core/group` and destroys the attributes, so a call placed after it would strip nothing in production while this test still passed. `$sanitization_was_called` proves the call happens, not that it happens early enough.

Recording the `$content` the stub receives and asserting it still contains the group delimiter with its access-control attributes would pin the ordering without needing the real implementation.

Separately, `plugins/newspack-listings/includes/utils.php:213` gets the same treatment with no test at all — the listings suite has 18 tests and none exercise `get_listing_excerpt()`. I verified that call site works at runtime, so this is about keeping it working.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -92,26 +114,96 @@ public static function filter_render_block( $block_content, $block ) {
...
 			? self::evaluate_gate_rules_for_user( $gate_ids, $user_id )
 			: self::evaluate_rules_for_user( $rules, $user_id );
 
-		if ( 'visible' === $visibility ) {
-			return $user_matches ? $block_content : '';
+		return 'visible' === $visibility ? ! $user_matches : $user_matches;
+	}
+
+	/**
+	 * Remove blocks that are withheld from a logged-out reader.
```
dkoo, line 149: **Suggestion:** The institution caveat has a concrete cache exposure worth naming — The docblock notes that the institution rule can vary between anonymous requests. The part it doesn't say is where that variance gets stored: `Newspack_Blocks_Caching` keys Homepage Posts markup on block attributes plus an index with no reader dimension (`class-newspack-blocks-caching.php:152-157`), for every visitor lacking `edit_posts`. `Institution::user_matches_institution()` evaluates IP ranges for user 0 when the visitor carries the IP-access cookie.

So an on-campus anonymous visitor can populate that cache with an institution-visible excerpt, which is then served to everyone for the TTL. Page caching is fine here — the cookie suppresses it — but the object cache isn't.

Having the excerpt path skip request-context-dependent rules would close it, and would also make the "evaluated against the anonymous reader" guarantee in the docblock literally true.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -92,26 +114,96 @@ public static function filter_render_block( $block_content, $block ) {
...
+	 * without a user dimension, so reader-varying output would be served across
+	 * readers. This does not guarantee an identical result for every anonymous
+	 * visitor, though: the "institution" access rule supports anonymous
+	 * evaluation and depends on request context (IP/cookie), so it can still
+	 * vary between two anonymous requests.
+	 *
+	 * @param string $content Serialized block content.
+	 * @return string Content with withheld blocks removed.
+	 */
+	public static function strip_blocks_hidden_from_public( $content ) {
```
dkoo, line 162: **Suggestion:** The strip runs twice per post on the hottest page — When Homepage Posts is on the page, this runs once from the priority-10 excerpt filter and again inside the priority-11 closure at `class-newspack-blocks.php:1180`, whose result discards the first pass entirely. Each call is a full `parse_blocks()` + recursive walk + `serialize_blocks()`, and a homepage typically carries several post loops.

The `strpos` early-out already keeps this off ungated content, so the cost only lands on posts that use the gate — but those are exactly the posts on gate-using sites. A static memo keyed on `md5( $content )` inside the method would make the repeat calls free and needs no coordination between the three call sites.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
```diff
@@ -92,26 +114,96 @@ public static function filter_render_block( $block_content, $block ) {
...
+				$inner_content = [];
+				$index         = 0;
+
+				foreach ( (array) $block['innerContent'] as $chunk ) {
+					if ( is_string( $chunk ) ) {
+						$inner_content[] = $chunk;
+						continue;
+					}
+					$child = $block['innerBlocks'][ $index++ ] ?? null;
+					if ( null === $child || self::is_hidden_for_user( $child, 0 ) ) {
```
dkoo: **Nit:** Each child is evaluated twice, and $stripped[0] assumes the two agree — `is_hidden_for_user( $child, 0 )` runs here as a guard and again at line 176 inside `strip_hidden( [ $child ] )`. The per-request cache makes that cheap, but `$stripped[0]` on line 195 is unguarded: if the two evaluations ever disagreed it would be an undefined-index warning plus a `null` in `innerBlocks`. They can't disagree today — the decision is memoized and pure within a request — so this is correctness by coincidence rather than a live bug.

Dropping the line-191 check and testing `if ( empty( $stripped ) ) { continue; }` instead collapses the two evaluations into one and makes the guard structural.

Also on line 185: `(array) $block['innerContent']` warns and yields `[]` if the key is absent, which would silently drop every child. `parse_blocks()` always sets it, but this is now public API another plugin could call with a hand-built array — `?? []` costs nothing.

plugins/newspack-blocks/tests/bootstrap.php
```diff
@@ -27,11 +27,33 @@ function newspack_blocks_manually_load_plugin() {
 }
 tests_add_filter( 'muplugins_loaded', 'newspack_blocks_manually_load_plugin' );
 
+/**
+ * Load test stubs for dependencies not available in the isolated test environment.
+ */
+require_once __DIR__ . '/class-newspack-tag-labels-stub.php';
```
dkoo: **Nit:** Stubs load before the autoloader, so the skip guards can never fire — The stubs are required before the composer autoloader is registered, so the `class_exists()` guard inside each stub file can't autoload a real implementation and the stub always wins. That makes the `markTestSkipped` guard at `test-homepage-posts-block.php:341` unreachable, and means a real `\Newspack\Block_Visibility` would be shadowed rather than preferred if newspack-plugin ever becomes loadable in this suite. Moving the two `require_once` calls below the autoloader block restores the intent.

Worth noting for anyone reading this thread later: `/tests` is in the plugin's `.distignore`, so the stub can't reach a build.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-block-visibility.php
```diff
@@ -420,6 +420,83 @@ public function test_non_editor_still_restricted_on_front_end() {
...
+	/**
+	 * The early-out's substring assumption holds for every gating shape.
+	 *
+	 * Sanitization returns early when the raw content does not contain the literal
+	 * 'newspackAccessControl'. That is a performance guard sitting on a security
+	 * boundary: if a gated block could ever serialize without the literal, the guard
+	 * would skip sanitization rather than merely skip work. Pin the assumption
+	 * instead of trusting it, and confirm the block is actually removed.
+	 */
+	public function test_gated_blocks_always_serialize_with_the_guard_literal() {
```
dkoo, line 450: **Nit:** Test docblock promises more than the body checks — The docblock says the test will confirm the block is actually removed, but the body only asserts that a serialized gated block contains the `newspackAccessControl` literal — it never calls `strip_blocks_hidden_from_public()`. Pinning that substring is worth doing, since the early-out at `class-block-visibility.php:152` depends on it; the claim just needs trimming to match. The `wp_set_current_user( 0 )` a few lines down is also inert here, as nothing in the test evaluates a reader.

dkoo (changes requested):
Release risk: HIGH — the excerpt fix is correct and well-tested, but its seam changes `get_the_excerpt` behavior on every Newspack site regardless of whether content gates are switched on, and the sanitized-clone guarantee is inherited rather than asserted. Both are reproducible and both have small fixes.

I verified the fix does what it claims. In an isolated env I confirmed withheld blocks are gone from the home-page teaser, `excerpt.rendered`, the Homepage Posts loop and the listings builder, that a fully-gated post yields a blank teaser, and that a hand-written excerpt is passed through untouched. `strip_hidden()`'s null-marker bookkeeping — the part most likely to go wrong — held across six block shapes and is idempotent. Nice work on the equivalence test across six gate configurations; asserting the excerpt matches the front end rather than merely asserting absence is what makes it catch over-stripping too.

Two things to weigh beyond the threads. First, both blockers are the same shape: a guarantee the code relies on but does not state. Neither is reachable through a path in this repo today; both become reachable through ordinary third-party code, and both fail silently and completely rather than partially. Second, and separate from this PR: on this base branch an anonymous REST read of `content.rendered` still returns withheld block content. That is pre-existing and untouched here, so it should not hold up this merge — but NPPM-3090 shouldn't close while it stands.

**Notes on code outside this PR's changed lines** (can't be posted inline, so collected here):
- `plugins/newspack-plugin/includes/content-gate/class-block-visibility.php:59` — **Suggestion:** Pre-existing: anonymous REST reads of content.rendered still return withheld blocks — Flagging this because of where it leaves the ticket, not because this PR caused it — the line is untouched by this diff and is present on `origin/release`, so it should not hold up the merge. `filter_render_block()` returns early whenever `REST_REQUEST` is defined. That constant is set for *public* REST traffic, not only editor traffic, and `WP_REST_Posts_Controller` builds `content.rendered` by running `the_content` over `post_content`. An anonymous `GET /wp-json/wp/v2/posts/<id>` therefore returns withheld block content in full, on a surface that is public by default. I confirmed it over real HTTP against this branch. Worth noting the distinction, since it is easy to miss: calling `rest_do_request()` in-process does *not* define `REST_REQUEST`, so an in-process check reports no exposure. Only a real request through `/wp-json/` exercises the branch. **To reproduce** 1. Publish a post with a gated `core/group` wrapping a marker. 2. Logged out, confirm the article page withholds it: `curl -sk 'https://<env>/<permalink>?cb=1' | grep MARKER` → only the ungated markers. 3. Logged out, read it over REST: `curl -sk 'https://<env>/wp-json/wp/v2/posts/<id>?_fields=content&cb=2' | grep MARKER` 4. Observed: the withheld marker is present, while `?_fields=excerpt` on the same post correctly withholds it after this PR. The bypass reads as intended for the editor (block renderer, previews, query-loop rendering), which a capability check would keep working while closing the public path. Since NPPM-3090 is a security ticket, it's worth confirming this is tracked before it closes.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
jason10lee: `293253847`. Worth recording that the obvious fix doesn't work: guarding `init()` on `is_newspack_feature_enabled()` would unregister the filter for the whole suite, since the constant is absent from the test bootstrap and `init()` runs at plugin load while tests define it in `setUp`.

Went with per-post scoping via a new `Block_Visibility::has_access_control()`, so the marker literal is named once. Baseline on the unfixed tree returned `'PUBLICMARK'` where `'THIRDPARTYEXCERPT'` should have survived.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
jason10lee: `f6470734f`, `$sanitized->filter = 'raw'` as you described. Baseline returned `'PUBLICMARK SECRETMARK'`, so the display-form post was bypassing sanitization completely.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
jason10lee: `9edcc1b9b` — returns `$text`.

The test calls `filter_get_the_excerpt()` directly rather than through `apply_filters`: another callback on this hook reads `->post_type` off the second argument and fatals on a bare ID, which masked the branch under test.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
jason10lee: `9edcc1b9b`, and I went further than the thread — the branch now passes `$resolved->post_excerpt` rather than `$text`.

Sanitizing `$text` would be ineffective: by the time it reaches here it is usually rendered HTML with no attributes left for the strip to match. Passing the post's own excerpt makes both branches refuse to trust `$text` for the same reason.

The cost is that on a gated post an upstream filter's substitution loses to the author's excerpt. That is a deliberate divergence from core, confined to gated posts — flag it if you would rather keep core's behavior and document the gap instead.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
jason10lee: `a0116d2da`. Reworded to "excerpts are trimmed twice", with a comment explaining why it is not a bypass. Left it on `Logger::log` rather than rerouting.

plugins/newspack-blocks/includes/class-newspack-blocks.php
jason10lee: Filed as NPPM-3141 rather than done here — it adds a public contract across two more plugins and this PR targets `release`. The ticket carries the listings coverage gap you noted.

plugins/newspack-blocks/tests/class-newspack-block-visibility-stub.php
jason10lee: `c514541b1`. The stub records `$received_content`, and the test asserts it still carries `newspackAccessControl` and `<!-- wp:group`.

Mutation-tested by moving the call after `excerpt_remove_blocks()`: it now fails, and the stub received `'<p class="wp-block-paragraph">PUBLICMARK</p><p class="wp-block-paragraph">SECRETMARK</p>'` — flattened, attributes gone, which is exactly the state the old assertions could not distinguish.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: Filed as NPPM-3142 with the Security label. Verified both halves: `get_cache_key()` carries no reader dimension, and caching is skipped only for logged-in users holding `edit_posts`.

Kept out of this PR since it concerns the method's guarantee generally rather than the excerpt path.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: `67c6d6a77`, memo keyed on `md5( $content )`, cleared alongside `rules_match_cache`.

No test asserts the repeat-call saving: `rules_match_cache` already suppresses the callbacks a counting test would measure, so such a test would report `1` with or without the memo. Asserted the reset behavior instead, and said which property is unobservable in the docblock.

plugins/newspack-plugin/includes/content-gate/class-block-visibility.php
jason10lee: `a0116d2da` — dropped the line-191 check for `if ( empty( $stripped ) )`, plus `?? []` on `innerContent`.

plugins/newspack-blocks/tests/bootstrap.php
jason10lee: `a0116d2da`. One correction: the move is right structurally but changes nothing observable today — `class_exists( '\Newspack\Block_Visibility' )` still returns false with the autoloader registered, so the stub wins and that `markTestSkipped` stays unreachable. Future-proofing rather than a live fix.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-block-visibility.php
jason10lee: `a0116d2da` — trimmed the claim to what the body asserts, and dropped the inert `wp_set_current_user( 0 )`.

jason10lee:
All twelve threads addressed across six commits, and the base is refreshed to `release`.

**Blockers.** `293253847` scopes the seam so ungated posts get core's contract untouched; `f6470734f` pins the clone to raw. Both have tests that fail on the unfixed tree.

**Suggestions.** `9edcc1b9b` (unresolvable-post and manual-excerpt branches), `67c6d6a77` (strip memo), `c514541b1` (stub ordering). The hook contract and the institution/cache exposure are filed as NPPM-3141 and NPPM-3142 rather than widening a hotfix.

**Nits.** `a0116d2da`.

Suites after the merge: newspack-plugin 447, newspack-blocks 126, phpcs clean. I resolved the four nit threads and left the blockers and suggestions open for you.

plugins/newspack-plugin/includes/content-gate/class-content-gate-excerpt.php
```diff
@@ -0,0 +1,99 @@
...
+		if ( ! $resolved instanceof \WP_Post ) {
+			return $text;
+		}
+
+		// Core returns a non-empty $text untouched; the branches below deliberately
+		// do not. Confine that difference to posts that actually use the gate, so on
+		// every other post -- including every post on a site with gates switched off,
+		// where this filter still replaces core's -- an excerpt supplied by a filter
+		// below priority 10 survives exactly as it would without Newspack installed.
+		if ( ! Block_Visibility::has_access_control( (string) $resolved->post_content ) ) {
```
dkoo, line 60: **Nit:** A post that merely mentions the attribute name loses the core pass-through — `has_access_control()` is a substring test, so a post whose prose happens to contain `newspackAccessControl` — documentation, a changelog entry, a support article — takes the gated branch and loses core's `$text` pass-through, even though it has no gated block.

I confirmed it: a post whose only occurrence of the string was inside a paragraph had a priority-6 filter's excerpt discarded.

This fails in the safe direction and the blast radius is a rounding error next to what the early-out fixes, so I would not change the check. Worth a clause in the docblock, since the substring test is now load-bearing for *whether to sanitize at all* rather than just for skipping work.

dkoo (approved):
Approving. Release risk: ELEVATED — no blockers remain, and the one thing keeping this above LOW is that the seam still displaces a core filter on every site. That is a contract change riding a hotfix, and it wants a release note rather than more code: `remove_filter( 'get_the_excerpt', 'wp_trim_excerpt' )` and `has_filter()` checks for it now silently do nothing, which is a documented and widely-copied idiom.

Both blockers are fixed and I confirmed neither reproduces. With gates switched off, a priority-6 filter's excerpt now survives on an ungated post, so core's contract is restored exactly where it was broken. A display-form post no longer discards the sanitized clone. The manual-excerpt branch reads the post's own excerpt, the unresolvable-post branch fails closed, and the stripping is memoized with the cache wired into the test reset. All six of my fixtures still behave, and the new early-out held against display-form posts, `post_content` filters, and parse/serialize round-trips.

The test work is better than what I asked for. The stub recording `$received_content` and asserting the block delimiter is still intact pins *ordering*, which is the property that made this a defect — and mutation-testing it by moving the call after `excerpt_remove_blocks()` is the right way to show it bites. Same for being explicit that the memo's saving is unobservable given `rules_match_cache`, and asserting the reset instead of writing a test that would pass either way. Your correction on the bootstrap ordering is right too: it is future-proofing, not a live fix.

NPPM-3141 and NPPM-3142 are the right calls for a PR targeting `release`.

One follow-up that is not about this diff: on this base branch an anonymous REST read of `content.rendered` still returns withheld block content, because the render-path bypass keys off `REST_REQUEST`, which is set for public REST traffic and not only editor traffic. I confirmed it over real HTTP against this branch — worth noting that an in-process `rest_do_request()` check does not reproduce it, since that never defines the constant. Pre-existing and untouched here, so it should not hold up this merge, but NPPM-3090 shouldn't close while it stands.

Review provenance: 4 of 4 reviewer passes (lineup: deep, standards, wp-expert, codex); round 2 re-verified at runtime; release risk: ELEVATED.

PR #833, opened 2026-08-09
ci(release): stop the plan gate failing when nothing needs releasing

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? — n/a in the usual sense: the change is a GitHub Actions workflow with no PHP or JS, so PHPCS does not cover it. It follows the surrounding file's conventions.
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? — no duplicate, but two open PRs do touch this file: #830 (removes `hotfix/**` and `epic/**` from the release trigger) and #811 (deploys the released version to wp.org). Neither touches the plan-gate lines and neither conflicts textually with this diff. See "Other information" for the one interaction with #830.

The release workflow's plan step now reports "nothing to release" instead of failing. That step exists to skip the multi-minute build on no-op pushes, and until now it has never been able to: the skip path was unreachable.

Two independent problems in the same step, both in `.github/workflows/release.yml`:

**The gate died instead of reporting no-release.** A bare `grep` ran with `set -e` active, so a no-match exited 1 and killed the script before `grep_status=$?` on the next line. Every branch below it — including `release_needed=false` and its "skipping build and release" message — was unreachable. It now uses `|| grep_status=$?`, which keeps the benign no-match from tripping `set -e` while leaving a real grep error (exit 2) to fall through to the failure branch, as the original comment intended.

**The tee check never fired.** `PIPESTATUS` is rebuilt after every command, and an assignment is a command, so reading `[1]` on the line after reading `[0]` returned empty. That is the `line 12: [: : integer expression expected` line present in every run's log, passing ones included. Both statuses are now copied in one go.

**The skip path now needs positive evidence.** Review flagged that trusting a *non-matching* grep leaves the gate fail-open: if `multi-semantic-release` reworded its dry-run output, the absence of "the next release version is" would read as "nothing to release" and skip a real release behind a green job. The skip branch now also requires the tool's own `Released 0 of N packages` line. Both strings would have to change together for a silent skip; if only one changes, the job fails loud and prints the whole log (a tail of a multi-package run shows only the last package). A genuine grep error, exit 2 or above, now reports as such rather than as a wording change. That restores the intent stated in #124 — fail loud on any dry-run, log, or grep error. A pre-existing `[ ! -s ]` guard already covered a missing or empty log; this covers a log that is present but unrecognizable.

**The jobs that follow a release now gate on the plan result.** With the gate fixed, a no-op push to `release` goes green, and four jobs that the red `Release` job had been skipping start running: `post-release` plus the three WordPress.org deploys. `post-release.sh` chooses between a merge and `git reset --hard release` plus `git push --force origin alpha` by reading the commit before HEAD, which it documents as the automated release commit — a commit a no-release run never creates. The `release` job now exposes `release_needed` as an output and all four dependents gate on it, so those pushes skip the jobs exactly as they did while the gate was red. It also drops the three redundant `pnpm install` + build + `release:archive` runs a green no-op push would otherwise trigger, on the one branch where those jobs run. The misread inside `post-release.sh` is untouched here and tracked as NPPM-3135.

Closes NPPM-3118.

1. Extract the `run:` block of the "Plan release" step into a file and replace the `pnpm exec multi-semantic-release …` call with `cat "$CANNED"`.
2. Run it with `CANNED` pointing at a log containing `The next release version is 1.2.3`. Expect exit 0 and `release_needed=true` in `$GITHUB_OUTPUT`.
3. Run it with `CANNED` pointing at a log containing `Released 0 of 18 packages, semantically!` Expect exit 0, `release_needed=false`, and the message `No package requires a release; skipping build and release.`
4. Repeat step 3 against this file from `release` for contrast: it exits 1 with `release_needed` unset.
5. Check this PR's own `Release` run. Pushing to a `hotfix/**` branch triggers the workflow, all three commits here are `ci:` (a non-releasing type under the analyzer's angular preset), and no package has releasing commits since its last tag — so the plan step finds nothing to release. That is precisely the case that fails on `release` today. A green run on this branch, logging the skip message, is the fix working end to end.
6. Run it with `CANNED` pointing at a log containing neither marker (e.g. `All done. Nothing was published this time.`). Expect exit 1, the "Neither a release version nor a zero-release marker found" message, and the whole log.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — not applicable: this is a workflow file, and the repo has no harness for executing workflow steps. Verified instead by extracting the step body and running it against canned logs, with the pre-fix version as a control.
* [x] Have you successfully run tests with your changes locally? — yes, via the extraction above: all three branches of the gate exercised (release needed, no release needed, dry-run failure), plus the `grep_status=2` error path in isolation.

**Interaction with #830.** That PR stops `hotfix/**` and `epic/**` from triggering releases at all, which removes one set of branches where this bug fires. It does not overlap with this fix: the gate still runs on `release` and `alpha`, where a push carrying only `chore:`/`docs:` commits hits the same failure. The two are independent and can land in either order. #830 is based on `main` and this is based on `release`, so they meet at the next promotion rather than in a merge conflict.


.github/workflows/release.yml
```diff
@@ -90,11 +94,31 @@ jobs:
...
           # next release version is X" for a package's first-ever release --
           # missing the latter would silently skip that release.
-          grep -qi 'the next release version is' msr-dryrun.log
-          grep_status=$?
+          # `|| grep_status=$?` rather than a bare call: under `set -e` a bare grep
+          # that matches nothing exits 1 and kills the script, so the "no release
+          # needed" branch below is never reached. Exit 2 (a real grep error) still
+          # falls through to the failure branch.
+          grep_status=0
+          grep -qi 'the next release version is' msr-dryrun.log || grep_status=$?
```
wil-gerken, line 109: **Suggestion:** A push to `release` that releases nothing currently fails right here, and that failure has been quietly holding back four downstream jobs. Once this lands they all run, and one of them rewrites shared branches.

Here is how that would play out. Someone promotes `alpha` into `release` to clear the queue, and it happens to carry only `chore:` and `ci:` commits, so nothing releases and — correctly, thanks to this fix — the run goes green. `post-release` fires, for the first time ever on a push like that. A day later a small `ci:` fix is merged to `release`. `post-release` runs again, looks one commit back, finds `Merge pull request #664 from Automattic/alpha`, concludes the release came from alpha, and force-pushes `alpha` back to `release`. Anything that landed on `alpha` in between is gone from the remote, and every run in the sequence is green.

The mechanism is an off-by-one that only appears when no release happens. `post-release.sh:26-28` documents its own assumption — "The last commit here is the automated release commit; the one before it carries the merge info" — and reads `git log -n 1 --skip 1`. On a no-release run no automated commit is created: the Release step is skipped and `finalize-package-versions.cjs:106-110` commits nothing. So that read lands one commit early. `post-release.sh:234` tests it against `^Merge .*alpha` to choose between the non-destructive merge at `:242` and `git reset --hard release --` plus `git push --force origin alpha` at `:237-239`.

On how likely this is, I want to be straight with you: I could not find a steady-state instance. `git log -n 1 --skip 1 --pretty=format:'%s' f937e9f69` does return `Merge pull request #664 from Automattic/alpha` today, so the shape is expressible — but both of those runs concluded `failure` and both predate this step existing at all, so that pair is an artefact of the week the workflow was being built rather than evidence of a promotion that released nothing. Since the gate landed there have been two alpha→release promotions, and the one that ran to completion did release. The precondition is therefore possible but unobserved, which is why this is a suggestion rather than something to hold the PR on.

What is not speculative is the rest of the chain. `alpha` currently holds 76 commits absent from `release`, 54 of them authored work, so if this ever does fire there is real work to lose. And branch protection would not catch it: `post-release.sh:244` already pushes directly to `alpha` under the same ruleset, and `alpha` carries release-bot commits from 2026-08-10 showing the token bypasses it.

Run 30862763213 is the before-picture: a `ci(release):` push to `release`, `Release` concluded `failure`, and `Post-release branch sync` plus all three WordPress.org jobs concluded `skipped`.

Worth stating plainly: this misfire predates your change. `post-release.sh:28` reads the wrong commit on any run, and the same path is already reachable from a *releasing* push whose predecessor is an alpha promotion. What this PR changes is the class of runs that get there, and it is specifically the no-release promotion that leaves a bare `Merge … alpha` at the tip of `release` to serve as the predecessor. I am raising it here because this is the change that takes the accidental safety net away.

The smallest in-scope guard is to make the plan result visible to dependent jobs — `outputs: { release_needed: "${{ steps.plan.outputs.release_needed }}" }` on the `release` job, then `&& needs.release.outputs.release_needed == 'true'` on the dependent jobs' conditions. That costs nothing in behaviour, because those pushes do not sync anything today either; the job is red. If you would rather fix it at the source, giving `post-release.sh` a positive "a release happened" signal — the tag snapshot at `release.yml:158` already exists for the Slack notifier — is the sturdier version. None of this has to happen in this PR — a follow-up ticket is the right home for it.

One smaller rider from the same run: `post-release.sh:29` resolves `LATEST_VERSION_TAG` with `git describe --tags --abbrev=0`, which on a no-release run returns a previous release's tag — at `f937e9f69`, an alpha prerelease tag. Any merge commit the sync creates would assert a release that did not happen. Cosmetic next to the force-push, and it only materialises when the target branch has drifted behind `release`, but it is the same root cause.

.github/workflows/release.yml
```diff
@@ -90,11 +94,31 @@ jobs:
...
+            # Absence of the version phrase is weak evidence on its own: a
+            # wording change upstream would read as "nothing to release" and
+            # silently skip a real one. Require multi-semantic-release's own
+            # zero-count line to corroborate before trusting the skip.
+            zero_status=0
+            # msr logs `Released 0 of 18 packages` when nothing is due and `Released 1
+            # of 18 packages` when something is, so the count is real corroboration.
+            # Matched loosely so a cosmetic reformat cannot redden a no-op push.
+            grep -qiE 'released +0 +of +[0-9]+ +packages?' msr-dryrun.log || zero_status=$?
+            if [ "$zero_status" -ne 0 ]; then
```
wil-gerken: **Nit:** The outer grep carefully separates exit 1 from exit 2 or above, but this inner check tests `-ne 0` and prints "the dry-run output may have changed" for both. On a genuine grep failure an unreadable or vanished log would be reported as an upstream wording change that did not happen.

Nothing is at risk either way, and the error branch is close to unreachable in practice — the `[ ! -s msr-dryrun.log ]` guard above has already established the log is present and non-empty, and nothing touches it in between. So this is purely about the message being honest about which case fired. Mirroring the outer branch's split would do it:

```bash
if [ "$zero_status" -eq 1 ]; then
  echo "Neither a release version nor a zero-release marker found; the dry-run output may have changed."
  cat msr-dryrun.log || true
  exit 1
elif [ "$zero_status" -ne 0 ]; then
  echo "Failed to scan the dry-run log for the zero-release marker (grep exit $zero_status)"
  exit "$zero_status"
fi
```

wil-gerken (approved):
Release risk: ELEVATED — one thing worth a follow-up: with the gate fixed, a no-op push to `release` would reach the post-release branch sync, which can force-push `alpha`.

Thank you, Jason! I was super happy to spot this PR.

I pulled the `Plan release` step body out of both this branch and `release` and ran them against canned dry-run logs under bash 5.1 and GNU grep 3.7, including the verbatim log from this PR's own run 31651405082. Twelve cases, all as advertised: on `release` that real log exits 1, on this branch it exits 0 and writes `release_needed=false`. The `[: : integer expression expected` line shows up on every base run and on none of yours, so the `PIPESTATUS` copy is doing exactly what the comment claims. The new zero-marker check holds up under pressure too — `Released 10 of 18` correctly fails to match, a log carrying both markers still yields `release_needed=true` because the version grep runs first, and the first-ever-release phrasing is matched. `|| grep_status=$?` preserves the exit-2 path, so #124's "fail loud" intent survives. All three changes do what you say they do, and the extraction-and-canned-log approach in your test plan is the right way to test this.

The one thing I would put on a follow-up is downstream of the diff, not in it. The `post-release` job and the three WordPress.org deploy jobs gate on the branch name and on the `release` job succeeding — not on `release_needed` — and the `release` job exposes no outputs, so they cannot currently gate on it. While the gate was broken, a no-op push to `release` turned the job red and those jobs were skipped; run 30862763213 shows exactly that. Once the job goes green they all run, and `post-release.sh` chooses between a merge and a `git reset --hard` plus `git push --force origin alpha` by reading the commit before HEAD — a commit it documents as "the automated release commit", which on a no-release run does not exist. Details and the smallest guard are in the inline comment.

The same missing gate has a second, milder effect worth folding into whatever you do about the first. The three WordPress.org deploy jobs are conditioned the same way, so the step that just logged "skipping build and release" is followed by three jobs each running a full `pnpm install`, `build:packages`, plugin build and `release:archive` before `release-wporg.sh:44` finds the SVN tag already present and exits. On the one branch where those jobs run, the multi-minute build this change exists to skip comes back three times over. The outcome is safe today, because that check is idempotent. I have filed the one wrinkle in it — `svn ls … > /dev/null 2>&1` inside an `if` cannot tell "tag absent" from "the lookup did not complete" — separately as NPPM-3132, since it is not your code and not sensibly fixable here.

To be clear about attribution: the misfire in `post-release.sh` is not something you introduced, and the failing gate was suppressing it by accident. I am raising it here because this is the change that removes the accident and the guard is small — not because it should hold up a fix you have clearly done carefully.

One note on the description rather than the code: test step 6 says the unknown-shape failure prints "the log tail", but `97953d8f` changed it to print the whole log, with a good reason in the comment. Worth a quick edit so the steps match what a reviewer will see.

.github/workflows/release.yml
jason10lee: Guard applied in a5163789e: `release_needed` is now a job output, and `post-release` plus the three WP.org jobs gate on it.

Two things I checked that add to the write-up. The trailing `--` in `git reset --hard release --` isn't a safety: with no pathspec after it, that's a plain hard reset returning 0, not the error `reset --hard <commit> -- <path>` produces. And every terminal path in the plan step either exits non-zero or writes `release_needed`, so an unset output can only mean the job already failed, in which case the dependents skip on `needs` anyway.

On what CI proves here: the run on this branch shows `Released 0 of 18 packages`, `Release` green, and all four dependents skipped. That skip is over-determined, though, since `github.ref == 'refs/heads/release'` is already false on a hotfix branch. The new condition parses and evaluates without error, but its first real exercise is a push to `release`.

The `post-release.sh` root cause and the `LATEST_VERSION_TAG` rider are written up for a follow-up ticket, not yet filed.

Thanks for chasing the post-release chain this far; it wasn't visible from the diff.

.github/workflows/release.yml
jason10lee: Applied in 829901274, matching the outer branch's 0/1/else split.

Exercised all four cases rather than reading them: `zero_status` 0 exits 0 and writes `release_needed=false`, 1 exits 1 with the wording message and the log, 2 and 3 exit with the matching code and name the grep failure.

.github/workflows/release.yml
jason10lee: Follow-up now filed as NPPM-3135 (Backlog, Medium), carrying the reproduction and the `LATEST_VERSION_TAG` rider.

PR #834, opened 2026-08-10
fix(content-gate): harmonize whole-post content gating

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? — PHPCS run from the workspace root against the root ruleset over every changed file, tests included; clean.
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? — no duplicate. The `tests/mocks/wc-mocks.php` overlap flagged earlier resolved in the base merge, which took main's two guarded mocks alongside this branch's `wc_get_page_id()`.

This makes whole-post content gating follow the same entitlement decision on REST reads that the front end already makes.

- **Extracts the entitlement decision.** `Content_Gate::get_restriction_for_post()` holds the feature flag, the Memberships deferral, the page exclusions and the restriction filters, with no dependency on query context.
- **Applies it over REST.** A `rest_prepare_{$post_type}` filter substitutes the teaser and gate into `content.rendered`, puts the teaser in `excerpt.rendered`, and reports `comment_status` as `closed` — matching what the front-end path assigns. Each item in a collection is decided independently.
- **Threads the post ID through gate-layout resolution**, so it resolves outside a singular view.
- **Opts reader-dependent responses out of shared caches**, including the responses this filter does not modify.
- **Short-circuits metering** for the duration of the REST evaluation, so an API read cannot spend a reader's article allowance.

Scope: reads served by `WP_REST_Posts_Controller`, plus the comment-creation path below.

**Added in review, each with its own reason for the shape it took:**

- **A body core withheld for a password is left alone.** The obvious test — is `content.rendered` present and empty — misreads every response that omits the field, which `context=embed` and any `_fields` list without it both do. Core withheld nothing there, and the excerpt still carries the real body, so that shape leaks it. Asking the controller's own `can_access_password_content()` is what settles it, because that is the predicate core used to decide whether to withhold.
- **REST no longer accepts a comment on a post it has just reported as closed.** The refusal sits inside the comments endpoint rather than in `comments_open()`: widening that would reach the admin comment screens and every front-end call for a post other than the one rendering, which is a far larger surface than the mismatch. Anonymous creation was never open — core refuses it independently of anything here.
- **`newspack_content_gate_has_restriction_source` answers in both directions again.** The feed path applies it to the combined answer rather than to the first-party half alone, so a publisher callback returning false is not overridden by Memberships being OR-ed in afterwards. This adds `Content_Gate::detect_first_party_restriction_source()` — worth naming because `newspack-manager` consumes this contract.
- **Page exclusions live in one list.** `restrict_post()` keeps the guards that need a query; the exclusions decidable from a post ID are `is_excluded_from_gating()`'s alone, so a fourth one does not have to be added twice to take effect on both paths.

**Deliberately not in this PR**, both filed: a gated post's comments remain readable through `/wp/v2/comments` (https://linear.app/a8c/issue/NPPM-3235 — excluding them at query level needs the restricted set up front, and dropping them afterwards desynchronises the pagination headers), and each restricted item in a collection re-renders the gate layout (https://linear.app/a8c/issue/NPPM-3236).

**Behaviour changes worth knowing before you review:**

- A reader whom metering would have admitted receives the teaser over REST rather than the full article. Deliberate — an API read should not silently consume an allowance — and it means REST is stricter than the front end for metered readers.
- `excerpt.rendered` for a gated post returns the teaser. That matches what the front-end path assigns on the singular view; other surfaces keep the editor's excerpt, so headless consumers will see a longer value than before.
- With gating configured, REST responses for gate-eligible post types are marked `no-store`. Sites with no published gate, and Memberships sites, are excluded from that.
- A reader the gate turns away now gets a 403 from `POST /wp/v2/comments` on that post. Entitled readers and ungated posts are unaffected.
- The metering short-circuit is defense in depth; the write it guards is not currently reachable through this path.

Also included: a test-harness fix. WordPress core snapshots the hook registry once per process and restores it between tests, which was silently unregistering these filters partway through a full-suite run. Tests are now re-established per test, and a `Content_Gate` static that persisted across test methods is reset.

Closes NPPM-3091.

1. Run the plugin's full suite: `n test-php` in `plugins/newspack-plugin`. Use the full suite rather than `--group content-gate`, which was not authoritative before the harness fix above.
2. `phpunit --filter Test_Content_Gate_Rest` runs the REST tests on their own (22).

The full run reports 3712 tests with nine errors and three failures, none of them in code this branch touches. Nine are the date-range segmentation integrations tests that arrived with #767 in the base merge. Two are premium-newsletter tests that read a sibling plugin by relative path, so they follow the checkout rather than the branch.

If you reach for `--filter Content_Gate` instead, expect two `Test_Advanced_Settings` feed tests to fail there and pass in a full run — on this branch and on a checkout without it. That pair is order-dependent rather than a feed defect: https://linear.app/a8c/issue/NPPM-3238.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — 22 tests in `tests/unit-tests/content-gate/class-content-gate-rest.php`, plus exclusion coverage in `content-gates.php`. Each was checked to fail when the behaviour it names is broken; the password-guard set includes one test that fails against the narrower fix considered for it, and the comment guard was checked at 201 without it and 403 with it.
* [x] Have you successfully run tests with your changes locally? — yes, full plugin suite; the residual failures are accounted for above.


plugins/newspack-plugin/tests/unit-tests/content-gate/content-gates.php
```diff
@@ -1210,6 +1210,89 @@ public function test_gated_post_locks_commenting() {
...
+		// Control: an ordinary post matching the same gate, not configured as
+		// any excluded page, is still gated.
+		$control_post_id  = $this->factory->post->create();
+		$this->post_ids[] = $control_post_id;
+
+		$this->assertIsArray(
+			Content_Gate::get_restriction_for_post( get_post( $control_post_id ) ),
+			'An ordinary post matching the gate must still be restricted.'
+		);
+	}
```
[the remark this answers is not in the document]
jason10lee: WordPress's test harness already covers this. `WP_UnitTestCase_Base` opens a transaction in `set_up()` (`abstract-testcase.php:142`) and issues `ROLLBACK` in `tear_down()` (`:173`), then flushes the object cache (`:429`, `:435`). The option write is therefore reverted in both the database and the `alloptions` cache before the next test method runs.

The concern isn't idle in this file, though. This branch fixes two genuine cross-test leaks of exactly this shape: WordPress snapshots the hook registry once per process and restores it between tests, which was silently unregistering these filters partway through a full-suite run, and a `Content_Gate` static persisted across test methods. Both needed explicit handling because the harness does not cover them. Options it does.

Leaving as-is.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,303 @@ public static function is_gating_active(): bool {
...
+	 * The `rest_prepare_{$post_type}` hook fires from WP_REST_Posts_Controller.
+	 * A post type declaring its own rest_controller_class never fires it; none
+	 * do in this workspace, and the boundary is documented rather than
+	 * assumed away.
+	 */
+	public static function register_rest_filters() {
+		foreach ( get_post_types( [ 'show_in_rest' => true ], 'names' ) as $post_type ) {
+			add_filter( "rest_prepare_{$post_type}", [ __CLASS__, 'filter_rest_response' ], 10, 3 );
+		}
+	}
```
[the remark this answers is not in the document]
jason10lee: The site-level half of this landed in `f2e444d6d`: sites with no published gate, and Memberships sites, now return before any evaluation or cache opt-in.

The per-post narrowing is deliberately not included. The obvious predicate — skipping posts with no gate bound via `Content_Restriction_Control::get_post_gates()` — reopens a bypass: `newspack_is_post_restricted` is a public, third-party-extensible filter, and a publisher plugin restricting posts without publishing a Newspack gate has no per-post equivalent to that check. Implementing it failed the regression test covering exactly that case, so it was reverted rather than shipped.

The residual cost is per-item evaluation and a `no-store` header on gate-eligible post types, on sites that have configured gating. Tracked as a follow-up rather than solved here, since the safe narrowing is already in and the unsafe one is a gate bypass.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+		// Known gap, not fixed here: if the body is password-protected AND
+		// genuinely empty (content.rendered === '' with no password supplied,
+		// or supplied-but-wrong), this whole block returns early and none of
+		// the substitutions below run — including comment_status, which stays
+		// whatever core reported rather than being forced to 'closed'. No
+		// content is disclosed either way (there was none to disclose), just a
+		// comment-status mismatch against the front end for this one combination.
+		if ( \post_password_required( $post ) ) {
+			$data = $response->get_data();
+			if ( empty( $data['content']['rendered'] ) ) {
```
adekbadek: **Blocker:** Password guard treats an absent `content` field as a withheld body — `empty( $data['content']['rendered'] )` is true in three cases, not the two the comment above it names: core withheld the body, core rendered an empty body, and the `content` field is not in the payload at all. Core decides the third — `filter_response_by_context()` strips `content` for `context=embed`, and `rest_is_field_included()` skips it for any `_fields` selection that omits it, both before this hook fires. So on a post that is both gated and password-protected, a caller supplying the correct password gets the early return, and `excerpt.rendered` ships the full body core built while its own `check_password_required` filter was still active. `comment_status` is not forced either.

Reproduced on the review env against a gated, password-protected post whose body carries a sentinel string:

**To reproduce**
1. Publish a post with `post_password` set and bind it to a published gate.
2. `curl "…/wp-json/wp/v2/posts/<id>?context=embed&password=secret"` — `excerpt.rendered` comes back as the full body, sentinel included.
3. `curl "…/wp-json/wp/v2/posts/<id>?_fields=excerpt,comment_status&password=secret"` — same body, and `comment_status` is `open`.
4. `curl "…/wp-json/wp/v2/posts/<id>?context=view&password=secret"` on the same post substitutes correctly, which is what makes the gap easy to miss.

Without the password (step 2 or 3 with it omitted) core emits an empty `excerpt.rendered` and nothing is disclosed, so the precondition is narrow: gated, password-protected, and the caller holds the password. Holding the password is not the gate's entitlement, which is the premise this block already argues.

Testing presence rather than emptiness keeps the no-password and wrong-password cases behaving exactly as they do now, while letting embed and `_fields` reach the substitutions below:

```php
if ( isset( $data['content']['rendered'] ) && '' === $data['content']['rendered'] ) {
```

The "Known gap" note above states "No content is disclosed either way (there was none to disclose)". That holds for the case it describes but not for this one, so it wants narrowing alongside the fix.

plugins/newspack-plugin/includes/content-gate/class-content-gate-advanced-settings.php
```diff
@@ -125,33 +125,20 @@ public static function init() {
...
-		 * restricting posts without publishing a gate or activating Memberships —
-		 * must return true here, or its restricted posts ship in full in the feed.
-		 *
-		 * @param bool $has_restriction_source Whether a first-party restriction source was detected.
-		 */
-		return (bool) apply_filters( 'newspack_content_gate_has_restriction_source', $has_restriction_source );
+		// Memberships is independent of Audience Management and keeps
+		// enforcing either way, so it isn't gated behind
+		// Content_Gate::is_gating_active() the way the first-party half is.
+		return Memberships::is_active() || Content_Gate::has_first_party_restriction_source();
```
adekbadek: **Suggestion:** `newspack_content_gate_has_restriction_source` lost its force-disable direction — The filter's input narrowed under an unchanged name. It previously received `Memberships::is_active() || ( gating && gates )` and its result was the final answer, so a publisher callback returning `false` stood feed restriction down on any site. It now receives only the first-party half, with `Memberships::is_active()` OR'd in after the filter on this line — so on a Memberships site a `false` return is ignored and the feed keeps restricting.

That is the safe direction and nothing is disclosed, but it is a behaviour change on a public filter that is not in the PR's declared list, and the filter's own docblock at `class-content-gate.php:323` still reads "Every caller of this predicate short-circuits entirely when it's false" — no longer true of this caller.

Either move the `apply_filters()` back to wrap the combined value inside `has_first_party_restriction_source()` (the REST path applies its own Memberships deferral unconditionally in `should_restrict_post()`, so it does not depend on the filter seeing that half), or name the Memberships carve-out in that docblock.

**Consider:** the blast radius is wider than a gating opt-in. `Memberships::is_active()` is `class_exists( 'WC_Memberships' ) && function_exists( 'wc_memberships' )` — "is the plugin loaded", with no `NEWSPACK_CONTENT_GATES` check — so this lands on every WooCommerce Memberships site, and the predicate gates five feed paths in this file. A publisher who used the documented opt-out to keep a full-text partner or syndication feed would start shipping truncated items, with no error and no admin notice to attribute it to.

**Risk if not addressed:** If a publisher used the documented opt-out to keep a full-text partner or syndication feed flowing, that feed starts shipping truncated items from the moment the release lands, with no error, no log line and no admin notice to attribute it to — the first signal is the syndication partner noticing.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+	 * keeps the gate off archives and secondary loops on the front end.
+	 *
+	 * @param \WP_Post $post Post to evaluate.
+	 * @return bool
+	 */
+	private static function should_restrict_post( $post ) {
+		if ( ! $post instanceof \WP_Post ) {
+			return false;
+		}
+		if ( ! self::is_newspack_feature_enabled() ) {
```
adekbadek, line 410: **Suggestion:** `should_restrict_post()` adds a feature-flag guard the front-end path did not have — The PR description says `restrict_post()` keeps every guard it had. That holds for the query guards, but `should_restrict_post()` also brings an `is_newspack_feature_enabled()` check that was not on the front-end path. Before the split, `restrict_post()` went from the query guards straight to `is_post_restricted()`, which is a bare seam — `apply_filters( 'newspack_is_post_restricted', false, $post_id )`, no feature-flag check of its own — so a publisher plugin answering that filter had effect on a site that never defined `NEWSPACK_CONTENT_GATES`.

Measured on the review env, A/B against `origin/main`'s copy of this file, with the constant undefined and an mu-plugin returning `true` from `newspack_is_post_restricted`: `origin/main` wraps the body in `newspack-content-gate__gate` markup and suppresses the comment form; this branch renders neither, and the comment form comes back. The rendered body is byte-identical in both — with no gate published there is no layout to truncate against — so what the guard costs is the gate wrapper and the comment lockdown, not body disclosure.

This is the same third-party seam `has_first_party_restriction_source()`'s docblock goes out of its way to keep escapable, so losing it on the front-end path reads as incidental to the refactor rather than chosen. If it is chosen, the constant check belongs in `filter_rest_response()` — where it already is, at line 513 — rather than in the shared decision, and this method's docblock should say the flag is new to `restrict_post()`; it currently lists the four things it holds without noting that one of them changes an existing caller.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+	 * serving that page's post ungated on the public front end. See
+	 * tests/unit-tests/content-gate/class-content-gate-rest.php's
+	 * test_in_process_rest_dispatch_during_page_render_does_not_disarm_front_end_gating(),
+	 * which fails against that shared-lock design and passes against this one.
+	 *
+	 * @param \WP_Post $post Post to evaluate.
+	 * @return array|null Array with 'teaser' and 'gate' keys, or null when the
+	 *                    post is not restricted for the current user.
+	 */
+	public static function get_restriction_for_post( $post ) {
```
adekbadek, line 488: **Suggestion:** Docblocks argue rejected designs rather than describe the code — The 90 new lines of production code carry roughly 197 lines of comment — 2.19:1, against this file's own 0.82:1 — and much of the surplus is design history rather than invariant. `get_restriction_for_post()` spends about 30 lines on a five-line method, most of it the case against an alternative implementation ("This split exists because the more obvious design … is unsound"). The seven-line comment at line 690 narrates why the refactor calls two methods instead of one. Lines 543-551 spend nine lines on what would happen if the premium-newsletter CPT were REST-exposed, and state in the same breath that it is not.

There is no "before" and no "split" for someone reading the merged file — only the code as it stands. What survives is short and genuinely load-bearing: this wrapper must not claim the render lock, because `rest_prepare_{$post_type}` fires for in-process dispatches too and claiming it there would let a concurrent front-end render's own `restrict_post()` bail. Keep that, keep the named regression test, and let the rejected-design argument live in this PR's discussion where it is already recorded.

One factual note while you are in there: the docblock names co-authors-plus, newspack-network's hub Woo store and newspack-community's moderation table as the dispatchers that make the hazard real, but the hazard needs one of them to dispatch a *posts* route, and neither the docblock nor the regression test shows that any of them does — the test constructs the sequence with a direct `rest_get( '/wp/v2/posts/…' )`. The constraint stands without the inventory.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+			return null;
+		}
+		return self::build_restriction( $post );
+	}
+
+	/**
+	 * Register the REST response filter for every post type exposed in REST.
+	 *
+	 * The `rest_prepare_{$post_type}` hook fires from WP_REST_Posts_Controller.
+	 * A post type declaring its own rest_controller_class never fires it; none
```
adekbadek: **Suggestion:** The `rest_controller_class` census is already false on `main` — "A post type declaring its own rest_controller_class never fires it; none do in this workspace" is wrong on both halves. `Institution::register_post_type()` sets `'rest_controller_class' => Institution_REST_Controller::class` in this same directory on `main` (`includes/content-gate/class-institution.php`), and that controller extends `\WP_REST_Posts_Controller`, so it fires `rest_prepare_np_institution` like any other. The branch predates that change, which is why the census read true when it was written; a rebase surfaces it.

No live hole — `np_institution` reads are capability-gated and institutions are never gate targets. The general statement is the part worth fixing: a custom controller escapes this hook only when it does *not* descend from `WP_REST_Posts_Controller`. Stating the boundary as a rule and dropping the census also stops the comment going stale the next time someone registers a post type.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+		// The restriction filter is not a pure predicate: metering records
+		// consumption as a side effect of granting access. A collection read
+		// would spend one view per item for articles the reader never opened,
+		// so metering is short-circuited for the whole REST path.
+		$short_circuit = static function () {
+			return true;
+		};
+		add_filter( 'newspack_content_gate_metering_short_circuit', $short_circuit );
+		try {
+			$restriction = self::get_restriction_for_post( $post );
```
adekbadek, line 610: **Suggestion:** Every restricted item in a collection renders the gate layout from scratch — `build_restriction()` calls `get_inline_gate_html( $post->ID )`, which runs the layout post through `newspack_gate_content` — `do_blocks()`, `wp_filter_content_tags()`, `do_shortcode()`, `wpautop()` — once per restricted item. A `per_page=100` read of a gated category pays 100 of those, and the response is now `no-store`, so a headless front end pays it on every request with no shared cache to absorb it.

Bounded by the guards already in place: only sites with a published gate reach this, and only items that actually restrict pay it. The gate HTML is a pure function of the resolved layout ID within one request — the reader is fixed and `annotate_gate_ctas()` is deterministic — so a request-scoped static in `get_inline_gate_html()` keyed on the layout ID collapses the collection case to one render. `get_restricted_post_excerpt()` necessarily stays per post.

**Consider:** narrowing the `no-store` opt-in in the same pass, or adding it only once the entitlement check has been reached for a gate-eligible type. `register_rest_filters()` hooks every `show_in_rest` post type, and core's default for `rest_send_nocache_headers` is "only when a reader is logged in" — so on a gating site this converts anonymous REST reads of attachments, reusable blocks and any plugin CPT from cacheable to uncacheable, which is a different cost from the per-item evaluation already tracked as a follow-up.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+		// above does not cover it.
+		$data = $response->get_data();
+		if ( isset( $data['content']['rendered'] ) ) {
+			$data['content']['rendered'] = $restriction['teaser'] . $restriction['gate'];
+		}
+		if ( isset( $data['excerpt']['rendered'] ) ) {
+			$data['excerpt']['rendered'] = $restriction['teaser'];
+		}
+		if ( isset( $data['comment_status'] ) ) {
+			$data['comment_status'] = 'closed';
```
adekbadek, line 652: **Suggestion:** `comment_status: closed` is not backed by the comments endpoint — The payload now reports comments closed, but nothing enforces that on `/wp/v2/comments`. Both front-end enforcers key on state REST never sets: `filter_comments_open()` and `filter_comments_array()` require `self::$is_content_locked`, which only `restrict_post()` sets, and also compare against `get_queried_object_id()`, which is `0` during a REST dispatch. So `GET /wp/v2/comments?post=<gated_id>` returns every approved comment on a post whose front-end thread is hidden, and `WP_REST_Comments_Controller::create_item_permissions_check()` gates creation on `comments_open( $post->ID )`, which still returns true — an anonymous caller can post a comment the same API has just told them is not accepted.

No article body is disclosed, and it predates the branch, so this is parity rather than a second instance of the reported defect. Worth either covering (a `rest_comment_query` filter plus a permissions guard keyed on the same restriction decision) or naming as a deliberate scope line the way `register_rest_filters()`'s docblock names its own boundary — forcing `comment_status` here signals an intent to reach comment parity that the response does not deliver.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+		$this->open_post_id = self::factory()->post->create(
+			[
+				'post_status'  => 'publish',
+				'post_title'   => 'Open article',
+				'post_content' => '<!-- wp:paragraph --><p>Freely readable.</p><!-- /wp:paragraph -->',
+			]
+		);
+
+		$this->gate_id = Content_Gate::create_gate( [ 'title' => 'REST Gate' ] );
+		Content_Gate::update_gate_settings(
```
adekbadek: **Suggestion:** The gate fixture is repeated verbatim five times — `Content_Gate::update_gate_settings()` appears here and at lines 367, 470, 668 and 968 as near-identical 25-line literals, differing only in `content_rules.value` and, in two cases, the `metering` / `custom_access` blocks — roughly 90 lines in which what each test actually varies is invisible.

A `publish_gate( array $post_ids, array $overrides = [] )` helper would make the varying part the only part a reader has to read.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+			'A password-protected post must stay empty, not be replaced with a gate teaser.'
+		);
+	}
+
+	/**
+	 * Supplying the correct password to a password-protected AND gated post
+	 * must still withhold the gated body and serve the teaser, not the full
+	 * content.
+	 *
+	 * Core's WP_REST_Posts_Controller::get_item() temporarily clears
```
adekbadek: **Suggestion:** The password-mechanism docblock describes something core does not do — This docblock says core "temporarily clears `$post->post_password` … then restores post_password before firing `rest_prepare_{$post_type}`". Core does neither: `WP_REST_Posts_Controller::prepare_item_for_response()` adds a `post_password_required` filter (`check_password_required()`) and removes it before the hook fires; `$post->post_password` is never written. That is what the production comment at `class-content-gate.php:568` says, so the two contradict each other. The method named here is wrong too — `prepare_item_for_response()`, not `get_item()`.

This is the docblock a reader consults to decide whether the "did core actually withhold the body" check is still needed after a core upgrade, so it should describe the filter add/remove rather than a mutation that does not happen.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+				$item['content']['rendered'],
+				'Every gated item in a collection must withhold its body.'
+			);
+		}
+	}
+
+	/**
+	 * An embed-context response keeps the shape core gives it.
+	 */
+	public function test_embed_context_does_not_gain_a_content_key() {
```
adekbadek: **Suggestion:** The embed test pins core's shape rather than the change — `test_embed_context_does_not_gain_a_content_key()` asserts `content` is absent from an embed response. That passes with this entire PR reverted — core never emits the key in embed context — so it exercises the framework, and it is the file's only embed coverage.

The assertion that would bite is on the field embed does carry: `excerpt.rendered` must be the teaser, not the auto-generated excerpt of the gated body. Adding it here is also what would have caught the password-guard gap on the other thread, since that early return is reached precisely because `content` is absent in this context.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -285,6 +286,355 @@ public static function is_gating_active(): bool {
...
+		// the empty string with a teaser the caller hasn't earned either the
+		// password or the gate for.
+		//
+		// post_password_required() alone can no longer be trusted to mean that,
+		// though: when the correct password IS supplied,
+		// WP_REST_Posts_Controller::prepare_item_for_response() temporarily adds
+		// a `post_password_required` filter (check_password_required(), gated by
+		// can_access_password_content() matching the request's ?password= — see
+		// class-wp-rest-posts-controller.php:1989-1996) so post_password_required()
+		// reports false while content.rendered is built from the real body, then
```
adekbadek: **Nit:** Core line-number citations will rot — `class-wp-rest-posts-controller.php:1989-1996` and `:2051-2053` are accurate against the WordPress in this workspace — I checked both — and will be wrong at the next core release that touches the file, at which point a reader cannot tell whether the mechanism changed or only the line numbers.

`check_password_required()` / `can_access_password_content()` and `prepare_item_for_response()` are already named in the same sentences, are stable, and locate it just as well. The invariant sentence and the "Known gap" paragraph both carry things the code cannot say and should stay.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+	 * very first test's tear_down() wipes register_rest_filters()'s
+	 * rest_prepare_post hook and it never comes back. Re-firing rest_api_init
+	 * per test mirrors what a real REST request does on every request, and
+	 * survives the hook restore regardless of run order.
+	 */
+	public function set_up() {
+		parent::set_up();
+
+		$GLOBALS['wp_rest_server'] = null;
+		do_action( 'rest_api_init', rest_get_server() );
```
adekbadek: **Nit:** `rest_api_init` fires twice per test — `rest_get_server()` already fires `rest_api_init` itself when `$GLOBALS['wp_rest_server']` is null, and PHP evaluates the argument before the call — so the explicit `do_action()` wrapped around it runs the whole registry a second time on each of the 18 tests.

Harmless here (`add_filter()` dedupes on callback plus priority, so `register_rest_filters()` is idempotent), but it also re-runs every other plugin's REST registration per test, and a callback with a non-idempotent side effect would double it. `rest_get_server();` on its own does what the docblock above describes.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+		$headers = $result['headers'];
+
+		if ( isset( $headers['Cache-Control'] ) ) {
+			$this->assertStringNotContainsString(
+				'no-store',
+				$headers['Cache-Control'],
+				'A site with no published gate must not force no-cache headers on REST reads.'
+			);
+		} else {
+			$this->assertArrayNotHasKey(
```
adekbadek: **Nit:** Tautological else branch — Inside the `else { }` of `if ( isset( $headers['Cache-Control'] ) )`, `assertArrayNotHasKey( 'Cache-Control', $headers )` cannot fail — the branch condition already established it. `assertStringNotContainsString( 'no-store', $headers['Cache-Control'] ?? '' )` covers both cases in one assertion.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
```diff
@@ -0,0 +1,1094 @@
...
+	/**
+	 * A publisher plugin that restricts a post by answering
+	 * `newspack_is_post_restricted` directly -- with no published Newspack
+	 * gate and no Memberships -- ships that post unrestricted over REST
+	 * unless it also opts back in via `newspack_content_gate_has_restriction_source`,
+	 * exactly mirroring the feed path's own documented, escapable guard (see
+	 * Test_Feed_Restriction::test_restriction_source_filter_opts_a_gateless_site_back_in()
+	 * and Content_Gate::has_first_party_restriction_source()'s docblock).
+	 *
+	 * Regression test for a real bug in an earlier version of this guard: it
```
adekbadek: **Nit:** Test docblocks narrate the branch's own development — Several passages describe how the code came to be rather than what the test guarantees: "Regression test for a real bug in an earlier version of this guard" here (the bug existed only mid-branch, so it never shipped); "which no longer calls this mid-test" at line 157; "Confirmed empirically: moving that `add_filter()` call to after the `null === $restriction` check … did not fail" at line 643; "Before the guard `filter_rest_response()` checks existed…" at line 791; "the wrong-post bug this replaces" at line 1084. And in `content-gates.php:1266`, "see the coordinator's review" cites a review a future reader of that file cannot consult.

The claims underneath are all worth keeping — that the filter seam must be honoured over REST; that only an *anonymous* entitled reader exercises the plugin's own `rest_send_nocache_headers` opt-in rather than core's `is_user_logged_in()` baseline (that one at line 636 is the counter-example: it carries a constraint the code cannot, and should stay); that a positive post ID compared against an all-non-positive exclusion array cannot fail, which is why the test reaches the private method by reflection. State those and drop the provenance.

adekbadek (changes requested):
Release risk: HIGH — one confirmed gate bypass remains in the new REST filter, reproduced on a review env. The fix itself verifies end to end against an isolated env, including the render-lock split under an in-process REST dispatch. Requesting changes on that one thread; the rest are suggestions and nits.

- `plugins/newspack-plugin/includes/content-gate/class-content-gate.php:666` — **Nit:** Three page exclusions now live in two lists — `restrict_post()` keeps `is_privacy_policy()`, the terms-page ID comparison and the accessibility-page ID comparison, then calls `should_restrict_post()`, which runs `is_excluded_from_gating()` over the same three plus the WooCommerce page IDs. `is_excluded_from_gating()`'s docblock explains why the drift between the two lists is safe, but not that a maintainer adding a fourth exclusion has to add it in both places. Dropping the three now-duplicated guards here and keeping only the genuinely query-context-only ones — `is_admin()`, `is_account_page()`, `is_cart()`, `is_checkout()` — would leave one list plus a stated reason for the remainder.

Review provenance: 4 of 5 reviewer passes (lineup: deep, standards, wp-expert, codex, opencode:kimi-k3); release risk: HIGH (degraded — skipped: codex).

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Reproduced on the env, and we went a different way than the suggested line.
`isset( … ) && '' === …` does close embed and `_fields`, but it also stops the
guard firing when there is no password at all, and the gate then writes its
teaser into `excerpt.rendered` for a caller who gets nothing from the front end.
Measured: embed with no password goes from empty to the teaser.

The guard now asks the controller that prepared the response, through
`can_access_password_content()` — the predicate core used to decide whether to
withhold in the first place. The assumption worth pushing on is the fallback: a
post type served by something that is not a `WP_REST_Posts_Controller` has no
such predicate, so it is reported as withheld, which leaves core's own output
alone.

Three tests. Two fail on the sentinel with the fix reverted; the third fails
under the one-liner, which is what it is there for. The known-gap note is
narrowed to the case that still returns early.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Gone. That block was rewritten for the blocker above and the two citations went
with it; the methods are named instead.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Corrected in `e2c9b13164`, which predates this round and was sitting unpushed.
`np_institution` is named as the one instance, with why nothing is lost there
(`public => false`, and its route is gated to an editing capability) and what
would make it a real gap.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Split this one, because the two halves measured differently.

Creation is fixed. The case is narrower than an anonymous caller, though: core
refuses anonymous REST comment creation through `rest_allow_anonymous_comments`
whatever `comments_open()` says, so the caller who gets through is a logged-in
reader the gate turns away. That now returns 403, hooked on
`rest_pre_insert_comment` rather than on `comments_open()` — widening that would
reach the admin comment screens and every front-end call for a post other than
the one rendering. Two tests, and the unentitled one goes 201 to 403 with the
hook removed.

Reads are filed as https://linear.app/a8c/issue/NPPM-3235. Dropping restricted
comments after the query desynchronises `X-WP-Total` and the pagination headers,
and excluding them in the query needs the restricted set computed up front,
which is a gate evaluation per candidate post. The issue also carries the
question worth settling before either: whether a gated post's comments should be
hidden over REST at all, or only its body.

plugins/newspack-plugin/includes/content-gate/class-content-gate-advanced-settings.php
jason10lee: Right about the contract, and we have restored it: `detect_first_party_restriction_source()`
is the raw predicate, `has_first_party_restriction_source()` stays the filtered
one for REST, and the feed path applies the filter to the combined answer. The
filter now resolves once per call site.

What we could not do is write a test for it, because we could not find a
reachable behaviour difference. `get_feed_restriction_mode()` returns
`FEED_MODE_OFF` whenever Memberships is active (NPPM-3204), and every consumer
of `has_restriction_source()` sits behind that mode. Measured with Memberships
active, a restricted post, and the filter forced both ways: mode is `off` in
both arms and the feed comes back identical. So the change is a contract repair
with nothing observable behind it today — worth having, but you should know
that is what it is before it reads as a bug fix.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Checked against the merge base and you are right about the code: `restrict_post()`
went straight to `is_post_restricted()`, and `init()` hooks `the_post`
unconditionally, so a site with the flag off does reach it.

It makes no difference to a reader, though. With `NEWSPACK_CONTENT_GATES`
undefined and a publisher plugin answering `newspack_is_post_restricted`, the
post renders in full either way: removing the guard flips
`should_restrict_post()` to true, and the page still serves the whole body,
because nothing downstream substitutes while the feature is off. So the guard
un-gates nothing, and it makes the shared predicate agree with
`is_gating_active()` and with what an external consumer is told. Left in.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Cut. `get_restriction_for_post()` loses the case against the alternative design
and keeps the mechanism plus the test that holds it; the render-lock comment
goes from seven lines to two; the premium-newsletter passage from eight to
three. The block you quoted hardest was already rewritten for the blocker.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
jason10lee: Confirmed, and filed as https://linear.app/a8c/issue/NPPM-3236. Counted
`newspack_gate_content` firings across a collection read: it tracks restricted
items rather than firing once, so the cost is as you describe. Fixing it means
memoizing rendered layout per gate for the request, which needs an invalidation
story of its own — more than this PR should carry.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: `configure_gate()` now takes the posts and any registration overrides; four of
the six sites are one line. The metering and custom_access ones stay written
out, since what they vary is the point.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: Rewritten to what core does: it adds a `post_password_required` override while
it builds the body and removes it before the hook, never touching
`$post->post_password`. It had also come to contradict the production comment in
this same PR, which says exactly that.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: Deleted. The two new embed tests exercise the change, so this is no longer the
file's only embed coverage.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: Dropped the `do_action()` wrapper — nulling the global is enough, and
`rest_get_server()` fires `rest_api_init` itself.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: One assertion with `?? ''`.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-content-gate-rest.php
jason10lee: All five rewritten in the present tense, including the `content-gates.php` one
citing a review — that now states the reasoning itself.

jason10lee:
Done. `restrict_post()` keeps only the guards that need a query — `is_admin()`,
`is_account_page()`, `is_cart()`, `is_checkout()` — and the three that can be
decided from the post ID now live only in `is_excluded_from_gating()`.

No test: the behaviour is identical by construction, so nothing can go red.
Verified by rendering the privacy-policy and accessibility pages anonymously
before and after (full body both times, with a plain gated page and a gated post
still gated as controls). The Terms and Conditions guard is the one we could not
exercise — it needs `wc_terms_and_conditions_page_id()`, so with no WooCommerce
on the env it was inert either way, and that one rests on reading the code.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Confirmed, and the objection to the one-liner is right: with no password in embed context there is no `content` key, `isset( … ) && '' === …` is false, and the substitution would have written the teaser into `excerpt.rendered` for a caller core gave nothing. `test_embed_context_without_the_password_is_left_to_core()` pins that.

`can_access_password_content()` is the right predicate to defer to — it is exactly what `prepare_item_for_response()` consults before adding the `post_password_required` override, so the guard now reproduces core's own decision rather than inferring it from the payload. The cookie case falls out correctly too: `post_password_required()` is already false when `wp-postpass_` matches, so the guard is skipped and the substitution runs, matching the full body core served.

The `WP_REST_Posts_Controller` fallback is the assumption I would keep an eye on, and reporting a foreign controller as withheld is the right default — a post type served by something else has no password contract for this filter to reason about, so leaving core's output alone is the only answer that cannot disclose.

plugins/newspack-plugin/includes/content-gate/class-content-gate-advanced-settings.php
adekbadek: Contract restored — `detect_first_party_restriction_source()` raw, `has_first_party_restriction_source()` filtered, and the feed path applying the filter to the combined value. A callback returning `false` stands feed restriction down on a Memberships site again.

No test needed. `get_feed_restriction_mode()` returning `FEED_MODE_OFF` under Memberships means there is nothing observable to assert against today, and a test written to the internals rather than to behaviour would only pin the implementation.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: Fair call, and agreeing the shared predicate with `is_gating_active()` is the better invariant.

Half the measurement stands, though: the body renders identically either way, but the gate wrapper and the comment lockdown do not. On `origin/main` with `NEWSPACK_CONTENT_GATES` undefined and a plugin answering `newspack_is_post_restricted`, the body is wrapped in `newspack-content-gate__gate` and the comment form is suppressed; on this branch neither happens and the form comes back. That is a behaviour change for a site in that configuration, not a no-op — one nobody is known to run, and in the direction the flag is supposed to mean, so noting it for the record rather than asking for another pass.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: `rest_pre_insert_comment` is the right seam and the 403 lands: `prepare_item_for_database()` applies this filter last, and `create_item()` returns the `WP_Error` straight through. Narrowing to logged-in callers is right too.

One thing to tidy: the guard calls `get_restriction_for_post()`, which runs `build_restriction()` — the gate layout through `do_blocks()`/`do_shortcode()` plus the teaser — then discards both to read a boolean. `should_restrict_post()` is private in the same class and answers exactly that question at no render cost. It also makes the metering posture consistent: `filter_rest_response()` wraps its call in `newspack_content_gate_metering_short_circuit` and this one does not, which is harmless today only because `get_gate_post_id()` will not resolve a gate outside `is_singular()`.

The filter also covers `PUT /wp/v2/comments/<id>` when the request carries `post`, since `prepare_item_for_database()` is shared with `update_item()`. Unreachable in practice — editing someone's comment needs `edit_comment`, which maps through `edit_post`, and `is_post_restricted()` returns false for anyone holding that — but in scope.

NPPM-3235 for reads, with the "hide a gated post's comments at all, or only its body" question attached, is the right shape.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
adekbadek: NPPM-3236 covers the per-item render. The second half of this thread is not in it: `register_rest_filters()` hooks every `show_in_rest` post type, so on a site with a published gate an anonymous `/wp/v2/media` read reaches the `rest_send_nocache_headers` opt-in and comes back uncacheable, where core's default would have cached it. Same class of cost, same fix window — worth adding to that issue rather than a separate one.

adekbadek (approved):
The gate bypass is closed, and deferring to `can_access_password_content()` is better than the suggested one-liner: it reproduces core's own withheld/not-withheld decision instead of inferring it from the payload, and `test_embed_context_without_the_password_is_left_to_core()` pins the disclosure that one-liner would have introduced. Every other thread is addressed, declined with a stated reason, or filed as NPPM-3235 / NPPM-3236.

Two non-blocking notes inline: the comment guard renders a gate layout it discards where `should_restrict_post()` would answer the same question, and the `no-store` opt-in reaching every `show_in_rest` post type belongs on NPPM-3236 alongside the per-item render.

PR #838, opened 2026-08-10
fix(content-gate): check both gate kinds in premium-list fast path

dkoo:
### All Submissions:

`main` is currently red: `Gating_Inertness_Test::test_premium_lists_reappear_in_signup_forms_while_inert` fails, so every PR's CI inherits the failure.

The cause is the fast path #769 added to `Premium_Newsletters::filter_subscription_lists()`:

```php
// If there are no premium newsletter gates, no list can be restricted, so
// skip the per-list check entirely.
if ( empty( self::get_gates() ) ) {
    return $lists;
}
```

`self::get_gates()` resolves to `Content_Gate::get_gates( GATE_CPT, 'publish', true )`, and that third argument **selects between** gate kinds rather than widening to both — the meta query is `EXISTS` vs `NOT EXISTS` on `is_newsletter`. So it returns premium newsletter gates only.

The per-list check it skips runs through `Content_Restriction_Control::is_post_restricted()`, which honours **any** gate whose content rules match the list. So on a site with ordinary content gates and no premium newsletter gates, the fast path returns early and lists that would have been filtered are handed back unfiltered. The premise in the comment — no premium newsletter gates means no list can be restricted — doesn't hold.

This makes the early return require both kinds to be absent. The optimization still fires for the common case of a site with no gates at all, and both calls are statically cached, so the fast path stays cheap.

Worth noting the failure direction is permissive: lists that should be hidden from signup forms are shown. How reachable that is in production depends on gate rules — real lists are the `newspack_nl_list` CPT, so a publisher needs a gate whose content rules match that post type, which is narrower than the test fixture (a plain `post`) suggests. The code paths genuinely diverge either way.

Not a flake, and not caused by #837: filtering CI runs to those that actually executed PHPUnit, `24c3870a3` was green and `00a9a791e` red, with #769 arriving between them via the release merge. Runs in between that look green never ran the job.

1. On `main`, run the failing test — it fails as a single isolated test, no ordering needed:
   ```
   vendor/bin/phpunit --filter test_premium_lists_reappear_in_signup_forms_while_inert
   ```
2. On this branch, run it again. It passes.
3. Run the full suite on both. `main` reports 1 failure; this branch is clean.
4. Sanity-check the fast path still short-circuits: on a site with no gates at all, `Premium_Newsletters::filter_subscription_lists()` returns its input without running the per-list check.

### Other information:

No new test: `Gating_Inertness_Test::test_premium_lists_reappear_in_signup_forms_while_inert` already covers this exactly, and is the test that caught it.

Before, on `main` @ `51f9172f3`: `Gating_Inertness_Test` — 10 tests, 1 failure. After, on this branch: clean. Full suite on this branch: 2844 tests / 8363 assertions / 0 failures / 10 environment skips, matching CI's counts. PHPCS clean against the root ruleset.

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
@@ -167,8 +167,9 @@ class Premium_Newsletters {
 		// than widening to both — the meta query is EXISTS vs NOT EXISTS — while the
 		// per-list check below runs through Content_Restriction_Control, which
 		// honours any gate whose content rules match the list. Testing only the
-		// newsletter half would skip lists an ordinary gate restricts. Both calls are
-		// statically cached, so the fast path stays cheap.
+		// newsletter half would skip lists an ordinary gate restricts. In non-test
+		// requests both gate lookups are cached (Content_Gate::get_gates cache is
+		// disabled under PHPUnit), so the fast path stays cheap.
 		if ( empty( self::get_gates() ) && empty( Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish' ) ) ) {
 			return $lists;
 		}
--- a/plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
@@ -161,9 +161,15 @@ class Premium_Newsletters {
 			return $lists;
 		}
 
-		// If there are no premium newsletter gates, no list can be restricted, so
-		// skip the per-list check entirely. get_gates() is statically cached.
-		if ( empty( self::get_gates() ) ) {
+		// With no gates at all, no list can be restricted, so skip the per-list
+		// check entirely. Both kinds have to be absent: get_gates()'s $is_newsletter
+		// argument selects between premium newsletter gates and ordinary ones rather
+		// than widening to both — the meta query is EXISTS vs NOT EXISTS — while the
+		// per-list check below runs through Content_Restriction_Control, which
+		// honours any gate whose content rules match the list. Testing only the
+		// newsletter half would skip lists an ordinary gate restricts. Both calls are
+		// statically cached, so the fast path stays cheap.
+		if ( empty( self::get_gates() ) && empty( Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish' ) ) ) {
 			return $lists;
 		}
 


plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
```diff
@@ -161,9 +161,16 @@ public static function filter_subscription_lists( $lists ) {
...
-		// If there are no premium newsletter gates, no list can be restricted, so
-		// skip the per-list check entirely. get_gates() is statically cached.
-		if ( empty( self::get_gates() ) ) {
+		// With no gates at all, no list can be restricted, so skip the per-list
+		// check entirely. Both kinds have to be absent: get_gates()'s $is_newsletter
+		// argument selects between premium newsletter gates and ordinary ones rather
+		// than widening to both — the meta query is EXISTS vs NOT EXISTS — while the
+		// per-list check below runs through Content_Restriction_Control, which
+		// honours any gate whose content rules match the list. Testing only the
+		// newsletter half would skip lists an ordinary gate restricts. In non-test
```
jason10lee, line 170: **Suggestion:** The comment describes a mechanism the code does not have — This comment is the justification for the added condition, so it is worth getting exactly right — and as written we think it inverts how gate selection works.

It says the per-list check "runs through `Content_Restriction_Control`, which honours **any** gate whose content rules match the list". `Content_Restriction_Control::get_post_gates()` picks the bucket from the post type first, before any content rule is consulted:

```php
$is_newsletter = false;
if ( class_exists( 'Newspack\Newsletters\Subscription_Lists' ) && Subscription_Lists::CPT && get_post_type( $post_id ) === Subscription_Lists::CPT ) {
 $is_newsletter = true;
}
$gates = Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish', $is_newsletter );
```

For a `newspack_nl_list` post that is always `true`, so ordinary gates are never fetched for a list, never mind matched against it. The sentence is true only of the test's plain-`post` fixture.

Our worry is what it leaves behind: the next reader of the file that decides which lists a reader sees comes away believing ordinary gates reach subscription lists. Could we say what is actually load-bearing — that the bucket selection in `get_post_gates()` is keyed on post type, and the fixture passes a non-CPT post, so the fast path has to stay permissive enough to reach the per-list check for such inputs? That would also let the comment shrink to two or three lines.

plugins/newspack-plugin/includes/content-gate/class-premium-newsletters.php
```diff
@@ -161,9 +161,16 @@ public static function filter_subscription_lists( $lists ) {
...
+		// requests both gate lookups are cached (Content_Gate::get_gates cache is
+		// disabled under PHPUnit), so the fast path stays cheap.
+		if ( empty( self::get_gates() ) && empty( Content_Gate::get_gates( Content_Gate::GATE_CPT, 'publish' ) ) ) {
```
jason10lee, line 173: **Suggestion:** `fix(` will ship a publisher-facing changelog entry for a change publishers cannot observe — Since old and new produce identical output in production, this has no publisher-visible effect — it unblocks CI. `AGENTS.md` reserves `feat`/`fix` for publisher-visible change precisely because semantic-release turns them into a patch release and a changelog line; everything else goes to `chore`/`test`/`ci`.

As `fix(content-gate):` this cuts a patch release whose notes tell publishers a content-gating bug was fixed, which would be the second time the description travels further than the change does. `chore(content-gate):` or `test(content-gate):` would match what actually happened.

No new commit needed — the squash title is editable at merge time.

jason10lee (approved):
Approved with comments!  This unblocks `main`, verified end-to-end.  Agentic review suggests the production bug described in the PR body may not be reachable, and the new comment states a mechanism the code doesn't have, neither worth holding up the fix.

Release risk: LOW — behavior-preserving in production, full suite green, CI green.

Further analysis:

`Content_Restriction_Control::get_post_gates()` picks the gate bucket from the *post type* before any content rule is consulted: a `newspack_nl_list` post only ever loads `is_newsletter EXISTS` gates. Every list reaching this filter is that CPT — `Subscription_Lists::get_all()` queries it exclusively, and it is the filter's only applier. A publisher cannot even route around it: `newspack_nl_list` is registered `show_in_rest => false`, and gate content rules only offer post types with `show_in_rest => true`. So an ordinary gate can never restrict a real list, and old and new code return identical output in production — the new conjunct only adds a lookup.

We localized the failure to the fixture: it pairs an ordinary gate with a plain `post` standing in for a list, a combination production cannot produce. That means the test has been passing vacuously with respect to the contract its docblock claims.

- `plugins/newspack-plugin/tests/unit-tests/content-gate/class-gating-inertness-test.php:189` — **Suggestion:** The fixture, not the fast path, is what diverges from production — Worth considering whether the fixture is the better place to fix this. The test creates an **ordinary** gate (`create_registration_gate()` defaults `$is_newsletter = false`) with a `post_types => ['post']` rule, and builds its "list" from `self::factory()->post->create()` — a plain `post` wrapped in an anonymous stub. Production cannot produce that pairing: `Subscription_Lists::get_all()` queries `newspack_nl_list` exclusively, and `newspack_nl_list` is registered `show_in_rest => false` while gate content rules only offer post types with `show_in_rest => true`. So as written the test does not guard the contract its own docblock claims ("a restricted list stops being filtered out") — it exercises a path production never takes. The sibling suite already builds the faithful shape: `Newspack_Test_Premium_Newsletters` creates its list as `Subscription_Lists::CPT` wrapped in a real `Subscription_List`, with a newsletter gate. The obstacle is real, though, and we do not want to hand-wave it: this file deliberately loads only `newsletters-mocks.php`, because the namespaced mock "defines a partial `Subscription_List` that breaks `Test_Newsletters_Access`". Without `Subscription_Lists`, `$is_newsletter` can never become true here. If that turns out to be genuinely blocking, keeping this widening as a deliberate concession is defensible — but then the comment should say so plainly, which is the ask above.
- `plugins/newspack-plugin/tests/mocks/newsletters-namespaced-mocks.php:25` — **Nit:** Mock CPT constant disagrees with the real one (pre-existing) — Not introduced here, and only relevant if the fixture above gets revisited: the mock declares `const CPT = 'np_newsletter_list'` while the real value is `newspack_nl_list`. Only equality is compared, so nothing breaks today — but since the whole question in this PR turns on a post-type comparison, a mock that disagrees with the thing it mocks is a trap for whoever next tries to reproduce this path in tests.
- `plugins/newspack-plugin/tests/unit-tests/content-gate/class-gating-inertness-test.php:96` — **Nit:** `Premium_Newsletters::$gates` is never reset between cases in this class — `reset_restriction_caches()` clears the three `Content_Restriction_Control` maps and flushes the gate cache, but not `Premium_Newsletters::$gates` — which memoizes unconditionally and is *not* subject to `is_gates_cache_enabled()`, so under PHPUnit it survives for the life of the process. The sibling suite resets it by reflection; this one does not. Nothing is wrong today: the first conjunct evaluates to "no newsletter gates" and stays there, and the assertions ride on the second. But it does mean a future case in this class that creates a newsletter gate would silently not see it. A reflection reset alongside the existing three would close that off.

dkoo:
Thanks, @jason10lee!

PR #303, opened 2026-06-13
ci: enforce SCSS lint per changed package (NPPM-290 Phase 2)

jason10lee:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guidelines](https://github.com/Automattic/newspack-workspace/blob/main/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Adds SCSS linting to CI as a required, non-bypassable gate. The `js` job already runs `lint:js`, `build`, and `test` per changed package; this adds a `lint:scss` step alongside `lint:js` so that staged SCSS in any plugin, theme, or package is linted on every PR.

`--if-present` means packages without a `lint:scss` script (e.g. `republication-tracker-tool`, `super-cool-ad-inserter`, the shared packages) are skipped, and the per-package change detection means the step only runs for packages a PR actually touches.

This completes the NPPM-290 enforcement: the pre-commit hook (#299) gives fast local feedback, and this CI gate is the backstop that can't be bypassed. Repo-wide SCSS is already clean (the one remaining item, a deliberate duplicate selector in `newspack-popups`, is annotated in #302).

**Merge ordering:**
- Merge **after #302** — otherwise the next PR touching `newspack-popups` SCSS would fail this gate on that pre-existing (now-annotated) lint.
- Ideally merge **after #299** so contributors have the local pre-commit hook before this hard CI gate begins blocking.

Part of [NPPM-290: Theme linting: PHP & CSS](https://linear.app/a8c/issue/NPPM-290/theme-linting-php-and-css).

### How to test the changes in this Pull Request:

1. Review the diff: the new step mirrors the existing `lint:js` step exactly (same `--if-present` filter and `continue-on-error: false`), inserted before `build`.
2. Note this PR changes only `ci.yml`, so its own run won't exercise the step (the `js` matrix is empty for a workflow-only change). It first runs on the next PR that changes a package's SCSS.
3. To see it gate: on a branch, introduce a deliberate SCSS error in any package with a `lint:scss` script (e.g. an empty block `.x {}`) and open a PR — the `JS / <package>` check fails on the `lint:scss` step. Reverting the error makes it pass.
[...]

PR #839, opened 2026-08-10
build(lint): make SCSS lint clean workspace-wide (NPPM-290)

jason10lee:
### All Submissions:

`pnpm -r run lint:scss` fails today on two packages: 62 errors in `newspack-plugin` and one in `newspack-newsletters`. That blocks #303, which adds the `lint:scss` step to CI, since turning the gate on would fail the next PR touching either package.

Most of the findings are the linter reading code it has no way to resolve, and `plugins/newspack-plugin/.stylelintrc.js` already covers all three cases. Two of its lists predate later additions to the colors package:

- `to-rgba` joins `to-rgb` in `ignoreFunctions`. Both are SCSS `@function` declarations in `packages/colors/colors.module.scss`, at lines 9 and 13. `to-rgb` was already listed; `to-rgba` has twelve call sites at lines 175–186, and those twelve are the `function-no-unknown` errors this removes. Stylelint resolves CSS functions only, so a locally-declared one reads as unknown.
- `/^overlay-/` joins the eight token families in `ignoreProperties`. These are entries in a CSS Modules `:export` block, which hands named values to JavaScript rather than declaring CSS properties.

That rc file is the only lint path that reaches the shared `packages/`: its glob is `'**/*.scss'`, and `plugins/newspack-plugin/packages/{colors,components,icons}` are symlinks tracked in git pointing at the workspace top level, while the shared packages define no `lint:scss` of their own. A second consumer of these tokens would need the same two entries.

The two `--wpds-*` findings in `packages/components/src/tabbed-navigation/style.scss` are a version gap rather than wrong token names. `@wordpress/stylelint-config@23.40.0` requires `@wordpress/theme` at `^0.15.0`, whose registry predates both tokens; `packages/components` depends on `^0.16.0`, where both exist. A pnpm override scoped to that subtree, `@wordpress/stylelint-config>@wordpress/theme`, points the rule at the newer registry. Scoping it to the subtree rather than the workspace leaves `packages/components`' own dependency free to move.

**Removing the override:** it stops being necessary once `@wordpress/stylelint-config` requires `@wordpress/theme` at 0.16 or later — 24.x requires `^1.1.0`. Until then it holds the rule's registry at 0.16.x, so a future `@wordpress/stylelint-config` major has to drop this entry in the same change, or the rule reads a registry older than the config expects.

That leaves five genuine findings, all `color-hex-length`: four eight-digit values in `packages/colors/colors.module.scss` and one in `plugins/newspack-newsletters/src/editor/style.scss`, shortened to their equivalent forms. Each is a doubled-nibble pair (`#00000033` → `#0003`), so the expansions are byte-identical; the non-reducible values were left alone.

No rendered output changes. The shortened hex values are equivalent, and the config changes affect only what stylelint reports. Prep for #303 (NPPM-290).

Verified across the workspace, not just the two edited packages: `pnpm -r --no-bail run --if-present lint:scss` exits 0, with all twelve packages that define a `lint:scss` script passing.

1. Check out the branch and run `pnpm install`. This applies the override, so `@wordpress/stylelint-config` resolves `@wordpress/theme` at 0.16.0.
2. Run `pnpm -r --no-bail run --if-present lint:scss`. It exits 0 and all twelve packages with a `lint:scss` script report `lint-style complete!`.
3. Run `pnpm run build:packages`, which regenerates `packages/components/colors/colors.module.scss` from the corrected source, then repeat step 2. It still exits 0.
4. For the before state, check out `main` and repeat steps 1 and 2. `newspack-plugin` reports 62 errors and `newspack-newsletters` reports 1.

### Other information:


dkoo (approved):
Works!

PR #840, opened 2026-08-10
feat(subscriptions): add a card to a manually created subscription

dkoo:

WooCommerce Subscriptions only offers the "change payment method" action when a subscription has a next-payment date and is active (`WC_Subscriptions_Change_Payment_Gateway::can_subscription_be_updated_to_new_payment_method()`). A subscription created by hand in wp-admin never gets a next-payment date, so the reader is offered no way to put a card on file and the subscription can never be paid. Part of NPPD-2170.

This opens the payment-method flow for exactly that gap, and no wider.

- **Eligibility** — `WooCommerce_Subscriptions::allow_add_payment_method_without_next_payment()`, a priority-20 filter on `woocommerce_can_subscription_be_updated_to_new-payment-method`. It reproduces every refusal WCS applies except the missing-next-payment / not-active one it targets: store-wide manual renewals, zero recurring total, no gateway that can take a customer card, a gateway that cannot cancel. Limited to `pending` / `on-hold` / `active`. `pending-cancel` is excluded — its next-payment date is always cleared, so the action would be a no-op for a subscription winding down (reactivating restores WCS's own flow).
- **Follow-through** — `schedule_next_payment_after_payment_method_added()`, on `woocommerce_subscription_payment_method_updated`. For an `active` subscription with no next-payment date it schedules the next payment, which both resumes billing and unlocks WCS's early renewal ("Renew now"), since `wcs_can_user_renew_early()` refuses on `subscription_no_next_payment`. Gated on `can_date_be_updated( 'next_payment' )`, so a gateway that owns the billing agreement and refuses date changes never gets a Woo-side schedule it won't honour. `pending` / `on-hold` are left to WooCommerce, which sets the date when their outstanding order is paid.
- **Label** — `subscription-header.php` now reads "Add payment method" when the subscription has no gateway yet and "Update payment method" when it does, matching WCS's own add-vs-change wording instead of a hardcoded "Update payment method".
- **Modal checkout** — `Checkout_Data::get_checkout_data()`. A subscription with no parent order now reads its own line items instead of calling `get_items()` on the `false` that `get_parent()` returns (a fatal), and returns empty for an item-less order rather than fatalling on a null product. The subscription is identified by `subscription_ids`, never as an `order_id`, so the post-checkout redirect lands on the subscription page rather than a view-order URL.

1. With WooCommerce Subscriptions and a subscriptions-capable gateway (e.g. Stripe test mode) active, create a subscription in wp-admin for a subscriber, attach a subscription product, and leave it with no payment method.
2. As the subscriber, open the subscription under My Account. An **Add payment method** action appears. Before this change there is none.
3. Add a card. The modal completes and returns to the subscription page, not a view-order page.
4. For an **active** subscription with no orders: a next-payment date is set afterwards and a **Renew now** action appears; renewing takes payment immediately.
5. For a **pending** subscription: the button remains after adding a card and no next-payment date is written. Payment is collected by paying the pending order.
6. The action does **not** appear on a store configured for manual renewals with the auto-renew toggle disabled, nor on a `pending-cancel` subscription.

Tests: `n test-php` in `newspack-plugin` (`--group WooCommerce_Subscriptions_Integration`) and in `newspack-blocks`.

### Other information:

<sub>🤖 Authored with [Claude Code](https://claude.com/claude-code).</sub>

### DIFF
--- a/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
+++ b/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
@@ -327,12 +327,16 @@ final class Checkout_Data {
 			}
 			$order_items = $items_source->get_items();
 			$order_item  = reset( $order_items ); // Use only the first item in the order.
-			if ( $order_item ) {
-				$product_id   = $order_item->get_product_id();
-				$variation_id = $order_item->get_variation_id();
-				$amount       = $order_item->get_subtotal();
+			if ( ! $order_item ) {
+				// No line item means no purchase to summarise, and the product and
+				// name lookups below would fatal on the resulting null. Return empty,
+				// the same "nothing to checkout" signal as an empty source above.
+				return $data;
 			}
-			$referrer = $items_source->get_meta( '_newspack_referer' );
+			$product_id   = $order_item->get_product_id();
+			$variation_id = $order_item->get_variation_id();
+			$amount       = $order_item->get_subtotal();
+			$referrer     = $items_source->get_meta( '_newspack_referer' );
 		}
 
 		// If we have no referrer, set it to the current path.
--- a/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
+++ b/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
@@ -269,7 +269,10 @@ final class Checkout_Data {
 
 		$cart_item    = null;
 		$order        = null;
+		$subscription = null;
 		$referrer     = '';
+		$product_id   = null;
+		$amount       = 0;
 		$variation_id = null;
 		$is_variable  = false;
 		$is_grouped   = false;
@@ -303,18 +306,33 @@ final class Checkout_Data {
 			$amount       = $cart_item['data']->get_price();
 			$referrer     = $cart_item['referer'] ?? '';
 		} elseif ( $source instanceof \WC_Order ) {
-			// If order as actually a subscription object, we need to get the original order.
+			// A subscription's purchase details normally live on the order it was
+			// bought through. A subscription created by hand in wp-admin has no parent
+			// order, and its own line items describe the same purchase, so read those
+			// instead (NPPD-2170).
+			//
+			// The two identities stay separate on purpose. A subscription is not an
+			// order, and reporting its ID as an order ID sends the modal to a
+			// view-order URL built from a subscription ID: a page nothing links to,
+			// which renders the read-only receipt from WooCommerce Subscriptions
+			// rather than anything the reader can act on.
 			if ( $source instanceof \WC_Subscription ) {
-				$order = $source->get_parent();
+				$subscription = $source;
+				$parent       = $source->get_parent();
+				$order        = $parent ? $parent : null;
+				$items_source = $parent ? $parent : $source;
 			} else {
-				$order = $source;
+				$order        = $source;
+				$items_source = $source;
 			}
-			$order_items  = $order->get_items();
-			$order_item   = reset( $order_items ); // Use only the first item in the order.
-			$product_id   = $order_item->get_product_id();
-			$variation_id = $order_item->get_variation_id();
-			$amount       = $order_item->get_subtotal();
-			$referrer     = $order->get_meta( '_newspack_referer' );
+			$order_items = $items_source->get_items();
+			$order_item  = reset( $order_items ); // Use only the first item in the order.
+			if ( $order_item ) {
+				$product_id   = $order_item->get_product_id();
+				$variation_id = $order_item->get_variation_id();
+				$amount       = $order_item->get_subtotal();
+			}
+			$referrer = $items_source->get_meta( '_newspack_referer' );
 		}
 
 		// If we have no referrer, set it to the current path.
@@ -387,6 +405,13 @@ final class Checkout_Data {
 					}
 				}
 			}
+		}
+
+		// A subscription with no parent order has no order to identify itself by, so
+		// name the subscription directly. Without this the modal has nothing to
+		// return the reader to once checkout finishes (NPPD-2170).
+		if ( $subscription && empty( $data['subscription_ids'] ) ) {
+			$data['subscription_ids'] = [ $subscription->get_id() ];
 		}
 
 		/**
--- a/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
+++ b/plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
@@ -414,7 +414,11 @@ final class Checkout_Data {
 		// A subscription with no parent order has no order to identify itself by, so
 		// name the subscription directly. Without this the modal has nothing to
 		// return the reader to once checkout finishes (NPPD-2170).
-		if ( $subscription && empty( $data['subscription_ids'] ) ) {
+		// Only when there is no parent order to identify the purchase by. With a
+		// parent present the block above owns the key, and widening this would set
+		// subscription_ids for product types it skips — a recurring donation resolves
+		// to `donation`, so it would silently change where those readers land.
+		if ( $subscription && ! $order && empty( $data['subscription_ids'] ) ) {
 			$data['subscription_ids'] = [ $subscription->get_id() ];
 		}
 
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
@@ -421,18 +421,34 @@ class WooCommerce_Subscriptions {
 	 * Schedule the next payment once a card is attached to a subscription that
 	 * had none.
 	 *
-	 * The other half of {@see allow_add_payment_method_without_next_payment()}.
-	 * Eligibility alone would leave the reader with a card on file and still no
-	 * billing: the subscription reached this state precisely because nothing was
-	 * ever scheduled, and WCS only recalculates the date when a payment
-	 * completes. Calculating it here is what turns "card added" into "renewals
-	 * resume".
+	 * The other half of {@see allow_add_payment_method_without_next_payment()},
+	 * and deliberately narrower than it. Eligibility opens the form for four
+	 * statuses; only `active` gets a date written here.
+	 *
+	 * An active subscription with no next payment date is already giving the
+	 * reader access with nothing scheduled to bill, so calculating the date is
+	 * what turns "card added" into "renewals resume" — and it is also what makes
+	 * WCS's early renewal available, since `wcs_can_user_renew_early()` refuses
+	 * on `subscription_no_next_payment`. That matters for a hand-made
+	 * subscription with no orders at all, where early renewal is the reader's
+	 * only self-serve route to paying.
+	 *
+	 * The other statuses are left to WooCommerce, which already handles them
+	 * correctly: paying the pending or failed order is what activates the
+	 * subscription and sets its date. Writing a date for them would be wrong
+	 * twice over — it hands the reader a billing period they never paid for, and
+	 * it withdraws the "Add payment method" action (this pair steps back once a
+	 * date exists, while WCS's own rule still refuses a non-active subscription),
+	 * leaving them with a card on file and no way back to the form.
 	 *
 	 * `calculate_date( 'next_payment' )` derives the date from the last payment
 	 * or the start date plus one billing period, keeps it far enough in the
 	 * future to survive a daylight-saving shift, and returns `0` when the next
 	 * period would fall past the subscription's end date. A `0` is respected:
 	 * a subscription winding down to a fixed end has nothing further to bill.
+	 * A `pending-cancel` subscription therefore stays untouched even though it
+	 * is eligible for the form; resuming its auto-renewal means clearing the
+	 * cancelled and end dates, which is a separate change.
 	 *
 	 * @param \WC_Subscription $subscription       The subscription that was updated.
 	 * @param string           $new_payment_method The gateway ID now attached.
@@ -458,7 +474,9 @@ class WooCommerce_Subscriptions {
 			return;
 		}
 
-		if ( ! $subscription->has_status( [ 'pending', 'on-hold', 'active', 'pending-cancel' ] ) ) {
+		// Active only. See the note above on why the other eligible statuses are
+		// left to WooCommerce's own payment-completes-then-schedules flow.
+		if ( ! $subscription->has_status( 'active' ) ) {
 			return;
 		}
 
--- a/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
+++ b/plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
@@ -49,7 +49,11 @@ class WooCommerce_Subscriptions {
 		// allow_add_payment_method_without_next_payment() for the whole story.
 		// Priority 20: WCS registers its own answer at 10 and we only override a no.
 		add_filter( 'woocommerce_can_subscription_be_updated_to_new-payment-method', [ __CLASS__, 'allow_add_payment_method_without_next_payment' ], 20, 2 );
-		add_action( 'woocommerce_subscription_payment_method_updated', [ __CLASS__, 'schedule_next_payment_after_payment_method_added' ], 10, 2 );
+		// Not `woocommerce_subscription_payment_method_updated`: that fires before the
+		// card is validated or charged, so a declined card would still leave a
+		// schedule behind. This filter runs after process_payment() and carries its
+		// result, so the schedule can follow the payment instead of preceding it.
+		add_filter( 'woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode', [ __CLASS__, 'schedule_next_payment_after_payment_method_added' ], 10, 2 );
 	}
 
 	/**
@@ -384,8 +388,9 @@ class WooCommerce_Subscriptions {
 			return $can_be_updated;
 		}
 
-		// Nothing recurring to charge, so nothing to add a card for. The 'edit'
-		// context reads the unfiltered total, as WCS's own check does.
+		// Nothing recurring to charge, so nothing to add a card for. WCS reads the
+		// filtered ('view') total here; 'edit' is deliberate, so a site filtering
+		// order totals cannot open the flow on a subscription with nothing to bill.
 		if ( (float) $subscription->get_total( 'edit' ) <= 0 ) {
 			return $can_be_updated;
 		}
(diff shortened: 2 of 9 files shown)


leogermani:
I have manually tested and confirm that the menu and the option to update the payment method appears with this PR applied.

There are a few things though, maybe not all related to this PR.

If I don't create a pending renewal order in the admin, the link to update the payment info from that little warning that pops up doesn't work.

And the modal that open from the link in the kebab menu in the Subscription page is broken with things at the bottom:

image

--- 

Probably not related to this PR but if I try to visit the link the admin can get as "user payment link" in the renewal order, the reader is taken to the standard checkout experience.

---

Definitely not related to this PR. If I try to visit the link above before logging in, I get los after going through the login flow and never land on where I was going.

---

I think for what this PR is concerned we need to understand what role the pending renewal order plays here. Do we need one? What if there's none? What should be the intruction for publishers?

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -43,6 +43,13 @@ public static function init() {
 		add_action( 'woocommerce_store_api_checkout_order_processed', [ __CLASS__, 'maybe_reactivate_pending_cancel_switch' ], 40 );
 		add_action( 'woocommerce_order_status_failed', [ __CLASS__, 'maybe_revert_reactivation_on_failed_switch' ] );
 		add_action( 'woocommerce_order_status_cancelled', [ __CLASS__, 'maybe_revert_reactivation_on_failed_switch' ] );
+
+		// Adding a card to a subscription that has no next payment date: eligibility
+		// plus the follow-through that resumes billing. See
+		// allow_add_payment_method_without_next_payment() for the whole story.
+		// Priority 20: WCS registers its own answer at 10 and we only override a no.
+		add_filter( 'woocommerce_can_subscription_be_updated_to_new-payment-method', [ __CLASS__, 'allow_add_payment_method_without_next_payment' ], 20, 2 );
+		add_action( 'woocommerce_subscription_payment_method_updated', [ __CLASS__, 'schedule_next_payment_after_payment_method_added' ], 10, 2 );
```
leogermani: **Blocker:** This hook fires before the card is validated or charged, so a declined card still gets a next-payment date.

In `WC_Subscriptions_Change_Payment_Gateway::change_payment_method_via_pay_shortcode()`, `update_payment_method()` runs at line 339 — which is what fires `woocommerce_subscription_payment_method_updated` — while `validate_fields()` is at line 345, `process_payment()` at line 351, and the early `return` on a non-success result at line 364. Newspack does not replace that handler (`Change_Payment_Gateway` only adjusts the return URL, success text and payload), so the modal checkout takes the same path.

I reproduced it against WCS 8.1.0: an `active` subscription with no next payment, taken through `update_payment_method()` without the success path, ends up with `next_payment` set and a pending `woocommerce_scheduled_subscription_payment` action queued in Action Scheduler. Before the attempt it had neither.

One correction worth noting, since it changes what happens next rather than whether it happens: the subscription is not left on manual renewal — `set_payment_method()` clears that flag for a subscriptions-capable gateway — so the queued job is an *automatic* renewal attempt against a card that was never accepted, rather than a manual invoice.

**Risk if not addressed:** a reader whose card is declined and who does not retry keeps access until the scheduled date, then a renewal attempt runs against a card that was never accepted; when it fails the subscription goes on hold and they lose access they had before touching the form.

**Consider:** writing the date from `woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode` and only when the result is a success, so the schedule follows the payment rather than preceding it.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -319,6 +326,237 @@ public static function maybe_revert_reactivation_on_failed_switch( $order_id ) {
...
+		if ( ! $subscription->payment_method_supports( 'subscription_cancellation' ) ) {
+			return $can_be_updated;
+		}
+
+		// The refusal we override, limited to statuses where putting a card on file
+		// leads somewhere: awaiting a first payment, suspended, or running.
+		// pending-cancel is excluded — its next payment is already cleared by
+		// design, so it would always match, and the action would do nothing for a
+		// subscription winding down (reactivating restores WCS's own flow).
+		if ( ! $subscription->has_status( [ 'pending', 'on-hold', 'active' ] ) ) {
```
leogermani, line 415: **Suggestion:** Eligibility opens for `pending` and `on-hold`, but the follow-through completes only `active`, and the reasoning given for that split does not hold for the subscriptions this PR targets.

The docblock explains that `pending` and `on-hold` are left to WooCommerce because "they carry an unpaid order, and paying it is what activates the subscription and sets its date". A subscription created by hand in wp-admin has no parent order at all — that is the premise of the `Checkout_Data` fix in this same PR — and `pending` is its default status, so it is also the status the ticket's reproduction steps produce.

I checked this in a running environment: a `pending` subscription with no parent order offers the action, accepts a card, and afterwards has no date, no order to pay, no order note and no "Renew now" — `wcs_can_user_renew_early()` still refuses. Same for `on-hold`.

**Risk if not addressed:** the ticket's own reproduction path ends where it started — the reader adds a card, is told it worked, and still cannot pay — now behind a button that looks like the fix.

Either narrowing eligibility to the status the pair can carry through, or leaving a marker for the publisher on the `pending` path, would close the gap.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -319,6 +326,237 @@ public static function maybe_revert_reactivation_on_failed_switch( $order_id ) {
...
+	 *
+	 * @return void
+	 */
+	public static function schedule_next_payment_after_payment_method_added( $subscription, $new_payment_method ) {
+		if ( ! ( $subscription instanceof \WC_Subscription ) ) {
+			return;
+		}
+
+		// No gateway attached means there is nothing to bill against.
+		if ( empty( $new_payment_method ) ) {
```
leogermani: **Suggestion:** `empty( $new_payment_method )` infers "a card is on file" from a string, and this hook is reached from more than the reader-facing form — the admin payment-method meta box and the update-all-subscriptions path both go through `update_payment_method()`. A non-empty but non-chargeable method would pass this guard and produce a Woo-side auto-renewal schedule plus an order note attributing it to the subscriber.

`$subscription->has_payment_gateway()` is the predicate the template change in this same PR already trusts for exactly this distinction; testing that (or `! $subscription->is_manual()`) after the update would make both halves agree on what "has a card" means.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -319,6 +326,237 @@ public static function maybe_revert_reactivation_on_failed_switch( $order_id ) {
...
+			$subscription->save();
+			$subscription->add_order_note(
+				sprintf(
+					/* translators: %s: the newly scheduled next payment date, localised. */
+					__( 'Next payment scheduled for %s after the subscriber added a payment method.', 'newspack-plugin' ),
+					$subscription->get_date_to_display( 'next_payment' )
+				)
+			);
+		} catch ( \Exception $e ) {
+			Logger::error(
```
leogermani: **Suggestion:** The one failure this method plans for leaves no trace on a normal site. `Logger::error()` goes through `Logger::log()`, which returns immediately unless `NEWSPACK_LOG_LEVEL` is defined and greater than zero (`includes/class-logger.php:40`) — undefined on a stock production site.

So a failed `update_dates()` produces no date, no order note and no log, on a subscription whose reader has just been told their payment method was saved. An order note here would be durable and lands where the publisher actually looks; `class-card-expiry-warning.php:440` is the nearby precedent for giving operationally significant events an explicit header.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/subscription-header.php
```diff
@@ -32,9 +32,14 @@
...
 }
 
-// Rename 'Change payment' action.
+// Rename the 'Change payment' action, keeping the distinction WCS makes: a
+// subscription with no gateway yet needs a card added, not an existing one
+// updated. Telling a reader with nothing on file to "update" their payment
+// method is what sent them looking for a card they never had (NPPD-2170).
 if ( ! empty( $actions['change_payment_method']['name'] ) ) {
-	$actions['change_payment_method']['name'] = __( 'Update payment method', 'newspack-plugin' );
+	$actions['change_payment_method']['name'] = $subscription->has_payment_gateway()
```
leogermani, line 40: **Suggestion:** The action now reads "Add payment method", but the modal it opens is still titled "Change payment method" — the title comes from `change_payment_method_title` (`class-my-account-ui-v1.php:116`), passed as a fixed string in `src/my-account/v1/modal-checkout.js:47`. A reader with no card lands under a heading telling them to change one, which is the wording problem NPPD-2170 describes, one step further in. The `.pay` buttons already show the pattern for a per-button title.

Separately, WCS picks its own "Add payment" wording on `! ( has_payment_gateway() && wc_get_payment_gateway_by_order( $subscription )->supports( 'subscriptions' ) )`, so a subscription carrying a gateway that does not support subscriptions gets "Add payment" from WCS but "Update payment method" here.

plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
```diff
@@ -389,6 +411,13 @@ function( $subscription ) {
 			}
 		}
 
+		// A subscription with no parent order has no order to identify itself by, so
+		// name the subscription directly. Without this the modal has nothing to
+		// return the reader to once checkout finishes (NPPD-2170).
+		if ( $subscription && empty( $data['subscription_ids'] ) ) {
```
leogermani: **Suggestion:** The comment scopes this to "a subscription with no parent order", but the condition is `$subscription && empty( $data['subscription_ids'] )`, which also fires for a subscription that *has* a parent when its product type is not `subscription` or `membership` — the `wcs_get_subscriptions_for_order()` branch above is gated on that type.

A recurring donation resolves to `donation`, so those payloads previously carried only `order_id` and `handleCheckoutComplete()` redirected to `/view-order/<id>`; they would now redirect to `/view-subscription/<id>`. That may well be the better destination, but it is a behaviour change for the most common Newspack recurring product, and neither new test covers it — both fixtures are `subscription`. Scoping the fallback to the parentless case, or adding a donation-subscription test, would make it deliberate.

plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
```diff
@@ -303,18 +306,37 @@ public static function get_checkout_data( $source ) {
...
+				$order        = $source;
+				$items_source = $source;
+			}
+			$order_items = $items_source->get_items();
+			$order_item  = reset( $order_items ); // Use only the first item in the order.
+			if ( ! $order_item ) {
+				// No line item means no purchase to summarise, and the product and
+				// name lookups below would fatal on the resulting null. Return empty,
+				// the same "nothing to checkout" signal as an empty source above.
+				return $data;
```
leogermani, line 334: **Nit:** This bail returns before `apply_filters( 'newspack_modal_checkout_data', $data, $source )`, so consumers of that filter no longer run for item-less orders. In-repo that is currently harmless, and the pre-existing `empty( $source )` bail above sets the same precedent — but `newspack-manager` is a listed consumer of this contract that could not be checked locally, so it is worth a deliberate call rather than an inherited one.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
```diff
@@ -319,6 +326,237 @@ public static function maybe_revert_reactivation_on_failed_switch( $order_id ) {
...
+		// The store has turned automatic payments off entirely — a publisher
+		// configuration choice, not a per-subscription state. static:: so a test
+		// subclass can override the WCS-dependent read.
+		if ( static::store_requires_manual_renewal() ) {
+			return $can_be_updated;
+		}
+
+		// Nothing recurring to charge, so nothing to add a card for. The 'edit'
+		// context reads the unfiltered total, as WCS's own check does.
+		if ( (float) $subscription->get_total( 'edit' ) <= 0 ) {
```
leogermani, line 394: **Nit:** The comment says the `'edit'` context matches what WCS's own check does, but `can_subscription_be_updated_to_new_payment_method()` calls `get_total()` with no argument — the `'view'` context, which passes through `woocommerce_order_get_total`. They agree in practice and diverge only near the zero boundary on a site that filters order totals. Since these comments are the map for keeping this mirrored against WCS, worth either matching the call or noting the divergence is deliberate.

leogermani (dismissed):
Thanks for this — the eligibility half is careful work, and the documentation of what it mirrors in WCS made it much easier to check.

I ran it in an isolated environment with WooCommerce Subscriptions 8.1.0 and a stub subscriptions-capable gateway, and read it against the WCS source. The refusals you preserve all hold: zero-total, terminal statuses and a store configured for manual renewals each still refuse, and eligibility is granted only with this change applied — I checked by removing the filter and watching it flip back. The date arithmetic is safe too: a subscription started thirteen months ago still gets a future date, and one near its end date gets none.

One thing needs changing before merge. `woocommerce_subscription_payment_method_updated` fires before WCS validates the card or takes payment, so a declined card still leaves a next-payment date and a real scheduled renewal behind, with nothing to roll it back. I reproduced this — after the declined-card path the subscription carries a pending renewal job it did not have before.

Alongside that, `pending` is the default status for a subscription created in wp-admin, which is the exact case in NPPD-2170, and the action now offered there completes without changing anything. Worth deciding whether that path should be opened at all in this PR.

The rest are smaller observations, inline.

dkoo:
@leogermani thanks for the review and manual QA notes! Re: this question:

> I think for what this PR is concerned we need to understand what role the pending renewal order plays here. Do we need one? What if there's none? What should be the intruction for publishers?

When a subscription owner adds a payment method to any subscription, this doesn't trigger a payment or future auto-renewal by itself. The publisher will need to "create a pending renewal order" for the subscription to compel the subscription owner to pay for the order. The subscription gets a new `next_payment` date for future auto-renewal after this first successful payment. 

The difference is, for subscriptions with an `active` status, the subscription owner can do this by themselves in My Account by clicking "Renew now" in the subscription actions kebab menu. For subscriptions with a `pending` status, this can't be done in My Account—the pending renewal order can only be created by an admin. However, I think the baseline instruction should be for publishers to create a pending renewal order in most cases because even active subscription owners have no incentive to click "Renew now" themselves—they'll have free/unrestricted access until that first payment sets the auto-renewal schedule.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
dkoo: Confirmed and fixed in e904edd5 — thank you, this was a real one and I had the ordering wrong.

I read it back in the WCS source: `update_payment_method()` at :349, `validate_fields()` at :353, `process_payment()` at :359, and the non-success `return` at :371. So the old hook was firing two steps before we knew the card was good.

Moved to the seam you suggested, `woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode`, which WCS applies after `process_payment()` and before it bails, and gated on `'success' === $result['result']`. Verified against an active subscription with no next payment:

```
before:                next_payment=0
after DECLINED card:   next_payment=0
after SUCCESSFUL card: next_payment=2026-09-12
```

`test_does_not_schedule_when_the_payment_failed()` pins it, and a second test asserts the filter returns the result unchanged either way. Your correction about `set_payment_method()` clearing manual renewal is what made the severity clear — an automatic renewal against a card that was never accepted is worse than no schedule at all.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
dkoo: Fixed in e904edd5. Now `! $subscription->has_payment_gateway() || $subscription->is_manual()`, so both halves of the PR use the same predicate for "has a card" and the admin and bulk-update paths are covered.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
dkoo: Fixed in e904edd5, and you were right about the reachability: `Logger::log()` returns immediately unless `NEWSPACK_LOG_LEVEL` is defined and non-zero (`class-logger.php:40`), which it is not on a stock site — so the one failure this method plans for was leaving no trace anywhere. Replaced with an order note naming the reason, which is durable and lands where the publisher looks.

plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
dkoo: Fixed in e904edd5 — and this is the second time this exact finding has been raised, because I reported it fixed in an earlier round and it was not. I separated the order and subscription identities but never scoped the fallback, so the condition still read `$subscription && empty( ... )`.

It is now `$subscription && ! $order && empty( $data['subscription_ids'] )`, and `test_donation_subscription_with_a_parent_keeps_order_identity()` covers the case both of us described — a donation-typed subscription with a parent keeps `order_id` and gets no `subscription_ids`. That test is the thing that should have existed the first time.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
dkoo: Fixed in e904edd5. You are right that WCS calls `get_total()` with no argument. Kept `'edit'` and reworded the comment to say the divergence is deliberate: a site filtering order totals should not be able to open the flow on a subscription with nothing to bill.

plugins/newspack-blocks/includes/modal-checkout/class-checkout-data.php
dkoo: Checked rather than inherited: `newspack-manager` has no references to `newspack_modal_checkout_data` or `Checkout_Data` at all, so nothing downstream loses a filter pass here. Leaving the bail as is, now on evidence rather than assumption. Thanks for flagging it as a call to make deliberately.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/subscription-header.php
dkoo: Both parts confirmed, and I have left them for a follow-up rather than folding them in.

The modal title comes from `change_payment_method_title` in `class-my-account-ui-v1.php:116`, passed as a fixed string from `src/my-account/v1/modal-checkout.js:47`, so fixing it means plumbing a per-button title through the modal-checkout data the way the `.pay` buttons already do — a JS/PHP change with its own testing, and not one I want to bundle into a fix that is already touching the payment path.

On the WCS parity point: agreed that `has_payment_gateway()` alone diverges from WCS, which also requires `->supports( 'subscriptions' )`. I kept `has_payment_gateway()` because it matches the submit button the reader actually lands on, but the divergence is real for a gateway that does not support subscriptions. I will capture both in the follow-up issue rather than leave it implied here.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/class-woocommerce-subscriptions.php
dkoo: This is the one I have not acted on yet, deliberately — dkoo has posted the mechanics above and we would like your read before changing behaviour.

Your reproduction matches mine: `pending` with no parent order offers the action, accepts a card, and afterwards has no date, no order and no "Renew now". Where I came out after digging into WCS: auto-creating the order from this flow looks wrong, because `WCS_Admin_Meta_Boxes::create_pending_renewal_action_request()` calls `$subscription->update_status( 'on-hold' )` before creating it. A reader who clicks "Add payment method" would lose access on the spot, which is exactly what support was protecting when they set the subscription on this ticket to active. It also sends a renewal invoice email and routes payment through `WCS_Cart_Renewal`, whose `all_items_required` failure path empties the cart and redirects to `/cart/` — the generic-error modal screen this ticket opened with.

What I would rather do is make the button honest: offer it for `active`, `on-hold`, and `pending` that has a payable order, and withhold it for `pending` with no order — so we never show a control that cannot complete. That leaves "create the pending renewal order" as the documented publisher step, which is dkoo's baseline instruction anyway.

Holding until you have weighed in, since it changes what we tell publishers.

dkoo:
Thanks — this was a genuinely useful review. The blocker was real, and one of your suggestions caught a finding I had reported as fixed in an earlier round and had not actually fixed. Pushed as e904edd5; replies are on each thread.

**Fixed**

- **The blocker.** Scheduling moved off `woocommerce_subscription_payment_method_updated` to the success-gated `woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode`. A declined card now leaves nothing behind — verified end to end: `next_payment` stays `0` after a declined result and is written only after a successful one.
- **`empty( $new_payment_method )`** → `has_payment_gateway() && ! is_manual()`, so both halves of the PR agree on what "has a card" means.
- **The silent `catch`.** You were right that `Logger::error()` is invisible on a stock site; it is an order note now.
- **The `subscription_ids` fallback**, properly scoped to the parentless case this time, with the donation-with-parent test that should have existed the first time.
- **The `'edit'` context comment**, reworded to say the divergence from WCS is deliberate.

**The broken modal from your QA notes** — fixed here, though it turned out not to be ours. I reproduced it on a subscription that was already eligible *without* this PR's filter, so the PR exposes it to a new population rather than causing it. Two things were being emitted into the checkout iframe: the content-gifting modal, which has no guards at all, and the modal-checkout container nesting inside its own iframe. Both are now guarded on `is_modal_checkout()`, following the eight existing precedents. The change-payment form still renders, and the parent My Account page keeps its gifting modal. I deliberately did not guard the reader-activation auth modals, since modal checkout uses those for registration.

**The `newspack_modal_checkout_data` nit** — checked rather than assumed: `newspack-manager` has no consumers of that filter or of `Checkout_Data`, so the early bail costs nothing downstream.

**Still open, and waiting on you**

The `pending` question is the one thing I have not acted on. @dkoo has [posted the mechanics](https://github.com/Automattic/newspack-workspace/pull/840#issuecomment-5272843069). We explored one idea that we ultimately didn't implement: auto-creating the order from this flow. This looks wrong because WCS's own action sets the subscription `on-hold` first — a reader clicking "Add payment method" would lose access immediately. Your call before we change anything else on this.

**Deferred to follow-ups**, both from your review: the modal title still reading "Change payment method" (needs a per-button title plumbed through the modal-checkout data), and the `has_payment_gateway()` vs WCS's two-part label predicate.

Verification on this round: 19 add-payment tests, 6 modal-checkout-data tests, full suites at 2913 and 151, PHPCS clean.

Authored by Claude.

leogermani (approved):
Approving it.

My only concern here is that a manually created subscription ends up with a broken state. We need to make sure to communicate this well somehow if there are no ways to put guardrails on that.

Also worth mentioned that I've tested and could not reproduce claims that a declined card would give the reader access to anything

dkoo:
> My only concern here is that a manually created subscription ends up with a broken state. We need to make sure to communicate this well somehow if there are no ways to put guardrails on that.

Yes, good call. I'll note that this PR does make manually-created subscriptions _less_ broken, since currently there's simply no way to attach a payment method/auto-renewal for them.

PR #842, opened 2026-08-11
feat(themes): offer Header and Body fonts in the editor font picker

thomasguillot:
### All Submissions:

The block editor's Font control never appears on the classic Newspack theme, so publishers cannot set a paragraph (or any other block) to the site's headline font without custom CSS. The control only shows when a theme registers font presets, and the classic theme has nowhere to do that without a `theme.json`.

This registers two presets, **Header** and **Body**, mirroring the Customizer's Header Font and Body Font settings. The Font control now offers them on every block that supports typography, in the base theme and all five style variations. The presets reference the theme's existing font variables rather than fixed font stacks, so changing a font in the Customizer updates any content using them, with no re-save.

Before and after: the Typography panel without and with the Font control offering Header and Body

1. On a site running the Newspack classic theme, edit a post and select a paragraph.
2. Open the Typography panel's options menu and enable Font. A Font dropdown appears with Default, Header, and Body.
3. Choose Header. The paragraph switches to the header font in the editor.
4. Save and view the post. The paragraph renders in the header font on the front end.
5. In the Customizer, set a different Header Font. The paragraph follows it without editing the post.

### Other information:

* [x] Have you written new tests for your changes, as applicable? New tests n/a: the theme has no PHPUnit suite. Verified live on a dev site.
* [x] Have you successfully run tests with your changes locally? PHPCS clean; no test suite exists for the theme.

Core hides the Font control behind the Typography panel's options menu by default, on block themes too. Step 2 covers it here; worth a line in help docs when this ships.

### DIFF
--- a/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
+++ b/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
@@ -33,11 +33,15 @@ if ( ! function_exists( 'newspack_font_stack' ) ) {
 	 * @return string
 	 */
 	function newspack_font_stack( $primary, $fallback_id ) {
-		$stacks = newspack_get_font_stacks();
-		$fonts  = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : [];
+		$stacks   = newspack_get_font_stacks();
+		$fonts    = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : [];
+		$generics = [ 'serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui' ];
 		array_unshift( $fonts, $primary );
 		foreach ( $fonts as &$font ) {
-			$font = '"' . $font . '"';
+			$font = str_replace( [ "\n", "\r", "\t", '<', '>' ], '', stripslashes( $font ) );
+			if ( ! in_array( strtolower( $font ), $generics, true ) ) {
+				$font = '"' . str_replace( [ '\\', '"' ], [ '\\\\', '\\"' ], $font ) . '"';
+			}
 		}
 		return implode( ',', $fonts );
 	}
--- a/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
+++ b/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
@@ -34,11 +34,11 @@ if ( ! function_exists( 'newspack_font_stack' ) ) {
 	 */
 	function newspack_font_stack( $primary, $fallback_id ) {
 		$stacks   = newspack_get_font_stacks();
-		$fonts    = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : [];
+		$fonts    = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : $stacks['serif']['fonts'];
 		$generics = [ 'serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui' ];
 		array_unshift( $fonts, $primary );
 		foreach ( $fonts as &$font ) {
-			$font = str_replace( [ "\n", "\r", "\t", '<', '>' ], '', stripslashes( $font ) );
+			$font = preg_replace( '/[\x00-\x1F\x7F<>]/', '', stripslashes( $font ) );
 			if ( ! in_array( strtolower( $font ), $generics, true ) ) {
 				$font = '"' . str_replace( [ '\\', '"' ], [ '\\\\', '\\"' ], $font ) . '"';
 			}
--- a/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
+++ b/plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
@@ -43,6 +43,7 @@ if ( ! function_exists( 'newspack_font_stack' ) ) {
 				$font = '"' . str_replace( [ '\\', '"' ], [ '\\\\', '\\"' ], $font ) . '"';
 			}
 		}
+		unset( $font );
 		return implode( ',', $fonts );
 	}
 }
--- a/themes/newspack-theme/newspack-theme/inc/typography.php
+++ b/themes/newspack-theme/newspack-theme/inc/typography.php
@@ -169,10 +169,11 @@ function newspack_get_font_stacks_as_select_choices() {
 /**
  * Prepare a font-family definition with a primary font and fallbacks.
  *
- * Font names are stripped of CSS string and function delimiters — these values
- * flow into inline styles and theme.json preset values, where the theme origin
- * is never passed through core's insecure-property filtering. Generic family
- * keywords stay unquoted so a stack always ends in a real generic.
+ * Values flow into inline styles and theme.json preset values, and core never
+ * passes the theme origin through its insecure-property filtering, so names
+ * are emitted as quoted CSS strings with string delimiters escaped and markup
+ * and control characters removed. Generic family keywords stay unquoted so a
+ * stack always ends in a real generic.
  */
 function newspack_font_stack( $primary_font, $fallback_id ) {
 	$stacks   = newspack_get_font_stacks();
@@ -180,9 +181,9 @@ function newspack_font_stack( $primary_font, $fallback_id ) {
 	$generics = array( 'serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui' );
 	array_unshift( $fonts, $primary_font );
 	foreach ( $fonts as &$font ) {
-		$font = str_replace( array( '"', "'", ';', '{', '}', '(', ')', '\\' ), '', $font );
+		$font = str_replace( array( "\n", "\r", "\t", '<', '>' ), '', stripslashes( $font ) );
 		if ( ! in_array( strtolower( $font ), $generics, true ) ) {
-			$font = '"' . $font . '"';
+			$font = '"' . str_replace( array( '\\', '"' ), array( '\\\\', '\\"' ), $font ) . '"';
 		}
 	}
 	return implode( ',', $fonts );
@@ -222,9 +223,17 @@ function newspack_font_family_stacks() {
 			'heading' => '"Fira Sans Condensed","Helvetica",sans-serif',
 		),
 	);
-	if ( isset( $variations[ get_stylesheet() ] ) ) {
-		$defaults = array_merge( $defaults, $variations[ get_stylesheet() ] );
+	$slug = isset( $variations[ get_stylesheet() ] ) ? get_stylesheet() : get_template();
+	if ( isset( $variations[ $slug ] ) ) {
+		$defaults = array_merge( $defaults, $variations[ $slug ] );
 	}
+
+	/**
+	 * Filter the default Header and Body font stacks.
+	 *
+	 * @param array $defaults Stacks keyed heading/body.
+	 */
+	$defaults = apply_filters( 'newspack_font_family_default_stacks', $defaults );
 
 	$mods   = array(
 		'heading' => array( 'font_header', 'font_header_stack' ),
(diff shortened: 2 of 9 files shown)


themes/newspack-theme/newspack-theme/inc/typography.php
```diff
@@ -170,3 +170,37 @@ function newspack_font_stack( $primary_font, $fallback_id ) {
...
 }
+
+/**
+ * Register the Customizer fonts as editor font family presets.
+ *
+ * The values reference the theme's own CSS variables rather than resolved
+ * stacks, so saved content tracks Customizer font changes with no re-save.
+ *
+ * @param WP_Theme_JSON_Data|WP_Theme_JSON_Data_Gutenberg $theme_json Theme JSON data.
+ * @return WP_Theme_JSON_Data
```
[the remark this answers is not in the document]
thomasguillot: Widened both annotations to the union in dddb14ab8. The class only exists with Gutenberg active, so the theme's phpstan stubs now declare it, keeping the union analyzable.

thomasguillot:
Release risk: LOW. No open findings after five rounds: no blockers at any point, all suggestions and nits fixed and re-verified live, cross-unit impact none, and the one prior reviewer comment on the PR is addressed and resolved. Two items are deliberately deferred outside this PR: unifying the default stacks with the newsletters email font constants (changes email rendering), and the newsletters test suite's order-dependent guards (pre-existing structure).

thomasguillot:
Final pass after the review: corrected the preset docblock, which cited lite pages as a context the var() fallback covers. Lite pages strip preset classes from content, so the fallback's value is feeds and other contexts that print global styles without the theme stylesheet (d3d59ab00). No code behavior changed.

plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
```diff
@@ -33,11 +33,15 @@ function newspack_get_font_stacks() {
 	 * @return string
 	 */
 	function newspack_font_stack( $primary, $fallback_id ) {
```
dkoo, line 35: **Suggestion:** Theme↔fixture font-stack duplication has no drift guard — This fixture mirrors the theme's `newspack_font_stack()`, and the newsletters `Test_Fonts` assertions call `newspack_font_stack()` — but in the test process both the expected and actual sides resolve to this fixture (the real theme isn't loaded), so the assertion stays green no matter how far the fixture drifts from the theme. If the theme's escaping evolves again, newsletter rendering could diverge from the theme with no failing test. The fixture docblock already points back to the theme; a matching pointer in the theme function, or a lightweight contract test, would make the coupling visible.

themes/newspack-theme/newspack-theme/inc/typography.php
```diff
@@ -160,13 +168,132 @@ function newspack_get_font_stacks_as_select_choices() {
...
 }
+
+/**
+ * The current Header and Body font stacks: the Customizer font when set,
+ * otherwise the active theme's default, mirroring
+ * sass/variables-site/_fonts.scss and each style variation's overrides.
+ *
+ * @return array Stacks keyed heading/body.
+ */
+function newspack_font_family_stacks() {
```
dkoo, line 210: **Suggestion:** Per-variation font stacks duplicate the SCSS, untested — `newspack_font_family_stacks()` hardcodes each style variation's Header/Body stacks, which today match `_variables-style.scss` and `_fonts.scss` exactly (verified). It is a second source of truth, though: a designer changing a variation's `--newspack-theme-font-*` in SCSS won't know to update this map, and no test ties them together. The docblock is upfront about the mirroring, so this is acceptable — a back-pointer comment in each variation's SCSS, and/or a follow-up to add coverage (the theme has no PHPUnit harness today), would reduce the maintenance trap.

themes/newspack-theme/newspack-theme/inc/typography.php
```diff
@@ -160,13 +168,132 @@ function newspack_get_font_stacks_as_select_choices() {
...
+ * the Gutenberg plugin active the resolver passes WP_Theme_JSON_Data_Gutenberg,
+ * a sibling class rather than a subclass, so the parameter stays untyped.
+ *
+ * @param WP_Theme_JSON_Data|WP_Theme_JSON_Data_Gutenberg $theme_json Theme JSON data.
+ * @return WP_Theme_JSON_Data|WP_Theme_JSON_Data_Gutenberg
+ */
+function newspack_font_family_presets( $theme_json ) {
+	$data     = $theme_json->get_data();
+	$families = $data['settings']['typography']['fontFamilies'] ?? array();
+	if ( isset( $families['theme'] ) ) {
```
dkoo, line 281: **Suggestion:** Dead `isset( $families['theme'] )` branch — On the `wp_theme_json_data_theme` filter, `$theme_json->get_data()` returns the raw theme.json-shaped array, where `settings.typography.fontFamilies` is a flat, indexed list — the origin keys (`theme`/`default`/`custom`) live only inside the merged `WP_Theme_JSON`, not this data object. So this branch never fires here, and if it somehow did, `$families = $families['theme']` would drop any sibling `default`/`custom` families. It's harmless as written; consider dropping it, or adding a one-line comment that it's purely defensive so a future reader doesn't puzzle over it.

themes/newspack-theme/newspack-theme/inc/typography.php
```diff
@@ -160,13 +168,132 @@ function newspack_get_font_stacks_as_select_choices() {
...
+ * @return string Comma-joined, pre-escaped CSS font-family list.
  */
 function newspack_font_stack( $primary_font, $fallback_id ) {
-	$stacks = newspack_get_font_stacks();
-	$fonts  = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : array();
+	$stacks   = newspack_get_font_stacks();
+	$fonts    = isset( $stacks[ $fallback_id ] ) ? $stacks[ $fallback_id ]['fonts'] : $stacks['serif']['fonts'];
+	$generics = array( 'serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui' );
 	array_unshift( $fonts, $primary_font );
 	foreach ( $fonts as &$font ) {
```
dkoo, line 193: **Nit:** `foreach ( $fonts as &$font )` leaves a dangling reference — The loop binds `$font` by reference with no `unset( $font )` afterward. It's harmless today because `$fonts` is `implode()`d immediately and not reused, but it's the classic reference-loop footgun — a later edit that reuses `$fonts` would corrupt the last element. The newsletters fixture has the same shape. A trailing `unset( $font )` makes it safe.

themes/newspack-theme/newspack-theme/inc/typography.php
```diff
@@ -160,13 +168,132 @@ function newspack_get_font_stacks_as_select_choices() {
...
+	return $stacks;
+}
+
+/**
+ * Register the Customizer fonts as editor font family presets.
+ *
+ * The values reference the theme's own CSS variables so saved content tracks
+ * Customizer font changes with no re-save; the resolved stack rides along as
+ * the var() fallback for feeds and other contexts that print global styles
+ * without the theme stylesheet. Existing theme-origin presets are appended
```
dkoo: **Nit:** Docblock sentence wraps awkwardly — The paragraph breaks mid-sentence — "...appended to, not replaced. With / the Gutenberg plugin active the resolver passes..." — leaving "With" stranded at a line end and splicing two thoughts together. Purely cosmetic; a sentence reflow would read more smoothly.

dkoo (approved):
Release risk: LOW — Approved with non-blocking comments. Verified end-to-end against an isolated env, no blockers, and a backward-compatible change on an admin-only surface. 

Browser testing confirmed the whole flow: the editor's Font control offers Default / Header / Body, choosing Header renders in the header font in both the editor and the front end, and a Customizer font change follows through to already-saved content with no re-save. The escaping consolidation reads as sound — markup and control characters are stripped and quotes escaped, so a font name can't break out of the CSS string or the `<style>` block, which makes the removed `wp_kses()` calls redundant rather than missed. The `wp_theme_json_data_theme` filter rebuilds from source on each fire, so presets don't accumulate, and each per-variation stack matches the SCSS exactly. 

Two things are worth a glance before merge, both maintainability rather than merge blockers: the `isset($families['theme'])` branch never fires for this filter's data shape, and the theme-to-newsletters font-stack duplication has no test guarding against drift.

### Verification walkthrough

Browser-tested end-to-end on the classic Newspack theme in an isolated environment.

**1. The Font control appears** once the theme registers presets (Typography ▸ options ▸ Font):

Font control enabled in the Typography panel

**2. The dropdown offers Default / Header / Body:**

Font dropdown showing Default, Header, Body

**3. Choosing Header applies it in the editor** — the paragraph gets `has-newspack-header-font-family` and computes to the header stack in the canvas:

Header font applied in the editor

**4. The saved post renders in the header font on the front end** — global styles emit `--wp--preset--font-family--newspack-header: var(--newspack-theme-font-heading, …)`:

Front end rendering the header font

**5. Changing the Customizer Header Font follows the saved post with no re-save** — set to Papyrus here, the paragraph re-renders as `Papyrus, Georgia, serif` after a plain reload:

Customizer font change tracked on the front end

themes/newspack-theme/newspack-theme/inc/typography.php
thomasguillot: Added `unset( $font )` after the loop, in both the theme and the newsletters fixture (ad3b84dbd).

themes/newspack-theme/newspack-theme/inc/typography.php
thomasguillot: Reflowed in ad3b84dbd.

themes/newspack-theme/newspack-theme/inc/typography.php
thomasguillot: This one does fire. `WP_Theme_JSON`'s constructor re-keys presets by origin ([`class-wp-theme-json.php:854-870`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-theme-json.php) in 7.0.2), and `WP_Theme_JSON_Data::get_data()` returns that raw internal array, so the flat indexed list is the shape going *in*, not the shape coming back out.

Checked against WP 7.0.2, constructing the data object the same way `get_theme_data()` does:

```
keys returned by get_data(): theme
isset($families['theme']): TRUE
without the branch, nesting depth: DOUBLE-NESTED (corrupt)
without the branch, slugs core would see: ["newspack-header"]
```

So dropping it would append to the origin-keyed array, core would wrap the whole thing again on the way back through `update_with()`, and any pre-existing preset would be lost.

You're right that nothing exercises it today: the classic theme has no `theme.json`, so the list arrives empty and the branch sits idle. It starts mattering the moment anything else puts a theme-origin family on this filter, whether that's another plugin at an earlier priority or a `theme.json` later. Kept it, with a comment in ad3b84dbd covering the origin keying, since the branch doesn't explain itself.

plugins/newspack-newsletters/tests/email-renderers/fixtures/theme-font-functions.php
thomasguillot: Added the pointer from the theme side in ad3b84dbd: `newspack_font_stack()` now names this fixture and says escaping changes have to be carried over, so the coupling is visible from whichever end someone edits.

Left the contract test out. The theme has no PHPUnit harness, and a newsletters-side test that reads the theme file would only work in the monorepo layout, skipping wherever the plugin runs standalone. That reads as coverage without being it.

themes/newspack-theme/newspack-theme/inc/typography.php
thomasguillot: Made it visible from the SCSS side in ad3b84dbd. `_fonts.scss` and each variation's `_variables-style.scss` carry a one-line pointer to `newspack_font_family_stacks()`, so a designer changing `--newspack-theme-font-heading` sees the PHP map named at the point of edit.

Coverage still needs a PHPUnit harness in the theme, which is a bigger piece of work than this change should carry.

thomasguillot:
All five threads are addressed in ad3b84dbd, and both items flagged in the review are covered.

The `isset( $families['theme'] )` branch turns out to fire: `WP_Theme_JSON` re-keys presets by origin in its constructor, and `get_data()` returns that internal array, so a non-empty list arrives nested under `theme`. It's idle today only because the classic theme ships no `theme.json`. Kept it with a comment, and the check against WP 7.0.2 is on the thread.

The font-stack duplication now has pointers at both ends: `newspack_font_stack()` names the newsletters fixture, and `_fonts.scss` plus each variation's `_variables-style.scss` name `newspack_font_family_stacks()`. Neither gets a test, since the theme has no PHPUnit harness, so the pointers are what a maintainer has to go on.

Thanks for the reference-loop catch, that one would have bitten quietly.

PR #844, opened 2026-08-11
fix(audience): lead the pricing rules screen with the rules

thomasguillot:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? PHP unchanged; JS, TS and SCSS lint clean.

Rebuilds the Pricing Rules screen around the rules themselves. The list is what the screen is named for, so it now comes first, and the catalogue impact reads as a summary underneath rather than a wall of reference data above.

A publisher with no rules used to land on an empty table with a toolbar over it. They now get an onboarding screen with the Add Rule action moved into it, matching what Access Control does for Institutions.

Pricing Rules screen: the rules table leads, with catalogue impact in a card beneath it

* **The rules table comes first.** The catalogue impact sits below it in a card, carrying the affected-product count as a headline number instead of a sentence, the same treatment the rule editor already uses for its Impact Preview.
* **The product table moved into a modal**, opened from a button and fetched only when asked for. It loads the full 50-row sample rather than the previous 10, so a store with 50 or fewer affected products now sees everything instead of a sample.
* **The screen loads as one unit.** A page-level spinner holds until both reads settle, so the header, the numbers and the table arrive together rather than in pieces.
* **An error no longer looks like an empty catalogue.** A failed read keeps the table shell and its notice, where before it rendered the same bare "No results" as a site with no rules at all.

Base is `fix/pricing-schedule-prices` (#826), not `main`.

Stacked on: #757 → #798 → #825 → #826 → this.

1. Open **Audience → Pricing Rules** on a site with WooCommerce and the dynamic pricing engine active. The rules table renders first, with a card beneath it showing the affected-product count.
2. Select **View Affected Products**. A modal opens with a spinner, then the product table. Close and reopen it. No second request is made, because a landed sample is cached.
3. Trash every pricing rule and reload. The onboarding screen replaces the table, the Add Rule button moves out of the header into it, and the breadcrumb shows no count. Untrash them afterwards.
4. Search the rules list for something that matches nothing. The DataViews "no results" treatment shows, not the onboarding screen. This is the distinction between an empty list and an empty search.
5. Trash a rule from the row actions. The table refreshes in place. The screen does not blank, because only the first load holds the page spinner.

### Other information:

* [x] Have you written new tests for your changes, as applicable? 857 tests across 89 suites, up from 824 at the branch point.
* [x] Have you successfully run tests with your changes locally? Tests, `lint:js`, `lint:scss` and the production build all pass. `tsc` reports 221 errors against 220 at the branch point; the extra one is the `start` and `end` grid-placement props on a `VStack` in `onboarding.tsx`, the layout convention six other views in this wizard already use.

---

The engine side has since landed, so both fields this screen reads are supplied. `sample_limit` now drives the truncation note in both previews, with the local constant as the fallback for a payload that omits it, and `audience` reaches the catalogue route, so the second headline number renders when the subscriptions layer is present.


thomasguillot:
## Code review

Release risk: LOW — no blockers, no shared-contract paths touched, and the changed paths carry tests. Nothing here should hold the merge. Before it goes out, three things are worth a close look, all of them in how the screen behaves once the engine is slower or less cooperative than the development site. Reopening the affected-products modal before its first request lands fires the screen's most expensive request a second time, and a late failure from the first request can clear a table that has already loaded. Trashing a rule refreshes the table but not the headline count or the cached product sample, so part of the screen keeps describing a rule that is gone. And because the page now waits on the catalogue read as well as the rules read, a catalogue route that hangs rather than fails leaves the whole screen a spinner even though the rules have arrived. The remaining items are smaller: the modal trusts a payload the sibling preview already guards, the loading states are silent to screen readers where the error state is not, and two different number formatters now render counts side by side on the same screen.

### Findings

**1. Reopening the modal mid-fetch double-requests, and a late rejection wipes a landed sample** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:43`

The `if ( detail ) { return; }` guard in `open` only knows about a settled fetch, not an in-flight one, and neither promise is tagged as latest-wins. Opening the modal starts request A; closing before it lands leaves A in flight with `detail` still null; opening again therefore fires request B. Because `.then( setDetail )` and `.catch( () => setHasError( true ) )` are both unconditional, if B resolves and A then rejects, the render falls through to the `role="alert"` error paragraph while `! hasError && detail` goes false — the table that had just loaded is replaced by "Could not load the affected products." This reproduces in a test driving exactly that order. Even without a failure the duplicate costs a second full `?limit=50` catalogue re-price, which the PR measured at roughly 1.7s, so the close-during-load window is wide enough to hit in normal use. A request counter that only the newest response may write closes both halves; the existing "keeps the sample across a close and reopen" test only covers the case where A has already resolved.

**Consider:** Track the in-flight request and ignore any response that is not the latest, so reopening the panel neither issues a duplicate of the most expensive request on this screen nor lets a stale failure clear a sample that has already loaded.

**2. Trashing a rule leaves the catalogue headline and the cached sample stale** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:144`

The catalogue read sits in a `useEffect` with an empty dependency array, so it runs once per mount, while `trashRule` refetches only the rules via `fetchData()`. Trashing a rule removes its products from the engine's active union — `catalog_affected()` walks `$repo->active()` — so "Products affected: N" keeps the pre-delete figure, and the `total_matching === 0` branch that swaps the button for "No active pricing rules are affecting products yet." can no longer be reached by deleting the last effective rule. The `detail` cached inside `CatalogImpact` goes stale the same way, so reopening the modal after a delete replays the trashed rule's products. Edit and create do not have this problem because `/edit/:id` is a separate route and the `Switch` remounts the list on return; trash is the one mutation that happens without a route change, which is why this is easy to miss in manual testing. Extracting the catalogue read into a `useCallback` and calling it alongside `fetchData()` in the trash handler would close it, and a `key` on the stats identity would reset the modal's cache for free.

**Consider:** Refetch the catalogue figures and drop the cached product sample in the same place the rules are refetched after a trash, so every number on the screen describes the same rule set.

**3. The whole screen is gated on the secondary read, with no failure path for a hang** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:79`

`isReady = hasResolved && statsResolved` withholds the rules table — the subject of the screen — until the catalogue read settles. A rejection is handled, since `.finally` flips `statsResolved` either way, but `apiFetch` sets no timeout, so a catalogue route that hangs rather than fails leaves the page a bare spinner indefinitely even though the rules payload has landed and the onboarding decision is already knowable. The `?limit=1` split does not bound this: `catalog_impact()` calls `catalog_affected( $union_ceiling )` before the sample loop, and the engine documents that walk as up to `num_rules × 1000` product loads, with `$limit` only bounding the pricing loop below it. So on a store with many rules and a large catalogue, time-to-first-rule equals the union walk regardless of the limit, and the 290ms measured on the development site reflects that site's catalogue size rather than a ceiling. Gating only `CatalogImpact` on `statsResolved` and letting `hasResolved` alone release the table and the header action keeps the "one unit" reading for the common case while removing the single point of failure.

**Consider:** Render the rules table as soon as the rules request settles and let the catalogue card carry its own spinner, and keep the Add Rule action available throughout — so a slow secondary read never makes the screen a dead end for work that does not depend on it.

**4. The modal trusts a payload the sibling preview already guards** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:93`

`detail` is handed straight to `ImpactTable` as `baseline={ detail.sample }` with only `segment_groups ?? []` defended. `ImpactTable` immediately runs `indexById( baseline )`, a `for…of` over the value, so an omitted `sample` is a render-time `TypeError` rather than an empty table, and `formatPrice` dereferences `currency.symbol` unconditionally. The modal also ignores `detail.supported` entirely, so an engine that answers the `?limit=50` read with `supported: false` — WooCommerce Subscriptions deactivated between the two calls, or the engine itself deactivated mid-session — renders as if it were a real sample. The engine lives in a plugin that is not in this repository and versions independently, and `rule-preview.tsx` already treats the same response type defensively with `! data?.supported` and `! data.sample?.length` before rendering the identical component. Mirroring that guard costs two lines and can reuse the error paragraph already in this file.

**Risk if not addressed:** An engine build that returns its minimal payload from the catalogue route — the shape it already returns from the per-rule route, carrying no sample and no currency — makes the table iterate over an undefined value, and because the wizard page is mounted with no error boundary above it, the throw unmounts the entire Pricing Rules screen to a blank page rather than breaking one panel.

**Consider:** Guard the modal payload the way the rule preview already guards the same response type — check the supported flag and treat a missing sample as the empty state — so an unexpected payload degrades to a message instead of taking the page down with it, since nothing above this component catches a render error.

**5. The limit-invariance claim covers two fields, but a third is rendered from the same response** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:123`

The comment justifies `?limit=1` on the grounds that `total_matching` and `count_limited` do not vary with the limit, which is the right check for the two fields it names. But this same response is passed into `CatalogImpact`, which forwards `stats.audience` to `ImpactStats`, and that component renders `audience.total`, `audience.caught` and `audience.protected` as headline numbers. Nothing establishes that the engine computes those independently of the sample size — if the subscriptions layer derives them by walking the priced sample, a `?limit=1` read would report "1 subscriber in scope" as a fact. `types.d.ts` marks `audience` as absent on the catalogue route today, which is exactly why it is worth pinning now: when the engine side ships it, the failure would surface as plausible-looking wrong numbers rather than an error. Either extend the comment to state the invariance being relied on for `audience`, or have the screen drop `audience` from the `?limit=1` payload and surface it only from the `?limit=50` read.

**6. Loading is announced on failure but not on success** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:72`

`Spinner` from `@wordpress/components` renders `role="presentation" focusable="false"`, so it contributes nothing to the accessibility tree. Opening the modal announces the dialog name "Affected Products" and then presents an empty region for the duration of the fetch, and when the table arrives nothing announces it — the `role="alert"` on the error branch covers the failure path only, making success the one outcome that is silent. The same applies to the new page-level gate in `list.tsx`, which now withholds the entire screen rather than just the impact panel. WCAG 2.2 AA 4.1.3 wants that status change conveyed. Wrapping each spinner in a `role="status"` container with visually hidden text would cover both without changing the visual design. A bare `<Spinner />` is the existing habit across the neighbouring audience views, so this is consistency rather than regression — but reaching for `role="alert"` on the error path makes the asymmetry look accidental.

**7. Query strings built by hand where `addQueryArgs` is the local convention** &nbsp;`suggestion`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:127`

Both new reads concatenate the query string directly — `` `${ IMPACT_PREVIEW_API_PATH }?limit=1` `` here and `` `${ API_PATH }?limit=${ IMPACT_SAMPLE_LIMIT }` `` in `catalog-impact.tsx`. `scope-targets.tsx` in this same directory already uses `addQueryArgs( path, { search, per_page: 20 } )`, as do the other data-fetching views under `src/wizards/`. The values are integer constants so there is no encoding hazard today, but a hand-built path breaks the moment a second parameter arrives or the constant grows a query string of its own. Worth noting that changing this means updating the literal path assertions in `list.test.jsx` and `catalog-impact.test.jsx`, which currently pin the hand-built form.

**8. The catalogue card has no accessible name** &nbsp;`nit`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:52`

The previous implementation opened the panel with an `<h3>` naming it "N products affected by your active pricing rules". The rebuilt card leads with `ImpactStats`, which renders only `<span>` elements inside an `HStack`, so the `Card.Root` region carries no heading and no accessible name. For someone navigating the page by heading, the section below the rules table is now unlabelled and the first thing encountered is a bare number. Leading with the stat is clearly the design intent, so a screen-reader-only heading or an `aria-label` on the card would restore the landmark without touching the visual treatment.

**9. The truncation caption is duplicated verbatim against `rule-preview.tsx`** &nbsp;`nit`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/catalog-impact.tsx:82`

This `_n( 'Showing a sample of %s product.', … )` block, its translator comment and the `preview_limited && sample_count >= IMPACT_SAMPLE_LIMIT` guard around it are byte-identical to the equivalent block in `rule-preview.tsx`. Two copies of the same translatable string are a maintenance hazard specific to i18n: editing one produces a second msgid that translators have to service separately, and the `.pot` generators in this workspace are manual, so the drift would not be caught by CI. Since this PR already extracted `IMPACT_SAMPLE_LIMIT` into `constants.ts` for exactly this reason, the note itself belongs alongside it; `impact-format.ts` already hosts `cycleMarkerNote()` as precedent for sharing a string so two hosts cannot drift apart. Separately, because the guard requires `sample_count >= 50`, the singular form is currently unreachable — harmless, but translators are servicing a string that cannot render.

**10. Two `formatCount` implementations now format counts on the same screen** &nbsp;`nit`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:25`

This file imports `formatCount` from `packages/components/src/breadcrumbs/format-count`, while `catalog-impact.tsx` and `impact-stats.tsx` import a same-named local from `./impact-format`. They are not equivalent: the shared one does `lang.replace( '_', '-' )` before handing the tag to `toLocaleString`, with a comment explaining that WordPress ships locales `Intl` rejects, whereas `impact-format.ts` passes `document.documentElement.lang` through untouched and relies on its `try`/`catch` to fall back to the browser locale. On a `pt_PT` site the header count is grouped for Portuguese and the "Products affected" figure beside it for whatever the browser is set to. Pointing `impact-format.ts` at the shared helper would make the two numbers on this screen agree.

**11. The `supported: false` branch is the one catalogue outcome no test pins** &nbsp;`nit`

`plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx:340`

`catalog-impact.test.jsx` and `list.test.jsx` between them cover the catalogue read landing, failing, and returning `total_matching: 0`, but nothing exercises `supported: false`. That branch is real and structurally different: the engine's `build_payload()` returns a bare `[ 'supported' => false ]` with `total_matching`, `count_limited` and `currency` all absent, so `stats?.supported` is the only thing standing between that payload and `CatalogImpact` reading `stats.total_matching` off `undefined`. `CatalogImpactResponse` declares those fields as required, so TypeScript gives no cover — the guard is purely a runtime contract. A case asserting the card is absent while the table still renders would pin it as cheaply as the existing rejection test does.

thomasguillot:
## Review follow-up

All eleven findings from the review above are addressed in `c857a36d5`, `c842d4790` and `92bc250d6`.

**The modal no longer races itself.** Reopening the panel before its first request lands was starting a second full catalogue re-price, and a late failure from the first could clear a table that had already loaded. A request generation counter now lets only the newest response write state, and an in-flight flag stops the duplicate request being issued at all. Retry-on-reopen after a genuine failure still works, and is still pinned by its test.

**The modal stands down instead of trusting the payload.** `supported` and an empty `sample` are both checked before the table walks the response, matching what `rule-preview.tsx` already did with the same response type. An engine that answers the catalogue read with its minimal payload now gets the existing empty treatment rather than a render-time `TypeError`.

**Trashing a rule refreshes the catalogue.** The headline count and the sample cached behind the card were both surviving a trash, so part of the screen kept describing a rule that was gone. The catalogue read is now a callback invoked alongside the rules refetch, and the card is keyed so its cached sample cannot outlive the rule set it describes.

**A stalled catalogue read no longer holds the page.** The screen still waits on both reads so it arrives as one unit, but the gate releases on its own after 8s. The engine's catalogue walk is not bounded by the `limit`, so a slow or wedged route was able to leave the rules behind a spinner indefinitely even though they had already arrived. Worth knowing this is visible: past the timeout the table appears first and the card follows when the read lands.

**Loading states are no longer silent.** All three spinners carry `role="status"` and a screen-reader-only label, so success is announced rather than being the one outcome nothing conveys. The card also regained a heading, screen-reader-only, so the section below the table is reachable by heading navigation again without changing how it looks.

Smaller items: both reads now build their query strings with `addQueryArgs` rather than by hand; the truncation caption is shared between the modal and the editor preview so the two cannot drift into separate msgids; `impact-format.ts` re-exports the shared count formatter instead of carrying its own, so the header count and the figures beside it group for the same locale; and the comment justifying `?limit=1` now names the `audience` block it also relies on.

Tests are up to 807 across 84 suites from 799, covering the reopen-mid-fetch case, both stand-down branches, the trash refetch, the gate release, and the announcements. `lint:js` is clean at 0 errors, `lint:scss` is clean, the production build compiles, and `tsc` is unchanged at its 218 pre-existing errors.

miguelpeixe:
> two things here need the dynamic pricing engine and cannot be done in this repo.

The engine side was updated, which settles the open question in `list.tsx:130` and opens up three small things here:

**`audience` is limit-invariant by construction.** `catalog_split()` never receives the `limit` param. It takes its scope from the catalogue walk the route already runs, plus its own subscriber ceiling, so `total`, `caught` and `protected` are identical at `limit=1` and at `limit=50`. Nothing compares the two limits in a test today, so if that's worth pinning rather than reasoning about, I'm glad to add it on the engine side.

**The hardcoded 50 can go.** The payload now echoes the cap the engine actually applied, so the guard at `catalog-impact.tsx:111` can read `detail.sample_limit` instead of `IMPACT_SAMPLE_LIMIT`. It's optional in the payload — `build_payload()` omits the key rather than guessing when a caller supplies no cap, since the per-rule and catalogue entry points default differently — so `sample_limit?: number` is the shape to declare. `audience` is worth adding to `CatalogImpactResponse` at the same time; it's still only on `RulePreviewResponse`.

**`count_limited` covers `caught` and `protected` too.** `ImpactStats` bounds `scope.total` but renders the other two through bare `formatCount`. When the ceiling is hit the loss isn't even: the subscription query orders by id ascending, so truncation keeps the oldest subscribers, which are the ones a `subscription_started_after` gate marks protected. A truncated result under-reports `caught` and over-reports `protected`, identically on every load, so the card reads a systematic skew as fact. Bounding all three, or dropping the derived pair when the flag is set, would close it.

**One note for the 8s gate.** The audience block rides on the same `?limit=1` read and isn't bounded by the limit either: one subscription query per active rule, then up to 2,000 subscription hydrations. The engine change removed a duplicate scope enumeration, so the walk itself got cheaper, but on a store with real subscriber volume this read is now heavier than the catalogue walk alone. Worth factoring into where that timeout sits.

thomasguillot:
Landed in `af7b6cb9d`.

**`sample_limit` is what both previews measure against now**, with `IMPACT_SAMPLE_LIMIT` demoted to the fallback for when the key is absent. The modal and the editor preview had each written out the same `preview_limited && sample_count >= 50` guard, so it moved into a single `sampleNote( payload )` rather than being corrected in two places. `sample_limit?: number` and `audience` are both on `CatalogImpactResponse`.

**The renewal split is bounded.** "Eligible at renewal" and its protected note take the same `N+` treatment the total already had, so a truncated split no longer sits beside a bounded total claiming to be exact.

**On the invariance:** the comment at `list.tsx:130` states it now instead of hedging. A test on the engine side would be welcome, since this screen reads the block from the `?limit=1` payload.

**On the 8s gate**, I'd rather leave it where it is. The gate is there to bound a read the limit doesn't bound, so a heavier read is an argument for keeping it short rather than lengthening it: past the timeout the screen degrades to table-first with the card following, which is what I'd want on a store where the subscriber walk is slow, whereas a longer gate just extends the blank. If arriving as one unit turns out to matter more than that at real volume, the better change is dropping the joint gate and giving the card its own spinner, which was your original suggestion.

Checked against current `trunk` locally: `sample_limit` echoes on both routes, and `audience` reaches the catalogue read with `application: mixed`.

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx
```diff
@@ -67,15 +73,27 @@ export default function PricingRulesList() {
...
+	// Remounts the card so its cached product sample cannot outlive a trash.
+	const [ statsVersion, setStatsVersion ] = useState( 0 );
 	const [ view, setView ] = useState< View >( DEFAULT_VIEW );
 	const [ segmentMap, setSegmentMap ] = useState< Record< number, string > >( {} );
 	const [ currency, setCurrency ] = useState< PricingRulesCurrency >( { code: '', symbol: '', decimals: 2 } );
 
+	const isReady = hasResolved && statsResolved;
+	// Raw payload, not the filtered view: a search that matches nothing keeps the
+	// DataViews treatment rather than claiming the site has no rules.
+	const isEmpty = isReady && ! hasError && data.length === 0;
```
miguelpeixe:
```suggestion
	const isEmpty = isReady && ! isLoading && ! hasError && data.length === 0;
```

Best not to risk presenting the onboarding screen while data is being fetched. Worth considering another UI state in this case?

miguelpeixe (changes requested):

plugins/newspack-plugin/src/wizards/audience/views/pricing-rules/list.tsx
thomasguillot: Applied, thanks.

`isReady` was already gating the first load, since `hasResolved` only flips in the first read's `finally`, but you are right that it says nothing about a refetch. With `! isLoading` in there, the empty state can no longer answer "does this site have rules" from a payload that is already being replaced.

On a second UI state: there is one, the page spinner behind `! isReady`, and it is deliberately first-load only. A refetch after trashing a rule keeps the table on screen rather than blanking it, which is what `hasResolved` separates from `isLoading`. So the three states are the spinner on first load, the table with its own loading treatment on a refetch, and the onboarding screen only once a settled read comes back with nothing.

miguelpeixe (approved):
Thank you!

PR #846, opened 2026-08-11
feat(blocks): reorder the content picked in Static mode (NPPD-946)

thomasguillot:
### All Submissions:

A Homepage Posts or Carousel block set to Static mode lets a publisher pick specific posts, but there has been no way to reorder them. Moving one out-of-place story to the top means deleting every token after it and retyping them.

A **Reorder Content** button now sits below the Content field and opens a modal listing the selection in order. Rows move by dragging, or with per-row up and down chevrons, so the feature works with a mouse, a keyboard and on touch. Save writes the new order back; closing with unsaved changes asks before discarding.

Reorder Content button in the block sidebar, highlighted in the post editor

Reorder Content modal open over the post editor, listing five posts in order

Discard confirmation raised over the reorder modal after a row was moved

Both blocks gain the feature from one change, since they share `src/components/query-controls.js`.

**No PHP changes.** `Newspack_Blocks::build_articles_query()` already sets `post__in` with `orderby => post__in`, and the editor preview endpoint does the same, so the stored array order was honoured end to end already. The gap was only in the editor UI.

Notes for reviewers:

* Rows are `@wordpress/ui` Card components. The package is in the dependency-extraction plugin's `BUNDLED_PACKAGES` list, so it bundles rather than resolving a `wp.ui` global — which matters, because WordPress does not ship one. Card's compiled CSS carries hardcoded fallbacks for every design token it reads, so no token stylesheet is imported; importing one placed a `:root` block after newspack-components' brand remap and reset 15 `--wpds-*` properties to upstream values across the editor.
* The modal owns its own Escape, ✕ and backdrop handling rather than using `Modal`'s. Core defers `onRequestClose` behind `closeModal()`, which plays the exit animation first, so a close vetoed by the discard prompt made the modal visibly fade out and back in.
* IDs are never re-derived from labels. Every save is a permutation of the array the modal opened with, so no path can drop, duplicate or retype an ID. A post that is later trashed keeps its slot as a `(no title)` row rather than being silently dropped from the block.

NPPD-946

1. Add a **Homepage Posts** block to a page and set Mode to **Static**.
2. Pick one post. **Reorder Content** stays disabled. Pick a second and it enables.
3. Open it, drag the last row to the top, and Save. The block preview redraws in the new order.
4. Reopen it. **Save** is disabled until you move something, and moving a row back to where it started disables it again.
5. Move a row, then press Escape. The discard prompt appears over the modal, which should not move, fade or flicker. Repeat with the ✕ and with a click on the backdrop.
6. Choose **Keep editing** to return to the list, or **Discard** to close. Reopen and confirm the original order survived.
7. Reorder with the chevrons only. Walking a row to the top moves focus to that row's other chevron once the one you pressed disables.
8. Publish the page and check the front end renders in the saved order.
9. Repeat step 3 on a **Carousel** block.
10. With a post selected, trash it in another tab, reload the editor and reopen the modal. A `(no title)` row holds its place, and reordering does not drop it.

### Other information:

101 tests across 6 suites. Coverage includes the array-move helper, the chevron reorder and focus handoff, the disabled thresholds, all four close paths, and the backdrop press-target logic that keeps a row drag released outside the modal from dismissing it.

### DIFF
--- a/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
+++ b/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
@@ -45,15 +45,18 @@ class AutocompleteTokenField extends Component {
 		if ( this.isFetchingInfoOnLoad() ) {
 			const { tokens, fetchSavedInfo } = this.props;
 
-			fetchSavedInfo( tokens ).then( results => {
-				const { validValues } = this.state;
+			fetchSavedInfo( tokens )
+				.then( results => {
+					const { validValues } = this.state;
 
-				results.forEach( suggestion => {
-					validValues[ suggestion.value ] = suggestion.label;
-				} );
+					results.forEach( suggestion => {
+						validValues[ suggestion.value ] = suggestion.label;
+					} );
 
-				this.setState( { validValues, loading: false } );
-			} );
+					this.setState( { validValues, loading: false } );
+				} )
+				// Without this the field spins forever when the request fails.
+				.catch( () => this.setState( { loading: false } ) );
 		}
 	}
 
--- a/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
+++ b/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
@@ -55,8 +55,17 @@ class AutocompleteTokenField extends Component {
 
 					this.setState( { validValues, loading: false } );
 				} )
-				// Without this the field spins forever when the request fails.
-				.catch( () => this.setState( { loading: false } ) );
+				.catch( () => {
+					// Without labels the field would render no tokens, and the next
+					// selection would then write back a list missing every saved ID.
+					const { validValues } = this.state;
+
+					tokens.forEach( token => {
+						validValues[ token ] = String( token );
+					} );
+
+					this.setState( { validValues, loading: false } );
+				} );
 		}
 	}
 
--- a/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
+++ b/plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
@@ -55,7 +55,9 @@ class AutocompleteTokenField extends Component {
 
 					this.setState( { validValues, loading: false } );
 				} )
-				.catch( () => {
+				.catch( error => {
+					// eslint-disable-next-line no-console
+					console.error( 'Newspack Blocks: could not load the titles for the saved selection.', error );
 					// Without labels the field would render no tokens, and the next
 					// selection would then write back a list missing every saved ID.
 					const { validValues } = this.state;
--- a/plugins/newspack-blocks/src/components/reorder-modal.js
+++ b/plugins/newspack-blocks/src/components/reorder-modal.js
@@ -38,7 +38,19 @@ export const moveItem = ( items, from, to ) => {
 // than the native attribute.
 const isAvailable = button => !! button && ! button.matches( ':disabled, [aria-disabled="true"]' );
 
+// Speech input matches on the visible label, so it has to open the accessible
+// name. Composing both from one translated string keeps that true in every locale.
+const nameWithTitle = ( label, title ) =>
+	sprintf(
+		/* translators: 1: the control's visible label. 2: title of the content being moved. */
+		__( '%1$s: %2$s', 'newspack-blocks' ),
+		label,
+		title
+	);
+
 const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
+	const moveUpLabel = __( 'Move Up', 'newspack-blocks' );
+	const moveDownLabel = __( 'Move Down', 'newspack-blocks' );
 	const [ items, setItems ] = useState( null );
 	const [ dragIndex, setDragIndex ] = useState( null );
 	const [ hasFetchError, setHasFetchError ] = useState( false );
@@ -48,6 +60,7 @@ const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
 	const pendingFocus = useRef( null );
 	const closeRef = useRef( null );
 	const preDragItems = useRef( null );
+	const dragBlocked = useRef( false );
 
 	// The order the modal opened with. Snapshotted so a change to `ids` from
 	// elsewhere in the editor cannot desynchronise it from `items`.
@@ -200,18 +213,33 @@ const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
 				size="medium"
 				className="newspack-blocks-reorder-modal"
 			>
-				<div className="newspack-blocks-reorder-modal__body">
-					{ ! items ? (
+				<div
+					className="newspack-blocks-reorder-modal__body"
+					aria-busy={ ! items }
+					// Rows are the only drop targets, so without this every gap between
+					// them reports a cancelled drag and would undo the reorder.
+					onDragOver={ event => event.preventDefault() }
+					onDrop={ event => event.preventDefault() }
+				>
+					{ ! items && (
 						<div className="newspack-blocks-reorder-modal__loading">
 							<Spinner />
 						</div>
-					) : (
+					) }
+					{ items && hasFetchError && (
 						<>
-							{ hasFetchError && (
-								<Notice status="error" isDismissible={ false } className="newspack-blocks-reorder-modal__notice">
-									{ __( 'The content could not be loaded, so the order cannot be saved.', 'newspack-blocks' ) }
-								</Notice>
-							) }
+							<Notice status="error" isDismissible={ false } className="newspack-blocks-reorder-modal__notice">
+								{ __( 'The content could not be loaded, so the order cannot be saved.', 'newspack-blocks' ) }
+							</Notice>
+							<div className="newspack-blocks-reorder-modal__footer">
+								<Button variant="tertiary" onClick={ requestClose }>
+									{ __( 'Cancel', 'newspack-blocks' ) }
+								</Button>
+							</div>
+						</>
+					) }
+					{ items && ! hasFetchError && (
+						<>
 							{ /* `list-style: none` drops list semantics in Safari, so the role is restated. */ }
 							{ /* eslint-disable-next-line jsx-a11y/no-redundant-roles */ }
 							<ul className="newspack-blocks-reorder-modal__list" role="list" ref={ listRef }>
@@ -224,10 +252,13 @@ const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
 											'is-dragging': dragIndex === index,
 										} ) }
 										draggable
+										// `dragstart` always fires at the draggable row, never at the
+										// chevron inside it, so the grab point is recorded separately.
+										onPointerDown={ event => {
+											dragBlocked.current = !! event.target.closest?.( 'button' );
+										} }
 										onDragStart={ event => {
-											// The chevrons live inside the drag source, so a click that
-											// drifts would otherwise drag the row instead of stepping it.
-											if ( event.target.closest?.( 'button' ) ) {
+											if ( dragBlocked.current ) {
 												event.preventDefault();
 												return;
 											}
@@ -258,14 +289,8 @@ const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
 												data-direction="up"
 												disabled={ 0 === index }
 												accessibleWhenDisabled
-												label={ __( 'Move Up', 'newspack-blocks' ) }
-												// The visible label has to stay a substring of the
-												// accessible name so speech input can reach it.
-												aria-label={ sprintf(
-													/* translators: %s: title of the content being moved. */
-													__( 'Move Up: %s', 'newspack-blocks' ),
-													item.label
-												) }
+												label={ moveUpLabel }
+												aria-label={ nameWithTitle( moveUpLabel, item.label ) }
 												onClick={ () => moveTo( index, index - 1, 'up' ) }
 											/>
 											<Button
@@ -274,12 +299,8 @@ const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
 												data-direction="down"
 												disabled={ index === items.length - 1 }
 												accessibleWhenDisabled
-												label={ __( 'Move Down', 'newspack-blocks' ) }
-												aria-label={ sprintf(
-													/* translators: %s: title of the content being moved. */
-													__( 'Move Down: %s', 'newspack-blocks' ),
-													item.label
-												) }
+												label={ moveDownLabel }
+												aria-label={ nameWithTitle( moveDownLabel, item.label ) }
 												onClick={ () => moveTo( index, index + 1, 'down' ) }
 											/>
 										</span>
--- a/plugins/newspack-blocks/src/components/reorder-modal.js
+++ b/plugins/newspack-blocks/src/components/reorder-modal.js
@@ -1,0 +1,292 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+import { speak } from '@wordpress/a11y';
+import { useEffect, useRef, useState } from '@wordpress/element';
+import {
+	Button,
+	Modal,
+	Spinner,
+	__experimentalConfirmDialog as ConfirmDialog, // eslint-disable-line @wordpress/no-unsafe-wp-apis
+} from '@wordpress/components';
+import { Icon, chevronDown, chevronUp, close, dragHandle } from '@wordpress/icons';
+import { Card } from '@wordpress/ui';
+
+/**
+ * Internal dependencies
+ */
+import './reorder-modal.scss';
+
+export const moveItem = ( items, from, to ) => {
+	if ( from === to || from < 0 || to < 0 || from >= items.length || to >= items.length ) {
+		return items;
+	}
+	const next = [ ...items ];
+	const [ moved ] = next.splice( from, 1 );
+	next.splice( to, 0, moved );
+	return next;
+};
+
+// Disabled controls stay focusable here, so they report `aria-disabled` rather
+// than the native attribute.
+const isAvailable = button => !! button && ! button.matches( ':disabled, [aria-disabled="true"]' );
+
+const ReorderModal = ( { title, ids, fetchItems, onSave, onClose } ) => {
+	const [ items, setItems ] = useState( null );
+	const [ dragIndex, setDragIndex ] = useState( null );
+	const [ isConfirmingDiscard, setIsConfirmingDiscard ] = useState( false );
+	const listRef = useRef( null );
+	const overlayRef = useRef( null );
+	const pendingFocus = useRef( null );
+	const closeRef = useRef( null );
+
+	// `ids` is the order the modal opened with: it remounts on every open.
+	const isDirty = !! items && items.some( ( item, index ) => item.id !== ids[ index ] );
+
+	useEffect( () => {
+		let cancelled = false;
+		const withLabels = labels => ids.map( id => ( { id, label: labels[ id ] || __( '(no title)', 'newspack-blocks' ) } ) );
+
+		fetchItems( ids )
+			.then( results => {
+				if ( cancelled ) {
+					return;
+				}
+				const labels = {};
+				results.forEach( ( { value, label } ) => {
+					labels[ value ] = label;
+				} );
+				setItems( withLabels( labels ) );
+			} )
+			.catch( error => {
+				// eslint-disable-next-line no-console
+				console.error( 'Newspack Blocks: could not prepare the content to reorder.', error );
+				if ( ! cancelled ) {
+					setItems( withLabels( {} ) );
+				}
+			} );
+
+		return () => {
+			cancelled = true;
+		};
+	}, [] ); // eslint-disable-line react-hooks/exhaustive-deps
+
+	// A chevron that reaches an end of the list stops accepting moves, so send
+	// focus to the row's other chevron.
+	useEffect( () => {
+		const pending = pendingFocus.current;
+		pendingFocus.current = null;
+		if ( ! pending || ! listRef.current ) {
+			return;
+		}
+		const row = listRef.current.querySelector( `[data-item-id="${ pending.id }"]` );
+		if ( ! row ) {
+			return;
+		}
+		const preferred = row.querySelector( `[data-direction="${ pending.direction }"]` );
+		const fallback = row.querySelector( `[data-direction="${ 'up' === pending.direction ? 'down' : 'up' }"]` );
+		if ( isAvailable( preferred ) ) {
+			preferred.focus();
+		} else if ( isAvailable( fallback ) ) {
+			fallback.focus();
+		}
+	}, [ items ] );
+
+	const moveTo = ( from, to, direction ) => {
+		const next = moveItem( items, from, to );
+		if ( next === items ) {
+			return;
+		}
+		pendingFocus.current = { id: items[ from ].id, direction };
+		setItems( next );
+		speak(
+			sprintf(
+				/* translators: 1: new position of the moved item. 2: total number of items. */
+				__( 'Moved to position %1$d of %2$d.', 'newspack-blocks' ),
+				to + 1,
+				items.length
+			)
+		);
+	};
+
+	const requestClose = () => {
+		if ( isDirty ) {
+			setIsConfirmingDiscard( true );
+			return;
+		}
+		onClose();
+	};
+
+	useEffect( () => {
+		closeRef.current = requestClose;
+	}, [ requestClose ] );
+
+	const handleKeyDown = event => {
+		if ( 'Escape' !== event.key || event.defaultPrevented ) {
+			return;
+		}
+		event.preventDefault();
+		requestClose();
+	};
+
+	// `Modal` plays its exit animation before it reports a close request, which
+	// would fade the modal out and back in whenever the confirmation vetoes one.
+	// Its own dismissal handlers are off, so drive the overlay press here.
+	useEffect( () => {
+		const overlay = overlayRef.current;
+		if ( ! overlay ) {
+			return;
+		}
+		let pressedOverlay = false;
+		const onPointerDown = event => {
+			pressedOverlay = event.target === overlay;
+			if ( pressedOverlay ) {
+				event.preventDefault();
+			}
+		};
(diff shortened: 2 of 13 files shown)


thomasguillot:
## Review: release risk LOW — approve

Verified the round-4 changes end to end, including reading the installed `@wordpress/components` 30.9.0 `Modal`, `Button`, `useModalExitAnimation`, `useHStack` and `VisuallyHidden` sources rather than reasoning from memory. Tests and linters re-run locally: 119 passed across 6 suites, ESLint and Stylelint clean on all changed files.

**No findings. The branch is clean.**

### What I checked and how it resolved

**The frame effect's lifetime and the frame's existence at effect time.** `Modal` merges `forwardedRef` onto the *overlay* div (`.components-modal__screen-overlay`), not the frame, so `overlayRef.current?.querySelector( '.components-modal__frame' )` is a correct descendant lookup rather than a self-match that would silently return `null`. `useMergeRefs` assigns synchronously inside a callback ref, which React invokes in the layout phase; `ReorderModal`'s `useEffect` is a passive effect and runs strictly after, in the same commit that mounts the portal. The frame is unconditionally rendered — `useModalExitAnimation` only toggles an overlay class and a CSS custom property, it never remounts or key-swaps the frame — so the node the listeners attach to is the node that lives for the modal's whole life. The early `return` on a missing frame is therefore unreachable in practice, not a silent no-op path.

**Leaks.** `accept` is stateless, both listeners are removed against the same captured node on unmount, and the node is never replaced, so there is no stale-node or duplicate-registration path. `dragover`/`drop` are not passive-by-default, so `preventDefault()` genuinely applies without an explicit `{ passive: false }`.

**Ordering against React's delegation.** React 18 delegates portal events at the portal container (`document.body`), which is an *ancestor* of the frame, so the native frame listener fires on the bubble at the frame **before** React dispatches the row's `handleDragOver`. Nothing reads `defaultPrevented` on drag events (`Modal`'s only `defaultPrevented` check is on `keydown`), and `handleDragOver` sets `dropEffect` after, so the reordering happens as before. The `effectAllowed = 'move'` set at `dragstart` means an un-set `dropEffect` on a cancelled dragover initialises to `move`, and `onDragEnd` only tests for the literal `'none'`, so the revert check stays correct even if a browser initialises it to `copy` instead.

**Whether anything inside the frame should still cancel.** No. The only drags that reach the frame with `dragIndex === null` are external ones (files, blocks, links). Cancelling those was already the behaviour for the body region; extending it to the header, padding and footer swallows a drop that would otherwise let the browser navigate away from an open editor, which is strictly safer. There is no region of the frame where a release should read as a cancel — that meaning now belongs solely to Escape and a release outside the frame, which the overlay test pins.

**The tokenfield catch-seed and the data-loss path.** It closes it. `getLabelsForValues` renders one token per saved ID, and `getValuesForLabels` maps each ID-as-label back to its own key, so a subsequent pick writes `[ ...savedIds, newId ]` rather than `[ newId ]`. No raw label leaks into the save: `getValuesForLabels` returns `Object.keys( validValues )` entries, which are IDs and never labels — and `Object.keys` returns integer-like keys in ascending numeric order, so a same-string label never resolves to a foreign key ahead of its own. A later successful `updateSuggestions` overwrites the seeded label in place under the same key, so a token silently upgrading from `"11"` to its real title keeps mapping back to `11`. I did consider the collision case (a searched post whose title is exactly the decimal string of another saved post's ID, under a failed titles fetch, making that suggestion unpickable) and am deliberately not filing it: it requires both conditions at once, and the label-keyed design already collides on two posts sharing a title, so the seed does not widen the class of failure in any way worth a change.

**The button's tooltip and accessible name.** Traced through `UnforwardedButton`: `shouldShowTooltip` needs `showTooltip && !!label`, so `label` is genuinely load-bearing; `tooltipProps.text` resolves to `children?.length && description ? description : label`, so with visible children present the tooltip renders the description alone. `aria-label` falls back to `label`, which is the same translated string as the visible text, so the accessible name is `Reorder Content` in every locale. The `description` prop landed in `@wordpress/components` 28.4.0, one minor *after* `accessibleWhenDisabled` (28.3.0) which this branch already relies on and which was settled earlier — so it raises no new runtime floor. The `VisuallyHidden` description div is `position: absolute`, and `useHStack` only wraps `Spacer` children in `FlexItem`, so it is an abspos child of the flex container rather than a flex item and contributes no extra `spacing={2}` gap under the button.

**Translator comment.** Changing only the comment leaves the msgid `%1$s: %2$s` untouched, so no existing translation is invalidated.

Files reviewed: `plugins/newspack-blocks/src/components/reorder-modal.js`, `plugins/newspack-blocks/src/components/specific-posts-control.js`, `plugins/newspack-blocks/src/components/autocomplete-tokenfield.js`, plus both test files.

<details>
<summary>Runtime verification behind this verdict</summary>

Exercised against a local site on the built bundle, not only in unit tests:

- **Chevron reorder round trip.** Walking the last row to the top and saving took `specificPosts` from `["3063","3025","3033","3043","2997"]` to `["2997","3063","3025","3033","3043"]` — same IDs, correct permutation. Reopening the modal showed the saved order.
- **Focus handoff.** When the pressed `Move Up` disabled at the top of the list, focus moved to that row's `Move Down`.
- **Drag guard, both directions, same source row and same destination.** Grabbing the chevron left the order unchanged with `Save` still disabled; grabbing the row body moved it from 5th to 2nd and enabled `Save`.
- **Drop surface.** A release in the modal frame's padding band — inside the dialog, outside the list — reported `dropEffect=move`, so the order survives. A release on the overlay reported `dropEffect=none`, so it correctly cancels.
- **Discard path.** Escape after a drag raised the confirmation with the modal held open; `Discard` closed it and left the stored order untouched.
- **Rendered attributes.** `aria-busy` on the body, `role="list"` on the list, the full headline in each row's `title`, and accessible names of the form `Move Up: <title>`.

</details>

thomasguillot:
## Final pass

Read the whole diff once more looking for anything left, including nits. No defects remained. Two clarity changes only, both verified behaviour-identical before and after:

- **`reorder-modal.js`** — on a failed titles fetch, `items` was populated with a full set of `(no title)` rows that the error view never renders. It now sets an empty list, with a comment saying the assignment exists only to leave the loading state. `isDirty`, `aria-busy` and the rendered output are unchanged.
- **`specific-posts-control.js`** — the comment on the shared titles request still described the cache being cleared unconditionally on rejection, which stopped being true when the eviction was scoped to the entry that owns the slot.

Also checked and left alone: `console.error` on the fetch failure stays an error rather than a warning now that the failure is surfaced in the UI; the unconditional `preventDefault` on drag events over the dialog is deliberate, since swallowing a foreign drop is safer than letting the browser navigate away from an open editor.

119 tests across 6 suites, ESLint and Stylelint clean, and the production bundle builds with no missing-export warnings.

### Summary of the review round

Four rounds. Three defects worth naming, all in the drag handling and none caught by tests, linters or static reading:

1. **A cancelled drag committed the hovered order.** Dragging a row and then pressing Escape, or releasing outside the window, left the list in whatever order the pointer had swept through.
2. **The chevrons sat inside the drag source.** Native drag-and-drop fires `dragstart` at the draggable row, never at the button inside it, so a guard keyed on the event target could not work — a click on a chevron that drifted a few pixels dragged the whole row instead of stepping it.
3. **The drop surface stopped short of the dialog's padding.** With rows as the only drop targets, releasing in the 32px band around the list reported a cancelled drag and reverted a reorder the user had just watched happen. The grip is the leftmost element in a row, so dragging by the handle put the pointer right beside that band.

Each was found by exercising the behaviour in a browser rather than reading the code, and 2 and 3 were introduced while fixing 1. The drag model is now covered by tests for the guard in both directions, the cancel-restore, and the drop surface.

plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
```diff
@@ -45,15 +45,27 @@ class AutocompleteTokenField extends Component {
...
+					this.setState( { validValues, loading: false } );
+				} )
+				.catch( () => {
+					// Without labels the field would render no tokens, and the next
+					// selection would then write back a list missing every saved ID.
+					const { validValues } = this.state;
 
-				this.setState( { validValues, loading: false } );
-			} );
+					tokens.forEach( token => {
```
thomasguillot, line 65: **Suggestion:** The fetch-failure fallback shows post IDs as titles, with nothing saying the titles failed — The mechanics here are right, and this closes a real gap: before it, a rejected fetch left `validValues` empty, `getTokens()` returned `[]`, and the next selection wrote back a list missing every saved ID. `String( token )` also keeps each label unique, so the IDs round-trip through `getValuesForLabels` unchanged, and a shared `(no title)` would have collapsed them.

What is missing is the signal. The field renders tokens reading `11`, `22`, `33` and `loading: false` removes the spinner, so it looks settled and correct. `fetchSavedInfoOnce` clears its shared slot on the same rejection, so opening the modal fires a fresh request that may succeed, leaving the modal showing real titles above a field still showing numbers. A publisher's reasonable read is that the block is broken, and the repair is to re-pick everything by hand. The modal handles the same failure with an explicit `Notice`; the field could say as much.

**Risk if not addressed:** If the taxonomy REST calls fail persistently for editors on a site, say because a security plugin or an edge proxy blocks `/wp/v2/categories` for logged-in roles, then every Content Loop and Carousel block renders Excluded categories and Excluded tags as unrecognisable numbers while the field looks fully loaded. An editor who tidies those numbers away writes an empty exclusion list, and since the exclusion only applies when the list is non-empty, categories the publisher deliberately kept off the homepage start appearing to readers. Nothing reaches the console, so there is no trace connecting the homepage change to the failing request. The fallback itself is an improvement, since without it the same failure wiped the exclusions on the next edit with no user intent at all; what it removes is the last remaining signal that anything went wrong.

**Consider:** Keeping the fallback is the right call; it is the silence that is worth a second look. Before this change a rejected saved-info fetch at least left an unhandled rejection in the console and a spinner that never stopped. A single `console.error` in the catch, matching what the reorder modal already does on its own fetch failure, would keep the cause findable if a publisher reports that a block's tokens turned into numbers or that an exclusion list emptied itself.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+	// The order the modal opened with. Snapshotted so a change to `ids` from
+	// elsewhere in the editor cannot desynchronise it from `items`.
+	const openedWith = useRef( ids );
+
+	const isDirty = !! items && items.some( ( item, index ) => item.id !== openedWith.current[ index ] );
+
+	useEffect( () => {
+		let cancelled = false;
+		const openIds = openedWith.current;
+		const withLabels = labels => openIds.map( id => ( { id, label: labels[ id ] || __( '(no title)', 'newspack-blocks' ) } ) );
```
thomasguillot, line 88: **Suggestion:** `(no title)` covers two different situations, and the row disagrees with the field above it — `withLabels` falls back to `(no title)` for any ID the fetch did not return, which is the same string `fetchSavedPosts` already gives a post whose title is genuinely empty (`query-controls.js:70`). The sibling helpers in that file keep the two apart: an unmatched category becomes `Deleted category` (`query-controls.js:153`) and an unmatched tag `Deleted tag` (`query-controls.js:194`).

In the editor, with one of five picked posts moved to the trash, the modal held its slot as a `(no title)` row and reordering preserved the ID, exactly as the description says. The Content field above it showed four tokens, so the two controls disagreed about what was selected, and the front end rendered four posts. Adding one post through the autocomplete then rewrote the attribute from `[17,5,8,11,14]` to `["17","5","8","14","32"]`, dropping ID 11. `getValuesForLabels` is untouched here, so that drop is older than this branch; what is new is a row that presents the orphan as ordinary, reorderable content.

**To reproduce**
1. Add a Content Loop block, set Mode to Static and pick five posts.
2. `wp post update <third-id> --post_status=trash`.
3. Reload the editor and open Reorder Content: the trashed post holds its slot as `(no title)`, while the Content field shows four tokens.
4. Type a title in the Content field and pick a suggestion. The trashed ID is gone from `specificPosts`.

A distinct label for an ID that did not resolve would follow the convention already in `query-controls.js` and give the publisher something to act on.

**Consider:** `fetchSavedPosts` is already being edited here, so it is the natural place to close the gap between the two controls. Mirroring `fetchSavedCategories` and `fetchSavedTags` a few dozen lines below, walking `postIDs` after the map and pushing an entry for any ID the endpoint did not return would make the ID render as a token, so a later edit preserves it, and resolve it in the modal so the row stops reading `(no title)`. One caveat while you are in there: the token field matches a label back to its ID by finding the first key with that label, so two IDs sharing an identical fallback label would collapse onto one. Putting the ID in the label, in the style the suggestion-collision handling already uses (`${ label } (${ value })`), avoids that.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+			overlay.removeEventListener( 'pointerup', onPointerUp );
+		};
+	}, [] );
+
+	// Rows are the only drop targets, and the frame's own padding is not one, so a
+	// release a few pixels wide of the list would report a cancelled drag and undo
+	// the reorder. Accepting the drop across the whole dialog leaves "cancelled"
+	// meaning only Escape or a release outside it.
+	useEffect( () => {
+		const frame = overlayRef.current?.querySelector( '.components-modal__frame' );
```
thomasguillot, line 218: **Suggestion:** Three behaviours rest on `Modal` forwarding its ref to the overlay, and all three fail quietly — `<Modal ref={ overlayRef } …>` assumes the forwarded ref lands on `.components-modal__screen-overlay`. Two things then depend on it: the pointer listeners bind to `overlayRef.current` directly (`:182`), and `const frame = overlayRef.current?.querySelector( '.components-modal__frame' )` walks down from it. That holds for the `@wordpress/components` copy Jest resolves, but at runtime the editor uses core's `wp.components` global, so the tests cannot vouch for the version publishers actually run.

If core ever forwards that ref to the frame or the content node instead, the `?.` and the `if ( ! overlay )` guard both swallow it: click-outside stops closing, an overlay press stops raising the discard prompt, and a drag released in the gutter between rows reverts the reorder. Deriving the overlay defensively, say `node.closest( '.components-modal__screen-overlay' )` with a document query behind it, would make the coupling explicit and survive a core change.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+				isDismissible={ false }
+				shouldCloseOnEsc={ false }
+				shouldCloseOnClickOutside={ false }
+				headerActions={ <Button size="compact" icon={ close } label={ __( 'Close', 'newspack-blocks' ) } onClick={ requestClose } /> }
+				size="medium"
+				className="newspack-blocks-reorder-modal"
+			>
+				<div className="newspack-blocks-reorder-modal__body" aria-busy={ ! items }>
+					{ ! items && (
+						<div className="newspack-blocks-reorder-modal__loading">
```
thomasguillot, line 260: **Suggestion:** Nothing is announced while the list loads — `Spinner` renders `role="presentation"` with no accessible name, and `aria-busy` on a plain `div` is not announced on its own. Someone using a screen reader hears the title and the Close button, then nothing, and nothing announces the list when it arrives, since `speak()` is only wired to moves. A visually hidden "Loading content…" plus a `speak()` when `items` resolves would close the loop.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+								</Button>
+							</div>
+						</>
+					) }
+				</div>
+			</Modal>
+			{ isConfirmingDiscard && (
+				<ConfirmDialog
+					isOpen
+					contentLabel={ __( 'Discard the new order?', 'newspack-blocks' ) }
```
thomasguillot, line 369: **Suggestion:** The discard dialog's accessible name and its whole body are the same sentence — `ConfirmDialog` forwards `contentLabel` to `Modal`'s `aria-label` and passes `__experimentalHideHeader`, so there is no visible heading and the string is announced as the dialog's name and then read again as the only body text. Past the repetition, it asks a question without stating the consequence. Something like "The order you set will be lost." as the body would remove the echo and tell both sighted and screen reader users what they are choosing between.

plugins/newspack-blocks/src/components/specific-posts-control.js
```diff
@@ -0,0 +1,91 @@
...
+
+	// The token field and the modal ask for the same titles, so the request is
+	// shared rather than made twice, and a failure leaves the slot open to retry.
+	const fetchSavedInfoOnce = useCallback(
+		ids => {
+			const key = [ ...ids ].sort().join( ',' );
+			if ( savedInfo.current.key !== key ) {
+				savedInfo.current = {
+					key,
+					promise: fetchSavedInfo( ids ).catch( error => {
```
thomasguillot, line 33: **Suggestion:** The cache's two documented invariants have no test behind them — Two non-obvious behaviours are asserted in comments and nowhere else: that a rejected fetch frees the slot so the next call retries, and that a late rejection from a superseded key does not evict the entry a newer request owns. `specific-posts-control.test.js` covers only the happy path. The same gap applies to the new `.catch` in `autocomplete-tokenfield.js:58`, where there is no test file at all, and that path decides what gets written back into the block attribute on the next edit. A rejecting stub would pin both cheaply, and they are the paths a publisher on a flaky connection hits.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+		const next = moveItem( items, from, to );
+		if ( next === items ) {
+			return;
+		}
+		pendingFocus.current = { id: items[ from ].id, direction };
+		setItems( next );
+		speak(
+			sprintf(
+				/* translators: 1: new position of the moved item. 2: total number of items. */
+				__( 'Moved to position %1$d of %2$d.', 'newspack-blocks' ),
```
thomasguillot: **Nit:** The move announcement gives the position but not the row — `filterMessage` in `@wordpress/a11y` appends a space when a message repeats, so consecutive identical announcements are re-read; that part is fine. The subject is what is missing. While the pressed chevron stays enabled focus does not move, so nothing re-announces the button's name and the user hears only "Moved to position 2 of 3". Core's block mover names the block alongside the positions, and `item.label` is already to hand via `nameWithTitle`.

plugins/newspack-blocks/src/components/reorder-modal.js
```diff
@@ -0,0 +1,358 @@
...
+									</Card.Root>
+								) ) }
+							</ul>
+							<div className="newspack-blocks-reorder-modal__footer">
+								<Button variant="tertiary" onClick={ requestClose }>
+									{ __( 'Cancel', 'newspack-blocks' ) }
+								</Button>
+								<Button
+									variant="primary"
+									disabled={ ! isDirty || hasFetchError }
```
thomasguillot: **Nit:** The `hasFetchError` term in the Save button's `disabled` cannot be reached — The button renders only inside `{ items && ! hasFetchError && ( … ) }` at `:253`, so `hasFetchError` is always false where this expression evaluates. It reads as a second line of defence but implies a state, error plus a rendered footer, that cannot occur, which makes the error branch harder to reason about than it is.

plugins/newspack-blocks/src/components/reorder-modal.test.js
```diff
@@ -0,0 +1,406 @@
...
+} );
+
+const rows = () => Array.from( document.querySelectorAll( '.newspack-blocks-reorder-modal__item' ) );
+
+describe( 'ReorderModal drag and drop', () => {
+	it( 'reorders as a row is dragged over another', async () => {
+		renderModal();
+		await screen.findByText( 'Alpha' );
+		const dt = dataTransfer();
+		fireEvent.dragStart( rows()[ 2 ], { dataTransfer: dt } );
```
thomasguillot, line 332: **Nit:** Every drag test moves a row upward — All four behavioural drag tests start at `rows()[ 2 ]` and drag over `rows()[ 0 ]`. Downward goes through a different shape of `handleDragOver`: after `moveItem( items, 0, 1 )` the element under the cursor becomes the dragged row itself, and the next `dragover` has to hit the `dragIndex === index` early return rather than swapping back. That self-stabilising step is the part most likely to regress if the handler is touched, and only the `moveItem` unit test covers its index arithmetic. One test dragging `rows()[ 0 ]` across `rows()[ 1 ]` and then `rows()[ 2 ]` would pin it.

thomasguillot (commented):
Release risk: LOW. No blockers, nothing touches a contract surface, and every path I exercised behaved as the description says it would.

I ran this in the block editor against a live site rather than reading it. The order round-trips end to end: reordering by chevron and by drag both write the permutation to `specificPosts`, and the editor preview and the front end follow it. Forty rapid chevron presses left a clean permutation with no dropped or duplicated ID. Save stays disabled until something moves and disables again when a row returns to where it started, and the button only enables at two items, with "Pick at least two items to reorder them." exposed to screen readers. All four close paths hold: Escape, the ✕, a backdrop press and a drag released outside the modal each leave the modal still, with the discard prompt appearing over it and no fade. Keep editing returns to the list with the move intact; Discard leaves the original order. Carousel gets the same control from the same change. The PHP claim holds too: `post__in` with `orderby => post__in` survives the front-end render, the preview endpoint and the deduplication branch, so no PHP change was needed. The `newspack/wizards` warning in the editor console is on main as well and is not from this branch.

The findings cluster on one theme, which is what happens to an ID the REST call does not return. The modal shows it as a `(no title)` row and writes it back on Save, while the Content field beside it drops it without a word, so the two controls disagree about what is selected. That drop is older than this branch, and `fetchSavedPosts` is already in the diff, so the backfill its sibling fetchers already do would close it here. The rest is polish, mostly around what assistive technology hears.

plugins/newspack-blocks/src/components/autocomplete-tokenfield.js
thomasguillot: Added in 16e9a12b8, worded to match the modal's. Verified by rejecting the posts endpoint at `window.fetch`: the field still renders the IDs, and the cause now reaches the console.

The new `autocomplete-tokenfield.test.js` also covers what the fallback writes back, since that is the part with teeth: after a rejection, removing one token still passes the remaining IDs through `onChange`.

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Closed at the source in e638854f5. `fetchSavedPosts` now walks `postIDs` after the map and pushes an entry for any ID the endpoint did not return, labelled `Unavailable content (%d)` so two unresolved IDs cannot collapse onto one label.

Ran the repro with post 8 trashed. The Content field shows five tokens instead of four, the modal row reads `Unavailable content (8)`, and adding a post through the autocomplete keeps the orphan: `["17","5","8","11","14","23"]`.

The modal keeps `(no title)` as a defensive default for a caller whose `fetchItems` skips an ID, but nothing reaches it through this path any more. New `query-controls.test.js` covers the backfill, the two-unavailable-IDs case, and an empty title alongside an unavailable one.

One pre-existing gap I left alone: two picked posts with genuinely empty titles still share `(no title)` and would collapse in `getValuesForLabels`. That predates this branch, and closing it would change label generation for categories, tags and authors too.

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Done in f5af168ae. `findOverlay()` takes whatever node `Modal` forwards and derives the overlay from it: `closest`, then a downward `querySelector`, then an `ownerDocument` query as the last resort. It goes through a stable `useCallback` ref, so the node is not dropped and re-derived on every render.

Unit tested from a descendant, from an ancestor, from the overlay itself, and from `null`.

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Both added in 3ee7a4e94: a `screen-reader-text` span beside the spinner, and a `speak()` when the list resolves, counted through `_n`.

Checked in the editor against a delayed response. The span computes to 1x1 and clipped inside the modal frame, `aria-busy` is `true` while it shows, and the polite region read `5 items ready to reorder.` once the rows arrived.

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Body is now `The order you set will be lost.` in d0dd0a5d5, with the question left as the dialog's accessible name.

In the editor the dialog reports `aria-label="Discard the new order?"` with the consequence as its only body text. The tests find it with `getByRole( 'dialog', { name: ... } )` now and assert the body does not repeat the name.

plugins/newspack-blocks/src/components/specific-posts-control.js
thomasguillot: Both invariants have tests in 48f1b8dc2, and I checked each one bites. Removing the `savedInfo.current.key === key` guard fails `leaves a newer request in place when an older one fails late`; caching the rejection outright fails `asks again rather than sharing the failure`.

The token field's `catch` has its own file now, in 16e9a12b8: the ID fallback, the log, and what a later edit writes back through `onChange`.

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Now `'%1$s moved to position %2$d of %3$d.'` with `item.label`, in 3ee7a4e94. Live, a chevron press announces `Culpa cupidatat enim culpa eius labore proident elit dicta id eu moved to position 2 of 5.`

plugins/newspack-blocks/src/components/reorder-modal.js
thomasguillot: Removed in 0c10c7e19. Confirmed in the editor by failing the fetch: the error branch renders the notice and a single Cancel button, so there is no footer for the term to guard.

plugins/newspack-blocks/src/components/reorder-modal.test.js
thomasguillot: Added in 04a52eb62. One test drags row 0 over row 1 and then over row 2, with an assertion in between that a second `dragover` on the dragged row leaves the order alone.

One thing that turned up while pinning it: the early return is not what makes that step safe. Deleting `if ( dragIndex === index ) { return; }` passes the whole suite, because `moveItem`'s `from === to` guard already covers it. What the new test does catch is a stale `dragIndex`. Dropping `setDragIndex( index )` fails this test and no other.

PR #850, opened 2026-08-11
Newsletters: Add ability to select draft/scheduled/pending posts in Posts Inserter

claudiulodro:
### All Submissions:

Closes NPMIG-2996.

As advertised, this PR adds the ability to search for and select draft/scheduled/pending posts in the Posts Inserter block when using the "Display specific posts" toggle.

It does **not** include draft/scheduled/pending posts in the latests posts mode of the block. That seems like it would have a lot of edge cases, it would make it difficult to do an accurate preview (unless you know when the newsletter will send, you don't know which scheduled posts will actually be live), and it would probably lead to a number of accidental newsletter sends with unpublished stories.

1. Add a Posts Inserter block in a newsletter. The auto-populating Latest X stories mode should only contain published stories.
2. Toggle on "Display specific posts" on the block. Search for a draft post and/or a scheduled post. Observe they show up in results and have a helpful label about their status:

Screenshot 2026-08-11 at 3 00 22 PM

4. Select the post. It should insert and preview nicely in the block:

Screenshot 2026-08-11 at 3 00 38 PM

6. Send a test email. It should include the draft/scheduled posts. If you click through a story, it should have a URL of the type `?p=181` which resolves nicely for staff when the story is unpublished and for readers if the story is published:

Screenshot 2026-08-11 at 3 02 19 PM

### Other information:

### DIFF
--- a/plugins/newspack-newsletters/src/editor/blocks/posts-inserter/consts.js
+++ b/plugins/newspack-newsletters/src/editor/blocks/posts-inserter/consts.js
@@ -1,2 +1,10 @@
 export const POSTS_INSERTER_BLOCK_NAME = 'newspack-newsletters/posts-inserter';
 export const POSTS_INSERTER_STORE_NAME = 'newspack-newsletters/posts-inserter-block';
+
+/**
+ * Post statuses a hand-picked post may have.
+ *
+ * Used both by the "Add posts" search and by the lookup of posts already saved on the
+ * block. Private posts are deliberately left out.
+ */
+export const SEARCHABLE_STATUSES = [ 'publish', 'future', 'draft', 'pending' ];
--- a/plugins/newspack-newsletters/src/editor/blocks/posts-inserter/query-controls.js
+++ b/plugins/newspack-newsletters/src/editor/blocks/posts-inserter/query-controls.js
@@ -16,20 +16,24 @@ import { Fragment, useState, useEffect } from '@wordpress/element';
 import apiFetch from '@wordpress/api-fetch';
 import { decodeEntities } from '@wordpress/html-entities';
 
-const fetchPostSuggestions = postType => search =>
-	apiFetch( {
-		path: addQueryArgs( '/wp/v2/search', {
-			search,
-			per_page: 20,
-			_fields: 'id,title',
-			subtype: postType,
-		} ),
-	} ).then( posts =>
-		posts.map( post => ( {
-			id: post.id,
-			title: decodeEntities( post.title ) || __( '(no title)', 'newspack-newsletters' ),
-		} ) )
-	);
+/**
+ * Internal dependencies
+ */
+import { formatPostLabel, getPostSearchPath, getPostStatusPath, mergePostStatuses } from './post-search';
+
+const fetchPostSuggestions = ( restBase, search ) =>
+	apiFetch( { path: getPostSearchPath( restBase, search ) } )
+		// Core rejects the whole request when the user can't edit this post type. Fall back
+		// to published posts rather than leaving the search empty.
+		.catch( () => apiFetch( { path: getPostSearchPath( restBase, search, false ) } ) )
+		.then( posts =>
+			posts.map( post => ( {
+				id: post.id,
+				title: decodeEntities( post.title?.rendered ) || __( '(no title)', 'newspack-newsletters' ),
+				status: post.status,
+			} ) )
+		)
+		.catch( () => [] );
 
 const SEPARATOR = '--';
 const encodePosts = posts => posts.map( post => [ post.id, post.title ].join( SEPARATOR ) );
@@ -45,9 +49,12 @@ const decodePost = encodedPost => {
 const QueryControlsSettings = ( { attributes, setAttributes } ) => {
 	const [ categoriesList, setCategoriesList ] = useState( [] );
 	const [ postTypesList, setPostTypesList ] = useState( [ { value: 'post', label: 'Posts' } ] );
+	const [ postTypeRestBases, setPostTypeRestBases ] = useState( { post: 'posts' } );
+	const [ postStatuses, setPostStatuses ] = useState( {} );
 	const [ showAdvancedFilters, setShowAdvancedFilters ] = useState( false );
 
 	const { categoryExclusions, tags, tagExclusions } = attributes;
+	const restBase = postTypeRestBases[ attributes.postType ];
 
 	useEffect( () => {
 		apiFetch( {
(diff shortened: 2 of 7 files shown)


plugins/newspack-newsletters/src/editor/blocks/posts-inserter/query-controls.js
```diff
@@ -14,22 +14,34 @@ import { Button, QueryControls, FormTokenField, SelectControl, ToggleControl, Sp
...
-			title: decodeEntities( post.title ) || __( '(no title)', 'newspack-newsletters' ),
-		} ) )
-	);
+/**
+ * Internal dependencies
+ */
+import { formatPostLabel, getPostSearchPath } from './post-search';
+import { selectSpecificPosts } from './specific-posts';
+
+const fetchPostSuggestions = ( restBase, search ) =>
```
dkoo, line 26: **Suggestion:** The security-relevant search fallback is untested — fetchPostSuggestions is the most failure-prone code in the PR and has no coverage: it retries *only* on error?.code === 'rest_forbidden_status' (dropping status → published-only), re-throws everything else, then swallows to []. A regression that widened the retry to any error — or drifted the code string — would fail open (or silently) and pass CI. A small apiFetch-mocking suite asserting (1) a forbidden error triggers exactly one no-status retry and (2) a non-forbidden error resolves to [] without a retry would lock the closed-fallback contract. The pure helpers around it are already well covered.

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/specific-posts.js
```diff
@@ -0,0 +1,45 @@
...
+ * Internal dependencies
+ */
+import { SEARCHABLE_STATUSES } from './consts';
+
+/**
+ * `getEntityRecords` query for the posts hand-picked on a block.
+ *
+ * @param {number[]} ids Post IDs.
+ * @return {Object} Query args.
+ */
```
dkoo, line 11: **Suggestion:** Hand-picked posts are capped at 10 by a missing per_page — getSpecificPostsQuery omits per_page, so getEntityRecords caps hand-picked results at core's default of 10 — select 11+ specific posts and the overflow silently never renders (index.js's reorder just skips the missing ids). This matches the old `{ include: ids }` query so it isn't a regression, but the new 20-result search makes picking many posts easier, so the ceiling is now easier to hit. `per_page: ids.length` (or 100) would close it.

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/consts.js
```diff
@@ -1,2 +1,10 @@
 export const POSTS_INSERTER_BLOCK_NAME = 'newspack-newsletters/posts-inserter';
 export const POSTS_INSERTER_STORE_NAME = 'newspack-newsletters/posts-inserter-block';
+
+/**
+ * Post statuses a hand-picked post may have.
+ *
+ * Used both by the "Add posts" search and by the lookup of posts already saved on the
+ * block. Private posts are deliberately left out.
```
dkoo, line 8: **Nit:** SEARCHABLE_STATUSES is used for more than search — The constant also drives the saved-token status lookup, which isn't a search. The docblock acknowledges the dual use; a name like SELECTABLE_STATUSES / PICKABLE_STATUSES would match both call sites. Minor.

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/specific-posts.js
```diff
@@ -0,0 +1,45 @@
...
+	// collection. Nothing is picked yet, so nothing is what we want.
+	if ( 0 === ids.length ) {
+		return [];
+	}
+
+	const { getEntityRecords, hasResolutionFailed } = select( 'core' );
+	const query = getSpecificPostsQuery( ids );
+
+	if ( hasResolutionFailed( 'getEntityRecords', [ 'postType', postType, query ] ) ) {
+		return getEntityRecords( 'postType', postType, { include: ids } );
```
dkoo, line 41: **Nit:** Make the closed-fallback reliance explicit — The `{ include: ids }` fallback (no status) keeps failing closed only because core's REST default status is `publish`. That's a stable core default and the behavior is correct, but the closed-fallback property is load-bearing and currently implicit — a one-line comment ("no status ⇒ core defaults to publish, the intended closed fallback") would flag it for a future editor tempted to add an explicit status here.

dkoo (approved):
Release risk: LOW. Tested and works as described! No blockers. The search leans on WP core's server-side capability checks (verified: a lower-privileged user sees only published posts plus their own drafts), and every permission fallback shows published-only. The suggestions inline are worth a look — especially the missing `per_page` that caps hand-picked posts at 10, and the untested error/fallback path in the search. One product consideration beyond the code: a newsletter actually *sent* while a hand-picked post is still unpublished will give subscribers dead ?p= links until that post publishes — maybe worth a send-time guard, but that's a follow-up, not for this PR.

- `plugins/newspack-newsletters/src/editor/blocks/posts-inserter/utils.js:32` — **Nit:** Pre-existing: post.link / title.rendered interpolated into HTML unescaped — `<a href="${ post.link }">${ post.title.rendered }</a>` is built without escaping. This predates the PR and title.rendered is server-side sanitized, so it's low-risk — noting only because the input set now includes unpublished, author-controlled draft titles. If this file is touched again, an escaped href / DOM construction would be the safer pattern.

Review provenance: 4 of 4 reviewer passes (lineup: deep, standards, wp-expert, security); release risk: LOW.

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/consts.js
claudiulodro: https://github.com/Automattic/newspack-workspace/pull/850/changes/9d6f69dcd1cfe564155eb8b010e03ba68f9e1563

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/specific-posts.js
claudiulodro: https://github.com/Automattic/newspack-workspace/pull/850/changes/9d6f69dcd1cfe564155eb8b010e03ba68f9e1563

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/specific-posts.js
claudiulodro: https://github.com/Automattic/newspack-workspace/pull/850/changes/7571f536dbee626e63467b57acb14e4eade1cdca

plugins/newspack-newsletters/src/editor/blocks/posts-inserter/query-controls.js
claudiulodro: We'll pass on this one for now since it's pre-existing and out of scope

PR #852, opened 2026-08-12
feat(audience): show the pricing rules impact as a grid of tiles

thomasguillot:
### All Submissions:

The Pricing Rules impact figures rendered as a single strip of value and label pairs inside one card, with "protected" reduced to a note hanging off a neighbouring stat. Four related numbers read as one run-on line, and the least self-explanatory of them had the least room.

They are now a grid of scorecard tiles, each with its own label, figure and description. Four tiles when the pricing engine reports an audience, one when it does not. A locked rule keeps its renewal tiles and shows the null glyph with the reason, rather than dropping the figures and leaving a sentence behind. The "View Affected Products" trigger moved into the tile whose number it explains.

The block is held to the wizard section width and shares a left edge with the rules table. Both the list screen and the rule editor's Impact Preview render it, from one component.

Two things follow from one component serving two screens. The product count means different things on each, so each caller supplies its own line for that tile: the list screen counts what active rules price today, the editor counts what the rule being edited would match. And the tile labels are headings, so both screens now read h1 to h2 to h3 with no level skipped.

The figures themselves are also read more carefully. Counts and prices cross a REST boundary owned by the pricing engine, which returns whatever `json_encode` made of them, so both are normalised before they are compared or formatted. A capped figure is spoken as "At least 500" rather than leaving the meaning to a "+" a screen reader may not announce.

Pricing Rules impact tiles

1. Open Audience Management > Pricing Rules on a site with at least one published rule. A row of tiles sits below the table.
2. Check the tiles line up with the rules table on the left and stop at the wizard section width.
3. Click "View Affected Products" inside the first tile. The modal opens and lists the affected products.
4. Narrow the browser below 1054px. The tiles reflow to two across, then to one below 744px, without any figure overflowing its tile.
5. Open any rule for editing and scroll to Impact Preview. The same tiles appear, spaced from the heading above and the price table below.
6. Set that rule to apply to new sign-ups only. Its last two tiles show an em-dash and "Applies to new sign-ups only".
7. Deactivate WooCommerce Subscriptions. Both screens fall back to a single full-width "Products affected" tile.
8. Trash every rule. The tile reads 0 and the trigger is replaced by "No active pricing rules are affecting products yet."

### Other information:

The tile stays local to the pricing-rules directory in this PR. The shared `StatCard` in `packages/components` (#864) covers the same anatomy, and adopting it is a follow-up under DSGNEWS-212 rather than part of this change.


thomasguillot:
## Review summary

**Release risk: LOW.** No blockers. No contract surface is touched, and there is no consumer outside this plugin's own view directory.

Two findings, both the same shape: hardening that stopped one function short.

### 1. Prices cross the same boundary as the counts but still call `.toFixed()` unguarded

`impact-format.ts:36`

`formatPrice` takes `amount: number` and calls `amount.toFixed( currency.decimals )` directly, two functions below the `finiteCount` doc block that explains why nothing from this engine can be assumed to be a number. `regular`, `adjusted` and `segments[].amount` arrive in the same payload, from the same plugin and the same `$wpdb` reads as the counts this branch normalises. They are typed `number`, so the compiler will not flag a string arriving.

**Risk if not addressed:** if a price arrives as a string it throws during render. There is no error boundary anywhere in the plugin's `src`, and the wizard mounts a single root (`src/wizards/index.tsx:107`), so the throw unmounts the whole admin page to a blank screen with no message. In the rule editor the preview sits inside the form (`rule-form.tsx:834`), so the same throw would also discard the in-progress rule the publisher had not yet saved.

**Consider:** routing `amount` through the existing normaliser and falling back to the em-dash treatment, so an unexpected price shape costs one cell rather than the page. The function has six call sites across three modules: `impact-table.tsx` formats engine payload values, while `model-sentence.ts` and `schedule-format.ts` format local form values. Guarding inside `formatPrice` covers the first without changing behaviour for the other two when the data is well formed.

### 2. The other `total_matching` comparison still reads the raw value

`rule-preview.tsx:89`

This branch hardened `catalog-impact.tsx` to `0 === finiteCount( stats.total_matching )`, and widened `ImpactStatsProps.totalMatching` to match. `rule-preview.tsx` still does `data.total_matching === 0` against the same field on the same response shape, so a `'0'` would slip past it.

The practical effect is nil: the `|| ! data.sample?.length` clause beside it catches every payload a string zero could describe, and a `'0'` alongside a non-empty sample is self-contradictory data with no sane rendering either way. Worth folding in for the consistency of having one way to read an engine count, not because a user can reach the difference.

### Also worth a look

The em-dash carries two meanings on the same tile shape. Under a locked rule it is paired with a visible reason line ("Applies to new sign-ups only"). When a count is unparseable, the explanatory word goes into screen-reader-only text while the glyph is hidden, so a sighted publisher sees only the dash in both cases.

That matters because the two states say opposite things: one means the figure does not apply, the other means it never arrived. Giving the missing-count tile a visible line, the way the locked tiles already carry one, would separate them.

thomasguillot:
## Final pass

Both findings addressed in `dbd450036`.

**1. Prices now cross the boundary through the same guard as the counts.** `formatPrice` normalises its amount and returns the em-dash when it is not a finite number, so an unexpected shape costs one cell instead of unmounting the page. Zero still formats as a price rather than a dash, which the table confirms (`$0.00` renders for a free product).

The helper is renamed `finiteCount` to `finiteNumber`, since it now serves prices as well as counts, and the em-dash glyph moved into `impact-format.ts` so the tile and the price formatter share one constant instead of two copies.

**2. `rule-preview.tsx` reads `total_matching` through that helper**, so both places that ask whether the catalogue is empty now read an engine count the same way.

`impact-format.test.js` gained direct coverage for the price guard alongside the existing normaliser and caption cases. The suite is at 91 files and 892 tests, up from 90 and 868 when this branch opened.

### Verified in a local environment

Both surfaces, against the built assets:

- **List screen**: heading outline reads h1 → h2 ("Catalog impact", screen-reader only) → four h3 tile labels, with no level skipped. The grid resolves to four columns, capped figures render as `2,000+` with "At least 2,000" as their spoken name, and the "View Affected Products" trigger sits in the first tile only.
- **Products modal**: opens from that trigger, lists 33 rows, captions itself "Showing a sample of 33 products.", and prices render across the cycle columns.
- **Rule editor**: Impact Preview reads h1 → h2 → four h3 tiles, and the product tile carries its own line, "This rule would price these products", rather than the list screen's.

The locked-rule and missing-count states are covered by tests rather than by this pass, since both need a payload the environment does not produce.

miguelpeixe (approved):
:chef-kiss:

PR #853, opened 2026-08-12
fix(newsletters): hide abandoned auto-drafts from lists (DSGNEWS-213)

thomasguillot:
### All Submissions:

"Add Newsletter" links to `post-new.php`, and WordPress inserts an auto-draft as soon as that page loads, before the layout picker opens. Leaving the picker without choosing a layout left that row in the Newsletters list as a `(no subject)` entry, because the list queried `auto-draft` alongside the writable statuses.

All three admin-shell lists (Newsletters, Newsletter Ads, Layouts) now exclude `auto-draft`, matching core's own Posts and Pages lists. Nothing is deleted here; core's daily cron force-deletes auto-drafts after 7 days.

**Any save promotes an auto-draft to `draft`** (core-data's `prePersistPostType` forces the status and clears the placeholder title), so a row still sitting in `auto-draft` has never been saved and its stored content is always empty. There is no in-progress newsletter to lose.

Two coupled pieces move with it. The REST draft buckets re-widened `auto-draft` back into `post_status`, so changing the JS alone would have done nothing. And advertiser term counts deliberately counted auto-draft ads to match the Ads draft bucket, so they drop it too, with the recount sentinel bumped so existing sites refresh.

The newsletters REST endpoint drops `auto-draft` from a `status` request rather than answering it. Asked for `auto-draft` and nothing else, it now returns an empty collection instead of the saved drafts a substituted status would have produced, and the selection always drives `post_status`, so the pre-deploy default set a cached bundle still sends no longer carries `auto-draft` through.

Closes DSGNEWS-213.

1. Note the count in the "All Newsletters" heading.
2. Go to Newsletters > Add New. The layout picker overlay opens.
3. Press the browser back button without choosing a layout.
4. Go to Newsletters > All Newsletters. No new `(no subject)` row appears and the count is unchanged.
5. Open the Status filter and choose Draft. Saved drafts appear; no auto-drafts do.
6. Save a newsletter with no subject. It still lists as `(no subject)`.
7. Repeat steps 2 to 4 for Newsletters > Advertising, then for Newsletters > Layouts.
8. Open Newsletters > Advertising and check the Advertiser filter. Each count matches the ads listed for that advertiser.
9. On a site with no newsletters, mis-click Add New, then return to the list. The empty state still shows.
10. Request `/wp-json/wp/v2/newspack_nl_cpt?status=auto-draft` as an administrator. The collection comes back empty.

### Other information:

* No deletion path ships here. Existing stray rows stay in the database, invisible, until core's 7-day cron clears them.
* The advertiser recount is a one-time pass per site, on the first admin request after this deploys by a user who can manage advertisers. It is belt-and-braces: advertiser terms are only written on save, and every save promotes an ad out of `auto-draft`, so no stored count is expected to move. Deleting the `newspack_nl_advertiser_count_recounted_v4` option re-runs it, which is the one-step repair if a site's counts ever look stale.

### DIFF
--- a/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
+++ b/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
@@ -118,6 +118,14 @@ class Newsletters_List_REST {
 	 */
 	public static function align_status_filter_with_scheduled_meta( $args, $request ) {
 		$values = self::parse_status_values( $request->get_param( 'status' ) );
+		// A legacy `auto-draft` selection resolves to the Draft bucket, as the
+		// list's own deep links do. Left as-is it would match no bucket, and the
+		// query would run unfiltered against `post_status = auto-draft`.
+		if ( in_array( 'auto-draft', $values, true ) ) {
+			$values   = array_diff( $values, [ 'auto-draft' ] );
+			$values[] = 'draft';
+			$values   = array_values( array_unique( $values ) );
+		}
 		if ( empty( $values ) ) {
 			return $args;
 		}
@@ -132,9 +140,6 @@ class Newsletters_List_REST {
 		if ( $wants_sent || $wants_draft || $wants_scheduled ) {
 			$widened = array_merge( $widened, [ 'publish', 'private' ] );
 		}
-		// `auto-draft` is deliberately absent: an abandoned "Add new" is empty
-		// by construction (any save promotes it to `draft`), and core's own
-		// lists hide it too.
 		if ( $wants_draft || $wants_scheduled ) {
 			$widened = array_merge( $widened, [ 'draft', 'pending' ] );
 		}
--- a/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
+++ b/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
@@ -123,7 +123,7 @@ class Newsletters_List_REST {
 		}
 
 		$wants_sent      = ! empty( array_intersect( $values, [ 'publish', 'private' ] ) );
-		$wants_draft     = ! empty( array_intersect( $values, [ 'draft', 'pending', 'auto-draft' ] ) );
+		$wants_draft     = ! empty( array_intersect( $values, [ 'draft', 'pending' ] ) );
 		$wants_scheduled = in_array( 'future', $values, true );
 		$wants_trash     = in_array( 'trash', $values, true );
 
@@ -132,8 +132,11 @@ class Newsletters_List_REST {
 		if ( $wants_sent || $wants_draft || $wants_scheduled ) {
 			$widened = array_merge( $widened, [ 'publish', 'private' ] );
 		}
+		// `auto-draft` is deliberately absent: an abandoned "Add new" is empty
+		// by construction (any save promotes it to `draft`), and core's own
+		// lists hide it too.
 		if ( $wants_draft || $wants_scheduled ) {
-			$widened = array_merge( $widened, [ 'draft', 'pending', 'auto-draft' ] );
+			$widened = array_merge( $widened, [ 'draft', 'pending' ] );
 		}
 		if ( $wants_scheduled ) {
 			$widened[] = 'future';
@@ -168,11 +171,10 @@ class Newsletters_List_REST {
 		}
 		if ( $wants_draft ) {
 			$bucket_clauses[] = $wpdb->prepare(
-				"( NOT EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = %s AND meta_value <> '' ) AND ( {$wpdb->posts}.post_status IN (%s, %s, %s) OR ( {$wpdb->posts}.post_status IN (%s, %s) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = %s AND meta_value <> '' ) ) ) )",
+				"( NOT EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = %s AND meta_value <> '' ) AND ( {$wpdb->posts}.post_status IN (%s, %s) OR ( {$wpdb->posts}.post_status IN (%s, %s) AND EXISTS ( SELECT 1 FROM {$wpdb->postmeta} WHERE post_id = {$wpdb->posts}.ID AND meta_key = %s AND meta_value <> '' ) ) ) )",
 				'sending_scheduled',
 				'draft',
 				'pending',
-				'auto-draft',
 				'publish',
 				'private',
 				'scheduling_error'
--- a/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
+++ b/plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
@@ -118,16 +118,15 @@ class Newsletters_List_REST {
 	 */
 	public static function align_status_filter_with_scheduled_meta( $args, $request ) {
 		$values = self::parse_status_values( $request->get_param( 'status' ) );
-		// A legacy `auto-draft` selection resolves to the Draft bucket, as the
-		// list's own deep links do. Left as-is it would match no bucket, and the
-		// query would run unfiltered against `post_status = auto-draft`.
-		if ( in_array( 'auto-draft', $values, true ) ) {
-			$values   = array_diff( $values, [ 'auto-draft' ] );
-			$values[] = 'draft';
-			$values   = array_values( array_unique( $values ) );
-		}
 		if ( empty( $values ) ) {
 			return $args;
+		}
+		$values = array_values( array_diff( $values, [ 'auto-draft' ] ) );
+		if ( empty( $values ) ) {
+			// Asked for auto-drafts and nothing else, so answer with nothing.
+			// Returning here instead would leave the posts controller's
+			// `post_status = auto-draft` running with no bucket clause.
+			return self::install_bucket_filter( $args, [ '1 = 0' ], '_newspack_nl_bucket_token' );
 		}
 
 		$wants_sent      = ! empty( array_intersect( $values, [ 'publish', 'private' ] ) );
@@ -146,10 +145,7 @@ class Newsletters_List_REST {
 		if ( $wants_scheduled ) {
 			$widened[] = 'future';
 		}
-		$widened = array_values( array_unique( $widened ) );
-		if ( $widened !== $values ) {
-			$args['post_status'] = $widened;
-		}
+		$args['post_status'] = array_values( array_unique( $widened ) );
 
 		global $wpdb;
 		$bucket_clauses = [];
--- a/plugins/newspack-newsletters/includes/ads/class-ads.php
+++ b/plugins/newspack-newsletters/includes/ads/class-ads.php
@@ -387,8 +387,6 @@ final class Ads {
 		global $wpdb;
 
 		$taxonomy_name = is_object( $taxonomy ) ? $taxonomy->name : $taxonomy;
-		// Include auto-draft to match the Ads list's draft bucket.
-		$statuses = [ 'publish', 'private', 'future', 'draft', 'pending', 'auto-draft' ];
 
 		foreach ( $terms as $tt_id ) {
 			$tt_id = (int) $tt_id;
@@ -398,15 +396,17 @@ final class Ads {
 					INNER JOIN {$wpdb->posts} p ON p.ID = tr.object_id
 					WHERE tr.term_taxonomy_id = %d
 					AND p.post_type = %s
-					AND p.post_status IN ( %s, %s, %s, %s, %s, %s )",
+					AND p.post_status IN ( %s, %s, %s, %s, %s )",
 					$tt_id,
 					self::CPT,
-					$statuses[0],
-					$statuses[1],
-					$statuses[2],
-					$statuses[3],
-					$statuses[4],
-					$statuses[5]
+					// `auto-draft` is absent to match the Ads list's draft
+					// bucket, which hides an abandoned "Add new" the way
+					// core's own lists do.
+					'publish',
+					'private',
+					'future',
+					'draft',
+					'pending'
 				)
 			);
 
@@ -420,10 +420,10 @@ final class Ads {
 	 * One-time recount of existing advertiser terms so counts predating
 	 * the custom count callback refresh without waiting for the next edit.
 	 * Bump the sentinel whenever the counted statuses change so already
-	 * recounted sites pick up the new semantics (v3: added auto-draft).
+	 * recounted sites pick up the new semantics (v4: dropped auto-draft).
 	 */
 	public static function maybe_recount_advertiser_terms() {
-		$option = 'newspack_nl_advertiser_count_recounted_v3';
+		$option = 'newspack_nl_advertiser_count_recounted_v4';
 		if ( get_option( $option ) ) {
 			return;
 		}
--- a/plugins/newspack-newsletters/includes/ads/class-ads.php
+++ b/plugins/newspack-newsletters/includes/ads/class-ads.php
@@ -26,6 +26,16 @@ final class Ads {
 	const ADVERTISER_TAX = 'newspack_nl_advertiser';
 
 	/**
+	 * Statuses an advertiser's term count covers. Must stay equal to the union
+	 * of the non-trash `post_status` sets in `Ads_List_REST::filter_rest_query`,
+	 * or the count disagrees with the list it describes. Bump the recount
+	 * sentinel in `maybe_recount_advertiser_terms` whenever this changes.
+	 *
+	 * @var string[]
+	 */
+	const COUNTED_STATUSES = [ 'publish', 'private', 'future', 'draft', 'pending' ];
+
+	/**
 	 * Ads already inserted in the newsletter.
 	 *
 	 * @var array[] Ad ids mapped by newsletter id.
@@ -399,14 +409,11 @@ final class Ads {
 					AND p.post_status IN ( %s, %s, %s, %s, %s )",
 					$tt_id,
 					self::CPT,
-					// `auto-draft` is absent to match the Ads list's draft
-					// bucket, which hides an abandoned "Add new" the way
-					// core's own lists do.
-					'publish',
-					'private',
-					'future',
-					'draft',
-					'pending'
+					self::COUNTED_STATUSES[0],
+					self::COUNTED_STATUSES[1],
+					self::COUNTED_STATUSES[2],
+					self::COUNTED_STATUSES[3],
+					self::COUNTED_STATUSES[4]
 				)
 			);
 
@@ -419,14 +426,22 @@ final class Ads {
 	/**
 	 * One-time recount of existing advertiser terms so counts predating
 	 * the custom count callback refresh without waiting for the next edit.
-	 * Bump the sentinel whenever the counted statuses change so already
-	 * recounted sites pick up the new semantics (v4: dropped auto-draft).
 	 */
 	public static function maybe_recount_advertiser_terms() {
 		$option = 'newspack_nl_advertiser_count_recounted_v4';
 		if ( get_option( $option ) ) {
 			return;
 		}
+		// `admin_init` also fires on admin-ajax.php, including for logged-out
+		// requests, so gate on the taxonomy's own management capability.
+		if ( ! current_user_can( 'manage_categories' ) ) {
+			return;
+		}
+		// Claim before the work. The recount is one `COUNT(*)` plus one `UPDATE`
+		// per term, so a concurrent burst would otherwise run it once each, and a
+		// pass that times out would re-arm it on every later request.
+		update_option( $option, 1 );
+		delete_option( 'newspack_nl_advertiser_count_recounted_v3' );
 
 		$term_ids = get_terms(
 			[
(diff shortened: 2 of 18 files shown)


plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/initial-filters.test.js
```diff
@@ -11,7 +11,7 @@ describe( 'getInitialFilters', () => {
 	} );
 
 	it( 'maps post_status=draft, pending, and auto-draft to the combined draft filter', () => {
```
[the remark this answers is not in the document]
thomasguillot: Addressed in 145ccd63. The title now calls out `auto-draft` as a legacy input and names the resulting `draft,pending` value, so it no longer reads as if `auto-draft` were part of the filter.

thomasguillot:
## Code review

**Release risk: LOW.** Approve with comments.

The part I most wanted to put under test was the three-way coupling, and it held. The JS status sets, both REST draft buckets, and the advertiser count query all move together, so the claim in the description that changing the JS alone would have been a no-op is correct: `align_status_filter_with_scheduled_meta` re-widened `auto-draft` straight back into `post_status`.

Verified on a local site with four pre-existing abandoned auto-drafts. Before the change, the All Newsletters heading read 117 with four `(no subject)` rows at the top. After it, the same page held at 113 through Add Newsletter, browser back, All Newsletters, with no new row, while the auto-drafts confirmably still landed in the database. JS suite 250/250, PHP suite 734/734, PHPCS and ESLint clean.

Concerns ruled out:

- **The paths that could have regressed did not.** `_update_term_count_on_transition_post_status` keeps advertiser counts accurate when an abandoned auto-draft is later saved or cleared by core's `wp_delete_auto_drafts`. The newsletters and layouts duplicate and restore actions create `draft` or `publish` rows rather than auto-drafts. Legacy `?post_status=auto-draft` deep links still resolve onto the Draft chip in both `initial-filters.js` files.
- **No empty `IN ()` or unbounded query.** `$post_status_set` in `filter_rest_query` is guarded by the early return on empty `$kinds`, and `$widened` always starts from a non-empty `$values`.
- **No cross-plugin surface.** All 14 files are internal to `newspack-newsletters`; no shared contract paths changed and no consumer references exist elsewhere in the workspace.
- **Nothing on the security side.** The `$wpdb->prepare()` calls are correctly parameterised and the only interpolations are `$wpdb` table properties. No capability, nonce, escaping or i18n regression in the diff.

Seven findings, none blocking.

---

### Suggestions

**Severity: suggestion** · `plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php:126`

**A status request of only `auto-draft` now installs no bucket filter at all.** Narrowing `$wants_draft` to `array_intersect( $values, [ 'draft', 'pending' ] )` means a request whose only status value is `auto-draft` leaves all four `wants_*` flags false. `$bucket_clauses` then comes out empty and `install_bucket_filter` returns `$args` untouched, so the request runs with the posts controller's `post_status = ['auto-draft']` and no `posts_where` at all. That is the one remaining path through this endpoint that returns abandoned rows, and it now returns them completely unbucketed, where the draft clause previously applied. It is not reachable from the shipped UI, since `initial-filters.js:17` rewrites the deep link to `draft,pending`, and it needs `edit_posts`, so this reads as an invariant gap rather than a leak. But `auto-draft` stays a valid value in the core `status` enum for anyone holding that cap, so a bookmarked REST URL or a future in-plugin caller walks straight through it. I think we should strip `auto-draft` inside `parse_status_values`, or treat it as an explicit no-match, which closes it without touching any other branch.

**Severity: suggestion** · `plugins/newspack-newsletters/includes/ads/class-ads.php:427`

**The re-armed recount is an uncapped per-term loop whose sentinel is written last.** The `_v3` to `_v4` bump makes every existing site re-run `maybe_recount_advertiser_terms` on its next `admin_init`. `get_terms()` takes no `number` cap, and `wp_update_term_count_now()` routes into `update_advertiser_term_count`, which issues one `SELECT COUNT(*)`, one `UPDATE`, and two `do_action` calls per term. `update_option( $option, 1 )` only runs after the whole loop completes, so nothing claims the work up front. The `update_option` return value is also discarded, so a failed write re-arms the full recount on every later admin request with no backoff. In practice the payload is probably empty, since assigning an advertiser term requires a save, which promotes the ad out of `auto-draft`, but the failure mode does not depend on a large payload to be worth closing.

**Risk if not addressed:** if a site with a large advertiser taxonomy takes a burst of traffic to `admin-ajax.php` during the upgrade window, every one of those requests runs the recount too, because `admin-ajax.php` fires the same hook before it checks who is calling and the recount has no capability check of its own. On such a site that means an uncapped term query plus one count and one update per advertiser, repeated per concurrent request, showing up as a database load spike right after the update rather than as wrong data.

**Consider:** claiming the sentinel before the loop instead of after it, and skipping the work for requests that cannot manage the taxonomy, so a burst of concurrent requests during the upgrade window does the recount once rather than once each.

**Severity: suggestion** · `plugins/newspack-newsletters/includes/ads/class-ads.php:399`

**The counted status set is hand-duplicated in three places.** `post_status IN ( %s, %s, %s, %s, %s )` hardcodes five placeholders against five literal arguments, and that set has to stay in lockstep with the Draft bucket in `Ads_List_REST::filter_rest_query` and `DEFAULT_STATUSES` in `ads-list/build-query.js`. This PR is the proof of the coupling: it had to edit all three and bump a sentinel. Nothing fails loudly when one is missed. The term count quietly disagrees with the list it is meant to describe, and a publisher meets that as a wrong number in the Advertiser filter.

**Consider:** deriving the counted status set from a single shared constant used by the count query, the REST draft bucket, and the JavaScript default. As written, a later change to one of the three leaves the advertiser counts disagreeing with the Ads list, and because counts are only recomputed when a term is edited, that disagreement persists silently until someone bumps the sentinel again.

### Nits

**Severity: nit** · `plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/fields.js:46`

**Stale `auto-draft` title branches and comments across the three list screens.** `fields.js:47` still branches on `'auto-draft' === item?.status` to render the `(no subject)` placeholder, as do `ads-list/fields.js:53` and `layouts-list/fields.js:121`, and `layouts-list/actions.js:23` still carries a comment explaining an auto-draft case. None of these can fire now that the lists never query the status. The `! raw ||` half of each condition still earns its keep for a saved draft with an empty subject, so the guards themselves are harmless. The comments are the problem: "New newsletters carry WordPress's 'Auto Draft' placeholder title" now describes a state the list cannot reach, and will mislead the next reader into thinking auto-drafts still land here.

**Severity: nit** · `plugins/newspack-newsletters/tests/test-newsletters-list-rest.php:666`

**The query half of `test_default_status_set_excludes_auto_draft` cannot fail.** `array_merge( $args, [ 'post_status' => $defaults ] )` overwrites whatever the widening produced with the literal five-status array, which excludes `auto-draft` by construction, so both assertions that follow are guaranteed and would pass identically on `main`. The discriminating assertion is the `assertNotContains` on line 662: on `main` the widening injected `auto-draft` into `$args['post_status']`, so that line does fail on a revert. It is worth keeping, and worth a comment saying so, because on the fixed code it reads like a guard on a key the widening deliberately never writes. Driving the query from `$args` rather than overriding it would make the second half earn its keep too. Separately, `$defaults` on line 654 is a hand-copy of `DEFAULT_STATUSES` in `build-query.js:16` with nothing tying the two together.

**Severity: nit** · `plugins/newspack-newsletters/src/admin-shell/screens/layouts-list/use-layouts-data.js:11`

**Layouts is the only changed list with nothing locking the new behaviour.** Newsletters and Ads each got a JS assertion that the default set does not contain `auto-draft`, plus PHP integration coverage. The Layouts change has neither, and the module exports only the `useLayoutsData` hook, so as written there is no seam to test it through. A revert of just this line would go unnoticed by CI. Exporting `buildPath` or the constant and asserting on it would match what the two sibling screens already do.

**Severity: nit** · `plugins/newspack-newsletters/includes/ads/class-ads.php:426`

**The superseded `_v3` sentinel is left orphaned in the options table.** The v3 row stays behind and stays autoloaded, so every bump of the status semantics adds one more permanently autoloaded option that is read on every front-end request and never again means anything. A `delete_option( 'newspack_nl_advertiser_count_recounted_v3' )` alongside the v4 write keeps the mechanism from accumulating cruft each time it is versioned.

---

### Spec alignment

DSGNEWS-213 asked for the abandoned auto-draft to be discarded on abandonment, and cautioned against changing what the list displays because "other in-progress auto-drafts still need to show". This PR takes the opposite route and says why: any save promotes an auto-draft to `draft` via core-data's `prePersistPostType`, so a row still sitting in `auto-draft` has never been saved and holds nothing to lose. That premise is sound and the reported symptom is resolved, so the divergence looks like the better call rather than a gap. The ticket posed its direction as a question, so it is worth a nod from the author before merge. DSGNEWS-214, where one Add Newsletter click creates two auto-drafts, stays open and is untouched here.

thomasguillot:
## Final pass

All seven findings are addressed in `e86b2e30` and `6eccefd5`, and the code comments this PR introduced are cut back to the ones carrying a reason.

**A status request of only `auto-draft` now resolves to the Draft bucket.** `align_status_filter_with_scheduled_meta` rewrites the value to `draft` before the bucket flags are computed, so the request takes the same route as the list's own legacy deep links instead of falling through to an unfiltered `post_status = auto-draft` query. Covered by `test_auto_draft_status_request_resolves_to_draft_bucket`.

**The recount claims its sentinel before doing the work,** and skips requests that cannot manage the taxonomy. `admin_init` fires on `admin-ajax.php` for logged-out requests too, so the capability gate is what keeps those out; claiming first is what stops a concurrent burst running the loop once each, and stops a pass that times out re-arming itself on every later request. Two tests cover it: one asserts the sentinel is already set by the time `get_terms` runs, one asserts a capability-less request does nothing.

**The counted status set has a single source.** `Ads::COUNTED_STATUSES` now feeds the count query, and `test_counted_statuses_match_the_list_buckets` asserts it equals the union of the non-trash `post_status` sets that `Ads_List_REST::filter_rest_query` produces. If either side moves alone the test fails, which is the loud failure that was missing. The placeholders stay written out rather than generated: building them with `array_fill` trips `PreparedSQLPlaceholders.ReplacementsWrongNumber`, and PHPCS warnings block the commit hook.

**The superseded `_v3` option is deleted** when v4 is claimed, so versioning the sentinel stops leaving an autoloaded row behind each time.

**The unreachable `auto-draft` title branches are gone** from all three list screens, leaving `raw || __( '(no subject)' )`, which still covers a saved draft with an empty subject. The stale rationale in `layouts-list/actions.js` is reworded to the reason that still holds, that `??` would keep an empty `title.raw` and produce "Copy of ".

**Layouts has a test seam.** `buildPath` is exported and `use-layouts-data.test.js` asserts the default status set, so a revert of that one line now fails CI like its two siblings.

**`test_default_status_set_excludes_auto_draft` is discriminating end to end.** It seeds `post_status` the way the posts controller does and drives the query from the returned `$args` rather than overwriting them, so both halves fail on a revert instead of only the first.

On the comments: the same "an abandoned Add new is empty by construction" note had been repeated in five files. It now appears once, next to the newsletters `DEFAULT_STATUSES` that defines the behaviour. What remains explains a reason that is not visible in the code: why the capability gate exists, why the sentinel is claimed early, why `??` is wrong in `copyTitle`, why the `auto-draft` key stays in the deep-link map. Comments that restated the code or a test name are gone.

Verification: JS 253/253 across 30 suites, PHP 738/738 with 3042 assertions, `composer phpcs` and `lint:js` clean.

plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
```diff
@@ -118,12 +118,20 @@ public static function filter_rest_query( $args, $request ) {
 	 */
 	public static function align_status_filter_with_scheduled_meta( $args, $request ) {
 		$values = self::parse_status_values( $request->get_param( 'status' ) );
+		// A legacy `auto-draft` selection resolves to the Draft bucket, as the
+		// list's own deep links do. Left as-is it would match no bucket, and the
+		// query would run unfiltered against `post_status = auto-draft`.
+		if ( in_array( 'auto-draft', $values, true ) ) {
```
dkoo: **Severity:** suggestion

**Suggestion:** Rewriting `auto-draft` to `draft` answers a different question than the caller asked — This filter runs on `rest_newspack_nl_cpt_query` for every request to the collection, not only the admin shell. The shell never sends `status=auto-draft` (`initial-filters.js` maps the legacy deep link to `draft,pending` before any request), so what this branch changes is the answer given to any other authenticated client asking for auto-drafts: in the env, `?status=auto-draft` returns the two saved drafts. That is a wrong result set rather than an empty one. Dropping the value so the request resolves to nothing would still close the unbucketed fall-through the earlier round found, and it keeps the endpoint honest to the caller. I found no in-repo client that sends it either way, so this is about API semantics, not a live break. WDYT?

plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
```diff
@@ -133,7 +141,7 @@ public static function align_status_filter_with_scheduled_meta( $args, $request
 			$widened = array_merge( $widened, [ 'publish', 'private' ] );
 		}
 		if ( $wants_draft || $wants_scheduled ) {
-			$widened = array_merge( $widened, [ 'draft', 'pending', 'auto-draft' ] );
+			$widened = array_merge( $widened, [ 'draft', 'pending' ] );
```
dkoo, line 143: **Severity:** nit

**Nit:** The `$widened !== $values` guard leaves `auto-draft` in `post_status` for the pre-deploy default set — For the old default set (`publish,private,future,draft,pending,auto-draft`), which a browser with the previous bundle cached still sends, the rewrite and the widening produce the same array, so the guard skips the assignment and core's `post_status` keeps `auto-draft`. In the env that request returns no auto-draft rows, but only because none of the bucket clauses can match one; the protection is incidental rather than designed. Assigning `$widened` unconditionally removes the coupling in one line.

plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -25,6 +25,16 @@ final class Ads {
...
 
+	/**
+	 * Statuses an advertiser's term count covers. Must stay equal to the union
+	 * of the non-trash `post_status` sets in `Ads_List_REST::filter_rest_query`,
+	 * or the count disagrees with the list it describes. Bump the recount
+	 * sentinel in `maybe_recount_advertiser_terms` whenever this changes.
+	 *
+	 * @var string[]
+	 */
+	const COUNTED_STATUSES = [ 'publish', 'private', 'future', 'draft', 'pending' ];
```
dkoo, line 36: **Severity:** nit

**Nit:** The `_v4` bump re-runs a recount the PR's own premise says cannot change a count — If a row still at `auto-draft` has never been saved, it carries no advertiser terms either, because terms are only written on save and every save promotes the status. I could not build a runtime path that tags an auto-draft ad (the new tests create that state by hand with `wp_set_object_terms`), so dropping `auto-draft` from `COUNTED_STATUSES` leaves every stored count as it was, and the bump buys one `COUNT(*)` plus one `UPDATE` per advertiser on each site's first admin load. Keeping it as belt-and-braces is reasonable; a sentence in the PR body saying so would spare the next reader the same trace.

**Risk if not addressed:** On a site with a large advertiser taxonomy, the pass is one count query plus one update per advertiser inside a single admin page load. If that request is cut short (PHP timeout, memory limit, a fatal from another plugin on the same hook), the flag is already written, so the recount never runs again and the advertisers it did not reach keep counts that still include abandoned auto-draft ads. There is no retry and no signal that it half-finished; each stale advertiser only corrects when one of its ads is next saved. The damage stops at the count column: advertiser pickers and filters fetch terms without hiding empty ones, so nothing drops out of a picker.

**Consider:** Because the flag is written before the recount rather than after, an interrupted pass cannot re-arm itself. Either process the advertiser terms in bounded batches with a stored cursor, or note in the PR body that deleting the `newspack_nl_advertiser_count_recounted_v4` option re-runs the recount, which gives support a one-step repair if a site's counts ever look stale.

plugins/newspack-newsletters/includes/ads/class-ads.php
```diff
@@ -419,14 +426,22 @@ public static function update_advertiser_term_count( $terms, $taxonomy ) {
...
 	 */
 	public static function maybe_recount_advertiser_terms() {
-		$option = 'newspack_nl_advertiser_count_recounted_v3';
+		$option = 'newspack_nl_advertiser_count_recounted_v4';
 		if ( get_option( $option ) ) {
 			return;
 		}
+		// `admin_init` also fires on admin-ajax.php, including for logged-out
+		// requests, so gate on the taxonomy's own management capability.
+		if ( ! current_user_can( 'manage_categories' ) ) {
```
dkoo: **Severity:** nit

**Nit:** Capability gate hard-codes `manage_categories` — The comment says "the taxonomy's own management capability" while the code checks the literal. They agree today because `register_taxonomy` passes no `capabilities` map, so `get_taxonomy( self::ADVERTISER_TAX )->cap->manage_terms` would keep the two in sync if a cap map is ever added. It is also narrower than the Ads screen's `edit_posts` gate, so on a site where only Authors work in newsletters the counts wait for an Editor or Admin page load. The PR body's "first `admin_init` after deploy" could say "by a user who can manage advertisers" to match.

dkoo (approved):
@thomasguillot I tested this in-browser and all testing steps behave as described.

**Release risk: LOW**—no blockers, no contract paths touched, browser-verified end to end.

The three-way coupling is what I most wanted to see hold, and it did: every list, both REST buckets, and the count query move together, and a fresh auto-draft landed in the database on each Add New without ever reaching a list. The recount behaved on a site staged as pre-upgrade: claimed `_v4`, dropped `_v3`, corrected a stale count on the first admin load. The legacy default set (with `auto-draft`) also returns nothing it shouldn't.

One design question and a few nits below. The suggestion is about what a raw REST `status=auto-draft` request should answer, where I read the tradeoff differently than the follow-up commit did; the rest are small.

plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
thomasguillot: Agreed, and changed in ce93a7277. `auto-draft` is now dropped from the parsed values rather than rewritten, so nothing else in the selection is disturbed, and a request whose only value was `auto-draft` installs a never-matching bucket clause instead of returning early. Returning early would have handed the request straight back to the posts controller's `post_status = auto-draft` with no clause at all, which is the fall-through we closed last round.

On the env, `?status=auto-draft` went from 100 rows of saved drafts to 0. `?status=auto-draft,draft` returns the drafts and no auto-drafts, so a mixed request still gets the half it asked for that we can answer.

Two tests: `test_auto_draft_only_status_request_matches_nothing` and `test_auto_draft_is_dropped_from_a_mixed_status_request`.

plugins/newspack-newsletters/includes/admin/class-newsletters-list-rest.php
thomasguillot: Guard gone in ce93a7277: `$args['post_status']` is now assigned unconditionally.

Confirmed on the env, sending the pre-deploy default set. Before, the filter returned `["publish","private","future","draft","pending","auto-draft"]`; now it returns `["publish","private","future","draft","pending"]`. `test_legacy_default_status_set_drops_auto_draft` covers that exact set, and the widening test's second half now asserts the incoming value is replaced rather than preserved.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Your trace matches mine, so the bump stays as belt-and-braces and the PR body now says so, along with the repair path: deleting `newspack_nl_advertiser_count_recounted_v4` re-runs the pass.

I left the batching out. With no runtime path that tags an auto-draft ad, the payload is a count and an update per advertiser over a taxonomy that stays small on real sites, and a cursor would add state that never gets exercised. The option delete covers the interrupted-pass case you describe, which is the one that needed an answer.

plugins/newspack-newsletters/includes/ads/class-ads.php
thomasguillot: Derived in bb93195c5: `get_taxonomy( self::ADVERTISER_TAX )->cap->manage_terms`, with a null guard in case the gate ever runs before registration. On the env it resolves to `manage_categories`, so behaviour is unchanged today and a future capabilities map carries automatically.

Left it narrower than the Ads screen's `edit_posts` on purpose. The pass writes term counts, so the cap that gates it should be the one that governs the terms. The PR body now says "by a user who can manage advertisers" rather than implying the first admin load of any kind.

thomasguillot:
## Round 2

All four threads addressed in ce93a7277 and bb93195c5, with per-thread detail inline.

**The design call went your way.** `auto-draft` is dropped from a `status` request rather than rewritten, so the endpoint no longer answers a question the caller did not ask. A request for `auto-draft` and nothing else installs a never-matching bucket clause, which is what keeps it from falling back to the posts controller's unfiltered `post_status = auto-draft`. On a local site with 95 saved drafts, `?status=auto-draft` went from 100 rows to 0, and `?status=auto-draft,draft` still returns the drafts.

**The `$widened !== $values` guard is gone,** so the selection always drives `post_status`. Sending the pre-deploy default set, the filter previously returned `["publish","private","future","draft","pending","auto-draft"]` and now returns the same list without `auto-draft`.

**The recount gate derives its capability from the taxonomy,** with a null guard for the case where it runs before registration. It resolves to `manage_categories`, so nothing changes today and a future capabilities map carries automatically.

The PR body picks up the two notes you asked for: the `_v4` bump is belt-and-braces because advertiser terms are only written on save, and deleting `newspack_nl_advertiser_count_recounted_v4` re-runs the pass if a site's counts ever look stale. The recount wording now says "by a user who can manage advertisers", and a tenth test step covers the REST behaviour.

Three new tests lock the changed paths: `test_auto_draft_only_status_request_matches_nothing`, `test_auto_draft_is_dropped_from_a_mixed_status_request` and `test_legacy_default_status_set_drops_auto_draft`. The widening test's second half now asserts the incoming `post_status` is replaced rather than preserved, so it fails on a revert of the guard removal.

Verification: the three list REST suites 105/105, PHPCS clean, and both findings reproduced on a local site before the fix and confirmed closed after.

Thanks for the REST semantics catch, that one was the right call.

PR #855, opened 2026-08-12
fix(content-gate): keep Memberships-exempt posts public after migration

dkoo:
### All Submissions:

Posts that WooCommerce Memberships marked always-public lose that exemption when a site moves to Access Control. `wp newspack migrate-membership-gates` carries over plans, rules, and layouts, but per-post exemptions are stored as `_wc_memberships_force_public` and stay behind. Those posts start being gated the moment Memberships is deactivated and the migrated gates take effect.

A post with no Access Control exemption of its own now defers to the Memberships flag. That covers the gate check and the "Disable access control restrictions for this post" toggle alike, so the editor shows what the gate will do. **A recorded exemption always wins, even a falsy one**, so turning the toggle off sticks.

The inferred value stays inferred. The block editor returns every meta value it read whenever any one of them is edited, so the exemption is stripped back out of those saves — otherwise opening a post and saving it would write the inference to the database, outliving the Memberships flag it came from. Sites that have finished migrating can switch the fallback off with the `newspack_content_gate_respect_memberships_force_public` filter.

1. Define `NEWSPACK_CONTENT_GATES` as `true` and enable Audience Management. Both are required for gating.
2. Create a published gate that puts all posts behind registration.
3. Open any post while logged out. The gate appears.
4. Run `wp post meta update <id> _wc_memberships_force_public yes` for that post.
5. Reload while logged out. The post is readable and no gate appears.
6. Open the post in the editor. The "Disable access control restrictions" toggle reads ON.
7. Save the post without touching that toggle. `wp post meta list <id>` shows no `newspack_content_restriction_is_exempt` row.
8. Run `wp post meta delete <id> _wc_memberships_force_public`. The gate returns.
9. Restore the flag, then turn the toggle off and save. The gate returns and stays.
10. On a second post, set the same meta to `no`. The gate still applies.
11. Set the flag on an attachment. It stays gated — the fallback covers only post types the exemption is registered for.

### Other information:

* [x] Have you successfully run tests with your changes locally? Full `newspack-plugin` suite: 2934 tests, no failures.

* Follow-up worth its own issue: `migrate-membership-gates` still doesn't record these exemptions, so they exist only by inference. Turning the opt-out filter on before backfilling real `newspack_content_restriction_is_exempt` rows would re-gate that content.
* Rollout risk: on a site that used force-public long ago and has since stopped using Memberships, those posts become exempt again. The filter above is the opt-out.
* `newspack-manager` is the only other consumer of the content-gate contract and does not read this meta, so no coordinated change is needed.

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-restriction-control.php
@@ -55,12 +55,21 @@ class Content_Restriction_Control {
 	const IS_EXEMPT_META_KEY = 'newspack_content_restriction_is_exempt';
 
 	/**
+	 * Post meta key WooCommerce Memberships uses to force a post public.
+	 *
+	 * @var string
+	 */
+	const WC_FORCE_PUBLIC_META_KEY = '_wc_memberships_force_public';
+
+	/**
 	 * Initialize hooks and filters.
 	 */
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_meta' ] );
 		add_action( 'init', [ __CLASS__, 'register_meta_guards' ] );
 		add_filter( 'newspack_is_post_restricted', [ __CLASS__, 'is_post_restricted' ], 10, 2 );
+		// Priority 20 so the fallback wins over the meta key's registered default.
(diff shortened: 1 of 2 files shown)


dkoo:
Low-risk change that proactively plugs a cross-platform data gap and prevents the need for post-migration backfills in the future.

PR #856, opened 2026-08-12
fix(subscribe): return a defined key set from the form response

jason10lee:
### All Submissions:

The subscribe block's form response now carries a defined set of fields rather than passing through whatever the email service provider returns.

Two constants in `src/blocks/subscribe/index.php` name the contract: `RESPONSE_KEYS` for the payload and `METADATA_KEYS` for the nested `metadata`. `send_form_response()` filters the success payload against both. The failure branch returns the reader-facing message on its own.

This gives the block a response shape that does not vary with the provider, so the front end depends on fields we define rather than on a provider's record layout.

The tests pin the contract in both directions. They fail if a key the block reads stops arriving, and they fail if a key arrives that the contract does not name, so widening the set is a deliberate act that updates the test.

One behavior change worth noting for release: custom JavaScript that reads provider fields off this response will stop receiving them. Every first-party consumer is covered by the tests here.

NPPM-3086.

1. On a site with a newsletter provider connected, add a Newsletter Subscription block to a post and publish it.
2. Open the post in a private window and submit the form with an address that is not yet subscribed. The success message appears.
3. Open the network panel and find the POST for that page. The JSON body contains only `message`, `newspack_newsletters_subscribe`, `newspack_newsletters_subscribed`, `metadata`, `registered`, `verified`, `verification_nonce`, and `email`, with no provider fields alongside them.
4. Check the `metadata` object in that same response. It contains only `current_page_url`, `newspack_popup_id`, `gate_post_id`, `newsletters_subscription_method`, `status`, `registration_method`, and `registered`.
5. Submit the form again with the email field empty. The response is a 400 carrying a message and nothing else.
6. With Reader Activation and email verification enabled, subscribe as a new reader. The verification prompt appears, which confirms `registered`, `verified`, `verification_nonce`, and `email` still reach the block.
7. Run `n test-php` from `plugins/newspack-newsletters`. The suite passes.

### Other information:

* [x] Have you written new tests for your changes, as applicable? 12 tests in `tests/test-subscribe-block-response.php`.
* [x] Have you successfully run tests with your changes locally? 736 tests, 3050 assertions, 0 failures.

<details>

**Files.** `src/blocks/subscribe/index.php` (constants and the two filters), `tests/test-subscribe-block-response.php`, plus `tests/mocks/class-subscribe-block-redirect-interrupt.php` and one line in `tests/bootstrap.php` supporting the redirect-branch test.

**Exits.** `send_form_response()` has four: JSON success (filtered), JSON failure (message only), the GET redirect (unchanged, carries a status flag and an error code), and a fall-through for a non-JSON, non-GET request that emits nothing.

**`metadata` is built locally** in `process_form()` and overwritten before the response is sent, so `METADATA_KEYS` is a second line of defense rather than the primary one. The closed-world test is what holds it.

**Two keys come from another plugin.** `verified` and `verification_nonce` are produced by `\Newspack\Reader_Activation::get_verification_payload()` and merged whole. A key added there needs an entry in `RESPONSE_KEYS` and a matching read in `view.js` before it reaches the front end; the `RESPONSE_KEYS` docblock records this.

**Test harness note.** `wp_send_json()` only raises a catchable `WPDieException` when `wp_doing_ajax()` is true, and it dispatches through `wp_die_ajax_handler` rather than the generic handler. The test's `set_up()` filters both.

</details>


jason10lee:
The red **Release** check is a known release-workflow bug, not a defect in this branch: the plan step fails when no package needs releasing. #833 fixes it and is awaiting review. PHPCS, PHPUnit, JS, Build ZIP and CI OK all passed.

If you test from artifacts: the prerelease zip `newspack-newsletters@3.38.6-hotfix-nppm-3086-subscribe-response-allowlist.1` was built from `71f07fcec` and predates the last two commits. Build from the branch head instead.

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
```diff
@@ -0,0 +1,385 @@
...
+	 * `$args_to_remove` — and this test is the only one that exercises that
+	 * branch, so such a regression would otherwise sit behind a green run
+	 * indefinitely. Catching only Subscribe_Block_Redirect_Interrupt closes
+	 * that gap: nothing else thrown in the branch is this exception type, so
+	 * it propagates and fails the test instead of being absorbed here.
+	 */
+	public function test_non_json_request_does_not_emit_json() {
+		unset( $_SERVER['HTTP_ACCEPT'] );
+		$_SERVER['REQUEST_METHOD'] = 'GET';
+
```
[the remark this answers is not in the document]
jason10lee: Fixed in `6bea031c9`. `REQUEST_METHOD` is now saved in `set_up()` and restored in `tear_down()`, matching how `HTTP_ACCEPT` is already handled.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
```diff
@@ -11,6 +11,48 @@
...
+ * caller. The consumer of record is `src/blocks/subscribe/view.js`: adding a key
+ * here without a matching read there has no effect, and removing one breaks the
+ * front end silently. Exception: `verified` and `verification_nonce` are read
+ * there too, but produced elsewhere — newspack-plugin's
+ * `\Newspack\Reader_Activation::get_verification_payload()` — and merged into
+ * the result wholesale below, so a key added on that side needs a matching
+ * read in view.js and an entry here before it can reach the front end.
+ */
+const RESPONSE_KEYS = [
+	'message',
```
adekbadek: **Suggestion:** `message` is the one allowlist entry with no first-party producer — `message` is the only entry here that the success path never sets itself. Everything else on that branch is assigned in this file — `newspack_newsletters_subscribed` at :468, `FORM_ACTION` at :646, `metadata` at :649, and `email`/`registered`/the verification payload at :659-661 — while the failure branch builds `compact( 'message' )` by hand at :465 and never consults this constant. So on the one branch the allowlist governs, the only thing that can supply `message` is the record the constant exists to bound.

I checked this on a running site: with a provider record carrying a `message` field, that value comes back in the 200 response. Nothing displays it — `view.js:147` renders the block's own `data-success-message` whenever `newspack_newsletters_subscribed` is truthy, and :468 forces it to `1` — so there is no visible problem today.

Dropping `'message'` from `RESPONSE_KEYS` costs nothing and narrows the contract to keys that are both producible and consumed.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
```diff
@@ -11,6 +11,48 @@
...
+
+/**
+ * Keys permitted inside the response's `metadata` member.
+ *
+ * Bounded for the same reason as RESPONSE_KEYS: allowlisting `metadata` as a
+ * whole would leave the response closed at the top level and open one level
+ * down. `gate_post_id` is never set by this block — it originates in
+ * newspack-plugin's content gate — but view.js reads it defensively.
+ */
+const METADATA_KEYS = [
```
adekbadek, line 63: **Suggestion:** The two lists answer different questions; the docblocks describe one — `RESPONSE_KEYS` states the governing rule at :19-21 — the consumer of record is `view.js`, and adding a key without a matching read there has no effect. `METADATA_KEYS` then lists `current_page_url` and `status`, which `view.js` never reads. They are produced by this block rather than by the provider, so keeping them is a reasonable call for a hotfix; the problem is that a later reader applying the stated rule would delete them.

Related, three lines up: the comment at :43 says `gate_post_id` "originates in newspack-plugin's content gate". Nothing in the monorepo puts it into this block's metadata — `$metadata` is built from a literal at :554 and assigned directly at :649 — so no current path supplies it. `view.js:162` does read it defensively, which is the accurate version of that sentence.

Describing this list as what the block itself produces, rather than what `view.js` reads, would make both docblocks true.

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
```diff
@@ -0,0 +1,432 @@
...
+	 * ever reaches wp_die() when wp_doing_ajax() is true (otherwise it calls a bare
+	 * die()); and once there, wp_die() dispatches on wp_doing_ajax() before anything
+	 * else, so the generic 'wp_die_handler' filter that WP_UnitTestCase wires up to
+	 * throw WPDieException is never consulted. Wiring the same handler onto
+	 * 'wp_die_ajax_handler' is what makes that exception catchable here.
+	 *
+	 * @param mixed $data Payload handed to send_form_response().
+	 * @return array Decoded response body.
+	 */
+	private function capture_response( $data ) {
```
adekbadek, line 86: **Suggestion:** The HTTP status is part of the stated contract but nothing pins it — The comment added at index.php:459-462 names the failure contract as the reader-facing message *and* the HTTP status, and `view.js` gates its whole success path on `res.status === 200` (:215, :225). No test pins either code — `capture_response()` returns only the decoded body — so swapping the `400` and `200` at index.php:465/478 leaves the suite green.

It is reachable from this helper. I confirmed in the test environment that `status_header()` still applies its `status_header` filter under CLI; only the `header()` call is skipped. So one filter captures the code:

```php
add_filter( 'status_header', function ( $header, $code ) use ( &$status ) { $status = $code; return $header; }, 10, 2 );
```

Returning it alongside the body would let `test_error_branch_returns_message_only` assert the 400 it is named for.

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
```diff
@@ -0,0 +1,432 @@
...
+	public function test_non_array_metadata_does_not_pass_through() {
+		$response = $this->capture_response(
+			[ 'metadata' => 'not-an-array' ]
+		);
+
+		$this->assertSame( [], $response['metadata'], 'A non-array metadata value must be normalized, not passed through.' );
+	}
+
+	/**
+	 * Every other test in this file pins RESPONSE_KEYS/METADATA_KEYS against
```
adekbadek, line 345: **Suggestion:** The suite ratchets against removal, but not against addition — This docblock, and the PR description, both say widening the allowlists "has to be a deliberate act that updates this test". I ran the mutation against the suite in a container and it does not hold: adding `'language'` to `RESPONSE_KEYS`, or `'location'` to `METADATA_KEYS`, leaves all 13 tests green. Both are real provider fields.

The cause is that `array_intersect_key()` only ever removes keys, so a widening is invisible unless the fixture happens to offer that key for the filter to preserve. `merge_fields` *is* caught — `test_provider_record_fields_are_not_returned` names it at :139 — but only because it sits in that fixture. Anything no fixture names passes. The comment at :375-379 already records this caveat for the metadata half; the same limit applies to the top half, where the docblock states the stronger claim.

Asserting the constants themselves alongside the response closes it, and holds regardless of what any fixture offers:

```php
$this->assertSame( $expected, array_values( \Newspack_Newsletters\Blocks\Subscribe\RESPONSE_KEYS ) );
```

(both sorted, and the same for `METADATA_KEYS`). The existing response assertion then proves the constant is actually applied, which is the half this test does well.

While you are in here: several of the surrounding tests exercise the same two `array_intersect_key()` calls with a different key name and are subsumed by this one. Consolidating them would make the file read more like the spec it is trying to be — though that is fine to leave for the forward-port rather than the hotfix.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
```diff
@@ -414,11 +456,25 @@ function send_form_response( $data ) {
...
+			// what a caller can see; view.js doesn't read that either, so this is
+			// deliberate, not an oversight.
 			$message = $data->get_error_message();
-			\wp_send_json( compact( 'message', 'data' ), 400 );
+			\wp_send_json( compact( 'message' ), 400 );
 			exit;
 		} else {
 			$data['newspack_newsletters_subscribed'] = 1;
+			$data                                    = array_intersect_key( $data, array_flip( RESPONSE_KEYS ) );
+			if ( isset( $data['metadata'] ) ) {
```
adekbadek: **Nit:** `isset()` lets a null `metadata` skip its own normalisation — `isset()` is false for `null`, so a `metadata` of `null` skips this branch and ships as `"metadata": null` — the one shape the normalisation below is here to prevent. I confirmed it comes back that way.

It is not reachable today: :649 always assigns the locally-built array, and `view.js` reads `metadata` with optional chaining throughout, so nothing breaks either way. `array_key_exists()` costs the same and makes the code do what the comment says.

plugins/newspack-newsletters/tests/mocks/class-subscribe-block-redirect-interrupt.php
```diff
@@ -0,0 +1,25 @@
...
+ * Thrown from the `wp_redirect` filter installed by
+ * Subscribe_Block_Response_Test::test_non_json_request_does_not_emit_json() to
+ * unwind out of send_form_response() before its `exit;` statement would end
+ * the test process for real. `wp_redirect()` applies this filter before it
+ * ever calls `header()`, so throwing here needs no cooperation from PHP's
+ * warning-to-exception conversion — unlike the WPDieException path the JSON
+ * branch relies on, this one fires deterministically regardless of whether
+ * headers have already been sent in this process.
+ */
+class Subscribe_Block_Redirect_Interrupt extends Exception {
```
adekbadek: **Nit:** Unprefixed global class name, and a phpcs:ignore without a reason — The class lands unprefixed in the global namespace and is loaded into every test run through `bootstrap.php`. The convention here is `Newspack_<Feature>`; the sibling mocks are unprefixed only because they stand in for third-party classes, which this one does not.

Separately, the `// phpcs:ignore WordPress.Files.FileName` on line 1 carries no `--` reason, unlike the other ignores in this diff, and the filename does match the class name. If the sniff genuinely fires, worth recording why; if not, it can go.

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
```diff
@@ -0,0 +1,432 @@
...
+			$interrupt           = new Subscribe_Block_Redirect_Interrupt( 'Intercepted for testing; not a real error.' );
+			$interrupt->location = $location;
+			throw $interrupt;
+		};
+		add_filter( 'wp_redirect', $intercept );
+
+		ob_start();
+		try {
+			send_form_response( [ 'merge_fields' => [ 'FNAME' => 'Ada' ] ] );
+			$this->fail( 'Expected the wp_redirect filter to interrupt send_form_response().' );
```
adekbadek: **Nit:** This `fail()` is unreachable, and the comment below explains a path that cannot happen — `send_form_response()` follows `wp_safe_redirect()` with `exit` at index.php:495-501, so had the filter not thrown, the process would have ended there and taken the test run with it — control never reaches this line. The comment at :422-424 then justifies the `finally` on the premise that this `fail()` "throws past the catch block", which cannot occur.

Either drop it and shorten the comment, or keep it and say what it actually guards against.

adekbadek (approved):
Release risk: LOW — no blockers. Verified end to end against a mock provider on a running site: a full contact record reaches the handler and none of it reaches the wire at either level, the registration path still delivers its verification keys, and the failure branch returns the message alone. The comments cover the contract's edges — one allowlist entry with no first-party producer, and a suite that ratchets against removal but not addition.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
jason10lee: Dropped in `294187b87`. Confirmed the whole chain: the failure branch hand-builds `compact( 'message' )` at `:465` and never consults the list, and on the success branch `view.js:147` renders `data-success-message` whenever `newspack_newsletters_subscribed` is truthy — which `:468` forces to `1` on every success. So the response's `message` was unreachable at both ends.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
jason10lee: Both docblocks rewritten in `95a8d3453`. `RESPONSE_KEYS` now says it governs the success branch only, and that a key `view.js` doesn't read is still not safe to add — what the list manages is what leaves the server.

One correction on `gate_post_id`: it does have a producer. `newspack-plugin/src/content-gate/gate.js:116` adds it as a hidden input to every form inside a gate. What's missing is the copy into this block's `$metadata`, which is built from a literal at `:554` that omits it — so `view.js:161` reads a key nothing populates.

That looks like a live bug rather than a documentation gap: the registration block reads the same value from the submitted body (`reader-registration/view.js:234`) and gets it, so gate attribution probably fires for registrations and not for newsletter signups. Filed as NPPM-3172 — https://linear.app/a8c/issue/NPPM-3172

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
jason10lee: The finding holds — swapping the 400 and the 200 leaves the suite green — but the mechanism can't work, one layer above where it was diagnosed. `wp_send_json()` wraps its status call in `if ( ! headers_sent() )`, and inside the suite `headers_sent()` measures **TRUE**: PHPUnit writes progress output before any test runs. So `status_header()` is never called and the filter has nothing to fire on. Your point about `status_header()` applying its filter under CLI is right; it just doesn't execute on this path.

I built it, watched it capture `null` on both branches, then measured `headers_sent()` to find out why — so this is tested rather than reasoned. Testing it needs process isolation with buffering established first, or a seam in `send_form_response()` reporting the status independently of the header layer. Ticketed as NPPM-3171 — https://linear.app/a8c/issue/NPPM-3171

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
jason10lee: Closed in `a60bb0185` with the constant assertions you suggested. Verified the gap first and then the fix: adding `'language'` to `RESPONSE_KEYS` passed the whole suite before, and now fails `test_response_key_set_is_closed`.

Left the consolidation for the forward-port as you suggested — NPPM-3173.

plugins/newspack-newsletters/src/blocks/subscribe/index.php
jason10lee: `array_key_exists()` in `95a8d3453`, with a test in `a60bb0185`. Reverting to `isset()` now fails it with `Failed asserting that null is identical to Array &0 ()` — the `"metadata": null` shape itself.

plugins/newspack-newsletters/tests/mocks/class-subscribe-block-redirect-interrupt.php
jason10lee: Both in `a60bb0185`. The class is `Newspack_Newsletters_Subscribe_Block_Redirect_Interrupt` now, matching `Newspack_Newsletters_Mailchimp_Api`; the unprefixed siblings are third-party stand-ins, as you say.

The ignore was inert — phpcs exits 0 without it on the original filename, so it went. Renaming the class did make `WordPress.Files.FileName` fire, so the file is renamed too and `bootstrap.php` follows.

plugins/newspack-newsletters/tests/test-subscribe-block-response.php
jason10lee: Removed in `a60bb0185`, and the `finally` comment now says what it actually guards: any throw other than the interrupt still has to close the buffer and drop the filter. Added a line recording why there's no `fail()` — the only path past the interrupt is the handler's own `exit`, which would end the process rather than reach an assertion.

PR #857, opened 2026-08-12
chore(bin): reject option-shaped names in env and worktree helpers

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? n/a: shell only, no PHP in this change.

`n env create --help` created an environment named `--help` instead of printing usage. It wrote `docker-compose.env---help.yml`, an `envs/--help/` directory and an `/etc/hosts` line, printed no usage text, and exited 0.

`AGENTS.md` tells people to run any command with `--help` for its full options, so the documented path is the one that misfires.

Two independent gaps combine. Neither is wrong on its own:

* `bin/env.sh` reads the environment name positionally, and no subcommand handles `-h` or `--help`.
* `validate_env_name` in `bin/_common.sh` allows a leading dash, so `--help` passes as an ordinary name.

This adds both guards:

* **`validate_env_name` rejects a leading dash.** One line in shared code, so it covers every subcommand at once, including `destroy`. It excludes a leading dash and nothing else: this validator also gates `up`, `down` and `destroy`, so a stricter first-character rule would strand an environment created under an older one, leaving the container, compose file, hosts entry and `envs/` directory removable only by hand.
* **Every `env` subcommand that reads `$2` handles `-h` and `--help`** — `create`, `up`, `down`, `destroy`, `list` and `cleanup` — printing the usage text each already had. An explicit help request exits 0; a missing name still exits 1, as before. `e2e-setup` is unchanged: it delegates to `bin/setup-local-e2e.sh`, which has its own `--help` arm.

`bin/_common.sh` carries a second validator, `validate_name`, with the same hole by a different route. It covers names that become git branches and path components, and `bin/worktree.sh` plus the `--worktree` parsing in `bin/env.sh` read those positionally through it, so `n worktree add --help` reached git with `--help` as the branch name. Git refuses that refname, so nothing was created, but the user got a git error rather than usage.

* **`validate_name` rejects a leading dash** on the same rule, so both validators exclude exactly that and nothing else. Slashed branches are the norm here and branches such as `_pr738` are in use, so both keep working. `..` and a leading `/` are rejected as before.

One behavior change, identical in both validators: a name beginning with `-` is rejected. Nothing else changes, so every name either validator accepted before is still accepted. `n env up --all` is unaffected.

One consequence worth stating: an environment whose name begins with `-` can only exist if the bug above created it, and it is no longer reachable through `n env` — `down` and `destroy` print usage rather than acting on it. Removing one is manual:

```sh
rm -f docker-compose.env---help.yml
rm -rf './envs/--help'
sudo newspack-manage-host host-remove '--help.test'
```

This overlaps #682 in `bin/env.sh`: it inserts at the same point above `case $1 in`, and it rewrites the `n env up` usage line this change replaces. Happy to rebase whichever lands second.

There's no linked issue; I hit this while using the tool.

**One-time cleanup, if you have an environment created by this bug.** `n env create --help` used to create an environment literally named `--help`. This change makes such a name unreachable through `n env`: `destroy` reads `--help` as a help request before the name is validated, and the tightened validator rejects any leading dash. `n env cleanup` doesn't help either — it lists environments by globbing the compose files but delegates removal back to `n env destroy`.

Removing one takes six steps by hand, matching what `destroy` does:

```bash
docker stop newspack_env___help; docker rm newspack_env___help
docker compose -f docker-compose.yml exec -T db mariadb -h localhost -u root -p"$MYSQL_ROOT_PASSWORD" \
  -e 'DROP DATABASE IF EXISTS `wordpress___help`; DROP DATABASE IF EXISTS `wp_tests___help`'
rm -rf envs/--help logs/env---help
sudo newspack-manage-host host-remove --help.test   # only if a hosts entry was added
rm -f docker-compose.env---help.yml
```

A `--` end-of-options form for `destroy` was considered and set aside: it would mean accepting a leading dash again on the one path that runs `rm -rf`, to serve debris from a bug this PR is the first to fix.

1. Run `./n env create --help`. The usage line prints and the command exits 0.
2. Check that nothing was created. There is no new `docker-compose.env-*.yml`, no new directory under `envs/`, and no new `/etc/hosts` line.
3. Repeat with `-h`, and with `up`, `down`, `destroy`, `list` and `cleanup`. Each prints its own usage and exits 0.
4. Run `./n env create` with no arguments. The usage line prints and the command exits 1.
5. Run `./n env create -foo`, then `.foo`, then `_foo`. Each is rejected.
6. Run `./n worktree add --help`. The name is rejected, rather than reaching git as a branch name.
7. Run `bash bin/tests/test-name-guards.sh`. All assertions pass.
8. Run `./n env create <name>` with an ordinary name, then `up` and `destroy` it. The cycle still works.

### Other information:

* [x] Have you written new tests for your changes, as applicable? `bin/tests/test-name-guards.sh` covers both validators and the help arms, and the Shell specs job runs it.
* [x] Have you successfully run tests with your changes locally? The 45 assertions pass. I confirmed the spec fails when each guard is removed on its own, when a validator is broken so that it exits non-zero without rejecting, and that with every guard reverted at once it still writes nothing outside its sandbox (`/etc/hosts` byte-identical across the run). Manual steps 1 through 6 also run clean. Step 8 I have not run end to end; the change does not touch that path beyond validation.

### DIFF
--- a/bin/_common.sh
+++ b/bin/_common.sh
@@ -7,9 +7,15 @@
 [[ -n "${NABSPATH:-}" && -f "$NABSPATH/n" ]] ||
     NABSPATH="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 
+# Names that become git branches and path components. Slashes are allowed here
+# and not in validate_env_name, since `fix/some-thing` is the normal branch
+# shape. The first character may not be a dash, so an option is never taken as
+# a name: bin/worktree.sh and the --worktree parsing in bin/env.sh both read
+# these positionally. A leading dot or underscore stays legal because branches
+# like _pr738 are in use, and git rejects the refnames that are truly invalid.
 validate_name() {
-    if [[ ! "$1" =~ ^[a-zA-Z0-9._/-]+$ ]] || [[ "$1" == *..* ]] || [[ "$1" == /* ]]; then
-        echo "Error: invalid $2 '$1' (only alphanumeric, dots, hyphens, underscores, slashes allowed; no '..' or leading '/')"
+    if [[ ! "$1" =~ ^[a-zA-Z0-9._][a-zA-Z0-9._/-]*$ ]] || [[ "$1" == *..* ]] || [[ "$1" == /* ]]; then
+        echo "Error: invalid $2 '$1' (must not start with '-'; only alphanumeric, dots, hyphens, underscores, slashes allowed; no '..' or leading '/')"
         exit 1
     fi
 }
--- a/bin/_common.sh
+++ b/bin/_common.sh
@@ -30,8 +30,13 @@ validate_name() {
 # environment created under an older, laxer one — unmanageable and removable
 # only by hand.
 validate_env_name() {
-    if [[ ! "$1" =~ ^[a-zA-Z0-9._][a-zA-Z0-9._-]*$ ]]; then
-        echo "Error: invalid environment name '$1' (must not start with '-'; only alphanumeric, dots, hyphens, underscores allowed)"
+    # The `..` clause matches validate_name's. `n env destroy ..` would otherwise
+    # validate and reach `rm -rf "$NABSPATH/envs/.."`, i.e. the workspace root.
+    # POSIX rm refuses a trailing `.` or `..` component, so that is not a live
+    # escape today -- but the guard then lives in rm rather than here, and moves
+    # out from under us the moment a call site builds the path differently.
+    if [[ ! "$1" =~ ^[a-zA-Z0-9._][a-zA-Z0-9._-]*$ ]] || [[ "$1" == *..* ]]; then
+        echo "Error: invalid environment name '$1' (must not start with '-' or contain '..'; only alphanumeric, dots, hyphens, underscores allowed)"
         exit 1
     fi
 }
--- a/bin/_common.sh
+++ b/bin/_common.sh
@@ -14,10 +14,14 @@ validate_name() {
     fi
 }
 
-# Stricter validation for env names — no slashes (Docker rejects them in container/service names)
+# Stricter validation for env names — no slashes (Docker rejects them in
+# container/service names), and an alphanumeric first character. The leading
+# character rule is what stops an option being read as a name: bin/env.sh takes
+# the name positionally, so without it `n env create --help` validates cleanly
+# and creates an environment called "--help" instead of printing usage.
 validate_env_name() {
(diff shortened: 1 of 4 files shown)


thomasguillot:
Funny timing, I hit this from the other end today and was about to open the same fix.

One more case for the pile: `n env destroy --help` runs the full teardown against an environment called `--help` and prints `Destroyed environment '--help'`. Every step it performs is `IF EXISTS` or silenced, so it reports success whatever the name was. Nothing was lost in my case, since those databases never existed, but it reads like it worked.

Two gaps this leaves, both in `n` rather than `bin/`:

- `n --help` and `n -h` still answer "Unknown command: --help", with the `help` case sitting at `n:495`.
- `n sites-add` (`n:383`) and `n sites-drop` (`n:397`) pass the name straight to the container unvalidated, same positional shape as `env`.

Happy to pick those up separately once this lands, rather than widening your diff.

jason10lee:
`destroy` is covered here—the case gained `|| is_help_arg "$env_name"`, and `validate_env_name` now rejects a leading dash, so `n env destroy --help` prints usage and exits 0 rather than tearing down. The "reports success whatever the name was" half is the sharper part of your report, and the dash rule closes it for `create`/`up`/`down`/`destroy` alike.

Both gaps confirmed and tracked as NPPM-3157 (https://linear.app/a8c/issue/NPPM-3157). `n --help` and `n -h` still answer "Unknown command", with the `help` case now at `n:498`. And `sites-add` (`n:383`) / `sites-drop` (`n:397`) interpolate `$2` and `${@:2}` unquoted into `sh -c`, so a name carrying a space or a `;` reaches the container's shell as syntax rather than being validated. Those two lines are also the repo's two standing `SC2145` shellcheck errors.

Chasing those turned up a third one, now NPPM-3156 (https://linear.app/a8c/issue/NPPM-3156): `n sites-drop` exits 0 and prints nothing when it has no TTY, because `n` only passes `-i` to `docker exec` from a terminal, so the confirmation `read` hits EOF and the script returns success without dropping anything. Same shape as yours, on a destructive command.

Agreed on keeping them out of this diff. I'll carry both as follow-ups once this lands. `n test-js` had the same shape on #410 and took `watch`'s form there: direct exec, quoted argument, no `sh -c`. `sites-drop` wants the same.

Shout if you hit more of them.

dkoo (approved):

PR #859, opened 2026-08-13
Pin a root-owned PATH in the passwordless host wrapper

jason10lee:
### All Submissions:

`bin/newspack-manage-host` runs as root through the sudoers drop-in that `bin/setup-networking.sh` installs, which is what lets `n start` and `n env create` set up loopback aliases and hosts entries without a password prompt (#67). This PR pins `PATH` to root-owned system directories before the first external call, so helper lookup no longer depends on the caller's environment. Homebrew prefixes stay out on purpose, since they can sit in a user-writable location.

Adds `bin/tests/test-newspack-manage-host-path.sh`, which the existing `Shell specs` job already picks up. It checks that a binary placed earlier in `PATH` is not used, that the pinned `PATH` still resolves every helper the wrapper calls, and that no command in the wrapper falls outside the pinned directories. The third check is what keeps this from drifting: it fails if a later change adds a helper the pin does not cover, rather than letting it surface as a runtime error under `set -e`.

Found by reading the code. We have no report of it occurring.

Run `bash bin/tests/test-newspack-manage-host-path.sh` — 3 passed, 0 failed; the `Shell specs` job runs it in CI alongside `test-lo0-alias-exists.sh`. `n env create <name>` still sets up networking without a password prompt.

### Other information:


jason10lee:
Copilot ran here and reported that it could not review any of the files in this PR. Both are shell, which its code review does not cover, so the usual automated gate is absent. We ran a multi-model AI review in its place: three models (Claude, Codex, Mistral) reviewing the diff independently, with voting and a skeptical re-review pass.

All three cleared the PATH pin itself. One singled out the wrapper's comment as the strongest part of the change, for naming the property the pin rests on rather than listing directories.

**Every finding landed on the test, not on the pin.** Four are addressed in 4b28d8f4f:

- **The guard meant to catch a future edit checked a hardcoded list of command names**, so it could only catch names someone had already thought of. It now derives the command set from the wrapper. Writing that turned up a second gap: the original scan missed `grep`, because it sits behind `!`.
- **The hijack probe had no positive control.** A `clean` result could not be told apart from a code path that was never reached. The identical probe now also runs against a copy with the pin stripped, and has to hijack there.
- **Widening the pin was invisible.** Deriving the value fixes staleness but not detection, since appending a directory still resolves everything. The spec now asserts the property the pin rests on: every pinned directory is root-owned and not group- or world-writable.
- **The probe ran the real `alias-add` path**, which mutates host state when the caller is root, as CI containers often are. It now uses an address the wrapper's validation accepts and `ifconfig` rejects, so the call is reached and nothing is applied.

Each guard is mutation-tested rather than only observed passing.

<details>
<summary>Mutation coverage</summary>

Each new guard was run against a wrapper mutated to contain the defect it exists to catch:

| Mutation | Result |
| --- | --- |
| Append `/opt/homebrew/bin` to the pin | Fails, naming `/opt/homebrew/bin` |
| Add a helper that resolves only outside the pin | Fails, naming `shellcheck` |
| Remove the `export PATH=` line | Fails on the pin-declared check |

The derivation has its own control: a command appended to a copy of the wrapper must be detected, so a derivation that silently stopped matching fails rather than returning an empty set and passing everything downstream.

</details>

<details>
<summary>Findings not adopted, and why</summary>

Three findings rested on assumptions about the repository the reviewers could not check:

- **Entrypoint interpreter hijack.** Reasoned from a `#!/usr/bin/env bash` shape; the shebang is `#!/bin/bash`, an absolute path, so there is no PATH-resolved interpreter. The recommendation to execute the wrapper directly rather than via `bash` was adopted anyway, since it matches how sudo invokes it.
- **"Both resolution checks fail on a Linux runner."** The `Shell specs` job log shows all assertions passing on ubuntu-latest.
- **"Nothing picks up a new `bin/tests/` file."** That job globs `bin/tests/*.sh`, which is why the spec lives there; it ran 8 seconds into the run.

Two further findings were left open deliberately: exercising the `set -e` failure contract, which tests shell behaviour more than this wrapper, and having the harness exit on first failure rather than accumulating, which a skeptical re-review pass already disputed.

</details>

bin/newspack-manage-host
```diff
@@ -11,6 +11,38 @@
...
+# this wrapper exists to serve — inherits that.
+#
+# Every directory here must be writable only by root, which is the property
+# doing the work. On Apple Silicon /opt/homebrew/bin is group-writable and owned
+# by the login user, so adding it would restore the same hole under a different
+# name; the same applies to any brew prefix. /usr/local/bin is left out on the
+# same principle even where it happens to be root-owned, because nothing needed
+# here lives there and its permissions vary by machine.
+#
+# The pin must also cover every external command below, or the script dies under
```
adekbadek: **Suggestion:** The stated failure mode for an uncovered command does not hold for `grep`, which fails open

The paragraph tells the next editor that a command outside the pinned directories makes the script "die under `set -e` with a confusing 'command not found'". That holds for two of the three commands it names: with the pin pointed at an empty directory, `alias-add` exits 127 with `ifconfig: command not found`, and `host-remove` exits 127 on `sed`.

`host-add` behaves differently. Its `grep` call at :85 sits in an `if !` condition, where `set -e` is suspended, and `2>/dev/null` swallows the shell's not-found message — so 127 inverts to true and the `echo … >> /etc/hosts` runs. Three consecutive runs exited 0 and left three identical entries in the hosts file.

This is not about exercising the `set -e` contract, which was declined; it is about what the sentence claims. Naming both shapes keeps the guidance usable to whoever adds the next helper: `ifconfig` or `sed` aborts, `grep` silently duplicates.

While rewriting this paragraph, the inventory in the next sentence is worth dropping. The spec's header explains that it derives the pin and the helper set from the wrapper "because a copy in this file would go stale in the direction that matters"; "Today that is ifconfig (/sbin), grep (/usr/bin) and sed (/usr/bin)" is that copy, fifteen lines above the calls it enumerates and checked by nothing. The rule that follows it stands on its own.

bin/newspack-manage-host
```diff
@@ -11,6 +11,38 @@
...
+# `set -e` with a confusing "command not found". Today that is ifconfig (/sbin),
+# grep (/usr/bin) and sed (/usr/bin). Adding a call to anything outside these
+# directories means reconsidering this line, not widening it reflexively.
+#
+# Two things outside this line carry the same weight. Line 1 is an absolute
+# `#!/bin/bash` on purpose: `#!/usr/bin/env bash` would have env search the
+# caller's forwarded PATH for an interpreter, as root, before this line runs,
+# and no pin here could reach that. And the sudoers grant must name the copy
+# bin/setup-networking.sh installs at /usr/local/bin, which it chowns to
```
adekbadek: **Suggestion:** The `chown root:wheel` clause does not establish what the sentence rests on it establishing

The comment gives the chown in `setup-networking.sh` as the reason the sudoers grant must name the `/usr/local/bin` copy rather than a checkout path — reading as: root ownership of the file is what stops the login user editing the granted target.

Whether an entry can be replaced is governed by the containing directory rather than by the entry's own ownership. `setup-networking.sh:41` chowns only the file, and `/usr/local/bin` is commonly created by Homebrew owned by the login user: on an Apple Silicon machine with the installer already run, the directory is owned by the login user at mode 755 while the wrapper inside it is `root:wheel 755`, exactly as the installer leaves it. So the contrast the sentence draws between the installed copy and a checkout path does not follow from the chown on that machine class.

This is a pre-existing property of `setup-networking.sh` rather than anything this PR introduces — the ask is only that the new comment not claim more than holds. Naming the invariant the grant depends on (the granted path sits outside anything the login user can write) rather than the chown that partly implements it would also give the next person editing that installer the right thing to check. Worth noting the paragraph at :25-27 already excludes `/usr/local/bin` from the pin because "its permissions vary by machine", and it is the same directory the grant points at.

bin/newspack-manage-host
```diff
@@ -11,6 +11,38 @@
...
+# root:wheel -- granting a path inside a developer's checkout would let the
+# login user edit this file rather than having to hijack PATH at all.
+#
+# bin/tests/test-newspack-manage-host-path.sh proves the pin holds, and pins
+# the shebang behaviourally with its own hijack probe.
+export PATH=/usr/sbin:/usr/bin:/sbin:/bin
```
adekbadek, line 53: **Suggestion:** The pin does not reach a machine that already ran `bin/setup-networking.sh`

`n` and `bin/env.sh` invoke `sudo newspack-manage-host`, which resolves the copy installed at `/usr/local/bin`. `bin/setup-networking.sh` snapshots the repo file there once at install time and nothing re-syncs it: every call site gates on `command -v newspack-manage-host` and none compares the installed file against the repo file.

On a machine set up before this branch, the deployed wrapper has no `export PATH=` line. Checked one that had run the installer — the installed copy is unpinned, so merging this changes nothing there until someone re-runs the script, and nothing in the PR tells them to. The spec asserts against `bin/newspack-manage-host` only, so it reports green either way.

README.md:342 already carries this instruction for the `.local` → `.test` migration. A matching line there, or a `cmp -s` in `n start` that prints one line when the installed copy differs from the repo copy, would make the pin take effect for people who already have the grant.

bin/tests/test-newspack-manage-host-path.sh
```diff
@@ -0,0 +1,171 @@
...
+BIN="$(cd "$(dirname "$0")/.." && pwd)"
+WRAP="$BIN/newspack-manage-host"
+FIX="$(mktemp -d)"; trap 'rm -rf "$FIX"' EXIT
+pass=0; fail=0
+ok(){ if [ "$2" = "$3" ]; then echo "  PASS  $1"; pass=$((pass+1)); else echo "  FAIL  $1 (got [$2] want [$3])"; fail=$((fail+1)); fi; }
+
+[ -x "$WRAP" ] || { echo "  FAIL  wrapper not found or not executable at $WRAP"; exit 1; }
+
+# --- the pinned value, read from the wrapper --------------------------------
+PIN="$(grep -m1 '^export PATH=' "$WRAP" | sed 's/^export PATH=//; s/[[:space:]]*$//')"
```
adekbadek: **Suggestion:** `grep -m1` reads only the first `export PATH=` line, so a re-widening below the pin is invisible

Every check that reads `$PIN` stops at the first match, so a second assignment further down the wrapper is never seen. Appending `export PATH=/opt/homebrew/bin:$PATH` on the line after the pin — which restores the caller's PATH in full — leaves this spec at 9 passed, 0 failed.

The mutation table in the PR comment covers the adjacent case (editing the pin line to append a directory), and that one does fail. Adding a line rather than editing one is the shape this check misses.

Reading the last match instead, or failing when more than one `export PATH=` is present, closes it.

bin/tests/test-newspack-manage-host-path.sh
```diff
@@ -0,0 +1,171 @@
...
+# property rather than the string is what makes widening visible.
+#
+# Symlinks are followed deliberately. Under usrmerge, /bin and /sbin are
+# symlinks into /usr and carry mode 777 like every symlink, so reading the link
+# rather than its target reports them world-writable and fails on Linux while
+# passing on macOS, where both are real directories. What governs lookup is the
+# target, and a link sitting in root-owned / cannot be replaced by a user.
+dir_root_only() {
+    local d="$1" owner mode g o
+    [ -d "$d" ] || return 0   # absent dir contributes nothing to lookup
```
adekbadek, line 64: **Suggestion:** An empty PATH element passes `dir_root_only`, but an empty element means the working directory

`[ -d "$d" ] || return 0` is commented "absent dir contributes nothing to lookup". That is right for a directory that does not exist and wrong for the empty string: an empty PATH element is the process's working directory at lookup time, not an absent directory.

With the pin rewritten as `export PATH=:/usr/sbin:/usr/bin:/sbin:/bin` — one stray leading colon — the wrapper resolved `ifconfig` from the caller's working directory and ran a file planted there, while the spec still reported 9 passed, 0 failed. The probe at :134 does not see it either, because it prepends to PATH rather than planting in cwd.

The comment at :41-45 names this check as the one that makes widening visible, and a stray colon is the likeliest shape an accidental widening takes. `[ -z "$d" ] && return 1` ahead of the `-d` test keeps the pass-through for genuinely missing directories while making an empty element red.

Two adjacent inputs behave the same way: a *trailing* colon is dropped by `IFS=: read -r -a` before the loop sees it, and if the pin is ever written quoted (`export PATH="/usr/sbin:..."`) the literal quote characters ride along on the first and last elements, neither of which is a directory.

bin/tests/test-newspack-manage-host-path.sh
```diff
@@ -0,0 +1,171 @@
...
+    local f="$1" funcs builtins tok
+    funcs=$(grep -oE '^[a-zA-Z_][a-zA-Z0-9_]*\(\)' "$f" | tr -d '()' | sort -u)
+    builtins='^(if|then|else|elif|fi|for|while|until|do|done|case|esac|function|return|local|echo|exit|set|export|shift|read|printf|test|eval|source|cd|true|false|break|continue|unset|trap)$'
+    sed 's/#.*//' "$f" \
+      | grep -oE '(^|[;&|(]|&&|\|\||!|\b(if|then|else|do|while|until)\b)[[:space:]]*!?[[:space:]]*[a-zA-Z_][a-zA-Z0-9_.-]*' \
+      | grep -oE '[a-zA-Z_][a-zA-Z0-9_.-]*$' \
+      | grep -vE "$builtins" \
+      | { if [ -n "$funcs" ]; then grep -vxF "$funcs"; else cat; fi; } \
+      | sort -u \
+      | while read -r tok; do command -v "$tok" >/dev/null 2>&1 && echo "$tok"; done
```
adekbadek, line 108: **Suggestion:** The derived helper set is a lower bound, so the resolution check can report clean on a set that never contained the helper

`derive_commands` filters candidate tokens through `command -v` on the machine running the spec. That filter does two jobs at once: it discards tokens that are not commands (case labels, variable names), and it discards real helpers that this particular machine does not have installed. The check at :120 then reports on whatever survived.

The wrapper runs only on macOS; `Shell specs` runs only on ubuntu-latest. Removing `/sbin` from the pin — which breaks `ifconfig`, the wrapper's primary helper — fails this spec on macOS (`got [ ifconfig]`) and passes 9 of 9 on ubuntu:24.04, where `ifconfig` is absent from the image and so never enters the derived set at all.

This is a different claim from the "both resolution checks fail on a Linux runner" one already discussed on the PR. These pass on Linux; the point is that one of them passes while covering less than its name says, and a green job log is consistent with both readings.

A second class of miss is platform-independent: the regex only recognises simple command position. A helper reached through backticks or a brace group is not derived even on macOS — adding a backtick-substituted `mkcert -CAROOT` and `{ mkcert -CAROOT; }` to the wrapper leaves the spec at 9 passed, 0 failed. `$(...)`, pipeline stages and `&&`/`||` chains are all caught. None of these shapes appear in the wrapper today.

Dropping the `command -v` filter and excluding non-command tokens some other way would let the check bite on the runner; alternatively, gating the run on Darwin would at least make the green report what it covered.

bin/tests/test-newspack-manage-host-path.sh
```diff
@@ -0,0 +1,171 @@
...
+    g=$(printf '%s' "$mode" | tail -c 2 | head -c 1)
+    o=$(printf '%s' "$mode" | tail -c 1)
+    case "$g" in [2367]) return 1 ;; esac
+    case "$o" in [2367]) return 1 ;; esac
+    return 0
+}
+
+writable=""
+IFS=: read -r -a PIN_DIRS <<< "$PIN"
+for d in "${PIN_DIRS[@]}"; do
```
adekbadek: **Nit:** Deleting the pin line aborts the spec on macOS's system bash before seven checks run

With the `export PATH=` line removed — the single regression this spec exists to catch — `$PIN` is empty, `IFS=: read -r -a PIN_DIRS` leaves a zero-length array, and `"${PIN_DIRS[@]}"` under `set -u` is an unbound variable on bash 3.2, which is `/bin/bash` on macOS. The run prints one FAIL and then `line 73: PIN_DIRS[@]: unbound variable`; the hijack probes are never reached. On bash 5.2 the same mutation reports 6 passed, 3 failed.

Exit status is non-zero either way, so CI is not misled. The cost is a confusing local run at the moment the spec has the most to say — and the file header's run instruction is a bare `bash bin/tests/...`, on the one spec here whose subject exists only on macOS. `PIN_DIRS=()` before the `read`, or guarding the loop on `${#PIN_DIRS[@]}`, restores the full report.

bin/tests/test-newspack-manage-host-path.sh
```diff
@@ -0,0 +1,171 @@
...
+      | while read -r tok; do command -v "$tok" >/dev/null 2>&1 && echo "$tok"; done
+}
+
+# Control for the derivation itself. If it silently stopped matching, it would
+# return nothing and every check below would pass while testing nothing.
+#
+# The probe command is created here rather than borrowed from the system: the
+# derivation drops tokens that resolve nowhere, so naming a real command would
+# make this control depend on that command being installed. An earlier version
+# used openssl and reported "not detected" on a bare Ubuntu image, which is the
```
adekbadek, line 117: **Nit:** The control's comment records the spec's own history

"An earlier version used openssl and reported 'not detected' on a bare Ubuntu image" describes how this control used to be written. Repo convention is that comments describe how things are; that belongs in the PR thread. The sentence before it already carries the constraint — naming a real command would make the control depend on that command being installed — so this one can go without losing anything.

adekbadek (approved):
Release risk: LOW — no blockers, no contract surface, and the shipped pin is correct as written. The comments are about the guard around it rather than the change itself: three edits that undo the pin leave the spec green, and the pin stays inert on any machine that already ran the installer until someone re-runs it.

bin/newspack-manage-host
jason10lee: Reproduced before changing anything: pin pointed at an empty directory, three
`host-add` runs, three exits of 0 and three identical lines in the hosts file.

`host-add` now reads grep's status rather than testing it in a condition — 0 is
present, 1 is absent, anything else is reported and exits 1. Same reproduction
after: three runs, three exits of 1, nothing written.

The paragraph you flagged claimed the general case. It now names the exception
instead: a command in a condition is
exempt from `set -e`, so its 127 becomes an ordinary false.

bin/newspack-manage-host
jason10lee: Narrowed. The comment still says why the grant must not name a checkout path,
but no longer rests that on the chown: it now states that replacement is
governed by the containing directory, that `/usr/local/bin` is owned by the
login user wherever Homebrew created it, and that the checkout path is the worse
of the two without the installed copy being unswappable.

bin/newspack-manage-host
jason10lee: This is the one that decided how much to do here, because a security fix that
skips any environment where the installer was run is not much of one.

`bin/_common.sh` gains `warn_if_manage_host_stale`, called from both `n` and
`bin/env.sh` at the point they shell out to the wrapper: it compares the
installed copy against the checkout's and points at `bin/setup-networking.sh`
when they differ. It warns rather than re-copying, because that file is
root-owned and a silent re-copy would need sudo at a moment the user did not
ask for it.

bin/tests/test-newspack-manage-host-path.sh
jason10lee: Fixed by counting first: the spec now fails when the wrapper carries more than
one `export PATH=`, and reads the last rather than the first. Your mutation —
appending `export PATH=/opt/homebrew/bin:$PATH` on the following line — went
from 9 passed, 0 failed to 7 passed, 3 failed.

bin/tests/test-newspack-manage-host-path.sh
jason10lee: Fixed, and this was the sharpest of the set: an empty element really is the
working directory, and `[ -d "" ]` reading as "absent" is what let it through.
`dir_root_only` now rejects an empty element before the `-d` test. Your
mutation — one stray leading colon — went from 9 passed, 0 failed to 9 passed,
1 failed.

bin/tests/test-newspack-manage-host-path.sh
jason10lee: Taken. The spec now names the helpers it expects (`ifconfig grep sed`), checks
that derivation finds each one this platform can see, and prints an explicit
SKIP line naming any it cannot, so a partial run says so instead of reporting a
clean pass.

Removing `/sbin` now fails on macOS. On a machine without ifconfig it still
reports 11 passed, 0 failed — that check genuinely cannot run there — but the
run now ends with `SKIP not installed here, so their resolution is unverified:
ifconfig`, which is the difference between an unverified check and an invisible
one.

bin/tests/test-newspack-manage-host-path.sh
jason10lee: Fixed with `"${PIN_DIRS[@]:-}"` and a comment saying why the fallback is there.
Confirmed on bash 3.2.57, the interpreter you named: with the pin line deleted
the run now completes at 5 passed, 5 failed instead of aborting on
`PIN_DIRS[@]: unbound variable` with seven checks unreached.

bin/tests/test-newspack-manage-host-path.sh
jason10lee: Rewritten to state what the control establishes rather than how it came to
exist.

jason10lee:
A note on sequencing:

`n` carries two pre-existing `SC2145` errors at the `sh -c "…${@:2}"` calls in
`sites-list` and `sites-drop`. They are not from this branch — with this PR's
one-line change to `n` stashed, shellcheck reports them at the same two lines.
#861 fixes them, by moving those call sites to direct argv.

That means both PRs now touch `n`, so whichever lands second will want a look at that file
rather than a clean merge. And #861 adds the CI gate that makes those errors
fail a build, so the ordering is worth choosing deliberately rather than
discovering.

jason10lee:
Correcting the sequencing comment above. We said whoever lands this and #861
second would want a look at `n` rather than a clean merge. That came from "both
touch the same file", not from an actual merge, and it is wrong.

The two hunks sit about 190 lines apart: this PR adds one line at 199, and #861
rewrites two `docker exec` calls at 392. Merged in either order they produce a
byte-identical tree, `n` included.

Running #861's gate over that merged tree, at the shellcheck version it pins,
comes back clean over all 93 shell files at error severity. So this branch
already passes the gate before the gate exists, and there is no coverage
argument for landing #861 first.

What does still hold is the re-run question. `main` requires only `CI OK` and
does not set a strict status policy, so neither PR's checks re-run when the
other lands. Whichever goes second is worth refreshing before it merges, so its
checks run against a `main` that contains the first.

PR #860, opened 2026-08-13
fix(content-gate): restrict who can read institution records

jason10lee:
### All Submissions:

Reading institution records through the REST API now requires a capability, and the access rules stored inside those records require a higher one.

The post type sets `show_in_rest` and maps every capability to `manage_options`. That map governs writes and did not reach reads. A `WP_REST_Posts_Controller` subclass, registered through `rest_controller_class`, supplies the read gate.

There are two tiers because the consumers need different things. Staff who set gating rules need institution names to populate a dropdown; only administrators need the domains, IP ranges and reader-data rules. So reads require `edit_others_posts`, and every registered meta field is removed unless the caller also holds `manage_options`. Removing all of them, rather than the three rule fields by name, keeps a field added later private by default.

**One visible change for support to know about.** Editors who previously saw an empty institution dropdown in the block editor will now see it populated. The panel is gated on `edit_others_posts`, but its fetch used an edit context that core gated on `manage_options`, so the list came back empty with no error. Admitting `edit_others_posts` at the REST layer repairs it. The underlying capability mismatch is tracked in NPPM-3128.

**This closes one half of NPPM-3084.** The access rules now require administrator rights. Institution *names* stay readable through a separate unauthenticated route, tracked in NPPM-3127 and out of scope here.

**The branch carries one unrelated commit on purpose.** `a53c5acf6` adds a static-cache reset to `class-premium-newsletters.php`, matching a pattern five sibling test files already use. It settles an intermittent, ID-sequence-dependent failure that this branch's new tests would otherwise keep triggering. Squash merge collapses its subject into this PR's title, so it is named here instead.

NPPM-3084.

1. On a site with the content gate enabled, create an institution in the Audience wizard with an email domain and an IP range.
2. Log out, then request `/wp-json/wp/v2/np_institution`. The response is a 401.
3. Log in as an Editor and request it again. The response is 200 and lists the institution. Its `meta` object is empty.
4. Log in as an administrator and request it again. The `meta` object now carries the email domain, IP range and reader-data values.
5. As an Editor, open a post in the block editor, add a Group block, and open the content-gate visibility panel. The institution dropdown lists the institution. Before this change it was empty.
6. In the Audience wizard, edit that institution and save it. The values persist.
7. Run `n test-php` from `plugins/newspack-plugin`. The suite passes.

### Other information:

* [x] Have you written new tests for your changes, as applicable? 20 tests in `tests/unit-tests/content-gate/class-institution-rest-controller.php`.
* [x] Have you successfully run tests with your changes locally? 2390 tests, 6782 assertions, 0 failures.

<details>

**Why the controller carries four methods rather than two, and why the test file has two unusual constructions, is explained in docblocks at each site.** They exist for reasons that are not obvious from the surrounding code, so please read them before simplifying any of them.

**Writes are unchanged and still derive from the post type's capability map** rather than from any capability restated in the controller. Three tests pin the refusals for update, create and delete.

**Completeness.** This route is the only one serving the post type. Core search excludes it. Nothing embeddable on a post carries the parent's meta, and no taxonomy is registered here. Two sibling routes in `class-ip-access-rule.php` also carry rule data, and both require `manage_options`; the third is the unauthenticated one tracked in NPPM-3127.

</details>


plugins/newspack-plugin/includes/content-gate/class-institution.php
```diff
@@ -74,17 +74,23 @@ public static function register_post_type() {
...
+				'label'                 => __( 'Institutions', 'newspack-plugin' ),
+				'public'                => false,
+				'show_ui'               => false,
+				'show_in_menu'          => false,
+				'show_in_rest'          => true,
+				'supports'              => [ 'title', 'excerpt', 'thumbnail', 'custom-fields' ],
 				/**
 				 * Institutions effectively grant access, so restrict all CRUD operations
 				 * (including via REST) to the `manage_options` user capability.
 				 */
```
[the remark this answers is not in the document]
jason10lee: Fixed in `6812bb51d`. The capability map's comment now says it governs writes, and points at the controller for the read requirement rather than contradicting it.

plugins/newspack-plugin/includes/content-gate/class-institution-rest-controller.php
```diff
@@ -0,0 +1,189 @@
...
+	 * get_value() returns — always serializes as an object, so leaving this
+	 * one as an array would make the field's JSON type vary by caller: valid
+	 * against the schema for one tier, a violation for the other.
+	 *
+	 * @param \WP_Post         $item    Post object.
+	 * @param \WP_REST_Request $request Request object.
+	 * @return \WP_REST_Response
+	 */
+	public function prepare_item_for_response( $item, $request ) {
+		$response = parent::prepare_item_for_response( $item, $request );
```
adekbadek: **Suggestion:** prepare_item_for_response() assumes the parent returned a WP_REST_Response — The parent's last statement is `apply_filters( "rest_prepare_{$this->post_type}", $response, $post, $request )`, and a filter can return anything. Core tolerates that — which is why `prepare_response_for_collection()` carries an explicit `instanceof WP_REST_Response` guard. Here `$response->get_data()` runs unguarded, so a plugin filter returning an array or a `WP_Error` turns the route into a fatal.

It fails closed rather than leaking, so it is not urgent, but on code headed for `release` it is a one-line hardening: `$response = \rest_ensure_response( parent::prepare_item_for_response( $item, $request ) );` with an `is_wp_error()` early return.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-institution-rest-controller.php
```diff
@@ -0,0 +1,471 @@
...
+		wp_set_current_user( $this->editor_id );
+		$request = new WP_REST_Request( 'GET', $this->route );
+		$request->set_param( 'context', 'edit' );
+		$request->set_param( '_fields', 'meta' );
+
+		$response = rest_filter_response_fields( rest_do_request( $request ), rest_get_server(), $request );
+		$data     = $response->get_data();
+
+		$this->assert_meta_withheld( $data[0]['meta'] );
+	}
```
adekbadek, line 545: **Suggestion:** Two collection-level guarantees sit outside anything the suite asserts — The strip resists a third-party filter re-adding the rules only because it runs *after* `rest_prepare_np_institution`. I verified that with a filter setting `meta.np_institution_ip_range` to a marker value: the Editor still received `meta: {}` while the administrator received the marker. Nothing pins the ordering, so swapping the two statements in `prepare_item_for_response()` would leave all 20 tests green.

Separately, an Editor is refused `?status=draft`, `?status=private` and `?status=any` — but by core's `sanitize_post_statuses()`, gated on this post type's `edit_posts`, not by anything in this class. Since `get_items_permissions_check()` is now replaced wholesale, that refusal is an undocumented dependency rather than a stated guarantee, and a later `get_collection_params()` override or `rest_np_institution_query` filter could open draft enumeration to the read tier with the suite still green.

Two short tests would close both — a `rest_prepare_np_institution` filter that tries to re-add `meta` and asserts `assert_meta_withheld()` still holds, and an Editor `?status=draft` expecting a refusal. The first is the higher-value of the pair.

plugins/newspack-plugin/includes/content-gate/class-institution-rest-controller.php
```diff
@@ -0,0 +1,189 @@
...
+	 * Mirrors the capability gating the block editor panel that consumes this
+	 * route, so the gate grants exactly what the only non-administrator consumer
+	 * needs and nothing wider.
+	 *
+	 * @var string
+	 */
+	const READ_CAPABILITY = 'edit_others_posts';
+
+	/**
+	 * Capability required to write to institutions, matching the post type's
```
adekbadek: **Nit:** RULES_CAPABILITY's docblock opens by naming it the write capability — The first sentence calls this the capability required to write to institutions, and the third concedes it is not consulted by the write gate. Its load-bearing uses are the read gate and the field tier in `prepare_item_for_response()`; writes take their value from the capability map in `Institution::register_post_type()`. Someone changing which tier sees the rule fields will read this as the write capability and hesitate — or change it believing they have changed the write gate. Describing it as the capability that admits the stored rule fields would leave the capability map as the single statement about writes.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-institution-rest-controller.php
```diff
@@ -0,0 +1,471 @@
...
+		$this->assertSame( 401, $response->get_status() );
+	}
+
+	/**
+	 * A logged-in caller with neither capability cannot read the collection.
+	 *
+	 * Every logged-in WordPress role, Subscriber included, holds the primitive
+	 * 'read' capability, so this — not the logged-out tests above — is what
+	 * proves the gate checks READ_CAPABILITY specifically rather than merely
+	 * "is someone logged in". Mutation #1 in the report shows this directly:
```
adekbadek: **Nit:** Docblock cites an artifact that is not in the repo — "Mutation #1 in the report" points at something a future reader has no way to find. The sentence before it already makes the point on its own.

adekbadek (approved):
Pre-approving with suggestions. 

The comments below are hardening and test-coverage asks; none of them blocks the hotfix.

- `plugins/newspack-plugin/includes/content-gate/class-institution.php:45` — **Suggestion:** A read-side meta guard is available, and would make the gate two-layered — The whole gate rests on `rest_controller_class` surviving registration. `register_post_type_args` runs before `WP_Post_Type` is constructed, so any plugin that rebuilds the args array drops the key, and `get_rest_controller()` then falls back to the default controller with no warning — reopening this exact disclosure silently. The ticket's conclusion that the meta registration cannot help is half right. `auth_callback` really is write-only: `WP_REST_Meta_Fields::get_value()` performs no capability check at all. But `show_in_rest` also accepts an array, and `prepare_value_for_response()` invokes the registered `prepare_callback` on every read. `'show_in_rest' => [ 'prepare_callback' => [ __CLASS__, 'redact_for_unauthorized' ] ]`, returning `''` without `manage_options`, gives a controller-independent read filter that survives both a hijacked registration and any future controller swap, in about three lines. **Consider:** if a `register_post_type_args` filter drops the key, the route silently becomes publicly readable again with nothing failing or logging; if instead the controller class is absent at registration time, core returns no controller and the route is never registered, so the Audience wizard's institutions list and the block editor's institution selector both stop loading. A meta-layer guard closes the first direction, and a test asserting an anonymous request is refused would catch either.

PR #861, opened 2026-08-13
ci: gate shell scripts on shellcheck errors

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? n/a: shell and CI config, no PHP in this change.

`bin/` and `n` have a test job but no linter, so a shell-only change gets no mechanical check. Several scripts already carry `# shellcheck source=` directives, so the tool is in use locally; nothing runs it in CI.

This adds a `shellcheck` step to the existing **Shell specs** job, which is already ungated on changed packages so a tooling-only PR runs it. `ci-ok` depends on that job, so the step gates merges.

It checks 90 files: every tracked `*.sh`, plus tracked extensionless files whose shebang names a shell. `n`, `bin/newspack-manage-host` and the ten per-package `.hooks/pre-push` scripts all carry no extension. An earlier revision of this branch enumerated only `*.sh` plus `bin/`, which silently skipped all ten hooks. The interpreter is matched by name rather than by a trailing `sh`, so a shebang such as `#!/usr/bin/env fish` is not swept in, and the list is NUL-delimited so a path containing whitespace cannot split into two arguments.

**`shellcheck` is pinned** to `v0.11.0` and downloaded in the step rather than taken from the runner image. The image's version changes when it is rebuilt, and a release that adds a check would redden `main` on an unrelated PR. Pinning makes the bump a deliberate commit that carries any resulting fixes with it.

**The floor is `error`, not `warning`, and that is deliberate.** After the fixes below the tree still carries 119 warning-level findings across 39 files, so a warning gate would fail on its first run. Errors are the class worth blocking on today; raising the floor is a cleanup that can happen file by file. The actively-maintained core (`bin/_common.sh`, `bin/env.sh`, `bin/worktree.sh`, `bin/newspack-manage-host`, `bin/tests/`) already passes at `warning`.

**A `.shellcheckrc` comes with it**, because there was none and the flags would otherwise live only in CI, giving anyone running shellcheck locally different results. It sets `external-sources=true` and `source-path=SCRIPTDIR`, which together make the five `# shellcheck source=` directives already in the tree resolve from any working directory: previously `bin/tests/test-lo0-alias-exists.sh` reported SC1091 from the repo root and resolved cleanly only from its own directory.

That file sets no severity on purpose. Pinning it to `error` there would hide the 119 warnings from editors too, which is the opposite of useful when those warnings are the deferred work. CI states its own floor with `-S error`, so the gate level stays a visible policy choice rather than a dotfile default.

**The seven errors it found are fixed here.** Two of them are one real latent bug; the other five are cosmetic:

* `n` built two commands as `sh -c "/var/scripts/sites-{list,drop}.sh ${@:2}"`, interpolating the arguments into a string that a second shell then parsed. Inside a quoted string `"$@"` expands to *separate words*, so arguments after the first became positional parameters of that inner shell rather than part of the command line. Both now invoke the script directly, the same form the `site-setup.sh` and `watch-repo.sh` calls already use elsewhere in the file, so nothing parses the arguments twice.

  This began as a narrower change that joined the arguments with `${*:2}`; review pointed out that it fixed the splitting while leaving them exposed to the inner shell. Checked against a running container: the old form turned an argument of `a;b` into `GOT=a` followed by `sh: 1: b: not found`, and split `has space` into two arguments, while the direct form passes both through intact. `n sites-list` returns the same output as before. The bug was latent either way, since `sites-list` takes no arguments and `sites-drop` takes one.
* `bin/composer.sh`, `bin/npm.sh` and `bin/test-php.sh` used the same pattern in their `echo "Running: …"` lines only. Their execution lines already pass `"${@:2}"` correctly, so this is display formatting.
* `bin/ncd.sh` and `bin/jn-functions.sh` are sourced, not executed, so having no shebang is correct. They get a `# shellcheck shell=bash` directive, which states that intent without changing how they run.

1. Install `shellcheck` (`brew install shellcheck`, or `apt-get install shellcheck`).
2. Run `shellcheck bin/tests/test-lo0-alias-exists.sh` from the repo root. It reports no SC1091, because the new `.shellcheckrc` resolves the `source=` directive. On `main` the same command reports one.
3. Run the same enumeration the CI step runs, against your own `shellcheck`:
   ```sh
   { git ls-files -z '*.sh'
     git ls-files -z | while IFS= read -r -d '' f; do
       case "${f##*/}" in *.*) continue ;; esac
       [ -f "$f" ] || continue
       head -1 "$f" 2>/dev/null \
         | grep -Eq '^#!(/[^ ]+)*/(env +)?(sh|bash|dash|ksh|zsh)( |$)' \
         && printf '%s\0' "$f"
     done
   } | sort -zu | xargs -0 shellcheck -x -S error -f gcc
   ```
   It reports nothing and exits 0. Confirm it lists 90 files by piping the same list through `tr '\0' '\n' | wc -l`.
4. Check out `main` and run the same enumeration. It reports the seven errors described above.
5. Run `./n help`, `./n root` and `./n sites-list`. All behave as before; `sites-list` needs a running container.
6. Confirm the new step appears under **Shell specs** in this PR's checks, and that it reports the file count it checked.

### Other information:

* [x] Have you written new tests for your changes, as applicable? The CI step is itself the check; it fails if the file list comes back empty, so it cannot pass by matching nothing.
* [x] Have you successfully run tests with your changes locally? The step's command runs clean over 90 files and `bash -n` passes on every edited script. I also checked that the gate can fail rather than only pass: reverting one of the fixes below turns it red with that file's error, and it returns to green when the fix is restored. The workflow YAML parses, and the pinned release URL returns HTTP 200.


n
```diff
@@ -391,11 +391,16 @@ case $1 in
...
+    # with ${*} keeps them in the string. Arguments containing spaces are still
+    # unsupported here, as documented for `n wp`.
     sites-list)
-        docker exec $DOCKER_IT $TARGET_CONTAINER sh -c "/var/scripts/sites-list.sh ${@:2}"
+        docker exec $DOCKER_IT $TARGET_CONTAINER sh -c "/var/scripts/sites-list.sh ${*:2}"
         ;;
     sites-drop)
-        docker exec $DOCKER_IT $TARGET_CONTAINER sh -c "/var/scripts/sites-drop.sh ${@:2}"
+        docker exec $DOCKER_IT $TARGET_CONTAINER sh -c "/var/scripts/sites-drop.sh ${*:2}"
         ;;
```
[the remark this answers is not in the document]
jason10lee: Taken — switched both to direct argv rather than joining with `${*:2}`, which as you say only fixed the splitting. `docker exec ... /var/scripts/sites-{list,drop}.sh "${@:2}"` matches the `site-setup.sh` and `watch-repo.sh` calls elsewhere in this file, so nothing parses the arguments a second time.

Verified against a running container: the old form turned `a;b` into `GOT=a` plus `sh: 1: b: not found` and split `has space` into two arguments; the direct form passes both through. `n sites-list` output is unchanged.

The wider `sh -c` pattern is used at 22 other call sites in `n` and I have left those alone as out of scope here.

.github/workflows/ci.yml
```diff
@@ -523,6 +523,59 @@ jobs:
...
+          # Bumping SHELLCHECK_VERSION means replacing this line too.
+          SHELLCHECK_SHA256: 8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198
+        # Errors only. The tree carries 119 warning-level findings across 39
+        # files, so a warning gate would fail on its first run; raising the floor
+        # is a separate cleanup. -x follows sourced files, honouring the
+        # `# shellcheck source=` directives already in the tree.
+        run: |
+          # Downloaded to a file rather than piped into tar: a stream cannot be
+          # checked before it is extracted.
+          curl -fsSL -o shellcheck.tar.xz "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
```
adekbadek: **Suggestion:** The download has no retry or timeout on a step that gates every merge — `Shell specs` is the one job in this workflow with no `if:` gate and no changed-packages filter, and `ci-ok` requires it, so this `curl` sits on the path of every PR in the repo. A transient failure reaching the release asset reddens a check on a PR that may not touch shell at all, with a message that says nothing about the PR; a stalled connection runs to the job's default timeout rather than failing fast.

The other network fetches in this file are equally bare, but they sit behind `steps.check.outputs.has_tests` and the package matrix, so they are not on the universal path this one is.

`--retry 3 --retry-all-errors --max-time 120` covers both cases. The sha256 makes the download safe to retry, and the verify-then-extract ordering is already right.

.github/workflows/ci.yml
```diff
@@ -523,6 +523,59 @@ jobs:
...
+
+          # Every tracked *.sh, plus tracked extensionless files whose shebang
+          # names a shell: `n`, bin/newspack-manage-host and the per-package
+          # .hooks/pre-push scripts all carry no extension. The interpreter is
+          # matched by name rather than by a trailing "sh", so a shebang like
+          # `#!/usr/bin/env fish` is not swept in. NUL-delimited throughout so a
+          # path containing whitespace cannot split into two arguments.
+          {
+            git ls-files -z '*.sh'
+            git ls-files -z | while IFS= read -r -d '' f; do
```
adekbadek: **Suggestion:** `.husky/pre-commit` is the one tracked shell script neither half of the gate reaches — It is extensionless and carries no shebang, so the shebang arm skips it, and `*.sh` in `.lintstagedrc.json` does not match it either. **Verified on the review env:** it is absent from the 91-file list, and it is clean at every severity today, so nothing is currently hidden.

It is worth closing anyway because of which file it is. Husky runs it on every commit in the repo, and a syntax error there does not fail loudly — it takes ESLint, Stylelint, PHPCS and now shellcheck down with it, silently, for everyone. That makes it the single script where this gate would pay for itself most.

The fix used for `bin/ncd.sh` and `bin/jn-functions.sh` does not help, since the file never reaches shellcheck to read a directive. Adding `#!/bin/sh` as its first line pulls it in through the existing shebang arm at no behavioural cost — the file is mode 644 and husky invokes it as `sh <file>`, so the line is parsed as a comment. Adding `git ls-files -z .husky/pre-commit` to the union does the same without touching the file.

.github/workflows/ci.yml
```diff
@@ -523,6 +523,59 @@ jobs:
...
+              case "${f##*/}" in *.*) continue ;; esac
+              [ -f "$f" ] || continue
+              head -1 "$f" 2>/dev/null \
+                | grep -Eq '^#!(/[^ ]+)*/(env +)?(sh|bash|dash|ksh|zsh)( |$)' \
```
adekbadek: **Suggestion:** `zsh` in the shebang alternation can only ever fail the gate — The alternation sweeps in files whose shebang names `sh|bash|dash|ksh|zsh`, but shellcheck does not support zsh — it rejects the file at `error` severity without analysing it, which the `-S error` floor then turns into a failed gate.

**Verified on the review env**, shellcheck v0.11.0:

```
probe-zsh:1:1: error: ShellCheck only supports sh/bash/dash/ksh/'busybox sh' scripts. Sorry! [SC1071] exit 1
probe-ksh exit 0
probe-dash exit 0
```

This is a failure mode the enumeration creates rather than inherits: a `.zsh` file would have escaped the `*.sh` glob entirely, so only an extensionless zsh script picks it up, and its author has no clean remedy — a disable directive cannot help a file shellcheck will not parse. It is latent today; all twelve extensionless matches in the tree are `sh` or `bash`.

The comment at `ci.yml:557` already gives the reasoning that resolves it — the interpreter is matched by name so `#!/usr/bin/env fish` is not swept in. Dropping `zsh` from the alternation applies the same rule to the other shell shellcheck cannot read.

.github/workflows/ci.yml
```diff
@@ -523,6 +523,59 @@ jobs:
...
+                && printf '%s\0' "$f"
+            done
+          } | sort -zu > shell-files.txt
+
+          count=$(tr -dc '\0' < shell-files.txt | wc -c | tr -d ' ')
```
adekbadek, line 657: **Suggestion:** The empty-list guard passes when only one arm of the union goes empty — The guard is a zero-check over a list built from two independent arms, so it catches only an enumeration that produced nothing at all. If the `*.sh` arm alone stops matching — a pathspec typo, a `git ls-files` failure — the extensionless arm still supplies its dozen files, `count` is non-zero, and the step prints `Checking 12 files.` and passes green while 79 scripts silently stop being checked. Nobody reads that number against a remembered baseline.

The usual hardening is not available here, which is worth knowing before reaching for it. The group's exit status is already discarded by the pipe into `sort`, and that discard is load-bearing: when the last tracked path is extensionless and not a shell script, the `while` body ends on the failing `head | grep -Eq && printf` list and the loop returns 1. Adding `set -o pipefail` — the convention the rest of this file uses at `ci.yml:53` — surfaces that benign status as a step failure, and whether it fires depends on which path happens to sort last.

**Verified on the review env:**

```
# same enumeration, list ending in an extensionless non-shell file
set -e -> exit 0, 1 file listed
set -e -o pipefail -> exit 1, no list produced
# same, list ending in a file with an extension
set -e -o pipefail -> exit 0, 1 file listed
```

On the current tree the last tracked path is `themes/newspack-theme/webpack.config.js`, which leaves the loop via `continue`, so the step is green today by accident of sort order rather than by construction.

Checking `${PIPESTATUS[@]}` after the pipe, or asserting that each arm produced output, closes the real gap without depending on that. Terminating the loop body with `|| true` would make its status deterministic if you also want the file's `set -euo pipefail` convention here.

.github/workflows/ci.yml
```diff
@@ -523,6 +523,59 @@ jobs:
...
+          SHELLCHECK_VERSION: v0.11.0
+          # sha256 of shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz. A
+          # release-asset URL keyed on a tag is not an immutable reference: tags
+          # move and assets can be replaced, and this binary then executes in a
+          # job that has the repository checked out. The project publishes no
+          # checksum file, so this records the artifact we verified; the same
+          # digest is pinned by selenium, mise, grafana/alloy and others.
+          # Bumping SHELLCHECK_VERSION means replacing this line too.
+          SHELLCHECK_SHA256: 8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198
+        # Errors only. The tree carries 119 warning-level findings across 39
```
adekbadek: **Nit:** The "119 findings across 39 files" count has no shelf life — It is accurate today — 119 confirmed on v0.11.0 against the branch's 91 files — which is exactly why it will be wrong soon. Nothing recomputes it, so the first shell edit or shellcheck bump moves it and the comment then misstates the tree it is attached to.

The sentence's job — explaining why the floor is `error` and not `warning` — survives without the figures. Either drop them and keep the claim, or say when they were measured. The exact number is preserved in this PR's history for whoever later raises the floor.

AGENTS.md
```diff
@@ -78,6 +78,7 @@ A husky hook runs `lint-staged` on every `git commit` (installed by `pnpm instal
 
 - **Auto-fix** with the package's own script for JS and SCSS: `pnpm --filter <package> run fix:js` / `format:scss`. For PHP use `composer phpcbf -- <path>` from the root instead; the packages' `fix:php` scripts can rewrite files across the whole monorepo (see above).
 - **Bypass** with `git commit --no-verify`, or `HUSKY=0` for one command or a whole shell. CI re-lints every PR, so skipping locally never lands unlinted code.
+- **Shell scope** is staged `*.sh`, gated at `error` to match CI. ShellCheck is a system package rather than something `pnpm install` provides, so a checkout without it **skips** this check with a notice instead of failing — the only linter here that does. CI runs the same gate on every PR and additionally sweeps tracked extensionless shell scripts (`n`, `bin/newspack-manage-host`, the `.hooks/pre-push` files), which `lint-staged` cannot glob. Config lives in `.shellcheckrc`, which leaves the local severity at `style` so editors still surface the warning-level findings the gate defers.
```
adekbadek: **Nit:** "which lint-staged cannot glob" dropped the qualifier that made it true — lint-staged's config keys are micromatch patterns, and a pattern containing a slash matches against the repo-relative path, so `n` and the `.hooks/pre-push` files can be matched by name or path. What it has no way to express is a shebang-aware selector.

`bin/precommit-shellcheck.sh:11` states this correctly — "cannot glob without matching every file in the repo". The AGENTS.md sentence dropped that qualifier. Since AGENTS.md is where contributors and agents read the conventions from, it is the copy worth having right.

bin/precommit-shellcheck.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+# Pre-commit ShellCheck helper (invoked by lint-staged for staged *.sh files).
+#
+# Gates at `error` only, matching the CI step in .github/workflows/ci.yml, so the
```
adekbadek: **Nit:** "matching the CI step" holds for the severity floor, not the version — The floor does match. The version does not: CI pins v0.11.0 and the comment at `ci.yml:527` gives the reason — a release that adds a check would redden `main` on an unrelated PR. The hook runs whatever the developer's package manager installed, checked against nothing, so a newer local build can block a commit CI would pass and an older one can miss an error CI will catch.

That is a fine trade-off to make, since shellcheck is a system package the repo cannot provision and the hook is a convenience with CI as the real gate. It is the sentence that needs narrowing, here and at `AGENTS.md:81`: say the severity floor matches CI and the local version is unpinned. Comparing `shellcheck --version` against the pin and warning on a mismatch is the alternative if the drift is worth surfacing.

n
```diff
@@ -391,11 +391,16 @@ case $1 in
         fi
 
         ;;
+    # No `sh -c` wrapper here: the arguments go straight into the container
```
adekbadek: **Nit:** The comment reads as a rule the file mostly does not follow — Five lines for two, three of which argue against a form these two lines do not use. On the merged branch a reader arrives at it with `sites-add` ten lines above at `n:384` and `build` at `n:409` both doing the interpolation the comment argues against, so it reads as a rule the file violates rather than a property of the code — and it raises a question it does not answer. That the other call sites stay as they are is settled and not what this is about.

The precedent it cites checks out: `n:361` and `n:430` do pass direct argv. What a reader of these two lines needs is the first sentence — arguments go straight into the container command's argv, so no second shell parses them. The rest is a reply to a review objection, which the PR thread already holds.

Placement is also slightly off: sitting between `sites-add`'s `;;` and the `sites-list)` label, it reads as governing only `sites-list`, though `sites-drop` takes the same form.

adekbadek (approved):
Release risk: LOW — no blockers; no product surface (no `plugins/`, `themes/` or `packages/` files touched); the gate was exercised end to end, including against a fresh merge into current `main`, where it passes over 98 files.

Approving. The gate does what it says and the seven fixes hold up under a real container. The comments are about failure modes the gate could develop later, plus some prose that will age; the one worth reading first is on the empty-list guard.

.github/workflows/ci.yml
jason10lee: Taken as written: `--retry 3 --retry-all-errors --max-time 120`, with a comment
saying why this fetch gets them and the others do not.

.github/workflows/ci.yml
jason10lee: Right about the gap, and the file is now checked — but not by adding a shebang.
Husky's own deprecation notice asks for that line to be *removed* from hook
files and says it "WILL FAIL in v10.0.0", so adding one would trade a coverage
gap for a version landmine on the script that runs on every commit.

What works instead is `# shellcheck shell=sh` as the first line, which is the
remedy shellcheck itself suggests: including the file bare fails
`SC2148: Tips depend on target shell and yours is unknown. Add a shebang or a
'shell' directive.` The directive answers that without reintroducing the line
husky wants gone. The enumeration gains a third, explicitly-named arm for shell
scripts carrying neither an extension nor a shebang. Gate now passes over 92
files rather than 91.

.github/workflows/ci.yml
jason10lee: Reproduced on shellcheck 0.11.0, same output you got — SC1071 at error severity
for zsh, exit 0 for ksh and dash. Dropped from the alternation, with a comment
recording that a disable directive cannot rescue a file the checker refuses to
parse.

.github/workflows/ci.yml
jason10lee: Taken. Each arm is enumerated into its own file and checked separately, and the
step prints the per-arm counts, so an arm that stops matching fails the job
rather than reducing a total nobody compares against a baseline.

.github/workflows/ci.yml
jason10lee: Dated rather than dropped: the sentence now reads "119 findings across 39 files
when measured 2026-08-27". Re-measured while making the change and it is still
exactly that over the 91 files the old enumeration produced, which is the last
moment either number will be trustworthy without a date on it.

AGENTS.md
jason10lee: Fixed, and with the reason rather than just the qualifier: lint-staged's keys
are micromatch patterns, so a path can be matched — what it cannot express is a
shebang-aware selector.

bin/precommit-shellcheck.sh
jason10lee: Narrowed to the severity floor, and the version drift is now stated: CI pins one
and the hook runs whatever the developer installed, so a newer local build can
flag what CI passes and an older one can miss what CI catches. Left the
`shellcheck --version` comparison alone — it is a real option, but it puts a
warning in every commit for a hook that is already the fast signal rather than
the gate.

n
jason10lee: Cut to the first sentence and moved inside the `sites-list)` arm, so it reads as
a property of those two lines rather than a rule the neighbours violate.

PR #864, opened 2026-08-13
feat(components): shared StatCard component (DSGNEWS-212)

thomasguillot:
### All Submissions:

Adds `StatCard` to `newspack-components`: a scorecard tile that shows one headline figure with a label above it, an optional change beside it, and a description below.

Pricing Rules and Insights each grew their own copy of the same tile, and the two drifted. Both settled on a figure that shrinks to fit its column, and both stand a glyph in for "no number", but only one of them names that glyph to screen readers in a way NVDA and VoiceOver read as text rather than a graphic. This is the shared anatomy, with that treatment as the default.

The component owns chrome, layout, the type scale and the null glyph. It owns no policy: the caller formats the figure and passes a string, so currency, locale and abbreviation stay with the screen that knows what the number means. `StatCard.Delta` holds the same line. The component draws the arrow and colours it, while the caller sets the direction and the tone separately, because a rise is not always good news and a refund rate climbing 2% wants an up arrow with a negative tone. Insights' `MetricCard` keeps its overlays, warming states and period-over-period maths, and composes these slots.

Seven slots: `Root`, `Label`, `Body`, `Value`, `Delta`, `Secondary` and `Footer`. Layout comes from `Stack` in `@wordpress/ui`, the same package the card chrome comes from, so the stylesheet carries type, colour and the container query rather than flex plumbing.

The words the card speaks for itself are the consumer's to set. `labels` on `StatCard.Root` replaces "Not applicable", "Up" and "Down" for every card underneath, which a bundle registered against another text domain needs because it never loads this package's strings. Every slot forwards a ref and passes unlisted props to the element it renders, so a caller can hang a `title` carrying the unabbreviated amount off an abbreviated `$1.2M` without wrapping the figure.

The hero step, 48px with a 1.125 ratio, lives in a new `packages/components/src/_variables.scss`. That is the package's first shared variables sheet, for the steps it needs above the `@wordpress/base-styles` scale. It holds one pair today and is the place to put the next one, so components stop reaching for bare pixel values.

Nothing consumes it yet. Adoption follows in Pricing Rules and Insights.

Closes DSGNEWS-212.

StatCard in the components demo: a figure with a green rise, a figure with a red fall and an info button, the null glyph, and the text variant

Four-up at 1512px puts each card at about 205px, where the container query holds the figure near 29px. The 48px ceiling is reached once a card passes roughly 343px, so on two-up, three-up or a wide screen.

1. Build the package and the plugin: `pnpm run build:packages && pnpm --filter newspack run build`.
2. Open **Newspack › Components Demo** in wp-admin and scroll to **StatCard**. Four cards render in a row.
3. Narrow the browser to about 900px. Each figure shrinks with its column instead of overflowing or wrapping, and stops at 20px rather than shrinking under its own label.
4. Turn on VoiceOver and move through the third card. The dash is announced as "Not applicable", not as an image.
5. Move through the second card. The figure is announced as "1.2 million dollars", the info button is not part of the heading, and its change reads "Down 4%" rather than as a glyph.
6. Compare the first two cards. The rise is green and the fall is red, and each keeps its arrow, so the direction survives without the colour.
7. Confirm the descriptions sit on one line across all four cards, whatever the body above them contains.
8. Check the info button on the second card. It sits on the first line of the label without pushing the row taller, and the call site applies no class of its own to get that.

### Other information:

This adds to the `newspack-components` public contract, so it lands in the next package release and reaches the plugins that depend on it. `STAT_CARD_NULL_GLYPH` is exported alongside the component so tables under a row of cards can show the same glyph.

The package publishes no TypeScript declarations: `main` points at the compiled output and there is no `types` field. A consumer installing from npm, `newspack-manager` among them, sees an untyped module, so the dev-mode console warnings are the enforcement outside this monorepo.

`lint:scss` fails on `packages/colors/colors.module.scss` and `packages/components/src/tabbed-navigation/style.scss`. Both are pre-existing on `main` and untouched here.

<details>

**Slots.** `Root` renders `Card.Root` from `@wordpress/ui` and owns `container-type: inline-size`. `Label`, `Body` and the figure row are `Stack`s; `Content` renders as a `Stack` through its `render` prop, so the card gains no wrapper element for its layout. `Body` takes the free space, which is what pins `Footer` to the bottom. Every part reads Root's context, and outside it throws in development and in tests, because a `Value` rendered loose would size its `14cqi` against whichever container it landed in and fail quietly. A production build warns and falls back to the default context instead: nothing in this package or in `newspack-plugin` puts an error boundary above these cards, so a mis-sized figure is the smaller failure of the two.

**Root declares its own `display: flex`,** duplicating what `Card.Root` already does. That rule sits inside `@layer wp-ui`, so the footer-pinning contract would otherwise rest on an upstream default that is free to move.

**`suffix` sits beside the heading, not inside it.** Inside the `<h3>` its text would join the heading's accessible name and the document outline. The card trims the 2px a 24px control adds to a 20px line, scoped to the label slot and derived from `$button-size-small` and `$font-line-height-small`, so a call site has no class of its own to remember and nothing to forget.

**The null glyph and the delta arrow are `aria-hidden`, with their meaning supplied as text** in a `VisuallyHidden` from `@wordpress/ui`, which brings its own CSS. ARIA prohibits naming a generic element, so `role="img"` is what makes an `aria-label` survive there, and it also makes screen readers announce "graphic" for a typographic placeholder. Hiding the glyph and supplying real text avoids both, and asks nothing of the host page: on wp-admin's `.screen-reader-text` every consumer's stylesheet would have been part of the contract. Insights currently uses the `role="img"` form in `MetricCard` and `SortableTable`; those converge on adoption.

**Spoken defaults ride the same context as the heading level.** `labels` on `Root` takes `{ notApplicable, up, down }` and is resolved per render, so a locale switched after the bundle evaluates still reaches it. `valueLabel`, `directionLabel` and `label` remain the per-instance override and still win. A blank entry anywhere falls back to the built-in string, so an empty translation cannot leave the glyph or the arrow announcing nothing.

**Every part forwards a ref and passes unlisted props through** to the element it renders, `Value` onto the figure rather than the row it shares with a `suffix`. Insights hangs a `title` carrying the unabbreviated amount off an abbreviated `$1.2M`; wrapping the figure to get that back would add a flex item to the body `Stack`.

**Delta colours** are `$success-600` and `$error-600` from `packages/colors`, both around 6.3:1 on the card. The `$alert-green` in `@wordpress/base-styles` is about 2.5:1 on white, which misses even the large-text floor, so it is not usable for text at this size.

**`variant="text"`** keeps the slot and drops the display scale, for a hero that is a phrase rather than a number ("0 of 17"). Without it a sentence wraps across three lines.

**A blank figure is a missing one.** `null`, `undefined` and a string that is empty or only whitespace all render the null glyph, so a field reporting "no data" as a padded sentinel gets the glyph rather than an empty hero. A zero is a figure and stays out of it.

**`Footer` gathers a run of text children** into one `<p>`, so `Applies to { count } products` is one sentence rather than three stacked paragraphs. Elements pass through untouched, which is how an action lands under the text. A Fragment is part of the run, since `createInterpolateElement` returns one; that keys on the wrapper rather than what it holds, so a group of non-text content needs a real element around it to stay out of the description styling.

**The card needs a definite inline size.** Inline-size containment means it contributes nothing to its own intrinsic width, so a grid track or a `flex: 1` item is fine but a shrink-to-fit parent collapses it.

Full API in [`packages/components/src/stat-card/README.md`](packages/components/src/stat-card/README.md).

</details>


thomasguillot:
## Code review

**Release risk: ELEVATED**, on one factor only: this adds `StatCard` and `STAT_CARD_NULL_GLYPH` to the `newspack-components` barrel, which 84 files across 8 units import. The change is additive and existing callers are unaffected, and there are no unresolved findings. Any PR that adds an export to the shared package carries this tier.

**What I verified.** 18 suites and 260 tests pass, including the new StatCard suite; `typescript:check` is clean; the plugin's `lint:js` reports 0 errors against 27 pre-existing warnings; Stylelint is clean; and a full package plus plugin build produces the expected rules in `commons.css`. The footer-pinning contract was checked against the CSS `@wordpress/ui` actually ships rather than assumed, as were the card's token fallbacks. The container-query hero scale, the null-glyph accessible name and the compound context guard all hold. The barrel addition brings in no new package surface, since `Page`, `Breadcrumbs` and `TabbedNavigation` already import `@wordpress/ui`.

**Addressed on the branch.** Four commits sit on top of the original:

- `valueLabel=""` disabled the accessible-name mechanism entirely. `??` let the empty string beat the default, then the truthiness check dropped the labelling branch, leaving a bare unnamed dash. Insights' `MetricCard` already uses `||` in the same position for this reason.
- `Footer` wrapped each child separately, so `Applies to { count } products` rendered as three stacked paragraphs. It now gathers a run of text into one `<p>`, matching `Drawer.Content`.
- `value` rejected `undefined` in TypeScript while the runtime, the tests and the README all supported `value={ data?.count }`. Insights' `MetricCard` declares `value?: number`, so it would have failed on its first call site.
- `Label` built its heading tag by string interpolation, so `heading={ 7 }` produced an `<h7>`, which carries no heading role. It now uses a lookup map with an `h3` fallback and a development warning.
- `Root` forwards a ref and passes `id`, `style` and `data-*` through, so a wrapper can reach the node. Prop types are exported, and the README records that they travel only through a path import.
- The stylesheet gained a floor on the hero figure, a text variant that is no longer smaller than its own label, matched line-height pairings, spacing that holds when a card has no label, an action rule that does not depend on stylesheet order, and its own accent and screen-reader-text rules, so the card no longer leans on ambient admin CSS.
- The demo's info control was a `<span>`, which cannot carry an accessible name. It is a labelled button now.

**Left alone, deliberately.** `justify-content: space-between` on the label row parks the info control at the card's right edge, roughly 180px from the text it qualifies on a narrow card. That is a layout call rather than a defect, and worth your eye before it becomes the standard for both consumers. The label row also stays a BEM flex row rather than `HStack`, matching `Drawer`, because the class is part of the styling contract consumers target.

thomasguillot:
## Final pass

No findings were left outstanding, so this pass was a comment audit across the component. No behaviour changed.

Comments that restated the code are gone and what remains explains why. `Body` no longer carries a note about reading the context for its invariant, so `Body`, `Secondary` and `Footer` now read the same way. The footer batching note, the heading fallback note and the container-query note each carry one idea instead of two, and the ARIA note beside the null glyph is down to a single line. Two test comments repeated rationale already stated in the source and were removed; three more were tightened to one line.

Seven comments remain in the source, each naming something a reader could not infer from the code:

- why `Footer` gathers a run of text into one paragraph
- why an out-of-range heading level falls back rather than rendering the tag it was given
- why a `Value` rendered outside a `Root` would size against the wrong container
- why the empty-label check uses `||` and not `??`
- why `Card` is aliased on import
- why the parts hang off a namespace object rather than this package's flat exports
- why the null glyph is hidden and named in text rather than given `role="img"`

The stylesheet carries none, per the house rule. The prop-type JSDoc stays, since it surfaces in editor tooltips, as do the import-group headers every file in the package uses.

Still green: 18 suites and 260 tests, `typescript:check`, `lint:js` at 0 errors against 27 pre-existing warnings, Stylelint, and a full package plus plugin build.

thomasguillot:
## Two changes since the review

`StatCard.Secondary` now takes the figure's colour and `heading-medium()`, so it reads as part of the headline rather than a note under it. The footer description keeps the quiet grey, and remains the muted line.

`StatCard.Label` and `StatCard.Footer` are now built from `Stack` in `@wordpress/ui` rather than hand-rolled flex rows. That supersedes the line in the review summary above about the label row staying a BEM flex row. `Stack` comes from the same package as `Card`, needs no experimental API, and its gap tokens ship fallbacks, so spacing holds outside the design-token sheet the same way the card chrome does. The BEM classes stay for consumers to target; they simply carry no flex declarations of their own now.

`Root`, `Content` and `Body` keep their own flex. Two of them are the card primitive's own elements, and `Body` leans on `flex: 1` to pin the footer, which is the documented exception to the layout rule.

260 tests, `typescript:check`, `lint:js`, Stylelint and a full build all still pass.

thomasguillot:
## What adopting this in Pricing Rules costs

Checked `StatCard` against `pricing-rules/stat-tile.tsx` and its two callers. The mapping is 1:1, and the two stylesheets are clearly written from the same source.

### Slot mapping

| `StatTile` prop | `StatCard` |
| --- | --- |
| `label` | `StatCard.Label` |
| `value: string \| null` | `StatCard.Value value` |
| `valueLabel` | `StatCard.Value valueLabel` |
| `description` | text child of `StatCard.Footer` |
| `secondary` | `StatCard.Secondary` |
| `actionLabel` + `onAction` | a `<Button>` element child of `StatCard.Footer` |

Nothing in `StatTile` lacks a slot, and nothing needs a workaround.

### Already identical, not merely compatible

* **Null glyph.** `EM_DASH` in `impact-format.ts` and `STAT_CARD_NULL_GLYPH` are the same character.
* **The "Not applicable" fallback.** Same string, same `_x` context, same `aria-hidden` plus `.screen-reader-text` pairing. `StatCard.Value` is slightly better: `||` rather than `??`, so an empty label cannot leave the glyph unnamed.
* **Heading level.** `StatTile` hardcodes `h3`; `StatCard.Root` defaults `heading = 3`. Same output, now configurable per card.
* **Label styling.** Both `@include wp-mixins.heading-large()`, `$gray-900`, `margin: 0`.
* **Card metrics.** Both `min-height: $grid-unit * 21`, both `container-type: inline-size`, both tabular figures at `$font-weight-medium` in `var(--wp-admin-theme-color)`, both growing at `14cqi`.

### Three differences, all visual

1. **The figure ceiling goes 44px to 48px.** Pricing Rules is `min(44px, 14cqi)`; this is `clamp($font-size-x-large, 14cqi, $font-size-3x-large)`. Same growth rate, higher ceiling, plus a 20px floor the local rule lacks. Four-up at 1512px puts each card near 205px, where `14cqi` lands around 29px and the two render the same. The difference shows only at two-up, three-up, or on a wide screen.
2. **The figure colour changes.** `StatCard` re-points `--wp-admin-theme-color` to `$primary-600` on the card, so the number becomes Newspack primary instead of inheriting the admin colour scheme. The most visible change on both screens.
3. **Vertical rhythm moves to `Stack` gaps.** Pricing Rules carries `padding-top: $grid-unit; gap: $grid-unit` on its body; here it comes from `Stack` (`sm` on the content, `xs` in `Body` and `Footer`). `StatCard.Footer` also merges adjacent text into one `<p>` while keeping elements separate, so a description plus an action button composes rather than stacking as bare siblings.

### Scope

Small. `stat-tile.tsx` goes, and `impact-stats.tsx` builds slot markup instead of `StatTileProps` objects, which is roughly a 30-line change since the tile array survives. The `Grid` wrapper stays as it is: `StatCard` supplies the card, not the layout above it.

The tail is larger than the swap. Around 50 lines of `__tile*` rules come out of `pricing-rules/style.scss`, `stat-tile.test.jsx` either goes or folds into `impact-stats.test.jsx`, and both surfaces want a visual pass for the colour and scale changes above.

### Sequencing

This branch is off `main`; the Pricing Rules tile is six PRs deep in the stack ending at #852. So adoption is a follow-up branch once both have landed, not something to fold into either side.

No release gate for this step: `packages/components` is a workspace symlink, so an in-monorepo consumer picks it up immediately. Insights is the one waiting on a package release.

thomasguillot:
## Code review

Reviewed at `ab256f328`.

**Release risk: ELEVATED**, on one factor: this adds `StatCard` and `STAT_CARD_NULL_GLYPH` to the `newspack-components` barrel, which 89 files across eight units import. The change is additive and existing callers are untouched. Any PR that adds an export to the shared package carries this tier, so it is the floor here rather than a signal about the code.

**What I verified**, against a local site serving this branch. All four demo cards render with the intended screen-reader treatment: the delta reads "Up 2%" with the arrow hidden, the abbreviated figure is hidden behind its spoken form, and the null glyph is hidden behind "Not applicable". Footers pin level across the row, there is no horizontal overflow, and the RTL stylesheet is identical to the LTR one. No console output traces to StatCard. In the production bundle the development throw in `useStatCardContext` is dead-code-eliminated, leaving the documented fallback and its warning, so dropping the throw in production behaves as described rather than only as intended. The published output carries the component's stylesheet and the new variables partial in both `dist/esm` and `dist/cjs`, so the path Insights will take resolves.

**Three things came back clean** that were worth checking. The new sheet is scoped entirely under `.newspack-stat-card`, and the eight theme-colour declarations inside it duplicate, at lower scope, values `src/style.scss` already sets on `:root` for the whole barrel, so nothing new reaches a consumer's controls. `@wordpress/ui` was already a dependency and is already imported by six barrel-exported components. And a 229KB gap between this branch's `commons.js` and the main checkout's turned out to be the two being built from different branches rather than a dropped dependency: the build emits no missing-export warnings.

**Six nits, no blockers.** What links most of them is that the API settles here, before either adopter arrives, so the gaps sit at that seam rather than in the paths the component runs today.

| Where | What |
| --- | --- |
| `root.tsx:21` | The labels memo is keyed on the object identity, so the inline form the README teaches defeats it. No slot is wrapped in `React.memo`, so the real cost is three `_x()` lookups per card per render rather than a re-render cascade. |
| `style.scss:28` | The label-row margin trim is calibrated for a 24px control but matches every button size. A default 36px button gets the same -2px, which neutralises nothing and grows the row from 40px to 60px. Both call sites today pass `size="small"`, so this is robustness rather than a live mismatch. |
| `value.tsx:65` | `newspack-stat-card__figure` is emitted but has no rule and no mention in the README, while `newspack-stat-card__action` runs the other way: styled and documented, never emitted. |
| `delta.tsx:69` | With `label` set, the children sit inside `aria-hidden`. `children` is typed `React.ReactNode`, so nothing stops an adopter putting a control there. |
| `index.test.js:477` | `label` winning over `directionLabel` is stated in the types and the README, but no test passes both. |
| `DEVELOPMENT.md:594` | The new `src/_variables.scss` convention is not in the styling guide. |

thomasguillot:
## Final pass

All six are addressed in `39061846e`.

`StatCard.Root` keys its memo on the three label strings rather than the object, so a caller writing `labels` inline gets a stable context. The label-row trim is scoped to `.components-button.is-small`, the size it was calculated for, so a default-size control in that slot is now left alone; I confirmed both in the browser. The class-name contract is documented in one table in the README, including that `__action` is the one class the card styles but never applies. `Delta` children are typed and documented as non-interactive, with the reason: `label` hides them from the accessibility tree while they would still take tab focus. Two tests pin `label` over `directionLabel`, including a blank `label` falling through to the caller's word rather than the built-in one. The styling guide now points at `src/_variables.scss`.

295 tests, typecheck and lint clean, and the four demo cards are unchanged on screen. Release risk stays ELEVATED, since that is the barrel export and none of the six was a defect.

PR #866, opened 2026-08-13
fix(reader-activation): keep the magic-link base on the site origin

jason10lee:
### All Submissions:

The reader magic-link email is built with a caller-supplied redirect as the base of the link. This restricts that base to the site's own origin, so the authentication link always points at the site itself.

A private helper, `Magic_Link::restrict_redirect_to_origin()`, validates the base: `wp_validate_redirect()` for parse-hardening, then a full-origin check — scheme, host and port — against `home_url()`. A base that resolves elsewhere collapses to `home_url()`; a same-site base or a relative path passes through unchanged. Both link builders, `generate_url()` and `send_email()`, route their base through it.

The check is stricter than `wp_validate_redirect()` alone in two ways. It does not inherit the platform-wide `allowed_redirect_hosts` list, which other flows populate for browser navigation (donation checkout, for one) — a broader allowance than an authentication link should take. And it compares the full origin rather than the host alone, so an `http://` base can't downgrade an `https` link.

An audit of every caller found no legitimate off-origin redirect: the auth form's JavaScript sets the redirect from `window.location.href`, and the two server-side callers derive theirs from the site's own URL. So this changes no behavior a reader or publisher relies on — only a base that resolves off-site is affected, and that is never produced by the normal flow.

NPPM-3041.

Run `n test-php` from `plugins/newspack-plugin` — the suite passes, including the five `magic-link.php` tests, which fail on the pre-fix code.

### Other information:

* [x] Have you written new tests for your changes, as applicable? 5 tests in `tests/unit-tests/magic-link.php` (the off-origin case is a data provider covering scheme mismatch, protocol-relative, backslash, and `mailto:`-style bases).
* [x] Have you successfully run tests with your changes locally? 2382 tests, 6786 assertions, 0 failures.

<details>

**Why the helper is a composite.** `wp_validate_redirect()` already handles the parser differentials a hand-rolled origin check gets wrong — `mailto:`/`javascript:` bases, protocol-relative `//host`, backslash authorities. The full-origin re-check on top is what drops the `allowed_redirect_hosts` inheritance and pins the scheme; neither layer is sufficient alone.

**Two builders, verified as the only surface.** A grep for every `add_query_arg` attaching the auth token confirms `generate_url()` and `send_email()` are the only two places the token becomes a URL base. A third construction (`newsletters_email_verification_email`) hard-codes `home_url()` as its base and carries the caller URL only as a query arg, so it is unaffected and correctly unchanged.

**Test coverage.** `generate_url()` returns its URL, so its cases assert on the returned base directly. `send_email()` embeds the URL in the email, so that case captures the rendered body through MockPHPMailer (quoted-printable-decoded) and asserts the token-bearing URL's host is the site host — not merely that `home_url()` appears in the body, which the email template satisfies independently.

**Scope.** Nonce, rate-limiting, token-binding and the `$_POST` intake are untouched; the builder-level gate covers every caller by construction, so the intake needs no change.

</details>


plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -405,6 +405,53 @@ public static function generate_secret( $user ) {
...
+		}
+
+		$target = \wp_parse_url( $candidate );
+
+		// A host-less result is a same-site path; wp_validate_redirect has already
+		// rejected the protocol-relative and backslash shapes that only look
+		// path-like.
+		if ( empty( $target['host'] ) ) {
+			return $candidate;
+		}
```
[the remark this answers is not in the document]
jason10lee: Fixed in `ac0b12c6d`. Added a `false === $target` guard after `wp_parse_url()` that falls back to `home_url()`, so the helper stays fail-closed even if an input ever parses in a way `wp_validate_redirect()` accepted. No test accompanies it: fuzzing found no input that passes `wp_validate_redirect()` yet fails `wp_parse_url()`, so the branch is unreachable today, and a test could only assert on a value nothing can produce.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
```diff
@@ -351,4 +352,177 @@ public function test_email_config_filter() {
...
+			// update this row (e.g. to an http:// input) or it silently stops testing the scheme check.
+			'scheme mismatch'     => [ 'https://example.org/x' ],
+			'protocol-relative'   => [ '//attacker.example/harvest' ],
+			'backslash authority' => [ 'https:\\\\attacker.example' ],
+			'schemeful host-less' => [ 'mailto:reader@example.com' ],
+			'userinfo confusion'  => [ 'https://example.org@attacker.example/x' ],
+			'suffix host'         => [ 'https://example.org.attacker.example/' ],
+			'bare authority'      => [ '://attacker.example' ], // Core normalises to a relative path.
+		];
+	}
```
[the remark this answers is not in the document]
jason10lee: Fixed in `b0c8edebb`. The row now derives the mismatch URL from `home_url()` at runtime, flipping only the scheme, so it exercises the scheme check regardless of whether the test site runs http or https.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
```diff
@@ -351,4 +352,177 @@ public function test_email_config_filter() {
...
+		reset_phpmailer_instance();
+
+		$sent = Magic_Link::send_email( get_user_by( 'id', self::$user_id ), 'https://attacker.example/harvest' );
+
+		$body = tests_retrieve_phpmailer_instance()->get_sent()->body;
+
+		remove_filter( 'newspack_magic_link_email_config', $name_callback );
+		remove_filter( 'newspack_email_configs', $config_callback );
+		Emails::reset_email_configs_cache();
+
```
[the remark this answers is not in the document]
jason10lee: Fixed in `b0c8edebb`. The `$sent` assertion now runs before the body is read, and the filter and cache setup and teardown are wrapped in `try`/`finally` so a failing assertion cannot leak state into later tests.

plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -405,6 +405,61 @@ public static function generate_secret( $user ) {
...
+		// malformed, but this is a security helper, so its fail-closed guarantee
+		// must hold even if that upstream assumption ever changes.
+		if ( false === $target ) {
+			return \home_url();
+		}
+
+		// A host-less result is a same-site path; wp_validate_redirect has already
+		// rejected the protocol-relative and backslash shapes that only look
+		// path-like.
+		if ( empty( $target['host'] ) ) {
```
adekbadek: **Suggestion:** The host-less branch returns its candidate unchanged, and the value's next stops are `add_query_arg()` and an email body — where a relative URL has nothing to resolve against.

The passthrough looks deliberate rather than overlooked (`test_generate_url_preserves_same_origin_base()` asserts `'Relative path preserved.'`), which is why it seems worth raising rather than treating as a slip: preserving a relative path is right for a value used in a browser redirect, but this one ends up in an email. `return \home_url( $candidate );` keeps the same-site intent, preserves an attached query string, and guarantees an absolute link.

Supporting detail is on NPPM-3041.

plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -405,6 +405,61 @@ public static function generate_secret( $user ) {
...
+	 * in two ways that matter here: it does not inherit the platform-wide
+	 * allowed_redirect_hosts list (which other flows populate for browser
+	 * navigation, e.g. donation checkout — a broader allowance than an
+	 * authentication link should take), and it pins the scheme, so an http://
+	 * base cannot downgrade an https link.
+	 *
+	 * @param string $url Candidate base URL.
+	 * @return string A same-origin base, or home_url() when $url is off-origin or empty.
+	 */
+	private static function restrict_redirect_to_origin( $url ) {
```
adekbadek, line 426: **Suggestion:** `send_email()` normalises a non-string redirect at line 539, but `generate_url()` has no equivalent, and both now funnel through this helper — a non-string value would fail inside `wp_validate_redirect()`'s `trim()` before any of the helper's own logic runs. No current caller passes one, so this is defence in depth rather than a live fault.

It matters a little more than usual because the comments here commit to a fail-closed guarantee holding "even if that upstream assumption ever changes", and the guarantee is currently uneven across the two entry points. Moving `if ( ! is_string( $url ) ) { $url = ''; }` into the helper and dropping the copy in `send_email()` makes it uniform and removes the duplication.

plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -405,6 +405,61 @@ public static function generate_secret( $user ) {
...
+		// path-like.
+		if ( empty( $target['host'] ) ) {
+			return $candidate;
+		}
+
+		$home          = \wp_parse_url( \home_url() );
+		$target_origin = \strtolower( $target['scheme'] ?? $home['scheme'] ) . '://' . \strtolower( $target['host'] ) . ( isset( $target['port'] ) ? ':' . $target['port'] : '' );
+		$home_origin   = \strtolower( $home['scheme'] ) . '://' . \strtolower( $home['host'] ) . ( isset( $home['port'] ) ? ':' . $home['port'] : '' );
+
+		return ( $target_origin === $home_origin ) ? $candidate : \home_url();
```
adekbadek: **Nit:** The origin comparison reads scheme, host and port only, so a `user:pass@` segment in an accepted base is carried through into the emailed link. No caller in the audited graph supplies credentials in a base, so unsetting `user` and `pass` before returning the candidate costs nothing and leaves every legitimate flow unchanged.

Supporting detail is on NPPM-3041.

plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -463,7 +518,7 @@ public static function generate_url( $user, $url = '' ) {
 				'token'  => $token_data['token'],
 				'secret' => self::generate_secret( $user ),
 			],
-			! empty( $url ) ? $url : \home_url()
+			self::restrict_redirect_to_origin( $url )
```
adekbadek, line 570: **Nit:** `generate_url()`'s `@param string $url Destination url. Default is home_url().` and `send_email()`'s `@param string $redirect_to` both read exactly as they did before this change, so a developer reading either signature has no way to learn that an off-origin value is now replaced by `home_url()`. The fallback is silent — no error, no log line. A sentence on both `@param` lines along the lines of "Must be same-origin; off-origin values fall back to `home_url()`" puts the contract where callers will read it.

Worth saying in the same breath: a destination on another domain is still supported after sign-in — it travels as a `redirect` query parameter inside a same-site link and is validated when the link is opened — so restricting the base does not remove that capability.

plugins/newspack-plugin/includes/class-magic-link.php
```diff
@@ -405,6 +405,61 @@ public static function generate_secret( $user ) {
...
+
+		// A host-less result is a same-site path; wp_validate_redirect has already
+		// rejected the protocol-relative and backslash shapes that only look
+		// path-like.
+		if ( empty( $target['host'] ) ) {
+			return $candidate;
+		}
+
+		$home          = \wp_parse_url( \home_url() );
+		$target_origin = \strtolower( $target['scheme'] ?? $home['scheme'] ) . '://' . \strtolower( $target['host'] ) . ( isset( $target['port'] ) ? ':' . $target['port'] : '' );
```
adekbadek: **Nit:** `https://<site>:443/x` builds `https://<site>:443` and compares unequal to a `home_origin` of `https://<site>`, so a legitimate destination collapses to the home page. The direction is safe and browsers never emit a default port in `location.href`, so this is cosmetic today — but normalising 80 and 443 against the scheme would make the docblock's full-origin claim true in both directions. Related: the data provider covers host and scheme mismatches but never a port mismatch, so that arm of the comparison has no coverage.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
```diff
@@ -351,4 +352,205 @@ public function test_email_config_filter() {
...
+			// change, the row keeps exercising a scheme mismatch on the same host
+			// instead of silently degrading to a host mismatch or a same-origin
+			// input).
+			'scheme mismatch'     => [ self::scheme_mismatch_base() ],
+			'protocol-relative'   => [ '//attacker.example/harvest' ],
+			'backslash authority' => [ 'https:\\\\attacker.example' ],
+			'schemeful host-less' => [ 'mailto:reader@example.com' ],
+			'userinfo confusion'  => [ 'https://example.org@attacker.example/x' ],
+			'suffix host'         => [ 'https://example.org.attacker.example/' ],
+			'bare authority'      => [ '://attacker.example' ], // Core normalises to a relative path.
```
adekbadek, line 421: **Nit:** The `'bare authority'` row passes with or without the helper: the resulting link's parsed host and scheme are both empty, so both assertions in `assertLinkStaysOnSite()` succeed either way. The `'backslash authority'` row (line 413) bites only on the scheme assertion, for the same reason — and would stop biting if the test site's `home_url()` ever became `https`.

Both rows document real core behaviour and are worth keeping. The point is only that neither asserts anything the fix uniquely provides, which is easy to read as coverage it isn't; a one-line comment saying so would settle it.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
```diff
@@ -351,4 +352,205 @@ public function test_email_config_filter() {
...
+		return $mismatched_home . '://' . $home['host'] . ( isset( $home['port'] ) ? ':' . $home['port'] : '' ) . '/x';
+	}
+
+	/**
+	 * A same-origin or same-site base is preserved unchanged.
+	 */
+	public function test_generate_url_preserves_same_origin_base() {
+		$user = get_user_by( 'id', self::$user_id );
+
+		$same_origin = Magic_Link::generate_url( $user, 'http://example.org/premium/' );
```
adekbadek: **Nit:** `'http://example.org/premium/'` bakes in both `WP_TESTS_DOMAIN` and its scheme — the same brittleness `scheme_mismatch_base()` was added a few lines above to avoid. Using `home_url( '/premium/' )` for the input and the `assertStringStartsWith()` expectation keeps the two halves of the file consistent.

adekbadek (approved):
Pre-approving, just some suggestions and nits.

plugins/newspack-plugin/includes/class-magic-link.php
jason10lee: Now returns `home_url( $path )`. Mutating it back reproduces the relative-URL artifact you predicted.

plugins/newspack-plugin/includes/class-magic-link.php
jason10lee: Moved into the helper, `send_email()`'s copy removed. Both entry points normalise identically now.

plugins/newspack-plugin/includes/class-magic-link.php
jason10lee: Dropped — the rebuild from `$target['path']` onward leaves them behind. Asserted in `test_generate_url_preserves_same_origin_base()`.

plugins/newspack-plugin/includes/class-magic-link.php
jason10lee: Both `@param` lines now carry the contract, including that a cross-domain destination still works via the `redirect` parameter.

plugins/newspack-plugin/includes/class-magic-link.php
jason10lee: New `url_origin()` normalises 80/443, and there's a `port mismatch` row now. Removing the normalisation fails exactly as you described.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
jason10lee: Comment added above both rows saying core rejects them before the origin check runs, so they don't cover the origin logic.

plugins/newspack-plugin/tests/unit-tests/magic-link.php
jason10lee: Built from `home_url()` now, input and expectation both.

jason10lee:
The threads on `:452`, `:460` and `:457` turned out to share one fix, so there aren't three separate patches to look for. 32 tests green, phpcs clean.

PR #867, opened 2026-08-13
refactor(components): move CardFeature to @wordpress/ui Card (DSGNEWS-215)

thomasguillot:
### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? #864 also touches `packages/components`, but a different component and no shared files.

`CardFeature` now renders on `@wordpress/ui`'s `Card` instead of WordPress core's. It is the first component to move as part of the `packages/components` audit, DSGNEWS-215.

The old implementation passed the whole card into `__experimentalCoreProps.header`, so a feature card was one large header carrying an inline `padding: 32` and no body at all. It now uses the compound slots for what they are: `Card.Header` holds the title, description and icon, `Card.Content` holds the action row. That removes `__experimentalCoreCard`, `__experimentalCoreProps` and `__experimentalHStack` from the component, along with three selectors that reached into `CoreCard`'s internals to force the layout.

The card takes the library's chrome: an 8px radius, a real 1px border in place of core's box-shadow ring, and the design system's 24px padding rather than the hardcoded 32px. The title now comes from `heading-x-large()`, and the title and description sit in a `Stack` with an 8px gap.

Comparing against `main` on a running site turned up three things worth naming:

- **Descriptions had lost their grey.** That colour came from `.newspack-card--core__header { color: $gray-700 }`, a rule the component no longer renders inside. Restored through the new `text-muted` mixin.
- **The `h2` and `p` browser margins came back** once the markup left core's card, pushing the title down and opening a gap that had not been there. Both are now zeroed, with the rhythm owned by the `Stack`.
- **The muted state's `border-color` had quietly become a no-op.** `@wordpress/theme@0.16.0` sets the card's own border to `#f0f0f0`, the exact grey the muted rule was painting, so the rule is gone and the README no longer promises a "lighter border". Muted cards are still signalled by the grey title, the error badge and the blocked button.

`text-align: left` is also gone from the title and description. All eight ancestors of the title compute `text-align: start`; the declaration existed only because `CoreCard`'s header could render as a `<button>`, which centres its text. Dropping it also fixes RTL, where the hardcoded `left` was wrong.

#### Accessibility

Three behavioural changes, all on the unmet-requirements state:

- **The blocked button stays reachable.** It carries `accessibleWhenDisabled`, so it keeps its place in the tab order and is `aria-disabled` rather than `disabled`. A keyboard or screen-reader user can now land on it instead of tabbing past it silently.
- **It says why it will not act.** `aria-describedby` points at the requirements badge, which is the only on-screen explanation. The reference is only set when `requirements` is set, so it never dangles.
- **The icon container is hidden from assistive tech.** A vendor mark passed as `node` carries no `aria-hidden` of its own, and the title already names the feature.

The per-card accessible names added in #757 are preserved: the primary button keeps `aria-label` in the form "Enable Metered countdown", and the overflow menu keeps "More options for %s".

#### New prop: `headingLevel`

The title was a hardcoded `h2`, which put every card at the same outline level regardless of what it sat under. It now takes `headingLevel`, defaulting to `3`. That default is already right for all three call sites, so none of them passes it: two render inside `WizardsTab` under its `h2`, and the third sits alongside an `h3`. The demo passes `4`, since its cards sit under `h3` section labels.

The name follows Core, which uses `headingLevel` wherever a component contains a heading whose outline position is the caller's to decide: `Page.Header` in `@wordpress/admin-ui`, and `ToolsPanel`, `ColorPalette` and `GradientPicker` in `@wordpress/components`. (Core's `Heading` uses the bare `level`, but there the component *is* the heading.)

`packages/components` currently spells this two ways, `heading` on `SectionHeader`, `ActionCard` and `EmptyState.Header`, and `titleLevel` on `CardForm` and `PluginSettings`. This PR does not close that split, it just avoids widening it; a follow-up will bring the rest onto `headingLevel`. The duplicated `HeadingLevel` union is now one shared type in `src/types.ts`, narrowed to `2 | 3 | 4 | 5 | 6`, and `CardForm` uses it too. Nothing passes `1`, and a card is never a page's only heading.

#### A long requirements badge no longer gets clipped

`Card.Root` sets `overflow: clip`, which core's card did not, and `.newspack-badge` is `flex: 0 0 auto`, so it could neither shrink nor wrap. Past a certain length the badge text was cut off mid-word, and that text is the only on-screen explanation of why the button will not act.

Ordinary data reaches that length. The Audience integrations view builds the string as `Requires %s` over a comma-joined plugin list, and WooCommerce extension names are long; `unsupportedReason` comes from the API and has no length ceiling at all. Measured on the demo, a three-plugin string overflowed the action row by 189px at 1680px wide and 246px at 1100px.

The action row now wraps, and the badge can shrink inside it. The badge drops onto its own line and wraps its text rather than being cut off. Cards in a row still bottom-align and still match heights.

#### New: `packages/components/src/_mixins.scss`

A shared partial holding `text-muted`, which reads the design-system token first and falls back to the `@wordpress/base-styles` grey, so a component still renders correctly outside `Page`, where the token sheet is not loaded.

A mixin rather than a utility class because roughly half the muted-text declarations in this package style elements *inside* third-party components through descendant selectors, where no class can be added. The partial ships with only the mixin that has call sites; the sweep that follows will add the others alongside their first real consumers.

The component README now records what the layout rests on: the two `Card.Root` properties the equal-height behaviour depends on, why the action row lives in `Card.Content` rather than a footer the library does not yet have, and why the title keeps `heading-x-large()` instead of `Card.Title`, which renders at 15px against this title's 20px.

The package's strings still carry the `newspack-plugin` text domain, which does not resolve for any consumer of `packages/components`. That is pre-existing and out of scope here, tracked in #942.

Part of DSGNEWS-215.

### Before and after

The components demo, `CardFeature` section.

**Before**

CardFeature before, on WP core Card

**After**

CardFeature after, on @wordpress/ui Card

1. Check out the branch and run `pnpm install`, then `pnpm run build:packages` followed by `pnpm --filter newspack run build`.
2. Open **Newspack → Components Demo** (`admin.php?page=newspack-components-demo`) and scroll to the **CardFeature** section. The wizard is hidden, so navigate to it directly. All eight cards render: the enabled, disabled and unmet-requirements states, the icon variants, custom labels, a custom badge and multiple dropdown controls.
3. Compare against the screenshots above. Cards carry an 8px radius and a 1px border, titles are 20px on a 24px line, descriptions are grey, and the action row sits at the bottom of each card.
4. Put two cards of different content lengths side by side in the **States** row. Both stay the same height and their action rows stay aligned.
5. Tab to the **Enable** button on the "Requires metering" card. It takes focus, reports itself as disabled, and does nothing on Enter. A screen reader reads the badge text after the button's name.
6. Lengthen that badge to check it stays readable. In the console, run `document.querySelector('.newspack-card-feature--muted .newspack-badge').textContent = 'Requires WooCommerce, WooCommerce Subscriptions, WooCommerce Memberships'`. The badge wraps to a second line inside the card. Narrow the window to 1100px and it still does.
7. Inspect any card title. It is an `h4` on the demo, which nests correctly under the `h3` section label above it.
8. If the site registers them, check the real call sites: **Newspack → Settings → Experimental tools**, the Audience integrations settings section, and the content-gates settings card. Cards there should match the demo. Neither tab was registered on the local site used for the screenshots, so the demo is the practical check.
9. Run `pnpm --filter newspack-components test` and `pnpm --filter newspack test`. Both suites pass.

### Other information:

* [x] Have you written new tests for your changes, as applicable? `card-feature/index.test.js` now covers the component's behaviour: structure and heading level, the primary button in each state, the blocked-but-reachable path and its description, badges, More-menu gating, and the icon variants. It absorbs the six accessible-name tests from #757 and the six More-menu tests that used to live in `card-feature-more-menu.test.js`, which is deleted as a duplicate now that `packages/components` runs its own suite in CI.
* [x] Have you successfully run tests with your changes locally? 22 suites / 303 tests in `newspack-components`, 92 suites / 880 tests in `newspack`, plus `typescript:check`, stylelint and eslint.


thomasguillot:
## Review

**Release risk: ELEVATED — from the shared-package surface, not from a defect.** `packages/components` is a contract with nine consumer units, and this PR adds to it: a new `_mixins.scss` partial, a `titleLevel` prop on `CardFeature`, and an optional `id` on `Badge`. All three are additive, so nothing existing breaks, but any change that widens that surface carries the tier. No finding in this review indicates an actual defect or an unexpected risk. Verdict: approve with comments.

### What was verified

The parts most likely to break silently were exercised rather than reasoned about:

- **Layout equalisation survives the migration.** `Card.Root` really is `display: flex; flex-direction: column`, so `margin-top: auto` on `.newspack-card-feature__actions` pins the action row. The unlayered rule beats `@wordpress/ui`'s layered `.header + .content { margin-block-start }`, and because the header is `:first-child` it keeps its full `padding-block-end`, so the header-to-actions gap never collapses.
- **The muted-title selector does what it looks like.** `&.newspack-card-feature--muted &__title` compiles to `.newspack-card-feature.newspack-card-feature--muted .newspack-card-feature__title`, confirmed against the built `dist/commons.css`.
- **`@wordpress/ui` is bundled, not externalised.** It sits in the dependency-extraction plugin's `BUNDLED_PACKAGES` list, so `Card` and `Stack` cannot arrive `undefined` at runtime on a host whose Gutenberg predates them.
- **The new SCSS partial ships.** `_mixins.scss` is present in both `dist/esm` and `dist/cjs`, and `@use "../mixins"` resolves at the same relative depth in both the source and published layouts, so an npm consumer such as `newspack-manager` resolves it too.
- **The blocked button is genuinely blocked.** With `accessibleWhenDisabled`, core computes `trulyDisabled = disabled && ! accessibleWhenDisabled`, drops the native attribute for `aria-disabled`, and installs `preventDefault` handlers spread after the consumer's `onClick`. Click, `mousedown` and keyboard activation all leave `onEnable` uncalled, while the button stays in the tab order.
- **The description association cannot dangle.** The button's `aria-describedby` and the badge's `id` key off the same `requirements` truthiness, verified in the locked, actionable and busy states, with exactly one element carrying the id.
- **Contrast.** Muted text resolves to `#707070`, about 4.95:1 on the card, up from `#757575`'s 4.60:1.

Suites and gates: 18 suites / 245 tests in `newspack-components`, 65 suites / 610 tests in `newspack`, `typescript:check`, eslint and stylelint all clean.

### What came back clean that was worth checking

The type-scale change (`heading-x-large()`, 20px on a 24px line at weight 499) is the correct step and deliberate — it is the SCSS twin of the design system's heading scale, and the 24px leading replaces an over-loose 32px. Removing the muted state's `border-color` is right rather than a loss: the card's own border already resolves to `#f0f0f0`, the exact grey that rule painted, so it had become a no-op. Dropping `text-align: left` is safe because it only ever compensated for a `<button>` header that no longer exists. And `overflow: clip` on the card does not clip the More menu, which portals to the document body.

### Worth a reviewer's eye

Two visual changes are deliberate and worth confirming against the screenshots in the description: card padding moves from a hardcoded 32px to the design system's 24px, and the title's leading tightens from 32px to 24px.

The unmet-requirements state is now carried by the title colour, the error badge and the blocked button, where it previously also had a border treatment. That reads clearly at the sizes involved, but it is a deliberate reduction in signals.

thomasguillot:
## Final pass

Everything from the review is now addressed, in `c19714f05`.

**Document outline.** The three call sites and the eight demo instances now pass `titleLevel={ 3 }`. This was the defect the prop was added for: `WizardsTab` renders its section title as an `h2`, so a card's own `h2` sat as a sibling of the heading introducing it rather than beneath it. Content Gates was the sharper case, where `SettingsCard` titles rendered `h2` while the gate cards immediately above them rendered `h3`, inverting the outline between two rows of equivalent cards. The change is visually inert, since `heading-x-large()` pins family, weight, size and line-height on the title's class rather than its tag. Confirmed on a running site: the titles now render as `h3` at an unchanged 20px / 24px / 499.

**Corrections to the docs.**

- The unmet-requirements section named the wrong callback. It said the lock leaves `onConfigure` with nothing to act on, when `onConfigure` is unreachable whenever `requirements` is set at all — the button only reads "Configure" with no outstanding requirement. The callback the lock actually neuters is `onEnable`, which is the one a consumer is most likely to have wired.
- That section also still described the button as "disabled". It is blocked but focusable, and described by the error badge, so the docs now say that and warn against wrapping it in a consumer's own `disabled` handling, which would undo the association.
- The canonical locked example used "Managed by site configuration", which renders the red error badge. A benign governance state shouldn't spend the loudest signal on the card, so the example now uses a genuine unmet requirement.
- `titleLevel`'s guidance was too generic to get chosen. It now names the two real cases: `3` under a `SectionHeader`, `2` when cards sit directly under a page's h1. The practical range is documented as 2-6, since wp-admin's `.wrap h1` rule outranks the title's class.
- `requirements` now doubles as the button's accessible description, which is an authoring constraint on the copy. Both the prop's JSDoc and the states table say so.

**Code tidy-up.** The description id was computed inline in two places that had to change together; it is now a single `describedById` next to the badge id, so the pairing is structural rather than incidental.

Two items are deliberately left for follow-ups rather than widened into this PR:

- **Heading conventions across the package.** `CardFeature` now has `titleLevel`, `SectionHeader` has `heading`, and `Drawer.Title` hardcodes an `h2`. `HeadingLevel` is also declared identically in `CardFeature` and `CardForm` without either exporting it, so a consumer writing a typed wrapper has to redeclare it. That wants one decision across the package rather than a third name added here.
- **A non-error badge for locked-but-fine states.** "Managed elsewhere, button blocked" is a real state, and forcing it through the error badge is the reason the example needed changing. Supporting it properly is a design call, not a doc fix.

Suites and gates after the pass: 18 suites / 245 tests in `newspack-components`, 65 suites / 610 tests in `newspack`, `typescript:check`, eslint and stylelint all clean.

thomasguillot (commented):
Three follow-ups pushed since the last round of changes.

**`9a8c928`** covers the state where a requirement is actionable and the feature is still off. That is what the Audience integrations view produces when a required plugin is installed but inactive, and it is where Activate lives. The suite reached the enabled and locked variants but not that one, and `settings-section.test.js` mocks `CardFeature`, so nothing exercised the button. The new test fails if the `requirementsActionable` term is dropped from the `disabled` expression.

**`bf6be27`** adds `dangerouslySetInnerHTML` to the `Omit` on `Badge`. It could never render unescaped HTML, since `text` is required and React throws its invariant, but the published type advertised an API that only ever crashes.

**`c67d5c6`** drops `headingLevel={ 3 }` from the three plugin call sites. Each was passing the component default, and that line was the only change in each file, so all three now match `main` and have left the diff. This PR is 11 files rather than 14.

Checked on a local site running WordPress 7.0.4:

- The blocked button reports `disabled=false`, `aria-disabled=true` and `tabIndex=0`, with `aria-describedby` resolving to a badge present on the page. Only the requirements badge carries an id, so the reference never dangles.
- Cards in a row hold equal height with unequal content, the shorter card's `margin-top: auto` absorbing 58.5px.
- A 72 character requirements badge wraps with no clipping and no scroll overflow at 1100px.
- Compiled LTR and RTL CSS for `.newspack-card-feature*` are byte identical, so no directional properties remain.
- `@wordpress/ui` is bundled rather than externalized, so there is no `wp.ui` undefined at runtime.

22 suites / 303 tests in `newspack-components`, 92 / 880 in `newspack`, plus `typescript:check`, eslint and stylelint.

PR #868, opened 2026-08-13
feat(components): adopt design system badges (DSGNEWS-215)

thomasguillot:
### All Submissions:

Badges across the admin now come from `@wordpress/ui`, so status labels match the design system. Three separate implementations become one: our own `Badge`, the markup `ActionCard` carried, and the plugin installer's status text.

The two APIs differ in more than name. Text moves from a `text` prop to children, and five severity levels map onto seven intents that describe attention priority rather than severity. `packages/components/src/badge/` is deleted and every call site moves over.

Most labels translate mechanically: `error` to `high`, `warning` to `medium`, `success` to `stable`, `info` to `informational`, `default` to `none`. Some do not, because severity was the wrong axis for them.

An inactive content gate is an unpublished draft post rather than a settled "off" state, so it reads `draft`. The same goes for the non-published state in the Emails, Pricing Rules and Plans status columns.

"Pending" is queued work, which the library documents as `low`, in both the activity log and the subscriber list. "Free" availability takes `low` as well, as something worth noticing rather than acting on.

Availability is a configuration fact rather than a problem, so "Private" is not a warning. It takes the `informational` the design system documents for it, which keeps Public, Free and Private reading apart from each other. "Canceled" in the activity log is a terminal state with nothing left to act on, so it reads `none` rather than sharing `high` with "Failed", which sits beside it as a separate filter.

Two spacing fixes come with the swap. The core card's small-card rule set a line-height on every descendant, which reached into the new badge and made it 28px instead of 24px; it now names the text elements it meant. `ActionCard`'s heading swaps a negative-margin gutter for a `gap`, which also fixes spacing when badges wrap.

`newspack-plugin` gains a direct `@wordpress/ui` dependency. That package sets `wpScript: false`, so webpack bundles it rather than WordPress providing it at runtime, and it should not lean on the workspace hoist the way core packages do.

Closes DSGNEWS-215.

CardFeature badges before and after

Before is on the left. Badges are heavier than they were: filled pills on an 8px radius, where the old ones were near-flat 2px chips with coloured text.

1. Build the components package and the plugin, then open Newspack > Audience > Access Control.
2. Confirm an inactive gate shows a grey "Inactive" badge, and an active one shows a green "Active".
3. Open the header kebab, then Gate Priority. The rows in that modal carry the same badges as the cards behind them.
4. Confirm the Metered Countdown card shows a red "Requires Metering" badge beside a blocked Enable button.
5. Tab to that blocked button. It stays reachable, and screen readers announce the badge as its description.
6. Open the Components Demo. All seven intents render with distinct colours on CardForm.
7. Find the two-badge action card in that demo. The badges sit 8px apart, matching the gap after the title, and each carries its own intent.
8. Scroll to the plugin installer. An installed plugin shows a green "Installed" badge against a green card edge.
9. On a site with subscriptions, open Audience > Plans. Public, Free and Private each render a distinct badge.
10. On the same site, check the subscriber list. Active reads `stable`, Pending `low`, On hold `medium`, Cancelled `high`.
11. On the same site, open Audience > Subscriptions and select the Subscriber-only products tab. The Status column renders Active and Paused badges.

### Other information:

The five card components used to describe a badge four different ways: a string or array of strings plus one shared intent, an array of `{ label, intent }`, an object keyed `text`, and a flattened `badgeText`/`badgeIntent` pair. They now share one shape, `{ label, intent }`, with `badges` where a component renders several and `badge` where it renders one. That shape is declared once, as `CardBadge` in `packages/components/src/types.ts`. Its label is optional, because every card drops a badge it cannot label rather than painting an empty pill, and three of the five used to declare the label required while guarding for its absence. That also closes a gap in `ActionCard`, which could not give two badges different intents.

`Badge` is no longer exported from `newspack-components`. Every call site in the monorepo moved with it, and the one consumer outside, `newspack-manager`, pins `~4.5.0` and never used it, so nothing needs a coordinated change.

The branch deliberately carries no `BREAKING CHANGE:` footer. The squashed commit spans this package and two plugins, and each unit's release analysis reads the whole commit body, so the footer would major `newspack` and `newspack-story-budget` alongside the library rather than just the package whose API changed. The `feat` title gives all three a minor instead, which is proportionate to what each actually gets.

`newspack-story-budget` styled `ActionCard`'s badge from outside with three `!important` rules. That block is removed. It had no effect either way: `Budget::get_stories_count()` returns a hardcoded `0`, and `ActionCard` treats `0` as falsy, so that badge never rendered.

The plugin installer's four status edges move from hardcoded colours to `--wpds-color-stroke-surface-*` tokens. Each keeps a fallback matching its token, so a bundle that loads the component without the design tokens renders the same colours.

The front-end `newspack-ui__badge` is a separate, reader-facing component and is untouched.


thomasguillot:
## Code review, release risk: LOW

Exercised against a local site running this branch: the gate cards on Access Control and the same cards inside the Gate Priority modal, the components demo (all seven intents, single and two-badge action cards, the plugin installer), and the Setup welcome cards. Measured rather than eyeballed wherever a number was the point.

What held up:

- **The header change does what it claims, and the fix for it was load-bearing.** Identical gate cards now render at 49px on the page and 48px in the Gate Priority modal. Before it they were 64px in the modal: `Modal` portals to `document.body`, which drops the `.newspack-wizard` ancestor that had been winning the specificity tie against Core's emotion `padding: 16px 24px`, while `min-height: 32px` on `__header-content` still applied.
- **Nothing is left on the old API.** `Badge`, `badgeLevel`, `badge.level`, `BadgeLevel`, `.newspack-badge` and `.newspack-action-card__badge` return zero hits across every plugin, theme and package, and across the standalone `newspack-manager` checkout.
- **The two deleted `display: inline-block` overrides were safe to remove.** Measured against the real DataViews grid rules (`min-height: 24px`, `padding-top: 2px`, `line-height: 20px`), the badge sits inside a 26px value box at 22px tall and hugs its label at 60px rather than stretching. That was the concern the overrides existed for, and it does not reproduce.
- **Badges stay styled without the token block.** Every `--wpds-*` reference in the library's badge CSS ships a literal fallback, for example `var(--wpds-color-background-surface-success,#c6f7cd)`, so the plugins that bundle `ActionCard` without ever rendering `Page` still get a styled badge.
- **The story-budget override really was dead.** `Budget::get_stories_count()` returns a hardcoded `0`, and `ActionCard` normalises `badge={ 0 }` to `badges = 0`, so `badges?.length` is `undefined` and no badge has ever rendered on that screen.

At HEAD: components 19 suites / 247 tests, plugin 65 suites / 610 tests, `typescript:check` clean, ESLint 0 errors with 27 pre-existing warnings, Stylelint clean on every changed stylesheet, both builds succeed. Plugin `tsc` moved from 223 errors to 222, with none newly introduced.

### Findings

**Severity: suggestion** — `packages/components/src/card/style-core.scss:82`

The 8px vertical padding is unconditional, so it also flattens Core's size-keyed padding on **icon** cards, where the 32px floor on `__header-content` does not reach because the icon is its sibling rather than its child. The Setup welcome cards and the Add prompt modal cards drop from 80px to 64px. Measured on the welcome screen: 64px header, 40px icon, 48px content, so content drives the height and nothing clips. This reads as a deliberate tightening rather than a defect, but it is a 16px shift on screens outside this PR's test plan.

**Severity: nit** — `packages/components/src/action-card/index.js:170`

`{ badges?.length && … }` renders a literal `0` when `badge` is an empty array. The guard is byte-identical on the base commit and no call site passes one, so this is pre-existing and currently unreachable rather than something this PR introduces.

### Deliberately unchanged

- The installed-plugin edge keeps `--wpds-color-background-surface-success`. It is a fill token doing a stroke's job at roughly 1.19:1 against the card, but the row already carries a visible "Installed" badge, so the stripe is redundant reinforcement rather than the only cue for that state, and its sibling stripes have always sat near 1.05:1.
- `newspack-components` is published to npm, and this removes an export and renames props under a patch bump. Every in-repo consumer moves within this PR, and the external exposure is accepted.
- `cancelled` stays `high`, `canceled` stays `medium`, and Availability keeps both `free` and `private` on `informational`.

thomasguillot:
## Final pass

Two changes on top of the review above, plus a description that now matches the code.

- `packages/components/src/action-card/index.js` guards the badge list with `badges?.length > 0` instead of `badges?.length`. With the old guard an empty `badge` array rendered a literal `0` into the heading. No call site passes one, and the guard predates this PR, but the JSX around it is already being rewritten here so it is cheaper to close now than to leave as a trap for whoever passes an array next.
- The description picks up the two things it was missing: why the header padding sits on a doubled class, and that cards with an icon tighten from 80px to 64px because the icon sits beside the header content rather than inside it, so the 32px floor does not reach them. Setup > Welcome and the Add prompt modal are the visible cases. Added a test step for the Gate Priority modal, which is the surface the specificity fix was for.

The icon-card shift is left as-is. It follows from making the padding uniform, and it measures correctly on the welcome screen: 64px header around a 40px icon and 48px of content, content driving the height.

Everything at HEAD: components 19 suites / 247 tests, plugin 65 suites / 610 tests, `typescript:check` clean, ESLint 0 errors with the 27 pre-existing warnings, Stylelint clean on every changed stylesheet, both builds succeed.

thomasguillot (commented):
Ran a verification pass over this branch in an isolated environment, with WooCommerce, Subscriptions and the pricing engine installed and the data each screen needs seeded, so every step of the test plan above could be run rather than read. That turned up one regression and a handful of smaller things, all fixed in the three commits at the top of the branch.

### The regression

Moving from a scalar `badge` prop to a `badges` array moved the falsy guard. `ActionCard` used to drop a falsy `badge`, but an array literal always has length 1, so a badge built from an absent value rendered as an empty pill. `getConnectionStatus()` returns `null` for a disabled gateway, so all three cards on Checkout & Payment showed one, which is the default state of that screen.

The guard now sits in `ActionCard` and filters on the label, which covers the two other call sites that build a badge from data that can be absent: `placementForPopup()` returns `undefined` for a placement in neither the placement map nor `custom_placements`, and `order.status` in the header bidding card comes straight off the GAM API. `SectionHeader`, `CardForm` and `CardSortableList` took the matching guard so the unified contract behaves the same way everywhere.

### Two intent collisions

Both resolved against the design system's own decision tree rather than by eye:

* **Canceled** had landed on `high`, the intent Failed already holds, so the activity log painted a deliberate stop and a failure identically while still offering them as separate filters. It now reads `none`, where the system files terminal states with nothing to act on.
* **Private** reading `draft` left it a faint border away from Public's `none`, which is the pairing that column most needs to separate. Private now takes the `informational` the system documents for it, and Free moves to `low`, so all three read apart.

### Smaller fixes

The Applied rules chip sizes to its pill instead of overhanging it by 2px, since the library badge computes to `display: inline` where ours was `block`. `CardSortableList` gets the `intent` default its three siblings already had. The installer status fallbacks each state their own token's value rather than a base colour that happens to coincide. The pricing rules intent lookup now degrades the same way the label beside it already did.

### Tests

18 assertions across five suites pin each deliberate departure this PR argues for, plus a rule that no two statuses in a map may share an intent, which is what let Canceled and Failed collide. Full suites pass: 920 in the plugin, 405 in components, 71 in story budget.

### One thing for a reviewer

The header bidding card in `newspack-ads` cannot render in the admin at all: its enqueue is gated on page slug `newspack-advertising-wizard`, which exists nowhere else in the repo, and the same is true on `main`. Nothing this branch caused, and the new guard covers that call site if the surface is ever revived, but it is worth someone deciding whether it should be fixed or removed.

PR #869, opened 2026-08-13
feat(components): Accordion becomes CollapsibleGroup (DSGNEWS-215)

thomasguillot:
### All Submissions:

`Accordion` wrapped core's legacy `PanelBody`, and its entire stylesheet existed to strip that component's chrome back off so the panels would sit flush with the surrounding column. This rebuilds it on `Collapsible` from `@wordpress/ui`, so the stylesheet is styling rather than counter-styling.

Renamed on the way, because the component is not an accordion: a W3C accordion coordinates its panels and moves focus between headers with the arrow keys, while these are independent disclosures. `Accordion` and `AccordionPanel` become `CollapsibleGroup` and `CollapsibleGroup.Item`, one barrel export instead of two.

The panels look as they did. Four things behave better:

* A collapsed item is hidden with `hidden="until-found"`, so the browser's find-in-page matches text inside a closed item and expands it to reveal the result.
* The trigger has a proper focus ring on the design system's focus token. It previously inherited one from the core `Button` inside `PanelBody`, and `Collapsible` ships no styles of its own, so without this it would have fallen back to whatever the browser draws.
* Item titles take their heading level from the group. `titleLevel` defaults to `2`, and both call sites pass `3`, because they sit under a section header rendered as `h2` and were previously repeating that level.
* The rule between items is the lightest grey, the same weight the Integrations screen already uses between its own sections.

Part of DSGNEWS-215.

1. Go to Components Demo, find the CollapsibleGroup section.
2. Expand and collapse each item. Only the clicked item changes state.
3. Confirm a divider sits between items but never after the last one.
4. Collapse every item, then use browser find-in-page for "independent". The matching item expands.
5. Tab to a trigger. A focus ring in the admin theme colour appears, offset from the text. Enter and Space both toggle it.
6. Inspect a trigger. It is a `button` inside an `h3`, with `aria-expanded` tracking its state. The `h3` is the demo passing `titleLevel={ 3 }`; the default is `h2`.
7. Check the "Single item" example renders open with no title and no trigger.
8. Go to Audience > Integrations, configure an ESP with grouped outbound fields.
9. Confirm the grouped checkbox sections render as before, first group open.
10. Set the browser to reduce motion. The chevron snaps instead of rotating.

### Other information:

`hidden="until-found"` can only match text that exists, so a collapsed item keeps its children mounted rather than unmounting them the way `PanelBody` did. Every item now renders on first paint, and a child's state and effects persist while it is closed. The README says so, and says to keep anything expensive or fetch-on-mount behind the item's own open state. The one call site renders checkboxes, so nothing there changes in practice.

`Accordion` and `AccordionPanel` were public exports of `newspack-components`, which is published, so this removes them from the package surface. Nothing imports them: every consumer in the workspace resolves `workspace:*`, and `newspack-manager`, the one repository outside it that installs the published package, imports `Button`, `Grid`, `SectionHeader`, `SelectControl`, `Card` and `DataViews` only. Worth a line in the release notes rather than a deprecated alias, given no consumer to keep working.

Open and close are deliberately not animated, matching `PanelBody` today. Animating means wiring `--collapsible-panel-height` and the base-ui starting/ending style hooks, which interacts with `hidden="until-found"`; only the chevron rotation animates, behind `prefers-reduced-motion`.

The 24px gap under a trigger sits on an inner wrapper rather than the panel. `hidden="until-found"` uses `content-visibility: hidden`, which keeps the panel's own box, so padding there would leave a gap under a collapsed item.

Verified on the demo page. Audience > Integrations is not reachable on the local env used here, so that call site is covered by its existing test suite rather than by hand.


thomasguillot:
## Verification pass

Release risk: **low**. No blockers, and nothing outstanding that indicates a defect. One documentation nit, addressed below.

I've read the full diff, both the round-3 commit and the accumulated state, verified the library internals the code depends on, and independently re-ran the components suite.

### Overview

Confirmed converged. The single-provider refactor of `root.tsx` is a strict improvement over the two-return shape: the `Provider` is now the stable outer element and only its child alternates between `div` and `VStack`, so the branch flip remounts exactly the same subtree it always did, with no new remount, no key change, no identity change. Hooks are unconditional in both `Root` (`useTitleLevel` before the ternary) and `Item` (before the `!title` early return). `cloneElement` preserves the `Children.toArray` key, and since `Item` has no `defaultProps`, `title: undefined` genuinely erases the title rather than falling back.

Things I checked that could have gone wrong and didn't:

- **Tests are not vacuous.** `packages/components` runs 17 suites / 229 tests green here, with no console errors from the new suite. The `data-panel-open` pair is self-validating: `toHaveAttribute` in the open test would fail if the attribute name were wrong, so the `not.toHaveAttribute` in the close test can't pass hollowly. I confirmed against base-ui 1.5.0 that `data-panel-open` really is the trigger attribute (`triggerOpenStateMapping`), so the chevron rotation selector is keyed correctly. The heading assertions run against headings that sit outside the panel and so are in the a11y tree; both nested tests genuinely exercise the two provider branches, since the outer item is either `defaultOpen` or untitled.
- **`hidden="until-found"`** is set imperatively by base-ui in a layout effect keyed on `[hidden, hiddenUntilFound]`, so it is re-applied on close. The close-test assertion is real, and the `beforematch` listener backing the README's find-in-page claim is wired.
- **No form-submit hazard.** base-ui's `useButton` applies `type: 'button'`, which matters because the audience call site sits in a settings form.
- **No runtime-externalization trap.** `@wordpress/ui` is in `dependency-extraction-webpack-plugin`'s `BUNDLED_PACKAGES`, so `Collapsible` is bundled, not resolved from a `wp.ui` global that WP core may not ship. This is the failure mode that bit `Badge`.
- **The bare `<button>` reset is complete.** `heading-large()` pulls in `_text-heading()`, which sets `font-family` and `font-weight`, so the trigger does not fall back to the UA button font. The switch from core's forced `color: #1e1e1e` to `color: inherit` lands on the wp-admin body colour, which is exactly what the adjacent `SectionHeader` headings already inherit, so the new value is more consistent with its surroundings, not less.
- **No lingering consumers.** A workspace-wide grep for `Accordion` returns only unrelated hits (`newspack-ui__accordion` front-end markup, the core/WooCommerce accordion blocks in the themes). The barrel export is alphabetically placed, `Divider`'s `tertiary` variant exists and its `margin: min(32px, 0px)` resolves to 0, and `.tsx`/`.ts` plus `copy-styles.js` are well-trodden ground in this package.

### Findings

**nit — `packages/components/src/collapsible-group/README.md:18`**

"Anything else, including bare text, is ignored" is true only for text, numbers and nullish children, which `Children.toArray().filter(isValidElement)` really does drop. Any other valid element, a wrapper component, or a `Fragment` holding two items, which `Children.toArray` does not flatten, is not ignored: it is counted as one item and rendered, and in the `hideSingleTitle` single-child path it is additionally cloned with `title: undefined`. The practical consequence is silently wrong divider placement rather than the documented no-op, which is the opposite of what a reader would brace for. Since this sentence was added as the round-2 fix for the unchecked cast, it is worth tightening to say non-`Item` elements render but break divider placement. Code unchanged either way.

thomasguillot:
## Final pass

Two changes, both in `ae5cee2f5`. No code behaviour changed.

**The children note was overstated.** It said anything other than `CollapsibleGroup.Item` is ignored. `Children.toArray().filter( isValidElement )` drops only text, numbers and nullish children; any other element still counts as one item and takes a divider, and `Children.toArray` does not flatten a `Fragment`, so a `Fragment` holding two items reads as one. The note now says that, since wrong divider placement is the opposite of the no-op a reader would have braced for.

**Comments trimmed** to the ones that carry a reason rather than restating the code. Four went: the heading-wraps-trigger and `hiddenUntilFound` notes in `item.tsx`, and the inherited-level and lone-item notes in `root.tsx`, all of which the README and the prop docs already cover. What stayed: the one-line note on why `Item` hangs off the group rather than forming a namespace object (the same shape `Drawer` documents), the `data-panel-open` note explaining why a test asserts a library-owned attribute, the two lint disables, and the prop-level JSDoc that consumers see in their editor. The stylesheet carries no prose comments.

Green at this commit: 17 suites / 229 tests in `newspack-components`, 65 suites / 610 tests in `newspack`, `typescript:check` clean, lint at the repository baseline, stylelint clean.

packages/components/src/index.js
```diff
@@ -14,6 +13,7 @@ export { default as CardSettingsGroup } from './card-settings-group';
 export { default as CardSortableList } from './card-sortable-list';
 export { default as IntegrationIcon } from './integration-icon';
 export { default as CategoryAutocomplete } from './category-autocomplete';
+export { default as CollapsibleGroup } from './collapsible-group';
```
thomasguillot, line 16: **Suggestion:** Retitle the squash subject to `feat(components):` so the removal cuts its own release

`Accordion` and `AccordionPanel` leave the published `newspack-components` surface here, and that part is settled: nothing in the workspace or outside it imports either one, and `newspack-manager`, the only consumer installing from npm, pins `~4.5.0`.

What is left is the release signal. `config/release-package.js` runs a stock `@semantic-release/commit-analyzer` with no `releaseRules`, so `refactor` is not a releasing type. `refactor` appears zero times in `packages/components/CHANGELOG.md`, which is the proof. Merged under the current title the removal waits on `main` until an unrelated `fix:` or `feat:` touching the package cuts 4.6.1 or 4.7.0, and the two exports disappear inside that release with nothing naming them.

`feat(components):` cuts 4.7.0 with its own changelog entry, which is the record worth having. It also clears the repo's publisher-visible bar on its own terms: find-in-page now reveals a collapsed field group on the Integrations screen, the trigger has a real focus ring, and the heading outline runs h1 to h2 to h3 rather than repeating h2. One side effect to know about: multi-semantic-release scopes by changed path rather than by scope label, so this cuts a minor for `newspack-plugin` as well.

**Risk if not addressed:** If a consumer later widens its dependency range, it crosses the removal inside what its manifest reads as a patch or minor bump, and any import of a dropped name resolves to `undefined` and throws at render rather than at build time, taking that admin screen out until someone traces it back to a version number that gave no warning.

**Consider:** If you want the stronger signal, a `BREAKING CHANGE:` footer in the squash body forces a major whatever the type, which is what semver asks for when public exports go. `0fc9822ec` is the precedent, and commitlint here is bare `config-conventional`, so a footer is allowed. With no importer to migrate, a minor is the proportionate choice.

plugins/newspack-plugin/src/wizards/audience/views/integrations/configure-view.test.js
```diff
@@ -46,9 +46,8 @@ jest.mock( '@wordpress/components', () => ( {
 	),
 } ) );
 jest.mock( '../../../../../packages/components/src', () => ( {
-	Accordion: ( { children } ) => children,
-	AccordionPanel: ( { children } ) => children,
 	Button: ( { children } ) => children,
+	CollapsibleGroup: Object.assign( ( { children } ) => children, { Item: ( { children } ) => children } ),
```
thomasguillot, line 54: **Suggestion:** The mock will hide unreachable checkboxes once a fixture has two groups

`Object.assign( ( { children } ) => children, { Item: … } )` mirrors the real compound-component shape, and it is faithful to what the tests do today: every fixture supplies exactly one `grouped_options` entry and the call site passes `hideSingleTitle`, so the real component also renders that item untitled and permanently open.

The gap opens at two groups. I loaded this screen against a live ESP whose `outgoing_metadata_fields` has two sections, and the second renders inside `hidden="until-found"`: its nine checkboxes are in the DOM, but `content-visibility: hidden` takes them out of the tab order, and calling `.focus()` on one leaves `document.activeElement` on `body`. The stub flattens that distinction away, so a multi-group test could assert `getByLabelText( … ).checked` on a checkbox nobody can reach in a browser and still pass.

A comment on the mock saying it only stands in for the single-group case, or dropping the flattening when a multi-group fixture lands, would keep that from going unnoticed.

packages/components/src/collapsible-group/root.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Internal dependencies
+ */
+import Divider from '../divider';
+import { TitleLevelContext, useTitleLevel } from './context';
+import type { CollapsibleGroupItemProps, CollapsibleGroupProps } from './types';
+import './style.scss';
+
+const Root = ( { children, className, gap = 'xl', hideSingleTitle = false, titleLevel }: CollapsibleGroupProps ) => {
+	const inheritedTitleLevel = useTitleLevel();
+	const items = Children.toArray( children ).filter( isValidElement ) as React.ReactElement< CollapsibleGroupItemProps >[];
```
thomasguillot, line 25: **Suggestion:** The line that keeps divider placement correct has no test

`Children.toArray( children ).filter( isValidElement )` is what makes `items.length - 1` on line 32 a count of *rendered* items: `toArray` strips `null`, `undefined` and booleans, and the `.filter` drops the strings and numbers it keeps. The call site's `{ outboundEnabled && outboundField && … }` shape makes conditional children realistic usage, and the README documents the contract at line 18.

None of the 12 tests passes anything other than a clean array of `Item`s, though. Dropping the `.filter`, or moving to `Children.map` (which does not flatten the same way), would put a divider after the last item and every test would still pass. One case rendering `{ false }` and a stray text node alongside two items, asserting one divider, would pin it.

packages/components/src/collapsible-group/style.scss
```diff
@@ -0,0 +1,47 @@
...
+			/* stylelint-disable-next-line plugin-wpds/no-unknown-ds-tokens -- the plugin's token list lags the prebuilt design-tokens.css names. */
+			outline: var( --wpds-border-width-focus, var( --wp-admin-border-width-focus, #{wp-vars.$border-width-focus-fallback} ) ) solid var( --wpds-color-stroke-focus, var( --wp-admin-theme-color ) );
+			outline-offset: wp-vars.$grid-unit-05;
+		}
+
+		svg {
+			flex-shrink: 0;
+		}
+
+		&[data-panel-open] svg {
```
thomasguillot: **Suggestion:** The chevron keys off a vendor data attribute rather than the ARIA state beside it

`&[data-panel-open] svg` depends on an attribute `@wordpress/ui` does not define. Its `collapsible/{root,trigger,panel}.tsx` are bare `forwardRef` pass-throughs with no props of their own, so `data-panel-open` comes from `@base-ui/react`'s `triggerOpenStateMapping`, two dependency hops down and outside anything the wrapper documents.

`aria-expanded` sits on the same element, is set unconditionally, and is guaranteed by the disclosure pattern rather than by a vendor. `[aria-expanded="true"] svg` costs nothing and drops the dependency. The test at `index.test.js:131` does pin `data-panel-open`, so a change underneath would fail loudly rather than silently, but that is a test you would not need to write against the ARIA contract.

Worth noting on the version range while we are here: `^0.16.0` on a `0.x` version resolves to `>=0.16.0 <0.17.0`, so a `@wordpress/ui` minor cannot arrive without a manifest change. Patch releases can still move Base UI underneath, which is what this guards against.

packages/components/src/collapsible-group/item.tsx
```diff
@@ -0,0 +1,42 @@
...
+import type { CollapsibleGroupItemProps } from './types';
+
+const Item = ( { children, className, defaultOpen = false, title }: CollapsibleGroupItemProps ) => {
+	const titleLevel = useTitleLevel();
+
+	if ( ! title ) {
+		return <div className={ classNames( 'newspack-collapsible-group__item', className ) }>{ children }</div>;
+	}
+
+	const Heading = `h${ titleLevel }` as const;
```
thomasguillot, line 25: **Nit:** titleLevel is enforced only by TypeScript, and neither call site is typed

The heading tag is built by interpolating `titleLevel` straight into a template literal. `HeadingLevel` in `types.ts` constrains it to `1 | 2 | 3 | 4 | 5 | 6`, but the barrel is `src/index.js` and both call sites (`configure-view.js:550` and `componentsDemo/index.js:396`) are plain JS, so nothing type-checks the prop where it is actually passed. A `titleLevel={ 7 }`, or a value read from data, renders `<h7>`, which React emits as an unknown element with no warning and no heading role. Clamping in `Root` before the value enters the provider would make the type documentation rather than the only defence.

packages/components/src/collapsible-group/style.scss
```diff
@@ -0,0 +1,47 @@
...
+		display: flex;
+		gap: wp-vars.$grid-unit-10;
+		justify-content: space-between;
+		padding: 0;
+		text-align: left;
+		width: 100%;
+
+		&:focus-visible {
+			/* stylelint-disable-next-line plugin-wpds/no-unknown-ds-tokens -- the plugin's token list lags the prebuilt design-tokens.css names. */
+			outline: var( --wpds-border-width-focus, var( --wp-admin-border-width-focus, #{wp-vars.$border-width-focus-fallback} ) ) solid var( --wpds-color-stroke-focus, var( --wp-admin-theme-color ) );
```
thomasguillot: **Nit:** The focus ring has no literal fallback, and is now the second copy of this declaration

Two small things about this line.

The width half bottoms out in a Sass literal, but the colour half ends at `var( --wp-admin-theme-color )` with nothing after it. If both custom properties were ever unset, the `var()` is guaranteed-invalid, which makes the whole `outline` shorthand invalid at computed-value time, so there would be no ring at all rather than a differently coloured one. `divider/style.scss` handles the same situation with an explicit Sass literal and a comment saying why. It is unreachable on both current call sites: I measured the ring live at `1.5px solid #003da5`, because the barrel pulls Page's token sheet.

The declaration is also now byte-identical to `tabbed-navigation/style.scss:94`, except this copy carries a `stylelint-disable` for `plugin-wpds/no-unknown-ds-tokens` and the older one does not. The older one does fail that rule today; it survives only because CI runs `lint:js` and never `lint:scss`, leaving the pre-commit hook as the sole gate on a file that is no longer staged. A shared focus-ring mixin would collapse both questions to a single site.

packages/components/src/collapsible-group/root.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+	const content =
+		hideSingleTitle && items.length === 1 ? (
+			<div className={ classNames( 'newspack-collapsible-group', className ) }>{ cloneElement( items[ 0 ], { title: undefined } ) }</div>
```
thomasguillot: **Nit:** The one-item and many-item branches render different element types

The `hideSingleTitle` branch emits a bare `div` carrying `newspack-collapsible-group`, while the general branch emits `Stack`, which merges its own class alongside the same name. So `.newspack-collapsible-group` is `display: block` in one case and a flex column with a 24px gap in the other. I confirmed both live. Nothing keys off that class for layout today, so this is really about the class being an unreliable styling hook, and about an item count crossing 1↔2 mid-session swapping the element type and remounting every item, discarding each one's uncontrolled open state. The old `Accordion` had the same shape, so it is not a regression; worth knowing rather than fixing.

packages/components/src/collapsible-group/types.ts
```diff
@@ -0,0 +1,22 @@
+import type { GapSize } from '@wordpress/theme';
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
```
thomasguillot, line 3: **Nit:** The new prop types are invisible to npm consumers

`CollapsibleGroupProps`, `CollapsibleGroupItemProps` and `HeadingLevel` reach in-monorepo consumers, which import `packages/components/src` directly, but nothing publishes them: `compile:esm`/`compile:cjs` run Babel with `@babel/preset-typescript`, which strips types without emitting declarations, `copy-styles.js` filters to `.scss` and `.svg`, and `package.json` has no `types` or `exports` field. I ran the publish build to check: the `.ts`/`.tsx` sources compile and the stylesheet is copied, so the tarball is correct, just untyped.

`drawer/` already set this precedent and 17 component directories are now TypeScript, so this PR does not introduce the gap. Raising it here because the package's TS surface has grown large enough to be worth a DSGNEWS-215 sibling rather than another silent instance.

thomasguillot (commented):
Release risk: ELEVATED. No blockers, and nothing can break today: two public exports leave a published package, but nothing in the workspace or outside it imports either one, and the only consumer installing that package from npm pins below the version that would carry the change. The tier is about release mechanics rather than the code: retitling the squash subject to `feat(components):` settles it, and that is the one call worth making before this merges.

I browser-tested both call sites against a live admin serving this branch. The disclosure markup holds up (`h3` wrapping a `button` with `aria-expanded`), find-in-page reveals a collapsed item and leaves its siblings shut, the focus ring resolves on the design system token, Enter and Space both toggle without scrolling the page, and the Integrations screen now runs h1 → h2 → h3 with no skipped level. Open state survives a parent re-render, and the checkboxes inside a collapsed section stay out of the tab order.

Two patterns across the rest. Both test notes are the same shape: the suite exercises a child arrangement the real component never sees, so a regression in divider placement or in reachability would pass unnoticed. And both stylesheet notes land on one declaration that now exists twice in the package, so a shared focus-ring mixin would settle them together.

packages/components/src/collapsible-group/root.tsx
thomasguillot: Unified on `Stack`. Both branches now render through the same `Stack` and the same keyed `Fragment`, with `hideTitle` deciding only whether to clone the title away, so the class is a consistent styling hook and the element type no longer changes with the item count.

It does not fix the remount, though. Mounting a group with an input inside and crossing from one item to two: without `hideSingleTitle` the open state and the typed value both survive, with it the input still resets. That comes from `Item` rather than the wrapper, since the title appearing swaps `<div>{children}</div>` for `<Collapsible.Root>` plus `<Collapsible.Panel>`, and the children remount whatever encloses them. Keeping them would mean `Item` always rendering the Collapsible tree with a forced-open root and no trigger, flipping it between controlled and uncontrolled. Not worth it for a crossing no call site makes.

packages/components/src/collapsible-group/types.ts
thomasguillot: Confirmed, and left for its own ticket. There are no `.d.ts` files anywhere in `dist`, `package.json` has no `types`, `typings` or `exports`, and `main` points straight at `dist/esm/index.js`. `types.ts` does compile, to `dist/esm/collapsible-group/types.js`, which is 10 bytes of `export {};`.

So the props are typed inside the monorepo and untyped for anyone installing from npm, which is part of why the heading level is now clamped at runtime rather than resting on the type alone. Emitting declarations and pointing `package.json` at them is a package-wide build change, so it wants its own ticket rather than riding along here.

PR #870, opened 2026-08-13
feat(cli): migrate premium newsletters to Access Control gates (NPPD-2079)

dkoo:
### All Submissions:

Adds `wp newspack migrate-premium-newsletters`, so premium newsletters stop being the one hand-clicked step in an otherwise scripted Access Control migration. Today a migration runs `migrate-membership-gates`, then someone opens Newsletters > Premium and recreates each premium newsletter gate by hand.

For every published WooCommerce Memberships plan carrying a `newspack_nl_list` content-restriction rule, the command writes the equivalent premium newsletter gate: the plan's lists as the gate's restricted lists, the plan's products as its paid-access rules. Plans restricting the same lists are grouped and share one gate, matching how `migrate-membership-gates` groups. Dry-run by default, `--live` to write.

NPPD-2079. Sibling of `migrate-membership-gates` (NPPD-2059), whose shape this follows.

Three things worth a reviewer's attention:

**Auto-signup is derived, and sometimes deliberately refused.** Before Access Control, activating a membership auto-subscribed the member to the plan's restricted lists, except lists shown in the post-checkout signup modal, which were left to reader opt-in. `newspack_premium_newsletters_auto_signup` is a single site-wide option, so that per-list distinction only survives when every restricted list falls on the same side. When a site's lists disagree, the command writes nothing and names both sides — one flag cannot express the split, and either guess has a victim: on subscribes readers who opted out, off drops readers who expected the list. A `--plan`-scoped run never writes the option at all, since its derivation would see only that plan's lists.

**Migrated gates are dormant until WooCommerce Memberships is deactivated**, so a gate that will not restrict looks migrated right up to cutover. Both passes check for that: `--live` re-reads each written gate against the conditions the evaluator applies, and dry-run runs the computable subset off the group data so the planning pass predicts rather than flatters.

**`migrate-membership-gates` no longer maps newsletter-list rules into content gates.** Those rules were inert there — the evaluator judges a list post against the premium newsletter bucket — but they still entered the rule fingerprint, so two plans differing only in their newsletter lists split into two content gates restricting identical content. Both commands also now decide, before writing anything, whether this run would create a gate alongside ones the same plans were migrated to individually — gate identity is the gate title, so any regrouping leaves the originals published and enforcing. That decision is made once up front and prompts at most once, so a run either declines to start or completes.

**`migrate-membership-gates` gains a prompt it did not have.** It previously ran start to finish unattended. It now asks for confirmation in the superseded-gate case, and hard-errors if it needs to ask when nothing can answer — so a non-interactive invocation such as `ssh host "wp newspack migrate-membership-gates --live"` needs `--yes`. Without it the run stops with a non-zero status and a message naming the flag, rather than proceeding. Two colliding gate titles are now a pre-flight error in both commands instead of a warning that continued and let the second group overwrite the first.

<details>

* A premium newsletter gate is an ordinary content gate carrying `is_newsletter` post meta. `Content_Gate::get_gates()` selects the bucket with an EXISTS / NOT EXISTS meta query, so the two never overlap — and a gate missing that flag never applies to a list, which is why verification checks it explicitly.
* No ID translation is needed for the rules themselves: `Content_Restriction_Control::rule_matches_post()` compares a `newsletters` rule's values against the list's own post ID, which is exactly what `$rule->get_object_ids()` returns.
* The auto-signup derivation does cross ID spaces. `Reader_Activation` settings hold public (ESP) list IDs while WooCommerce carries post IDs, so it resolves through `Subscription_List::get_public_id()`, the same hop the pre-Access-Control integration makes.
* Registration mode is always activated; paid access only when every plan in a group requires a purchase. The two gate modes compose with AND for a logged-in reader while WooCommerce Memberships grants access from either plan, so a mixed group stays registration-gated and the free-signup plan's members keep their lists at cutover.
* Product IDs are merged across the group, and anything that cannot grant access is dropped with a warning: non-positive values (a garbage `_product_ids` entry becomes `0`, and `variation_id` is `0` on every simple-product line item, so a `[0]` rule would match any active subscription) and IDs that no longer resolve to a product. Variation IDs are kept rather than stripped, because `WC_Subscription::has_product()` matches them directly and the parent would also admit sibling variations the plan never granted. The gate editor's product picker lists parents only, so a variation rule is invisible there — the run says so.
* `--plan` rejects a bare flag as well as a non-integer value. WP-CLI strips a valueless flag before the command runs, so without the argv check a bare `--plan --live` would widen to every plan on the site.
* Re-running overwrites a matching gate's content rules and mode settings but never its layouts — this command does not author them, and a publisher may have customized one in the Premium wizard.
* `get_plans()` and `group_requires_purchase()` are duplicated from the sibling rather than shared, so a change to one migration's plan selection cannot silently move the other's.

</details>

<details>
<summary>Dry-run output on a fixture site (4 newsletter-restricting plans, 2 without)</summary>

```
*** DRY RUN MODE — no data will be modified. Pass --live to write. ***

Found 6 membership plan(s). Starting migration…

Grouped into 3 gate(s) after deduplication.

=== DRY RUN SUMMARY ===

+-------------------------------------------+----------------------------------------+-----------+-------+-------------+
| Plan Name                                 | Action                                 | Gate ID   | Lists | Access Type |
+-------------------------------------------+----------------------------------------+-----------+-------+-------------+
| Premium Newsletter Purchase Plan          | created (dry-run)                      | (pending) | 1     | purchase    |
| Premium Newsletter Signup Plan            | created (dry-run)                      | (pending) | 1     | signup      |
| Premium Newsletter Shared Plan (Purchase) | Premium Newsletter Shared Plan (Signup) | created (dry-run) | (pending) | 1 | signup |
| Registration Wall                         | skipped (restricts no newsletter list) | —         | 0     | signup      |
| Golden Plan                               | skipped (restricts no newsletter list) | —         | 0     | purchase    |
+-------------------------------------------+----------------------------------------+-----------+-------+-------------+

Auto-signup is already on; leaving it unchanged.

Success: Done. 3 gate(s) would be created/updated.
```

Two rows are worth reading closely.

The third row is the grouping path: two plans restricting the same list share one gate, and the group resolves to `signup` even though one of the two is a purchase plan. That is `group_requires_purchase()` keeping the most permissive plan's requirement, so the free-signup plan's members do not lose the list at cutover.

That row also shows a cosmetic wart: the group title joins plan names with `" | "`, which is the summary table's own column delimiter, so a grouped row renders with a phantom extra column and the borders stop lining up. The data is correct. `migrate-membership-gates` joins the same way and has always done this; changing the separator would be a follow-up across both commands.

</details>

Needs WooCommerce Memberships, WooCommerce Subscriptions, and `newspack-newsletters` active, with Audience Management set up.

1. Create newsletter lists and membership plans covering four shapes: a purchase plan restricting one list; a signup plan restricting a different list; two plans restricting the same list; and a plan restricting only posts.
2. Run `wp newspack migrate-premium-newsletters`. The summary names the gates it would create, the two same-list plans appear as one grouped row, and the posts-only plan is skipped as restricting no newsletter list. Nothing is written.
3. Run `wp newspack migrate-premium-newsletters --live`. The gates appear under Newsletters > Premium with the right restricted lists, and the purchase-plan gate carries its products as paid-access rules.
4. Add one of the restricted lists to the post-checkout newsletter signup modal (Audience > Reader Activation) and re-run with `--live`. Auto-signup warns that the lists disagree and leaves the setting alone.
5. Re-run with `--plan=<id>` for one of the plans. The auto-signup line says the setting is left alone because the run is scoped to one plan.
6. Deactivate WooCommerce Memberships. A reader without the subscription no longer sees the premium list in signup forms; a subscriber does.
7. Run `wp newspack migrate-membership-gates`. The plan restricting only newsletter lists reports that it belongs to `migrate-premium-newsletters` rather than the misleading "no restrictions".
8. Run either command non-interactively with `--live` in a case that needs confirmation, e.g. `wp newspack migrate-premium-newsletters --live < /dev/null`. It stops with a non-zero status naming `--yes`, rather than exiting silently part-way through.

### Other information:

Full `newspack-plugin` suite passes at 3060 tests, 0 failures, 0 errors. PHPCS clean under the root ruleset.

`group_plans_by_lists()` and the gate-write loop have no unit tests: both need `WC_Memberships_Membership_Plan`, which the test harness does not load. That is the same limitation the sibling command documents, and it is why steps 1-7 above matter. Everything else in the command is covered, including the auto-signup derivation, gate verification, `get_plans()`, and the `--plan` validation.

Two follow-ups deliberately left out of this PR: the ESP expected-vs-actual delta report described in the ticket's verify sub-step, and a durable fix for gate identity surviving regrouping — this PR warns about that case rather than solving it.


plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,952 @@
...
+			// and would silently overwrite each other's rules.
+			if ( isset( $titles_written[ $gate_key ] ) ) {
+				WP_CLI::warning(
+					sprintf(
+						'Two plan groups resolve to the gate title "%s" (same plan name(s), different lists). The later group overwrites the earlier one — rename one of the plans and re-run.',
+						$gate_title
+					)
+				);
+			}
+			$titles_written[ $gate_key ] = true;
```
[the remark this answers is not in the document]
dkoo: Fixed in `3c682dd22`, as a pre-flight error rather than a mid-run abort. Aborting once the loop is running leaves every gate written before the collision in place, so the operator still has a partial migration to unpick. The collision is computable from the grouping output before anything is written, so both commands now hard-error up front and name the colliding plans.

The same restructure moved the superseded-gate prompt above the write loop, so each command now either declines to start or runs to completion.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,952 @@
...
+	 */
+	public function migrate_premium_newsletters( $args, $assoc_args ) {
+		$dry_run = ! (bool) \WP_CLI\Utils\get_flag_value( $assoc_args, 'live', false );
+
+		// A mistyped --plan must never silently widen the run to every plan, so an
+		// unusable value is a hard error rather than a fallback to "no filter".
+		$plan_arg = \WP_CLI\Utils\get_flag_value( $assoc_args, 'plan', null );
+		$plan_id  = 0;
+		if ( null !== $plan_arg ) {
+			if ( ! is_numeric( $plan_arg ) || (int) $plan_arg <= 0 ) {
```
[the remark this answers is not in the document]
dkoo: `ctype_digit` in `6ae1ca544`, in both commands.

Tracing it turned up a larger hole in the same validation: a bare `--plan` with no value. WP-CLI warns and strips a valueless flag before the command runs, so `get_flag_value()` returns `null`, `$plan_id` stays `0`, and this block never executes. `--plan --live` migrated every plan on the site, and in `migrate-premium-newsletters` also wrote the site-wide auto-signup option. The same commit scans `$_SERVER['argv']` and hard-errors, following the `--limit` guard in `class-teams-migration.php`.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,952 @@
...
+					'gate_id'     => '—',
+					'lists'       => count( $list_ids ),
+					'access_type' => $access_type,
+				];
+				continue;
+			}
+
+			// Only lists behind a gate this run actually wrote feed the site-wide
+			// auto-signup derivation; a group whose write failed migrated nothing.
+			$migrated_lists = array_merge( $migrated_lists, $list_ids );
```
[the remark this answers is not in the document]
dkoo: Fixed in `b7e99a788`, by suppressing the write rather than widening the derivation.

Two things came out of working through it. The option defaults to on, so the example case usually lands on "already on, leaving it unchanged" and writes nothing; the damaging direction is the mirror, where a modal group succeeds, a non-modal group fails, and the command turns auto-signup **off** site-wide. Including every source group's lists is not safe either: when every group fails, an empty set becomes a determinate write derived from lists that got no gate at all.

So any group that fails to write, or that verification flags as a WARN, now puts the run on the same report-only path `--plan` already used. A re-run after fixing the failure settles the setting.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,952 @@
...
+			if ( null !== $unresolvable ) {
+				$issues[] = $unresolvable;
+			}
+		}
+
+		$registration  = \Newspack\Content_Gate::get_registration_settings( $gate_id );
+		$custom_access = \Newspack\Content_Gate::get_custom_access_settings( $gate_id );
+		if ( empty( $registration['active'] ) && empty( $custom_access['active'] ) ) {
+			$issues[] = 'neither the registration nor the paid access mode is active';
+		}
```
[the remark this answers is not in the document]
dkoo: Confirmed on the create path and fixed in `3b1083743`.

The update-path half turned out to work differently: a trashed layout post leaves `gate_layout_id` truthy rather than falsy, so the gate restricts and renders nothing instead of going dormant. `verify_migrated_gate()` now checks both conditions, that the ID is set and that the post still exists, and `compute_pre_write_issues()` predicts the update-path case before writing.

The docblock claiming the omission was safe is gone. It held only while the layout inserts succeed, and `create_gate()` gives the caller no way to know they didn't.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,952 @@
...
+		return $public_id ? (string) $public_id : null;
+	}
+
+	/**
+	 * The public list IDs shown in the post-checkout newsletter signup modal.
+	 *
+	 * Mirrors the lookup the pre-Access-Control WooCommerce Memberships integration
+	 * used to decide which lists to leave to reader opt-in. With custom lists off the
+	 * modal offers every list rather than a chosen set, so the saved selection is not
+	 * a carve-out and the set is empty.
```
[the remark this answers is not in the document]
dkoo: Correct, and fixed in `ea4069561`.

The error came from reading `get_available_newsletter_lists()`, which does return every list when custom lists are off. That path feeds the registration form; the post-checkout modal sits behind `is_newsletters_signup_available()`, which is `(bool) get_setting( 'use_custom_lists' )`. Both docblocks now give the real reason.

plugins/newspack-plugin/tests/unit-tests/content-gate/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,804 @@
...
+
+		$this->assertNull( $derived['value'] );
+		$this->assertSame( [ $this->list_a ], $derived['modal'] );
+		$this->assertSame( [ $this->list_b ], $derived['non_modal'] );
+	}
+
+	/**
+	 * With custom lists off, the modal shows every list rather than a chosen set, so
+	 * the saved list selection is not the carve-out and must be ignored.
+	 */
```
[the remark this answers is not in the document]
dkoo: Fixed alongside the production docblock in `ea4069561`.

dkoo:
All six review comments are addressed, and running them down turned up four more problems in the same areas. Ready for another look.

Four of the six were correct as written. Two were right about the problem and off about the cause, and both are answered in their threads.

What the round found beyond the review comments:

- **A bare `--plan` widened the run to every plan.** WP-CLI strips a valueless flag before the command runs, so the validation never executed. With `--live` that also wrote the site-wide auto-signup option. Present in `migrate-membership-gates` too.
- **A product ID of `0` granted the gate to every subscriber.** A garbage `_product_ids` entry survived `absint`, and `variation_id` is `0` on every simple-product line item, so the rule matched any active subscription. Verification passed it because the rule was not empty. Also fixed in the sibling.
- **A prompt could evaporate.** `WP_CLI::confirm()` reads `false` from a closed STDIN and exits at status 0, so a non-interactive `--live` run stopped mid-migration and reported success. Both commands now decide before writing and hard-error if they would need to ask with nothing to answer.
- **Stale gates were unreported.** The superseded-gate check matched single plan names only, missing regroupings, renames, and deleted plans. A stale gate wins over a newer one, and for a premium newsletter that unsubscribes the reader at the ESP rather than showing a paywall.

`migrate-membership-gates` now needs `--yes` for non-interactive runs. It had no prompts before this branch, so anyone re-running it the way they ran it the first time should know. Noted in the description.

Verified against a fixture site with WooCommerce Memberships and six plans: after deactivating Memberships, the purchase gate admits only the subscribing reader, and the mixed purchase/signup group stays registration-only so the free plan's members keep the list. Bad product IDs were injected and confirmed dropped from the write. Suite is at 3060 tests, PHPCS clean.

Two cases are covered by unit tests but not yet exercised end to end: the duplicate-title pre-flight error, and the list count on a manual-only plan. Neither fixture exists on the test site.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+			// '(pending)' on a dry run, and the write path below fills it in on --live.
+			$action  = array_key_exists( $gate_key, $existing_gates ) ? 'updated' : 'created';
+			$gate_id = $existing_gates[ $gate_key ] ?? null;
+
+			$write_error = '';
+			if ( ! $dry_run ) {
+				if ( null === $gate_id ) {
+					$result = \Newspack\Content_Gate::create_gate(
+						[
+							'title'               => $payload['title'],
```
adekbadek, line 340: **Blocker:** Migrated gates publish immediately, and auto-signup subscribes readers to every premium list before cutover — `create_gate()` is called with `'status' => 'publish'`, so a migrated premium newsletter gate joins the published bucket the moment `--live` runs — while WooCommerce Memberships is still active, which is the normal state, since the command needs Memberships to read the plans.

`Premium_Newsletters::check_access()` guards only on `Content_Gate::is_gating_active()` (`class-premium-newsletters.php:312`), not on whether Memberships is still active. It then asks `Content_Restriction_Control::is_post_restricted()` per list — which returns early while Memberships is active (`class-content-restriction-control.php:342`) — and reads that `false` as entitlement, sending the list to `$lists_to_add` when auto-signup is on (`:350-352`) and on to `Newspack_Newsletters_Contacts::add_and_remove_lists()` (`:247`). This runs unattended on the hourly queue and on reader events.

So the run's closing banner — that these gates "do not restrict anything yet" and should be checked before deactivating Memberships — is true of rendering but not of the ESP sync half. In the window between `--live` and cutover, every reader entering the access-check queue is subscribed at the provider to every list the migration just gated.

**To reproduce**
1. On a site with Memberships active and Audience Management on, run `wp newspack migrate-premium-newsletters --live`.
2. Confirm the state the check runs under: `wp eval 'var_dump( Newspack\Content_Gate::is_gating_active(), (bool) get_option( "newspack_premium_newsletters_auto_signup", 1 ) );'` — both true.
3. For a reader with no membership and no subscription, evaluate each gated list: `wp eval '$r = get_user_by( "email", "reader@example.com" ); foreach ( Newspack\Premium_Newsletters::get_restricted_lists() as $l ) { var_dump( $l, Newspack\Content_Restriction_Control::is_post_restricted( false, $l, $r->ID ) ); }'`
4. Every list returns `false`, which is the branch `check_access()` treats as auto-signup. On my fixture site all five migrated lists (11, 12, 13, 46, 47) came back `false` for a reader holding zero memberships and zero subscriptions.

Either keep migrated gates out of the published bucket until cutover, or have the access-check path stand down while Memberships is active — the same condition the restriction filter already tests.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	 * @return int[] Deduplicated newsletter list post IDs, in the order WC returned them.
+	 */
+	private static function extract_list_ids( array $wc_rules ): array {
+		$list_cpt = self::get_list_cpt();
+		$list_ids = [];
+		foreach ( $wc_rules as $rule ) {
+			if ( $list_cpt !== $rule->get_content_type_name() ) {
+				continue;
+			}
+			foreach ( $rule->get_object_ids() as $object_id ) {
```
adekbadek, line 1102: **Blocker:** A plan restricting every newsletter list migrates to nothing and is reported as restricting none — `extract_list_ids()` builds the gate's lists from `$rule->get_object_ids()` alone. A Memberships content-restriction rule with `content_type_name = newspack_nl_list` and an empty object list is how Memberships expresses "all lists" — its "Search… or leave blank to apply to all" behaviour — and the list CPT is registered `public` precisely so Memberships can target it, so this shape is reachable from the plan editor.

Such a plan falls through to the `empty( $list_ids )` branch at `:1473` and is written into the summary as `skipped (restricts no newsletter list)` with `Lists = 0`. No gate is created and nothing warns, so at cutover every premium list on that site goes open.

**To reproduce**
1. Create a membership plan with a newsletter-list content restriction and leave the list selection blank.
2. Confirm Memberships is really restricting: `wp eval 'wp_set_current_user( 0 ); var_dump( wc_memberships_is_post_content_restricted( <list_post_id> ) );'` — returns `true`.
3. Run `wp newspack migrate-premium-newsletters`. The plan's row reads `skipped (restricts no newsletter list)`, Lists `0`.

Memberships exposes `has_objects()` on the rule for exactly this distinction. Reading a blank object list as "every list" — or failing the run on it — keeps the plugin's own reading of its own rule; reporting it as a no-op is the one option that loses the lists silently.

**Risk if not addressed:** If a plan leaves the list selection blank — the way Memberships expresses "all lists" — no gate is written for it, so after the flip the access check returns before touching anything and the lists reappear in public signup forms. Every premium newsletter on that site becomes free to subscribe to and free to read, and the run's output says the plan restricted nothing rather than warning.

**Consider:** Memberships exposes `has_objects()` on the rule for exactly this distinction — a newsletter rule with no object IDs targets the whole content type.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	 * subscription from everyone. WooCommerce Memberships grants access to a holder
+	 * of either plan, so the signup plan's members would silently lose their lists
+	 * at cutover. Keeping the most-permissive plan's requirement is the faithful
+	 * migration.
+	 *
+	 * @param array[] $group Plan descriptors, each carrying an 'access_method' key.
+	 *
+	 * @return bool
+	 */
+	private static function group_requires_purchase( array $group ): bool {
```
adekbadek, line 1168: **Blocker:** A group mixing a purchase plan with a signup plan drops the paywall, and neither warning path can fire — Groups are keyed by list fingerprint, so a purchase plan and a signup plan restricting the same list land in one group. `group_requires_purchase()` then returns false, `build_gate_payload()` writes `custom_access.active = false` with no subscription rule, and any reader with an account passes.

Keeping the most-permissive plan's requirement is the right call and the docblock argues it well — activating paid access would strip the signup plan's members. The problem is that the operator is never told. The two checks that name this exact outcome, `verify_migrated_gate()` at `:1310` and `compute_pre_write_issues()` at `:1355`, both sit inside `if ( $has_purchase )`, which is false for precisely this case, so neither can fire. The only signal is the word `signup` in the Access Type column, on a row counted as a plain success.

**To reproduce**
1. Create plan A (purchase, product P) and plan B (free signup), both restricting the same list L.
2. Run `wp newspack migrate-premium-newsletters --live`. One gate is written, titled `A | B`, Access Type `signup`, reported as created with no warning.
3. Inspect it: `wp eval 'var_export( get_post_meta( <gate_id>, "custom_access", true ) );'` — `active => false`, `access_rules => []`.
4. Deactivate Memberships and evaluate L for a registered reader holding no membership and no subscription — returns `false`, i.e. unrestricted. On my fixture this was gate 37 for list 13.

Note the asymmetry: a purchase group that loses all its product IDs (`custom_access.active = true`, no rules) does get a WARN row for the same reader-facing outcome. Warning here, naming the purchase plan whose requirement was dropped, would make the two consistent — and would let `count_incomplete_groups()` keep the group out of the auto-signup derivation.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+					'plan_name'   => $plan_name,
+					'action'      => 'skipped (restricts no newsletter list)',
+					'gate_id'     => '—',
+					'lists'       => '0',
+					'access_type' => $access_method,
+				];
+				continue;
+			}
+
+			$fingerprint                   = self::compute_list_fingerprint( $list_ids );
```
adekbadek, line 1782: **Blocker:** Plans whose list sets overlap without matching produce two gates, and the stricter one takes the list away — Grouping keys on `compute_list_fingerprint()`, which is exact set equality, so two plans whose restricted lists intersect but do not match land in separate groups and get separate gates. That is safe under Memberships, which grants on the union of a reader's plans, but `Content_Restriction_Control::is_post_restricted()` is restrictive-wins: it returns `true` at the first gate that restricts (`class-content-restriction-control.php:432-435`). A list named by two gates is therefore restricted unless the reader satisfies every gate naming it.

Nothing in the pre-flight detects the overlap — the run reports both gates as clean.

**To reproduce**
1. Create plan X (purchase, product P) restricting lists E and F, and plan Y (free signup) restricting only list E.
2. Run `wp newspack migrate-premium-newsletters --live`. Two gates are written, both reported clean: X with lists E+F and a `subscription` rule on P, Y with list E and no paid access.
3. Deactivate WooCommerce Memberships.
4. Evaluate list E for a registered reader who holds no subscription: `wp eval '$r = get_user_by( "email", "reader@example.com" ); var_dump( Newspack\Content_Restriction_Control::is_post_restricted( false, <list_E_id>, $r->ID ) );'`
5. Returns `true` — plan Y's members lose list E, which Memberships granted them. On my fixture this was gate 53 (lists 46, 47) and gate 56 (list 46), and list 46 evaluated `true` for a plain registered reader.

The loss is not only a render decision: `Premium_Newsletters::check_access()` puts a restricted list in `$lists_to_remove` and pushes the removal to the provider, so those readers are also unsubscribed.

**Risk if not addressed:** If two plans restrict overlapping but unequal list sets, then once WooCommerce Memberships is deactivated every reader whose only entitlement was the weaker plan is both blocked from the shared list and unsubscribed from it at the email provider on the next access-check run. Nothing warns before the flip, the unsubscribes are per reader at the provider, and re-subscribing them needs fresh consent rather than a re-run.

**Consider:** A pre-flight that stops the run when two groups' list sets intersect without matching, naming the shared lists, would catch this. Grouping is the only place the two access models can be compared while nothing has been written yet.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+					$incomplete_groups
+				)
+			);
+			return;
+		}
+		if ( $dry_run ) {
+			WP_CLI::line( sprintf( 'Auto-signup would be set to %s (currently %s).', $derived_label, $current_label ) );
+			return;
+		}
+		\update_option( 'newspack_premium_newsletters_auto_signup', $derived['value'] ? 1 : 0, false );
```
adekbadek, line 1955: **Blocker:** A live run turns site-wide auto-signup back on over a publisher’s explicit off — `report_auto_signup()` writes `newspack_premium_newsletters_auto_signup` whenever the derived value differs from the stored one, with no prompt. The derivation returns `true` whenever no restricted list is a post-checkout-modal list, which is the default outcome on any site with `use_custom_lists` off, because `get_modal_public_ids()` returns `[]` there.

So a publisher who deliberately turned auto-signup off in Newsletters > Premium has it turned back on by a migration run. The direction matters: on is the setting that generates outbound subscribe calls at the email provider. The docblock at `:1498-1524` reasons carefully about on→off being the dangerous direction and does not consider off→on.

**To reproduce**
1. `wp option update newspack_premium_newsletters_auto_signup 0`
2. Leave Reader Activation's custom signup lists off, so the derivation is determinate.
3. `wp newspack migrate-premium-newsletters --live --yes < /dev/null`
4. Output reads `Auto-signup set to on (was off).` and `wp option get newspack_premium_newsletters_auto_signup` returns `1`. No prompt was shown, and stdin was closed throughout.

The command already has `confirm_or_error()` and `--yes` for the strictly less consequential superseding-gates case.

**Risk if not addressed:** If a site has custom signup lists switched off, a live run turns auto-signup on site-wide, and the next access-check run adds every entitled reader to every restricted list at the email provider — including readers who had deliberately unsubscribed. The additions go out per reader with no record of the previous state, so undoing them means restoring each reader's list membership by hand.

**Consider:** `get_option( ..., 1 )` cannot tell "the publisher chose on" from "never set", so the comparison that decides whether to write can never see a deliberate choice. Reading the raw stored value and writing only when the option is absent, or confirming with the current and derived values named, would close that.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+
+		WP_CLI::line( sprintf( 'Found %d membership plan(s). Starting migration…', $total ) );
+		WP_CLI::line( '' );
+
+		// Pre-load existing premium newsletter gates indexed by lower-cased title. Only
+		// published gates are considered: the frontend enforces nothing else, so writing
+		// into a draft or trashed title match would produce a gate that never restricts.
+		$existing_gates = [];
+		foreach ( \Newspack\Content_Gate::get_gates( \Newspack\Content_Gate::GATE_CPT, 'publish', true ) as $gate ) {
+			$existing_gates[ trim( strtolower( $gate['title'] ) ) ] = $gate['id'];
```
adekbadek, line 181: **Suggestion:** Two published gates sharing a title collapse into one, and the twin stays enforcing — `$existing_gates[ trim( strtolower( $gate['title'] ) ) ] = $gate['id']` is last-write-wins, and nothing in `Content_Gate` enforces title uniqueness within a bucket — two published premium gates can share a title, which is easy to produce by hand in Newsletters > Premium.

The run then updates whichever `get_gates()` returned last and leaves the other published and still restricting, with no report: `report_stale_gates()` filters on the title, which is in `$titles_written`, so the abandoned twin is invisible there too.

The design rests on gate identity being the title, and the run already hard-errors when two *groups* collide on one (`:186`). The symmetric check on existing gates belongs in the same pre-flight.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+			WP_CLI::error(
+				sprintf(
+					'Two or more plan groups resolve to the same gate title: %s. A gate is identified by its title, so the second group would replace the first group\'s content rules outright and leave that group\'s lists behind no gate at all. Rename one of the plans and re-run. Nothing has been written.',
+					implode( ', ', array_map( fn( $title ) => sprintf( '"%s"', $title ), $collisions ) )
+				)
+			);
+		}
+
+		foreach ( $payloads as $payload ) {
+			self::report_product_id_issues( $payload );
```
adekbadek, line 275: **Suggestion:** Product-ID warnings fire for gates that carry no product rule — `report_product_id_issues()` runs for every payload, but `build_gate_payload()` only emits a subscription rule when `$has_purchase` is true. For a mixed purchase/signup group the collected product IDs are computed and discarded, yet the operator is still told the gate "keeps product variation ID(s) …, which is what the plan granted" and that a dropped `0` "would grant the gate to every reader with an active subscription".

Both statements are false for a gate with no access rules. Gating these warnings on `$payload['has_purchase']` keeps them true.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+				array_merge( $summary, $skipped )
+			),
+			[ 'Plan Name', 'Action', 'Gate ID', 'Lists', 'Access Type' ]
+		);
+
+		WP_CLI::line( '' );
+		self::report_stale_gates( $titles_written, $dry_run, (bool) $plan_id );
+
+		WP_CLI::line( '' );
+		self::report_auto_signup(
```
adekbadek, line 432: **Suggestion:** Auto-signup is derived from this run’s lists but governs every gate on the site — `derive_auto_signup()` sees only `$migrated_lists`, but `check_access()` applies the option to `get_restricted_lists()` — the union of the `newsletters` rules on every published premium gate, including ones this run never touched.

A pre-existing untouched gate's lists are exactly the "missing list [that] can only turn a genuine disagreement into a decision" the method's own docblock warns about, so a site with one hand-built gate on a non-modal list can have auto-signup written from a partial view. `report_stale_gates()` enumerates those gates on the line immediately above, but nothing connects the two.

Either derive over all published newsletter gates' lists, or take the report-only path when `report_stale_gates()` found anything — the same treatment `--plan` and incomplete groups already get.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	 * value can never fire on a real invocation, and a run the operator scoped to one
+	 * plan would silently widen to every plan on the site (and, under --live, write the
+	 * site-wide auto-signup setting). Reading the raw argv is the only place the
+	 * mistake is still visible.
+	 *
+	 * @param string[]|null $argv Raw argument vector; defaults to $_SERVER['argv'].
+	 *
+	 * @return string[] The value-requiring flags present without a value.
+	 */
+	public static function get_valueless_value_flags( $argv = null ): array {
```
adekbadek, line 492: **Suggestion:** Eleven methods are byte-identical across the two commands, where the stated rationale covers two — The PR describes deliberately duplicating `get_plans()` and `group_requires_purchase()`. In fact eleven methods are byte-for-byte identical between the two classes: `get_valueless_value_flags`, `is_valid_plan_arg`, `prompt_is_unanswerable`, `stdin_is_a_tty`, `confirm_or_error`, `gate_title`, `find_colliding_gate_titles`, `find_superseding_groups`, `find_superseded_gates`, plus the two declared ones. That is three coherent subsystems — argv validation, the TTY/`--yes` prompt apparatus, and title collision/supersession.

The rationale that holds for a small `get_posts()` wrapper does not obviously extend to the prompt apparatus, whose whole point is a subtle WP-CLI EOF behaviour that should not drift between two commands run back to back from the same runbook. The cost is visible in the tests too: 19 test methods share a name across the two test files, and around eleven are verbatim copies exercising verbatim-copied code.

The same split shows up in naming: `get_newsletter_list_cpt()` (gates) and `get_list_cpt()` (newsletters) are the same lookup, and `report_dropped_product_ids()` / `report_product_id_issues()` are the same reporting step. A shared trait carrying those eleven, with one set of tests, would cut both files and remove the drift risk; keeping `get_plans()` duplicated on purpose after that is an easier argument to make.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+				'slug'  => 'newsletters',
+				'value' => array_map( 'strval', $list_ids ),
+			],
+		];
+
+		$access_rules = $has_purchase && ! empty( $product_ids )
+			? [
+				[
+					[
+						'slug'  => 'subscription',
```
adekbadek, line 839: **Suggestion:** Plans granted by a one-time product migrate to a subscription rule those buyers cannot satisfy — `build_gate_payload()` emits a `subscription` rule for every purchase plan, whatever the product is. Memberships plans can be granted by simple or variable one-time products, and `Access_Rules::has_active_subscription()` only matches active subscriptions — so on such a plan the gate is written clean and the paying members fail it at cutover, losing the list and being unsubscribed at the provider.

Access Control already registers a `one_time_purchase` rule for this (`class-access-rules.php:132-143`), with its own duration handling; neither migration command uses it.

I confirmed the rule emission and the evaluator's contract by reading the code, but did not stage a one-time purchase end-to-end, so I have not observed the access failure itself. What would settle it: whether any site queued for migration has a premium newsletter plan granted by a non-subscription product. If so this is a blocker rather than a suggestion. At minimum, detecting the product type and warning would keep the run honest — `migrate-membership-gates` has the same gap, so a fix likely belongs in both.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	 *
+	 * @param array[] $group Plan descriptors, each carrying a 'product_ids' key.
+	 *
+	 * @return array 'product_ids' are the surviving IDs, in the order they appeared;
+	 *               'variations' is the subset of them that are product variations;
+	 *               'dropped' holds 'invalid' (did not normalize to a positive integer —
+	 *               a non-numeric meta value therefore appears as 0) and 'unresolvable'
+	 *               (no product post with that ID).
+	 */
+	private static function resolve_product_ids( array $group ): array {
```
adekbadek, line 923: **Suggestion:** `resolve_product_ids()` means the opposite thing in the two sibling commands — Same name, same signature, same `'dropped'` return shape — but this copy keeps product variations in the rule value while the gates copy at `class-membership-gates-migration.php:856` drops them, and the return shapes differ accordingly (`variation_ids` here, `dropped.variations` there). Both are new in this PR.

The docblock here argues the case convincingly, and the reasoning applies unchanged to the content-gate path: `Access_Rules::customer_bought_product_after()` matches on either `product_id` or `variation_id` (`class-access-rules.php:737-740`), so keeping the variation admits exactly the readers the plan admitted while dropping it narrows them. That argument indicts the sibling's behaviour, which its own docblock defends only as the behaviour it has always had.

Two files a reviewer reads side by side should not use one identifier for two opposite policies. Either name them apart, or unify the policy and file the parity ticket.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+		$superseding = self::find_superseding_groups( $plan_groups, $existing_gates );
+		foreach ( $superseding as $gate_title => $superseded ) {
+			WP_CLI::warning(
+				sprintf(
+					'"%s" merges plans that were migrated separately before. Creating it leaves these gates in place, still restricting the same lists: %s. Retire them after this run.',
+					$gate_title,
+					implode(
+						', ',
+						array_map(
+							fn( $title, $id ) => sprintf( '"%s" (gate %d)', $title, $id ),
```
adekbadek, line 296: **Nit:** The supersession warning prints gate titles lower-cased — `find_superseded_gates()` keys its return map on `trim( strtolower( $plan['name'] ) )` and the warning prints `array_keys( $superseded )`, so the operator is asked to retire gates under a name that does not match what Newsletters > Premium shows. In my run it read:

`Creating it leaves these gates in place, still restricting the same lists: "insider purchase plan" (gate 31), "insider signup plan" (gate 34).`

while the gates are titled "Insider Purchase Plan" and "Insider Signup Plan". The stale-gate warning in the same run prints correct casing, because it carries `$gate['title']` rather than the lookup key. Carrying the display title through would make the two consistent.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+					'Lists'       => $row['lists'],
+					'Access Type' => $row['access_type'],
+				],
+				array_merge( $summary, $skipped )
+			),
+			[ 'Plan Name', 'Action', 'Gate ID', 'Lists', 'Access Type' ]
+		);
+
+		WP_CLI::line( '' );
+		self::report_stale_gates( $titles_written, $dry_run, (bool) $plan_id );
```
adekbadek: **Nit:** Under `--plan`, the stale-gate report names every other gate on the site — A `--plan` run writes at most one title, so `find_stale_newsletter_gates()` returns every other published premium gate and the warning tells the operator to "retire the ones no plan accounts for". The sentence explaining why is appended after the list, which is the wrong order for a warning an operator skims.

Since a plan-scoped run structurally cannot compute staleness, skipping the report under `--plan` — with a one-line note that staleness is only computed on a full run — would be both quieter and more accurate.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	/**
+	 * Ask the operator to confirm, or hard-error when nothing can answer.
+	 *
+	 * @param string $question   The yes/no question.
+	 * @param array  $assoc_args The command's named args, passed through to WP-CLI so
+	 *                           --yes answers the prompt.
+	 *
+	 * @return void
+	 */
+	private static function confirm_or_error( string $question, array $assoc_args ) {
```
adekbadek: **Nit:** `confirm_or_error()` has no test, and four helpers are missing `: void` — `prompt_is_unanswerable()` gets three tests in each test file, but `confirm_or_error()` — the method that actually chooses between erroring out and prompting — has none, and the `WP_CLI::confirm()` mock added to `wp-cli-mocks.php` is never asserted on. Because `stdin_is_a_tty()` is always false under PHPUnit, the `WP_CLI::confirm()` branch is unreachable as written; passing the TTY result in as an argument, the way `describe_inactive_gating()` already passes its two conditions, would make both branches testable and let one test pin the thing that matters — that a superseding group under `--live` reaches the prompt before the write loop.

Separately, `confirm_or_error()`, `report_product_id_issues()` (`:780`), `report_stale_gates()` (`:976`) and `report_auto_signup()` (`:1534`) take typed parameters but declare no return type, while their neighbours in the same class do.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
```diff
@@ -0,0 +1,1611 @@
...
+	/**
+	 * Existing gates this group's plans were migrated to individually.
+	 *
+	 * Gate identity is the gate title, and a group's title is its plan names joined.
+	 * When regrouping merges plans a previous run migrated separately, the merged
+	 * title matches no existing gate — so the run creates a new gate while the
+	 * originals stay published and keep restricting the same lists. Naming them lets
+	 * the operator retire them.
+	 *
+	 * Entries whose value is null are skipped: those titles were claimed by this run
```
adekbadek: **Nit:** `find_superseded_gates()` documents and tests a state the code can no longer reach — The docblock says entries whose value is null are skipped, "claimed by this run rather than found on the site", and `:922` uses `isset()` to implement it. But `$existing_gates` is now built once from `Content_Gate::get_gates()` and never mutated — this PR removed both the null claim-marker and the write-back from the gates command, because the collision hard-error made them unnecessary.

So no null can reach here in either class, and `test_find_superseded_gates_ignores_titles_claimed_by_this_run` (in both test files) pins behaviour that is unreachable in production. The same stale paragraph is at `class-membership-gates-migration.php:556`. Worth checking `find_superseding_groups()`'s `array_key_exists()` at `:610` in the same pass, since it was presumably chosen to distinguish null from absent.

adekbadek (changes requested):
Release risk: HIGH — five reader-facing defects that only surface after WooCommerce Memberships is deactivated, each reproduced on an isolated env rather than inferred.

The command's shape is right and the reasoning behind it is unusually careful: the pre-flight/write split genuinely means an interrupted run writes nothing, the dry run predicts what `--live` does rather than flattering it, and every load-bearing claim about Newspack internals in the description checks out against the code. The eight steps in the test plan all pass.

What the testing found is a gap between what the run reports and what the site ends up in. Two of the blockers are silent by construction: the mixed purchase/signup group and the blank "all lists" rule both produce a row that reads as success, and in the mixed case the two checks that exist to catch exactly that outcome sit behind a condition that is false for it. The other three write or publish something the operator did not ask for.

The thread to read first is the pre-cutover one. Migrated gates publish immediately, and the reader-facing consequence of a premium newsletter gate is not a paywall render but a subscribe or unsubscribe at the email provider — so the run's closing reassurance that the gates "do not restrict anything yet" is true of rendering and not of the half that touches readers.

One suggestion is unresolved rather than refuted: whether any site queued for migration has a premium newsletter plan granted by a one-time product. If so it becomes a blocker, and that is a question about the fleet rather than the code.

Worth considering before any site is flipped: the first thing that acts on a migrated gate is not a paywall render but a subscription change at the email provider, and neither direction records the reader's previous list membership. A per-reader comparison of restricted lists before and after — even a sampled one — is the only check available ahead of the flip, and it has to run while Memberships is still active, because Access Control's restriction filter returns early for as long as it is.

- `plugins/newspack-plugin/includes/cli/class-membership-gates-migration.php:39` — **Suggestion:** A plan restricting both articles and lists loses its list rules with no row and no warning — `map_rules_to_ac_format()` now drops newsletter-list rules, which changes a command that sits in a published migration runbook. The explanatory summary row — `skipped (newsletter lists only — run migrate-premium-newsletters)` — only appears for plans that restrict nothing else. A plan restricting both articles and lists gets its list rules dropped silently: no row, no warning, and the description at this line still reads "Sets content rules from the shared restriction rules" unqualified, so `wp help migrate-membership-gates` never mentions the second command. That is the shape where an operator working from the runbook misses a step and leaves lists ungated at cutover. **Consider:** A WARN row naming the dropped list rules would make the dependency on the second command visible to an operator working from the existing runbook.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Fixed in `acce537fa`, taking the second option: the access-check path stands down while Memberships is active, rather than holding migrated gates out of the published bucket. Unpublished gates would add a second cutover step for someone to forget, and they are worth inspecting before the flip.

That condition now sits in `Premium_Newsletters::is_access_control_active()` and guards all four chokepoints: the access check, the queue write, the scheduled-event arming, and the renewal snapshot. So nothing accumulates during the window either.

Measured on a fixture site before the change, with both of your step 2 conditions true: all three restricted lists returned `false` for a reader holding nothing, so all three were headed for `$lists_to_add`. After the change the enqueue is a no-op.

The closing banner now covers the ESP half rather than only rendering.

One gap worth naming: this direction has no unit test. Making `Memberships::is_active()` true needs a global `WC_Memberships` class and `wc_memberships()` function, and once defined they hold for the rest of the PHPUnit run, so every `class_exists()` guard in the suite would silently take the Memberships path. That reasoning is recorded in the gating-inertness suite header, and I verified this direction against real Memberships instead.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Fixed in `2366bb3c4`, as a pre-flight error rather than reading a blank rule as every list. A gate names the lists it covers, so a snapshot of "all" is already wrong for the first list created after the run. Refusing keeps that visible instead of encoding a set that quietly goes out of date.

`restricts_all_lists()` uses `has_objects()`. I confirmed the reading in Memberships' own matcher first: `class-wc-memberships-rules.php` sets `$apply_rule = true` when the content type matches and `empty( $rule_object_ids )`, so the rule does apply to every post of the type.

The check sits after the manual-only skip, since those plans write no gate and an unbounded rule on one costs nothing.

Verified end to end against real Memberships with a blank-selection plan: the run now errors and names the plan, where it previously wrote `skipped (restricts no newsletter list)` with `Lists = 0`.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Fixed in `74a06853e`. The payload now carries `dropped_paywalls`, the purchase plans in a group that did not end up requiring a purchase, and the report loop warns on it outside the `$has_purchase` condition, naming those plans.

The permissive behaviour is unchanged; the only difference is that the operator hears about it.

On the `count_incomplete_groups()` half, I left the group counted as complete. A mixed group is a faithful migration of what Memberships grants, so pulling it out of the auto-signup derivation would drop lists that migrated correctly, and a missing list can only turn a genuine disagreement into a determinate write. The warning closes the visibility gap without changing what the derivation sees. Happy to revisit if you read that trade the other way.

It fires on the fixture env's pre-existing mixed group, which was silent before.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Fixed in `2366bb3c4`, as the pre-flight you describe. `find_lists_shared_across_groups()` runs on the grouping output and stops the run, naming each shared list and both gate titles.

It counts one entry per group rather than per plan. Plans inside a group share a gate by construction, so counting per plan would refuse every ordinary multi-plan group; that is what the third test pins, and switching it back to per-plan counting fails it.

Verified end to end with a plan overlapping two existing groups. The error names both shared lists, the gate each is reached from, and what to do about it (make the overlapping plans restrict the same set, or move the shared list onto one of them).

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Fixed in `74a06853e`, reading the raw stored value as you suggest. A null default separates absent from stored, and the run writes only when the option is absent. A stored value is reported and left alone.

I went with a `--set-auto-signup` flag rather than `confirm_or_error()`. This step runs after the write loop, and the command's pre-flight contract is that every prompt is answered before anything is written, so a prompt here would break it. A flag also survives the non-interactive invocation in your step 3, which is where this would most likely bite.

The suppression is reported before the dry-run branch, so a dry run predicts it rather than promising a write.

Three tests cover it: a stored value left alone, an absent option written, and the flag writing over a stored value. Removing the guard fails the first.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Treating this as a blocker: of the three sites queued for this migration, two have a premium newsletter plan granted by a non-subscription product. One grants on a single simple product; the other grants on a mix of subscription products and one simple product.

Fixed in `74a06853e`. Products are split by kind, and the paid access mode carries one rule group per kind, so a plan granting on both emits two groups. Groups are OR'd, so either purchase path admits the reader; flattening them into one would demand both and admit nobody.

The duration comes from each plan's own access length rather than a flag, so two plans with different lengths stay different. Unlimited maps to `forever`, and the units convert exactly into the ones the rule evaluates: a week is 7 days, a year is 12 months. `--one-time-duration` covers the one shape with no duration relative to the purchase, a plan whose access ends on a fixed calendar date. Without it, such a plan stops the run rather than migrating to a rule nobody satisfies.

You are right that `migrate-membership-gates` has the same gap. I have left it out of this PR rather than widening it further: it needs its own classification path, duration flag, and tests. Happy to open that as a follow-up, or fold it in here if you would rather keep them together.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
dkoo: Not renamed, and the reason is a change in shape underneath this thread rather than a disagreement with it.

The one-time purchase gap you raise below turned out to be a blocker, so that fix landed in both commands. Its shared machinery went into `includes/cli/trait-one-time-purchase-migration.php` rather than being copied across, since copying seven more methods into the second command would have made your duplication point worse.

`resolve_group_duration()` in that trait has to know which purchase plans actually contribute a one-time product, and answering that means validating each plan's product IDs, which is the policy-divergent step itself. It calls `self::resolve_product_ids()`, so both classes have to answer to one name. Renaming them apart means restructuring the trait so it stops calling back: the caller would pre-filter the contributing plans and pass those in, which pushes the `group_requires_purchase()` guard out of the trait and back into both commands, re-duplicating it.

What is there instead is that shared name as a declared contract. The trait's docblock records that each command defines the method for itself, that they disagree about product variations, and that the disagreement is deliberate. That puts the divergence in the one place a reader of either command reaches, which the two files could not do for each other.

That is a third option rather than the one you asked for, so say the word if you would rather have the rename and I will restructure the trait for it.

On the parity question underneath: I think you are right on the merits. `customer_bought_product_after()` matching either ID means keeping the variation admits exactly the readers the plan admitted, and the gates copy narrows them on reasoning its own docblock does not really defend. Changing what article gates grant is wider than this PR set out to be, so I have left it out; happy to file it as a follow-up if you agree that is the right call.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Confirmed. `is_access_control_active()` is the same condition `Content_Restriction_Control::is_post_restricted()` tests on its own first line, so the access path and the restriction filter now stand down together rather than disagreeing about what `false` means. All four call sites are converted — no bare `is_gating_active()` is left in the class — so nothing accumulates in the queue during the window either.

Standing the access path down rather than holding gates back is the better of the two, and for a reason beyond the cutover step: an unpublished gate is invisible to `verify_migrated_gate()`, `report_stale_gates()` and the overlap pre-flight, all of which read the published bucket. Keeping them published keeps the run's own checks honest about them.

The missing test is the right call to name rather than paper over. Defining `WC_Memberships` globally would leak into every `class_exists()` guard for the rest of the run, and a test that has to be quarantined to be correct is worse than the note you left.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Refusing is the right answer, and I agree with the reason: a snapshot of "all" is wrong for the first list created after the run, and encoding one would be the quiet failure rather than the loud one.

`has_objects()` checks out — it is a documented alias for `has_object_ids()` on `WC_Memberships_Membership_Plan_Rule`, returning `is_array( $object_ids ) && ! empty( $object_ids )`, so the reading is the plugin's own.

Two follow-ups, neither blocking:

A manual-only plan with a blank list selection still reports `lists 0` and no warning. `restricts_all_lists()` is checked after the manual-only skip, so such a plan takes the `report_manual_only_plan()` path with an empty `$list_ids`, which is exactly the branch that stays quiet. The argument for the ordering holds for the gate (none is written either way), but the row is the operator's only view of that plan, and "0 lists" is the opposite of what it restricts. Naming it in the manual-only warning would close that.

`restricts_all_lists()` itself has no unit test, and it is the check whose failure mode is every premium list going open. The anonymous rule stub in the test file is one method away from supporting it: `has_objects()` returning `! empty( $this->object_ids )` would let two tests pin both directions without WooCommerce Memberships.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Confirmed: `dropped_paywalls` is computed on the payload and `report_dropped_paywall()` runs outside the `$has_purchase` condition, so the mixed group finally has a voice. The message names the purchase plans and says what changes at cutover, which is the part the Access Type column could never convey.

I read the `count_incomplete_groups()` trade your way too. A mixed group migrated faithfully, so pulling it out of the derivation would drop lists that are fine and bias the setting from a smaller view — the opposite of what that guard is for.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Confirmed, and the per-group counting is right. Plans inside a group share a fingerprint, so `$group[0]['list_ids']` stands for the whole group and counting per plan would refuse every ordinary multi-plan group. The error naming both gate titles and both remedies is what makes it actionable rather than just a stop.

One boundary worth knowing rather than fixing: the check compares groups within the run, so it cannot see a published gate this run does not write — a hand-built one, or one left by an earlier `--plan` run — whose lists overlap a gate being created. `report_stale_gates()` names those gates and says the first restricting gate wins, so the operator has the pieces, but they are in two different messages. Fine as it stands.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Confirmed. The null default separates absent from stored, the write is gated on absent, and `--set-auto-signup` is the deliberate override. The flag over `confirm_or_error()` is the better choice for the reason you give: this step runs after the write loop, so a prompt there would break the contract that every question is answered before anything is written, and it would not survive the non-interactive invocation that is most likely to hit this.

Reporting the suppression before the dry-run branch matters more than it looks — a dry run that promised a write and a live run that declined it would be the confusing pair. The three tests pin the two directions and the override.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Addressed in both commands: `find_duplicate_gate_titles()` runs on the `get_gates()` result before anything is written, and the buckets stay separate because the `$is_newsletter` argument is EXISTS vs NOT EXISTS rather than a widening, so the gates command cannot trip over a newsletter gate's title.

Making it a hard error rather than a warning is a behaviour change for a command already in the runbook: a site with two same-named published content gates, unrelated to any plan, now cannot run `migrate-membership-gates` until one is renamed. That is the right side to err on given the alternative is updating one and abandoning its twin, and the message says what to do about it.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Fixed here — `report_product_id_issues()` returns early when the payload has no purchase, and two tests pin both directions.

The same defect is in the sibling, though, and it arrived in this PR's last commit. `report_dropped_product_ids()` is called unconditionally from `migrate-membership-gates`' pre-flight loop, while `build_access_rules()` is only reached inside `if ( $has_purchase && null !== $layouts['custom_access'] )`. So a mixed group there still gets told its gate "dropped product ID(s) N ... writing one would grant the gate to every reader with an active subscription", about a rule that was never written. The guard you added here ports across as-is.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Addressed. `report_stale_gates()` now returns whether it found anything and that answer feeds `report_auto_signup()` as `$untouched_gates`, taking the report-only path with a message that connects the two: the gates named above restrict lists the derivation never saw, and the setting governs those too. That is the version of the fix that keeps the two lines of output related rather than adjacent.

Deriving over every published newsletter gate would have been the other route, but this one is consistent with how `--plan` and incomplete groups are already handled, and it fails toward leaving the setting alone.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Confirmed as a blocker fix, and the last commit went further than your reply here says: `migrate-membership-gates` uses the trait too, splits its products by kind and builds its own two-group access rules, so both commands are covered rather than one. That answers the parity half of this thread as well.

I checked the parts I could not take on trust:

- The value shape matches what `Access_Rules::sanitize_one_time_purchase_value()` expects, and `update_custom_access_settings()` writes raw meta, so the associative value survives the round trip that a flat product-ID list used to make.
- Two rule groups do behave as OR: `evaluate_rules()` normalises, then returns true on the first group that passes, with AND inside a group. So a plan granting on both kinds admits either path, as you describe.
- The unit conversions are exact against the rule evaluator: `has_one_time_purchase()` handles only `days`, `months` and `forever`, which is what `derive_one_time_duration()` emits, and anything else fails closed. Weeks to days and years to months land inside that set rather than near it.
- `duration_rank()`'s 30-day month only orders candidates; the value stored is the plan's own, so the approximation never reaches a reader.

Refusing a fixed-end-date plan rather than migrating it to a rule nobody satisfies is the right default, and `--one-time-duration` is the right escape hatch for it.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: The third option works. Once `resolve_group_duration()` has to ask which plans contribute a one-time product, the divergent step is inside the trait's call graph, and restructuring to avoid that would push `group_requires_purchase()` back into both commands — trading one shared name for two copied methods, which is the wrong direction. The trait docblock stating that the two definitions disagree deliberately puts the divergence where a reader of either command lands, which is more than the two files were doing for each other. No rename needed.

One way to make that contract enforced rather than only documented: declare the two methods the trait calls as abstract on the trait itself (`abstract private static function resolve_product_ids( array $group ): array;` and the same for `group_requires_purchase()`, both legal as private abstracts since PHP 8.0). A class using the trait without them then fails at compile time instead of at the first call.

On parity: agreed it belongs in a follow-up rather than here. Changing what article gates grant is a different blast radius from adding a rule type, and it deserves its own testing.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Fixed in both commands: `find_superseded_gates()` keys the return on `$plan['name']` while still matching on the lower-cased key, so the operator is asked to retire gates under the name they will actually see. The comment saying why the two differ is worth keeping.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Fixed. A `--plan` run now returns early from `report_stale_gates()` with one line saying staleness cannot be computed from a run that writes one gate, and pointing at the full run. The reason leads instead of trailing the list, and `false` propagates to `report_auto_signup()`, so a plan-scoped run does not read as "no untouched gates" either.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Both fixed. `confirm_or_error()` takes the TTY result as an argument and has three tests in each file covering the prompt, the abort and `--yes`, so the branch PHPUnit could never reach is now the one under test. `: void` is on `confirm_or_error()` and `report_product_id_issues()`, `report_auto_signup()` declares `: void`, and `report_stale_gates()` declares `: bool` now that it returns one.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Fixed. The null claim-marker paragraph is gone from both docblocks, the `test_find_superseded_gates_ignores_titles_claimed_by_this_run` tests are gone with it, and `$titles_written` no longer exists in the gates command at all — the collision hard-error made both unnecessary. `find_superseding_groups()`'s `array_key_exists()` now reads as a plain presence check against a map that only ever holds gate IDs, which is what it should be.

plugins/newspack-plugin/includes/cli/class-premium-newsletters-migration.php
adekbadek: Leaving this open as a follow-up rather than a condition on the PR.

The one-time work took the right route — `trait-one-time-purchase-migration.php` rather than seven more copies — which settles the question of whether the mechanism fits these two classes. The count in this thread has gone up rather than down, though: twelve methods are now byte-identical across the two commands, `find_duplicate_gate_titles()` having joined `get_valueless_value_flags`, `is_valid_plan_arg`, `prompt_is_unanswerable`, `stdin_is_a_tty`, `confirm_or_error`, `gate_title`, `find_colliding_gate_titles`, `find_superseding_groups`, `find_superseded_gates`, `group_requires_purchase` and `get_plans`.

That is the argv/prompt/title-collision cluster, and this round is a good illustration of the drift risk: the `has_purchase` guard on the product-ID warnings landed on one side only, because the two copies had already diverged. A second trait alongside the first, carrying those twelve and one set of tests, would be mechanical now that the pattern exists — and it would cut roughly a dozen duplicated test methods with them. Happy for it to be its own PR.

adekbadek (approved):
All five blockers are fixed and verified against the code they depend on, not just the diff: the access path now stands down on the same condition the restriction filter tests, `has_objects()` is the plugin's own reading of a blank rule, the one-time rule's value shape and OR-group evaluation check out against `Access_Rules`, and the auto-signup write is gated on the option being absent. Six suggestion and nit threads landed too, several without a reply; each is answered in its thread and resolved.

Two things are worth doing before or after merge, neither blocking:
- `report_dropped_product_ids()` in `migrate-membership-gates` still fires for a mixed group that writes no paid access rule — the same defect fixed on the newsletter side in this PR, and the guard ports across unchanged.
- `wp help migrate-membership-gates` still describes content rules unqualified, so a plan restricting both articles and lists gives an operator no hint that a second command owns the list half.

The duplication thread stays open as a follow-up; it is now twelve byte-identical methods, and the one-sided fix above is what that costs.

dkoo:
Thanks, @adekbadek! I took your last two suggestions in 31ded80 and updated merge conflicts with `main`. Merging after CI reports green.

dkoo:
The failing CI jobs are due to defects currently on `main`, so they're unrelated to the changes in this PR. Merging without green CI.

PR #871, opened 2026-08-14
fix(network): escape node-supplied output in hub admin list tables

jason10lee:
### All Submissions:

The hub admin list tables display data reported by nodes across the network: event summaries, display names, product and plan names, node URLs. This PR escapes every node-supplied value at output across the hub admin surfaces, using `esc_html()` for text and `esc_url()` for links.

Tracked in NPPM-3042.

The surfaces: the Event Log, Users, Subscriptions, Orders and Membership Plans tables, and the WooCommerce Memberships row action.

**Behavior changes beyond escaping.** Two things change alongside the escaping. Routing the Subscriptions and Orders columns through the shared helper replaces `target="blank"` with `target="_blank"`; the old value opened links in a reused window named "blank" rather than a new tab. The Membership Plans catch-all now guards on `is_scalar()`, so an array column renders empty instead of `Array` plus a PHP notice.

<details>
<summary>Technical notes</summary>

* Files: `includes/hub/admin/class-event-log-list-table.php`, `includes/class-users.php`, `includes/hub/admin/class-subscriptions.php`, `includes/hub/admin/class-membership-plans-table.php`, `includes/woocommerce-memberships/class-admin.php`.
* The repeated `<a href>` output goes through a `print_link( $url, $text )` helper that escapes the href and text at the point of output. It lives on the shared `Woo` base rather than on one table, because Orders and Subscriptions render equivalent anchors from the same `Woo_Item` getters. Because the escaping is inside the helper where PHPCS can see it, both files drop their `phpcs:ignore` comments rather than annotating them.
* The event-log data column measures the raw JSON when choosing between its inline and collapsed branches, since escaping inflates each quote sixfold and would push ordinary payloads over the threshold. The collapsed branch escapes with `esc_textarea()`: `js/event-log.js` copies that element's `.value` to the clipboard, and `esc_html()` does not double-encode, so data containing `&amp;` would reach the clipboard as `&`.
* `formatted_total` stays plain text (`esc_html`), matching the sibling orders table and the source, which strips tags before syncing. No `wp_kses` is used anywhere.
* Column-method signatures are unchanged, and no other unit consumes them, so there is no cross-plugin impact.

</details>

Run `n test-php` from `plugins/newspack-network` — the six new escaping test classes pass. (The suite has one pre-existing, unrelated failure, `test_toggle_jetpack_photon`, which this PR does not touch.)

### Other information:

* [x] Have you written new tests for your changes, as applicable? Six test classes, one per admin surface.
* [x] Have you successfully run tests with your changes locally? `n test-php` is green for the changed areas; the suite is otherwise green aside from the one pre-existing unrelated failure noted above.


plugins/newspack-network/includes/hub/admin/class-subscriptions.php
```diff
@@ -81,46 +81,55 @@ public static function posts_columns_values( $column, $post_id ) {
...
+	/**
+	 * Echo an anchor with an escaped href and escaped text. Both arguments are
+	 * node-supplied and treated as untrusted.
+	 *
+	 * @param string $url    The href.
+	 * @param string $text   The link text.
+	 * @param string $target The link target attribute.
+	 * @return void
+	 */
+	private static function print_link( $url, $text, $target = '_blank' ) {
```
adekbadek: **Suggestion:** `print_link()` is the right primitive, but as a `private static` on `Subscriptions` it cannot be reached from `Orders`, which extends the same `Woo` base and renders equivalent anchors from the same getters. Promoting it to `protected static` on `Woo` covers both tables in one place and gives any future node-sourced column on either of them a safe default to reach for — which is what the ticket meant by a shared escaping helper for the hub list-table columns.

plugins/newspack-network/includes/hub/admin/class-event-log-list-table.php
```diff
@@ -143,7 +143,7 @@ protected function extra_tablenav( $which ) {
 	 * @return string
 	 */
 	protected function get_data_column( $item ) {
-		$str = wp_json_encode( $item->get_data(), JSON_PRETTY_PRINT );
+		$str = esc_html( wp_json_encode( $item->get_data(), JSON_PRETTY_PRINT ) );
```
adekbadek: **Suggestion:** `$str` is escaped before `300 > strlen( $str )` picks between the `<pre><code>` and `<textarea>` branches, so the threshold is applied to the inflated length rather than the JSON's real size. Each `"` costs six bytes instead of one, so ordinary payloads cross the line: I measured a `subscription_changed`-shaped payload at 192 bytes raw and 302 escaped, which flips it from the compact inline render to the collapsed textarea. Keeping the raw string for the test and escaping only at output restores the intended behaviour: `$json = wp_json_encode( … ); $str = esc_html( $json ); if ( 300 > strlen( $json ) ) { … }`.

Display-only either way — the textarea branch is the richer of the two, so nothing is lost. While you are on this line: the textarea's value is read by `js/event-log.js` and copied to the clipboard, and `esc_html()` does not double-encode, so an event whose data legitimately contains `&amp;` is copied back as `&`. `esc_textarea()` round-trips exactly and is the conventional escaper for that context.

plugins/newspack-network/tests/unit-tests/test-subscriptions-columns-escaping.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+class Test_Subscriptions_Columns_Escaping extends WP_UnitTestCase {
+
+	/**
+	 * Create a subscription CPT post carrying payload-bearing meta.
+	 *
+	 * @param string $payload The XSS marker payload.
+	 * @return int Post ID.
+	 */
+	private function make_subscription( $payload ) {
```
adekbadek, line 22: **Suggestion:** Every URL in these fixtures descends from `get_node_url()`, which falls back to `get_bloginfo( 'url' )` when no node post exists — so each one is already a well-formed local URL that `esc_url()` returns unchanged. Removing the `esc_url()` calls from `print_link()`, `Membership_Plans_Table::column_default()` and the memberships row action would leave this suite green, so half of the change is unpinned. The memberships test is the cheapest place to close that: the node site URL is plain post meta, so a fixture value that `esc_url()` rewrites would make the assertion bite.

plugins/newspack-network/includes/class-users.php
```diff
@@ -59,7 +59,7 @@ public static function manage_users_custom_column( $value, $column_name, $user_i
 					'<a href="%swp-admin/user-edit.php?user_id=%d">%s</a>',
 					trailingslashit( esc_url( $remote_site ) ),
```
adekbadek: **Nit:** Escaping first and mutating after is the fragile order: `esc_url()` returns `''` for a value it rejects, and `trailingslashit( '' )` is `'/'`, so the href would resolve relative to the hub rather than coming out empty. `esc_url( trailingslashit( $remote_site ) )` escapes last and avoids that. This line is unchanged by the PR and I could not find a writer of `newspack_remote_site` that stores such a value, so it is only worth folding in because you are already editing the two lines below it.

plugins/newspack-network/includes/hub/admin/class-subscriptions.php
```diff
@@ -81,46 +81,55 @@ public static function posts_columns_values( $column, $post_id ) {
...
-				$output = sprintf(
-					'<a href="%s" target="_blank">%s</a> for %s on <a href="%s" target="_blank">%s</a>',
-					$item->get_edit_link(),
-					$item->get_title(),
-					$item->get_user_name(),
-					$item->get_node_url(),
-					$item->get_node_url()
-				);
-				echo $output; // phpcs:ignore
+				self::print_link( $item->get_edit_link(), $item->get_title() );
```
adekbadek, line 84: **Nit:** `Woo_Item::get_edit_link()` bare-returns `null` when `remote_id` or `node_url` is missing, and `esc_url()` has no string cast before `ltrim()`. I confirmed the deprecation fires on PHP 8.3: `ltrim(): Passing null to parameter #1 ($string) of type string is deprecated`. The old `sprintf( '%s', null )` did not. In practice `create_item()` always writes `remote_id`, so this is latent rather than reachable — a `(string)` cast on `$url` inside `print_link()` closes it for free, and covers the sibling column too once the helper moves up.

plugins/newspack-network/includes/hub/admin/class-subscriptions.php
```diff
@@ -81,46 +81,55 @@ public static function posts_columns_values( $column, $post_id ) {
...
-				echo $output; // phpcs:ignore
 				break;
 			case 'orders':
 				$link = sprintf(
 					'%s/wp-admin/edit.php?post_status=all&post_type=shop_order&_subscription_related_orders=%d',
 					$item->get_node_url(),
 					$item->get_remote_id()
 				);
-				printf( '<a href="%s" target="blank">%s</a>', $link, $item->get_payment_count() ); // phpcs:ignore
+				self::print_link( $link, $item->get_payment_count() );
```
adekbadek, line 103: **Nit:** Routing this column through `print_link()` picks up the helper's `_blank` default, replacing the old `target="blank"` — which opened links in a reused window named "blank" rather than a new tab. It is the right value, but it is a behaviour change riding inside a PR presented as pure escaping, on a branch that ships to production. One line in the description would save the next reader working it out from the diff. The `is_scalar()` guard in the plans table is the other one; that one at least has a test documenting it.

plugins/newspack-network/tests/unit-tests/test-event-log-list-table-escaping.php
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Build a stub event-log item exposing the getters column_default reads.
+	 *
+	 * @param string $summary     Summary text.
+	 * @param string $node_url    Node URL.
+	 * @param string $action_name Action name.
+	 * @param mixed  $data        Decoded data payload.
+	 * @return object
+	 */
+	private function make_item( $summary, $node_url, $action_name, $data ) {
```
adekbadek, line 24: **Nit:** The stub and its per-getter docblocks run from line 24 to 92 — over half the file — to return four values. The sibling test in this same PR builds a *real* event-log row with `$wpdb->insert` and exercises the production item class, which is both shorter and closer to what ships. Reusing that approach here would roughly halve the file and remove the risk of the stub drifting from `Abstract_Event_Log_Item`.

plugins/newspack-network/tests/unit-tests/test-membership-plans-table-escaping.php
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+
+	/**
+	 * A non-scalar value in the default case renders empty, not "Array" + notice.
+	 */
+	public function test_default_case_guards_non_scalar() {
+		$table = new Membership_Plans_Table();
+		$item  = $this->item( 'x' );
+		$item['some_future_column'] = [ 'a', 'b' ];
```
adekbadek: **Nit:** `some_future_column` is not a key `get_columns()` produces, so the test documents a hypothetical. The real path the `is_scalar()` guard protects is `network_pass_discrepancies` — an array that reaches the catch-all whenever `network_pass_id` is falsy, and which previously rendered `Array` plus an "Array to string conversion" warning. Asserting `''` for that column with an empty `network_pass_id` would pin the behaviour the guard actually changed.

plugins/newspack-network/tests/unit-tests/test-subscriptions-columns-escaping.php
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	/**
+	 * The subscription, items and total columns must be escaped.
+	 */
+	public function test_columns_are_escaped() {
+		$payload = '<img src=x onerror=NPPM3042>';
+		$post_id = $this->make_subscription( $payload );
+
+		foreach ( [ 'subscription', 'items', 'orders', 'total' ] as $column ) {
```
adekbadek, line 78: **Nit:** Looping one payload over four columns with one shared assertion means a pass cannot say which sink earned it. I removed `esc_html()` from `print_link()` and re-ran: the suite does catch it, but on the `items` column — the `subscription` column's two assertions still passed, because the marker is satisfied by the escaped `get_user_name()` in the adjacent `printf()` and the title is neutralised by `wp_filter_post_kses` on post creation rather than by the escaping under test. Coverage is real; only the attribution is loose. Placing the payload in the title via `wp_update_post` after creation would make that column's assertion mean what it appears to mean.

adekbadek (changes requested):
The Orders table issue is the one blocker.

plugins/newspack-network/includes/hub/admin/class-subscriptions.php
jason10lee: Moved to `Woo` as `protected static`; Orders routes through it now, covered by `test-orders-columns-escaping.php`. Checked the rest of hub admin while there — the remaining suppressions are all `NonceVerification`, so Orders was the last one.

plugins/newspack-network/includes/hub/admin/class-event-log-list-table.php
jason10lee: Both taken. The threshold measures the raw JSON now, and the collapsed branch uses `esc_textarea()`. A `(string)` cast came with it: `wp_json_encode()` can return false, which the old ordering hid from `strlen()`.

plugins/newspack-network/tests/unit-tests/test-subscriptions-columns-escaping.php
jason10lee: Confirmed for `print_link()` — removing its `esc_url()` leaves the subscriptions suite green. Not for the other two: the tag-shaped payload reaches the href as well as the text, so both tests fail. Their real gap was a quote-only breakout, and the memberships and plans tests each cover one now.

plugins/newspack-network/includes/class-users.php
jason10lee: Folded in — `esc_url( trailingslashit( … ) )`, so a rejected value leaves the href empty instead of pointing at the hub.

plugins/newspack-network/includes/hub/admin/class-subscriptions.php
jason10lee: Closed by a `(string)` cast inside the helper, which covers both tables now that it sits on `Woo`.

plugins/newspack-network/includes/hub/admin/class-subscriptions.php
jason10lee: Documented in the description, together with the `is_scalar()` guard.

plugins/newspack-network/tests/unit-tests/test-event-log-list-table-escaping.php
jason10lee: Holding off on this one. It rewrites a passing test on a branch heading for production, and the drift it guards against hasn't happened yet. Better done when the file next needs a change for another reason.

plugins/newspack-network/tests/unit-tests/test-membership-plans-table-escaping.php
jason10lee: Switched to `network_pass_discrepancies` with a falsy `network_pass_id`. Removing the guard now errors on the array-to-string conversion; the old fixture passed with or without it.

plugins/newspack-network/tests/unit-tests/test-subscriptions-columns-escaping.php
jason10lee: `wp_update_post` runs kses too, so the title still landed empty — I probed create, update, and update-as-administrator, and only the last gets the payload through. With that, removing `esc_html()` reports on `subscription` instead of `items`.

adekbadek (approved):
The Orders blocker is fixed: the `order` column routes both anchors through the shared `Woo::print_link()`, and stripping either escaper out of that helper turns the Orders and Subscriptions suites red, so the coverage is real rather than incidental. Also checked the event-log threshold/`esc_textarea` change, the membership-plans `is_scalar` guard and the users-list `esc_url( trailingslashit() )` reordering — 16 new tests pass and 5 of 6 mutations were caught. The sixth: reverting the users-list escape ordering leaves the suite green, so that is the one fix here without a regression test, and it is worth pinning given the rejected-URL case it fixes otherwise resolves to the hub's own user-edit screen for the remote id. Unrelated to this branch, `test_toggle_jetpack_photon` fails at the base commit too.

PR #872, opened 2026-08-14
refactor(audience): tidy the Integrations settings layout (DSGNEWS-215)

thomasguillot:
### All Submissions:

Layout fixes to the Audience > Integrations configure view, found while testing the `CollapsibleGroup` rebuild in #869. None of them is caused by that change, so they sit here rather than in it.

**Empty containers no longer take up space.** Two of them. The inbound options list rendered its container whenever inbound sync was on, even with no options in it, leaving a 24px gap under the toggle with nothing in it. And core's `SelectControl` returns nothing when it has no options, but the Newspack wrapper still rendered its `div` around that nothing, so a field with no choices left a gap where the field would have been.

**A required setting stays on screen when it has nothing to pick.** Dropping every option-less field would also drop the ESP master list. For ActiveCampaign and Constant Contact that is the section's only setting, so the Settings heading would go with it. A required list now renders with an explanation instead of disappearing, so the Enable modal's "Open settings" button always lands on the setting it named.

There are two ways a list arrives with nothing to pick, and they need to be treated alike. `get_list_options()` returns an empty array when the call fails, but a *successful* call against an account with no audiences still returns the prepended `None` entry, so counting options reads that second case as a working list. The Enable modal already asked the right question, whether any option carries a non-empty value, so that predicate now lives in `settings-field.js` as `hasSelectableOption()` and both screens share it.

This matters because the card still reads Enabled, since that flag comes from the stored option rather than from any setting. The sync does not fail loudly either: an empty master list makes `is_set_up()` return false, `get_active_configured_integrations()` filters the integration out, and both contact sync and the hourly health check skip it before anything is raised. So the page would look configured while nothing is sent and nothing is logged.

Settings section with the required master list disabled, reading No options available

**The unusable list stays reachable.** It is marked `aria-disabled` rather than `disabled`, with a no-op change handler. A disabled `select` is not focusable, so a keyboard user would never land on the one field the Enable flow sent them to, nor on the help text explaining why it is empty. It is styled to render exactly as `disabled` did. Its help keeps the field's own description and appends the explanation, rather than replacing it and taking any `help_url` link with it.

**Settings selects now use core's control.** The Newspack `SelectControl` is marked `@deprecated` in its own docblock, and what it contributed here was an unconditional `div` around core's, which is the empty container the render guard was written to suppress. The call site uses core's directly and the wrapper is gone from the page.

**The field groups sit in the section's rhythm.** A divider now separates the last outbound setting from the first field group, so the group's first heading is divided from what precedes it the same way its own items are divided from each other. The Inbound and Outbound columns move from 16px to 24px between items, matching Settings, so the three sections share one rhythm.

**The columns are `Stack`.** Five single-column `Grid`s become `Stack` from `@wordpress/ui`, the design system's own layout primitive: `gap="xl"` for the three section columns and `gap="sm"` for the two checkbox lists. `Grid` stays where the layout is a real grid. Worth knowing for the next one of these: `Stack` sets only `display: flex`, so a vertical stack needs an explicit `direction="column"` or it lays out as a row, and its `gap` is on the token scale rather than a 4px multiplier.

**Each inbound operator select carries its own accessible name.** They shared one hidden label, so ten enabled fields announced as "Segment as" ten times over. Each now names its field: "Segment VIP as". Visuals are unchanged.

Part of DSGNEWS-215.

1. Go to Audience > Integrations and configure an ESP with grouped outbound fields.
2. In the Outbound section, confirm a divider sits between the last setting and the first field group.
3. Confirm the spacing between items is even down the Settings, Inbound and Outbound columns.
4. Expand and collapse the field groups. Only the clicked group changes state.
5. Inspect the section columns. Each is a `Stack` laying out as a column, not a row.
6. On an ESP whose account has no lists, look at the Settings section. The master list field is still there, greyed, reading "No options available", under a Settings heading. The same holds whether the list call failed or simply returned no audiences.
7. Tab to that field. It takes focus, and its help reads the field's own description followed by "No options are available. Check the connection to this integration."
8. Confirm the Inbound section shows no empty gap under its toggle when there are no incoming fields.
9. Enable an inbound field and inspect its operator select. Its accessible name names the field, for example "Segment VIP as".

### Other information:

* [x] Have you written new tests for your changes, as applicable? Adds `settings-field.test.js`, and extends `configure-view.test.js` to cover both render guards, the required-field exception, the two empty-list shapes, focusability, the group divider, the column gaps, the per-field operator labels and the legacy no-picker payload.

The divider is placed relative to the `CollapsibleGroup` introduced in #869, which has since merged, so this now sits directly on `main`.

`newspack-plugin` imports `Stack` straight from `@wordpress/ui` without declaring it, which matches how it already imports `@wordpress/icons` across 52 files: the plugin declares no `@wordpress/*` dependency at all, and the root `.npmrc` sets `public-hoist-pattern[]=@wordpress/*` so it doesn't have to. Both packages sit in the dependency-extraction plugin's `BUNDLED_PACKAGES`, so they compile into `dist/commons.js` rather than resolving to a `wp.*` global at runtime. No dependency or lockfile change.

`withMargin={ false }` also reaches the Enable modal, which renders `SettingsField` too. Nothing there changes in practice: the prop only touches the text, number and password fields, the modal only collects fields marked `required`, and every required field the plugin declares is a `select`. Checked against a Mailchimp ESP, where the modal's single control computes zero top and bottom margin.

The `hidden` case is guarded twice, at `settingsFieldRenders()` and again in the render switch. Those fields hold OAuth access and refresh tokens and their values reach the client, so a fall-through to the `text` default would print them in an editable input.

`settings-field.scss` carries one literal colour for the unusable select. It is the value core's own `COLORS.ui.textDisabled` resolves to, chosen so the control renders identically to the `disabled` state it replaces; core does not expose it as a custom property.

Tightening the predicate has one effect beyond the required case: a *non-required* select whose only option carries an empty value is now dropped rather than rendered. No built-in integration hits this, since both non-required selects ship hardcoded options.

Verified on a local environment against a Mailchimp-configured ESP with four field groups, including both empty-list shapes.


thomasguillot:
## Review

I put the parts most likely to break under test rather than reading the diff alone.

The claim this change rests on holds. At the pinned `@wordpress/components` 30.9.0, `SelectControl` returns `null` when it has no options (`build/select-control/index.js:93`), so the new guard removes an empty wrapper rather than something a publisher could previously see. Both empty states are reachable rather than theoretical: `get_list_options()` returns `[]` whenever the ESP list call errors, and `get_grouped_default_fields()` starts from `[]`, skips every section that contributes no field, and is filterable at the end. `@wordpress/ui` sits in `dependency-extraction-webpack-plugin`'s `BUNDLED_PACKAGES`, so importing `Stack` directly bundles rather than externalising, with no runtime `undefined` risk.

What came back clean, including the points called out in the description: the gap tokens match the stated intent exactly, `xl` being 24px and `sm` 8px, so the only numeric change is the intended 16px to 24px. Fragments still flatten into flex items, so the section toggle spaces as it did inside the grid. Heading order is right, with `titleLevel={ 3 }` under the section `h2`.

**Release risk: low.** No blockers. Neither changed file sits in a shared contract path, so nothing outside `newspack-plugin` is affected.

One real regression surfaced and is fixed. Newspack's `TextControl` defaults `withMargin` to true, which stamps `margin: 32px 0` on the control root. The old single-column `Grid` suppressed that through `.newspack-grid > * { margin: 0 !important }`; `Stack` has no equivalent, and these controls are no longer grid children. `metadata_prefix` is a text field in the Outbound column, so this shipped rather than staying latent, and it is invisible on the Settings column because both fields there are selects. `withMargin={ false }` restores the intended spacing, and also drops an unmatched 32px top gap in the Enable modal, the other screen `SettingsField` feeds.

One thing deliberately left out of scope. When an ESP returns no lists, the Enable modal warns "No options are available yet. Configure this integration to complete setup" and sends the publisher to this screen, which then has nothing to configure. That predates this change, since the core control already rendered nothing for an empty option set, so it wants its own issue rather than widening a layout PR.

thomasguillot:
## Final pass

What changed since the review above, in `9c2d811` through `522c743`:

- **Section guards count rendered output, not declarations.** A group whose every field renders nothing no longer leaves a heading above an empty column. `settingsFieldRenders` is exported from `settings-field.js` and is the single source of truth: `SettingsField` returns early on it, so the predicate and the component cannot drift apart.
- **The outbound guards test `grouped_options.length` rather than `!! outboundField`.** `CollapsibleGroup` renders its own wrapper for zero items, so the earlier form left an empty container and a divider separating the last setting from nothing.
- **The controls that relied on the grid reset now opt in explicitly.** `__nextHasNoMarginBottom` where it was missing, and `__next40pxDefaultSize` on the inbound operator select, which sat at 36px while every sibling select was 40px.
- **Tests.** A new `settings-field.test.js`, plus coverage for the inbound list, the field-group divider, an empty group list, the Settings section, and the legacy payload that predates the per-direction toggles. Each guard was checked by reverting it on its own and confirming a test fails, so the coverage is real rather than incidental.
- **Comments trimmed** to those that explain why rather than what.

622 tests across 66 suites, 229 across 17 in `newspack-components`, `typescript:check` clean, lint and build at their existing baselines.

Still worth a visual pass before merge: `withMargin={ false }` also reaches the Enable modal, which is a second screen this PR does not otherwise touch. The modal's own stylesheet already zeroes `.components-base-control` bottom margin because spacing there is meant to come from the `VStack` gap, so this brings the top half into line with that intent, but it is a change to a screen outside the stated scope.

thomasguillot (commented):
A second pass over the branch, against a local site running it.

No blockers. The layout work holds up: all five new `Stack`s lay out as columns with the intended 24px and 8px gaps, the group divider lands between the last outbound setting and the first field group, the Inbound section no longer leaves an empty container under its toggle, and each enabled inbound field gets its own accessible name.

Seven things came out of it, all now addressed on the branch:

1. **The empty-list check missed a connected ESP with no audiences.** `get_list_options()` returns `[]` only on error; a successful call against an account with zero audiences still gets the prepended `None` entry, so `options` had length 1 and the new branch never fired. The Enable modal reads the same payload the opposite way and offers *Open settings*, which landed the publisher on an ordinary enabled dropdown whose only choice was None. Both screens now share one `hasSelectableOption()` predicate.
2. **The empty-list help replaced the field's own description** rather than extending it, taking any `help_url` link with it. It now appends.
3. **`disabled` took the kept field out of the tab order,** so a keyboard user never reached the control the Enable flow pointed them at, nor its help. It is now `aria-disabled` with a no-op change handler, styled to match what `disabled` rendered.
4. **Half the render guard existed to work around a `@deprecated` wrapper.** The Newspack `SelectControl` emitted an unconditional div around core's, which already returns nothing for an empty list. The call site now uses core's directly and the wrapper is gone from the page.
5. **The escape hatch keyed on `required` alone,** so an optionless select with a stored value was still dropped. It now also stays when it has a value.
6. **The helper's note described the payload as it will be, not as it is.** `get_settings_config()` currently sets both `options` and `grouped_options` on the outbound metadata field.
7. **The per-field operator label had no test,** which is the one change here that regresses invisibly. There is one now.

One correction for the description rather than the code: an unset ESP master list does not make the server refuse syncs with an error. `is_set_up()` returns false, `get_active_configured_integrations()` filters the integration out, and both contact sync and the hourly health check skip it before anything is raised, so the failure mode is silence rather than a logged refusal. That is a slightly worse place to land than the description suggests, and it is what made the first point worth acting on.

913 tests and 94 suites pass, ESLint and Stylelint are clean, and each new guard fails the suite when mutated.

PR #873, opened 2026-08-14
fix(reader-activation): escape the row-action URLs in the users list

jason10lee:
### All Submissions:

Two Users-list (`wp-admin/users.php`) row-action links — the magic-link "Send authentication link" action and the contact-sync "Resync contact to ESP" action — now wrap their URL in `esc_url()` at output, matching the sibling profile-UI links in the same file that already do so.

Tracked in NPPM-3043.

Run `n test-php --filter Newspack_Test_User_Row_Actions_Escaping` from `plugins/newspack-plugin` — two tests, which fail on the pre-fix code.

### Other information:

* [x] Have you written new tests for your changes, as applicable? A new PHPUnit class covers both row actions and asserts the `href` is properly escaped; each test fails on the pre-fix code.
* [x] Have you successfully run tests with your changes locally? The new tests plus the existing Magic_Link and Contact_Sync suites pass.


plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,120 @@
...
+		wp_set_current_user( $this->admin_id );
+		$reader = get_user_by( 'id', $reader_id );
+		$this->assertInstanceOf( \WP_User::class, $reader, 'Precondition: the reader account was created.' );
+
+		$_SERVER['REQUEST_URI'] = self::PAYLOAD_URI;
+		$actions                = Magic_Link::user_row_actions( [], $reader );
+
+		$this->assertArrayHasKey( 'newspack-magic-link-send', $actions, 'Precondition: the magic-link row action is present.' );
+		$this->assert_href_not_breakable( $actions['newspack-magic-link-send'] );
+	}
```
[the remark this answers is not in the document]
jason10lee: Already in as of `95de82f39`: `set_up()` snapshots `$_SERVER['REQUEST_URI']` (`:55`) and `tear_down()` restores it, unsetting when there was nothing to restore (`:68-72`). That went up seven minutes after this review, so it wasn't in the revision under review here. `tear_down()` rather than `finally`, matching `class-ip-access-rule.php`.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,120 @@
...
+	 */
+	public function tear_down() {
+		$_SERVER['REQUEST_URI'] = '/';
+		remove_filter( 'newspack_reader_activation_enabled', '__return_true' );
+		remove_filter( 'newspack_reader_activation_is_syncing_allowed', '__return_true' );
+		delete_option( Integrations::OPTION_NAME );
+		$this->reset_integrations();
+		Integrations::register_integrations();
+		parent::tear_down();
+	}
```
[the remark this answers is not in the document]
jason10lee: Same commit — `tear_down()` resets the screen to `'front'` at `:73`.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,144 @@
...
+	 * The request URI captured before a test overrides it.
+	 *
+	 * @var string|null
+	 */
+	private $original_request_uri;
+
+	/**
+	 * Enable reader activation and act as an admin before each test.
+	 */
+	public function set_up() {
```
adekbadek, line 53: **Suggestion:** Contact-sync test is order-dependent on a shared static — `Sample_Integration` carries mutable static state, and `integrations/class-test-integrations.php` sets `Sample_Integration::$can_sync_error_codes` at `:951` and `:977` while resetting it only in its own `set_up()`. `Contact_Sync_Admin::user_row_actions()` reaches `Sync::has_one_syncable_integration()`, which calls `$integration->can_sync( true )` unconditionally (`class-sync.php:201`) regardless of the `$return_errors` argument — so a dirty static makes the row action absent and this test fails on its precondition rather than on its subject.

It does not bite today, only because unrelated later tests in that class happen to reset the static through their own `set_up()`. Sibling files that use this fixture guard against it explicitly: `integrations/class-test-integrations.php:39` and `integrations/my-account-dispatch.php:30` both call `Sample_Integration::reset()` in `set_up()`.

**To reproduce**
1. Add a throwaway test class under `tests/unit-tests/` whose name sorts before `user-row-actions-escaping.php`, setting `Sample_Integration::$can_sync_error_codes = [ 'ras_esp_master_list_id_not_found' ];` and leaving it set.
2. Run both classes in one process.
3. `test_contact_sync_row_action_escapes_request_uri` fails at line 125: `Failed asserting that an array has the key 'newspack-esp-sync'`.

Adding `Sample_Integration::reset()` to `set_up()` makes the test independent of whatever ran before it.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,144 @@
...
+	 * A well-formed `<a href="…">…</a>` carries exactly the two href-delimiter
+	 * quotes. Any reflected value that escapes the attribute introduces a third,
+	 * so the quote count is the escaping invariant — robust against a partial
+	 * filter that perturbs `<svg` but leaves the attribute-closing `"` intact,
+	 * and against breakout vectors that use no `<` (e.g. `" onmouseover=…`).
+	 *
+	 * @param string $anchor The rendered row-action anchor markup.
+	 */
+	private function assert_href_not_breakable( $anchor ) {
+		$this->assertSame( 2, substr_count( $anchor, '"' ), 'The row-action href must not be breakable out of: ' . $anchor );
```
adekbadek, line 154: **Suggestion:** The quote-count invariant passes vacuously on an empty href — `assert_href_not_breakable()` asserts only that the anchor carries two double quotes, which an anchor with an **empty** href also satisfies. That state is reachable in the code under test, not hypothetical: both `get_admin_action_url()` implementations return `''` when `! is_admin()` (`class-magic-link.php:1014-1016`, `class-contact-sync-admin.php:62-64`), while the row action is added regardless. So if the `set_current_screen( 'users' )` call in `set_up()` ever stops making `is_admin()` true, both tests stay green while asserting nothing about escaping — and the `assertArrayHasKey` preconditions do not catch it, because the action is registered either way.

**To reproduce**
1. Comment out `set_current_screen( 'users' );` at line 59.
2. Run `n test-php --filter Newspack_Test_User_Row_Actions_Escaping` — both tests pass.
3. Temporarily change the expected count in `assert_href_not_breakable()` from `2` to `999`; the failure message prints the anchor it was happy with: `<a href="">Send authentication link</a>`.

The test also never proves the other half of the contract — that the escaped link still works. Asserting the href is non-empty and still carries `uid=` and `_wpnonce=` closes the vacuous-pass gap and pins the `&` → `&#038;` behaviour in the same line, which is the part a future regression is most likely to break.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,144 @@
...
+	 * Restore global state touched by these tests.
+	 */
+	public function tear_down() {
+		if ( null === $this->original_request_uri ) {
+			unset( $_SERVER['REQUEST_URI'] );
+		} else {
+			$_SERVER['REQUEST_URI'] = $this->original_request_uri;
+		}
+		set_current_screen( 'front' );
+		remove_filter( 'newspack_reader_activation_enabled', '__return_true' );
```
adekbadek: **Nit:** Three tear_down restorations duplicate what the framework already undoes — `WP_UnitTestCase` restores the hook registry from its backup and rolls back the per-test DB transaction in `parent::tear_down()`, so the two `remove_filter()` calls at `:74-75` and the `delete_option()` at `:76` are no-ops by the time they run. The restorations that earn their place are the ones the framework does not cover: the `$_SERVER['REQUEST_URI']` snapshot at `:68-72`, `set_current_screen( 'front' )` at `:73`, and `reset_integrations()` at `:77`, since `Integrations::$integrations` is static state outside both the DB and the hook backup. Dropping the three redundant lines would leave the rest reading as a deliberate inventory of what actually leaks — worth doing only if the file is touched again for the suggestions above.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
```diff
@@ -0,0 +1,144 @@
...
+		$property->setValue( null, [] );
+	}
+
+	/**
+	 * The magic-link "Send authentication link" row action must escape the URL.
+	 */
+	public function test_magic_link_row_action_escapes_request_uri() {
+		// register_reader() must run for a logged-out visitor; then act as the admin.
+		wp_set_current_user( 0 );
+		$reader_id = Reader_Activation::register_reader( 'reader-nppm3043@example.com', 'Reader NPPM3043' );
```
adekbadek: **Nit:** Magic-link fixture goes through the full registration pipeline — The magic-link test builds its subject with `Reader_Activation::register_reader()`, which forces the logged-out/re-login step at `:98-100` and pulls welcome email, data-event dispatch and ESP sync into a test whose subject is one `esc_url()` call. `can_magic_link()` needs only reader activation enabled plus `is_user_reader()` (`class-magic-link.php:78-102`), and `is_user_reader()` is satisfied by the `np_reader` user meta (`class-reader-activation.php:1057`) — so `self::factory()->user->create()` plus one `update_user_meta()` gives the same coverage with none of the coupling. On a branch aimed at `release`, the narrower fixture is the one less likely to fail for a reason unrelated to escaping.

adekbadek (approved):
Pre-approving, just some test files suggestions and nits.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
jason10lee: Added `\Sample_Integration::reset()` to `set_up()`, matching the guard in the sibling fixtures. Confirmed the mechanism rather than assuming it: dirtying `$can_sync_error_codes` immediately after the reset fails `test_contact_sync_row_action_escapes_request_uri` with `Failed asserting that an array has the key 'newspack-esp-sync'` — the precondition failure rather than the subject, as you described.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
jason10lee: `assert_href_not_breakable()` now pulls the href out and asserts it is non-empty, still carries `uid=` and `_wpnonce=`, and encodes the separator as `&#038;`.

Ran your repro: with `set_current_screen( 'users' )` commented out, both tests now fail with `An empty href satisfies the quote count without proving anything: <a href="">Send authentication link</a>`. Both passed before. Assertions went 9 → 25.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
jason10lee: Dropped both `remove_filter()` calls and the `delete_option()`, and rewrote the docblock to record why the remaining three stay: the hook backup and the per-test transaction rollback cover what was removed, while the superglobal, the current screen and the static registry survive both.

plugins/newspack-plugin/tests/unit-tests/user-row-actions-escaping.php
jason10lee: Swapped to `self::factory()->user->create()` plus the `np_reader` meta. The precondition now asserts `Magic_Link::can_magic_link()` directly, which is the condition the test actually depends on.

jason10lee:
All four addressed in af22378d3. The file runs green alongside the integrations suite (128 tests, 364 assertions), so the order-dependence is closed in practice as well as in principle, and phpcs is clean against the root ruleset.

PR #875, opened 2026-08-14
fix(popups): preview links keep preview mode under WP 7.1 (NEWS-2889)

dkoo:
### All Submissions:

Links inside a prompt preview keep the preview going. Before this, the first click on any link inside the preview landed on the ordinary published page: the prompt stopped following the editor and unsaved settings stopped being reflected.

WordPress 7.1 serves the block editor with `Document-Isolation-Policy`, which places it in its own agent cluster and severs synchronous access to the preview iframe — even though that frame is same-origin. The editor used to reach into the frame and rewrite the links from outside, so under 7.1 it threw before rewriting anything, uncaught, on every preview open:

```
SecurityError: Failed to read a named property 'document' from 'Window':
Blocked a frame with origin "https://<site>" from accessing a cross-origin frame.
    at onLoad (…/newspack-popups/dist/editor.js)
```

The fix moves the work into the previewed document, which may always rewrite its own links. No cross-document access remains, so the error is gone at its source rather than caught and swallowed.

- `src/view/preview-links.js` (new) — the previewed page carries its own preview params onto same-origin links. It runs only inside the preview iframe, confines itself to `http(s)` (so an opaque-path href like `blob:` is left alone rather than flattened into a page URL), skips in-page anchors and off-site links, uses `searchParams.set` so a second pass cannot duplicate params, and leaves each href in the shape the page wrote it so a relative link stays relative.
- `src/view/index.js` — runs it on `domReady`, regardless of the prompt-disabled flag, since a preview has to survive navigation even where prompts are switched off. Wrapped, and ordered after the pageview push, so an admin-only nicety cannot cost a reader their pageview or their prompts.
- `src/editor/Preview.js` — drops the cross-frame rewrite.
- `includes/class-newspack-popups-inserter.php` — localizes the preview param names to the view script only where the request is an actual prompt preview: a `pid` that names a prompt, held by a user who may preview one. `pid` is a common campaign parameter, so the check matters.
- `includes/class-newspack-popups.php` — that predicate now lives in one place, `Newspack_Popups::can_preview_popup()`, shared with the renderer (`Newspack_Popups_Model::retrieve_preview_popup()`). Two copies of a security-relevant check is how they drift: a gate stricter than the renderer yields a preview whose links go nowhere, and a looser one leaks preview state onto ordinary pages.

Closes NEWS-2889.

Requires **Chrome 137 or later over HTTPS**, signed in as a user who can upload files. Cross-origin isolation is only sent to Chromium, so this will not reproduce in curl, Firefox, or Safari.

1. On WordPress 7.1, open any prompt in the editor.
2. Open the browser console. Click **Preview** in the sidebar. No `SecurityError` appears.
3. Dismiss the overlay shown inside the preview.
4. Click any link in the preview, such as a category or a post title.
5. The preview stays in preview mode: the prompt still carries its PREVIEW badge, and the front-end admin bar stays suppressed.
6. Copy the preview's own URL into a new top-level tab and click a link there. Preview mode does **not** follow you: the links on that page are ordinary ones. Preview stickiness belongs to the preview frame.

On `main` the same steps throw a `SecurityError` at step 2, and at step 5 the preview drops out to the ordinary page.

Step 5 with this branch applied — the preview has followed a category link and is still in preview mode, prompt and PREVIEW badge intact, front-end admin bar still suppressed. On `main` the badge is gone by this point and the front-end admin bar is back.

Prompt preview still in preview mode after following a link

### Other information:

Eighteen unit tests cover `propagatePreviewParams`: same-origin rewriting, the top-level and no-preview and no-matching-param cases, an absent global, off-site links, in-page anchors, `mailto:`/`tel:`, an unparseable href, SVG anchors, href shape (root-relative, absolute, path-relative, opaque-path and protocol-relative), fragment and unrelated-param preservation, and idempotency. All four negative cases assert that nothing is traversed at all, so reordering the early returns fails loudly rather than quietly walking every anchor on every page load.

Three more cover the entry point, because the propagation call shares a `domReady` callback with every reader-facing prompt handler: they pin that a throw in propagation still leaves prompt display, prompt analytics and the pageview push intact.

Six PHP tests cover the localization gate — a prompt manager previewing a prompt, a `pid` that is not a prompt, a reader and an author on a well-formed preview URL, no `pid` at all, and the localized key itself. The capability cases use an *editor*, not an administrator, matching the convention in `test-presets.php`: an administrator satisfies any capability the gate might later be tightened to, which would hide the regression.

The two behavioural fixes in the latest round were mutation-tested: removing the `http(s)` guard fails the opaque-path test, and removing the `try`/`catch` fails the entry-point tests.

Verified by hand in Chrome 151 against WordPress 7.1-RC3 with `crossOriginIsolated: true`.

The frame check the fix now depends on was measured rather than reasoned about, since the whole fix would silently no-op if `Document-Isolation-Policy` collapsed the frame tree. In a 7.1-RC3 editor, reading only cross-origin-allowlisted properties:

| frame | `document` readable from the editor | frame's own `self === top` |
| --- | --- | --- |
| `editor-canvas` (Gutenberg) | yes | false |
| `web-preview` (prompt preview) | **throws `SecurityError`** | false |

So isolation is genuinely in effect on the preview frame — that `SecurityError` is the original bug — while the frame still reports `self !== top`, with `top` and `parent` both resolving to the editor window. The guard does not fire inside a preview. As a side effect this also confirms Gutenberg's own editor canvas stays reachable under isolation.

Locally: the full JS suite passes (245 tests, 13 suites), the full PHP suite passes (417 tests, 1103 assertions), and PHPCS and ESLint are clean on every changed file.

**Not covered:** AMP. The param names are localized alongside `view.js`, which the inserter only registers on non-AMP requests, so a site in AMP standard mode keeps the pre-7.1 behaviour. Restoring it there would mean rewriting the links server-side, since AMP forbids arbitrary inline script. The module's docblock says so.

### DIFF
--- a/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
@@ -1009,30 +1009,43 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
+	 * The preview slice of the view script's localized data.
+	 *
+	 * Separate from preview_param_names() so the localized *key* is assertable, not
+	 * just the list: the front end reads `newspack_popups_view.preview_param_names`,
+	 * and renaming either side alone would leave both suites green while previews
+	 * silently stopped surviving a click. Named for the shape it has — a flat list of
+	 * param names, where newspack_popups_data.preview_query_keys is a
+	 * meta-key => param map.
+	 *
+	 * @return array Empty when this is not a prompt preview.
+	 */
+	public static function get_view_script_preview_data() {
+		$names = self::preview_param_names();
+		return empty( $names ) ? [] : [ 'preview_param_names' => $names ];
+	}
+
+	/**
 	 * The preview query param names to hand the previewed document, so it can carry
 	 * them onto same-origin links and keep the preview alive across a click.
 	 *
-	 * All three checks earn their place, because previewed_popup_id() only reports
-	 * that the request carries a `pid` — not that anyone may preview, nor that the
-	 * id is a prompt. `pid` is a common campaign parameter, so without
-	 * is_user_admin() a reader arriving on `?pid=…` would have it stamped onto every
-	 * link for the rest of their session while seeing no prompts at all; without the
-	 * post-type check an editor who follows an ad or newsletter link carrying an
-	 * unrelated `pid` would get the same. Together they make this mean "an actual
-	 * prompt preview", matching the gate on
-	 * Newspack_Popups_Model::retrieve_preview_popup(), which is what renders one.
+	 * Both checks earn their place, because previewed_popup_id() only reports that
+	 * the request carries a `pid` — not that anyone may preview, nor that the id is a
+	 * prompt. `pid` is a common campaign parameter, so without can_preview_popup() a
+	 * reader arriving on `?pid=…` would have it stamped onto every link for the rest
+	 * of their session while seeing no prompts at all, and an editor who follows an
+	 * ad or newsletter link carrying an unrelated `pid` would get the same.
 	 * is_preview_request() is the wrong test here — it also covers preset, view-as
 	 * and customizer previews, which pass their state differently.
+	 *
+	 * The JS half of this contract has its own guard: propagation only runs inside
+	 * the preview frame. See src/view/preview-links.js.
 	 *
 	 * @return array Param names, empty when this is not a prompt preview.
 	 */
 	public static function preview_param_names() {
 		$previewed_popup_id = Newspack_Popups::previewed_popup_id();
-		if (
-			! $previewed_popup_id
-			|| ! Newspack_Popups::is_user_admin()
-			|| Newspack_Popups::NEWSPACK_POPUPS_CPT !== get_post_type( $previewed_popup_id )
-		) {
+		if ( ! $previewed_popup_id || ! Newspack_Popups::can_preview_popup( $previewed_popup_id ) ) {
 			return [];
 		}
 		return array_merge(
@@ -1121,12 +1134,7 @@ final class Newspack_Popups_Inserter {
 				$script_data['donor_landing_page'] = $donor_landing_page;
 			}
 
-			// Named for the shape it has: a flat list of param names, where
-			// newspack_popups_data.preview_query_keys is a meta-key => param map.
-			$preview_param_names = self::preview_param_names();
-			if ( ! empty( $preview_param_names ) ) {
-				$script_data['preview_param_names'] = $preview_param_names;
-			}
+			$script_data = array_merge( $script_data, self::get_view_script_preview_data() );
 
 			\wp_localize_script( $script_handle, 'newspack_popups_view', $script_data );
 			\wp_enqueue_script( $script_handle );
--- a/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
@@ -1009,6 +1009,39 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
+	 * The preview query param names to hand the previewed document, so it can carry
+	 * them onto same-origin links and keep the preview alive across a click.
+	 *
+	 * All three checks earn their place, because previewed_popup_id() only reports
+	 * that the request carries a `pid` — not that anyone may preview, nor that the
+	 * id is a prompt. `pid` is a common campaign parameter, so without
+	 * is_user_admin() a reader arriving on `?pid=…` would have it stamped onto every
+	 * link for the rest of their session while seeing no prompts at all; without the
+	 * post-type check an editor who follows an ad or newsletter link carrying an
+	 * unrelated `pid` would get the same. Together they make this mean "an actual
+	 * prompt preview", matching the gate on
+	 * Newspack_Popups_Model::retrieve_preview_popup(), which is what renders one.
+	 * is_preview_request() is the wrong test here — it also covers preset, view-as
+	 * and customizer previews, which pass their state differently.
+	 *
+	 * @return array Param names, empty when this is not a prompt preview.
+	 */
+	public static function preview_param_names() {
+		$previewed_popup_id = Newspack_Popups::previewed_popup_id();
+		if (
+			! $previewed_popup_id
+			|| ! Newspack_Popups::is_user_admin()
+			|| Newspack_Popups::NEWSPACK_POPUPS_CPT !== get_post_type( $previewed_popup_id )
+		) {
+			return [];
+		}
+		return array_merge(
+			[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ],
+			array_values( Newspack_Popups::PREVIEW_QUERY_KEYS )
+		);
+	}
+
+	/**
 	 * Enqueue the assets needed to display the popups.
 	 */
 	public static function enqueue_scripts() {
@@ -1088,22 +1121,11 @@ final class Newspack_Popups_Inserter {
 				$script_data['donor_landing_page'] = $donor_landing_page;
 			}
 
-			// In a prompt preview the previewed document carries its own preview
-			// params onto same-origin links, so the preview survives navigation.
-			// It needs the param list to know which of its query params those are.
-			//
-			// Gated on the prompt preview specifically, not on is_preview_request():
-			// that covers preset, view-as and customizer previews too, and it is not
-			// a capability check — previewed_popup_id() reads `pid` from any request,
-			// signed in or not. `pid` is also a common campaign parameter, so without
-			// is_user_admin() an ordinary reader arriving on `?pid=…` would have it
-			// stamped onto every link for the rest of their session while seeing no
-			// prompts at all. preset_popup_id() guards itself the same way.
-			if ( Newspack_Popups::previewed_popup_id() && Newspack_Popups::is_user_admin() ) {
-				$script_data['preview_query_params'] = array_merge(
-					[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ],
-					array_values( Newspack_Popups::PREVIEW_QUERY_KEYS )
-				);
+			// Named for the shape it has: a flat list of param names, where
+			// newspack_popups_data.preview_query_keys is a meta-key => param map.
+			$preview_param_names = self::preview_param_names();
+			if ( ! empty( $preview_param_names ) ) {
+				$script_data['preview_param_names'] = $preview_param_names;
 			}
 
 			\wp_localize_script( $script_handle, 'newspack_popups_view', $script_data );
--- a/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
@@ -1088,6 +1088,18 @@ final class Newspack_Popups_Inserter {
 				$script_data['donor_landing_page'] = $donor_landing_page;
 			}
 
+			// In a prompt preview the previewed document carries its own preview
+			// params onto same-origin links, so the preview survives navigation.
+			// It needs the key list to know which of its query params those are.
+			// Sent only on a preview request: outside one there is nothing to
+			// propagate, and the keys would be dead weight on every page.
+			if ( Newspack_Popups::is_preview_request() ) {
+				$script_data['preview_query_keys'] = array_merge(
+					[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ],
+					array_values( Newspack_Popups::PREVIEW_QUERY_KEYS )
+				);
+			}
+
 			\wp_localize_script( $script_handle, 'newspack_popups_view', $script_data );
 			\wp_enqueue_script( $script_handle );
 		}
--- a/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
+++ b/plugins/newspack-popups/includes/class-newspack-popups-inserter.php
@@ -1090,11 +1090,17 @@ final class Newspack_Popups_Inserter {
 
 			// In a prompt preview the previewed document carries its own preview
 			// params onto same-origin links, so the preview survives navigation.
-			// It needs the key list to know which of its query params those are.
-			// Sent only on a preview request: outside one there is nothing to
-			// propagate, and the keys would be dead weight on every page.
-			if ( Newspack_Popups::is_preview_request() ) {
-				$script_data['preview_query_keys'] = array_merge(
+			// It needs the param list to know which of its query params those are.
+			//
+			// Gated on the prompt preview specifically, not on is_preview_request():
+			// that covers preset, view-as and customizer previews too, and it is not
+			// a capability check — previewed_popup_id() reads `pid` from any request,
+			// signed in or not. `pid` is also a common campaign parameter, so without
+			// is_user_admin() an ordinary reader arriving on `?pid=…` would have it
+			// stamped onto every link for the rest of their session while seeing no
+			// prompts at all. preset_popup_id() guards itself the same way.
+			if ( Newspack_Popups::previewed_popup_id() && Newspack_Popups::is_user_admin() ) {
+				$script_data['preview_query_params'] = array_merge(
 					[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ],
 					array_values( Newspack_Popups::PREVIEW_QUERY_KEYS )
 				);
--- a/plugins/newspack-popups/src/view/index.js
+++ b/plugins/newspack-popups/src/view/index.js
@@ -14,12 +14,12 @@ import { domReady, logPageview, getPrompts } from './utils';
 import './merge-tags';
 
 domReady( () => {
-	// Runs regardless of the prompt-disabled flag: a preview has to survive
-	// navigation even on a page where prompts are switched off.
-	propagatePreviewParams();
-
 	window.newspackRAS = window.newspackRAS || [];
 	window.newspackRAS.push( logPageview ); // Pageviews should be logged whether or not prompts are enabled.
+
+	// After the pageview push on purpose: link rewriting is a preview nicety, and
+	// nothing about it should be able to cost a real reader their pageview.
+	propagatePreviewParams();
 
 	if ( ! newspack_popups_view?.has_disabled_prompts ) {
 		// Fetch all prompts on the page just once.
--- a/plugins/newspack-popups/src/view/index.js
+++ b/plugins/newspack-popups/src/view/index.js
@@ -8,11 +8,16 @@ import './patterns.scss';
 import { handleSegmentation } from './segmentation';
 import { handleAnalytics } from './analytics/ga4';
 import { handleContextualPromptAnalytics } from './analytics/contextual-prompt';
+import { propagatePreviewParams } from './preview-links';
 import { domReady, logPageview, getPrompts } from './utils';
 
 import './merge-tags';
 
 domReady( () => {
+	// Runs regardless of the prompt-disabled flag: a preview has to survive
+	// navigation even on a page where prompts are switched off.
+	propagatePreviewParams();
+
 	window.newspackRAS = window.newspackRAS || [];
 	window.newspackRAS.push( logPageview ); // Pageviews should be logged whether or not prompts are enabled.
 
--- a/plugins/newspack-popups/src/view/index.js
+++ b/plugins/newspack-popups/src/view/index.js
@@ -17,9 +17,17 @@ domReady( () => {
 	window.newspackRAS = window.newspackRAS || [];
 	window.newspackRAS.push( logPageview ); // Pageviews should be logged whether or not prompts are enabled.
 
-	// After the pageview push on purpose: link rewriting is a preview nicety, and
-	// nothing about it should be able to cost a real reader their pageview.
-	propagatePreviewParams();
+	// Wrapped, and after the pageview push, on purpose. Everything below runs for
+	// every reader, while link rewriting is an admin-only preview nicety; a throw
+	// here would otherwise take prompt display and prompt analytics down with it.
+	// Nothing in it can throw today, so this keeps that a property of the structure
+	// rather than of the current implementation.
+	try {
+		propagatePreviewParams();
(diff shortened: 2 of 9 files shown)


plugins/newspack-popups/includes/class-newspack-popups-inserter.php
```diff
@@ -1088,6 +1088,24 @@ public static function enqueue_scripts() {
...
+			// It needs the param list to know which of its query params those are.
+			//
+			// Gated on the prompt preview specifically, not on is_preview_request():
+			// that covers preset, view-as and customizer previews too, and it is not
+			// a capability check — previewed_popup_id() reads `pid` from any request,
+			// signed in or not. `pid` is also a common campaign parameter, so without
+			// is_user_admin() an ordinary reader arriving on `?pid=…` would have it
+			// stamped onto every link for the rest of their session while seeing no
+			// prompts at all. preset_popup_id() guards itself the same way.
+			if ( Newspack_Popups::previewed_popup_id() && Newspack_Popups::is_user_admin() ) {
```
chickenn00dle: **Suggestion:** Preview stickiness now escapes the editor iframe into ordinary admin browsing — The gate is on the request rather than on being inside the preview modal, so propagation runs for any admin whose URL happens to carry `pid` — not only inside `WebPreview`. An editor who opens a preview link in a new tab, reloads the iframe URL top-level, or follows one pasted into chat then has `pid` plus every `n_*` param stamped onto every same-origin link for the rest of that browsing session.

Confirmed at runtime on 7.1-RC3, in a plain top-level request with no iframe involved:

| request | `preview_query_params` | admin bar |
| --- | --- | --- |
| admin, `/` | absent | present |
| admin, `/?pid=10` | all 25 localized | **absent** |

That second row is what makes this worth raising: preview mode suppresses the admin bar, so the usual cue that you are not looking at the real site is gone, and there is no exit affordance short of hand-editing the URL. It is also self-sustaining — every link on the page carries the params forward.

Before this change the stickiness was structurally confined to the frame, because the rewriting code lived in the editor. Scoping propagation to `window.self !== window.top`, or having `WebPreview` add a marker param the gate also requires, would restore the old blast radius for a couple of lines.

plugins/newspack-popups/src/view/index.js
```diff
@@ -16,6 +17,10 @@ domReady( () => {
 	window.newspackRAS = window.newspackRAS || [];
 	window.newspackRAS.push( logPageview ); // Pageviews should be logged whether or not prompts are enabled.
 
+	// After the pageview push on purpose: link rewriting is a preview nicety, and
+	// nothing about it should be able to cost a real reader their pageview.
+	propagatePreviewParams();
```
chickenn00dle: **Suggestion:** Preview-only code now runs ahead of every visitor's prompt logic — The comment correctly protects the pageview by pushing it first, but everything *after* this line — `getPrompts()`, `handleSegmentation()`, `handleAnalytics()`, `handleContextualPromptAnalytics()` — is in the same `domReady` callback and is not protected. A throw inside `propagatePreviewParams()` would take prompt display and prompt analytics down for all front-end traffic, not just previews.

Nothing in it can realistically throw today: the `a[href]` selector guarantees `getAttribute` returns a string, `new URL` is already wrapped, and the PHP always localizes a list when it localizes at all. So this is defence-in-depth rather than a live bug. But the asymmetry is stark — an admin-only nicety sitting in front of the revenue-relevant path for every reader — and right now the safety is a property of the current implementation rather than of the structure. A `try/catch` around the call, or moving it to the end of the callback, makes it structural and survives future edits to the module.

plugins/newspack-popups/src/view/preview-links.js
```diff
@@ -0,0 +1,68 @@
...
+ * preview lands on the ordinary published page, so the prompt stops following
+ * the editor and unsaved edits stop being reflected.
+ *
+ * This runs in the previewed document rather than in the editor on purpose.
+ * WordPress 7.1 serves the block editor with `Document-Isolation-Policy`, which
+ * places it in its own agent cluster and severs synchronous access to the
+ * preview iframe even though the frame is same-origin — so the editor can no
+ * longer reach in and rewrite these links from outside. A document may always
+ * rewrite its own. See NEWS-2889.
+ */
```
chickenn00dle, line 31: **Suggestion:** The same WP 7.1 breakage is left standing in two newspack-plugin previews — Out of scope for this PR, but worth an issue while the context is fresh — and worth deciding before this closes NEWS-2889, since the ticket is about the class of problem.

`newspack-plugin/src/content-gate/editor/preview.js:65` does the identical `iframeEl.contentWindow.document.querySelectorAll` reach-in from a `WebPreview` `onLoad`, and its own comment says it mirrors the popups prompt preview. It is a block-editor sidebar preview, so it hits the same `Document-Isolation-Policy` boundary under 7.1; it has a try/catch, so it degrades to a console warning rather than an uncaught error, but its preview params stop surviving navigation — the exact bug fixed here.

`newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js:53` is the worse of the two: no try/catch, and the frame access happens *before* `setIsOpen( true )` and `onLoad( iframeEl )`, so a throw leaves the modal never opening at all. That screen is a wizard rather than the block editor so it may not be isolated today, but the ordering is fragile either way.

The localization gate here is `previewed_popup_id()`, which is `pid`-specific and deliberately excludes preset, view-as and customizer previews, so neither is picked up by this fix. Generalizing `propagatePreviewParams()` so the previewed document handles all three would be the cheaper end state.

plugins/newspack-popups/src/view/preview-links.test.js
```diff
@@ -0,0 +1,142 @@
...
+const setLinks = html => {
+	document.body.innerHTML = html;
+};
+const hrefs = () => [ ...document.querySelectorAll( 'a' ) ].map( anchor => anchor.getAttribute( 'href' ) );
+
+describe( 'propagatePreviewParams', () => {
+	beforeEach( () => {
+		global.newspack_popups_view = { preview_query_params: [ 'pid', 'n_bc' ] };
+		setSearch( '' );
+		setLinks( '' );
```
chickenn00dle, line 31: **Suggestion:** The negative tests don't pin the property that matters most — `does nothing outside a preview` and `does nothing when no preview param is present` both assert only that hrefs are unchanged. Those would still pass if the function walked every anchor on the page and happened to rewrite nothing — which is precisely the regression worth guarding against, given this now executes on every front-end page load.

The contract worth pinning is "returns before any DOM traversal": a `jest.spyOn( document, 'querySelectorAll' )` assertion in those two cases would fail loudly if someone later reorders the early returns. Two smaller gaps while you are in here — the `catch` branch has no test, and nothing pins the href normalization noted separately, so a change in serialization would go unnoticed.

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
```diff
@@ -1088,6 +1088,24 @@ public static function enqueue_scripts() {
...
+				$script_data['preview_query_params'] = array_merge(
```
chickenn00dle: **Nit:** `pid` is accepted verbatim, so the collision the comment describes still applies to admins — The gating comment explains that `pid` is a common campaign parameter and that a reader arriving on `?pid=…` would otherwise get it stamped onto every link — correct, and `is_user_admin()` fixes that for readers. But `previewed_popup_id()` still returns any non-empty `pid`, so a logged-in editor who follows an ad or newsletter link carrying an unrelated `pid` gets the same stickiness, and can paste a param-laden URL somewhere.

`Newspack_Popups_Model::retrieve_preview_popup()` already rejects a `pid` that isn't a prompt CPT. Matching that check here — or reusing its result — would make the gate mean "this is actually a prompt preview" rather than "this request has a `pid`", which is what the surrounding comment already reasons as if it meant.

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
```diff
@@ -1088,6 +1088,24 @@ public static function enqueue_scripts() {
...
+					[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ],
```
chickenn00dle: **Nit:** `preview_query_params` and `preview_query_keys` are near-identical names for different shapes — `newspack_popups_view.preview_query_params` is a flat list of param names; `newspack_popups_data.preview_query_keys` is the meta-key → abbreviated-param map. Two globals, two nearly identical names, two different shapes — and newspack-plugin's gate preview reuses the `preview_query_keys` name for its own map as well. Something like `preview_param_names` on the view side would make the distinction self-evident at the call site.

plugins/newspack-popups/src/view/preview-links.js
```diff
@@ -0,0 +1,68 @@
...
+			// Resolve against the document base so relative hrefs work and a <base>
+			// tag is respected.
+			url = new URL( href, document.baseURI );
+		} catch ( e ) {
+			return;
+		}
+
+		// Off-site links, and schemes like mailto: and tel: that resolve to a
+		// null origin, are none of our business.
+		if ( url.origin !== window.location.origin ) {
```
chickenn00dle, line 111: **Nit:** Origin comparison widens scope versus the old site-URL prefix match — The removed editor code matched `a[href^="' + frontendUrl + '"]`, where `frontendUrl` was `get_site_url()`; this matches on `url.origin`. On a subdirectory install, or a subdirectory multisite network, those are not the same set — same-origin links pointing outside the WordPress install, or at a sibling network site that also runs newspack-popups, now get the params stamped on. A super-admin following such a link lands on site B in preview mode carrying site A's post ID, which resolves to nothing useful.

Harmless in practice, and arguably an improvement for multibranded sites where brands are same-origin taxonomy paths. Flagging it only because it is an unremarked scope change; comparing against the localized site URL prefix would preserve the original behaviour if that was deliberate.

plugins/newspack-popups/src/view/preview-links.js
```diff
@@ -0,0 +1,68 @@
...
+			return;
+		}
+
+		present.forEach( ( [ key, value ] ) => url.searchParams.set( key, value ) );
```
chickenn00dle, line 115: **Nit:** `url.toString()` normalizes every href, so previewed markup differs from production — `url.toString()` re-serializes the whole URL, not just the appended params. Relative hrefs become absolute, and existing query strings get re-encoded by `URLSearchParams` — `%20` → `+`, `(` → `%28`, `~` → `%7E`, `!` → `%21`. My run against real preview HTML shows the absolute-URL conversion plainly.

Servers treat those forms as equivalent, so actual breakage is limited to links whose query is compared or signed byte-for-byte, and I found none of those in the monorepo's front-end output. The subtler cost is that theme or plugin code keyed on href *shape* — `a[href^="/"]` selectors, "current item" JS comparing href to `location.pathname` — can behave differently inside a preview than on the live page. That is the exact class of false positive a preview exists to avoid. Appending to `url.search` manually, or restoring the original relative form, would keep previewed markup closer to production.

chickenn00dle (commented):
Release risk: ELEVATED — no blockers, but the change widens where preview state reaches, and its central premise (Chrome's isolation behaviour) could not be verified without a browser.

The approach is the right one. Moving the rewrite into the previewed document removes the cross-document access rather than catching around it, which is what makes it survive `Document-Isolation-Policy` at all. The new module is careful in the places that bite — reading the `href` attribute rather than the property so SVG `<a>` doesn't produce a garbage path, resolving against `document.baseURI`, and using `searchParams.set` so a second pass can't duplicate.

I verified what I could without Chrome. Against an isolated env on 7.1-RC3: the admin gate holds in both directions over real HTTP — anonymous plain load and anonymous `?pid=10` both carry no `preview_query_params`, an admin without `pid` carries none, and an admin with `?pid=10` receives all 25. Running `propagatePreviewParams()` against the actual rendered preview HTML rewrote 35 of 35 same-origin links, left both off-site links and all three in-page anchors alone, and changed nothing on a second pass. The full popups Jest suite passes here (12 suites, 236 tests) — the six failures mentioned in the description don't reproduce, so they look local to that toolchain.

What I could not verify is the part the fix exists for: the actual Chrome 137+ isolation behaviour. That still needs a human pass in Chrome.

Comments inline — the first is the one I'd most want your read on.

- `plugins/newspack-popups/includes/class-newspack-popups-inserter.php:1052` — **Suggestion:** AMP endpoints lose preview-link propagation, and nothing says so — The whole `$script_data` / `wp_localize_script` block sits inside `if ( ! self::is_amp() )`, so on a site in AMP standard mode — or a paired-mode preview that resolves to the AMP endpoint — `view.js` is never registered and the params are never propagated. Confirmed by structure: the `! is_amp()` branch opens at this line and encloses both the `preview_query_params` assignment and the `wp_localize_script` call. The old implementation was AMP-agnostic precisely because the JS lived in the editor rather than the previewed document — it rewrote the frame's links from outside regardless of what the frame served. This is an unavoidable consequence of moving the work into the document (AMP forbids arbitrary inline JS), so I am not asking you to solve it here. But it is a real behaviour difference that neither the comment here nor the module docblock mentions. A sentence in one of them would save the next person an afternoon on "preview links don't work on this site"; if AMP still matters for anyone on this path, the fallback would be filtering the rendered links server-side under the same gate.

chickenn00dle (approved):

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
dkoo: Taken — this was the right call, and it's the one finding here that changed behaviour for a real person.

I checked whether confining to the frame could cost a legitimate flow before doing it, since "it only ever runs in an iframe" is the kind of assumption that ages badly. It holds: `pid` preview URLs are constructed in exactly one place, `src/editor/Preview.js`, which renders `WebPreview`; `WebPreview` renders an `<iframe>` and offers no open-in-new-tab affordance. So `window.self === window.top` loses nothing real and restores the pre-#875 blast radius exactly.

Your admin-bar point is what tipped it. Sticky params with no admin bar and no exit affordance is a worse state than the bug I was fixing.

Mirrored into the gate module in #877, and there's a test for the top-level case in both.

plugins/newspack-popups/src/view/index.js
dkoo: Taken. Wrapped in `try`/`catch`, so the property is structural rather than a consequence of the current implementation:

```js
try {
	propagatePreviewParams();
} catch ( e ) {
	console.warn( 'Prompt preview: could not propagate preview params.', e );
}
```

Agreed on the asymmetry being the real argument — an admin-only nicety should not sit in front of prompt display and prompt analytics for every reader, whatever today's code can prove about itself. #877 does the same thing at the gate's call site, and adds two tests pinning that a throw in propagation cannot stop the gate initialising.

plugins/newspack-popups/src/view/preview-links.js
dkoo: Already handled — #877 is stacked on this branch and fixes both.

The segmentation preview turned out to be the worse of the two for a reason beyond the ordering you spotted: because `setIsOpen( true )` never ran on a throw, the effect that recomputes the decorated URL never re-ran either, so every reopened preview reused the first session ID. The state update now runs in a `finally`.

While rewriting it to match this module I also found a live bug that had nothing to do with isolation. It matched links by prefixing `frontendUrl`, which falls back to `'/'`; under that fallback a protocol-relative off-site href like `//example.com/page` also "starts with" the prefix, so it was decorated with the segment and session IDs and carried them to a third party. It compares origins now.

On generalising `propagatePreviewParams()` to cover preset, view-as and customizer previews — agreed that's the cheaper end state, but it widens what each preview type localizes, so I'd rather not fold it into a hotfix. Noted as follow-up.

plugins/newspack-popups/src/view/preview-links.test.js
dkoo: Taken, all three.

The traversal contract is now asserted through a helper, so the three negative cases fail loudly if someone reorders the early returns:

```js
const expectNoTraversal = run => {
	const spy = jest.spyOn( document, 'querySelectorAll' );
	run();
	expect( spy ).not.toHaveBeenCalled();
	spy.mockRestore();
};
```

It covers the new top-level case as well as the two you named. Also added a test for the `catch` branch — `href="http://["` throws in `new URL` and the pass has to carry on to the next anchor rather than abandoning the page — and a `href shape` block pinning normalization now that the serialization changed (see the thread below).

Sixteen tests on the module now, and the full popups suite is 240 passing.

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
dkoo: Taken. `Newspack_Popups_Model::retrieve_preview_popup()` was the right precedent to match — the two gates disagreeing about what a preview *is* was the actual defect, and your reading of the comment as describing something stronger than the code was correct.

The gate now lives in one testable place:

```php
public static function preview_param_names() {
	$previewed_popup_id = Newspack_Popups::previewed_popup_id();
	if (
		! $previewed_popup_id
		|| ! Newspack_Popups::is_user_admin()
		|| Newspack_Popups::NEWSPACK_POPUPS_CPT !== get_post_type( $previewed_popup_id )
	) {
		return [];
	}
	...
}
```

Extracting it also let me put PHP tests on it, which this PR was missing entirely: an admin previewing a prompt, a `pid` that names an ordinary post, a reader on a well-formed preview URL, and no `pid` at all. Full popups PHP suite passes — 415 tests, 1099 assertions.

Worth noting the gate side in #877 already had both checks; only this one was loose.

plugins/newspack-popups/includes/class-newspack-popups-inserter.php
dkoo: Taken — `preview_param_names` on the view side, in both plugins. Both fields are unreleased, so this was the last cheap moment to fix it.

Kept `preview_query_keys` as it is: it's the pre-existing map, it's read by the editor, and renaming it would be churn beyond this hotfix. The list is the thing that needed the clearer name.

plugins/newspack-popups/src/view/preview-links.js
dkoo: Kept the origin comparison, and documented it as deliberate — which I think was your actual complaint, since you make the case for it being an improvement yourself.

Reasoning for the record: same-origin-but-outside-the-install is what makes multibranded sites work, where brands are same-origin paths, and a stray `pid` on a page that isn't a prompt preview resolves to no prompt and does nothing. Restoring the prefix match would mean re-localizing `frontend_url`, which this PR removed as dead — a moving part reintroduced for a subdirectory-multisite case where the failure mode is a no-op.

The comment now says so, so the next reader doesn't have to work out whether it was considered.

plugins/newspack-popups/src/view/preview-links.js
dkoo: Taken, in part — and I think this one is worth slightly more than a nit.

Adding query params leaves an `a[href^="/"]` selector matching; absolutizing the href stops it matching at all. Those aren't the same size of change, and a preview that breaks the theme code it exists to exercise is the failure you describe. So relative hrefs stay relative:

```js
const isAbsolute = /^(?:[a-z][a-z0-9+.-]*:)?\/\//i.test( href );
anchor.setAttribute( 'href', isAbsolute ? url.toString() : url.pathname + url.search + url.hash );
```

Two residual differences I did not chase, both now named in the comment: a path-relative href comes back root-relative, since resolving it is what told us where it points, and `URLSearchParams` still re-encodes an existing query. Preserving that byte-for-byte means hand-editing query strings with a regex, which buys less than it risks.

New `href shape` tests pin all three cases so a future change to the serialization fails a test rather than a theme.

dkoo:
Thanks — that was a useful pass. Replies are on each thread; seven of the eight are taken, and the one I didn't take is now documented instead of silent. Pushed in 95da8e7.

**On the AMP note** (no thread to reply to, so here): confirmed, and you read the structure correctly — `if ( ! self::is_amp() )` opens above the `$script_data` block and encloses both the assignment and the `wp_localize_script` call, so on a site in AMP standard mode nothing is localized and no propagation happens. It is an unavoidable consequence of moving the work into the document, as you say, and I've taken your suggestion of writing it down rather than solving it. The module docblock now carries it:

> AMP pages are not covered. The param list is localized alongside `view.js`, which the inserter only registers on non-AMP requests, so a site in AMP standard mode keeps the pre-7.1 behaviour. Closing that would mean rewriting the links server-side, since AMP forbids arbitrary inline script.

**On the Chrome gap.** That half is verified by hand: Chrome 151 against 7.1-RC3 with `crossOriginIsolated: true` in the editor, preview opened with a clean console, and a click through a category link staying in preview mode with the PREVIEW badge and suppressed admin bar intact. The screenshot in the description is that state. What is still unverified is the *gate* preview in #877 — no content gating configured in the environment — and I've flagged that in its description as the thing worth a manual pass.

**Two additions beyond your list.** This PR had no PHP tests at all, which the tightened gate made worth fixing: four now cover it, and I ran the full popups PHP suite for the first time on this branch — 415 tests, 1099 assertions, green. The JS suite is 240 passing, and the six failures the description used to mention don't reproduce, matching what you found; I've removed that claim.

One thing to know if you re-review: #877 needed a force-push when its base moved, but this branch was fast-forward only, so your comment anchors here are intact.

dkoo:
Superseded for the WP 7.1 ship by #896, a single hotfix against `release`.

This stack is built on `main`, so it can't be retargeted to `release` — merging it there would carry 47 unrelated `main` commits with it. #896 is the same net change replayed onto `release` (two lines differ, both because `release` predates Contextual Prompts), re-verified against that tree rather than inheriting these results.

Leaving this open until #896 is green. The review history here is the record for that change, so it is worth reading rather than re-deriving — and the one open thread, on PHP type declarations, is still an open question rather than a settled decline.

dkoo:
Closing: this landed as a hotfix in #896, squash-merged to `release` as f71194149 to make the WP 7.1 ship.

#896 is the same net change replayed onto `release` and re-verified against that tree — two lines differ, both because `release` predates Contextual Prompts. Not superseded on the merits: the review here is what the change was built against, and the reasoning in these threads is the record for it.

One thing that does not travel automatically: the post-release sync of `release` into `main` will conflict in `plugins/newspack-popups/includes/class-newspack-popups-inserter.php`, because #656 added an A/B config block adjacent to this change's `$script_data` line on `main`. The resolution is to keep both additions; everything else auto-merges. That is noted in #896 too.

PR #876, opened 2026-08-14
fix: three WordPress 7.1 regressions in Newsletters

dkoo:
### All Submissions:

Three regressions that WordPress 7.1 introduces in Newsletters, found while testing the release candidate (NEWS-2699). They're small and independent, but they share a cause — 7.1 changing behaviour we relied on — so they're grouped to land together before 7.1 ships. One commit each.

**Button widths survive a re-save (NEWS-2852).** WordPress 7.1 moved `core/button` width from a top-level `width` attribute to the `dimensions` block support, so the editor rewrites stored markup the first time a newsletter is re-saved. The MJML renderer only read the old shape, so a half-width button silently became full-width in the sent email. Width reads now go through one helper that accepts either shape.

Widths arrive in more than one shape, so the helper handles all of them: a literal percentage, a `var:preset|dimension|<slug>` reference resolved to its size, and absolute units like `200px`, which no MJML column can express and which fall back to the shared default. Values outside 1-100 fall back too, since a column at `-50%` disturbs the rest of its row.

Preset references get looked at harder than core looks at them, deliberately. On the web an unresolved reference still renders, because the style engine falls back to the `--wp--preset--dimension--*` custom property theme.json emits; email clients have no custom properties, so an unresolved reference here just loses the width. The lookup therefore reads the root `dimensions.dimensionSizes` group as well as the `blocks.core/button` node core reads, includes the `blocks` origin, and tolerates a kebab-cased reference to a camelCase slug.

Where both shapes are present the legacy attribute wins, but only if it yields a usable width — otherwise it falls through to the `dimensions` value rather than discarding it. Partially-migrated content is exactly what this helper exists to survive.

Two things worth knowing about this one: the renderer gap is not version-specific — feed it the new shape on 7.0.x and it fails the same way — so any newsletter already migrated is affected today. And there was no test coverage for button width in either shape, which is why the suite stayed green through the regression. This adds it.

The WooCommerce rendering engine carries the same regression and is **deliberately not fixed here**: it sits behind a feature flag that is off by default, and the fix there is entangled with a broader question about that engine's rendering path. Tracked in NPPM-3139.

Before and after, same newsletter, email preview on 7.1-RC3 — half-width and quarter-width buttons collapse to content size, then keep their widths:

NEWS-2852 before and after: email preview where button widths collapse to content size, then keep their half and quarter widths

**Grid positioning holds (NEWS-2866).** The grid positioned children with `[start="2"]` / `[end="4"]` attribute selectors, which relied on those props reaching the DOM. The emotion-styled components in newer `@wordpress/components` filter props through `@emotion/is-prop-valid` first, and its allowlist carries `start` but not `end` — so a `VStack` asking for `end` lost its `grid-column-end` and collapsed to one column. Positioning now keys off `data-` attributes, which are never filtered, at both the medium and wide breakpoints.

This is a component-level filter rather than a WordPress-wide change: a plain `<div start="2" end="4">` still emits both. The bare selectors are kept for those plain children, but they do not rescue an emotion-styled consumer still passing `end` — that attribute never reaches the DOM to be matched.

This affects 5 call sites and only one is in Newsletters — the other four are Audience and content-gates onboarding screens in newspack-plugin.

Newsletters → Advertising empty state, before and after. It measures 228px with `grid-column-end: auto`, then 487px with `grid-column-end: 4` — the same figure the screen had before 7.1:

NEWS-2866 before and after: newsletter ads empty state, cramped at one column then restored to the full middle span

**Drafts can be saved again (NEWS-2888).** Newsletters holds a post-saving lock while a newsletter fails validation. 7.1 added that lock to core's "Save draft" disabled condition, so authors could not save work in progress until they had filled in sender name, sender email and a list. The lock is removed; sending stays gated by the Send button's own validation, which is the control that actually dispatches to the ESP.

Editor toolbar on a newsletter with unsaved changes and no sender or list. Note Send stays disabled on both sides — only saving is unblocked:

NEWS-2888 before and after: editor toolbar with Save draft greyed out, then available, with Send disabled in both

Closes NEWS-2852.
Closes NEWS-2866.
Closes NEWS-2888.

Set up a site on WordPress 7.1-RC3 with an ESP connected. An isolated env works: `n env create <name> --worktree newspack-newsletters:fix/wp-71-newsletter-regressions`, then upgrade WordPress with `wp core update --version=7.1-RC3 --force`.

**Button widths (NEWS-2852)**

1. Create a newsletter with three buttons in one Buttons block. Set the first to 50% width and the second to 25%, and leave the third alone.
2. Save, then reopen the newsletter and edit the content so the post is dirty.
3. Save again. WordPress 7.1 rewrites the button markup at this point.
4. Open the email preview. The buttons keep their relative widths.

On `main` the buttons all render full width from step 3 onward.

**Grid positioning (NEWS-2866)**

1. Go to Newsletters → Advertising with no ads created.
2. The empty state is centred and spans the middle half of the page.
3. Narrow the window to around 900px. It spans the full width rather than one column.

On `main` it renders about half as wide and sits left of centre. The same check applies to Audience → Content gates onboarding in newspack-plugin.

**Saving drafts (NEWS-2888)**

1. Create a newsletter and leave the Campaign sidebar untouched, so sender and list stay empty.
2. Edit the content.
3. "Save draft" is available, and saves.
4. The Send button stays disabled until sender and list are set.

On `main` "Save draft" is greyed out at step 3 and does nothing when clicked.

### Other information:

Verified on WordPress 7.1-RC3 against live Mailchimp and ActiveCampaign test accounts:

| | Result |
| --- | --- |
| PHPUnit (newspack-newsletters) | 763 passing, 0 failures — on both WordPress 7.0.4 (what CI installs) and 7.1-RC3 |
| Grid empty state | 487px and `grid-column: 2 / 4` at 1400px; full-width `1 / -1` at 900px (228px and `auto` before the fix) |
| Draft saving | Saving unlocked and the draft persists, with the Send button still disabled |
| Button widths | A newsletter re-saved through the 7.1 editor stores the `dimensions` shape and renders `50%, 25%, 100%` |

Two notes for reviewers:

The tests only cover the renderer. The grid and saving fixes are a stylesheet change and an effect removal, which unit tests cover poorly, so both were verified in a browser on 7.1-RC3 at both breakpoints — measurements above.

Note for anyone rebuilding: `packages/components` is consumed from its built `dist/`, so a change to its `src/` needs `npm run build` in the package before the consuming plugin picks it up.

This PR was written with AI assistance.

### DIFF
--- a/packages/components/src/grid/style.scss
+++ b/packages/components/src/grid/style.scss
@@ -83,11 +83,13 @@
 		&__columns-4 {
 			grid-template-columns: repeat(2, 1fr);
 
-			> [start] {
+			> [start],
+			> [data-start] {
 				grid-column-start: 1;
 			}
 
-			> [end] {
+			> [end],
+			> [data-end] {
 				grid-column-end: -1;
 			}
 		}
@@ -108,16 +110,19 @@
 		&__columns-4 {
 			grid-template-columns: repeat(4, 1fr);
 
-			// Positioning is keyed off `data-` attributes because WordPress 7.1
-			// stopped forwarding unrecognised props through to the DOM: `start`
-			// survives (it is a real attribute, on `<ol>`), `end` does not, so a
-			// child asking for `end` silently lost its `grid-column-end` and
-			// collapsed to a single column (NEWS-2866). `data-` attributes are
-			// always forwarded, so they hold regardless of the component in use.
+			// Positioning is keyed off `data-` attributes because the emotion-styled
+			// components in newer `@wordpress/components` filter props through
+			// `@emotion/is-prop-valid` before they reach the DOM. Its allowlist
+			// carries `start` (a real attribute, on `<ol>`) but not `end`, so a
+			// `VStack` asking for `end` silently lost its `grid-column-end` and
+			// collapsed to one column (NEWS-2866). `data-` attributes are never
+			// filtered, so they hold whatever component renders them.
 			//
-			// The bare-attribute selectors are kept so any consumer still passing
-			// `start`/`end` renders as it did before, rather than breaking on the
-			// way past.
+			// This is a component-level filter, not a WordPress-wide one: a plain
+			// `<div start="2" end="4">` still emits both attributes. The bare
+			// selectors below are kept for those plain children — they do NOT
+			// rescue an emotion-styled consumer still passing `end`, which never
+			// reaches the DOM to be matched.
 			> [start="2"],
 			> [data-start="2"] {
 				grid-column-start: 2;
--- a/packages/components/src/grid/style.scss
+++ b/packages/components/src/grid/style.scss
@@ -108,11 +108,23 @@
 		&__columns-4 {
 			grid-template-columns: repeat(4, 1fr);
 
-			> [start="2"] {
+			// Positioning is keyed off `data-` attributes because WordPress 7.1
+			// stopped forwarding unrecognised props through to the DOM: `start`
+			// survives (it is a real attribute, on `<ol>`), `end` does not, so a
+			// child asking for `end` silently lost its `grid-column-end` and
+			// collapsed to a single column (NEWS-2866). `data-` attributes are
+			// always forwarded, so they hold regardless of the component in use.
+			//
+			// The bare-attribute selectors are kept so any consumer still passing
+			// `start`/`end` renders as it did before, rather than breaking on the
+			// way past.
+			> [start="2"],
+			> [data-start="2"] {
 				grid-column-start: 2;
 			}
 
-			> [end="4"] {
+			> [end="4"],
+			> [data-end="4"] {
 				grid-column-end: 4;
 			}
 		}
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
@@ -903,26 +903,82 @@ final class Newspack_Newsletters_Renderer {
 	 * Buttons carry their width in one of two shapes. Newsletters authored before
 	 * WordPress 7.1 store a bare `width` attribute; WordPress 7.1 moved
 	 * `core/button` to the `dimensions` block support, so the editor rewrites the
-	 * markup to a percentage string under `style.dimensions.width` the first time a
-	 * newsletter is re-saved. Both forms can coexist in the same newsletter, so
-	 * every width read goes through here (NEWS-2852).
+	 * markup under `style.dimensions.width` the first time a newsletter is
+	 * re-saved. Both forms can coexist in the same newsletter, so every width read
+	 * goes through here (NEWS-2852).
 	 *
-	 * Only percentages translate to MJML column widths. Any other unit is treated
-	 * as "no width set", which lets the button fall back to the shared default
-	 * rather than producing a nonsense column.
+	 * The `dimensions` shape has three variants, and WordPress resolves all of them
+	 * before rendering (`wp-includes/blocks/button.php`): a literal percentage, a
+	 * `var:preset|dimension|<slug>` reference to a `dimensionSizes` preset (core
+	 * ships 25/50/75/100 for buttons, so this is reachable on every site), and an
+	 * absolute unit such as `200px`. Only percentages map onto an MJML column, so
+	 * presets are resolved first and absolute units are treated as "no width set" —
+	 * the button then takes the shared default rather than producing a nonsense
+	 * column.
+	 *
+	 * Widths outside 1–100 are also treated as unset. Nothing in the block editor
+	 * can produce them, but hand-edited or migrated content can, and an
+	 * `mj-column` with a negative or >100% width breaks the whole row rather than
+	 * just that button.
+	 *
+	 * Safe to call with either raw parsed-block attributes or the output of
+	 * `process_attributes()`, which touches neither width key.
 	 *
 	 * @param array $attrs A core/button block's attributes.
-	 * @return int|null Width as a percentage, or null when unset.
+	 * @return float|null Width as a percentage, or null when unset.
 	 */
 	private static function get_button_width( $attrs ) {
-		if ( isset( $attrs['width'] ) && '' !== $attrs['width'] ) {
-			return intval( $attrs['width'] );
+		$width = null;
+
+		if ( isset( $attrs['width'] ) && is_scalar( $attrs['width'] ) && '' !== $attrs['width'] ) {
+			$width = $attrs['width'];
+		} elseif ( isset( $attrs['style']['dimensions']['width'] ) && is_scalar( $attrs['style']['dimensions']['width'] ) ) {
+			$width = self::resolve_dimension_preset( (string) $attrs['style']['dimensions']['width'] );
 		}
-		$dimensions_width = $attrs['style']['dimensions']['width'] ?? null;
-		if ( null === $dimensions_width || ! preg_match( '/^\s*(\d+(?:\.\d+)?)\s*%\s*$/', (string) $dimensions_width, $matches ) ) {
+
+		if ( null === $width || ! preg_match( '/^\s*(\d+(?:\.\d+)?)\s*%?\s*$/', (string) $width, $matches ) ) {
 			return null;
 		}
-		return intval( $matches[1] );
+
+		$percentage = (float) $matches[1];
+
+		return ( $percentage > 0 && $percentage <= 100 ) ? $percentage : null;
+	}
+
+	/**
+	 * Resolve a `dimensions.width` value to its literal size.
+	 *
+	 * Mirrors the preset lookup WordPress does in `wp-includes/blocks/button.php`:
+	 * a `var:preset|dimension|<slug>` reference is looked up against the
+	 * `dimensionSizes` presets, searching custom, then theme, then default origins.
+	 * Any other value is returned unchanged.
+	 *
+	 * @param string $width The raw `style.dimensions.width` value.
+	 * @return string The resolved width.
+	 */
+	private static function resolve_dimension_preset( $width ) {
+		$prefix = 'var:preset|dimension|';
+		if ( 0 !== strpos( $width, $prefix ) ) {
+			return $width;
+		}
+
+		$slug    = substr( $width, strlen( $prefix ) );
+		$presets = wp_get_global_settings( [ 'dimensions', 'dimensionSizes' ], [ 'block_name' => 'core/button' ] );
+
+		if ( is_array( $presets ) ) {
+			foreach ( [ 'custom', 'theme', 'default' ] as $origin ) {
+				if ( empty( $presets[ $origin ] ) || ! is_array( $presets[ $origin ] ) ) {
+					continue;
+				}
+				foreach ( $presets[ $origin ] as $preset ) {
+					if ( isset( $preset['slug'] ) && (string) $preset['slug'] === $slug ) {
+						return (string) ( $preset['size'] ?? $width );
+					}
+				}
+			}
+		}
+
+		return $width;
 	}
 
 	/**
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
@@ -907,73 +907,105 @@ final class Newspack_Newsletters_Renderer {
 	 * re-saved. Both forms can coexist in the same newsletter, so every width read
 	 * goes through here (NEWS-2852).
 	 *
-	 * The `dimensions` shape has three variants, and WordPress resolves all of them
-	 * before rendering (`wp-includes/blocks/button.php`): a literal percentage, a
-	 * `var:preset|dimension|<slug>` reference to a `dimensionSizes` preset (core
-	 * ships 25/50/75/100 for buttons, so this is reachable on every site), and an
-	 * absolute unit such as `200px`. Only percentages map onto an MJML column, so
-	 * presets are resolved first and absolute units are treated as "no width set" —
-	 * the button then takes the shared default rather than producing a nonsense
-	 * column.
+	 * The legacy attribute is preferred when it yields a usable width, but an
+	 * unusable one falls through to the `dimensions` value rather than discarding
+	 * it — partially-migrated content is exactly what this helper exists to
+	 * survive.
 	 *
-	 * Widths outside 1–100 are also treated as unset. Nothing in the block editor
-	 * can produce them, but hand-edited or migrated content can, and an
-	 * `mj-column` with a negative or >100% width breaks the whole row rather than
-	 * just that button.
+	 * A width is usable only if it resolves to a percentage in 1-100. Absolute
+	 * units have no meaning as an MJML column share, and a column outside that
+	 * range disturbs the rest of its row; either way the button falls back to the
+	 * default share, which is what a width-less button gets.
 	 *
 	 * Safe to call with either raw parsed-block attributes or the output of
 	 * `process_attributes()`, which touches neither width key.
 	 *
 	 * @param array $attrs A core/button block's attributes.
-	 * @return float|null Width as a percentage, or null when unset.
+	 * @return float|null Width as a percentage, or null when unset or unusable.
 	 */
-	private static function get_button_width( $attrs ) {
-		$width = null;
+	private static function get_button_width( array $attrs ): ?float {
+		$candidates = [];
 
-		if ( isset( $attrs['width'] ) && is_scalar( $attrs['width'] ) && '' !== $attrs['width'] ) {
-			$width = $attrs['width'];
-		} elseif ( isset( $attrs['style']['dimensions']['width'] ) && is_scalar( $attrs['style']['dimensions']['width'] ) ) {
-			$width = self::resolve_dimension_preset( (string) $attrs['style']['dimensions']['width'] );
+		if ( isset( $attrs['width'] ) && is_scalar( $attrs['width'] ) ) {
+			$candidates[] = (string) $attrs['width'];
+		}
+		if ( isset( $attrs['style']['dimensions']['width'] ) && is_scalar( $attrs['style']['dimensions']['width'] ) ) {
+			$candidates[] = self::resolve_dimension_preset( (string) $attrs['style']['dimensions']['width'] );
 		}
 
-		if ( null === $width || ! preg_match( '/^\s*(\d+(?:\.\d+)?)\s*%?\s*$/', (string) $width, $matches ) ) {
-			return null;
+		foreach ( $candidates as $candidate ) {
+			if ( ! preg_match( '/^\s*(\d+(?:\.\d+)?)\s*%?\s*$/', $candidate, $matches ) ) {
+				continue;
+			}
+			$percentage = (float) $matches[1];
+			if ( $percentage > 0 && $percentage <= 100 ) {
+				return $percentage;
+			}
 		}
 
-		$percentage = (float) $matches[1];
-
-		return ( $percentage > 0 && $percentage <= 100 ) ? $percentage : null;
+		return null;
 	}
 
 	/**
-	 * Resolve a `dimensions.width` value to its literal size.
+	 * Resolve a `dimensions.width` preset reference to its literal size.
 	 *
-	 * Mirrors the preset lookup WordPress does in `wp-includes/blocks/button.php`:
-	 * a `var:preset|dimension|<slug>` reference is looked up against the
-	 * `dimensionSizes` presets, searching custom, then theme, then default origins.
-	 * Any other value is returned unchanged.
+	 * WordPress stores a chosen preset as `var:preset|dimension|<slug>` rather than
+	 * a size. On the web an unresolved reference still renders, because the style
+	 * engine falls back to the `--wp--preset--dimension--<slug>` custom property
+	 * that theme.json emits. Email clients do not support custom properties, so the
+	 * reference has to be resolved here or the width is simply lost — which is why
+	 * this looks harder for the preset than core does.
+	 *
+	 * Core's own lookup (`render_block_core_button()`) reads only the
+	 * `blocks.core/button` node, with the origins and exact-slug match mirrored
+	 * below. That is enough for core because of the custom-property fallback; it
+	 * isn't enough for us, so this also reads the root `dimensions.dimensionSizes`
+	 * group, includes the `blocks` origin that `wp_theme_json_data_blocks` writes
+	 * to, and tolerates a kebab-cased reference to a camelCase slug.
 	 *
 	 * @param string $width The raw `style.dimensions.width` value.
-	 * @return string The resolved width.
+	 * @return string The resolved width, or the input unchanged when it isn't a
+	 *                preset reference or names an unknown preset.
 	 */
-	private static function resolve_dimension_preset( $width ) {
+	private static function resolve_dimension_preset( string $width ): string {
 		$prefix = 'var:preset|dimension|';
-		if ( 0 !== strpos( $width, $prefix ) ) {
+		if ( ! str_starts_with( $width, $prefix ) ) {
 			return $width;
 		}
+		$slug = substr( $width, strlen( $prefix ) );
 
-		$slug    = substr( $width, strlen( $prefix ) );
-		$presets = wp_get_global_settings( [ 'dimensions', 'dimensionSizes' ], [ 'block_name' => 'core/button' ] );
+		// Read the settings once and index into them. `wp_get_global_settings()`
+		// returns the *whole* settings array when the requested path is absent, so
+		// asking it for a path and testing `is_array()` on the result never fails —
+		// it just hands back a different array to walk.
+		$settings = wp_get_global_settings();
+		$groups   = [];
+		if ( isset( $settings['blocks']['core/button']['dimensions']['dimensionSizes'] ) ) {
+			$groups[] = $settings['blocks']['core/button']['dimensions']['dimensionSizes'];
+		}
+		if ( isset( $settings['dimensions']['dimensionSizes'] ) ) {
+			$groups[] = $settings['dimensions']['dimensionSizes'];
+		}
 
-		if ( is_array( $presets ) ) {
-			foreach ( [ 'custom', 'theme', 'default' ] as $origin ) {
+		foreach ( $groups as $presets ) {
+			if ( ! is_array( $presets ) ) {
+				continue;
+			}
+			// Origin priority, most specific first.
+			foreach ( [ 'custom', 'theme', 'blocks', 'default' ] as $origin ) {
 				if ( empty( $presets[ $origin ] ) || ! is_array( $presets[ $origin ] ) ) {
 					continue;
 				}
 				foreach ( $presets[ $origin ] as $preset ) {
-					if ( isset( $preset['slug'] ) && (string) $preset['slug'] === $slug ) {
-						return (string) ( $preset['size'] ?? $width );
+					if ( ! is_array( $preset ) || ! isset( $preset['slug'] ) || ! is_scalar( $preset['slug'] ) ) {
+						continue;
 					}
+					$preset_slug = (string) $preset['slug'];
+					if ( $preset_slug !== $slug && _wp_to_kebab_case( $preset_slug ) !== $slug ) {
+						continue;
+					}
+					$size = $preset['size'] ?? null;
+					return is_scalar( $size ) ? (string) $size : $width;
 				}
 			}
 		}
@@ -1405,6 +1437,11 @@ final class Newspack_Newsletters_Renderer {
 						$button_attrs['css-class'] = $attrs['className'];
 					}
 
+					// NOTE: the width below goes on $default_button_attrs, which is what
+					// the emit at the end of this case actually serializes. The
+					// $button_attrs built just above is dead (pre-existing) — if that
+					// is ever cleaned up, move this width with it or buttons silently
+					// lose their width again.
 					$column_attrs['css-class'] = 'mj-column-has-width';
 					$column_width              = $default_width;
 					$button_width              = self::get_button_width( $attrs );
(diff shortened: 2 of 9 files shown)


plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,34 @@ private static function cleanup_rdb_html( $content ) {
...
+	private static function get_button_width( $attrs ) {
+		if ( isset( $attrs['width'] ) && '' !== $attrs['width'] ) {
+			return intval( $attrs['width'] );
+		}
+		$dimensions_width = $attrs['style']['dimensions']['width'] ?? null;
+		if ( null === $dimensions_width || ! preg_match( '/^\s*(\d+(?:\.\d+)?)\s*%\s*$/', (string) $dimensions_width, $matches ) ) {
+			return null;
+		}
+		return intval( $matches[1] );
+	}
```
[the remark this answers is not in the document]
dkoo: Good catch — confirmed and fixed in b5e323032.

You're right that this was a behaviour change: `empty( $block['attrs']['width'] )` treated `0` as unset, and the helper didn't. Reproduced across all three shapes before fixing (`0`, `"0"`, and `"0%"` each produced a `width="0%"` column).

Rather than special-case zero, the helper now normalises the whole range: a width is used only when it resolves to a percentage in 1-100, so negatives and values above 100 fall back to the shared default as well. Those two were already possible on `main` — a legacy `-50` produced `width="-50%"` — and an `mj-column` with a nonsense width breaks the whole row rather than just that button, so it seemed worth closing together.

Two related things came out of testing it:

- The legacy branch was passing everything through `intval()`, so `"50px"` silently became a 50% column. It now has to match the same percentage pattern as the `dimensions` branch.
- `style.dimensions.width` can also hold `var:preset|dimension|<slug>`, which WordPress resolves against the `dimensionSizes` presets before rendering. Core ships 25/50/75/100 for buttons, so that shape is reachable on any site; the helper now resolves it the same way.

Test coverage went from 3 cases to 12, including zero, negative, above-100, absolute units, an unresolvable preset, and `dimensions` with no `width` key.

packages/components/src/grid/style.scss
```diff
@@ -108,11 +108,23 @@
...
+			// Positioning is keyed off `data-` attributes because WordPress 7.1
+			// stopped forwarding unrecognised props through to the DOM: `start`
+			// survives (it is a real attribute, on `<ol>`), `end` does not, so a
+			// child asking for `end` silently lost its `grid-column-end` and
+			// collapsed to a single column (NEWS-2866). `data-` attributes are
+			// always forwarded, so they hold regardless of the component in use.
+			//
+			// The bare-attribute selectors are kept so any consumer still passing
+			// `start`/`end` renders as it did before, rather than breaking on the
+			// way past.
```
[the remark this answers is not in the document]
dkoo: Confirmed and fixed in 8b41881c0 — thanks, this was a real gap.

Verified before fixing by reading the rules the browser had actually loaded: the 744px band carried only `> [start]` / `> [end]`, so an element with just `data-start`/`data-end` got neither `grid-column-start: 1` nor `grid-column-end: -1`. The change had traded a bug at ≥1054px for one at 744-1053px.

Both selectors at that breakpoint now match the `data-` attributes too. Measured on the Newsletters → Advertising empty state:

| Viewport | Band | `grid-column` | Width |
| --- | --- | --- | --- |
| 900px | 744-1053 | `1 / -1` | 864 / 864 |
| 1400px | ≥1054 | `2 / 4` | 487 / 1006 |

Investigating this also turned up something worth correcting in the comment I'd left above those selectors. It said the bare selectors keep any consumer still passing `start`/`end` working — that isn't true for `end`. The emotion-styled components in newer `@wordpress/components` filter props through `@emotion/is-prop-valid`, and its HTML allowlist carries `start` but not `end` (`end` appears only in its SVG list). So a `VStack` still passing `end` never emits the attribute, and no selector can match it — the bare selectors only help plain, non-emotion children. The comment now says that, and names the filter rather than attributing it to WordPress generally, since a plain `<div start="2" end="4">` still emits both.

dkoo:
Pushed three commits addressing the review. Both inline threads are answered and resolved.

**Button widths — 8b41881c0, b5e323032**

The width helper now normalises the whole range rather than reading a raw value: a width applies only when it resolves to a percentage in 1-100. That covers the zero case raised in review, and also negatives and values above 100, which were already possible on `main`.

Two more shapes came out of testing it. `style.dimensions.width` can hold `var:preset|dimension|<slug>`, which WordPress resolves against the `dimensionSizes` presets before rendering — core ships 25/50/75/100 for buttons, so a newsletter using one would have lost its width entirely. And the legacy branch was routing everything through `intval()`, so `"50px"` silently became a 50% column. Both are handled now, and fractional widths keep their precision (`33.9%` was truncating to 33, which left three equal thirds summing to 99%).

**Grid positioning — 8b41881c0**

The `data-` selectors were only added at the 1054px breakpoint, so the change had traded a bug at wide widths for one between 744-1053px. Both breakpoints now match. Measured on the Advertising empty state: `1 / -1` and full width at 900px, `2 / 4` and 487px of 1006px at 1400px.

The comment explaining the change was also wrong in a way worth fixing: the bare `[start]`/`[end]` selectors don't rescue a consumer still passing `end`, because `@emotion/is-prop-valid` filters that prop before it reaches the DOM. They only help plain, non-emotion children. There's one such consumer left in `newspack-manager`, which needs the same migration in its own repo — it isn't fixed by this PR.

**Editor — 735421cca**

Dropped a prop left dead by the lock removal, and moved the explanatory comment to the top of the component. It had also claimed saving isn't gated at all, which is too broad — `editor/mjml` still takes a transient lock around the post-save HTML refresh, and that would misdirect anyone triaging a future "Save draft is greyed out" report.

**Verification**

PHPUnit 761 passing, 0 failures — button-width coverage went from 3 cases to 12, including the shapes above. PHPCS and ESLint clean. Both grid breakpoints and the draft-saving path re-checked in a browser on 7.1-RC3.

**Still out of scope, deliberately:** the WooCommerce rendering engine carries the same button-width regression. It's behind a feature flag that's off by default, and the fix there is entangled with a broader question about that engine's rendering path, so it's tracked separately in NPPM-3139 rather than folded in here.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,90 @@ private static function cleanup_rdb_html( $content ) {
...
+	 * @return string The resolved width.
+	 */
+	private static function resolve_dimension_preset( $width ) {
+		$prefix = 'var:preset|dimension|';
+		if ( 0 !== strpos( $width, $prefix ) ) {
+			return $width;
+		}
+
+		$slug    = substr( $width, strlen( $prefix ) );
+		$presets = wp_get_global_settings( [ 'dimensions', 'dimensionSizes' ], [ 'block_name' => 'core/button' ] );
```
chickenn00dle: **Blocker:** Preset lookup misses theme-defined presets, silently dropping the width — `wp_get_global_settings( [ 'dimensions', 'dimensionSizes' ], [ 'block_name' => 'core/button' ] )` resolves against the `blocks/core/button` node only — it does not fall back to the root `dimensions.dimensionSizes`, which is where core registers the preset group and the natural place for a theme to declare its own. The editor that writes the attribute does fall back (it tries the block node, then the root), so a theme can offer a width in the button control, store `var:preset|dimension|<slug>`, and have this helper never resolve it. The value then falls through the percentage regex and returns `null`, so the button silently renders at the default width — the exact NEWS-2852 symptom, in the code added to fix it.

Reproduced on WordPress 7.1-RC3 in an isolated env. With a theme declaring `dimensionSizes` at the root, the root lookup returns the presets (`theme: half => 50%`) while the block-scoped lookup returns only core's `default` origin, and `var:preset|dimension|half` renders as **`100%`** instead of `50%`.

A second, independent miss in the same loop: core kebab-cases slugs when it builds the reference (`_wp_to_kebab_case`), so a theme slug `buttonHalf` is written as `var:preset|dimension|button-half`, while line 974 compares against the raw stored `buttonHalf`. Confirmed separately at the block node — `var:preset|dimension|button-half` renders `100%` while the raw `buttonHalf` renders `50%`. The origin loop also omits `blocks`, which is where presets injected via `wp_theme_json_data_blocks` land.

Worth noting the `is_array( $presets )` guard on line 968 is not actually guarding: when the path is absent `wp_get_global_settings()` returns the *entire* settings array, so the check always passes and the loop goes on to walk `settings.custom` — a real top-level theme.json key that Newspack itself uses.

Reading the settings once and picking the node explicitly addresses all of these at the same time.

plugins/newspack-newsletters/tests/test-renderer.php
```diff
@@ -304,6 +304,184 @@ public function test_render_embed_blocks() {
...
+
+	/**
+	 * A `dimensions` width naming a preset resolves to that preset's size.
+	 *
+	 * WordPress stores a chosen preset as `var:preset|dimension|<slug>` rather
+	 * than a literal size, and resolves it at render time. Core ships 25/50/75/100
+	 * presets for buttons, so this shape is reachable on every site — not only
+	 * where a theme defines its own.
+	 */
+	public function test_button_width_dimensions_preset() {
```
chickenn00dle, line 441: **Blocker:** New preset test is red on CI — it assumes presets the test environment doesn't ship — `test_button_width_dimensions_preset()` asserts that `var:preset|dimension|50` resolves to `50%`, but nothing in the test registers that preset — it relies on core shipping it. CI installs the latest **stable** WordPress (`.github/workflows/ci.yml` reads `api.wordpress.org/core/version-check` and takes `offers[0].version`), which has no `dimensionSizes` presets for `core/button`, so the reference never resolves and the button falls back to `100%`.

This is failing on the current head (735421cc): run 31837351805, `PHPUnit / newspack-newsletters` — `Tests: 761, Assertions: 3069, Failures: 1`, with `Expected a preset width reference to resolve to the preset size` / `'50%'` vs `'100%'` at this line. I confirmed the mechanism from both sides: on the env's original 7.0.4 the lookup yields no presets, and after upgrading the same env to 7.1-RC3 core's 25/50/75/100 appear and the same input renders `50%`.

Registering the preset in the test via a `wp_theme_json_data_theme` filter (added in `set_up`, removed in `tear_down`) would make it hermetic and version-independent — and would also surface the root-node gap above, since a theme-declared preset is exactly what the helper currently misses.

One small thing for the PR description while you're in here: it reports 761 passing with 0 failures, which was true before the last two commits but no longer matches the current head. Worth refreshing so the next reader isn't misled.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,90 @@ private static function cleanup_rdb_html( $content ) {
...
+	 * Safe to call with either raw parsed-block attributes or the output of
+	 * `process_attributes()`, which touches neither width key.
+	 *
+	 * @param array $attrs A core/button block's attributes.
+	 * @return float|null Width as a percentage, or null when unset.
+	 */
+	private static function get_button_width( $attrs ) {
+		$width = null;
+
+		if ( isset( $attrs['width'] ) && is_scalar( $attrs['width'] ) && '' !== $attrs['width'] ) {
```
chickenn00dle: **Suggestion:** Branch is picked by presence, so an unusable legacy width discards a valid `dimensions` one — The `if` / `elseif` chooses the branch on `isset()` and only validates afterwards, so a block carrying an unusable legacy `width` alongside a valid `style.dimensions.width` loses the good value entirely rather than falling through to it.

Confirmed on 7.1-RC3 against the real renderer:

| attributes | renders |
| --- | --- |
| `width: 0` + `dimensions.width: '50%'` | `100%` |
| `width: 150` + `dimensions.width: '50%'` | `100%` |
| `width: '50px'` + `dimensions.width: '50%'` | `100%` |
| `width: ''` + `dimensions.width: '50%'` | `50%` (correct — the `'' !==` check catches it) |

The empty-string case already falls through correctly, which suggests the intent was there. `test_button_width_legacy_wins_over_dimensions` pins which *valid* value wins, and that ordering is worth keeping — resolving legacy first and only falling through when it yields `null` preserves it while closing this hole. Contrived on content the editor produces, but partially-migrated content is exactly what this helper exists to survive, and the zero case is reachable from the same legacy `0` that came up earlier in review.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,90 @@ private static function cleanup_rdb_html( $content ) {
...
+
+		$percentage = (float) $matches[1];
+
+		return ( $percentage > 0 && $percentage <= 100 ) ? $percentage : null;
+	}
+
+	/**
+	 * Resolve a `dimensions.width` value to its literal size.
+	 *
+	 * Mirrors the preset lookup WordPress does in `wp-includes/blocks/button.php`:
```
chickenn00dle: **Suggestion:** Docblocks point at a function that doesn't do the resolution — Both docblocks (here and at line 911) say WordPress resolves these values in `wp-includes/blocks/button.php`. `render_block_core_button()` only walks the tag processor to drop empty buttons and returns the content untouched — preset references become `var(--wp--preset--dimension--<slug>)` via the style engine and only resolve through theme.json-generated CSS custom properties. That is precisely why this helper needs to exist at all: many email clients don't support CSS variables, so the value has to be resolved server-side. Pointing at `WP_Theme_JSON::PRESETS_METADATA` / `get_settings_values_by_slug()` instead, and stating the email reason, would save the next maintainer a wasted trip and makes the helper's justification stronger.

plugins/newspack-newsletters/tests/test-renderer.php
```diff
@@ -304,6 +304,184 @@ public function test_render_embed_blocks() {
...
+
+	/**
+	 * Sibling buttons keep their own widths, whichever form each one uses.
+	 *
+	 * A newsletter can hold both forms at once: re-saving rewrites the buttons the
+	 * editor parsed, while untouched ones keep the legacy attribute. Widths are
+	 * chosen so the two cases can't coincide - if `dimensions` were ignored, the
+	 * 30% button would fall back to the 25% default share.
+	 */
+	public function test_button_width_mixed_forms() {
```
chickenn00dle, line 599: **Suggestion:** The multi-button accumulation change isn't pinned by any new test — The two callbacks at renderer lines 1296–1316 are the riskiest part of this change — `$total_defined_width` moves from `intval()` accumulation to float, and the width-less count now keys off the helper — but no new fixture distinguishes old from new behaviour.

`test_button_width_mixed_forms`'s fixture (`width: 50`, `dimensions: 30%`, `{}`) produces `['50%', '30%', '25%']` under both the old and new accumulation, because `max( 25, … )` clamps the computed share to 25 either way. Its docblock says the width-less button takes "the remaining share", which isn't what happens here — the remaining share is 20 and the clamp supplies 25.

A fixture that does discriminate: `[ dimensions 60%, {} ]` renders `['60%', '40%']` on this branch, where the pre-change accumulation would give `['60%', '50%']`. Adding that case would pin the behaviour the refactor actually changed.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,90 @@ private static function cleanup_rdb_html( $content ) {
...
+	 * The `dimensions` shape has three variants, and WordPress resolves all of them
+	 * before rendering (`wp-includes/blocks/button.php`): a literal percentage, a
+	 * `var:preset|dimension|<slug>` reference to a `dimensionSizes` preset (core
+	 * ships 25/50/75/100 for buttons, so this is reachable on every site), and an
+	 * absolute unit such as `200px`. Only percentages map onto an MJML column, so
+	 * presets are resolved first and absolute units are treated as "no width set" —
+	 * the button then takes the shared default rather than producing a nonsense
+	 * column.
+	 *
+	 * Widths outside 1–100 are also treated as unset. Nothing in the block editor
```
chickenn00dle: **Nit:** New docblock promises a row-total guarantee the multi-button path doesn't hold — The docblock justifies rejecting widths outside 1–100 because "an `mj-column` with a negative or >100% width breaks the whole row rather than just that button." The per-button value is now bounded, but the row total still isn't: `[ width: 90, {} ]` renders `['90%', '25%']` — 115% — because the `max( 25, … )` floor overrides the remaining-share calculation. Confirmed on 7.1-RC3.

This is pre-existing and I wouldn't ask you to fix it here, but the new comment now promises something the adjacent code doesn't deliver. Softening the wording (or opening a follow-up) would keep it honest. Related: `$is_multi_row` is assigned twice and never read anywhere in the file — whatever wrapping it was meant to trigger looks like it was never implemented.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -897,6 +897,90 @@ private static function cleanup_rdb_html( $content ) {
...
+	 * a `var:preset|dimension|<slug>` reference is looked up against the
+	 * `dimensionSizes` presets, searching custom, then theme, then default origins.
+	 * Any other value is returned unchanged.
+	 *
+	 * @param string $width The raw `style.dimensions.width` value.
+	 * @return string The resolved width.
+	 */
+	private static function resolve_dimension_preset( $width ) {
+		$prefix = 'var:preset|dimension|';
+		if ( 0 !== strpos( $width, $prefix ) ) {
```
chickenn00dle: **Nit:** `str_starts_with()` and type declarations on the new helpers — `0 !== strpos( $width, $prefix )` is the pre-8.0 idiom; `! str_starts_with( $width, $prefix )` says the same without the falsy-zero trap. Both helpers are new and private, so `private static function get_button_width( array $attrs ): ?float` and `private static function resolve_dimension_preset( string $width ): string` can't break a caller — the surrounding file being untyped is a reason to start rather than to skip. Also worth guarding `$preset['size']` on line 975: a malformed array value would `(string)`-cast to `"Array"` rather than being rejected.

The null and scalar handling itself is sound — `isset( $preset['slug'] )` on a string element returns false rather than warning, and the `is_scalar()` guards keep arrays out of the casts.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
```diff
@@ -1322,8 +1407,9 @@ function ( $block ) {
 
 					$column_attrs['css-class'] = 'mj-column-has-width';
 					$column_width              = $default_width;
-					if ( ! empty( $attrs['width'] ) ) {
-						$column_width                  = $attrs['width'];
+					$button_width              = self::get_button_width( $attrs );
+					if ( null !== $button_width ) {
+						$column_width                  = $button_width;
 						$default_button_attrs['width'] = '100%'; // Buttons with defined width should fill their column.
```
chickenn00dle, line 1451: **Nit:** New width assignment works because a neighbouring variable is dead — `$button_attrs` is built just above (merging `self::get_colors( $attrs )` and, for outline buttons, a `css-class`) and then never used — the emit uses `self::array_to_attributes( $default_button_attrs )`. That's the same on `origin/main`, so it isn't introduced here. But the new `$default_button_attrs['width'] = '100%'` only takes effect *because* the merged variable is dead: if someone later "fixed" the merge, the width would be silently dropped. Since this PR is editing these exact lines, either deleting `$button_attrs` or leaving a one-line comment would stop the next reader from turning a cleanup into a regression.

chickenn00dle (approved):
Release risk: HIGH — the PR's own new test is failing on CI at the current head, and the preset resolution it adds silently drops width for theme-defined presets, which is the same symptom NEWS-2852 exists to fix.

The core of this work is solid and I verified it end to end on 7.1-RC3: legacy widths, literal `dimensions` widths and core's built-in presets all render the right `mj-column`, unusable values fall back cleanly, and the multi-button share-out is correct. The grid migration is complete — no consumer anywhere in the monorepo or `repos/` still passes `start`/`end`, and both breakpoints carry the `data-` selectors. Removing the save lock also looks safe: `newspack-newsletters-post-lock` has no other consumer, the Send button still gates on `validateNewsletter()`, and a publish that slips past the UI is still stopped server-side.

Two things to settle before merge. The preset lookup is scoped to the `core/button` node only, so a theme that declares `dimensionSizes` at the root — the natural place, and where core registers the preset group — is invisible to it, and the width is silently dropped. Slugs are also compared raw, while WordPress kebab-cases them when it writes the reference. Both reproduce on 7.1-RC3. And the new preset test asserts against presets that only exist on 7.1, while CI installs the latest stable WordPress, so it fails there today.

A few smaller things inline, the most useful being that the branch is picked by presence rather than usability, so a block carrying `width: 0` plus a valid `dimensions` width renders at 100% instead of 50%.

Verified against an isolated environment on WordPress 7.1-RC3, driving the renderer directly. I could not do a visual pass (no browser available in this session), so the grid evidence is compiled-CSS and DOM-attribute level rather than a screenshot.

**Noting I am going to pre-approve assuming the blockers will be addressed and this will go through self review since the blocking feedback here is relatively small and simple to fix.**

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: Fixed in 731bd5933 — and you were right to push on it, though the reason turned out to be more interesting than "the lookup is wrong".

The lookup was a deliberate mirror of core: `render_block_core_button()` reads the same `blocks.core/button` node with no root fallback, walks the same `custom > theme > default` origins, compares slugs raw, and carries the same ineffective `is_array()` guard. I copied it on the theory that matching core was the safe choice.

Your reproduction shows why that theory is wrong for us specifically. Core can afford that lookup because when it fails it still emits `width: var(--wp--preset--dimension--<slug>)`, and theme.json's generated CSS resolves it in the browser. Email has no custom properties, so the same miss costs us the width outright. Matching core is exactly the wrong bar here — we have to resolve where core can afford not to.

So the helper now reads the settings once and indexes explicitly: the `blocks.core/button` node first, then the root `dimensions.dimensionSizes` group, across `custom > theme > blocks > default`, matching a slug either raw or kebab-cased, with a real `isset()` on the path rather than an `is_array()` on whatever came back. Confirmed your point about that guard, incidentally — the root call returns the entire settings array when the path is absent, so it was walking `settings.custom`.

Verified: a theme preset declared at the root resolves (that is what the new test registers), a kebab-cased reference resolves against a camelCase slug, core's built-in 25/50/75/100 still resolve on 7.1, and an unknown slug still falls back to the default share.

plugins/newspack-newsletters/tests/test-renderer.php
dkoo: Fixed in 731bd5933. Thanks for catching this — and for pinning it to the run, which made it unambiguous.

You were right about both the failure and the cause. I ran the suite only against the 7.1-RC3 env, where core ships those presets, so the test passed locally for the one reason it could not pass on CI.

The preset tests now register their own presets through a `wp_theme_json_data_theme` filter in a helper, removed on tear down with the theme.json cache cleared either side. That makes them independent of what the installed WordPress ships — and, per your other thread, it puts the preset at the theme layer's root, so the test also covers the node core's own lookup does not read.

Verified against both versions rather than assuming: I reinstalled the test library at `latest` (7.0.4, what CI takes) and ran the full suite, then reinstalled at 7.1-RC3 and ran it again. 763 passing, 0 failures on both.

PR description refreshed to 763 and to state both versions.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: Fixed in 731bd5933 — this was a real bug and your table reproduces exactly here.

The branch is no longer chosen by presence. Both shapes are collected in preference order, then each is validated in turn and the first usable one wins, so an unusable legacy value falls through to a valid `dimensions` one instead of discarding it. The legacy-wins ordering `test_button_width_legacy_wins_over_dimensions` pins is preserved, since legacy is still tried first.

| attributes | before | now |
| --- | --- | --- |
| `width: 0` + `dimensions 50%` | `100%` | `50%` |
| `width: 150` + `dimensions 50%` | `100%` | `50%` |
| `width: "50px"` + `dimensions 50%` | `100%` | `50%` |
| `width: 50` + `dimensions 25%` | `50%` | `50%` |

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: Docblocks rewritten in 731bd5933, though I want to flag one part of this that does not hold on 7.1 so it does not get carried forward.

`render_block_core_button()` does do the resolution there. On 7.1-RC3, `wp-includes/blocks/button.php:67-85` reads `style.dimensions.width`, detects the `var:preset|dimension|` prefix, looks the slug up against `dimensionSizes`, and only then branches on whether the resolved value ends in `%`. I suspect the read was against latest stable, where `core/button` still has the plain `width` attribute and none of that block exists yet — which is also the version difference behind the CI failure on the other thread.

The substance of your point stands and is the more useful half: the docblock explained *what* the helper mirrors without explaining *why we need it at all*. It now leads with the email reason — an unresolved reference still renders on the web through the `--wp--preset--dimension--*` custom property, and email clients have none — and then says where our lookup deliberately goes beyond core's, which is the part a future maintainer would otherwise "simplify" back down to core's version.

plugins/newspack-newsletters/tests/test-renderer.php
dkoo: Added in 731bd5933 as `test_button_width_shares_remainder()`, using your fixture.

You were right that `test_button_width_mixed_forms` cannot discriminate: with 50 + 30 + one width-less, the leftover is 20 and `max( 25, … )` supplies 25 either way, so the assertion passes on both the old and new accumulation. `[ dimensions 60%, {} ]` renders `[ "60%", "40%" ]` here, and 40 is above the floor, so a wrong total cannot be masked by the clamp.

I also corrected the `mixed_forms` docblock, which claimed the width-less button takes "the remaining share" — as you noted, it lands on the floor rather than the 20% actually remaining. It now says so and points at the new test for the arithmetic itself.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: Softened in 731bd5933.

Confirmed the case: `[ width: 90, {} ]` renders `[ "90%", "25%" ]`, 115% total, because the `max( 25, … )` floor overrides the remaining share. The docblock no longer claims a row-level guarantee — it now says only that a column outside 1-100 disturbs the rest of its row, which is what the per-button clamp actually delivers.

Leaving the floor behaviour itself and the unused `$is_multi_row` alone here, as you suggested — both predate this PR. I will open a follow-up so the row-total case is tracked rather than just noted in a thread.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: All three applied in 731bd5933: `str_starts_with()` instead of the `0 !== strpos()` idiom, `private static function get_button_width( array $attrs ): ?float` and `resolve_dimension_preset( string $width ): string`, and `$preset["size"]` now goes through an `is_scalar()` check before the cast so a malformed array value is rejected rather than stringified to `"Array"`.

Agreed on the reasoning for the type hints — new private methods can not break a caller, and the untyped surroundings are a reason to start somewhere.

plugins/newspack-newsletters/includes/class-newspack-newsletters-renderer.php
dkoo: Added a comment in 731bd5933 rather than deleting, since removing `$button_attrs` touches behaviour beyond this PR's scope (the outline `css-class` it builds is currently lost, which is its own small bug).

The note sits directly above the width assignment and says what you identified: the width lands on `$default_button_attrs` because that is what the emit serializes, `$button_attrs` beside it is dead, and anyone cleaning that up needs to move the width with it or buttons silently lose their width again. That is the trap worth flagging — it turns a tidy-up into a regression, and nothing else in the file hints at it.

dkoo:
Both blockers and all the suggestions are addressed in 731bd5933. Every thread has a reply with the reproduction; summarising the two that changed my thinking.

**The preset lookup was a mirror of core, and that was the mistake.** `render_block_core_button()` reads the same `blocks.core/button` node with no root fallback, walks the same origins, compares slugs raw, and carries the same ineffective `is_array()` guard — I copied it deliberately, on the theory that matching core was safest.

Core can afford that lookup because when it fails it still emits `width: var(--wp--preset--dimension--<slug>)`, which theme.json's CSS resolves in the browser. Email has no custom properties, so the same miss costs us the width outright. The helper now reads the settings once and indexes explicitly: the block node, then the root `dimensionSizes` group, across `custom > theme > blocks > default`, matching a slug raw or kebab-cased, guarded on the path rather than on the shape of whatever came back.

**CI was red and I had not noticed**, because I only ran the suite against the 7.1-RC3 env where core ships the presets the test assumed. The preset tests now register their own through a `wp_theme_json_data_theme` filter, so they no longer depend on the installed version — and they declare it at the theme layer's root, which is exactly the node the old lookup missed.

Verified on both versions rather than assuming: test library reinstalled at `latest` (7.0.4, what CI takes) and at 7.1-RC3, full suite run on each. 763 passing, 0 failures both times.

Also fixed: an unusable legacy width no longer discards a valid `dimensions` one (`width: 0` + `dimensions: 50%` rendered 100%, now 50%, with legacy precedence preserved); `test_button_width_shares_remainder()` pins the accumulation with the `[ 60%, {} ]` fixture that actually discriminates; and the `str_starts_with()`, type-hint, and `$preset['size']` guard points are in.

Two corrections went the other way, for the record. `render_block_core_button()` does resolve presets on 7.1 — `wp-includes/blocks/button.php:67-85` — so the docblock pointer was accurate for the version this PR targets, though the underlying point was right that it explained *what* without explaining *why*, and it now leads with the email reason. And the row-total case is real (`[ width: 90, {} ]` gives 115%) but predates this PR, so the comment was softened rather than the clamp changed; tracked as NPPM-3152 along with the unused `$is_multi_row` it belongs with.

PR description refreshed — it still said 761 passing, which was true two commits ago.

PR #877, opened 2026-08-14
fix(content-gate): keep gate preview links in preview mode under WP 7.1

dkoo:
> **Stacked on #875.** Base is `hotfix/prompt-preview-cross-origin-isolation`, so this diff shows only the follow-up. Merge #875 first, or retarget this to `main` once it lands.

### All Submissions:

#875 fixed prompt previews under WordPress 7.1. Reviewing it turned up the same cross-frame pattern in two other places, so this applies the same fix where it is needed and hardens the one place where it is not.

**Content-gate preview — the real one.** The gate's layout editor reached into the preview iframe to append preview params to links, exactly as the prompt editor did. WordPress 7.1's `Document-Isolation-Policy` severs that access. This one failed *quietly*: the call sat inside a `try`/`catch` written for genuinely cross-origin setups, so under 7.1 link rewriting simply stopped and left only a console warning. A gate preview still opened, but the first click inside it landed on the ordinary page.

The fix mirrors #875. `propagateGatePreviewParams()` runs in the previewed document, which may always rewrite its own links, and the editor no longer touches the frame. The gate's front-end script now also loads on a preview request where no gate renders — an archive, the home page, a search — because a click through one of those pages has to keep the previewed params too. Its stylesheet still loads only where a gate actually renders.

A gate preview no longer gets served from the page cache, matching what the two sibling classes already do.

**Opaque-path hrefs survive a preview.** `blob:` and friends report their *inner* origin, so they passed the origin test, failed the "is this absolute" test, and were rewritten into ordinary page URLs — destroying the link. The rewrite is now confined to `http(s)`, matching the sibling module in newspack-popups. The two files' rewriting logic is byte-identical again; only their comments and the exported names differ.

**Archive previews no longer misreport unrelated posts.** `filter_is_post_restricted()` compared `get_queried_object_id()` against post IDs. On the archive, home and search views this PR newly supports, that value is a *term* or *user* ID, so any post whose ID happened to match the queried term was force-restricted — and consumers act on that, `exclude_restricted_posts_from_feed()` among them. It now requires both `is_singular()` and the queried object being a `WP_Post`. The type check is what makes the ID comparison sound; `is_singular()` confines the forcing to views that render one post, since the blog posts index on a static-front-page site and a 404 whose `pagename` matched an unpublished page both hand back a `WP_Post` without being singular. Those two are inert today — every render path already guards on `is_singular()` — so that half is hardening that keeps them inert without relying on callers staying careful. Preview-only and privileged-only, so not a security matter; it made the preview lie about posts it was never previewing.

**Segmentation preview — hardening, and one real fix.** The Audience wizard's "view as" preview uses the same construct, but wizards run on `admin.php`, and cross-origin isolation is hooked only to the post-editor, site-editor and widgets screens. The cross-frame access is not broken today. Two things were worth changing anyway:

- It was unguarded, and `setIsOpen( true )` came after the rewrite, so any throw left the modal stuck on "Loading…" — and, because the effect that recomputes the URL never re-ran, every reopened preview reused the first session ID. The state update now runs in a `finally`.
- It matched links by prefixing the site URL, which falls back to `/`. Under that fallback a protocol-relative off-site href such as `//example.com/page` also "starts with" the prefix, so it was decorated with the segment and session IDs and carried them to a third party. It now compares origins.

**Newsletters preview — deliberately not changed.** It also reads `contentDocument`, but for a different purpose (scoping a CSS class and awaiting stylesheet load) and against Gutenberg's own editor-canvas iframe, which stays reachable under isolation. Verified in a live 7.1-RC3 editor: `contentDocument` returns a live document with cross-origin isolation active. Nothing to fix.

Also removed `frontend_url` from the gate's editor preview payload, which the deleted handler was the only consumer of.

Requires **Chrome 137 or later over HTTPS** on WordPress 7.1, signed in as a user who can edit others' pages. Cross-origin isolation is only sent to Chromium, so this will not reproduce in curl, Firefox, or Safari.

1. Enable content gating and create a gate layout, with at least two restricted posts.
2. Open the gate layout in the editor and open the browser console.
3. Click **Preview**. No `Gate preview: could not rewrite in-iframe links` warning appears.
4. Inside the preview, click a link to another restricted post.
5. The preview stays in preview mode — the gate still reflects the layout's unsaved settings.
6. Click through to an archive or the home page, then on to another restricted post. Preview mode survives the trip.

On the base branch, step 3 logs the warning and step 4 drops out of preview.

For the segmentation preview, open **Audience → Segments**, use "view as" on any segment, and confirm the preview modal still opens and links are still decorated.

### Other information:

Thirteen unit tests cover `propagateGatePreviewParams`, matching the coverage added in #875: same-origin rewriting, the top-level and no-preview and no-matching-param cases, an absent global, off-site links, in-page anchors, `mailto:`, an unparseable href, SVG anchors, href shape, fragment and unrelated-param preservation, and idempotency. Two more pin the wiring itself — that a throw in propagation cannot stop the gate from initialising, which is the property the comment there claims.

Seven PHP tests cover the seams this PR introduces. Three pin the localized payload: an admin previewing gets the param names and no gate metadata, a subscriber on the same URL gets neither, and a request without the param gets neither. Four more pin the behaviour, and each was mutation-tested to confirm it fails when the fix is reverted:

- the script still loads on an archive during a preview, and does **not** on an ordinary archive — reverting the enqueue decision to `renders_gate` alone fails this;
- an archive preview does not force-restrict a post sharing the queried term's ID — dropping the `WP_Post` guard fails this, reproducing the reported behaviour as a unit test;
- cache prevention applies only to a user who may preview.

The enqueue decision is asserted through `get_frontend_script_context()` rather than by driving `enqueue_scripts()`, because that path needs built assets for its `filemtime()` cache-busters and CI runs PHPUnit without a build — a test that called it directly would pass locally and error in CI.

Locally: the full newspack-plugin JS suite passes (627 tests, 67 suites), the full PHP suite passes (3028 tests, 8922 assertions, 10 pre-existing environmental skips), and ESLint and PHPCS are clean on every changed file.

**Not verified at runtime:** the gate preview itself. The available environment has no content gating configured, so the gate path was verified by unit tests, build inspection, and PHP class resolution rather than by clicking through a real preview. Worth one manual pass before merge, since that is the half of this PR that actually changes behaviour.

### DIFF
--- a/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
@@ -772,8 +772,11 @@ class Content_Gate {
 	 *
 	 * @return array{enqueue: bool, renders_gate: bool, is_preview: bool}
 	 */
-	public static function get_frontend_script_context() {
-		$renders_gate = self::has_gate() && is_singular() && self::is_post_restricted();
+	public static function get_frontend_script_conditions() {
+		// is_singular() first, matching enqueue_content_banner_assets(): during a
+		// preview on an archive, has_gate() would otherwise scan the gate list and have
+		// the result thrown away by the very next operand.
+		$renders_gate = is_singular() && self::has_gate() && self::is_post_restricted();
 		$is_preview   = Content_Gate\Gate_Preview::is_preview_request();
 		return [
 			'enqueue'      => $renders_gate || $is_preview,
@@ -788,7 +791,7 @@ class Content_Gate {
 	public static function enqueue_scripts() {
 		self::enqueue_content_banner_assets();
 
-		$context      = self::get_frontend_script_context();
+		$context      = self::get_frontend_script_conditions();
 		$renders_gate = $context['renders_gate'];
 		$is_preview   = $context['is_preview'];
 
@@ -824,7 +827,7 @@ class Content_Gate {
 	 *
 	 * Split out from enqueue_scripts() so the payload is assertable without
 	 * touching the filesystem for asset versions. Whether the script loads at all
-	 * is decided in get_frontend_script_context().
+	 * is decided in get_frontend_script_conditions().
 	 *
 	 * The two keys are independently optional: gate.js reads `metadata`, and
 	 * preview-links.js reads `preview_param_names`. A preview on a view where no
--- a/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
@@ -824,7 +824,7 @@ class Content_Gate {
 		// own check already requires the preview capability, so this does not ship
 		// to ordinary readers.
 		if ( $is_preview ) {
-			$script_data['preview_query_params'] = array_merge(
+			$script_data['preview_param_names'] = array_merge(
 				[ Content_Gate\Gate_Preview::PREVIEW_QUERY_PARAM ],
 				array_values( Content_Gate\Gate_Preview::PREVIEW_QUERY_KEYS )
 			);
--- a/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
@@ -760,20 +760,39 @@ class Content_Gate {
 	}
 
 	/**
+	 * Whether the gate's front-end script should load on this request, and why.
+	 *
+	 * The decision lives here, not just its payload, so it can be asserted without
+	 * building assets: the script also has to load on a gate preview where no gate
+	 * renders — archives, home, search — so the previewed params survive a click
+	 * through one of those pages. Gate them out and the preview ends silently at
+	 * the first non-singular view, which is not what the pre-7.1 editor-side
+	 * handler did: it re-ran on every navigation inside the preview frame.
+	 * Returning both inputs alongside the verdict saves the caller recomputing them.
+	 *
+	 * @return array{enqueue: bool, renders_gate: bool, is_preview: bool}
+	 */
+	public static function get_frontend_script_context() {
+		$renders_gate = self::has_gate() && is_singular() && self::is_post_restricted();
+		$is_preview   = Content_Gate\Gate_Preview::is_preview_request();
+		return [
+			'enqueue'      => $renders_gate || $is_preview,
+			'renders_gate' => $renders_gate,
+			'is_preview'   => $is_preview,
+		];
+	}
+
+	/**
 	 * Enqueue frontend scripts and styles for gated content.
 	 */
 	public static function enqueue_scripts() {
 		self::enqueue_content_banner_assets();
 
-		$renders_gate = self::has_gate() && is_singular() && self::is_post_restricted();
-		$is_preview   = Content_Gate\Gate_Preview::is_preview_request();
+		$context      = self::get_frontend_script_context();
+		$renders_gate = $context['renders_gate'];
+		$is_preview   = $context['is_preview'];
 
-		// The script also loads on a gate preview where no gate renders — archives,
-		// home, search — so the previewed params survive a click through one of
-		// those pages. Gate them out and the preview ends silently at the first
-		// non-singular view, which is not something the pre-7.1 editor-side handler
-		// did: it re-ran on every navigation inside the preview frame.
-		if ( ! $renders_gate && ! $is_preview ) {
+		if ( ! $context['enqueue'] ) {
 			return;
 		}
 
@@ -803,13 +822,17 @@ class Content_Gate {
 	/**
 	 * Data localized to the front-end gate script.
 	 *
-	 * Split out from enqueue_scripts() so the preview branch — the one seam that
-	 * decides whether the preview params reach the page — is assertable without
-	 * touching the filesystem for asset versions.
+	 * Split out from enqueue_scripts() so the payload is assertable without
+	 * touching the filesystem for asset versions. Whether the script loads at all
+	 * is decided in get_frontend_script_context().
+	 *
+	 * The two keys are independently optional: gate.js reads `metadata`, and
+	 * preview-links.js reads `preview_param_names`. A preview on a view where no
+	 * gate renders carries only the second.
 	 *
 	 * @param bool $renders_gate Whether a gate renders on this request.
 	 * @param bool $is_preview   Whether this is a gate preview request.
-	 * @return array
+	 * @return array{metadata?: array, preview_param_names?: string[]}
 	 */
 	public static function get_frontend_script_data( $renders_gate, $is_preview ) {
 		$script_data = [];
--- a/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
@@ -780,13 +780,23 @@ class Content_Gate {
 			true
 		);
 		\wp_script_add_data( $handle, 'async', true );
-		\wp_localize_script(
-			$handle,
-			'newspack_content_gate',
-			[
-				'metadata' => self::get_gate_metadata(),
-			]
-		);
+		$script_data = [
+			'metadata' => self::get_gate_metadata(),
+		];
+
+		// On a gate preview the previewed document carries its own preview params
+		// onto same-origin links, so the preview survives navigation. It needs the
+		// param list to know which of its query params those are. Gate_Preview's
+		// own check already requires the preview capability, so this does not ship
+		// to ordinary readers.
+		if ( Content_Gate\Gate_Preview::is_preview_request() ) {
+			$script_data['preview_query_params'] = array_merge(
+				[ Content_Gate\Gate_Preview::PREVIEW_QUERY_PARAM ],
+				array_values( Content_Gate\Gate_Preview::PREVIEW_QUERY_KEYS )
+			);
+		}
+
+		\wp_localize_script( $handle, 'newspack_content_gate', $script_data );
 		\wp_enqueue_style(
 			$handle,
 			Newspack::plugin_url() . '/dist/content-gate.css',
--- a/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
+++ b/plugins/newspack-plugin/includes/content-gate/class-content-gate.php
@@ -765,12 +765,18 @@ class Content_Gate {
 	public static function enqueue_scripts() {
 		self::enqueue_content_banner_assets();
 
-		if ( ! self::has_gate() ) {
+		$renders_gate = self::has_gate() && is_singular() && self::is_post_restricted();
+		$is_preview   = Content_Gate\Gate_Preview::is_preview_request();
+
+		// The script also loads on a gate preview where no gate renders — archives,
+		// home, search — so the previewed params survive a click through one of
+		// those pages. Gate them out and the preview ends silently at the first
+		// non-singular view, which is not something the pre-7.1 editor-side handler
+		// did: it re-ran on every navigation inside the preview frame.
+		if ( ! $renders_gate && ! $is_preview ) {
 			return;
 		}
-		if ( ! is_singular() || ! self::is_post_restricted() ) {
-			return;
-		}
+
 		$handle = 'newspack-content-gate';
 		\wp_enqueue_script(
 			$handle,
@@ -780,29 +786,51 @@ class Content_Gate {
 			true
 		);
 		\wp_script_add_data( $handle, 'async', true );
-		$script_data = [
-			'metadata' => self::get_gate_metadata(),
-		];
+		\wp_localize_script( $handle, 'newspack_content_gate', self::get_frontend_script_data( $renders_gate, $is_preview ) );
 
-		// On a gate preview the previewed document carries its own preview params
-		// onto same-origin links, so the preview survives navigation. It needs the
-		// param list to know which of its query params those are. Gate_Preview's
-		// own check already requires the preview capability, so this does not ship
-		// to ordinary readers.
-		if ( Content_Gate\Gate_Preview::is_preview_request() ) {
-			$script_data['preview_query_params'] = array_merge(
-				[ Content_Gate\Gate_Preview::PREVIEW_QUERY_PARAM ],
-				array_values( Content_Gate\Gate_Preview::PREVIEW_QUERY_KEYS )
-			);
+		// Only a rendering gate needs the styles.
+		if ( ! $renders_gate ) {
+			return;
 		}
-
-		\wp_localize_script( $handle, 'newspack_content_gate', $script_data );
 		\wp_enqueue_style(
 			$handle,
 			Newspack::plugin_url() . '/dist/content-gate.css',
 			[],
 			filemtime( dirname( NEWSPACK_PLUGIN_FILE ) . '/dist/content-gate.css' )
 		);
+	}
+
+	/**
+	 * Data localized to the front-end gate script.
+	 *
+	 * Split out from enqueue_scripts() so the preview branch — the one seam that
+	 * decides whether the preview params reach the page — is assertable without
+	 * touching the filesystem for asset versions.
+	 *
+	 * @param bool $renders_gate Whether a gate renders on this request.
+	 * @param bool $is_preview   Whether this is a gate preview request.
+	 * @return array
+	 */
+	public static function get_frontend_script_data( $renders_gate, $is_preview ) {
+		$script_data = [];
+
+		if ( $renders_gate ) {
+			$script_data['metadata'] = self::get_gate_metadata();
+		}
+
+		// On a gate preview the previewed document carries its own preview params
+		// onto same-origin links, so the preview survives navigation. It needs the
+		// param list to know which of its query params those are. Gate_Preview's
+		// own check already requires the preview capability, so this does not ship
+		// to ordinary readers.
+		if ( $is_preview ) {
+			$script_data['preview_query_params'] = array_merge(
+				[ Content_Gate\Gate_Preview::PREVIEW_QUERY_PARAM ],
+				array_values( Content_Gate\Gate_Preview::PREVIEW_QUERY_KEYS )
+			);
+		}
+
+		return $script_data;
 	}
 
 	/**
--- a/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
+++ b/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
@@ -8,6 +8,7 @@
  */
 
 const mockPropagate = jest.fn();
+let warnSpy;
 
 jest.mock( './preview-links', () => ( {
 	propagateGatePreviewParams: ( ...args ) => mockPropagate( ...args ),
@@ -27,11 +28,13 @@ describe( 'gate.js preview-param wiring', () => {
 		// the path a footer/async script actually takes.
 		Object.defineProperty( document, 'readyState', { value: 'interactive', configurable: true } );
 		document.body.innerHTML = '<div class="newspack-content-gate__gate"></div>';
-		jest.spyOn( console, 'warn' ).mockImplementation( () => {} );
+		// Held in a variable rather than reached for as `console.warn`, which the
+		// no-console rule rejects.
+		warnSpy = jest.spyOn( console, 'warn' ).mockImplementation( () => {} );
 	} );
 
 	afterEach( () => {
-		console.warn.mockRestore();
+		warnSpy.mockRestore();
 	} );
 
 	it( 'runs propagation on load', () => {
@@ -48,6 +51,6 @@ describe( 'gate.js preview-param wiring', () => {
 		// The throw is swallowed rather than aborting module evaluation, so the
 		// gate's own registration below it still runs.
 		expect( () => require( './gate' ) ).not.toThrow();
-		expect( console.warn ).toHaveBeenCalled();
+		expect( warnSpy ).toHaveBeenCalled();
 	} );
 } );
--- a/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
+++ b/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
@@ -24,6 +24,7 @@ describe( 'gate.js preview-param wiring', () => {
 		jest.resetModules();
 		mockPropagate.mockReset();
 		global.newspack_content_gate = { metadata: {} };
+		window.newspackRAS = [];
 		// `interactive` makes gate.js's domReady() invoke synchronously, which is
 		// the path a footer/async script actually takes.
 		Object.defineProperty( document, 'readyState', { value: 'interactive', configurable: true } );
@@ -48,9 +49,12 @@ describe( 'gate.js preview-param wiring', () => {
 			throw new Error( 'isolation' );
 		} );
 
-		// The throw is swallowed rather than aborting module evaluation, so the
-		// gate's own registration below it still runs.
 		expect( () => require( './gate' ) ).not.toThrow();
 		expect( warnSpy ).toHaveBeenCalled();
+		// Assert the gate actually initialised rather than inferring it from module
+		// evaluation completing: gate.js pushes onto newspackRAS on the gate path, so
+		// this fails if someone ever moves that registration above the propagation
+		// block — the arrangement this test exists to protect.
+		expect( window.newspackRAS.length ).toBeGreaterThan( 0 );
 	} );
 } );
--- a/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
+++ b/plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
@@ -52,9 +52,12 @@ describe( 'gate.js preview-param wiring', () => {
 		expect( () => require( './gate' ) ).not.toThrow();
 		expect( warnSpy ).toHaveBeenCalled();
 		// Assert the gate actually initialised rather than inferring it from module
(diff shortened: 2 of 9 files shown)


plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
```diff
@@ -0,0 +1,53 @@
...
+		global.newspack_content_gate = { metadata: {} };
+		// `interactive` makes gate.js's domReady() invoke synchronously, which is
+		// the path a footer/async script actually takes.
+		Object.defineProperty( document, 'readyState', { value: 'interactive', configurable: true } );
+		document.body.innerHTML = '<div class="newspack-content-gate__gate"></div>';
+		jest.spyOn( console, 'warn' ).mockImplementation( () => {} );
+	} );
+
+	afterEach( () => {
+		console.warn.mockRestore();
```
chickenn00dle: **Blocker:** CI is red: two `no-console` errors in this new test file — `JS / newspack` is failing on the current head — not on tests, on ESLint:

```
src/content-gate/gate-preview-wiring.test.js
  34:3   error  Unexpected console statement  no-console
  51:11  error  Unexpected console statement  no-console
```

Reproduced locally on this branch — same two lines, 2 errors, 0 warnings. The tests themselves are fine: the full newspack-plugin suite passes here (67 suites, 621 tests), and PHPUnit and PHPCS are green in CI too. Lint is the only thing red.

The fix is a line you already use twice in this same PR — `// eslint-disable-next-line no-console` appears in `src/content-gate/gate.js:403` and `src/wizards/audience/components/segmentation-preview/index.js:85` — so this reads as an oversight rather than a disagreement with the rule. Capturing the spy in a variable (`const warn = jest.spyOn( console, 'warn' )…`) and asserting on that would also do it without a disable.

While you are updating the description: it currently says ESLint is clean on every changed file, which no longer matches this head.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -765,12 +765,18 @@ public static function enqueue_block_editor_assets() {
 	public static function enqueue_scripts() {
```
chickenn00dle, line 791: **Suggestion:** The second commit is absent from the description and from the test plan — The "Changes proposed" section covers only the first commit's three sections. The second commit — restructuring `enqueue_scripts()` so the gate script loads on every front-end view of a preview session, adding `get_frontend_script_data()`, and adding the `template_redirect` cache-cancel hook — is described nowhere, despite being the majority of the PHP and touching the paywall's enqueue path.

The "How to test" has the same gap, and it matters more: step 4 is a singular-to-singular hop, which the *first* commit alone already fixes. Nothing asks the tester to click through an archive, category or home view, which is precisely what the second commit exists for. Since the description already flags that the gate preview is not verified at runtime, the manual pass being asked for will not exercise the new code unless the plan says to.

One factual correction while you are in there: the description says the segmentation rewrite "now runs after the modal opens". It does not — the rewrite is in the `try` and `setIsOpen( true )` is in the `finally`, so it still runs before the modal opens. The code comment states this correctly ("The `finally` keeps the original ordering"); only the description disagrees.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -765,12 +765,18 @@ public static function enqueue_block_editor_assets() {
...
 		self::enqueue_content_banner_assets();
 
-		if ( ! self::has_gate() ) {
-			return;
-		}
-		if ( ! is_singular() || ! self::is_post_restricted() ) {
+		$renders_gate = self::has_gate() && is_singular() && self::is_post_restricted();
```
chickenn00dle: **Suggestion:** Whole-session preview navigation reaches pages the overlay gate doesn't guard — Making the preview survive navigation across every view means an admin can now click from the preview into `/my-account/`, `/cart/` or `/checkout/` with `ngp_id` still set. Those are singular views, so `Gate_Preview::filter_is_post_restricted()` force-restricts the queried object and `$renders_gate` evaluates true there.

`Content_Gate::restrict_post()` is careful about this — it bails on the privacy policy, the account page, the terms page, cart and checkout — so inline gating is safe. `render_overlay_gate()` is not: its only guards are `has_gate()`, the `newspack_can_render_overlay_gate` filter, `is_singular() && is_post_restricted()`, the metering bail and the style check. During a preview every one of those is either forced true or short-circuited, and none of the page bails exist. So previewing an overlay layout and navigating to My Account renders the overlay over the account page.

The mechanism is pre-existing — the old editor-side handler also rewrote same-site links on every frame load — but this PR promotes cross-view navigation from an accident to a designed feature, which is what makes it worth a decision now rather than later. Either that is correct preview behaviour and the comment should say so, or `render_overlay_gate()` should get the same page guards `restrict_post()` already has.

plugins/newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js
```diff
@@ -49,13 +49,42 @@ const SegmentationPreview = props => {
...
+				if ( href.startsWith( '#' ) ) {
+					return;
+				}
+				let target;
+				try {
+					target = new URL( href, frameDoc.baseURI );
+				} catch ( e ) {
+					return;
+				}
+				// Compare origins rather than prefix-matching `frontendUrl`, which
```
chickenn00dle, line 78: **Suggestion:** This change is more than hardening — it closes an off-site leak worth naming in the description — The description calls this hardening only, on the grounds that wizard screens are not isolated today. The `try`/`catch`/`finally` restructuring is indeed defensive, but two of the changes here fix live bugs, and framing them as hardening risks them being dropped if this PR is ever trimmed.

The origin comparison replaces `href.indexOf( frontendUrl ) === 0`, where `frontendUrl` is `window?.newspackAudienceCampaigns?.frontend_url || '/'`. Under that `'/'` fallback the prefix test matches a protocol-relative off-site href such as `//example.com`, which would then be decorated by `decorateUrl()` and carry the segment and session IDs to a third party. Your inline comment says exactly this; it just is not reflected in the PR description.

The `finally` is the second one: as your comment notes, a throw previously left `isOpen` false, so the effect recomputing the decorated URL never re-ran and every reopened preview reused the first session ID.

No change requested in the code — I would just move both into the description so they land in review notes and changelog rather than being discovered later.

plugins/newspack-plugin/tests/unit-tests/content-gate/gate-preview.php
```diff
@@ -500,4 +500,47 @@ public function test_admin_bar_hidden_during_preview() {
...
+	/**
+	 * The preview param list reaches the front-end script only on a preview
+	 * request, which by definition means a user who can preview.
+	 *
+	 * The list itself is not secret — it is constant param names — but shipping it
+	 * is what makes the previewed document rewrite every same-origin link. A
+	 * reader who received it would carry the preview params for the rest of their
+	 * session.
+	 */
+	public function test_preview_params_localized_only_for_a_previewing_user() {
```
chickenn00dle, line 513: **Suggestion:** The new PHP tests cover the payload, not the seam the commit is named after — The two behaviours this PR actually introduces are the enqueue decision in `class-content-gate.php` (`if ( ! $renders_gate && ! $is_preview ) { return; }`) and `Gate_Preview::prevent_preview_caching()`. Neither is exercised.

All three new tests call `Content_Gate::get_frontend_script_data()` directly with hand-supplied booleans, so they cover a pure two-branch array builder and never reach `enqueue_scripts()`. They sit *downstream* of the branch — they would still pass if the early return were reverted to `! $renders_gate` alone, i.e. if the fix were undone. That is the specific regression a test here should catch.

The docblock says the split was made "so the preview branch — the one seam that decides whether the preview params reach the page — is assertable". The seam that decides that is the early return, and the split did not make it assertable: `$renders_gate` and `$is_preview` are still computed inside the untested function.

A test that sets the preview param, drives an archive query, calls `enqueue_scripts()` and asserts `wp_script_is( 'newspack-content-gate', 'enqueued' )` is true while `wp_style_is( ... )` is false would pin both new lines directly — and the `filemtime()` calls are satisfiable in the test env since `dist/` exists after a build. Worth pairing with one asserting `prevent_preview_caching()` no-ops without the capability. The existing capability, sanitization and autosave coverage is genuinely thorough, which is why this gap stands out.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
```diff
@@ -795,6 +800,39 @@ public static function enqueue_scripts() {
...
+	 *
+	 * Split out from enqueue_scripts() so the preview branch — the one seam that
+	 * decides whether the preview params reach the page — is assertable without
+	 * touching the filesystem for asset versions.
+	 *
+	 * @param bool $renders_gate Whether a gate renders on this request.
+	 * @param bool $is_preview   Whether this is a gate preview request.
+	 * @return array
+	 */
+	public static function get_frontend_script_data( $renders_gate, $is_preview ) {
```
chickenn00dle, line 840: **Nit:** No type declarations on the newly public surface — `get_frontend_script_data( $renders_gate, $is_preview )` is newly public specifically so tests can assert on it, and `prevent_preview_caching()` is a new hook callback; neither carries parameter or return types. The surrounding file is untyped throughout, so this is consistency rather than a defect — but new API is the cheap place to start, and neither has an external caller to break.

`public static function get_frontend_script_data( bool $renders_gate, bool $is_preview ): array` and `public static function prevent_preview_caching(): void`. The bare `@return array` could also become `@return array{metadata?: array, preview_query_params?: string[]}`, which documents the two mutually-optional keys that `gate.js` and `preview-links.js` each depend on.

plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
```diff
@@ -0,0 +1,53 @@
...
+	} );
+
+	it( 'runs propagation on load', () => {
+		require( './gate' );
+
+		expect( mockPropagate ).toHaveBeenCalled();
+	} );
+
+	it( 'still initialises the gate when propagation throws', () => {
```
chickenn00dle, line 47: **Nit:** Test name claims more than the assertions check — The title says "still initialises the gate", but the assertions only prove `require( './gate' )` did not throw and that `console.warn` fired. Module evaluation completing does imply the second `domReady` was reached, so the test is not wrong — it is indirect, and it would break silently if someone later moved the gate registration above the propagation block, which is exactly the arrangement it exists to protect.

`gate.js` runs `initReloadHandler()` on the gate path, which pushes onto `window.newspackRAS`; asserting `window.newspackRAS.length` (or that a listener landed on the gate element) would pin the stated property rather than a proxy for it. Worth folding into the same edit that fixes the two `no-console` errors above.

plugins/newspack-plugin/src/content-gate/preview-links.js
```diff
@@ -0,0 +1,74 @@
...
+ * lands on the ordinary page and the editor loses what they were previewing.
+ *
+ * This runs in the previewed document rather than in the editor on purpose.
+ * WordPress 7.1 serves the block editor with `Document-Isolation-Policy`, which
+ * places it in its own agent cluster and severs synchronous access to the
+ * preview iframe even though the frame is same-origin — so the editor can no
+ * longer reach in and rewrite these links from outside. A document may always
+ * rewrite its own. Same fix as NEWS-2889, applied to the gate preview.
+ *
+ * Kept deliberately in step with newspack-popups/src/view/preview-links.js,
```
chickenn00dle, line 16: **Nit:** The duplication is the right call; the stated rationale is not quite right — I agree with duplicating rather than sharing, but not with the reason as written. "A shared package would tie two independently released plugins to one version" does not hold here: `newspack-popups/package.json` already declares `newspack-colors` and `newspack-components` as `workspace:*`, and those are bundled into each plugin's own `dist/` at build time, so there is no runtime version coupling to create.

The honest justification is smaller and sufficient — about 50 lines of dependency-free DOM code, two different global names, and a `packages/` set that is React components and build tooling rather than view-layer utilities. Worth correcting because the current wording would discourage someone from sharing code in a case where sharing genuinely is fine.

The two files are byte-identical apart from the export name and comments, so the reciprocal pointer comment plus a test file on each side is the right minimum protection against drift — and both are here.

plugins/newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js
```diff
@@ -49,13 +49,42 @@ const SegmentationPreview = props => {
...
+				// falls back to '/' — under that fallback a protocol-relative
+				// off-site href like //example.com also "starts with" it, and would
+				// carry the segment and session IDs to a third party.
+				if ( target.origin !== new URL( frontendUrl, frameDoc.baseURI ).origin ) {
```
chickenn00dle: **Nit:** Reference origin rebuilt once per anchor — `new URL( frontendUrl, frameDoc.baseURI ).origin` is constructed inside the `forEach`, so it is recomputed for every anchor on a page that can easily carry a few hundred links. Both operands are loop-invariant — hoisting it above the `querySelectorAll` is a one-liner, and it also makes the origin-comparison intent you explain just above read more clearly.

Equally minor, in the same block: the `finally` is only distinguishable from putting `setIsOpen`/`onLoad` after the `catch` if the `catch` body itself throws, which it cannot here. Sequential statements would express the same guarantee with less machinery — though I take the point that `finally` states the intent more loudly, so this is a genuine toss-up.

chickenn00dle (approved):
Release risk: ELEVATED — CI is red on lint, and the archive previews this PR newly supports surface a queried-object assumption that makes the preview misreport unrelated posts.

The gate fix is the strongest of the three WP 7.1 PRs in this series. It mirrors #875 faithfully — the two `preview-links.js` files are identical apart from the export name and the global they read — and it goes further than popups did in one respect worth calling out: `previewed_layout_id()` validates that the ID actually points at a gate-layout CPT, which is exactly the hardening suggested for popups' `pid`.

Since this is a paywall surface I spent the verification budget on the gating, against an env on 7.1-RC3 with `NEWSPACK_CONTENT_GATES` enabled. It held in every case I could construct:

| request | preview state |
| --- | --- |
| anonymous, singular + valid layout id | refused |
| anonymous, archive + valid layout id | refused |
| admin, singular + valid layout id | honoured |
| admin, archive + valid layout id | honoured |
| admin, id pointing at a normal post | refused |
| admin, no `ngp_id` | refused |

Two things to settle before merge, both inline. One thing worth not losing: the segmentation-preview change is described as hardening only, but it also closes a real off-site leak and fixes a session-ID reuse bug — that deserves to be in the description rather than found later.

I could not verify the Chrome 137+ isolation behaviour or click through a real gate preview (no browser in this session), so the manual pass you flagged is still worth doing.

- `plugins/newspack-plugin/includes/content-gate/class-gate-preview.php:397` — **Suggestion:** `filter_is_post_restricted()` treats the queried object as a post, which breaks on the archives this PR now supports — `$queried_id = (int) get_queried_object_id();` is only a post ID on singular views. This PR makes archive, home and search views an explicitly supported part of a preview session, and on a category, tag or author archive that call returns a **term or user ID** instead. The guard below it then fires on an ID that was never a post. The docblock just above says "nothing is forced when there is no singular queried object" — that is the intended behaviour, but it is not what happens, because a term ID is non-zero and so passes the `$queried_id &&` test. Reproduced at runtime with a preview active on a category archive (term ID 3): | call | result | expected | | --- | --- | --- | | `is_post_restricted()` (no id, out of loop) | `true` | `false` | | `is_post_restricted( 3 )` — the *Privacy Policy page* | `true` | `false` | | `is_post_restricted( 8 )` — non-colliding post | `false` | `false` | The second row is the one that bites: an unrelated page is force-restricted purely because its ID collides with the queried term ID, and consumers act on that — `Content_Gate_Advanced_Settings::exclude_restricted_posts_from_feed()` drops such a post from feeds. It is preview-only and privileged-only, so this is not a security issue; it makes the preview lie about posts it was never previewing. A single guard ahead of the lookup makes the seam match its docblock.

Note that I am pre-approving since the blocker is relatively minor.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
dkoo: Taking the first branch of your either/or: this is intended preview behaviour, and the comment now says so. 92f7250ea.

I verified the mechanism before deciding rather than taking the report at face value — `restrict_post()` does bail on the privacy policy, account, terms, cart and checkout pages, and `render_overlay_gate()`'s only content guard is `is_singular() && is_post_restricted()`, the second of which a preview forces. So the overlay does render over My Account, exactly as you describe.

The reasoning for leaving it: the editor asked to see this layout, and a preview that silently declined to render on certain pages would be a worse lie than one that renders everywhere — the failure mode we spent the rest of this PR eliminating. It is preview-only and limited to users who may preview, so no reader is affected.

You are right that this PR is what makes the decision worth recording, since cross-view navigation goes from accident to feature here. The note lives in `Gate_Preview::filter_is_post_restricted()`, where the forcing originates, and names `render_overlay_gate()` so anyone arriving from that function's missing bails finds the answer.

Resolving on that basis — if you would rather it were the second branch (give `render_overlay_gate()` the same page guards), say so and I will reopen it.

plugins/newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js
```diff
@@ -49,15 +49,28 @@ const SegmentationPreview = props => {
...
+		setIsOpen( true );
+		onLoad( iframeEl );
+
+		try {
 			[ ...iframeEl.contentWindow.document.querySelectorAll( 'a' ) ].forEach( anchor => {
 				const href = anchor.getAttribute( 'href' );
 				if ( href.indexOf( frontendUrl ) === 0 ) {
 					anchor.setAttribute( 'href', decorateUrl( href ) );
 				}
 			} );
```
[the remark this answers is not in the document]
dkoo: Fixed. The selector is `a[href]` and the value is read with `getAttribute()`, so a non-link anchor can no longer throw and abort the rest of the pass.

Worth noting the guard was hiding a second problem in the same expression. `href.indexOf( frontendUrl ) === 0` used `frontendUrl`, which falls back to `'/'`; under that fallback a protocol-relative off-site href such as `//example.com/page` also "starts with" the prefix, so it was decorated by `decorateUrl()` and carried the segment and session IDs to a third party. The loop now compares origins instead, which fixes both at once.

plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
dkoo: Already fixed — I hit this myself before your review landed, so the current head is green on lint.

I took your second suggestion rather than the disable comment: the spy is held in a variable and asserted on, which reads better than `expect( console.warn )` and does not need an exemption from a rule that is usually right.

You were right about the description too. It has been rewritten, and it now claims lint is clean only where that is true.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
dkoo: Both gaps closed in the rewritten description.

The second commit is now described — the enqueue restructuring, `get_frontend_script_data()`, and the `template_redirect` cache-cancel — and the test plan has a step 6 that clicks through an archive or the home page and on to another restricted post, which is the path only that commit fixes. You were right that without it the manual pass would have exercised only the first commit.

Thanks for the factual correction on the segmentation ordering. That sentence was prose left over from an earlier version of the change: the rewrite is in the `try` and `setIsOpen( true )` is in the `finally`, exactly as the code comment says. Corrected.

plugins/newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js
dkoo: Agreed, and moved. The description now names both as live fixes rather than hardening: the protocol-relative off-site href that the `'/'` prefix fallback let through, with the segment and session IDs going to a third party, and the session-ID reuse that followed from `isOpen` staying false on a throw.

Your reason for asking is the right one — framing them as hardening is how they would have been dropped if this PR were ever trimmed.

plugins/newspack-plugin/tests/unit-tests/content-gate/gate-preview.php
dkoo: This was the most useful finding in the review, and you were right on the substance: all three tests sat downstream of the branch and would have passed with the early return reverted.

I did not take the suggested shape, though, because it would have failed in CI. The `PHPUnit / newspack` job checks out, installs composer and runs — there is no build step, so `dist/content-gate.js` does not exist and the `filemtime()` calls in `enqueue_scripts()` would error. A test that drove `enqueue_scripts()` directly would pass locally, where `dist/` exists after a build, and go red in CI — on the same check that blocked this PR.

So the decision itself is extracted instead, into `Content_Gate::get_frontend_script_context()`, which returns `enqueue` alongside `renders_gate` and `is_preview`. That answers the part of your comment I think matters most — "`$renders_gate` and `$is_preview` are still computed inside the untested function" — because they are now computed in a tested one, with no filesystem dependency.

Four tests, and I mutation-tested each rather than assume it was load-bearing:

| mutation | result |
| --- | --- |
| `enqueue` reverted to `renders_gate` alone | archive-during-preview test fails |
| `WP_Post` guard removed from `filter_is_post_restricted()` | unrelated-post test fails |

`prevent_preview_caching()` now returns whether it acted, so the capability gate is assertable — both its side effects are invisible under PHPUnit, since `nocache_headers()` bails on `headers_sent()` and `batcache_cancel()` is not loaded. The action discards the value and the docblock says why it exists.

plugins/newspack-plugin/includes/content-gate/class-content-gate.php
dkoo: Taking the docblock half, declining the signatures.

`@return array{metadata?: array, preview_param_names?: string[]}` is in, and it earns its place: the two keys are independently optional, and that is exactly the thing a caller gets wrong.

On the parameter and return types — you already name the reason to leave them, that the file is untyped throughout. Newspack has not adopted typed signatures anywhere in this codebase, so adding them to two methods makes this file inconsistent with itself rather than more consistent with a convention. Happy to be overruled if the team wants to start somewhere, but I would rather that be a deliberate sweep than a side effect of a hotfix.

plugins/newspack-plugin/src/content-gate/gate-preview-wiring.test.js
dkoo: Fixed, and you identified the right instrument. The test now asserts `window.newspackRAS.length`, which `initReloadHandler()` pushes onto along the gate path, so it pins the stated property instead of inferring it from module evaluation completing.

That closes the exact hole you described: moving the gate registration above the propagation block now fails the test, where before it would have kept passing.

plugins/newspack-plugin/src/content-gate/preview-links.js
dkoo: You are right, and I have rewritten it in both plugins.

I checked the claim before changing it: `newspack-popups/package.json` does declare `newspack-colors` and `newspack-components` as `workspace:*`, and those are bundled into each plugin's own `dist/` at build time, so there is no runtime version coupling for a shared module to create. My stated reason was wrong.

The replacement is the smaller, honest one you suggested — `packages/` is React components and build tooling rather than view-layer utilities, and this is roughly 50 lines of dependency-free DOM code reading a differently-named global in each plugin — with an explicit parenthetical that sharing would *not* create version coupling, so the next person is not discouraged from sharing in a case where sharing is fine.

plugins/newspack-plugin/src/wizards/audience/components/segmentation-preview/index.js
dkoo: Hoisted — `referenceOrigin` is computed once above the `querySelectorAll()`, and you are right that it also makes the origin-comparison intent read more clearly.

On `finally`, I am keeping it, while agreeing it is a toss-up. Your analysis is correct that it is indistinguishable today, since the `catch` body cannot throw. What it buys is that the guarantee is expressed where the guarantee is made: the reason those two lines must always run is that a throw previously left `isOpen` false and every reopened preview reused the first session ID. With sequential statements, a future early return added to the `catch` body would silently reintroduce that bug and nothing would flag it. Cheap insurance for a failure mode this PR exists to fix — but I would not argue if you prefer the flatter form.

dkoo:
Superseded for the WP 7.1 ship by #896, a single hotfix against `release`.

This stack is built on `main`, so it can't be retargeted to `release` — merging it there would carry 47 unrelated `main` commits with it. #896 is the same net change replayed onto `release` (two lines differ, both because `release` predates Contextual Prompts), re-verified against that tree rather than inheriting these results.

Leaving this open until #896 is green. The review history here is the record for that change, so it is worth reading rather than re-deriving — and the one open thread, on PHP type declarations, is still an open question rather than a settled decline.

dkoo:
Closing: this landed as a hotfix in #896, squash-merged to `release` as f71194149 to make the WP 7.1 ship.

#896 is the same net change replayed onto `release` and re-verified against that tree — two lines differ, both because `release` predates Contextual Prompts. Not superseded on the merits: the review here is what the change was built against, and the reasoning in these threads is the record for it.

One thing that does not travel automatically: the post-release sync of `release` into `main` will conflict in `plugins/newspack-popups/includes/class-newspack-popups-inserter.php`, because #656 added an A/B config block adjacent to this change's `$script_data` line on `main`. The resolution is to keep both additions; everything else auto-merges. That is noted in #896 too.

PR #880, opened 2026-08-14
fix(newsletters): check edit capability in the bulk visibility handler

jason10lee:
### All Submissions:

The `is_public` meta controls whether a newsletter has a public page. It is not a label: the active service provider watches it and moves the post between `private` and `publish` to match.

Who may set it is now one rule, `Newspack_Newsletters::current_user_can_set_public_status()` — `edit_post` for either direction, plus `publish_post` when making a page public, since that is the same escalation as publishing any other content.

The rule is enforced where the write happens, on `update_post_metadata`, so it covers every writer rather than each entry point separately. There are three: the list-table bulk actions, the classic quick edit, and the registered REST meta field. Writes with no user (WP-CLI, cron, migrations) pass through, since a capability belongs to a user and every HTTP route reaching this write authenticates first.

The bulk actions also consult the rule directly, so a newsletter the user cannot act on is skipped and left out of the reported count rather than attempted and blocked. They additionally skip ids that are not newsletters, and ids in the trash — where the write would change nothing now and take effect silently on restore.

Tracked in NPPM-3085. A follow-up on the notice wording is NPPM-3191.

1. From `plugins/newspack-newsletters`, run `n test-php --filter 'Bulk_Actions_Test|Public_Status_Capability_Test'`. 15 tests pass.
2. Run each class on its own as well. The CPT registration in `set_up()` is what lets the capability assertions resolve, and a full-suite run masks its absence because another test file registers the CPT first.
3. As a user who can edit but not publish a newsletter of their own, attempt to set the public-page flag through the REST meta route and confirm it does not take effect. Repeat in the opposite direction and confirm turning a page non-public still works.

### Other information:

* [x] Have you written new tests for your changes, as applicable? `Bulk_Actions_Test` covers the capability skips, the trash skip, the non-newsletter skip, both visibility directions and the counting semantics; `Public_Status_Capability_Test` covers the rule itself, the guard on the write, ownership, programmatic writes and both quick-edit directions.
* [x] Have you successfully run tests with your changes locally? Yes — 15 tests across the two classes, each run in isolation and in the full newsletters suite (809 tests, 3215 assertions, 0 failures). PHPCS clean against the root ruleset.


plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
```diff
@@ -58,21 +58,44 @@ public static function bulk_action_handler( $redirect_to, $action_name, $post_id
...
+		$updated = 0;
+		foreach ( $post_ids as $post_id ) {
+			if ( \Newspack_Newsletters::NEWSPACK_NEWSLETTERS_CPT !== get_post_type( $post_id ) ) {
+				continue;
+			}
+			if ( ! current_user_can( 'edit_post', $post_id ) ) {
+				continue;
+			}
+			update_post_meta( $post_id, 'is_public', (bool) $is_public );
+			$updated++;
```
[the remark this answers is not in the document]
jason10lee: The premise holds — `update_post_meta()` does return false on an unchanged value, confirmed directly: creating the meta returns the meta_id, re-writing the same value returns false, flipping it returns true.

Counting it anyway is deliberate, though. The notice reads "%d newsletters now have public page available", which describes the resulting state. Select five newsletters where three are already public, count only changed rows, and it reports two — understating a bulk action that did exactly what was asked. An already-public newsletter belongs in that number.

What this PR did fix about the count is the other half: it was `count( $post_ids )`, so ids the caller could not edit were reported as successes. Those are skipped now, pinned at `test-bulk-actions.php:53` and `:115`.

Two changes out of this thread. The `@return` line said "actually updated", which is what invites the row-count reading; it now states what the number counts and why an unchanged newsletter is in it. And `test_bulk_handler_counts_a_newsletter_already_in_the_requested_state` pins the semantics — it fails against the suggested version with `newsletters_public_count=0`, and none of the existing four catch that.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
```diff
@@ -58,21 +58,46 @@ public static function bulk_action_handler( $redirect_to, $action_name, $post_id
...
+	 *             already in the requested state counts, since the notice reports
+	 *             the resulting state rather than the number of rows changed.
+	 */
+	private static function set_public_status( $post_ids, $is_public ) {
+		$updated = 0;
+		foreach ( $post_ids as $post_id ) {
+			if ( \Newspack_Newsletters::NEWSPACK_NEWSLETTERS_CPT !== get_post_type( $post_id ) ) {
+				continue;
+			}
+			if ( ! current_user_can( 'edit_post', $post_id ) ) {
```
dkoo: **Suggestion:** `edit_post` still lets a Contributor make their own private newsletter public — Core maps `edit_post` on a user's own non-published post to plain `edit_posts`, so a Contributor who authored a newsletter that is now `private` passes this check. Flipping `is_public` to true then reaches the active ESP's `updated_post_meta()`, which moves the post to `publish`. That is a Contributor making a page public without `publish_posts`. No email goes out (publish↔private stays inside the controlled statuses), so the impact is exposure, not a send.

I reproduced it in the env:

**To reproduce**
1. Set an ESP (`wp option update newspack_newsletters_service_provider mailchimp`; no key needed for this path).
2. Create a `private` newsletter authored by a Contributor with `is_public=0`. Confirm `current_user_can( 'publish_post', $id )` is false for them.
3. As that Contributor, submit the **Make newsletter pages public** bulk action for that id.
4. The redirect reports `newsletters_public_count=1`, `post_status` is `publish`, and an anonymous GET of the permalink returns 200.

The quick-edit path this PR mirrors has the same shape, and this is narrower than the gap the PR closes, so not a blocker. It's small enough to fold in here if you'd rather not leave it for a follow-up.

**Consider:** Since making a newsletter public can promote its post status, consider requiring `publish_post` for the make-public direction specifically, leaving make-non-public on `edit_post`. That keeps the de-escalating action available to anyone who can edit and puts the escalating one behind the same bar as publishing any other content. The quick-edit path would need the same line.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
```diff
@@ -58,21 +58,46 @@ public static function bulk_action_handler( $redirect_to, $action_name, $post_id
...
+				continue;
+			}
+			update_post_meta( $post_id, 'is_public', (bool) $is_public );
```
dkoo, line 106: **Nit:** Trashed newsletters are written and counted — The bulk actions also appear on the Trash view. A trashed newsletter passes both guards (still the CPT; an admin passes `edit_post` via the trash branch), so the meta is written and counted while `updated_post_meta()` skips the status change for `trash`. The notice then reports a newsletter as public when nothing became public, and the meta takes effect if it is later restored. Pre-existing, mentioned only because this PR's goal includes counting what was actually updated.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
```diff
@@ -58,21 +58,46 @@ public static function bulk_action_handler( $redirect_to, $action_name, $post_id
...
+			$updated++;
```
dkoo, line 107: **Nit:** Skipped newsletters are silent — When a selection includes newsletters the user can't edit, those ids drop out and the notice reports only the applied count, so an editor can see "0 newsletters now have public page available" with no hint why. Core's `bulk_edit_posts()` tracks a `skipped` list for the same reason. Fine for a hotfix; a `newsletters_skipped_count` query arg (added to `register_removable_args()`) would make it self-explaining.

**Consider:** The count can now legitimately be zero, which renders as "0 newsletters now have public page available" with no explanation. Reporting a skipped count alongside the applied count would make the outcome self-explaining.

plugins/newspack-newsletters/tests/test-bulk-actions.php
```diff
@@ -0,0 +1,148 @@
...
+ * capability, so a user cannot flip `is_public` on a newsletter they cannot edit.
+ */
+class Bulk_Actions_Test extends WP_UnitTestCase {
+
+	/**
+	 * Ensure the newsletter CPT (and its capability mapping) is registered.
+	 */
+	public function set_up() {
+		parent::set_up();
+		\Newspack_Newsletters::register_cpt();
```
dkoo, line 23: **Nit:** `register_cpt()` in `set_up()` is redundant — The WP test suite only resets post types under `WP_RUN_CORE_TESTS`, and `Newspack_Newsletters` registers the CPT on `init` unconditionally, so it is already registered here; the other newsletters test files create `newspack_nl_cpt` posts without this call. Re-registering is harmless, but the docblock ("Ensure the newsletter CPT ... is registered") reads as if the test would break without it, which would mislead the next person copying the boilerplate.

plugins/newspack-newsletters/tests/test-bulk-actions.php
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	/**
+	 * Create a published newsletter owned by an administrator.
+	 *
+	 * @return int Newsletter post ID.
+	 */
+	private function make_admin_newsletter() {
```
dkoo: **Nit:** `make_admin_newsletter()` is used by one test while the others inline the same fixture — Only the first test calls the helper; the other four build the same admin-owned newsletter inline because they also need the admin id for `wp_set_current_user()`. A helper that takes the author id (`make_newsletter( $author )`) would serve all five. Readability only.

dkoo (approved):
Release risk: LOW. No blockers; the change closes the gap the ticket describes and I verified it at runtime.

I replayed the bulk action over HTTP as a Contributor against an admin-owned newsletter, an own draft, and a plain post: only the own draft changed, the count reported 1, and the admin newsletter's page stayed public. The two new skip tests fail on the pre-fix handler and pass here.

One thing worth deciding before merge, on the `edit_post` thread: a Contributor who authored a newsletter that is now private still passes that check and can make its page public, which promotes the post to `publish` when an ESP is active. I reproduced it in the env. It is a narrower case than the one this PR fixes and the quick-edit path has the same shape, so I've left it as a suggestion: fix it here with a `publish_post` check on the make-public direction, or track it as a follow-up.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
jason10lee: Folded in, `5ef7cbc8b3`. The make-public direction now requires `publish_post`; make-non-public de-escalates and stays on `edit_post`. The quick-edit path had the same shape and got the same line. Two tests pin both directions, and the escalating one fails against the pre-fix handler.

I've minimized your comment and copied its detail to NPPM-3085. The analysis is right and worth having; it's the repro steps I'd rather not leave on a public repo until this ships.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
jason10lee: Fixed in `286fb4d260`: trashed newsletters are now skipped, before the capability checks. Confirmed the constant behind it — `$controlled_statuses` is `[ 'publish', 'private' ]`, so the provider never moves a trashed post and the write only ever mattered on restore. `test_bulk_handler_skips_trashed_newsletters` pins it and fails against the old handler.

plugins/newspack-newsletters/includes/class-newspack-newsletters-bulk-actions.php
jason10lee: Filed as NPPM-3191 rather than done here. A query arg, a notice string and a new translatable message widen a hotfix past its fix, and this is a clarity gap rather than a wrong result. The ticket records your `newsletters_skipped_count` shape, and the open question of whether one count is enough or whether a capability skip and a trash skip deserve different messages.

plugins/newspack-newsletters/tests/test-bulk-actions.php
jason10lee: Removed rather than re-documented. The suite is green without it, so it was redundant as you thought. `set_up()` is gone entirely now, since registering the CPT was all it did.

plugins/newspack-newsletters/tests/test-bulk-actions.php
jason10lee: Replaced with `make_newsletter( $author, $status = 'publish' )` in `e59df8da56`, now used by all eight tests. It had got worse in the meantime: the two tests we added for the `publish_post` gap were inlining the fixture too.

jason10lee:
A self-review before re-requesting found the capability rule enforced on the two classic-admin writers of `is_public`, but not on the REST write of the same meta, which is the route the admin shell and the block editor actually use. Two reviewers found it independently and a refutation pass could not disprove it.

Rather than add a third check, the rule now has one definition — `Newspack_Newsletters::current_user_can_set_public_status()` — enforced on `update_post_metadata`, so it covers every writer including any added later. The bulk handler and quick-edit still consult it directly, because they need to know what they skipped in order to report a count. Writes with no user pass through: a capability belongs to a user, and WP-CLI, cron and migrations have none.

A correction on the `register_cpt()` thread above. I said the suite was green without it, so it was redundant. That was wrong, and the evidence was too: the bulk tests fail 7 of 8 when the class runs on its own, and passed only because a full-suite run had another test file register the CPT first. It is restored, with a comment saying why it is not redundant.

Tests are 15 across two classes, each run in isolation as well as in the full suite (809 tests, green). The mechanism detail is on NPPM-3085.

This is now well past what you approved — a new filter, a new public method, both entry points refactored, and a new test file — so re-requesting rather than carrying the earlier approval.

PR #884, opened 2026-08-14
fix(newsletters): stop AC trash cleanup deleting an unrelated campaign

dkoo:
### All Submissions:

Trashing a newsletter on a site connected to ActiveCampaign could delete an **unrelated campaign** in that account, automation campaigns and already-sent campaigns included. This removes the code path that did it.

`trash()` passed the post's `ac_message_id` to the v1 `campaign_delete` endpoint. Messages and campaigns are separate ActiveCampaign entities numbered from separate sequences, so the call deleted whichever campaign happened to carry that number. A publisher who creates roughly one campaign per message advances both sequences in step, which makes a collision likely rather than rare.

Three things compounded it:

* **A `message_view` lookup guarded the call**, but it can only confirm that a *message* with that ID exists. It cannot tell whether a campaign of the same number belongs to us, so it never caught the mismatch.
* **The branch called `api_v1_request()` directly** instead of `delete_campaign()`, skipping the `DELETABLE_STATUSES` guard that keeps sent and sending campaigns out of reach.
* **The campaign and message blocks were independent `if`s**, not `if`/`else`, so the wrong deletion fired on every trashed newsletter carrying an `ac_message_id`, whether or not the post had a campaign of its own.

Cleanup now covers the post's own campaign and its test campaigns, which is how the Mailchimp and Constant Contact providers already work. Switching the call to `message_delete` would not be safer: one message backs every send a post ever made, so deleting it damages the history of campaigns already out the door. A message left behind in ActiveCampaign is inert.

Introduced in the ActiveCampaign tags-support work in December 2022 and unchanged since, so it reaches every release on every channel. Deletions were silent: nothing surfaced in WordPress, and the call's return value was never checked.

New tests in `tests/test-active-campaign-trash.php` pin the behavior. They fail on `release` with a message ID reaching `campaign_delete`, which is the reported symptom reproduced.

CI covers the unit tests. To confirm the behavior against a real account without deleting anything:

1. Connect a test site to an ActiveCampaign account.
2. Add a logger for the `newspack_log` action so you can read outbound API calls. The provider fires it on every v1 request with the action name in `data.action`.
3. Create a newsletter post, pick a list and sender, and save it. Saving the email HTML gives the post an `ac_message_id`.
4. Confirm the post has `ac_message_id` and no `ac_campaign_id` in `wp_postmeta`.
5. Trash the post. No `campaign_delete` or `message_view` call is logged.
6. On `release`, the same steps log a `campaign_delete` carrying the message ID.

Do step 6 on a throwaway account. It deletes a real campaign if one shares that number.

### Other information:

* [x] Have you successfully run tests with your changes locally? Full `newspack-newsletters` suite green: 747 tests, 3049 assertions, 3 skipped.


dkoo:
Pushed `7146cef` after a self-review pass.

**What changed**

- `trash()` now clears `ac_campaign_id`, but only when the delete succeeded. A refusal means `delete_campaign()` declined a campaign past draft, so it still exists in ActiveCampaign and the post should keep pointing at it. Mailchimp and Constant Contact both do this, as does the `ac_test_campaign` loop directly above. Without it, restoring and re-sending a trashed newsletter reaches `send()` with a dead ID and fires a forced `campaign_delete` against it.
- Corrected the `@param` on `trash()`. It was typed `string` and described its argument as a campaign, which is the conflation this PR exists to untangle.
- Two assertions: the stored ID is cleared when the campaign goes, and kept when the delete is refused.

**Self-review**

No blockers, and the reasoning in the docblock checked out against the code. The tests are real regression coverage rather than tautology: 4 of 6 fail against `release` with the reported signature, all 6 pass here, and the two new assertions fail with the meta-clear removed. Full suite green at 747 tests / 3051 assertions / 3 skipped, root PHPCS clean.

Worth a reviewer's attention: no live ActiveCampaign account was exercised, since no working credentials are available. That is acceptable for a removal, which adds no new call that could be wrong, but it does mean the fix ships unobserved against the real API.

Two things this PR deliberately leaves alone, both pre-existing and worth their own ticket:

- No provider registers `before_delete_post`. Core's `wp_trash_post()` short-circuits to a permanent delete before firing its action when `EMPTY_TRASH_DAYS` is falsy, so a site with trash disabled never gets ESP cleanup at all. Shared by all three providers.
- The `pre_http_request` mock in this test file is the third hand-rolled copy across the ActiveCampaign tests. A shared trait would stop the three drifting.

The review was AI-assisted, as was the fix.

leogermani (approved):

PR #885, opened 2026-08-15
fix(network): validate peer image sideloads against blocked ranges

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? — PHPCS clean from the workspace root over every changed file, tests included.

newspack-network fetches two peer-supplied image URLs server-side — the content-distribution thumbnail and the propagated-user avatar — through `media_sideload_image()`. This validates both against private and reserved address ranges before the fetch.

The check lives in a shared `Utils\Network` helper used by both call sites:

- Rejects any URL that resolves (A or AAAA) into a private, loopback, link-local or reserved range, including the ranges PHP's filter flags miss (CGNAT `100.64.0.0/10`, benchmarking `198.18.0.0/15`). WordPress core added the link-local range to its own check in 7.0; routing through this helper keeps the plugin covered independently of the WordPress version.
- Validates each redirect hop through the same guard (via the Requests `before_redirect` bridge); legitimate images behind a redirect still load.
- Uses the version-appropriate Requests exception, so the redirect refusal is a clean skip rather than a fatal on WordPress < 6.2.

Tracked in NPPM-3144.

**Added in review.** Each of these was reproduced on a running environment before it was changed, and the reproduction is in the thread it belongs to.

- **The AAAA half of the resolver failed open.** `dns_get_record()` returns `false` when a lookup errors and `[]` when a host genuinely has no AAAA records, and both were read as "no v6 addresses" — so a host with a public A record and an internal AAAA cleared the guard whenever the AAAA query failed. It now refuses when the answer is unknown. `NXDOMAIN` returns `[]`, so an ordinary missing host is unaffected; this needs a real resolver error.
- **A non-string URL fatalled instead of refusing.** Both URLs arrive as peer JSON, and the `string` type hint raised a `TypeError` before the guard could run. The wrapper now takes a mixed value and returns a `WP_Error`.
- **The v4-in-v6 spellings of a blocked address passed.** PHP's reserved-range flags understand `::ffff:0:0/96` and nothing else, so `64:ff9b::a9fe:a9fe` and `2002:a9fe:a9fe::` — both naming 169.254.169.254 — were allowed. Blocked now; a public v6 address still passes.
- **A network on private addressing lost image sync with no way back.** This guard runs after `wp_http_validate_url()` and overrides it, so answering `http_request_host_is_external` no longer sufficed. The new `newspack_network_blocked_sideload_ip` filter is the way back in, defaulting to today's behaviour.
- **A refused avatar left no production record.** `Debugger::log()` writes nothing unless `NEWSPACK_NETWORK_DEBUG` is defined. The avatar refusal now routes through `newspack_log` the way the thumbnail refusal already did.

**Known and deliberately not fixed here:** a thumbnail sideload that fails for any reason pins the old featured image permanently, because the stored payload records the URL that failed and every later distribution matches it and bails. That predates this PR and reaches every failure mode, not just a refusal — https://linear.app/a8c/issue/NPPM-3239 carries the fix and the `wp newspack network distribute post` workaround.

Run `n test-php --filter TestNetworkUtils` from `plugins/newspack-network` — 20 tests. `TestIncomingPost` covers the sideload paths for regressions.

The full newspack-network suite reports 286 tests with one failure, `test_toggle_jetpack_photon`. It is pre-existing and unrelated: it sideloads a real external image, and the test environment cannot complete the TLS handshake to fetch it. It fails identically with every commit here reverted.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — `TestNetworkUtils` is now 20 tests. The four added in review cover the hostname branch no IP-literal test reaches, the opt-in filter in both directions, the v4-in-v6 range table, and the production hook registration. The last was mutation-checked: renaming the hook in `sideload_peer_image()` turns it red while the rest of the file stays green.
* [x] Have you successfully run tests with your changes locally? — yes; the one residual failure is accounted for above.


plugins/newspack-network/tests/unit-tests/test-network-utils.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * The redirect-hop validator throws on a target in a blocked range. This is the callback
+	 * Requests invokes before following each 302 during a peer sideload, so a redirect to the
+	 * metadata address is aborted rather than fetched.
+	 */
+	public function test_redirect_validator_refuses_internal_target() {
+		$this->expectException( \WpOrg\Requests\Exception::class );
+		Network::assert_safe_redirect( 'http://169.254.169.254/latest/meta-data/#x.jpg' );
+	}
```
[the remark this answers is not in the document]
jason10lee: Addressed in cc12aa7c3 — the tests now assert whichever Requests exception the running WordPress provides, matching the version-safe throw in `assert_safe_redirect()`.

plugins/newspack-network/includes/content-distribution/class-incoming-post.php
```diff
@@ -461,13 +462,16 @@ protected function upload_thumbnail() {
...
-		if ( ! function_exists( 'media_sideload_image' ) ) {
-			require_once ABSPATH . 'wp-admin/includes/media.php';
-			require_once ABSPATH . 'wp-admin/includes/file.php';
-			require_once ABSPATH . 'wp-admin/includes/image.php';
+		// The thumbnail URL comes from the peer's payload and is fetched server-side. Core's
+		// wp_safe_remote_get blocks RFC1918 and loopback but not the 169.254.0.0/16
+		// cloud-metadata range, so refuse any URL that resolves into a private or reserved
+		// range before making the request (SSRF).
+		if ( ! Network::is_safe_sideload_url( $thumbnail_url ) ) {
+			self::log( 'Featured image URL is not a valid external URL, skipping sideload for post ' . $this->ID );
```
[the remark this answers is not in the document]
jason10lee: Kept as `self::log()` to match the other eight call sites in this class. `log()` is an instance method, but calling it via `self::` from within another instance method (`upload_thumbnail`) preserves object context, so `$this` is forwarded and there is no PHP 8 deprecation. Switching just this line to `$this->log()` would leave the file mixing both styles.

jason10lee:
🟢 Non-blocking: a follow-up automated scan over this branch flagged the redirect guard's scope. Recording it as information, not a change request.

`sideload_peer_image()` attaches `assert_safe_redirect` with `add_action()`, so the callback is registered **process-wide** for the length of the sideload rather than against this fetch's own request. Another HTTP request running inside that window has its redirects checked by `is_safe_sideload_url()` too. A hook firing during `media_sideload_image()` would qualify, as would a cache warm.

The scan measured this shape on the same code path. An unrelated `wp_remote_get()` issued inside the window returned a `WP_Error`, where the identical call outside the window completed. The `finally` bounds it to the sideload, so nothing persists, and the effect is refusal rather than permission.

An alternative is to register the callback on that request's own hooks object, which is how core scopes its own redirect validation. That carries its own cost. If another request runs between the point where you identify your fetch and the point where Requests exposes its hooks object, the guard drops for that fetch and falls back to core's checks.

Both shapes cost something. One can reach a neighbouring request and never misses its own; the other never touches a neighbour and can miss its own. They behave the same for the case this PR covers.

plugins/newspack-network/includes/content-distribution/class-incoming-post.php
```diff
@@ -461,13 +462,16 @@ protected function upload_thumbnail() {
...
 
-		if ( ! function_exists( 'media_sideload_image' ) ) {
-			require_once ABSPATH . 'wp-admin/includes/media.php';
-			require_once ABSPATH . 'wp-admin/includes/file.php';
-			require_once ABSPATH . 'wp-admin/includes/image.php';
+		// The thumbnail URL comes from the peer's payload and is fetched server-side. Core's
+		// wp_safe_remote_get blocks RFC1918 and loopback but not the 169.254.0.0/16
+		// cloud-metadata range, so refuse any URL that resolves into a private or reserved
+		// range before making the request (SSRF).
+		if ( ! Network::is_safe_sideload_url( $thumbnail_url ) ) {
```
adekbadek, line 469: **Suggestion:** Duplicated SSRF check invites removing the wrapper's own gate — `is_safe_sideload_url()` runs here, again at the top of `sideload_peer_image()` (`class-network.php:94`), and again at `class-users.php:142`.

Reading as "callers validate", that pattern invites a later edit to drop the wrapper's check as redundant, reopening this CVE while both call sites still look correct.

Make the wrapper the sole gate and delete the pre-checks, widening its signature from `string $url` first, since both sources are unvalidated peer JSON and an array would raise a `TypeError` instead of a refusal.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+	 *
+	 * @param mixed $url Candidate URL from the network payload.
+	 *
+	 * @return bool True only for a valid external http(s) URL clear of private/reserved ranges.
+	 */
+	public static function is_safe_sideload_url( $url ): bool {
+		// wp_http_validate_url enforces the http(s) scheme, the allowed ports, and core's
+		// own RFC1918/loopback checks, and rejects any host containing ':' — so IPv6-literal
+		// URLs never reach the resolution below. The reserved-range check covers what it misses.
+		if ( ! is_string( $url ) || ! wp_http_validate_url( $url ) ) {
```
adekbadek, line 175: **Suggestion:** Give `is_blocked_sideload_ip()` a filter, as core gives `wp_http_validate_url()` one — `http_request_host_is_external` lives inside `wp_http_validate_url()` (wp-includes/http.php:635), so line 161 still honours it. What overrides it is the later `is_blocked_sideload_ip()` check on the resolved addresses: it re-refuses a private host core just approved, and honours no filter. Networks on private addressing lose avatar sync. Suggest wrapping that return in a `newspack_network_blocked_sideload_ip` filter, defaulting to today's behaviour.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+	private static function resolve_host( string $host ): array {
+		$ips = gethostbynamel( $host );
+		$ips = ( false === $ips ) ? [] : $ips;
+
+		// Peers control the hostname, so an unresolvable one would emit an E_WARNING per event;
+		// the empty/false return is handled fail-closed below, so the warning carries nothing.
+		$aaaa = @dns_get_record( $host, DNS_AAAA ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+		if ( is_array( $aaaa ) ) {
+			foreach ( $aaaa as $record ) {
+				if ( ! empty( $record['ipv6'] ) ) {
```
adekbadek: **Suggestion:** A failed AAAA lookup reads as "no v6 records" — `dns_get_record()` returns `false` on lookup failure, and the `is_array()` test treats that the same as a host with no AAAA record. So a host whose A record resolves passes the range check with its v6 half never examined, and the zone owner decides whether the AAAA query succeeds. Refuse when `false === $aaaa`, the way an unresolvable host is already refused, so "could not tell" stays fail-closed.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+	 * @param string $ip IPv4 or IPv6 address.
+	 *
+	 * @return bool True if the address is private or reserved and must be refused.
+	 */
+	public static function is_blocked_sideload_ip( string $ip ): bool {
+		if ( false === filter_var( $ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE ) ) {
+			return true;
+		}
+
+		foreach ( [ '100.64.0.0/10', '198.18.0.0/15' ] as $cidr ) {
```
adekbadek: **Suggestion:** IPv6 gets no equivalent to the explicit v4 range list — The CIDR loop closes two IPv4 gaps, but nothing does the same for IPv6, so v6 rests on `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE` alone. On the running env, `64:ff9b::a9fe:a9fe` (NAT64) and `2002:a9fe:a9fe::` (6to4) both pass, while `::ffff:169.254.169.254` is blocked: PHP covers one v4-in-v6 form and none of the others. Not a live hole — none of those prefixes is routable from an Atomic host, which has no NAT64 gateway, 6to4 relay or Teredo client — so this is defence-in-depth of the same kind as the CGNAT and benchmarking entries, not a gap to fix before this ships. Add the matching v6 prefixes, or decode any embedded v4 and re-check.

plugins/newspack-network/includes/utils/class-users.php
```diff
@@ -135,23 +135,18 @@ public static function maybe_sideload_avatar( $user_id, $user_data, $overwrite )
...
-			// (media_sideload_image -> download_url -> wp_remote_get, which does not block
-			// private/reserved hosts), so a peer could otherwise make this site request an
-			// internal address (SSRF). Reject anything that isn't a valid external URL.
-			if ( ! is_string( $avatar_url ) || ! wp_http_validate_url( $avatar_url ) ) {
+			// via media_sideload_image -> download_url -> wp_safe_remote_get. Core blocks
+			// RFC1918 and loopback there but not the 169.254.0.0/16 cloud-metadata range, so
+			// a peer could otherwise make this site request that internal address (SSRF).
+			// is_safe_sideload_url() adds the reserved-range check core omits.
+			if ( ! Network::is_safe_sideload_url( $avatar_url ) ) {
 				Debugger::log( 'Avatar URL is not a valid external URL, skipping sideload' );
```
adekbadek, line 158: **Suggestion:** `Debugger::log()` leaves no record in production — `Debugger::log()` returns immediately unless `NEWSPACK_NETWORK_DEBUG` is defined, so this new refusal of unresolvable hosts is invisible in production and a transient resolver failure skips the avatar with nothing to diagnose. Log it the way `Incoming_Post::log()` does, through the `method_exists( 'Newspack\Logger', 'newspack_log' )` fallback.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
```diff
@@ -0,0 +1,157 @@
...
+		$this->assertTrue( Network::is_blocked_sideload_ip( 'fd00::1' ), 'unique-local' );
+		$this->assertFalse( Network::is_blocked_sideload_ip( '2606:4700:4700::1111' ), 'public v6' );
+	}
+
+	/**
+	 * The peer sideload refuses an unsafe initial URL up front, returning a WP_Error rather
+	 * than making the request.
+	 */
+	public function test_sideload_peer_image_refuses_unsafe_initial_url() {
+		$result = Network::sideload_peer_image( 'http://169.254.169.254/x.jpg', 0, null, 'id' );
```
adekbadek, line 211: **Suggestion:** Untested hostname branch in `is_safe_sideload_url()` — Every test passes a literal IP, so `resolve_host()` is never entered. Invert the `empty( $ips )` fail-closed check at `class-network.php:178` and an unresolvable host clears the guard, with all 13 tests still green.

`assertFalse( Network::is_safe_sideload_url( 'http://no-such-host.invalid/x.jpg' ) )` pins that branch and stays hermetic: RFC 2606 guarantees NXDOMAIN, and a resolver-less CI container lands on the same assertion.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
```diff
@@ -0,0 +1,157 @@
...
+		$this->assertTrue( true, 'A public redirect target must not throw.' );
+	}
+
+	/**
+	 * The validator is wired to the hook core's own HTTP stack fires on each redirect. Drive
+	 * that bridge (WP_HTTP_Requests_Hooks::dispatch) the way WP_Http does and assert the
+	 * validator is reached and its exception propagates — this is what makes the redirect
+	 * refusal real rather than a hook registered on a name nothing fires.
+	 */
+	public function test_before_redirect_bridge_reaches_validator() {
```
adekbadek, line 245: **Suggestion:** Test never reads the production hook registration — The test writes the hook name itself at :144, so it never reads the registration at `class-network.php:106`. Rename that hook and all 13 tests still pass while the redirect guard does nothing in production, the exact failure the docblock says this test rules out. Asserting `has_action()` on the production registration during a real `sideload_peer_image()` call closes the gap.

adekbadek (commented):
Release risk: LOW. No blockers, the ticket's requirements are met, and the newly-refused set is narrow. Verified at runtime that the metadata payload is refused with zero outbound requests while a public URL still fetches. Two comments are worth weighing before this lands on release: the AAAA half of the resolver fails open where the rest of the guard fails closed, and a refused avatar writes nothing to any log a publisher or support engineer would read. 

- `plugins/newspack-network/includes/content-distribution/class-incoming-post.php:443` — **Suggestion:** Refused thumbnail is never retried on later distributions — Pre-existing, though this PR widens the inputs that reach it. When the guard refuses, `upload_thumbnail()` returns early but the payload still records the refused URL (:863), so every later distribution matches that stored URL and takes the early bail at :443 without retrying the sideload. A post whose new featured image was refused during a transient failure keeps showing the old image indefinitely. Keying the early bail on the current thumbnail actually having come from the incoming URL would close it, as a follow-up rather than in this hotfix.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+	 * redirect still load; only hops into a blocked range are stopped.
+	 *
+	 * @param string      $url     External image URL.
+	 * @param int         $post_id Parent post ID (0 for none).
+	 * @param string|null $desc    Image description.
+	 * @param string      $return  media_sideload_image return type ('html', 'src' or 'id').
+	 *
+	 * @return int|string|\WP_Error Whatever media_sideload_image returns; WP_Error if a hop is refused.
+	 */
+	public static function sideload_peer_image( string $url, int $post_id = 0, ?string $desc = null, string $return = 'html' ) {
```
adekbadek: **Nit:** Missing return types on the two new public methods — `sideload_peer_image()` and `assert_safe_redirect()` (:126) have no return type, while the rest of the file, including both new private helpers, does. `int|string|\WP_Error` and `void` are accurate.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+		}
+
+		if ( ! function_exists( 'media_sideload_image' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/media.php';
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+			require_once ABSPATH . 'wp-admin/includes/image.php';
+		}
+
+		// Requests fires this bridge action before following each redirect; the callback
+		// throws on an unsafe target, which Requests turns into a WP_Error for the caller.
```
adekbadek: **Nit:** WP_Error conversion happens in WP_Http, not Requests — Requests only propagates the exception; `WP_Http::request()` is what catches it and returns the WP_Error. Worth naming that layer in the comment, since the no-fatal behaviour rests on it.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+	 * Registered on the Requests before_redirect bridge during a peer sideload. Throwing
+	 * aborts the redirect the way core's own reject_unsafe_urls check does, but through
+	 * is_safe_sideload_url(), which blocks the ranges core misses on older WordPress.
+	 *
+	 * @param string $location Redirect target URL.
+	 *
+	 * @throws \WpOrg\Requests\Exception If the target is unsafe, on WordPress 6.2+.
+	 * @throws \Requests_Exception       If the target is unsafe, on WordPress < 6.2.
+	 */
+	public static function assert_safe_redirect( $location ) {
```
adekbadek: **Nit:** Stranded `$wp_current_filter` entry after a refused hop — Throwing out of a `do_action` callback skips the `array_pop( $wp_current_filter )` in `do_action_ref_array()`, so `current_filter()` returns `'requests-requests.before_redirect'` for the rest of the request, once per refused hop.

Confined in practice: every `current_filter()`/`doing_action()` check in the workspace tests a named hook, and the request is already failing. There is also no cleaner option here, since core's `$options['hooks']->register(...)` route is not reachable from a plugin.

Worth a sentence in the docblock recording that tradeoff.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+		if ( is_string( $location ) && ! self::is_safe_sideload_url( $location ) ) {
+			// The namespaced Requests exception only exists on WordPress 6.2+, but this guard
+			// protects older versions too, where the class is Requests_Exception. Throw whichever
+			// the running core provides so WP_Http catches it and returns a WP_Error, rather than
+			// fataling with class-not-found on the exact versions we target.
+			if ( class_exists( '\WpOrg\Requests\Exception' ) ) {
+				throw new \WpOrg\Requests\Exception( esc_html__( 'Refused a sideload redirect to a private or reserved address.', 'newspack-network' ), 'newspack_network_unsafe_sideload_redirect' );
```
adekbadek: **Nit:** Escaping exception messages mangles translations — These messages are never output as HTML: they land in `WP_Error::get_error_message()` and a log line, while `esc_html__()` mangles apostrophes in translations. WPCS's `EscapeOutput` sniff is what forces it, so `__()` plus a scoped `phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped` is the accurate form. Same at :135, and it matches the `WP_Error` at :95.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+			}
+			throw new \Requests_Exception( esc_html__( 'Refused a sideload redirect to a private or reserved address.', 'newspack-network' ), 'newspack_network_unsafe_sideload_redirect' );
+		}
+	}
+
+	/**
+	 * Whether a peer-supplied URL is safe to fetch server-side with media_sideload_image.
+	 *
+	 * The sideload runs through wp_safe_remote_get, which blocks RFC1918 and loopback
```
adekbadek: **Nit:** Deduplicate the core-gap rationale into one place — The "core does not block 169.254.0.0/16" rationale is repeated in this docblock, class-network.php:226, class-incoming-post.php:465, class-users.php:137, and the TestNetworkUtils class docblock. WordPress 7.1 core now rejects that range explicitly. Keep the guard: it also covers CGNAT 100.64.0.0/10, benchmarking 198.18.0.0/15, and older WordPress. State the rationale once here; have the others name the guard.

plugins/newspack-network/includes/utils/class-network.php
```diff
@@ -68,4 +68,199 @@ public static function get_networked_urls_from_list( array $urls ): array {
...
+		if ( ! is_string( $url ) || ! wp_http_validate_url( $url ) ) {
+			return false;
+		}
+
+		$host = wp_parse_url( $url, PHP_URL_HOST );
+		if ( ! is_string( $host ) || '' === $host ) {
+			return false;
+		}
+
+		// A bracketed IPv6 literal would arrive with its brackets; strip them before
```
adekbadek: **Nit:** Unreachable IPv6 bracket strip — `wp_parse_url( …, PHP_URL_HOST )` strips the brackets itself, and `wp_http_validate_url()` has already rejected any host containing `:`, as the comment at :158 says. So these two lines can never do anything, and their own comment admits it. Suggest deleting them; if they are future-proofing against core relaxing the host check, say that instead.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/**
+	 * The CGNAT and benchmarking ranges PHP's filter flags omit are blocked, and addresses
+	 * just outside them are not.
+	 */
+	public function test_cgnat_and_benchmarking_ranges_are_blocked() {
+		$this->assertTrue( Network::is_blocked_sideload_ip( '100.64.0.1' ), '100.64/10 low' );
+		$this->assertTrue( Network::is_blocked_sideload_ip( '100.100.100.200' ), 'Alibaba metadata' );
+		$this->assertTrue( Network::is_blocked_sideload_ip( '100.127.255.255' ), '100.64/10 high' );
```
adekbadek, line 84: **Nit:** Range table as a data provider — These six methods all assert `is_blocked_sideload_ip( $ip )` against an expected boolean with a label. A `@dataProvider` (used elsewhere in this plugin) would name the failing address in the PHPUnit output instead of just the family, and make the boundary pairs individually visible.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
```diff
@@ -0,0 +1,157 @@
...
+		Network::assert_safe_redirect( 'http://169.254.169.254/latest/meta-data/#x.jpg' );
+	}
+
+	/**
+	 * The redirect-hop validator lets a public target through, so legitimate images behind a
+	 * redirect still load.
+	 */
+	public function test_redirect_validator_allows_public_target() {
+		Network::assert_safe_redirect( 'http://93.184.216.34/x.jpg' );
+		$this->assertTrue( true, 'A public redirect target must not throw.' );
```
adekbadek: **Nit:** assertTrue( true ) is not an assertion — The message carries the intent; the assertion only silences PHPUnit's risky-test warning. `$this->expectNotToPerformAssertions()` says that outright, or assert the return value.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Reproduced, then fixed. `fd00--1.sslip.io` pinned to a public A record in
`/etc/hosts` with its real AAAA (`fd00::1`) in DNS: refused with the resolver
healthy, allowed with the resolver black-holed — same host, same A record. So the
v6 half does real work and silently stops running whenever the query errors.

`resolve_host()` now separates the two answers, returning null for "we never
learned" against an empty array for "no records", and the caller refuses on null.
Worth knowing for the availability question: `NXDOMAIN` returns an empty array,
not false, so an ordinary missing host never reaches the fail-closed path — it
takes a real resolver error.

The cost of failing closed is a refused sideload on a transient error, and that
lands on the retry gap in your body note, which is why it is filed rather than
left as an observation. https://linear.app/a8c/issue/NPPM-3239

plugins/newspack-network/includes/utils/class-users.php
jason10lee: Right, and only on this path: `Incoming_Post::log()` already routes through
`Newspack\Logger::newspack_log()`, so the thumbnail refusal was visible and the
avatar one was not. The avatar refusal now follows the same route, keeping the
`Debugger::log()` line for a local trace.

plugins/newspack-network/includes/content-distribution/class-incoming-post.php
jason10lee: Went a different way on this one. The `string $url` hint was the real hazard —
peer JSON reaching it as an array raised a TypeError before the guard could
refuse, which I reproduced — so the wrapper now takes a mixed value and returns a
`WP_Error` for anything that is not a safe string URL.

The pre-checks stay. Each one carries a log line the wrapper cannot write: the
post ID at the thumbnail site, and the production `newspack_log` entry at the
avatar site. Their comments now say the wrapper is the authoritative gate and
refuses again regardless, which is the misreading you were guarding against,
stated where someone would act on it.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Taken, and the reasoning changed my mind: this check runs after
`wp_http_validate_url()` and overrides it, so a network answering
`http_request_host_is_external` for its own addressing loses avatar and thumbnail
sync with no way back. `newspack_network_blocked_sideload_ip` now wraps the
verdict, defaulting to today's behaviour. Verified that a filter opting in
10.0.0.5 leaves 169.254.169.254 refused, with a test for both.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Confirmed your measurement and closed it rather than deferring — it is twelve
lines and squarely what this PR is for. `64:ff9b::/96` and `2002::/16` are
blocked now, so the NAT64 and 6to4 spellings of 169.254.169.254 refuse like the
IPv4-mapped one already did, and a public v6 address still passes. Covered by a
data provider.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Added: `int|string|\WP_Error` and `void`.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Right, and checked: `WP_Http::request()` builds it in its catch block at
`class-wp-http.php:425`. The comment names that layer now.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Reproduced — after the throw, `current_filter()` returns the hook name and
`doing_action()` stays true for the rest of the request. The docblock records the
tradeoff and why there is no cleaner option.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Both throws use `__()` now, with a `phpcs:ignore` naming the reason so the sniff
does not pull the escaping back in.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Consolidated onto `is_safe_sideload_url()`, which now says it is the one place
the reasoning lives; the call sites name the guard instead. Took the WordPress
7.1 correction too — the docblock no longer claims core misses 169.254.0.0/16
outright, and says what the guard still covers that core does not.

plugins/newspack-network/includes/utils/class-network.php
jason10lee: Half of this does not hold, and it is the half the decision rests on:
`wp_parse_url( 'https://[::1]/x', PHP_URL_HOST )` returns `[::1]` — brackets and
all. The lines are unreachable, but only because `wp_http_validate_url()` rejects
`:` hosts first, so they are exactly what would be needed if core relaxed that.
Kept, with the comment corrected to say that rather than claiming core strips
them.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
jason10lee: Added, using your `.invalid` suggestion for the hermetic property.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
jason10lee: Added a test that drives a real `sideload_peer_image()` and reads `has_action()`
from inside the request via `pre_http_request`, then asserts the hook is removed
afterwards. Mutation-checked: renaming the hook in production turns it red and
leaves the rest of the file green, which is the failure you described.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
jason10lee: Used one for the new v4-in-v6 cases. The existing CGNAT and benchmarking table I
left alone: converting it is churn on a hotfix, and the boundary pairs it asserts
are already individually labelled.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
jason10lee: `expectNotToPerformAssertions()`.

jason10lee:
Filed as https://linear.app/a8c/issue/NPPM-3239, and it turned out wider than the
note suggests. The bail keys on the stored payload's URL rather than on why the
fetch failed, so a `WP_Error` from `sideload_peer_image()` — a 404 on the peer, a
timeout — pins the old image exactly as a refusal does. That is live on `release`
today, independent of this PR.

Keying the bail on the attachment's real source turned out to be the expensive
half of it: `sideload_peer_image()` records no source URL, so that shape needs a
fallback for every attachment predating the meta or the first distribution after
deploy re-sideloads every thumbnail on every node. Not storing a URL we failed to
fetch is five lines and reaches the same place, so that is what the issue proposes
— with the trap written down, because blanking the field instead makes
`insert()` call `delete_post_thumbnail()` on the next update.

Support has `wp newspack network distribute post` in the meantime.

plugins/newspack-network/includes/content-distribution/class-incoming-post.php
adekbadek: Fair counter, and the log lines carry it: the post ID at the thumbnail site and the `newspack_log` entry at the avatar site are both things the wrapper has no way to write. The hazard I was actually guarding against was the `string $url` hint, and that is gone — `sideload_peer_image()` now takes a mixed value and refuses with a `WP_Error`, so a caller that skips its pre-check gets a refusal rather than a `TypeError`. The comments at both call sites name the wrapper as authoritative, which is what stops the "redundant, delete it" edit.

plugins/newspack-network/includes/utils/class-network.php
adekbadek: `64:ff9b::a9fe:a9fe` matches `64:ff9b::/96` and `2002:a9fe:a9fe::` matches `2002::/16`; `2606:4700:4700::1111` matches neither.

One interaction, not a change request: both prefixes are blocked wholesale, so a site behind DNS64/NAT64 gets the `64:ff9b::` form back for every v4-only peer host and loses image sync entirely. `newspack_network_blocked_sideload_ip` is the exit — a prefix test in the filter rather than a per-address one.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
adekbadek: Original point closed.

One follow-up on the new test: `$registered` starts as `null`, and `assertNotFalse( null )` passes, so the test passes whenever `pre_http_request` never fires. It doesn't fire in a container with no resolver — `is_safe_sideload_url( 'https://example.com/x.jpg' )` refuses up front because `resolve_host()` returns null, and `sideload_peer_image()` returns a `WP_Error` before registering the hook. The removal assertion also passes, since nothing was added. That's the same environment the `.invalid` test was made hermetic against, and the class docblock still says these tests resolve no DNS.

`assertNotNull( $registered )` before `assertNotFalse`, or an IP-literal URL, closes it.

plugins/newspack-network/includes/utils/class-network.php
adekbadek: You are right and I was wrong on the half the decision rests on: `parse_url( 'https://[::1]/x', PHP_URL_HOST )` returns `[::1]`, brackets and all. Keeping the strip with a comment that says it is unreachable only because `wp_http_validate_url()` rejects `:` hosts first is the right call.

adekbadek (approved):
All fifteen threads are addressed; the duplicated pre-checks and the IPv6 bracket strip came back with a counter, and both counters were right.

Verified the new v6 prefix math independently: `64:ff9b::a9fe:a9fe` and `2002:a9fe:a9fe::` match their blocks, a public v6 address matches neither, and the avatar refusal reaches `newspack_log` with the expected signature.

One nit left open in its thread: the hook-registration test passes vacuously when `pre_http_request` never fires. Worth folding in, not worth another round.

plugins/newspack-network/tests/unit-tests/test-network-utils.php
jason10lee: Reproduced before changing anything: pointed the call at a host that cannot
resolve, and the test still reported OK with 2 assertions, both satisfied on a
probe that never ran.

We took both of your suggestions, since they close different halves. The URL is
a literal IP now, which is what the class docblock claims for the whole file and
what this test had quietly stopped doing, so the interceptor fires wherever the
suite runs rather than only where DNS happens to work. The `assertNotNull` sits
ahead of the `assertNotFalse`, so if that ever stops holding the test says so
instead of passing.

Mutation-checked both directions. With the unresolvable host restored it now
fails on "must reach the HTTP layer, or this test asserts nothing"; with the
production hook renamed it still fails on your original point. TestNetworkUtils
is 20 tests, 40 assertions green, and the docblock now records why the literal
IP is load-bearing rather than incidental.

PR #887, opened 2026-08-17
feat(components): rebuild the info button on Popover (DSGNEWS-215)

thomasguillot:
### All Submissions:

The info button beside a settings label held its explanation in a tooltip attached to a plain `<span>`. Nothing could reach that explanation: no keyboard focus, no accessible name, and no way to open it by tapping. Anyone not using a mouse got nothing.

The component is rebuilt on `Popover` from `@wordpress/ui`, which the design system prescribes for info affordances because a tooltip never opens from a touch pointer. Hover still opens it on a desktop, a tap opens it on a tablet, and Escape closes it and returns focus. The explanation is carried as a description rather than as the button's name, so each trigger is named for the setting it explains instead of repeating "More information" once per section.

The appearance moves onto the design system's tokens rather than staying literal: the popup surface is `#1e1e1e` with a 4px radius, the text is 13px and left aligned, and the mark is 20px inside its 24px hit target. The `text` prop becomes `description`, and the component gains a README.

The popup also portals into the design system's shared overlay slot, a container the library reserves above the older z-index map, so it clears a core WordPress popover instead of tying with it. `Tooltip`, `Select`, `Combobox` and `Autocomplete` already opt into that slot; `Popover` is the one overlay that does not by default. Off a WordPress screen the slot is absent and the component's own z-index applies.

Part of DSGNEWS-215.

The same criteria before and after, opened on the same screen. Before on the left, after on the right.

Segment editor info affordance, before on the left and after on the right

1. Go to Audience > Campaigns > Segments and add a segment. Each criteria label shows an info icon.
2. Hover an info icon. A dark tooltip opens with that criteria's description.
3. Hover a long description, such as Donation. The text wraps and the tooltip stays 320px wide.
4. Tab to an info icon. A focus ring appears. Press Enter. The tooltip opens.
5. Press Escape. The tooltip closes and focus returns to the icon.
6. Emulate a touch device and tap an info icon. The tooltip opens.
7. Inspect an icon. Its accessible name names the setting, for example "More information about Donation".
8. Inspect an open tooltip. It renders inside a `[data-wp-compat-overlay-slot]` container.
9. Run `pnpm --filter newspack-components test`. All 24 suites pass.

### Other information:

`InfoButton` is exported from the published `newspack-components` package, and `text` is removed rather than aliased. Nothing in the org imports it apart from `SettingsSection`, which this PR migrates: `newspack-manager` and `newspack-newsletters` declare the package but never reference the component, and `newspack-manager-admin` does not depend on it at all. Landing as `feat` cuts a minor, so the removed prop arrives on a version boundary.


thomasguillot:
Pushed three changes.

**The popup portals into the design system's shared overlay slot** rather than the default container. It had been relying on `--wp-ui-popover-z-index: 1000000`, which is exactly the value of `.components-popover`, so against a core popover the winner would have been DOM order. `Tooltip`, `Select`, `Combobox`, and `Autocomplete` already default to that slot; `Popover` is the one overlay that does not. Nothing collides on the segment editor today, since it has no core popover-anchored control, but this is a published component and the next screen to adopt it may not be so lucky.

**The README no longer cites the design system for a limit it does not set.** The library text is tooltip-specific and says close to the opposite of how it was being used: it rules out hiding an important description behind a tooltip, and names a popover as the alternative when space is tight. The stricter line is ours, so it now reads that way.

**Three tests added.** `if ( ! description ) return null` was the component's only branch and nothing covered it, so removing it left the suite green. The same applied to the `typeof title === 'string'` fallback in `SettingsSection`, which only ever saw a string title. Both are now pinned in each direction.

One thing deliberately left alone: the per-component import in `SettingsSection` does not change any bundle yet. Its two siblings still import from the barrel, and the only consumer imports the barrel directly, so the package's `:root` overrides land in that bundle regardless. Making it real means those two plus roughly twenty more call sites, which is its own sweep rather than part of an info button change. `DEVELOPMENT.md` now says so, and the rule reads as the bar for new and edited modules rather than a description of the package as it stands.

24 suites and 300 tests pass, along with the TypeScript check and both linters. After the portal change I re-checked it on a local site: hover opens and holds when the pointer moves onto the popup, Enter and Escape still work with focus returning to the trigger, and the 320px cap is unchanged.

PR #889, opened 2026-08-17
fix(experimental-tools): keep template placeholders through sanitizing

adekbadek:
### All Submissions:

Experimental-tool field values are sanitized with `sanitize_textarea_field()`, which deletes a percent sign followed by two hex digits as a URL-encoded octet. A field holding a template therefore loses any placeholder whose name opens with two hex characters: `%CACHE_KEY%` is stored as `CHE_KEY%`. The owning tool cannot repair it either, because the saved-fields action carries the already-sanitized value.

Placeholders are now held out of the sanitizer and restored afterwards, for every field. Everything else is sanitized exactly as before, and a bare `%CA` stays an octet rather than becoming a placeholder.

`wp_kses_post` would be the obvious alternative, but it entity-encodes plain `<` and `>` and still eats angle-bracket type markers such as `<integer>`, so it trades one corruption for another while widening every field to post HTML.

**Rollout note.** Values already corrupted are not migrated; they have to be re-saved.

1. Register a tool with a `textarea` field through the `newspack_experimental_tools` filter.
2. Save the value `Use %CACHE_KEY% here` into that field.
3. Read the stored value. Before this change it is `Use CHE_KEY% here`; after it, unchanged.
4. Save `a <script>alert(1)</script> %CA b` into the same field.
5. Read it. The markup and the bare `%CA` octet are both gone.
6. Run `n test-php --filter Newspack_Test_Experimental_Tools`. Thirteen tests pass.

### Other information:

### DIFF
--- a/plugins/newspack-plugin/includes/experimental-tools/class-experimental-tools.php
+++ b/plugins/newspack-plugin/includes/experimental-tools/class-experimental-tools.php
@@ -199,12 +199,6 @@ class Experimental_Tools {
 	/**
 	 * Get registered tools from the filter (raw, without saved state).
 	 *
-	 * Each tool declares `slug`, `label` and optionally `fields`. A field carries
-	 * `key`, `type` and `label`; a `type` of `display` is read-only and never
-	 * stored. A field may also declare `sanitize_callback`, which
-	 * save_tool_fields() applies instead of the default sanitizer — see
-	 * sanitize_field_value() for when a tool needs that.
-	 *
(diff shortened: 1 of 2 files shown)


adekbadek:
Consumer side: Automattic/newspack-manager#632 declares `sanitize_callback` on the Editorial Assistant prompt fields, which store publisher-written templates containing `%CONTENT%` and `%CATEGORIES%`.

PR #890, opened 2026-08-17
feat(components): replace Newspack's Notice with core's (DSGNEWS-215)

thomasguillot:
### All Submissions:

Two related pieces of work. The first replaces the package's own `Notice` with core's. The second gives page-level notices somewhere to live, which is the problem the first one exposed.

Part of DSGNEWS-215.

#### 1. Core's `Notice` replaces ours

The package's own `Notice` was a bare `div`: no role, no live region, no `speak()`, so admin notices were silent to assistive technology. Half the admin already imported core's `Notice` directly, so two visual treatments coexisted, and mixing the two APIs had left real casualties: three `newspack-ads` notices rendered blank, and the Advanced Settings error message never displayed.

This started as a thin wrapper around core's component. It ends with the wrapper gone: `packages/components/src/notice/` is deleted, along with its barrel export, so `newspack-components` no longer exports `Notice`. Every `Notice` across `packages/`, `plugins/` and `themes/` now comes from `@wordpress/components`.

The wrapper's two house defaults are written out at the call sites instead. `isDismissible={ false }`, because core defaults it to `true` and would otherwise render a close button that does nothing. And `spokenMessage=""` on `info` and `warning` notices, so contextual content stays out of the live region. Only `error` and `success` announce, which is the policy the wrapper enforced.

`htmlToText` moves to `packages/components/src/utils/html-to-text.js`. The components that render raw HTML (`action-card`, `with-wizard`, `WizardError`, `handoff-message`) use it so their announcements still have tags stripped and entities decoded. `@wordpress/a11y` writes the announcement with `textContent`, so an entity left encoded is read out as its source.

Core derives its own announcement with `renderToString` during render, which corrupts the hook order of anything it walks, so a notice holding a component throws on its next render. Any notice whose children are not plain text has to pass `spokenMessage` itself. The ad units "created custom targeting keys" notice now does, with an explicit `sprintf` string, because its children are an array holding a component.

All `.newspack-notice` CSS is gone: the component's own stylesheet, plus the five stylesheets whose only rules cancelled its 32px margin. `HandoffMessage` and a new `DebugMode` component take over the two variants that were never notices. `SettingsCard`'s notification level is now a prop rather than hardcoded, overridable per instance and still defaulting to `info`.

#### 2. Page-level notices get one region

Deleting the wrapper took the last shared spacing rule with it, which made an older problem obvious: notices that belong to the page rather than to the section beneath it were spread across four mounting points at three depths, two of them straddling the page header. There was no single place meaning "page-level notice", so each new one picked a position by hand.

`GlobalNotices` widens from a query-parameter reader into the region itself: a `Stack` holding its own query notices plus a slot that everything else fills. `Wizard` and `withWizard` each render it, `HandoffMessage` sits above it so it takes none of the region's padding, and `WizardBanner` folds into the new slot. The region renders nothing at all when it has no notices and no fills, so screens with nothing to show gain no vertical space.

The inert-gating notice arrives as a fill rather than as a member of `GlobalNotices`, which keeps the component from having to know what a wizard is. It owns exactly one source, the `newspack-notice` query parameter; everything else fills the slot.

This moves `GlobalNotices` down from the very top of the page on the Dashboard, Settings and the setup wizard. That is a visible change and it is the point: one position rather than three.

#### Three defects this surfaced, fixed here

* **The debug bar rendered twice.** `Wizard` renders `DebugMode` itself, and the Dashboard and Settings views each rendered another. Two bars were in the DOM on both.
* **Nested wizards ended up with two regions.** Seven audience and newsletter views render a `Wizard` inside a `withWizard`-wrapped component, and both shells render the region, so those pages had two of them at different depths, each parsing the query string independently. `withWizard`'s error notice resolved to the outer one and sat above the header rather than in the region below it. `Wizard`'s slot provider now passes through to an ancestor when it has one, so a nested wizard shares a single registry, and the region renders only when it owns the live registration for its own name. Core's base `Slot` already guards itself the same way.
* **Query notices were re-announced on every tab switch.** `TabbedNavigation` unmounts the panel it is not showing, and core's `Notice` announces from a mount effect, so a "Settings saved" notice was read out again on each click. The parameter now comes out of the URL once the region has mounted and announced, which also stops the notice reappearing on a manual reload.

#### Other behaviour changes worth flagging

* **`GlobalNotices` no longer renders raw HTML.** Its text comes straight off the `newspack-notice` query parameter, so rendering it as markup meant a crafted admin URL could inject HTML into wp-admin. It is now plain text.
* **Trashed donation product names render as text.** That notice built an HTML string from product titles; it now uses `createInterpolateElement`, so a product title cannot carry markup into the page.
* **Two notices changed tier.** A disabled webhook endpoint and a disconnected Nextdoor account were both neutral before and are now `warning`. Both read as conditions worth attention, so this is intended.
* **`WizardError` imported `Notice` from the barrel** through a bare relative hop (`from '../..'`), so dropping the export left it undefined and the wizard's error states blank. Fixed. No test covered that path, which is why the suite stayed green.
* **A notice at the top of a modal sat flush against the first field**, so the "URL" label read as belonging to the notice. Notices in a modal have their own bottom margin again, and stacked ones sit 8px apart.

1. Run `pnpm -w run build:packages`, then build `newspack-plugin`, `newspack-ads`, `newspack-newsletters`, `newspack-popups` and `newspack-story-budget`.
2. Open Settings. There should be exactly one debug bar, not two, and the notice region sits below the page header rather than above it.
3. Load Settings with `&newspack-notice=Settings%20saved`. The notice appears once, the parameter disappears from the URL, and switching tabs does not bring it back or re-announce it.
4. Open a nested wizard, for example Audience > Access Control or Audience > Subscriptions. There should be exactly one `.newspack-global-notices` in the DOM, below the header.
5. Open Audience > Campaigns, which has no page-level notices. The region should not be in the DOM at all, and the content should sit flush under the header with no extra gap.
6. Open Settings > Connections > Webhooks, add an endpoint, and send a test request with a URL that fails. The error notice sits above the URL field with clear space between them.
7. Open the setup wizard with the setup already completed. The "Heads up!" notice renders in the region below the header.
8. Open Advertising > Display Ads with no GAM credentials. The "No credentials provided" error renders with its text visible, not as an empty bar.

### Other information:

Tests pass: newspack-components 437, newspack-plugin 1014, newspack-newsletters 328, newspack-popups 289, newspack-story-budget 71. TypeScript, ESLint and Stylelint are clean, and all five affected plugins build.

Two exported symbols change, so this needs a major. `newspack-components` no longer exports `Notice`, and it no longer exports `WizardBanner`; `GlobalNoticeFill` replaces the latter and fills the same region from anywhere inside a wizard. Every consumer in the workspace is converted here, and neither standalone checkout imports either symbol. Our commit-analyzer runs the angular preset, which gives no release at all for `refactor` and does not read the `!` shorthand in a subject line, so the subject is typed `feat` and the squashed commit needs a `BREAKING CHANGE:` footer for the major to be cut.


thomasguillot:
## Code review — round 4 · release risk: LOW

**Overview.** Fourth and converging round of review on the Notice rebuild. Every finding from rounds 1–3 has been fixed and independently re-verified: the six-plus call sites that had migrated onto the shared package inside editor entry graphs are back on core's `Notice` (barrel reachability is now identical to `main` across all 101 webpack entries of the six plugins, with editor bundles back to their `main` sizes byte-for-byte); the reflected-XSS vector in `GlobalNotices` (a query-string value rendered as raw HTML, pre-existing behaviour) is closed and the parameter handling hardened; screen-reader announcements are derived only for `error`/`success` notices, markup-stripped, with an explicit opt-out, so simultaneous and static notices no longer misfire; the wrapper carries a `NoticeProps` declaration; and the one dismissal regression (a handler wired with no close button to trigger it) is fixed, with a repo-wide sweep confirming no other site has that shape. Test suites (components 234, plugin 610, newsletters 275, popups 223, story-budget 71) and full builds are green.

**Context.** `packages/components` is the shared `newspack-components` package consumed by every product plugin's admin UI. This PR settles the Notice row of the DSGNEWS-215 audit (following Badge #868, Accordion #869, CardFeature #867, InfoButton #887): wrap core's component rather than keep the bespoke div or adopt the design system's 0.x Notice now.

**Spec alignment: fulfilled.** The ticket asks for a per-component verdict with notes on what a swap breaks; this delivers the notice/global-notices row (wrap core) with the consequences recorded in DEVELOPMENT.md and a typed prop surface. The behaviour fixes found en route (blank notices, dead props, the XSS vector) go beyond the ticket's letter and motivate the `fix:` release type.

**Findings.**

- **suggestion** — `plugins/newspack-plugin/src/wizards/audience/views/donations/configuration/index.tsx:106` — the trashed-products error notice's only child renders through `dangerouslySetInnerHTML`, so the announcement derivation finds no reachable text and the message is never spoken. It is the sole `error`/`success` notice in the migrated set that silently opts out of the announcement behaviour this PR establishes. Not a regression (the old component announced nothing either); one explicit `spokenMessage` with the plain-text sentence closes it. **Risk if not addressed:** if rendering server-supplied HTML into notices spreads as a habit, a growing set of failure messages look announced by policy but are silent in practice, which reads as a guarantee the code does not keep. **Consider:** an error or success notice that derives an empty announcement is a call site that needs an explicit `spokenMessage`, not a notice that should be silent; a development-only warning on that branch would catch this case and anything shaped like it later.
- **nit** — `plugins/newspack-network/src/story-budget/components/pull-story.js:7` — the file's only remaining change is reordering `Notice` within the `@wordpress/components` import; it stays on core's Notice correctly. Restoring the original order drops the file from the diff entirely.

consequences: verdict low
lineup: 3/3 required reviewer passes ran

**Codex (verbatim):**

> No actionable correctness regressions were identified. The Notice API migration appears consistent across known consumers, and the components TypeScript check passes.

Review provenance: 3 of 3 reviewer passes ran (deep, WP-expert, Codex), plus a consequences-synthesis pass (verdict low); blocker adjudication skipped (zero blockers after the fan-out); release-risk rubric: **LOW** — approvable.

thomasguillot:
**Final pass after the low-risk review** (0b0357234)

- The trashed-products notice on the donations configuration screen now passes an explicit `spokenMessage` with the plain-text sentence, so screen readers hear the message even though the visible content renders linked markup via `dangerouslySetInnerHTML`.
- Restored the original import order in `pull-story.js`; its remaining hunk was order-only, so the file drops out of the diff.

The suggested development-only warnings (a notice given `onRemove` without `isDismissible`, and an announced status deriving an empty announcement) are deliberately left out to keep the wrapper thin; the one concrete case of each is fixed on this branch.

thomasguillot (commented):
Release risk: LOW. No blockers. The announcement path, the injection fix and the flush-modal spacing all hold up in the browser, every editor bundle stays free of `newspack-components`, and the call-site sweep is genuinely complete on props.

Two things are worth a look before merge. The 32px vertical rhythm applies to every notice, including the ones that render outside `.newspack-page`, where it should be zero. And notices present at first paint never reach the live region, which is every message `GlobalNotices` renders. Both have threads below with repro steps.

Everything else is second-order, and it clusters in one place. The derivation helper and the docs around it promise a little more than the code delivers, in four small ways that would each cost a line or two to close.

### What I exercised

On WP 7.1, with this branch’s `newspack-plugin`, `newspack-ads` and `newspack-newsletters` built and served: the segment editor’s Referrer Sources note renders as a core info notice; “No endpoints found” renders core-styled; the webhook modal’s error notice is the container’s first child at `margin-top: 0` and its text reaches `#a11y-speak-assertive`; Advanced Settings loads two notices with both live regions empty; and `?newspack-notice=_error_` renders as escaped text, with no element created and no handler fired. JS suites pass (341 + 907) and ESLint is clean.

I did not exercise a live Nextdoor error or a real GAM failure in the Ad Unit block, since both need a failing third-party round-trip. I read those conversions instead.

One test-plan note beyond the one on the demo thread: step 4’s modal notice only appears for an endpoint that is errored or disabled, so the Add Endpoint modal opens without one on a new endpoint.

### Three smaller things with no line in the diff to hang off

- `plugins/newspack-plugin/src/wizards/setup/views/welcome/index.js:119` concatenates a raw `err.message` into the `newspack-notice` URL with no `encodeURIComponent`. Since `GlobalNotices` splits on `,`, a REST error reading “Failed to fetch, please retry” renders as an error notice plus a green success notice, and an `&` or `#` truncates it. Worth encoding at the writer, and `rawurlencode`-ing the `add_query_arg` value at `includes/class-newspack.php:348` while you are there.
- A few screens keep both `Notice` sources side by side. `newspack-newsletters/src/admin-shell/screens/settings/local-list-modal.js:172` still renders core’s while its three siblings moved, and in Story Budget `story-field.js:119` and `stories-edit.js:191` still render core’s while `stories.js` and `story.js` moved, so identical markup on one screen no longer shares the same vertical rhythm. Either finishing the sweep or noting in the README that the policy is wrapper-only would settle it.
- `packages/components/package.json` does not declare `@wordpress/a11y`, which `notice/index.test.js` imports and mocks. It resolves via the hoisted root copy today, but for a published package a `devDependencies` entry is worth having.

packages/components/src/global-notices/index.js
```diff
@@ -6,18 +6,27 @@ import { parse } from 'qs';
...
+import Notice from '../notice';
 
+/**
+ * Renders notices passed through the `newspack-notice` query parameter.
+ *
+ * The text comes off the query string, so it is rendered as plain text rather than
+ * markup. Do not reach for `__unstableHTML` here to match the other notices: that
+ * would let a crafted admin URL inject markup into wp-admin.
+ */
 const GlobalNotices = () => {
```
thomasguillot: **Severity:** suggestion — GlobalNotices messages never reach the live region

Every message this component renders is `error` or `success`, so by the new policy every one should be announced. In the browser none of them are.

The announcement path itself is fine: mounting core's `Notice` with a string `spokenMessage` after page load puts the text straight into `#a11y-speak-assertive`. What fails is announcing at first paint, and `GlobalNotices` reads its text off a query parameter after a server-side redirect, so first paint is the only time it ever renders.

There is a second issue underneath, which the fix should probably take in one go: the component splits on `,` and renders a list, and `speak()` clears both regions before writing. So even once first-paint announcing works, only the last message would survive and the first — usually the one that matters — would be lost. `README.md:75` prescribes the fix and two call sites already apply it, so `spokenMessage={ 0 === i ? text : '' }` would bring the package in line with its own rule.

**To reproduce**
1. Open `/wp-admin/admin.php?page=newspack-settings&newspack-notice=_error_Something%20went%20wrong%20saving`.
2. The red "Something went wrong saving" notice renders.
3. In the console: `document.getElementById('a11y-speak-assertive').textContent` → `""`.
4. For contrast, still on that page, mount a notice by hand:
   `const el=document.createElement('div');document.body.appendChild(el);wp.element.createRoot(el).render(wp.element.createElement(wp.components.Notice,{status:'error',spokenMessage:'PROBE'},'PROBE'))`
5. Re-read the same node → `"PROBE"`.

packages/components/src/notice/style.scss
```diff
@@ -2,112 +2,17 @@
...
-			&:hover {
-				color: inherit;
-			}
-
-			&:focus {
-				box-shadow: 0 0 0 var(--wp-admin-border-width-focus) currentcolor;
-			}
-		}
-	}
+	margin: wp-vars.$grid-unit-40 0;
```
thomasguillot: **Severity:** suggestion — Notices outside newspack-page still take the 32px vertical rhythm

`margin: $grid-unit-40 0` applies to every `.newspack-notice` on the page, but the 32px rhythm belongs to content inside `.newspack-page`. A notice rendered outside it should carry no top or bottom margin.

That case is not hypothetical: `GlobalNotices` is a sibling of `Wizard`, so it always renders outside the page container. On `admin.php?page=newspack-settings&newspack-notice=_error_…` the notice resolves to `margin-top: 32px; margin-bottom: 32px` while sitting above the page header, with `closest(.newspack-page)` returning `null`. Same shape on the dashboard (`views/dashboard/index.tsx:30`) and the setup welcome screen (`views/setup/views/welcome/index.js:296`).

Scoping the margin to `.newspack-page .newspack-notice` and leaving the bare class at `margin: 0` would make the rule structural rather than something each out-of-page call site opts out of by hand with `newspack-notice--flush`.

**To reproduce**
1. Open `/wp-admin/admin.php?page=newspack-settings&newspack-notice=_error_Margin%20check%20notice`.
2. In the console: `const n=document.querySelector(.newspack-notice); [n.closest(.newspack-page), getComputedStyle(n).marginTop, getComputedStyle(n).marginBottom]`
3. Returns `[null, 32px, 32px]`.

packages/components/src/notice/index.js
```diff
@@ -3,69 +3,95 @@
...
+ * props core reads, with two house defaults documented on `NoticeProps`:
+ * not dismissible, and announced only for `error`/`success` statuses, because
+ * contextual content does not belong in the live region and simultaneous
+ * announcements cancel each other.
+ *
+ * @param {import('./notice.d').NoticeProps} props Component props.
+ */
+const Notice = ( { children, className, isDismissible = false, spokenMessage, status = 'info', ...otherProps } ) => {
+	const announced = ANNOUNCED_STATUSES.includes( status );
+	const resolved = spokenMessage ?? ( announced ? stripTags( toText( children ) ) : '' );
```
thomasguillot: **Severity:** suggestion — A caller-supplied spokenMessage is forwarded unchecked

`resolved = spokenMessage ?? (…)` hands a caller value straight to core, which then runs `typeof message === 'string' ? message : renderToString( message )`. The README states the guarantee without qualification ("The wrapper resolves the text itself and always hands core a string"), but that only holds on the derivation branch.

Most call sites are `.js`, so `notice.d.ts` can't catch it: one passing an element re-opens exactly the hook-corruption crash the wrapper exists to prevent. All eleven `spokenMessage` values in the repo are strings today, so nothing is reachable now. `typeof spokenMessage === 'string' ? spokenMessage : …` would make the docblock true.

packages/components/src/notice/index.js
```diff
@@ -3,69 +3,95 @@
...
-			</div>
+/**
+ * Strips markup from a derived announcement. The live region filters markup on its
+ * own, but stripping here (with the same tag pattern it uses) keeps the derived
+ * value plain text with tidy whitespace rather than relying on that behaviour.
+ *
+ * @param {string} text The derived announcement.
+ * @return {string} The announcement without tags, with collapsed whitespace.
+ */
+const stripTags = text =>
```
thomasguillot: **Severity:** suggestion — The derived announcement loses angle brackets and HTML entities

`stripTags` runs on every derived announcement, not just the `__unstableHTML` ones it was written for, and the regex can't tell markup from content. A server error like `Unexpected token < in <body>` loses the bracketed span from the spoken text — reachable through `webhooks/modals/upsert.tsx:122`, which concatenates a server-supplied `disabled_error` into an `error` notice.

The `__unstableHTML` case has the mirror-image problem: entities such as `&amp;` or `&#8217;` are decoded for display by `RawHTML`, but the derived string only strips tags, and `@wordpress/a11y` assigns it via `textContent` — so the entity is announced literally.

Gating the strip on `__unstableHTML` handles the first, and decoding entities on that same branch handles the second.

packages/components/src/notice/index.js
```diff
@@ -3,69 +3,95 @@
...
+ * contextual content does not belong in the live region and simultaneous
+ * announcements cancel each other.
+ *
+ * @param {import('./notice.d').NoticeProps} props Component props.
+ */
+const Notice = ( { children, className, isDismissible = false, spokenMessage, status = 'info', ...otherProps } ) => {
+	const announced = ANNOUNCED_STATUSES.includes( status );
+	const resolved = spokenMessage ?? ( announced ? stripTags( toText( children ) ) : '' );
+
+	if ( 'production' !== process.env.NODE_ENV && announced && spokenMessage === undefined && ! resolved && children ) {
```
thomasguillot: **Severity:** suggestion — The silent-announcement warning only fires when nothing is readable

The condition requires `! resolved`, so it catches the all-or-nothing case only. `<Notice status="error">{ 'Failed: ' }<SomeComponent /></Notice>` derives `'Failed:'`, which is truthy, so no warning fires while the component's own text never reaches the live region.

Both `notice.d.ts:28` and `README.md:69` describe the warning as catching children "the derivation cannot read", which reads as covering partial loss too. No current call site hits it. Worth either tightening the guard or narrowing the doc claim.

packages/components/src/global-notices/index.js
```diff
@@ -6,18 +6,27 @@ import { parse } from 'qs';
...
-	const notice = parse( window.location.search )[ 'newspack-notice' ];
-	if ( ! notice ) {
+	const notice = parse( window.location.search, { ignoreQueryPrefix: true } )[ 'newspack-notice' ];
+	if ( typeof notice !== 'string' || ! notice ) {
 		return null;
 	}
 	return notice.split( ',' ).map( ( text, i ) => {
```
thomasguillot: **Severity:** suggestion — The security-relevant change is the one without a test

`notice/index.test.js` is thorough at 179 lines, but `GlobalNotices` has no test file at all — and it is where the raw-HTML removal, the new `typeof notice !== 'string'` guard and the `ignoreQueryPrefix` fix all live.

Two short cases would pin the behaviour that matters most: a `?newspack-notice=_error_` render asserting the text is escaped, and `?newspack-notice[]=a` asserting `null`. Without them, a future "let's use `__unstableHTML` here for consistency" edit reopens the injection path silently.

plugins/newspack-plugin/src/wizards/audience/views/donations/configuration/index.tsx
```diff
@@ -104,22 +103,33 @@ export const DonationAmounts = ( { hideHeader = false }: { hideHeader?: boolean
...
+						// Translators: %s is a comma-separated list of trashed product names.
+						__(
+							'One or more donation products is in trash. Please restore the product(s) to continue using donation features: %s',
+							'newspack-plugin'
+						),
+						trashed.join( ', ' )
+					) }
+				>
+					<span>
+						{ createInterpolateElement(
```
thomasguillot: **Severity:** suggestion — createInterpolateElement on a translated string, unguarded

The new call passes a translated string carrying `<a>` straight to `createInterpolateElement`. `@wordpress/element` 8.x has no try/catch around the tokeniser: its `closer` branch pops the stack and then reads `stackTop.prevOffset`, so a translation that keeps `</a>` but loses the opener throws a `TypeError` mid-render. An unbalanced opener degrades gracefully; a stray closer does not.

This repo already diagnosed this and built `interpolateOrPlainText` for it at `packages/components/src/wizard/index.js:70`, whose docblock spells out the same failure. The Donations tab has no error boundary above it either. It needs a broken `.po` rather than anything an attacker or normal operation can reach, so it is not urgent — but exporting that helper and using it here costs one import.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
```diff
@@ -343,7 +343,8 @@ export default function SubscriberList() {
 	// offer a retry.
 	if ( error ) {
 		return (
-			<Notice isError noticeText={ sprintf( __( 'Could not load subscribers: %s', 'newspack-plugin' ), error ) }>
+			<Notice status="error">
```
thomasguillot: **Severity:** suggestion — Action button labels fold into the derived announcement

`toText` walks the whole children array and reads `props.children` regardless of element type, so these screens announce "Could not load subscribers: … Retry" rather than just the message. `GroupList.jsx` and `PersonProfile.jsx` have the same shape.

The cause is that the button sits in `children` rather than in core's `actions` prop. Moving it to `actions`, or passing the message as an explicit `spokenMessage`, keeps the live region to the message.

plugins/newspack-plugin/src/wizards/componentsDemo/index.js
```diff
@@ -486,14 +487,6 @@ class ComponentsDemo extends Component {
...
 						</Card>
-						<Card>
-							<h2>{ __( 'Notice', 'newspack-plugin' ) }</h2>
-							<Notice noticeText={ __( 'This is an info notice.', 'newspack-plugin' ) } />
-							<Notice noticeText={ __( 'This is an error notice.', 'newspack-plugin' ) } isError />
-							<Notice noticeText={ __( 'This is a help notice.', 'newspack-plugin' ) } isHelp />
-							<Notice noticeText={ __( 'This is a success notice.', 'newspack-plugin' ) } isSuccess />
-							<Notice noticeText={ __( 'This is a warning notice.', 'newspack-plugin' ) } isWarning />
-						</Card>
 						<Card>
```
thomasguillot, line 578: **Severity:** nit — The demo loses its Notice showcase, and the test plan still points at it

The card that rendered the statuses side by side was removed rather than converted, leaving only an incidental `<Notice>` in the autocomplete demo at `:806`. On the running site the Components Demo page has no "Notice" heading and shows three notices with no `success` and none dismissible.

That makes step 8 of the PR's test plan ("the Notice card shows the four statuses and a dismissible example") impossible to follow. Since this is the PR that changes what every status looks like, the demo screen is where that is easiest to check — converting the card to `status="error|warning|success|info"` costs about five lines, and step 8 then does what it says.

packages/components/src/notice/README.md
```diff
@@ -0,0 +1,111 @@
...
+anything it renders, so a notice holding a component that uses hooks throws on
+its next render, not on the first. The wrapper resolves the text itself and
+always hands core a string, which is the branch that renders nothing.
+
+**Anyone importing core's `Notice` directly has to handle this.** Pass a string
+`spokenMessage` whenever the children are not a plain string, or pass `''` to
+stay out of the live region. There is a regression test for the crash in
+`index.test.js`.
+
+The flattener reads strings and host elements. Text held inside a component, or
```
thomasguillot: **Severity:** nit — The docs say raw HTML is unreachable; the test asserts the opposite

`README.md:69`, `notice.d.ts:29` and the warning string at `index.js:80` all state that text "rendered as raw HTML" cannot be reached and leaves the notice silent. But `__unstableHTML` string children *are* reached — `toText` returns the string and `stripTags` cleans it — and `index.test.js:86` asserts exactly that. Only `dangerouslySetInnerHTML` is genuinely unreachable.

As written, someone would add a redundant `spokenMessage` to all five `__unstableHTML` notices.

packages/components/src/notice/notice.d.ts
```diff
@@ -0,0 +1,45 @@
+export interface NoticeAction {
+	className?: string;
+	disabled?: boolean;
```
thomasguillot: **Severity:** nit — NoticeAction declares two fields core ignores and omits one it reads

Core destructures only `className`, `label`, `isPrimary`, `variant`, `noDefaultClasses`, `onClick` and `url` from each action. `disabled` and `openInNewTab` are never read — the button gets neither a `disabled` attribute nor a `target` — while `isPrimary`, which core does read, is missing from the interface.

Publishing a type that invites a call site to pass a prop that silently does nothing is the same class of bug this PR is fixing.

packages/components/src/notice/notice.d.ts
```diff
@@ -0,0 +1,45 @@
...
+	 * component, or rendered as raw HTML, cannot be reached and leaves the notice
+	 * silent, so pass this explicitly for those. Announcements also replace one
+	 * another, so where several notices mount together, silence all but one.
+	 */
+	spokenMessage?: string;
+	status?: 'error' | 'warning' | 'success' | 'info';
+	politeness?: 'polite' | 'assertive';
+	/** Without this, `isDismissible` renders a close button that does nothing. */
+	onRemove?: () => void;
+	/** Deprecated core alias for `onRemove`. */
```
thomasguillot: **Severity:** nit — onDismiss is documented as a deprecated alias, which contradicts core

Core's own source comments the opposite: "onDismiss is a callback executed when the notice is dismissed. It is distinct from onRemove, which _looks_ like a callback but is actually the function to call to remove the notice from the UI", and `onDismissNotice()` calls both.

The rest of this `.d.ts` is unusually good at recording exact core behaviour, so this one line stands out — it would lead someone to think passing `onDismiss` alone is enough.

packages/components/src/settings/SettingsCard.js
```diff
@@ -8,11 +8,11 @@ import classnames from 'classnames';
  */
 import { Grid, ActionCard } from '../';
 
-const SettingsCard = ( { children, className, columns = 3, gutter = 32, noBorder, rowGap, ...props } ) => {
+const SettingsCard = ( { children, className, columns = 3, gutter = 32, noBorder, notificationLevel = 'warning', rowGap, ...props } ) => {
```
thomasguillot: **Severity:** nit — The new warning default has no call site to exercise it

Flagged in the description as intentional, so this is only to put the trade-off on the record. Making `notificationLevel` a forwarded prop is clearly right; it is the choice of `'warning'` for the default that is worth a second look.

The only `SettingsCard` in the repo that passes `notification` is `single-segment.js:271`, which this PR pins to `info` — so the new default ships unexercised, and the next settings note written without an explicit level comes out amber rather than grey. If the intent was "keep the old shape unless overridden", `'info'` is the default that says so.

packages/components/src/global-notices/index.js
thomasguillot: Half of this reproduces, half does not, so I fixed the half that does.

The multi-notice loss is real and now fixed in 4890eede2. With `?newspack-notice=_error_First message that matters,Second message` on this branch before the fix, `#a11y-speak-assertive` was empty and `#a11y-speak-polite` held `Second message`: the error was written, then cleared by the success notice mounting after it. So the message that mattered was the one lost, exactly as `README.md:75` predicts. `spokenMessage={ 0 === i ? message : '' }` now applies, and the same URL leaves `First message that matters` in the assertive region with polite empty.

The first-paint half does not reproduce. I ran the repro five times against freshly loaded pages and got the message in `#a11y-speak-assertive` five times out of five, and a 20ms poll from an init script shows the write landing at ~1.3s, a little after `window.load`. Reading `textContent` immediately after the page settles is too early: the React app has not mounted yet at that point, and a backgrounded tab throttles the timers further. The regions are created by `@wordpress/a11y`'s `domReady( setup )`, which runs before our bundle mounts, so `speak()` finds them.

packages/components/src/notice/style.scss
thomasguillot: Fixed in 4890eede2, though not by scoping the selector: `GlobalNotices` now renders with `newspack-notice--flush`, so it resolves to `margin: 0` above the page header. All three out-of-page sites you name are `<GlobalNotices />` render sites (`newspack/views/settings/index.tsx:29`, `newspack/views/dashboard/index.tsx:30`, `setup/views/welcome/index.js:299`), so the one component covers them.

I tried `.newspack-page .newspack-notice` first and backed it out, because `.newspack-page` is not the only container that should carry the rhythm:

- Modals portal to `<body>`, so `closest( '.newspack-page' )` on a modal notice returns `null`. Under the scoped rule the webhook modal's notice kept its flush top and lost its 32px bottom, jamming it against the form.
- `newspack-newsletters`' admin shell only renders inside `Page` in bundled mode. Standalone it renders `.newspack-newsletters-admin` with no `.newspack-page` anywhere, so every notice on those screens would have gone to zero.

Naming both containers in a shared package means the package knowing about one plugin's shell class, and still missing the next shell that does the same thing. The modifier keeps the opt-out at the one call site that needs it. I have recorded the reasoning in the README so the next person does not retry the scoped version.

While there I moved `--flush`, `--spaced-top` and `--spaced-bottom` out of `newspack-plugin`'s `admin/style.scss` into the component's own stylesheet, since they are the component's spacing.

packages/components/src/notice/index.js
thomasguillot: Fixed in 590ec3b48. The caller's value is taken only when `typeof spokenMessage === 'string'`; anything else falls through to the derivation, so core always gets a string and the docblock holds. The warning now keys off the same check rather than `=== undefined`.

packages/components/src/notice/index.js
thomasguillot: The entity half is fixed in 590ec3b48; the angle-bracket half turned out not to be reachable.

Tag stripping is now gated on `__unstableHTML`, and that branch runs `decodeEntities` as well, so a `&amp;` or `&#8217;` is announced as the character `RawHTML` displays rather than as its source. Everywhere else the derivation only collapses whitespace.

The `Unexpected token < in <body>` case ends up the same either way, though. `@wordpress/a11y`'s `filterMessage` applies `message.replace( /<[^<>]+>/g, ' ' )` to every message it speaks, which is the same pattern, so `<body>` is stripped downstream whatever we hand it. Gating the strip is still worth doing for the entities and for not touching content we have no reason to touch, but it does not recover the bracketed span. I have said so in the README rather than implying a fix that is not there.

packages/components/src/notice/index.js
thomasguillot: Narrowed the doc claim rather than the guard, in a887cbfdc. Both `notice.d.ts` and the README now say the warning catches a notice that derives nothing at all, and state plainly that partial loss goes unwarned.

I prototyped tightening it and did not keep it. Distinguishing "a component holding text we cannot read" from "a component contributing no text" is not decidable from the element: `<Notice status="error"><Icon icon={ x } />Failed</Notice>` has the same shape as your example, and would warn on a notice whose announcement is completely fine. A dev warning that fires on correct code gets ignored, which costs more than the case it catches, and as you note no call site hits it today.

packages/components/src/global-notices/index.js
thomasguillot: Added in 4890eede2. `global-notices/index.test.js` covers six cases, including the two you asked for: `_error_` asserting no `img` element is created and the text renders escaped, and `?newspack-notice[]=a` asserting the component renders nothing. It also pins the success/error split and the announce-the-first-only rule, so the `speak` behaviour has a test rather than only a README line.

Components suite is 347 now, up from 341.

plugins/newspack-plugin/src/wizards/audience/views/donations/configuration/index.tsx
thomasguillot: Fixed in 58dae4863. `interpolateOrPlainText` is now exported from `wizard/index.js` and re-exported from the barrel, and the trashed-products notice uses it, so a stray `</a>` in a translation degrades to the message with its tags stripped instead of throwing through a tab with no error boundary above it. `createInterpolateElement` is no longer imported in that file.

plugins/newspack-plugin/src/wizards/subscribers/screens/SubscriberList.jsx
thomasguillot: Fixed in 98067fbb0, by the second route you offer: the message is lifted into a local and passed as an explicit `spokenMessage` on all three screens, so the live region gets the message without the button label. `PersonProfile.jsx` has two of these and both are covered.

I did not move the button into `actions`, even though that is the semantically right home, because core recomputes the variant: `if ( variant !== 'primary' && ! noDefaultClasses ) computedVariant = ! url ? 'secondary' : 'link'`. The inline link buttons would come back as 40px secondary buttons in `.components-notice__actions` below the text unless every call site also passed `noDefaultClasses`. That is a visual change to three screens in a PR about announcements, so it is worth doing on its own if we want it.

plugins/newspack-plugin/src/wizards/componentsDemo/index.js
thomasguillot: Restored in 6de6e75a8. The card is back with all four statuses and a dismissible example, so step 8 does what it says. The `success` and `error` demos pass `spokenMessage=""` so opening the demo screen does not announce anything, and the dismissible one swaps for a button that brings it back, which keeps the card from emptying out once you have tried it.

Checked on the running site: five notices under a `Notice` heading, statuses `info`, `warning`, `success`, `error`, `info`, only the last with a close button, and both live regions empty on load.

packages/components/src/notice/README.md
thomasguillot: Fixed in a887cbfdc. The README, `notice.d.ts` and the warning string now all name `dangerouslySetInnerHTML` as the unreachable case, and the README says explicitly that `__unstableHTML` children are reached, with the tags removed and the entities decoded. Nobody should now add a redundant `spokenMessage` to those five call sites.

packages/components/src/notice/notice.d.ts
thomasguillot: Fixed in a887cbfdc. `disabled` and `openInNewTab` are gone, `isPrimary` is in. I also documented the `variant` recomputation on the interface, since that is the same class of surprise: a passed `variant` is overridden unless `noDefaultClasses` comes with it.

packages/components/src/notice/notice.d.ts
thomasguillot: Fixed in a887cbfdc. It now says `onDismiss` runs alongside `onRemove` rather than instead of it, and that on its own it leaves the notice on screen, which matches `onDismissNotice()` calling both.

packages/components/src/settings/SettingsCard.js
thomasguillot: Agreed, changed to `info` in 91282662d. `main` had `notificationLevel="info"` hardcoded, so `warning` was a silent behaviour change with nothing exercising it; keeping the old shape as the default is the honest version of "forwarded prop, same behaviour unless overridden".

With the default back, `single-segment.js` no longer needs the explicit prop, so I dropped it and that file now matches `main` exactly and leaves the diff.

thomasguillot:
**Follow-up on the review** (590ec3b48…6de6e75a8)

All thirteen threads have replies with the commit that closes them. Two of them push back rather than agree: the margin fix takes a different route than the one suggested, and the first-paint half of the `GlobalNotices` announcement finding did not reproduce. Details are on those threads. Here are the three items from the review body.

### The unencoded `newspack-notice` value

Fixed in 398aa140c, at both writers. `welcome/index.js:119` now wraps the whole value in `encodeURIComponent`, so a REST error reading "Failed to fetch, please retry" arrives as one error notice instead of an error plus a green success notice, and an `&` or `#` no longer truncates it. `class-newspack.php:348` now passes the value through `rawurlencode`, which is needed because `add_query_arg` builds its query with `build_query()`, and that calls `_http_build_query( $data, null, '&', '', false )` with encoding off. Confirmed against the running site: the message survives a comma, an ampersand and a hash intact.

### Screens with both `Notice` sources side by side

Documented rather than swept, because all three remaining call sites have a reason, and converting them would cost more than the inconsistency does.

`newspack-story-budget`'s `story-field.js` and `stories-edit.js` are both reachable from that plugin's editor entry: `editor/index.js` → `story-field-panel` → `story-field-panel-row` → `story-field`, and `editor/index.js` → `hooks/index.js` → `stories-edit`. Moving them onto the wrapper puts `newspack-components` into an editor bundle, which is the property this PR spent a round protecting.

`newspack-newsletters`' `local-list-modal.js` is the one I expected to convert, and I did, then reverted it. Its `admin-shell` siblings already carry the barrel, but the file is also reachable from the `wizard-bridge` entry via `local-list-modal-host.js`, and nothing else in that graph imports this package. Converting it pulls the whole barrel into that bundle, and with it `Page`'s `:root` block of token overrides and a second registration of the `newspack/wizards` store on a screen where `newspack-plugin`'s own bundle has already registered it.

So the rule is not "wrapper only", it is "the wrapper only from a bundle that already carries the package". That is now written down in `notice/README.md` with all three call sites named, so the next person reads it as policy rather than as an oversight.

### `@wordpress/a11y` not declared

Added to `devDependencies` in d7e0132f7.

### Test plan

Step 8 works again now that the demo card is back. I have not corrected step 4 in the description yet, since the modal notice needing an errored or disabled endpoint is a fair point and I would rather fix the wording once you have looked at the rest.

### Checks

Components 347 (was 341, the six new ones are `GlobalNotices`), `newspack-plugin` 907. ESLint, Stylelint and PHPCS clean on everything touched. Verified in the browser on WP 7.1: the multi-notice case leaves the error in `#a11y-speak-assertive` with polite empty; global notices resolve to `margin: 0` above the page header while in-page notices stay at 32px and modal notices keep flush-top with a 32px bottom; the demo card shows four statuses and a working dismiss; and the segment editor, Connections, Advanced Settings, Donations, Subscribers and the setup wizard all render with no console errors.

thomasguillot:
**Correcting one thing from the comment above: `@wordpress/a11y` cannot be declared here** (a8c470c60)

I said the `devDependencies` entry was added in d7e0132f7. It was, CI went red on it, and I have reverted it. The entry is not safe to add, and the reason is worth writing down so it does not get suggested again.

`packages/components` builds against `@wordpress/components@^30.6.0`, which resolves to 30.9.0, and that resolves `@wordpress/a11y@4.48.0`. Declaring `"@wordpress/a11y": "^4.48.0"` makes pnpm install 4.49.0 into `packages/components/node_modules`, so the package now has a copy that core does not use. `jest.mock( '@wordpress/a11y' )` resolves from the test file and mocks 4.49.0; core's `Notice` imports 4.48.0 and calls the real `speak`. The mock records nothing:

```
● Notice › announces string content
    Expected: "Something failed", "assertive"
    Number of calls: 0
```

Eight tests across `notice/index.test.js` and `global-notices/index.test.js` fail that way, which is every test that asserts an announcement.

An exact `4.48.0` pin does dedupe and does keep the suite green, but it has to match another package's transitive resolution to work, with nowhere in JSON to say so, so the next `@wordpress/components` bump re-breaks the suite with a failure that reads as "the component stopped announcing". Given `devDependencies` are not installed for consumers of a published package, and this package is only ever developed inside this workspace where the root already provides `a11y`, the entry buys little and costs that. `pnpm-lock.yaml` is back to matching `main` exactly, so the PR carries no lockfile change.

Worth noting the local/CI gap that hid this: CI checks out the PR **merge** commit, so it was running `main`'s new `stat-card` suite too, which is where the 421-vs-347 test count came from. I have since run the merge locally: 27 suites and 421 tests pass in `newspack-components`, 97 suites and 974 tests in `newspack-plugin`, and `main` merges clean, including `componentsDemo/index.js`, which we both touched.

All 43 checks are green.

PR #893, opened 2026-08-17
fix(wizard): refetch data after the installer adds required plugins

dkoo:
### All Submissions:

Setting up Donations on a fresh site left publishers looking at an empty Configuration screen the moment they activated WooCommerce and WooCommerce Subscriptions from the wizard. Every option appeared once they clicked Save, with nothing on screen to suggest that was the way out.

A wizard fetches its data once, when it mounts. With `requiredPlugins` set, that fetch runs while the plugins are still missing, and endpoints that depend on them answer empty or fail outright. The plugin installer then reported success by flipping React state alone, so nothing re-asked the server and the section rendered against the plugin-less response.

The shared `Wizard` now records that the installer reported the requirements unmet, and invalidates its `getWizardAPIData` resolution once they are satisfied. Wizards whose plugins were already active when the page loaded are left alone: that fetch saw the real site, and refetching would cost them a second request on every load.

Of the four `requiredPlugins` call sites across the monorepo and the standalone repos, Donations is the only one that passes an `apiSlug`, so it is the only screen whose behavior changes.

Closes NPPM-3123.

1. Start from a site with WooCommerce and WooCommerce Subscriptions inactive, and the reader revenue platform left at its default. Audience Management does not need to be set up.
2. Go to **Audience > Donations**. The wizard shows the required-plugins screen.
3. Activate WooCommerce, then WooCommerce Subscriptions, from that screen.
4. The Configuration form appears on its own. Suggested donation amounts and the minimum donation field are populated, and no save was needed. Before this change the screen was empty.
5. Reload the page. The form matches what step 4 showed.

### Other information:

**Verified in a local environment**, as an A/B where the only variable was the compiled bundle. Same site state both rounds: WooCommerce and WooCommerce Subscriptions present but inactive, no Audience Management setup, no donation products, and the wizard endpoint returning `400 newspack_missing_required_plugin` on both runs.

* **Before:** activating both plugins left the Configuration tab showing its heading and nothing else. Clicking Save filled the form in, matching the report.
* **After:** the form rendered populated the moment activation finished, with no Save. The network trace is the mechanism exactly — `GET /newspack/v1/wizard/newspack-audience-donations` returns 400 on page load, then 200 after activation, and no POST is involved.
* Reloading afterward produced an identical form, no console errors, and no new entries in `debug.log`.

**Known limitation, checked.** Data localized into the page at render time is not reconciled here, because `wp_localize_script` output cannot be invalidated. On this screen that is `can_use_name_your_price`, which gates the Donation Type selector. Name Your Price is not one of the wizard's required plugins, so this flow never activates it. The post-activation form and a freshly reloaded one matched field for field in testing, so this does not bite on a fresh install. A site carrying Name Your Price while WooCommerce was inactive would still need a reload before the selector appears.

**Scope.** The Subscriptions and Newsletters wizards use the same install-and-continue flow but pass no `apiSlug`, so they are unchanged. Subscriptions reads its state from a page-load global rather than the wizard store, so a store refetch would not repair it there.

**Follow-ups, filed separately rather than carried by a hotfix:**

* The wizard store records a failed fetch as a successful resolution. `FETCH_FROM_API` catches the rejection into `setError` and resolves `undefined`, so the resolver still marks the slug resolved. Any failed wizard fetch pins that screen to empty until a page reload. The plugin-install flow is one door into this; this change closes that one.
* `Audience_Donations::api_get_donation_settings()` returns a bare 400 when WooCommerce is inactive, while the same class of failure with Subscriptions inactive returns 200 with a `donation_data.errors` payload both clients already degrade on.
* `WizardError` only mounts inside `TabbedNavigation`, which needs more than one visible section, so a single-section wizard shows nothing when its fetch fails. Mounting it unconditionally would be worse: the donations error carries `level: fatal`, which renders as a blocking modal, so it would land over the required-plugins screen. The fix has to suppress or downgrade errors while requirements are unmet.

<details>

**Mechanism, at `origin/release`:**

1. `Audience_Donations::api_get_donation_settings()` returns a `WP_Error` (`newspack_missing_required_plugin`, 400) when the reader revenue platform is `wc`, which is the default from `Donations::get_platform_slug()`, and WooCommerce is inactive.
2. `Wizard` triggers its only fetch on mount through the `getWizardAPIData` resolver. Resolution is cached per slug, so it runs exactly once.
3. The store's `FETCH_FROM_API` control swallows the rejection into `setError` and resolves `undefined`, so `apiData[ slug ]` ends up empty and the resolution is recorded as finished.
4. `PluginInstaller` completion only called `setPluginRequirementsSatisfied`.
5. `DonationAmounts` early-returns `null` on `! wizardData.donation_data`, and `wizardData.donation_page` is undefined, so the section rendered its title and nothing else.

Saving repaired it because `saveWizardSettings` writes the POST response into the store.

**Why a ref rather than comparing state:** `pluginRequirementsSatisfied` starts `false` for any non-empty `requiredPlugins`, so a state comparison cannot tell "the installer just activated something" from "the plugins were already active." `PluginInstaller` reports `complete: true` on its very first callback in the already-active case, which is what the ref distinguishes.

`setPluginRequirementsSatisfied` runs before the invalidation so that a failure in the dispatch could not strand someone on a required-plugins screen for a plugin they had just installed. Ordering is not load-bearing for correctness: both are synchronous, and the re-render re-runs the `useSelect` either way.

**Verification:** the new test reproduces the mechanism (the section renders nothing after activation) and passes with the change. Mutation testing confirms both tests discriminate: reverting the component fails only the refetch test, and defeating the ref guard so it always invalidates fails only the already-active control. `packages/components` 14/14 suites and 97 tests; `newspack-plugin` 60/60 suites and 567 tests.

**Known edge, not guarded:** `PluginInstaller.retrievePluginInfo` keeps only slugs the plugins endpoint returns, and `Object.values( … ).every( … )` is vacuously true for an empty object. A slug the plugin manager does not know reports `complete: true` on the first callback, so the latch never sets and no refetch happens. Pre-existing, and the first thing to check if this is ever reported for one specific slug.

</details>

### DIFF
--- a/packages/components/src/wizard/index.js
+++ b/packages/components/src/wizard/index.js
@@ -150,6 +150,10 @@ const Wizard = (
 	// wizard a second request on every load.
 	const requirementsWereUnmet = useRef( false );
 	const onPluginStatus = ( { complete } ) => {
+		// Leave the installer first: whatever happens to the refetch, the user
+		// should not be stranded on a required-plugins screen for a plugin they
+		// just installed.
+		setPluginRequirementsSatisfied( complete );
 		if ( ! complete ) {
 			requirementsWereUnmet.current = true;
 		} else if ( requirementsWereUnmet.current ) {
@@ -158,7 +162,6 @@ const Wizard = (
 				invalidateResolution( 'getWizardAPIData', [ apiSlug ] );
 			}
 		}
-		setPluginRequirementsSatisfied( complete );
 	};
 
 	if ( ! pluginRequirementsSatisfied ) {
--- a/packages/components/src/wizard/index.js
+++ b/packages/components/src/wizard/index.js
@@ -126,6 +126,7 @@ const Wizard = (
 	const isQuietLoading = useSelect( select => select( WIZARD_STORE_NAMESPACE ).isQuietLoading() );
 	const headerData = useSelect( select => select( WIZARD_STORE_NAMESPACE ).getHeaderData() );
 	const notices = useSelect( select => select( WIZARD_STORE_NAMESPACE ).getNotices() );
+	const { invalidateResolution } = useDispatch( WIZARD_STORE_NAMESPACE );
 	const { actions, backNav, badges, sectionDescription, sectionMenu, sectionName, sectionTitle, sectionPrimaryAction, sectionSecondaryAction } =
 		headerData;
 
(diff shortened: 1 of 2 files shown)


dkoo:
Passed manual QA by a teammate—merging!

PR #897, opened 2026-08-18
fix(data-events): announce readers created during Store API checkouts

wil-gerken:
### All Submissions:

When a reader creates an account during checkout, Newspack dispatches `newspack_registered_reader_via_woo`. Two subscribers hang off it: the `reader_registered` data event, which carries the reader to the other sites in a network and to the publisher's email service, and session hydration, which binds the analytics client ID and sets the reader auth cookie.

Checkouts through WooCommerce's Store API never dispatch it. The existing check waits for a signal only the classic checkout emits, and the Store API does not emit it, so the account is created and both subscribers are skipped. Wallet payments joined that transport when woocommerce-gateway-stripe 9.2.0 made the Blocks API its default for express checkout, so no Newspack code changed. This affects express wallet buttons (Apple Pay, Google Pay) and the block checkout page alike. On a site using the block checkout, that is most checkout registrations.

The existing handler now also subscribes to the Store API's own pre-account-creation signal, so both pipelines announce the accounts they create, with the same campaign metadata.

Reference: NPPM-3147.

**Setup.** All three are required; none is the default on a stock install.

1. On a site with WooCommerce and reader activation, create a page containing the Checkout block. The default checkout page uses the `[woocommerce_checkout]` shortcode, which posts through the classic pipeline and will not exercise this change.
2. Confirm `woocommerce_enable_signup_and_login_from_checkout` is `yes`. Newspack's WooCommerce setup already sets it, and the Store API refuses to create an account without it.
3. Enable an offline payment gateway (WooCommerce → Settings → Payments → Cash on delivery) and make sure a purchasable product exists. Neither is present on a stock install.

**Reproduce the problem** on `release`.

3. Add a product to the cart. Check out on the block page with a brand-new email.
4. The order completes and an account is created for that email. No `reader_registered` event fires — though other events will, and seeing them confirms the logging works. Watch the hub event log, or drop a mu-plugin that hooks `newspack_data_event` and writes each event name to `debug.log`.

**Verify the fix** on this branch.

5. In a new private window, repeat step 3 with a second brand-new email. A `reader_registered` event fires carrying `registration_method: woocommerce`.
6. In a new private window, add the product to the cart again, then check out with a third new email through the classic checkout form at `/checkout/?modal_checkout=1`. The event fires exactly once.
7. Still signed in as the account from step 6, add the product again and check out on the block page. No `reader_registered` event fires.

The block checkout shows no "Create an account" checkbox, and that is expected: reader activation turns guest checkout off, so WooCommerce requires an account and stops offering the choice. Every guest checkout creates one. Each completed checkout also signs the tester in as the account it created, which is why steps 5 and 6 each start in a new private window.

Express wallet buttons submit through the same Store API transport as the block checkout, so steps 3 and 5 exercise the wallet path without configuring Stripe.

### Release risk: High

| Signal | Score | Evidence |
|--------|-------|----------|
| Contract surface | Medium | `data-events`; an existing event fires under wider conditions, no signature change |
| Surface type | Medium | Hook subscription inside a checkout flow. Decides no charge or access outcome, but does start session hydration where it previously did not run |
| Publisher exposure | High | Foundation code, no flag; every site with WooCommerce, reader activation, and a Store API checkout |
| Change shape | Medium | +33/-9 production in one file, which took a prior fix in #4310; seven tests added |

High because of publisher exposure: foundation code with no flag, on every site with WooCommerce, reader activation, and a Store API checkout. The other three signals are Medium. Verified locally against the base commit in one environment: without the change, a Store API checkout creates the order and the account and announces nothing; with it, `reader_registered` fires once carrying `registration_method: woocommerce`. Two review findings — both cases of checkout state outliving the checkout that set it, reachable because the block checkout routinely batches several Store API requests into one PHP process — were reproduced, fixed, and covered by tests.

Worth testing harder on: announcement volume on a site using the block checkout. The subscribed hook is not wallet-specific, so counts rise on first deploy by design.

### Other information:

Seven new tests cover both pipelines, the campaign-metadata harvest, batch isolation, checkout-state consumption, and the no-signal guard. Full suite 2751 passing.

**Scope of the change.** This restores the registration dispatch and both of its subscribers, on both checkout pipelines. Membership access is unchanged: access to a network-wide tier travels on a separate event that fires when the membership record is written, regardless of checkout transport. Verified on a hub-and-node pair, where a buyer with no registration announcement still arrived on the node as an active member of the network tier.

**Two known limits, both pre-existing and out of scope.** Store API registrations sync to the ESP without a `REGISTRATION_PAGE` value: the referer-based branch needs request parameters the JSON transport doesn't carry. Order-completed sync already pushes the same empty value, so nothing regresses. And WooCommerce's delayed-account-creation flow, where an account is created later from the order-confirmation page, is announced by neither pipeline; that gap predates this change.

### DIFF
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -37,6 +37,14 @@ final class Woo_User_Registration {
 		// is processing checkout?
 		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'checkout_process' ] );
 
+		// The Store API checkout — the transport express wallets like Apple Pay and
+		// Google Pay submit through — never fires woocommerce_checkout_process. This
+		// action is that pipeline's equivalent signal: it fires once per checkout
+		// POST, before the account is created, while the cart is loaded for the
+		// metadata harvest. It also fires on the pay-for-existing-order route, where
+		// no account is ever created, so the flag set there is never consumed.
+		add_action( 'woocommerce_store_api_checkout_update_customer_from_request', [ __CLASS__, 'checkout_process' ], 10, 0 );
+
 		// created a user?
 		add_action( 'woocommerce_created_customer', [ __CLASS__, 'created_customer' ], 1 );
 
@@ -58,15 +66,20 @@ final class Woo_User_Registration {
 		 *
 		 * Here, we are going to read the same information from the cart and use it to send the metadata to Newspack on registration events.
 		 */
-		foreach ( \WC()->cart->get_cart() as $cart_item_key => $values ) {
-			if ( ! empty( $values['newspack_popup_id'] ) ) {
-				self::$metadata['newspack_popup_id'] = $values['newspack_popup_id'];
-			}
-			if ( ! empty( $values['prompt_title'] ) ) {
-				self::$metadata['prompt_title'] = $values['prompt_title'];
-			}
-			if ( ! empty( $values['referer'] ) ) {
-				self::$metadata['referer'] = $values['referer'];
+		// Cart presence is the calling pipeline's guarantee, not ours — without
+		// one there is simply no campaign metadata to harvest, and the flag
+		// below must still be set so the created account is announced.
+		if ( ! empty( \WC()->cart ) ) {
+			foreach ( \WC()->cart->get_cart() as $cart_item_key => $values ) {
+				if ( ! empty( $values['newspack_popup_id'] ) ) {
+					self::$metadata['newspack_popup_id'] = $values['newspack_popup_id'];
+				}
+				if ( ! empty( $values['prompt_title'] ) ) {
+					self::$metadata['prompt_title'] = $values['prompt_title'];
+				}
+				if ( ! empty( $values['referer'] ) ) {
+					self::$metadata['referer'] = $values['referer'];
+				}
 			}
 		}
 
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -59,6 +59,11 @@ final class Woo_User_Registration {
 	 */
 	public static function checkout_process() {
 
+		// One signal, one checkout: the Store API's batch route serves several
+		// checkouts in a single PHP process, so metadata harvested for an
+		// earlier one must not be attributed to this reader.
+		self::$metadata = [];
+
 		/**
 		 * On Newspack\Donations::process_donation_request(), we add these values to the cart.
 		 *
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -102,6 +102,12 @@ final class Woo_User_Registration {
 			return;
 		}
 
+		// A checkout announces the one account it creates. Consuming the flag
+		// here keeps an account created later in the same process — the Store
+		// API batch route serves several checkouts in one — from being
+		// announced on this checkout's behalf.
+		self::$processing_checkout = false;
+
 		$user = get_user_by( 'id', $user_id );
 
 		if ( ! $user ) {
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -15,11 +15,33 @@ use Newspack\Reader_Activation;
 final class Woo_User_Registration {
 
 	/**
-	 * Whether the current request is processing a checkout.
+	 * Whether a classic checkout is in flight.
+	 *
+	 * Raised by woocommerce_checkout_process and never lowered. The classic
+	 * pipeline runs one checkout per request and can create more than one
+	 * account in it — Subscriptions Gifting creates the recipient on the
+	 * order-status transition — and both are that checkout's to announce.
 	 *
 	 * @var boolean
 	 */
 	private static $processing_checkout = false;
+
+	/**
+	 * Whether a Store API checkout is in flight.
+	 *
+	 * @var boolean
+	 */
+	private static $store_api_checkout = false;
+
+	/**
+	 * Folded billing email the in-flight Store API checkout expects to create.
+	 *
+	 * Empty means the signal carried none, in which case the first account
+	 * created stands in — see claim_account().
+	 *
+	 * @var string
+	 */
+	private static $store_api_expected_email = '';
 
 	/**
 	 * Metadata to send with the registration event.
@@ -41,9 +63,9 @@ final class Woo_User_Registration {
 		// Google Pay submit through — never fires woocommerce_checkout_process. This
 		// action is that pipeline's equivalent signal: it fires once per checkout
 		// POST, before the account is created, while the cart is loaded for the
-		// metadata harvest. It also fires on the pay-for-existing-order route, where
-		// no account is ever created, so the flag set there is never consumed.
-		add_action( 'woocommerce_store_api_checkout_update_customer_from_request', [ __CLASS__, 'checkout_process' ], 10, 0 );
+		// metadata harvest. It carries the customer, whose billing email is what
+		// tells the created account apart from any other in the same process.
+		add_action( 'woocommerce_store_api_checkout_update_customer_from_request', [ __CLASS__, 'store_api_checkout_process' ], 10, 2 );
 
 		// created a user?
 		add_action( 'woocommerce_created_customer', [ __CLASS__, 'created_customer' ], 1 );
@@ -53,11 +75,42 @@ final class Woo_User_Registration {
 	}
 
 	/**
-	 * During the checkout process, set the processing_checkout flag and store metadata.
+	 * Note a classic checkout is under way.
 	 *
 	 * @return void
 	 */
 	public static function checkout_process() {
+		self::harvest_cart_metadata();
+		self::$processing_checkout = true;
+	}
+
+	/**
+	 * Note a Store API checkout is under way, and whose account it expects to create.
+	 *
+	 * Unlike the classic signal this one carries the customer, so the account
+	 * this checkout creates can be told apart from any other created in the same
+	 * process. WooCommerce writes the posted billing email onto the customer
+	 * before firing this, and refuses to create an account without one, so an
+	 * empty email here means no account will follow.
+	 *
+	 * @param object|null $customer WC_Customer the checkout is building.
+	 * @param object|null $request  The checkout request. Unused; part of the action signature.
+	 * @return void
+	 */
+	public static function store_api_checkout_process( $customer = null, $request = null ) {
+		self::harvest_cart_metadata();
+		self::$store_api_checkout       = true;
+		self::$store_api_expected_email = is_object( $customer ) && method_exists( $customer, 'get_billing_email' )
+			? self::fold_email( $customer->get_billing_email() )
+			: '';
+	}
+
+	/**
+	 * Read the campaign attribution the current checkout's cart carries.
+	 *
+	 * @return void
+	 */
+	private static function harvest_cart_metadata() {
 
 		// One signal, one checkout: the Store API's batch route serves several
 		// checkouts in a single PHP process, so metadata harvested for an
@@ -72,8 +125,8 @@ final class Woo_User_Registration {
 		 * Here, we are going to read the same information from the cart and use it to send the metadata to Newspack on registration events.
 		 */
 		// Cart presence is the calling pipeline's guarantee, not ours — without
-		// one there is simply no campaign metadata to harvest, and the flag
-		// below must still be set so the created account is announced.
+		// one there is simply no campaign metadata to harvest. Announcing the
+		// reader does not depend on it; the callers set their state either way.
 		if ( ! empty( \WC()->cart ) ) {
 			foreach ( \WC()->cart->get_cart() as $cart_item_key => $values ) {
 				if ( ! empty( $values['newspack_popup_id'] ) ) {
@@ -87,26 +140,69 @@ final class Woo_User_Registration {
 				}
 			}
 		}
-
-		self::$processing_checkout = true;
 	}
 
 	/**
-	 * If a user was created during the checkout by Woo, fire an event.
+	 * Decide whether the in-flight checkout is the one that created this account.
+	 *
+	 * The two pipelines answer this differently, on purpose. The classic signal
+	 * carries nothing identifying the buyer, so it keeps the long-standing rule
+	 * — any account created while a classic checkout is in flight is that
+	 * checkout's — and announces every one of them. The Store API signal fires
+	 * on every checkout POST, including a logged-in shopper's where no account
+	 * follows, so leaving it standing would let it speak for an unrelated
+	 * account later in the same process. It announces only the account matching
+	 * the email it carried, and stands down once it has.
+	 *
+	 * @param string $email The created account's email address.
+	 * @return bool Whether this checkout announces this account.
+	 */
+	private static function claim_account( $email ) {
+		if ( self::$store_api_checkout ) {
+			// No email on the signal, so nothing to match against and the first
+			// account created stands in. Unreachable on the announce path today:
+			// WooCommerce will not create an account without a billing email, so a
+			// signal without one is followed by no account at all.
+			if ( '' === self::$store_api_expected_email ) {
+				self::$store_api_checkout = false;
+				return true;
+			}
+
+			if ( self::fold_email( $email ) === self::$store_api_expected_email ) {
+				self::$store_api_checkout       = false;
+				self::$store_api_expected_email = '';
+				return true;
+			}
+		}
+
+		return self::$processing_checkout;
+	}
+
+	/**
+	 * Fold an email address for comparison.
+	 *
+	 * The two sides arrive by different routes — one off the customer object as
+	 * posted, the other off the saved user record — so they are compared folded
+	 * rather than raw. A mismatch would drop the announcement silently and
+	 * restore the original bug with the suite still green.
+	 *
+	 * @param string $email Email address.
+	 * @return string
+	 */
+	private static function fold_email( $email ) {
+		return strtolower( trim( (string) $email ) );
+	}
+
+	/**
+	 * Announce an account a checkout created, with what is known about where it came from.
 	 *
 	 * @param int $user_id The ID of the created user.
 	 * @return void
 	 */
 	public static function created_customer( $user_id ) {
-		if ( ! self::$processing_checkout ) {
+		if ( ! self::$processing_checkout && ! self::$store_api_checkout ) {
 			return;
 		}
-
-		// A checkout announces the one account it creates. Consuming the flag
-		// here keeps an account created later in the same process — the Store
-		// API batch route serves several checkouts in one — from being
-		// announced on this checkout's behalf.
-		self::$processing_checkout = false;
 
 		$user = get_user_by( 'id', $user_id );
 
@@ -114,7 +210,13 @@ final class Woo_User_Registration {
 			return;
 		}
 
-		// If a user is created later on the request by woocommerce_created_customer(), it will at least have this data.
+		if ( ! self::claim_account( $user->user_email ) ) {
+			return;
+		}
+
+		// Every account a checkout creates carries the method. The page and
+		// campaign below are best-effort on top of it, and depend on what the
+		// cart happened to hold.
 		self::$metadata['registration_method'] = 'woocommerce';
 
 		// For modal checkout, the referer is actually what we want to capture as the registration page.
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -59,12 +59,7 @@ final class Woo_User_Registration {
 		// is processing checkout?
 		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'checkout_process' ] );
 
-		// The Store API checkout — the transport express wallets like Apple Pay and
-		// Google Pay submit through — never fires woocommerce_checkout_process. This
-		// action is that pipeline's equivalent signal: it fires once per checkout
-		// POST, before the account is created, while the cart is loaded for the
-		// metadata harvest. It carries the customer, whose billing email is what
-		// tells the created account apart from any other in the same process.
+		// The Store API's equivalent signal; see store_api_checkout_process().
 		add_action( 'woocommerce_store_api_checkout_update_customer_from_request', [ __CLASS__, 'store_api_checkout_process' ], 10, 2 );
 
 		// created a user?
@@ -87,8 +82,14 @@ final class Woo_User_Registration {
 	/**
 	 * Note a Store API checkout is under way, and whose account it expects to create.
 	 *
-	 * Unlike the classic signal this one carries the customer, so the account
-	 * this checkout creates can be told apart from any other created in the same
+	 * The Store API checkout — the transport express wallets like Apple Pay and
+	 * Google Pay submit through — never fires woocommerce_checkout_process. The
+	 * action below is that pipeline's equivalent signal: it fires once per
+	 * checkout POST, before the account is created, while the cart is loaded for
+	 * the metadata harvest.
+	 *
+	 * Unlike the classic signal it carries the customer, so the account this
+	 * checkout creates can be told apart from any other created in the same
 	 * process. WooCommerce writes the posted billing email onto the customer
 	 * before firing this, and refuses to create an account without one, so an
 	 * empty email here means no account will follow.
--- a/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
+++ b/plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
@@ -36,8 +36,8 @@ final class Woo_User_Registration {
 	/**
 	 * Folded billing email the in-flight Store API checkout expects to create.
 	 *
-	 * Empty means the signal carried none, in which case the first account
-	 * created stands in — see claim_account().
+	 * Empty means the signal carried none, which means no account will follow
+	 * it, so nothing is announced against it — see claim_account().
 	 *
 	 * @var string
 	 */
@@ -159,13 +159,13 @@ final class Woo_User_Registration {
 	 */
 	private static function claim_account( $email ) {
 		if ( self::$store_api_checkout ) {
-			// No email on the signal, so nothing to match against and the first
-			// account created stands in. Unreachable on the announce path today:
-			// WooCommerce will not create an account without a billing email, so a
-			// signal without one is followed by no account at all.
+			// A signal without an email creates no account: WooCommerce refuses a
+			// customer without a valid billing address. So any account arriving
+			// while this is empty belongs to something else in the request, and
+			// announcing it would name the wrong reader — the same harm this
+			// keying exists to prevent.
 			if ( '' === self::$store_api_expected_email ) {
-				self::$store_api_checkout = false;
-				return true;
+				return false;
 			}
 
 			if ( self::fold_email( $email ) === self::$store_api_expected_email ) {
--- a/plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
+++ b/plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
@@ -169,6 +169,37 @@ class Newspack_Test_Woo_User_Registration extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Each checkout signal harvests its own campaign metadata. The Store API
+	 * batch route serves up to 25 requests in one PHP process, so a second
+	 * checkout must not inherit the first one's attribution.
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_second_checkout_does_not_inherit_campaign_metadata() {
+		$this->stub_wc_with_cart(
+			[
+				'item1' => [
+					'newspack_popup_id' => 123,
+					'referer'           => '/donate/',
+				],
+			]
+		);
+		do_action( 'woocommerce_store_api_checkout_update_customer_from_request', null, new WP_REST_Request( 'POST', '/wc/store/v1/checkout' ) );
+		do_action( 'woocommerce_created_customer', self::factory()->user->create( [ 'user_email' => 'batch-first@example.test' ] ) );
+
+		// Second checkout in the same process, with nothing to harvest.
+		WC()->cart = new WC_Cart( [] );
+		do_action( 'woocommerce_store_api_checkout_update_customer_from_request', null, new WP_REST_Request( 'POST', '/wc/store/v1/checkout' ) );
+		do_action( 'woocommerce_created_customer', self::factory()->user->create( [ 'user_email' => 'batch-second@example.test' ] ) );
+
+		$this->assertCount( 2, $this->fired );
+		$this->assertArrayNotHasKey( 'newspack_popup_id', $this->fired[1]['metadata'], 'The second checkout should not inherit the first one\'s campaign attribution.' );
+		$this->assertArrayNotHasKey( 'referer', $this->fired[1]['metadata'] );
+		$this->assertSame( 123, $this->fired[0]['metadata']['newspack_popup_id'], 'The first checkout keeps its own attribution.' );
+	}
+
+	/**
 	 * Classic checkout keeps announcing (regression control).
 	 *
 	 * @runInSeparateProcess
--- a/plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
+++ b/plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
@@ -223,6 +223,27 @@ class Newspack_Test_Woo_User_Registration extends WP_UnitTestCase {
 	}
 
 	/**
+	 * A checkout announces the one account it creates. The Store API batch
+	 * route serves several sub-requests in a single PHP process, so the state a
+	 * checkout raises must not still be standing when an account is created
+	 * later in that process without a checkout of its own.
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_checkout_state_is_consumed_by_the_account_it_announces() {
+		$this->stub_wc_with_cart();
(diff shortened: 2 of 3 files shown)


plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
```diff
@@ -58,15 +66,20 @@ public static function checkout_process() {
...
-			}
-			if ( ! empty( $values['prompt_title'] ) ) {
-				self::$metadata['prompt_title'] = $values['prompt_title'];
-			}
-			if ( ! empty( $values['referer'] ) ) {
-				self::$metadata['referer'] = $values['referer'];
+		// Cart presence is the calling pipeline's guarantee, not ours — without
+		// one there is simply no campaign metadata to harvest, and the flag
+		// below must still be set so the created account is announced.
+		if ( ! empty( \WC()->cart ) ) {
```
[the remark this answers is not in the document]
wil-gerken: Good catch, and it turned out to be reachable rather than defensive. The Store API's batch route (`src/StoreApi/Routes/V1/Batch.php`) serves up to 25 `/wc/store` requests inside one PHP process with `permission_callback => __return_true`, so two checkouts really can share this static state — the second reader would then be announced with the first one's campaign attribution.

Added a test that fires the signal twice in one process, harvesting `newspack_popup_id` and `referer` on the first and nothing on the second. It failed on the second announcement inheriting `newspack_popup_id`, which is the leak you described. `self::$metadata` is now cleared at the top of `checkout_process()`, and the test passes.

The processing flag stays where it was, at the end of the method: it marks the request as a checkout after the harvest, and both signals set it the same way.

Fixed in 459e12afd.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
```diff
@@ -37,6 +37,14 @@ public static function init() {
 		// is processing checkout?
 		add_action( 'woocommerce_checkout_process', [ __CLASS__, 'checkout_process' ] );
 
+		// The Store API checkout — the transport express wallets like Apple Pay and
+		// Google Pay submit through — never fires woocommerce_checkout_process. This
+		// action is that pipeline's equivalent signal: it fires once per checkout
+		// POST, before the account is created, while the cart is loaded for the
+		// metadata harvest. It also fires on the pay-for-existing-order route, where
+		// no account is ever created, so the flag set there is never consumed.
+		add_action( 'woocommerce_store_api_checkout_update_customer_from_request', [ __CLASS__, 'checkout_process' ], 10, 0 );
```
adekbadek: **Blocker:** A checkout that creates no account leaves the flag raised for the rest of the request — The comment notes that the signal also fires on the pay-for-existing-order route, where no account is created, and treats the flag left standing there as harmless. Nothing lowers it — and the exposure is wider than that route, because an ordinary checkout reaches the same state.

**Within a single `/wc/store/v1/checkout` request**, `Checkout.php` runs: `:442` `update_customer_from_request()`, which fires the signal at `:793` unconditionally — flag raised, cart metadata harvested; then `:452` `process_customer()`, which decides at `:849` whether to create an account, and `should_create_customer_account()` returns false for a logged-in shopper (`:892`); then `:557` `process_payment()`. On success the order reaches `processing` in that same request, firing `woocommerce_order_status_processing`, where a third-party plugin can call `wc_create_new_customer()` — Subscriptions gifting does, at priority 8. That account is then announced as this checkout's reader.

So for any logged-in reader buying anything through the block checkout, the flag is raised, never consumed, and still standing when payment runs. No batching is involved. The batch route is an amplifier on top: `/wc/store/v1/batch` serves up to 25 sub-requests in one process, so the raised flag additionally survives into later sub-requests.

The flag consumed at line 109 cannot help here — there is no account to consume it.

**Measured on the review env:**
1. A logged-in Store API checkout fired the signal, logged no `woocommerce_created_customer` at all, and proceeded to `processing` — the raised-and-unconsumed state, reached by one ordinary request.
2. Firing the signal with `newspack_popup_id` and `referer` in the cart, creating no account on it, then creating an unrelated customer in the same PHP process announced that unrelated reader carrying `{"newspack_popup_id":4242,"referer":"/support-us/","registration_method":"woocommerce"}` — a real reader attributed to a campaign they never saw.

The gifting half is read from source rather than driven end to end: the hook registrations and the call into `wc_create_new_customer()` were verified in the checked-out plugin, and on-site gateways complete payment in-request, but I did not run a real gift purchase.

The same standing-flag shape does exist on the classic pipeline today for a logged-in shopper, so this is an extension rather than an invention — onto the transport that now carries most checkouts. The action passes the request, so keying the raised state to the identity the checkout is for, or clearing it once the route has passed the point at which it would have created an account, closes it without touching the classic path.

**Risk if not addressed:** That announcement also drives the session subscriber, which writes the announced reader's email address into the responding browser's reader cookie and binds that browser's analytics client id to the announced reader's id. The purchaser's browser therefore finishes the checkout displaying a different reader's email as the signed-in reader, and the purchaser's own session-hydration call then fails against the rebound id, so the post-checkout refresh never completes and the front-end reader state stays stale until the next full page load. It fails closed rather than open — the hydration endpoint compares the stored id against the logged-in user and returns 403 on a mismatch, so no session is handed to the wrong person — but on a network the recipient is simultaneously propagated to every site carrying a checkout attribution that is not theirs.

**Consider:** A checkout that creates no account leaves the state raised for the rest of the request, which is what lets a later account creation claim it. Scoping the raised state to the identity the checkout is for, or clearing it once the route has passed the point at which it would have created an account, closes that window without changing the classic path.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
```diff
@@ -84,6 +102,12 @@ public static function created_customer( $user_id ) {
 			return;
 		}
 
+		// A checkout announces the one account it creates. Consuming the flag
+		// here keeps an account created later in the same process — the Store
+		// API batch route serves several checkouts in one — from being
+		// announced on this checkout's behalf.
+		self::$processing_checkout = false;
```
adekbadek: **Blocker:** Consuming the flag here drops the second account of a two-account checkout — Consuming `self::$processing_checkout` at the top of `created_customer()` means a checkout that creates **two** accounts announces only the first. WooCommerce Subscriptions ships gifting in-plugin (`class-wc-subscriptions-plugin.php:320` loads it unconditionally), and `WCSG_Recipient_Management::maybe_create_recipient` is registered on `woocommerce_order_status_processing`/`_completed` at priority 8 (`class-wcsg-recipient-management.php:53-54`), reaching `wc_create_new_customer()` via `WCS_Gifting::create_recipient_user()` (`class-wcs-gifting.php:650`) for a recipient who has no account yet. For a payment that completes in-request, `WC_Checkout::process_checkout()` fires `woocommerce_checkout_process` (:1303), creates the purchaser (:1167), then reaches `processing` inside the same call (:1370), so the recipient's `woocommerce_created_customer` arrives second. Newspack supports gifting first-party (`class-woocommerce-subscriptions-gifting.php:103`, `newspack-blocks/includes/class-modal-checkout.php:192`).

Measured on the review env, driving the real class: with this branch, one checkout signal followed by two `wc_create_new_customer()` calls announces only the first. Restoring the pre-change flag state between them (the old code never cleared it) announces both.

The pre-change behaviour was not right either, which is why the answer is not a revert: `self::$metadata` is static and was never reset, so the recipient was announced carrying the **purchaser's** `newspack_popup_id`/`referer`, and `Session_Hydration::on_woo_customer_created` (`class-session-hydration.php:38`) rebound the purchaser's analytics CID to the recipient's user id and wrote the recipient's auth cookie into the purchaser's response (`class-reader-activation.php:996`). So a two-account checkout is handled by neither version — this one drops a real reader registration (no `REGISTRATION_METHOD`/`REGISTRATION_PAGE` meta, no ESP contact, no network `maybe_create_user`), the old one announced it as somebody else.

A bool that says *a checkout is happening* cannot answer *which account this checkout is expecting*. Keying the state on the identity the signal already has — the billing email, or the draft order id — lets `created_customer()` match rather than trust arrival order, and closes the same gap the comment at line 44 describes.

**To reproduce** (review env, `newspack_env_pr_897_store_api_reader`):
1. `wp eval` a capture hook on `newspack_registered_reader_via_woo`.
2. `do_action( 'woocommerce_checkout_process' );`
3. `wc_create_new_customer( 'first@example.test', 'first', wp_generate_password() );`
4. `wc_create_new_customer( 'second@example.test', 'second', wp_generate_password() );`
5. Observed: only `first@example.test` is announced. Setting `processing_checkout` back to `true` before step 4 announces both.

**Risk if not addressed:** Registration is the only path that pushes a gift recipient to the email service provider at creation: the subscription-side sync targets the subscription's customer, which for a gift is the purchaser, and the recurring contact sync only stages users who are logged in, which a gift recipient is not until they visit the site themselves. The same announcement is also what creates the reader on the other sites of a network. So for as long as this release is out, gift recipients accumulate as accounts that exist on one site, are absent from the publisher's audience, and never receive the welcome and newsletter flows the gift is meant to grant — noticed only when someone asks why gift recipients never hear from them.

**Consider:** Which account belongs to this checkout is safer decided against the account itself than against a process-wide flag — for example by comparing the created user's email to the email the checkout is for, so a second account created later in the request is judged on its own terms rather than on whether it arrived first. Whichever way a second announcement is admitted, note that it carries a side effect on the responding browser: the session subscriber writes the announced reader's email into that browser's reader cookie and rebinds its analytics client id, so a recipient announced during a purchaser's request must not be allowed to do that.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	/**
+	 * A Store API checkout (the wallet transport) announces the customer it
+	 * creates: the pre-creation signal arrives, then the account exists.
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_store_api_checkout_announces_created_customer() {
```
adekbadek, line 151: **Suggestion:** The two announce tests state one guarantee twice — `test_store_api_checkout_announces_created_customer` and `test_classic_checkout_announces_created_customer` (line 252) are the same test with a different hook name and email. A `@dataProvider` over `['woocommerce_checkout_process', 'woocommerce_store_api_checkout_update_customer_from_request']` collapses them into one method whose name can state the actual guarantee — both pipelines announce identically — which is what the change is for, and a stronger spec than a pair whose relationship the reader has to infer. `@runInSeparateProcess` applies per data set, so isolation survives.

`test_store_api_checkout_harvests_cart_campaign_metadata` (line 153) overlaps the same way with `test_second_checkout_does_not_inherit_campaign_metadata`: same fixture, same first half, and line 222 re-asserts line 167. Only the `referer` assertion on the first checkout is unique to it, and folding that into the batch test loses nothing.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
```diff
@@ -0,0 +1,273 @@
...
+		$this->stub_wc_with_cart();
+		do_action( 'woocommerce_store_api_checkout_update_customer_from_request', null, new WP_REST_Request( 'POST', '/wc/store/v1/checkout' ) );
+		do_action( 'woocommerce_created_customer', self::factory()->user->create( [ 'user_email' => 'checkout-account@example.test' ] ) );
+
+		// No second checkout signal: nothing announces this account.
+		do_action( 'woocommerce_created_customer', self::factory()->user->create( [ 'user_email' => 'unrelated-account@example.test' ] ) );
+
+		$this->assertCount( 1, $this->fired, 'Only the account its checkout created should be announced.' );
+		$this->assertSame( 'checkout-account@example.test', $this->fired[0]['email'] );
+	}
```
adekbadek, line 262: **Suggestion:** The lifecycle branch the code documents in prose has no test — `test_checkout_state_is_consumed_by_the_account_it_announces` pins the consumed path. The unconsumed one — a checkout signal that creates no account, leaving the flag for whatever comes next — is named outright in the production comment at `class-woo-user-registration.php:44` and asserted nowhere. That is the branch behind the pay-for-order finding, and it reproduces today.

The test that would have caught it: fire `woocommerce_store_api_checkout_update_customer_from_request`, create no customer, then fire `woocommerce_created_customer` for an unrelated account and assert nothing was announced. It fails against this branch, which is the point — on a hotfix, the one lifecycle the author documented is the one with no spec.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
```diff
@@ -0,0 +1,273 @@
...
+
+	/**
+	 * Classic checkout keeps announcing (regression control).
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_classic_checkout_announces_created_customer() {
```
adekbadek, line 276: **Suggestion:** The classic regression control would pass against the reverted code — `test_classic_checkout_announces_created_customer` is labelled a regression control, but it stubs an empty cart and asserts only `registration_method`, which `created_customer()` sets unconditionally at line 118 whatever the cart held. It therefore passes against `origin/release` unchanged, and would also pass if the new `self::$metadata = []` reset at line 65 broke the classic harvest — which is the one classic behaviour this diff could plausibly damage. Giving it the cart fixture from `test_store_api_checkout_harvests_cart_campaign_metadata` and asserting `newspack_popup_id` survives makes it bite. As it stands no test in the file covers the classic pipeline's campaign metadata at all.

plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
```diff
@@ -0,0 +1,273 @@
...
+ *
+ * The checkout-signal tests run in separate processes because the watcher's
+ * signal handler reads WC()->cart, so they stub WC() — which must not leak
+ * into the rest of the suite, whose WooCommerce-absent tests rely on
+ * function_exists( 'WC' ) staying false. Pattern follows
+ * my-account.php::test_woocommerce_delegation.
+ *
+ * @group data-events
+ */
+class Newspack_Test_Woo_User_Registration extends WP_UnitTestCase {
```
adekbadek: **Nit:** File and class name diverge from the data-events test convention — The repo keeps data-events tests in `tests/unit-tests/data-events/` — five files, including `class-listeners.php`, which covers the listener consuming this very action — and names their classes `Newspack_Test_Data_Events_*`. This one lands flat as `data-events-woo-user-registration.php` with class `Newspack_Test_Woo_User_Registration`. The recursive suite picks it up either way, so it costs nothing today; it just makes the next person grepping that directory miss it.

adekbadek (changes requested):
Two blockers and some suggestions. Also:

- `plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php:56` — **Suggestion:** Both docblocks understate what the methods now do — `checkout_process()`'s docblock still reads "During the checkout process, set the processing_checkout flag and store metadata" — it does not say the metadata is reset first, nor that the method now serves two unrelated WooCommerce pipelines, which is the whole change. `created_customer()`'s (line 95) does not say it consumes the flag, so the one-shot semantics introduced at line 109 are invisible to anyone reading the contract rather than the body. Related: the six lines of hook-contract prose at lines 40-45 carry facts the code cannot and earn their place, but they live in `init()`, where a reader arriving at `checkout_process()` from a stack trace will never see them. They would do more work on the handler's own docblock.
- `plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php:117` — **Suggestion:** This comment now states the opposite of what the code guarantees — "If a user is created later on the request by `woocommerce_created_customer()`, it will at least have this data" is no longer true, and it is untrue about exactly the invariant this change introduces: with the flag consumed at line 109, a user created later in the same request returns at the guard on line 101 and never reaches this line. It sits eight lines below a new comment saying the opposite, so the next reader gets two accounts of the same lifecycle on one screen. Either delete it or restate it as what the line does guarantee — that an announced account always carries a `registration_method`.
- `plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php:121` — **Suggestion:** Registration-page attribution never reaches the readers this change announces — `Modal_Checkout::is_modal_checkout()` cannot return true on a Store API request. It reads `$_REQUEST['modal_checkout']`, `$_REQUEST['post_data']` and, via `is_express_checkout()`, `$_POST['express_payment_type']` (`newspack-blocks/includes/class-modal-checkout.php:2313-2351`); a Store API checkout POSTs a JSON body to `/wp-json/wc/store/v1/checkout`, so all three are empty. So `current_page_url` is never derived from the harvested `referer`, `Reader_Activation::REGISTRATION_PAGE` is never written at line 130, and the ESP sync reads that field back empty (`reader-activation/sync/contact-metadata/class-registration.php:66`). Confirmed on the review env: after a real Store API checkout that created a reader, `np_reader_registration_page` on that user is `''` while `np_reader_registration_method` is `woocommerce`. Under the pre-9.2 form-encoded express transport these registrations did carry it, so for the wallet population this change rescues, the announcement comes back with strictly less than it used to. newspack-blocks already solved this shape for the same transport: `scrub_store_api_checkout_address` uses `is_modal_checkout_referer()` (`class-modal-checkout.php:1832`), reading the iframe URL out of `HTTP_REFERER`. Related, and probably the same follow-up: `checkout_order_processed()` is registered on `woocommerce_checkout_order_processed` (line 52), which the Store API never fires — its equivalent is `woocommerce_store_api_checkout_order_processed` — so the class is half-wired to the pipeline it now listens to. The PR body already names a `REGISTRATION_PAGE` limitation as pre-existing and out of scope. Worth being explicit that it applies to this path too, so it lands as a known follow-up rather than a second bug found later from the same ticket.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
wil-gerken: Fixed in `d6e023fa6` and `f51069c0e`.

The registration signal is now scoped to the billing email from the Store API checkout, so it can only announce the account created for that checkout. A later account created during payment — such as a gift recipient — won't match and won't inherit the original reader's campaign attribution.

Two notes from working through this:

* The billing email is the only usable identifier at signal time. Of the two identifiers suggested, the draft order isn't created until step #4 (`Checkout.php:575`), while the signal fires during step #3 (`:568 → :936`). The billing email has already been written to `WC()->customer` from `$request['billing_address']['email']` at `:915`.
* If an account is created, that email is present. `wc_create_new_customer()` rejects an empty or invalid email (`wc-user-functions.php:57-59`), and account creation at `:987` uses the same request value.

So there's no fallback needed here: if there's no billing email, no account is created and there's nothing to announce.

<details>
<summary>Detail — verification, and a second route that fires the same signal</summary>

**Measured on `nppm3147`, WooCommerce 10.9.4:**

| Scenario | Signal | Account | Announced |
| --- | --- | --- | --- |
| Guest, valid billing email | fires, email present | user created | `reader_registered` dispatched |
| Billing `email` key omitted | fires, email empty | refused, `registration-error-invalid-email` | nothing |
| Signed-in shopper | fires, email present | none created | nothing |

The boundary probe reads `processing_checkout=false` at creation time on the guest run, so the classic flag is demonstrably not what announced it — the email-keyed state is.

**A second emitter.** `woocommerce_store_api_checkout_update_customer_from_request` fires from two places in 10.9.4: `Checkout.php:936` and `CheckoutOrder.php:226`, the pay-for-an-existing-order route. `CheckoutOrder::process_customer()` (`:280-282`) only calls `sync_customer_data_with_order()` and creates no account, despite a docblock at `:128` saying it will. Email-keying covers that route for free: state is set, no `woocommerce_created_customer` follows, and the statics die with the request.

**Unit spec:** `test_store_api_checkout_does_not_announce_a_different_account` fails against the previous commit (announced 1, expected 0) and passes now. It also asserts the inverse, which is deliberate: a non-match leaves the state armed, because the buyer's own account arriving after an unrelated one is still that checkout's to announce.

Comparison is folded with `strtolower( trim( … ) )` on both sides, covered by `test_store_api_checkout_matches_the_account_regardless_of_email_case`. Making the fold a no-op fails that test.

**A signal with no email announces nothing** (`f51069c0e`). An earlier revision let that case fall back to announcing the first account created, as insurance in case wallet payloads arrived without a billing address. Driving the route settled that they do not need to: since account creation requires a valid address, a signal without one is followed by no account of its own, so anything arriving while it stands belongs to something else in the request. Announcing it would name the wrong reader. Covered by `test_store_api_signal_without_an_email_announces_nothing`.
</details>

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
wil-gerken: Fixed in `d6e023fa6` by removing flag consumption from the classic checkout path. Classic now announces every account created during checkout, as it did on `release`.

The reproduction is a gift subscription through classic checkout: the purchaser account is created first, then the recipient when the order reaches `processing`. On `e6817fece`, consuming the flag after the first account caused the recipient to be skipped entirely — no ESP contact, network account, or welcome flow.

Two notes from working through this:

* Gifting is loaded on a normal install. `class-wc-subscriptions-plugin.php:56` hooks `init_gifting` on `plugins_loaded`; the early return at `:305-310` applies only when the standalone gifting plugin is present. So the reproduction holds.
* This is the classic checkout path. The reproduction runs through `woocommerce_checkout_process`, so this was a regression introduced here rather than a Store API issue. Reverting the consumption on this path is the right fix.

Two-account checkouts are still imperfect on both paths, in different ways. On classic, the recipient is announced carrying the purchaser's `newspack_popup_id` and `referer`; that predates this change. On Store API, the purchaser is announced and the recipient is not — `release` announced neither, so announcing one and not the other is new here rather than inherited. I've held both out of this fix to keep the change to the two blockers, and haven't filed a follow-up, but say the word and I will.

<details>
<summary>Detail — the split, and why classic is not email-keyed</summary>

The two pipelines now hold different state, on purpose:

| Path | State | Lifecycle |
| --- | --- | --- |
| Classic (`woocommerce_checkout_process`) | plain bool | raised, never lowered — matches `release` |
| Store API (`…_update_customer_from_request`) | folded expected billing email | announces on match, stands down on match |

Classic is not email-keyed because its signal takes no arguments (`class-wc-checkout.php:1314`), so keying it would mean reading `$_POST['billing_email']` — a behaviour change on the pipeline this PR sets out not to touch, and one that would stop classic gift recipients being announced at all.

**On "no change to classic," precisely.** The classic handler is not byte-identical to `release`: it now calls a shared `harvest_cart_metadata()`, which carries the `self::$metadata = []` reset and the cart null-guard that `release` did not have. Both are no-ops for classic — one checkout per request, and `validate_cart_not_empty()` guarantees the cart — so there is no behavioural change, but the diff does show that handler moving.

**Unit spec:** `test_classic_checkout_announces_every_account_it_creates` fails against the previous commit (announced 1, expected 2) and passes now.
</details>

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
wil-gerken: I left these as separate tests after working through the fix, because the two checkout paths are no longer symmetrical.

Classic now raises a bool that it never consumes, while Store API keys the state to the buyer's billing email and clears it once that email matches. A `@dataProvider` over the two hook names would therefore need different fixtures and assertions for each case, which would add conditionals and make the combined test harder to follow.

There is still some overlap between `test_store_api_checkout_harvests_cart_campaign_metadata` and `test_second_checkout_does_not_inherit_campaign_metadata`. I left those separate as well because the first gives us a standalone assertion around the metadata itself, while the second covers metadata isolation across two checkouts in one process.

Happy to collapse that pair if you would still rather.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
wil-gerken: Added, and it went red first against the commit this thread was written on: 1 account was announced when 0 was expected.

`test_store_api_checkout_does_not_announce_a_different_account` covers the case from the suggestion: it fires the Store API signal without creating a customer, then creates an unrelated account and verifies that nothing is announced.

I added one more assertion to cover what happens next: if the buyer's own account is then created, it is announced. A non-match intentionally leaves the state active, so account creation order doesn't determine the outcome.

Writing this also exposed a gap in the existing Store API tests. Five tests were passing `null` as the customer, which meant they were exercising the empty-email branch rather than the normal Store API path. Those now pass a `WC_Customer` with the matching billing email.

One intentional `null` case remains: `test_store_api_signal_without_an_email_announces_nothing`, which verifies that no account is announced when the signal has no email to match against.

<details>
<summary>Detail — how the coverage was checked</summary>

Forcing the email comparison to never match fails 7 of 11 tests — every Store API test — and spares exactly the four that should not care: the two classic tests, the no-signal guard, and the named fallback. Before the five tests were converted, five of those seven survived that mutation.
</details>

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
wil-gerken: Rebuilt as suggested. `test_classic_checkout_announces_created_customer` now includes a cart fixture and verifies that `newspack_popup_id`, `prompt_title`, and `referer` survive alongside `registration_method`.

I also verified that the test catches the regression: removing the metadata harvest from the classic signal causes it to fail on the missing `newspack_popup_id`; restoring it passes.

You were right that we didn't have coverage for classic campaign metadata before. This test now provides it.

plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
wil-gerken: Done. Moved the test to `tests/unit-tests/data-events/class-woo-user-registration.php`, renamed the class to `Newspack_Test_Data_Events_Woo_User_Registration`, and updated the mock require path.

One note from checking this: the move doesn't make the test runnable directly by file path. `phpunit tests/unit-tests/data-events/class-listeners.php` behaves the same way for the existing test there — PHPUnit looks for a class matching the filename, and filenames diverge from class names throughout this directory. The suite is unaffected because it discovers tests by directory scan. The main benefit is keeping the file and class naming consistent with the rest of `data-events`.

wil-gerken:
Picking up the out-of-diff notes from the review body since they don't have inline threads.

**`:56` and `:95` docblocks — rewritten.** Both described behavior that changed when the checkout paths were split. `checkout_process()` and the new `store_api_checkout_process()` now document their own behavior, and the hook-specific detail that was in `init()` has moved to the relevant handler.

**`:117` comment — replaced.** You were right that it described the opposite of the actual guarantee. It now reflects that every announced account has a `registration_method`, while page and campaign metadata are best-effort. I also found and removed a stale reference to "the flag below" in the extracted metadata helper.

**`REGISTRATION_PAGE` on Store API — confirmed.** `Modal_Checkout::is_modal_checkout()` cannot return true during a Store API request, so `current_page_url` is not populated and the registration page is currently empty on this path. That's tracked in NPPM-3168, including `is_modal_checkout_referer()` as the likely approach. This PR still improves the Store API path by announcing those readers, but without the registration page metadata that the older form-encoded checkout carried.

**`checkout_order_processed()` half-wiring — confirmed, not changed here.** It listens to `woocommerce_checkout_order_processed` (`:52`), which Store API does not fire; the Store API equivalent is `woocommerce_store_api_checkout_order_processed` (`Checkout.php:675`). That code handles modal-checkout reload prevention rather than the registration announcement, so I left it out of this PR to keep the change scoped to the announcement path. Happy to fold it in if you would rather it went with this.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
adekbadek: Closed.

`Checkout::process_customer()` creates the account with `$request['billing_address']['email']` (`Checkout.php:850`), and `update_customer_from_request()` writes that same value onto the customer via `set_billing_email()` before firing the signal (`:718`, `:776`) — so the matched email is the one the account is created from, not a parallel copy that could drift.

Verified on the review env at this head:

| Probe | Result |
| --- | --- |
| Force the comparison to always succeed (`if ( true )` in `claim_account()`) | `test_store_api_checkout_does_not_announce_a_different_account` fails alone: announced 1, expected 0 |
| Full `newspack-plugin` suite | 2755 tests green |
| PHPCS, monorepo ruleset, both changed files | clean |

The logged-in path now stands down: the signal carries the buyer's own address, no account follows it, and an account created later in the request fails the comparison. The gift recipient going unannounced on this pipeline is the residual; `release` announced neither, so it is not a regression — see the other blocker thread for the follow-up.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
adekbadek: Closed. `claim_account()` returns `self::$processing_checkout` without clearing it, so a classic checkout announces every account it creates. Reinstating the consumption on that return fails `test_classic_checkout_announces_every_account_it_creates` alone (announced 1, expected 2).

On the follow-ups: yes, please file them, as two tickets rather than one.

- Classic announces the recipient carrying the purchaser's `newspack_popup_id` and `referer`, and the session subscriber then writes the recipient's email into the purchaser's browser cookie. Pre-existing, and the more damaging of the two.
- Store API announces the purchaser and not the recipient, so the recipient reaches no ESP and no network site. New here only in the sense that `release` announced neither.

Holding both out of a hotfix is right; they just shouldn't live only in this thread.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
adekbadek: Agreed, and the reason is the fix rather than the tests. When the suggestion was written both pipelines answered the same question the same way; they no longer do, so a data provider over the two hook names would have to branch on fixture and assertion inside the method body. That is worse than the duplication it removes.

Leaving the metadata pair separate too. Withdrawn.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
adekbadek: Closed. The `null`-customer find is the more valuable half: five tests were exercising the empty-email branch while reading as coverage of the normal one.

The converted tests bite. Folding the comparison to a no-op (`return (string) $email;` in `fold_email()`) fails `test_store_api_checkout_matches_the_account_regardless_of_email_case` alone; forcing the match to always succeed fails `test_store_api_checkout_does_not_announce_a_different_account` alone.

One nit for the next time this file is touched: `reset_watcher_state()` resets `processing_checkout` and `metadata` but not `store_api_checkout` or `store_api_expected_email`. Latent rather than live — every Store API test is `@runInSeparateProcess` and the single main-process test never raises them. Still worth closing, since the helper's docblock says it resets the watcher's request-scoped state and it now covers half of it.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
adekbadek: Closed, and it bites in both directions. Dropping `harvest_cart_metadata()` from `checkout_process()` fails this test alone on the missing `newspack_popup_id`; dropping the `self::$metadata = []` reset fails `test_second_checkout_does_not_inherit_campaign_metadata` alone. The classic harvest and the per-signal reset each have exactly one test standing over them.

plugins/newspack-plugin/tests/unit-tests/data-events-woo-user-registration.php
adekbadek: Closed.

On running it by file path: that is PHPUnit matching class name against filename, and it is how the whole of `data-events/` already behaves. Not something this move introduced, and not worth diverging from the directory to fix.

adekbadek (approved):
Both blockers are closed, each pinned by a test that fails when the fix is backed out. Five mutations against this head each failed exactly one test; the full `newspack-plugin` suite is green at 2755 tests, PHPCS clean on both files. Detail is in the six resolved threads.

Two items from the previous review's out-of-line notes have no thread to carry them, and both belong with the two gift-checkout follow-ups requested in the second blocker thread rather than in this PR:

- `REGISTRATION_PAGE` is never written on the Store API path. `Modal_Checkout::is_modal_checkout()` reads `$_REQUEST['modal_checkout']`, `$_REQUEST['post_data']` and `$_POST['express_payment_type']`, none of which a JSON POST to `/wc/store/v1/checkout` carries, so `current_page_url` is never derived from the harvested `referer` and the ESP sync reads the field back empty. `newspack-blocks` handles the same transport with `is_modal_checkout_referer()`.
- `checkout_order_processed()` is registered on `woocommerce_checkout_order_processed`, which the Store API never fires. Its equivalent is `woocommerce_store_api_checkout_order_processed`, so the class is still half-wired to the pipeline it now listens to.

plugins/newspack-plugin/includes/data-events/class-woo-user-registration.php
wil-gerken: I filed this as two follow-ups: NPPM-3232 for classic checkout and NPPM-3233 for block and wallet checkout.

NPPM-3232 covers both classic-path issues: the recipient inheriting the purchaser's page and campaign attribution, and the recipient announcement updating the purchaser's browser session. It also notes that the purchaser is not actually signed in as the recipient; the session check rejects that mismatch.

NPPM-3233 covers the block and wallet path, where the recipient is not announced at all.

Both issues note that the gifting behavior was verified from source, but a real gift purchase has not yet been run through either checkout path. Reproducing that flow is the first step on both.

plugins/newspack-plugin/tests/unit-tests/data-events/class-woo-user-registration.php
wil-gerken: This has been fixed here rather than leaving it for later: `reset_watcher_state()` now resets all four statics.

This was latent as described, so it doesn't change current behavior. I included it because the helper's docblock says it resets the watcher's request-scoped state, and leaving two statics behind could make a future same-process test inherit state from the previous one.

wil-gerken:
Thanks for that rapid review!

The two out-of-diff items are now tracked.

`REGISTRATION_PAGE` on this path remains in NPPM-3168, as noted earlier.

I also added the `checkout_order_processed()` Store API gap there rather than including it in this PR or filing a separate issue. It's another classic-only hook in the same class, so it fits the scope of NPPM-3168.

I left it as a verify-first item because we haven't yet confirmed whether a wallet checkout actually causes a visible reload. If it does, I'll split that into its own issue since it would be a separate reader-facing bug.

PR #900, opened 2026-08-18
fix(newspack-plugin): restore the major revision controls

adekbadek:
### All Submissions:

Revisions control lets a publisher mark a revision as major so the retention limit never deletes it. Those controls stopped loading in v5.0.0, when the enqueue moved to `dist/` during the `@wordpress/scripts` migration (Automattic/newspack-plugin#3181) with no webpack entry to build it. Both assets have 404'd since, so the flag can only be set in the database.

This adds the entry and moves the source under `src/` alongside every other entry, so the build produces what the enqueue already asks for. The button also moves beside "Restore This Revision" and picks up its size. One line beyond that, the AJAX callback used `jQuery.parseJSON`, which is gone in jQuery 4.

`e2e/tests/revisions.spec.ts` covers the round trip and fails when the asset is missing. Revisions control is off until configured, so `e2e-setup.sh` turns it on for the suite.

Closes NPPM-3155.

**Before** — the screen as it ships today, with no controls:

Revisions screen with no Mark as a major revision button

**After** — the same screen on WordPress 7.1-RC4 with the asset built:

Revisions screen on WordPress 7.1 with the major revision button beside Restore This Revision

1. Check out this branch and build newspack-plugin.
2. Run `wp option update newspack_revisions_control '{"active":true,"number":5,"min_age":"-1 week"}' --format=json`. This activates the feature and is required.
3. Publish a post. Edit and save it twice.
4. Open the post revisions screen at `/wp-admin/revision.php?revision=<latest revision ID>`. A "Mark as a major revision" button appears beside "Restore This Revision".
5. Open the browser console. No 404 appears for `dist/revisions-control.js` or `.css`.
6. Select "Mark as a major revision". The button becomes "Unmark as a major revision" and a "Major revision" label appears.
7. Reload the page. The revision is still marked, and its tickmark above the slider is colored.
8. Select "Unmark as a major revision", then reload. The revision is no longer marked.

To run the new e2e spec against a local env: `cd e2e && SITE_URL=https://<env>.test USE_SETUP=true npx playwright test --project="setup-vanilla"`, then the same command without `USE_SETUP` and with `-g "Mark a revision as major"`.

### Other information:

* [x] Have you written new tests for your changes, as applicable? e2e coverage added; verified red without the built asset and green with it.
* [x] Have you successfully run tests with your changes locally? The 17 existing revisions-control PHP tests pass, and the vanilla e2e suite is green on desktop and mobile.

<details>
<summary>For AI</summary>

- Entry: `plugins/newspack-plugin/webpack.config.js`, `'revisions-control': resolveSource( 'src', 'revisions-control', 'index' )`. Output is `dist/revisions-control.{js,css,-rtl.css,asset.php}`.
- Enqueue is `Newspack\Major_Revisions::enqueue_scripts()` in `includes/revisions-control/class-major-revisions.php`; unchanged, and `Newspack::asset_version( 'revisions-control' )` now resolves a real `.asset.php` instead of falling back to the plugin version.
- The script patches the Backbone revisions app (`wp.revisions.view.Tickmarks`, `wp.revisions.view.MetaTo`) on `revision.php`. WordPress 7.1's paginated timeline and autosave labels are in the in-editor React UI, so nothing here is affected. Checked on 7.1-RC4.
- The in-editor revisions UI reads revisions over REST, which does not run `wp_prepare_revision_for_js`, so a major revision has no indicator there and cannot be marked. Out of scope; worth a follow-up if the classic screen is ever retired.
- `.distignore` excludes `/src/**/*.js` and `/src/**/*.scss`, so the moved source no longer ships in the zip; `dist/` does.
- The tickmark loop maps `model.revisions.models[i]` onto `.revisions-tickmarks > div[i]`, and core renders one fewer tickmark than revisions, so the newest revision has no tickmark to color. Pre-existing, not touched here, and the reason the e2e spec asserts on the label rather than the tickmark.

</details>


leogermani (approved):
All good

PR #901, opened 2026-08-18
fix(newsletters): persist list appearance and cut load time (NPPM-3140)

thomasguillot:
### All Submissions:

Two problems a publisher reported on the newsletters admin screens.

**Appearance settings did not survive a reload.** Only items per page was stored, so density, visible columns and their Title, Media and Description toggles, column order, widths, sort and layout had to be set again on every visit. All of it now persists per user, on all four list screens. Page, search and filters stay out: that is a query, not a configuration.

**The lists were slow.** They asked for `_links`, the only way to make `_embed` expand anything, and core answers that by computing target hints for every row's `self` link, re-resolving the whole REST route map per row. Author and term names now come from dedicated REST fields, and "All" fetches in far fewer requests. On a local site with 613 newsletters, 8 requests and 3,381 KB become 2 requests and 934 KB, and the table is ready in 3.2s instead of 5.9s.

Loading is left to DataViews. Our overlay sat on top of it, so two indicators animated at once. The first paint, which has no list behind it, gains a label.

Newsletters list first paint: spinner with a Fetching newsletters label

Part of NPPM-3140.

1. On Newsletters, set Density to Compact in View options, toggle on Categories, and move a column.
2. Reload. Density, the column and the new order are all still applied.
3. Repeat on Advertising, Advertisers and Layouts. On Layouts, also switch to Table and change Preview size.
4. Watch the network panel on a reload. No request carries `_embed` or `_links`.
5. Set items per page to All. The whole list arrives in one or two requests.
6. Confirm Author still shows a name and avatar, and Categories and Tags still show names.
7. Open Quick Edit on a newsletter with categories or tags. The pickers are pre-filled.
8. Open Quick Edit on an ad with an advertiser, placement and category. Change only the status, save, reload. All three survived.
9. Throttle the network and reload. The first paint shows a spinner labelled "Fetching newsletters…".
10. On Layouts, switch to Table and reload. The table comes back as a table, with no preview thumbnails in the rows.
11. Logged out, request `/wp/v2/newspack_nl_cpt?per_page=1000`. It returns a 400; `per_page=100` still works.

### Other information:

The `per_page` ceiling is raised to 1000 on the four collections these screens read, and only those; `/wp/v2/posts` still rejects anything above 100. Anything above core's 100 also requires `edit_posts`, since the newsletters CPT is public and an unauthenticated caller could otherwise ask one request to render ten times as many rows as core allows.

Step 8 is the one worth doing. The ads Quick Edit pickers seed entirely from the new terms field, so a field that ever went missing would show three empty pickers; each taxonomy now travels only when it changed, which is what stops a status-only save clearing the rest.

One behaviour change worth flagging for networks: `usermeta` is network-global, so the stored preference key now carries the site prefix and a single list configuration can no longer be shared across every site on a network. Anyone who had already set items per page will have that one value reset once. It was the only thing persisted before this, so that is the whole scope.

Does not close NPPM-3140. The per-status counts and small-screen column widths are still open there.

### DIFF
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
@@ -10,6 +10,7 @@ namespace Newspack\Newsletters\Admin;
 use Newspack_Newsletters;
 use Newspack_Newsletters\Ads;
 use Newspack_Newsletters_Layouts;
+use WP_Error;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -24,9 +25,20 @@ defined( 'ABSPATH' ) || exit;
  * count, not the query, is what a publisher waits on. Raising the
  * ceiling lets the client walk the same collection in far fewer trips.
  *
+ * The cost this trades against is memory: `WP_Query` selects whole rows,
+ * so a page holds that many `WP_Post` objects, `post_content` included,
+ * whatever `_fields` the client asks for. Measured at roughly 3.3KB per
+ * row before content, so the ceiling itself is a few megabytes and the
+ * rest scales with how long a site's newsletters are.
+ *
  * The walk still chunks rather than asking for everything at once, so a
  * site with thousands of newsletters can't turn one request into a
  * multi-megabyte response.
+ *
+ * The raise is gated on the capability the list screens already require:
+ * the newsletters CPT is public, so without a gate any anonymous caller
+ * could ask one request to render ten times as many rows through the
+ * full `the_content` chain as core's ceiling allows.
  *
  * This is not the per-page value the controls offer — that stays at 100
  * (`Admin_Shell_Preferences::PER_PAGE_MAX`).
@@ -38,6 +50,11 @@ class Admin_Shell_Collection_Params {
 	 * together.
 	 */
 	const MAX_PER_PAGE = 1000;
+
+	/**
+	 * Core's own ceiling, and the most an unprivileged caller may ask for.
+	 */
+	const CORE_MAX_PER_PAGE = 100;
 
 	/**
 	 * Boot hooks.
@@ -80,8 +97,44 @@ class Admin_Shell_Collection_Params {
 	 */
 	public static function raise_per_page_cap( $params ) {
 		if ( isset( $params['per_page'] ) && is_array( $params['per_page'] ) ) {
-			$params['per_page']['maximum'] = self::MAX_PER_PAGE;
+			$params['per_page']['maximum']           = self::MAX_PER_PAGE;
+			$params['per_page']['validate_callback'] = [ __CLASS__, 'validate_per_page' ];
 		}
 		return $params;
+	}
+
+	/**
+	 * Hold unprivileged callers to core's ceiling.
+	 *
+	 * The capability check can't live in `raise_per_page_cap()`: collection
+	 * params are filtered on `rest_api_init`, and resolving the current
+	 * user that early pre-empts `rest_authentication_errors`, which breaks
+	 * application-password auth. A validate callback runs at dispatch,
+	 * after authentication has settled.
+	 *
+	 * @param mixed            $value   Submitted value.
+	 * @param \WP_REST_Request $request Incoming request.
+	 * @param string           $param   Parameter name.
+	 * @return true|WP_Error
+	 */
+	public static function validate_per_page( $value, $request, $param ) {
+		$valid = rest_validate_request_arg( $value, $request, $param );
+		if ( is_wp_error( $valid ) ) {
+			return $valid;
+		}
+
+		if ( (int) $value > self::CORE_MAX_PER_PAGE && ! current_user_can( 'edit_posts' ) ) {
+			return new WP_Error(
+				'rest_invalid_param',
+				sprintf(
+					/* translators: %d: the highest per_page value available to this user. */
+					__( 'per_page must be %d or fewer.', 'newspack-newsletters' ),
+					self::CORE_MAX_PER_PAGE
+				),
+				[ 'status' => 400 ]
+			);
+		}
+
+		return true;
 	}
 }
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * Admin shell — REST collection parameters.
+ *
+ * @package Newspack_Newsletters
+ */
+
+namespace Newspack\Newsletters\Admin;
+
+use Newspack_Newsletters;
+use Newspack_Newsletters\Ads;
+use Newspack_Newsletters_Layouts;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Raises the `per_page` ceiling on the collections the admin-shell list
+ * screens read.
+ *
+ * Core caps `per_page` at 100, which is what makes the "All" option fan
+ * out into one request per 100 rows. At that size the response itself is
+ * cheap (roughly 0.13s of server time per 100 newsletters) while each
+ * extra round trip costs the full WordPress bootstrap, so the request
+ * count, not the query, is what a publisher waits on. Raising the
+ * ceiling lets the client walk the same collection in far fewer trips.
+ *
+ * The walk still chunks rather than asking for everything at once, so a
+ * site with thousands of newsletters can't turn one request into a
+ * multi-megabyte response.
+ *
+ * This is not the per-page value the controls offer — that stays at 100
+ * (`Admin_Shell_Preferences::PER_PAGE_MAX`).
+ */
+class Admin_Shell_Collection_Params {
+	/**
+	 * The `per_page` ceiling these collections accept. Mirrored client
+	 * side by `FETCH_ALL_CHUNK_SIZE` in `utils/per-page.js`; change both
+	 * together.
+	 */
+	const MAX_PER_PAGE = 1000;
+
+	/**
+	 * Boot hooks.
+	 *
+	 * Registered on `rest_api_init` rather than at load time because the
+	 * CPTs and taxonomies these filters name are registered on `init`.
+	 */
+	public static function init(): void {
+		add_action( 'rest_api_init', [ __CLASS__, 'register_filters' ] );
+	}
+
+	/**
+	 * Hook the cap onto every collection the list screens read.
+	 */
+	public static function register_filters(): void {
+		foreach ( self::get_collections() as $rest_base ) {
+			add_filter( 'rest_' . $rest_base . '_collection_params', [ __CLASS__, 'raise_per_page_cap' ] );
+		}
+	}
+
+	/**
+	 * The post types and taxonomies whose collections back a list screen.
+	 *
+	 * @return array<string>
+	 */
+	public static function get_collections(): array {
+		return [
+			Newspack_Newsletters::NEWSPACK_NEWSLETTERS_CPT,
+			Ads::CPT,
+			Ads::ADVERTISER_TAX,
+			Newspack_Newsletters_Layouts::NEWSPACK_NEWSLETTERS_LAYOUT_CPT,
+		];
+	}
+
+	/**
+	 * Raise the collection's `per_page` maximum.
+	 *
+	 * @param array $params Collection parameters.
+	 * @return array
+	 */
+	public static function raise_per_page_cap( $params ) {
+		if ( isset( $params['per_page'] ) && is_array( $params['per_page'] ) ) {
+			$params['per_page']['maximum'] = self::MAX_PER_PAGE;
+		}
+		return $params;
+	}
+}
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
@@ -70,8 +70,8 @@ class Admin_Shell_Collection_Params {
 	 * Hook the cap onto every collection the list screens read.
 	 */
 	public static function register_filters(): void {
-		foreach ( self::get_collections() as $rest_base ) {
-			add_filter( 'rest_' . $rest_base . '_collection_params', [ __CLASS__, 'raise_per_page_cap' ] );
+		foreach ( self::get_collections() as $collection ) {
+			add_filter( 'rest_' . $collection . '_collection_params', [ __CLASS__, 'raise_per_page_cap' ] );
 		}
 	}
 
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -119,17 +119,18 @@ class Admin_Shell_Preferences {
 	 * @return array
 	 */
 	private static function get_prefs_schema(): array {
+		// Sanitised rather than rejected: DataViews owns this shape, and
+		// schema validation fails the whole request on the first unknown
+		// key, so one added property would silently stop every appearance
+		// setting persisting. `sanitize_column_style()` allowlists the same
+		// four keys on write and on read.
 		$column_style = [
-			'type'                 => 'object',
-			'additionalProperties' => false,
-			'properties'           => [
+			'type'       => 'object',
+			'properties' => [
 				'width'    => [ 'type' => [ 'string', 'integer' ] ],
 				'minWidth' => [ 'type' => [ 'string', 'integer' ] ],
 				'maxWidth' => [ 'type' => [ 'string', 'integer' ] ],
-				'align'    => [
-					'type' => 'string',
-					'enum' => self::ALIGNMENTS,
-				],
+				'align'    => [ 'type' => 'string' ],
 			],
 		];
 
@@ -423,7 +424,15 @@ class Admin_Shell_Preferences {
 	 * @return string
 	 */
 	public static function get_user_meta_key( string $screen_key ): string {
-		return self::USER_META_KEY_PREFIX . $screen_key;
+		global $wpdb;
+
+		// `usermeta` is network-global, so on multisite a bare key would
+		// share one list configuration across every site in the network.
+		// Single-site installs keep the unprefixed key so preferences saved
+		// before this survive.
+		$prefix = is_multisite() ? $wpdb->get_blog_prefix() : '';
+
+		return $prefix . self::USER_META_KEY_PREFIX . $screen_key;
 	}
 
 	/**
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -22,9 +22,9 @@ defined( 'ABSPATH' ) || exit;
  * each other via a shared read-modify-write.
  *
  * The stored shape is the presentation half of the DataViews `view`:
- * layout type, sort, visible fields (in their display order), and the
- * per-layout settings (density, column widths, grid preview size), plus
- * items per page. Query state — page, search and filters — is
+ * layout type, sort, visible fields (in their display order and with
+ * their toggles), and the per-layout settings (density, column widths,
+ * grid preview size), plus items per page. Query state — page, search and filters — is
  * deliberately not stored, matching classic Screen Options: a saved
  * list configuration, not a saved query.
  *
@@ -126,17 +126,15 @@ class Admin_Shell_Preferences {
 	 * @return array
 	 */
 	private static function get_prefs_schema(): array {
-		// Sanitised rather than rejected: DataViews owns this shape, and
-		// schema validation fails the whole request on the first unknown
-		// key, so one added property would silently stop every appearance
-		// setting persisting. `sanitize_column_style()` allowlists the same
-		// four keys on write and on read.
+		// Sanitised rather than rejected: schema validation fails the whole
+		// request on the first unknown key, so one property added upstream
+		// would silently stop every appearance setting persisting.
 		$column_style = [
 			'type'       => 'object',
 			'properties' => [
-				'width'    => [ 'type' => [ 'string', 'integer' ] ],
-				'minWidth' => [ 'type' => [ 'string', 'integer' ] ],
-				'maxWidth' => [ 'type' => [ 'string', 'integer' ] ],
+				'width'    => [ 'type' => [ 'string', 'number' ] ],
+				'minWidth' => [ 'type' => [ 'string', 'number' ] ],
+				'maxWidth' => [ 'type' => [ 'string', 'number' ] ],
 				'align'    => [ 'type' => 'string' ],
 			],
 		];
@@ -445,10 +443,9 @@ class Admin_Shell_Preferences {
 	public static function get_user_meta_key( string $screen_key ): string {
 		global $wpdb;
 
-		// `usermeta` is network-global, so on multisite a bare key would
-		// share one list configuration across every site in the network.
-		// Single-site installs keep the unprefixed key so preferences saved
-		// before this survive.
+		// `usermeta` is network-global, so a bare key would share one list
+		// configuration across a whole network. Single-site installs keep the
+		// unprefixed key so preferences saved before this survive.
 		$prefix = is_multisite() ? $wpdb->get_blog_prefix() : '';
 
 		return $prefix . self::USER_META_KEY_PREFIX . $screen_key;
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -51,6 +51,13 @@ class Admin_Shell_Preferences {
 	const PER_PAGE_ALL = -1;
 	const PER_PAGE_MAX = 100;
 
+	/**
+	 * DataViews renders these alongside the field checkboxes in View
+	 * options, so they persist with them. Mirrored client side by
+	 * `VISIBILITY_FLAGS` in `use-persisted-view.js`.
+	 */
+	const VISIBILITY_FLAGS = [ 'showTitle', 'showMedia', 'showDescription' ];
+
 	const LAYOUT_TYPES = [ 'table', 'grid', 'list' ];
 	const DENSITIES    = [ 'compact', 'balanced', 'comfortable' ];
 	const ALIGNMENTS   = [ 'start', 'center', 'end' ];
@@ -134,7 +141,7 @@ class Admin_Shell_Preferences {
 			],
 		];
 
-		return [
+		$schema = [
 			'perPage'    => [ 'type' => 'integer' ],
 			'type'       => [
 				'type' => 'string',
@@ -177,6 +184,12 @@ class Admin_Shell_Preferences {
 				],
 			],
 		];
+
+		foreach ( self::VISIBILITY_FLAGS as $flag ) {
+			$schema[ $flag ] = [ 'type' => 'boolean' ];
+		}
+
+		return $schema;
 	}
 
 	/**
@@ -292,6 +305,12 @@ class Admin_Shell_Preferences {
 			}
 			// An empty array is meaningful — every column hidden.
 			$clean['fields'] = $fields;
+		}
+
+		foreach ( self::VISIBILITY_FLAGS as $flag ) {
+			if ( isset( $prefs[ $flag ] ) && is_bool( $prefs[ $flag ] ) ) {
+				$clean[ $flag ] = $prefs[ $flag ];
+			}
 		}
 
 		$sort = isset( $prefs['sort'] ) && is_array( $prefs['sort'] ) ? self::sanitize_sort( $prefs['sort'] ) : null;
--- a/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
+++ b/plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
@@ -13,13 +13,26 @@ use WP_REST_Server;
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Per-user view preferences for the admin-shell list screens (items per
- * page), stored in user meta so they follow the user across browsers.
- * Values are bootstrapped into the localized `newspackNewslettersAdmin`
- * global (see `Admin_Shell_Assets`), so the REST route is write-only in
- * practice. Each screen gets its own user-meta key so a save only ever
- * touches that screen's value — two tabs saving different screens can't
- * clobber each other via a shared read-modify-write.
+ * Per-user view preferences for the admin-shell list screens, stored in
+ * user meta so they follow the user across browsers. Values are
+ * bootstrapped into the localized `newspackNewslettersAdmin` global (see
+ * `Admin_Shell_Assets`), so the REST route is write-only in practice.
+ * Each screen gets its own user-meta key so a save only ever touches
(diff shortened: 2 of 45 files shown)


thomasguillot:
Went back over this branch end to end before putting it up for review, and pushed the fixes that came out of it.

What I put under test first was the ads Quick Edit seeding path, since that is the part I flagged as highest risk in the description, and the caching behind the two new REST fields. On the local site with 613 newsletters the terms field adds no extra queries at either 10 or 100 rows, so the author and term lookups really are reading the caches `WP_Query` primes rather than going per row. Per-row memory is roughly 3.3KB before `post_content`. No shared contract files are touched, so nothing outside this plugin is affected.

**Release risk: LOW.** Three things needed fixing, all now fixed.

**The raised `per_page` ceiling applied to callers who were not logged in.** The newsletters CPT is registered `public => true, show_in_rest => true`, and `WP_REST_Posts_Controller::get_items_permissions_check()` gates only the `edit` context. So `GET /wp/v2/newspack_nl_cpt?per_page=1000` with `_fields` omitted would render a thousand newsletters through the full `the_content` chain instead of a hundred. Anything above core's 100 now requires `edit_posts`, checked in a `validate_callback` so the capability resolves after authentication rather than during `rest_api_init`.

**A restored table layout kept the grid-only `mediaField`.** On Layouts a stored `type: 'table'` merged over a default carrying `mediaField: 'preview'`, and the cleanup that strips it only ran on a user-driven change, never at mount. The table renders the media field in its primary column, so a restored table mounted a `BlockPreview` iframe in every row. That reconciliation is now shared between the initial view and later changes.

**Ads Quick Edit sent all three taxonomies on every save.** They seed entirely from the new terms field, so if that field were ever absent, a status-only save would clear the ad's advertiser, placement and categories. That matters more than it first looks: `Ads::get_ads()` skips the category check for an ad that has none, so a cleared list makes the ad eligible for every newsletter rather than none. Each taxonomy now travels only when it was actually changed.

Alongside those, preference saves can no longer trip schema validation and silently stop persisting, the unload flush always goes out on a keepalive request rather than relying on one that navigation can tear down, and the Title, Media and Description toggles now persist with the rest of the appearance state.

**Severity: nit** in `includes/admin/class-admin-shell-preferences.php`: the column-style schema declares the width keys as `[ 'string', 'integer' ]`, so a fractional width would fail validation and take the rest of the payload with it. Nothing in DataViews 16 writes `layout.styles`, so it cannot happen today.

788 PHP tests and 302 JS tests pass, the build is clean, and PHPCS, ESLint and Stylelint are clean on the touched files. No browser pass since these changes.

thomasguillot:
Final pass on top of the above.

**Fractional column widths.** The column-style schema declared the width keys as `[ 'string', 'integer' ]`, so a fractional value would have failed validation and taken items-per-page, layout type, fields, sort and density down with it, then been re-sent on every later change. They are `[ 'string', 'number' ]` now, which matches the float branch `sanitize_column_style()` already carried. Latent rather than live: nothing in DataViews 16 writes `layout.styles`.

**Comment sweep.** Went through every comment this branch adds and cut it back to the ones that explain why rather than what, 139 lines down to 91, none in SCSS. What survives is mostly the cross-file couplings that have to change together (`VISIBILITY_FLAGS`, the two per-page ceilings, `ALIGNMENTS`) and the handful of decisions that are not obvious from the code: why `_links` is left out, why the ads save is gated per taxonomy, why `speak()` owns the loading announcement instead of a live region.

**Description.** Brought it up to date with the capability gate on `per_page`, the Title, Media and Description toggles now persisting, and two more test steps: restoring the Layouts table, and confirming the ceiling rejects a logged-out request.

788 PHP tests and 302 JS tests pass, build clean, PHPCS and ESLint clean on the touched files.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -64,19 +108,88 @@ public static function register_routes(): void {
 							'type'                 => 'object',
 							'required'             => true,
 							'additionalProperties' => false,
```
adekbadek: **Suggestion:** `additionalProperties => false` on `prefs` is the failure mode this class argues against — The comment at `:130` states the policy — schema validation fails the whole request on the first unknown key, so one property added upstream would silently stop every appearance setting persisting — and applies it to `$column_style`. `prefs` itself and `layout` at `:167` are still strict, so a single unrecognised key 400s the save and every appearance setting stops persisting, which is the publisher-visible symptom NPPM-3140 opens with. Not reachable today: `toPrefs()` in `use-persisted-view.js:96` builds the payload key by key. It becomes reachable the first time a screen forwards a fuller `view`, or a DataViews bump adds a key ahead of the PHP schema. Consider dropping `additionalProperties => false` on both and letting `sanitize_prefs()`, which already whitelists key by key, do the rejecting; `test_rejects_query_state` stays green either way, since `search` / `page` / `filters` sanitise to an empty array and hit `invalid_preference_error()`.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -121,25 +253,202 @@ public static function get_preferences(): array {
...
 	 * @return string
 	 */
 	public static function get_user_meta_key( string $screen_key ): string {
-		return self::USER_META_KEY_PREFIX . $screen_key;
+		global $wpdb;
+
+		// `usermeta` is network-global, so a bare key would share one list
+		// configuration across a whole network. Single-site installs keep the
+		// unprefixed key so preferences saved before this survive.
+		$prefix = is_multisite() ? $wpdb->get_blog_prefix() : '';
```
adekbadek, line 459: **Suggestion:** The multisite key prefix orphans saved preferences, and the test covers the other branch — On a network, `get_blog_prefix()` now prefixes the meta key, so a preference saved before this lands is unreadable and stays in `usermeta`; that includes the network's main site, where the prefix is `wp_` rather than empty. The blast radius is one reset items-per-page value per user per site, which is why this reads as acceptable rather than a fault — but it is a behaviour change with no note in the PR body. The test that would catch it, `test_the_meta_key_is_scoped_per_site_only_on_multisite` at `tests/test-admin-shell-preferences.php:427`, cannot: the multisite branch is unreachable in a single-site PHPUnit run, and the branch that does execute asserts `USER_META_KEY_PREFIX . 'newsletters-list'`, which is what `get_user_meta_key()` returned before this change — so it passes with the change reverted. Either run that file under `WP_TESTS_MULTISITE` as a selectable group, or drop the test and note the one-time reset in the PR body.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.test.js
```diff
@@ -51,15 +47,9 @@ describe( 'ItemsPerPage', () => {
...
-		const { rerender } = render( <ItemsPerPage value={ PER_PAGE_ALL } onChange={ jest.fn() } progress={ { loaded: 100, total: 1200 } } /> );
-		expect( speak ).toHaveBeenCalledTimes( 1 );
-
-		rerender( <ItemsPerPage value={ PER_PAGE_ALL } onChange={ jest.fn() } progress={ { loaded: 200, total: 1200 } } /> );
-		expect( speak ).toHaveBeenCalledTimes( 1 );
-		expect( screen.getByText( 'Loading 200 of 1,200…' ) ).toBeInTheDocument();
-
-		rerender( <ItemsPerPage value={ PER_PAGE_ALL } onChange={ jest.fn() } progress={ null } /> );
-		expect( speak ).toHaveBeenCalledTimes( 2 );
+	it( 'renders no loading indicator of its own', () => {
```
adekbadek, line 53: **Suggestion:** `renders no loading indicator of its own` passes against the pre-change component — Rendered with no `progress` prop, the previous `ItemsPerPage` took `progress = null` by default, so its `{ progress && … }` branch was falsy, `wasLoadingAllRef.current` already matched `isLoadingAll` and no `speak()` fired, and `usePopoverSlot` returned null — the container was empty and `.newspack-newsletters-fetch-all-progress` absent, which is exactly what this test asserts. It cannot go red if the overlay comes back. If the guarantee worth pinning is "a fetch-all walk adds no second indicator", passing `progress={ { loaded: 100, total: 1200 } }` makes the assertion bite on a prop the component no longer honours; otherwise the SCSS and component deletions already cover the removal and the test can go.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
```diff
@@ -1,77 +1,354 @@
...
+const isPreviewSize = value => Number.isInteger( value ) && value >= 0 && value <= MAX_PREVIEW_SIZE;
+
+const isNonEmptyString = value => typeof value === 'string' && value.length > 0;
+
+// Mirrors `Admin_Shell_Preferences::ALIGNMENTS`.
+const ALIGNMENTS = [ 'start', 'center', 'end' ];
+
+// Picked key by key rather than forwarded whole: DataViews owns this
+// shape, and one unknown key would fail validation for the whole payload.
+function toColumnStyles( styles ) {
```
adekbadek, line 66: **Suggestion:** Column-width persistence has no writer in the bundled DataViews — `toColumnStyles`, the `$column_style` schema branch, `sanitize_column_style()`, `MAX_STYLED_COLUMNS` / `MAX_WIDTH_LENGTH` and three tests all persist `view.layout.styles`, but the resolved `@wordpress/dataviews@16.0.0` only reads it: `components/dataviews-layouts/table/index.mjs:142` and `:374` destructure `width`/`minWidth`/`maxWidth`/`align` onto the cell styles, and the package's only `styles:` assignment is in `dataform-layouts/normalize-form.mjs`, which is DataForm rather than the list view. There is no resize handle. The screens pass `defaultLayouts={ { table: {} } }`, so in practice `view.layout.styles` is always undefined. Density and `previewSize` are genuinely user-settable, so the feature works — this is the width slice only. Worth either saying in the docblock that it is forward-compat for a DataViews version that writes widths, or dropping it. Related: the module docblock at `:12` justifies the debounce with "column resizing streams changes", which is the one interaction that cannot happen; `previewSize` is a RangeControl and does stream, so the debounce is right but anchored on the wrong control.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
```diff
@@ -1,77 +1,354 @@
...
+ * @param {Object} view DataViews view state.
+ * @return {Object} Preferences payload.
+ */
+function toPrefs( view = {} ) {
+	const prefs = {};
+
+	if ( isValidPerPage( view.perPage ) ) {
+		prefs.perPage = view.perPage;
+	}
+	if ( isNonEmptyString( view.type ) ) {
```
adekbadek: **Suggestion:** `type` and `fields` are the two payload keys without a client-side mirror — `SORT_DIRECTIONS`, `DENSITIES`, `ALIGNMENTS` and `isPreviewSize` all mirror the server enums so `toPrefs` cannot emit a value the schema rejects, but `type` is passed through on `isNonEmptyString` alone and `fields` is not capped at `MAX_FIELDS` at `:109`. Because `additionalProperties => false` and `enum` fail the whole request, either would 400 the payload and stop every appearance setting persisting — the failure mode `get_prefs_schema()`'s own comment warns about. Not reachable today: `layoutTypes` is `Object.keys( DEFAULT_LAYOUTS )` on every screen, so `view.type` is only ever `table` or `grid`, and no screen defines close to 50 fields. `toPrefs` already runs where `layoutTypes` is in scope, so gating on it closes the asymmetry cheaply.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
```diff
@@ -1,77 +1,354 @@
...
+				data: { screen: screenKey, prefs: JSON.parse( payload ) },
+				...( keepalive ? { keepalive: true } : {} ),
+				...( controller ? { signal: controller.signal } : {} ),
 			} )
 				.then( () => {
-					lastSavedRef.current = perPage;
+					lastSavedRef.current = payload;
 				} )
 				.catch( () => {
 					failed = true;
```
adekbadek, line 332: **Suggestion:** A rejected save leaves no trace — If the server starts rejecting the payload — a DataViews upgrade adding a view property the route schema does not know, the drift the schema comment in `Admin_Shell_Preferences` worries about — persistence stops with no notice, no console entry and no server-side record: just a 400 in the network tab and a retry that also fails quietly. That is the same publisher-visible symptom this PR fixes, with nothing to diagnose it by.

The shell already has the right surface for this: `src/admin-shell/notices.js` exports a snackbar `createErrorNotice`, and `app.js` renders the `SnackbarList`. Two constraints shape how it should be used here. The save fires on every view change, so if the schema has drifted every save fails and a naive notice becomes a snackbar on each density toggle and column move; latching it behind a ref so it fires once per screen per session avoids that. And the unload flush is a keepalive request, so skip the notice when `unloadingRef.current` is set — there is no page left to show it on.

Worth noting the catch takes no argument (`.catch( () => { failed = true; } )`), so the error is discarded before anything could report it — capturing it is the prerequisite either way.

**Risk if not addressed:** If a dependency bump lands unattended, the regression reaches every site at once and stays for a release cycle. With no client notice and no record of the rejected payload, there is no way afterwards to tell when it started or which key caused it.

**Consider:** Surface the exhausted retry to the person it affects rather than to the console: one latched error snackbar per screen, plus the response code kept for diagnosis. A publisher who sees "couldn't save your view settings" reports something actionable; a silent failure reproduces the complaint that opened this ticket.

plugins/newspack-newsletters/src/admin-shell/utils/per-page.js
```diff
@@ -1,16 +1,26 @@
...
-// Mirrors `Admin_Shell_Preferences::PER_PAGE_MAX` (the REST cap the
-// server validates stored preferences against) — change both together.
-export const FETCH_ALL_CHUNK_SIZE = 100;
+// Mirrors `Admin_Shell_Preferences::PER_PAGE_MAX`, which validates
+// stored preferences — change both together.
+export const MAX_SELECTABLE_PER_PAGE = 100;
+
+// Mirrors `Admin_Shell_Collection_Params::MAX_PER_PAGE`, the ceiling
+// those collections accept — change both together.
+export const FETCH_ALL_CHUNK_SIZE = 1000;
```
adekbadek, line 22: **Suggestion:** One chunk size for four screens with very different row weights — `FETCH_ALL_CHUNK_SIZE` is shared by all four screens through `utils/build-query.js:47`, but `layouts-list/use-layouts-data.js:25` is the one `_fields` set that ships `content.raw`, so an All walk there asks for 1000 layouts of full block markup in one response and one request's peak memory, where the other three strip content. `FETCH_ALL_CONCURRENCY = 3` means three such pages can be in flight. `Admin_Shell_Collection_Params`' docblock says the walk "still chunks rather than asking for everything at once, so a site with thousands of newsletters can't turn one request into a multi-megabyte response" — true for the other three `_fields` sets, not for layouts. Total bytes transferred are unchanged either way; this is peak per-response size and peak memory. Letting a screen override the chunk, or simply using a smaller one for layouts, bounds the worst request without giving up the round-trip savings elsewhere.

plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	public static function init(): void {
+		add_action( 'rest_api_init', [ __CLASS__, 'register_filters' ] );
+	}
+
+	/**
+	 * Hook the cap onto every collection the list screens read.
+	 */
+	public static function register_filters(): void {
+		foreach ( self::get_collections() as $rest_base ) {
```
adekbadek: **Nit:** `$rest_base` holds a post-type name, not a REST base — `get_collections()` returns post-type and taxonomy names, which is what `rest_{$post_type}_collection_params` and `rest_{$taxonomy}_collection_params` are keyed on — a REST base would be wrong there, so only the name is misleading. The feature's own test makes the distinction: `rest_base_for()` at `tests/test-admin-shell-collection-params.php:192` exists because `newspack_nl_ad_placement` overrides `rest_base`. A reader who trusts the variable name would reach for `$object->rest_base` here and all four filters would silently stop matching. `$object_name` or `$collection` reads truer.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -121,25 +253,202 @@ public static function get_preferences(): array {
...
+	private static function sanitize_column_style( array $style ): ?array {
+		$clean = [];
+
+		foreach ( [ 'width', 'minWidth', 'maxWidth' ] as $key ) {
+			if ( ! isset( $style[ $key ] ) ) {
+				continue;
+			}
+			$value = $style[ $key ];
+			if ( is_int( $value ) || is_float( $value ) ) {
+				$clean[ $key ] = (int) $value;
```
adekbadek: **Nit:** Fractional widths pass the schema, then get truncated on store — The schema was widened to `[ 'string', 'number' ]` at `:135-137` so a fractional width stops failing validation, but this branch casts floats with `(int)`, so 217.5 comes back as 217 on the next load. Impact is sub-pixel and, as the finding on `use-persisted-view.js:49` notes, nothing currently writes widths at all — so this is a consistency question rather than a bug: either keep the float (`$value + 0`) or the schema could have stayed `integer`. As written the two halves of the fix disagree about what a fractional width means.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
```diff
@@ -121,25 +253,202 @@ public static function get_preferences(): array {
...
 	 * @param string $screen_key Screen identifier.
 	 * @return string
 	 */
 	public static function get_user_meta_key( string $screen_key ): string {
-		return self::USER_META_KEY_PREFIX . $screen_key;
+		global $wpdb;
+
+		// `usermeta` is network-global, so a bare key would share one list
+		// configuration across a whole network. Single-site installs keep the
+		// unprefixed key so preferences saved before this survive.
```
adekbadek: **Nit:** Comment dates itself against the change — "Single-site installs keep the unprefixed key so preferences saved before this survive" reads as a note about this branch; once merged, "before this" has no referent. The constraint underneath is worth keeping — state it as the rule, that single-site keys stay unprefixed to remain compatible with existing stored values. The same sentence is repeated verbatim in the test docblock at `tests/test-admin-shell-preferences.php:425`.

plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
```diff
@@ -22,10 +22,19 @@
...
 	use Rest_Status_Field;
+	use Rest_Terms_Field;
 	use Status_Filter_Builder;
 
 	const STATUS_QUERY_PARAM = 'newspack_newsletters_ad_status';
 
+	/**
+	 * Taxonomies behind the Advertiser, Ad placement and Categories
+	 * columns. Quick Edit seeds all three from the same payload and
+	 * sends them back on every save, so the list asks for this field
```
adekbadek: **Nit:** Docblock describes the pre-fix save behaviour — "Quick Edit … sends them back on every save" stopped being true in a945cffb05 — the panel now sends only the taxonomies the editor changed. The half that is still true is the reason the field stays unconditional: Quick Edit has no other source for the names, so a hidden column would leave its pickers empty. The same stale clause survives in the `test_terms_field_covers_every_quick_edit_taxonomy` docblock in `tests/test-ads-list-rest.php`.

plugins/newspack-newsletters/includes/admin/trait-rest-terms-field.php
```diff
@@ -0,0 +1,127 @@
...
+	 * @param array            $post_array  Prepared post, as passed by the controller.
+	 * @param string           $field_name  Registered field name.
+	 * @param \WP_REST_Request $request     Incoming request.
+	 * @param string           $object_type Object type the field is registered on.
+	 * @return array<string, array<array{id: int, name: string}>>
+	 */
+	public static function rest_get_terms( $post_array, $field_name = '', $request = null, $object_type = '' ): array {
+		return self::get_terms_payload(
+			isset( $post_array['id'] ) ? (int) $post_array['id'] : 0,
+			self::$terms_field_taxonomies[ $object_type . ':' . $field_name ] ?? []
```
adekbadek: **Nit:** The taxonomy registry is written under `$cpt` and read under `$object_type` — `register_terms_field()` stores taxonomies at `$cpt . ':' . $field_name` at `:55` while `rest_get_terms()` looks them up at `$object_type . ':' . $field_name`, relying on `WP_REST_Posts_Controller::get_object_type()` returning the post-type slug. That holds for both current callers, but a miss returns `array_fill_keys( [], [] )`, so the taxonomy columns and the ads Quick Edit pickers would come back blank rather than raising anything. Keying on `$field_name` alone, or falling back to the single registered entry when the composite key misses, would make the failure loud — worth doing only if a third caller is expected.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
```diff
@@ -8,22 +8,19 @@
...
  * into the popover when it opens (anchored on the same class names the
- * package styles against). This component itself is mounted in the
- * DataViews `header` slot; while a fetch-all walk runs it overlays a
- * centered spinner + progress message on the list (the popover is
- * closed during a walk, and the header offers no room for it).
+ * package styles against). It is mounted in the DataViews `header` slot.
+ *
+ * Loading is left entirely to DataViews, which pulses the list and sets
+ * `aria-busy` while a fetch is in flight. A second, custom indicator on
+ * top of that read as two competing animations.
```
adekbadek: **Nit:** Docblock describes the code that was removed — "A second, custom indicator on top of that read as two competing animations" is past tense about an overlay no longer in the file. The preceding sentence already carries why the component renders no loading state ("Loading is left entirely to DataViews, which pulses the list and sets `aria-busy`").

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/index.js
```diff
@@ -31,19 +30,26 @@ const DEFAULT_VIEW = {
 	filters: [],
 	titleField: 'title',
 	fields: [ 'status', 'date', 'send_date', 'send_list', 'author', 'public_page' ],
-	...getInitialView(),
 };
 
 const DEFAULT_LAYOUTS = { table: {} };
 
+// `urlPatch` is read at module scope so a forwarded legacy link seeds
```
adekbadek: **Nit:** The `urlPatch` comment gives a reason that isn't the mechanism — `usePersistedView` reads `urlPatch` only inside the lazy `useState` initializer at `hooks/use-persisted-view.js:251`, so it is applied once whether the options object is built at module scope or inline per render. Module scope avoids re-parsing `window.location.search` on each render; it is not what stops the patch being re-applied. Either state the parse cost or drop the second clause.

plugins/newspack-newsletters/tests/test-admin-shell-collection-params.php
```diff
@@ -0,0 +1,203 @@
...
+
+		$this->assertSame( 400, rest_do_request( $request )->get_status() );
+	}
+
+	/**
+	 * The newsletters CPT is public, so without a capability gate the
+	 * raised ceiling would let any anonymous caller ask one request to
+	 * render ten times as many rows as core allows.
+	 */
+	public function test_a_logged_out_caller_is_held_to_the_core_cap() {
```
adekbadek: **Nit:** Three tests assert one guarantee — This, `test_a_subscriber_is_held_to_the_core_cap` at `:142` and `test_the_gate_covers_every_raised_collection` at `:157` all assert that an unprivileged caller asking above 100 gets a 400, differing only in who the caller is and which collection is hit. A data provider over (role, collection) says it once and reads as the spec. The two neighbours that look redundant are not: `test_a_logged_out_caller_can_still_reach_the_core_cap` at `:129` catches a `>=` / `>` slip in `validate_per_page`, and `test_unrelated_collections_keep_the_core_cap` at `:104` catches the raise being hooked onto `rest_post_collection_params`.

plugins/newspack-newsletters/tests/test-admin-shell-preferences.php
```diff
@@ -160,4 +160,280 @@ public function test_save_does_not_touch_other_screens_meta_key() {
...
+		$sanitized = Admin_Shell_Preferences::sanitize_prefs( [ 'fields' => [ 'status', 'author', 'status' ] ] );
+
+		$this->assertSame( [ 'status', 'author' ], $sanitized['fields'] );
+	}
+
+	/**
+	 * Preferences written by an older version stay readable: the extra
+	 * keys simply arrive absent.
+	 */
+	public function test_reads_back_a_per_page_only_value_saved_by_an_earlier_version() {
```
adekbadek: **Nit:** `test_reads_back_a_per_page_only_value_saved_by_an_earlier_version` restates an existing assertion — `test_get_preferences_sanitizes_stored_meta` at `:133` already writes `[ 'perPage' => 25 ]` straight into user meta and asserts it reads back unchanged, which is the same guarantee under a different name. The one thing this test names that the other does not — that a payload written before the new keys existed still round-trips — is carried by the older test's fixture already.

adekbadek (approved):
Release risk: LOW — no blockers; every claim in the test plan reproduced on the review env, and the REST surface is correctly gated (subscriber and anonymous callers get a 400 above per_page 100, and neither new field reaches an anonymous response).

Approving. The comments below are all optional: a few latent edges around the preferences round-trip, one memory bound worth stating, and some comment and test cleanups.

One note on scope: an ad whose categories are cleared becomes eligible for every newsletter rather than none, because `Ads::get_ads()` skips the category check when an ad has no categories. That predates this branch and this PR narrows the exposure by sending each taxonomy only when it changed, so it is context rather than a request.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Dropped on `prefs`, `sort` and `layout` in 6c0e772e4. `sanitize_prefs()` was already whitelisting key by key, so an unrecognised key is now dropped rather than fatal. 214e4d066 adds the test that was missing: a known key alongside an unknown one, asserting the rest still lands.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Retargeted in 6c0e772e4, then made to assert the composition directly in 214e4d066 so the prefixed key is the expectation wherever the suite runs. The one-time reset is now noted in the description.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.test.js
thomasguillot: 72ece0d08 passes `progress` explicitly, so it asserts against the prop the overlay used to render from rather than passing either way.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
thomasguillot: Kept as forward-compat and said so in the docblock in 6c0e772e4, along with re-anchoring the debounce rationale on the preview-size slider, which is the control that actually streams.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
thomasguillot: Both closed. `fields` is capped at `MAX_FIELDS` and `type` gated on the screen's `layoutTypes` in 6c0e772e4; 2ce352762 also intersects it with a mirrored `LAYOUT_TYPES`, which covers the ungated default branch too.

plugins/newspack-newsletters/src/admin-shell/hooks/use-persisted-view.js
thomasguillot: Added in 6c0e772e4: the error is captured, one latched snackbar per screen, silent on the keepalive unload flush. 2ce352762 then clears the latch on a later success, so a recovered blip doesn't leave a false claim standing or spend the latch, and stops the retry re-sending a payload the server rejected.

plugins/newspack-newsletters/src/admin-shell/utils/per-page.js
thomasguillot: Layouts walks in 200-row chunks from 82911aa8b. 61dba7c0b then removed the second call site entirely: `useCollectionData` reads `per_page` back off the path it was given, so the request size and the page ceiling can no longer drift apart.

plugins/newspack-newsletters/includes/admin/class-admin-shell-collection-params.php
thomasguillot: Renamed to `$collection` in 72ece0d08, in the test's loop as well.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Float kept in 6c0e772e4. That turned out to let `INF` through too, which `wp_json_encode` cannot encode, so it would have printed the localized global as a syntax error and stopped the admin shell booting for that user. `is_finite()` guards it in 214e4d066, and a7a2c9d60 falls back to an empty payload at the localize site so no future value can do the same.

plugins/newspack-newsletters/includes/admin/class-admin-shell-preferences.php
thomasguillot: Reworded to state the rule rather than date itself, in 6c0e772e4, test docblock included.

plugins/newspack-newsletters/includes/admin/class-ads-list-rest.php
thomasguillot: Both reworded in 72ece0d08.

plugins/newspack-newsletters/includes/admin/trait-rest-terms-field.php
thomasguillot: 769065cbb binds the taxonomies into the `get_callback` closure, so there is no lookup left to miss. Keying on `$field_name` alone was not an option: both CPTs register the same field name over different taxonomies.

plugins/newspack-newsletters/src/admin-shell/components/items-per-page/index.js
thomasguillot: Dropped in 72ece0d08.

plugins/newspack-newsletters/src/admin-shell/screens/newsletters-list/index.js
thomasguillot: Reworded to the parse cost in 72ece0d08.

plugins/newspack-newsletters/tests/test-admin-shell-collection-params.php
thomasguillot: Consolidated into a data provider over the two caller kinds in 72ece0d08. b63ba3471 then put `CORE_MAX_PER_PAGE + 1` back alongside the raised ceiling, which the consolidation had dropped and which is what pins where the gate sits.

plugins/newspack-newsletters/tests/test-admin-shell-preferences.php
thomasguillot: Dropped in 6c0e772e4.

PR #907, opened 2026-08-18
fix(editor): stop console warnings from Button, cards and Collections

chickenn00dle:
### All Submissions:

Clears five console errors and one layout defect found while testing WordPress 7.1 RC. **None is a 7.1 regression** — every one reproduces identically on 7.0.4 — but they are noisy in the editor and wizards.

* `Button` forwarded its own `loading` prop to the DOM, which React rejects as a non-boolean attribute. It now maps to `@wordpress/components`' equivalent `isBusy`, covering all six `loading={...}` call sites.
* `Button` was not a `forwardRef`, so core's Tooltip could not attach a ref to it. The same change fixes this.
* `CoreCard` passed `hideLabelFromVision` to `ToggleControl`, which does not support it, so it landed on the `<input>`. The card's own CSS already hides that label visually, so the prop was redundant.
* The Collections block declared `editorScript` in `block.json`, so WordPress registered the raw unbuilt ES module as a classic script and the browser threw a `SyntaxError`. The block's editor code already ships in the bundled `dist/blocks.js`.
* The wizard store registered without a guard, so a second bundle loading it logged `Store "newspack/wizards" is already registered`.
* The corrections modal's Delete button sat flush against the description field (measured 0px gap).

Found while testing NEWS-2696.

1. Enable the Collections module: `wp newspack optional-module activate collections`.
2. Open any post in the block editor with the console open.
3. Confirm no `Cannot use import statement outside a module` error, and no `src/blocks/collections/index.js` script tag in the page source.
4. Confirm the Collections block still inserts and renders.
5. Go to Audience > Access Control and open a gate for editing.
6. Confirm no `hideLabelFromVision` and no `Function components cannot be given refs` warnings.
7. Confirm the gate card's toggle still works and is still labelled for screen readers.
8. Open Audience > Configuration > Emails, click Settings, and save.
9. Confirm no `non-boolean attribute 'loading'` warning, and the Save button still shows its busy state.
10. Edit a post with a correction, open Manage corrections, and confirm the Delete button has a gap above it and still deletes.
11. Rebuild newspack-blocks and newspack-popups from this branch too, since each bundles its own copy of the wizard store. In the editor, the `Store "newspack/wizards" is already registered.` error is gone.
12. A production build stays silent. A development build now logs a deliberate `Wizard store: "newspack/wizards" is already registered` note naming the collision, so run `n build` rather than `n watch` for step 11.

### Other information:

`packages/components` is consumed by nine units, so the `loading` prop is kept as this component's public API and translated internally rather than renamed. The wizard-store guard only takes effect once each consuming plugin is rebuilt, since each bundles its own copy.

Verified on WordPress 7.1-RC4: the Collections script tag and both Access Control warnings are gone from a real browser session, and the corrections Delete button measures an 8px gap where it previously measured 0px. Re-verified in an isolated env with newspack-plugin, newspack-blocks and newspack-popups all built from this branch, which is what the wizard-store step needs. `main` is merged in, and both suites pass on the result: 414 tests in `packages/components` and 1258 in newspack-plugin.


packages/components/src/button/index.tsx
```diff
@@ -67,8 +67,13 @@ const Button = ( { href, loading = undefined, onClick, ...otherProps }: Props )
...
 		otherProps.disabled = true;
 	}
+	// `loading` is this component's own prop name; @wordpress/components' Button
+	// expresses the same state as `isBusy`. Forwarding `loading` verbatim would
+	// land it on the DOM node, which React rejects as a non-boolean attribute.
 	// @ts-expect-error - @wordpress/components' Button can only have either href or onClick, not both.
-	return <BaseComponent loading={ loading ? true : undefined } { ...otherProps } />;
-};
+	return <BaseComponent ref={ ref } isBusy={ loading ? true : undefined } { ...otherProps } />;
+} );
```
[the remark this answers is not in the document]
chickenn00dle: Added in de04807 — `packages/components/src/button/index.test.js` now covers the `loading` → `isBusy` mapping (busy state rendered, no `loading` attribute on the DOM node, no React warning logged) and the ref forwarding the same change introduced. Both assertions fail against the pre-fix component; the whole 224-test suite passes with it.

packages/components/src/button/index.test.js
```diff
@@ -141,3 +146,40 @@ describe( 'Button with a javascript: href alone', () => {
...
+// which React rejects with an unknown-prop warning.
+describe( 'Button loading prop', () => {
+	const button = () => screen.getByRole( 'button', { name: 'Save' } );
+
+	it( 'renders the core busy state without putting `loading` on the DOM node', () => {
+		const error = jest.spyOn( console, 'error' ).mockImplementation( () => {} );
+		render( <Button loading>Save</Button> );
+
+		expect( button() ).toHaveClass( 'is-busy' );
+		expect( button() ).not.toHaveAttribute( 'loading' );
```
adekbadek: **Suggestion:** The `loading` attribute assertion passes on the pre-change code — `expect( button() ).not.toHaveAttribute( 'loading' )` is not what makes this test bite: React 18 does not write a boolean-valued unknown attribute to the DOM — `validateProperty` warns and returns "remove" for that case — so the attribute is absent under a revert too. The assertion that actually catches a revert is `expect( error ).not.toHaveBeenCalled()` on the next line, and it is worth anchoring to the message it guards: `expect( error ).not.toHaveBeenCalledWith( expect.stringContaining( 'non-boolean attribute' ), ... )`. That names the guarantee, and it matters beyond documentation — React dedupes that warning per property name for the lifetime of the module registry, so an unanchored "no errors at all" guard silently stops biting the moment any earlier test in the file renders `<Button loading>`. The same argument applies to line 182, where the message is `Function components cannot be given refs`. The `toHaveClass( 'is-busy' )` assertion above is sound and does bite.

packages/components/src/wizard/store/index.js
```diff
@@ -160,4 +160,18 @@ const store = createReduxStore( WIZARD_STORE_NAMESPACE, {
...
+/**
+ * Register the wizard store, unless it already exists.
+ *
+ * This package is bundled separately into each consuming plugin, so more than one
+ * bundle can call this on the same page (e.g. the block editor loads assets from
+ * both newspack-blocks and newspack-manager). @wordpress/data keeps a single
+ * registry on `wp.data`, so the second call is a no-op that logs
+ * `Store "newspack/wizards" is already registered.` — guard it instead.
+ */
+export default () => {
```
adekbadek, line 177: **Suggestion:** Guard removes the only signal that two bundle versions are live — The guard is correct and behaviour-preserving — `registerStoreInstance` already returned the existing store rather than replacing it, so the first bundle won before this change too. What it removes is the sole indication that two copies of `packages/components` at different versions are on the same page. That collision is not hypothetical: the package declares no `sideEffects`, so any bundle importing the barrel pulls in `wizard/index.js`, whose module scope calls `registerStore()`; `plugins/newspack-newsletters/webpack.config.js:61` documents this exact case and stubs the module out to avoid it. Since consumers rebuild and release on their own schedules, an older bundle can win the race, and the wizard store's selector set grows over time — `getWizardData` is newer than `getWizardAPIData`, and there are around forty call sites reading it.

**Consider:** log a `console.warn` from the guard when it fires, wrapped in a `'production' !== process.env.NODE_ENV` check, matching the pattern already used in this package at `button/index.tsx:43-47`. Developers keep the breadcrumb and publishers still get a clean console.

packages/components/src/button/index.test.js
```diff
@@ -141,3 +146,40 @@ describe( 'Button with a javascript: href alone', () => {
 		expect( link() ).toHaveAttribute( 'href', '/wp-admin/admin.php?page=next' );
 	} );
 } );
+
+// `loading` is this component's own prop name; the core Button expresses the same
```
adekbadek: **Nit:** Comment duplicates the one in the implementation — These three lines are the comment added at `packages/components/src/button/index.tsx:70-72`, reworded. The implementation is the right home for it, since that is where the constraint binds, and the `describe` plus the test name on line 156 already state the guarantee here. Two copies of the same reasoning drift apart on the next edit.

packages/components/src/button/index.test.js
```diff
@@ -141,3 +146,40 @@ describe( 'Button with a javascript: href alone', () => {
...
+	const button = () => screen.getByRole( 'button', { name: 'Save' } );
+
+	it( 'renders the core busy state without putting `loading` on the DOM node', () => {
+		const error = jest.spyOn( console, 'error' ).mockImplementation( () => {} );
+		render( <Button loading>Save</Button> );
+
+		expect( button() ).toHaveClass( 'is-busy' );
+		expect( button() ).not.toHaveAttribute( 'loading' );
+		expect( error ).not.toHaveBeenCalled();
+		error.mockRestore();
```
adekbadek: **Nit:** A failing assertion leaves console.error mocked for the rest of the run — `error.mockRestore()` is the last statement in the test rather than in a `finally`, and the shared Jest config sets no `restoreMocks`, so if the assertion on line 162 fails the spy is never restored and every subsequent test in the file runs with `console.error` swallowed — turning one real failure into a quiet run. Line 183 has the same shape, and the pre-existing `warn.mockRestore()` on line 141 does too, so this is a convention the file is already inconsistent about rather than something the diff introduces. A single `afterEach( () => jest.restoreAllMocks() )` covers all three.

packages/components/src/button/index.test.js
```diff
@@ -141,3 +146,40 @@ describe( 'Button with a javascript: href alone', () => {
...
+	} );
+
+	it( 'leaves the button idle without the prop', () => {
```
adekbadek: **Nit:** Idle-state test passes with the change reverted — `leaves the button idle without the prop` asserts that core's Button renders no `is-busy` class when nothing asks for it, which was equally true before the change — the old code forwarded `loading={ undefined }`, which also produced no class. It specs `@wordpress/components`' default rather than the `loading`→`isBusy` mapping. It does read as a deliberate control for the assertion above it; if that contrast is worth keeping, `rerender`ing from `<Button />` to `<Button loading />` inside the first test states it in one place rather than as a second case that can never go red.

packages/components/src/button/index.test.js
```diff
@@ -141,3 +146,40 @@ describe( 'Button with a javascript: href alone', () => {
...
+		render( <Button>Save</Button> );
+
+		expect( button() ).not.toHaveClass( 'is-busy' );
+	} );
+} );
+
+// A ref on a plain function component is dropped with a warning, so the wizard
```
adekbadek: **Nit:** Ref comment names a consumer that does not exist — "the wizard screens that measure or focus a button need it forwarded" describes a use case with no call site — nothing in the monorepo passes a ref to this Button. The real driver is core's Button wrapping itself in a Tooltip when it has a `label` and no children, which needs a ref on the child. Verified call sites that do that with this Button: `plugins/newspack-plugin/src/wizards/newspack/views/settings/experimental-tools/configure-view.tsx:261` (`icon` plus `label`, no children) and `plugins/newspack-plugin/src/wizards/audience/components/segment-group/index.js:67` (`showTooltip`). Naming one of those gives the comment a fact the code cannot carry; as written it sends the next reader looking for a consumer that is not there.

adekbadek (approved):
Release risk: LOW — no blockers; the `packages/components` contract change is compatible across its nine consumers (the public `loading` prop is unchanged and `forwardRef` is additive); browser-verified on WordPress 7.1. Approving with comments. 

packages/components/src/button/index.test.js
chickenn00dle: Confirmed, and the check was worth running: with the fix reverted and the `is-busy` assertion removed, `not.toHaveAttribute( 'loading' )` still passes and only the console spy goes red, with `Received `true` for a non-boolean attribute `loading``.

Both assertions now name their message through a small `messagesFrom()` helper, so the dedupe hazard you describe cannot quietly disarm them. I dropped the attribute assertion rather than keeping it alongside, since it never had teeth. Verified each message assertion fails on its own by removing the assertions above it.

bf0f27a4d

packages/components/src/wizard/store/index.js
chickenn00dle: Added the dev-only warn in dfada81a2, matching the `button/index.tsx` pattern. It names the collision and says the first bundle to load wins.

One thing worth recording: `select()` on an unregistered store is silent in `@wordpress/data@10.48.0` (it falls through to `parent?.select()`), so the guard adds no output of its own on the normal path.

Testing this needs more rebuilding than the PR body claimed. In a default install three plugins register the store in the editor, not two: newspack-plugin, newspack-blocks and newspack-popups. Updated the test steps.

packages/components/src/button/index.test.js
chickenn00dle: Removed. The implementation comment is the one that stays.

packages/components/src/button/index.test.js
chickenn00dle: Replaced all three with a single `afterEach( () => jest.restoreAllMocks() )`.

That change caused the lint failure on this branch just now: the two `javascript:` href tests bound a spy they never asserted on, so dropping their `mockRestore()` left `warn` unused. They no longer bind it.

packages/components/src/button/index.test.js
chickenn00dle: Folded into the first test with a `rerender` from `<Button>` to `<Button loading>`, so the contrast is stated once and the idle case cannot pass on a revert. Suite is 223 tests now.

packages/components/src/button/index.test.js
chickenn00dle: Rewrote the comment. You are right that no call site passes a ref: the only `ref=` on this Button in the monorepo is the test I added.

It now names the Tooltip as the driver and points at the two call sites you found, both of which I confirmed: `experimental-tools/configure-view.tsx:261` (`icon` plus `label`, no children) and `audience/components/segment-group/index.js:67` (`showTooltip`).

PR #912, opened 2026-08-19
fix(accessibility): stop duplicate statement pages appearing (NEWS-2948)

thomasguillot:
### All Submissions:

Reading the accessibility statement page created one. `get_page()` made a draft whenever no page ID was stored, and the classic theme footer calls it on every front-end render, so ordinary visits and 404s minted pages. The ID also lived in a theme mod, which every theme and style variation keeps its own copy of, so each switch started the count again. Sites ended up with a pile of identical drafts nobody asked for.

**Reading is now a pure lookup.** Creation happens only on the explicit POST, and hands back the existing page instead of a second one. The ID moved to a site-wide option, adopted on upgrade from whatever a theme mod was holding, including a theme that has since been deleted. That old pointer stays readable until the upgrade runs, and again if the plugin is rolled back, so a published statement keeps its footer link and its content-gate exemption throughout. Both the read and the write only see the active theme, so that cover applies to a site still on the theme that stored the ID; one that had already switched gets its link back when the upgrade runs. Widening the fallback would put a scan of every theme's mods on every front-end render, which is a worse trade than the gap it closes.

The wizard now separates a page that was deleted from one that was never created. That first state was unreachable before, because loading the screen created the page.

Advanced Settings: the new empty state, reading You do not have an accessibility statement page yet

Closes NEWS-2948.

1. Activate a classic Newspack theme. This puts the footer call in play.
2. Open Newspack > Settings > Advanced Settings and create the page. Pages shows one draft.
3. Switch to another style variation, for example `newspack-scott`.
4. Load a front-end page, then a URL that 404s. Pages still shows one draft.
5. Load the front end while logged out. Pages still shows one draft.
6. Reopen Advanced Settings. The card still points at the same page.
7. Publish the page. The footer link appears in the site footer.
8. Trash the page, then reopen Advanced Settings. The notice says it was trashed.
9. Click Create Page. Exactly one new draft appears.
10. Delete both `newspack_accessibility_statement_*` options, and permanently delete every accessibility statement page the steps above made, trash included. Each theme still holds a mod pointing at one of them, and those are only followed to a page that still exists, so deleting the pages is what leaves the site with no pointer at all.
11. Reload Advanced Settings. The notice now says no page exists yet, rather than claiming one was trashed.
12. For the upgrade path, delete those options again and set the `accessibility_statement_page_id` theme mod to a published page's ID.
13. Load the front end logged out. The footer link is there before any admin visit.
14. Log in as a subscriber and open `/wp-admin/profile.php`. Neither option is written: the upgrade waits for someone who could create the page, and the footer link keeps working meanwhile.

### Other information:

Sites that already collected duplicates keep them. The upgrade adopts one page, preferring a published one, and leaves the rest alone rather than deleting publisher content on update, so support should expect to tell affected publishers to bin the extras.

The `force_create` parameter is gone from the POST endpoint. It existed only to bypass the guard this PR makes reliable.

Where the GET finds no page it now answers with a `reason` of `none` or `missing`, rather than putting either word in `status`. `status` stays a post status, so a client cannot mistake a post status registered under one of those names for an absent page. The wizard is the only consumer, and the classic theme reads `get_page()`, which still returns `false`.

### DIFF
--- a/plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
+++ b/plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
@@ -10,40 +10,131 @@ namespace Newspack;
 /**
  * Accessibility Statement Page class.
  */
-class Accessibility_Statement_Page {
+final class Accessibility_Statement_Page {
+
+	/**
+	 * Option holding the ID of the page the site uses.
+	 */
+	const OPTION_NAME = 'newspack_accessibility_statement_page_id';
+
+	/**
+	 * Where the page ID used to live. Theme mods are per theme, so every theme
+	 * and style variation kept its own, and a site lost the page each time it
+	 * switched. Still read until the upgrade has run, and still written on
+	 * create, so a plugin rolled back to before this change finds the page
+	 * where it looks for it instead of making a second one.
+	 */
+	const LEGACY_THEME_MOD = 'accessibility_statement_page_id';
+
+	/**
+	 * Marks the upgrade as done so the theme scan runs once per site, including
+	 * on sites that turn out to have nothing to migrate. Autoloaded, so the
+	 * debug reset clears it together with the pointer rather than leaving a
+	 * site that believes it has already migrated and has nothing to show.
+	 */
+	const MIGRATION_FLAG = 'newspack_accessibility_statement_migrated';
+
+	/**
+	 * The slug reserved for the page.
+	 */
+	const PAGE_SLUG = 'accessibility-statement';
 
 	/**
 	 * Add hooks.
 	 */
-	public static function init() {
+	public static function init(): void {
 		// Register REST routes.
-		add_action( 'rest_api_init', [ __CLASS__, 'register_rest_routes' ] );
+		add_action( 'rest_api_init', [ __CLASS__, 'register_rest_routes' ], 10 );
+		add_action( 'admin_init', [ __CLASS__, 'migrate_legacy_theme_mod' ], 10 );
 		// Add post status to accessibility statement page.
 		add_filter( 'display_post_states', [ __CLASS__, 'post_status' ], 10, 2 );
 	}
 
 	/**
-	 * Create an accessibility statement page.
+	 * Get the ID of the page the site uses, whether or not it still exists.
 	 *
-	 * @param bool $force_create Whether to force creation of a new page even if theme_mod exists.
-	 * @return array|null|WP_Error The created page data, null if it has been trashed or deleted, or error.
+	 * Falls back to the legacy theme mod when no usable page is stored. That
+	 * covers the window between the plugin updating and the next admin request,
+	 * which on an auto-updating site can be days, and it covers a page an older
+	 * version created while the plugin was rolled back.
+	 *
+	 * @return int The stored page ID, which may point at a page that has since
+	 *             been trashed or deleted, or 0 if no pointer was ever stored.
 	 */
-	public static function create_page( $force_create = false ) {
-		// Check if page ID is stored in theme_mods.
-		$page_id = get_theme_mod( 'accessibility_statement_page_id' );
-		if ( $page_id && ! $force_create ) {
-			$page = get_post( $page_id );
-			// Check if page exists and is either published or draft.
-			if ( $page && 'page' === $page->post_type && in_array( get_post_status( $page->ID ), [ 'publish', 'draft' ] ) ) {
-				$page_data = [
-					'editUrl' => get_edit_post_link( $page->ID, 'raw' ),
-					'status'  => get_post_status( $page->ID ),
-					'pageUrl' => get_permalink( $page->ID ),
-				];
-				return $page_data;
-			}
-			// If page doesn't exist or is in trash, return null but keep the ID stored.
+	public static function get_page_id(): int {
+		$page_id = (int) get_option( self::OPTION_NAME, 0 );
+		if ( $page_id && self::usable_page( $page_id ) ) {
+			return $page_id;
+		}
+
+		$legacy_id = (int) get_theme_mod( self::LEGACY_THEME_MOD );
+		if ( $legacy_id && self::usable_page( $legacy_id ) ) {
+			return $legacy_id;
+		}
+
+		return $page_id;
+	}
+
+	/**
+	 * Get the stored page, if it is still there to be used.
+	 *
+	 * @return \WP_Post|null The page, or null if none is stored, it has been
+	 *                       deleted, or it sits in the trash.
+	 */
+	private static function get_stored_page(): ?\WP_Post {
+		$page_id = self::get_page_id();
+		if ( ! $page_id ) {
 			return null;
+		}
+
+		return self::usable_page( $page_id );
+	}
+
+	/**
+	 * Resolve a page ID to a page the site can still use.
+	 *
+	 * @param int $page_id The page ID.
+	 * @return \WP_Post|null The page, or null if it is gone, trashed, or not a page.
+	 */
+	private static function usable_page( int $page_id ): ?\WP_Post {
+		$page = get_post( $page_id );
+		if ( ! $page || 'page' !== $page->post_type || 'trash' === $page->post_status ) {
+			return null;
+		}
+
+		return $page;
+	}
+
+	/**
+	 * Shape a page for the wizard and the theme.
+	 *
+	 * @param \WP_Post $page The page.
+	 * @return array{editUrl: ?string, pageUrl: string|false, status: string, title: string} The page data.
+	 */
+	private static function page_data( \WP_Post $page ): array {
+		return [
+			'editUrl' => get_edit_post_link( $page->ID, 'raw' ),
+			'pageUrl' => get_permalink( $page->ID ),
+			'status'  => $page->post_status,
+			'title'   => get_the_title( $page->ID ),
+		];
+	}
+
+	/**
+	 * Create an accessibility statement page, unless the site already has one.
+	 *
+	 * Idempotent against a stored pointer, which is what a repeat click hits.
+	 * Two genuinely simultaneous requests can still each insert: WordPress has
+	 * no atomic option claim, and a second request's write is invisible to this
+	 * one anyway, because the first miss is cached for the rest of the request.
+	 *
+	 * @return array{editUrl: ?string, pageUrl: string|false, status: string, title: string}|\WP_Error
+	 *         The page data, or an error if it could not be created.
+	 */
+	public static function create_page(): array|\WP_Error {
+		$existing = self::get_stored_page();
+		if ( $existing ) {
+			return self::page_data( $existing );
 		}
 
 		// Get the Accessibility Statement boilerplate content.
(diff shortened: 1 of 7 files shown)


plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
-			if ( $page && 'page' === $page->post_type && in_array( get_post_status( $page->ID ), [ 'publish', 'draft' ] ) ) {
-				$page_data = [
-					'editUrl' => get_edit_post_link( $page->ID, 'raw' ),
-					'status'  => get_post_status( $page->ID ),
-					'pageUrl' => get_permalink( $page->ID ),
-				];
-				return $page_data;
-			}
-			// If page doesn't exist or is in trash, return null but keep the ID stored.
+	public static function get_page_id(): int {
```
adekbadek, line 131: **Suggestion:** A pointer at a deleted page re-queries on every call — `WP_Post::get_instance()` caches a hit but not a miss, so once the stored page is permanently deleted every `usable_page()` call goes back to the database. `get_page_id()` resolves the ID, then `get_stored_page()` resolves the same ID a second time.

Measured on the review env with `SAVEQUERIES`: `get_page()` costs 2 uncached queries when the option points at a deleted page, 3 when a dead legacy theme mod is present as well, against 1 query total for two calls when the pointer resolves. That lands on every uncached classic-theme render, and — because `post_status()` now calls `get_page_id()` — once per row on the Pages list table and Appearance > Menus, where the old code did a cached `get_theme_mod()` and no post lookup.

This state is not transient: keeping the stale pointer is what lets `api_get_page()` tell `missing` from `none`, so it persists until the publisher creates a new page. A request-scoped static on the resolved `WP_Post|null`, reset in `create_page()`, collapses all of it to one lookup; having `get_page_id()` hand the resolved post to `get_stored_page()` removes the duplicate resolution on its own.

**Consider:** the added lookups only cost anything when the stored ID points at a page that no longer exists, so a site with a deleted accessibility statement page is the one configuration worth checking before release.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
 		);
 
 		if ( is_wp_error( $page_id ) ) {
 			return $page_id;
 		}
 
-		// Save the page ID.
-		set_theme_mod( 'accessibility_statement_page_id', $page_id );
+		update_option( self::OPTION_NAME, $page_id, true );
+		set_theme_mod( self::LEGACY_THEME_MOD, $page_id );
```
adekbadek, line 213: **Nit:** How-to-test steps 10-11 do not reach the state they describe — This line writes the rollback breadcrumb on every create, and the breadcrumb lives in `theme_mods_<stylesheet>`, not in either `newspack_accessibility_statement_*` option. Step 10 deletes only the two options, so the breadcrumb survives, the migration re-runs with its flag gone and re-adopts the page step 9 created.

Ran the steps literally on the review env: after step 9 the option is 9 and the theme mod is 9; step 10 clears the option and the mod still reads 9; step 11 shows the draft with "not yet published" and an Edit button, not the "no page exists yet" notice the step predicts.

The code is behaving as designed — re-adoption is the point. It is the steps that need one more action, either `wp theme mod remove accessibility_statement_page_id` or deleting the page, or whoever verifies this will hit the mismatch at step 11 and may bounce the PR back.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
+
+	/**
+	 * Move the page ID from theme mods to a site-wide option.
+	 *
+	 * @return void
+	 */
+	public static function migrate_legacy_theme_mod(): void {
+		// admin_init also fires on admin-ajax.php and admin-post.php, both
+		// reachable logged out. Anonymous requests should not pay for the scan.
+		if ( wp_doing_ajax() || ! is_user_logged_in() ) {
```
adekbadek: **Suggestion:** The migration runs for any logged-in user, at any role — The guard admits every authenticated request that reaches `admin_init`, so the options-table scan and the three writes that follow it — `update_option` twice plus `set_theme_mod`/`remove_theme_mod`, the last of which normally sits behind `edit_theme_options` — can be paid for by a reader.

Verified on the review env: a Subscriber loading `/wp-admin/profile.php` ran the migration to completion, writing the option, the flag and the active theme's mod. Nothing here is attacker-controlled and the flag makes it once per site, so this is not an escalation — but a reader-facing write is the class of thing this PR exists to remove.

The plugin already has the shape for this: `includes/reader-activation/class-my-account.php:259` gates its own `admin_init` self-healing write on a capability check. Adding `! current_user_can( 'edit_pages' )` to the same early return costs the upgrade path nothing, because the legacy theme-mod fallback in `get_page_id()` covers the window until a privileged user visits.

**Consider:** a short transient set before the scan begins and cleared when it finishes would cap how often the scan can repeat if the completion flag does not read back, while keeping the retry-after-failure behaviour the comment above the flag describes.

**Risk if not addressed:** if the completion flag does not read back as set on a site whose stored pointer stays empty, every logged-in request into the admin repeats the options-table scan and its per-row reads. The repeat writes are no-ops because the values do not change, so the outcome is added query load on admin requests rather than data damage, and it stops the moment the flag reads back.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
+			return;
 		}
 
-		$page = get_post( $page_id );
-		if ( ! $page || 'page' !== $page->post_type || 'trash' === get_post_status( $page->ID ) ) {
-			// If page doesn't exist, isn't a page, or is in trash, return false but keep theme_mod.
-			return false;
+		$page_id = self::find_legacy_page_id();
+		if ( $page_id ) {
+			update_option( self::OPTION_NAME, $page_id, true );
```
adekbadek: **Suggestion:** The migration can overwrite a pointer written while it was scanning — The decision to write is made from the `get_option()` at :204, but the write at :221 is unconditional and a multi-query scan runs in between. A `create_page()` landing in that gap has its pointer replaced by the scan's result.

Narrow — `admin_init` runs during the wizard's own render, well before the REST POST fires on click, so it needs two concurrent admin sessions and a slow scan. `add_option()` closes it without a lock: core implements it as `INSERT … ON DUPLICATE KEY UPDATE option_name = VALUES(option_name)` and returns false rather than overwriting. Two concurrent migrations are already safe, since the scan is deterministic and both write the same value.

**Risk if not addressed:** if an editor clicks Create Page while another admin request is inside the scan, the migration's write lands after the create and points the site at the older page. The site then holds two accessibility statement drafts and shows the older one, which is the duplicate this change exists to prevent, recoverable by trashing the extra page.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -154,19 +337,27 @@ public static function api_create_page( $request ) {
...
 	 */
-	public static function api_get_page() {
+	public static function api_get_page(): \WP_REST_Response|\WP_Error {
 		$page_data = self::get_page();
-		return rest_ensure_response( $page_data );
+		if ( $page_data ) {
+			return rest_ensure_response( $page_data );
+		}
+
+		return rest_ensure_response( [ 'status' => self::get_page_id() ? 'missing' : 'none' ] );
```
adekbadek: **Suggestion:** The status key carries two unrelated meanings — `status` is a WordPress post status when a page exists and a sentinel when one does not, so the client cannot test for a page — it can only subtract values from the post-status vocabulary. `accessibility-statement.tsx:32` has to write `status !== 'none' && status !== 'missing'`, which misclassifies any post status registered under either name.

A separate key keeps both meanings intact: `[ 'page' => null, 'reason' => 'missing' ]`, or `[ 'exists' => false, 'trashed' => true ]`, with `isPage()` testing the discriminator.

Bounding this: the route requires `edit_pages` and this wizard view is its only consumer, so nothing external breaks today. Worth noting that the GET returned `false` here before and now returns a truthy object in the same case, so a falsiness check anywhere off this route would invert. The classic theme is unaffected — it reads `get_page()`, which still returns `false`.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
+	const OPTION_NAME = 'newspack_accessibility_statement_page_id';
+
+	/**
+	 * Where the page ID used to live. Theme mods are per theme, so every theme
+	 * and style variation kept its own, and a site lost the page each time it
+	 * switched. Still read until the upgrade has run, and still written on
+	 * create, so a plugin rolled back to before this change finds the page
+	 * where it looks for it instead of making a second one.
+	 */
+	const LEGACY_THEME_MOD = 'accessibility_statement_page_id';
```
adekbadek, line 30: **Nit:** The rollback breadcrumb does not survive a theme switch — This docblock reads as unconditional, but `set_theme_mod()` writes to whichever theme is active at write time, and `get_theme_mod()` reads only the active one. A publisher who switches theme or style variation after the page is created — the exact sequence the ticket describes, and the reason five style variations count as five themes — leaves a rolled-back plugin nothing to read.

So the dual write covers rollback-without-a-theme-switch, not rollback generally. Worth saying that here, or dropping the write, since `migrate_legacy_theme_mod()` re-adopts from the option going forward regardless.

**Risk if not addressed:** if the plugin is rolled back on a site that switched theme or style variation since the pointer was written, the older version finds no pointer and returns to creating a draft on every front-end render, including logged-out visits and 404s. The site collects duplicate drafts again until the plugin rolls forward.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
-				return $page_data;
-			}
-			// If page doesn't exist or is in trash, return null but keep the ID stored.
+	public static function get_page_id(): int {
+		$page_id = (int) get_option( self::OPTION_NAME, 0 );
+		if ( $page_id && self::usable_page( $page_id ) ) {
+			return $page_id;
+		}
+
+		$legacy_id = (int) get_theme_mod( self::LEGACY_THEME_MOD );
```
adekbadek: **Nit:** The legacy fallback is narrower than the docblock and the description say — Two limits worth recording where the fallback lives.

It reads the active theme only. The PR description says the old pointer "stays readable until the upgrade runs … so a published statement keeps its footer link and its content-gate exemption throughout" — that holds when the mod is on the active theme, and not for a site that switched, which is the population the ticket names. Verified on the review env: with a published page whose pointer sat only on an inactive theme's mods, `get_page_id()` returned 0 and the anonymous front end rendered no footer link until the migration ran, after which the link came back. Not a regression — on `main` those sites got a fresh draft whose `draft` status failed the theme's `'publish' ===` guard, so also no link — so the sentence is what wants narrowing, not the fallback. Widening the fallback to scan inactive themes on every front-end render would be far worse than the gap it closes.

It also rejects a trashed page, via `usable_page()`, while `find_legacy_page_id()` deliberately keeps one and says why in its own docblock. Worth having the two paths agree, or narrowing this docblock to the option branch it actually describes.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -10,141 +10,324 @@
...
 			'/wizard/newspack-settings/accessibility-statement',
 			[
-				'methods'             => 'GET',
-				'callback'            => [ __CLASS__, 'api_get_page' ],
-				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+				[
+					'methods'             => \WP_REST_Server::READABLE,
+					'callback'            => [ __CLASS__, 'api_get_page' ],
+					'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+					'args'                => [],
```
adekbadek: **Nit:** The empty args arrays are no-ops — Collapsing the two registrations into one with two method entries is the right WP form and matches the constants used across the plugin. Neither callback reads a parameter, though, and these are the only two empty `args` arrays in `includes/`. Dropping both lines keeps the registration consistent with the rest of the plugin; leaving them is harmless if they are a placeholder for the arg schema.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
```diff
@@ -180,13 +371,16 @@ public static function api_permissions_check() {
...
-	 * @param WP_Post $post The post object.
+	 * @param mixed $post_states The post states, normally an array.
+	 * @param mixed $post The post object, normally a WP_Post.
 	 * @return array The post states.
 	 */
-	public static function post_status( $post_states, $post ) {
-		$page_id = get_theme_mod( 'accessibility_statement_page_id' );
-		if ( $page_id === $post->ID ) {
+	public static function post_status( $post_states, $post = null ): array {
+		if ( ! is_array( $post_states ) ) {
```
adekbadek: **Nit:** Coercing post_states to an empty array discards earlier filters' work — Core cannot reach this branch: `get_post_states()` builds `$post_states` as an array and returns before applying the filter unless `$post instanceof WP_Post` (`wp-admin/includes/template.php`). Where it does fire — another plugin calling `apply_filters( 'display_post_states', … )` with the wrong shape — replacing the input with `[]` drops every state contributed before it, turning that plugin's bug into missing badges on this screen.

Returning the input untouched when it is not an array preserves both, at the cost of widening the `: array` return type. `tests/unit-tests/accessibility-statement-page.php:463` pins the discarding behaviour, so it moves with the change. The `instanceof \WP_Post` check on the next line is worth keeping either way — it is what makes the identity comparison safe.

plugins/newspack-plugin/src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
```diff
@@ -16,83 +17,93 @@ interface AccessibilityStatementProps {
...
+	};
 
-	// Function to fetch fresh data
-	const fetchFreshData = () => {
+	const handleError = ( error: { message?: string } ) => {
+		setLocalPageData( null );
+		setIsPageMissing( false );
+		setLocalIsFetching( false );
+		// The notice is a plain div, so nothing else announces the failure.
+		speak( error?.message ?? __( 'Something went wrong.', 'newspack-plugin' ), 'assertive' );
```
adekbadek: **Nit:** The announced error text is not entity-decoded — `speak()` gets the raw message while the same text rendered into the `Notice` goes through `decodeEntities()` in the hook (`use-wizard-api-fetch.ts:242`). A server message carrying an encoded entity — apostrophes in translated strings being the usual source — is announced with the entity intact. Passing the hook's `errorMessage`, or wrapping with `decodeEntities` here, keeps announced and displayed text identical.

plugins/newspack-plugin/tests/unit-tests/accessibility-statement-page.php
```diff
@@ -0,0 +1,578 @@
...
+		$before = $this->count_pages();
+
+		$this->assertFalse( Accessibility_Statement_Page::get_page() );
+		$this->assertSame( $before, $this->count_pages() );
+	}
+
+	/**
+	 * Repeated reads must stay free of side effects.
+	 */
+	public function test_repeated_reads_create_nothing() {
```
adekbadek: **Nit:** Some tests go red on the same mutation as their neighbour — The migration and rollback cases are the strongest part of this file — `test_a_theme_mod_from_a_deleted_theme_is_adopted` and `test_a_live_legacy_page_beats_a_stored_pointer_that_is_gone` each pin a decision that is not readable from the code.

Four earn less than they cost, on the trim pass the guidelines ask for. `test_repeated_reads_create_nothing` is `test_get_page_does_not_create_anything` in a loop once the method is a pure read. `test_the_stored_id_is_an_integer_for_the_content_gate` asserts `assertIsInt` on a method carrying an `int` return type. `test_post_state_tolerates_an_unexpected_argument` covers a branch core cannot reach. `test_a_theme_mod_from_an_inactive_theme_is_adopted` is subsumed by the deleted-theme case, which is the one that bites against a `wp_get_themes()` implementation.

Against that, the guard at `class-accessibility-statement-page.php:200` — the line deciding who pays for the scan and who performs the theme-mod write — is never executed: every test calls the method directly with an administrator set in `set_up()`. Deleting the guard leaves all 29 green. `wp_set_current_user( 0 )` and asserting the option and flag stay unset is the test that would notice.

adekbadek (approved):
Approving — no blockers. But the caching issues should be addressed. 

Release risk: LOW. No confirmed blockers; the duplicate-creation path is closed and verified end to end on a review env, including adoption of a pointer left on an inactive theme; the consequences pass returned low; `content-gate` is the only contract touched and has no consumer references.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5. `get_page_id()` and `get_stored_page()` now share one `resolve()`, memoised for the request and keyed on the option plus the theme mod, with `clean_post_cache` dropping it when the page is trashed, restored or deleted. Measured the same way: 40 calls against a deleted pointer went from 60 queries to 0. Removing the `clean_post_cache` line turns the trashed-pointer test red, so the invalidation is pinned.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Fixed in the description. Step 10 now permanently deletes the pages as well as the two options, since a theme mod is only followed to a page that still exists, and that is what leaves the site with no pointer at all.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5. The guard is now `! current_user_can( 'edit_pages' )`, with `test_migration_waits_for_a_user_who_could_create_the_page` covering a logged-out request and a subscriber leaving both the option and the flag unset while `get_page_id()` still finds the page. It goes red on the old guard. Left the transient out: the flag write is the last thing the method does, and the retry-after-failure behaviour is worth more than capping a repeat that only a failing write can cause.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5, `add_option` in both write paths.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5. No page now answers `[ 'reason' => 'missing' | 'none' ]`, `status` stays a post status, and `isPage()` discriminates on the `reason` key. The shape change is recorded in the description.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Narrowed the docblock in 7b40c3a5. Kept the write: it costs nothing and still covers rollback without a theme switch, which is the common shape.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Narrowed both in 7b40c3a5. The docblock now says the fallback reads the active theme only and skips a trashed page where `find_legacy_page_id()` keeps one, and the description sentence carries the same qualification.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Dropped both in 7b40c3a5.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5. The filter hands back what it was given when that is not an array, and the test asserts the pass-through instead of the empty array.

plugins/newspack-plugin/src/wizards/newspack/views/settings/advanced-settings/accessibility-statement.tsx
thomasguillot: Done in 7b40c3a5, `speak()` gets the message through `decodeEntities`.

plugins/newspack-plugin/tests/unit-tests/accessibility-statement-page.php
thomasguillot: Done in 7b40c3a5. Dropped the loop, the `assertIsInt` and the inactive-theme case. Kept the unexpected-argument one, since it now pins the pass-through rather than the discard. Added `test_migration_waits_for_a_user_who_could_create_the_page` for the guard, verified red against the old `is_user_logged_in()` check.

plugins/newspack-plugin/includes/advanced-settings/class-accessibility-statement-page.php
thomasguillot: Correcting this: the `add_option` change I described did nothing, and I have replaced it in fcf0f77d and 07ab8e48.

Two things were wrong with it. Current core overwrites on the duplicate-key branch, not just the option name: `INSERT ... ON DUPLICATE KEY UPDATE option_name = VALUES(option_name), option_value = VALUES(option_value), autoload = VALUES(autoload)`. What normally keeps `add_option()` non-destructive is the existence check above that query, and the check is skipped whenever the option sits in `notoptions`. This method calls `get_option( self::OPTION_NAME, 0 )` before the scan, and a miss there writes `notoptions[ OPTION_NAME ] = true` for the rest of the request, so the check was deterministically bypassed and a concurrent pointer was overwritten exactly as before.

It also had a failure of its own. In the one case where `add_option()` did refuse, a pre-existing row holding `0` or `""`, no pointer was written but the migration flag was recorded immediately after, so the site could never adopt. There is a test for that now, red against the old code.

The write is a small `store_page_id()` instead: it drops the three option cache entries, re-reads, and only writes when the pointer is still empty, returning whichever ID the site ends up on so the rollback breadcrumb follows it rather than the page the scan found. That narrows the window to the re-read rather than closing it, and the docblock says so. Closing it properly needs `INSERT IGNORE` and a raw query, which is not worth it here: the loser of the race is a new empty draft and the winner is the site's real statement page. `test_a_pointer_stored_during_the_scan_is_not_overwritten` covers the branch, and fails if either the cache deletes or the return-value gating is removed.

PR #915, opened 2026-08-19
feat(group-subscription): make invite links belong to the subscription (NPPD-2157)

adekbadek:
### All Submissions:

An invite link belonged to the manager who minted it: the URL carried that user ID and validation required them to still be a manager, so the link died on any owner or manager change.

The link is now a property of the subscription: the URL carries only `subscription` and `key`, the meta holds one entry per subscription instead of one per manager, and validation matches the key alone. Every manager sees the same link; Regenerate or Disable clears every key at once.

Removing a manager no longer takes down the links they shared, so the demotion notice now names the control that does: "To stop the invite link for this group, regenerate it."

Links already in inboxes still carry the `manager` arg; it is ignored. Keys in the old shape keep working on their original terms — only while the manager who minted them still manages the group — until the next Regenerate or Disable rewrites the meta.

Closes NPPD-2157.

1. Enable `NEWSPACK_CONTENT_GATES` and Reader Activation. Activate WooCommerce and WooCommerce Subscriptions.
2. Create an active group subscription for a reader account.
3. Sign in as that reader. Go to My Account > Group.
4. Select **Copy invite link**. The URL contains `subscription` and `key` only. It has no `manager` parameter.
5. Transfer the subscription to a second reader account. The first account is no longer a manager.
6. Sign in as a third reader account. Open the copied link. The reader joins the group.
7. Sign in as the new owner. Go to My Account > Group. The page shows the same link the first account created.
8. Promote a member to manager, then select **Remove manager**. The notice tells you to regenerate the invite link.

Legacy URLs:

1. Set the subscription's `newspack_group_subscription_link_invites` meta by hand to the old per-manager shape: `[ <user_id> => [ 'key' => 'testkey', 'created_at' => 1000 ] ]`. Use a `<user_id>` that manages the group.
2. Sign in as a reader who is not in the group.
3. Open `?action=group_invite_link&subscription=<id>&manager=<user_id>&key=testkey`. The reader joins the group.
4. Repeat from step 1 with a `<user_id>` that does not manage the group. The reader does not join, and sees the invalid-link notice.

### Other information:

Full suite green: 3098 tests, 9089 assertions, 0 failures. PHPCS clean.

For the release note: removing a manager is no longer a way to revoke an invite link. Once a group's link has been regenerated under the new shape, Regenerate and Disable are the only ways to stop a link in circulation. A link still stored in the old per-manager shape keeps the old behaviour until then — revoked while its creator is not a manager, working again if that person is re-added.


plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -49,8 +49,18 @@ class Group_Subscription_Invite {
...
+	 * The subscription meta key for the invite-link entry.
+	 * Stored as: [ 'key' => string, 'created_at' => int, 'created_by' => int ].
+	 *
+	 * Only `key` is load-bearing. `created_at` and `created_by` are an audit trail: invite links do
+	 * not expire, and the link is deliberately not scoped to its creator.
+	 *
+	 * The link belongs to the subscription, not to the manager who minted it, so it keeps working
+	 * when the owner or managers change. Subscriptions written before that carry the legacy
+	 * per-manager shape, [ $manager_user_id => [ 'key' => string, 'created_at' => int ] ]. A legacy
+	 * key stays valid only while its creator still manages the group — the terms it was minted
```
dkoo: **Suggestion:** PR description promises the opposite of the shipped legacy behavior — The PR body says legacy keys "stay valid on the same terms as new ones — including keys minted by someone who has since stopped managing the group". The code deliberately rejects those keys (`test_validate_link_invite_rejects_a_legacy_key_of_a_departed_manager` pins it, and the rejection reproduces on a live site), and this docblock explains why that's the right call. The body looks like it predates the "keep revoked links revoked" commit — worth updating the description so a changelog or release note doesn't repeat the stale claim.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -207,57 +217,126 @@ public static function get_invites( $subscription, $show_expired = true ) {
...
+		// Legacy per-manager shape, from before the link belonged to the subscription. A legacy key
+		// is honoured only while its creator still manages the group, which is exactly what the old
+		// per-manager validation did.
+		//
+		// Keeping that condition matters because removing a manager was how an institution revoked
+		// the links that person had circulated: honouring their keys now would silently hand paid
+		// access back to everyone holding one, on existing production data, with nobody told. New
+		// links get the subscription-wide semantics — they outlive any change of manager — and a
+		// legacy link whose creator is still a manager now survives that manager's demotion too
+		// (NPPD-2120). What does not happen is a dead link coming back to life.
```
dkoo: **Suggestion:** "A dead link coming back to life" overstates the guarantee — `get_link_invite_entries()` re-evaluates `user_is_manager()` on every read, so a legacy key is only dead while its creator isn't a manager: re-promoting a previously removed manager silently revives every legacy link that person circulated — no UI signal, nothing logged. The old per-manager validation behaved the same way, and Disable/Regenerate now clear the whole map, so behavior is a net improvement. But this comment (and the `LINK_META` docblock's matching claim) is load-bearing for the PR's security argument, so it should say "revoked while its creator is not a manager" rather than assert permanence. If permanence is actually wanted, only a one-time migration of the legacy map gives it.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -784,9 +869,34 @@ public static function process_link_invite_request() {
...
+			// link names only the subscription and the reader who clicked it — and the
+			// question being asked in the weeks after an upgrade is usually "whose link
+			// is this, and why did it stop working?". Null when the key matches nothing,
+			// which is itself the answer.
+			// Guarded on the subscription itself: validation also fails when the URL
+			// names a subscription that does not exist, and there is nothing to read
+			// entries from in that case.
+			$entry = null;
+			if ( $subscription ) {
+				foreach ( self::get_link_invite_entries( $subscription ) as $candidate ) {
```
dkoo: **Suggestion:** Invalid-link log's `created_by` is null in the case it was built for — The lookup loop reads `get_link_invite_entries()`, which has already dropped any legacy entry whose creator no longer manages the group — and on the `not_found` path a match is impossible by construction, since validation failed precisely because no key from those same entries matched. So a revoked departed-manager key — the "whose link is this, and why did it stop working?" support case the comment names — logs `created_by: null`, indistinguishable from a garbage key.

Verified on a live site: a legacy entry keyed to a user who never managed the subscription, clicked while logged in, logged `{"subscription_id":12,"member_id":5,"created_by":null}` while the raw meta still named the creator. This path only runs after validation has already failed, so a fix can't affect access decisions.

**Consider:** Reading the creator from the raw stored entries rather than from the already-filtered list would let this log answer the question it was added for: a link that stopped working because its creator left the group is exactly the case that currently records no creator at all.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -207,57 +217,126 @@ public static function get_invites( $subscription, $show_expired = true ) {
...
+	 */
+	private static function get_link_invite_entries( $subscription ) {
+		$stored = $subscription->get_meta( self::LINK_META, true );
+		if ( ! is_array( $stored ) || empty( $stored ) ) {
+			return [];
+		}
+
+		// Current shape: one entry for the whole subscription. Its key stays valid however the
+		// managers change -- that is the point of storing the link against the subscription.
+		if ( ! empty( $stored['key'] ) ) {
```
dkoo: **Nit:** Shape detection is correct by accident; a corrupt key value matches as `'Array'` — `! empty( $stored['key'] )` conflates "current shape with an empty key" with "legacy map" — the result is right (the empty-key test pins it) but by accident of the heuristic. And if corrupted meta ever holds a non-string under `key`, `array_map( 'strval', array_column( … ) )` emits an Array-to-string warning and yields the literal string `'Array'`, which `?key=Array` would then match. Not reachable by an attacker (requires already-writable meta), but `array_key_exists( 'key', $stored )` plus an `is_string()` check would classify by shape first and fail closed on corruption.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -207,57 +217,126 @@ public static function get_invites( $subscription, $show_expired = true ) {
...
+			return [ $stored ];
+		}
```
dkoo, line 253: **Nit:** A plugin rollback can mint a key the upgraded code then hides — If a site rolls back to the pre-change plugin after a current-shape entry exists, the old `generate_link_invite()` merges an integer-keyed entry into the flat one; on re-upgrade the `key`-present branch returns only the flat entry, so the rollback-minted key dies silently until the next regenerate or disable rewrites the meta. Mixed-version windows are usually out of scope — noting it because "the link I made last week stopped working" is a recognizable support-ticket shape.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -207,57 +217,126 @@ public static function get_invites( $subscription, $show_expired = true ) {
...
+	 *
+	 * Every manager of a subscription shares one link, so this takes no user.
 	 *
 	 * @param \WC_Subscription|int $subscription The subscription object or ID.
-	 * @param int                  $user_id      The manager user ID.
 	 *
 	 * @return array|null The link-invite entry, or null if missing or subscription invalid.
 	 */
-	public static function get_link_invite( $subscription, $user_id ) {
+	public static function get_link_invite( $subscription ) {
```
dkoo, line 324: **Nit:** Three public signatures changed on a shared integration surface — `get_link_invite()`, `get_link_invite_url()` and `validate_link_invite()` dropped their `$user_id` parameter. Every caller is updated in this diff, and searches across the monorepo and the standalone manager checkout found no external references, so nothing known breaks — flagging only because the class sits on a shared integration surface, and any out-of-tree caller of the old arity would fatal on PHP 8.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -277,32 +356,26 @@ public static function generate_link_invite( $subscription, $user_id ) {
...
 		$entry = [
 			'key'        => wp_generate_password( 32, false ),
-			'created_at' => $now,
+			'created_at' => time(),
+			'created_by' => $user_id,
 		];
-		$all[ $user_id ] = $entry;
 
-		$subscription->update_meta_data( self::LINK_META, $all );
+		$subscription->update_meta_data( self::LINK_META, $entry );
```
dkoo, line 408: **Nit:** Concurrent regenerates: last write wins, silently — Two managers clicking Regenerate around the same time is now a lost update: each mints a key and saves the whole entry, and the loser's clipboard holds a URL that no longer validates while their UI reports success (`generateLink()` in `src/my-account/v1/group-subscriptions.js` reports from its own response only). Under the per-manager map both writes survived. Inherent to one shared link and self-healing on re-click — noting it rather than asking for a change.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
```diff
@@ -324,26 +397,33 @@ public static function delete_link_invite( $subscription, $user_id ) {
 			);
 		}
 
-		$all = $subscription->get_meta( self::LINK_META, true );
-		if ( ! is_array( $all ) || ! isset( $all[ $user_id ] ) ) {
+		// Nothing stored means nothing to disable. Skipping the write keeps a no-op click off the
+		// full woocommerce_update_subscription cascade, which this request reaches from My Account.
+		// The check is on the raw meta rather than on get_link_invite_entries(), so a corrupt or
+		// filtered-out value — which no reader could revoke any other way — is still cleared.
+		if ( '' === $subscription->get_meta( self::LINK_META, true ) ) {
```
dkoo, line 447: **Nit:** Two newly justified branches have no test — This guard is documented as existing so "a corrupt or filtered-out value … is still cleared", but no test stores a corrupt value and asserts the delete happens (a literal empty-string meta value is also treated as absent and left behind — harmless, since it holds no key). And nothing covers the invalid-link log payload; a test asserting `created_by` for a departed-manager legacy key would have caught the null-diagnostic issue directly.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/group-subscription-members.php
```diff
@@ -365,7 +365,7 @@ class="newspack-ui__tabs__panel<?php echo 'invites' === $active_tab ? ' selected
 
 				<section class="newspack-ui__modal__content">
 						<p>
-							<?php esc_html_e( 'The current link will stop working. You\'ll get a new link to share, and anyone who hasn\'t joined yet will need the new link.', 'newspack-plugin' ); ?>
+							<?php esc_html_e( 'Any invite link for this group will stop working. You\'ll get a new link to share, and anyone who hasn\'t joined yet will need the new link.', 'newspack-plugin' ); ?>
```
dkoo, line 368: **Nit:** Disable modal and snackbar copy still describe a single link — This Regenerate copy was updated to "Any invite link for this group will stop working", but the Disable modal (line 442) still says "The current link will stop working", and the `invite_link_regenerated` snackbar (`class-my-account-ui-v1.php:119`, "The old one no longer works") keeps the singular framing. Both actions now clear every stored key — including peer managers' legacy keys the viewer was never shown — so the three strings make the same promise and only one was reworded.

plugins/newspack-plugin/tests/unit-tests/content-gate/group-subscriptions.php
```diff
@@ -1444,15 +1455,183 @@ public function test_validate_link_invite_valid() {
...
+			Group_Subscription_Invite::validate_link_invite( $refetched, Group_Subscription_Invite::get_link_invite( $refetched )['key'] ),
+			'The freshly minted key is the only one left.'
+		);
+	}
+
+	/**
+	 * A keyless URL must never be accepted, from either direction -- hash_equals( '', '' ) is true.
+	 * The two guards are redundant by design, so this pins the behaviour rather than either guard.
+	 */
+	public function test_validate_link_invite_rejects_empty_key() {
```
dkoo, line 1637: **Nit:** Empty-key test exercises only one of the two directions its docblock claims — Both `assertWPError` calls pass `''` as the URL key, so both trip the supplied-key guard; the stored-key direction never reaches `validate_link_invite()`'s comparison. The `assertNull( get_link_invite( … ) )` next to them does pin the entry-filtering side, so the coverage exists — just not where the docblock says. Asserting `validate_link_invite( $sub, 'anything' )` against an empty stored key would close the loop.

dkoo (approved):
Release risk: LOW — no blockers; the three changed public signatures have no callers outside this diff (checked across the monorepo and the standalone manager checkout), and the behavior contract was exercised end-to-end on a live site: a fresh URL carries only `subscription` + `key`, the link survives an ownership transfer, the new owner sees the same link, a legacy `manager=` URL still joins, and a departed manager's legacy key stays revoked. Two things worth a look before merge, both with inline comments: the PR description still promises the opposite of the shipped departed-manager behavior, and the new `created_by` log field is null in exactly the case it was added for. The rest is copy and test polish.

Removing a manager used to kill the links that person had circulated, and after this change it no longer does for links minted in the new shape. Since the Remove manager control and the invite-link controls sit on the same page, a short line there — or in the removal confirmation — pointing the owner at Regenerate invite link would stop the two from being confused when someone is trying to cut off access.

Worth stating plainly in the release note as well: once a group's link has been regenerated under the new shape, removing a manager no longer revokes it, and Regenerate or Disable becomes the only way to stop a link that is already in circulation.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Reworded rather than migrating. The comment, the `LINK_META` docblock and `get_link_invite_keys()`'s docblock now say a legacy key is valid only while its creator manages the group, that the check runs on every read so revocation follows manager status in both directions — re-adding a removed manager revives their keys — and that Regenerate and Disable are the permanent revocation because they clear the map. Nothing observable changed; migrating the legacy map would, so that stays a separate call.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Fixed. The meta parsing now sits in `get_stored_link_invite_entries()`, which normalises both shapes and returns everything the subscription holds; `get_link_invite_entries()` is the thin filter on top that drops a legacy entry whose creator no longer manages the group. The log lookup reads the unfiltered list, so a revoked departed-manager key now logs its creator instead of null. `test_process_link_invite_request_logs_the_creator_of_a_revoked_key()` pins it.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Classification is now by shape: `array_key_exists( 'key', $stored )` picks the current shape, and the entry is only used when `key` is a non-empty string. Corrupt meta is read as a broken current-shape entry and dropped rather than falling through to the legacy branch, so `?key=Array` matches nothing — `test_validate_link_invite_rejects_a_corrupt_key()` covers it. The legacy branch also requires a numeric map key and a string `key`.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Handled rather than noted. The reader no longer returns early on the current shape: it takes the flat entry when there is one, then walks the numeric keys for legacy entries, so a meta value holding both yields both keys. A rollback-minted key survives the re-upgrade instead of dying until the next regenerate. `test_validate_link_invite_accepts_both_shapes_in_one_meta_value()` builds that merged value directly.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Both covered. `test_delete_link_invite_clears_an_unusable_value()` stores a departed manager's legacy key — no usable entry, nothing any reader could revoke otherwise — and asserts the row leaves the database, which is what pins the guard to the raw meta. `test_process_link_invite_request_logs_the_creator_of_a_revoked_key()` asserts the log payload's `created_by` for exactly that key.

plugins/newspack-plugin/includes/plugins/woocommerce/my-account/templates/v1/group-subscription-members.php
adekbadek: All three now match. The Disable modal reads "Any invite link for this group will stop working", and the `invite_link_regenerated` snackbar is "New invite link copied. Any older link no longer works." — in the PHP labels and in the JS fallback string that shadows them.

plugins/newspack-plugin/tests/unit-tests/content-gate/group-subscriptions.php
adekbadek: Closed the loop: the empty-stored-key half now also asserts `validate_link_invite( $group_sub, 'anything' )` is an error, so a non-empty URL key reaches the comparison and finds nothing to match.

adekbadek:
Stated in the PR description, which becomes the squash commit body and so the release note: removing a manager no longer revokes an invite link, and once a group's link has been regenerated under the new shape, Regenerate and Disable are the only ways to stop a link already in circulation. It also states what holds until then — a link still in the old per-manager shape is revoked while its creator is not a manager, and re-adding that person makes it work again.

Two stale legacy claims are corrected in the same pass: the description said existing keys stay valid "including keys minted by someone who has since stopped managing the group", and the legacy test steps said such a link still admits the reader.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: The description has been rewritten since that review. It now says keys in the old shape keep working "on their original terms — only while the manager who minted them still manages the group", and the release-note paragraph says the same in publisher terms. Nothing in the body still promises the unconditional behaviour.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Re-checked on this branch: the callers are the My Account templates and the REST controller, both in this diff. No shim keeps the old arity on purpose — an out-of-tree caller should fail loudly rather than pass a user ID into a validation path that no longer reads it.

plugins/newspack-plugin/includes/plugins/woocommerce-subscriptions/group-subscription/class-group-subscription-invite.php
adekbadek: Left as it is, and now written down rather than only known: `generate_link_invite()`'s docblock records that two simultaneous regenerates are a lost update, that both callers are told they succeeded, and that re-copying from the page hands back the live link. The window is inherent to one link per subscription, and the alternative is per-manager links again.

adekbadek:
Added the pointer you asked for. Removing a manager and stopping a link in circulation are now separate actions, so the demotion notice names the second one:

> group-joiner is no longer a manager. To stop the invite link for this group, regenerate it.

The sentence is appended only when the group has a live link, and the read runs after the demotion, so a legacy key the removal has just revoked does not trigger it. `test_demotion_notice_points_at_the_invite_link_control()` pins both halves, and the notice was walked through in a browser.

There is no confirm step on Remove manager (it is an instant action by design), so the notice is the only place this fits without adding one.

PR #916, opened 2026-08-19
fix(newsletters): keep subscription lists on a cold ESP cache

leogermani:
### All Submissions:

Groups and tags are read from the Mailchimp cache while the subscription list set is assembled. A cache miss returns an empty array, which is indistinguishable from an audience that genuinely has none. On a site whose cache has not warmed yet, that partial set was treated as authoritative: the garbage collector deleted every stored group and tag list, and the partial set was cached for the full TTL, so the settings screen kept showing audiences only long after the data landed.

The cache now records a cold read and reports it through a new `newspack_newsletters_lists_are_complete` filter. When the set is partial, `get_lists()` skips both the garbage collector and the transient write. A complete read behaves as before.

A second commit fixes the other half. WordPress records a missing option in its `notoptions` cache and answers later reads from that record without querying the database. Because a cache miss dispatches a refresh that writes the option moments later from its own request, that record hid the write: on a site with a persistent object cache it outlived the request, every later read was told the option was still missing, and the data never appeared. The cache now clears the record as soon as it misses.

Other ESPs are unaffected: the filter is registered from the Mailchimp cache, which only initializes when Mailchimp is the configured provider.

1. Connect a Mailchimp account with an audience that has groups or tags. This setup is what the bug needs.
2. Go to Newsletters > Settings. Group and tag rows are listed under the audience.
3. Record the stored list count: `wp post list --post_type=newspack_nl_list --format=count`.
4. Clear the audience cache: `wp option delete newspack_nl_mailchimp_cache_<audience_id> && wp cache flush`. Editing the row in SQL needs the flush too.
5. Delete the list transient: `wp transient delete newspack_newsletters_lists_mailchimp`.
6. Reload Newsletters > Settings. Only the audience shows, which is expected while the cache warms.
7. Re-run the count from step 3. It is unchanged, so no lists were deleted.
8. Wait a few seconds for the background refresh, then reload. Groups and tags are listed again.

### Other information:

Step 4 needs `wp cache flush` on any site with a persistent object cache, so the run starts from a coherent state. Changing the option outside WordPress, in SQL, needs it as well: the object cache keeps its own copy either way.

The provider test double left `$service` unset, and the garbage collector matches stored lists on that slug, so the collector was a silent no-op in every existing `get_lists()` test. The double now sets it, which is what lets the new tests exercise deletion.

### DIFF
--- a/plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php
+++ b/plugins/newspack-newsletters/includes/class-newspack-newsletters-subscription.php
@@ -529,15 +529,36 @@ class Newspack_Newsletters_Subscription {
 			);
 
 			/**
+			 * Filters whether the ESP reported its full set of lists.
+			 *
+			 * A provider that serves sublists from its own cache returns a partial
+			 * set while that cache is still warming. Acting on a partial set is
+			 * destructive, so providers report it here and this method stands down.
+			 *
+			 * @param bool $is_complete Whether the list set is complete.
+			 */
+			$is_complete = (bool) apply_filters( 'newspack_newsletters_lists_are_complete', true );
+
+			/**
 			 * Remove from the local DB lists that no longer exist in the ESP.
 			 * This also cleans up the DB in case we accidentally created more than one list for the same ESP list.
+			 *
+			 * Skipped on a partial set: the collector deletes every stored remote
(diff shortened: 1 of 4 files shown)


leogermani:
closed in favor of https://github.com/Automattic/newspack-workspace/pull/921

PR #917, opened 2026-08-19
fix(newsletters): refresh only Mailchimp audiences in use (NPPM-3070)

leogermani:
### All Submissions:

Every ten minutes, the Mailchimp cache cron dispatched one async admin-ajax request per audience in the account, each making about five API calls. Accounts hold far more audiences than a site sends to, so the work scaled with the account rather than with the site. On thestranger.com this was the largest single admin-ajax cost, roughly 19.9 minutes of cumulative request time in the observed window.

The cron now refreshes only the audiences in use: those behind a configured subscription list, as the audience itself or as the parent of a group or tag, and those a newsletter sends to. Audiences left out still self-heal, because a read dispatches its own refresh when the cache is missing or stale.

`newspack_newsletters_mailchimp_audiences_to_refresh` overrides the set, so an account with an unusual shape can pin an explicit list.

Closes NPPM-3070.

1. Connect a Mailchimp account with several audiences, only one of which is configured as a subscription list.
2. Enable logging so cron dispatches are visible: `wp config set NEWSPACK_NEWSLETTERS_DEBUG_MODE true --raw`.
3. Run the cron: `wp cron event run newspack_nl_mailchimp_refresh_cache`.
4. Check the log. A refresh is dispatched for the configured audience only, not for the others.
5. Set a newsletter's send list to a second, unconfigured audience, then run the cron again.
6. Check the log. That audience is now dispatched too.
7. Open the Newsletters editor on an audience the cron skipped. Its groups and tags load, confirming reads still self-heal.

### Other information:

Stacked on #916, which is what makes this safe: before it, an audience left cold could be deleted from the stored subscription lists on the next settings-screen load. Review and merge #916 first; this PR targets that branch.

Only the audience-scoping lever from NPPM-3070 is here. Caching campaign folders once per account, merging the segments and tags calls, the interest-groups N+1, and the cron interval are all untouched.

### DIFF
--- a/plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
+++ b/plugins/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp-cached-data.php
@@ -8,6 +8,7 @@
 defined( 'ABSPATH' ) || exit;
 
 use Newspack_Newsletters_Mailchimp_Api as Mailchimp;
+use Newspack\Newsletters\Subscription_Lists;
 
 /**
  * Mailchimp cached class data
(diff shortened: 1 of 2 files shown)


leogermani:
closed in favor of https://github.com/Automattic/newspack-workspace/pull/921

PR #918, opened 2026-08-19
fix(network): distribute dynamic galleries as resolved images

rbcorrales:
### All Submissions:

WordPress 7.1's dynamic gallery saves no image IDs. It saves "show the images attached to this post", which core resolves at render time against the post being rendered. That instruction distributes intact but means something different on a node, where the only attached image is the sideloaded featured image. The gallery arrives showing a single unrelated image, or nothing, with no error.

This resolves the images on the origin, so the block distributes in the same shape as an ordinary gallery.

The node's copy is static. Attaching an image on the origin does not resync on its own, so the node updates on the next save of the post.

Closes NEWS-2951.

1. On a hub, create a post and upload three images from inside the post so they attach.
2. Set a featured image that is not one of those three.
3. Delete the three Image blocks. The images stay attached to the post.
4. Add an empty Gallery block and click **Use attached images**.
5. Confirm the block offers **Detach Gallery**, which means it is in dynamic mode.
6. Publish, then distribute to a node from the Distribute sidebar.
7. Pull on the node and open the post. Expect the three original images.
8. Repeat without a featured image. Expect three images, where before there were none.
9. Distribute a post with an ordinary gallery. Expect no change.
10. Distribute a dynamic gallery nested inside a Group block. Expect three images.

### Other information:

The new tests skip on WordPress below 7.1, which is what CI currently runs, so they report as skipped rather than passing.

`aspectRatio` does not travel: a gallery pinned to a ratio renders at natural ratios on the node.

### DIFF
--- a/plugins/newspack-network/includes/content-distribution/blocks/class-block-processor.php
+++ b/plugins/newspack-network/includes/content-distribution/blocks/class-block-processor.php
@@ -76,15 +76,16 @@ class Block_Processor {
 	/**
 	 * Process an outgoing block.
 	 *
-	 * @param array $block The block to process.
+	 * @param array $block   The block to process.
+	 * @param int   $post_id The ID of the post being distributed.
 	 *
 	 * @return array The processed block.
 	 */
-	public function process_outgoing_block( $block ) {
+	public function process_outgoing_block( $block, $post_id = 0 ) {
 		if ( ! is_callable( $this->outgoing_callback ) ) {
 			return $block;
 		}
-		return call_user_func( $this->outgoing_callback, $block );
+		return call_user_func( $this->outgoing_callback, $block, $post_id );
 	}
 
 	/**
--- a/plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
+++ b/plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
@@ -27,6 +27,7 @@ class Blocks {
 		// Register block processors.
 		self::register_block_processor( 'jetpack/slideshow', [ __CLASS__, 'process_jetpack_galleries' ] );
 		self::register_block_processor( 'jetpack/tiled-gallery', [ __CLASS__, 'process_jetpack_galleries' ] );
+		self::register_block_processor( 'core/gallery', [ __CLASS__, 'process_dynamic_gallery' ] );
 	}
 
 	/**
@@ -60,11 +61,25 @@ class Blocks {
 	/**
 	 * Process an outgoing block.
 	 *
-	 * @param array $block The block to process.
+	 * Recurses into inner blocks so a block nested inside a Group, Columns or
+	 * similar container is processed too. Without this, a gallery only gets
+	 * handled when it sits at the top level of the post.
+	 *
+	 * @param array $block   The block to process.
+	 * @param int   $post_id The ID of the post being distributed.
 	 *
 	 * @return array The processed block.
 	 */
-	public static function process_outgoing_block( $block ) {
+	public static function process_outgoing_block( $block, $post_id = 0 ) {
+		if ( ! empty( $block['innerBlocks'] ) ) {
+			$block['innerBlocks'] = array_map(
+				function ( $inner_block ) use ( $post_id ) {
+					return self::process_outgoing_block( $inner_block, $post_id );
+				},
+				$block['innerBlocks']
+			);
+		}
+
 		$block_name = $block['blockName'];
 
 		$processors = self::get_block_processors( $block_name );
@@ -73,7 +88,7 @@ class Blocks {
 		}
 
 		foreach ( $processors as $processor ) {
-			$block = $processor->process_outgoing_block( $block );
+			$block = $processor->process_outgoing_block( $block, $post_id );
 		}
 		return $block;
 	}
@@ -120,5 +135,98 @@ class Blocks {
 	public static function process_jetpack_galleries( $block ) {
 		unset( $block['attrs']['ids'] );
 		return $block;
+	}
+
+	/**
+	 * Flatten a dynamic gallery into ordinary image blocks for distribution.
+	 *
+	 * A gallery in dynamic mode stores no image IDs. It stores the instruction
+	 * "show the images attached to this post", which core resolves at render time
+	 * against whichever post is being rendered. That instruction travels intact,
+	 * but it means something different on a node, where the only attached image is
+	 * the sideloaded featured image. The gallery then arrives showing a single
+	 * unrelated image, or nothing at all, with no error either way.
+	 *
+	 * Resolving the images here, on the origin, leaves the block in the same shape
+	 * an ordinary gallery is saved in, which already distributes correctly. The
+	 * node's copy stops following new attachments, which matches how every other
+	 * edit reaches a node: on the next sync.
+	 *
+	 * @param array $block   The block to process.
+	 * @param int   $post_id The ID of the post being distributed.
+	 *
+	 * @return array The processed block.
+	 */
+	public static function process_dynamic_gallery( $block, $post_id = 0 ) {
+		if ( empty( $block['attrs']['dynamicContent'] ) ) {
+			return $block;
+		}
+
+		// Dynamic galleries arrived in WordPress 7.1. Leave the block alone on older versions.
+		if ( ! function_exists( 'block_core_gallery_resolve_dynamic_source' ) ) {
+			return $block;
+		}
+
+		$post_id = $post_id ? $post_id : get_the_ID();
+		if ( ! $post_id ) {
+			return $block;
+		}
+
+		$attachment_ids = block_core_gallery_resolve_dynamic_source(
+			$block['attrs']['dynamicContent'],
+			new \WP_Block( $block, [ 'postId' => $post_id ] )
+		);
+		if ( empty( $attachment_ids ) ) {
+			return $block;
+		}
+
+		$size_slug = $block['attrs']['sizeSlug'] ?? 'large';
+		$link_to   = $block['attrs']['linkTo'] ?? 'none';
+
+		$images = [];
+		foreach ( $attachment_ids as $attachment_id ) {
+			$url = wp_get_attachment_image_url( $attachment_id, $size_slug );
+			if ( ! $url ) {
+				continue;
+			}
+			$images[] = sprintf(
+				"<!-- wp:image %1\$s -->\n<figure class=\"wp-block-image size-%2\$s\"><img src=\"%3\$s\" alt=\"%4\$s\" class=\"wp-image-%5\$d\"/></figure>\n<!-- /wp:image -->",
+				wp_json_encode(
+					[
+						'id'              => (int) $attachment_id,
+						'sizeSlug'        => $size_slug,
+						'linkDestination' => $link_to,
+					]
+				),
+				esc_attr( $size_slug ),
+				esc_url( $url ),
+				esc_attr( (string) get_post_meta( $attachment_id, '_wp_attachment_image_alt', true ) ),
+				(int) $attachment_id
+			);
+		}
+
+		if ( empty( $images ) ) {
+			return $block;
+		}
+
+		$attrs = $block['attrs'];
+		unset( $attrs['dynamicContent'] );
+
+		// Keep the gallery's own wrapper so its classes and layout survive.
+		$figure_open = '<figure class="wp-block-gallery has-nested-images columns-default is-cropped">';
+		if ( ! empty( $block['innerHTML'] ) && preg_match( '/<figure[^>]*>/', $block['innerHTML'], $matches ) ) {
+			$figure_open = $matches[0];
+		}
+
+		$markup = sprintf(
+			"<!-- wp:gallery %1\$s -->\n%2\$s\n%3\$s\n</figure>\n<!-- /wp:gallery -->",
+			wp_json_encode( (object) $attrs ),
+			$figure_open,
+			implode( "\n", $images )
+		);
+
+		$parsed = parse_blocks( $markup );
+
+		return $parsed[0] ?? $block;
 	}
 }
--- a/plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
+++ b/plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
@@ -7,6 +7,8 @@
 
 namespace Newspack_Network\Content_Distribution;
 
+use Newspack_Network\Debugger;
+
 /**
  * Blocks class.
  */
@@ -148,9 +150,24 @@ class Blocks {
 	 * unrelated image, or nothing at all, with no error either way.
 	 *
 	 * Resolving the images here, on the origin, leaves the block in the same shape
-	 * an ordinary gallery is saved in, which already distributes correctly. The
-	 * node's copy stops following new attachments, which matches how every other
-	 * edit reaches a node: on the next sync.
+	 * an ordinary gallery is saved in, which already distributes correctly.
+	 *
+	 * The image and wrapper construction mirrors core's own dynamic render
+	 * (`block_core_gallery_render_dynamic_image()` and the `dynamicContent` branch
+	 * of `block_core_gallery_render()`), reusing core's helpers so links, captions,
+	 * aspect ratio and responsive markup match what the origin renders. Blocks are
+	 * built as arrays and left for `serialize_blocks()` to encode, so attribute
+	 * values never pass through a hand-written block-comment delimiter.
+	 *
+	 * Two caveats worth knowing:
+	 *
+	 * - The node's copy is static. Attaching a new image on the origin does not by
+	 *   itself resync the post, so the node updates on the next save of the post,
+	 *   not on the upload.
+	 * - When nothing resolves, the block is flattened to an empty gallery rather
+	 *   than shipped as-is. Passing the dynamic instruction through would let the
+	 *   node resolve it against its own attachments, which is the bug this exists
+	 *   to prevent.
 	 *
 	 * @param array $block   The block to process.
 	 * @param int   $post_id The ID of the post being distributed.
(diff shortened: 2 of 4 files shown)


plugins/newspack-network/includes/content-distribution/blocks/class-block-processor.php
```diff
@@ -76,15 +76,16 @@ public function set_incoming_callback( $callback = null ) {
...
 	 * @return array The processed block.
 	 */
-	public function process_outgoing_block( $block ) {
+	public function process_outgoing_block( $block, $post_id = 0 ) {
 		if ( ! is_callable( $this->outgoing_callback ) ) {
 			return $block;
 		}
-		return call_user_func( $this->outgoing_callback, $block );
+		return call_user_func( $this->outgoing_callback, $block, $post_id );
 	}
```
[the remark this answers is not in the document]
rbcorrales: Checked this on PHP 8.5. Extra arguments to user-defined functions, methods and closures are ignored; only *internal* functions throw `ArgumentCountError`:

```
static method, 2 args passed: ok:2
closure,       2 args passed: ok:2
internal fn,   2 args passed: THREW ArgumentCountError
```

Every registered outgoing callback is a user-defined static method, so the two-argument call is safe. Keeping the signature as it is.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +136,97 @@ public static function process_jetpack_galleries( $block ) {
 		unset( $block['attrs']['ids'] );
 		return $block;
```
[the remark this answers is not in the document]
rbcorrales: Same as the thread on `class-block-processor.php`. `process_jetpack_galleries` is a user-defined static method, so the extra argument is ignored rather than fatal. Only internal functions throw `ArgumentCountError`.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +136,97 @@ public static function process_jetpack_galleries( $block ) {
...
+
+		$post_id = $post_id ? $post_id : get_the_ID();
+		if ( ! $post_id ) {
+			return $block;
+		}
+
+		$attachment_ids = block_core_gallery_resolve_dynamic_source(
+			$block['attrs']['dynamicContent'],
+			new \WP_Block( $block, [ 'postId' => $post_id ] )
+		);
```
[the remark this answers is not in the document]
rbcorrales: Agreed, and added in ffa79db0a. `tests/unit-tests/content-distribution/test-dynamic-gallery.php` covers resolution, wrapper settings, the gallery caption, links and lightbox, attribute injection, failing closed when nothing resolves, a gallery nested in a Group, idempotency, and the post ID reaching a registered callback.

Worth flagging for anyone reading the CI output: the PHPUnit environment runs WordPress 7.0.2, which predates dynamic galleries, so nine of those skip with an explicit message until the test WordPress is bumped. The behaviour is verified against 7.1-RC3 separately.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +136,97 @@ public static function process_jetpack_galleries( $block ) {
...
+						'id'              => (int) $attachment_id,
+						'sizeSlug'        => $size_slug,
+						'linkDestination' => $link_to,
+					]
+				),
+				esc_attr( $size_slug ),
+				esc_url( $url ),
+				esc_attr( (string) get_post_meta( $attachment_id, '_wp_attachment_image_alt', true ) ),
+				(int) $attachment_id
+			);
```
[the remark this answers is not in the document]
rbcorrales: Good catch, and it was reachable. A `className` containing `"} -->` broke the delimiter and the injected markup reached the node, which stores distributed content with `content_save_pre` filters removed.

Fixed in ffa79db0a, by a different route than suggested: the processor now builds block arrays and lets `serialize_blocks()` do the encoding, so no block-comment delimiter is hand-written anywhere. The same payload now round-trips intact, with no raw markup in the distributed content and no attribute loss.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +138,241 @@ public static function process_jetpack_galleries( $block ) {
...
+			block_core_gallery_dynamic_image_link_attributes( $attachment_id, $attributes )
+		);
+
+		// Build the `<img>` the way the image block *saves* it, not the way core
+		// renders it. `wp_get_attachment_image()` adds width, height, srcset, sizes,
+		// loading and decoding, which the block's `save()` never emits, so persisting
+		// them makes the block fail validation the moment an editor opens the post.
+		// Those attributes are added at render time on the node, as they are for any
+		// other distributed image.
+		$image_markup = sprintf(
```
wil-gerken, line 334: **Blocker:** Flattened images drop aspectRatio and scale, so a cropped gallery loses its ratio on the node — A photo editor crops a gallery to 16:9 on the hub, checks it, and publishes. On every node the same gallery comes back at a row height derived from the photos rather than the ratio the editor chose, so a 16:9 band on the hub arrives at some other proportion downstream. Nothing reports it, and the block still describes itself as cropped.

Say the hub publishes a council-meeting story with a 16:9 strip of three photos and distributes it. On the hub the strip stays uniform. On every node the same three photos fill a row whose height follows the images rather than the 16:9 the editor chose, and the desk there has no way to tell a ratio was ever set.

Verified on WordPress 7.1-RC4, in the browser rather than only in markup:

| | inline style | computed `aspect-ratio` |
| --- | --- | --- |
| Hub | `aspect-ratio:16/9;object-fit:cover` | `16 / 9` |
| Node | *(none)* | `auto 800 / 600` |

`build_gallery_image_block()` emits only `src`, `alt` and `class` (lines 316-321), where core's `block_core_gallery_render_dynamic_image()` sets both the `safecss_filter_attr( 'aspect-ratio:%s;object-fit:cover;' )` inline style and the `aspectRatio` + `scale` block attributes (`wp-includes/blocks/gallery.php:227-236`, `:253-255`). `ffa79db0a` had both; `3c0e53045` removed them alongside the `wp_get_attachment_image()` swap.

The gallery-level `is-cropped` class does not compensate. It is ratio-agnostic — `blocks/gallery/style.css` gives the images `flex: 1 0 0%; height: 100%; object-fit: cover`, which equalises heights within a row but reproduces no particular ratio. With `imageCrop` at its default the node still crops, just to a content-dependent row height rather than the ratio the editor chose.

**One correction worth flagging before you fix it.** `aspectRatio` and `scale` are source-less in `blocks/image/block.json`, which makes it tempting to restore just the attributes. That would reintroduce the validation break `3c0e53045` fixed: the image block's `save()` derives `style={{ ...aspectRatio, objectFit: scale }}` from those attrs, so attrs without the matching persisted style put `save()` and the stored markup back out of step. Restoring both together, mirroring `gallery.php:230-236` for the style and `:253-255` for the attributes, is what stays valid.

No test sets `aspectRatio` — not even the `dynamic_gallery()` helper — which is why the suite stayed green across this change. Worth adding one in the same commit as the fix: a gallery at a non-`auto` ratio, asserting both the `aspectRatio`/`scale` attributes and the inline style on the flattened image, so the pair cannot silently come apart again.

**To reproduce**
1. On WordPress 7.1, create a post, attach three images to it, and add a Gallery block in dynamic mode.
2. In the gallery inspector set **Aspect ratio** to 16:9. The images render as a uniform strip.
3. Run the post through the outgoing processor: `wp eval 'echo serialize_blocks( array_map( fn( $b ) => \Newspack_Network\Content_Distribution\Blocks::process_outgoing_block( $b, POST_ID ), parse_blocks( get_post( POST_ID )->post_content ) ) );'`
4. Store that output on a second post with no attachments and open it. The `` tags carry no `style`, and the row renders at a height derived from the images rather than at 16:9.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -60,11 +63,25 @@ public static function reset_block_processors( $block_name ) {
...
+	 * handled when it sits at the top level of the post.
+	 *
+	 * @param array $block   The block to process.
+	 * @param int   $post_id The ID of the post being distributed.
 	 *
 	 * @return array The processed block.
 	 */
-	public static function process_outgoing_block( $block ) {
+	public static function process_outgoing_block( $block, $post_id = 0 ) {
+		if ( ! empty( $block['innerBlocks'] ) ) {
```
wil-gerken, line 79: **Suggestion:** The only change that ships to WordPress 7.0.x sites has no test — Until a site updates to 7.1, every line of the gallery work is inert behind the `function_exists` guard, so the change this PR delivers to a live site first is the `innerBlocks` recursion — and that is the one part with no test.

Concretely: a hub running 7.0.x that distributes a story with a Jetpack slideshow inside a Group block gets different output the day this merges than it did the day before, because `process_jetpack_galleries()` now reaches nested blocks and strips `attrs['ids']` from them. That is very likely correct — origin attachment ids are meaningless on a node, which is exactly why the top-level path already strips them — but it is a behaviour change arriving on `release` with nothing exercising it.

Recursion is covered only through `core/gallery`, and those tests skip below 7.1. The one Jetpack fixture, `tests/unit-tests/content-distribution/post-content/jetpack-slideshow.html`, is top-level. A nested-Jetpack-gallery test would run on every WordPress version and cover the change that ships now.

Worth a docblock line too: the recursion depends on processors returning one block per block. A processor that returned a different count would desynchronise the parent's `innerContent` null placeholders.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +138,241 @@ public static function process_jetpack_galleries( $block ) {
...
+	 * @return array The processed block.
+	 */
+	public static function process_dynamic_gallery( $block, $post_id = 0 ) {
+		if ( empty( $block['attrs']['dynamicContent'] ) ) {
+			return $block;
+		}
+
+		// Dynamic galleries arrived with the WordPress 7.1 gallery block. Leave the
+		// block untouched on older versions, where nothing can resolve it anyway.
+		if (
```
wil-gerken, line 191: **Suggestion:** The version guard fails open in the one case that reaches it — If a site ever holds 7.1-authored content without 7.1 core, this branch puts the gallery's dynamic instruction back on the wire and the original bug returns — silently, with no error and no version bump marking the moment.

Picture a hub rolled back after a bad deploy, or a story imported into a site still on 7.0.x. The gallery still carries `dynamicContent`. Distribution would pass it straight through, the node would resolve it against its own attachments, and a reader would see one unrelated photo — the exact symptom this function was written to eliminate.

That is not a hypothetical corner of this branch; it is the only case that reaches it. By the time control gets here the block has already cleared `if ( empty( $block['attrs']['dynamicContent'] ) )` above, so an editor on 7.1 or later produced this block. A site with the attribute but without the core functions is a downgrade, an import or a migration — and in every one of those, passing the block through is the failure.

The docblock states the invariant at lines 167-171: passing the dynamic instruction through would let the node resolve it against its own attachments, "which is the bug this exists to prevent." The two runtime branches below honour it and fail closed. This one — whose only reachable case is the dangerous one — returns the block untouched.

Unsetting `dynamicContent` before returning here would make all three failure paths agree, and costs a line.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
```diff
@@ -400,8 +400,11 @@ protected function get_raw_post_content() {
 			return $this->post->post_content;
 		}
 
-		$blocks = array_map(
-			[ Blocks::class, 'process_outgoing_block' ],
+		$post_id = $this->post->ID;
```
wil-gerken, line 403: **Suggestion:** Flattened gallery images never reach media_data, so the node's lightbox degrades — When a reader on the node clicks a distributed gallery image, the enlarged view opens a smaller file than the same click on the hub does, and at unknown dimensions. Images from a static gallery on the same page behave correctly, so the difference shows up side by side.

`get_raw_post_content()` now resolves via the explicit `$post_id` (this line), while `get_post_media_data()` at line 531 still builds `media_data` by scanning the original, unflattened content re-rendered through `the_content`. The two disagree. Measured on a post with three attached images (24, 25, 26) and an unattached featured image (27):

```
media_data keys:            27
raw_content image ids:      24, 25, 26
```

A static gallery on the same site returns its own image ids in `media_data`.

The consequence is narrower than it first looks, and worth stating precisely. `filter_content_image_attributes()` keys on `data-id` before it reads `media_data` (`blocks/class-image-block.php:221-227`) and the saved markup carries only `class="wp-image-24"`, so it skips these images either way — static gallery images are skipped for the same reason already. Core does not mis-resolve the id against the node's own library either: `wp_calculate_image_srcset()` only builds a srcset once a filename from the attachment metadata appears in the image `src`, and the `src` here is a hub URL, so core adds nothing. Confirmed in the browser — the node's `` tags carry no `srcset`.

What does depend on it is `render_lightbox()`, which looks the id up in `media_data` at `class-image-block.php:107`. The flattened blocks do carry `id`, so the lookup happens, misses, and falls back to the `src` in the markup with `targetWidth`/`targetHeight` of `'none'`. The lightbox opens the `large` size at unknown dimensions where a static gallery image opens the hub's full-size file.

Deriving `media_data` from `get_raw_post_content()`'s output rather than re-deriving it from `get_the_content()` would close the gap for both.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +138,241 @@ public static function process_jetpack_galleries( $block ) {
...
+		) {
+			return $block;
+		}
+
+		if ( ! $post_id ) {
+			Debugger::log( 'Dynamic gallery: no post ID given, flattening to an empty gallery.' );
+			return self::flatten_gallery( $block, [] );
+		}
+
+		$attachment_ids = block_core_gallery_resolve_dynamic_source(
```
wil-gerken, line 204: **Nit:** Resolution runs twice per payload — `get_payload()` calls `get_raw_post_content()` directly and again through `get_processed_post_content()`, so the attachment resolution and the image loop each run twice per payload; `get_payload_hash()` costs another two. Measured with a `posts_pre_query` counter: 2 resolution queries per `get_payload()`. `_prime_post_caches()` makes the second image pass mostly cache hits, so this is structural doubling this PR makes more expensive rather than a new N+1, and it runs on the `shutdown` of whatever request saved the post. Memoizing `get_raw_post_content()` would fix it for both. Noting it rather than asking for it in a hotfix — the resolved set can legitimately change mid-request, so memoization has its own wrinkles.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +138,241 @@ public static function process_jetpack_galleries( $block ) {
...
+		foreach ( $attachment_ids as $attachment_id ) {
+			$image_block = self::build_gallery_image_block( $attachment_id, $block['attrs'] );
+			if ( $image_block ) {
+				$image_blocks[] = $image_block;
+			}
+		}
+
+		remove_filter( 'jetpack_photon_override_image_downsize', '__return_true' );
+
+		if ( empty( $image_blocks ) ) {
```
wil-gerken, line 235: **Nit:** Empty-gallery fallback drops the saved caption, and its log is off in production — No images are lost here — the hub renders nothing for an unresolved gallery too (`wp-includes/blocks/gallery.php:326-328` returns `''`), so the node's empty `<figure>` is a cosmetic difference rather than content loss. Two smaller things did come out of checking it.

The saved gallery caption diverges. In the "resolved N, none produced markup" branch core falls through to `gallery.php:376-378` and keeps the caption on an image-less gallery, while `flatten_gallery()` drops it at line 264 (`$caption = $image_blocks ? trim( ... ) : ''`). Reproduced: hub caption present, node caption absent.

And `Debugger::log()` is a no-op unless `NEWSPACK_NETWORK_DEBUG` is defined (`includes/class-debugger.php:32-34`), so these three messages are invisible in production. This is the established outgoing-side convention and the PR adds signal where `class-outgoing-post.php` previously had none, so it is not a regression — but `Incoming_Post::log()` (`class-incoming-post.php:181-200`) prefers `\Newspack\Logger::newspack_log()` with a context array when the Newspack plugin is present, and that channel would survive with the constant off.

The `$post_id === 0` branch is unreachable in production: the only non-test caller passes `$this->post->ID`, and the constructor throws unless that is set.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
```diff
@@ -121,4 +138,241 @@ public static function process_jetpack_galleries( $block ) {
...
+	 * @param string $classes Additional classes for the wrapper.
+	 *
+	 * @return string The wrapper attributes.
+	 */
+	private static function get_block_wrapper_attributes( $block, $classes ) {
+		if ( ! class_exists( '\WP_Block_Supports' ) || ! function_exists( 'get_block_wrapper_attributes' ) ) {
+			return sprintf( 'class="%s"', esc_attr( $classes ) );
+		}
+
+		$previous                            = \WP_Block_Supports::$block_to_render;
```
wil-gerken, line 392: **Nit:** The WP_Block_Supports swap has no try/finally — `WP_Block_Supports::$block_to_render` is public static, so the swap is legal and the restore is correct on the normal path. If `get_block_wrapper_attributes()` throws — a third-party block support is enough, since it calls `apply_block_supports()`, which `call_user_func`s the `apply` callback of every registered support — the restore is skipped.

The leaked static costs little on its own, since distribution runs on `shutdown` (`class-content-distribution.php:46`) and there is nothing left in the request to render. The throw is the part that matters: `distribute_queued_posts()` loops every queued post with no `try`/`catch`, so one failure would abort the handler and the remaining queued posts in that request would never distribute. Because distribution is triggered by save, they would stay that way until someone saved them again, with nothing surfacing an error.

That loop predates this branch — what is new is the throw surface inside it, a `WP_Block` construction and two core calls with third-party filters attached. A `try`/`finally` here costs two lines and keeps this swap from being what widens it.

plugins/newspack-network/tests/unit-tests/content-distribution/test-dynamic-gallery.php
```diff
@@ -0,0 +1,266 @@
...
+		$caption = '<figcaption class="blocks-gallery-caption wp-element-caption">Scenes from the parade</figcaption>';
+		$output  = $this->flatten( $this->dynamic_gallery( [], $caption ) );
+
+		$this->assertStringContainsString( 'Scenes from the parade', $output );
+	}
+
+	/**
+	 * Linked galleries keep their links, and the lightbox keeps working.
+	 */
+	public function test_preserves_links_and_lightbox() {
```
wil-gerken, line 159: **Nit:** Two link assertions can't fail, and the `_blank` branch has no test — Two assertions in `test_preserves_links_and_lightbox()` would not catch the regressions they look like they guard, and a third link mode is not tested at all. Core's `block_core_gallery_dynamic_image_link_attributes()` sets `linkDestination => 'none'` and `lightbox => [ 'enabled' => true ]` for `linkTo => 'lightbox'`, which is what these can be pinned against.

- Line 167, `assertStringContainsString( 'lightbox', $lightbox )`, passes on the bare word, so it would still pass on `"lightbox":{"enabled":false}`. Asserting `'"lightbox":{"enabled":true}'` closes that.
- Line 163 counts three `<a href=` but never checks where they point. Core distinguishes `wp_get_attachment_url()` for `media` from `get_attachment_link()` for `attachment`, so a swapped mapping still yields three anchors and still passes.
- The `linkTarget => '_blank'` branch has no test. That one matters more than the other two, because `class-blocks.php:327-328` builds the `target` and `rel` attributes by hand rather than delegating to core.

wil-gerken (changes requested):
Release risk: HIGH — one confirmed regression in the most recent commit, on a branch targeting `release`.

I built an isolated environment on WordPress 7.1-RC4 and ran this end to end: the flattening path, the full newspack-network suite, and a hub-shaped post distributed into a node-shaped copy holding no attachments of its own. The core of the fix holds up perfectly. Resolving on the hub is the only side this can work from, since a node has no attachments to resolve against, and it matches how `process_jetpack_galleries()` already strips origin-specific ids on the way out. Images, captions, alt text and links all arrive intact. The `serialize_blocks()` rewrite closes the injection hole properly, and the hub's stored `post_content` is never touched, which is the most important property of the design. All 11 of the PR's own tests pass on 7.1-RC4; PHPCS is clean; the one failure elsewhere in the newspack-network suite is a remote-image sideload that cannot reach the network from the container.

One thing to fix before this ships, on the aspect-ratio thread — with a correction worth reading first, since restoring the attributes on their own would reintroduce the validation break `3c0e53045` was written to solve.

One pattern worth naming across the smaller threads: until a site updates to 7.1, the entire gallery path is inert behind its `function_exists` guard. The change this PR delivers to a live site first is the `innerBlocks` recursion, and that is the one part with no test.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Fixed in 24815f85b, and thank you for the correction. I had dropped the pair because restoring the attributes alone risked reintroducing the validation break, which was the wrong conclusion: it is restoring them *without* the style that breaks. Both now travel together, mirroring `gallery.php:230-236` for the style and `:253-255` for the attributes.

Verified through core's editor bundles in jsdom before writing the test: `aspectRatio` at 16/9, 4/3 with columns, 1 with `linkTo:media`, and `auto` all parse `isValid: true` with every nested image valid. `safecss_filter_attr()` drops the trailing semicolon, which is what keeps the persisted style byte-identical to what `save()` regenerates.

Added `test_preserves_aspect_ratio()`, which asserts the attributes and the inline style on all three images, plus `test_auto_aspect_ratio_adds_nothing()` so the `auto` path cannot start emitting a style.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Fixed in 24815f85b. You are right that the only reachable case is the dangerous one, and it now strips `dynamicContent` and logs, so all three failure paths agree.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Good catch, and the sharpest point in the review: the part that reaches live sites first was the part with no coverage.

Added in 24815f85b, both running on every WordPress version rather than behind the 7.1 gate:

- `test_recursion_reaches_a_nested_jetpack_gallery()`, a `jetpack/tiled-gallery` inside a Group, asserting `ids` is stripped and the block survives.
- `test_recursion_preserves_block_structure()`, a Group of two paragraphs that must round-trip byte-identical, which is the one-block-per-block contract.

Also documented that contract on `process_outgoing_block()`.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Fixed in 24815f85b with a `try`/`finally`. The point about `distribute_queued_posts()` having no catch of its own is the one that persuaded me: silently dropping the rest of the queue until someone saves those posts again is a bad failure mode to widen.

plugins/newspack-network/tests/unit-tests/content-distribution/test-dynamic-gallery.php
rbcorrales: Fixed in 24815f85b. The lightbox assertion now pins `"lightbox":{"enabled":true}` rather than the bare word, the link assertions check the actual targets (`wp_get_attachment_url()` for media, `get_attachment_link()` for attachment) so a swapped mapping fails, and there is now a `linkTarget: '_blank'` case covering the hand-built `target`/`rel`.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
rbcorrales: Confirmed, and your tracing is more precise than mine was. I had this as a srcset problem; you are right that `filter_content_image_attributes()` skips these images either way on `data-id`, exactly as it skips static gallery images, and that core adds no srcset because the `src` is a hub URL. The real consequence is `render_lightbox()` finding the `id`, missing in `media_data`, and falling back with `targetWidth`/`targetHeight` of `'none'`.

Not fixing it here. It is pre-existing in shape (`get_post_media_data()` has always re-derived from `get_the_content()`) and the fix you describe, feeding it from `get_raw_post_content()`, changes behaviour for every block processor rather than just this one. That is more than I want inside a hotfix aimed at `release`. Tracking it as a follow-up.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Agreed on all of it, including that memoization has its own wrinkles when the resolved set can change mid-request. Leaving it out of this hotfix and tracking it with the `media_data` item, since both are about `Outgoing_Post` rebuilding the payload rather than about the gallery.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Both noted. On the caption: core keeps it on an image-less gallery and `flatten_gallery()` drops it, so we do diverge. Leaving it, because a caption with no images to caption is the case core itself calls meaningless when it returns `''` from the dynamic branch, and changing it is not worth the risk in a hotfix. Worth revisiting alongside the empty-`<figure>` difference.

On the logging: fair, and `Incoming_Post::log()` routing through `\Newspack\Logger` when available is the better pattern. Following the outgoing-side convention here rather than introducing a second one mid-hotfix, but it is the right cleanup for the same follow-up.

plugins/newspack-network/includes/content-distribution/class-outgoing-post.php
rbcorrales: Tracked as [NPPM-3181](https://linear.app/a8c/issue/NPPM-3181).

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Tracked as [NPPM-3181](https://linear.app/a8c/issue/NPPM-3181), grouped with the `media_data` item since both are `Outgoing_Post` rebuilding the payload.

plugins/newspack-network/includes/content-distribution/blocks/class-blocks.php
rbcorrales: Splitting these two. The logging point is tracked as [NPPM-3181](https://linear.app/a8c/issue/NPPM-3181). The caption divergence is gallery-scoped and two lines, so it belongs in this PR rather than a follow-up. Fixing it here.

wil-gerken (approved):
Verified on a 7.1-RC4 env. The images come through at 16:9 now, and the editor opens the block cleanly, so putting the attributes back didn't reintroduce the validation problem you fixed in `3c0e53045`.

Thanks for gating both halves on the same `$has_ratio` and adding the comment explaining why.

I also ran it on 7.0.4 — both recursion tests pass there.

Good call keeping `media_data` and the logging as follow-ups rather than growing the hotfix.

Thanks!

PR #919, opened 2026-08-19
feat(content-gate): share one metering allowance across gates (NPPD-2191)

thomasguillot:
### All Submissions:

Metering counters are keyed by gate, so a site that needs several gates meters each section separately: a reader crossing sections gets the free-view allowance **once per gate**, and the countdown banner resets with it. Sites migrating from Woo Memberships, which kept a single meter, see their allowance double.

Gates now share one site-wide allowance, with a per-audience opt-out. Enablement stays on the gate, so a hard wall and a metered gate coexist against the same pool. The allowance follows the **reader** rather than the gate: signed-out views draw on one count and signed-in views on another, so crossing sections cannot hand anyone a second allowance.

A one-time routine adopts the existing configuration where published gates agree. Where they disagree it keeps each gate's own allowance, and it pins only the audience paths that would otherwise change, so no reader's allowance moves on upgrade. Premium newsletter gates are kept out of that vote and keep their own allowance, since no publisher-facing screen can show or reconcile them. The routine seeds only settings that have not been written yet, so a value a publisher has already saved always stands, and it defers while Woo Memberships is active, so a migrating site adopts after cutover, from its Access Control gates.

Existing per-gate reader counters are not merged into the shared one. Merging unions the posts already read, which can exceed the shared count and lock a reader out mid-period. The cost runs the other way: on the day adoption completes, every reader starts the shared counter with a full allowance, whatever they had spent per gate. Pages cached before the switch still name the per-gate counter, so deploys to metered sites should purge the page cache.

Countdown banner settings move into Metering, since the banner counts down the allowance.

Metering settings: one shared allowance, and the countdown banner that counts it down

`migrate-membership-gates` is unchanged, so a Woo Memberships site does not yet seed the site meter from its legacy meta; with adoption deferred until cutover, a later change can carry the legacy counts onto the new gates and adoption will read them.

Closes NPPD-2191.

1. Enable Access Control by defining `NEWSPACK_CONTENT_GATES`. This is required; the metering settings do not appear without it.
2. Create two gates, each metering on Paid Access.
3. Open Audience > Access Control. Both gates report the same allowance, marked "site meter".
4. Open Metering, set signed-out free views to 1, and save.
5. Sign out and read a gated post behind the first gate, then one behind the second. The second post is gated. Before this change it was free.
6. In the first gate, set Meter to "This gate only" and give it 3 free views. Save.
7. Read gated posts behind each gate again from a signed-out browser. The first gate allows 3, the second still allows 1.
8. Open Metering. A notice names the first gate as ignoring these settings.
9. Change a free-view count without saving. A second notice names the gates whose layout wording still quotes the previous number.
10. Set Banner to Enabled and save, then reload a gated post. The countdown reflects the shared allowance.
11. Set Banner to Disabled. The banner settings grey out rather than disappearing.
12. On an existing site with one metered gate, load any admin screen. The gate's allowance is unchanged.

### Other information:

`SectionHeader` gains a `size="hidden"` variant, its back nav is now hidden below 1280px where the breadcrumb already provides navigation, and badges and the kebab lose their 8px top offset. All three are shared-component changes that affect every wizard.

Content Gifting's settings page follows the same layout as Metering: the hidden section header, the same section spacing, and the recipient banner preview under its fields.

<details>

- `Newspack\Site_Meter` (new) owns the two counts and the reset period as `newspack_site_meter_*` options. Adoption runs once per site, on `admin_init` (requests from users who can edit posts) and on the wizard's config read, and records completion in its own option outside that prefix. Its seed is add-only per key: a publisher's save always outlives a concurrently finishing run.
- `Metering::get_meter_key()` returns the shared key or the gate ID, and drives the user-meta key plus the localised `meter_key` that `src/content-gate/metering.js`, `content-banner.js` and the countdown `view.js` read.
- `Metering::resolve_settings()` resolves count and period against a `scope` stored on each audience path's metering array. Enablement stays on the gate.
- Scope is stored **per audience path** (`registration.metering.scope`, `custom_access.metering.scope`) rather than per gate, which is where the data and the two existing controls already live. The ticket describes a per-gate toggle, so this is finer-grained than written.
- Questions about the gate, such as whether any gate meters at all, resolve without reading the current visitor's session, so site-wide surfaces answer the same for everyone. The count key follows the chosen audience path, so a reader held at a verification wall is described by the wall's own allowance.
- Legacy Woo Memberships gates are exempt throughout and keep their per-gate counters.
- REST: `site_meter` in the gates config payload, plus `POST /wizard/newspack-audience-access-control/site-meter`.

</details>


adekbadek (approved):
Release risk: LOW — the review's blocker and each runtime-reproduced finding are fixed on the branch: adoption's seed is add-only per key so a saved allowance outlives a late-finishing run, adoption defers while Woo Memberships is active, the admin_init path requires `edit_posts`, a save equal to the defaults now writes a real row, and a reader held at a verification wall resolves against the wall's own allowance. Each fix landed with a regression test that ran red first, and the adoption scenarios were exercised on an isolated env (agree, conflict, late save, Memberships active, anonymous admin-ajax).

For transparency: the merge of `main` and the fix commits (980d2b9e, db56d53b) are reviewer-authored, pushed to the branch instead of requesting changes — worth a read-through. One thread stays open by design: the Metering screen is editable while Woo Memberships is still active, and a value saved there pre-cutover becomes the shared meter at adoption. Published gates that differ get pinned, so no reader's allowance changes; a notice like the one feeds use would close it.

Review provenance: 4 of 5 reviewer passes (lineup: deep, standards, wp-expert, codex, opencode:kimi-k3) — codex skipped (CLI auth failure).

PR #914, opened 2026-08-19
feat(cli): add fix-memberships command (NPPM-386)

adekbadek:
### Changes proposed in this Pull Request:

Ports the `newspack fix-memberships` WP-CLI command into the monorepo. It was reviewed, tested, and merged upstream in legacy `newspack-plugin` [#4733](https://github.com/Automattic/newspack-plugin/pull/4733), but the daily legacy→monorepo sync has been failing since ~July 22 and #4733 (merged July 23) never landed here.

The command reconciles `wc_user_membership` posts against WooCommerce Subscriptions — repairing memberships left in a stale/paused state by a renewal race condition (NPPM-386).

Faithful port: the class file is byte-identical to the merged legacy version, and registration matches #4733. The only monorepo-specific change is extending two `posts_per_page => -1` phpcs-ignores to also cover VIPWPCS 3.1.0's `NoPaging` sniff, using the justification convention this repo established in #741 (the legacy repo runs an older VIPWPCS that does not flag it).

Transitional by design: registration and execution are gated on `function_exists( 'wc_memberships' )`, so the command self-hides once a site is migrated off WooCommerce Memberships to Access Control.

Ref: NPPM-386. Legacy merge: `ef8dd95`.

### How to test the changes in this Pull Request:

1. On a site with WooCommerce Memberships active, run `wp newspack fix-memberships --help`. Confirm the command is registered.
2. Run `wp newspack fix-memberships` (dry run). Confirm it reports discrepancies without mutating data.
3. Run `wp newspack fix-memberships --live --limit=1`. Confirm exactly one membership is repaired.
4. Deactivate WooCommerce Memberships. Confirm `wp newspack fix-memberships` is no longer registered.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable? (Port of already-reviewed upstream code; no new tests.)
* [x] Have you successfully run tests with your changes locally? (`php -l` + root PHPCS clean; command logic verified end-to-end in #4733 on identical code.)

PR #920, opened 2026-08-19
ci: unbreak the type gate and stop PHPUnit hanging on a stalled download

dkoo:
### All Submissions:

Two unrelated CI failures on `main`, both of which currently block every open PR.

**The type gate is red on `main`.** `rule-form.tsx` holds the pricing rule status as `useState< 'publish' | 'draft' >`, and the toggle's `onChange` widened it back to `string` via `String( v )`. That landed in #825, shortly after #914 turned the type check into a gate, so `JS / newspack` has failed on every run since:

```
src/wizards/audience/views/pricing-rules/rule-form.tsx(632,34): error TS2345:
Argument of type 'string' is not assignable to parameter of type 'SetStateAction<"publish" | "draft">'.
```

The setter now narrows to the same union the state declares, matching what the initializer on line 233 already does. No behaviour change: the toggle only ever emits `publish` or `draft`.

**PHPUnit can hang for six hours and report a red job that ran no tests.** Every download in "Install WordPress test suite" talked to wordpress.org with no time limit, and `svn` applies none of its own. When wordpress.org stalls, the step sits on a silent connection until the six-hour job limit kills it; the PHPUnit step is then skipped, so the run reports a PHPUnit failure that never executed a test.

Two runs were caught in that state today within minutes of each other, on unrelated PRs, while a run started 20 minutes later passed in 4½ minutes — an intermittent upstream stall rather than anything in the branches. The step now has a 10 minute cap, `curl` and `svn` retry with per-attempt timeouts, and an empty version lookup fails with a message that says so instead of building URLs that hang.

1. Check out this branch and run `pnpm --filter newspack run typescript:check` from the repo root. It passes. On `main` the same command fails with the TS2345 above.
2. Open the pricing rule editor in the Audience wizard and switch the Status toggle between Published and Draft. The value saves as before.
3. Read the `Install WordPress test suite` step in `.github/workflows/ci.yml`. Confirm `timeout-minutes: 10`, the `--max-time`/`--retry` flags on both `curl` calls, the empty-version guard, and the `svn_checkout` retry helper.
4. Watch this PR's own `PHPUnit / newspack` job complete in the usual few minutes.

### Other information:

* [ ] Have you written new tests for your changes, as applicable? — n/a: a one-line type narrowing and a CI workflow step, neither of which the suites cover.

Verification I ran, rather than assuming:

- `tsc --noEmit` exits 0 with the fix. As a control, reverting just that line reproduces CI's exact error, same file and line, so the local check genuinely exercises the gate.
- The workflow parses as YAML, the step's script passes `bash -n`, and `shellcheck` is clean.
- The empty-version guard and the retry helper were exercised in isolation: `""` and `null` both refuse with the intended message while a real version proceeds, and the retry loop stops after exactly three attempts and returns non-zero.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -345,14 +345,30 @@ jobs:
           fi
       - name: Install WordPress test suite
         if: steps.check.outputs.has_tests == 'true'
+        # Every download here talks to wordpress.org, and none of them had a time
+        # limit. A stalled connection therefore held the step open until the six-hour
+        # job limit killed it, and the run reported a red PHPUnit that had never
+        # executed a test. The cap turns that into a failure within minutes, and the
+        # retries below let an ordinary blip recover on its own.
+        timeout-minutes: 10
         run: |
+          set -eo pipefail
           # svn isn't on ubuntu-latest runners by default; the WP develop
           # repo's phpunit test library is published over svn.
           sudo apt-get update -qq
           sudo apt-get install -y -qq subversion
           # Download WP core and test library.
           mkdir -p "$WP_TESTS_DIR"
-          WP_TESTS_TAG=$(curl -s "https://api.wordpress.org/core/version-check/1.7/" | jq -r '.offers[0].version')
+          WP_TESTS_TAG=$(curl --fail --silent --show-error --location \
+            --retry 3 --retry-delay 5 --max-time 30 \
+            "https://api.wordpress.org/core/version-check/1.7/" | jq -r '.offers[0].version')
+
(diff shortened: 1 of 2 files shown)


dkoo:
Preempted by #826. That PR fixed the main blocking issue addressed in this PR, which was the failing JS test. The other fix proposed in this PR (to set a 10-minute timeout for the phpunit CI job) is a definite improvement, but unless we start seeing timeouts happening frequently across multiple commits and PRs, I think we can table the change for now.

PR #923, opened 2026-08-19
ci(release): skip the plan gate when a newer push supersedes it

jason10lee:
### All Submissions:

* [x] Does your code follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)? — n/a in the usual sense: a GitHub Actions workflow with no PHP or JS, so PHPCS does not cover it. The step body is clean under `shellcheck -S warning`.
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change? — no duplicate. Two open PRs touch this file: #895 and #811. See "Other information" for how they interact.

The release plan gate now reports "nothing to release" when the push it was started for has already been replaced, instead of failing.

When two pushes land on a branch minutes apart, the older `Release` run still holds the earlier SHA. multi-semantic-release refuses to publish from a checkout behind the branch, so the dry-run in `Plan release` exits non-zero with `The local branch <branch> is behind the remote one, therefore a new version won't be published.` and the run goes red. Two examples from 2026-08-19: runs 32210068269 and 32210067677. Both branches were refreshed against `release` at 02:51Z, then refreshed again at 02:53Z after `release` moved twice in eight seconds, and the 02:53 runs both passed. `concurrency` sets `cancel-in-progress: false` on purpose, since a real release must never be interrupted, so a superseded run is never cancelled and always runs to failure.

**A moved tip is not on its own a reason to skip.** This workflow only runs on branches that publish, so a skip that nothing else covers would drop a release with no trace. The gate therefore requires two conditions before skipping, and both are checked against the remote:

* The new tip is a descendant of this run's SHA. A force-push that rewound the branch leaves this run as the only one that saw its commits, so there is no newer run to inherit the work.
* The new tip is not a `[skip ci]` commit. That is how semantic-release pushes its own version bump, and such a push starts no run, so whatever this run was handed would never be released.

Anything else keeps today's behaviour and fails loudly, including a remote tip that cannot be read: a false "superseded" would silently skip a real release, which is the outcome this step's original comment refuses to risk. On a skip it writes `release_needed=false`, so the existing gating on `steps.plan.outputs.release_needed` carries the skip through the rest of the job with no other change.

Closes NPPM-3176.

1. Extract the supersession block from the top of the `Plan release` step into a file, and create a scratch repo with a bare remote: a base commit `A`, a descendant `B` with a normal message, a descendant `C` whose message contains `[skip ci]`, and a commit `D` on a branch rooted at `A` for the rewind case.
2. Point the remote branch at `A` and run with `GITHUB_SHA=A`. Expect no output and no `release_needed` written: the run is current and falls through to the dry-run.
3. Point it at `B`, still with `GITHUB_SHA=A`. Expect a notice naming both SHAs, `release_needed=false` in `$GITHUB_OUTPUT`, and exit 0. This is the case that fails red today.
4. Point it at `C`, with `GITHUB_SHA=A`. Expect a warning and no skip, because `[skip ci]` means no newer run exists.
5. Point it at `D`, with `GITHUB_SHA=B`. Expect a warning and no skip: `D` is not a descendant of `B`.
6. Set `origin` to a path that is not a repository and run again. Expect the "could not read the remote tip" warning and no skip.

Note on what CI can show here. Since #830, this workflow runs only on `release` and `alpha`, so it does not run on this PR and the checks below say nothing about this change. The fixture above is how it was exercised; all five cases pass.

### Other information:

* [x] Have you written new tests for your changes, as applicable? — not applicable in the repo's harness sense; verified by the extracted-block fixture described above, which covers all five paths through the check.
* [x] Have you successfully run tests with your changes locally? — yes: the five fixture cases, plus `shellcheck -S warning` and a YAML parse of the file.

**Scope after #830.** That PR removed `hotfix/**` and `epic/**` from this workflow's triggers on `main`, which retires the branch class where these two failures happened. `release` still carries the old triggers, so the failure is live there until the next promotion, and the case that survives on `main` is two merges to `release` or `alpha` in quick succession, where the queued older run starts after the newer push has landed. That is rarer than what prompted the ticket, and it is why NPPM-3176 is filed Low rather than treated as urgent.

**Interaction with #895 and #811.** #895 touches the same step, but its diff re-adds what `main` already has: its base predates the back-merge that brought #833 into `main`, so most of its release.yml hunk is content rather than change. My insertion sits above the `set +e` line it treats as context, so the two should merge cleanly, and identically-either-order. #811 is based on `release` while this is based on `main`, so they meet at a promotion rather than in a conflict.

**Base branch.** `main`, following #830's precedent for this file, and because the branches the check still protects (`release`, `alpha`) receive it at the same promotion either way.


.github/workflows/release.yml
```diff
@@ -80,6 +80,44 @@ jobs:
...
+          # signal a missed release needs.
+          fetched=true
+          git fetch --quiet --no-tags origin "+refs/heads/$GITHUB_REF_NAME" || fetched=false
+          remote_tip=''
+          if [ "$fetched" = true ]; then
+            remote_tip=$(git rev-parse FETCH_HEAD 2>/dev/null) || remote_tip=''
+          fi
+          if [ -z "$remote_tip" ]; then
+            echo "::warning::Could not read the remote tip of $GITHUB_REF_NAME; treating this run as current."
+          elif [ "$remote_tip" != "$GITHUB_SHA" ]; then
```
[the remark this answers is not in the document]
jason10lee: Agreed, and the distinction is the point rather than the phrasing: "could not check" and "checked, and the run is current" need different responses from whoever reads the log, and the old wording collapsed them into the reassuring one.

Reworded in f60449f:

```
Could not read the remote tip of $GITHUB_REF_NAME, so the supersession check
did not run. Continuing: if this push has been superseded, the dry-run below
fails rather than skipping.
```

Behaviour is unchanged, since continuing is still the right call when the check cannot run. Only the log now says which of the two happened.

.github/workflows/release.yml
```diff
@@ -80,6 +80,44 @@ jobs:
...
+          git fetch --quiet --no-tags origin "+refs/heads/$GITHUB_REF_NAME" || fetched=false
+          remote_tip=''
+          if [ "$fetched" = true ]; then
+            remote_tip=$(git rev-parse FETCH_HEAD 2>/dev/null) || remote_tip=''
+          fi
+          if [ -z "$remote_tip" ]; then
+            echo "::warning::Could not read the remote tip of $GITHUB_REF_NAME, so the supersession check did not run. Continuing: if this push has been superseded, the dry-run below fails rather than skipping."
+          elif [ "$remote_tip" != "$GITHUB_SHA" ]; then
+            if git merge-base --is-ancestor "$GITHUB_SHA" "$remote_tip" &&
+              ! git log -1 --format=%B "$remote_tip" | grep -qF '[skip ci]'; then
```
adekbadek: **Suggestion:** `! git log -1 --format=%B "$remote_tip" | grep -qF '[skip ci]'` gives the right answer only because this step inherits GitHub's default `bash -e {0}`, which does not set `pipefail`. Under `-o pipefail`, `grep -q` exits at its first match and closes the pipe, `git log` dies of SIGPIPE (141), that status becomes the pipeline's, and the leading `!` reads it as "not a `[skip ci]` commit" — so the guard skips, writes `release_needed=false`, and the release is dropped on a green run.

It only inverts once the message exceeds the 64 KiB pipe buffer, which here is the ordinary case rather than an edge: this repo's own `chore(release): … [skip ci]` commits measure 80 KB, 87 KB and 175 KB. Checked against a real one (`fd7f7638`, 87,604 bytes) — the same fixture returns "continue" under `bash -e` and "skip" under `bash -e -o pipefail`.

Nothing is wrong as merged: this step sets no `shell:` and the file sets no `defaults.run.shell`. What makes it worth changing anyway is that the file's own idiom is the trigger — the `run` blocks at :206 and :302 both open with `set -euo pipefail`, so an editor adding that line here would get no signal that they had inverted a release gate.

Capturing the message first drops the pipe and behaves identically under either shell:

```sh
tip_message=$(git log -1 --format=%B "$remote_tip")
case "$tip_message" in
  *'[skip ci]'*) tip_starts_run=false ;;
  *)             tip_starts_run=true ;;
esac
if git merge-base --is-ancestor "$GITHUB_SHA" "$remote_tip" && [ "$tip_starts_run" = true ]; then
```

adekbadek: **Suggestion:** The guard's stated invariant is that it skips only when the newer commit gets a run of its own. GitHub suppresses a push-triggered run for five spellings — `[skip ci]`, `[ci skip]`, `[no ci]`, `[skip actions]`, `[actions skip]` — and for a `skip-checks:true` / `skip-checks: true` trailer. This matches the first only.

Repo automation never writes the others: `config/release-helpers.js:52`, `packages/scripts/scripts/release.js:102` and `.github/scripts/post-release.sh:91` all use `[skip ci]`, so no machine push reaches this. The path that does is a human one, and `release` and `alpha` take human pushes — a promotion merge or a base-refresh, the same class of push the guard exists for. A merge whose message says `[ci skip]` starts no run, and the guard reads it as run-starting and skips.

Widening the match closes the one case here where the skip is unrecoverable:

```sh
*'[skip ci]'*|*'[ci skip]'*|*'[no ci]'*|*'[skip actions]'*|*'[actions skip]'*)
```

.github/workflows/release.yml
```diff
@@ -80,6 +80,44 @@ jobs:
...
+              echo "::notice::$GITHUB_SHA is no longer the tip of $GITHUB_REF_NAME (now $remote_tip); the newer push has a run of its own, so this one has nothing to do."
+              echo "release_needed=false" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+            echo "::warning::$GITHUB_SHA is behind $GITHUB_REF_NAME (now $remote_tip), but the newer commit is a rewind or a [skip ci] commit that starts no run. Continuing, so a release this run cannot publish fails loudly."
```
adekbadek: **Suggestion:** The Release workflow has no `pull_request` trigger, so nothing runs this block before it reaches `release` or `alpha` — and there its failure mode is a silent skip rather than a red run, which is the outcome the comment at :91-102 says must not happen.

Both halves of a fix already exist in the repo. `bin/tests/*.sh` runs in the "Shell specs" CI job, which is deliberately ungated on changed packages so that a tooling-only PR like this one still gets a run; `.github/scripts/` is the established home for extracted release shell (`notify-release.sh`, `post-release.sh`).

The case worth pinning is the one above: a `[skip ci]` tip whose message exceeds 64 KiB must not skip. A spec written against a short fixture message passes under both shell modes and proves nothing, so the message size is the test rather than incidental to it. The rewind case earns the same treatment.

The five-case fixture described in the PR body is most of this already.

.github/workflows/release.yml
```diff
@@ -80,6 +80,44 @@ jobs:
...
+          # silently. Only skip when the
+          # newer commit will get a run of its own, which needs two things to
+          # hold. It must be a descendant of this SHA -- a force-push that
+          # rewound the branch leaves this run as the only one that ever saw
+          # these commits. And it must not be a [skip ci] commit -- that is how
+          # semantic-release pushes its own version bump, and it starts no run,
+          # so whatever this run was handed would never be released. Either way
+          # the old behaviour stands: the dry-run below fails, which is the loud
+          # signal a missed release needs.
+          fetched=true
```
adekbadek, line 146: **Suggestion:** The check runs once, at the top of the plan step. A push landing after it still reds the run: multi-semantic-release's own behind-remote check fires in the dry-run below, and again in the `Release` step ten-plus minutes later after the build. For the trigger NPPM-3176 names — a base-refresh landing during checkout or `pnpm install` — the common case is covered, so this reduces the rate rather than eliminating it, and the ticket is worth closing on those terms rather than as fully resolved.

`alpha` also keeps a superseded-red case of its own that this cannot reach. `post-release.sh` resets `alpha` onto `release` and force-pushes, leaving `chore(release): restore workspace:* deps after release [skip ci]` at the tip (`post-release.sh:91`). An `alpha` run in flight sits in a different concurrency group from the `release` run driving that push, so it is not serialised against it: it finds its SHA is an ancestor of a `[skip ci]` tip, falls through, and goes red even though its commits were released by the other branch's run. Failing loud is the defensible choice there — worth a line in the comment so a later reader does not take the guard as having ended superseded reds.

.github/workflows/release.yml
```diff
@@ -80,6 +80,44 @@ jobs:
...
+          # apart does it -- and the older run still holds the older SHA, so
+          # multi-semantic-release below refuses from a stale checkout and exits
+          # non-zero with "The local branch <branch> is behind the remote one".
+          # cancel-in-progress stays false on purpose (an in-flight real release
+          # must never be interrupted), so the check has to live inside the job.
+          #
+          # "The tip moved" is not on its own a reason to skip: this workflow
+          # only runs on branches that publish (the list lives in
+          # config/release.js), so a skip nothing else covers loses a release
+          # silently. Only skip when the
```
adekbadek: **Nit:** Two small things in an otherwise well-pitched comment.

The wrap goes ragged mid-paragraph: :94 ends at column 40 where every other line in the block runs to about 77. It reads as a leftover from the reword in f60449f — rewrapping :91-95 settles it.

At :100-101, "Either way the old behaviour stands" points at the change rather than at the code. On the merged branch there is no old behaviour to stand, and the repo's in-repo prose rule is that a comment describes how things are. "Either way this run continues and the dry-run below fails, which is the loud signal a missed release needs" says the same thing without the backward reference.

adekbadek (approved):
Release risk: LOW — the guard is correct as it will actually run, and I verified that rather than reasoning about it: the block was extracted and exercised against a git fixture, eight of eight cases behaving as the PR describes under GitHub's default `bash -e {0}`. No blockers. Approving.

The one thing worth weighing before merge is not a defect in the diff. This step is the only automatic alert these two branches have, and the change trades it for a green square: a superseded run used to go red, and now it succeeds silently. Nothing downstream compensates — the Slack step runs but exits without posting when no tags appeared, and branch sync and the three WordPress.org deploys are gated off the same flag. That is the right call for the case this targets, and it is why the two comments about the skip conditions are worth closing even though neither is live today.

Two measurements are on the threads rather than here, because both cut against a first reading: the oversized-commit trigger is ordinary rather than exotic (8 of 600 sampled `[skip ci]` commits on these branches exceed the pipe buffer), and the alternative skip markers have never once been used on `release` or `alpha` in 4,252 such commits — which is why that one stays a suggestion.

- `.github/workflows/release.yml:313` — **Nit:** This step is gated on `github.ref_name == 'release'` alone, not on `release_needed`, so the new skip path reaches it with a checkout that is by construction behind the remote — a state that used to be an error rather than a routine outcome. It is a no-op today, traced end to end: the tree is clean, so `finalize-package-versions.cjs` stages nothing and returns at its `staged.length === 0` guard without committing; `git push origin HEAD:release` is then rejected non-fast-forward, the fallback `pull --rebase` fast-forwards, and the retry pushes nothing. Adding `steps.plan.outputs.release_needed == 'true'` to the condition makes the skip a true no-op, and removes the need to re-derive that chain the next time the step changes.

.github/workflows/release.yml
jason10lee: Reproduced before changing anything. Against `fd7f7638` (87,604 bytes) the same
fixture returns "continue" under `bash -e` and "skip" under `bash -e -o
pipefail`; a short `[skip ci]` message returns "continue" under both. So the
message size is the trigger rather than incidental, exactly as you described.

Fixed the way you suggested: capture the message, match with `case`. The block
now contains no pipelines at all, so no shell option can reach it, and the same
fixture returns "continue" under both modes at the new head.

jason10lee: Widened to all five. Your premise holds here too, though our counts are not
directly comparable: measuring message lines rather than commits across
`origin/release` and `origin/alpha`, `[skip ci]` appears 4,319 times and
`[ci skip]`, `[no ci]`, `[skip actions]`, `[actions skip]` and both
`skip-checks` forms appear zero times each. No machine push writes anything but
the first, so this guards the human one.

.github/workflows/release.yml
jason10lee: Taken, and deferred rather than done, because the specific hazard is now
structurally absent instead of merely fixed: with the pipeline gone there is no
shell option that can change the result. What remains is regression cover
against a future edit reintroducing a pipe or narrowing the marker set.

Filed as NPPM-3355, with your constraint: the fixture message has to exceed
64 KiB, since a short one passes under both modes and proves nothing. It also
carries the four cases to pin and both halves of the mechanism you identified.

.github/workflows/release.yml
jason10lee: Both are in the comment now. It says the check runs once so a later push still
reds the run, and it records the `alpha` case: `post-release.sh` resets onto
`release` and force-pushes, the in-flight `alpha` run sits in a different
concurrency group so it is not serialised against it, and it goes red although
the other branch released its commits. The comment calls that the right
outcome, not a gap.

.github/workflows/release.yml
jason10lee: Both taken. The paragraph is rewrapped, and the sentence now describes what the
code does rather than what it used to: "Either way this run continues and the
dry-run below fails, which is the loud signal a missed release needs."

jason10lee:
The out-of-diff nit from the review body is folded in: `Sync package.json
versions` now carries `steps.plan.outputs.release_needed == 'true'` alongside
the branch check.

The comment above the condition records why it was a no-op: on the skip path
the tree is clean, `finalize-package-versions.cjs` returns at its
`staged.length === 0` guard without committing, and the push that follows has
nothing to send.

Two commits: `e9241408d2` for the marker matching, `6c3ac1ba41` for this.

